PR #14258, opened 2025-03-05
[Attention] FlashAttn MLA

LucasWilkinson:
Use latest FlashAttention code to compute decode MQA in MLA

```
VLLM_USE_V1=0 lm_eval --model vllm --model_args pretrained=deepseek-ai/DeepSeek-V2-Lite-Chat,tensor_parallel_size=2,dtype=auto,gpu_memory_utilization=0.9,trust_remote_code=True,max_model_len=16384 --task gsm8k --num_fewshot 5  --batch_size auto

|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.6725|±  |0.0129|
|     |       |strict-match    |     5|exact_match|↑  |0.6657|±  |0.0130|
```

Co-authored-by: @MatthewBonanni

PR #26541, opened 2025-10-09
[Attention][Spec Decode] FlashMLA spec decode support

MatthewBonanni:
## Purpose
This PR implements speculative decoding support for the FlashMLA backend.

**NOTE**: the comment about the intermittent test failure was true prior to this PR, I just made a note of it.

cc @LucasWilkinson 

## Test Plan
`pytest tests/v1/attention/test_mla_backends.py`

## Test Result
Passes

---
<details>

</details>


tests/v1/attention/test_mla_backends.py
```diff
@@ -397,11 +445,41 @@ def test_backend_correctness(dist_init, batch_spec_name: str, model: str):
...
+                )
+                within_threshold = q_len <= threshold if threshold else False
+                if (
+                    within_threshold
+                    and query_len_support == QueryLenSupport.UNIFORM
+                    and i > 0
+                ):
+                    first_q_len = query_lens[0]
+                    within_threshold = q_len == first_q_len
+                is_decode.append(within_threshold)
```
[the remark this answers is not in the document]
MatthewBonanni: addressed

vllm/v1/attention/backends/mla/common.py
```diff
@@ -455,14 +474,13 @@ class MLACommonMetadataBuilder(AttentionMetadataBuilder[M]):
...
-    # when reordering the batch. Non-uniform decode requests will
-    # fall back to prefill in this case.
-    supports_uniform_spec_as_decode: ClassVar[bool] = False
+    # Defines the level of query length support for this backend.
+    # - SINGLE_ONLY: Only single-token queries (no spec decode support)
+    # - UNIFORM: Supports uniform multi-token queries (spec decode with uniform lengths)
+    # - VARLEN: Supports variable-length queries (spec decode with mixed lengths)
+    # If set to UNIFORM or VARLEN, this will increase `reorder_batch_threshold` when
+    # speculative decoding is enabled.
+    query_len_support: ClassVar[QueryLenSupport] = QueryLenSupport.SINGLE_ONLY
```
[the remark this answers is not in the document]
MatthewBonanni: fixed

LucasWilkinson (approved):
Nice! LGTM but lets have @benchislett look too

vllm/v1/attention/backends/mla/common.py
```diff
@@ -599,11 +617,23 @@ def __init__(
...
         )
 
+        # Validate consistency between query_len_support and reorder_batch_threshold
+        if self.query_len_support == QueryLenSupport.SINGLE_ONLY:
+            assert self.reorder_batch_threshold == 1, (
+                f"reorder_batch_threshold must be 1 when query_len_support is "
+                f"SINGLE_ONLY, got {self.reorder_batch_threshold}"
+            )
+        else:
+            assert self.reorder_batch_threshold > 1, (
```
benchislett: This should crash when spec decoding is disabled? In such cases I expect `query_len_support` can vary but `reorder_batch_threshold` will stay as 1 since there will never be any spec tokens.

vllm/v1/attention/backends/mla/flashmla.py
```diff
@@ -62,6 +67,8 @@ class FlashMLAMetadata(MLACommonMetadata[FlashMLADecodeMetadata]):
 
 class FlashMLAMetadataBuilder(MLACommonMetadataBuilder[FlashMLAMetadata]):
     cudagraph_support: ClassVar[AttentionCGSupport] = AttentionCGSupport.UNIFORM_BATCH
+    query_len_support: ClassVar[QueryLenSupport] = QueryLenSupport.UNIFORM
+    reorder_batch_threshold: int = 512  # TODO(matt): tune this
```
benchislett: What is the point of this flag if the backend only supports decodes? Shouldn't `_init_reorder_batch_threshold` already set this to the max possible value dynamically?

Is it in order to support small prefills using the decode pathway? If so, please add a small comment to annotate this behaviour as it is not obvious to me as a reader

benchislett (changes requested):
minor issue flagged. otherwise LGTM, please ping me when resolved

vllm/v1/attention/backends/mla/common.py
MatthewBonanni: This is related to your other comment: all backends that support spec decode implicitly support `reorder_batch_threshold > 1`. For these backends, we also want small prefills to use the decode pathway, with the threshold tuned for optimal performance, so it's `> 1` even with spec decode disabled. It's unlikely that setting `reorder_batch_threshold` to `1` would ever be the optimal choice for such a backend, but you're right that if so, it would cause this assertion to fail. I'll remove it.

vllm/v1/attention/backends/mla/flashmla.py
MatthewBonanni: Yes, we want small prefills to use the decode pathway. This was first implemented in the FlashAttention MLA backend in #14258 (you can see the tuning results in the comments). I'll add a small comment as you suggest.

benchislett (approved):
LGTM, Thanks!

PR #26545, opened 2025-10-09
[ROCM] MoE fp4 CK kernel

maleksan85:
server:
```
VLLM_DISABLE_COMPILE_CACHE=1 \
AMDGCN_USE_BUFFER_OPS=1 \
VLLM_ROCM_USE_AITER=1 \
VLLM_TRITON_FP4_GEMM_USE_ASM=0 \
VLLM_ROCM_USE_AITER_MLA=1 \
VLLM_ROCM_USE_TRITON_ROPE=1 \
VLLM_ROCM_USE_CK_MXFP4_MOE=1 \
vllm serve /data/models/deepseek-ai/DeepSeek-R1-0528-WMXFP4-AMXFP4-Scale-UINT8-MoE-Quant-ASQ \
    --host localhost \
    --port 8000 \
    --swap-space 64 \
    --disable-log-requests \
    --dtype auto \
    --max-model-len 8192 \
    --tensor-parallel-size 8 \
    --max-num-seqs 1024 \
    --trust-remote-code \
    --block-size 1 \
    --gpu-memory-utilization 0.90 \
    --max-num-batched-tokens 131072 \
    --no-enable-prefix-caching \
    --async-scheduling \
    --compilation-config='{"pass_config":{"enable_attn_fusion":true,"enable_noop":true,"enable_fusion":true},"cudagraph_mode":"FULL","custom_ops":["+rms_norm","+silu_and_mul","+quant_fp8"],"splitting_ops":[]}'
```
client
```
curl http://localhost:8000/v1/chat/completions \
-d '{
    "model": "/data/models/deepseek-ai/DeepSeek-R1-0528-WMXFP4-AMXFP4-Scale-UINT8-MoE-Quant-ASQ",
    "messages": [
        {"role": "user", "content": "Who won the world series in 2023?"}
    ]}'
```

or 
```
VLLM_DISABLE_COMPILE_CACHE=1 \
AMDGCN_USE_BUFFER_OPS=1 \
VLLM_ROCM_USE_AITER=1 \
VLLM_TRITON_FP4_GEMM_USE_ASM=0 \
VLLM_ROCM_USE_AITER_MLA=1 \
VLLM_ROCM_USE_TRITON_ROPE=1 \
VLLM_ROCM_USE_CK_MXFP4_MOE=1 \
python /data/vllm-scripts/llm_test.py \
  --model /data/models/deepseek-ai/DeepSeek-R1-0528-WMXFP4-AMXFP4-Scale-UINT8-MoE-Quant-ASQ \
  --dataset-path /data/models/ShareGPT_Vicuna_unfiltered/ShareGPT_V3_unfiltered_cleaned_split.json \
  --batch-size 32 \
  --swap-space 64 \
  --dtype auto \
  --max-model-len 8192 \
  --tensor-parallel-size 8 \
  --max-num-seqs 1024 \
  --trust-remote-code \
  --block-size 1 \
  --gpu-memory-utilization 0.90 \
  --max-num-batched-tokens 131072 \
  --no-enable-prefix-caching \
  --async-scheduling \
  --compilation-config='{"pass_config":{"enable_attn_fusion":true,"enable_noop":true,"enable_fusion":true},"cudagraph_mode":"FULL","custom_ops":["+rms_norm","+silu_and_mul","+quant_fp8"],"splitting_ops":[]}'
```
returns reasonable results.

Correctness:
```
lm_eval --model local-completions   --model_args "base_url=http://localhost:8000/v1/completions,pretrained=/data/models/deepseek-ai/DeepSeek-R1-0528-WMXFP4-AMXFP4-Scale-UINT8-MoE-Quant-ASQ,tensor_parallel_size=8,add_bos_token=true,trust_remote_code=true"   --tasks gsm8k --num_fewshot 5 --limit 250   --batch_size 64
```

```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.972|±  |0.0105|
|     |       |strict-match    |     5|exact_match|↑  |0.972|±  |0.0105|
```


vllm/envs.py
```diff
@@ -926,6 +927,10 @@ def get_vllm_port() -> Optional[int]:
     "VLLM_ROCM_QUICK_REDUCE_MAX_SIZE_BYTES_MB": lambda: maybe_convert_int(
         os.environ.get("VLLM_ROCM_QUICK_REDUCE_MAX_SIZE_BYTES_MB", None)
     ),
+    # If set, enables CK fp4 MoE
+    "VLLM_ROCM_USE_CK_MXFP4_MOE": lambda: (
+        os.getenv("VLLM_ROCM_USE_CK_MXFP4_MOE", "False").lower() in ("true", "1")
```
mawong-amd: Should be enabled by default.

vllm/model_executor/layers/quantization/quark/quark_moe.py
```diff
@@ -566,6 +567,22 @@ def create_weights(
         layer.register_parameter("w13_weight_scale", w13_weight_scale)
         layer.register_parameter("w2_weight_scale", w2_weight_scale)
 
+    def process_weights_after_loading(self, layer):
+        if use_fp4_aiter_moe():
```
SageMoore: What's the expected behavior on a machine that doesn't support fp4? Will we crash if the environment variable is set?

tjtanaa: Is it possible to infer the datatype from the quark config if `VLLM_ROCM_USE_CK_MXFP4_MOE` is always set to true so that non-fp4 will not go into this path?

https://github.com/vllm-project/vllm/blob/d254b86794ae422aa4c941391d7bc50faab6f667/vllm/model_executor/layers/quantization/quark/quark_moe.py#L570-L584

vllm/model_executor/layers/quantization/quark/quark_moe.py
```diff
@@ -626,17 +641,38 @@ def apply(
...
-            inplace=True,
-            activation=activation,
-            global_num_experts=global_num_experts,
-            apply_router_weight_on_input=apply_router_weight_on_input,
-            expert_map=expert_map,
-            quant_config=self.moe_quant_config,
-        )
+        if use_fp4_aiter_moe():
+            from aiter import ActivationType, QuantType
+            from aiter.fused_moe import fused_moe
```
tjtanaa, line 650: Does it require new AITER version? I remember that the AITER commit in the `docker/Dockerfile.rocm_base` still requires the op to be wrapped in vllm's `direct_register_custom_ops`.

What's the aiter commit tested in this PR?

CC. @gshtras

vllm/model_executor/layers/quantization/quark/quark_moe.py
```diff
@@ -626,17 +641,38 @@ def apply(
...
+                x,
+                layer.w13_weight,
+                layer.w2_weight,
+                topk_weights,
+                topk_ids,
+                quant_type=QuantType.per_1x32,
+                w1_scale=layer.w13_weight_scale,
+                w2_scale=layer.w2_weight_scale,
+                activation=(
+                    ActivationType.Silu if activation == "silu" else ActivationType.Gelu
```
mgoin: Could you assert the activation instead of assuming else gelu?

vllm/envs.py
tjtanaa: If this is enabled by default, is it possible to infer the datatype from the quark config if `VLLM_ROCM_USE_CK_MXFP4_MOE` is always set to true so that non-fp4 will not go into the fp4 path to shuffle the weights for non-fp4 in quark moe?

https://github.com/vllm-project/vllm/blob/d254b86794ae422aa4c941391d7bc50faab6f667/vllm/model_executor/layers/quantization/quark/quark_moe.py#L570-L584

vllm/model_executor/layers/quantization/quark/quark_moe.py
maleksan85: well, to my knowledge there should be emulation. However deeper look showed that emulation is not yet implemented: https://github.com/vllm-project/vllm/blob/ff4810ba73168237f090b0195fbe0ae4d64e0730/vllm/model_executor/layers/quantization/quark/quark_moe.py#L475 cc @fxmarty-amd @BowenBao please comment if emulation support is planned.

vllm/model_executor/layers/quantization/quark/quark_moe.py
maleksan85: I'm running on pretty old version from rocm/aiter/355_wip: 
Name: aiter
Version: 0.1.5.dev128+g2d33708e2

vllm/envs.py
maleksan85: @tjtanaa well, this code only called on ocp_mx case: 
https://github.com/vllm-project/vllm/blob/ff4810ba73168237f090b0195fbe0ae4d64e0730/vllm/model_executor/layers/quantization/quark/quark_moe.py#L73
 for non fp4 cases other classes will be used.

vllm/model_executor/layers/quantization/quark/quark_moe.py
maleksan85: refactored, thanks!

vllm/model_executor/layers/quantization/quark/quark_moe.py
tjtanaa: Thank you for the explanation.

BowenBao: Hi @maleksan85, emulation is already supported. when `self.emulate = True` it simulates in bf16 matmul, with mxfp4 weight dequant and performs qdq over activation. This is part of today's main.

Please modify the else-branch setting `self.emulate=False` now that native kernel is being added. The else-branch was there to unblock running on gfx950 when there were no native kernel support in upstream.
https://github.com/vllm-project/vllm/blob/ff4810ba73168237f090b0195fbe0ae4d64e0730/vllm/model_executor/layers/quantization/quark/quark_moe.py#L484-L491

maleksan85: corrected the logic, thanks

BowenBao (approved):
Thanks for updates, LGTM on quark side.

mgoin (approved):
LGTM, thanks!

vllm/model_executor/layers/fused_moe/rocm_aiter_fused_moe.py
```diff
@@ -47,6 +47,11 @@ def is_rocm_aiter_moe_enabled() -> bool:
     )
 
 
+@cache
+def use_fp4_aiter_moe() -> bool:
```
fxmarty-amd:
```suggestion
def use_mxfp4_aiter_moe() -> bool:
```

vllm/model_executor/layers/quantization/quark/quark_moe.py
```diff
@@ -470,22 +471,19 @@ def __init__(
...
 
-        if not current_platform.supports_mx():
-            self.emulate = True
+        self.emulate = not current_platform.supports_mx() or not (
+            use_fp4_aiter_moe() and self.ocp_mx_scheme == "w_mxfp4_a_mxfp4"
+        )
+        if self.emulate:
             logger.warning_once(
-                "The current platform does not support native MXFP4/MXFP6 "
+                "The current mode does not support native MXFP4/MXFP6 "
```
fxmarty-amd: We can add `current_platform.supports_mx()`, `use_fp4_aiter_moe()`, `self.ocp_mx_scheme` values in the log

vllm/model_executor/layers/quantization/quark/quark_moe.py
```diff
@@ -626,17 +640,44 @@ def apply(
...
+
+            aiter_acts = {
+                ActivationType.No.name.lower(): ActivationType.No,
+                ActivationType.Silu.name.lower(): ActivationType.Silu,
+                ActivationType.Gelu.name.lower(): ActivationType.Gelu,
+            }
+            assert activation in aiter_acts, (
+                f"Aiter CK fp4 MoE doesn't support activation {activation}"
+            )
+            out = fused_moe(
```
fxmarty-amd, line 660: Why are we not using `torch.ops.vllm.rocm_aiter_fused_moe` or `vllm.model_executor.layers.fused_moe.rocm_aiter_fused_moe.rocm_aiter_fused_experts` like done in https://github.com/vllm-project/vllm/blob/00417f4e44378bc5c5c5c6698b4d8a86017198d3/vllm/model_executor/layers/quantization/quark/quark_moe.py#L310-L314 or in https://github.com/vllm-project/vllm/blob/00417f4e44378bc5c5c5c6698b4d8a86017198d3/vllm/model_executor/layers/quantization/fp8.py#L1246-L1252 or https://github.com/vllm-project/vllm/blob/00417f4e44378bc5c5c5c6698b4d8a86017198d3/vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors_moe.py#L1090-L1108

? I am not sure whether it is a good idea to deviate from the common implementation in https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/fused_moe/rocm_aiter_fused_moe.py.

For example, at the end of the day, `compressed-tensors` could also produce MXFP4 models that would be consumed by AITER MOE backend.

fxmarty-amd (approved):
cc @mgoin if you have an opinion on the above. Otherwise LGTM, think we should use `mxfp4` instead of `fp4` to clarify against nvfp4.

vllm/model_executor/layers/quantization/quark/quark_moe.py
maleksan85: sure

vllm/model_executor/layers/quantization/quark/quark_moe.py
maleksan85: not sure I got your message but mentioned MoEs are a8w8 MoEs, not fp4.

PR #26518, opened 2025-10-09
[Bugfix][Multi Modal] Fix incorrect output in Molmo

piood:
## Purpose

Fix incorrect spatial positioning output in Molmo multimodal model as reported in issue #26451.

The root cause is that PR #12966 removed the `image_input_idx` field during the TensorSchema migration, which eliminated the critical patch reordering logic. This caused image patches to remain in crop processing order instead of proper spatial order, leading to incorrect coordinate predictions for visual grounding tasks.

## Background

Molmo processes high-resolution images by splitting them into multiple crops (tiles) that are processed separately. The original implementation used `image_input_idx` to maintain spatial relationships between patches:

1. **Image tiling**: High-res images are divided into overlapping crops (e.g., 4×3 grid = 12 crops + 1 global thumbnail)
2. **Patch processing**: Each crop generates patches processed by the vision backbone
3. **Spatial reordering**: `image_input_idx` maps patches back to their correct spatial positions in the final sequence

Without this reordering, patches remain in crop processing order, breaking spatial continuity and causing coordinate prediction errors.

## Changes Made

1. **Restored `image_input_idx` field** to `MolmoImageInputs` TensorSchema as an optional field
2. **Updated multimodal field configuration** to include `image_input_idx` in the processing pipeline  
3. **Reimplemented patch reordering logic** in `_process_image_input()` method
4. **Added fallback behavior** when `image_input_idx` is unavailable to maintain backward compatibility

## Test Plan

**Test Script:**
```python
from vllm import LLM
from vllm.sampling_params import SamplingParams
import requests
from PIL import Image

model = LLM(
    model="allenai/Molmo-7B-D-0924",
    trust_remote_code=True,
    dtype='bfloat16',
    gpu_memory_utilization=0.95,
)

sampling_params = SamplingParams(max_tokens=64, temperature=0)

image_url = "https://www.visitscotland.com/binaries/content/gallery/visitscotland/cms-images/2022/06/24/clashnessie-bay-car-road"
image = Image.open(requests.get(image_url, stream=True).raw)

inputs = [{
    "prompt": "Point to the car.",
    "multi_modal_data": {"image": image},
}]

outputs = model.generate(inputs, sampling_params=sampling_params)
print(outputs[0].outputs[0].text)
```

## Test Results

**Before fix (incorrect coordinates):**
```xml
<point x="84.0" y="51.5" alt="car">car</point>
```

**After fix (correct coordinates):**
```xml
<point x="69.0" y="48.3" alt="car">car</point>
[...]

PR #12966, opened 2025-02-08
[VLM] Merged multi-modal processor for Molmo

DarkLight1337:
- Adopt merged multi-modal processor for Molmo
- Support different `num_crops` within the batch
- Support Molmo-O (thanks @Isotr0py!)

**Note:** The model implementation can handle multi-image input now, but we need to refactor the prompt replacement logic to enable end-to-end support in vLLM as the image tokens are inserted at the start of the string without replacing anything. This is left to future work.

### Example script

V0 output:
```
# python examples/offline_inference/vision_language.py -m molmo
# ----
 The image showcases a stunning view of the Tokyo Tower in Japan, captured from a low angle looking upwards. The iconic tower's distinctive white and yellow color scheme is prominently featured, with its cylindrical structure and observation deck clearly visible against a clear blue sky.

In the foreground, cherry blossom trees are in full bloom, their delicate
# ----
 The image captures a stunning view of the Tokyo Tower, a prominent landmark in Tokyo, Japan. The tower's distinctive architecture is on full display, with its tall, slender structure rising into the sky. The tower's white exterior contrasts beautifully with the vibrant blue sky, creating a striking visual.

The photograph is taken from a
# ----
 The image captures a stunning view of a tall, white tower with a distinctive spherical structure at its peak, resembling the Space Needle. This impressive architectural landmark is set against a clear blue sky, creating a striking contrast.

In the foreground, cherry blossom trees are in full bloom, their delicate pink flowers framing the tower beautifully.
# ----
 The image captures a stunning view of the Tokyo Tower in Japan, framed beautifully by cherry blossom branches. The tower's distinctive white and yellow structure stands tall against a clear blue sky, its iconic spherical middle section clearly visible. The cherry blossoms in the foreground are in full bloom, displaying their characteristic pink and white petals with
```

V1 output:
```
# VLLM_USE_V1=1 python examples/offline_inference/vision_language.py -m molmo
# ----
 The image captures a stunning view of a tall, white tower with a distinctive spherical structure at its peak, resembling a space needle. This impressive architectural feature stands out against a clear blue sky, creating a striking contrast.

In the foreground, cherry blossom trees are in full bloom, their delicate pink flowers providing a beautiful natural frame
# ----
 The image captures a stunning view of the Tokyo Tower, a prominent landmark in Tokyo, Japan. The tower's distinctive architecture is on full display, with its white and yellowish-white structure rising majestically against a clear blue sky.

The tower's unique design features a series of circular sections, starting with a wider base
# ----
[...]

PR #26563, opened 2025-10-10
[Bugfix][Multi Modal] Fix incorrect Molmo image processing

sangho-vision:
## Purpose
Restore correct Molmo outputs by re-introducing image patch re-ordering using `image_input_idx`.
In PR #12966, the re-ordering that maps image features to their corresponding patch tokens was removed and replaced with a boolean mask (`feat_is_patch`). That mask only indicates whether each image feature is **used** as a patch token in the multimodal input sequence, not *where* each valid patch feature should be placed.

In addition, the method `get_num_image_tokens` in `MolmoProcessingInfo` did not account for image start/end tokens as well as column separator tokens.

This PR fixes that regression by:
- Restoring correct patch-token placement using `image_input_idx`
- Correcting token-count computation to include start/end and column tokens.

FIX #26518

**Note**
PR #26451 also addresses the patch re-ordering issue, but it does not include the correction for `get_num_image_tokens` (which ensures image start/end and column tokens are properly counted).
This PR provides a more complete fix covering both aspects.

## Background
- Issue #26451 reported incorrect pointing results in Molmo.
- As discussed in #26518, removing `image_input_idx` broke the alignment between image features and patch-token positions.
- The boolean array `feat_is_patch` lacks the per-feature index mapping needed to reconstruct the original patch order.

## Changes made
1. Replace `feat_is_patch` field in `MolmoImageInputs` TensorSchema with `image_input_idx`
2. Update multimodal field configuration to include `image_input_idx` in the processing pipeline
3. Re-order patches back to their correct spatial positions in the final sequence using `image_input_idx`
4. Fix `get_num_image_tokens` to include start/end and column separator tokens in the total count

## Test Plan
Used the same test script from #26518:

```python
from vllm import LLM
from vllm.sampling_params import SamplingParams
import requests
from PIL import Image

model = LLM(
    model="allenai/Molmo-7B-D-0924",
    trust_remote_code=True,
    dtype='bfloat16',
    gpu_memory_utilization=0.95,
)

sampling_params = SamplingParams(max_tokens=448, temperature=0)

image_url = "https://www.visitscotland.com/binaries/content/gallery/visitscotland/cms-images/2022/06/24/clashnessie-bay-car-road"
image = Image.open(requests.get(image_url, stream=True).raw)

inputs = [{
    "prompt": "Point to the car.",
    "multi_modal_data": {"image": image},
}]

outputs = model.generate(inputs, sampling_params=sampling_params)
print(outputs[0].outputs[0].text)
```

## Test Result

```python
 <point x="69.0" y="48.6" alt="car">car</point>
```
The coordinates correctly align with the car in the image.

## Related
- See discussion in #26518
- Concurrent bug fix: #26451 
- Reverts the unintended effect introduced by #12966

### DIFF
--- a/vllm/model_executor/models/molmo.py
+++ b/vllm/model_executor/models/molmo.py
@@ -114,11 +114,11 @@ class MolmoImageInputs(TensorSchema):
         TensorShape("bn", "nc", "np", dynamic_dims={"nc"}),
     ]
 
-    feat_is_patch: Annotated[
+    image_input_idx: Annotated[
         Union[torch.Tensor, list[torch.Tensor]],
         TensorShape("bn", "nc", "tp", dynamic_dims={"nc"}),
     ]
-    # A boolean mask indicating which image features correspond to patch tokens.
+    # An index tensor that maps image features to their corresponding patch tokens.
     num_crops: Annotated[torch.Tensor, TensorShape("bn")]
 
 
@@ -1177,7 +1177,7 @@ class MolmoProcessorWrapper:
             num_crops = torch.tensor(tilings).prod(-1) + 1
             assert num_crops.sum() == len(feat_is_patch)
 
-            outputs["feat_is_patch"] = feat_is_patch
+            outputs["image_input_idx"] = image_input_idx
             outputs["num_crops"] = num_crops
             outputs["img_patch_id"] = self.image_patch_id
 
@@ -1211,8 +1211,9 @@ class MolmoProcessingInfo(BaseProcessingInfo):
         image_token_length_w = processor.image_token_length_w
         image_token_length_h = processor.image_token_length_h
 
-        extra = image_token_length_w * image_token_length_h
-        joint = ((ncols + 1) // pooling_size) * ((nrows + 1) // pooling_size)
+        # Calculate total tokens: 2 for start/end + (w+1)*h for column separators
+        extra = 2 + (image_token_length_w + 1) * image_token_length_h
+        joint = 2 + ((ncols + 1) // pooling_size + 1) * ((nrows + 1) // pooling_size)
 
         return extra + joint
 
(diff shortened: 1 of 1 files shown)


vllm/model_executor/models/molmo.py
```diff
@@ -1162,7 +1162,7 @@ def __call__(
         input_ids: torch.Tensor = outputs.pop("input_ids")
         outputs["input_ids"] = input_ids.unsqueeze(0)
 
-        image_input_idx = outputs.pop("image_input_idx", None)
+        image_input_idx = outputs.get("image_input_idx", None)
```
DarkLight1337: Is there a need to change this?

vllm/model_executor/models/molmo.py
```diff
@@ -1494,13 +1495,14 @@ def _process_image_input(
...
-        return [
-            feats[f_is_patch]
-            for feats, f_is_patch in zip(
-                image_features_flat.split(num_crops.tolist()),
-                feat_is_patch_flat.split(num_crops.tolist()),
-            )
-        ]
+        # Re-order the features using the image_input_idx tensor
+        results = []
+        for feats, img_idx in zip(image_features_flat.split(num_crops.tolist()), image_input_idx_flat.split(num_crops.tolist())):
```
DarkLight1337: Call `tolist` just once for efficiency

vllm/model_executor/models/molmo.py
```diff
@@ -1494,13 +1495,14 @@ def _process_image_input(
...
+            valid = img_idx >= 0
```
DarkLight1337: Rename to `is_valid`

vllm/model_executor/models/molmo.py
sangho-vision: yes, we need to return `image_input_idx`.

vllm/model_executor/models/molmo.py
sangho-vision: I pushed the fix.

vllm/model_executor/models/molmo.py
sangho-vision: I pushed the fix.

vllm/model_executor/models/molmo.py
DarkLight1337: Isn't this set in L1180?

sangho-vision: oh yeah you are right. I think I mixed up several patches I made.
~Should I fix this line, or can I just leave it as is? We can also delete L1180.~
Nvm. I reverted it.

DarkLight1337 (approved):
Thanks, LGTM

PR #24486, opened 2025-09-09
[Hybrid]: Decouple Kernel Block Size from KV Page Size

zhiyuan1i:
## Purpose
 This PR introduces a hybrid cache architecture that separates logical kernel block size from
  physical page size, enabling more flexible memory management. Key changes include:

  - Added kernel_block_size field to CacheConfig for logical block sizing
  - Enhanced platform-specific configurations for CUDA and ROCm to support hybrid blocks
  - Implemented block table conversion logic between physical and logical representations
  - Added support for different physical/logical block size ratios in V1 worker components

  This hybrid model decoupling enables independent development of high-performance operators
  without being constrained by linear attention mechanisms like Mamba, addressing performance
  bottlenecks discussed in issues https://github.com/vllm-project/vllm/issues/24280 and
  https://github.com/vllm-project/vllm/issues/23161.

## Test Plan
  Added comprehensive tests in tests/v1/worker/test_gpu_model_runner.py to verify:
  - Block table conversion between physical and logical representations
  - Proper handling of different block size ratios
  - Integration with existing GPU model runner functionality
  - Platform-specific configurations for CUDA and ROCm

## Test Result

pytest tests/v1/worker/test_gpu_model_runner.py - 20 passes
```
tests/v1/worker/test_gpu_model_runner.py ....................                                                                                                                        [100%]

===================================================================================== warnings summary =====================================================================================
../../../../opt/conda/envs/vllm-upstream/lib/python3.12/site-packages/torch/cuda/__init__.py:63
  /opt/conda/envs/vllm-upstream/lib/python3.12/site-packages/torch/cuda/__init__.py:63: FutureWarning: The pynvml package is deprecated. Please install nvidia-ml-py instead. If you did not install pynvml directly, please report this to the maintainers of the package that installed pynvml for you.
    import pynvml  # type: ignore[import]

<frozen importlib._bootstrap>:488
  <frozen importlib._bootstrap>:488: DeprecationWarning: builtin type SwigPyPacked has no __module__ attribute

<frozen importlib._bootstrap>:488
  <frozen importlib._bootstrap>:488: DeprecationWarning: builtin type SwigPyObject has no __module__ attribute

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
======================================================================== 20 passed, 3 warnings in 89.20s (0:01:29) =========================================================================
sys:1: DeprecationWarning: builtin type swigvarlink has no __module__ attribute
```
---
<details>
[...]

PR #26564, opened 2025-10-10
[BUG] Qwen3-next MTP. Fix attn metadata build bug

vadiklyutiy:
## Purpose
After fixing #24486 Qwen3-next with FlashInfer full attn start working without MTP. 
But with MTP it fails. 

The reason we choose incorrect attn metadata type for draft model (choose GDN instead of full attn). 

Fix it. 

## Test Result
Qwen3-next with MTP works now.


LucasWilkinson (approved):
Oh nice; theres already infra, LGTM!

PR #25103, opened 2025-09-17
Separate MLAAttention class from Attention

therealnaveenkamal:
## Purpose
This PR implements the first step of #24620 by separating Multi-Head Latent Attention into its own dedicated `AttentionLayerBase` subclass. 

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #26568, opened 2025-10-10
[Refactor][MLA]: Independently passing q_nope & q_rope

### LINKED ISSUE #26567: [Refactor][MLA]: Independently pass q_nope & q_rope

This PR: https://github.com/vllm-project/vllm/pull/25103 shows custom op for MLA

Original issue (https://github.com/vllm-project/vllm/issues/24620) wants to
> pass q_nope and q_rope independently instead of concatenated.

This will require a sizable refactor in order to split the two in every backend, so that both are passed truly independently.

This issue tracks progress towards this task, the current pr is this: https://github.com/vllm-project/vllm/pull/28368

cc @ProExpertProg

ghost:
This is outdated, I messed up src control. Here is the updated pr: https://github.com/vllm-project/vllm/pull/28368

I track in this issue: https://github.com/vllm-project/vllm/issues/26567

> open up the opportunity for MLA to its own custom op instead of unified_attention allowing for to potentially explore passing q_nope and q_rope independently instead of concatenated

Reference: https://github.com/vllm-project/vllm/issues/24620 and https://github.com/vllm-project/vllm/pull/25103

The buildthedocs fails because of the changes in flashmla_sparse.py. This will prob need to be overwritten.

cc @ProExpertProg @MatthewBonanni @LucasWilkinson

### DIFF
--- a/tests/v1/test_outputs.py
+++ b/tests/v1/test_outputs.py
@@ -1,0 +1,101 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+from unittest import TestCase
+
+from vllm.v1.outputs import LogprobsLists
+
+
+class TestLogprobsLists(TestCase):
+    def setUp(self):
+        self.logprobsLists = LogprobsLists(
+            logprob_token_ids=[
+                [1, 2],  # Request 0 token 0
+                [3, 4],  # Request 0 token 1
+                [5, 6],  # Request 1 token 0
+                [7, 8],  # Request 1 token 1
+                [9, 10],  # Request 1 token 2
+                [11, 12],  # Request 2 token 0
+                [13, 14],  # Request 2 token 1
+                [15, 16],  # Request 2 token 2
+                [17, 18],  # Request 2 token 3
+            ],
+            logprobs=[
+                [0.1, 0.2],
+                [0.3, 0.4],
+                [0.5, 0.6],
+                [0.7, 0.8],
+                [0.9, 1.0],
+                [1.1, 1.2],
+                [1.3, 1.4],
+                [1.5, 1.6],
+                [1.7, 1.8],
+            ],
+            sampled_token_ranks=[1, 3, 5, 7, 9, 11, 13, 15, 17],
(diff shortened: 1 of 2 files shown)


vllm/attention/backends/abstract.py
```diff
@@ -241,7 +241,7 @@ def __init__(
     def forward(
         self,
         layer: AttentionLayer,
-        hidden_states_or_cq: torch.Tensor,
+        hidden_states_or_cq: tuple[torch.Tensor, torch.Tensor],
```
ProExpertProg: Can we just make these separate parameters and name them appropriately?

vllm/v1/attention/backends/mla/flashmla_sparse.py
```diff
@@ -461,7 +461,7 @@ def _forward_fp8_kv(
     def forward(
         self,
         layer: AttentionLayer,
-        q: torch.Tensor,
+        q: Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]],
```
ProExpertProg: Why do we need to support both concat and separate here?

ProExpertProg (commented):
Let's split the tuple and update flashmla_sparse to the new interface as well! Also cc @MatthewBonanni and @LucasWilkinson - do we expect a win here somewhere because `q` no longer needs to be concatenated?

vllm/attention/backends/abstract.py
ghost: The tuple is used pretty much throughout the mla backends. it is also passed around as a single object in those backends. should I split the two in every backend?

vllm/v1/attention/backends/mla/flashmla_sparse.py
ghost: Pretty sure whichever kernel (I am looking at this: https://github.com/vllm-project/vllm/blob/main/vllm/attention/ops/flashmla.py#L121C1-L122C1 and https://github.com/vllm-project/vllm/blob/main/vllm/attention/ops/flashmla.py#L211 )  still expect the packed representation. 

I guess my primary answer to both your question is just that it requires a significantly larger refactor across both the mla backends

PR #25959, opened 2025-09-30
[NIXL][non-cuda] Add install script for nixl with non-cuda ucx

xuechendi:
## Purpose

Alternative solution for https://github.com/vllm-project/vllm/pull/25911
With nixl>=0.5.1, libcuda.so will be missing in UCX package while UCX will still link to libcuda.so. => Which leads to non-cuda platform will not be able to use pre-built NIXL wheel for UCX backend.

In this PR, propose to provide a build script for non-cuda platform NIXL.

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #26570, opened 2025-10-10
[XPU] Upgrade NIXL to remove CUDA dependency

zhenwei-intel:
> https://github.com/vllm-project/vllm/pull/25959 With nixl>=0.5.1, libcuda.so will be missing in UCX package while UCX will still link to libcuda.so. => Which leads to non-cuda platform will not be able to use pre-built NIXL wheel for UCX backend.

Update Dockerfile.xpu to build and install NIXL from source code, eliminating the libcuda dependency for better compatibility with XPU environments.

- Use `with-ze=no` to disable Level Zero support, the UCX build currently fails with Level Zero

### DIFF
--- a/tools/install_nixl_from_source_ubuntu.py
+++ b/tools/install_nixl_from_source_ubuntu.py
@@ -135,6 +135,7 @@ def build_and_install_prerequisites(args):
         "--enable-devel-headers",
         "--with-verbs",
         "--enable-mt",
+        "--with-ze=no",
     ]
     run_command(configure_command, cwd=ucx_source_path)
     run_command(["make", "-j", str(os.cpu_count() or 1)], cwd=ucx_source_path)
--- a/vllm/platforms/xpu.py
+++ b/vllm/platforms/xpu.py
@@ -54,6 +54,14 @@ class XPUPlatform(Platform):
         has_sink: bool,
         use_sparse,
     ) -> str:
+        from vllm.v1.attention.backends.utils import set_kv_cache_layout
+
+        set_kv_cache_layout("NHD")
+        logger.info(
+            "Setting VLLM_KV_CACHE_LAYOUT to 'NHD' for XPU; "
+            "only NHD layout is supported by XPU attention kernels."
+        )
+
         from vllm.attention.backends.registry import _Backend
 
         if use_sparse:
@@ -190,13 +198,6 @@ class XPUPlatform(Platform):
                 vllm_config.scheduler_config.max_model_len,
                 DEFAULT_MAX_NUM_BATCHED_TOKENS,
             )
-        from vllm.v1.attention.backends.utils import set_kv_cache_layout
-
-        set_kv_cache_layout("NHD")
-        logger.info(
-            "Setting VLLM_KV_CACHE_LAYOUT to 'NHD' for XPU; "
-            "only NHD layout is supported by XPU attention kernels."
-        )
 
     @classmethod
     def support_hybrid_kv_cache(cls) -> bool:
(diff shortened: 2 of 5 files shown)


vllm/platforms/xpu.py
```diff
@@ -54,6 +54,13 @@ def get_attn_backend_cls(
         has_sink: bool,
         use_sparse,
     ) -> str:
+        from vllm.v1.attention.backends.utils import set_kv_cache_layout
```
jikunshang, line 57: why move to here?

zhenwei-intel: aligned with cuda https://github.com/vllm-project/vllm/blob/main/vllm/platforms/cuda.py#L307

jikunshang: got. what format does tirton attention backend need? maybe we need to set it seprately.

tools/install_nixl_from_source_ubuntu.py
```diff
@@ -135,6 +135,7 @@ def build_and_install_prerequisites(args):
         "--enable-devel-headers",
         "--with-verbs",
         "--enable-mt",
+        "--with-ze=no",
```
jikunshang, line 138: it will build with ze by defualt?

jikunshang (approved):
overall LGTM.  trigger CI.

tools/install_nixl_from_source_ubuntu.py
zhenwei-intel: yes, enable ze by default.
Use `with-ze=no` to disable Level Zero support,

vllm/platforms/xpu.py
zhenwei-intel: `cuda.py` appears to be in the default format: `NHD`

tools/install_nixl_from_source_ubuntu.py
xuechendi: what is the problem build with ze?

vllm/platforms/xpu.py
xuechendi: That is not true, we had recent discussion on hpu side.
Hetero TP only works when prefill is HND. only when num_kv_heads is at outer dimension, nixl can use data_ptr to offset for different TP. Suggest don't limit XPU to NHD.

zhenwei-intel: The current PD on XPU does not support Hetero TP, and current XPU kernel only supports NHD format.

tools/install_nixl_from_source_ubuntu.py
zhenwei-intel:
```
copy/ze_copy_iface.c:199:30: error: initialization of 'void (*)(struct uct_iface *)' from incompatible pointer type 'void (*)(void)' [-Werror=incompatible-poi
nter-types]
  199 |     .iface_vfs_refresh     = ucs_empty_function,
      |                              ^~~~~~~~~~~~~~~~~~
copy/ze_copy_iface.c:199:30: note: (near initialization for 'uct_ze_copy_iface_internal_ops.iface_vfs_refresh')
copy/ze_copy_iface.c:200:30: error: initialization of 'ucs_status_t (*)(struct uct_ep *, uct_ep_attr_t *)' {aka 'ucs_status_t (*)(struct uct_ep *, struct uct_
ep_attr *)'} from incompatible pointer type 'ucs_status_t (*)(void)' [-Werror=incompatible-pointer-types]
  200 |     .ep_query              = ucs_empty_function_return_unsupported,
      |                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
copy/ze_copy_iface.c:200:30: note: (near initialization for 'uct_ze_copy_iface_internal_ops.ep_query')
copy/ze_copy_iface.c:201:30: error: initialization of 'ucs_status_t (*)(struct uct_ep *, unsigned int)' from incompatible pointer type 'ucs_status_t (*)(void)
' [-Werror=incompatible-pointer-types]
```
Cherry is working on UCX with ze.

PR #19482, opened 2025-06-11
[V1] Logit processors for rejection sampler

southfreebird:
Hi team,
Thank you for your amazing work!
This PR was prepared by the Nebius team.

This is a draft of the PR, so your suggestions are welcome! I would be happy to improve it or add anything I missed.

## Purpose
This PR adds logit processors to the rejection sampler, which allows the use of bad_words, penalties, and allowed_token_ids with speculative decoding. Previously, these parameters were silently ignored. The problem was described in issue: https://github.com/vllm-project/vllm/issues/18902

## Test Plan
You can test this functionality using the following pytest command:
```
pytest tests/v1/sample/test_rejection_sampler.py -k "test_frequency_penalties or test_bad_words or test_allowed_token_ids"
```
Or use the script provided in the issue. 

## Test Result

Server command:
```
vllm serve meta-llama/Llama-3.1-8B-Instruct --speculative-config '{"method": "eagle3", "model": "yuhuili/EAGLE3-LLaMA3.1-Instruct-8B", "num_speculative_tokens": 4}' --max-model-len 2048 --no-enable-prefix-caching
```

Client command:
```
vllm bench serve --model meta-llama/Llama-3.1-8B-Instruct --dataset-name sharegpt --num-prompts 200 --dataset-path ShareGPT_V3_unfiltered_cleaned_split.json --seed 0
```

Baseline:
```
============ Serving Benchmark Result ============
Successful requests:                     200       
Benchmark duration (s):                  10.76     
Total input tokens:                      42659     
Total generated tokens:                  43382     
Request throughput (req/s):              18.59     
Output token throughput (tok/s):         4032.80   
Peak output token throughput (tok/s):    4726.00   
Peak concurrent requests:                200.00    
Total Token throughput (tok/s):          7998.39   
---------------Time to First Token----------------
Mean TTFT (ms):                          887.85    
Median TTFT (ms):                        854.29    
P99 TTFT (ms):                           1306.54   
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          28.76     
Median TPOT (ms):                        18.41     
P99 TPOT (ms):                           159.41    
---------------Inter-token Latency----------------
Mean ITL (ms):                           26.70     
Median ITL (ms):                         24.23     
P99 ITL (ms):                            207.50    
==================================================
```

Measurements with a frequency-penalty enabled:
```
============ Serving Benchmark Result ============
Successful requests:                     200       
Benchmark duration (s):                  11.50     
[...]

PR #26586, opened 2025-10-10
[Bugfix] Fix error with penalties when speculative decoding and structural output are enabled

southfreebird:
Fix an error that appears after https://github.com/vllm-project/vllm/pull/19482 when logit processors (such as penalties) are enabled together with speculative decoding and structural output. The example of the error:

```
File "/vllm/model_executor/layers/utils.py", line 45, in get_token_bin_counts_and_mask
     bin_counts.scatter_add_(1, tokens, torch.ones_like(tokens))
RuntimeError: Expected index [24, 4162] to be no larger than self [21, 201089] apart from dimension 1 and to be no larger size than src [24, 4162]
```


vllm/v1/worker/gpu_model_runner.py
```diff
@@ -783,7 +783,14 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> None:
...
                 # NOTE(woosuk): `num_tokens` here may include spec tokens.
                 self.input_batch.num_tokens[req_index] += num_spec_tokens
-                self.input_batch.spec_token_ids[req_index] = spec_token_ids
+
+            # Note: When speculative decoding and structured output are enabled
+            # simultaneously, the scheduler can drop draft tokens that do not
+            # meet the structural schema. This means that
+            # scheduler_output.scheduled_spec_decode_tokens might be empty,
+            # even when speculative decoding is enabled. So, we moved this line
+            # from the 'if' block above.
```
benchislett: Please rephrase the comment so that it explains the state of the code and not the change to the code. Comments about moved lines can become less meaningful over time with refactoring

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -783,7 +783,14 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> None:
...
+            self.input_batch.spec_token_ids[req_index] = spec_token_ids
```
benchislett, line 800: Is there a case where spec token ids is None here, or is it always a list?

southfreebird: Based on annotations, `scheduled_spec_decode_tokens: dict[str, list[int]]` [here](https://github.com/vllm-project/vllm/blob/main/vllm/v1/core/sched/output.py#L146)
And by the code, it is always a list.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -783,7 +783,13 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> None:
                 ] = spec_token_ids
                 # NOTE(woosuk): `num_tokens` here may include spec tokens.
                 self.input_batch.num_tokens[req_index] += num_spec_tokens
-                self.input_batch.spec_token_ids[req_index] = spec_token_ids
+
+            # When speculative decoding is used with structured output,
```
benchislett, line 795: I don't think it makes sense to have this comment located here. It doesn't seem related to the assignment in any way. Would it make more sense to have it on line 775 next to the `.get()`?

benchislett: Nevermind, I forgot what the refactor did :facepalm:

benchislett (approved):
LGTM, Thanks!

njhill (approved):

PR #26587, opened 2025-10-10
[Misc] Refactor `get_kv_cache_spec` into `AttentionLayerBase`

NickLucche:
This PR modifies the `AttentionLayerBase` interface to add a new `get_kv_cache_spec` method.
This allows different attention layers to define their own KV Cache spec, by making the spec entirely transparent to the Model Runner.

As a consequence, the runner can now limit itself to collect the specs without having to handle different attention types and/or model-specific hacks such as the one for DSv32 Indexer. 
It also makes the code much simpler as all ENCODER,ENCODER_ONLY and ENCODER_DECODER type management is moved to a method dispatch system.

cc @heheda12345 who clearly defined the task
 
PS this used to be a TODO in code from @LucasWilkinson https://github.com/vllm-project/vllm/blob/releases/v0.11.0/vllm/v1/worker/gpu_model_runner.py#L4065

### DIFF
--- a/vllm/attention/layer.py
+++ b/vllm/attention/layer.py
@@ -172,7 +172,6 @@ class Attention(nn.Module, AttentionLayerBase):
             kv_cache_dtype = "auto"
             block_size = 16
             calculate_kv_scales = False
-        self.block_size = block_size
         self.kv_cache_torch_dtype = kv_cache_dtype_str_to_dtype(
             kv_cache_dtype, vllm_config.model_config
         )
@@ -409,6 +408,8 @@ class Attention(nn.Module, AttentionLayerBase):
         return self.attn_backend
 
     def get_kv_cache_spec(self, vllm_config: VllmConfig) -> KVCacheSpec:
+        # Block size may get updated after model loading, refresh it
+        block_size = vllm_config.cache_config.block_size
         # Should not be called for enc-dec or encoder-only attention.
         assert self.attn_type == AttentionType.DECODER
         if self.sliding_window is not None:
@@ -416,7 +417,7 @@ class Attention(nn.Module, AttentionLayerBase):
                 "MLA is not supported for slidingwindow"
             )
             return SlidingWindowSpec(
-                block_size=self.block_size,
+                block_size=block_size,
                 num_kv_heads=self.num_kv_heads,
                 head_size=self.head_size,
                 dtype=self.kv_cache_torch_dtype,
@@ -424,7 +425,7 @@ class Attention(nn.Module, AttentionLayerBase):
             )
         else:
             return FullAttentionSpec(
-                block_size=self.block_size,
+                block_size=block_size,
                 num_kv_heads=self.num_kv_heads,
                 head_size=self.head_size,
                 dtype=self.kv_cache_torch_dtype,
@@ -624,7 +625,6 @@ class MLAAttention(nn.Module, AttentionLayerBase):
             block_size = 16
             calculate_kv_scales = False
         self.kv_cache_dtype = kv_cache_dtype
-        self.block_size = block_size
 
         dtype = torch.get_default_dtype()
         self.attn_backend = get_attn_backend(
@@ -791,7 +791,7 @@ class MLAAttention(nn.Module, AttentionLayerBase):
             self.kv_cache_dtype, vllm_config.model_config.dtype
         )
         return MLAAttentionSpec(
-            block_size=self.block_size,
+            block_size=vllm_config.cache_config.block_size,
             num_kv_heads=1,
             head_size=self.head_size,
             dtype=kv_cache_dtype,
--- a/vllm/attention/layer.py
+++ b/vllm/attention/layer.py
@@ -787,7 +787,7 @@ class MLAAttention(nn.Module, AttentionLayerBase):
 
     def get_kv_cache_spec(self, vllm_config: VllmConfig) -> KVCacheSpec:
         kv_cache_dtype = kv_cache_dtype_str_to_dtype(
-            self.kv_cache_dtype, vllm_config.model_config.dtype
+            self.kv_cache_dtype, vllm_config.model_config
         )
         return MLAAttentionSpec(
             block_size=vllm_config.cache_config.block_size,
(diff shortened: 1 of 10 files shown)


vllm/attention/layers/chunked_local_attention.py
```diff
@@ -99,3 +101,13 @@ def __init__(
...
         )
+
+    def get_kv_cache_spec(self, vllm_config: VllmConfig) -> Optional[KVCacheSpec]:
+        assert self.attention_chunk_size
+        return ChunkedLocalAttentionSpec(
+            block_size=vllm_config.cache_config.block_size,
+            num_kv_heads=self.num_kv_heads,
+            head_size=self.head_size,
+            dtype=self.kv_cache_torch_dtype,
+            attention_chunk_size=self.attention_chunk_size,
```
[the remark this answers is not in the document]
heheda12345: @NickLucche I think this is a bug

vllm/attention/layers/encoder_only_attention.py
```diff
@@ -99,3 +101,7 @@ def __init__(
             attn_type=AttentionType.ENCODER_ONLY,
             **kwargs,
         )
+
+    def get_kv_cache_spec(self, vllm_config: VllmConfig) -> Optional[KVCacheSpec]:
```
heheda12345: Note that we also have `EncoderOnlyAttentionSpec`. We skip it in get_kv_cache_spec because these layers doesn't need kv cache, but add it back in `may_add_encoder_only_layers_to_kv_cache_config` as we need to build attention metadata for these layers. Can you try to make a better abstraction (leaving it as it is now may also be fine if no better idea)

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -4539,109 +4533,12 @@ def get_kv_cache_spec(self) -> dict[str, KVCacheSpec]:
...
-                )
-
-        ds_indexer_layers = get_layers_from_vllm_config(
-            self.vllm_config, DeepseekV32IndexerCache
-        )
-        for layer_name, ds_indexer_module in ds_indexer_layers.items():
-            kv_cache_spec[layer_name] = ds_indexer_module.get_kv_cache_spec()
+            # Skip modules that don't need KV cache (eg encoder-only attention)
+            if spec := attn_module.get_kv_cache_spec(self.vllm_config):
+                kv_cache_spec[layer_name] = spec
```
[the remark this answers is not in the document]
heheda12345: @NickLucche I think this make sense.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -4539,109 +4533,12 @@ def get_kv_cache_spec(self) -> dict[str, KVCacheSpec]:
...
-                    mamba_type=attn_module.mamba_type,
-                    num_speculative_blocks=(
-                        self.speculative_config.num_speculative_tokens
-                        if self.speculative_config
-                        else 0
-                    ),
-                )
-
-        ds_indexer_layers = get_layers_from_vllm_config(
-            self.vllm_config, DeepseekV32IndexerCache
```
heheda12345, line 4679: do you need to implement `get_kv_cache_spec` for `DeepseekV32IndexerCache`?

NickLucche: it already had one, just changed its signature

vllm/v1/worker/gpu_model_runner.py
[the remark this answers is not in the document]
NickLucche: nice catch, my bad!

vllm/model_executor/models/deepseek_v2.py
```diff
@@ -469,7 +469,7 @@ def __init__(
             raise ValueError(f"Duplicate layer name: {prefix}")
         compilation_config.static_forward_context[prefix] = self
 
-    def get_kv_cache_spec(self) -> KVCacheSpec:
+    def get_kv_cache_spec(self, vllm_config: VllmConfig) -> Optional[KVCacheSpec]:
```
NickLucche: @heheda12345 deepseek change

vllm/attention/layer.py
```diff
@@ -775,6 +812,18 @@ def calc_kv_scales(
     def get_attn_backend(self) -> type[AttentionBackend]:
         return self.attn_backend
 
+    def get_kv_cache_spec(self, vllm_config: VllmConfig) -> Optional[KVCacheSpec]:
+        kv_cache_dtype = kv_cache_dtype_str_to_dtype(
+            self.kv_cache_dtype, vllm_config.model_config.dtype
+        )
+        return MLAAttentionSpec(
```
LucasWilkinson, line 792: I dont think this is the right spec; the MLA spec should be used for:

https://github.com/vllm-project/vllm/blob/96ad65b7fe515663da8ede09a1aa7f74aa500c97/vllm/attention/layer.py#L567

I think this layer (`MultiHeadAttention`) multimodal models but tbh im not exactly sure where it is used

vllm/v1/spec_decode/eagle.py
ILikeIneine, line 23: I noticed here the instance are loaded directly from vllm's models, which is not registered by plugin. Would this be possible to fetch from Model Registry?

vllm/attention/layer.py
NickLucche: @LucasWilkinson not sure what's wrong with github preview, this change is actually to the 
`class MLAAttention(nn.Module, AttentionLayerBase):`

MHA is a simple nn.Module, it doesn't even implement the attn interface

NickLucche: like if you go to line 815 you can see it belongs to the mla class

vllm/v1/spec_decode/eagle.py
NickLucche: nice catch, I think it's outside the scope of the PR but we have to change that too

vllm/attention/layer.py
LucasWilkinson: oh sorry ya you are correct my bad! thats really weird that the GitHub preview didn't show that 🤔

LucasWilkinson (approved):
LGTM

PR #26601, opened 2025-10-10
[Log] Optimize Startup Log

yewentao256:
## Purpose

1. remove duplicate logs when using multi-cards
2. debug or delete some useless logs

### DIFF
--- a/vllm/distributed/device_communicators/pynccl.py
+++ b/vllm/distributed/device_communicators/pynccl.py
@@ -18,7 +18,6 @@ from vllm.distributed.device_communicators.pynccl_wrapper import (
     ncclRedOpTypeEnum,
     ncclUniqueId,
 )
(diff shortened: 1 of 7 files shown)


jeejeelee (approved):
LGTM

vllm/distributed/device_communicators/pynccl.py
```diff
@@ -106,7 +107,8 @@ def __init__(
         self.disabled = False
 
         self.nccl_version = self.nccl.ncclGetRawVersion()
-        logger.info("vLLM is using nccl==%s", self.nccl.ncclGetVersion())
+        if is_global_first_rank():
```
njhill: Do we still want to log on each node in multi-node case, i.e. should this instead be local first rank?

yewentao256: Fixed using rank 0, thanks!

PR #26602, opened 2025-10-10
Add support for the /rerank endpoint in vllm bench serve

maxdebayser:
The /rerank API can be support both by embedding models and native reranker models. However, with reranker models the query is concatenated with each document with a separator token in between. Therefore the amount of tokens that passes through the model has to be accounted for differently in each case.

Because of these details, this PR, adds a specialized random dataset to generates requests which send the expected amount of tokens. So when the use sets `random-input-len`, `num-prompts` and `random-batch-size`, in both cases we will generate requests such that the total amount of tokens is prompts*input-len in batches of size batch-size*input-len.

Here is an example of how this works. With the server running an reranker or embedding model
```
vllm serve BAAI/bge-reranker-v2-m3
```
run a benchmark using the vllm-rerank backend and the `random-rerank` dataset:
```
vllm bench serve --model BAAI/bge-reranker-v2-m3 --tokenizer BAAI/bge-reranker-v2-m3 \
  --backend vllm-rerank --endpoint /v1/rerank --dataset-name random-rerank \
  --random-input-len 512 --num-prompts 10 --random-batch-size 5
```


docs/contributing/benchmarks.md
```diff
@@ -878,6 +879,49 @@ vllm bench serve \
...
+  --backend vllm-rerank \
+  --endpoint /v1/rerank \
+  --dataset-name random-rerank \
+  --tokenizer BAAI/bge-reranker-v2-m3 \
+  --random-input-len 512 \
+  --num-prompts 10 \
+  --random-batch-size 5
+```
+
+For reranker models, this will create 2 rerank requests with 5 "documents" each where each
```
DarkLight1337: I think you should state explicitly that the number of requests is `num_prompts / random_batch_size` before using the example

vllm/benchmarks/datasets.py
```diff
@@ -638,6 +639,117 @@ def generate_token_sequence(
...
+class RandomDatasetForReranking(RandomDataset):
+    """
+    Random dataset specialized for the needs of scoring:
+    - Batches of inputs
+    - Inputs composed of pairs
+    """
+
+    # Default values copied from benchmark_serving.py for the random dataset.
+    DEFAULT_RANGE_RATIO = 0.0
+    DEFAULT_INPUT_LEN = 1024
```
ZJY0516: I think this can be derived from `RandomDataset`, so the definition here may not be necessary.

docs/contributing/benchmarks.md
maxdebayser: I've reworded this part, I hope it's clearer now.

docs/contributing/benchmarks.md
```diff
@@ -878,6 +879,51 @@ vllm bench serve \
...
 
+#### Reranker Benchmark
+
+Benchmark the performance of rerank requests in vLLM.
+
+<details class="admonition abstract" markdown="1">
+<summary>Show more</summary>
+
+Unlike generative models which use Completions API or Chat Completions API,
+you should set `--backend vllm-rerank openai-embeddings` and `--endpoint /v1/rerank` to use the Reranker API.
```
DarkLight1337:
```suggestion
you should set `--backend vllm-rerank` and `--endpoint /v1/rerank` to use the Reranker API.
```

docs/contributing/benchmarks.md
```diff
@@ -878,6 +879,51 @@ vllm bench serve \
...
+  --endpoint /v1/rerank \
+  --dataset-name random-rerank \
+  --tokenizer BAAI/bge-reranker-v2-m3 \
+  --random-input-len 512 \
+  --num-prompts 10 \
+  --random-batch-size 5
+```
+
+For reranker models, this will create `num-prompts / random-batch-size` requests with
+`random-batch-size` "documents" where each one has close to `random-input-len` tokens.
```
DarkLight1337:
```suggestion
For reranker models, this will create `num_prompts / random_batch_size` requests with
`random_batch_size` "documents" where each one has close to `random_input_len` tokens. 
```

docs/contributing/benchmarks.md
```diff
@@ -878,6 +879,51 @@ vllm bench serve \
...
+```
+
+For reranker models, this will create `num-prompts / random-batch-size` requests with
+`random-batch-size` "documents" where each one has close to `random-input-len` tokens. 
+In the example above, this results in 2 rerank requests with 5 "documents" each where
+each document has close to 512 tokens.
+
+Please note that the `/v1/rerank` is also supported by embedding models. So if you're running
+with an embedding model, also set `--no-reranker`. Because in this case the query is
+treated as a individual prompt by the server, here we send `random-batch-size - 1` documents
```
DarkLight1337:
```suggestion
treated as a individual prompt by the server, here we send `random_batch_size - 1` documents
```

I think snake_case is more idiomatic

vllm/benchmarks/datasets.py
```diff
@@ -638,6 +639,113 @@ def generate_token_sequence(
...
+        query_lens, _, query_offsets = self.get_sampling_params(
+            1, range_ratio, query_len_param, 0, tokenizer
+        )
+
+        query_len = int(query_lens[0])
+
+        if not is_reranker:
+            assert num_requests > 1 and batchsize > 1
+            num_requests = num_requests - 1
+            batchsize = batchsize - 1
```
DarkLight1337:
```suggestion
            num_requests -= 1
            batchsize -= 1
```

vllm/benchmarks/datasets.py
```diff
@@ -638,6 +639,113 @@ def generate_token_sequence(
...
+
+        if token_mismatch_total != 0:
+            sign = "more" if token_mismatch_total > 0 else "fewer"
+            logger.warning(
+                "Across all generated prompts, there were %d %s tokens "
+                "than expected after decoding and re-encoding. This is "
+                "expected due to the imperfect nature of the sampling "
+                "procedure.",
+                abs(token_mismatch_total),
+                sign,
```
DarkLight1337:
```suggestion
            logger.warning(
                "Across all generated prompts, there were %d %s tokens "
                "than expected after decoding and re-encoding. This is "
                "expected due to the imperfect nature of the sampling "
                "procedure.",
                abs(token_mismatch_total),
                "more" if token_mismatch_total > 0 else "fewer",
```

Inline this

DarkLight1337 (approved):

Issue #25277, opened 2025-09-19
[Bug]: Sequence Parallelism and Async TP disabled by default


ProExpertProg:
Currently, due to compilation issues, we only enable sequence parallelism (and the dependent AsyncTP) for static compile sizes (and not by default). That's because sequence parallelism splits the residual tensor into smaller pieces, which breaks with piecewise compilation and dynamic shapes. #21031 addressed this but got stuck on an Inductor bug that caused extreme memory pressure. That Inductor bug has since been resolved with PyTorch 2.9.

The course of action:
1. [x] Pick up #21031 and verify that [torch==2.9](https://dev-discuss.pytorch.org/t/pytorch-2-9-rc1-produced-for-pytorch/3230) resolved the memory issue (compare end-to-end activation memory with pass disabled and enabled).
2. [x] Test #21031 with changes in #24281 and `-O. use_inductor_graph_partition=True` to check that Inductor partitioning full compilation works with sequence parallelism.
3. [x] Check end-to-end performance of sequence parallelism alone as well as async TP on a dense unquantized and dense quantized model (both Hopper and Blackwell). Make sure to use full cudagraphs where available.
4. [ ] Merge the PR guarding on torch 2.9 and enable seq par and async tp by default to get performance on day 0 of torch 2.9 release.

Additionally, the padding requirement should be re-evaluated: we should benchmark to see the performance reduction in padding `num_tokens` and compare it to just manually padding with `-num_tokens % tp_size` around the sequence parallel section, or by doing uneven work across TP ranks by manipulating the sizes returned by `reduce_scatter` (likely too complicated)

linwang-aviva:
As discussed with Luka on Slack, leaving the following comment here:

Need to bring your attention for this:  previously we use ```vllm bench serve```  and we didn't see any claimed gain on Async TP + SP even without this PR [#21031](https://github.com/vllm-project/vllm/pull/21031).   Here is the full analysis report: 

[\[GPU\] Async Tensor Parallelism + Sequence Parallelism.pdf](https://github.com/user-attachments/files/22520298/GPU.Async.Tensor.Parallelism.%2B.Sequence.Parallelism.pdf)


We identified two key problems:

1. Lack of proper and comprehensive E2E testing in the original Async TP + SP PR:

 - It used the outdated and inaccurate vllm_latency.py script.

- Only one data point was tested: input length = 8K, output length = 1.

2. We did NSys analysis, and it shows expected fused kernels were not generated by Async TP.
[...]

PR #21031, opened 2025-07-16
Enable sequence parallelism for full cuda graph without specifying compile sizes

cascade812:
Previously it has to specify `compile_sizes` in `CompilationConfig` to enable sequence parallelism.
This PR is to remove this limitation for full cuda graph compilation.

PR #26604, opened 2025-10-10
[compile] Add patched_fused_scaled_matmul_reduce_scatter

angelayi:
## Purpose

In https://github.com/vllm-project/vllm/issues/25277#issuecomment-3386818025, while testing out https://github.com/vllm-project/vllm/pull/21031, we realized that `symm_mem::fused_scaled_matmul_reduce_scatter` is not properly symint-ified, meaning that if we trace a graph containing this op with dynamic shapes, the dynamic shape will become specialized. The proper fix is https://github.com/pytorch/pytorch/pull/165086, but this won't make it into the torch 2.9 release. So as a patch we will register a custom op in vllm which has the correct symint schema, fixing the specialization issue.

## Test Plan
`pytest -sv tests/compile/test_async_tp.py -k test_async_tp_pass_replace`

cc @zou3519 @ProExpertProg @cascade812

### DIFF
--- a/tests/compile/test_async_tp.py
+++ b/tests/compile/test_async_tp.py
@@ -142,7 +142,7 @@ class TestScaledMMRSModel(_BaseScaledMMModel):
         return [torch.ops.vllm.reduce_scatter.default]
 
     def ops_in_model_after(self):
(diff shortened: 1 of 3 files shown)


ProExpertProg (approved):
No notes, great PR!

PR #26622, opened 2025-10-11
[WIP][CustomOp] Make deepseek indexer custom op.

whx-sjtu:
## Purpose
This PR tries to make indexer in deepseek also a custom op in order to support oot backends. Currently I don't move `Indexer` into some files in `model_executor/layers` because it is used for deepseek only and is coupled with deepseek. Although I know that it might not be a good idea to introduce CustomOp into modeling files, I haven't thought of a better solution yet. If any reviewers have any better ideas, welcome to comment and I will continue to work on this, thanks.
cc @wangxiyuan @LucasWilkinson @ProExpertProg @Yikun 
## Test Plan
No need to add new tests.
## Test Result
All current tests should pass.
---
<details>

</details>

### DIFF
--- a/vllm/model_executor/models/deepseek_v2.py
+++ b/vllm/model_executor/models/deepseek_v2.py
@@ -47,6 +47,7 @@ from vllm.distributed import (
 )
 from vllm.forward_context import get_forward_context
 from vllm.logger import init_logger
+from vllm.model_executor.custom_op import CustomOp
 from vllm.model_executor.layers.activation import SiluAndMul
 from vllm.model_executor.layers.attention_layer_base import AttentionLayerBase
 from vllm.model_executor.layers.fused_moe import SharedFusedMoE
@@ -754,7 +755,8 @@ direct_register_custom_op(
 )
 
 
-class Indexer(nn.Module):
+@CustomOp.register("deepseek_indexer")
+class DeepseekIndexer(CustomOp):
     def __init__(
         self,
         vllm_config: VllmConfig,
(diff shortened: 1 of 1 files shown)


vllm/model_executor/models/deepseek_v2.py
```diff
@@ -754,7 +755,8 @@ def sparse_attn_indexer_fake(
 )
 
 
-class Indexer(nn.Module):
+@CustomOp.register("deepseek_indexer")
```
wangxiyuan, line 758: This is really out of scope for CustomOps IMO. Can you tell more about why we need to override Indexer? Maybe we can make Indexer more common first to solve the problem

whx-sjtu: The main problem here is the quantizing implementation of indexer. In current version, quantization of fp8 with ue8m0 is hard-coded which fails to work in backends that doesn't support fp8.

whx-sjtu: Maybe it's better to specially add a quantization backend for indexer?

wangxiyuan: Got your point, Indexer here is hard-code to use fp8 and triton. So IMO, the right direction is to make the ops in Indexer to support `customOps` and `quant_config` way like others do. cc @youkaichao

wangxiyuan: After more work, I find that the Indexer is used by MLA module which is custom op already. You can add your indexer inner mla, so this PR is not required IMO.

whx-sjtu: Currently I can implement an `IndexerWrapper` class in custom MLA module,  which accepts the `Indexer` here as a initialization parameter, inherits the module weights defined in `Indexer`, and rewrites its forward method. But I don’t think this is a formal solution. Do we still need to design a formal technical solution to make `Indexer` more general? cc @wangxiyuan @youkaichao. If current state looks good to you, I will close this PR.

ProExpertProg (changes requested):
Agree with @wangxiyuan here, this should just be handled inside the custom MLA

PR #26623, opened 2025-10-11
Improve enable chunked_prefill & prefix_caching logic.

noooop:
## Purpose

Improve enable chunked_prefill & prefix_caching logic.

## Test Plan

pytest -s -vvv tests/test_config.py::test_is_chunked_prefill_supported
pytest -s -vvv tests/test_config.py::test_is_prefix_caching_supported
pytest -s -vvv tests/models/language/pooling/test_auto_prefix_cache_support.py

## Test Result

pass

---
<details>

</details>


tests/models/language/pooling/test_auto_prefix_cache_support.py
```diff
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+# ruff: noqa: SIM117
```
DarkLight1337: Is this really needed?

vllm/model_executor/models/interfaces_base.py
```diff
@@ -212,3 +212,17 @@ def func(model: _T) -> _T:
 
 def get_default_pooling_type(model: type[object] | object) -> str:
     return getattr(model, "default_pooling_type", "LAST")
+
+
+def attn_type(attn_type: str):
+    """Decorator to set `VllmModelForPooling.attn_type`."""
+
+    def func(model: _T) -> _T:
+        model.attn_type = attn_type  # type: ignore
```
DarkLight1337, line 235: Please add this field to the interface definition

vllm/config/model.py
```diff
@@ -1764,6 +1764,68 @@ def get_and_verify_max_len(self, max_model_len: int):
         logger.info("Using max model len %s", max_model_len)
         return max_model_len
 
+    @property
+    def attn_type(self) -> str:
```
DarkLight1337: Make the attention type a `typing.Literal`?

tests/models/language/pooling/test_auto_prefix_cache_support.py
noooop: If you remove this, ruff will complain about the code below.

```
    with pytest.raises(ValueError):
        with vllm_runner(
```

If changed to the code below, pytest.raises cannot capture ValueError.

```
with pytest.raises(ValueError) and vllm_runner() as vllm_model:
  ...
```

noooop: If chunked_prefill or prefix_caching is not supported but setting, now will raise ValueError. 

Because this setting will definitely cause errors in the executor or kv cache, it's best to completely prevent users from trying.

DarkLight1337: We should just put the noqa on that specific line then, instead of ignoring the whole file

vllm/config/utils.py
```diff
@@ -293,3 +293,28 @@ def get_hash_factors(config: ConfigT, ignored_factors: set[str]) -> dict[str, ob
...
+            raise ValueError(self.reason)
+
+
+class BoolWithReasonGroup(dict):
+    def __setattr__(self, key: str, value: BoolWithReason):
+        assert isinstance(value, BoolWithReason)
+        self[key] = value
+
+    def __getattr__(self, key: str) -> BoolWithReason:
+        return self[key]
```
noooop: @DarkLight1337 

Does vllm have an implementation similar constant group?

This implementation has a single issue where the IDE cannot jump to the definition. Set as a module variable can.

DarkLight1337: Maybe you can just create a class and set the constants as class variables

DarkLight1337: Like

```
class cp_reasons:
    GENERATIVE_MODELS_SUPPORT_CHUNKED_PREFILL = BoolWithReason(
        value=True, reason="Generative models support chunked prefill."
    )
    PREFILL_CONTEXT_PARALLEL_NOT_SUPPORT_CHUNKED_PREFILL = BoolWithReason(
        value=False,
        reason="Prefill context parallel does not compatible with chunked prefill.",
    )

```

zhuohan123 (changes requested):
Why do we need a new class for bool? I think these kind of classes add mental overhead for no good reason. Will leave more comments tmrw.

hmellor (commented):
I agree, all the `reason`s should just be warnings

vllm/engine/arg_utils.py
```diff
@@ -250,6 +253,8 @@ def _compute_kwargs(cls: ConfigType) -> dict[str, dict[str, Any]]:
                 default = (
                     default.default
                     if default.default_factory is None
+                    # FIXME: A default vllm_config will be constructed here.
+                    # Info logs will be output, which is confusing.
                     else default.default_factory()
                 )
         elif field.default_factory is not MISSING:
```
noooop: @DarkLight1337 @hmellor 

A default vllm_config will be constructed here, including the creation of a default SchedulerConfig, and the following info log will be output, which is confusing.

For example, BAAI/bge-base-en does not support chunked prefill, so chunked prefill is not used by default.

vllm serve BAAI/bge-base-en


```
INFO 11-22 15:35:41 [config/scheduler.py:207] Chunked prefill is enabled with max_num_batched_tokens=2048.
```

vllm/config/cache.py
```diff
@@ -73,7 +73,7 @@ class CacheConfig:
     sliding_window: int | None = None
     """Sliding window size for the KV cache. This is primarily set in
     `ModelConfig` and that value should be manually duplicated here."""
-    enable_prefix_caching: bool = True
+    enable_prefix_caching: bool | None  = None
     """Whether to enable prefix caching."""
```
noooop: @DarkLight1337 


After #29186, EngineArgs.enable_prefix_caching: bool | None = None is not working, EngineArgs defaults to True always.

We still need None here, otherwise we can't distinguish between user-enforced input and the default.

vllm/config/cache.py
```diff
@@ -73,7 +73,7 @@ class CacheConfig:
     sliding_window: int | None = None
     """Sliding window size for the KV cache. This is primarily set in
     `ModelConfig` and that value should be manually duplicated here."""
-    enable_prefix_caching: bool = True
+    enable_prefix_caching: bool | None = None
```
DarkLight1337: We should set the defaults to non-None values before initializing `CacheConfig`, otherwise it makes initialization more complicated.

noooop: I can't think of a better way

***

After #29186, EngineArgs.enable_prefix_caching: bool | None = None is not working, EngineArgs defaults to True always. by

```
        cache_group.add_argument(
            "--enable-prefix-caching", **cache_kwargs["enable_prefix_caching"]
        )
```

We still need None here, otherwise we can't distinguish between user-enforced input vs the default.

DarkLight1337: Oops, I forgot to add the default here. It should be added like

```
        scheduler_group.add_argument(
            "--enable-chunked-prefill",
            **{
                **scheduler_kwargs["enable_chunked_prefill"],
                "default": None,
            },
        )
```

noooop:
@zhuohan123 @DarkLight1337 @hmellor 
Ready for review

***

@tdoublep
Please help review this PR. This PR is also related to #29393

zhuohan123 (approved):
Thanks for the hard work and the fix!

tests/engine/test_arg_utils.py
```diff
@@ -277,19 +277,24 @@ def test_prefix_cache_default():
     parser = EngineArgs.add_cli_args(FlexibleArgumentParser())
     args = parser.parse_args([])
 
-    # should be None by default (depends on model).
+    # should default to on.
     engine_args = EngineArgs.from_cli_args(args=args)
-    assert engine_args.enable_prefix_caching is None
+    vllm_config = engine_args.create_engine_config()
+    assert vllm_config.cache_config.enable_prefix_caching, (
```
DarkLight1337: Why do we need to change this? It should be testing `EngineArgs` only (being in `arg_utils.py`), not `VllmConfig`

noooop: Because engine_args.create_engine_config() will call _set_default_chunked_prefill_and_prefix_caching_args, otherwise enable_prefix_caching remains none, which is trivial and does not need testing.

DarkLight1337: The test is to test whether `EngineArgs.from_cli_args` is working properly, not the conversion into config.

noooop: Okay, I'm reverting this change

vllm/config/model.py
```diff
@@ -1752,6 +1757,114 @@ def get_and_verify_max_len(self, max_model_len: int):
...
+                        "Pooling models with mean pooling does not support "
+                        "chunked prefill."
+                    )
+                    return False
+                elif pooling_type == "mean":
+                    logger.debug(
+                        "Pooling models with all pooling does not support "
+                        "chunked prefill."
+                    )
+                    return False
```
nwaughachukwuma: 1. The logging and comments are mixed up. Please clean it up. 

For example, `pooling_type == 'all'` should have a logging.debug message of "Pooling models with all pooling...", same for `pooling_type == 'mean'`.

2. The comment "# for decoder models (causal attn)" before `pooling_type == 'all'` doesn't seem appropriate there. No?

nwaughachukwuma (changes requested):
Thanks for doing an excellent job here. 

However, there're so many not-too-related changes in this PR across multiple touch points. To make reviewing and merging faster, please break this down into multiple PRs, each with relative changes.

Thank you.

vllm/config/model.py
noooop: Oh, I just remembered there's also step pooling.

vllm/engine/arg_utils.py
```diff
@@ -1984,6 +1933,29 @@ def _set_default_args(
                 scope="local",
             )
```
noooop, line 1934: here

PR #26625, opened 2025-10-11
[Feature] Optimize Prefill Phase: Add Hybrid Chunked Prefill Support

Ther-LF:
# Optimize Prefill Phase: Add Hybrid Chunked Prefill Support 

## Description
This PR introduces **Hybrid Chunked Prefill**, an optimization designed to dynamically switch between **continuous prefill** and **chunked prefill** in the vLLM serving pipeline.

**Why this matters:**  
Today users enable **chunked prefill** mainly to reduce inter-token latency (ITL) when prefill and decode overlap. But chunking also splits long prefill segments, increasing launch/coord overhead and **hurting throughput**. The current strategy applies chunked prefill unconditionally, so we keep paying the throughput tax **even when no decode requests are running**.
**Hybrid Chunked Prefill** fixes this by enabling chunking **only when decode is active**; otherwise it falls back to **continuous prefill, recovering baseline throughput** while still preserving the ITL benefits when needed.

This feature enables vLLM to achieve **higher throughput** and **lower latency** in both low and high concurrency scenarios.

---

## Purpose
- Implement **adaptive hybrid prefill scheduling** that improves efficiency across different concurrency levels.  
- Reduce **prefill fragmentation** and **launch overhead** when decode traffic is low.  
- Enhance **scalability and stability** for large models such as QwQ-32B in multi-request serving.

---

## Test Plan
- **Model:** `QwQ-32B`
- **Dataset:** 20 representative prompts (mixed-length, real inference-like workloads)
- **Setup:** Evaluate with `--concurrency {1, 2, 4, 8}`
- **Comparison:**  
  1. `--enable-chunked-prefill --max-num-batched-tokens 1024`  
  2. `--enable-hybrid-chunked-prefill --max-num-batched-tokens 1024`
- **Metrics:**
  - Request throughput (req/s)
  - Token throughput (tok/s)
  - Time to First Token (TTFT)
  - Time per Output Token (TPOT)
  - End-to-End Latency (E2E)

---

## Test Result

| Concurrency | Mode | Request Throughput (req/s) | Token Throughput (tok/s) | Mean TTFT (ms) | Mean TPOT (ms) | Mean E2E (ms) |
|:------------:|:------|:--------------------------:|:-------------------------:|:---------------:|:---------------:|:---------------:|
| **1** | Hybrid Chunked (1024) | 0.26 | **3643.7** | **1085.9** | 7.21 | **3851.0** |
|  | Chunked (1024) | 0.26 | 3580.4 | 1368.8 | 7.20 | 3910.3 |
| **2** | Hybrid Chunked  (1024) | 0.35 | **4928.4** | 1410.0 | 9.79 | **5418.1** |
|  | Chunked (1024) | 0.34 | 4811.6 | 1400.5 | 9.55 | 5549.8 |
| **4** | Hybrid Chunked (1024) | 0.45 | **6336.2** | 1601.6 | **15.9** | **8080.0** |
|  | Chunked (1024) | 0.45 | 6281.1 | 1645.5 | 15.9 | 8107.3 |
| **8** | Hybrid Chunked (1024) | **0.52** | **7335.9** | **2830.5** | **30.5** | **13556.9** |
|  | Chunked (1024) | 0.51 | 7206.1 | 2847.5 | 31.3 | 13599.8 |

---

## Observations
- **+2–5% higher total token throughput** across all concurrency levels.  
- **10–20% lower TTFT**, especially under low concurrency.  
- **Stable scaling up to concurrency = 8**, maintaining high throughput without latency degradation.  
- Prefill efficiency improves since hybrid mode avoids unnecessary chunk splitting when decode is idle.

---

Hybrid Chunked Prefill achieves **higher throughput with equal or better latency**, demonstrating its effectiveness on large-scale inference workloads.

---

## Comment
This PR adds an **adaptive prefill mechanism** to vLLM that dynamically balances throughput and latency.  
By intelligently enabling chunked prefill only when necessary, it significantly improves efficiency for models like **QwQ-32B** under real concurrent serving scenarios.

---

<details>

- [x] Purpose: Introduce hybrid prefill scheduling to optimize serving performance.
- [x] Test Plan: Detailed benchmark procedure and dataset.
- [x] Test Results: Comparative performance table and observations.
- [ ] (Optional) Documentation update (`supported_features.md`).
- [ ] (Optional) Release notes update for upcoming vLLM version.
</details>


vllm/v1/core/sched/scheduler.py
```diff
@@ -190,7 +202,14 @@ def schedule(self) -> SchedulerOutput:
...
         req_to_new_blocks: dict[str, KVCacheBlocks] = {}
         num_scheduled_tokens: dict[str, int] = {}
-        token_budget = self.max_num_scheduled_tokens
+
+        # Check if there are any requests in the decode phase in the running queue.
+        has_decode_requests = self._has_decode_requests()
+        if self.scheduler_config.enable_hybrid_chunked_prefill and not has_decode_requests:
+            token_budget = self.prefill_max_num_scheduled_tokens
+        else:
+            token_budget = self.max_num_scheduled_tokens
```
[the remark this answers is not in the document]
Ther-LF: This is addressed by modifying max_num_tokens directly in vllm/v1/worker/gpu_model_runner.py, which ensures that all downstream buffer allocations in the worker stack are consistently updated — so the stack size is uniformly increased when hybrid prefill raises the per-step token budget.

vllm/config/scheduler.py
```diff
@@ -37,6 +37,11 @@ class SchedulerConfig:
     This config has no static default. If left unspecified by the user, it will
     be set in `EngineArgs.create_engine_config` based on the usage context."""
 
+    prefill_max_num_batched_tokens: SkipValidation[int] = None
```
hmellor: Why `SkipValidation`? If it's because this will be set later by us if `None` you can do:

```suggestion
    prefill_max_num_batched_tokens: int = Field(default=None)
```

This will skip validateion for the default `None` but validate passed values.

vllm/config/scheduler.py
```diff
@@ -83,6 +88,11 @@ class SchedulerConfig:
     """If True, prefill requests can be chunked based
     on the remaining max_num_batched_tokens."""
 
+    enable_hybrid_chunked_prefill: SkipValidation[bool] = None  # type: ignore
```
hmellor: Why `SkipValidation`? If it's because this will be set later by us if `None` you can do:

```suggestion
    enable_hybrid_chunked_prefill: bool = Field(default=None)
```

This will skip validateion for the default `None` but validate passed values.

hmellor (dismissed):
Please fix the following before review:

- new config args
- pre-commit
- dco
- docs build

vllm/config/scheduler.py
```diff
@@ -37,6 +37,11 @@ class SchedulerConfig:
     This config has no static default. If left unspecified by the user, it will
     be set in `EngineArgs.create_engine_config` based on the usage context."""
 
+    prefill_max_num_batched_tokens: int | None = Field(default=None)
```
hmellor: If we are going to type hint `| None` then the `Field` is not necessary

```suggestion
    prefill_max_num_batched_tokens: int | None = None
```

vllm/config/scheduler.py
```diff
@@ -83,6 +88,11 @@ class SchedulerConfig:
     """If True, prefill requests can be chunked based
     on the remaining max_num_batched_tokens."""
 
+    enable_hybrid_chunked_prefill: bool | None = Field(default=None)  # type: ignore
```
hmellor: If we are going to type hint `| None` then the `Field` is not necessary

```suggestion
    enable_hybrid_chunked_prefill: bool | None = None
```

Ther-LF:
@hmellor Thanks for the suggestion! I’ve updated both fields to use `Field(default=None)` and annotated them as `int | None` / `bool | None`, so defaults aren’t validated but user-provided values are. `None` will be resolved in `EngineArgs.create_engine_config` based on the usage context.

```py
from pydantic import Field

prefill_max_num_batched_tokens: int | None = Field(default=None)
enable_hybrid_chunked_prefill: bool | None = Field(default=None)
```

vllm/config/scheduler.py
```diff
@@ -37,6 +38,11 @@ class SchedulerConfig:
     This config has no static default. If left unspecified by the user, it will
     be set in `EngineArgs.create_engine_config` based on the usage context."""
 
+    prefill_max_num_batched_tokens: int = None  # type: ignore
```
hmellor: Is mypy the reason that `type: ignore` is needed?

```suggestion
    prefill_max_num_batched_tokens: int = None  # type: ignore
```

vllm/engine/arg_utils.py
```diff
@@ -475,6 +475,7 @@ class EngineArgs:
     ignore_patterns: str | list[str] = get_field(LoadConfig, "ignore_patterns")
 
     enable_chunked_prefill: bool | None = SchedulerConfig.enable_chunked_prefill
+    enable_hybrid_chunked_prefill: bool = SchedulerConfig.enable_hybrid_chunked_prefill
```
hmellor: Do we need a different flag for this? 

I would argue that `enable_chunked_prefill` is sufficient.

Ther-LF: The hybrid chunked prefill feature is designed to switch between chunked prefill and continuous prefill, offering significant benefits by optimizing performance based on workload. It helps reduce unnecessary overhead and improves throughput, especially when there are no decode requests, making it a valuable addition for efficiency in various scenarios.

However, to implement this functionality, a separate flag is necessary. Enabling hybrid prefill will trigger continuous prefill, which changes the max_num_tokens to max_model_len in gpu_model_runner.py, leading to higher memory usage. Since many systems use chunked prefill due to memory constraints, not having an extra flag could potentially cause out-of-memory (OOM) issues in those environments.

vllm/config/scheduler.py
Ther-LF: Yes, mypy is the reason # type: ignore is needed here.

vllm/engine/arg_utils.py
hmellor: I agree that this is a valuable feature.

So are you saying that

```py
enable_chunked_prefill = True
enable_hybrid_chunked_prefill = True
```

will behave differently to

```py
enable_chunked_prefill = False
enable_hybrid_chunked_prefill = True
```

for a hybrid model?

Ther-LF: They will behave the same way.

Both 
```
enable_chunked_prefill = True
enable_hybrid_chunked_prefill = True 
```
as well as 
```
enable_chunked_prefill = False 
enable_hybrid_chunked_prefill = True 
```
will enable hybrid chunked prefill. Currently, vLLM defaults to having enable_chunked_prefill = True.
However, it might be necessary to add an assertion check for the case?
```
enable_chunked_prefill = False 
enable_hybrid_chunked_prefill = True
```
My understanding is that chunked prefill should be enabled first before hybrid chunked prefill can be activated. This ensures that the hybrid model is used correctly.

Ther-LF:
```
enable_chunked_prefill = True 
enable_hybrid_chunked_prefill = False 
```
will always enable chunked prefill, not the hybrid model.
```
enable_chunked_prefill = False 
enable_hybrid_chunked_prefill = False
```
 will always use continuous prefill.

vllm/config/scheduler.py
```diff
@@ -310,4 +346,12 @@ def _verify_args(self) -> Self:
                 f"max_num_partial_prefills ({self.max_num_partial_prefills})."
             )
 
+        if self.enable_hybrid_chunked_prefill and not self.chunked_prefill_enabled:
```
Ther-LF: I have added the validation for hybrid chunked prefill here

vllm/engine/arg_utils.py
hmellor: I still don't understand why we can't use `enable_chunked_prefill` to enable chunked prefill in hybrid models

Ther-LF: My understanding is that we should first ensure the existing chunked prefill functionality remains unchanged (since chunked prefill also helps reduce memory usage), and then consider optimizing for hybrid mode within chunked prefill (as it brings additional memory overhead).

Ther-LF: This is also why I think this situation:
```
enable_chunked_prefill = False
enable_hybrid_chunked_prefill = True
```
should not occur. Chunked prefill should be enabled first, and then we should consider whether to enable hybrid mode within chunked prefill, rather than considering enabling chunked prefill within hybrid mode.

vllm/config/scheduler.py
hmellor: It's possible to trick mypy by using `Field(default=None)` here so that you can remove the ignore.

vllm/engine/arg_utils.py
hmellor: Ok normal chunked prefill and hybrid chunked prefill are 2 different features that can coexist in the same model?

This is the question I've been wondering about.

Ther-LF: Yes, both of them can coexist because, although the current chunked prefill always performs chunk-based calculations for requests, causing performance loss when there are no decode requests, it still limits the maximum number in `max_num_tokens` in `gpu_model_runner.py` to not exceed the chunk size. This helps reduce memory usage by keeping it from growing to `max_model_len` and instead limits it to `max_num_batched_tokens`.

However, with hybrid mode enabled, since we switch to continuous prefill when there are no decode requests, the `max_num_tokens` in `gpu_model_runner.py` becomes `max_model_len`, which leads to additional memory overhead.

Therefore, I believe we should ensure the existing chunked prefill functionality remains unchanged, as some systems rely on chunked prefill to reduce memory overhead. For systems with larger memory capacity, they can enable hybrid mode to improve their performance.

hmellor: Oh I'm sorry I think I misunderstood. I thought hybrid was referring to Mamba!

vllm/config/scheduler.py
```diff
@@ -181,6 +201,12 @@ def __post_init__(self, is_encoder_decoder: bool) -> None:
                 " prefix caching; disabling both."
             )
 
+        if self.enable_hybrid_chunked_prefill is None:
+            self.enable_hybrid_chunked_prefill = False
```
hmellor: If we immediately set it to `False` can we just set the actual default to `False` instead of `None`?

vllm/config/scheduler.py
```diff
@@ -83,6 +89,11 @@ class SchedulerConfig:
     """If True, prefill requests can be chunked based
     on the remaining max_num_batched_tokens."""
 
+    enable_hybrid_chunked_prefill: bool = None  # type: ignore
```
```suggestion
    enable_hybrid_chunked_prefill: bool = False
```

vllm/config/scheduler.py
```diff
@@ -163,6 +174,15 @@ def compute_hash(self) -> str:
         hash_str = hashlib.md5(str(factors).encode(), usedforsecurity=False).hexdigest()
         return hash_str
 
+    @field_validator(
+        "prefill_max_num_batched_tokens", "enable_hybrid_chunked_prefill", mode="wrap"
+    )
+    @classmethod
```
hmellor, line 201:
```suggestion
    @field_validator("prefill_max_num_batched_tokens", mode="wrap")
    @classmethod
```

vllm/config/scheduler.py
Ther-LF: I have already used `Field(default=None)` for `enable_hybrid_chunked_prefill` and `prefill_max_num_batched_tokens` in the latest commit,, and removed `# type: ignore` in the .

vllm/config/scheduler.py
Ther-LF: Thank you for your suggestion, I have set it to False in the latest commit.

vllm/config/scheduler.py
```diff
@@ -37,6 +38,11 @@ class SchedulerConfig:
     This config has no static default. If left unspecified by the user, it will
     be set in `EngineArgs.create_engine_config` based on the usage context."""
 
+    prefill_max_num_batched_tokens: int = Field(default=None)
```
hmellor: This doesn't seem to be configurable at init time because of the immediate

```py
        self.prefill_max_num_batched_tokens = max(
            self.max_model_len, DEFAULT_MAX_NUM_BATCHED_TOKENS
        )
```

So let's mark it as `init=False`. It doesn't need a default this way either.

```suggestion
    prefill_max_num_batched_tokens: int = Field(init=False)
```

vllm/config/scheduler.py
```diff
@@ -163,6 +174,13 @@ def compute_hash(self) -> str:
         hash_str = hashlib.md5(str(factors).encode(), usedforsecurity=False).hexdigest()
         return hash_str
 
+    @field_validator("prefill_max_num_batched_tokens", mode="wrap")
+    @classmethod
+    def _skip_none_validation(cls, value: Any, handler: Callable) -> Any:
+        if value is None:
+            return value
+        return handler(value)
+
```
```suggestion
```

hmellor (approved):
This LGTM overall (once the config nits have been changed).

Let's wait for a second review from someone who knows the scheduler better than I do. cc @njhill

njhill:
> > Thanks @Ther-LF I like this change but wondering why the `enable_hybrid_chunked_prefill` flag is needed at all?
> > Can we not just have `prefill_max_num_batched_tokens` default to `max_num_batched_tokens` and if you want the new behaviour you just explicitly set it higher?
> 
> Thanks for the suggestion, @njhill. To confirm: under the current chunked prefill logic, at each scheduling step we’d first check whether any decode requests are active; if yes, set this step’s `token_budget` to `max_num_batched_tokens`, otherwise set it to `prefill_max_num_batched_tokens`. This is indeed cleaner. The only minor concern is that it requires checking for decode presence every step (likely negligible), whereas a flag would confine that check to when hybrid mode is enabled.
> 
> You’re also right that setting `prefill_max_num_batched_tokens` to the max model length by default isn’t ideal—it can introduce unnecessary memory overhead and reduces flexibility. I’ll proceed with your approach and update the PR accordingly. Thanks!

Sounds good... yes I guess the validation that would make sense would be
```
max_num_batched_tokens <= prefill_max_num_batched_tokens <= max_model_length
```

Perhaps allow `prefill_max_num_batched_tokens == -1` => `prefill_max_num_batched_tokens = max_model_length`.

> The only minor concern is that it requires checking for decode presence every step (likely negligible)

I think this could be tracked more efficiently during the existing schedule loop rather than an independent loop over the requests.

njhill:
As well as bounding ITL, setting a batch token limit also benefits throughput for long context since we can constrain the memory headroom needed for activations meaning the kv cache can be larger.

Related to this, we also need to make sure that a larger `prefill_max_num_batched_tokens` is taken into account during initial memory profiling.

Ther-LF:
Hi @njhill, thank you for your constructive suggestions. I’ve summarized your points as follows. Could you confirm if I’ve understood correctly?

1. We add a parameter `prefill_max_num_batched_tokens`, which sets the `token_budget` to this value when there are no decode requests. Regarding its setting:

   * If not set, the default is `None`, and it behaves like `max_num_batched_tokens`.
   * If set to an integer, it should be within the range: `max_num_batched_tokens <= prefill_max_num_batched_tokens <= max_model_length`.
   * If set to `-1`, we should set it to a maximum value. For this maximum, should we use `max_model_length`, or should we base it on memory profiling, possibly selecting a value larger than `max_model_length`? If the latter, does `max_model_length` need to be adjusted accordingly?

2. For determining whether there is a decode request, we use a `bool` variable to track requests in the running list, rather than iterating over each step.

3. When performing initial memory profiling, we consider cases with larger memory consumption, like `prefill_max_num_batched_tokens`. This is currently handled in the PR by setting `GPUModelRunner`'s `max_num_tokens`, which is used in its `profile_run()` method’s `_dummy_run()` function to perform memory profiling. In the future, we can set this to `max(max_num_tokens, prefill_max_num_batched_tokens)`.

Let me know if this matches your intention!


> As well as bounding ITL, setting a batch token limit also benefits throughput for long context since we can constrain the memory headroom needed for activations meaning the kv cache can be larger.
> 
> Related to this, we also need to make sure that a larger `prefill_max_num_batched_tokens` is taken into account during initial memory profiling.

hmellor:
> So if prefill_max_num_batched_tokens is -1 (i.e., effectively unbounded) or ≥ max_model_length, that would mean we could prefill more than one max_model_length sequence concurrently.

Yes, exactly.

vllm/config/scheduler.py
```diff
@@ -37,6 +37,11 @@ class SchedulerConfig:
     This config has no static default. If left unspecified by the user, it will
     be set in `EngineArgs.create_engine_config` based on the usage context."""
 
+    prefill_max_num_batched_tokens: int = Field(init=False)
```
hmellor: `False`?

Ther-LF: Hi @hmellor,

The `prefill_max_num_batched_tokens` part hasn’t been updated yet—I only added the optimization to track whether there are decode requests. I’m currently considering the behavior when `prefill_max_num_batched_tokens` is set to `-1`. My earlier view was that `-1` implies the user wants a default that maximizes throughput, so we would set it to `max_num_reqs * max_model_length`. However, many users likely aren’t assuming a scenario where every request in `max_num_reqs` has length `max_model_length`; more realistically, there are `x` prefill requests (with `x` not very large) and `max_num_reqs - x` decode requests. Setting it directly to `max_num_reqs * max_model_length` could therefore waste a lot of memory, significantly reduce KV-cache allocation, and even risk OOM.

I’m wondering if we should instead only require `prefill_max_num_batched_tokens >= max_num_batched_tokens`, and let users control the upper bound themselves—as long as it doesn’t cause OOM.

njhill:
@Ther-LF apologies I got sidetracked and then was sick the last couple of days. I will get back to it this week.

Ther-LF:
@njhill Hi, looking at it again, I feel like the output from `log_schedule_capacity_profile` might be a bit too verbose. Do you have any thoughts on this?

Ther-LF:
@njhill Hi, hope you're feeling better!
I've resolved the conflicts and updated the branch with the latest main. Regarding my previous question about the logging verbosity (log_schedule_capacity_profile), I'm happy to adjust it if you feel it's too noisy, or we can address it in a follow-up if it's not a blocker.
Since this PR brings significant throughput improvements (2-5%) and latency reduction for chunked prefill, I’d love to get your thoughts on the final implementation when you have a moment. Thanks!

PR #26626, opened 2025-10-11
[Bugfix][Qwen3VL] fix deepstack in qwen3vl

JJJYmmm:
## Purpose

quick fix deepstack for qwen3vl. If deepstack features is not none, we should set it properly. 🤗
related to https://github.com/QwenLM/Qwen3-VL/issues/1576

@ywang96

### DIFF
--- a/vllm/model_executor/models/qwen3_vl.py
+++ b/vllm/model_executor/models/qwen3_vl.py
@@ -1701,7 +1701,7 @@ class Qwen3VLForConditionalGeneration(
             is_multimodal=is_multimodal,
         )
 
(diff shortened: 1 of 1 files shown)


DarkLight1337 (commented):
Could you elaborate how `deepstack_input_embeds is not None` doesn't imply `self.use_deepstack` already (and vice versa)?

DarkLight1337 (commented):
Can you revert the `if` condition to satisfy gemini's comment?

DarkLight1337 (approved):

PR #26247, opened 2025-10-05
Convert formatting to use `ruff` instead of `yapf` + `isort`

hmellor:
Closes #17657

This is a massive change and would be impossible to merge during the week. Any PR's merged while this was waiting for CI would cause merge conflicts which would be time consuming to solve in this PR.

Therefore, the plan to getting this PR merged quickly is as follows:

- First commit:
  - Make the changes to the pre-commit
  - Run `pre-commit run -a` and note all the things `ruff` couldn't automatically fix
  - Add these as `pre-file-ignores` to `pyproject.toml`
- Second commit:
  - Run `pre-commit run -a`
  - `ruff` and `ruff format` should pass immediately

Then, these ignores will be systematically removed in smaller more manageable PRs afterwards.

PR #26633, opened 2025-10-11
Update `Optional[x]` -> `x | None` and `Union[x, y]` to `x | y`

hmellor:
Removes all `UP` rule ignores from `pyproject.toml` now that we no longer support Python 3.9 https://github.com/vllm-project/vllm/pull/26247.

This will be the last repo-wide formatting PR for a long time.

Thank you everyone for persevering with these changes. They will be good for the long-term health of the codebase.


DarkLight1337 (approved):

youkaichao (changes requested):
can we rethink the necessity of this change?

we have many downstream users, who will not easily migrate the python version they use, and such PR will cause great pain for them to rebase / merge new code from main, which badly hurts vllm adoption.

youkaichao (approved):
okay, given that there’s already #26247 landed, let’s get this in as well then. reverting would be too complicated.

Issue #26366, opened 2025-10-07
[Feature]: Improve config validation using Pydantic


hmellor:
### 🚀 The feature, motivation and pitch

All the configs in https://github.com/vllm-project/vllm/tree/main/vllm/config are decorated with `pydantic.dataclass`. However, many of them include manual validation in `__post_init__`.

This issue is a call for help to improve the configs in the following ways:

- Use [`pydantic.Field`](https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.Field) to apply simple constraints, such as `gt=0`, directly to the default values of the dataclass fields ([usage docs](https://docs.pydantic.dev/latest/concepts/fields/))
- Use [`pydantic.field_validator`](https://docs.pydantic.dev/latest/api/functional_validators/#pydantic.functional_validators.field_validator) to apply more complex field validation to individual fields. ([usage docs](https://docs.pydantic.dev/latest/concepts/validators/#field-validators))
- Use [`pydantic.model_validator`](https://docs.pydantic.dev/latest/api/functional_validators/#pydantic.functional_validators.model_validator) to apply more complex field validation to interdependent fields ([usage docs](https://docs.pydantic.dev/latest/concepts/validators/#model-validators))
- Consolidate all validation (that does not depend on external factors) to happen in these three places

---

Pull requests:

- [x] #26390
- [x] #26413 
- [ ] #26415
- [x] #26417
- [ ] #26519
- [x] #26629
- [ ] #26637
- [x] #26726
- [ ] #27156
- [ ] compilation - WIP, no PR
- [ ] model - WIP, no PR
- [x] kv_events? not needed
- [x] pooler? not needed
- [x] speech_to_text? not needed

vrdn-23:
I'd be happy to help with some of this @hmellor

robert-za:
I'd like to work on this one. Could you please assign me?

hmellor:
Thanks!

Just reference this issue with "Part of #26366" and I'll get your PRs added to the list in the description.

@vrdn-23's PR is a good reference for how the expected changes should look.

hmellor:
For future readers, all configs are now accounted for. There is no more work to pick up for this issue.

Navya1707:
@hmellor - thanks for assigning it to me, I've raised a PR for speculative.py - https://github.com/vllm-project/vllm/pull/27156
Could you please review and let me know if any changes?

sciafri:
Hi,

[...]

PR #26637, opened 2025-10-11
[Feature]: Use pydantic validation in observability.py config

cern1710:

- Introduced @field_validator to `show_hidden_metrics_for_version` to ensure proper version formatting
- Introduced @field_validator to `otlp_traces_endpoint` to ensure it starts with either 'http://' or 'https://'
- Added @model_validator to validate dependency between `collect_detailed_traces` and `otlp_traces_endpoint`

## Test Result

---
<details>

</details>


vllm/config/observability.py
```diff
@@ -79,25 +80,51 @@ def compute_hash(self) -> str:
...
+            )
+        return value
+
+    @field_validator("otlp_traces_endpoint", mode="after")
+    @classmethod
+    def _validate_endpoint(cls, value: Optional[str]) -> Optional[str]:
+        if value is None:
+            return value
+        if not (value.startswith("http://") or value.startswith("https://")):
+            raise ValueError("otlp_traces_endpoint must start with http:// or https://")
```
hmellor: This is not strictly true, we support tracing using `grpc` too. Let's leave this one as is.

```suggestion
```

vllm/config/observability.py
```diff
@@ -79,25 +80,51 @@ def compute_hash(self) -> str:
...
 
         from vllm.tracing import is_otel_available, otel_import_error_traceback
 
-        if not is_otel_available() and self.otlp_traces_endpoint is not None:
+        if not is_otel_available():
```
hmellor: This will force people to have `opentelemetry` installed
```suggestion
        if not is_otel_available() and self.otlp_traces_endpoint is not None:
```

vllm/config/observability.py
```diff
@@ -79,25 +80,51 @@ def compute_hash(self) -> str:
...
+        return value
+
+    @field_validator("collect_detailed_traces", mode="before")
+    @classmethod
+    def _parse_collect_traces(cls, value: Any) -> Optional[list[DetailedTraceModules]]:
+        if value is None:
+            return None
+
+        assert isinstance(value, list)
+        return [v.strip() for item in value for v in str(item).split(",") if v.strip()]
```
hmellor: This assumes every element in the list is a comma separated string, which it is not.

hmellor (changes requested):

vllm/config/observability.py
```diff
@@ -79,13 +80,57 @@ def compute_hash(self) -> str:
...
-    def __post_init__(self):
-        if (
-            self.collect_detailed_traces is not None
-            and len(self.collect_detailed_traces) == 1
-            and "," in self.collect_detailed_traces[0]
-        ):
-            self._parse_collect_detailed_traces()
+    @field_validator("show_hidden_metrics_for_version", mode="before")
+    @classmethod
+    def _normalize_version(cls, value: str | None) -> str | None:
```
hmellor:
```suggestion
    def _validate_version(cls, value: str | None) -> str | None:
```

vllm/config/observability.py
```diff
@@ -79,13 +80,57 @@ def compute_hash(self) -> str:
...
+        parts = value.split(".")
+        if len(parts) not in (2, 3) or not all(p.isdigit() for p in parts):
+            raise ValueError(
+                "show_hidden_metrics_for_version must look like '0.7' or '0.7.0'"
+            )
+        return value
+
+    @field_validator("collect_detailed_traces", mode="before")
+    @classmethod
+    def _parse_collect_traces(cls, value: Any) -> list[DetailedTraceModules] | None:
```
hmellor:
```suggestion
    def _validate_collect_detailed_traces(cls, value: Any) -> list[DetailedTraceModules] | None:
```

vllm/config/observability.py
```diff
@@ -79,13 +80,57 @@ def compute_hash(self) -> str:
...
+        out: list[str] = []
+        seen: set[str] = set()
+        for item in items:
+            if item and item not in seen:
+                seen.add(item)
+                out.append(item)
+
+        if not out:
+            return None
+        return cast(list[DetailedTraceModules], out)
```
hmellor: This still isn't quite right. You can change it to an `after` validator because the input will always pass pydantic's built in validation. We just need a post processing step. 

`collect_detailed_traces` will always be a `list[str]` or `None`. The special case is if the list only has one element, and that element is a comma separated list. This is why we split the 0th element in the original implementation.

hmellor (commented):
Thanks for the PR, can you:

- move `_parse_collect_detailed_traces` into the if block and make that an after field validator
- make the rest of `__post_init__` into an after field validator for otlp_traces_endpoint

vllm/config/observability.py
cern1710: Okay, on it!

hmellor:
Thank you again for the PR, I've pushed a commit that simplifies the logic of the new validators you wrote.

It's not worth writing a fancy validator for the comma separated string use case because it's legacy and I hope to deprecate some day.

hmellor (approved):

PR #26502, opened 2025-10-09
[FrontEnd] UNREVERT CompilationConfig overhaul (#20283): deprecate use_inductor in favor of backend, simplify custom_ops 

morrison-turnansky:
Original PR: #26113
Revert of original PR: #26472

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #26113, opened 2025-10-02
[Frontend] CompilationConfig overhaul (#20283): deprecate use_inductor in favor of backend, simplify custom_ops

morrison-turnansky:
Reverted in #26472, unrevert in #26502.

## Purpose
update custom_ops, backend, and use_inductor refactor as discussed in https://github.com/vllm-project/vllm/issues/20283#event-20038024543
## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #25845, opened 2025-09-28
[torch.compile] Make inductor partition rules respect splitting_ops #25691

baonudesifeizhai:
### Key Changes

- **Preserve user-specified `splitting_ops`**: When `use_inductor_graph_partition=True`, user-provided `splitting_ops` are now preserved in `_user_specified_splitting_ops` instead of being cleared
- **Dynamic partition rules**: Implement `_setup_dynamic_partition_rules()` using PyTorch 2.9+'s `register_should_partition_rule` API to register custom partition points
- **Robust fallback mechanism**: If dynamic rule registration fails for any user-specified operation, the system falls back to traditional splitting behavior with preserved `splitting_ops`
- **Comprehensive logging**: Add detailed debug logging to track partition rule setup, registration success/failure, and fallback behavior
- **PyTorch version compatibility**: Update requirements to support PyTorch 2.10.0.dev20250927+cu128 for the new partition rule API

### Technical Implementation

1. **Operation mapping**: Map user-provided `splitting_ops` (including aliases like "flash_attention") to `torch._ops.OpOverload` objects
2. **Rule registration**: Use `register_should_partition_rule(op_overload, partition_function)` for each resolved operation
3. **Duplicate prevention**: Track registered overloads globally to prevent duplicate registrations
4. **Alias resolution**: Handle common operation aliases (e.g., "flash_attention" → "vllm.unified_attention")

### Unit Tests
- Updated `test_splitting_ops_dynamic()` to verify `splitting_ops` preservation behavior
- Added comprehensive test coverage for dynamic partition rule setup
- Updated existing compilation tests to reflect new default behavior

### Integration Tests
- Test with various `splitting_ops` configurations:
  ```bash
  python -m vllm.entrypoints.openai.api_server \
    --model Qwen/Qwen2.5-7B-Instruct \
    --compilation-config '{"use_inductor_graph_partition": true, "splitting_ops": ["flash_attention", "addmm", "aten.bmm.default"]}' \
    --host 0.0.0.0 --port 8000
  ```
- Verify fallback behavior when PyTorch version doesn't support dynamic rules
- Test with mixed resolvable/unresolvable operations

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
[...]

PR #26645, opened 2025-10-11
[compile] Fix inductor partition config

angelayi:
When trying to run the following cmd, I ran into [this assertion error](https://github.com/baonudesifeizhai/vllm/blob/5bc26c438d096bac6148b49f5aec7ac7736d72d0/vllm/v1/cudagraph_dispatcher.py#L43-L54) because [self.backend](https://github.com/baonudesifeizhai/vllm/blob/5bc26c438d096bac6148b49f5aec7ac7736d72d0/vllm/config/compilation.py#L713) is equal to `""`.

```
vllm bench latency \
    --model=RedHatAI/Meta-Llama-3.1-8B-Instruct-FP8 \
    --output-len 1 --input-len 8192 --batch-size 1 \
    --tensor-parallel-size 8 --load-format dummy \
    --num_iters_warmup 5 --num_iters 15 \
    -O '{"level": 3, "pass_config": {"enable_async_tp": true, "enable_sequence_parallelism": true}, "use_inductor_graph_partition": true, "custom_ops":["+quant_fp8"], "cudagraph_mode":"FULL_AND_PIECEWISE"}' \
    --no-enable-prefix-caching
```

cc @ProExpertProg @zou3519 @BoyuanFeng @baonudesifeizhai

### DIFF
--- a/vllm/config/compilation.py
+++ b/vllm/config/compilation.py
@@ -709,9 +709,7 @@ class CompilationConfig:
             return self.level == CompilationLevel.PIECEWISE
 
         # Inductor partition case
-        return (
-            self.level > CompilationLevel.NO_COMPILATION and self.backend == "inductor"
-        )
+        return self.level > CompilationLevel.NO_COMPILATION and self.use_inductor
 
     def custom_op_log_check(self):
         """


ProExpertProg (approved):
Yep, bad merge of #25845 after #26113 was reverted in #26472. @morrison-turnansky can you fix in #26502?

Issue #26606, opened 2025-10-10
[Performance]: Lite-Profiler Framework for Model Analysis


namanlalitnyu:
### Existing Implementation
Currently, VLLM offers a few built-in tools to help identify performance improvement opportunities:

- torch.profiler: This tool provides detailed callstacks and tracks interactions across both CPU and GPU, making it useful for capturing major system bottlenecks. However:
   - Performance Overhead: The profiler adds noticeable overhead, so the measured elapsed time for each component may differ from actual runtime.
   - Limited Scope: To keep profiling files manageable, we often reduce profiling duration, which means torch.profiler alone may not capture the full system picture or catch performance outliers (such as sudden or periodic regressions).

- Client-side benchmark data: Benchmark scripts provide a high-level overview of system performance, including metrics like Request/Token Throughput, TTFT (Time to First Token), and TPOT (Time per Output Token). However, these topline metrics can be affected by client/server interactions and do not pinpoint specific bottlenecks for optimization.

### Proposal
To address these gaps, we propose a simple, low-cost method to capture the overhead of critical system components. This framework is controlled by the environment variable `VLLM_LITE_PROFILER_LOG_PATH`, which specifies where logs are stored for later aggregation.

When enabled, the framework uses the existing `record_function_or_null_context()` to wrap key function calls. For each call, a `LiteScope` object records the duration. Right now, we are dumping the log to the file, after each file write. But, we plan to improve this performance in the future.

After logs are generated, a script aggregates them into a summary, showing the time spent in each function. This helps us identify areas for optimization.

The breakdown includes two types of metrics:
- `Top Line Metrics`: Time spent in major stages such as “Input:Process”, “Step:Schedule”, “Step:Model”, and “Step:Output”.
- `Detailed Model Metrics`: Within the worker process, we track phases like “UpdateState”, “PrepareInput”, “Forward”, “Postprocess”, “Sample”, “Bookkeep”, “EPLB”, and “Draft”.

<img width="1041" height="160" alt="Image" src="https://github.com/user-attachments/assets/47b3d871-67a5-4c6d-b3ff-347d2166bb04" />

### Action Items:
[ ] The framework is implemented in this PR ([26648](https://github.com/vllm-project/vllm/pull/26648)), with a detailed test plan and outputs from different models (TP1 and TP8) to ensure consistent behavior across multi-processing.
[ ] Add documentation for lite-profiling in the vLLM Documentation Page.
[ ] Add profiling support for Bench serving as well.
[...]

PR #26648, opened 2025-10-11
[Core] Lite weight profiler implementation

namanlalitnyu:
## Purpose
In this PR, we are implementing a lite-weight profiling framework to capture the overhead of critical system components. This framework is controlled by the environment variable `VLLM_LITE_PROFILER_LOG_PATH`, which specifies where the logs are going to be stored, The logs will then be used to generate a summary representing the time spent by each function. This helps us identify areas for optimization.

Explained about the implementation with more details in the issue here: [26606](https://github.com/vllm-project/vllm/issues/26606)

The final generated output looks like this:
Screenshot 2025-10-15 at 2 00 44 PM

TODO: 
1. Need to implement support for Bench serving command as well, making sure that there is no data leakage.
2. Try to optimize the logging approach.

## Test Plan

1. Verify that there is no regression when the environment variable is disabled (not set). There should be no performance overhead  in the benchmark results like throughput and latency.
3. Verify that there is no impact on the eval with the new changes.
4. Verify that the profiler report is getting generated correctly once lite-profiling is enabled.

## Test Result

The framework is tested on 3 different LLM Models, with different configurations: `facebook/opt-125m`, `meta-llama/Meta-Llama-3.1-8B`, and `meta-llama/Llama-4-Maverick-17B-128E`. 

1. **Benchmark Throughput** tests on three different scenarios (main-branch, vllm_lite_profiler branch with profiling disabled, and vllm_lite_profiler branch with profiling enabled). Did 5 runs and took the average.
    -> From results, we observed that there is no performance regression when profiling is disabled. Moreover, when profiling is enabled, the performance overhead is very minimal, leading for us to easily identify the possible time consuming sections in the code.
   
    Screenshot 2025-10-15 at 1 57 26 PM

   -> Commands for different models, with configurations:
   1.  meta-llama/Meta-Llama-3.1-8B: `VLLM_LITE_PROFILER_LOG_PATH=/tmp/vllm-lite-profiler-llama3-8b.log vllm bench throughput --backend vllm --model meta-llama/Meta-Llama-3.1-8B-Instruct --dataset-name sharegpt --dataset-path ~/ShareGPT_V3_unfiltered_cleaned_split.json --num-prompts 200`
   2. Facebook/opt-125m: `VLLM_LITE_PROFILER_LOG_PATH=/tmp/vllm-lite-profiler-facebook-125m.log vllm bench throughput --backend vllm --model facebook/opt-125m --dataset-name sharegpt --dataset-path ~/ShareGPT_V3_unfiltered_cleaned_split.json --num-prompts 200`
   3. meta-llama/Llama-4-Maverick-17B-128E: `VLLM_LITE_PROFILER_LOG_PATH=/tmp/vllm-lite-profiler-llama4.log vllm bench throughput --model meta-llama/Llama-4-Maverick-17B-128E-Instruct-FP8 --tensor-parallel-size 8 --load-format dummy --dataset ~/ShareGPT_V3_unfiltered_cleaned_split.json --num-prompts 200 --backend vllm --max-model-len 8192`
  
3. **Eval testing**
    (Results are the same, no change) 
    `lm_eval --model vllm --model_args pretrained=meta-llama/Meta-Llama-3.1-8B-Instruct,tensor_parallel_size=1 --tasks gsm8k  --num_fewshot 5 --batch_size auto --limit 250` 

a) vllm_lite_profiler branch
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.768|±  |0.0268|
|     |       |strict-match    |     5|exact_match|↑  |0.740|±  |0.0278|

b) Main branch
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.768|±  |0.0268|
|     |       |strict-match    |     5|exact_match|↑  |0.740|±  |0.0278|

---
<details>

</details>

### DIFF
--- a/docs/contributing/profiling.md
+++ b/docs/contributing/profiling.md
@@ -235,3 +235,41 @@ Leverage VLLM_GC_DEBUG environment variable to debug GC costs.
 - VLLM_GC_DEBUG=1: enable GC debugger with gc.collect elpased times
 - VLLM_GC_DEBUG='{"top_objects":5}': enable GC debugger to log top 5
   collected objects for each gc.collect
+
+## Lite Profiler
+
+The lite profiler is a lightweight, minimal-overhead profiling tool designed for capturing the time spent by the critical components of the system.
+
+### How It Works
+
+The lite profiler uses context managers to time code blocks and writes timing data to a log file with minimal overhead (~50KB buffer). After execution completes, it generates an aggregate summary report showing time spent in different pipeline stages.
+
+### Environment Variable
+
+Enable lite profiling by setting the `VLLM_LITE_PROFILER_LOG_PATH` environment variable to the desired log file path:
+
+```bash
+export VLLM_LITE_PROFILER_LOG_PATH=/tmp/vllm_lite_profile.log
+```
+
+### Example Commands and Usage
+
+#### Throughput Benchmark
+
+Profile throughput with the lite profiler:
+
+```bash
+VLLM_LITE_PROFILER_LOG_PATH=/tmp/vllm_lite_profile.log \
+vllm bench throughput \
+    --model meta-llama/Llama-3.1-8B-Instruct \
+    --dataset-name random \
+    --num-prompts 1000 \
+    --input-len 128 \
+    --output-len 128
+```
+
+The profiler will automatically generate a summary report at the end showing time breakdowns for:
+
+- **Top-level pipeline events**: Schedule, Model execution, Output processing
+- **Model events**: State updates, input preparation, forward pass, postprocessing, bookkeeping
+```
--- a/docs/contributing/profiling.md
+++ b/docs/contributing/profiling.md
@@ -242,7 +242,7 @@ The lite profiler is a lightweight, minimal-overhead profiling tool designed for
 
 ### How It Works
 
-The lite profiler uses context managers to time code blocks and writes timing data to a log file with minimal overhead (~50KB buffer). After execution completes, it generates an aggregate summary report showing time spent in different pipeline stages.
+The lite profiler uses context managers to time code blocks and writes timing data to a log file with minimal overhead. After execution completes, it uses the logs from the log file to generate an aggregate summary report showing time spent in different pipeline stages. Currently, its enabled for both the `vllm bench throughput` and `vllm bench latency` commands. 
 
 ### Environment Variable
 
--- a/docs/contributing/profiling.md
+++ b/docs/contributing/profiling.md
@@ -242,7 +242,7 @@ The lite profiler is a lightweight, minimal-overhead profiling tool designed for
 
 ### How It Works
 
-The lite profiler uses context managers to time code blocks and writes timing data to a log file with minimal overhead. After execution completes, it uses the logs from the log file to generate an aggregate summary report showing time spent in different pipeline stages. Currently, its enabled for both the `vllm bench throughput` and `vllm bench latency` commands. 
+The lite profiler uses context managers to time code blocks and writes timing data to a log file with minimal overhead. After execution completes, it uses the logs from the log file to generate an aggregated summary report showing time spent in different execution stages. Currently, it's enabled for both the `vllm bench throughput` and `vllm bench latency` commands. 
 
 ### Environment Variable
 
@@ -270,6 +270,6 @@ vllm bench throughput \
 
 The profiler will automatically generate a summary report at the end showing time breakdowns for:
 
-- **Top-level pipeline events**: Schedule, Model execution, Output processing
-- **Model events**: State updates, input preparation, forward pass, postprocessing, bookkeeping
+- **Top-level pipeline events**: schedule, model execution, output processing
+- **Model events**: state updates, input preparation, forward pass, post-processing, bookkeeping
 ```
--- a/vllm/envs.py
+++ b/vllm/envs.py
@@ -1387,8 +1387,10 @@ environment_variables: dict[str, Callable[[], Any]] = {
     ),
     # Log path for the lightweight timing profiler.
     # If this path is set (not None), lightweight profiling will be enabled,
-    # providing detailed analysis of the execution time for each function call.
-    "VLLM_LITE_PROFILER_LOG_PATH": lambda: os.getenv("VLLM_LITE_PROFILER_LOG_PATH"),
+    # providing detailed execution latency breakdown by segment.
+    "VLLM_LITE_PROFILER_LOG_PATH": lambda: os.getenv(
+        "VLLM_LITE_PROFILER_LOG_PATH", None
+    ),
 }
 
 # --8<-- [end:env-vars-definition]
--- a/vllm/envs.py
+++ b/vllm/envs.py
@@ -214,6 +214,7 @@ if TYPE_CHECKING:
     VLLM_USE_FBGEMM: bool = False
     VLLM_GC_DEBUG: str = ""
     VLLM_DISABLE_SHARED_EXPERTS_STREAM: bool = False
+    VLLM_LITE_PROFILER_LOG_PATH: str | None = None
 
 
 def get_default_cache_root():
@@ -1384,6 +1385,10 @@ environment_variables: dict[str, Callable[[], Any]] = {
     "VLLM_DISABLE_SHARED_EXPERTS_STREAM": lambda: os.getenv(
         "VLLM_DISABLE_SHARED_EXPERTS_STREAM", False
     ),
+    # Log path for the lightweight timing profiler.
+    # If this path is set (not None), lightweight profiling will be enabled,
+    # providing detailed analysis of the execution time for each function call.
+    "VLLM_LITE_PROFILER_LOG_PATH": lambda: os.getenv("VLLM_LITE_PROFILER_LOG_PATH"),
 }
 
 # --8<-- [end:env-vars-definition]
--- a/vllm/envs.py
+++ b/vllm/envs.py
@@ -1388,9 +1388,7 @@ environment_variables: dict[str, Callable[[], Any]] = {
     # Log path for the lightweight timing profiler.
     # If this path is set (not None), lightweight profiling will be enabled,
     # providing detailed execution latency breakdown by segment.
-    "VLLM_LITE_PROFILER_LOG_PATH": lambda: os.getenv(
-        "VLLM_LITE_PROFILER_LOG_PATH", None
-    ),
+    "VLLM_LITE_PROFILER_LOG_PATH": lambda: os.getenv("VLLM_LITE_PROFILER_LOG_PATH"),
 }
 
 # --8<-- [end:env-vars-definition]
--- a/vllm/utils/lite_profiler.py
+++ b/vllm/utils/lite_profiler.py
@@ -1,0 +1,144 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""Minimal helpers for opt-in lightweight timing collection."""
+
+from __future__ import annotations
+
+import atexit
+import multiprocessing
+import os
+import time
+from contextlib import suppress
+from types import TracebackType
+from typing import TextIO
+
+import vllm.envs as envs
+from vllm.logger import init_logger
+
+logger = init_logger(__name__)
+
+
+def _should_log_results() -> bool:
+    """Check if the current process should log results.
+    Only the data-parallel rank 0 engine core and worker 0 should emit logs in
+    multi-process deployments so that we avoid duplicating identical timing
+    data.
+    """
+    process = multiprocessing.current_process()
+    return process.name in ("EngineCore_DP0", "VllmWorker-0")
+
+
+# Cache for log file handle
+_log_file: TextIO | None = None
+log_results = _should_log_results()
+
+
+def _write_log_entry(name: str, elapsed_us: int) -> None:
+    """Write a profiler entry using cached file handle for optimal performance.
+
+    This function implements an efficient caching approach where the file handle
+    is opened once and reused for all subsequent writes. This eliminates the
+    significant overhead of opening/closing files for every profiler entry,
+    which is crucial for maintaining the lightweight nature of the profiler.
+
+    The cached file handle is automatically closed on program exit via atexit.
+    """
+    global _log_file
+    _LOG_PATH = envs.VLLM_LITE_PROFILER_LOG_PATH
+
+    if not log_results or _LOG_PATH is None:
+        return
+
+    # Handle case where file handle was opened in parent but we're in the
+    # child process. The file descriptor may become invalid after fork
+    if _log_file is not None:
+        try:
+            # Verify if the file handle is still valid
+            _log_file.tell()
+        except (OSError, ValueError):
+            # File handle is stale, clear and reopen
+            _log_file = None
+
+    # Write the log entry
+    log_line = f"{name}|{elapsed_us}\n"
+    if _log_file is None:
+        directory = os.path.dirname(_LOG_PATH)
+        if directory:
+            os.makedirs(directory, exist_ok=True)
+        # ruff: noqa: SIM115 - intentionally keeping file handle cached globally
+        _log_file = open(_LOG_PATH, "a", buffering=50000)
+        atexit.register(_log_file.close)
+
+    _log_file.write(log_line)
+
+
+class LiteScope:
+    """Lightweight context manager for timing code blocks with minimal overhead.
+
+    This class provides a simple way to measure and log the execution time of
+    code blocks using Python's context manager protocol (with statement). It's
+    designed for high-frequency profiling with minimal performance impact.
+    """
+
+    def __init__(self, name: str) -> None:
+        self._name = name
+        self._start_time: int | None = None
+
+    def __enter__(self) -> None:
+        self._start_time = time.perf_counter_ns()
+
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_value: BaseException | None,
+        traceback: TracebackType | None,
+    ) -> bool:
+        if self._start_time is not None and exc_type is None:
+            elapsed_ns = time.perf_counter_ns() - self._start_time
+            # Use integer microseconds for better performance
+            elapsed_us = elapsed_ns // 1000
+            _write_log_entry(self._name, elapsed_us)
+        return False
+
+
+def maybe_emit_lite_profiler_report() -> None:
+    """Generate and display a summary report of profiling data if available.
+
+    This function serves as the main entry point for analyzing and displaying
+    profiling results. It checks if profiling was enabled and a log file exists,
+    then delegates to the lite_profiler_report module to generate statistics
+    like function call counts, timing distributions, and performance insights.
+    """
+
+    log_path = envs.VLLM_LITE_PROFILER_LOG_PATH
+    if log_path is None:
+        return
+
+    if not os.path.exists(log_path):
+        logger.warning(
+            "Lite profiler log not found. Ensure the profiled process sets "
+            "the expected path."
+        )
+        return
+
+    try:
+        from vllm.utils import lite_profiler_report
+    except Exception as exc:  # pragma: no cover - import error should not crash
+        logger.error("Failed to import lite profiler report helper: %s", exc)
+        return
+
+    logger.info("")
+    logger.info("Lite profiler summary (%s):", log_path)
+    try:
+        # Generate and display the summary report
+        lite_profiler_report.summarize_log(log_path)
+
+        # Clear the log file to avoid accumulating data from multiple runs
+        with suppress(OSError):
+            directory = os.path.dirname(log_path)
+            if directory:
+                os.makedirs(directory, exist_ok=True)
+            with open(log_path, "w"):
+                pass
+    except Exception as exc:  # pragma: no cover - avoid crashing benchmarks
+        logger.error("Failed to summarize lite profiler log %s: %s", log_path, exc)
--- a/vllm/utils/lite_profiler.py
+++ b/vllm/utils/lite_profiler.py
@@ -8,7 +8,6 @@ import atexit
 import multiprocessing
 import os
 import time
-from contextlib import suppress
 from types import TracebackType
 from typing import TextIO
 
@@ -30,7 +29,6 @@ def _should_log_results() -> bool:
 
 # Cache for log file handle
 _log_file: TextIO | None = None
-log_results = _should_log_results()
 
 
 def _write_log_entry(name: str, elapsed_us: int) -> None:
@@ -46,7 +44,7 @@ def _write_log_entry(name: str, elapsed_us: int) -> None:
     global _log_file
     _LOG_PATH = envs.VLLM_LITE_PROFILER_LOG_PATH
 
-    if not log_results or _LOG_PATH is None:
+    if not _should_log_results() or _LOG_PATH is None:
         return
 
     # Handle case where file handle was opened in parent but we're in the
@@ -66,7 +64,7 @@ def _write_log_entry(name: str, elapsed_us: int) -> None:
         if directory:
             os.makedirs(directory, exist_ok=True)
         # ruff: noqa: SIM115 - intentionally keeping file handle cached globally
-        _log_file = open(_LOG_PATH, "a", buffering=50000)
+        _log_file = open(_LOG_PATH, "w", buffering=50000)
         atexit.register(_log_file.close)
 
     _log_file.write(log_line)
@@ -92,13 +90,12 @@ class LiteScope:
         exc_type: type[BaseException] | None,
         exc_value: BaseException | None,
         traceback: TracebackType | None,
-    ) -> bool:
+    ) -> None:
         if self._start_time is not None and exc_type is None:
             elapsed_ns = time.perf_counter_ns() - self._start_time
             # Use integer microseconds for better performance
             elapsed_us = elapsed_ns // 1000
             _write_log_entry(self._name, elapsed_us)
-        return False
 
 
 def maybe_emit_lite_profiler_report() -> None:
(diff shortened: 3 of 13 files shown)


vllm/utils/lite_profiler.py
```diff
@@ -0,0 +1,144 @@
...
+    multi-process deployments so that we avoid duplicating identical timing
+    data.
+    """
+    process = multiprocessing.current_process()
+    return process.name in ("EngineCore_DP0", "VllmWorker-0")
+
+
+# Cache for log file handle
+_log_file: TextIO | None = None
+log_results = _should_log_results()
```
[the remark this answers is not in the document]
namanlalitnyu: Verified that the worker processes are correctly setting  the value for logging.

vllm/envs.py
```diff
@@ -1391,6 +1392,12 @@ def get_vllm_port() -> Optional[int]:
     # - VLLM_GC_DEBUG='{"top_objects":5}': enable GC debugger with
     #                                      top 5 collected objects
     "VLLM_GC_DEBUG": lambda: os.getenv("VLLM_GC_DEBUG", ""),
+    # Log path for the lightweight timing profiler.
+    # If this path is set (not None), lightweight profiling will be enabled,
+    # providing detailed analysis of the execution time for each function call.
```
linzebing: Nit: 
```suggestion
    # providing detailed execution latency breakdown by segment.
```

vllm/utils/lite_profiler_report.py
```diff
@@ -0,0 +1,121 @@
...
+from collections import defaultdict
+from collections.abc import Iterable, Sequence
+
+from vllm.logger import init_logger
+
+logger = init_logger(__name__)
+
+
+def _extract_event_us(filename: str) -> dict[str, list[int]]:
+    """Collect the microsecond timings for every scope in ``filenames``."""
```
linzebing:
```suggestion
    """Collect the microsecond timings for every scope in ``filename``."""
```

vllm/utils/lite_profiler.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def __enter__(self) -> None:
+        self._start_time = time.perf_counter_ns()
+
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_value: BaseException | None,
+        traceback: TracebackType | None,
+    ) -> bool:
```
linzebing: bool or None?

vllm/utils/lite_profiler.py
```diff
@@ -0,0 +1,144 @@
...
+            "Lite profiler log not found. Ensure the profiled process sets "
+            "the expected path."
+        )
+        return
+
+    try:
+        from vllm.utils import lite_profiler_report
+    except Exception as exc:  # pragma: no cover - import error should not crash
+        logger.error("Failed to import lite profiler report helper: %s", exc)
+        return
```
linzebing: When will this happen?

vllm/utils/lite_profiler.py
```diff
@@ -0,0 +1,144 @@
...
+        # Generate and display the summary report
+        lite_profiler_report.summarize_log(log_path)
+
+        # Clear the log file to avoid accumulating data from multiple runs
+        with suppress(OSError):
+            directory = os.path.dirname(log_path)
+            if directory:
+                os.makedirs(directory, exist_ok=True)
+            with open(log_path, "w"):
+                pass
```
linzebing: I think we log file clear is not needed, but change the "append" mode to "overwrite" mode above

vllm/utils/lite_profiler.py
```diff
@@ -0,0 +1,144 @@
...
+            _log_file = None
+
+    # Write the log entry
+    log_line = f"{name}|{elapsed_us}\n"
+    if _log_file is None:
+        directory = os.path.dirname(_LOG_PATH)
+        if directory:
+            os.makedirs(directory, exist_ok=True)
+        # ruff: noqa: SIM115 - intentionally keeping file handle cached globally
+        _log_file = open(_LOG_PATH, "a", buffering=50000)
```
linzebing: I think "overwrite" is better than "append".

vllm/utils/lite_profiler.py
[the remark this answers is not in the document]
linzebing: I have the same question as codex. It does feel we should check inside `_write_log_entry`. How did this work at import time?

vllm/utils/lite_profiler_report.py
```diff
@@ -0,0 +1,121 @@
...
+    event_us_sum: dict[str, int],
+    events: Sequence[str],
+) -> list[str]:
+    total_us = sum(event_us_sum.get(event, 0) for event in events)
+    cells = []
+    for event in events:
+        cells.append(_format_duration_us(event_us_sum.get(event, 0), total_us))
+    # Convert microseconds to seconds
+    total_seconds = total_us / 1_000_000 if total_us else 0.0
+    cells.append(f"{total_seconds:.2f}s")
```
linzebing: Maybe 3 digits or use milliseconds

namanlalitnyu: I saw that for bigger models, the time in seconds was a bit large enough, so if we replace that with milliseconds, then it wasn't looking that good. Updated it with 3 digits. 
But, can update it to milliseconds, if that is more suited, and helpful.

vllm/utils/lite_profiler.py
namanlalitnyu: Thanks for the suggestion, verified that once we put this in write mode, then we won't have to explicitly clear the file, at the end of the benchmarking.

vllm/utils/lite_profiler.py
namanlalitnyu: This was just an extra cautious check, but this is actually not needed, as we would be able to import the file. Updated the changes with its removal.

vllm/utils/lite_profiler.py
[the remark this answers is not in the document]
namanlalitnyu: Moved the check inside the function.

vllm/utils/lite_profiler.py
```diff
@@ -0,0 +1,129 @@
...
+    is opened once and reused for all subsequent writes. This eliminates the
+    significant overhead of opening/closing files for every profiler entry,
+    which is crucial for maintaining the lightweight nature of the profiler.
+
+    The cached file handle is automatically closed on program exit via atexit.
+    """
+    global _log_file
+    _LOG_PATH = envs.VLLM_LITE_PROFILER_LOG_PATH
+
+    if not _should_log_results() or _LOG_PATH is None:
```
linzebing: We can assert _LOG_PATH is not None?

vllm/utils/lite_profiler/lite_profiler.py
```diff
@@ -0,0 +1,129 @@
...
+
+    # Write the log entry
+    log_line = f"{name}|{elapsed_us}\n"
+    if _log_file is None:
+        directory = os.path.dirname(_LOG_PATH)
+        if directory:
+            os.makedirs(directory, exist_ok=True)
+        # ruff: noqa: SIM115 - intentionally keeping file handle cached globally
+        _log_file = open(_LOG_PATH, "w", buffering=50000)
+        atexit.register(_log_file.close)
```
linzebing, line 76: This only closes the file when Python interpreter shuts down. This makes me think whether it's possible for us to read incomplete data if we don't shut down the server (due to buffered data not being flushed yet).

This is a bit tricky. One way is to explicitly close in `maybe_emit_lite_profiler_report`, but this only works for `benchmark_latency` and `benchmark_throughput`. `benchmark serve` runs server and client separately.

If we can't find a good way, we can skip `benchmark serve` for now.

vllm/utils/lite_profiler/lite_profiler.py
```diff
@@ -0,0 +1,129 @@
...
+    Only the data-parallel rank 0 engine core and worker 0 should emit logs in
+    multi-process deployments so that we avoid duplicating identical timing
+    data.
+    """
+    process = multiprocessing.current_process()
+    return process.name in ("EngineCore_DP0", "VllmWorker-0")
+
+
+# Cache for log file handle
+_log_file: TextIO | None = None
```
Jialin, line 33: Hmmm, as we might wire logs from different processors, I'm wondering if there's anything we need to do to synchronize the write accordingly.

I might feel safer to reuse logger instead. And currently, just have a knob to control whether to enable the logging.

vllm/utils/lite_profiler.py
```diff
@@ -0,0 +1,129 @@
...
+        self,
+        exc_type: type[BaseException] | None,
+        exc_value: BaseException | None,
+        traceback: TracebackType | None,
+    ) -> None:
+        if self._start_time is not None and exc_type is None:
+            elapsed_ns = time.perf_counter_ns() - self._start_time
+            # Use integer microseconds for better performance
+            elapsed_us = elapsed_ns // 1000
+            _write_log_entry(self._name, elapsed_us)
```
Jialin: > Use integer microseconds for better performance

Do you mind to elaborate on this a bit?

Also, can we directly use elpased_ns intead?

vllm/utils/lite_profiler_report.py
```diff
@@ -0,0 +1,121 @@
...
+MODEL_EVENTS = [
+    "Model:UpdateState",
+    "Model:PrepareInput",
+    "Model:Forward",
+    "Model:Postprocess",
+    "Model:Sample",
+    "Model:Bookkeep",
+    "Model:EPLB",
+    "Model:Draft",
+]
```
Jialin, line 86: Maybe we could keep the original events first without introducing new ones in this PR.

vllm/v1/engine/core.py
```diff
@@ -822,15 +828,17 @@ def _process_input_queue(self):
...
 
         if waited:
             logger.debug("EngineCore loop active.")
 
         # Handle any more client requests.
         while not self.input_queue.empty():
             req = self.input_queue.get_nowait()
-            self._handle_client_request(*req)
+            with record_function_or_nullcontext("Input:Process"):
+                self._handle_client_request(*req)
```
Jialin: Input process is not isolated from core process (e.g. running in parallel), maybe we could leave as is and not track their performance for now.

vllm/utils/lite_profiler/lite_profiler.py
```diff
@@ -0,0 +1,129 @@
+# SPDX-License-Identifier: Apache-2.0
```
Jialin, line 1: Let's create a dedicated path utils/lite_profiler instead.

vllm/utils/lite_profiler/lite_profiler.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+def _should_log_results() -> bool:
+    """Check if the current process should log results.
+    Only the data-parallel rank 0 engine core and worker 0 should emit logs in
+    multi-process deployments so that we avoid duplicating identical timing
+    data.
+    """
+    process = multiprocessing.current_process()
+    return process.name in ("EngineCore_DP0", "VllmWorker-0")
```
Jialin, line 29: Maybe we should just let all ranks logs currently, and iterating on it later on.

The logging overhead, should be relatively small per rank, and we could leave the heavy lifting filtering to the reporting and reduce the branching and computations to runtime as much as possible.

vllm/utils/lite_profiler/lite_profiler.py
namanlalitnyu: We can try to flush the file data before going for generating the summary, but I think that needs a bit of testing for serving part.
I'll skip the serving part in this PR, and will take it up a new PR with a few more changes.

vllm/utils/lite_profiler.py
namanlalitnyu: I used integer microseconds instead of nanoseconds because smaller numbers are faster to format and write to logs, reducing the profiler's overhead. Since we ultimately convert to seconds for the final report, microsecond precision (1µs = 0.001ms) is sufficient and avoids unnecessary nanosecond granularity. I verified through throughput measurements that using microseconds provides slightly better performance compared to logging raw nanoseconds.
But, please let me know your thoughts as well.

Jialin: I'm a bit leaning toward to keep it in ns unless `ns -> us` give us a noticeable improvements. Just feel that we've pay enough cost to get a granular ns-level counter, truncate it to loss precision sounds wasting resources in the first place.

vllm/utils/lite_profiler/lite_profiler.py
```diff
@@ -111,19 +111,27 @@ def maybe_emit_lite_profiler_report() -> None:
...
 
-    from vllm.utils import lite_profiler_report
+    from vllm.utils.lite_profiler import lite_profiler_report
 
     logger.info("")
     logger.info("Lite profiler summary (%s):", log_path)
     try:
         # Generate and display the summary report
         lite_profiler_report.summarize_log(log_path)
+        os.remove(log_path)
```
namanlalitnyu, line 133: Since the file is opened in append mode while profiling to avoid having processes overwrite data, once report generation completes, we need to remove the file, so that there is no data duplication for the future runs.

vllm/utils/lite_profiler/lite_profiler.py
namanlalitnyu: Thanks, created a new folder path containing lite_profiler related files.

vllm/utils/lite_profiler.py
namanlalitnyu: Sure, I have updated the implementation to use nano_seconds.

docs/contributing/profiling.md
```diff
@@ -231,3 +231,41 @@ Leverage VLLM_GC_DEBUG environment variable to debug GC costs.
...
 - VLLM_GC_DEBUG='{"top_objects":5}': enable GC debugger to log top 5
   collected objects for each gc.collect
+
+## Lite Profiler
+
+The lite profiler is a lightweight, minimal-overhead profiling tool designed for capturing the time spent by the critical components of the system.
+
+### How It Works
+
+The lite profiler uses context managers to time code blocks and writes timing data to a log file with minimal overhead. After execution completes, it uses the logs from the log file to generate an aggregate summary report showing time spent in different pipeline stages. Currently, its enabled for both the `vllm bench throughput` and `vllm bench latency` commands.
```
linzebing:
```suggestion
The lite profiler uses context managers to time code blocks and writes timing data to a log file with minimal overhead. After execution completes, it uses the logs from the log file to generate an aggregated summary report showing time spent in different execution stages. Currently, it's enabled for both the `vllm bench throughput` and `vllm bench latency` commands. 
```

docs/contributing/profiling.md
```diff
@@ -231,3 +231,41 @@ Leverage VLLM_GC_DEBUG environment variable to debug GC costs.
...
+    --dataset-name random \
+    --num-prompts 1000 \
+    --input-len 128 \
+    --output-len 128
+```
+
+The profiler will automatically generate a summary report at the end showing time breakdowns for:
+
+- **Top-level pipeline events**: Schedule, Model execution, Output processing
+- **Model events**: State updates, input preparation, forward pass, postprocessing, bookkeeping
```
linzebing:
```suggestion
- **Top-level pipeline events**: schedule, model execution, output processing
- **Model events**: state updates, input preparation, forward pass, post-processing, bookkeeping
```

vllm/utils/lite_profiler/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+
+from vllm.utils.lite_profiler.lite_profiler import (
+    LiteScope,
+    maybe_emit_lite_profiler_report,
+)
+
+__all__ = [
+    "LiteScope",
+    "maybe_emit_lite_profiler_report",
+]
```
linzebing, line 13: For my understanding, what's the use of this?

vllm/utils/lite_profiler/lite_profiler.py
```diff
@@ -0,0 +1,140 @@
...
+    if _log_file is not None:
+        _log_file.close()
+        _log_file = None
+
+    if not os.path.exists(log_path):
+        logger.warning(
+            "Lite profiler log not found. Ensure the profiled process sets "
+            "the expected path."
+        )
+        return
```
linzebing: This should be merged with line 115 if statement?

vllm/utils/lite_profiler/lite_profiler.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # Write the log entry
+    log_line = f"{name}|{elapsed_ns}\n"
+    if _log_file is None:
+        directory = os.path.dirname(_LOG_PATH)
+        if directory:
+            os.makedirs(directory, exist_ok=True)
+        # ruff: noqa: SIM115 - intentionally keeping file handle cached globally
+        # Currently, we are flushing the file handle after every write. This
+        # is done to ensure safety so that there is no data leakage.
```
linzebing:
```suggestion
        # is done to ensure safety so that there is no data loss.
```

vllm/utils/lite_profiler/lite_profiler.py
```diff
@@ -0,0 +1,140 @@
...
+        # TODO: We can optimise this further, to ensure performance overhead
+        # can be reduced in future.
+        _log_file = open(_LOG_PATH, "a", buffering=1)
+        atexit.register(_log_file.close)
```
linzebing, line 76: Do we still need this close hook if we explicitly close in `maybe_emit_lite_profiler_report`?

vllm/utils/lite_profiler/lite_profiler.py
```diff
@@ -0,0 +1,140 @@
...
+    if _log_file is None:
+        directory = os.path.dirname(_LOG_PATH)
+        if directory:
+            os.makedirs(directory, exist_ok=True)
+        # ruff: noqa: SIM115 - intentionally keeping file handle cached globally
+        # Currently, we are flushing the file handle after every write. This
+        # is done to ensure safety so that there is no data leakage.
+        # TODO: We can optimise this further, to ensure performance overhead
+        # can be reduced in future.
+        _log_file = open(_LOG_PATH, "a", buffering=1)
```
linzebing:
```suggestion
        _log_file = open(_LOG_PATH, "a", buffering=0)
```

linzebing: Given now we use append, we need to explicit remove the file at the beginning of `maybe_emit_lite_profiler_report`?

vllm/envs.py
```diff
@@ -1391,6 +1392,12 @@ def get_vllm_port() -> Optional[int]:
     # - VLLM_GC_DEBUG='{"top_objects":5}': enable GC debugger with
     #                                      top 5 collected objects
     "VLLM_GC_DEBUG": lambda: os.getenv("VLLM_GC_DEBUG", ""),
+    # Log path for the lightweight timing profiler.
+    # If this path is set (not None), lightweight profiling will be enabled,
+    # providing detailed execution latency breakdown by segment.
+    "VLLM_LITE_PROFILER_LOG_PATH": lambda: os.getenv(
+        "VLLM_LITE_PROFILER_LOG_PATH", None
```
Jialin: default `None` should be the default behavior, maybe simply use os.getenv("VLLM_LITE_PROFILER_LOG_PATH")

vllm/utils/lite_profiler/__init__.py
Jialin: Maybe not needed?

vllm/utils/lite_profiler/lite_profiler.py
```diff
@@ -0,0 +1,140 @@
...
+from types import TracebackType
+from typing import TextIO
+
+import vllm.envs as envs
+from vllm.logger import init_logger
+
+logger = init_logger(__name__)
+
+
+def _should_log_results() -> bool:
```
Jialin, line 22: maybe we should tag it with functools.cache, as the result should not change after process got initialized properly?

vllm/utils/lite_profiler/lite_profiler.py
```diff
@@ -0,0 +1,140 @@
...
+
+    This function implements an efficient caching approach where the file handle
+    is opened once and reused for all subsequent writes. This eliminates the
+    significant overhead of opening/closing files for every profiler entry,
+    which is crucial for maintaining the lightweight nature of the profiler.
+
+    The cached file handle is automatically closed on program exit via atexit.
+    """
+    global _log_file
+    _LOG_PATH = envs.VLLM_LITE_PROFILER_LOG_PATH
```
Jialin: I think it should be safe to directly use `envs.VLLM_LITE_PROFILER_LOG_PATH`

The captialized _LOG_PATH looks like some global variable.

vllm/utils/lite_profiler/lite_profiler.py
```diff
@@ -0,0 +1,140 @@
...
+        try:
+            # Verify if the file handle is still valid
+            _log_file.flush()
+            _log_file.tell()
+        except (OSError, ValueError):
+            # File handle is stale, clear and reopen
+            _log_file = None
+
+    # Write the log entry
+    log_line = f"{name}|{elapsed_ns}\n"
```
Jialin: nit: Get rid of log_line and directly use in _log_file.write(f"{name}|{elapsed_ns}\n")

vllm/utils/lite_profiler/lite_profiler.py
```diff
@@ -0,0 +1,140 @@
...
+        # is done to ensure safety so that there is no data leakage.
+        # TODO: We can optimise this further, to ensure performance overhead
+        # can be reduced in future.
+        _log_file = open(_LOG_PATH, "a", buffering=1)
+        atexit.register(_log_file.close)
+
+    _log_file.write(log_line)
+
+
+class LiteScope:
```
Jialin: nit: maybe spell out the fullname LiteProfilerScope?

vllm/utils/lite_profiler/lite_profiler.py
```diff
@@ -0,0 +1,140 @@
...
+    def __enter__(self) -> None:
+        self._start_time = time.perf_counter_ns()
+
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_value: BaseException | None,
+        traceback: TracebackType | None,
+    ) -> None:
+        if self._start_time is not None and exc_type is None:
```
Jialin: Maybe we don't need to check exc_type?

vllm/utils/lite_profiler/lite_profiler.py
```diff
@@ -0,0 +1,140 @@
...
+    """Lightweight context manager for timing code blocks with minimal overhead.
+
+    This class provides a simple way to measure and log the execution time of
+    code blocks using Python's context manager protocol (with statement). It's
+    designed for high-frequency profiling with minimal performance impact.
+    """
+
+    def __init__(self, name: str) -> None:
+        self._name = name
+        self._start_time: int | None = None
```
Jialin: add time unit to the variable name for better readability

_start_time_ns

vllm/utils/lite_profiler/lite_profiler_report.py
```diff
@@ -0,0 +1,115 @@
...
+    "Step:Output",
+]
+
+MODEL_EVENTS = [
+    "Model:UpdateState",
+    "Model:PrepareInput",
+    "Model:Forward",
+    "Model:Postprocess",
+    "Model:Bookkeep",
+]
```
Jialin, line 86: None blocking: Should we add extract and accumulate the events automatically during line processing?

vllm/utils/lite_profiler/__init__.py
namanlalitnyu: I observed from the failed CI runs that this is something which we need to create whenever we are creating a new directory in this repo. If we don't create this __init__.py file, then the CI runs are failing. Moreover, i think its just representing the functions which are being referenced by different files from this directory.

This was the failing CI with the error: https://app.readthedocs.org/api/v2/build/29947204.txt
Screenshot 2025-10-16 at 1 41 42 PM

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2619,7 +2621,7 @@ def propose_draft_token_ids(sampled_token_ids):
             # as inputs, and does not need to wait for bookkeeping to finish.
             propose_draft_token_ids(sampler_output.sampled_token_ids)
 
-        with record_function_or_nullcontext("Bookkeep"):
+        with record_function_or_nullcontext("Model:Bookkeep"):
```
Jialin, line 2634: Hmmm, unfortunately, the current record place might not be quite accurate, but no need to address it for now.
Currently, Model:Forward, Model:Sample only capture the CPU time, but the model forward computation and sampling computation is still in-process on GPU side, until the sampled tokens d2h event happen to copy sampled tokens from GPU to CPU in 

https://github.com/vllm-project/vllm/blob/aa255ff55aecfb1a90f1359a2ead315f4d29cff7/vllm/v1/worker/gpu_model_runner.py#L2291

But just keep this in mind, as we might see high bookkeep cost vs small model/sampler costs :/

linzebing (approved):

vllm/utils/lite_profiler/lite_profiler.py
namanlalitnyu: Its just a extra safety check I added, so that even if our program crashes due to some reason, we have proper file cleanup.

vllm/v1/worker/gpu_model_runner.py
namanlalitnyu: Oh I see..thanks for mentioning about it. We can pick it up as a fix in the future.

vllm/utils/lite_profiler/lite_profiler.py
namanlalitnyu: Sure, actually this variable was being used at a couple of places in the file, so I just thought that it be easy to have a shorter variable name, instead of reusing the long name. I have updated it with the small case notation to avoid confusion.

vllm/utils/lite_profiler/lite_profiler_report.py
namanlalitnyu: Oh, sorry, I didn't fully get this, could you please elaborate on it.

Jialin: When analyzing the log file, we should be able to collect all logged events instead of specify the event here. So in the future, when people add new events to the code, they don't need to update lite_profiler_report to add the new events.

Jialin (approved):

vllm/utils/lite_profiler/lite_profiler_report.py
namanlalitnyu: Ah I see. Thank for explaining. 
Actually right now, we are collecting all the metrics in the log file which are referenced by the record_function() in the vllm code, and for different report generation components (like model and step breakdown), we are just picking up the specific events mentioned in this list.  
But, I think we can update our approach later on to being closer to the one that you mentioned.

namanlalitnyu:
@yeqcharlotte could you please review these changes, whenever you have some time.
Thanks!

Jialin:
> @yeqcharlotte could you please review these changes, whenever you have some time. Thanks!

CC @yeqcharlotte @houseroad for reviews for the liteweight profiler. @linzebing and I had finished the initial review.

PR #25135, opened 2025-09-18
Llamas 3.1 405B fp4 changes upstreaming from 355_wip

maleksan85:
Perf is the same:
upstream tp1
```
============ Serving Benchmark Result ============
Successful requests:                     320
Maximum request concurrency:             64
Benchmark duration (s):                  413.07
Total input tokens:                      326905
Total generated tokens:                  327680
Request throughput (req/s):              0.77
Output token throughput (tok/s):         793.27
Total Token throughput (tok/s):          1584.66
---------------Time to First Token----------------
Mean TTFT (ms):                          5240.27
Median TTFT (ms):                        5150.59
P99 TTFT (ms):                           13542.02
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          75.58
Median TPOT (ms):                        76.05
P99 TPOT (ms):                           79.35
---------------Inter-token Latency----------------
Mean ITL (ms):                           75.58
Median ITL (ms):                         66.27
P99 ITL (ms):                            73.04
----------------End-to-end Latency----------------
Mean E2EL (ms):                          82558.95
Median E2EL (ms):                        82994.12
P99 E2EL (ms):                           89413.31
==================================================
```

355_wip tp1
```
============ Serving Benchmark Result ============
Successful requests:                     320
Maximum request concurrency:             64
Benchmark duration (s):                  415.51
Total input tokens:                      326905
Total generated tokens:                  327680
Request throughput (req/s):              0.77
Output token throughput (tok/s):         788.62
Total Token throughput (tok/s):          1575.37
---------------Time to First Token----------------
Mean TTFT (ms):                          5262.51
Median TTFT (ms):                        5156.42
P99 TTFT (ms):                           14143.78
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          76.03
Median TPOT (ms):                        76.44
P99 TPOT (ms):                           79.86
---------------Inter-token Latency----------------
Mean ITL (ms):                           76.03
Median ITL (ms):                         66.64
P99 ITL (ms):                            69.13
----------------End-to-end Latency----------------
Mean E2EL (ms):                          83044.74
Median E2EL (ms):                        83387.81
P99 E2EL (ms):                           90413.53
==================================================
```

Command:
```
HIP_VISIBLE_DEVICES=7 \
VLLM_DISABLE_COMPILE_CACHE=1 \
USE_FASTSAFETENSOR=1 \
SAFETENSORS_FAST_GPU=1 \
VLLM_USE_V1=1 \
[...]

PR #24490, opened 2025-09-09
[RFC][ROCm][AITER] Keep all AITER kernels in `_aiter_ops` class like `_custom_ops` and `_ipex_ops`

vllmellm:
## Purpose

This PR introduces `_aiter_ops.py` as proposed in the [RFC here](https://github.com/vllm-project/vllm/issues/21504). The `aiter_ops` namespace provides several key benefits:

- Centralized kernel registration: Ensures that kernels from the aiter package are properly registered

- Environment availability checks: Encapsulates aiter support detection and environment compatibility validation

- Reduced code duplication: Eliminates the need for duplicate helper functions, namely checking device compability and environment varible enability checks across different vLLM modules.

This implementation establishes the foundation for future refactoring efforts, where existing kernels throughout the vLLM repository will be migrated to use this unified approach for better maintainability and consistency.

This PR uses [5ee37dce](https://github.com/ROCm/aiter/commit/5ee37dced6f1bde0229b2c77ce079433549aa25f) commit from `aiter` repo.

## Test Plan

Test models that are afftected by this change, using lm_eval on gsm8k dataset.

**environment setting**

Step 1: run vllm serve

`
VLLM_USE_V1=1 \
VLLM_ROCM_USE_AITER=1 \
SAFETENSORS_FAST_GPU=1 \
VLLM_DISABLE_COMPILE_CACHE=1 \
vllm serve $MODEL_NAME --compilation-config '{"cudagraph_mode": "FULL_AND_PIECEWISE", "cudagraph_capture_sizes": [1,2,4,8,16,24,32]}'  --trust-remote-code --swap-space 16 --distributed-executor-backend mp`

Step 2: run lm_eval

`lm_eval --model local-completions
--tasks gsm8k
--model_args model=$MODEL_NAME,base_url=http://localhost:8000/v1/completions
--trust_remote_code
--num_fewshot 5
--batch_size 256`

### deepseek-ai/DeepSeek-V3 -tp 8 --block-size 1  --max-model-len 32768 --max_seq_len_to_capture 32768 

|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9500|±  | 0.006|
|     |       |strict-match    |     5|exact_match|↑  |0.9507|±  | 0.006|

[...]

Issue #21504, opened 2025-07-24
[RFC] [ROCm] [AITER]: Propose a `_aiter_ops` class like `_custom_ops` and `_ipex_ops`


tjtanaa:
### Motivation.

This RFC proposes the creation of a new module, `vllm/_aiter_ops.py`, to centralize the management, conditional loading, and registration of AITER kernels for ROCm. This new module will be analogous to the existing `_custom_ops.py` and `_ipex_ops.py`, providing a single, authoritative source for all AITER-related operations. This change will improve code organization, prevent circular dependencies, simplify the developer experience, and streamline testing.

As the integration of AITER kernels into vLLM deepens, the current ad-hoc approach of importing and checking for these kernels across different parts of the codebase is becoming unsustainable. This leads to several challenges:

- **Code Scatttering & Discoverability**: AITER kernels and their enabling logic are spread across various files, such as `vllm/model_executor/layers/linear.py` (for unquantized layers) and `vllm/model_executor/layers/quantization/utils/fp8_utils.py ` (for quantized layers). This makes it difficult for developers to find where a specific kernel is implemented or to understand the full scope of AITER integration.

- **Risk of Circular Imports**: As more modules begin to depend on AITER kernels, and those kernels potentially need utilities from other modules, the risk of introducing circular import dependencies increases significantly. A centralized, low-level ops module breaks these potential dependency cycles.

- **Inconsistent Kernel Usage**: A single AITER kernel might be applicable to multiple contexts (e.g., both unquantized and FP8 quantized linear layers). Without a central location, we lack a clear pattern for defining such a kernel once and reusing it, leading to potential code duplication or complex import paths.

- **Redundant Environment Variable Checks**: The logic to enable AITER kernels, often controlled by environment variables like `VLLM_ROCM_USER_AITER_*`, is repeated in multiple places. These checks are fundamentally op-level switches, not platform-level configurations, and should be co-located with the operations they control.

- **Testing Complexity**: Aggregating all AITER operations in one place simplifies the process of writing targeted unit and integration tests. We can easily import all available ops and mock their availability without navigating a complex module hierarchy.
By creating _aiter_ops.py, we can adopt the same clean architectural pattern used for other specialized backends, making the codebase more modular, robust, and easier to maintain.

<details>
   <summary> draft </summary>

As AITER kernels are more tightly integrated, we have been observing it is starting to get harder to avoid circular imports and potentially confuses developer with import of kernels from different ops. 

E.g. 
[...]

Issue #17067, opened 2025-04-23
[RFC]: All Ops should be determined during init and wrapped in a Layer Module to avoid envs.ENVIRON overhead


tjtanaa:
### Motivation.

Accessing envs.ENVIRON has non-negligible overhead. Given that LLM models have many ops and layers. The overhead from accessing envs.ENVIRON could spike to 0.1 ~ 1ms overhead per token. I have observed a huge overhead in MLA prefill forward pass when using envs.ENVIRON in kernel selection logic (where `if-else` statement is involved).

Proposed action:
1. Layer Module is suggested to store the selected kernel ops as a property of the layer.
`@cache` is discourage due to the increasing complexity that it is causing to clear the cache as there are many properties depending on envs.
`@cache` is discouraged in several PRs review as there is a usecase as such: 
Users instantiate multiple LLMs in a single python program. Each LLM instance uses different sets of ENV variables.

2. Document the overhead issue down in vLLM documentation page under Contribution section to remind developers of the abstract and the overhead caused by envs.ENVIRON invocation.

## Overhead experiments

```
Average time per accessing envs.ENVIRON : 1.0514259338378907e-06 seconds
Average time per accessing class method access : 3.0994415283203126e-08 seconds
```

Script:
```
import time
import torch
import vllm.envs as envs


class Layer:
    @staticmethod
    def forward():
        return torch.randn(1024, 1024, dtype=torch.float16, device='cuda')

def test_envs_timing():
    
    # Time the operation
    num_runs = 100
    start_time = time.time()
    for _ in range(num_runs):
        envs.VLLM_ROCM_USE_AITER_LINEAR
    end_time = time.time()
    
    # Calculate average time per run
    avg_time = (end_time - start_time) / num_runs
    print(f"Average time per accessing envs.ENVIRON : {avg_time} seconds")


def test_class_method():
    
    # Time the operation
    num_runs = 100
    start_time = time.time()
    for _ in range(num_runs):
        Layer.forward
    end_time = time.time()
    
    # Calculate average time per run
    avg_time = (end_time - start_time) / num_runs
    print(f"Average time per accessing class method access : {avg_time} seconds")

if __name__ == "__main__":
    test_envs_timing()
    test_class_method()
```

### Proposed Change.

As an example, in model layers, such as linear layers, they should use the following abstraction:
`vllm/model_executor/layers/quantization/fp8.py`
```
[...]

PR #26670, opened 2025-10-13
[ROCm] Add AMD GPU support on Deepseek v3.2 and SparseMLA

ganyi1996ppo:
## Purpose
The PR add Deepseek v3.2 support on ROCm platforms. The main change in this PR include: 
- Replace all hardcode float8_e4m3fn to platform supported fp8 dtype, and add FP8_E4M3FNUZ enum to cpp kernels.
- Add torch impl to deepgemm 
- Add rocm_aiter_mla_sparse backend and dispatch it in rocm platform

## Test Plan
Verify its accuracy on gsm8k and wikitext

## Test Result
```
# wikitext
| Tasks  |Version|Filter|n-shot|    Metric     |   |Value |   |Stderr|
|--------|------:|------|-----:|---------------|---|-----:|---|------|
|wikitext|      2|none  |     0|bits_per_byte  |↓  |0.3691|±  |   N/A|
|        |       |none  |     0|byte_perplexity|↓  |1.2915|±  |   N/A|
|        |       |none  |     0|word_perplexity|↓  |3.9272|±  |   N/A|

# gsm8k

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9629|±  |0.0052|
|     |       |strict-match    |     5|exact_match|↑  |0.9621|±  |0.0053|
 
```

---
<details>

</details>


HAIAI (changes requested):
Please revert the `FP8_E4M3FNUZ` part for kv cache, we had uniformed the fp8 KV cache format since 3.2024, and OCP is the only format we support from Quark as well.

vllm/v1/attention/backends/mla/indexer.py
```diff
@@ -53,6 +54,8 @@ def get_kv_cache_stride_order() -> tuple[int, ...]:
 
     @classmethod
     def get_supported_kernel_block_size(cls) -> list[int | MultipleOf]:
+        if current_platform.is_rocm():
+            return [MultipleOf(1)]
```
HAIAI: why `1`

vllm/platforms/rocm.py
```diff
@@ -243,6 +241,13 @@ def get_attn_backend_cls(
             from vllm.v1.attention.backends.mla.rocm_aiter_mla import (
                 is_aiter_mla_enabled,
             )
+            if use_sparse:
+                if kv_cache_dtype.startswith("fp8"):
+                    raise ValueError(f"ROCMAiterMLASparseBackend dose not support kv_cache_dtype == fp8.")
```
HAIAI: fix typo

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
```diff
@@ -0,0 +1,404 @@
...
+    slot_mapping: torch.Tensor
+
+    block_table: torch.Tensor
+    req_id_per_token: torch.Tensor
+    block_size: int = 64
+    topk_tokens: int = 2048
+
+
+
+def ref_convert_to_gloabl(req_id: torch.Tensor,
```
HAIAI: typo

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
```diff
@@ -0,0 +1,404 @@
...
+    topk_indices,
+    kv_lora_rank=512
+):
+    from aiter.ops.triton.unified_attention_sparse_mla import unified_attention_sparse_mla
+
+    qs, qh, _ = q.shape
+    out = torch.randn([qs, qh, kv_lora_rank], dtype=torch.bfloat16, device=q.device)
+    block_table = torch.empty_like(topk_indices)    # no use
+    num_seq = cu_seqlens_q.size(0) - 1
+    seqused_k = torch.randn([num_seq], device=q.device)  # no use
```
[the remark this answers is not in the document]
HAIAI: same question

HAIAI (changes requested):
lint: few more comments
And please format accordingly.

vllm/platforms/rocm.py
SageMoore: Nit: This looks like a RuntimeError

vllm/v1/attention/backends/mla/indexer.py
SageMoore: I have the same question

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
```diff
@@ -0,0 +1,404 @@
...
+      q: torch.Tensor,
+      kv_c_and_k_pe_cache: torch.Tensor,
+      topk_indices: torch.Tensor,
+      attn_metadata: ROCMAiterMLASparseMetadata,
+    ) -> torch.Tensor:
+      num_tokens = q.shape[0]
+      kv_c_and_k_pe_cache = kv_c_and_k_pe_cache.view(
+          -1, 1, kv_c_and_k_pe_cache.shape[-1])
+
+      # NOTE(Chen): kernel requires num_local_head to be a multiple of
```
SageMoore: Nit: Let's update the comment or remove it.

vllm/envs.py
```diff
@@ -894,9 +894,9 @@ def get_vllm_port() -> int | None:
         os.getenv("VLLM_ROCM_USE_AITER_FP4_ASM_GEMM", "False").lower() in ("true", "1")
     ),
     # Whether to use aiter rope.
-    # By default is disabled.
+    # By default is enabled.
     "VLLM_ROCM_USE_TRITON_ROPE": lambda: (
-        os.getenv("VLLM_ROCM_USE_TRITON_ROPE", "False").lower() in ("true", "1")
+        os.getenv("VLLM_ROCM_USE_TRITON_ROPE", "True").lower() in ("true", "1")
```
tjtanaa: @ganyi1996ppo 
To actually call the AITER triton rope code path, we also have to set both `VLLM_ROCM_USE_TRITON_ROPE=1` and  `--compilation-config '{"custom_ops": ["+rotary_embedding"]}' `. Only by setting the `custom_ops` that the `forward_hip` function will be called, else it will call `forward_native` by default. I think you will encounter a bug with the DeepSeekRotary Embedding class when trying to use the AITER triton rope.

For my understanding, this AITER triton rope has not been validated for all ROPE classes. It was introduced and tested on Llama-3-405B only. https://github.com/vllm-project/vllm/pull/25135 . So it is better to set it to False for now. Or open another PR to bugfix this and validate on all ROPE classes.

vllm/platforms/rocm.py
ganyi1996ppo: Hi, this `ValueError` is meant to tell the users that we don't support fp8 kv cache for now, so we throw value error here.

vllm/v1/attention/backends/mla/indexer.py
ganyi1996ppo: The current triton kernel `deepgemm_fp8_paged_mqa_logits_stage1` we used to replace deepgemm only support `block-size=1` case and so dose our MLA kernel, although it not used in deepseekv3.2. We already have a solution on this, but not in this PR.

vllm/envs.py
ganyi1996ppo: This PR only enables eager mode for dsv3.2, we haven't test graph mode for now. The interesting facts is that Mi300/Mi350(we only test these 2) can only get correct result when using triton rope.

And of course we will investigate this case, and since rope will be broke up to small ops and get fused by torch.inductor in graph mode, this change actually won't effect the default path in graph mode.

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
ganyi1996ppo: Done, irrelevant comments have been removed.

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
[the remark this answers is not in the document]
ganyi1996ppo: Done

vllm/v1/attention/backends/mla/indexer.py
```diff
@@ -335,10 +338,10 @@ def build(
             requires_padding = (decode_lens_cpu.max() > decode_lens_cpu.min()).item()
 
             seq_lens = common_attn_metadata.seq_lens[:num_decodes]
-
-            self.scheduler_metadata_buffer[:] = get_paged_mqa_logits_metadata(
-                seq_lens, self.kv_cache_spec.block_size, self.num_sms
-            )
+            if current_platform.is_cuda():
```
SageMoore: Looking at the code, it appears that we only need this when running with DeepGemm? Can you update the check to `if has_deep_gemm()`

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
```diff
@@ -0,0 +1,376 @@
...
+        output_block_scale: torch.Tensor | None = None,
+    ) -> torch.Tensor:
+        # NOTE(lucas): for the sparse FlashMLA kernels the kernels want to use
+        # MQA 576/512 approach for both prefill and decode
+
+        assert output is not None, "Output tensor must be provided."
+
+        if output_scale is not None or output_block_scale is not None:
+            raise NotImplementedError(
+                "fused output quantization is not yet supported for MLACommonImpl"
```
SageMoore: Nit: Can you update the string to reference this class

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
```diff
@@ -0,0 +1,376 @@
...
+            )
+        else:
+            # Multiply (N, B, P) x (N, P, L) -> (N, B, L)
+            ql_nope = torch.bmm(q_nope, self.W_UK_T)
+            # Convert from (N, B, L) to (B, N, L)
+            ql_nope = ql_nope.transpose(0, 1)
+
+        topk_indices = self.topk_indices_buffer[:num_actual_toks]
+
+        # Note: the above triton kernel may triggers some strange unexpected
```
SageMoore: What kernel is this comment referring to?

vllm/utils/deep_gemm.py
```diff
@@ -204,7 +298,7 @@ def fp8_mqa_logits(
     """
     _lazy_init()
     if _fp8_mqa_logits_impl is None:
-        return _missing()
+        return fp8_mqa_logits_torch(q, kv, weights, cu_seqlen_ks, cu_seqlen_ke)[0]
```
SageMoore: Doing this dispatching inside of a file that's just for wrapping Deep Gemm seems incorrect. Let's move this dispatching somewhere else. Looks like indexer.py could be a good spot? CC @LucasWilkinson

vllm/utils/deep_gemm.py
```diff
@@ -260,8 +354,30 @@ def fp8_paged_mqa_logits(
         `torch.float32`.
     """
     _lazy_init()
+    if current_platform.is_rocm() and envs.VLLM_ROCM_USE_AITER:
```
SageMoore: Same idea here. Let's get this out of deep_gemm.py

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
```diff
@@ -0,0 +1,376 @@
...
+    from vllm.model_executor.models.deepseek_v2 import Indexer
+logger = init_logger(__name__)
+
+if is_rocm_aiter_fp8bmm_enabled():
+    from aiter.ops.triton.batched_gemm_a8w8_a_per_token_group_prequant_w_per_batched_tensor_quant import (  # noqa: E501
+        batched_gemm_a8w8_a_per_token_group_prequant_w_per_batched_tensor_quant as aiter_triton_fp8_bmm,  # noqa: E501
+    )
+
+    def dynamic_per_batched_tensor_quant(
+        x: torch.Tensor, dtype: torch.dtype = torch.float8_e4m3fn
```
rasmith: Why not `torch.dtype = current_platform.fp8_dtype()`?

csrc/cache_kernels.cu
```diff
@@ -552,7 +552,11 @@ __global__ void indexer_k_quant_and_cache_kernel(
 #ifndef USE_ROCM
   __syncwarp();
 #endif
+#if defined(__gfx942__)
```
gshtras, line 555: Better way would be to check for HIP_FP8_TYPE_OCP from amd_hip_fp8.h

vllm/envs.py
```diff
@@ -107,7 +107,7 @@
     VLLM_ROCM_USE_AITER_MLA: bool = True
     VLLM_ROCM_USE_AITER_MHA: bool = True
     VLLM_ROCM_USE_AITER_FP4_ASM_GEMM: bool = False
-    VLLM_ROCM_USE_TRITON_ROPE: bool = False
+    VLLM_ROCM_USE_TRITON_ROPE: bool = True
```
divakar-amd: nit: can we rename this to `VLLM_ROCM_USE_AITER_TRITON_ROPE` ?

vllm/utils/deep_gemm.py
ganyi1996ppo: Sorry for the delay. Yes, I think we can make it a custom_op and dispatch it outside rather than put it inside the deep_gemm.py

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
ganyi1996ppo: This refers to this triton kernel `triton_convert_req_index_to_global_index`, I'm using a reference torch path to replace it now. But I think we should go `triton_convert_req_index_to_global_index` eventually for performance consideration, and I'm working on it now.

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
ganyi1996ppo: Sure, thanks to ping that out~

vllm/v1/attention/backends/mla/indexer.py
ganyi1996ppo: I'll refactor this place, thanks~

vllm/utils/deep_gemm.py
ganyi1996ppo: ok~

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
ganyi1996ppo: Sorry for the delay, got caught up by something else. Yes, there should be using platform related dtype, thanks to point that out!

csrc/cache_kernels.cu
ganyi1996ppo: Sorry for the delay, got caught up by something else. Thanks for the suggestion.

vllm/envs.py
ganyi1996ppo: Don't know if that's necessary for this PR. I prefer to leave it untouched in this PR and figure out why disable triton rope may cause accuracy issue.

divakar-amd: [This PR](https://github.com/vllm-project/vllm/pull/27373) resolves the accuracy issue. It was caused due to mis-match  of in-place/out-place return values.

ganyi1996ppo: That's Great! Thanks for this message! I'll remove this line.

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
ganyi1996ppo: done

vllm/platforms/rocm.py
ganyi1996ppo: done

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
ganyi1996ppo: Hi @SageMoore , triton crash issue have already been fixed in newest commit, please take a look.

vllm/envs.py
ganyi1996ppo: Hi @divakar-amd , this change have been removed, please take a look

vllm/utils/deep_gemm.py
ganyi1996ppo: I have isolate the sparse_attn_indexer function out of the deepseek modeling file. This should look clean than previous version, please take a look again.

vllm/utils/deep_gemm.py
ganyi1996ppo: Done

csrc/cache_kernels.cu
ganyi1996ppo: Changed to `HIP_FP8_TYPE_OCP` already, please take a look

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
ganyi1996ppo: This function is unnecessary, removed.

vllm/envs.py
divakar-amd: LGTM

vllm/envs.py
ganyi1996ppo: FYI, piecewise cudagraph is now supported

wuhuikx: Agree with @tjtanaa , we should set False by default.

vllm/v1/attention/backends/mla/indexer.py
wuhuikx: Can you add some comments here?

wuhuikx (approved):

wuhuikx:
Could you please help review this PR again? Many thanks. @tjtanaa  @gshtras  @HaiShaw

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
```diff
@@ -0,0 +1,329 @@
...
+        batched_gemm_a8w8_a_per_token_group_prequant_w_per_batched_tensor_quant as aiter_triton_fp8_bmm,  # noqa: E501
+    )
+
+
+class ROCMAiterMLASparseBackend(AttentionBackend):
+    accept_output_buffer: bool = True
+
+    @staticmethod
+    def get_name() -> str:
+        return "ROCMAITERMLA_SPARSE"
```
tjtanaa: NITS: can you rename this to `ROCM_AITER_MLA_SPARSE` to align with the convention used in other backend?

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
```diff
@@ -0,0 +1,329 @@
...
+        # Inputs and outputs may be padded for CUDA graphs
+
+        q = q[:num_actual_toks, ...]
+        k_c_normed = k_c_normed[:num_actual_toks, ...]
+        k_pe = k_pe[:num_actual_toks, ...]
+
+        q_nope, q_pe = q.split([self.qk_nope_head_dim, self.qk_rope_head_dim], dim=-1)
+        # Convert from (B, N, P) to (N, B, P)
+        q_nope = q_nope.transpose(0, 1)
+        if is_rocm_aiter_fp8bmm_enabled():
```
tjtanaa: Please reuse the flag from `_aiter_ops.py` `rocm_aiter_ops.is_fp8bmm_enabled()` 

It is recommended to evaluate this condition in the `def __init__()` and reuse it as recommended in RFC https://github.com/vllm-project/vllm/issues/17067 as a best practice.

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
```diff
@@ -0,0 +1,329 @@
...
+            # Multiply+Transpose (N, B, P)x(N, P, L)->(N, B, L)->(B, N, L)
+            ql_nope = aiter_triton_fp8_bmm(
```
tjtanaa: `aiter_triton_fp8_bmm` has been moved to `_aiter_ops.py`, because it is an aiter function that is used for multiple files and to ease the imports of a function with a very long name.

You can call `rocm_aiter_ops.triton_fp8_bmm(...)`

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
```diff
@@ -0,0 +1,329 @@
...
+    AttentionMetadataBuilder,
+    CommonAttentionMetadata,
+)
+from vllm.v1.kv_cache_interface import AttentionSpec
+
+if TYPE_CHECKING:
+    from vllm.model_executor.models.deepseek_v2 import Indexer
+logger = init_logger(__name__)
+
+if is_rocm_aiter_fp8bmm_enabled():
```
tjtanaa: This import statement can be removed.

`aiter_triton_fp8_bmm` has been moved to `_aiter_ops.py`, because it is an aiter function that is used for multiple files and to ease the imports of a function with a very long name.

vllm/attention/ops/rocm_aiter_mla_sparse.py
```diff
@@ -0,0 +1,204 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+import torch
+
+import vllm.envs as envs
+from vllm.logger import init_logger
+from vllm.platforms import current_platform
+
+if current_platform.is_cuda_alike():
```
tjtanaa: This statement seems to be unused code.

vllm/attention/ops/rocm_aiter_mla_sparse.py
```diff
@@ -0,0 +1,204 @@
...
+            block indices to physical blocks in the paged cache.
+        schedule_metadata: Returned by `get_paged_mqa_logits_metadata`;
+            used to distribute work across SMs.
+        max_model_len: Maximum sequence length used to size the logits output.
+
+    Returns:
+        Logits tensor of shape [B * next_n, max_model_len], dtype
+        `torch.float32`.
+    """
+    if current_platform.is_rocm() and envs.VLLM_ROCM_USE_AITER:
```
tjtanaa: Do we need a branching logic here given that this backend is used only when `VLLM_ROCM_USE_AITER` is enabled?

vllm/attention/ops/rocm_aiter_mla_sparse.py
```diff
@@ -0,0 +1,204 @@
...
+        from aiter.ops.triton.pa_mqa_logits import deepgemm_fp8_paged_mqa_logits_stage1
```
tjtanaa, line 188: It seems this kernel is not found in the AITER version specified in the `Dockerfile.rocm_base` https://github.com/vllm-project/vllm/blob/684f2545851ee0ee49be9a80545ed497324f1a96/docker/Dockerfile.rocm_base#L10

We should comment this out first and add the `TODO comment` like in 
`rocm_fp8_mqa_logits` .

CC @wuhuikx

ganyi1996ppo:
> @ganyi1996ppo Sorry for the long comments.
> 
> This AITER OPs refactor has landed #24490 , it addressed this RFC #21504, the **aiter ops and custom op registration** should be placed under `_aiter_ops.py` to **avoid circular import and duplicated registration**. Moreover, it serves as a centralised place to manage **all of the AITER environment variables** and `is_xxx_enabled()` function. This way, kernel integrators is aware that there is already such a flag combination `envs.VLLM_ROCM_USE_AITER and envs.VLLM_ROCM_USE_AITER_LINEAR` so that we don't re-define the `is_xxx_enabled()` function again in the file.
> 
> This RFC #17067 highlighted the cost of calling `envs.VARIABLES` in forward pass as the cost is non-negligible. The `_aiter_ops.py` will evaluate the `envs.VARIABLES` and stored within the class variable to reduce the CPU cost. Kernel integrator can just call `rocm_aiter_ops.is_xxxx_enabled()`
> 
> It is encouraged to try to avoid calling dispatcher functions in the forward pass and retrieve the necessary function in the class’s `__init__()` that we are working with so that the forward pass is simple and could help avoid unnecessary branching.
> 
> For ops related to attention backend that are not going to be reused elsewhere can live within the attention ops and attention backend class. E.g. `vllm/attention/ops/rocm_aiter_mla_sparse.py` which is exclusive to attention backend, we can store it in its own file, we don't need to store it in `_aiter_ops.py`

Thanks for this info, I'll aligned with the RFC's design

vllm/attention/ops/rocm_aiter_mla_sparse.py
ganyi1996ppo: I use the `hasattr` to decide which path to choose, please take a look

vllm/attention/ops/rocm_aiter_mla_sparse.py
ganyi1996ppo: Removed

vllm/attention/ops/rocm_aiter_mla_sparse.py
ganyi1996ppo: removed

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
ganyi1996ppo: Done

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
ganyi1996ppo: good work! Done

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
ganyi1996ppo: Done

vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py
ganyi1996ppo: Good suggestion! Done

vllm/attention/ops/rocm_aiter_mla_sparse.py
```diff
@@ -0,0 +1,203 @@
...
+            block indices to physical blocks in the paged cache.
+        schedule_metadata: Returned by `get_paged_mqa_logits_metadata`;
+            used to distribute work across SMs.
+        max_model_len: Maximum sequence length used to size the logits output.
+
+    Returns:
+        Logits tensor of shape [B * next_n, max_model_len], dtype
+        `torch.float32`.
+    """
+    has_paged_mqa_logits = hasattr("aiter.ops.triton", "pa_mqa_logits")
```
tjtanaa: AITER in `Dockerfile.rocm_base` has been upgraded to commit `59bd8ff2` which has this function now.

ganyi1996ppo: That's for the info, I'll test cudagraph on new nightly docker image.

tjtanaa: @ganyi1996ppo can you update the PR description to share the command to run the model?

When I am trying to validate this PR I ran the following command and encounter issue

```bash
#!/bin/bash

rm -rf /root/.cache/vllm

export VLLM_USE_V1=1
export SAFETENSORS_FAST_GPU=1
export VLLM_ROCM_USE_AITER=1
export VLLM_ROCM_USE_AITER_MOE=1
export NCCL_DEBUG=WARN
export VLLM_RPC_TIMEOUT=1800000
export VLLM_ROCM_USE_AITER_FUSION_SHARED_EXPERTS=1


# cache dirs
export VLLM_CACHE_ROOT=/root/.cache/vllm
export TORCHINDUCTOR_CACHE_DIR=/root/.cache/inductor

rm -rf /root/.cache/

export HF_HUB_CACHE=/app/model

model_path=deepseek-ai/DeepSeek-V3.2-Exp
export AITER_ENABLE_VSKIP=0
vllm serve $model_path \
    --tensor-parallel-size 8 \
    --max-num-batched-tokens 32768 \
    --trust-remote-code \
    --no-enable-prefix-caching \
    --disable-log-requests \
    --gpu_memory_utilization 0.9 \
    --compilation-config '{"cudagraph_mode": "PIECEWISE"}' \
    --async-scheduling \
    2>&1 | tee server-deepseek-ai_DeepSeek-V3.2-Exp-2.log
 ```

```
 File "/app/rocmvllm/dsv3.2_rocm_support/vllm/v1/worker/gpu_worker.py", line 398, in initialize_from_config                                                                                                
self.model_runner.initialize_kv_cache(kv_cache_config)                                                                                                                                                  
  File "/app/rocmvllm/dsv3.2_rocm_support/vllm/v1/worker/gpu_model_runner.py", line 4776, in initialize_kv_cache                                                                                            
   kernel_block_sizes = self._prepare_kernel_block_sizes(kv_cache_config)                                                                                                                                  
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                  
  File "/app/rocmvllm/dsv3.2_rocm_support/vllm/v1/worker/gpu_model_runner.py", line 4549, in _prepare_kernel_block_sizes                                                                                    
  selected_kernel_size = self.select_common_block_size(                                                                                                                                                   
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                   
File "/app/rocmvllm/dsv3.2_rocm_support/vllm/v1/worker/gpu_model_runner.py", line 4431, in select_common_block_size                                                                                       
  raise ValueError(f"No common block size for {kv_manager_block_size}. ")                                                                                                                                 
ValueError: No common block size for 16.                 
```

Comment (ganyi1996ppo) [reply vllm/attention/ops/rocm_aiter_mla_sparse.py]: We might still need `block_size=1` if we want to use `pa_mqa_logits`, this kernel helps a lot on performance.

Comment (ganyi1996ppo) [reply vllm/attention/ops/rocm_aiter_mla_sparse.py]: New version pushed up, `block-size=1` support added to the indexer if its rocm platform,  please replace `block_size` to 1, and try again

File: vllm/platforms/rocm.py [hunk 23]
```diff
@@ -222,7 +222,16 @@ def get_attn_backend_cls(
...
 
         if use_sparse:
-            raise NotImplementedError("Sparse Attention is not supported on ROCm.")
+            if kv_cache_dtype.startswith("fp8"):
+                raise ValueError(
+                    "ROCMAiterMLASparseBackend doesn't support fp8 kv_cache_dtype."
+                )
+
+            logger.info_once("Using Sparse MLA backend on V1 engine.")
+            return (
```
tjtanaa, line 236: @ganyi1996ppo what do you think about adding an assertion of block-size 1 when we are using this backend?

ganyi1996ppo: Sounds good

ganyi1996ppo: Done, please take a look.

vllm/attention/ops/rocm_aiter_mla_sparse.py
```diff
@@ -176,7 +180,9 @@ def rocm_fp8_paged_mqa_logits(
         Logits tensor of shape [B * next_n, max_model_len], dtype
         `torch.float32`.
     """
-    has_paged_mqa_logits = hasattr("aiter.ops.triton", "pa_mqa_logits")
+    has_paged_mqa_logits = (
+        importlib.util.find_spec("aiter.ops.triton.pa_mqa_logits") is not None
```
tjtanaa: this line of code is very slow. I know that this is a temporary solution while waiting for aiter main to merge the feature. So, in the meantime, can you write a caching function to cache the results? Then add a TODO to remind us to remove it when AITER main has merged the feature.

vllm/attention/ops/rocm_aiter_mla_sparse.py
```diff
@@ -80,7 +82,9 @@ def rocm_fp8_mqa_logits(
     """
 
     # TODO(ganyi): Uncomment this after aiter merge this kernel into main
-    has_mqa_logits = hasattr("aiter.ops.triton", "fp8_mqa_logits")
+    has_mqa_logits = (
+        importlib.util.find_spec("aiter.ops.triton.fp8_mqa_logits") is not None
```
tjtanaa: this line of code is very slow. I know that this is a temporary solution while waiting for aiter main to merge the feature. So, in the meantime, can you write a caching function to cache the results? Then add a TODO to remind us to remove it when AITER main has merged the feature.

vllm/attention/ops/rocm_aiter_mla_sparse.py
tjtanaa: @ganyi1996ppo I have checked, it seems this function existed in the AITER commit used in the` Dockerfile.rocm_base`. Let me test it by removing this checks and get back to you. If it works, let's remove this line

vllm/attention/ops/rocm_aiter_mla_sparse.py
tjtanaa: This feature has been merged into AITER main, but the AITER commit used in the `Dockerfile.rocm_base` still does not contain this function, so we still need to check if the module exists or not.

ganyi1996ppo: Oh, you are right, thanks for the suggestion. I add `lru_cache` for the function and also comments the future plan as TODO

vllm/attention/ops/rocm_aiter_mla_sparse.py
ganyi1996ppo: It works, I verified my own before.  I'll remove it.

ganyi1996ppo: But we still need the ref path right? for non-aiter path

tjtanaa (approved):

tjtanaa:
@ganyi1996ppo can you try rebasing again. Thank you so much.

ganyi1996ppo:
> @ganyi1996ppo can you try rebasing again. Thank you so much.

Sure, but the error seems mainly related to llama and oom issue, Should not caused by this PR right?

PR #26674, opened 2025-10-13
[Core] Initialize LoRA support for tower and connector in multi-modal models

### LINKED ISSUE #27916: [Feature]: Does the latest version support LoRa for visual models?

When I loaded the QWEN2.5-VL model fine-tuned by LoRa using vllm version 0.8.4, I encountered the following prompt:

> Regarding multimodal models, vLLM currently only supports adding LoRA to language model, visual.blocks.31.mlp.up_proj will be ignored.

I found an issue https://github.com/vllm-project/vllm/issues/26422 with a similar problem, but it seems the PR hasn't been merged into master. How can I enable loading visual-side LORA parameters and using VLLM to accelerate inference?

Looking forward to your reply

### Alternatives

_No response_

### Additional context

_No response_

jeejeelee:
## Purpose
FIX https://github.com/vllm-project/vllm/issues/26422
FIX https://github.com/vllm-project/vllm/issues/27916
FIX https://github.com/vllm-project/vllm/issues/29635

### Summary

This PR enables LoRA support for the  tower modules and connector in multi-modal models. Previously, a single global Punica wrapper was used, which was insufficient because the input structures for the tower/connector differ from those of the language model.

### Key Changes

To address this, the following changes were implemented:

* **Decoupled Punica Wrappers:** Instead of sharing a single wrapper, separate Punica wrappers are now instantiated for the language model, tower model , and connector respectively.
* **New `ProcessingInfo` Methods:** Added `get_num_mm_encoder_tokens` and `get_num_mm_connector_tokens` to the `ProcessingInfo` interface. These are used to calculate the actual input token counts for the vision tower and connector.

### Why add `get_num_mm_*_tokens`?

The number of multi-modal tokens represented in the language model does not necessarily match the input length required by the linear layers in the vision tower or connector. Since the `lora_mapping` requires the precise input token length prior to activation, these helper functions are necessary to bridge the discrepancy and calculate the correct lengths. 

### Implementation Details

Currently, this PR implements the logic for the **Qwen2/2.5/3 VL and Idefics3**. Below is an example of the implementation:

```python
def get_num_mm_encoder_tokens(
    self,
    num_image_tokens: int,
) -> int:
    hf_config = self.get_hf_config()
    vision_config = hf_config.vision_config
    merge_size = vision_config.spatial_merge_size

    return num_image_tokens * merge_size**2
```

**Note:** To support LoRA for the tower and connector of other multi-modal models, these two functions must be implemented in their respective classes.

## Test Result

---
<details>

</details>


shangzhensen0620-ui:
image
这个分支测试后报错如下：
image

vllm/config/lora.py
```diff
@@ -55,6 +55,9 @@ class LoRAConfig:
     per prompt. When run in offline mode, the lora IDs for n modalities
     will be automatically assigned to 1-n with the names of the modalities
     in alphabetic order."""
+    enable_mm_lora: bool = False
```
jeejeelee: I think we should use `enable_tower_connector_lora` here

vllm/config/lora.py
```diff
@@ -55,6 +55,9 @@ class LoRAConfig:
...
+    """If `True`, LoRA support for multimodal models will be enabled. Currently, 
+    only the qwenvl series models support this feature. The default is False."""
```
jeejeelee: The related string also need to be modified

vllm/engine/arg_utils.py
```diff
@@ -1006,6 +1007,10 @@ def add_cli_args(parser: FlexibleArgumentParser) -> FlexibleArgumentParser:
             "--lora-dtype",
             **lora_kwargs["lora_dtype"],
         )
+        lora_group.add_argument(
```
jeejeelee, line 1000: DITTO

vllm/lora/layers/row_parallel_linear.py
```diff
@@ -63,6 +63,10 @@ def forward(
             input_parallel = splitted_input[self.tp_rank].contiguous()
 
         # Matrix multiply.
+        # Only fuse bias add into GEMM for rank 0 (matches base
+        # RowParallelLinear behavior). This ensures bias will not get
+        # added more than once in TP>1 case and matches the numerical
+        # behavior of the unwrapped layer
```
jeejeelee: I remember we had fixed this issue,right? @prashanth058

prashanth058: Hi @jeejeelee , correct this is fixed.
above comment was probably left over while resolving merge conflicts. we should remove the comment.

vllm/lora/layers/column_parallel_linear.py
```diff
@@ -340,7 +340,12 @@ def can_replace_layer(
         packed_modules_list: list,
         model_config: PretrainedConfig | None = None,
     ) -> bool:
-        return type(source_layer) is QKVParallelLinear and len(packed_modules_list) == 1
+        # Vision tower QKV has packed_modules_list=[] (already packed in checkpoint)
+        # Language models have packed_modules_list=[module_name]
+        # (single LoRA for qkv_proj)
+        return type(source_layer) is QKVParallelLinear and (
```
jeejeelee: @prashanth058  I think this change isn't reasonable, if  the tower QKV already packed in checkpoint, we should implement ` packed_modules_mapping` like:
```
packed_modules_mapping = {
        "query_key_value": ["query_key_value"],
        "dense_h_to_4h": ["dense_h_to_4h"],
    }
````

Comment (prashanth058) [reply vllm/lora/layers/column_parallel_linear.py]: @jeejeelee was not sure if it made sense to add those identity mappings for multiple models but if we are already using similar mappings in other places we can change to that. PR to change : https://github.com/jeejeelee/vllm/pull/11

**Comment by jeejeelee** [comment]
@DarkLight1337  @Isotr0py  As an initial PR to for  tower and connector LoRA support, it's readt for preview, pleas take a look , thank you very much

File: vllm/model_executor/models/qwen2_5_vl.py [hunk 6]
```diff
@@ -1023,6 +1023,25 @@ def get_replacement_qwen2vl(item_idx: int, modality: str):
...
 
+    def get_num_mm_encoder_tokens(
+        self,
+        num_image_tokens: int,
+    ) -> int:
+        hf_config = self.get_hf_config()
+        vision_config = hf_config.vision_config
+        merge_size = vision_config.spatial_merge_size
+
+        return num_image_tokens * merge_size**2
```
Isotr0py: Hmm, I think `num_image_tokens * merge_size**2` doesn't present actual `vision_seq_lens` for Qwen-VL family. Because `pos_info.length` can include non-embeds tokens as well. 🤔

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2129,6 +2145,69 @@ def _execute_mm_encoder(
...
+            token_lora_mapping = []
+            lora_requests = set()
+
+            for req_id, (_, pos_info) in zip(encoder_req_ids, mm_hashes_pos):
+                req_idx = self.input_batch.req_id_to_index[req_id]
+                lora_id = int(self.input_batch.request_lora_mapping[req_idx])
+
+                num_tokens = self.info.get_num_mm_encoder_tokens(  # type: ignore[attr-defined]
+                    pos_info.length
+                )
```
Isotr0py: If we want to know how many tokens cost by mm embeddings, I think we can calculate it from `pos_info.is_embed.sum()`

vllm/lora/model_manager.py
```diff
@@ -82,34 +95,106 @@ def __init__(
...
+        mm_budget = MultiModalBudget(
+            model_config,
+            vllm_config.scheduler_config,
+            MULTIMODAL_REGISTRY,
+        )
+        limit_per_prompt: int = max(self.info.get_allowed_mm_limits().values())
+
+        # For vision tower
+        num_encoder_tokens = self.info.get_num_mm_encoder_tokens(
+            mm_budget.get_encoder_budget()
```
DarkLight1337, line 181: Maybe we can be conservative and directly use `mm_budget.get_encoder_budget()` to avoid the edge cases @Isotr0py mentioned.

DarkLight1337: The buffer size used in `PunicaWrapperBase` doesn't seem to be that large anyway

vllm/v1/worker/gpu_model_runner.py
linitra24: Good catch. Thanks for pointing that out. I've fixed the issue.

vllm/lora/model_manager.py
linitra24: Sorry, I'm not sure I fully understood. If we use mm_budget.get_encoder_budget(), wouldn't that just return the maximum token count? I think @Isotr0py  was referring to the actual number of MM tokens in the current input.

vllm/v1/worker/gpu_model_runner.py
prashanth058: for my own understanding. are you referring to `pos_info.length` including the `|vision_start|`, `|vision_end|` special tokens in the QwenVL models? : )

@B-201 looks like there is a `pos_info.get_num_embeds()` method that we can use?

linitra24: > pos_info.get_num_embeds()

You are right. It's better to use `get_num_embeds()`. Thanks for pointing that out.

vllm/lora/model_manager.py
DarkLight1337: Oops, I missed that `get_num_mm_encoder_tokens` isn't only used for buffer allocation. Then indeed we need the precise number of tokens in that case.

vllm/model_executor/models/qwen2_5_vl.py
```diff
@@ -1007,6 +1007,25 @@ def get_replacement_qwen2vl(item_idx: int, modality: str):
             for modality in ("image", "video")
         ]
 
+    def get_num_mm_encoder_tokens(
```
DarkLight1337: Can you add this to the base class? With the default raising `NotImplementedError`

Anexdeus: Looks like we don't need these methods in the *MultiModalProcessor classes
They are always called from ProcessingInfo like `self.info.get_num_mm_encoder_tokens()`

linitra24: @DarkLight1337 @Anexdeus Agreed, adding a default NotImplementedError in the base class makes perfect sense. However, since only the Qwen-VL series has implemented these methods so far, I’m concerned that adding them to the base class might break other multi-modal models. Should we consider some special handling to avoid this?

vllm/lora/model_manager.py
```diff
@@ -71,6 +81,10 @@ def __init__(
             lora_config: the LoRA configuration.
         """
         self.model: SupportsLoRA = model
+        self.supported_lora_modules = get_supported_lora_modules(self.model)
+        assert self.supported_lora_modules, "No supported LoRA modules found in"
+        f" {self.model.__class__.__name__}."
```
DarkLight1337:
```suggestion
        assert self.supported_lora_modules, (
            "No supported LoRA modules found in {self.model.__class__.__name__}."
        )
```

Accidentally made the string a no-op

linitra24: This has been addressed.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2150,6 +2168,72 @@ def _execute_mm_encoder(
...
+                    index_mapping=tuple(new_token_indices.tolist()),
+                    prompt_mapping=last_mapping.prompt_mapping,
+                    is_prefill=last_mapping.is_prefill,
+                    type=LoRAMappingType.CONNECTOR,
+                )
+
+                self.lora_manager.set_active_adapters(
+                    lora_requests,
+                    connector_mapping,
+                )
```
Isotr0py, line 2216: QQ: When we enable tower and connector lora, seems that the encoder cache will still cache mm embeddings based on `mm_hash` tied to mm inputs. 

I think this will cause incorrect cache hitting for requests with different lora enabled? If so, I think we might have to disable mm encoder cache for this case.

vllm/model_executor/models/qwen2_5_vl.py
Anexdeus: @DarkLight1337 @B-201 If we don't need to add these methods to other multimodal models yet, we can implement the mixin specifically for the Qwen-VL series now and generalize it later once other models are supported. I proposed the changes [here](https://github.com/jeejeelee/vllm/pull/12)

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -589,6 +590,15 @@ def __init__(
             pin_memory=self.pin_memory,
         )
 
+        # Multimodal LoRA support
+        self.enable_tower_connector_lora = False
+        if self.supports_mm_inputs and self.lora_config:
+            self.info = self.mm_registry.create_processor(self.model_config).info
```
DarkLight1337: Can we avoid creating the whole processor?

vllm/model_executor/models/qwen2_5_vl.py
DarkLight1337: > I’m concerned that adding them to the base class might break other multi-modal models.

As long as they aren't marked `abstractmethod` it should be fine. The point is to define the interface so we get some uniformity and also tell developers what is the purpose of this method.

> If we don't need to add these methods to other multimodal models yet, we can implement the mixin specifically for the Qwen-VL series now

I would avoid using mixins as it makes the inheritance tree even more complicated.

DarkLight1337: Also, personally I prefer defining these methods in the model class itself (part of `SupportsLoRA` interface) instead of inside `ProcessingInfo` because they don't actually influence input processing.

DarkLight1337: Since the model itself receives `hf_config` as part of `vllm_config`, we should still be able to calculate the appropriate values

Anexdeus: Thanks for the review! 
I implemented the methods within the model class and extended the SupportsMultiModal interface.

vllm/v1/worker/gpu_model_runner.py
prashanth058: You are right. Good catch. I assumed `lora_name` is included in the mm_hash similar to text LoRA KV cache but looks like thats not the case.

Instead of disabling mm encoder cache, would something like this(similar to text LoRA prefix caching) work? WDYT?  

```
def _get_mm_identifier(
    mm_hash: str,
    lora_request: LoRARequest | None,
) -> str:
    if lora_request is None:
        return mm_hash
    return f"{lora_request.lora_name}:{mm_hash}" 

mm_features.append(
      MultiModalFeatureSpec(
          data=decoder_mm_inputs[modality][idx],
          modality=modality,
-         identifier=decoder_mm_hashes[modality][idx],
+         identifier=_get_mm_identifier(
+             decoder_mm_hashes[modality][idx],
+             lora_request,
+         ),
          mm_position=decoder_mm_positions[modality][idx],
      )
  )

```

vllm/v1/worker/gpu_model_runner.py
linitra24: Fixed in the latest commit.

vllm/v1/worker/gpu_model_runner.py
linitra24: @prashanth058 I've pushed a commit attempting to fix this, but your approach definitely looks cleaner and more reasonable. Could you take a look at the latest changes? If they don't seem ideal, I’m happy to refactor the code according to your suggestion.

prashanth058: Thanks for the fix @B-201

I wonder if we could simplify this? Instead of threading lora_kwargs through the entire preprocessing pipeline, maybe the approach I suggested above would be less invasive?

Also looks like using hash(lora_request) instead of `lora_request.name` is more robust if someone overrides LoRARequest.__hash__, especially given that lora_name uniqueness doesn't seem to be enforced. WDYT? 

That said, I would defer to @Isotr0py 's opinion/suggestion.

tests/lora/test_qwenvl.py
```diff
@@ -175,3 +205,80 @@ def test_qwen25vl_lora(qwen25vl_lora_files):
...
+        model_path=QWEN2VL_MODEL_PATH,
+        lora_path=qwen2vl_vision_tower_lora_files,
+    )
+    tester = Qwen2VLTester(config)
+    for lora_id in [1, 2]:
+        tester.run_test(
+            TEST_IMAGES,
+            expected_outputs=EXPECTED_OUTPUTS_VISION_NO_CONNECTOR,
+            lora_id=lora_id,
+        )
```
Isotr0py: I think we can consolidate vision/language lora requests into one test reusing same instance, so that we can also test mm_encoder_cache with different lora requests. WDYT?

linitra24: This has been addressed.

Isotr0py:
> For now, perhaps we can disable the multi-modal processor cache when enable_tower_connector_lora is active? Could we address this more thoroughly in a follow-up PR?

Sure, feel free to disable the cache when enabling mm lora! We can address it in following PRs.

Isotr0py (approved):
LGTM now!

PR #26315, opened 2025-10-06
[Attention][UX][1/N] Add AttentionConfig and change attention env vars to CLI arguments

MatthewBonanni:
## Purpose
As part of the effort to reduce env variables (#25700), this PR introduces `AttentionConfig` and the `--attention-config` CLI argument group. The following environment variables are converted to CLI arguments:

```
VLLM_ATTENTION_BACKEND                        -> --attention-config.backend
VLLM_FLASH_ATTN_VERSION                       -> --attention-config.flash_attn_version
VLLM_V1_USE_PREFILL_DECODE_ATTENTION          -> --attention-config.use_prefill_decode_attention
VLLM_FLASH_ATTN_MAX_NUM_SPLITS_FOR_CUDA_GRAPH -> --attention-config.flash_attn_max_num_splits_for_cuda_graph
VLLM_USE_CUDNN_PREFILL                        -> --attention-config.use_cudnn_prefill
VLLM_USE_TRTLLM_RAGGED_DEEPSEEK_PREFILL       -> --attention-config.use_trtllm_ragged_deepseek_prefill
VLLM_USE_TRTLLM_ATTENTION                     -> --attention-config.use_trtllm_attention
VLLM_DISABLE_FLASHINFER_PREFILL               -> --attention-config.disable_flashinfer_prefill
VLLM_FLASHINFER_DISABLE_Q_QUANTIZATION        -> --attention-config.disable_flashinfer_q_quantization
```
Due to its significance, the attention backend may also be specified by `--attention-backend`

Backwards compatibility is maintained with warnings for removal in v0.14.

Many of the examples and tests will need to be updated. This will be done in follow-ups.

## Test Plan
CI should suffice

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #26680, opened 2025-10-13
remove attn output view kernel

### LINKED ISSUE #22293: [Feature]: Optimize RoPE

Recently, we have noticed that rope is relatively slow with vLLM based on profiling.

goal is to optimize rope by:
- investigate why torch.compile's result is not optimal
- investigate alternative kernels (e.g. FlashInfer's rope)

### Alternatives

_No response_

### Additional context

_No response_

BoyuanFeng:
Before this PR, attention output is allocated and initialized with 0 (due to `torch.zeros`), and the view into a shape, before the output tensor is used by any other ops. This becomes a triton kernel of ~1 us latency, which is on-par with a rope/layer norm (~1.6us) latency.

This PR changes to allocate with `torch.empty` which only allocates the tensor and does not initialize it. This allocation will be removed by cudagraph so it is free.

As a result, this PR removes the attn_out_view kernel at the end of this qwen3-0.6b trace.
image

See https://github.com/vllm-project/vllm/pull/26682#issue-3508480559 for perf win.


vllm/config/model.py
```diff
@@ -179,6 +179,12 @@ class ModelConfig:
     - 1k -> 1000\n
     - 1K -> 1024\n
     - 25.6k -> 25,600"""
+    init_attn_out: bool = False
+    """When true, initialize attention output with zero at the cost of an
+    extra cuda/triton kernel which cannot be removed by cudagraph. By
+    default, initialize attention output with `torch.empty` which can be
+    removed by cudagraph.
+    """
```
hmellor: This should go in the new `AttentionConfig` that's currently being added (and it should probably be overriden by the specific backends which require it rather than requiring the user to do it)

BoyuanFeng: yes it should be attention specific configs. any estimated timeline for AttentionConfig or reference pr?

hmellor: https://github.com/vllm-project/vllm/pull/26315 is the PR, I'm not sure how fast we can get it in

BoyuanFeng: since this PR is simple, do you mind we land it first and move the config to #26315 when it is ready?

BoyuanFeng: updated the pr to always initializing with 0 for profiling runs and use `torch.empty` for other runs. No need for the config now.

ProExpertProg: This sounds good. @hmellor this is on the milestone and Attentionconfig probably won't make into this release anyway.

ProExpertProg (approved):
Can you just check the attention fusion pass? I think it might rely on the fact that we use `zeros` for attention output

zou3519 (approved):

PR #26681, opened 2025-10-13
[compile] Enable sequence parallelism for full cuda graph without specifying compile sizes

angelayi:
## Purpose

Addresses https://github.com/vllm-project/vllm/issues/25277
Porting over the changes from @cascade812 in https://github.com/vllm-project/vllm/pull/21031 which enabled SP without needing to specify compile_sizes in CompilationConfig, + adding some special handling so that async TP/SP can be enabled if just `use_inductor_graph_partition=True`

## Test Result

Here are the updated numbers:

For meta-llama/Llama-3.1-70B-Instruct, updated the command:
```
vllm bench latency --model=meta-llama/Llama-3.1-70B-Instruct --output-len 1 --input-len 8192 --batch-size 1 --tensor-parallel-size 8 --load-format dummy --num_iters_warmup 5 --num_iters 15 -O '{"level": 3, "pass_config": {"enable_async_tp": true, "enable_sequence_parallelism": true}, "use_inductor_graph_partition": false, "splitting_ops":[], "cudagraph_mode": FULL}' --no-enable-prefix-caching --gpu_memory_utilization 0.6
```

Image

For RedHatAI/Meta-Llama-3.1-8B-Instruct-FP8:

```
vllm bench latency --model=RedHatAI/Meta-Llama-3.1-8B-Instruct-FP8 --output-len 1 --input-len 8192 --batch-size 1 --tensor-parallel-size 8 --load-format dummy --num_iters_warmup 5 --num_iters 15 -O '{"level": 3, "pass_config": {"enable_async_tp": true, "enable_sequence_parallelism": true}, "use_inductor_graph_partition": false, "splitting_ops":[], "custom_ops":["+quant_fp8"], "cudagraph_mode": FULL}' --no-enable-prefix-caching
```

Image

For nvidia/Llama-3.3-70B-Instruct-FP8:

Image

### DIFF
--- a/vllm/compilation/collective_fusion.py
+++ b/vllm/compilation/collective_fusion.py
@@ -435,7 +435,10 @@ class AsyncTPPass(VllmPatternMatcherPass):
         # This pass is applied on top of the sequence parallelism pass.
         # It inherits the same applicability condition as `SequenceParallelismPass`.
         # See `SequenceParallelismPass.is_applicable` for more details.
-        if self.splitting_ops is None or self.splitting_ops == []:
+        if (
+            not self.compilation_config.splitting_ops
+            or self.compilation_config.use_inductor_graph_partition
+        ):
             return True
         tp_size = get_tensor_model_parallel_world_size()
         return shape is not None and shape % tp_size == 0
--- a/vllm/compilation/collective_fusion.py
+++ b/vllm/compilation/collective_fusion.py
@@ -431,8 +431,12 @@ class AsyncTPPass(VllmPatternMatcherPass):
 
         self.dump_patterns(config, self.patterns)
 
-    def is_applicable_for_shape(self, shape: int | None) -> bool:
-        # only do replace for specific shapes
+    def is_applicable(self, shape: int | None) -> bool:
+        # This pass is applied on top of the sequence parallelism pass.
+        # It inherits the same applicability condition as `SequenceParallelismPass`.
+        # See `SequenceParallelismPass.is_applicable` for more details.
+        if self.splitting_ops is None or self.splitting_ops == []:
+            return True
         tp_size = get_tensor_model_parallel_world_size()
         return shape is not None and shape % tp_size == 0
 
--- a/vllm/compilation/sequence_parallelism.py
+++ b/vllm/compilation/sequence_parallelism.py
@@ -482,7 +482,22 @@ class SequenceParallelismPass(VllmPatternMatcherPass):
             ).register(self.patterns)
         self.dump_patterns(config, self.patterns)
 
-    def is_applicable_for_shape(self, shape: int | None) -> bool:
+    def is_applicable(self, shape: int | None) -> bool:
+        # When sequence parallelism is enabled, the residual tensor from RMSNorm
+        # needs to be split along the sequence dimension. However, this dimension
+        # is symbolic during piecewise compilation, and splitting symbolic shapes
+        # is not supported.
+        #
+        # This pass is therefore only applied when the sequence dimension is
+        # concrete:
+        # 1. In full-graph compilation mode (no splitting ops are used).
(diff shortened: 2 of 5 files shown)


vllm/compilation/vllm_inductor_pass.py
```diff
@@ -28,6 +28,7 @@ class VllmInductorPass(InductorPass):
     """Keep track of pass index for debug dump ordering."""
 
     def __init__(self, config: VllmConfig):
+        self.compilation_config = config.compilation_config
```
ProExpertProg: This will create a cycle, can you use a `weakref`?

vllm/compilation/sequence_parallelism.py
```diff
@@ -482,7 +482,23 @@ def __init__(self, config: VllmConfig):
...
-    def is_applicable_for_shape(self, shape: int | None) -> bool:
+    def is_applicable(self, shape: int | None) -> bool:
+        # When sequence parallelism is enabled, the residual tensor from RMSNorm
+        # needs to be split along the sequence dimension. However, this dimension
+        # is symbolic during piecewise compilation, and splitting symbolic shapes
+        # is not supported.
+        #
+        # This pass is therefore only applied when the sequence dimension is
+        # concrete:
+        # 1. In full-graph compilation mode (no splitting ops are used).
```
ProExpertProg: nit: no Dynamo splitting

vllm/compilation/collective_fusion.py
```diff
@@ -431,8 +431,13 @@ def __init__(self, config: VllmConfig):
 
         self.dump_patterns(config, self.patterns)
 
-    def is_applicable_for_shape(self, shape: int | None) -> bool:
-        # only do replace for specific shapes
+    def is_applicable(self, shape: int | None) -> bool:
+        # This pass is applied on top of the sequence parallelism pass.
```
ProExpertProg, line 435: Fine for now but @cascade812 isn't this pass technically fine no matter the shape? Obviously it won't match anything if sequence parallelism didn't run, but still

ProExpertProg (approved):
Nice work! Can you post the benchmarking numbers in this PR as well?

vllm/compilation/vllm_inductor_pass.py
```diff
@@ -28,12 +29,10 @@ class VllmInductorPass(InductorPass):
     """Keep track of pass index for debug dump ordering."""
 
     def __init__(self, config: VllmConfig):
+        self.compilation_config = weakref.ref(config.compilation_config)
```
ProExpertProg: Nit: if you use a weakref.proxy you don't need to call it! And this instance should always live as compilation config will always outlive the pass instances.

vllm/compilation/collective_fusion.py
cascade812: yes, removing this implementation would work as well.  But it would trigger matching logic for this pass which could add some overhead.

vllm/compilation/vllm_inductor_pass.py
angelayi: ooh ok! thanks!

ProExpertProg (approved):

mgoin (approved):

zou3519:
Btw, this PR causes the PostGradPassManager object to hold onto the CompilationConfig.

This is not good, because Inductor configs need to be serializable and deepcopy-able, the PostGradPassManager is added to the Inductor config, and the CompilationConfig holds some nn.Modules. Internally, we saw deepcopy fail on the nn.Modules and OOMs due to this

ProExpertProg:
Yeah we used to only save `PassConfig` to avoid a cycle but avoiding saving the `CompilationConfig` seems like a good idea as well. We can just save the relevant properties in `__init__`

PR #26682, opened 2025-10-13
use combo kernel to fuse qk-norm and qk-rope

### LINKED ISSUE #22293: [Feature]: Optimize RoPE

BoyuanFeng:
This PR enables horizontal fusion from inductor. This is helpful for fusing q-norm & k-norm into 1 kernel; and q-rope & k-rope into 1 kernel. It was NOT fused before since q and k have different shapes, which prevent some optimizations to happen.

The following trace comes from qwen3-0.6b.

Before:

image

After (together w/ #26680):
image

After qkv_proj, we reduces from 5 kernels to 2 kernels: 1 for qk-norm and 1 for qk-rope.

### Performance 
Applying this PR + #26680

Qwen/Qwen3-0.6B

Before:
image

After:
image

### DIFF
--- a/vllm/config/compilation.py
+++ b/vllm/config/compilation.py
@@ -367,6 +367,11 @@ class CompilationConfig:
     since we know all keys are in a range [0, max_capture_size],
     we can optimize it to list[int] for better lookup performance."""
 
+    use_horizontal_fusion = True
+    """Whether use horizontal fusion. This is
+    useful for fusing qk-norm and qk-rope when query and key have
+    different shapes."""
+
     # keep track of enabled and disabled custom ops
     enabled_custom_ops: Counter[str] = field(default_factory=Counter, init=False)
     """custom ops that are enabled"""
@@ -497,6 +502,10 @@ class CompilationConfig:
 
         if isinstance(self.pass_config, dict):
             self.pass_config = PassConfig(**self.pass_config)
+
+        if self.use_horizontal_fusion and is_torch_equal_or_newer("2.9.0.dev"):
+            self.inductor_compile_config["combo_kernels"] = True
+            self.inductor_compile_config["benchmark_combo_kernel"] = True
 
         # migrate the deprecated flags
         if not self.use_cudagraph:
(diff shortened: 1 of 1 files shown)


vllm/config/compilation.py
```diff
@@ -367,6 +367,11 @@ class CompilationConfig:
     since we know all keys are in a range [0, max_capture_size],
     we can optimize it to list[int] for better lookup performance."""
 
+    use_horizontal_fusion = True
```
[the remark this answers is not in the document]
BoyuanFeng: We want to enable by default since it benefits models in general.

zou3519 (approved):

mlazos (approved):

vllm/config/compilation.py
```diff
@@ -498,6 +498,16 @@ def __post_init__(self) -> None:
...
 
+        if (
+            is_torch_equal_or_newer("2.9.0.dev")
+            and "combo_kernels" not in self.inductor_compile_config
+            and "benchmark_combo_kernel" not in self.inductor_compile_config
+        ):
+            # use horizontal fusion, which is useful for fusing qk-norm and
+            # qk-rope when query and key have different shapes.
+            self.inductor_compile_config["combo_kernels"] = True
+            self.inductor_compile_config["benchmark_combo_kernel"] = True
```
mgoin, line 509: I would appreciate a doc pointer to how this works so I can understand for future work. Currently this is very opaque

BoyuanFeng: we have some [doc](https://github.com/pytorch/pytorch/blob/main/torch/_inductor/config.py#L735-L739) here. I will followup with more pytorch docs.

ProExpertProg (approved):
Yolo here seems acceptable we can add tests in the future

mgoin (approved):

PR #26695, opened 2025-10-13
[Misc] rename torch_dtype to dtype

### LINKED ISSUE #26293: [RFC]: replace deprecated torch_dtype by dtype
### Motivation.

Huggingface recently updated the argument from `torch_dtype` to a simple `dtype` to be coherent with PyTorch (see PR [here](https://github.com/huggingface/transformers/pull/39782)). 

Although their implementation is backward compatible, we could consider also making the change `torch_dtype` -> `dtype` to not trigger the warning:

```
`torch_dtype` is deprecated! Use `dtype` instead!
```
The warning is triggered for every call that involves huggingface (even if it is just for loading the model). 

For example:
```
vllm bench latency --model "any_model_from_hf_hub" . . .
```
triggers the above warning. 

### Proposed Change.

Following changes are suggested: 
- Replace `torch_dtype` with `dtype` to silence above warning and be prepared to when `torch_dtype` is fully removed in hf. 
- Ensure backwards compatibility from vLLM user perspective. 
- (Optional) add a test for backwards compatibility. 

### Feedback Period.

_No response_

### CC List.

_No response_

### Any Other Things.

_No response_

wangxiyuan:
## Purpose

Now, the log ``torch_dtype` is deprecated! Use `dtype` instead!`` is printed once running vLLM. it because  that `torch_dtype` has been deprecated from transformers 4.56.0. This PR replace it to `dtype` to clean up the noice log.

https://github.com/huggingface/transformers/commit/d8f6d3790acdd7033e199875b5b57691adf10735

FIX #26293 

## Test Result

---
<details>

</details>


vllm/entrypoints/llm.py
```diff
@@ -117,9 +117,9 @@ class LLM:
             execution with tensor parallelism.
         dtype: The data type for the model weights and activations. Currently,
             we support `float32`, `float16`, and `bfloat16`. If `auto`, we use
-            the `torch_dtype` attribute specified in the model config file.
-            However, if the `torch_dtype` in the config is `float32`, we will
-            use `float16` instead.
+            the `torch_dtype` or `dtype` attribute specified in the model config
```
ProExpertProg: Is this valid? I thought we just use style in the code now - or does the `transformers` package automatically looks at both for compatibility?

wangxiyuan: both of them work now. https://github.com/huggingface/transformers/blob/d7c9fbdb64842c541eebeaf0b11d1883ccf10be4/src/transformers/configuration_utils.py#L415-L423

hmellor (changes requested):
In order to make this change you must update the minimum required transformers version to `4.56.1`

vllm/config/model.py
mgoin, line 1: Should we still search for both dtype and torch_dtype here if some models use it?

hmellor: We don't need to. `torch_dtype` was a field of `PretrainedConfig` so the backward compatibility applies to all models (regardless of what is in their `config.json`)

https://github.com/huggingface/transformers/blob/e7d351cebad5f6dcdd169b0c034fdee0a000e6a9/src/transformers/configuration_utils.py#L264-L281

mgoin (approved):
Nice, I've been annoyed by this log for a while

tests/conftest.py
```diff
@@ -342,7 +342,7 @@ def _init(
         )
 
         model_kwargs = model_kwargs if model_kwargs is not None else {}
-        model_kwargs.setdefault("torch_dtype", torch_dtype)
+        model_kwargs.setdefault("dtype", torch_dtype)
```
DarkLight1337: Change the init argument name as well?

wangxiyuan: Done

DarkLight1337 (approved):

hmellor (approved):

AbdulmalikDS (commented):
image

PR #23734, opened 2025-08-27
[Feature] Support Decode Context Parallel (DCP) for MLA

youzhedian:
This PR adds Decode Context Parallel (DCP) support for MLA inference, fully compatible with chunked prefill and APC. 

**You can enable DCP with `--decode-context-parallel-size/-dcp xxx` (only support flashmla backend now), and tp_size needs to be divisible by dcp_size**, because the world size does not change by dcp, it simply reuse the GPUs of TP group, and split one TP group into tp_size//dcp_size DCP groups. e.g.
```
with -tp 8 -dcp 8 , we use 8 GPUs
with -tp 8 -dcp 4 , we use 8 GPUs
with -tp 4 -dcp 4 -pp 2 , we use 8 GPUs

and kvcache token budget always increased by `dcp` times.
```

This DCP implement store kvcache with an **interleave style**, the kvcache for **the token whose `token_idx` is `i` is always stored on the GPU whose `dcp_rank` equals to `i % dcp_world_size`**:   
```
e.g. DCP2, req with prompt_len=5, generation_len=4:
kvcache store in dcp_rank0: 0, 2, 4, 6, 8 
kvcache store in dcp_rank1: 1, 3, 5, 7,
```

deepseek-ai/DeepSeek-V2-Lite-Chat gsm8k eval:
```
- TP4PP2DCP4
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.6687|±  | 0.013|
|     |       |strict-match    |     5|exact_match|↑  |0.6657|±  | 0.013|

-  TP4PP2DCP1
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.6657|±  | 0.013|
|     |       |strict-match    |     5|exact_match|↑  |0.6611|±  | 0.013|
```

more info pls ref [introduce Doc](https://docs.google.com/document/d/1C93PuBSJmq8eUM16kD7CftzYlbNJ9j6Ut2W2kVuXgo4/edit?tab=t.0#heading=h.bc5zpep7g7ln)

Future work (These items will be tackled in follow-up PRs; community contributions are warmly welcomed.):
[...]

PR #26696, opened 2025-10-13
[DCP] Support dcp kv_cache interleave size > 1

zhangsicheng5:
### 1. cp_kv_cache_interleave_size support
In dcp scenario, kv_cache is split across dcp ranks, current implementation ([#23734](https://github.com/vllm-project/vllm/pull/23734)) split kv_cache with a token-level interleave style: token_idx i is stored on GPU whose dcp_rank == i % dcp_world_size.

For the convenience of pd disaggregate support, we add the cp_kv_cache_interleave_size argument to control the interleave size of kv_cache split size: store interleave_size tokens on dcp i, then store next interleave_size tokens on dcp i+1. The default value of cp_kv_cache_interleave_size is 1, which is same as original token-level interleave implementation. By setting cp_kv_cache_interleave_size to block_size, we can split kv_cache with a block-level interleave style, and makes it easy to support pd disaggregate with dcp > 1: D nodes only need to pull the corresponding kv_cache blocks, without need to rearange tokens in blocks.

Only dcp with cp_kv_cache_interleave_size is supported now, but the case of (p)cp is also considered and is easy to extend in the future.

### 2. Move dcp_local_seq_lens computation to utils
Move dcp_local_seq_lens computation to utils and pass it by metadata, so other attn backends can reuse it.

## Test Plan
Model: DeepSeek-V2-Lite-Chat
Dataset: gsm8k
```python
vllm serve DeepSeek-V2-Lite-Chat --gpu-memory-utilization 0.9 --tensor-parallel-size 2 --decode-context-parallel-size 2 --cp-kv-cache-interleave-size 64
```

## Test Result
tp2 dcp2, original code
| dataset | version | metric | mode | vllm-api-stream-chat |
| gsm8k | 7cd45e | accuracy | gen | 67.85 |

tp2 dcp2, interleave_size = 1
| dataset | version | metric | mode | vllm-api-stream-chat |
| gsm8k | 7cd45e | accuracy | gen | 67.85 |

tp2 dcp2, interleave_size = 64
| dataset | version | metric | mode | vllm-api-stream-chat |
| gsm8k | 7cd45e | accuracy | gen | 67.55 |

---
<details>

</details>


vllm/engine/arg_utils.py
```diff
@@ -1480,6 +1486,10 @@ def create_engine_config(
             enable_chunked_prefill=self.enable_chunked_prefill,
             disable_log_stats=self.disable_log_stats,
         )
+        if speculative_config is not None and self.cp_kv_cache_interleave_size != 1:
+            raise ValueError(
+                "MTP with cp_kv_cache_interleave_size > 1 is not supported now."
+            )
```
hmellor: Can this config live in `VllmConfig` with the other validation?

tests/distributed/test_context_parallel.py
```diff
@@ -52,6 +53,7 @@ def detailed(
         tp_base: int = 4,
         pp_base: int = 1,
         dcp_base: int = 1,
+        cp_kv_cache_interleave_size: int = 1,
```
LucasWilkinson: please call this `dcp_kv_cache_interleave_size`

vllm/v1/worker/block_table.py
```diff
@@ -119,7 +119,10 @@ def swap_row(self, src: int, tgt: int) -> None:
         self.block_table.np[src_tgt] = self.block_table.np[tgt_src]
 
     def compute_slot_mapping(
-        self, req_indices: np.ndarray, positions: np.ndarray
+        self,
+        req_indices: np.ndarray,
+        positions: np.ndarray,
+        cp_kv_cache_interleave_size: int = 1,
```
LucasWilkinson: since this is a constant pass it via the init

vllm/utils/__init__.py
```diff
@@ -3426,3 +3426,35 @@ def unique_filepath(fn: Callable[[int], Path]) -> Path:
         if not p.exists():
             return p
         i += 1
+
+
+def get_dcp_local_seq_lens(
```
LucasWilkinson: we should find a better spot for this; this is too broad of a utils file for a feature specific utility

vllm/engine/arg_utils.py
zhangsicheng5: Thanks for review, we have moved this validation to VllmConfig. Looking forward to your future reviews

LucasWilkinson (changes requested):
Thanks for your contribution! Requesting changes to prevent merging until the test results can be obtained 👍

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1276,6 +1287,14 @@ def _prepare_inputs(
                 logits_indices
             )
 
+        # update seq_lens of decode reqs under DCP.
+        if self.dcp_world_size > 1:
+            self.dcp_local_seq_lens.gpu[:num_reqs] = get_dcp_local_seq_lens(
```
LucasWilkinson: I think it might actually be better to compute `get_dcp_local_seq_lens` using host buffers and then do a non-blocking copy to  `self.dcp_local_seq_lens.gpu` (see: `CpuGpuBuffer.copy_to_gpu`)

LucasWilkinson: (then when async scheduling is enabled it will be overlapped)

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -256,6 +258,11 @@ def __init__(
         self.is_multimodal_pruning_enabled = False
         self.max_model_len = model_config.max_model_len
         self.dcp_world_size = self.parallel_config.decode_context_parallel_size
+        try:
+            self.dcp_rank = get_dcp_group().rank_in_group
```
youzhedian: delay to `get_dcp_local_seq_lens ` calling is better?

vllm/v1/worker/block_table.py
youzhedian: nit: maybe no default val is better

youzhedian (approved):
LGTM.

tests/distributed/test_context_parallel.py
zhangsicheng5: After prefill cp ([#25852](https://github.com/vllm-project/vllm/pull/25852/)) is supported, this kv_cache_interleave_size will be used for both dcp and pcp, shall we keep this name for future usage?

vllm/v1/worker/block_table.py
zhangsicheng5: Thanks for review, we have already passed it via init

vllm/utils/__init__.py
zhangsicheng5: Now we put this function in `vllm/v1/attention/backends/utils.py`, same place as `CommonAttentionMetadata.dcp_local_seq_lens` definition, this should be a more appropriate spot

vllm/v1/worker/gpu_model_runner.py
zhangsicheng5: Modified as suggested, thanks for review

vllm/v1/worker/block_table.py
zhangsicheng5: Thanks for review, now we pass this arg via init, since it's a constant

vllm/v1/worker/gpu_model_runner.py
zhangsicheng5: In some cases we might need to know how seq_len is split globally, instead of only local seq_len on current dcp_rank, for example in [our current npu mla impl](https://github.com/LookAround0301/vllm-ascend/blob/long_seq_dev/vllm_ascend/attention/mla_v1.py#L1319), we need the global seq_len split message to calculate a mask for following update_lse (if no kv_cache is stored on some (d)cp_ranks, then there's no need to do corresponding update_lse), so we think it's better to return the full seq_len split result from `get_dcp_local_seq_lens`, and each dcp_rank can select their corresponding part as needed.

vllm/v1/attention/backends/utils.py
```diff
@@ -992,3 +992,35 @@ def compute_causal_conv1d_metadata(query_start_loc_p: torch.Tensor):
         nums_dict[BLOCK_M]["token_chunk_offset_ptr"] = token_chunk_offset_ptr  # type: ignore
 
     return nums_dict, batch_ptr, token_chunk_offset_ptr
+
+
+def get_dcp_local_seq_lens(
```
LucasWilkinson, line 1081: nit: can you simplify this and have it only compute for the current dcp_rank? and pass in dcp rank

tests/distributed/test_context_parallel.py
LucasWilkinson: sure; by that logic we should update `dcp_local_seq_lens` to `cp_local_seq_lens` too but we can do that in the pcp PR

vllm/v1/worker/gpu_model_runner.py
LucasWilkinson: nit: I think we can simplify this to:
```
self.dcp_rank = 0 if self.dcp_world_size <= 1 else get_dcp_group().rank_in_group
```

that way we'll still get the benefit of the assert in `get_dcp_group()` an if a test sets `self.dcp_world_size > 1` it should be initializing the dcp group anyways

LucasWilkinson (approved):
Overall looks good to me! Thanks for the contribution, left a few nits.

vllm/v1/worker/gpu_model_runner.py
zhangsicheng5: better way to get dcp_rank 👍 Modified as suggested

vllm/v1/attention/backends/utils.py
zhangsicheng5: Same reason as [this comment](https://github.com/vllm-project/vllm/pull/26696#discussion_r2428390768), it might be more flexible to return the full seq_len split result and each dcp_rank can select their own part as needed. Since the size of `seq_len` is only `max_num_reqs`, multiply it by `dcp_size` will increase little computation/storage overhead and we think it should be acceptable.

FENP:
Does `dcp_kv_cache_interleave_size` affect `reorg_kvcache`, especially the calculation of `cp_target_rank`?
https://github.com/vllm-project/vllm/blob/e20eba753bbced43837aa92f747e6c50ee36ce09/vllm/v1/attention/backends/mla/common.py#L976-L1042

LucasWilkinson (changes requested):
Great catch @FENP ; please address: https://github.com/vllm-project/vllm/pull/26696#issuecomment-3414535257

minosfuture:
> > Why not use `common_attn_metadata.dcp_local_seq_lens` for flash_attn? might be related to the `- query_kv_lens_cpu` part of `dcp_context_kv_lens_cpu = seq_lens_cpu - query_kv_lens_cpu`
> 
> @LucasWilkinson yes, `common_attn_metadata.dcp_local_seq_lens` is about the whole seq with both context and current chunks, but we only need the context chunks here.
> 
> I have thoroughly examined the subsequent usage of `dcp_context_kv_lens` but was unable to dive deeper into the operator-side analysis. I believe we need to cc some experts proficient in the logic of the `flash_attn` operator to help explain the meaning of `seqused_k` and identify the root cause of the issue. I would greatly appreciate assistance from specialists in this area.

seqused_k is the whole kv length (context len + query len).  We need to fix the usage in _forward_with_dcp.
I double checked we correctly used kv len in code path of MLA and non-dcp attention.

vllm/v1/attention/backends/mla/common.py
```diff
@@ -995,27 +993,32 @@ def reorg_kvcache(
...
-        cp_target_rank = (chunk_context_len - 1) % cp_world_size
+
         cur_seq_len = 0
         for rank in range(cp_world_size):
-            if rank > cp_target_rank and cp_chunk_seq_len:
-                real_cp_chunk_seq_len = cp_chunk_seq_len - 1
-            else:
-                real_cp_chunk_seq_len = cp_chunk_seq_len
-            if real_cp_chunk_seq_len:
+            real_cp_chunk_seq_len = local_context_lens[rank]
```
LucasWilkinson: nit: can we make this naming consistent; i.e. `real_cp_chunk_seq_len` -> `local_contex_len`; also appears that we went from using `cp_chunk_seq_len` to `origin_context_lens`; can you please explain the naming here?

vllm/v1/attention/backends/mla/common.py
```diff
@@ -995,27 +993,32 @@ def reorg_kvcache(
...
+        interleave_size: Interleave size of kv_cache storage.
     """
     kv_c_segments = []
     k_pe_segments = []
     src_token_idx = 0
     max_seq_len_check = 0
-    for cp_chunk_seq_len, origin_context_len in zip(
-        cp_chunk_seq_lens_lst, origin_context_lens
+    local_context_lens_allrank = get_dcp_local_seq_lens(
+        torch.Tensor(origin_context_lens),
```
LucasWilkinson: why do we need to construct a tensor here? we should really avoid constructing tensors inside the forward pass. This should be moved to the metadata building

vllm/v1/attention/backends/utils.py
LucasWilkinson: sure; this seems acceptable :+1:

vllm/v1/attention/backends/mla/common.py
pisceskkk: Yes, I have renamed some variables according to the following conventions:

1. For original lengths ​​not partitioned by DCP​​, retain the original names without any prefix.
2. For lengths specific to a particular DCP rank, use the ​​`local_`​​ prefix (as opposed to a potential `​​global_`​​ prefix).
3. For different types of lengths, clarify whether they refer to ​​`chunk`, `​​context`, or ​​whole `seq`.

Please refer to the diagram.
dcp block

vllm/v1/attention/backends/mla/common.py
pisceskkk: sure, I move this tensor to metadata building and clean some unused codes and params.

vllm/v1/attention/backends/mla/common.py
```diff
@@ -851,32 +838,52 @@ def build(
...
                 if self.dcp_world_size > 1:
+                    local_context_lens_allrank = get_dcp_local_seq_lens(
+                        context_lens_cpu,
+                        self.dcp_world_size,
+                        None,
+                        self.dcp_kv_cache_interleave_size,
+                    )
+                    # Note(qcs): The context lengths in the perspective of dcp rank0
+                    # padded to `dcp_kv_cache_interleave_size`.
+                    local_context_lens_cpu = (
```
LucasWilkinson: is this correct? im surprised this is dcp_rank agnostic, am i missing something? 
right now its 
```
local_context_lens_cpu = cdiv(context_lens_cpu, self.dcp_world_size * self.dcp_kv_cache_interleave_size) * self.dcp_kv_cache_interleave_size
```

but i would have thought it should be something like:
```
virtual_block_size = self.dcp_world_size * self.dcp_kv_cache_interleave_size
partial_virtual_block = context_lens_cpu % virtual_block_size
base = context_lens_cpu // virtual_block_size
extra = min(max(0, partial_virtual_block - self.dcp_rank * self.dcp_kv_cache_interleave_size),
            self.dcp_kv_cache_interleave_size)
local_context_lens_cpu = base + extra
```

(also can we use cdiv here to make more readable?, i.e. `from vllm.utils import cdiv`)

vllm/v1/attention/backends/mla/common.py
```diff
@@ -851,32 +838,52 @@ def build(
...
                     # be divisible by dcp_world_size, because DCP use
                     # cp_gather_cache which not require `cp_chunk_starts`
                     # aligned to page_size.
                     assert max_context_chunk % self.dcp_world_size == 0
-                    cp_max_context_chunk = max_context_chunk // self.dcp_world_size
-                    cp_chunk_starts = (
+                    local_max_context_chunk = (
+                        max_context_chunk
+                        // (self.dcp_world_size * self.dcp_kv_cache_interleave_size)
+                    ) * self.dcp_kv_cache_interleave_size
```
LucasWilkinson: nit: ` self.dcp_world_size * self.dcp_kv_cache_interleave_size` is used alot; maybe we can add something like
```
self.dcp_virtual_block_size = self.dcp_world_size * self.dcp_kv_cache_interleave_size
self.dcp_local_block_size = self.dcp_kv_cache_interleave_size
```

then this could be shortened to:
```
local_max_context_chunk = (max_context_chunk // self.dcp_virtual_block_size) * self.dcp_local_block_size
```

LucasWilkinson (commented):
Overall looking really good! Thanks for all the cleanup; couple final comments

vllm/v1/attention/backends/mla/common.py
LucasWilkinson: thank you!!! this makes it so much more readable! my head hurts alot less 😄

vllm/v1/attention/backends/mla/common.py
pisceskkk: make sense, changed

vllm/v1/attention/backends/mla/common.py
pisceskkk: I've changed this comment to: "The max local context lengths padded to dcp_local_block_size."

> is this correct? im surprised this is dcp_rank agnostic, am i missing something? 

Yes, this value is dcp_rank agnostic. The reason I initially mentioned rank 0 is because in the current implementation, the sequence on rank 0 is typically the longest by default. Replacing that description with "maximum length" is indeed more accurate.

LucasWilkinson: won't this over gather on everything other than rank 0? previously there was this correction code in `reorg_kvcache` to correct to the current rank

```
            if rank > cp_target_rank and cp_chunk_seq_len:
                real_cp_chunk_seq_len = cp_chunk_seq_len - 1
            else:
                real_cp_chunk_seq_len = cp_chunk_seq_len
```

also this not the rank 0 context len any more since we could partial block on rank 0, previously:

```
            # Note(hc): The context lengths in the perspective of dcp rank0.
            cp_context_lens_cpu = torch.ceil(
                context_lens_cpu.float() / self.dcp_world_size
            ).int()
```
was correct because with an interleave size 1 there could never be a partial block on rank 1

am i missing something?

pisceskkk: sorry for delay, we've identified another bug and are working on a fix.

> won't this over gather on everything other than rank 0? 

Yes, the current approach would cause over-gathering on other ranks. We've changed `local_context_lens_cpu` to accurately reflect the KVCache length on each device to avoid unnecessary gathering (1a94cb75fb98474df54ba64dd21a45d114af5a37). 
However, one question remains: are the performance of `all_gather` and `all_gatherv` consistent? To my knowledge, `all_gatherv` suffers from performance degradation on NPUs. I'm uncertain whether similar issues exist on GPUs or other devices.

**Test results**
DeepSeek-V2-Lite-Chat with FlashMLA
TP8DCP8 concurrency=256

interleave 1
| dataset | version | metric | mode | vllm-api-stream-chat |
| gsm8kdataset | - | accuracy | gen | 69.14 |

interleave 8
| dataset | version | metric | mode | vllm-api-stream-chat |
| gsm8kdataset | - | accuracy | gen | 69.37 |

interleave 64
| dataset | version | metric | mode | vllm-api-stream-chat |
| gsm8kdataset | - | accuracy | gen | 68.69 |

**new bug**

Regarding the newly discovered bug, it occurs in the **tritonMLA** backend and has existed prior to this PR (tested on 650b51f9f90dcd0393fe33a55e25bd2688874b25). Specifically, when testing with the gsm8k dataset, we observe some outputs consisting entirely of "!" as concurrency increases (appearing in 2-3 cases at concurrency=5, and dozens of cases at concurrency=256, leading to a 6% drop in accuracy). This indicates NaN values occurring during the prefill phase, though the root cause has not yet been identified. I think resolving this issue may require significant time for fixes, and plan to address it in a subsequent PR.
CC @youzhedian

LucasWilkinson: got it; makes sense! You are correct we should go with the allgather; that was my bad 🤦 I didn't realize `reorg_kvcache` was unpadding. I think we need to improve the readability of this padded all-gather then un-padding logic (apologies this is basically tech debt from the orignal DCP PR not you; i was never-very-happy / totally-understood the reorg_kvcache code and was hoping to refactor it but haven't had a chance)

got it; makes sense, I understand now, good call lets go back to the all_gather approach but try to make it more readable to avoid confusions. Some suggestions:

`local_max_context_chunk` -> `padded_local_max_context_chunk_across_ranks`
`local_chunk_seq_lens` -> `padded_local_chunk_seq_lens`
`local_cu_chunk_seq_lens_cpu` -> `padded_local_cu_chunk_seq_lens_cpu`
`local_cu_seq_lens` -> `padded_local_cu_seq_lens`
`local_context_lens_allrank` -> `local_context_lens_allranks`

`reorg_kvcache` <- Add a comment indicating this removes padding added to make the all-gather a fixed size across ranks or maybe rename `reorg_kvcache` -> `unpad_kvcache`

pisceskkk: Indeed, it took us considerable effort to understand this function initially. I've updated the comments and added examples for clarity.

Sure, I've renamed those variables based on your feedback. Thanks for the review

LucasWilkinson (approved):
LGTM! Thanks for all the hardwork and sorry about the back and forth

we should refactor `reorg_kvcache` at somepoint (maybe use a triton kernel) but that can be done in the future; i appreciate the clarify comments/renaming!

Butterfingrz:
Hello! I have a question: If I am not using PD disaggregation—specifically when deploying an MLA-based model with DCP8—would setting cp_kv_cache_interleave_size to 64 yield any performance gains compared to the default of 1?

PR #26698, opened 2025-10-13
[CI][Release][Arm64]: Build arm64 release for gpu arch 8.9

cyb70289:
## Purpose
Support RTX-40xx card (arch=8.9) for Arm64 release.

## Test Result
Verified manually on Arm Neoverse-N2 server with RTX-4090 card.


mgoin (approved):
Thanks, this makes sense as a PCIe card

Issue #26703, opened 2025-10-13
[RFC]: [gpt-oss] Supporting Any MCP Server Tools


alecsolder:
### Motivation.

There are three main motivations:
- vLLM should support integrating with all MCP servers
  - TODO: Test with enterprise ones like Github
- Clean up gpt-oss tool specific code paths in vLLM
  - Tools and models change, but MCP is a protocol meant to support all models
  - Sets the stage for full MCP integration for all tool calling models
- Avoid losing information due to lossy OpenAI types for output
  - An example below

```
# What you can return to the user on the ResponsesAPI for this tool call with the OpenAI types.
class ActionOpenPage(TypedDict, total=False):
    type: Required[Literal["open_page"]]
    url: Required[str]
# The interface for the actual tool call the model does
open_link(ctx: Context,
    id: Union[int, str] = -1,
    cursor: int = -1,
    loc: int = -1,
    num_lines: int = -1,
    view_source: bool = False,
    source: Optional[str] = None)
```

This feature is similar to the concept of “connectors” because we still don’t support users specifying arbitrary MCP servers by URL yet, the list is static and owned by the server, which is more similar to the concept of [OpenAI MCP Connectors](https://platform.openai.com/docs/guides/tools-connectors-mcp#quickstart).

### Tool Namespaces
- For gpt-oss, a namespace represents a group of tools. For OpenAI Harmony, some namespaces get treated in unique ways 
- A MCP server has a collection of tools. The MCP server can be identified by its **serverInfo.name** 
- For the ResponsesAPI, the MCP Tool type has a **server_label** field to identify the MCP server.

In the codebase these were three different concepts, but by using it as a foreign key we simplify the codebase and still provide needed functionality. MCP servers and the Responses API are generic and able to support all models, you should be able to use the same MCP server for Deepseek and gpt-oss. 

As RL for tools gets more popular, more models will have special case behavior for tools. Different fine tunes of the same model may have different special case behavior for tools. So this special case handling has been moved from hard coded paths to be based on **namespaces**. 

These are defined by the server/model and I’ll refer to namespaces that need special case handling as **elevated**. (Needs a new name)

There is currently special case code in vLLM to treat the container namespace as elevated. Other use cases may not want the same thing. So now, one deployment can elevate the namespace with an env variable, and the other deployment doesn’t need to. 

### Standardize around MCP 
[...]

PR #25724, opened 2025-09-25
[responsesAPI] add better error messaging for long prompts

qandrew:
## Purpose
We notice these max_tokens wrong value errors started happening last week. Error message example

"ClientException(message=\"Engine encountered error for request: error=ErrorInfo(message='max_tokens must be at least 1, got -12806.', type='BadRequestError', param=None, code=400)\")", type_: Unknown })"

This PR returns a more descriptive errors and returns an error that says that prompt is larger than max_model_length, as current error message is very confusing.

## Test Plan
Send a request with a long engine prompt
added a unit test

## Test Result
`{"error":{"message":"The engine prompt length 67 exceeds the max_model_len 131072. Please reduce prompt.","type":"invalid_request_error","param":null,"code":400}} `

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #26704, opened 2025-10-13
[Frontend][gpt-oss] Support all MCP servers for gpt-oss

alecsolder:
## Purpose
This PR is an implementation of #26703 

Note: This PR is really big, but there was a lot that had to be cleaned up at the same time. A large amount of the code is from adding tests to existing code and a MCP server to test with and new MCP tests.

There are three main motivations:
- vLLM should support integrating with all MCP servers
  - TODO: Test with enterprise ones like Github
- Clean up gpt-oss tool specific code paths in vLLM
  - Tools and models change, but MCP is a protocol meant to support all models
  - Sets the stage for full MCP integration for all tool calling models
- Avoid losing information due to lossy OpenAI types for output
  - An example below

```
# What you can return to the user on the ResponsesAPI for this tool call with the OpenAI types.
class ActionOpenPage(TypedDict, total=False):
    type: Required[Literal["open_page"]]
    url: Required[str]
# The interface for the actual tool call the model does
open_link(ctx: Context,
    id: Union[int, str] = -1,
    cursor: int = -1,
    loc: int = -1,
    num_lines: int = -1,
    view_source: bool = False,
    source: Optional[str] = None)
```

This feature is similar to the concept of “connectors” because we still don’t support users specifying arbitrary MCP servers by URL yet, the list is static and owned by the server, which is more similar to the concept of [OpenAI MCP Connectors](https://platform.openai.com/docs/guides/tools-connectors-mcp#quickstart).

## Test Plan
Run the provided simple Memory MCP server using:
```
python tests/entrypoints/openai/memory_mcp_server.py 
# 8765 is the default port
```
Then register it with your vLLM responses API server with:
```
CUDA_VISIBLE_DEVICES=6 vllm serve openai/gpt-oss-20b --tool-server=localhost:8765
```
And then try it out with something like:
```
curl http://localhost:8000/v1/responses \
  -H "Content-Type: application/json" -N \
  -d '{
        "input": "Show me that the memory tool works.",
        "tools": [
          {
            "type": "mcp",
            "server_label": "memory", 
            "server_url": "ignored"
          }
        ],
        "enable_response_messages": true
      }'
```

## Test Result
Passed


vllm/entrypoints/openai/serving_responses.py
```diff
@@ -594,6 +599,8 @@ async def responses_full_generator(
                 tool_output_tokens=num_tool_output_tokens,
             ),
         )
+        print("INPUT MESSAGES \n\n", input_messages)
+        print("\n\nOUTPUT MESSAGES \n\n", output_messages)
```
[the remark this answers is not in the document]
alecsolder: thank you gemini :)

tests/entrypoints/openai/memory_mcp_server.py
```diff
@@ -0,0 +1,183 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""
+Standalone Memory MCP Server
```
qandrew, line 5: nit: this file should go in the mcp folder

tests/entrypoints/openai/test_response_api_mcp_tools_elevated.py
```diff
@@ -0,0 +1,143 @@
...
+@pytest.mark.parametrize("model_name", [MODEL_NAME])
+async def test_memory_mcp_elevated(memory_elevated_client, model_name: str):
+    """Test Memory MCP tool as elevated.
+
+    When memory IS in GPT_OSS_SYSTEM_TOOL_MCP_LABELS:
+    - Tool should be in system message (not developer message)
+    - Tool calls should be on analysis channel
+    - Tool responses should be on analysis channel
+    """
+    response = await memory_elevated_client.responses.create(
```
qandrew, line 75: i think i general we shouldn't need to do E2E tests for all functionality testing, ideally we can mock the client/server and reduce CI load. You might run into CI memory issues (i had some here https://github.com/vllm-project/vllm/pull/25724/commits/eb6dd74d025c2fcc5e5c6e0114881ac3e14bcdd3) by spinning up more servers within the same CI step

qandrew (commented):
thanks for putting this together! However the PR is a bit large to review thoroughly, can you split into smaller pieces? if you have 3 main motivations, can you split them into 3 separate PRs? a 4th pr could add memory_mcp_server, test_memory_mcp_server logic too

tests/entrypoints/openai/test_response_api_mcp_tools_custom.py
```diff
@@ -0,0 +1,214 @@
...
+@pytest.mark.asyncio
+@pytest.mark.parametrize("model_name", [MODEL_NAME])
+async def test_memory_mcp_custom(memory_custom_client, model_name: str):
+    """Test Memory MCP tool as custom (not elevated).
+
+    When memory is NOT in GPT_OSS_SYSTEM_TOOL_MCP_LABELS:
+    - Tool should be in developer message (not system message)
+    - Tool calls should be on commentary channel
+    - Tool responses should be on commentary channel
+    """
```
bbrowning, line 76: These comments (and the similar ones for the elevated test case) are the most concise description of custom tools vs elevated/system/builtin (naming is hard) tools that I've seen, and help me really understand the difference. And as someone that has studied the Harmony format, I immediately understand the implications of this.

It feels like we need some user-facing documentation about this as well, with something to help the user know if they want their tools to be in `GPT_OSS_SYSTEM_TOOL_MCP_LABELS` or not and what the implications of that are on the tool calls generated or on the execution of those tool calls.

StevenSong:
finding this PR after being very confused by the documentation: https://docs.vllm.ai/projects/recipes/en/latest/OpenAI/GPT-OSS.html#tool-use
> Please note that the default options are simply for demo purposes. For production usage, vLLM itself can act as MCP client to multiple services. [...] The URLs are expected to be MCP SSE servers that implement instructions in server info and well documented tools. The tools will be injected into the system prompt for the model to enable them.

and the original announcement blog: https://blog.vllm.ai/2025/08/05/gpt-oss.html
> Alternatively, users can launch an MCP-compliant external tool server, to let vLLM use the tool server instead of directly leveraging the gpt-oss toolkit. This modular architecture simplifies the creation of scalable tool-calling libraries and services, requiring no internal changes to vLLM.

It had sounded like the support for arbitrary MCP servers with `gpt-oss` had already been enabled, but the instructions (explicitly mentioning the `VLLM_GPT_OSS_SYSTEM_TOOL_MCP_LABELS` env var) in this example (https://docs.vllm.ai/en/stable/examples/online_serving/openai_responses_client_with_mcp_tools/) finally led me to these pieces in the responses endpoint: 

https://github.com/vllm-project/vllm/blob/6283021142bbf5ee324395dad5e80b8661400329/vllm/entrypoints/openai/responses/utils.py#L223-L237

and

https://github.com/vllm-project/vllm/blob/6283021142bbf5ee324395dad5e80b8661400329/vllm/entrypoints/openai/responses/serving.py#L403-L424

(and that the env var restricts values to only those 3 built-ins given the custom logic to handle those tools)

Based on all this, it seems like currently, only the built-in tools are supported for `gpt-oss`, hence the need for this PR? Would love to see this PR move forward, sad to see it's gone a bit stale, but hopefully this comment can at least help clarify the current status for anyone else looking to bring other MCP servers to use with `gpt-oss` in vLLM.

for my use local use case, will just try to wrap the mcp tools and make my own agent loop to handle them

PR #18969, opened 2025-05-30
[P/D] NixlConnector use cache device index for memory registration

ptarasiewiczNV:
Currently `NixlConnecotr` uses `rank` to specify device for NIXL memory registration. This doesn't work in multi-node case. This PR changes that to use torch device index of the cache directly.

PR #26709, opened 2025-10-13
[Misc] Remove use of CUDA_VISIBLE_DEVICES for device selection (fix DP slow startup time &c)

ilmarkov:
Warmup nccl allreduce (in `pynccl.py`) are quite slow in DP setup (up to 60-180 seconds for 1 element allreduce). This is caused by the setup of DP processes where multiple workers have the same cuda device id (with different `CUDA_VISIBLE_DEVICES` setting per node). This confuses NCCL and leads to slow startup times.

In this PR we get rid of `CUDA_VISIBLE_DEVICES` setting in `CoreEngineProcManager`, only use `torch.cuda.set_device` for proper DP setup.

## Test Result

---
<details>

</details>


vllm/v1/engine/utils.py
```diff
@@ -130,15 +130,9 @@ def __init__(
 
         self._finalizer = weakref.finalize(self, shutdown, self.processes)
 
-        data_parallel = vllm_config.parallel_config.data_parallel_size > 1
         try:
-            for proc, local_dp_rank in zip(self.processes, local_dp_ranks):
-                with (
-                    set_device_control_env_var(vllm_config, local_dp_rank)
```
tlrmchlsmth, line 138: Is `set_device_control_env_var` still used anywhere? Or could it be removed now?

vllm/distributed/parallel_state.py
```diff
@@ -1193,6 +1193,7 @@ def init_distributed_environment(
         # setting, where we can use rank as local rank
         local_rank = envs.LOCAL_RANK if distributed_init_method == "env://" else rank
     global _WORLD, _NODE_COUNT
+
```
SageMoore: Nit: remove new line

vllm/distributed/device_communicators/cuda_communicator.py
```diff
@@ -55,6 +55,7 @@ def __init__(
 
         self.pynccl_comm: PyNcclCommunicator | None = None
         if self.world_size > 1:
+            print(f"Creating PyNcclCommunicator for unique_name: {self.unique_name}")
```
SageMoore: Nit: Remove print

SageMoore (commented):
Thanks for fixing this @ilmarkov!

vllm/executor/uniproc_executor.py
```diff
@@ -53,6 +53,8 @@ def _distributed_args(self) -> tuple[str, int, int]:
         # set local rank as the device index if specified
         device_info = self.vllm_config.device_config.device.__str__().split(":")
         local_rank = int(device_info[1]) if len(device_info) > 1 else 0
+        if self.vllm_config.parallel_config.data_parallel_size > 1:
+            local_rank = self.vllm_config.parallel_config.data_parallel_rank_local
```
njhill: Should we do this only in the case that the device isn't specified explicitly in the config (i.e. `len(device_info) <= 1` above)?

vllm/v1/engine/utils.py
njhill: This is also for non-cuda platforms in which case it may be a different variable to `CUDA_VISIBLE_DEVICES`. I think we may need to keep the existing behaviour for those?

njhill (commented):
Thanks @ilmarkov ...

As well as the inline questions, I think we may still need tol honor the `current_platform.device_id_to_physical_device_id` mapping as is currently done by `set_device_control_env_var()`, via `get_device_indices()`.

This is to properly handle cases where folks may run vllm with e.g. `CUDA_VISIBLE_DEVICES=2,4,7,8` .. we should map the dp local ranks 0-4 onto these.

ilmarkov:
@njhill 
> This is to properly handle cases where folks may run vllm with e.g. CUDA_VISIBLE_DEVICES=2,4,7,8 .. we should map the > dp local ranks 0-4 onto these.

In this case the env variable `CUDA_VISIBLE_DEVICES` will already be set and `torch.cuda.set_device(id)` will do the mapping accordingly.

smarterclayton:
I'm seeing a NIXL crash when I try to use this configured for kv transfer:

```
(Worker_DP4_EP4 pid=2261) ERROR 10-15 10:51:19 [v1/executor/multiproc_executor.py:695] Traceback (most recent call last):
(Worker_DP4_EP4 pid=2261) ERROR 10-15 10:51:19 [v1/executor/multiproc_executor.py:695]   File "/app/vllm/vllm/v1/executor/multiproc_executor.py", line 690, in worker_busy_loop
(Worker_DP4_EP4 pid=2261) ERROR 10-15 10:51:19 [v1/executor/multiproc_executor.py:695]     output = func(*args, **kwargs)
(Worker_DP4_EP4 pid=2261) ERROR 10-15 10:51:19 [v1/executor/multiproc_executor.py:695]              ^^^^^^^^^^^^^^^^^^^^^
(Worker_DP4_EP4 pid=2261) ERROR 10-15 10:51:19 [v1/executor/multiproc_executor.py:695]   File "/app/vllm/vllm/v1/worker/worker_base.py", line 327, in initialize_from_config
(Worker_DP4_EP4 pid=2261) ERROR 10-15 10:51:19 [v1/executor/multiproc_executor.py:695]     self.worker.initialize_from_config(kv_cache_config)  # type: ignore
(Worker_DP4_EP4 pid=2261) ERROR 10-15 10:51:19 [v1/executor/multiproc_executor.py:695]     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(Worker_DP4_EP4 pid=2261) ERROR 10-15 10:51:19 [v1/executor/multiproc_executor.py:695]   File "/app/vllm/vllm/v1/worker/gpu_worker.py", line 344, in initialize_from_config
(Worker_DP4_EP4 pid=2261) ERROR 10-15 10:51:19 [v1/executor/multiproc_executor.py:695]     self.model_runner.initialize_kv_cache(kv_cache_config)
(Worker_DP4_EP4 pid=2261) ERROR 10-15 10:51:19 [v1/executor/multiproc_executor.py:695]   File "/app/vllm/vllm/v1/worker/gpu_model_runner.py", line 4535, in initialize_kv_cache
(Worker_DP4_EP4 pid=2261) ERROR 10-15 10:51:19 [v1/executor/multiproc_executor.py:695]     kv_transfer_group.register_kv_caches(kv_caches)
(Worker_DP4_EP4 pid=2261) ERROR 10-15 10:51:19 [v1/executor/multiproc_executor.py:695]   File "/app/vllm/vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py", line 226, in register_kv_caches
(Worker_DP4_EP4 pid=2261) ERROR 10-15 10:51:19 [v1/executor/multiproc_executor.py:695]     self.connector_worker.register_kv_caches(kv_caches)
(Worker_DP4_EP4 pid=2261) ERROR 10-15 10:51:19 [v1/executor/multiproc_executor.py:695]   File "/app/vllm/vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py", line 894, in register_kv_caches
(Worker_DP4_EP4 pid=2261) ERROR 10-15 10:51:19 [v1/executor/multiproc_executor.py:695]     self.nixl_wrapper.register_memory(descs, backends=self.nixl_backends)
(Worker_DP4_EP4 pid=2261) ERROR 10-15 10:51:19 [v1/executor/multiproc_executor.py:695]   File "/app/venv/lib/python3.12/site-packages/nixl/_api.py", line 377, in register_memory
(Worker_DP4_EP4 pid=2261) ERROR 10-15 10:51:19 [v1/executor/multiproc_executor.py:695]     self.agent.registerMem(reg_descs, handle_list)
(Worker_DP4_EP4 pid=2261) ERROR 10-15 10:51:19 [v1/executor/multiproc_executor.py:695] nixl._bindings.nixlBackendError: NIXL_ERR_BACKEND
```

Happens on every start.  I know that there are interactions between NIXL and which GPUs are visible, does this mean visibility is not working or it is?

vllm/executor/uniproc_executor.py
```diff
@@ -53,6 +53,12 @@ def _distributed_args(self) -> tuple[str, int, int]:
         # set local rank as the device index if specified
         device_info = self.vllm_config.device_config.device.__str__().split(":")
         local_rank = int(device_info[1]) if len(device_info) > 1 else 0
+        if (
+            self.vllm_config.parallel_config.data_parallel_size > 1
+            and len(device_info) > 1
```
njhill: shouldn't this be `<=1`?

vllm/v1/engine/utils.py
```diff
@@ -129,13 +129,12 @@ def __init__(
             )
 
         self._finalizer = weakref.finalize(self, shutdown, self.processes)
-
```
njhill, line 133: nit: revert unrelated change

vllm/executor/uniproc_executor.py
njhill: As discussed offline I think the change here is actually redundant

vllm/v1/worker/gpu_worker.py
```diff
@@ -168,6 +168,13 @@ def init_device(self):
         if self.device_config.device.type == "cuda":
             # This env var set by Ray causes exceptions with graph building.
             os.environ.pop("NCCL_ASYNC_ERROR_HANDLING", None)
+            if self.vllm_config.parallel_config.data_parallel_size > 1:
+                self.local_rank = (
+                    self.parallel_config.data_parallel_rank_local
+                    * self.parallel_config.world_size
+                    + self.local_rank % self.parallel_config.world_size
+                ) % torch.cuda.device_count()  # type: ignore[has-type]
```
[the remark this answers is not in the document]
njhill: @ilmarkov could you double-check this?

njhill:
The default CI tests all passed ... I've unblocked a bunch of distributed tests, running now.

vllm/v1/engine/utils.py
```diff
@@ -135,7 +135,7 @@ def __init__(
             for proc, local_dp_rank in zip(self.processes, local_dp_ranks):
                 with (
                     set_device_control_env_var(vllm_config, local_dp_rank)
-                    if (data_parallel)
+                    if (data_parallel and not current_platform.is_cuda_alike())
```
njhill, line 142: @ilmarkov could you add a comment here why we set the env var for cuda but not for non-cuda?

vllm/v1/worker/gpu_worker.py
```diff
@@ -168,6 +168,27 @@ def init_device(self):
...
+                # Use local DP rank if available, otherwise use global DP rank.
+                data_parallel_local_rank: int = (
+                    self.parallel_config.data_parallel_rank_local
+                    if self.parallel_config.data_parallel_rank_local is not None
+                    else self.parallel_config.data_parallel_rank
+                ) % self.parallel_config.data_parallel_size_local
+                tp_pp_world_size = (
+                    self.parallel_config.pipeline_parallel_size
+                    * self.parallel_config.tensor_parallel_size
+                )
```
njhill, line 185: Can just use `self.parallel_config.world_size` for this

vllm/v1/worker/gpu_worker.py
```diff
@@ -168,6 +168,27 @@ def init_device(self):
...
+                # DP_LOCAL_RANK * TP_PP_WORLD_SIZE + TP_RANK % TP_PP_WORLD_SIZE
+
+                self.local_rank = (
+                    data_parallel_local_rank * tp_pp_world_size
+                    + self.local_rank % tp_pp_world_size  # type: ignore[has-type]
+                ) % torch.cuda.device_count()
+
```
tlrmchlsmth, line 192: This is a dense block of code and readability would be improved by factoring out things like `dp_rank`, `dp_rank_local`, `tp_size` into local variables.

Also could you consistently use `self.parallel_config` instead of mixing `self.parallel_config` and `self.vllm_config.parallel_config`?

vllm/v1/worker/gpu_worker.py
```diff
@@ -168,6 +168,27 @@ def init_device(self):
         if self.device_config.device.type == "cuda":
             # This env var set by Ray causes exceptions with graph building.
             os.environ.pop("NCCL_ASYNC_ERROR_HANDLING", None)
+            if (
+                self.vllm_config.parallel_config.data_parallel_size > 1
+                and self.parallel_config.data_parallel_size_local > 0
+            ):
```
tlrmchlsmth, line 176: When is `data_parallel_size_local` equal to 0? It doesn't seem like that should happen

vllm/v1/worker/gpu_worker.py
```diff
@@ -168,6 +168,27 @@ def init_device(self):
...
+                tp_pp_world_size = (
+                    self.parallel_config.pipeline_parallel_size
+                    * self.parallel_config.tensor_parallel_size
+                )
+                # DP_LOCAL_RANK * TP_PP_WORLD_SIZE + TP_RANK % TP_PP_WORLD_SIZE
+
+                self.local_rank = (
+                    data_parallel_local_rank * tp_pp_world_size
+                    + self.local_rank % tp_pp_world_size  # type: ignore[has-type]
+                ) % torch.cuda.device_count()
```
njhill: I think this would suffice ... local_rank will always be < world_size, and the resulting index should always be < device_count (if not there's some bug/error)
```suggestion
                self.local_rank += data_parallel_local_rank * self.parallel_config.world_size
```

vllm/v1/worker/gpu_worker.py
```diff
@@ -168,6 +168,27 @@ def init_device(self):
...
+            if (
+                self.vllm_config.parallel_config.data_parallel_size > 1
+                and self.parallel_config.data_parallel_size_local > 0
+            ):
+                # Use local DP rank if available, otherwise use global DP rank.
+                data_parallel_local_rank: int = (
+                    self.parallel_config.data_parallel_rank_local
+                    if self.parallel_config.data_parallel_rank_local is not None
+                    else self.parallel_config.data_parallel_rank
+                ) % self.parallel_config.data_parallel_size_local
```
njhill: suggest to simplify ... and I don't think we need the modulo
```suggestion
                dp_rank_local = self.parallel_config.data_parallel_rank_local
                if dp_rank_local is None:
                    dp_rank_local = self.parallel_config.data_parallel_rank
```

vllm/v1/worker/gpu_worker.py
tlrmchlsmth: This looks a little suspicious, could you explain what's happening here?

njhill (commented):
@ilmarkov looking at this some more, I think it would be better to set this higher up i.e. where we assign local_rank before launching the worker process, and also to align other platforms (XPUWorker, CPUWorker, etc.) and ray with this (so that the call to set_device_control_env_var will be removed after all).

We should maybe also consider aligning the handling of `rank` similarly. So that the adjustment isn't required here: https://github.com/vllm-project/vllm/blob/e5b438a24786decebd0b69340f024f18ea184d44/vllm/distributed/parallel_state.py#L1152-L1156

I'm not sure whether there are other implications of doing this though.

cc @youkaichao who may have other comments/insight.

vllm/v1/worker/gpu_worker.py
ilmarkov: For external launcher we [multiply](https://github.com/vllm-project/vllm/blob/main/vllm/config/parallel.py#L474) by DP_size

vllm/v1/worker/gpu_worker.py
ilmarkov: In the second example [here](https://docs.vllm.ai/en/v0.10.0/serving/data_parallel_deployment.html#internal-load-balancing):
> This will run DP=4 with only the API server on the first node and all engines on the second node

vllm/v1/worker/gpu_worker.py
ilmarkov: Can't `dp_rank` be out of `dp_size_local `?

vllm/v1/worker/gpu_worker.py
ilmarkov: Shouldn't reach this part of the code, though

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -876,7 +878,7 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
                         "All kv cache tensors must have the same size"
                     )
                 caches_data.append(
-                    (base_addr, curr_tensor_size_bytes, self.tp_rank, "")
```
NickLucche, line 990: was this altogether incorrect?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -923,7 +925,8 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
                 block_offset = block_id * self.block_len_per_layer[i]
                 addr = base_addr + block_offset
                 # (addr, len, device id)
-                blocks_data.append((addr, kv_block_len, self.tp_rank))
+                assert self.device_id is not None
```
NickLucche: can we assert once out of the loops and keep mypy happy that way?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
ilmarkov: It has to be the same as device id of the tensor.

vllm/v1/worker/gpu_worker.py
njhill: possibly `dp_rank` but not `dp_rank_local`.

vllm/v1/worker/gpu_worker.py
tlrmchlsmth: oof that's tough. Should we try to make that consistent? @youkaichao

vllm/v1/worker/gpu_worker.py
```diff
@@ -168,6 +168,29 @@ def init_device(self):
         if self.device_config.device.type == "cuda":
             # This env var set by Ray causes exceptions with graph building.
             os.environ.pop("NCCL_ASYNC_ERROR_HANDLING", None)
+            if (
+                self.parallel_config.data_parallel_size > 1
+                and self.parallel_config.data_parallel_size_local > 0
+            ):
+                # Use local DP rank if available, otherwise use global DP rank.
+                dp_local_rank = self.parallel_config.data_parallel_rank_local
+                if dp_local_rank is None:
```
tlrmchlsmth, line 179: When can dp_local_rank be None? Do we know?

ilmarkov: In case of [external launcher](https://github.com/vllm-project/vllm/blob/main/vllm/config/parallel.py#L481), I think. We set `self.data_parallel_rank` but not dp_local_rank

vllm/v1/worker/gpu_worker.py
njhill: Yes it feels that there is some inconsistency here that we should rationalize/simplify.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
njhill: Yes I think this was only partially addressed in https://github.com/vllm-project/vllm/pull/18969.

This updated logic looks good to me.

vllm/v1/worker/gpu_worker.py
njhill: And there is also the [`world_size_across_dp`](https://github.com/vllm-project/vllm/blob/7bb736d00e81c6cd893007323641dd974d921715/vllm/config/parallel.py#L294) property...

ilmarkov: Isn't using `self.parallel_config.pipeline_parallel_size * self.parallel_config.tensor_parallel_size` in this PR safe enough?

njhill: It's more that we should streamline/unify things as much as possible.

vllm/v1/worker/gpu_worker.py
```diff
@@ -168,6 +168,30 @@ def init_device(self):
...
+                self.parallel_config.data_parallel_size > 1
+                and self.parallel_config.data_parallel_size_local > 0
+                and self.parallel_config.data_parallel_backend != "ray"
+            ):
+                # Use local DP rank if available, otherwise use global DP rank.
+                dp_local_rank = self.parallel_config.data_parallel_rank_local
+                if dp_local_rank is None:
+                    dp_local_rank = (
+                        self.parallel_config.data_parallel_rank
+                        % self.parallel_config.data_parallel_size_local
```
njhill: The reason for not having the modulo here is that we should be explicitly falling back to `data_parallel_rank` in this case. It should be less than `data_parallel_size_local` or else `data_parallel_rank_local` should be set.

njhill (approved):
Thanks again for finding and fixing this @ilmarkov!

As discussed we can look at making things more consistent in a follow-on PR, w.r.t. other platforms and local_rank computation.

SageMoore (approved):
I'm very excited about this change. Thanks for the contribution!

PR #26715, opened 2025-10-13
[Model] Always use Transformers backend for PaliGemma and Gemma3-MM

DarkLight1337:
## Purpose

Since vLLM doesn't support the special attention mask used by PaliGemma and Gemma3-MM (not to be confused with Gemma3n), this PR removes our custom implementations so Transformers backend is used for these models.

cc @hmellor 

@NickLucche it would be great if you could test if gemma3 works with Transformers backend on TPU!

## Test Plan

Transformers backend tests should pass.

## Test Result

---
<details>

</details>


vllm/model_executor/models/registry.py
hmellor, line 1: If you add the removed entries in `_TRANSFORMERS_SUPPORTED_MODELS` then vLLM will use the transformers backend by default. (no warning when not passing `--model-impl transformers`)

DarkLight1337: Thanks for pointing that out, done.

vllm/model_executor/models/registry.py
```diff
@@ -403,6 +398,11 @@
     "SmolLM3ForCausalLM": ("transformers", "TransformersForCausalLM"),
     # Multimodal models
     "Emu3ForConditionalGeneration": ("transformers", "TransformersForMultimodalLM"),  # noqa: E501
+    "Gemma3ForConditionalGeneration": ("transformers", "TransformersForMultimodalLM"),  # noqa: E501
+    "PaliGemmaForConditionalGeneration": (
+        "transformers",
+        "TransformersForMultimodalLM",
+    ),  # noqa: E501
```
hmellor: I don't think you're winning this fight with `ruff`

```suggestion
    ),
```

hmellor (approved):
LGTM

PR #26726, opened 2025-10-13
[Feature] Change vllm.py with pydantic validation

VladOS95-cyber:
## Purpose
## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/config/vllm.py
+++ b/vllm/config/vllm.py
@@ -151,17 +151,16 @@ class VllmConfig:
     @model_validator(mode="after")
     def validate_all2all_backend_for_dbo(self) -> Self:
         if self.parallel_config.enable_dbo:
-            a2a_backend = envs.VLLM_ALL2ALL_BACKEND
+            a2a_backend = self.parallel_config.all2all_backend
             allowed = {"deepep_low_latency", "deepep_high_throughput"}
 
             if a2a_backend not in allowed:
                 raise ValueError(
-                    "Microbatching currently only supports the 'deepep_low_latency' "
-                    "and 'deepep_high_throughput' all2all backends. "
-                    f"Got '{a2a_backend}'.\n"
-                    "To fix this, set the environment variable:\n"
-                    "  export VLLM_ALL2ALL_BACKEND=deepep_low_latency\n"
-                    "and install the DeepEP kernels."
+                    "Microbatching currently only supports the deepep_low_latency and "
+                    f"deepep_high_throughput all2all backend. {a2a_backend} is not "
+                    "supported. To fix use --all2all-backend=deepep_low_latency or "
+                    "--all2all-backend=deepep_high_throughput and install the DeepEP"
+                    " kernels."
                 )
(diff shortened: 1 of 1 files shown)


vllm/config/vllm.py
```diff
@@ -55,33 +56,31 @@ class VllmConfig:
     simplifies passing around the distinct configurations in the codebase.
     """
 
-    # TODO: use default_factory once default constructing ModelConfig doesn't
-    # try to download a model
-    model_config: ModelConfig = None  # type: ignore
```
VladOS95-cyber, line 60: Hey @hmellor, I am not sure about this one but I did not find in the ModelConfig where the model is downloaded, probably I just missed that

vllm/config/vllm.py
```diff
@@ -55,33 +56,31 @@ class VllmConfig:
...
+    device_config: DeviceConfig = Field(default_factory=DeviceConfig)
     """Device configuration."""
-    load_config: LoadConfig = field(default_factory=LoadConfig)
+    load_config: LoadConfig = Field(default_factory=LoadConfig)
     """Load configuration."""
-    lora_config: LoRAConfig | None = None
+    lora_config: LoRAConfig | None = Field(default=None)
     """LoRA configuration."""
-    speculative_config: SpeculativeConfig | None = None
+    speculative_config: SpeculativeConfig | None = Field(default=None)
```
hmellor: We don't need to start using `Field` if it's not necessary (there are other places we could revert this too)

vllm/config/vllm.py
hmellor: It's possible that this is fixed now. The issue was that if you instantiated `ModelConfig()` it would start downloading the default model

hmellor: Looks like the docs build has given us the answer https://app.readthedocs.org/projects/vllm/builds/29935677/, it's trying to access the HF Hub repo. We should leave this as `None`. But you can do so using `Field(default=None)` and remove the `type: ignore` (Pydantic doesn't validate derfault values of `Field`s by default)

hmellor: Can you add the comment back explaining why it's `None`?

vllm/config/vllm.py
```diff
@@ -103,11 +101,11 @@ class VllmConfig:
     # some opaque config, only used to provide additional information
     # for the hash computation, mainly used for testing, debugging or out of
     # tree config registration.
-    additional_config: dict | SupportsHash = field(default_factory=dict)
+    additional_config: dict | SupportsHash = Field(default_factory=dict)
     """Additional config for specified platform. Different platforms may
     support different configs. Make sure the configs are valid for the platform
     you are using. Contents must be hashable."""
-    instance_id: str = ""
+    instance_id: str = Field(default="")
```
hmellor:
```suggestion
    instance_id: str = ""
```

hmellor (approved):
LGTM! Thanks for taking the time to see how we could improve this.

PR #23917, opened 2025-08-29
[P/D]`kv_output_aggregator` support heterogeneous

LCAIZJ:
## Purpose
1. D is memory bound so greater TP size will yield better performance. However, in the MLA scenario, each TP worker maintains the full kvcache, which prevents actual GPU memory savings. As a result, when MLA is enabled, the TP size for P often exceeds that of D.
2. When the TP size for P exceeds that of D, the current mechanism of kv_output_aggregator fails to function correctly. As illustrated in the figure below, both finished_sending and finished_recving should be 2 in this scenario. However, the kv_output_aggregator incorrectly initializes them to world_size, causing finished_sending to remain at 4 and preventing the release of P's kvcache.
image

3. To resolve this issue, we introduce a new interface get_finished_count() in the connector's base class. This method is implemented by specific connector subclasses to return the correct values for finished_sending and finished_recving. When different executors invoke get_finished_count() to initialize the kv_output_aggregator, it ensures the counters are configured correctly.

## Test Plan
Model：DeepSeek-R1
Distributed Strategy：1P(TP16)1D(DP4、TP4)

## Test Result
Successfully processed 10,000 benchmark prompts with 4K input tokens and 1.5K output tokens per request. 
image
Since the primary goal was to verify overall functionality, we did not enable many optimization features during benchmark execution. As a result, the TTFT and TPOT metrics are suboptimal, though the verification of the core functionality was successfully completed.

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #26734, opened 2025-10-13
[P/D] Dynamic `kv_output_aggregator` collect size

NickLucche:
This PR expands the approach presented in https://github.com/vllm-project/vllm/pull/23917, referring to the same issue, but specifically addressing the cases where the number of requests to collect isn't known at startup time.
image

This is the case for discovery-based connectors such as NixlConnector, whereby the topology isn't fully propagated top-down, but rather discovered by handshaking with peers.
 
Hence, this PR enables connectors to output an extra-optional field named `expected_finished_reqs` which allows to set the number of actual transfers that are happening and that are actually reported from worker->executor.

Mind that this approach could superseed the changes introduced in https://github.com/vllm-project/vllm/pull/23917, as long as this value is returned on the first (or on every) call to get_finished, from which the kv conector output is built from. cc @LCAIZJ @zzy-ContiLearn as I don't mean to break previous integration, but merely stating a possiblity here. 

I've added a new test to showcase functionality.


vllm/distributed/kv_transfer/kv_connector/utils.py
```diff
@@ -154,6 +157,19 @@ def update_finished_set(
...
+            if (
+                output.expected_finished_reqs > 0
+                and output.expected_finished_reqs != self._expected_finished_reqs
+            ):
+                logger.debug(
+                    "Expected finished requests updated from %d to %d",
+                    self._expected_finished_reqs,
+                    output.expected_finished_reqs,
+                )
+                self._expected_finished_reqs = output.expected_finished_reqs
```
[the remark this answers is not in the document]
NickLucche: Yeah I was planning to entrust connector with the responsability of returning the same value among workers, but we can have it so we only look at the first one for simplicity. 
This value shouldn't vary often.

vllm/distributed/kv_transfer/kv_connector/utils.py
```diff
@@ -141,7 +141,10 @@
             finished_set: set[str],
         ) -> None:
             for req_id in req_ids or ():
-                remaining_count_dict[req_id] -= 1
+                remaining_count = remaining_count_dict.get(
+                    req_id, self._expected_finished_reqs
+                )
+                remaining_count_dict[req_id] = remaining_count - 1
```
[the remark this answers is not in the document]
NickLucche: this is intended, workers can still report for older pending requests.

vllm/v1/outputs.py
```diff
@@ -88,6 +88,7 @@ class KVConnectorOutput:
     # IDs of externally computed KV blocks that failed to load.
     # Requests referencing these blocks should be rescheduled to recompute them.
     invalid_block_ids: set[int] = field(default_factory=set)
+    expected_finished_reqs: int = 0
```
markmc: Could do with a brief comment here that it overrides `connector.get_finished_count()` and also a comment about that in `get_finished_count()` docs

vllm/distributed/kv_transfer/kv_connector/utils.py
```diff
@@ -127,8 +126,9 @@ class KVOutputAggregator:
     def __init__(self, world_size: int):
         # Complete transfer tracker. Used to track finished requests
         # [req_id -> n_remaining_workers]
-        self._recv_remaining_count = defaultdict[str, int](lambda: world_size)
-        self._send_remaining_count = defaultdict[str, int](lambda: world_size)
+        self._recv_remaining_count = dict[str, int]()
+        self._send_remaining_count = dict[str, int]()
+        self._expected_finished_reqs = world_size
```
markmc: It would be nice to have the relationship between 'get_finished_count()' and 'kv_connector_output.expected_finished_reqs' expressed somewhere in this file

For a start, `s/world_size/finished_count/` here would help

But also maybe

```
class KVOutputAggregator:

    @classmethod
    def new_from_connector(cls, connector: KVConnectorBase, world_size: int):
        return cls(connector.get_finished_count() or world_size)
```

The only reason I like this is it calls attention to `get_finished_count()` when you're looking at `KVOutputAggregator`

so engine:

```
self.model_executor.init_kv_output_aggregator(self.scheduler.connector)
```

and executor:

```
    def init_kv_output_aggregator(self, connector: KVConnectorBase) -> None:
        self.kv_output_aggregator = KVOutputAggregator.from_connector(connector, self.parallel_config.world_size)
```

tests/v1/kv_connector/unit/test_output_aggregator.py
```diff
@@ -16,11 +16,13 @@ def __init__(
         finished_sending: set[str] | None = None,
         finished_recving: set[str] | None = None,
         invalid_block_ids: set[int] | None = None,
+        expected_finished_reqs: int = 0,
```
markmc: It would be really nice to fix the typo in this file's name :rofl:

tests/v1/kv_connector/unit/test_output_aggregator.py
```diff
@@ -158,3 +160,40 @@ def test_async_aggregate_workers_output():
...
+
+    assert not aggregated.kv_connector_output.finished_sending
+
+    # Req2 is finished
+    assert "req2" not in aggregator._recv_remaining_count
+    assert aggregated.kv_connector_output.finished_recving == {"req2"}
+
+    # Req1 is still waiting for 2 more acks (expected_finished_reqs has no effect)
+    # NOTE: This is to showcase dynamic update. Workers are responsible for
+    # ensuring "req1" termination in this case
```
markmc, line 198: The commentary is very useful, and the test very useful to head off e.g. the review comment from codex. But the scenario seems a bit wild! I can understand the expected count changing at handshake time, but after requests are in flight? (I'm probably missing something)

tests/v1/kv_connector/unit/test_output_aggregator.py
NickLucche: lol

tests/v1/kv_connector/unit/test_output_aggregator.py
NickLucche: So the idea is that this number can vary based on external factors (the reads issued from D),  so if D configuration was to change at runtime (a scenario I am not interested to support in nixl rn) the connector would have maximum flexibility to implement their own logic to react here.

But I understand this may be scary in terms of how much freedom the connector has over it.
I am also open to just limit the number of changes to 1.

vllm/v1/outputs.py
njhill: Agree very much with @markmc about the comment. It would be good to include the fact that it applies to both sending and recving cases. And also making it clear that this is not actually kv connector output/events like the other field but more like current "configuration" information and that the value would be expected to remain static in most cases.

Also, I'm wondering whether it makes sense to have a default here and if so whether 0 makes sense. IIRC this will only be instantiated by vllm itself so perhaps we should make this field (and others?) required, for less confusion.

Finally I am wondering whether there's a more accurate name for this, since `expected_finished_reqs` I think could be a bit misleading. Since it's not really a count of requests but rather a count of notifications to expect per request, right? Would something like `workers_to_aggregate` or `expected_finished_notifs_per_request` be better? (not that I necessarily like either of those!)

njhill (commented):
Thanks @NickLucche and sorry for taking ages to get to it. Though it's a small change it takes me a while to context switch back to this and understand / think carefully about the implications!

Looks great overall. Main thing is that I feel the purpose of the field is a bit unintuitive and so very clear naming and comments would be good.

vllm/v1/outputs.py
NickLucche: > vllm itself so perhaps we should make this field (and others?) required

I think the default-based output struct currently comes from the fact that output is yielded first and then only populated on finally: in the mixin.

vllm/distributed/kv_transfer/kv_connector/utils.py
NickLucche: good point, was just waiting to get nick's ack on general idea, will implement this

vllm/executor/executor_base.py
```diff
@@ -257,10 +260,10 @@ async def check_health_async(self) -> None:
         exception."""
         self.check_health()
 
-    def init_kv_output_aggregator(self, finished_count: int | None) -> None:
+    def init_kv_output_aggregator(self, connector: KVConnectorBase_V1) -> None:
```
markmc:
```
vllm/executor/executor_base.py:263: in ExecutorBase
    def init_kv_output_aggregator(self, connector: KVConnectorBase_V1) -> None:
E   NameError: name 'KVConnectorBase_V1' is not defined
```

tests/v1/kv_connector/unit/test_output_aggregator.py
```diff
@@ -158,3 +160,40 @@ def test_async_aggregate_workers_output():
     assert aggregated.finished_sending is None
     assert aggregated.finished_recving == {"req2"}
     assert aggregated.invalid_block_ids == {3, 4, 5}
+
+
+def test_aggregate_workers_output_with_expected_finished_reqs():
+    # We create the aggregator expecting to collect from 4 workers
+    aggregator = KVOutputAggregator(world_size=4)
```
markmc:
```
>       aggregator = KVOutputAggregator(world_size=4)
E       TypeError: KVOutputAggregator.__init__() got an unexpected keyword argument 'world_size'

tests/v1/kv_connector/unit/test_output_aggregator.py:167: TypeError

```

markmc:
```
$ git grep -n KVOutputAggr.*world_size tests/
tests/v1/kv_connector/unit/test_nixl_connector.py:706:    aggregator = KVOutputAggregator(world_size=3)
tests/v1/kv_connector/unit/test_nixl_connector.py:771:    aggregator = KVOutputAggregator(world_size=3)
tests/v1/kv_connector/unit/test_output_aggregator.py:38:    aggregator = KVOutputAggregator(world_size=2)
tests/v1/kv_connector/unit/test_output_aggregator.py:90:    aggregator = KVOutputAggregator(world_size=2)
tests/v1/kv_connector/unit/test_output_aggregator.py:167:    aggregator = KVOutputAggregator(world_size=4)
```

njhill (approved):
Thanks @NickLucche!

PR #26737, opened 2025-10-13
[Core] Streamline some structured output related code

njhill:
Simplify some data structures, redundant fields, etc.

Things noticed while reading through the code to understand it.

### DIFF
--- a/vllm/v1/core/sched/output.py
+++ b/vllm/v1/core/sched/output.py
@@ -165,9 +165,8 @@ class SchedulerOutput:
     # freed from the encoder cache.
     free_encoder_mm_hashes: list[str]
 
-    # Dict of request ids to their index within the batch
-    # for filling the next token bitmask
-    structured_output_request_ids: dict[str, int]
+    # ids of structured outputs requests included in the bitmask, in order.
+    structured_output_request_ids: list[str]
     # the bitmask for the whole batch
     grammar_bitmask: "npt.NDArray[np.int32] | None"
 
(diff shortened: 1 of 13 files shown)


vllm/v1/core/sched/output.py
```diff
@@ -165,9 +165,8 @@ class SchedulerOutput:
     # freed from the encoder cache.
     free_encoder_mm_hashes: list[str]
 
-    # Dict of request ids to their index within the batch
-    # for filling the next token bitmask
-    structured_output_request_ids: dict[str, int]
+    # ids of structured outputs requests included in the bitmask, in order.
```
russellb, line 168:
```suggestion
    # ids of structured outputs requests included in the bitmask.
    # The index of the id in this list corresponds to the index into
    # grammar_bitmask.
```

is this right?

russellb (approved):
just one minor comment suggestion, otherwise lgtm. thanks!

vllm/v1/core/sched/output.py
njhill: Actually not necessarily in spec decoding case. The requests will be in the order of the list but there may be more than one row per request so the index in the list won't match index into the bitmask. It would be true in non-spec decode case though.

russellb: Oh, right -- clarifying the order definition here would be nice in any case

njhill: I will do this in follow-on PR to avoid running the whole CI again!

PR #9632, opened 2024-10-23
[torch.compile] support moe models

youkaichao:
moe models will read config file to determine the triton config to run.

reading files during forward is a disaster for `torch.compile` .

this pr wraps the config reading part inside a custom op, so that it can pass `torch.compile` (although `torch.compile` will not be able to optimize it).

PR #26739, opened 2025-10-13
[torch.compile] Unwrap fused_marlin_moe custom op

varun-sundar-rabindranath:
## Purpose
Unwrap unnecessary CustomOP'ing of `fused_marlin_moe`. 

`fused_marlin_moe` was first wrapped into a custom op in the PR https://github.com/vllm-project/vllm/pull/9632 as `fused_marlin_moe` was doing file reads to get optimal triton config. This function has evolved since and no longer needs file reads and could be unwrapped. 

## Test Plan
CI


bnellnm (approved):

mgoin (approved):
Nice find, LGTM

jeejeelee (approved):

PR #26740, opened 2025-10-13
[Log] Optimize Startup Log

yewentao256:
## Purpose

We have too much log printed in the startup, this PR:

1. debug or delete some useless information
2. only print in the first rank


vllm/v1/engine/core.py
```diff
@@ -726,7 +725,6 @@ def startup_handshake(
         )
 
         # Receive initialization message.
-        logger.info("Waiting for init message from front-end.")
```
mgoin, line 716: Why not leave as debug? This seems useful for timeout debugging

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2846,7 +2846,8 @@ def load_model(self, eep_scale_up: bool = False) -> None:
         Args:
             eep_scale_up: the model loading is for elastic EP scale up.
         """
-        logger.info("Starting to load model %s...", self.model_config.model)
+        if is_global_first_rank():
+            logger.info_once("Starting to load model %s...", self.model_config.model)
```
mgoin: What if we make a new version of warning/info/debug to handle this case? I'd like to do this in a lot of places. Such as `logger.info_once_global("Starting to load model %s...", self.model_config.model)`

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1137,7 +1137,7 @@ def __init__(
             )
             self.local_num_experts = local_num_experts
             self.register_buffer("expert_map", expert_map)
-            logger.info_once(
+            logger.debug_once(
                 "[EP Rank %s/%s] Expert parallelism is enabled. Expert "
                 "placement strategy: %s. Local/global"
                 " number of experts: %s/%s. Experts local to global index map:"
```
mgoin, line 1218: I feel like this one is pretty important and not on by default

vllm/v1/engine/core.py
```diff
@@ -92,7 +91,7 @@ def __init__(
         load_general_plugins()
 
         self.vllm_config = vllm_config
-        if is_global_first_rank():
+        if vllm_config.parallel_config.data_parallel_rank == 0:
```
mgoin, line 92: Why not is_global_first_rank?

vllm/distributed/parallel_state.py
mgoin, line 1: cc @njhill @tlrmchlsmth

vllm/v1/engine/core.py
yewentao256: `is_global_first_rank` doesn't work here, since the group hasn't been initialized

vllm/v1/worker/gpu_model_runner.py
yewentao256: I think it would make it not that easy to maintain, firstly we can't have it in the `logger.py` (circular import), if we put it somewhere else then we actually make things complicated.

Also, we also need to introduce `.info_once_local_node`, and what's worse, sometimes when it is not initialized, `info_once_global` won't work and we may need to use `vllm_config.parallel_config.data_parallel_rank == 0`

@mgoin CC

vllm/model_executor/layers/fused_moe/layer.py
yewentao256: fixed

vllm/v1/engine/core.py
yewentao256: fixed

vllm/model_executor/model_loader/default_loader.py
```diff
@@ -311,10 +312,12 @@ def load_weights(self, model: nn.Module, model_config: ModelConfig) -> None:
             loaded_weights = load_weights_and_online_quantize(self, model, model_config)
 
         self.counter_after_loading_weights = time.perf_counter()
-        logger.info(
-            "Loading weights took %.2f seconds",
-            self.counter_after_loading_weights - self.counter_before_loading_weights,
-        )
+        if is_global_first_rank():
```
tlrmchlsmth: I think this should check the local rank instead of the global one. Makes sense to print this in every pod IMO

vllm/distributed/parallel_state.py
```diff
@@ -1322,7 +1322,7 @@ def initialize_model_parallel(
         group_ranks, get_world_group().local_rank, backend, group_name="ep"
     )
 
-    logger.info(
+    logger.debug(
```
tlrmchlsmth: I'm slightly in favor of leaving this as info, as I find it pretty informative

vllm/v1/engine/core.py
tlrmchlsmth: In that case how does this work on current main?

vllm/distributed/parallel_state.py
yewentao256: Updated to `info_once`

vllm/model_executor/model_loader/default_loader.py
yewentao256: Make sense, added `is_local_first_rank`

vllm/v1/engine/core.py
yewentao256: This actually doesn't work in main now, eg : `vllm serve nvidia/DeepSeek-R1-FP4 -dp 4  --enable-expert-parallel --port 9256` we will see the config log 4 times.

vllm/compilation/backends.py
mgoin, line 1: @ProExpertProg are you okay with making these debug? I personally think they should be since we already have the global "torch.compile took X time"

mgoin (approved):
LGTM, just want @ProExpertProg to sign off on the compile change

vllm/compilation/backends.py
```diff
@@ -606,7 +606,7 @@ def __call__(
         if disable_cache:
             logger.info("vLLM's torch.compile cache is disabled.")
         else:
-            logger.info(
+            logger.debug(
```
zou3519: To check, we're squashing this log on all ranks, or all but the 0th?

The location of the cache folder is useful to know.

vllm/compilation/backends.py
```diff
@@ -245,9 +245,9 @@ def compile(
             if graph_index == 0:
                 # adds some info logging for the first graph
                 if runtime_shape is None:
-                    logger.info("Cache the graph for dynamic shape for later use")
+                    logger.debug("Cache the graph for dynamic shape for later use")
```
zou3519: Same question as before. For visibility, I think it's important for the user to know if they're reading from the cache or doing a cold compile. We've had issues where people complain about compile time, but it turns out they're doing a cold compile and not a warm compile.

zou3519 (changes requested):
I think the compile time (and which mode of compilation we're on - like if we're loading from the cache, or starting from scratch) should be printed by default. The compilation is a part of the programming model and at this point I'm worried hiding it from users means we have to work harder to explain it. If you think there are too many logs, can we work to figure out the minimal amount of things to communicate?

vllm/compilation/backends.py
yewentao256: Fixed, thanks!
Here we update to log one on rank 0 for each local node

vllm/compilation/backends.py
yewentao256: This log is for writing, in my perspective is not that important.

zou3519: I think this is good for the programming model. This informs the user that the next time they load this, it'll be cached.

zou3519 (commented):
I would prefer the logs for an artifact being cached to also be in the startup log. I think it's important enough (and helps the user pick up on the programming model).

other than that this lgtm, thank you

vllm/compilation/backends.py
yewentao256: OK, updated to local rank 0 print the `info` log, thanks!

vllm/compilation/backends.py
ProExpertProg: I think this is already at a minimum, I agree with @zou3519's comments. Taking a look now through it

vllm/compilation/backends.py
```diff
@@ -620,7 +623,7 @@ def __call__(
         from .monitor import torch_compile_start_time
 
         dynamo_time = time.time() - torch_compile_start_time
-        logger.info("Dynamo bytecode transform time: %.2f s", dynamo_time)
+        logger.debug("Dynamo bytecode transform time: %.2f s", dynamo_time)
```
ProExpertProg: This is relied upon for figuring out startup times for different components, we should keep it

vllm/v1/worker/gpu_model_runner.py
ProExpertProg: You can just import locally in the function to avoid circular imports? And can the function not handle the fallback to `vllm_config.parallel_config.data_parallel_rank == 0` automatically - check if parallel state is initialized or not?

ProExpertProg: Also fallback is not that important it's only used in one place

ProExpertProg (commented):
Mostly LGTM, I agree with @mgoin about the utility

vllm/v1/worker/gpu_model_runner.py
yewentao256: The main point I am considering is we will need at least `logger.info_once_local` and `logger.info_once_global`, and similarly for `debug` `warning` `error`. What's more, if we want to fall back like `vllm_config.parallel_config.data_parallel_rank == 0`, we will need pass `config` to the log function as well.
So perhaps it is not a very good idea. @ProExpertProg

vllm/compilation/backends.py
yewentao256: OK, converted to local first rank.

vllm/v1/worker/gpu_model_runner.py
ProExpertProg: What about adding `local=True` `global=True` params to the `info_once` function? And no fallback - the one case that uses it can remain manual?

yewentao256: How about `scope="process" | "global" | "local"`?

mgoin: I like `scope` since it naturally reads well with `once`

ProExpertProg (approved):

zou3519 (approved):
lgtm, thank you

Issue #26582, opened 2025-10-10
[Bug]: which triton-kernels version for MXFP4 Triton backend?


matkle:
### Your current environment

vllm v0.11.0 installed via `uv pip install vllm --torch-backend=auto`

triton + triton-kernels at different commits installed from source

### 🐛 Describe the bug

**Which triton + triton-kernels version does one have to install to run GPT-OSS with the MXFP4 Triton backend?**

No matter which version I try, I always get an error `Failed to import Triton kernels. Please make sure your triton version is compatible.`

Clearly, the latest triton-kernels will not work since the code in `vllm.model_executor.layers.fused_moe.gpt_oss_triton_kernels_moe` tries to import from `triton_kernels.routing`, but `triton_kernels.routing` has been deprecated (cf. https://github.com/triton-lang/triton/commit/30ede52aa2aecfd2ab3d6672ed21bbf4eb6438b3).

But also with older versions I get errors like `ImportError: cannot import name 'triton_key' from 'triton.compiler.compiler` or `Error: No module named 'triton.language.target_info`.

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

PR #26748, opened 2025-10-14
[gpt-oss] Update triton_kernels mxfp4 backend to be compatible with triton_kernels upstream

zyongye:
Recently, triton_kernels introduced routing function [refactoring](https://github.com/triton-lang/triton/pull/8375). We need to incorporate it into vllm. 

Partial address #26582

Accuracy bench
openai/gpt-oss-120b, TP=2, GPQA:
| Reasoning effort | Score |
|:--------|:-----:|
| Low | 65.84  | 
| Medium  | 72.2  |

### DIFF
--- a/vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
+++ b/vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
@@ -83,9 +83,7 @@ def pack_bitmatrix(
         tl.store(bitmatrix_ptrs, y, mask=offsets_m[:, None] < n_rows)
 
 
-def legacy_routing_from_bitmatrix(
-    bitmatrix, expt_scal, expt_indx, n_expts_tot, n_expts_act
-):
+def routing_from_bitmatrix(bitmatrix, expt_scal, expt_indx, n_expts_tot, n_expts_act):
     sparse_logits = SparseMatrix(
         indx=expt_indx,
         vals=expt_scal,
@@ -110,7 +108,7 @@ def legacy_routing_from_bitmatrix(
     return routing_data, gather_idx, scatter_idx
 
 
-def legacy_routing(logits, n_expts_act, sm_first=False, expt_indx=None, n_rows=None):
+def routing(logits, n_expts_act, sm_first=False, expt_indx=None, n_rows=None):
     if sm_first:
         logits = torch.softmax(logits, dim=-1)
     sparse_logits = triton_topk(
@@ -120,7 +118,7 @@ def legacy_routing(logits, n_expts_act, sm_first=False, expt_indx=None, n_rows=N
         y_indx=expt_indx,
         n_rows=n_rows,
     )
-    return legacy_routing_from_bitmatrix(
+    return routing_from_bitmatrix(
         sparse_logits.mask,
         sparse_logits.vals,
         sparse_logits.indx,
@@ -142,7 +140,7 @@ def triton_kernel_moe_forward(
     global_num_experts: int = -1,
     expert_map: torch.Tensor | None = None,
 ) -> torch.Tensor:
-    routing_data, gather_idx, scatter_idx = legacy_routing(
+    routing_data, gather_idx, scatter_idx = routing(
         gating_output, topk, sm_first=not renormalize
     )
 
@@ -269,7 +267,7 @@ def make_routing_data(
 
     # matmul_ogs expects invalid topk_weights to be -1s
     topk_weights = torch.where(topk_ids == -1, -1.0, topk_weights)
-    routing_data, gather_indx, scatter_indx = legacy_routing_from_bitmatrix(
+    routing_data, gather_indx, scatter_indx = routing_from_bitmatrix(
         bitmatrix, topk_weights, topk_ids, num_local_experts, num_topk
     )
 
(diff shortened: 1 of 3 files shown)


vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
```diff
@@ -71,6 +83,50 @@ def pack_bitmatrix(
         tl.store(bitmatrix_ptrs, y, mask=offsets_m[:, None] < n_rows)
 
 
+def routing_from_bitmatrix(bitmatrix, expt_scal, expt_indx, n_expts_tot, n_expts_act):
```
bnellnm: Could you add type annotations to this function?

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
```diff
@@ -71,6 +83,50 @@ def pack_bitmatrix(
...
+        n_expts_tot,
+        n_expts_act,
+        ragged_batch_metadata,
+    )
+    gather_idx = GatherIndx(combine_indx, dispatch_index)
+    scatter_idx = ScatterIndx(dispatch_index, combine_indx)
+    return routing_data, gather_idx, scatter_idx
+
+
+def routing(logits, n_expts_act, sm_first=False, expt_indx=None, n_rows=None):
```
bnellnm: Ditto

bnellnm (approved):
LGTM

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
zyongye: All comments addressed.

eppaneamd:
Hello @zyongye! Just FYI (looking into this): 

```
(APIServer pid=7223) INFO 10-17 04:46:12 [model.py:658] Resolved architecture: Qwen3VLMoeForConditionalGeneration
(APIServer pid=7223) INFO 10-17 04:46:12 [model.py:1745] Using max model len 262144
(APIServer pid=7223) ERROR 10-17 04:46:12 [gpt_oss_triton_kernels_moe.py:40] Failed to import Triton kernels. Please make sure your triton version is compatible. Error: No module named 'triton_kernels.tensor'
```

zyongye:
@eppaneamd Is the import error still present for models that don't require `triton_kernels`?

mgoin:
I agree with Varun, is there a reason not to target the triton release that matches torch 2.9 and add triton_kernels as a proper dep at that point?

zyongye:
> I agree with Varun, is there a reason not to target the triton release that matches torch 2.9 and add triton_kernels as a proper dep at that point?

No particular reason. We can defer merging this after a new torch release. But we definitely need trying to make triton_kernels as a dep and pin to the triton release commit.

zyongye:
> > @eppaneamd Is the import error still present for models that don't require `triton_kernels`?
> 
> That's a good question, unfortunately not sure. Haven't been testing that extensively, but briefly with `Qwen3VLMoeForConditionalGeneration`. I have been looking for possible solutions/workarounds for getting it work, as I am encountering an import error with some of these (and so far has been difficult to find a commit where all of these work as vllm currently expects)
> 
> ```
> triton_kernels.routing.routing_from_bitmatrix
> triton.compiler.compiler.triton_key
> triton_kernels.tensor
> triton_kernels.tensor.SparseMatrix
> ```

@eppaneamd I think you need to update `triton` to main and use corresponding `triton_kernels` in that commit.

eppaneamd:
> > > @eppaneamd Is the import error still present for models that don't require `triton_kernels`?
> > 
> > 
> > That's a good question, unfortunately not sure. Haven't been testing that extensively, but briefly with `Qwen3VLMoeForConditionalGeneration`. I have been looking for possible solutions/workarounds for getting it work, as I am encountering an import error with some of these (and so far has been difficult to find a commit where all of these work as vllm currently expects)
> > ```
> > triton_kernels.routing.routing_from_bitmatrix
> > triton.compiler.compiler.triton_key
> > triton_kernels.tensor
> > triton_kernels.tensor.SparseMatrix
> > ```
> 
> @eppaneamd I think you need to update `triton` to main and use corresponding `triton_kernels` in that commit.

That yields triton_key import error unfortunately

eppaneamd:
> Context: At the moment, to get `triton_kernels` working with vllm we need to install `triton_kernels` explicitly. For DP, we just have to build from source as it requires torch nightly - more information here #26582
> 
> Discussed with @zyongye over DM: This PR introduces changes so vllm works with triton_kernels from triton `main`. I believe we shouldn't target triton `main` as it is too volatile - IIUC, `torch2.9` ships with Triton branch `release/3.5.x` and we should also target `triton_kernels` in `release/3.5.x` so it works for when vllm starts supporting `torch2.9`. For now, the workarounds mentioned in issue #26582 should be sufficient.
> 
> fwiw, I tried the following to see if `triton_kernels` works with vllm for gpt-oss,
> 
> ```
>  # build vllm from source using torch2.9 
>  uv pip install torch==2.9 
>  uv pip install torchaudio 
>  uv pip install torchvision 
>  git clone https://github.com/vllm-project/vllm.git
>  cd vllm/
>  python use_existing_torch.py
>  uv pip install requirements/common.txt 
>  uv pip install -r requirements/common.txt 
>  uv pip install -r requirements/cuda.txt 
>  uv pip install -r requirements/build.txt 
>  uv pip install -vvv  -e .  --no-build-isolation 
>  
> # Installed triton_kernels like,
> git clone https://github.com/triton-lang/triton.git 
> cd triton
> git checkout release/3.5.x
> cd ../
> uv pip install triton/python/triton_kernels --no-deps 
> ```
> 
> I see that it works for both TP and DP on vllm `main` - when we switch to torch2.9, we could add `triton_kernels` as a dependency.
> 
> @zou3519 @ProExpertProg @mgoin - Is it correct to target Triton branch `release/3.5.x` ? @zyongye I believe I've captured our conversation correctly. Please feel free to correct me.

@varun-sundar-rabindranath do you have any estimate when vllm starts to officially support `torch2.9`?

PR #26754, opened 2025-10-14
[Bugfix][Core] running queue index leakage exception

CLFutureX:
When an already scheduled request is evicted, an element will be removed from the range [0, req_index) of the running queue. 
At this point, all unscheduled request will shift forward by one position, and req_index needs to be adjusted accordingly by moving it forward one step (i.e., req_index -= 1). Failure to do so will result in the next request being skipped.
This defect was previously identified by me, but due to other factors, it has not been properly resolved to date.


vllm/v1/core/sched/scheduler.py
```diff
@@ -274,6 +274,7 @@ def schedule(self) -> SchedulerOutput:
                         token_budget += num_scheduled_tokens[preempted_req.request_id]
                         req_to_new_blocks.pop(preempted_req.request_id)
                         num_scheduled_tokens.pop(preempted_req.request_id)
+                        req_index -= 1
```
njhill, line 281: Thanks for catching this @CLFutureX and sorry for not getting to it sooner!

I think whether we decrement `req_index` here should depend on whether the index of the preempted request is less than `req_index`?

njhill: Oh nevermind, I see that it will only be in `scheduled_running_reqs` iff this is the case, so I think this looks correct.

njhill (approved):
Thanks @CLFutureX

Issue #26744, opened 2025-10-13
[RFC]: Nixl Connector Heterogeneous BlockSize support


xuechendi:
### Motivation.

To support scenarios when prefill and decode have their own preferred block_size. Ex: Prefill with CUDA(16 as block_size) and decode with Intel Gaudi(128 as block_size)

### Proposed Change.
**edition2(latest proposal)**

- update add_remote_agent to prepare xfer_buffer with **local block_size**
- update read_block to map remote_block_ids/local_block_ids
- update in get_finished for necessary pos-process

One liner: Always using larger block_size as copy unit, do post processing after receiving to split/merge blocks

Reason:
For HND, memory is contiguous as block0_head0, block0_head1,... block1_head0, block1_head1..
Copy multiple blocks will leads to post-processing work after receiving

case1:
prefill Block_size < decode BlockSize

<img width="2389" height="504" alt="Image" src="https://github.com/user-attachments/assets/e30be4f4-2103-4328-8fad-43d52f0067ac" />

<img width="2288" height="648" alt="Image" src="https://github.com/user-attachments/assets/3ab5b6f6-a16e-4381-9947-84eaed62a976" />

case2:
prefill block_size > decode blocksize

<img width="2331" height="468" alt="Image" src="https://github.com/user-attachments/assets/8aea786a-2cbc-4a0f-8a8b-7d821d2c30ce" />

<img width="2337" height="652" alt="Image" src="https://github.com/user-attachments/assets/226d5c50-f91f-4533-a621-7f8f751fd43a" />

==
old proposal => proved wrong after hands on (not working for HND)
---
Scenario1:
prefill blk_size // decode blk_size = 1/4;
prefill TP // decode TP = 1/2

<img width="1872" height="908" alt="Image" src="https://github.com/user-attachments/assets/bbb20c28-5851-4b57-9854-f2977cec0e49" />

Potential Overhead:
Redundant blocks will be read when req num_tokens can't be divided by local_block_size. 
For example, remote block_size is 4, local block_size is 16, req_num_tokens=24, total 6 blocks used in remote. while total 2 blocks used in local which takes 32 tokens.

Comparing with using remote block_size to read:
codes change will be less using local block_size => using remote block_size will need significant change for build_connector_metadata and workflow.

---
Scenario2:
prefill blk_size // decode blk_size = 4;
prefill TP // decode TP = 1/2

[...]

PR #26410, opened 2025-10-08
[Nixl] Minor refactor to handshake related metadata

NickLucche:
This PR is aimed at improving code clarity by grouping some of the more obscure handshake logic behind Heterogenous/Homogeneous TP, related to how kv cache addresses are split.

It does two basic things:
 - group validation inside `add_remote_agent` in its own function. We have multiple invariants we're checking that need motivation, and I feel this is breaking the flow of the code.
 - group KV address related properties inside `KVInfo`: this is to clarify the use of `tp_ratio` or `target_remote_rank` as well as allowing a cleaner implementation as we move to support both D TP > P TP and the opposite, as this will entail modifying both.

PR #26759, opened 2025-10-14
[NIXL] heterogeneous block_size support

xuechendi:
## Purpose

To support scenarios when prefill and decode have their own preferred block_size. Ex: Prefill with CUDA(16 as block_size) and decode with Intel Gaudi(128 as block_size)

More details describe in #26744 

### Current status:
- Prefill BlockSize < DecodeBlockSize
    - NHD => accuracy is good
    - HND => accuracy is good
- Prefill BlockSize > DecodeBlockSize
    - NHD => accuracy is good
    - NHD => accuracy is good
  
CMD for test:

**use case 1:**
```
DECODE_BLOCK_SIZE=64 bash run_accuracy_test.sh
# expected 0.41
```
Task | Metric | Value
-- | -- | --
gsm8k | exact_match,strict-match | 0.45703125
gsm8k | exact_match_stderr,strict-match | 0.03119537
gsm8k | exact_match,flexible-extract | 0.44921875
gsm8k | exact_match_stderr,flexible-extract | 0.031149309

Accuracy is Ok.

```
DECODER_TP_SIZE=2 DECODE_BLOCK_SIZE=64 bash run_accuracy_test.sh
# expected 0.41
```
Task | Metric | Value
-- | -- | --
gsm8k | exact_match,strict-match | 0.44921875
gsm8k | exact_match_stderr,strict-match | 0.0311493
gsm8k | exact_match,flexible-extract | 0.4375
gsm8k | exact_match_stderr,flexible-extract | 0.031065

Accuracy is Ok.

**use case 2:**
```
PREFILL_BLOCK_SIZE=64 bash run_accuracy_test.sh 
# expected 0.41
```

Task | Metric | Value
-- | -- | --
gsm8k | exact_match,strict-match | 0.453125
gsm8k | exact_match_stderr,strict-match | 0.031173
gsm8k | exact_match,flexible-extract | 0.44921875
gsm8k | exact_match_stderr,flexible-extract | 0.0311493

when setting gpu_utilization=0.8, accuracy is good
When setting gpu_utilization=0.3, we might used up all block_ids, and we can't use tail block_ids for temp buffer to store prefill block before permute, accuracy might gets slight impact => detail please refer to below [Prefill Block Size < Decode Block Size ] - 3.get_finished()

```
DECODER_TP_SIZE=2 PREFILL_BLOCK_SIZE=64 bash run_accuracy_test.sh 
# expected 0.41
```

Task | Metric | Value
-- | -- | --
gsm8k | exact_match,strict-match | 0.4609375
gsm8k | exact_match_stderr,strict-match | 0.031215
gsm8k | exact_match,flexible-extract | 0.44921875
gsm8k | exact_match_stderr,flexible-extract | 0.031149

#### Case 1: nP < nD
PREFILL block size < Decode block size: (example block_size_ratio = 0.25)

1. add_remote_agent
    1.1 we register remote address using remote layout
    1.2 we create a new local_xfer_handler using remote block_len, so it can do "one on one" remote to local copy
```
# remote:               | 0| 1| 2| 3| 4| 5| 6| 7| 8| 9|10|11|12|
# local origin:|          0|          1|          8|         12|
# local mapped:| 0| 1| 2| 3| 4| 5| 6| 7| 8| 9|10|11|12|13|14|15|
```

2. read_blocks
    2.1 remap local_block_ids: block [1, 2, 3, 4] => block [4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
    -> In that case, local blocks and remote blocks are size and count alligned 

3. get_finished()
    3.1 For HND, do permute on 4 blocks buffer
```
# remote is |h0-b0|h1-b0|h2-b0|h3-b0|h0-b1|h1-b1|h2-b1|h3-b1|...
# local is  |h0-b0..................|h1-b0..................|...
```
```
          # permute is to:
          # 1. view => view remote as n_blocks * remote_shape(H,remoteN,D)
          # 2. permute => (H, nblocks, remoteN, D)
          # 3. flatten => (H, localN, D)
```
   3.2 For NHD, no permute needed
        

#### Case 2: nP > nD
For Prefill Block Size < Decode Block Size (example block_size_ratio = 4)

1. add_remote_agent:
```
# remote: |         0|          1|          8|         12|
# local:          | 0| 1| 2| 3| 4| 5| 6| 7| 8| 9|10|11|12|
```

2. read_blocks
    2.1 re-map remote_block_ids => block[0] => block [1, 2, 3, 4]
    -> In that case, local blocks and remote blocks are size and count alligned 
    2.2 If len(local_block_ids) < len(after_mapping_remote_block_ids)
    -> allocate block_id from end of the blockAllocator to local_block_ids
    -> For ex: [1, 2, pad, pad ] => [1, 2, 12416, 12415] (Reason is at 3.1)

3. get_finished()
    3.1 For HND, do permute on 4 blocks buffer
          ///// is not used token, it's hidden in copied buffer, we will permute to move them to the tail
          Ex: we don't need block3, but with larger block_size, it was part of each head. We need to firstly
               copy entire buffer, then permute to move block3 to tail and get rid of it.
               That is also why need a temp buffer in local to have enough to store remote buffer
```
# remote is  |h0-b0............./////|h1-b0............./////|h3-b0............./////|h4-b0............./////|
# local is   |h0-b0|h1-b0|h2-b0|h3-b0|h0-b1|h1-b1|h2-b1|h3-b1|h0-b2|h1-b2|h2-b2|/////|/////|/////|/////|/////|
```

```
          # 1. view => view remote as (-1, H, n_blocks, localN, D)
          # 2. permute => (-1, nblocks, H, localN, D)
          # 3. flatten => (-1, H, localN, D)
```
   3.2 For NHD, no permute needed

## Test Result

---
<details>

</details>


vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1094,40 +1096,37 @@ def add_remote_agent(
...
             if tp_ratio > 1:
                 # Heterogeneous TP expects same kv_cache_layout.
                 assert nixl_agent_meta.kv_cache_layout == self.kv_cache_layout
                 if self.device_type == "xpu":
                     raise ValueError("Heterogeneous TP is not supported on XPU")
 
-            assert remote_block_len == self.block_len_per_layer[0] * tp_ratio, (
+            assert (
+                remote_block_len
+                == self.block_len_per_layer[0] * tp_ratio * block_size_ratio
```
NickLucche: let's get this validation changes in after https://github.com/vllm-project/vllm/pull/26410 so it's easier grouping

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1078,43 +1095,10 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
...
-            # could create fewer, but then _get_block_descs_ids needs to
-            # select agent_meta.num_blocks instead of self.num_blocks for
-            # local descr, and that makes handling regular flow less clean.
-            for block_id in range(self.num_blocks):
-                block_offset = block_id * self.block_len_per_layer[i]
-                addr = base_addr + block_offset
-                # (addr, len, device id)
-                blocks_data.append((addr, kv_block_len, self.device_id))
+        self.seen_base_addresses = seen_base_addresses
+        self.src_xfer_side_handle = self.register_local_xfer_handler(self.block_size)
```
xuechendi, line 1253: self.register_local_xfer_handler might get called twice, 
- during register kv
- after handshake with remote(only happened when blksizeP < blkSizeD)

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -823,12 +817,17 @@ def __init__(self, vllm_config: VllmConfig, engine_id: str):
 
         # nixl_prepped_dlist_handle.
         self.src_xfer_side_handle: int = 0
+        self.src_xfer_side_handles: dict[int, int] = {}
```
NickLucche, line 893: note: to merge after P TP > D TP where this is a `dict[int, list[int]]`

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -947,6 +946,24 @@ def _nixl_handshake(
             remote_agent_name = self.add_remote_agent(
                 metadata, p_remote_rank, remote_tp_size
             )
+            if metadata.block_size < self.block_size:
+                # when prefill with small block_size, we need to init a
+                # new handler with same block_len to match
```
NickLucche: this if-else handling different block_size logic can be moved into `add_remote_agent`

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -947,6 +946,24 @@ def _nixl_handshake(
...
+                self.src_xfer_side_handles[metadata.block_size] = (
+                    self.register_local_xfer_handler(metadata.block_size)
+                )
```
NickLucche: The `src_xfer_side_handles` indexing here will clash with P TP > D TP one: here we index by remote block_size, while in the other PR we index by tp_ratio (or remote tp_size equivalently).

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -947,6 +946,24 @@ def _nixl_handshake(
             remote_agent_name = self.add_remote_agent(
                 metadata, p_remote_rank, remote_tp_size
             )
+            if metadata.block_size < self.block_size:
+                # when prefill with small block_size, we need to init a
```
NickLucche: nit: 

```suggestion
                # When prefill block_size is smaller, we need to init a
```

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1155,43 +1172,10 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
...
-        )
+        self.seen_base_addresses = seen_base_addresses
+        self.src_xfer_side_handle = self.register_local_xfer_handler(self.block_size)
 
-        descs = self.nixl_wrapper.get_xfer_descs(blocks_data, self.nixl_memory_type)
-        # NIXL_INIT_AGENT to be used for preparations of local descs.
-        self.src_xfer_side_handle = self.nixl_wrapper.prep_xfer_dlist(
-            "NIXL_INIT_AGENT", descs
-        )
+        self.src_xfer_side_handles[self.block_size] = self.src_xfer_side_handle
```
NickLucche, line 1255: this clean-up is nice :)

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1227,8 +1211,51 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
             kv_cache_layout=self.kv_cache_layout
             if not self.use_host_buffer
             else self.host_buffer_kv_cache_layout,
+            block_size=self.block_size,
+        )
+
+    def register_local_xfer_handler(
+        self,
+        block_size: int,
+    ) -> int:
```
NickLucche, line 1297: nit: we can add a docstring here as to explain "default" registering vs different block_size one

NickLucche: I think I can re-use `register_local_xfer_handler` for local descs I need here https://github.com/vllm-project/vllm/pull/27274/files#diff-23f718da8b27c870b58872baf0388020c9c4c47c831545481295608897749823R1269.
Nice one.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1297,13 +1324,38 @@ def add_remote_agent(
 
         # Create dst descs and xfer side handles. TP workers have same #blocks
         # so we only register once per engine_id.
+        # All attn in vLLM uses blocks starts with 1st(0 is for empty)
+        # For hetero block size case, block 0 should always remote block_len
```
NickLucche: I didn't get the comment on this particular line

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1297,13 +1324,38 @@ def add_remote_agent(
...
+        # block_size_ratio < 1:
+        # remote:               | 0| 1| 2| 3| 4| 5| 6| 7| 8| 9|10|11|12|
+        # local origin:|          0|          1|          8|         12|
+        # local mapped:| 0| 1| 2| 3| 4| 5| 6| 7| 8| 9|10|11|12|13|14|15|
+        # block_size_ratio > 1:
+        # remote:     |          0|          1|          8|         12|
+        # new remote: | 0| 1| 2| 3| 4| 5| 6| 7| 8| 9|10|11|12|13|14|15|
+        # local:               | 0| 1| 2| 3| 4| 5| 6| 7| 8| 9|10|11|12|
+        remote_block_size = nixl_agent_meta.block_size
+        block_size_ratio = remote_block_size / self.block_size
```
NickLucche: what do you think of a uniform tp/block_ratio semantics, with negative block/tp_ratio meaning remote > local?

https://github.com/vllm-project/vllm/pull/27274/files#diff-23f718da8b27c870b58872baf0388020c9c4c47c831545481295608897749823R593

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
NickLucche: even if block 0 is empty and just padding, I don't think we should care at this level of abstraction right? It just won't be passed in from scheduler

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
NickLucche: don't have a super strong opinion on the standard, as long as we choose one recognizable way

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1297,13 +1324,38 @@ def add_remote_agent(
...
+        self.dst_block_size_ratio[engine_id] = block_size_ratio
+
+        expanded_num_blocks = math.ceil(block_size_ratio)
+
```
NickLucche, line 1423:
```suggestion
remote_blocks_adjusted = nixl_agent_meta.num_blocks * expanded_num_blocks
```
or a similar name

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1319,27 +1371,41 @@ def add_remote_agent(
...
             )
             for block_id in range(nixl_agent_meta.num_blocks):
                 block_offset = block_id * nixl_agent_meta.block_lens[i]
-                # For each block, grab the heads chunk belonging to rank_i
-                # of size remote_nheads // tp_ratio, which correspond to
-                # self.block_len == remote_block_len//tp_ratio bytes.
-                addr = base_addr + block_offset + rank_offset
-                # (addr, len, device id)
-                blocks_data.append((addr, kv_block_len, nixl_agent_meta.device_id))
+                for sub_block_id in range(expanded_num_blocks):
```
NickLucche: can you comment some like "this is >1 just for remote block_size > local"?
I feel like nesting fors keep increasing complexity in this method

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1358,7 +1424,7 @@ def add_remote_agent(
         return remote_agent_name
 
     def _validate_remote_agent_handshake(
-        self, nixl_agent_meta: NixlAgentMetadata, remote_tp_size: int
+        self, nixl_agent_meta: NixlAgentMetadata, remote_tp_size: int, num_blocks: int
```
NickLucche: I think we should check remote block size is a nice multiple of local block size

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -823,12 +817,17 @@ def __init__(self, vllm_config: VllmConfig, engine_id: str):
...
         # nixl_prepped_dlist_handle.
         self.src_xfer_side_handle: int = 0
+        self.src_xfer_side_handles: dict[int, int] = {}
         # Map of engine_id -> nixl_prepped_dlist_handle (int)].
         self.dst_xfer_side_handles: dict[EngineId, int] = {}
 
         # Map of engine_id -> num_blocks. All ranks in the same deployment will
         # have the same number of blocks.
         self.dst_num_blocks: dict[EngineId, int] = {}
+        self.dst_block_size_ratio: dict[EngineId, float] = {}
```
NickLucche: we could have a similar helper method to `kv_topo.tp_size` here with `kv_topo.block_size_ratio`

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1542,6 +1686,20 @@ def get_finished(self) -> tuple[set[str], set[str]]:
                 self.sync_recved_kv_to_device(req_id, meta)
             if self.enable_permute_local_kv:
                 block_ids_to_permute += meta.local_block_ids
+
+            # post processing for heteroblocksize
+            block_size_ratio = self.dst_block_size_ratio[meta.remote_engine_id]
+            if block_size_ratio not in block_ids_for_blocksize_post_process:
+                block_ids_for_blocksize_post_process[block_size_ratio] = []
```
NickLucche: this is a defaultdict(list)

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1777,14 +1967,18 @@ def _get_block_descs_ids(
                 region_ids = np.arange(layer_idx, layer_idx + 1)
 
         num_blocks = self.dst_num_blocks[engine_id]
+        num_blocks = np.full((self.num_regions), num_blocks)
 
         # Compute the desc ids for each block.
-        region_ids = region_ids[:, None]
         block_ids = np.array(block_ids)[None, :]
-        descs_ids = region_ids * num_blocks + block_ids
+        region_nblocks = region_ids * num_blocks
```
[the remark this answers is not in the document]
NickLucche: looks legit

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1926,6 +2165,74 @@ def shutdown(self):
...
+        self.available_block_ids: list[int] = [-id for id in range(total_num_blocks)]
+        self.assigned_num_blocks = assigned_num_blocks
+        logger.info(
+            "BlockAllocatorForHeteroBlockSize is initialized, using %d - %d blocks "
+            "as buffer blocks for temporary remote larger block_size cache",
+            (total_num_blocks - assigned_num_blocks),
+            total_num_blocks,
+        )
+
+        heapq.heapify(self.available_block_ids)
```
NickLucche: why do we need a heap here?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1926,6 +2165,74 @@ def shutdown(self):
         self._registered_descs.clear()
 
 
+class BlockAllocatorForHeteroBlockSize:
+    def __init__(self, total_num_blocks: int, assigned_num_blocks: int):
+        assert total_num_blocks > 0, "No Available blocks"
+        self.available_block_ids: list[int] = [-id for id in range(total_num_blocks)]
+        self.assigned_num_blocks = assigned_num_blocks
```
NickLucche: this looks to complex impl for a a tmp block "buffer" class

NickLucche (changes requested):
Thanks for the good work on this PR @xuechendi !

I think the code is already mostly in good shape.
What worries me the most is the added complexity that this (hard) feature brings in.  
In particular in the `BlockAllocatorForHeteroBlockSize` I think we're making two mistakes here: 
- it's a complex implementation for a staging area  
- block allocation should be handled by the scheduler, so I feel like we're crossing responsibilities here since we also have a scheduler connector we could use.

Most importantly, it seems most of the complexities stems from supporting remote (P) block_size > local block_size (D) where you need more space than initially allocated, while afaiu the use-case you primarily care about is the opposite (remote < local).

In the interest of maintainability, I would suggest we split the PR:
- PR1: small refactor with `register_local_xfer_handler` that we can land quickly and re-use in the P TP > D TP case
- PR2: remote block_size < local block_size which looks easier to maintain and already in good state here
- PR3: remote block_size > local block_size, *to really understand the importance of the use-cases for this one first* (given complexity). Then likely moving the blocks allocation logic into the scheduler or just use a fixed stage area. 

Let me know what you think!

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: How about I update it to self.src_xfer_side_handles: dict[int, dict[int, list[int]] later? (block_size, tp_ratio, list of handles)

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: Done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: Yes, remove the confusing comment now, previously, I plan to skip block 0 but I realized that was not necessary.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: Right! updated!

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: remove nP > nD path

tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh
```diff
@@ -49,6 +49,8 @@ NUM_DECODE_INSTANCES=${NUM_DECODE_INSTANCES:-1}   # Default to 1
 PREFILLER_TP_SIZE=${PREFILLER_TP_SIZE:-1}
 DECODER_TP_SIZE=${DECODER_TP_SIZE:-1}
 GPU_MEMORY_UTILIZATION=${GPU_MEMORY_UTILIZATION:-0.2}
+PREFILL_BLOCK_SIZE=${PREFILL_BLOCK_SIZE:-16}
+DECODE_BLOCK_SIZE=${DECODE_BLOCK_SIZE:-16}
```
NickLucche, line 53: nit: can we default to 128? Block size can significantly impact our transfer efficiency, just having it as kind of suggested value

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -686,13 +681,36 @@ def tp_ratio(
...
+            Calculate the block size ratio between local and remote TP.
+            """
+            assert (
+                self.block_size % remote_block_size == 0
+                or remote_block_size % self.block_size == 0
+            ), (
+                f"Local block size {self.block_size} is not divisible "
+                f"by remote block size {remote_block_size} or vice versa."
+            )
+            return remote_block_size / self.block_size
```
NickLucche: did you consider inverting the ratio like we do for tp_ratio? https://github.com/vllm-project/vllm/pull/27274/files#diff-23f718da8b27c870b58872baf0388020c9c4c47c831545481295608897749823R562
`local/remote`

Don't have a strong preference, but we should definitely pick one

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1227,8 +1220,51 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
...
             if not self.use_host_buffer
             else self.host_buffer_kv_cache_layout,
+            block_size=self.block_size,
+        )
+
+    def register_local_xfer_handler(
+        self,
+        block_size: int,
+    ) -> int:
+        block_size_ratio = self.block_size // block_size
```
NickLucche, line 1309: let's make this consistent with `kv_topo.block_size_ratio`

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
NickLucche: still need to doc here :)

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1355,6 +1406,13 @@ def add_remote_agent(
             remote_agent_name, descs
         )
 
+        if block_size_ratio < 1:
+            # when prefill with small block_size, we need to init a
```
NickLucche: nit: has smaller block_size

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1401,33 +1462,26 @@ def _validate_remote_agent_handshake(
...
-                "KV cache sizes must match between P and D when replicated"
-            )
-            remote_block_size = remote_block_len // (self.slot_size_per_layer[0])
+            for i in range(len(self.block_len_per_layer)):
+                assert (
+                    self.block_len_per_layer[i] * block_size_ratio
+                    == nixl_agent_meta.block_lens[i]
+                ), "KV cache sizes must match between P and D when replicated"
         else:
             # When MLA is not used, this is a list of the same block length
```
NickLucche, line 1471: pretty sure we still need this

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1511,6 +1565,59 @@ def permute_device_kv(self, block_ids: list[int]):
                 )
                 cache.index_copy_(0, indices, permuted_blocks)
 
+    def blocksize_post_process(self, block_ids_per_ratio: dict[float, list[list[int]]]):
+        def _process_local_gt_remote(blocks_to_update, block_size_ratio):
+            n_kv_heads, block_size, head_size = blocks_to_update.shape[1:]
+            remote_block_size = int(block_size * block_size_ratio)
+            n_blocks = int(1 / block_size_ratio)
```
NickLucche: probably would be easier with local/remote

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1511,6 +1565,59 @@ def permute_device_kv(self, block_ids: list[int]):
                 )
                 cache.index_copy_(0, indices, permuted_blocks)
 
+    def blocksize_post_process(self, block_ids_per_ratio: dict[float, list[list[int]]]):
+        def _process_local_gt_remote(blocks_to_update, block_size_ratio):
+            n_kv_heads, block_size, head_size = blocks_to_update.shape[1:]
```
NickLucche, line 1652: @xuechendi have you tested with MLA? this line wouldn't work here

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1511,6 +1565,59 @@ def permute_device_kv(self, block_ids: list[int]):
...
+                .permute(0, 2, 1, 3, 4)
+                .flatten(2, 3)
+            )
+            return permuted_blocks
+
+        split_k_and_v = not (self.use_mla or self._use_pallas or self._use_flashinfer)
+        sample_cache = list(self.device_kv_caches.values())[0][0]
+        for block_size_ratio, block_ids_list in block_ids_per_ratio.items():
+            if len(block_ids_list) == 0:
+                continue
```
NickLucche: should we be checking this condition? I don't think we would add a block if it's size 0

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1511,6 +1565,59 @@ def permute_device_kv(self, block_ids: list[int]):
...
+            if block_size_ratio < 1:
+                fn = _process_local_gt_remote
+                block_ids_list = [
+                    [item for sublist in block_ids_list for item in sublist]
+                ]
+            for block_ids in block_ids_list:
+                if len(block_ids) == 0:
```
NickLucche: same here

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1511,6 +1565,59 @@ def permute_device_kv(self, block_ids: list[int]):
...
+            )
+            return permuted_blocks
+
+        split_k_and_v = not (self.use_mla or self._use_pallas or self._use_flashinfer)
+        sample_cache = list(self.device_kv_caches.values())[0][0]
+        for block_size_ratio, block_ids_list in block_ids_per_ratio.items():
+            if len(block_ids_list) == 0:
+                continue
+            if block_size_ratio < 1:
+                fn = _process_local_gt_remote
```
NickLucche: there's only one function now

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1710,6 +1828,15 @@ def _read_blocks(
         dst_engine_id: str,
         request_id: str,
     ):
+        block_size_ratio = self.kv_topo.block_size_ratio_from_engine_id(dst_engine_id)
+        block_size_ratio_inv = None
+        if block_size_ratio < 1:
+            block_size_ratio_inv = int(1 / block_size_ratio)
```
NickLucche: this could also be removed if we did local/remote

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1710,6 +1828,15 @@ def _read_blocks(
...
+            local_block_ids = self.get_mapped_blocks(
+                np.asarray(local_block_ids), block_size_ratio_inv
+            )
+            if len(local_block_ids) > len(remote_block_ids):
```
NickLucche, line 1925: qq: when is this happening?

NickLucche (commented):
Looking cleaner already @xuechendi !
Left some comments, among other things to check: MLA and local/remote ratio order.

tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh
xuechendi: cuda block_size is currently default to 16, should I change to a large number

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: Since now get_mapped_blocks will always expand block_ids for n times.
when prefill only needs 10 blocks with block_size as 4 =>[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
For decode with large block_size =16, it will be [1, 2, 3]
and get_mapped_blocks will expland the list [1, 2, 3] to [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
Then we still need to clip local from [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12] to [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: I didn't get it, which line we need to keep?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: I tested MLA, since MLA kv_layout is None, so it will never enter _process_local_gt_remote path,
To make sure it also not fail for HND, I added a check here, so for MLA, no post_process is needed
image

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: right, I removed the check

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: fixed

xuechendi:
> Looking cleaner already @xuechendi ! Left some comments, among other things to check: MLA and local/remote ratio order.

For MLA, I added a skip for post-process since we don't need it for single kv_head. Verified with GSM8K accuracy test locally and with both VLLM_KVCACHE_LAYOUT=HND and NHD

tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh
NickLucche: yep can we set it to 128 at these lines?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -686,13 +689,37 @@ def tp_ratio(
...
+            """
+            assert (
+                self.block_size % remote_block_size == 0
+                or remote_block_size % self.block_size == 0
+            ), (
+                f"Local block size {self.block_size} is not divisible "
+                f"by remote block size {remote_block_size} or vice versa."
+            )
+            ret = self.block_size / remote_block_size
+            return ret if ret < 0 else int(ret)
```
NickLucche: we don't support `remote_block_size > self.block_size`, so I think we should just assume ret >=1.
Also we should just check `self.block_size % remote_block_size == 0` otherwise we're effectively allowing remote>local.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
NickLucche: good!

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1511,6 +1585,53 @@ def permute_device_kv(self, block_ids: list[int]):
...
+                .permute(0, 2, 1, 3, 4)
+                .flatten(2, 3)
+            )
+            return permuted_blocks
+
+        split_k_and_v = not (self.use_mla or self._use_pallas or self._use_flashinfer)
+        sample_cache = list(self.device_kv_caches.values())[0][0]
+        for block_size_ratio, block_ids_list in block_ids_per_ratio.items():
+            assert block_size_ratio > 1, "Only nP < nD supported currently."
+            fn = _process_local_gt_remote
```
NickLucche: nit: can we drop `fn` and just use `_process_local_gt_remote` directly?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1542,6 +1664,20 @@ def get_finished(self) -> tuple[set[str], set[str]]:
...
             if self.enable_permute_local_kv:
                 block_ids_to_permute += meta.local_block_ids
+
+            # post processing for heteroblocksize
+            block_size_ratio = self.kv_topo.block_size_ratio_from_engine_id(
+                meta.remote_engine_id
+            )
+            if (
+                not self.use_mla
+                and block_size_ratio != 1
```
NickLucche: nit: `block_size_ratio > 1` due to original remote < local constraint. It just helps keeping things clearer and consistent

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1834,8 +1991,26 @@ def _read_blocks(
                 self.nixl_wrapper.release_xfer_handle(handle)
             self._failed_recv_reqs.add(request_id)
 
+    def get_mapped_blocks(self, block_ids, block_size_ratio):
+        """
+        Calculates the new set of block IDs by mapping every element
+        in the (potentially sparse) input array.
```
NickLucche:
```suggestion
        Calculates the new set of block IDs by mapping every element
        in the (potentially sparse) input array.
        Example: block_ids=[0, 2], block_size_ratio=2
      get_mapped_blocks    0     1     [2     3]     4     5
            # remote is |h0-b0|h1-b0||h0-b1|h1-b1||h0-b1|h1-b1||
            # local is  |h0-b0......||h1-b0......||h2-b0........
      local_block_ids         0           [1]           2        
```

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
NickLucche: I see, nice one, I think it could use a comment line as it might be not obvious

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1762,13 +1910,17 @@ def _read_blocks(
         # Get descs ids.
         local_block_descs_ids: np.ndarray
         remote_block_descs_ids: np.ndarray
+
         if not self.block_window_per_layer:
             # Default case: assume global attention
             remote_block_descs_ids = self._get_block_descs_ids(
-                dst_engine_id, remote_block_ids
+                dst_engine_id,
+                remote_block_ids,
```
NickLucche, line 1998: nit: was this reformat needed?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1752,7 +1895,12 @@ def _read_blocks(
             remote_block_ids = remote_block_ids[-num_local_blocks:]
 
         # Get side handles.
-        local_xfer_side_handle = self.src_xfer_side_handle
+        block_size = (
+            self.block_size
+            if block_size_ratio <= 1
+            else int(self.block_size // block_size_ratio)
+        )
+        local_xfer_side_handle = self.src_xfer_side_handles[block_size]
```
NickLucche:
```suggestion
        remote_block_size = self.kv_topo.remote_block_size[dst_engine_id]
        local_xfer_side_handle = self.src_xfer_side_handles[remote_block_size]
```

NickLucche (commented):
Left some comments, thanks for your patience here @xuechendi !

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: yes, assert remote_block_size <= self.block_size is done in `self._nixl_handshake`, check is before calling `add_remote_agent`, so nP > nD is not allowed.

For This function, I would like to keep this option open since the nP > nD will never arrive to this point.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: Sure

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: Comment added

tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh
xuechendi, line 53: I switched to 128 but switched back to 16 again.
I tested with 128, and noticed that even on origin/main, accuracy is not correct at this moment.
Will see if I can find out main reason in separate PR.

xuechendi:
@NickLucche , comments are mostly resolved, only one is for default block_size, I explained in another comments, I verified accuracy on this Branch and current main, both accuracy is not correct. So I switched back to 16

tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh
xuechendi: I verified setting block_size=64 works, but somehow using block_size=128 for CUDA gets tensor Nan.
@NickLucche , do you want me to set as 64? Actually I refer to use current default for CUDA which is 16/

tests/v1/kv_connector/nixl_integration/test_accuracy.py
```diff
@@ -52,6 +52,7 @@ def test_accuracy():
         model="local-completions",
         model_args=model_args,
         tasks=TASK,
+        limit=256,
```
NickLucche: can you revert this line?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
NickLucche: I see your point, but I still find it could be confusing to readers.
 
```suggestion
            assert self.block_size % remote_block_size == 0, (
                f"Local block size {self.block_size} is not divisible "
                f"by remote block size {remote_block_size}."
            )
            return self.block_size // remote_block_size
```

Happy to add it back if/when we land the opposite case.

NickLucche (approved):
LGTM, thanks for your patience @xuechendi! Left two minor comments but it would be nice if you could address those quickly before we turn auto-merge on.

tests/v1/kv_connector/nixl_integration/test_accuracy.py
xuechendi: yes

PR #26761, opened 2025-10-14
Fix qwen3 omni audio lengths #26630

baonudesifeizhai:
FIX #26630
Fix audio processing bug in Qwen3-Omni model that caused split_with_sizes RuntimeError when processing audio inputs. The issue was caused by incorrect calculation of audio_feature_lengths in the model's audio processing pipeline.
 
 

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/models/qwen3_omni_moe_thinker.py
+++ b/vllm/model_executor/models/qwen3_omni_moe_thinker.py
@@ -710,21 +710,12 @@ class Qwen3OmniMoeThinkerMultiModalProcessor(
                 x = np.pad(x, (0, pad_length), mode="constant", constant_values=0)
             return x
 
-        feature_extractor = None
-        original_chunk_length = None
-
         # NOTE: WhisperFeatureExtractor cannot handle empty list of audios
         if audios:
             # NOTE: Qwen3-Omni processor accept "audio"
             # To make sure the cache works with padding=True, we pre-padded
             # the audio to multiple of hop_length.
-            feature_extractor = self.info.get_feature_extractor()
-            hop_length = feature_extractor.hop_length
-
-            # Disable chunk_length limit to support long audio (>30s)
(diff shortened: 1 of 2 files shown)


vllm/model_executor/models/qwen3_omni_moe_thinker.py
```diff
@@ -733,25 +733,6 @@ def pad_to_hop_length(x: np.ndarray, hop_length: int) -> np.ndarray:
...
-                num_frame = (
-                    (audio_length // hop_length)
-                    if audio_length % hop_length == 0
-                    else (audio_length // hop_length - 1)
-                )
-                audio_num_frames.append(num_frame)
-            hf_inputs["feature_attention_mask"] = [
-                torch.ones(num_frame) for num_frame in audio_num_frames
-            ]
-            hf_inputs["audio_feature_lengths"] = torch.tensor(audio_num_frames)
```
Isotr0py, line 754: Hmmm, this is intentional to make sure mm_cache can hit audio item correctly, because `padding=True` will create different attention mask when audios are padded/unpadded.

You can simply run the following test for checking:
```
pytest -s -v tests/models/multimodal/processing/test_common.py -k qwen3-omni -v
```
```
  Full diff:
    {
        'audio_feature_lengths': tensor([5, 4]),
  -     'feature_attention_mask': [
  +     'feature_attention_mask': tensor([[1, 1, 1, 1, 1],
  -         tensor([1, 1, 1, 1, 1], dtype=torch.int32),
  ?         -------             ^
  +         [1, 1, 1, 1, 0]], dtype=torch.int32),
  ?                      ^^
  -         tensor([1, 1, 1, 1], dtype=torch.int32),
  -     ],
```

baonudesifeizhai: my bad 
if they chunk the audio less then 30s  , it works ..

PR #24002, opened 2025-08-31
[V1][CUDA Graph] Fix attention metadata tensor sizes for padded batches

ayushsatyam146:
Fixes #23789. Move CUDA graph padding before attention metadata construction to ensure tensor sizes match execution environment. Fixes crashes in FlashMLA and other attention backends caused by metadata/execution size mismatches.

- Pass padded token count to _prepare_inputs()
- Update slot_mapping and num_actual_tokens to use padded sizes
- Eliminates need for backend-specific workarounds

Issue #23789, opened 2025-08-28
[Attention]: Pad for cudagraphs before constructing attention metadata


LucasWilkinson:
### 🚀 The feature, motivation and pitch

Currently we pad for cudagraphs:

https://github.com/vllm-project/vllm/blob/853c371fc33e7c99aa2ab9f6e2cd7cbd1cadcf99/vllm/v1/worker/gpu_model_runner.py#L1497

after constructing the metadata:

https://github.com/vllm-project/vllm/blob/853c371fc33e7c99aa2ab9f6e2cd7cbd1cadcf99/vllm/v1/worker/gpu_model_runner.py#L820-L909

This creates headaches for making attention backends with ahead-of-time schedulers cudagraph compatible e.g. FlashMLA and FlashAttn. This is because the scheduler might be called with a batch size smaller then what the graph is running with; e.g. we recently had to work around this in FlashMLA (https://github.com/vllm-project/FlashMLA/pull/3). Moving the padding calculation before metadata construction and passing padded `CommonAttentionMetadata` is much better long term solution though since the we won't have to work around this in future attention kernels.

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

ayushsatyam146:
@LucasWilkinson I can work on it. Please assign it to me, and I will start working on it right away!

LucasWilkinson:
Thank you!

ayushsatyam146:
Hi @LucasWilkinson the PR is up for this issue. PTAL thanks!

PR #26764, opened 2025-10-14
[Core] Separate out attention metadata building logic from prepare inputs

LucasWilkinson:
Preparatory refactor PR for: https://github.com/vllm-project/vllm/pull/24002 (https://github.com/vllm-project/vllm/issues/23789)

Separate out the attention metadata building logic from prepare inputs so we can re-use it for dummy_runs and to setup for padding for cuda-graphs before building attention metadata (see https://github.com/vllm-project/vllm/issues/23789 for motivation)

Moving `_compute_cascade_attn_prefix_lens` out of metadata building loops since whether or not we are using cascade  attention is needed by cudagraph dispatcher that will happen before attention metadata building in https://github.com/vllm-project/vllm/pull/24002.

NOTE: renamed `kv_cache_group_id` to `kv_cache_gid` (common acronym in linux) to make more lines fit within the formatter max line width

### DIFF
--- a/vllm/v1/worker/gpu_model_runner.py
+++ b/vllm/v1/worker/gpu_model_runner.py
@@ -1489,7 +1489,6 @@ class GPUModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
 
         return attn_metadata, spec_decode_common_attn_metadata
 
-
     def _compute_cascade_attn_prefix_lens(
         self,
         num_scheduled_tokens: np.ndarray,
--- a/vllm/v1/worker/gpu_model_runner.py
+++ b/vllm/v1/worker/gpu_model_runner.py
@@ -1490,7 +1490,7 @@ class GPUModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
     ) -> list[list[int]] | None:
         """
         :return: Optional[common_prefix_lens]
-                   common_prefix_lens is 2D: [kv_cache_group_id][attn_group_idx],
+                   common_prefix_lens is 2D: ``[kv_cache_group_id][attn_group_idx]``,
                    None if we should not use cascade attention
         """
         use_cascade_attn = False
--- a/vllm/v1/worker/gpu_model_runner.py
+++ b/vllm/v1/worker/gpu_model_runner.py
@@ -1320,7 +1320,7 @@ class GPUModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
         use_spec_decode: bool = False,
         for_cudagraph_capture: bool = False,
         scheduled_encoder_inputs: dict[str, list[int]] | None = None,
-        common_prefix_lens: list[list[int]] | None = None,
+        cascade_attn_prefix_lens: list[list[int]] | None = None,
     ) -> tuple[PerLayerAttnMetadata, CommonAttentionMetadata | None]:
         """
         :return: tuple[attn_metadata, spec_decode_common_attn_metadata]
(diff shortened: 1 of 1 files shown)


vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1387,45 +1399,66 @@ def _prepare_inputs(
...
+                    prefix_len = self._compute_cascade_attn_prefix_len(
+                        num_scheduled_tokens,
+                        num_common_prefix_blocks[kv_cache_gid],
+                        attn_group.kv_cache_spec,
+                        attn_group.get_metadata_builder(),
+                    )
+                common_prefix_lens[kv_cache_gid].append(prefix_len)
+                use_cascade_attn |= prefix_len > 0
+
+        return common_prefix_lens if use_cascade_attn else None
```
fhl2000: Why change the logic for `use_cascade_attn`?  I think previously `use_casecade` given by each attention builder is not only determined by the `prefix_len` but may also be other factors  (like in https://github.com/vllm-project/vllm/blob/49c00fe304d4890f87dbcd99aad9f46222c22111/vllm/v1/attention/backends/flash_attn.py#L827, though I have no idea why it is unused now). And some attention backends just do not support cascade attention.

fhl2000 (commented):
Hi @LucasWilkinson, thank you for refactoring this! Just left one question, otherwise looking great.

vllm/v1/worker/gpu_model_runner.py
fhl2000: Sorry for this. Just realized that you are right.  It already uses `attn_metadata_builder.use_cascade_attention()` inside the `self._compute_cascade_attn_prefix_len`, so prefix_len is already the corrected version.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1384,45 +1396,66 @@ def _prepare_inputs(
...
+                    prefix_len = 0
+                else:
+                    prefix_len = self._compute_cascade_attn_prefix_len(
+                        num_scheduled_tokens,
+                        num_common_prefix_blocks[kv_cache_gid],
+                        attn_group.kv_cache_spec,
+                        attn_group.get_metadata_builder(),
+                    )
+                common_prefix_lens[kv_cache_gid].append(prefix_len)
+                use_cascade_attn |= prefix_len > 0
```
fhl2000: Although it's correct, could you add comments here explaining why `prefix_len>0` is the only source for `use_cascade_attn`, as it can be confusing because the `prefix_len` is not the actual prefix length but rather a processed value after some justifications.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -3328,55 +3390,14 @@ def _dummy_run(
...
-                        for layer_name in attn_group.layer_names:
-                            attn_metadata[layer_name] = attn_metadata_i
+            # Build attention metadata using the unified method
+            attn_metadata, _ = self._build_attention_metadata(
+                total_num_scheduled_tokens=num_tokens,
+                max_num_scheduled_tokens=max_query_len,
+                num_reqs=num_reqs,
+                ubatch_slices=ubatch_slices,
+                for_cudagraph_capture=True,
+            )
```
fhl2000, line 3538: I'm not sure why we use `num_tokens` for attention building while using `num_token_after_padding` for CUDA graph dispatching below.  Would it cause a potential hang for Flashinfer because it previously hung due to attn_metadata not being correctly prepared?

fhl2000 (commented):
Overall looks good! Any chance to also separate the `attn_metadata` building logic into a function for the Eagle drafter? So we can later reuse them for both the drafter execution and its dummy run (prepared for full CUDA graph support).

vllm/v1/worker/gpu_model_runner.py
LucasWilkinson: padding before attention metadata building is the follow up PR: https://github.com/vllm-project/vllm/pull/24002 you'll see in there we remove the FlashInfer workaround hack :+1:

vllm/v1/worker/gpu_model_runner.py
LucasWilkinson: added a comment and renamed common_prefix_lens for clarity

mgoin (approved):
LGTM, nothing clearly stood out as missed

Issue #26711, opened 2025-10-13
[Bug]: DeepSeek V3.2 running with MTP will cause DeepseekV32IndexerMetadata parsing error


RekingQin:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
Collecting environment information...
==============================
        System Info
==============================
OS                           : Ubuntu 22.04.5 LTS (x86_64)
GCC version                  : (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Clang version                : Could not collect
CMake version                : version 4.1.0
Libc version                 : glibc-2.35

==============================
       PyTorch Info
==============================
PyTorch version              : 2.8.0+cu128
Is debug build               : False
CUDA used to build PyTorch   : 12.8
ROCM used to build PyTorch   : N/A

==============================
      Python Environment
==============================
Python version               : 3.12.11 (main, Jun  4 2025, 08:56:18) [GCC 11.4.0] (64-bit runtime)
Python platform              : Linux-5.4.241-1-tlinux4-0017.7-x86_64-with-glibc2.35

==============================
       CUDA / GPU Info
==============================
Is CUDA available            : True
CUDA runtime version         : Could not collect
CUDA_MODULE_LOADING set to   : LAZY
GPU models and configuration : 
GPU 0: NVIDIA H20
GPU 1: NVIDIA H20
GPU 2: NVIDIA H20
GPU 3: NVIDIA H20
GPU 4: NVIDIA H20
GPU 5: NVIDIA H20
GPU 6: NVIDIA H20
GPU 7: NVIDIA H20

Nvidia driver version        : 535.161.08
cuDNN version                : Could not collect
HIP runtime version          : N/A
MIOpen runtime version       : N/A
Is XNNPACK available         : True

==============================
          CPU Info
==============================
Architecture:                    x86_64
CPU op-mode(s):                  32-bit, 64-bit
Address sizes:                   52 bits physical, 57 bits virtual
Byte Order:                      Little Endian
CPU(s):                          384
On-line CPU(s) list:             0-383
Vendor ID:                       AuthenticAMD
BIOS Vendor ID:                  Advanced Micro Devices, Inc.
Model name:                      AMD EPYC 9K84 96-Core Processor
BIOS Model name:                 AMD EPYC 9K84 96-Core Processor                
CPU family:                      25
Model:                           17
Thread(s) per core:              2
Core(s) per socket:              96
Socket(s):                       2
[...]

PR #25109, opened 2025-09-17
[Spec-Decode] Support piecewise cudagraphs for Eagle head

LucasWilkinson:
## Purpose

Support PIECEWISE cudagraphs with eagle head; in-between fix until https://github.com/vllm-project/vllm/pull/23679 can be refactored and landed. This should get use most of the performance of that with alot less complexity while the gpu model runner is refactored.

image

## Test Plan

```
vllm serve Qwen/Qwen3-Next-80B-A3B-Instruct -tp 4 --tokenizer-mode auto --speculative-config '{"method": "qwen3_next_mtp", "num_speculative_tokens": 2}' --port 3331 --max-num-batched-tokens 512 --gpu-memory-utilization 0.75

(vllm2) lwilkinson@H100-GPU17:~/code/vllm2$ python tests/evals/gsm8k/gsm8k_eval.py --port 3331
Running GSM8K evaluation: 1319 questions, 5-shot
Evaluating: 100%|████████████████████████████████████████████████████████████████████████████████████| 1319/1319 [03:48<00:00,  5.77it/s]

Results:
Accuracy: 0.835
Invalid responses: 0.000
Total latency: 228.584 s
Questions per second: 5.770
```

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #26779, opened 2025-10-14
[Bugfix] DeepSeek V3.2 MTP metadata & CUDA graph issues

xiaohajiayou:
## Purpose
(see [Issue #26711](https://github.com/vllm-project/vllm/issues/26711)).  
1. **Fix CUDA Graph Capture Crash in EAGLE**  
   Resolve `IndexError: list index out of range` when accessing `self.cudagraph_batch_sizes[-1]` during CUDA Graph capture 
   - **Root Cause**:  
     - When using PIECEWISE capture, `gpu_model_runner` passes `use_cudagraphs=True`, but MTP heads of draft models (e.g., DeepSeek MTP) lack native CUDA graph support.  
     - EAGLE sets `self.use_cuda_graph = False` and initializes `self.cudagraph_batch_sizes` as an empty list for such models.  
     - PR #25109 ignored `self.use_cuda_graph` in `dummy_run` arguments, leading to out-of-range access of the empty `cudagraph_batch_sizes` list.  
   - **Change**: In `vllm/v1/spec_decode/eagle.py:dummy_run`, gate padding and `cudagraph_runtime_mode` with `(use_cudagraphs and self.use_cuda_graph)` to prevent empty-list indexing and align with runner behavior.

2. **Fix DeepSeek V3.2 MTP Metadata & Sparse MLA Layer Selection**  
   Resolve incorrect layer classification and metadata mapping for DeepSeek V3.2 Sparse MLA.  
   - **Root Cause**:  
     - PR https://github.com/vllm-project/vllm/pull/25103 Separate MLAAttention class from Attention, but the Indexer layers were not excluded from attention layer lists, causing mismatches between MTP metadata and model components.  
     - Sparse MLA architectures require strict separation of attention and indexer layers to ensure proper metadata propagation for expert routing.  
   - **Change**: In `vllm/v1/spec_decode/eagle.py:load_model`, filter indexer layers from attention layer lists using `draft_attn_layer_names - draft_indexer_layer_names`, ensuring accurate metadata selection for DeepSeek MTP heads.

## Test Plan
For the scenarios described in Issue #26711 (DeepSeek V3.2 MTP metadata anomalies and CUDA graph capture crashes in EAGLE mode), run tests with and without the fix applied to verify if the issues recur.

## Test Result
Verification for Issue #26711 has been completed. After the fix: DeepSeek V3.2 MTP metadata mapping works normally, and there are no crashes in CUDA graph capture under EAGLE mode. The original issues have been resolved.

### DIFF
--- a/vllm/v1/spec_decode/eagle.py
+++ b/vllm/v1/spec_decode/eagle.py
@@ -109,6 +109,7 @@ class EagleProposer:
             else []
         )
 
+        self.use_cuda_graph = self.use_cuda_graph and bool(self.cudagraph_batch_sizes)
         # persistent buffers for cuda graph
         self.input_ids = torch.zeros(
             self.max_num_tokens, dtype=torch.int32, device=device
@@ -939,7 +940,7 @@ class EagleProposer:
             self.vllm_config, DeepseekV32IndexerCache
         )
         draft_indexer_layer_names = indexer_layers.keys() - target_indexer_layer_names
-        self.attn_layer_names = list(draft_attn_layer_names)
+        self.attn_layer_names = list(draft_attn_layer_names - draft_indexer_layer_names)
         self.indexer_layer_names = list(draft_indexer_layer_names)
 
         if self.indexer_layer_names:
@@ -1051,9 +1052,7 @@ class EagleProposer:
         use_cudagraphs=True,
     ) -> None:
         # Determine if CUDA graphs should be used for this run.
-        cudagraphs_enabled = (
-            use_cudagraphs and self.use_cuda_graph and bool(self.cudagraph_batch_sizes)
-        )
+        cudagraphs_enabled = use_cudagraphs and self.use_cuda_graph
         if cudagraphs_enabled and num_tokens <= self.cudagraph_batch_sizes[-1]:
             num_tokens = self.vllm_config.pad_for_cudagraph(num_tokens)
 
--- a/vllm/v1/spec_decode/eagle.py
+++ b/vllm/v1/spec_decode/eagle.py
@@ -1050,16 +1050,22 @@ class EagleProposer:
         num_tokens: int,
         use_cudagraphs=True,
     ) -> None:
-        if use_cudagraphs and num_tokens <= self.cudagraph_batch_sizes[-1]:
+        if (
+            use_cudagraphs
+            and self.use_cuda_graph
+            and num_tokens <= self.cudagraph_batch_sizes[-1]
+        ):
             num_tokens = self.vllm_config.pad_for_cudagraph(num_tokens)
 
         with set_forward_context(
             None,
             self.vllm_config,
             num_tokens=num_tokens,
-            cudagraph_runtime_mode=CUDAGraphMode.PIECEWISE
-            if use_cudagraphs
-            else CUDAGraphMode.NONE,
+            cudagraph_runtime_mode=(
+                CUDAGraphMode.PIECEWISE
+                if (use_cudagraphs and self.use_cuda_graph)
+                else CUDAGraphMode.NONE
+            ),
         ):
             if self.supports_mm_inputs:
                 input_ids = None
--- a/vllm/v1/spec_decode/eagle.py
+++ b/vllm/v1/spec_decode/eagle.py
@@ -1050,11 +1050,11 @@ class EagleProposer:
         num_tokens: int,
         use_cudagraphs=True,
     ) -> None:
-        if (
-            use_cudagraphs
-            and self.use_cuda_graph
-            and num_tokens <= self.cudagraph_batch_sizes[-1]
-        ):
+        # Determine if CUDA graphs should be used for this run.
+        cudagraphs_enabled = (
+            use_cudagraphs and self.use_cuda_graph and bool(self.cudagraph_batch_sizes)
+        )
+        if cudagraphs_enabled and num_tokens <= self.cudagraph_batch_sizes[-1]:
             num_tokens = self.vllm_config.pad_for_cudagraph(num_tokens)
 
         with set_forward_context(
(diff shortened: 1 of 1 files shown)


vllm/v1/spec_decode/eagle.py
```diff
@@ -1049,16 +1061,20 @@ def dummy_run(
         num_tokens: int,
         use_cudagraphs=True,
     ) -> None:
-        if use_cudagraphs and num_tokens <= self.cudagraph_batch_sizes[-1]:
+        # Determine if CUDA graphs should be used for this run.
+        cudagraphs_enabled = (
```
benchislett: I think it would be easier to just set `self.use_cuda_graph = self.use_cuda_graph and bool(self.cudagraph_batch_sizes)` in `__init__` since they are both unchanging over the lifetime of the object I think.

benchislett: What is the scenario where `use_cuda_graph` is True and `cudagraph_batch_sizes` is empty? I wonder if this might be a symptom of a deeper issue

xiaohajiayou: Thanks for the suggestion! I applied it and finalized the flag in `__init__`, and all runtime gating now only checks this flag.

xiaohajiayou: - Potential scenarios include: after configuration initialization, cudagraph_mode is overridden to include PIECEWISE mode while the capture size list remains empty—such as when the model's enforce_eager mechanism blocks the size generation process, users explicitly configure an empty list, or sizes are filtered to have no valid entries.

It seems this situation should not occur at present. The addition of this check field is mainly to ensure no out-of-bounds issues arise, while making the drafter's behavior more consistent and safe. @benchislett

xiaohajiayou:
The issue referenced in #26711 is now fixed, with no test issues. Mind reviewing if we can merge this? @benchislett

vllm/v1/spec_decode/eagle.py
benchislett: @xiaohajiayou could you check if https://github.com/vllm-project/vllm/pull/26821 solves the issue, or if this additional change is also necessary?

vllm/v1/spec_decode/eagle.py
```diff
@@ -938,7 +939,7 @@ def load_model(self, target_model: nn.Module) -> None:
             self.vllm_config, DeepseekV32IndexerCache
         )
         draft_indexer_layer_names = indexer_layers.keys() - target_indexer_layer_names
-        self.attn_layer_names = list(draft_attn_layer_names)
+        self.attn_layer_names = list(draft_attn_layer_names - draft_indexer_layer_names)
```
benchislett, line 943: could you help me understand how `attn_layer_names` is used, and why `draft_indexer_layer_names` must be excluded?

xiaohajiayou: Sure! Here’s how the pieces fit together:

- When we build the drafter layers in 
https://github.com/vllm-project/vllm/blob/ab3e80042eac24dd362408e6d63ad98768046359/vllm/v1/spec_decode/eagle.py#L934-L940
we grab every module that inherits `AttentionLayerBase`. DeepSeek’s Lightning Indexer (DeepseekV32IndexerCache) does that too, so its layer names got lumped into `draft_attn_layer_names`.

- Later, _get_attention_metadata_builder looks at the very first name in self.attn_layer_names, finds the backend for that layer, and caches its metadata builder（might be indexer metadata builder） for the whole set.
https://github.com/vllm-project/vllm/blob/ab3e80042eac24dd362408e6d63ad98768046359/vllm/v1/spec_decode/eagle.py#L1078-L1088. 

- After that we loop over self.attn_layer_names and hand that builder’s output to every entry 
https://github.com/vllm-project/vllm/blob/ab3e80042eac24dd362408e6d63ad98768046359/vllm/v1/spec_decode/eagle.py#L256-L262

- The snag is that Lightning Indexer expects DeepseekV32IndexerMetadata, produced by DeepseekV32IndexerMetadataBuilder .
https://github.com/vllm-project/vllm/blob/ab3e80042eac24dd362408e6d63ad98768046359/vllm/v1/attention/backends/mla/indexer.py#L215
the standard attention backends expect completely different metadata. If an indexer sneaks into `self.attn_layer_names`, `_get_attention_metadata_builder` can lock onto the indexer backend, and the loop then feeds indexer metadata to the true attention layers while the indexer never reaches its dedicated path .

- So we subtract `draft_indexer_layer_names` when we finalize `self.attn_layer_names` , like:
```self.attn_layer_names = list(draft_attn_layer_names - draft_indexer_layer_names)```
That guarantees the first entry really is an attention layer.  `_get_attention_metadata_builder` picks the correct attention backend, the drafter attention layers share that metadata as intended, and the indexer layers stay in `self.indexer_layer_names`, where they go through the metadata builder their backend expects.


Hope that clarifies why draft_indexer_layer_names has to be excluded.

vllm/v1/spec_decode/eagle.py
xiaohajiayou: I’d lean toward keeping it. #26821 already handled the “drafter forces eager ⇒ empty cudagraph_capture_sizes” situation. What’s left are the cases where self.use_cuda_graph still flips to True while self.cudagraph_batch_sizes ends up empty—for example:

- someone starts vLLM with `--compilation-config '{"level": 3, "cudagraph_capture_sizes": []}'`;
- the default capture sizes get filtered away by max_num_batched_tokens, sequence parallel, or similar constraints.

In both scenarios, the guard `self.use_cuda_graph &= bool(self.cudagraph_batch_sizes)` automatically disables the drafter’s CUDA graph when the list is empty. The drafter simply falls back to eager, which costs the graph speedup but keeps the model running instead of crashing on an index error. 

That trade-off’s much friendlier than a hard failure.

benchislett (approved):
LGTM, thanks!

PR #22188, opened 2025-08-04
[P/D][Nixl] Introduce `KVTransferMetrics` and aggregation strategy

NickLucche:
This PR provides support for a general `KVTransferMetrics` object while laying the ground for recording stats, but without tracking actual metrics just yet, waiting for NIXL to expose actual telemetry. 
Adding those should be a matter of expanding `record_transfer` and providing `aggregate`+`reduce` operations.

Therefore, this PR is focusing on interfaces and plumbing required for aggregating and reducing KVStats across multiple different kvconnectors (when `MultiKVConnector` is used)  as well as across TP ranks.

Aggregation across connectors happens in  the MultiKVConnector class (newly added `get_kv_transfer_stats` interface) at the worker-level.
Aggregation across ranks happens in the MultiProcExecutor (same as finished_reqs sets) on the main/fe process.
Final reduction happens on logger in main/fe process. 

A few notes on the `KVTransferStats` interface:
 - `.aggregate` is for "fusing" two stats objects into one
 - `.reduce` is for computing representative values (eg avg/median..) from the collected series, for printing/storing.

With this initial enablement PR we're also setting the stage for more complex metric management, including exposing them to Prometheus. For now, we just print them to stdout with the default logger.

Example:
```
(APIServer pid=3026493) INFO 08-05 17:32:58 [loggers.py:126] Engine 000: Avg prompt throughput: 53.2 tokens/s, Avg generation throughput: 15.0 tokens/s, Running: 0 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.1%, Prefix cache hit rate: 0.0%
(APIServer pid=3026493) INFO 08-05 17:32:58 [metrics.py:122] KVConnectorType: KVConnectorType.NIXL, KV Transfer metrics: num_successful_transfers=1
```
### Test with
```
tests/v1/kv_connector/unit/test_nixl_connector.py::test_kv_transfer_stats
tests/v1/kv_connector/unit/test_nixl_connector.py::test_kv_transfer_stats_aggregation
```

Update
---

I've re-worked this PR a bit to address the disalignment between the async READ xfer-related telemetry collection and the scheduler->logger workflow.
Basically it can happen that telemetry from nixl is received on a scheduler step where no ECOs are to be sent to the AsyncLLM. ~~To address that, we introduce a buffer (which can conveniently use `.aggregate`) to address this lag. 
Buffered stats will then only be forwarded (and reset) when the scheduler is set to produce ECOs (checking `num_scheduled_tokens`).~~
[...]

PR #26811, opened 2025-10-14
[KVConnector] Add metrics to Prometheus-Grafana dashboard

NickLucche:
Follow-up to the work and discussion in this PR https://github.com/vllm-project/vllm/pull/22188.
Here we expose the metrics we're currently tracking for the nixl connector to prometheus.

image

cc @markmc  

Update:
@markmc generalized the PR to work for KVConnectorStats more broadly


vllm/v1/metrics/loggers.py
```diff
@@ -804,6 +804,104 @@ def __init__(
...
+                documentation="Histogram of transfer post time for NIXL KV"
+                " Cache transfers.",
+                buckets=buckets[1:],
+                labelnames=labelnames,
+            )
+            self.nixl_histogram_post_time = make_per_engine(
+                nixl_histogram_post_time, engine_indexes, model_name
+            )
+            # uniform 2kb to 16gb range
+            buckets = [2**10 + i for i in range(1, 24, 2)]
```
markmc: Phancy! ITYM

```
# uniform 2kib to 8gib range
buckets = [2**(10 + i) for i in range(1, 24, 2)]
```

Checking it out:

```
>>> def human_size(bytes, units=[' bytes','KB','MB','GB','TB', 'PB', 'EB']):
...     """ Returns a human readable string representation of bytes """
...     return str(bytes) + units[0] if bytes < 1024 else human_size(bytes>>10, units[1:])
... 
>>> [human_size(2**(10+i)) for i in range(1, 25, 2)]
['2KB', '8KB', '32KB', '128KB', '512KB', '2MB', '8MB', '32MB', '128MB', '512MB', '2GB', '8GB']
```

vllm/v1/metrics/loggers.py
```diff
@@ -804,6 +804,104 @@ def __init__(
...
+                labelnames=labelnames,
+            )
+            self.nixl_histogram_xfer_time = make_per_engine(
+                nixl_histogram_xfer_time, engine_indexes, model_name
+            )
+            nixl_histogram_post_time = self._histogram_cls(
+                name="vllm:nixl_post_time_seconds",
+                documentation="Histogram of transfer post time for NIXL KV"
+                " Cache transfers.",
+                buckets=buckets[1:],
```
markmc: This suggests we expect transfer times to have a smaller minimum than post times? I had understood the opposite?

markmc (changes requested):
Hey @NickLucche 

I wanted to make sure the per-connector metrics was going to work out. I didn't feel good about adding a bunch of NIXL-specific metrics into `vllm/v1/metrics`. So I worked it out at NickLucche/vllm/pull/4

Couple of small inline comments on buckets too :+1:

markmc (approved):

simon-mo (approved):
@markmc reviewed

Issue #19329, opened 2025-06-08
[RFC]: Graceful Error Handling for KV Connector Load Failures


sdavidbd:
### 🧩 Summary  
This RFC proposes adding fault recovery support to the KV connector infrastructure introduced in #15960. The goal is to make vLLM resilient to KV block load failures (e.g., due to eviction or network disconnections) by detecting failures and rescheduling only affected requests for recomputation.

### 🧠 Background  
vLLM recently introduced the KV connector abstraction, providing a pluggable interface for integrating external solutions for KV cache offload and transfer. This enables supporting use cases like KV cache sharing and prefill-decode (P-D) disaggregation.
Notable examples include the LMCache connector (PR #16625) and NIXL connector (PR #17751).

The KV loading protocol involves:
- start_load_kv() → invoked before the forward pass to initiate (asynchronous) loading of all required KV blocks.
- wait_for_layer_load() → invoked before each layer’s attention kernel to ensure the relevant blocks are available.

### ⚠️ Problem  
Currently, vLLM lacks a mechanism to detect or recover from KV load failures. If one or more KV blocks fail to load, it may lead to system instability (crash or hang) or incorrect outputs (silent corruption of the KV cache). This undermines reliability, especially in large-scale or disaggregated deployments.

---

### Proposed Change.

Even if KV load failures occur, the forward pass should complete. Post-execution, vLLM will:

- **Detect failed KV blocks**, and
- **Reschedule only the affected requests** for recomputation, using the longest valid prefix.

#### 💡 Key Considerations:
- Preserve progress for valid blocks in affected requests and for unaffected requests in the same batch.
- Prevent deadlocks in tensor-parallel setups (e.g., all-reduce ops).
- In some connector implementations attention kernels may have already been launched and cannot be "rolled back."

### 🛠️ Suggested Implementation
1. KV Connector API Extension
    - Add `get_block_ids_with_load_errors()` to `KVConnectorBase_V1` to allow connectors to report failed block IDs.

2. Executor Reporting
    - `GPUModelRunner` retrieves the failed block IDs and includes them in `ModelRunnerOutput` (should be aggregated across all workers in TP setups).

3. Scheduler Recovery Logic

    - Detects affected requests that used failed blocks

    - Truncates the number of computed blocks for affected requests to their longest valid prefix

    - Reschedules affected requests for recomputation

---

### 🔮 Future Enhancements
- Skip attention computations that depend on failed KV blocks (early detection)
[...]

PR #26813, opened 2025-10-14
[P/D] KV Load Failure Recovery/Abort Configuration

wseaton:
## Purpose

In some situations an operator may not want to allow KV load failure recovery to result in a local prefill on a decode node at all costs. This provides plumbing to make KV load failures bubble up to the api server as a 500  that can be properly handled (either at the proxy layer in a P/D setup, or by clients).

We introduce a new `FINISHED_ERROR` `RequestStatus` that the API server process can check for to throw the correct semantic error.

## Test Plan

Added unit tests, also manually spun up a 1P/1D H100 deployment using the NixlConnector and injected faults in UCX. PR behaves as expected.


vllm/v1/core/sched/scheduler.py
```diff
@@ -1487,29 +1510,78 @@ def _handle_invalid_blocks(self, invalid_block_ids: set[int]) -> set[str]:
...
 
         total_requests_to_reschedule += len(sync_affected_req_ids)
         total_tokens_to_reschedule += num_tokens_to_reschedule
 
-        if total_requests_to_reschedule:
+        if not total_requests_to_reschedule:
+            return set()
+
+        # --- Take action based on policy ---
+        if abort:
```
wseaton: Undecided on extracting this branch into a helper method

vllm/entrypoints/openai/serving_chat.py
```diff
@@ -1074,6 +1074,14 @@ async def chat_completion_stream_generator(
 
                     # if the model is finished generating
                     else:
+                        # check for error finish reason and abort streaming
+                        if error_data := self._handle_streaming_error_finish_reason(
+                            output.finish_reason, request_id
+                        ):
```
NickLucche: do you want to add the same check in speech_to_text.py (audio endpoints) for completeness?

vllm/v1/request.py
```diff
@@ -247,4 +248,5 @@ def get_finished_reason(status: "RequestStatus") -> FinishReason | None:
     RequestStatus.FINISHED_LENGTH_CAPPED: FinishReason.LENGTH,
     RequestStatus.FINISHED_ABORTED: FinishReason.ABORT,
     RequestStatus.FINISHED_IGNORED: FinishReason.LENGTH,
+    RequestStatus.FINISHED_ERROR: FinishReason.ERROR,
```
NickLucche: nit: doesn't this feel a bit too generic to just end up with a hardocoded BAD_GATEWAY?

vllm/v1/core/sched/scheduler.py
```diff
@@ -1471,11 +1467,38 @@ def _update_requests_with_invalid_blocks(
...
         return affected_req_ids, total_affected_tokens
 
-    def _handle_invalid_blocks(self, invalid_block_ids: set[int]) -> set[str]:
-        total_requests_to_reschedule = 0
-        total_tokens_to_reschedule = 0
+    def _handle_invalid_blocks(
+        self, invalid_block_ids: set[int], outputs: dict[int, list[EngineCoreOutput]]
+    ) -> set[str]:
+        should_abort = (
+            self.vllm_config.kv_transfer_config is not None
```
NickLucche: eg we shouldnt have to check this condition in these methods

vllm/v1/core/sched/scheduler.py
```diff
@@ -1483,29 +1510,78 @@ def _handle_invalid_blocks(self, invalid_block_ids: set[int]) -> set[str]:
...
+                request.status = RequestStatus.FINISHED_ERROR
+                kv_transfer_params = self._free_request(request)
+
+                # create EngineOutput for the aborted request
+                outputs[request.client_index].append(
+                    EngineCoreOutput(
+                        request_id=req_id,
+                        new_token_ids=[],
+                        finish_reason=request.get_finished_reason(),
+                        kv_transfer_params=kv_transfer_params,
```
markmc: Why would we return transfer params for an aborted request?

vllm/v1/core/sched/scheduler.py
```diff
@@ -1483,29 +1510,78 @@ def _handle_invalid_blocks(self, invalid_block_ids: set[int]) -> set[str]:
...
+
+                # track which queue the request was in
+                if request.status == RequestStatus.RUNNING:
+                    stopped_running_reqs.add(request)
+                else:
+                    stopped_waiting_reqs.append(request)
+
+                # abort and free the request
+                request.status = RequestStatus.FINISHED_ERROR
+                kv_transfer_params = self._free_request(request)
```
markmc: Any reason not to use `finish_requests()` here? At a glance, it would replace much of this logic?

vllm/v1/core/sched/scheduler.py
```diff
@@ -1483,29 +1510,78 @@ def _handle_invalid_blocks(self, invalid_block_ids: set[int]) -> set[str]:
             )
         )
 
-        total_requests_to_reschedule += len(async_affected_req_ids)
-        total_tokens_to_reschedule += num_tokens_to_reschedule
-
-        # Mark requests with async KV load failures; they will be rescheduled
-        # once loading completes.
-        self.failed_recving_kv_req_ids |= async_affected_req_ids
+        total_requests_to_reschedule = len(async_affected_req_ids)
```
markmc: "requests to reschedule" no longer seems appropriate naming

vllm/v1/core/sched/scheduler.py
```diff
@@ -1483,29 +1510,78 @@ def _handle_invalid_blocks(self, invalid_block_ids: set[int]) -> set[str]:
...
+                else:
+                    stopped_waiting_reqs.append(request)
+
+                # abort and free the request
+                request.status = RequestStatus.FINISHED_ERROR
+                kv_transfer_params = self._free_request(request)
+
+                # create EngineOutput for the aborted request
+                outputs[request.client_index].append(
+                    EngineCoreOutput(
```
markmc: AFAICS `EngineCoreOutput` instances are only every created in `update_from_output()` at the moment - I think it would be nicer if we could maintain that ... just return the aborted request IDs from this function?

vllm/config/kv_transfer.py
```diff
@@ -64,6 +64,11 @@ class KVTransferConfig:
     enable_permute_local_kv: bool = False
     """Experiment feature flag to enable HND to NHD KV Transfer"""
 
+    kv_load_retry_policy: Literal["recompute", "abort"] = "recompute"
+    """Policy for handling KV cache load failures.
+    'recompute': reschedule the request to recompute failed blocks (default)
+    'abort': immediately abort the request with an error finish reason"""
```
markmc: AFAIU #24520 mentions a similar need for policy in the preemption case?

vllm/entrypoints/openai/serving_engine.py
```diff
@@ -732,6 +732,35 @@ def create_streaming_error_response(
...
         return json_str
 
+    def _handle_error_finish_reason(
+        self, outputs: list[CompletionOutput], request_id: str
+    ) -> ErrorResponse | None:
+        """handle error finish reason by logging and returning 502 if found"""
+        for output in outputs:
+            if output.finish_reason == "error":
+                logger.error(
+                    "KV cache load failure detected for request %s",
```
markmc: I wouldn't be so specific here about the cause of the error - can "KV cache load failure" be bubbled up from the lower level and used here?

tests/entrypoints/openai/test_completion_502_error.py
```diff
@@ -0,0 +1,195 @@
...
+        trace_headers,
+        priority,
+    ):
+        return dict(engine_prompt), {}
+
+    serving_completion._process_inputs = AsyncMock(side_effect=_fake_process_inputs)
+    return serving_completion
+
+
+@pytest.mark.parametrize("stream", [False, True])
```
kfirwolfson: consider 2 separate test functions as most of the current function code is branched according to boolean value

vllm/config/kv_transfer.py
kfirwolfson: > AFAIU https://github.com/vllm-project/vllm/pull/24520 mentions a similar need for policy in the preemption case

More or less. In #24520 we added a field that gives the calling entity (e.g. router) control over how much recompute is allowed.

vllm/v1/core/sched/scheduler.py
wseaton: Good point! Let's remove

vllm/entrypoints/openai/serving_engine.py
wseaton: It can be, it would require making a new specific finish reason (or propogating some type of error context object in the engineoutput to switch on). Do you have a preference on approach? The error context approach I tried previously, there are just not a lot of errors that require custom error handling today at the API server level; I still quite like the idea as a "future infra" investment.

vllm/config/kv_transfer.py
wseaton: Is it correct to think of `cache-hit-threshold` basically an intermediate option between these two extremes? The impetus for landing this is that nixl_connector now defaults to "recompute" in all cases, and we need that tunable, and more importantly following correct client semantics (eg. not returning empty output)

vllm/v1/core/sched/scheduler.py
wseaton: renamed also to "affected" since I think it matches better

vllm/config/kv_transfer.py
kfirwolfson: I guess so. As you suggested offline, the enum can be changed to have a third option of kv_cache threshold. Like I mentioned in another comment, if loading succeeded for the first 95% of the tokens, you may prefer "recompute" rather than "abort" behavior.

wseaton:
@kfirwolfson sounds good on abort language. I'm also fine swapping to 503 on error code. Did you have any thoughts on the error context ideas? I really don't like the trend of needing custom finish/stop reasons for individual internal error cases. It seems like we should just have 1 type for each semantic (eg. request rejected because of internal error).

kfirwolfson:
Looking again through the code, @wseaton, I see why the new `finish_reason` is required - it's a way of propagating the error back to the engine. It just won't return to the user, and the caller (router) will still need some way of discerning how to handle the failure. See previous comment regarding `error.message`.

sdavidbd:
@wseaton Thanks for contributing this! I had noted the idea of making the recovery behavior configurable (e.g., allowing request failure instead of recompute for disaggregated P-D setups) in #19329, so it’s really nice to see it implemented.

vllm/v1/core/sched/scheduler.py
```diff
@@ -1055,6 +1051,23 @@ def update_from_output(
...
+                        request_id=req_id,
+                        new_token_ids=[],
+                        finish_reason=request.get_finished_reason(),
+                        kv_transfer_params=None,
+                        trace_headers=request.trace_headers,
+                        num_cached_tokens=request.num_cached_tokens,
+                    )
+                )
+                self._free_request(request)
```
sdavidbd: I see two potential issues with this logic:

**1. Abort case:**
If the connector performs async loading, it may still be transferring KV data into some of the request blocks we’re freeing here. Since async loading can be progressive, freeing these blocks prematurely could lead to a race condition if they’re reallocated before the transfer completes (and the request is marked finished).

**2. Recompute case:**
Here, `failed_kv_load_req_ids` includes sync-loading requests with failed blocks that we intend to recompute. These requests are still running and should not be freed.

wseaton: Thanks for the review! I'll work on a test scenario for these two cases.

wseaton: 1) I see in nixl connector we implement `delay_free_blocks`, which the scheduler already properly handles. Added a test for this case
2) We can simply check if the `request.is_finished()`, added a test case for this, and this check I think solves the issue cleanly

vllm/config/kv_transfer.py
```diff
@@ -64,6 +64,11 @@ class KVTransferConfig:
     enable_permute_local_kv: bool = False
     """Experiment feature flag to enable HND to NHD KV Transfer"""
 
+    kv_load_retry_policy: Literal["recompute", "abort"] = "recompute"
```
njhill: IMO we should reserve the "abort" term to mean client-aborted if possible. So could this be `"fail"` or `"fail_request"` instead?

vllm/v1/core/sched/scheduler.py
```diff
@@ -942,7 +942,7 @@ def update_from_output(
         for req_id, num_tokens_scheduled in num_scheduled_tokens.items():
             assert num_tokens_scheduled > 0
             if failed_kv_load_req_ids and req_id in failed_kv_load_req_ids:
-                # Skip requests that were recovered from KV load failure
+                # skip aborted or rescheduled requests from KV load failure
```
njhill: if we use other terminology
```suggestion
                # skip failed or rescheduled requests from KV load failure
```

vllm/v1/core/sched/scheduler.py
```diff
@@ -1468,44 +1497,68 @@ def _update_requests_with_invalid_blocks(
...
+        )
+
+        # Handle async KV loads (WAITING_FOR_REMOTE_KVS)
         async_load_reqs = (
             req
             for req in self.waiting
             if req.status == RequestStatus.WAITING_FOR_REMOTE_KVS
         )
-        async_affected_req_ids, num_tokens_to_reschedule = (
+        async_affected_req_ids, num_affected_tokens = (
```
njhill: I think we should name it something like `num_failed_tokens`. "affected" is a bit ambiguous

vllm/v1/core/sched/scheduler.py
```diff
@@ -1468,44 +1497,68 @@ def _update_requests_with_invalid_blocks(
...
+            # mark requests as finished and remove from queues without freeing
+            # we don't free yet as outputs will be created in update_from_output
+            self.finish_requests(
+                all_affected_req_ids,
+                RequestStatus.FINISHED_ERROR,
+                free_immediately=False,
+            )
+
+            return all_affected_req_ids
+        else:
```
njhill: the `else` is redundant here

vllm/v1/engine/__init__.py
```diff
@@ -18,24 +18,27 @@
...
 class FinishReason(enum.IntEnum):
     """
-    Reason a request finished - stop, length, or abort.
+    Reason a request finished - stop, length, abort, or error.
 
     Int rather than Str for more compact serialization.
 
     stop - a stop string was emitted
     length - max_tokens was consumed, or max_model_len was reached
     abort - aborted for another reason
```
njhill: I know preexisting but we could update this description too
```suggestion
    abort - aborted by client
```

vllm/v1/core/sched/scheduler.py
```diff
@@ -1051,6 +1051,26 @@ def update_from_output(
             # This is a rare case and unlikely to impact performance.
             self.waiting.remove_requests(stopped_preempted_reqs)
 
+        if failed_kv_load_req_ids:
```
njhill: I'm not keen though on how the finish/free gets decoupled (i.e the `free_immediately` arg). I have been thinking it through and wonder if this would be better:

Instead of calling `self.finish_requests` in `_handle_invalid_blocks`, just call it here if `failed_kv_load_req_ids` is non-empty and if the policy is to "fail" ("abort").

We can change `finish_requests` to return `valid_requests`, and here that can then be looped over to add the `EngineCoreOutput`s.

vllm/v1/core/sched/scheduler.py
```diff
@@ -1051,6 +1051,26 @@ def update_from_output(
...
+                # recompute case: requests remain RUNNING and will be rescheduled.
+                if request is None or not request.is_finished():
+                    continue
+
+                outputs[request.client_index].append(
+                    EngineCoreOutput(
+                        request_id=req_id,
+                        new_token_ids=[],
+                        finish_reason=request.get_finished_reason(),
+                        kv_transfer_params=None,
```
njhill: nit: can just omit this one since it has a default
```suggestion
```

njhill (commented):
Thanks for this @wseaton it looks pretty good to me!

vllm/v1/core/sched/scheduler.py
wseaton: I like this, way cleaner!

vllm/entrypoints/openai/serving_completion.py
```diff
@@ -280,6 +281,20 @@ async def create_completion(
...
             final_res_batch_checked = cast(list[RequestOutput], final_res_batch)
 
+            for final_res in final_res_batch_checked:
+                for output in final_res.outputs:
+                    if output.finish_reason == "error":
+                        logger.error(
+                            "Request-level error for request %s: %s",
+                            request_id,
+                            output.stop_reason or "unknown",
+                        )
```
njhill: Might be better to omit the logging here given we already log where the error occurs?

vllm/entrypoints/openai/serving_completion.py
```diff
@@ -428,6 +443,19 @@ async def completion_stream_generator(
                     finish_reason = output.finish_reason
                     stop_reason = output.stop_reason
 
+                    if finish_reason == "error":
+                        logger.error(
+                            "Request-level error for request %s: %s",
+                            request_id,
+                            stop_reason or "unknown",
+                        )
```
njhill: Same comment as above

vllm/v1/core/sched/scheduler.py
```diff
@@ -1468,44 +1503,59 @@ def _update_requests_with_invalid_blocks(
...
+        if should_fail:
             logger.warning(
-                "Recovered from KV load failure: "
-                "%d request(s) rescheduled (%d tokens affected).",
-                total_requests_to_reschedule,
-                total_tokens_to_reschedule,
+                "Failing %d request(s) due to KV load failure "
+                "(retry_policy=fail, %d tokens affected).",
+                total_failed_requests,
+                total_failed_tokens,
```
njhill, line 1790: Could change this one to error since we're failing the requests. Could possibly also log the set of request ids if we don't log them individually in the front-end.

vllm/v1/core/sched/scheduler.py
```diff
@@ -1171,11 +1195,20 @@ def finish_requests(
         self,
         request_ids: str | Iterable[str],
         finished_status: RequestStatus,
-    ) -> None:
+    ) -> list[Request]:
```
njhill: I guess the return type needs to be changed in `SchedulerInterface` too

vllm/v1/core/sched/scheduler.py
```diff
@@ -1051,6 +1051,30 @@ def update_from_output(
             # This is a rare case and unlikely to impact performance.
             self.waiting.remove_requests(stopped_preempted_reqs)
 
+        if failed_kv_load_req_ids:
+            should_fail = (
+                self.vllm_config.kv_transfer_config is not None
+                and self.vllm_config.kv_transfer_config.kv_load_retry_policy == "fail"
+            )
```
njhill: Could have a bool field in the constructor like `self.recompute_kv_load_failures` set to this.

Then the check here would be cleaner, like
```python
if failed_kv_load_req_ids and not self.recompute_kv_load_failures:
    # ...
```

vllm/config/kv_transfer.py
```diff
@@ -64,6 +64,11 @@ class KVTransferConfig:
     enable_permute_local_kv: bool = False
     """Experiment feature flag to enable HND to NHD KV Transfer"""
 
+    kv_load_retry_policy: Literal["recompute", "fail"] = "recompute"
```
njhill: Maybe change the name since neither of these involve retrying?
```suggestion
    kv_load_failure_policy: Literal["recompute", "fail"] = "recompute"
```

I guess a separate question is whether we should also retry to get the blocks from the connector. It could make sense to do that (or not do it) in conjunction with either of these.

I'm also wondering (just thinking aloud) whether it would make sense for this to be controlled instead via a "strict decode-only mode" parameter.

vllm/entrypoints/openai/serving_chat.py
```diff
@@ -1074,6 +1075,21 @@ async def chat_completion_stream_generator(
...
                     # if the model is finished generating
                     else:
+                        # check for error finish reason and abort streaming
+                        # finish_reason='error' indicates a retryable error
+                        if output.finish_reason == "error":
+                            logger.error(
+                                "Request-level error for request %s: %s",
+                                request_id,
+                                output.stop_reason or "unknown",
+                            )
```
njhill: same logging comment

vllm/entrypoints/openai/serving_chat.py
```diff
@@ -1268,6 +1284,21 @@ async def chat_completion_full_generator(
...
 
+        # Check for error finish reason and return 503 error
+        # finish_reason='error' indicates a retryable request-level internal error
+        for output in final_res.outputs:
+            if output.finish_reason == "error":
+                logger.error(
+                    "Request-level error for request %s: %s",
+                    request_id,
+                    output.stop_reason or "unknown",
+                )
```
njhill: same logging comment

vllm/entrypoints/openai/serving_completion.py
```diff
@@ -280,6 +281,20 @@ async def create_completion(
...
+            for final_res in final_res_batch_checked:
+                for output in final_res.outputs:
+                    if output.finish_reason == "error":
+                        logger.error(
+                            "Request-level error for request %s: %s",
+                            request_id,
+                            output.stop_reason or "unknown",
+                        )
+                        return self.create_error_response(
+                            "Service temporarily unavailable",
```
njhill: What about like "Completion request failed" for the message (same in all places)?

vllm/entrypoints/openai/serving_completion.py
```diff
@@ -428,6 +443,19 @@ async def completion_stream_generator(
...
 
+                    if finish_reason == "error":
+                        logger.error(
+                            "Request-level error for request %s: %s",
+                            request_id,
+                            stop_reason or "unknown",
+                        )
+                        error_data = self.create_streaming_error_response(
+                            "Service temporarily unavailable"
+                        )
```
njhill: I think we should pass the err_type and status_code here too or else they will default to BadRequest which is incorrect (same for chat completion).

njhill (commented):
Thanks @wseaton I like your updates, also have a few more comments

vllm/v1/core/sched/scheduler.py
```diff
@@ -1055,6 +1051,30 @@ def update_from_output(
...
+
+            if should_fail:
+                # finish requests (which frees them) and create outputs
+                finished_reqs = self.finish_requests(
```
sdavidbd: After discussing offline with @kfirwolfson, we realized that in the **sync-loading** case, KV blocks are inserted into the cache **before** the load completes. As a result, failing requests on load failure leaves these **invalid blocks** in the cache, which risks them being **reused by future requests**. (All currently active requests sharing the invalid blocks will also fail.)

A straightforward fix could be to introduce a flag in `finish_requests()` to **evict all blocks** associated with the failed requests. However, this would also remove **valid blocks** — those successfully loaded or already cached locally — and we need to ensure **we don’t evict valid blocks shared by active requests** that didn’t fail.

A safer alternative would be to **evict only the invalid blocks** (the ones that failed to load and any subsequent ones within the same failed requests) in __update_requests_with_invalid_blocks()_.

vllm/entrypoints/openai/serving_chat.py
```diff
@@ -1268,6 +1284,21 @@ async def chat_completion_full_generator(
 
         assert final_res is not None
 
+        # Check for error finish reason and return 503 error
+        # finish_reason='error' indicates a retryable request-level internal error
+        for output in final_res.outputs:
+            if output.finish_reason == "error":
```
chaunceyjiang: Could this be defined as a function?  
That way, it could be reused—for example, as in https://github.com/vllm-project/vllm/pull/27064.

vllm/v1/engine/__init__.py
```diff
@@ -18,24 +18,27 @@
...
+    Reason a request finished - stop, length, abort, or error.
 
     Int rather than Str for more compact serialization.
 
     stop - a stop string was emitted
     length - max_tokens was consumed, or max_model_len was reached
-    abort - aborted for another reason
+    abort - aborted by client
+    error - retryable request-level internal error (e.g., KV load failure).
+            Invariant: always converted to 503 Service Unavailable.
```
markmc: Just digging into 500 vs 503 ...

500 Internal Server Error is ... an unexpected error, worthy of a bug report. With an idempotent inferencing requesting, it's perfectly sensible for a client to retry, but if you repeatedly get 500s you'd assume it's not going to complete successfully until the server is fixed.

503 implies "we're at some capacity limit or in maintenance, back off and try again later". I think proxies can even choose to cache this response (maybe only if Retry-After header included) to reduce load on the server. I can see this being appropriate if the common case for a request error-ing is a maintenance scenario - "we had to cut your request short because of a maintenance operation, please try again".

I'd tend to say we should default to 500, and only use 503 when we actually know it's a -in-maintenance/over-capacity error.

vllm/entrypoints/openai/serving_chat.py
```diff
@@ -6,6 +6,7 @@
 import time
 from collections.abc import AsyncGenerator, AsyncIterator
 from collections.abc import Sequence as GenericSequence
+from http import HTTPStatus
```
markmc: Need to check carefully where else `RequestOutput` is handled by the frontend, and add support for this new finish reason

(Interestingly, for "pooling requests" we just discard the finish reason - see output processor `make_request_output()`)

Other places to look at `RequestOutput` handling:

* `serving_engine.py` - see `beam_search()`
* `serving_responses.py` - see `responses_full_generator()`

vllm/entrypoints/openai/serving_chat.py
```diff
@@ -1074,6 +1075,21 @@ async def chat_completion_stream_generator(
...
+                        # check for error finish reason and abort streaming
+                        # finish_reason='error' indicates a retryable error
+                        if output.finish_reason == "error":
+                            logger.error(
+                                "Request-level error for request %s: %s",
+                                request_id,
+                                output.stop_reason or "unknown",
+                            )
+                            error_data = self.create_streaming_error_response(
+                                "Service temporarily unavailable"
```
markmc: This defaults to `400 BAD_REQUEST`

vllm/v1/core/sched/scheduler.py
```diff
@@ -1206,11 +1239,13 @@ def finish_requests(
         if waiting_requests_to_remove:
             self.waiting.remove_requests(waiting_requests_to_remove)
 
-        # Second pass: set status and free requests
+        # Set status and free all valid requests
         for request in valid_requests:
             request.status = finished_status
             self._free_request(request)
 
+        return valid_requests
```
markmc: I don't think we need to return these - in the case you're using the return value, all requests passes in should be returned ... they should never hit the `not in self.requests or not request.is_finished()` case

vllm/v1/core/sched/scheduler.py
```diff
@@ -1468,44 +1503,59 @@ def _update_requests_with_invalid_blocks(
...
+        """
+        Handle requests affected by invalid KV cache blocks.
 
-        # --- Handle async KV loads (WAITING_FOR_REMOTE_KVS) ---
+        Returns:
+            Set of affected request IDs to skip in update_from_output main loop.
+        """
+        should_fail = (
+            self.vllm_config.kv_transfer_config is not None
+            and self.vllm_config.kv_transfer_config.kv_load_retry_policy == "fail"
```
markmc: return `should_fail` from this function? Just a bit duplicative to have this policy in two places

vllm/v1/core/sched/scheduler.py
wseaton: agree, I don't like deriving the same config in two places :)

vllm/v1/core/sched/scheduler.py
markmc: Oh, I see - we've no way of looking up a request ID after it has been freed

Just a naming thing, but I'd call the return value "freed_requests"

vllm/v1/core/sched/scheduler.py
wseaton: we are going to tack it onto self, see nick's comment above

vllm/config/kv_transfer.py
wseaton: > I guess a separate question is whether we should also retry to get the blocks from the connector. It could make sense to do that (or not do it) in conjunction with either of these.

I agree, I think we'd need a separate config for this that can be overridden, and let the connectors define their own default behavior (which includes retry policy). It seems out of scope for this PR though

vllm/entrypoints/openai/serving_chat.py
wseaton: Fixed as per nick's comment

vllm/config/kv_transfer.py
kfirwolfson: @njhill @wseaton, I think a "strict decode-only mode" may be too harsh in many cases. It blocks the option for a router to decide not to disaggregate a request (e.g. due to high cache hit on a decode node)
Also in case of load failures, if say 95% of the tokens were loaded successfully and only a small tail at the end failed, it may make sense to perform some prefill on the decode node (similar to the decision not to disaggregate). This is one of the use-cases for the cache_hit_threshold feature #24520

vllm/v1/engine/__init__.py
kfirwolfson: Good points. Makes sense to use 500 then.

I found 3 places in the repo that already return 500 today, all of which not relevant, so I don't see a backward compatibility problem in returning 500.

details:
Two of the 500 return codes are in disagg_proxy_demo.py which is an example script (and AFAIU above vLLM anyway).
Only other place is in elastic expert parallelism flow, which is not relevant for the current discussion: https://github.com/vllm-project/vllm/blob/bd66b8529bb00821bf14565e8b7663cbdc286d91/vllm/entrypoints/openai/api_server.py#L1121

vllm/v1/core/sched/scheduler.py
wseaton: @sdavidbd I think I have a decent fix for this, as well as a test case that isolates (and reproduces) the problem. Ended up going with Option B): 

> evict only the invalid blocks (the ones that failed to load and any subsequent ones within the same failed requests) in __update_requests_with_invalid_blocks()_.

We do this now regardless of retry policy which I think is correct, there's a race condition in the recompute case, and a cache pollution issue in the failure case.

vllm/v1/engine/__init__.py
wseaton: as we have consensus I went ahead and made this change

vllm/v1/core/sched/scheduler.py
```diff
@@ -1468,44 +1502,66 @@ def _update_requests_with_invalid_blocks(
...
+        total_failed_requests += len(sync_failed_req_ids)
+        total_failed_tokens += num_failed_tokens
+
+        if not total_failed_requests:
+            return set()
+
+        # evict invalid blocks from the prefix cache to prevent future
+        # requests from reusing corrupted data. this must happen regardless
+        # of retry policy to prevent cache pollution race conditions.
+        self.kv_cache_manager.evict_blocks(invalid_block_ids)
```
njhill: Async loading blocks won't be cached at this pint, does that matter?

njhill (commented):
Would be good for @heheda12345 and @sdavidbd to also check the kv cache manager / eviction changes.

Looks like there are some precommit errors too.

vllm/v1/core/sched/scheduler.py
wseaton: I belive it to be a noop and safe, but we should fix it to only do this manual eviction on the sync path

wseaton: found a cheap fix for this in f479682a11612d79f20314339318e609e5ede698, before eviction we can simply check if the blocks have hashes in the block table, if they don't they can be ignored.

vllm/entrypoints/openai/serving_chat.py
```diff
@@ -1077,6 +1075,18 @@ async def chat_completion_stream_generator(
 
                     # if the model is finished generating
                     else:
+                        # check for error finish reason and abort streaming
+                        # finish_reason='error' indicates a retryable error
+                        if output.finish_reason == "error":
```
chaunceyjiang: Could this be defined as a function?  
That way, it could be reused—for example, as in
https://github.com/vllm-project/vllm/pull/27064/files#diff-190c665c438d34a7190da9a4d9bc1ed24bed8b13ee1b3f20c6da5c8aa52b0f3bR739-R742

wseaton: Yes, let's adapt your interface, it'll make merging a lot easier since our PRs do similar things

vllm/v1/core/sched/scheduler.py
```diff
@@ -1465,44 +1502,70 @@ def _update_requests_with_invalid_blocks(
...
-        if total_requests_to_reschedule:
-            logger.warning(
-                "Recovered from KV load failure: "
-                "%d request(s) rescheduled (%d tokens affected).",
-                total_requests_to_reschedule,
-                total_tokens_to_reschedule,
+        if not total_failed_requests:
+            return set()
+
+        # invalidate cached blocks by eviction to prevent poisoning
```
sdavidbd: It may not be correct to evict the blocks unconditionally. When the retry policy is **recompute**, any request sharing those blocks will only access them **after** they’re recomputed. In `_update_requests_with_invalid_blocks()`, only the **first request** (by priority) sharing invalid blocks will trigger recomputation, while the others will treat those blocks as already computed (though they will still retry computation of downstream blocks depending on them).

vllm/v1/core/sched/scheduler.py
```diff
@@ -1465,44 +1502,70 @@ def _update_requests_with_invalid_blocks(
...
+        # NOTE: this naturally excludes async requests, as they have no
+        # hash in the block table at this point
+        cached_invalid_blocks = self.kv_cache_manager.get_cached_block_ids(
+            invalid_block_ids
+        )
+        if cached_invalid_blocks:
+            self.kv_cache_manager.evict_blocks(cached_invalid_blocks)
```
sdavidbd: `invalid_block_ids` includes only the blocks that failed to load in the connector, but there may also be **subsequent blocks** that become invalid because they were computed based on those failed ones.
I suggest adding a flag to `_update_requests_with_invalid_blocks()` indicating whether invalid request blocks should be evicted, and when iterating over `req_block_ids`, track the first affected block and evict `req_block_ids[first_affected:]`.

tests/v1/kv_connector/unit/test_cache_pollution_prevention.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+
+    recompute_scheduler.update_from_output(scheduler_output, model_runner_output)
+
+    # verify request still running (recompute policy)
+    assert request1.status == RequestStatus.RUNNING
+
+    # critical assertion: invalid block should be evicted from cache
+    # the block's hash should be None after eviction
+    assert block.block_hash is None, (
```
sdavidbd: Perhaps add an assert that the invalid block hash is not `None` before the call to `update_from_output`.

tests/v1/kv_connector/unit/test_cache_pollution_prevention.py
```diff
@@ -0,0 +1,133 @@
...
+        f"invalid block {invalid_block_id} should have been evicted "
+        f"(hash reset to None), but hash is {block.block_hash}"
+    )
+
+    # verify the block is not in the cached_block_hash_to_block map
+    # try to look it up - should return None
+    cached_blocks = (
+        recompute_scheduler.kv_cache_manager.block_pool.cached_block_hash_to_block
+    )
+    assert len(cached_blocks) == 0 or invalid_block_id not in [
```
sdavidbd: With the eviction logic I suggested, you could strengthen the assertion here: instead of just verifying that the failed block is not in the cache, you could also assert that the cache contains **exactly the valid blocks** (i.e., all blocks before the first affected one). That would make the test cover both correctness and completeness of the eviction behavior.

tests/v1/kv_connector/unit/test_cache_pollution_prevention.py
```diff
@@ -0,0 +1,133 @@
...
+        invalid_block_ids=invalid_block_ids,
+        use_eos=False,
+    )
+
+    recompute_scheduler.update_from_output(scheduler_output, model_runner_output)
+
+    # verify request still running (recompute policy)
+    assert request1.status == RequestStatus.RUNNING
+
+    # critical assertion: invalid block should be evicted from cache
```
sdavidbd: Following my previous comment, this check could be extended to **all subsequent blocks** since those become invalid as well (computed based on the failed block).

vllm/v1/core/sched/scheduler.py
sdavidbd: @wseaton Thanks for the fix, and I really appreciate how you consistently accompany every change with a dedicated test! I left a couple of inline comments in the code with some follow-up thoughts:

- We might not want to evict blocks unconditionally — with a **recompute** policy, shared requests will only access those blocks once they’ve been recomputed.

- Also, `invalid_block_ids` currently only includes the blocks that failed to load, but dependent downstream blocks are also invalid.

I suggested in the comments a possible approach to handle this by adding a flag to `_update_requests_with_invalid_blocks()` and evicting `req_block_ids[first_affected:]` when appropriate.

vllm/v1/core/sched/scheduler.py
wseaton: Good catch, I've implemented your logic for this!

tests/v1/kv_connector/unit/test_cache_pollution_prevention.py
wseaton: Updated all the tests based on your feedback

vllm/v1/core/sched/scheduler.py
sdavidbd: Overall the fix LGTM! Just noting that we’re still evicting blocks unconditionally, which might not be needed in the **recompute** case as I mentioned earlier.

vllm/v1/core/sched/scheduler.py
```diff
@@ -1055,6 +1060,24 @@ def update_from_output(
             # This is a rare case and unlikely to impact performance.
             self.waiting.remove_requests(stopped_preempted_reqs)
 
+        if failed_kv_load_req_ids and not self.recompute_kv_load_failures:
+            # finish requests (which frees them) and create outputs
+            finished_reqs = self.finish_requests(
```
markmc: I'm sceptical of the changes to `finish_requests()`

In what cases should a request at this point fail one of these tests?

```
            if request is None or request.is_finished():
                # Invalid request ID.                                                                                                                                           
                continue
```

I'm pretty sure the answer is "never", in which case I'd much rather we don't have the appearance of this being a valid code path ... it would just be something for future readers to puzzle over

If we felt, for example, there might be some subtle race condition where a request would already be finished at this point, I'd add an assertion to enforce the invariant - if it was ever triggered, we'd learn something from it and probably fix the bug that allowed it to happen

vllm/v1/core/sched/scheduler.py
```diff
@@ -1470,6 +1511,7 @@ def _update_requests_with_invalid_blocks(
                     continue
 
                 marked_invalid_block = True
+                first_affected_idx = idx
```
markmc: Why couldn't the `blocks_to_evict.update()` happen here?

vllm/v1/core/sched/scheduler.py
```diff
@@ -1487,50 +1529,83 @@ def _update_requests_with_invalid_blocks(
...
     def _handle_invalid_blocks(self, invalid_block_ids: set[int]) -> set[str]:
-        total_requests_to_reschedule = 0
-        total_tokens_to_reschedule = 0
+        """
+        Handle requests affected by invalid KV cache blocks.
+
+        Returns:
+            Set of affected request IDs to skip in update_from_output main loop.
+        """
+        should_fail = (
```
markmc: Why not just use `not self.recompute_kv_load_failures` ?

vllm/v1/core/sched/scheduler.py
```diff
@@ -1397,36 +1431,43 @@ def _update_from_kv_xfer_finished(self, kv_connector_output: KVConnectorOutput):
...
+        invalid_block_ids: set[int],
+        evict_blocks: bool = True,
+    ) -> tuple[set[str], int, set[int]]:
         """
         Identify and update requests affected by invalid KV cache blocks.
 
         This method scans the given requests, detects those with invalid blocks
         and adjusts their `num_computed_tokens` to the longest valid prefix.
-        For observability, it also accumulates the total number of tokens that
-        will need to be recomputed across all affected requests.
```
markmc, line 1621: Curious why you've removed this (and another below) mention of observability?

vllm/v1/core/kv_cache_manager.py
```diff
@@ -342,6 +342,33 @@ def free(self, request: Request) -> None:
...
+                cached_block_ids.add(block_id)
+        return cached_block_ids
+
+    def evict_blocks(self, block_ids: set[int]) -> None:
+        """evict blocks from the prefix cache by their block IDs.
+
+        Args:
+            block_ids: Set of block IDs to evict from cache.
+        """
+        self.coordinator.evict_blocks(block_ids)
```
markmc: It looks like `KVCacheManager` interacts directly with the `BlockPool` in several other cases, and the coordinator doesn't need to be involved ... I suggest removing the `KVCacheCoordinator.evict_blocks()` API

vllm/v1/core/kv_cache_manager.py
```diff
@@ -342,6 +342,33 @@ def free(self, request: Request) -> None:
         """
         self.coordinator.free(request.request_id)
 
+    def get_cached_block_ids(self, block_ids: set[int]) -> set[int]:
+        """filter a set() of block IDs to only those that are cached
```
markmc: I don't think we need this method since `_maybe_evict_cached_block()` is already tolerant of uncached blocks, it seems fine for `evict_blocks()` to be tolerant also

vllm/entrypoints/openai/serving_chat.py
```diff
@@ -1102,6 +1102,16 @@ async def chat_completion_stream_generator(
...
                     else:
+                        # check for error finish reason and abort streaming
+                        # finish_reason='error' indicates a retryable error
+                        error_data = self._handle_streaming_error_finish_reason(
+                            output.finish_reason, request_id
+                        )
+                        if error_data:
+                            yield f"data: {error_data}\n\n"
+                            yield "data: [DONE]\n\n"
+                            return
```
markmc: Why wouldn't we translate the error finish reason into an exception, and re-use the existing exception handler for this:

```
        except Exception as e:
            # TODO: Use a vllm-specific Validation Error                                                                                                                        
            logger.exception("Error in chat completion stream generator.")
            data = self.create_streaming_error_response(str(e))
            yield f"data: {data}\n\n"
```

vllm/entrypoints/openai/serving_chat.py
```diff
@@ -1296,6 +1306,15 @@ async def chat_completion_full_generator(
...
         assert final_res is not None
 
+        # Check for error finish reason and return 500 error
+        # finish_reason='error' indicates a retryable request-level internal error
+        for output in final_res.outputs:
+            error_response = self._handle_error_finish_reason(
+                output.finish_reason, request_id
+            )
+            if error_response:
+                return error_response
```
markmc: I'd be inclined to use the exception handling pattern here too? i.e. we have a few cases just above where we're handling exceptions and turning them into error responses?

markmc: This gets me thinking about the `LLM` API, e.g. `LLM.generate()`

The proposal here is that we add `finish_reason = error` meaning that one of the returned `RequestOutput` could have an error and, by adding this new finish reason, we're possibly requiring users to update their code to handle it?

But when we put this behind an HTTP API, any error causes a `500` return? Not saying which is right or wrong, just observing the asymmetry as worth thinking about

tests/v1/kv_connector/unit/test_invalid_blocks_race_conditions.py
```diff
@@ -0,0 +1,542 @@
...
+    return create_scheduler(vllm_config)
+
+
+def test_async_fail_blocks_not_freed_before_transfer_complete(
+    fail_scheduler: Scheduler,
+):
+    """
+    Test Issue 1: Async fail case - blocks protected by delay_free_blocks.
+
+    When failing a request with async KV loading and invalid blocks:
```
markmc: Something doesn't seem right here ... `delay_free_blocks` is a thing in P to keep the blocks around until D loads the. But async KV loading is a thing on the D side? Why would they interact?

This is about the call to `Scheduler.finish_requests()` if there is a KV load error, but what blocks are we protecting in this case?

(Maybe I'm completely misunderstanding!)

tests/v1/kv_connector/unit/test_invalid_blocks_race_conditions.py
```diff
@@ -0,0 +1,542 @@
...
+    assert len(allocated_blocks[0]) > 0, (
+        "Blocks should be protected from freeing when delay_free_blocks=True"
+    )
+
+
+def test_sync_recompute_blocks_not_freed_for_running_requests(
+    recompute_scheduler: Scheduler,
+):
+    """
+    Test Issue 2: Sync recompute case - blocks incorrectly freed for running requests.
```
markmc: I wouldn't call this a "race condition"? (as in, not sure about the filename choice)

vllm/entrypoints/openai/serving_chat.py
markmc: To be clear:

1. I think I'm strongly in favour of reusing the exception handling here - but this could be an exception raised by the error handling helpers you have added
2. I'd like us to consider whether `LLM.generate()` etc. should raise an exception rather than encoding the error in `RequestOutput` ... but I anticipate good arguments for and against that approach

markmc (changes requested):
This is looking good, sorry for coming in so late with comments. It took me a while to get my head into the block eviction stuff

vllm/v1/core/sched/scheduler.py
wseaton: we can assert here, but I'm a little confused by the suggestion, this is not something I've modified. There's also similar code in `update_draft_token_ids` which may benefit from a similar assertion

vllm/v1/core/sched/scheduler.py
wseaton: it can, I like this suggestion because it removes an else branch

tests/v1/kv_connector/unit/test_invalid_blocks_race_conditions.py
wseaton: s/race condition/correctness seems better

vllm/v1/core/sched/scheduler.py
markmc: What I mean is, we should be able to just do:

```
        if failed_kv_load_req_ids and not self.recompute_kv_load_failures:
            self.finish_requests(
                failed_kv_load_req_ids,
                RequestStatus.FINISHED_ERROR,
            )
            for request in failed_kv_load_req_ids_reqs:
                outputs[request.client_index].append(
                    EngineCoreOutput(
                        request_id=request.request_id,

```

but if we were particularly concerned about those requests that `finish_requests()` might filter out, we could:

```
        if failed_kv_load_req_ids and not self.recompute_kv_load_failures:
            for request in failed_kv_load_req_ids_reqs:
                assert (request.request_id in self.requests and
                        not request.is_finished())
                self.finish_requests(request.request_id, RequestStatus.FINISHED_ERROR)
                outputs[request.client_index].append(
                    EngineCoreOutput(
                        request_id=request.request_id,
```

I'm not convinced the assertion is all that useful, but if it's not useful then it's also not useful to use `finish_requests()` to filter requests in this situation

tests/v1/kv_connector/unit/test_invalid_blocks_race_conditions.py
wseaton: You are right, this is a really bad description for what this test actually does... what it actually tests is probably also not relevant for this PR (the P side contract on blocks), so I think it's worth cutting it.

wseaton: Initially I thought this test was useful; but that was based on a misunderstanding of the purpose of `delay_free_blocks`, so looking into your question helped clear that up. It's probably more useful to test this in integration

vllm/v1/core/sched/scheduler.py
wseaton: incidental deletion, will true up

vllm/v1/core/kv_cache_manager.py
wseaton: perfect, this cleans things up quite a bit

vllm/entrypoints/openai/serving_chat.py
wseaton: For 2) I think I'm in favor of LLM.generate() throwing an exception (probably the same one as you mention), as encoding the error in RequestOutput is not really spec conforming. It would be consistent for us never to return errors in output for the user.

vllm/v1/core/sched/scheduler.py
wseaton: Totally get it now, I think you are right. We should trust the invariant holds and not do a redundant assert. I'll clean this up

vllm/entrypoints/openai/serving_chat.py
wseaton: I have streamlined the error handling to be control flow based in [1467f5c](https://github.com/vllm-project/vllm/pull/26813/commits/1467f5ce96403c473396cd59a0173db6d0350dce) like you suggest, I also had to add quite a few tests for coverage

vllm/v1/core/sched/scheduler.py
```diff
@@ -1175,11 +1194,19 @@ def finish_requests(
...
         For example, the API server can abort a request when the client
         disconnects.
+
+        Args:
+            request_ids: Request IDs to finish.
+            finished_status: The finished status to set.
+
+        Returns:
+            List of freed requests. These are freed from the
+            scheduler PoV, but can be used to create outputs.
```
markmc: No need for the changes to `finish_requests()` anymore, including the interface definition

markmc (changes requested):
Sorry, just the `finish_requests()` thing

vllm/v1/core/sched/scheduler.py
wseaton: covered in 8a1865f

markmc (approved):

wseaton:
Rebasing this to try to get working CI

vllm/v1/core/sched/scheduler.py
```diff
@@ -1108,6 +1113,20 @@ def update_from_output(
...
 
+        if failed_kv_load_req_ids and not self.recompute_kv_load_failures:
+            for req_id in failed_kv_load_req_ids:
+                request = self.requests[req_id]
+                self.finish_requests(req_id, RequestStatus.FINISHED_ERROR)
+                outputs[request.client_index].append(
+                    EngineCoreOutput(
+                        request_id=request.request_id,
+                        new_token_ids=[],
+                        finish_reason=request.get_finished_reason(),
```
markmc, line 1193: Per-request engine core events are relevant, even if the request fails:

```
events=request.take_events(),
```

vllm/entrypoints/openai/serving_completion.py
```diff
@@ -294,6 +298,15 @@ async def create_completion(
 
             final_res_batch_checked = cast(list[RequestOutput], final_res_batch)
 
+            try:
+                for final_res in final_res_batch_checked:
+                    for output in final_res.outputs:
+                        self._handle_error_finish_reason(
+                            output.finish_reason, request_id
```
markmc: Why not do this in `request_output_to_completion_response()` and catch `GenerationError` at the same level as the other exception handlers?

vllm/entrypoints/openai/serving_chat.py
```diff
@@ -1305,6 +1318,14 @@ async def chat_completion_full_generator(
 
         assert final_res is not None
 
+        # Check for error finish reason and return 500 error
+        # finish_reason='error' indicates a retryable request-level internal error
+        try:
+            for output in final_res.outputs:
+                self._handle_error_finish_reason(output.finish_reason, request_id)
+        except GenerationError as e:
+            return self._convert_generation_error_to_response(e)
```
markmc: I'd expect this exception handler to be at the call site of `chat_completion_full_generator()`

vllm/entrypoints/openai/serving_chat.py
```diff
@@ -1305,6 +1318,14 @@ async def chat_completion_full_generator(
 
         assert final_res is not None
 
+        # Check for error finish reason and return 500 error
+        # finish_reason='error' indicates a retryable request-level internal error
+        try:
+            for output in final_res.outputs:
+                self._handle_error_finish_reason(output.finish_reason, request_id)
```
markmc: Why not do this below where we're already iterating this list?

vllm/entrypoints/openai/serving_responses.py
```diff
@@ -581,40 +585,57 @@ async def responses_full_generator(
...
+                if request.enable_response_messages:
+                    raise NotImplementedError(
+                        "enable_response_messages is currently only "
+                        "supported for gpt-oss"
+                    )
+                # Calculate usage.
+                assert final_res.prompt_token_ids is not None
+                num_tool_output_tokens = 0
+        except GenerationError as e:
+            return self._convert_generation_error_to_response(e)
```
markmc: The call sites of `responses_full_generator()` look to be the more natural place to have this exception handling?

tests/entrypoints/openai/test_chat_error.py
```diff
@@ -0,0 +1,227 @@
...
+class MockModelConfig:
+    task = "generate"
+    runner_type = "generate"
+    tokenizer = MODEL_NAME
+    trust_remote_code = False
+    tokenizer_mode = "auto"
+    max_model_len = 100
+    tokenizer_revision = None
+    multimodal_config = MultiModalConfig()
+    hf_config = MockHFConfig()
```
markmc, line 42: Need to add:

```
logits_processors: list[str] | None = None
```

to match #27560

Test failures:

```
FAILED tests/entrypoints/openai/test_chat_error.py::test_chat_error_non_stream - AttributeError: 'MockModelConfig' object has no attribute 'logits_processors'
FAILED tests/entrypoints/openai/test_chat_error.py::test_chat_error_stream - AttributeError: 'MockModelConfig' object has no attribute 'logits_processors'
```

markmc (dismissed):

vllm/v1/core/sched/scheduler.py
wseaton: noted, thanks!

vllm/entrypoints/openai/serving_completion.py
```diff
@@ -498,8 +506,12 @@ async def completion_stream_generator(
             # report to FastAPI middleware aggregate usage across all choices
             request_metadata.final_usage_info = final_usage_info
 
+        except GenerationError as e:
+            logger.exception("Error in completion stream generator.")
+            yield f"data: {self._convert_generation_error_to_streaming_response(e)}\n\n"
         except Exception as e:
             # TODO: Use a vllm-specific Validation Error
+            logger.exception("Error in completion stream generator.")
```
njhill, line 513: Curious why we are now logging the exception here but not in other cases?

vllm/entrypoints/openai/serving_engine.py
```diff
@@ -769,6 +801,37 @@ def create_streaming_error_response(
         )
         return json_str
 
+    def _handle_error_finish_reason(
```
njhill: wdyt about different name? I think it would make the code a bit easier to understand because it's then clear what the method does
```suggestion
    def _raise_if_error(
```

vllm/entrypoints/openai/serving_responses.py
```diff
@@ -640,6 +646,12 @@ async def responses_full_generator(
                     status = "incomplete"
                 elif context.finish_reason == "abort":
                     status = "cancelled"
+                elif context.finish_reason == "error":
+                    logger.error(
+                        "Request %s failed with internal error during generation",
+                        request.request_id,
+                    )
+                    raise GenerationError("Internal server error")
```
njhill: Use same method here?
```suggestion
                else:
                    self._handle_error_finish_reason(context.finish_reason, request.request_id)
```

vllm/entrypoints/openai/serving_responses.py
```diff
@@ -1044,6 +1061,9 @@ async def _run_background_request_stream(
             async for event in generator:
                 event_deque.append(event)
                 new_event_signal.set()  # Signal new event available
+        except GenerationError as e:
+            logger.exception("Background request failed for %s", request.request_id)
+            response = self._convert_generation_error_to_response(e)
```
njhill: Similar to other comment, it feels like the error logging is a bit inconsistent. We should ideally log in a single equivalent place in all cases (perhaps that's actually within `_convert_generation_error_to_response`?)

vllm/entrypoints/openai/serving_completion.py
```diff
@@ -514,6 +526,10 @@ def request_output_to_completion_response(
         tokenizer: TokenizerLike | None,
         request_metadata: RequestResponseMetadata,
     ) -> CompletionResponse:
+        for final_res in final_res_batch:
+            for output in final_res.outputs:
+                self._handle_error_finish_reason(output.finish_reason, request_id)
```
njhill: Call this in the loop below instead? Since it will be rare, we don't really care if some unused work is done, probably better than having an additional loop over all of the outputs on the happy path.

njhill (commented):
@wseaton really sorry for taking so long to get back to this. Thanks for all of the great work and perseverance/patience!

And thanks a lot to @markmc @sdavidbd @kfirwolfson for the really thorough reviews.

Just have a few minor comments. I guess the main observation is that our logging of these erorrs seems a bit inconsistent. I'm sure we can (finally) get this merged this week.

vllm/entrypoints/openai/serving_completion.py
wseaton: good callout, will remove

vllm/entrypoints/openai/serving_responses.py
wseaton: Makes sense, we will get to remove a lot of call site logging and this will also make it so streaming logs (it doesn't currently)

wseaton: I assume we should add the same exception logging to `_convert_generation_error_to_streaming_response()` right?

njhill: Yes I think so assuming we add it in `_convert_generation_error_to_response`, but I didn't fully inspect all the paths to determine whether this does actually make the most sense.

njhill: When these errors occur, is an error already logged earlier on before the exception propagates? If so then I think better not to log here, if not then we could add the log statements to these functions.

wseaton: In the case of `GenerationError` in the scheduler we log: 
```python
  logger.error(
      "Failing %d request(s) due to KV load failure "
      "(failure_policy=fail, %d tokens affected). Request IDs: %s",
      total_failed_requests,
      total_failed_tokens,
      all_failed_req_ids,
  )
```
So another generic `logger.error` in the request path seems sufficient.

I'm still not super happy with the current helpers we have for error handling, I'm thinking about further consolidating into something like this:

```python
def _make_error_response(self, e: Exception) -> ErrorResponse:
      if isinstance(e, GenerationError):
          logger.error("Generation error: %s", e)
          err_type = "InternalServerError"
          status_code = getattr(e, "status_code", HTTPStatus.INTERNAL_SERVER_ERROR)
      else:
          logger.exception("Error in request")
          err_type = "BadRequestError"
          status_code = HTTPStatus.BAD_REQUEST
      return self.create_error_response(str(e), err_type=err_type, status_code=status_code)

..then later..

  except Exception as e:
      return self._make_error_response(e)
```

njhill (approved):

vllm/entrypoints/openai/serving_responses.py
njhill: As discussed, would be good to address this in a follow-on PR

PR #26821, opened 2025-10-14
[Bugfix] skip cuda graph for drafter when running with eager

benchislett:
## Purpose

Patch for `speculative_config.enforce_eager` not being respected in the draft model dummy run.

This flag was introduced for DSV3.2 MTP which currently must run in eager mode, so it is not a common crash.


pavanimajety (approved):

benchislett:
disabling auto-merge, forgot I still need a stamp

PR #26826, opened 2025-10-14
[bugfix] Fix test using nm-testing/TinyLlama-1.1B-Chat-v1.0-FP8-e2e

angelayi:
## Purpose

Fixes the following test failure when I run on main with regards to the model `nm-testing/TinyLlama-1.1B-Chat-v1.0-FP8-e2e`. I'm not sure if the ModelConfig specific to this model is really necessary so I just removed it, and the tests seem to be fine?

image

## Test Plan
`pytest -sv tests/compile/test_sequence_parallelism.py` passes now

cc @ProExpertProg @zou3519

### DIFF
--- a/tests/compile/test_async_tp.py
+++ b/tests/compile/test_async_tp.py
@@ -329,13 +329,7 @@ def async_tp_pass_on_test_model(
         ),
     )
     vllm_config.device_config = DeviceConfig(device=torch.device("cuda"))
-
-    # this is a fake model name to construct the model config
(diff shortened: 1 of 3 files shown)


ProExpertProg (approved):

mgoin (changes requested):
This fix is wrong at least for test_fusion_all_reduce.py‎ because we need an fp8 static per-tensor model. I've pushed a proper fix here https://github.com/vllm-project/vllm/pull/26816

PR #26832, opened 2025-10-14
create is_in_the_same_node on cpu

helunwencser:
Summary: `in_the_same_node_as` is only called on the host. This PR makes `is_in_the_same_node` to be created on the host directly; otherwise, this function fails if `torch.set_default_device()` is called during device initialization because `is_in_the_same_node` will be created on the device by default.

Test Plan:
Before:
vLLM fails to initialize distributed group due to the following error:
```RuntimeError: No backend type associated with device type XXX```

After:
vLLM can initialize distributed group successfully

Differential Revision: D84630888


houseroad (commented):
Looks good. 

Can we fix the linter (pre-commit)?

Also can we add a unittest?

houseroad (approved):

PR #26844, opened 2025-10-14
Guard SM100 CUTLASS MoE macro to SM100 builds

jonahbernard:
## Purpose
Fix symbol issues (https://github.com/vllm-project/vllm/issues/26843) on SM120 builds where grouped_mm_c3x_sm100.cu is not compiled but ENABLE_CUTLASS_MOE_SM100 was defined by unrelated blocks.

Make ENABLE_CUTLASS_MOE_SM100 accurately reflect SM100 grouped MoE availability:

Defined only on SM100 arches (10.0a/10.1a/10.3a on CUDA 12.8–12.9, or 10.0f on CUDA ≥13.0).

**Note:** I may be misinterpreting the purpose of  `ENABLE_CUTLASS_MOE_SM100` as it may be being used not literally to mean only SM100. However, if this is the case, this imprecision should be addressed to fix the code and reduce confusion. 

## Test Plan

I’m unable to build for all affected architectures; it should be tested on SM90, SM100, SM110, and SM120.

## Test Result

Not available.
---
<details>

</details>


Jonahcb:
Hi! @mgoin thoughts on this:
> Note: I may be misinterpreting the purpose of `ENABLE_CUTLASS_MOE_SM100` as it may be being used not literally to mean only SM100. However, if this is the case, this imprecision should be addressed to fix the code and reduce confusion.

CMakeLists.txt
mgoin: Should we remove 12.x from this block?

mgoin (commented):
I agree we should only set ENABLE_CUTLASS_MOE_SM100 if we are compiling for 10.x, maybe even more strict. I'm not sure I understand this fix though as some of these cases already specialize for 10.x

CMakeLists.txt
jonahbernard: @mgoin Yeah, that sounds good. I am relatively new to NVIDIA and can't find much information about this online, but is 11.x compatible with 10.x? If not, shouldn't we remove 11.x as well? Thanks!

johnnynunez:
LGTM @mgoin

CMakeLists.txt
johnnynunez: yes, this is a mistake

huydhn (approved):

CMakeLists.txt
```diff
@@ -594,9 +594,9 @@ if(VLLM_GPU_LANG STREQUAL "CUDA")
...
   # FP4 Archs and flags
   if(${CMAKE_CUDA_COMPILER_VERSION} VERSION_GREATER_EQUAL 13.0)
-    cuda_archs_loose_intersection(FP4_ARCHS "10.0f;11.0f;12.0f" "${CUDA_ARCHS}")
+    cuda_archs_loose_intersection(FP4_ARCHS "10.0f" "${CUDA_ARCHS}")
   else()
-    cuda_archs_loose_intersection(FP4_ARCHS "10.0a;10.1a;12.0a;12.1a" "${CUDA_ARCHS}")
+    cuda_archs_loose_intersection(FP4_ARCHS "10.0a;10.1a" "${CUDA_ARCHS}")
   endif()
   if(${CMAKE_CUDA_COMPILER_VERSION} VERSION_GREATER_EQUAL 12.8 AND FP4_ARCHS)
     set(SRCS
```
mgoin, line 603: I'm not sure if some of these sources are needed for 12.0, such as the nvfp4 quant source

mgoin (approved):
This looks good to me, I'm just unsure about the quant utils that might be shared with 11 and 12

CMakeLists.txt
ECMGit: I think we should keep 11.0f and 12.0f which are required for sm120 and sm110 device

ECMGit: These source are actually required for nvfp4 for sm120 and sm110. 10.0f will be still compiled if we leave them as they are.

CMakeLists.txt
```diff
@@ -716,9 +716,9 @@ if(VLLM_GPU_LANG STREQUAL "CUDA")
   endif()
 
   if(${CMAKE_CUDA_COMPILER_VERSION} VERSION_GREATER_EQUAL 13.0)
-    cuda_archs_loose_intersection(SCALED_MM_ARCHS "10.0f;11.0f;12.0f" "${CUDA_ARCHS}")
+    cuda_archs_loose_intersection(SCALED_MM_ARCHS "10.0f" "${CUDA_ARCHS}")
   else()
-    cuda_archs_loose_intersection(SCALED_MM_ARCHS "10.0a;10.1a;10.3a;12.0a;12.1a" "${CUDA_ARCHS}")
```
ECMGit, line 722: In CUDA 13.0, I remember 12.0a and 12.1a should be 12.0f, 12.0f is good enough for both archs. I would suggest keep it as it is.

CMakeLists.txt
johnnynunez: @Aidyn-A what do you think?

johnnynunez:
> ## Purpose
> Fix symbol issues (#26843) on SM120 builds where grouped_mm_c3x_sm100.cu is not compiled but ENABLE_CUTLASS_MOE_SM100 was defined by unrelated blocks.
> 
> Make ENABLE_CUTLASS_MOE_SM100 accurately reflect SM100 grouped MoE availability:
> 
> Defined only on SM100 arches (10.0a/10.1a/10.3a on CUDA 12.8–12.9, or 10.0f on CUDA ≥13.0).
> 
> **Note:** I may be misinterpreting the purpose of `ENABLE_CUTLASS_MOE_SM100` as it may be being used not literally to mean only SM100. However, if this is the case, this imprecision should be addressed to fix the code and reduce confusion.
> 
> ## Test Plan
> I’m unable to build for all affected architectures; it should be tested on SM90, SM100, SM110, and SM120.
> 
> ## Test Result
> ## Not available.
> Essential Elements of an Effective PR Description Checklist

could you check why CI is not passing? cc @mgoin

eugr:
@johnnynunez - looks like this recent dependency issue with apache-tvm-ffi. The main branch build is also failing. From the CI job logs:

```
No solution found when resolving dependencies:
[2025-10-29T05:55:33Z] 1.626   ╰─▶ Because there is no version of apache-tvm-ffi==0.1.0b15 and
[2025-10-29T05:55:33Z] 1.626       flashinfer-python==0.4.1 depends on apache-tvm-ffi==0.1.0b15, we can
[2025-10-29T05:55:33Z] 1.626       conclude that flashinfer-python==0.4.1 cannot be used.
[2025-10-29T05:55:33Z] 1.626       And because you require flashinfer-python==0.4.1, we can conclude that
[2025-10-29T05:55:33Z] 1.626       your requirements are unsatisfiable.
[2025-10-29T05:55:33Z] 1.626
[2025-10-29T05:55:33Z] 1.626       hint: `apache-tvm-ffi` was requested with a pre-release marker (e.g.,
[2025-10-29T05:55:33Z] 1.626       apache-tvm-ffi==0.1.0b15), but pre-releases weren't enabled (try:
[2025-10-29T05:55:33Z] 1.626       `--prerelease=allow`)
```

eugr:
> mmm maybe he has to merge main to his PR due that apache tvm ffi now has stable release

This also happens in main, because of dependency on flashinfer==0.4.1 that triggers this apache dependency. There is a new version of flashinfer==0.5.0, I wonder if they should use that instead (assuming no breaking changes).

johnnynunez:
> > mmm maybe he has to merge main to his PR due that apache tvm ffi now has stable release
> 
> This also happens in main, because of dependency on flashinfer==0.4.1 that triggers this apache dependency. There is a new version of flashinfer==0.5.0, I wonder if they should use that instead (assuming no breaking changes).

there is an open PR. Be patient

Azure-Tang:
> On Spark, you need two things:
> 
> 1. Apply the changes from this PR.
> 2. Set environment variables before building VLLM (this is also essential, otherwise cmake will not pick up correct architecture):
> 
> export TORCH_CUDA_ARCH_LIST=12.0f export TRITON_PTXAS_PATH=/usr/local/cuda/bin/ptxas

I tried to export this two env then compile

but still encounter error. 
```
NotImplementedError: No compiled get_cutlass_moe_mm_data: no cutlass_scaled_mm kernel for CUDA device capability: 121. Required capability: 90 or 100
```
Is this means we cannot run moe model on spark now? @eugr

Azure-Tang:
> > I tried to export this two env then compile
> 
> Have you applied the changes from this PR to CMakeLists.txt?

sure

johnnynunez:
> > > I tried to export this two env then compile
> > 
> > 
> > Have you applied the changes from this PR to CMakeLists.txt?
> 
> sure. And I successfully run qwen3-30b-a3b. But moe models are keep failing

di you tried new pytorch 2.9.1 and triton 3.5.1? That fix some issues for spark related to triton kernels

swtb3-ryder:
> > > di you tried new pytorch 2.9.1 and triton 3.5.1? That fix some issues for spark related to triton kernels
> > 
> > 
> > I used pytorch 2.9.0. But I don't think thats the case cause the cutlass_scaled_fp4_mm is a cuda kernel? And maybe it seems we don't get some thing like "cutlass_scaled_fp4_mm_sm120f" or "cutlass_scaled_fp4_mm_sm121" in vllm, which related to spark's arch.
> > image
> 
> yes, there is a lack of support for spark right now, we are working on it

@johnnynunez im interested in following developments with spark support, where can I do that?

eugr:
> it pulls in the CPU-only version of torch

Run `python use_existing_torch.py` before building.

trias702:
> > it pulls in the CPU-only version of torch
> 
> Run `python use_existing_torch.py` before building.

That doesn't work though, because all `python use_existing_torch.py` does is just erase torch from pyproject.toml. Then when you run `python -m build --wheel` the isolated environment fails immediately because it doesn't download any torch at all, neither the cpu or cuda one, so you get an ImportError "no module torch".

`python -m build -wheel` creates an isolated environment for the build, completely separate from the local environment, and it pulls everything in from the toml, so basically just like how `uv run` works. So if you remove torch from the toml, the isolated build will fail as there's no torch at all anymore.

eugr:
Any reason you can't use --no-build-isolation?

trias702:
> Any reason you can't use --no-build-isolation?

I can, in which case it works just fine, however, that is only because my local Python environment already mirrors the contents of the toml file. The point of an isolated build is that it is essentially running `uv run python -m build --wheel` under the bonnet, and this is super useful because it means I don't have to worry if my local python environment matches what's necessary to build. I could have a different local cuda or torch version, whatever. But I can still build vllm any time by creating an isolated environment on the fly that always matches the toml. It's super useful and is the correct way to build complex packages as there is guaranteed no version mismatches.

Only on the DGX Spark it's not working correctly because the toml is pointing to the CPU-only install for torch, and I'm wondering how we change that to be the cuda version instead. As it stands currently, I believe it's a legitimate bug, and should affect any unix box, not just the Spark, but I don't have another unix box with a GPU to know for certain.

eugr:
> I could have a different local cuda or torch version, whatever.

Won't venv give you that? venv works well with uv, just use uv venv to create it and then activate as usual.

trias702:
> > I could have a different local cuda or torch version, whatever.
> 
> Won't venv give you that? venv works well with uv, just use uv venv to create it and then activate as usual.

venv is what uv uses under the bonnet, same with python build. Unfortunately, the toml file in vllm isn't telling the venv system to use the cuda variant of torch. You need to explicitly tell it to do that, like how Nemo-RL does it: https://github.com/NVIDIA-NeMo/RL/blob/2951ce37ae1e568ab5ff063997da478a85551dfa/pyproject.toml#L185

eugr:
@Jonahcb  - looks like you need to merge main into your PR for checks to pass...

CMakeLists.txt
```diff
@@ -717,9 +717,9 @@ if(VLLM_GPU_LANG STREQUAL "CUDA")
...
+    cuda_archs_loose_intersection(SCALED_MM_ARCHS "10.0a;10.1a;10.3a" "${CUDA_ARCHS}")
```
wrmedford, line 722: This is going to break in PTX 9.0 I think. 10.1a is going to be 11.0a at that point.

CMakeLists.txt
```diff
@@ -669,7 +669,7 @@ if(VLLM_GPU_LANG STREQUAL "CUDA")
   endif()
 
   if(${CMAKE_CUDA_COMPILER_VERSION} VERSION_GREATER_EQUAL 13.0)
-    cuda_archs_loose_intersection(SCALED_MM_ARCHS "10.0f;11.0f" "${CUDA_ARCHS}")
```
johnnynunez: here mantain 11.0f

CMakeLists.txt
```diff
@@ -717,9 +717,9 @@ if(VLLM_GPU_LANG STREQUAL "CUDA")
   endif()
 
   if(${CMAKE_CUDA_COMPILER_VERSION} VERSION_GREATER_EQUAL 13.0)
-    cuda_archs_loose_intersection(SCALED_MM_ARCHS "10.0f;11.0f;12.0f" "${CUDA_ARCHS}")
```
johnnynunez: share here: mantain 11.0f

CMakeLists.txt
johnnynunez: in < 13.0 is 10.1, in >=13.0 is 11.0

CMakeLists.txt
Aidyn-A: Looking at the corresponding CUTLASS kernel configs I see only `Sm100`:
https://github.com/vllm-project/vllm/blob/db56a59970a84842da2adc3aa64e436f42448b48/csrc/quantization/w8a8/cutlass/moe/grouped_mm_c3x_sm100.cu#L13-L62

I might be wrong, but these kernels will work on `sm_110f`, but definitely will not work on `sm_120f`.

@johnnynunez, have you tested MOE kernels on DGX Spark when you added `sm_120f`?

CMakeLists.txt
Aidyn-A: Agree, `11.0f` should stay.

CMakeLists.txt
Aidyn-A: It is worth keeping it, as long as there are CUTLASS kernels for `sm_120`. But are there any?

mgoin (changes requested):
Blocking due to unresolved comments. For now prioritizing https://github.com/vllm-project/vllm/pull/28938

PR #26845, opened 2025-10-14
[CI] Fix mypy for `vllm/executor`

yewentao256:
## Purpose

Part of the https://github.com/vllm-project/vllm/issues/26533

## Test

```bash
(vllm-user-6) vllm-user-6@centralia:~/vllm-source$ pre-commit run --hook-stage manual mypy-3.10 -a
```


DarkLight1337 (commented):
This directory is V0 code which will be removed soon anyways, I think there is not much value in fixing mypy here

hmellor:
Given that the work has been done and the tests pass, could we merge anyway?

The timeline for removal of all remaining V0 code is uncertain and doing this would allow us to push forward with running mypy directly instead of via our complicated script.

DarkLight1337 (approved):
OK then

PR #26849, opened 2025-10-14
[Compressed Tensors] Always clone output for compile robustness

kylesayrs:
## Purpose ##
* Ensure that compressed tensors transforms logic compiles consistently across hardware

## Background ##
Output transforms are annoying to execute performantly, specifically because they need to be applied individually to each of the fused layer output partitions. These output partitions are sharded along the column dimension, meaning that they cannot be cleanly separated into contiguous blocks without a copy operation.

```python3
if self.output_transform is not None:
  for part_id, (start, length) in enumerate(self.partition_ranges):
      x[:, start : start + length] = self.output_transform(
          x[:, start : start + length].contiguous(), part_id=part_id
      )
```

However, for some reason, on certain hardware, torch.compile does not recognize this `contiguous` call as a copy operation, and releases an error complaining about ambiguous memory addresses. @brian-dellabetta

```
(EngineCore_DP0 pid=889)   File "/root/.clearml/venvs-builds/3.10/lib/python3.10/site-packages/vllm/model_executor/layers/quantization/compressed_tensors/transform/linear.py", line 146, in apply
(EngineCore_DP0 pid=889)     x[:, start:start + length] = self.output_transform(
(EngineCore_DP0 pid=889) RuntimeError: unsupported operation: some elements of the input tensor and the written-to tensor refer to a single memory location. Please clone() the tensor before performing the operation.
```

We aim to make this code more consistent by always copying values. The table below shows that there are no performance differences as a result of this change, and output transforms are not the performant target for transforms anyways.

Model | Before | After
-- | -- | --
Base | 0.461 | 0.461
w4a16 | 0.363 | 0.363
quip-w4a16 | 0.441 | 0.441
quipv-w4a16 | 0.389 | 0.388

## Changes ##
* Replace `contiguous` with `clone` when fetching the inputs to compressed tensors output transforms


yewentao256 (commented):
Would this affect speed? Could you have a test for E2E throughput using `vllm bench serve...`

kylesayrs:
@yewentao256 See the table in the PR description! I suspect we were always cloning in most cases

brian-dellabetta (approved):
Glad the latencies all match! That means our assumption that it was always non-contiguous is correct

kylesayrs:
@yewentao256 Not really, as the hadamard kernels are require contiguous memory to work properly, and and unfortunately output partitions are split along the column dimension (not contiguous memory). Therefore, some operations to make the memory contiguous, then split it up again after the operation is completed, are required.

yewentao256 (approved):
LGTM, thanks for the work!

mgoin (approved):
LGTM

PR #26853, opened 2025-10-14
[Misc] Update TritonLanguagePlaceholder to have attributes that are used by Flash Linear Attention ops.

madongfly:
Summary:
Update TritonLanguagePlaceholder to have attributes that are used by Flash Linear Attention ops.

Signed-off-by: Xudong Ma [mxd@meta.com](mailto:mxd@meta.com)

Differential Revision: D84470467

### DIFF
--- a/vllm/triton_utils/importing.py
+++ b/vllm/triton_utils/importing.py
@@ -98,3 +98,6 @@ class TritonLanguagePlaceholder(types.ModuleType):
         self.int64 = None
         self.int32 = None
         self.tensor = None
+        self.exp = None
+        self.log = None
+        self.log2 = None


madongfly:
> @madongfly could you paste the test you are fixing and its result?

vllm/model_executor/layers/fla/ops/op.py has code pieces like:

```
@triton.jit
        def div_normal(x, y):
            return x / y
        div = div_normal
        exp = tl.exp
```

which triggers AttributeError: module 'triton.language' has no attribute 'exp' when triton is not installed.

yeqcharlotte:
> > @madongfly could you paste the test you are fixing and its result?
> 
> vllm/model_executor/layers/fla/ops/op.py has code pieces like:
> 
> ```
> @triton.jit
>         def div_normal(x, y):
>             return x / y
>         div = div_normal
>         exp = tl.exp
> ```
> 
> which triggers AttributeError: module 'triton.language' has no attribute 'exp' when triton is not installed.

let's double check which unit tests could have captured this and update it?

madongfly:
> > > @madongfly could you paste the test you are fixing and its result?
> > 
> > 
> > vllm/model_executor/layers/fla/ops/op.py has code pieces like:
> > ```
> > @triton.jit
> >         def div_normal(x, y):
> >             return x / y
> >         div = div_normal
> >         exp = tl.exp
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which triggers AttributeError: module 'triton.language' has no attribute 'exp' when triton is not installed.
> 
> let's double check which unit tests could have captured this and update it?

IIUC, this class is a dummy class to work around some CI environment that doesn't have triton, now the code piece I provided above shows vllm/model_executor/layers/fla/ops/op.py reference to it, then it is imported by vllm/model_executor/layers/fla/ops/chunk_delta_h.py through "from .op import exp".

I think the logic of why this is needed is pretty clear, do we really want to artificially add unused import to trigger the import chain in a random unit test? For such a minor fix on a dummy class whose whole purpose is to work around other tests?

yeqcharlotte (approved):
LGTM. sigh ... we should probably look into why we are importing triton when triton is not available.

PR #26862, opened 2025-10-15
[Docs] Move build.inc into arm.inc

windsonsea:
Based on https://github.com/vllm-project/vllm/pull/26747#issuecomment-3400640926, move the build.inc content into arm.inc page.

cc @hmellor


DarkLight1337 (approved):
Thanks

Issue #26133, opened 2025-10-02
[RFC]: Support Context Parallelism with Fully Sharded KV Cache and Ring Attention


qiruiyangmeta:
### Motivation.

Context parallelism introduces an additional degree of parallelism to LLM inference. While tensor parallelism and pipeline parallelism focus on distributing model weights and layers across devices, context parallelism specifically targets the parallel processing of multiple input contexts or sequences. By combining context parallelism with parallelisms, systems can achieve more scalable and efficient inference, leveraging all forms of parallelism to maximize hardware utilization and reduce latency. 
<img width="1166" height="550" alt="Image" src="https://github.com/user-attachments/assets/9c9a5f66-6822-4681-834e-38a46ea943c3" />
Context parallelism improves performance as the context length grows by distributing both computation and the KV cache across multiple GPUs. This approach effectively lowers processing latency and can also decrease the memory required per GPU potentially, especially when dealing with extremely large KV caches (such as sequence lengths on the order of 1 million tokens).

### Proposed Change.

Within the model, attention is the only component that has dependency on the sequence dimension, since each token must attend to all previous tokens in the same sequence. In contrast, FFN and element-wise operations are performed independently for each token. For a more in-depth understanding of context parallelism in LLM inference, including partial attention, read the MLSys paper available at[ https://arxiv.org/pdf/2411.01783](https://arxiv.org/pdf/2411.01783).
To implement context parallelism in vLLM, the design needs to:

- Be aware of these dependencies to minimize synchronization overhead,
- Remain flexible to support various backends, and
- Avoid major changes to core components to ensure system stability.

## Partition Sequence Dimension

Causal attention imposes a varying computational load for each token, as shown in the following figure. To ensure an even workload distribution, tokens should be partitioned across different context parallelism (CP) ranks. Specifically, the sequence is divided into 2 × cp_world_size chunks. Each CP rank i is assigned both the i-th chunk and the (2 × cp_world_size - i - 1)-th chunk. This approach helps balance the compute load among all CP ranks.

<img width="521" height="463" alt="Image" src="https://github.com/user-attachments/assets/b188c8d9-5a93-43c2-a5d2-a46bcd21f3c7" />

## Prefill

[...]

Issue #25749, opened 2025-09-26
[RFC]: Support Prefill Context Parallel (PCP)


pisceskkk:
### Motivation.

A follow-up to add prefill context parallel(PCP) after DCP (PR [#23734](https://github.com/vllm-project/vllm/pull/23734)) to optimize TTFT.

In DCP, the KV Cache is partitioned along the sequence dimension into several segments (interleave style), achieving context parallelism during the decode phase. 

---

<img width="3484" height="1004" alt="Image" src="https://github.com/user-attachments/assets/fa5edbe2-3e84-4720-a576-5d80c1a0c6e6" />

--- 

<img width="1988" height="864" alt="Image" src="https://github.com/user-attachments/assets/c57a50d0-e295-423b-b92d-31a11a300680" />

---

Similarly, PCP will split the entire request along the sequence dimension during the prefill phase and further partitions the KVCache along the sequence dimension.

### Proposed Change.

Aside from the attention module, almost all other modules do not involve contextual dependencies and can naturally adapt to context parallelism.

In the attention module, we adopt the following strategy to implement PCP (chunked prefill is not considered for now), marked as **red.** 
First, for the KV, we perform an AllGather op along the sequence dimension within the PCP group to obtain the complete KV values. Then, the kvcache is stored according to the slot_mapping. 
For attention computation, since we have obtained the complete KV, we only need to carefully design the custom mask and perform normal attention.

<img width="1661" height="592" alt="Image" src="https://github.com/user-attachments/assets/1830b122-9acc-4903-9d7c-3d7beda957e1" />

During the decode phase, modules other than attention involve redundant computations (since num new tokens = 1) across PCP group.
In the attention module, we first execute the original DCP computation logic within the respective PCP group. 
Then, before updating the attention output using lse, we perform an AllGather within the PCP group to obtain complete sequence information.  And the subsequent steps remain unaffected. 

<img width="1505" height="467" alt="Image" src="https://github.com/user-attachments/assets/1369518a-e0e9-4ae5-b922-c6fafa45be8a" />

**Summary**
- **The creation of Prefill Context Parallel communication domains.** Unlike DCP communication domains, which are subdivisions within the TP domain, PCP communication domains stand alongside DP, PP, and TP, and affect the total device count allocation.
[...]

PR #25438, opened 2025-09-23
[DCP] Support Decode Context Parallel (DCP) for GQA with Flashinfer

gjc0824:
## Purpose
This PR adds Decode Context Parallel (DCP) support for GQA follwing PR [#23734](https://github.com/vllm-project/vllm/pull/23734) and PR [#24864](https://github.com/vllm-project/vllm/pull/24864). Current implementation based on FlashInfer Attention.

FlashInfer inserts the current query KV into the cache before computation. Each query then attends to both its own KV and the context KV on the local device, with LSE applied to correct the attention outputs.

- In the prefill/partial-prefill stage, custom mask is added to support interleaved KV cache with FlashInfer.  
```
q_lens = 8, total_lens = 25 , group_size = 4, local_rank = 0

stored kv cache
rank0: 0 4 8 12 16 20 24
rank1: 1 5 9 13 17 21
rank2: 2 6 10 14 18 22
rank3: 3 7 11 15 19 23

rank0 custom mask
q\kv    0      4      8     12    16      20      24
17   True,  True,  True,  True,  True,  False, False
18   True,  True,  True,  True,  True,  False, False
19   True,  True,  True,  True,  True,  False, False
20   True,  True,  True,  True,  True,  True,  False
21   True,  True,  True,  True,  True,  True,  False
22   True,  True,  True,  True,  True,  True,  False
23   True,  True,  True,  True,  True,  True,  False
24   True,  True,  True,  True,  True,  True,  True
```
- In the decode stage, this PR follows the DCP decode approach from MLA, i.e., all-gathering Q and lse, then correcting the attn out before performing reduce-scatter.

## Test Plan
Qwen/Qwen3-235B-A22B
```
export VLLM_ATTENTION_BACKEND='FLASHINFER'
vllm serve Qwen/Qwen3-235B-A22B --gpu-memory-utilization 0.9 --tensor-parallel-size 8 --decode-context-parallel-size 2
```
## Test Result
- gsm8k eval
```
dcp=1
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8578|±  |0.0068|
|     |       |strict-match    |     5|exact_match|↑  |0.8415|±  |0.0071|

dcp=2
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
[...]

PR #26864, opened 2025-10-15
[Feature] Support Prefill Context Parallel (PCP) for GQA flashinfer

LookAround0301:
# Purpose
This PR adds the Prefill Context Parallelism (PCP) feature, which corresponds to DCP. For specific implementation details, please refer to the RFC [#25749](https://github.com/vllm-project/vllm/issues/25749).  

- Modified `ModelRunner.py` for CP partitioning logic for tokens;  
- Modified `flashinfer.py` to adapt the FlashInfer backend for GQA to PCP.
- Modified `block_tables.py` to extend the KV cache storage based on DCP&PCP;  
- Added a communication group `cp_group` for PCP;  
- Added necessary command-line arguments to control parallelism for PCP;  

# Test Plan
Qwen/Qwen3-32B

```bash
vllm serve Qwen/Qwen3-32B --gpu-memory-utilization 0.9 --tensor-parallel-size 2 --context-parallel-size 2
```

# Test Result
gsm8k eval

TP2 batchsize 256 max_out_len 1024

PCP1
| dataset | version | metric | mode | vllm-api-stream-chat |
| gsm8kdataset | - | accuracy | gen | 88.02 |

PCP2
| dataset | version | metric | mode | vllm-api-stream-chat |
| gsm8kdataset | - | accuracy | gen | 87.87 |

# TODOs
- [x] UT for PCP;
- [ ] Make PCP of flashinfer compatible with DCP after PR [#25438](https://github.com/vllm-project/vllm/pull/25438)  is merged;
- [ ] Make block-level interleaved KV cache storage compatible after PR [#26696](https://github.com/vllm-project/vllm/pull/26696) is merged;

**Feature works** (These items will be tackled in follow-up PRs; community contributions are warmly welcomed.):
- PCP support for MLA and other backends;
- PCP support for chunked-prefill and prefix caching features;
- PCP support for MTP;
- PCP support for CUDAFullGraph;
- Ring-CP style attention backend algorithm, ref RFC [#26133](https://github.com/vllm-project/vllm/issues/26133).

### DIFF
--- a/tests/distributed/test_context_parallel.py
+++ b/tests/distributed/test_context_parallel.py
@@ -64,7 +64,7 @@ class CPTestSettings:
         for eager_mode_val in [False]:
             for pp_multiplier in [1]:
                 # TODO(qcs): Test the effect of mixed activation
-                # when CP and DCP are compatible.
+                # when PCP and DCP are compatible.
                 for pcp_multiplier, dcp_multiplier in zip([1, 2, 1], [0.5, 1, 1]):
                     for chunked_prefill_val in [True]:
                         parallel_setups.append(
--- a/tests/distributed/test_context_parallel.py
+++ b/tests/distributed/test_context_parallel.py
@@ -30,6 +30,7 @@ class ParallelSetup(NamedTuple):
     tp_size: int
     pp_size: int
     dcp_size: int
+    pcp_size: int
     eager_mode: bool
     chunked_prefill: bool
 
@@ -37,6 +38,7 @@ class ParallelSetup(NamedTuple):
 class CPTestOptions(NamedTuple):
     multi_node_only: bool
     load_format: str | None = None
+    attn_backend: str = "FLASH_ATTN"
 
 
 @dataclass
@@ -52,20 +54,25 @@ class CPTestSettings:
         tp_base: int = 4,
         pp_base: int = 1,
         dcp_base: int = 1,
+        pcp_base: int = 1,
         multi_node_only: bool = False,
         runner: RunnerOption = "auto",
         load_format: str | None = None,
+        attn_backend: str = "FLASH_ATTN",
     ):
         parallel_setups = []
         for eager_mode_val in [False]:
             for pp_multiplier in [1]:
-                for dcp_multiplier in [0.5, 1]:
+                # TODO(qcs): Test the effect of mixed activation
+                # when CP and DCP are compatible.
+                for pcp_multiplier, dcp_multiplier in zip([1, 2, 1], [0.5, 1, 1]):
                     for chunked_prefill_val in [True]:
                         parallel_setups.append(
                             ParallelSetup(
                                 tp_size=tp_base,
                                 pp_size=pp_multiplier * pp_base,
                                 dcp_size=int(dcp_multiplier * tp_base),
+                                pcp_size=int(pcp_multiplier * pcp_base),
                                 eager_mode=eager_mode_val,
                                 chunked_prefill=chunked_prefill_val,
                             )
@@ -75,7 +82,9 @@ class CPTestSettings:
             distributed_backends=["mp"],
             runner=runner,
             test_options=CPTestOptions(
-                multi_node_only=multi_node_only, load_format=load_format
+                multi_node_only=multi_node_only,
+                load_format=load_format,
+                attn_backend=attn_backend,
             ),
         )
 
@@ -108,11 +117,12 @@ def _compare_cp_with_tp(
         tp_size,
         pp_size,
         dcp_size,
+        pcp_size,
         eager_mode,
         chunked_prefill,
     ) = parallel_setup
 
-    multi_node_only, load_format = test_options
+    multi_node_only, load_format, attn_backend = test_options
 
     model_info = HF_EXAMPLE_MODELS.find_hf_info(model_id)
     model_info.check_transformers_version(on_fail="skip")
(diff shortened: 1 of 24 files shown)


tests/distributed/test_context_parallel.py
```diff
@@ -29,13 +29,15 @@ class ParallelSetup(NamedTuple):
     tp_size: int
     pp_size: int
     dcp_size: int
+    cp_size: int
```
LucasWilkinson: I think we should call the `pcp_size`, thoughts?

tests/distributed/test_context_parallel.py
```diff
@@ -190,6 +201,8 @@ def _compare_cp_with_tp(
         str(pp_size),
         "--decode-context-parallel-size",
         str(dcp_size),
+        "--context-parallel-size",
```
LucasWilkinson: ditto earlier comment; I think this should be `--prefill-context-parallel-size`

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -944,9 +1057,25 @@ def _prepare_inputs(
...
+                   )
+            np.add(self.input_batch.num_computed_tokens_cpu[req_indices],
+                   positions_cp[:total_num_scheduled_tokens],
+                   out=positions_np)
+        else:
+            np.add(self.input_batch.num_computed_tokens_cpu[req_indices],
+                   arange,
+                   out=positions_np)
+            req_indices_for_slotmapping = req_indices
+            positions_np_for_slotmapping = positions_np
```
LucasWilkinson: this feels a bit messy; maybe something like this would be cleaner?:
```
        num_scheduled_tokens = np.array(tokens, dtype=np.int32)
        ...
        # E.g., [2, 5, 3] -> [0, 0, 1, 1, 1, 1, 1, 2, 2, 2]
        req_indices = np.repeat(self.arange_np[:num_reqs],
                                num_scheduled_tokens)
        ...

        np.add(self.input_batch.num_computed_tokens_cpu[req_indices],
               arange,
               out=positions_np)

        self.input_batch.block_table.compute_slot_mapping(
            req_indices, positions_np)
        self.input_batch.block_table.commit_slot_mapping(
            total_num_scheduled_tokens)

        ...
        num_scheduled_tokens, positions_np = self._update_tokens_for_cp( # This would have to be modified to filter 
             num_scheduled_tokens, positions_np)
```

LucasWilkinson (commented):
Thanks for the contribution! Excited for this!

I'll do a more thorough review later but my initial comment is:

Can we try reduce amount of changes overall. I dont think this needs to be this invasive especially in the gpu_model_runner. The gpu model runner is already a very complex piece of code making it very hard for contributors to keep, we should try out best to not add to the complexity.

tests/distributed/test_context_parallel.py
pisceskkk: yes, I have renamed variables like `cp_*`/`context_parallel_*` to `pcp_*`/`prefill_context_parallel_*` to ensure they are distinct from dcp.

vllm/v1/worker/gpu_model_runner.py
pisceskkk: Thank you for the review. Moving `compute_slot_mapping` up to `_update_tokens_for_cp` indeed makes it clearer. I have made the changes and look forward to further review.

riou-chen:
good，it run successfully after merge the new commit. but the performance decrease compare to the base. 
I run the pcp test as above command，and client send request 8K/1K ，2k/1k，1k/1k input/ouput，the test result as follows：
**the 1K/1K  base data:**
image

**the 1K/1K pcp data**
image


**the 2K/1K base data**
image

the 2K/1K PCP data
image



Is there any error in my testing method? Why isn't PCP effective?

pisceskkk:
> but the performance decrease compare to the base.

@riou-chen  During our GPU performance testing, we also identified performance degradation issues. Preliminary analysis suggests two potential causes:

1. The custom mask computation introduces significant overhead. Although we have implemented initial optimizations (please ref [commit](https://github.com/vllm-project/vllm/pull/26864/commits/f0ab17cc6e2708656100931266476ea9b4c1b214)), there still remains some performance loss.
2. We suspect that flashinfer may have performance degradation when handling custom mask computations (though there is no concrete evidence yet).

Therefore, the current GPU implementation still has room for optimization. 

Additionally, based on our implementation and performance testing on NPU, DCP & PCP incur performance losses for short sequences (comparing CP vs. DP under the same TP parallelism). Performance gains are only observed in scenarios with long-sequence inputs exceeding 32K, and the benefits become more pronounced as the sequence length increases.

pisceskkk:
> * The custom mask computation introduces significant overhead. Although we have implemented initial optimizations (please ref [commit](https://github.com/vllm-project/vllm/pull/26864/commits/f0ab17cc6e2708656100931266476ea9b4c1b214)), there still remains some performance loss.
> * We suspect that flashinfer may have performance degradation when handling custom mask computations (though there is no concrete evidence yet).

We have updated the calculation method of flashinfer backend with PCP to avoid the usage of custom_mask. Some performance tests proof that PCP indeed improves TTFT.


**Test results**
tp8
| dataset | version | metric | mode | vllm-api-stream-chat |
| gsm8kdataset | - | accuracy | gen | 88.17 |

tp4pcp2
| dataset | version | metric | mode | vllm-api-stream-chat |
| gsm8kdataset | - | accuracy | gen | 87.80 |

TTFT(ms)
| input_len | 32K | 64K | 128K |
|----- | ----- | ----- | ----- | 
| tp8 |  996 | 2522 | 7131 |
| tp4pcp2 | 851 | 2115 | 5961 |

PR #23391, opened 2025-08-22
[Core] Restructure core loop for async input preparation 

WoosukKwon:
Closes #23233 

This PR restructures the core loop for both synchronous and asynchronous scheduling.

For sync scheduling, this allows overlapping bitmask construction with the model execution.
For async scheduling, this allows overlapping input preparation (and bitmask) with the model execution. Currently, only deserialization of scheduler outputs is overlapped. This could be expanded in future PRs.
Besides, this PR refactors the async scheduling loop to be easier to understand, and enables structured outputs support.

PR #23224, opened 2025-08-20
[PERF] Async structured outputs

vadiklyutiy:
## Purpose
Continue working on optimizing structured outputs started in #21862.
This PR implements asynchronous calculation of structured outputs to improve performance of structured output generation by moving heavy computation to a separate background process.

### Key Changes:
1. Multiprocess Architecture: Introduces a new 3-tier architecture:
- `StructuredOutputManager` (Main process): Coordinates operations via multiprocessing queues
- `StructuredOutputGateway` (Child process): Background process that receives and executes tasks
- `StructuredOutputExecutor` (Child process): Performs actual grammar compilation, bitmask generation, and token acceptance

2. Asynchronous Grammar Bitmask Calculations:
- Grammar bitmask generation now happens asynchronously in the background process
- Main process receives a `GrammarBitmaskPlaceholder` and can continue execution
- Actual bitmask is retrieved when needed via shared memory

3. Communication:
- Uses shared memory for large bitmask data transfer
- Three dedicated queues for different types of communication:
   - task_queue: Main → Child for task submission
   - batch_validate_result_queue: Child → Main for token validation results
   - grammar_init_notification_queue: Child → Main for grammar initialization notifications

4. Introduce Batch Processing for:
- submit_batch_accept_tokens(): Batches token acceptance operations
- submit_batch_validate_tokens(): Batches token validation for speculative decoding
to reduce amount of communications

## Test Plan
```
vllm serve Qwen/Qwen3-1.7B  --no-enable-prefix-caching
```
```
python3 benchmarks/benchmark_serving_structured_output.py --backend vllm --model Qwen/Qwen3-1.7B --structured-output-ratio 1.0 --request-rate 100 --num-prompts 2000 --json-schema-path ./test3.json  --output-len 2048
```
<details><summary> test3.json </summary>

```
{
    "description": "Schema for representing structured information about an animal species.",
    "properties": {
      "animal_name": {
        "description": "The most common English name for the species. Keep it under 50 characters. Avoid including subspecies or variation labels unless critical for identification. Do not invent or decorate the name.",
        "title": "Animal Name",
        "type": "string"
      },
      "short_summary": {
        "description": "One brief sentence (no more than 20 words) describing the animal’s main characteristic or behavior.",
        "title": "Short Summary",
        "type": "string"
      },
      "distinctive_trait": {
        "description": "The single feature that most clearly separates this species from close relatives (e.g., pattern, call, or body shape).",
        "title": "Distinctive Trait",
        "type": "string"
      },
[...]

PR #26866, opened 2025-10-15
[Core] Async scheduling + structured outputs compatibility

njhill:
Following similar approach to https://github.com/vllm-project/vllm/pull/23391.

Throughput benchmarks using the same json schema as https://github.com/vllm-project/vllm/pull/23224:
```
vllm serve Qwen/Qwen3-1.7B --uvicorn-log-level=error  --no-enable-prefix-caching

python3 benchmarks/benchmark_serving_structured_output.py --backend vllm --model Qwen/Qwen3-1.7B --structured-output-ratio $ratio --request-rate 200 --max-concurrency 800 --num-prompts 4000 --json-schema-path ./test3.json  --output-len 128
```

|Test|Executor / pct struct reqs ->|0.0|0.2|0.8|1.0|
|--|--|--|--|--|--|
|main|uniproc|103.16|92.57|70.68|69.36|
|This PR|uniproc|103.19|99.67|87.90|85.28|
|This PR + `--async-scheduling` |uniproc|132.72|106.08|93.59|90.34|
|This PR + `--async-scheduling` |multiproc|133.31|114.67|96.08|93.42|

This is a breaking change for the model runner and scheduler interfaces.


vllm/v1/engine/core.py
```diff
@@ -318,10 +318,15 @@ def step(self) -> tuple[dict[int, EngineCoreOutputs], bool]:
...
         scheduler_output = self.scheduler.schedule()
-
+        future = self.model_executor.execute_model(scheduler_output, non_block=True)
+        assert isinstance(future, Future)
+        grammar_output = self.scheduler.get_grammar_bitmask(scheduler_output)
         with self.log_error_detail(scheduler_output):
-            model_output = self.model_executor.execute_model(scheduler_output)
+            model_output = future.result()
+            if model_output is None:
+                model_output = self.model_executor.sample_tokens(grammar_output)  # type: ignore[assignment]
```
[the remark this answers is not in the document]
njhill: async scheduling always uses `step_with_batch_queue` with `max_concurrent_batches == 2`

vllm/v1/executor/multiproc_executor.py
```diff
@@ -183,12 +180,31 @@ def execute_model(  # type: ignore[override]
...
+    def sample_tokens(  # type: ignore[override]
+        self,
+        grammar_output: GrammarOutput | None,
+        non_block: bool = False,
     ) -> ModelRunnerOutput | Future[ModelRunnerOutput]:
+        return self.do_execute_model(  # type: ignore[return-value]
+            "sample_tokens", grammar_output, non_block=non_block
+        )
+
+    def do_execute_model(
```
WoosukKwon: Can we find a different name? `do_execute_model` is a bit confusing

njhill: renamed to `_execute_with_aggregation`

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2585,9 +2603,45 @@ def execute_model(
...
+            hidden_states,
+            sample_hidden_states,
+            aux_hidden_states,
+            kv_connector_output,
+        )
+        return None
+
+    @torch.inference_mode
+    def sample_tokens(
+        self, grammar_output: "GrammarOutput | None"
```
WoosukKwon, line 2623:
```suggestion
        return ExecuteModelState(
            scheduler_output,
            logits,
            spec_decode_metadata,
            spec_decode_common_attn_metadata,
            hidden_states,
            sample_hidden_states,
            aux_hidden_states,
            kv_connector_output,
        )

    @torch.inference_mode
    def sample_tokens(
        self, state: ExecuteModelState, grammar_output: "GrammarOutput | None"
```

njhill: Discussed offline that we'll keep this as-is. Have added some comments to the methods in `WorkerBase` explaining that the two-stage design may change in future.

WoosukKwon (approved):
LGTM! Thanks for the effort.

njhill:
> Hi @njhill, I found some performance drop for pipeline-parallism scenarios after your pr merged. Do you have some ideas about it, thanks in advance for your great support.
> 
> And below is the command to launch the server:
> 
> ```
> VLLM_USE_V1=1 VLLM_ALLOW_LONG_MAX_MODEL_LEN=1 python3 -m vllm.entrypoints.openai.api_server --model deepseek-ai/DeepSeek-R1-Distill-Qwen-32B --enforce-eager --port 8000 --host 0.0.0.0 -pp 2 --distributed_executor_backend=mp --trust-remote-code --gpu-memory-util=0.9 --no-enable-prefix-caching --max-num-batched-tokens=8192 --disable-log-requests --max-model-len=8192 --block-size 64 --quantization fp8    --dtype=float16   -tp=2
> ```
> 
> The command for send the request:
> 
> ```
> python3 -m vllm.entrypoints.cli.main bench serve  --model deepseek-ai/DeepSeek-R1-Distill-Qwen-32B --ready-check-timeout-sec 1 --dataset-name random --random-input-len=1024 --random-output-len=512 --ignore-eos --port=8000 --host 0.0.0.0 --num-prompt 30 --request-rate inf --backend vllm --trust-remote-code
> ```
> 
> And the perf drop from 617.26 tok/s to 384.40 tok/s.

Thanks @ys950902. Which commit exactly were you testing? There was some known perf regression from this PR which was subsequently fixed in https://github.com/vllm-project/vllm/pull/28012. Unfortunately, that PR was just reverted due to a compatibility bug, but the re-apply of it https://github.com/vllm-project/vllm/pull/28319 should be merged to main soon.

It would be great if you could check whether the degraded performance still shows up when that PR is included (if it wasn't already in your test). If so could you open a new issue with the above detail and we can investigate further.

vllm/v1/engine/core.py
```diff
@@ -376,20 +379,47 @@ def step_with_batch_queue(
...
+                # Block-wait for execute to return (continues running async on the GPU).
+                with self.log_error_detail(scheduler_output):
+                    exec_result = exec_future.result()
+                    assert exec_result is None
+            else:
+                # We aren't waiting for any tokens, get any grammar output immediately.
+                grammar_output = self.scheduler.get_grammar_bitmask(scheduler_output)
+                # Block-wait for execute to return (continues running async on the GPU).
+                with self.log_error_detail(scheduler_output):
+                    exec_result = exec_future.result()
```
weireweire, line 403: why do we blocking before batch queue is full? won't this break the batch queue behavior?

weireweire: Could you have a look? PP mode will block here, none parallel will happen. Even though here is intent to wait for the model_execute, but the previous sample_tokens task should also in the queue.

weireweire: @ys950902 is your PP perf issue solved? is it also related to the blocking here?

nvpohanh: @njhill Could you help answer this question? Thanks!

weireweire: draft fix: https://github.com/vllm-project/vllm/pull/28286

PR #26872, opened 2025-10-15
[Attention][DSA] Abstrat DSA topk-indices-buffer to AttentionBackend

MengqingCao:
## Purpose
Abstrat DSA topk-indices-buffer to `AttentionBackend`, seperated from #26656. Some attention backends do not need to create this buffer, such as the attention backend in vLLM-Ascend. This seems to be a logic closely related to the attention backend, and abstracting it into `AttentionBackend` will benefit a wide range of backend implementations
## Test Plan
Test with DeepSeek-V3.2-Exp

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/models/deepseek_v2.py
+++ b/vllm/model_executor/models/deepseek_v2.py
@@ -75,7 +75,7 @@ from vllm.model_executor.model_loader.weight_utils import (
 from vllm.model_executor.models.utils import sequence_parallel_chunk
 from vllm.platforms import current_platform
 from vllm.sequence import IntermediateTensors
-from vllm.utils import cdiv, direct_register_custom_op
+from vllm.utils import cdiv, direct_register_custom_op, resolve_obj_by_qualname
 from vllm.utils.deep_gemm import fp8_mqa_logits, fp8_paged_mqa_logits
 from vllm.v1.attention.backends.mla.indexer import (
     DeepseekV32IndexerBackend,
@@ -1168,7 +1168,7 @@ class DeepseekV2Model(nn.Module):
 
         self.vocab_size = config.vocab_size
         self.is_v32 = hasattr(config, "index_topk")
-        if self.is_v32:
+        if self.is_v32 and enable_dsa_topk_indices_buffer():
             topk_tokens = config.index_topk
             topk_indices_buffer = torch.empty(
                 vllm_config.scheduler_config.max_num_batched_tokens,
@@ -1508,3 +1508,19 @@ def get_spec_layer_idx_from_weight_name(
             if weight_name.startswith(f"model.layers.{layer_idx + i}."):
                 return layer_idx + i
     return None
+
+
+def enable_dsa_topk_indices_buffer():
+    return resolve_obj_by_qualname(
+        current_platform.get_attn_backend_cls(
+            selected_backend=None,
+            head_size=None,
+            dtype=None,
+            kv_cache_dtype=None,
+            block_size=None,
+            use_v1=True,
+            use_mla=True,
+            has_sink=False,
+            use_sparse=True,
+        )
+    ).enable_dsa_topk_indices_buffer()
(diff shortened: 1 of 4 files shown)


vllm/model_executor/models/deepseek_v2.py
```diff
@@ -1168,7 +1168,7 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = ""):
 
         self.vocab_size = config.vocab_size
         self.is_v32 = hasattr(config, "index_topk")
-        if self.is_v32:
+        if self.is_v32 and enable_dsa_topk_indices_buffer():
```
LucasWilkinson, line 1171: if we do this we also need to gate constructing the `Indexer` since that would be useless without the `topk_indices_buffer` buffer

vllm/model_executor/models/deepseek_v2.py
```diff
@@ -1508,3 +1508,19 @@ def get_spec_layer_idx_from_weight_name(
...
+            head_size=None,
+            dtype=None,
+            kv_cache_dtype=None,
+            block_size=None,
+            use_v1=True,
+            use_mla=True,
+            has_sink=False,
+            use_sparse=True,
+        )
+    ).enable_dsa_topk_indices_buffer()
```
LucasWilkinson, line 1526: This is quite messy; I don't think we should resolve the backend multiple times in different spots with potentially different parameters (this is ripe for bugs)

I see two potential alternatives:

1) add a CLI argument indicating you want to run with dense attention (and we will omit this and the indexer)
2) move `topk_indices_buffer` into `vllm.attention.layer.MLAAttention` that way we can resolve the backend first; see it it supports sparse or not then choose to allocate the buffer, we would then need to move the `Indexer` creation to after MLA so we don't create it if `vllm.attention.layer.MLAAttention` doesn't expose a `topk_indices_buffer`

cc @heheda12345

MengqingCao: Agree, actually I also realize that this pr is not very reasonable, and perfer to move it into attention kvcache spec, and refactor the magic number 656 of MLAAttentionSpec.
but maybe next week I could work on this

heheda12345: @MengqingCao why ascend doesn't need this buffer?
@LucasWilkinson do you think it is a good idea to merge DeepseekV32IndexerMetadata and FlashMLASparseMetadata (as well as their backend builder) into one metadata (builder) and use it for both two components? The other backend can just return an empty metadata.
With this approach, the topk indices buffer becomes an internal buffer of sparse attention and we can run the common building logic only once (not sure how much common logics now)

MengqingCao: Sorry for the delay, we've allocated another buffer in vllm-ascend to hold the keys passed to the indexer.

heheda12345: Can you explain more? Personally I don't like the vllm's current approach and want to get some better ideas.

MengqingCao: On the point of kvcache, we actually allocate `FullAttentionSpec` for deepseek-v3.2, and use the half of it for indexer k cache. This is not good as there is still some cache wasted in current approach, thus we also hope the kvcachespec of deepseek v3.2 could be more general. This maybe should make `MLAAttentionSpec` aware of the k cache size of indexer of the runtime platform.

On the point of attention metadata, we make an unified `AscendSFAMetadata` (SFA refers to sparse flash attention), but I must point out that it dosen't implement the chunk of prefill metadata, and it more likely a normal `MLAMetadata.` When caching the k cache in index selector,  we make a small modify to the slot mapping to make sure of the accuracy
https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/attention/sfa_v1.py#L530

PR #20874, opened 2025-07-13
[Frontend] OpenAI Responses API supports Tool/Function calling

chaunceyjiang:
OpenAI Responses API supports Tool/Function calling

Follow up https://github.com/vllm-project/vllm/pull/20504

PR #20504, opened 2025-07-04
Implement OpenAI Responses API [1/N]

WoosukKwon:
This PR implements the basic support for the OpenAI Responses API. 

- [x] Basic stateless functionalities with stream=False
- [x] Reasoning parser support
- [x] Structured outputs on normal outputs (not on reasoning part)
- [x] Stateful APIs (with just a python dictionary)

TODOs:
- [ ] Tool/functional calling support
- [ ] Output token streaming
- [ ] Structural tag in reasoning
- [ ] Multimodal support

PR #26874, opened 2025-10-15
[Frontend] OpenAI Responses API supports Tool/Function calling - non-harmony 

chaunceyjiang:
## Purpose
rebase https://github.com/vllm-project/vllm/pull/20874

OpenAI Responses API supports Tool/Function calling

Follow up https://github.com/vllm-project/vllm/pull/20504

## Test Result

---
<details>

</details>


yeqcharlotte:
i think we are quite aligned that we don't want responses api to only work for gpt-oss https://github.com/vllm-project/vllm/issues/26703 ;)

could you share some e2e vllm serve command how you use this without other models?

vllm/entrypoints/openai/serving_engine.py
```diff
@@ -1099,13 +1099,17 @@ async def _preprocess_chat(
...
+                msg = (
+                    "Tool usage is only supported for Chat Completions API "
+                    "or Responses API requests."
+                )
                 raise NotImplementedError(msg)
 
-            request = tool_parser(tokenizer).adjust_request(  # type: ignore
-                request=request
-            )
+            if isinstance(request, ChatCompletionRequest):
```
qandrew: question: what do we if the request is ResponsesRequest? do we need to do some adjust_request stuff too?

tests/v1/entrypoints/openai/serving_responses/test_function_call.py
```diff
@@ -0,0 +1,187 @@
...
+
+    assert len(response.output) >= 1
+    tool_call = response.output[0]
+
+    assert tool_call.type == "function_call"
+    assert json.loads(tool_call.arguments) is not None
+
+
+@pytest.mark.asyncio
+async def test_named_tool_use(client: openai.AsyncOpenAI):
```
qandrew, line 137: nit: I think ideally the test name should correspond with the original file. I think this file should be here?
`tests/entrypoints/openai/serving_responses/test_function_call` 
we could create multiple files under serving_responses if that file has a lot of different tests.

vllm/entrypoints/openai/protocol.py
```diff
@@ -294,6 +296,91 @@ def get_logits_processors(
     return None
 
 
+def get_json_schema_from_tool(
```
qandrew: nit: if you're refactoring code, maybe this should go in utils.py instead?

vllm/entrypoints/openai/serving_responses.py
```diff
@@ -785,15 +798,120 @@ def _make_response_output_items(
...
+                        status="completed",
+                        name=tool_call.name,
+                        arguments=tool_call.arguments,
+                    )
+                    for tool_call in function_calls
+                ]
+            )
+        else:
+            if reasoning_item:
+                outputs.append(reasoning_item)
```
qandrew: question: why are we not outputting reasoning_item / message_item if there is a function call? in harmony mode we still output this, we should be consistent?

vllm/entrypoints/openai/serving_engine.py
chaunceyjiang: Since `ResponsesRequest` does not have a `skip_special_tokens` attribute, I'm temporarily skipping `adjust_request` for `ResponsesRequest` here.  
I'll fix this issue in the next PR. @qandrew

vllm/entrypoints/openai/protocol.py
chaunceyjiang: I tried, but this function can only stay in the `protocol.py` file—if I move it to another file like `utils.py`, it causes a circular import issue.  

The reason is that the type annotations used in `get_json_schema_from_tool`, such as `ChatCompletionNamedToolChoiceParam`, are defined in this `protocol.py` file.
@qandrew

tests/v1/entrypoints/openai/serving_responses/test_function_call.py
chaunceyjiang: good idea.

Currently, all tests in this file are placed under the `tests/v1/entrypoints/openai/responses` directory to reuse the already running vLLM instance. If we move them to a new directory, a new vLLM instance would need to be started.  

How about renaming the current directory from `tests/v1/entrypoints/openai/responses` to `tests/v1/entrypoints/openai/serving_responses`?

vllm/entrypoints/openai/serving_responses.py
chaunceyjiang: Fixed.

Added a test for it.

vllm/entrypoints/openai/protocol.py
```diff
@@ -294,6 +296,91 @@ def get_logits_processors(
...
+            for tool in tools
+            if isinstance(tool, ChatCompletionToolsParam)
+        }
+        if tool_name not in tool_map:
+            raise ValueError(f"Tool '{tool_name}' has not been passed in `tools`.")
+        return tool_map[tool_name].function.parameters
+
+    if tool_choice == "required":
+
+        def extract_tool_info(
```
alecsolder: I think this should be turned into a standalone method, I think it will be useful outside of the context of this method and should also be able to be tested by itself. Same with the rest of these methods. 

We will need to figure out how to tie it into the tool definitions from MCP servers like we do when converting from MCP -> harmony tool definitions as well

vllm/entrypoints/openai/serving_responses.py
```diff
@@ -785,15 +801,123 @@ def _make_response_output_items(
...
+                        name=tool_call.name,
+                        arguments=json.dumps(tool_call.parameters, ensure_ascii=False),
+                    )
+                    for tool_call in tool_calls
+                ]
+            )
+            content = None  # Clear content since tool is called.
+        elif request.tool_choice == "auto" or request.tool_choice == "none":
+            try:
+                tool_parser = self.tool_parser(tokenizer)
```
alecsolder: Should we be doing this for every request, or can we just do it once at server startup time?

vllm/entrypoints/openai/protocol.py
alecsolder: Does it make sense to split this into two methods? One for going from tools to dictionaries, and then one for filtering based on the tool_choice param 

It may be useful to standardize around MCP tool definitions vs just a json dict for this protocol.py file, like we convert from FunctionTool or ChatCompletionToolsParam to the MCP tool definition, and then from MCP tool definitions to json schemas for structured output purposes.

This would help us migrate more logic out of this file as well, since we would no longer be relying on protocol.py types, and we will need to add support for MCP tools anyway.

alecsolder (commented):
I think this is great, it really gets us started with supporting other models on responses API

Comparing this to gpt-oss, for gpt-oss we handle the conversion of the output tokens to "messages" in entrypoints/context.py. The Harmony library happens to be a tool_parser reasoning_parser and tokenizer all at the same time, but it is nice that it all happens in one location. This matches needing to convert from responses types to completions types so we can use the parsers for other models.

I think the implementation in this PR as-is works, but once we want to support server side tool calling with models besides gpt-oss, we will likely need to move the parsing logic to entrypoints/context.py as well so it can happen in the tool calling "loop" from _generate_with_builtin_tools() in serving_engine.

Ideally we'd be able to start pulling conversion logic out of serving_responses into their own files like with harmony_utils.py, and continue to standardize more around something like creating the right context object for serving_engine.

vllm/entrypoints/openai/serving_responses.py
chaunceyjiang: > Should we be doing this for every request,

yes.

chaunceyjiang:
> I think the implementation in this PR as-is works, but once we want to support server side tool calling with models besides gpt-oss, we will likely need to move the parsing logic to entrypoints/context.py as well so it can happen in the tool calling "loop" from _generate_with_builtin_tools() in serving_engine.

> Ideally we'd be able to start pulling conversion logic out of serving_responses into their own files like with harmony_utils.py, and continue to standardize more around something like creating the right context object for serving_engine.

Hi, @alecsolder 

Your suggestions are excellent, and I completely agree. However, the main goal of this PR is to enable non-gpt-oss models to use tool calling with the Responses API.

Essentially, this PR constructs chat-completion-style messages with tool calling, and then extracts tools from the output using the `tool_parser`.

You can think of it as similar to how the Responses API currently handles Harmony.

The new `_construct_chat_message_with_tool_call` function is analogous to the existing `_construct_input_messages_with_harmony` function — both are responsible for constructing input messages.

Similarly, the new `_extract_tool_calls` function is analogous to the existing `parse_output_message` function — both are responsible for parsing output messages.

Regarding what you mentioned about **server-side tool calling** and **MCP**, I believe that should be handled by `SimpleContext`. Therefore, for non-gpt-oss models to support MCP, we’ll likely need a new `Context`. I plan to submit a separate PR to introduce that new `Context`.

However, I still think the `_construct_chat_message_with_tool_call` and `_extract_tool_calls` functions are necessary for constructing input messages and parsing output messages. These functions do not conflict with the `Context`, since currently the `Context` is not responsible for message construction or parsing.

vllm/entrypoints/openai/protocol.py
chaunceyjiang: > Does it make sense to split this into two methods? One for going from tools to dictionaries, and then one for filtering based on the tool_choice param

Done.

>  MCP tool definitions


Can you remind me what `MCP tool definitions` are? Is it this one `from openai_harmony import ToolDescription`?

vllm/entrypoints/openai/protocol.py
chaunceyjiang: > I think this should be turned into a standalone method, I think it will be useful outside of the context of this method and should also be able to be tested by itself. Same with the rest of these methods.

Done.

> We will need to figure out how to tie it into the tool definitions from MCP servers like we do when converting from MCP -> harmony tool definitions as well


I agree with what you said, but at the moment I'm not sure whether non-gpt-oss models can support MCP, or how to support it. Should this be done in a separate PR, and then, if needed, refactor this part in that other PR?

What do you think?

vllm/entrypoints/openai/protocol.py
chaunceyjiang: If by “MCP tool definitions” you mean
`from openai.types.responses.tool import Mcp`,
then I still believe this should be refactored when implementing MCP for non-gpt-OSS models in the Responses API — not now. @alecsolder

alecsolder: I had meant https://github.com/vllm-project/vllm/blob/141d3b9fc51442743ac10f4d3e042af0853c1215/vllm/entrypoints/tool_server.py#L15 

Using the types from the actual MCP library from Anthropic, which the tools are defined like this https://github.com/modelcontextprotocol/python-sdk/blob/1200ba0082e05daf81ff77f36d24afb5b1b2a424/src/mcp/types.py#L871

Because that is what we use to map to the harmony ToolDescription.

alecsolder (approved):
Hey @chaunceyjiang, Completely agree with everything in your comment, I am fully aligned on moving things to the context classes, and all of it definitely doesn't need to be in this PR. I was mostly just trying to say these things to make sure we are aligned, which we are! 

Other models can definitely support MCP, and the implementation from here is basically just us deciding when to handle the parsed FunctionCall from within vLLM, vs when to return it to the client. So we are right on track!

I'll try to put together the list of things that still need to be done to support the same arbitrary MCP integration as here #26704 for other models :)

vllm/entrypoints/openai/protocol.py
yeqcharlotte: i think most of the tool parsing logic shouldn't belong to protocol.py. how much can we push to tool parser implementation. cc: @aarnphm

chaunceyjiang: > i think most of the tool parsing logic shouldn't belong to protocol.py. how much can we push to tool parser implementation. cc: @aarnphm

goo idea, I’ll submit a new PR to handle this. @yeqcharlotte

chaunceyjiang: > i think most of the tool parsing logic shouldn't belong to protocol.py. how much can we push to tool parser implementation. cc: @aarnphm

Done.

tests/v1/entrypoints/openai/serving_responses/test_function_call.py
qandrew: @chaunceyjiang i don't see any files in tests/v1/entrypoints/openai/responses; they're all just in tests/v1/entrypoints/openai. 

I'm supportive of moving files into tests/v1/entrypoints/openai/serving_responses

chaunceyjiang:
@yeqcharlotte PTAL.

vllm/entrypoints/openai/serving_responses.py
```diff
@@ -806,15 +822,122 @@ def _make_response_output_items(
...
                 content=[output_text],
                 role="assistant",
                 status="completed",
                 type="message",
             )
-            output.append(message)
-        return output
+        outputs = []
+
+        if reasoning_item:
```
yeqcharlotte, line 852: i think the logic here is quite heavy to put inside serving_responses.py. any way we can clearly identify the shared components between chat completion and responses. 

cc: @aarnphm

vllm/entrypoints/openai/serving_responses.py
```diff
@@ -806,15 +822,122 @@ def _make_response_output_items(
...
+                    status="completed",
+                    name=tool_call.name,
+                    arguments=tool_call.arguments,
+                )
+                for tool_call in function_calls
+            ]
+            outputs.extend(tool_call_items)
+        return outputs
+
+    def _parse_tool_calls_from_content(
```
yeqcharlotte: most logic here doesn't seem to require access to `OpenAIServingResponses`. wonder if we can extract those logic to utils? we can have separate utils for chat completion and responses. and see if we can share most logic.

vllm/entrypoints/openai/serving_responses.py
chaunceyjiang: There’s no logic here at all — it’s just initializing the classes related to Response.

It’s completely different from Chat Completion, isn’t it?
@yeqcharlotte

chaunceyjiang:
@yeqcharlotte Ready for review.

yeqcharlotte (approved):
thanks for the change. let's keep iterate on this to also support minimax-m2 cc: @qandrew

PR #26878, opened 2025-10-15
[BugFix] Patch inductor memory plan logic

BoyuanFeng:
This PR fixes a test failure for `test_multi_graph_piecewise_compile`. This is a pytorch issue and is fixed in https://github.com/pytorch/pytorch/pull/165514. This PR monkey patches vLLM for PyTorch-2.9.0 only to unblock.


vllm/env_override.py
```diff
@@ -21,3 +22,75 @@
...
+            self.lines[i] = line.plan(planning_states[-1])
+        elif isinstance(line, EnterSubgraphLine):
+            planning_states.append(MemoryPlanningState())
+        elif isinstance(line, ExitSubgraphLine):
+            past_planning_states.append(planning_states.pop())
+    past_planning_states.append(planning_states.pop())
+    assert len(planning_states) == 0
+
+
+if _is_torch_equal("2.9.0"):
```
ProExpertProg: Should we use `is_torch_equal` instead (no leading underscore)?

ProExpertProg (approved):

vllm/env_override.py
BoyuanFeng: yes!

PR #26883, opened 2025-10-15
Support block size of 256 used by Intel HPU

mandy-li:
This PR adds block size of 256 to the list which is used by Intel HPU fp8 models


hmellor (approved):

yewentao256 (approved):
Thanks for the work! Please also merge from main to fix some of the CI issue

PR #26915, opened 2025-10-15
[Feature] publisher default set zmq in kv_event config

lengrongfu:
## Purpose

FIX: https://github.com/vllm-project/vllm/issues/26772

## Test Result

---
<details>

</details>


vllm/config/kv_events.py
```diff
@@ -17,7 +17,7 @@ class KVEventsConfig:
     Events can be published externally by zmq using the event publisher config.
     """
 
-    publisher: str = "null"
+    publisher: str = "zmq"
```
hmellor: Either here or `KVEventsPublisher = Literal["null", "zmq"]` on line ~9

```suggestion
    publisher: Literal["null", "zmq"] = "zmq"
```

vllm/distributed/kv_events.py
```diff
@@ -353,12 +353,12 @@ def create(
         cls, config: KVEventsConfig | None, data_parallel_rank: int = 0
     ) -> EventPublisher:
         """Create publisher from a config mapping."""
-        if not config:
+        if not config or not config.enable_kv_cache_events:
```
hmellor:
```suggestion
        if not (config and config.enable_kv_cache_events):
```

lengrongfu:
> Thinking about it, what is the purpose of the `null` value?

I think main when not enable kv_cache, can keep code in scheduler.py, but it not anthing action:
```
        # publish collected KV cache events
        if events:
            batch = KVEventBatch(ts=time.time(), events=events)
            self.kv_event_publisher.publish(batch)
```

hmellor:
Should we leave the default as `"null"` and add a `pydantic.model_validator` like:

```python
@model_validator
def _validate_kv_events_config(self):
    if self.enable_kv_cache_events and self.publisher == "null":
        raise ...
```

vllm/config/kv_events.py
```diff
@@ -17,7 +19,7 @@ class KVEventsConfig:
     Events can be published externally by zmq using the event publisher config.
     """
 
-    publisher: str = "null"
+    publisher: Literal["null", "zmq"] = "zmq"
```
hmellor:
```python
from pydantic import Field
```

```suggestion
    publisher: Literal["null", "zmq"] = Field(default=None)
```

Then in `__post_init__`:

```python
if self.publisher is None and self.enable_kv_cache_events:
    self.publisher = "zmq"
else:
    self.publisher = "null"
```

hmellor (commented):
In that case, should we do this suggestion?

vllm/config/kv_events.py
lengrongfu: This code has the same effect as the current pr.
- new code exec login, we define "1" name

|publisher/enable_kv_cache_events | None | "null" | "zmq" |
|----------|------|--------|-------|
|False | use null | use null | use null |
|True  | use zmq  | use null | use zmq|

- current pr exec login, we define "2" name

|publisher/enable_kv_cache_events | "null" | "zmq" |
|-----|--------|-------|
|False | use null | use null |
|True  | use null | use zmq|

e.g:
- `--kv-events-config.enable_kv_cache_events true`, exec "1" login after, use zmq publisher, exec "2" login after, then use zmq publisher.
- `--kv-events-config.enable_kv_cache_events true --kv-events-config.publisher null`, exec "1" login after, use null publisher, exec "2" login after, then use null publisher.
- `--kv-events-config.enable_kv_cache_events false --kv-events-config.publisher null`, exec "1" login after, use null publisher, exec "2" login after, then use null publisher.
- `--kv-events-config.enable_kv_cache_events true --kv-events-config.publisher zmq`,exec "1" login after, use zmq publisher, exec "2" login after, then use zmq publisher.

So their execution results are the same.

hmellor: The ultimate behaviour is the same. But in the current pr the actual value of `publisher` is always `zmq` unless the user manually changes it.

IMO it makes more sense to have the value of the config represent what will actually happen.

lengrongfu: make sense.

vllm/distributed/kv_events.py
```diff
@@ -353,12 +353,12 @@ def create(
         cls, config: KVEventsConfig | None, data_parallel_rank: int = 0
     ) -> EventPublisher:
         """Create publisher from a config mapping."""
-        if not config:
+        if not (config and config.enable_kv_cache_events):
```
hmellor: ?

```suggestion
        if config is None or config.publisher == "null":
```

lengrongfu: Oh, i mis this code.

hmellor (approved):

vllm/config/kv_events.py
```diff
@@ -47,3 +50,9 @@ class KVEventsConfig:
     """The topic to use for the event publisher. Consumers can subscribe to
     this topic to receive events.
     """
+
+    def __post_init__(self):
+        if self.publisher is None and self.enable_kv_cache_events:
+            self.publisher = "zmq"
+        elif self.publisher is None:
+            self.publisher = "null"
```
hmellor:
```suggestion
        if self.publisher is None:
            self.publisher = "zmq" if self.enable_kv_cache_events else "null"
```

PR #26917, opened 2025-10-15
Lower severity of log when model info cache misses due to exception

hmellor:


mgoin (approved):
Thanks, appreciate it

ProExpertProg (approved):

PR #24604, opened 2025-09-10
[torch.compile] Enable attention and allreduce fusion without custom ops enabled

ProExpertProg:
## Purpose

This PR enables matching the torch implementations of custom ops QuantFP8 and RMSNorm. On `main`, fusion currently requires enabling custom ops, but they are slower than their torch counterparts, so the benefit of custom fusion passes is reduced.

We add a bunch of "matcher util" objects which can be called in patterns and get traced to the same fx nodes as the custom op they correspond to in both enabled and disabled form automatically.

This PR also adds additional debugging utilities and adds E2E fusion tests to verify fusions happen in models end-to-end instead of just in unit tests.

## Test Plan

Unit tests, added more fusion E2E tests.

## Test Result
Tests all pass

## Performance numbers

Below are B200 numbers (with flashinfer) from `vllm bench serve` on the following serve command:

```
vllm serve redhatai/meta-llama-3.1-70B-Instruct-FP8 --no-enable-prefix-caching --load-format dummy --kv-cache-dtype=fp8 -O.splitting_ops=[] -O.cudagraph_mode=FULL_DECODE_ONLY
```

We test the following regimes with corresponding additional arguments:
1. `none`: `-O.custom_ops='["none"]' -O.pass_config={"enable_fi_allreduce_fusion":false,"enable_attn_fusion":false,"enable_noop":true}`
1. `none_fusion_attention`: `-O.custom_ops='["none"]' -O.pass_config={"enable_fi_allreduce_fusion":false,"enable_attn_fusion":true,"enable_noop":true}`
1. `none_fusion_attention_allreduce`: `-O.custom_ops='["none"]' -O.pass_config={"enable_fi_allreduce_fusion":true,"enable_attn_fusion":true,"enable_noop":true}`
1. `rms_quant`: `-O.custom_ops='["none", "+quant_fp8", "+rms_norm"]' -O.pass_config={"enable_fi_allreduce_fusion":false,"enable_attn_fusion":false,"enable_noop":true}`
1. `rms_quant_fusion_attention`: `-O.custom_ops='["none", "+quant_fp8", "+rms_norm"]' -O.pass_config={"enable_fi_allreduce_fusion":false,"enable_attn_fusion":true,"enable_noop":true}`
1. `rms_quant_fusion_attention_allreduce`: `-O.custom_ops='["none", "+quant_fp8", "+rms_norm"]' -O.pass_config={"enable_fi_allreduce_fusion":true,"enable_attn_fusion":true,"enable_noop":true}`

[...]

PR #26933, opened 2025-10-15
[compile] Enable sequence parallelism matching w/o custom ops enabled

angelayi:
Superseeded by https://github.com/vllm-project/vllm/pull/27126

## Purpose

Based on https://github.com/vllm-project/vllm/pull/24604, modified sequence-parallelism pass to do custom op matching w/o needing to enable the custom op

## Test Plan
`pytest -sv tests/compile/test_sequence_parallelism.py -k TestQuantModel`


vllm/utils/flashinfer.py
```diff
@@ -357,7 +357,8 @@ def bmm_fp8(
     ) -> torch.Tensor:
         from flashinfer import bmm_fp8 as bmm_fp8_
 
-        return bmm_fp8_(A, B, A_scale, B_scale, dtype, None, backend)
+        result = bmm_fp8_(A, B, A_scale, B_scale, dtype, None, backend)
+        return result.reshape(A.shape[1], B.shape[2])
```
angelayi, line 361: I moved the reshape into the kernel because it was causing a lot of trouble in the pattern matching wrt dynamic shapes.. hope this is ok

PR #26935, opened 2025-10-15
[Feature] Migrate DeepGEMM API from `get_m_alignment_for_contiguous_layout` to `get_mk_alignment_for_contiguous_layout`

yewentao256:
## Purpose

`get_m_alignment_for_contiguous_layout` is deprecated, we should migrate to newer API

## Test

`lm_eval   --model vllm   --model_args "pretrained=Qwen/Qwen3-30B-A3B-FP8,max_model_len=32768,enforce_eager=True"   --trust_remote_code   --tasks gsm8k   --num_fewshot 5   --batch_size auto`

```bash
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8575|±  |0.0096|
|     |       |strict-match    |     5|exact_match|↑  |0.8969|±  |0.0084|
```

### DIFF
--- a/tests/kernels/moe/test_block_fp8.py
+++ b/tests/kernels/moe/test_block_fp8.py
@@ -22,12 +22,12 @@ from vllm.model_executor.layers.fused_moe.fused_moe import (
 )
 from vllm.platforms import current_platform
 from vllm.utils import has_deep_gemm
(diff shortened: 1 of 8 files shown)


mgoin (approved):
LGTM, although I think `get_mk_alignment_for_contiguous_layout` is not as clear as `deep_gemm_block_shape` in code

PR #26937, opened 2025-10-15
[Perf] Enable full CUDA graphs for spec decoding with FlashInfer

### LINKED ISSUE #26856: [Performance]: FalshInfer attn backend. Use dynamic AttentionCGSupport
### Proposal to improve performance

Right now FlashInfer attn backend uses `AttentionCGSupport.UNIFORM_SINGLE_TOKEN_DECODE` that doesn't allow to use FULL cudagraph in case if speculative decoding. 

In fact FlashInfer contains several attn those chose dynamically. TRT_LLM Gen backend (one of the fastest) supports speculative decoding (`AttentionCGSupport.UNIFORM_BATCH`) and it does make sense to enable cudagraph for decode phase.

benchislett:
## Purpose

TRTLLM-gen kernels support full cuda graphs, but are only used with FlashInfer on Blackwell under certain conditions.
It might not be safe to change FlashInfer's `cudagraph_support` to `UNIFORM_BATCH` always, but we can still set it when we know TRTLLM-gen backend will be used.

Also update the docs to reflect the FlashInfer and FlashInferMLA cuda graph compatibility

FIX https://github.com/vllm-project/vllm/issues/26856

## Test Plan

Ran Llama 3.1 8B-Instruct with EAGLE3 and confirmed that lm_eval-gsm8k is unchanged compared to main, and when TRTLLM attention is force disabled. Confirmed via torch profile that full graphs are now issued for verification when TRTLLM attention is enabled

#### TRTLLM on:

```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.7726|±  |0.0115|
|     |       |strict-match    |     5|exact_match|↑  |0.7013|±  |0.0126|
```

#### TRTLLM off:

```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.7726|±  |0.0115|
|     |       |strict-match    |     5|exact_match|↑  |0.7013|±  |0.0126|
```

## Benchmarks

MT-Bench at concurrency 1 sees a minimal speedup (~2%)

```bash
vllm serve meta-llama/Llama-3.1-8B-Instruct --speculative-config '{"method": "eagle3", "model": "yuhuili/EAGLE3-LLaMA3.1-Instruct-8B", "num_speculative_tokens": 3}' &

vllm bench serve --dataset-name hf --dataset-path philschmid/mt-bench --num-prompts 80 --max-concurrency 1 --model meta-llama/Llama-3.1-8B-Instruct --base-url http://0.0.0.0:8049
```

#### Before:

```
Successful requests:                     80        
Maximum request concurrency:             1         
Benchmark duration (s):                  42.58     
Total input tokens:                      8133      
Total generated tokens:                  16955     
Request throughput (req/s):              1.88      
Output token throughput (tok/s):         398.23    
Peak output token throughput (tok/s):    186.00    
Peak concurrent requests:                4.00      
Total Token throughput (tok/s):          589.25    
Mean TTFT (ms):                          12.11     
Median TTFT (ms):                        11.88     
P99 TTFT (ms):                           14.29     
Mean TPOT (ms):                          2.45      
Median TPOT (ms):                        2.45      
P99 TPOT (ms):                           3.33      
Mean ITL (ms):                           5.36      
Median ITL (ms):                         5.36      
P99 ITL (ms):                            5.61      
```

#### After:

```
Successful requests:                     80        
Maximum request concurrency:             1         
Benchmark duration (s):                  41.73     
Total input tokens:                      8133      
Total generated tokens:                  16795     
Request throughput (req/s):              1.92      
Output token throughput (tok/s):         402.47    
Peak output token throughput (tok/s):    190.00    
Peak concurrent requests:                4.00      
Total Token throughput (tok/s):          597.37    
Mean TTFT (ms):                          11.86     
Median TTFT (ms):                        11.75     
P99 TTFT (ms):                           14.93     
Mean TPOT (ms):                          2.43      
Median TPOT (ms):                        2.37      
P99 TPOT (ms):                           3.39      
Mean ITL (ms):                           5.26      
Median ITL (ms):                         5.25      
P99 ITL (ms):                            5.48      
```


vllm/v1/attention/backends/flashinfer.py
```diff
@@ -354,7 +355,10 @@
         else:
             self.q_data_type = self.model_config.dtype
 
+        # If using trtllm attention, we can support uniform_batch speculative decoding
         self._init_reorder_batch_threshold(1, supports_spec_as_decode=can_use_trtllm)
+        if can_use_trtllm:
+            self.cudagraph_support = AttentionCGSupport.UNIFORM_BATCH
```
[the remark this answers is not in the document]
mgoin: Agreed, it seems we need to lock into using trtllm attention once we've made this choice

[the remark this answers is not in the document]
benchislett: I'll look into refactoring to make this a hard commitment. For now, I think it's mostly safe, since having speculative decoding enabled will force trtllm to be used (there's a condition inside of `use_trtllm_attention`). This behaviour is already implicitly depended on since we set `reorder_batch_threshold` based on this condition.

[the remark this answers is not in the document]
benchislett: Updated to make this dependency explicit.

vllm/v1/attention/backends/cpu_attn.py
```diff
@@ -348,7 +348,7 @@ def get_seq_len_block_table_args(
 
 
 class TorchSDPAMetadataBuilderV1(AttentionMetadataBuilder[TorchSDPAMetadata]):
-    reorder_batch_threshold: int = 1
+    reorder_batch_threshold: int
```
LucasWilkinson, line 351: nit: I still think this needs to be set?

vllm/v1/attention/backends/xformers.py
```diff
@@ -207,7 +207,7 @@ def decode_metadata(self) -> Optional["XFormersAttentionMetadata"]:
 class XFormersAttentionMetadataBuilder(
     AttentionMetadataBuilder[XFormersAttentionMetadata]
 ):
-    reorder_batch_threshold: int = 1
+    reorder_batch_threshold: int
```
LucasWilkinson, line 210: nit: is this still needed?

vllm/v1/attention/backends/mla/indexer.py
```diff
@@ -213,11 +212,7 @@ def split_prefill_chunks(
 
 
 class DeepseekV32IndexerMetadataBuilder(AttentionMetadataBuilder):
-    cudagraph_support: ClassVar[AttentionCGSupport] = (
-        AttentionCGSupport.UNIFORM_SINGLE_TOKEN_DECODE
-    )
-
-    reorder_batch_threshold: int = 1
+    reorder_batch_threshold: int
```
LucasWilkinson, line 215: nit: is this still needed?

LucasWilkinson (commented):
LGTM; there are some nits that should be addressed (specifically for the CPU backend I think we should still keep the reorder_batch_threshold = 1)

it is a bit harder to see where `cudagraph_support` is set now :/ I guess the alternative would be use a function; i.e. add a `get_cudagraph_support()` function in the base class (I think the current implementation is better but im also flip-flopping haha)

vllm/v1/attention/backends/cpu_attn.py
benchislett: It is set in the constructor: `_init_reorder_batch_threshold(1, False)`

The type annotation is left to indicate that it will never be "None" on this class and its subclasses. This is a common pattern in the changes in this PR

vllm/v1/attention/backends/mla/indexer.py
benchislett: yes, as type annotation, see previous comment

vllm/v1/attention/backends/xformers.py
benchislett: yes, as type annotation, see previous comment

fhl2000:
Hi @benchislett, it is no longer safe to dynamically update `cudagraph_support` inside `__init__()` since #27427 merged, because now we resolve cudagraph mode before actually initializing the builder instance.  So making it an instance property is not a good idea.  Instead, I would make a class method `get_cudagraph_support()` for this.

fhl2000:
The right logic should be "determine cudagraph_support of each backend(builder class)"-> "resolve cudagraph mode" -> "initial cudagraph-relative stuff of each backend"  anyway.  I think since the cudagraph_support of the specific backend is fixed after its initialization, can we extract how we determine the cudagraph_support out as a static method (I think passing what passed to the `builder_class. __init__()` is enough).  Alternative may be delaying the "initial cudagraph-relative stuff of each backend" after the backend initialization but before the first build() call (may be triggered at build_for cudagraph_capturing?). so the flow becomes  `initial each backend (also determine cudagraph support here)" -> resolve cudagraph mode -> "trigger cudagraph initialization at build_for_cudagraph_capturing".   I think the first option is easier, but let's see if you and @LucasWilkinson have other concerns on it.

PR #26938, opened 2025-10-15
[TPU] move torch_xla out of tpu.txt

yaochengji:
## Purpose

Move the legacy torch_xla requirement out of tpu.txt to avoid conflicts with the new tpu-inference backend.


yaochengji:
>  @yaochengji as the bot says, will this affect the dockerfile and should we update that too?

@mgoin thanks for reminding me! Yes, we need to make the dockerfile for the torch_xla backend correct.

Issue #26558, opened 2025-10-10
CUDA illegal memory access in fused_marlin_moe with GLM-4.6-GPTQ MoE model (V1 engine, TP=4)


chriswritescode-dev:
### Describe the bug

CUDA illegal memory access error during model initialization with GLM-4.6-GPTQ-Int4-Int8Mix MoE model when using the V1 engine with tensor parallelism.

### Environment

- vLLM version: 0.11.0rc2.dev371+gaafb99a4d.d20251010
- Platform: CUDA 13 (tried 12.8 12.9)
- GPU: 4x GPU RTX 6000 pro
- Python: 3.12
- PyTorch CUDA version: with nccl==2.27.3

### Configuration

```bash
vllm serve models/GLM-4.6-GPTQ-Int4-Int8Mix \
  --trust-remote-code \
  --served-model-name GLM-4.6 \
  --tensor-parallel-size 4 \
  --swap-space 16.0 \
  --enable-prefix-caching \
  --tool-call-parser glm45 \
  --reasoning-parser glm45 \
  --enable-auto-tool-choice
```

### Error Traceback

The error occurs during the profiling run in `determine_available_memory()`:

```python
torch.AcceleratorError: CUDA error: an illegal memory access was encountered
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.
```

**Stack trace:**
```
File "/home/ioplex/vllm/vllm/model_executor/layers/fused_moe/fused_marlin_moe.py", line 243, in fused_marlin_moe
    return torch.sum(intermediate_cache3.view(-1, topk, K), dim=1, out=output)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
torch.AcceleratorError: CUDA error: an illegal memory access was encountered
```

**Full traceback chain:**
1. `EngineCoreProc.__init__` → `_initialize_kv_caches()`
2. `model_executor.determine_available_memory()`
3. `gpu_worker.determine_available_memory()` → `model_runner.profile_run()`
4. `gpu_model_runner._dummy_run()` with compiled model
5. `fused_marlin_moe()` → illegal memory access

### Additional Context

[...]

PR #23770, opened 2025-08-27
[BugFix] Async scheduling and PP compatibility with DP

njhill:
Also fixes issue with finished requests not being processed in async scheduling and PP cases, and possible worker race conditions.

Should resolve/replace:
- https://github.com/vllm-project/vllm/pull/22543
- https://github.com/vllm-project/vllm/pull/21244
- https://github.com/vllm-project/vllm/pull/19558

Some possible follow-on work, out of scope for this PR:
- Avoid busy cpu spin in worker response waiting thread
- Single reactor loop for engine core - wait on input queue and worker response at the same time
- Uniproc async scheduling for TP=1

PR #18779, opened 2025-05-27
[V1] Support DP with Ray

ruisearch42:
This PR adds support for DP with Ray, with support for multi-node and API server scale out.

We reuse ZMQ communication mechanism between frontend and engine cores, as in #15977 , and the same API server scale out mechanism, as in https://github.com/vllm-project/vllm/pull/17546

Main differences from those PRs:
* The handshake between frontend and engine cores are greatly simplified, thanks to the Ray API
* We can launch all DP ranks just on the head node

### Examples

This will run DP=4 on the head node.

```
# Head node  (with ip address 10.99.48.128)
vllm serve $MODEL --data-parallel-size 4 --data-parallel-size-local 4 \
                  --data-parallel-address 10.99.48.128 --data-parallel-rpc-port 13345
                  --data-parallel-backend ray
```

This will run DP=4 with DP ranks 0 and 1 on the head node and ranks 2 and 3 on other nodes.

```
# Head node  (with ip address 10.99.48.128)
vllm serve $MODEL --data-parallel-size 4 --data-parallel-size-local 2 \
                  --data-parallel-address 10.99.48.128 --data-parallel-rpc-port 13345
                  --data-parallel-backend ray
```

This will run DP=4 with only the API server on the head node and all engines other nodes:

```
# Head node  (with ip address 10.99.48.128)
vllm serve $MODEL --data-parallel-size 4 --data-parallel-size-local 0 \
                  --data-parallel-address 10.99.48.128 --data-parallel-rpc-port 13345
                  --data-parallel-backend ray
```

### Design

See the following illustration. DP Coordinator is omitted, but is the same as https://github.com/vllm-project/vllm/pull/17546 .

image

PR #26953, opened 2025-10-15
Fix GPTQ Marlin MoE mixed precision support

chriswritescode-dev:
Fixes several critical issues preventing GPTQ Marlin MoE from working correctly with 8-bit quantization:

1. Shared memory calculation bug: The marlin_moe_wna16 kernel incorrectly calculated sh_b_size for 8-bit quantization, causing memory corruption. Fixed by accounting for b_thread_vecs based on bit width.

2. Workspace allocation: Added bit-width-aware workspace allocation (max_blocks_per_sm = 8 for 8-bit, 4 for 4-bit).

3. Cache validation: Added proper size validation for intermediate_cache2 and intermediate_cache13 buffers to prevent out-of-bounds access.

4. Quantization metadata: Enhanced GPTQMarlinMoEMethod to properly initialize and track per-layer quantization configuration including pack_factor and bit width.

5. SM 12.0 support: Added Blackwell architecture support to scaled_mm in CMakeLists.txt.

Includes new test coverage for mixed precision GPTQ Marlin MoE.

## Purpose
Resole issue #26558 CUDA illegal memory access in fused_marlin_moe with GLM-4.6-GPTQ MoE model (V1 engine, TP=4) 

### Test 1: test_gptq_marlin_moe_method_uses_dynamic_bits_override
                                                                                                                               
  Dynamic quantization override (8-bit) for layers 0-2  
                                                                                                                               
     • Validates that quant_metadata["bits"] = 8
     • Validates that pack_factor = 4 (32 // 8)
     • Validates that weight shapes are correct (64 // 4 = 16)
### Test 2: test_gptq_marlin_moe_method_retains_base_bits_without_override
                                                                                                                               
    Base 4-bit quantization for layers without override (layer 5)
                                                                                                                               
     • Validates that quant_metadata["bits"] = 4
     • Validates that pack_factor = 8 (32 // 4)
     • Validates that weight shapes are correct (64 // 8 = 8)  

## Test Result
tests/quantization/test_gptq_marlin_moe_mixed_precision.py::test_gptq_marlin_moe_method_uses_dynamic_bits_override PASSED    [ 50%]
tests/quantization/test_gptq_marlin_moe_mixed_precision.py::test_gptq_marlin_moe_method_retains_base_bits_without_override PASSED [100%]

---
<details>

</details>


yewentao256 (commented):
Let's run ci and see what can we get

joninco:
## Test Failure Analysis: `test_async_llm_dp.py[True-ray-RequestOutputKind.DELTA]`

  This PR is being blocked by a **pre-existing bug** in vLLM that was recently exposed. The failure is **not caused by changes in this PR**.

  ---

  ### 🔍 The Problem

  The test `v1/distributed/test_async_llm_dp.py::test_load[True-ray-RequestOutputKind.DELTA]` fails with:

  ValueError: Currently, async scheduling only supports mp or uni
  distributed executor backend, but you choose ray.

  **Root Cause:** Incorrect logic in `vllm/config/parallel.py` lines 533-538 conflates two **independent** configuration options:

  | Config Option | Purpose |
  |--------------|---------|
  | `data_parallel_backend` | How to spawn **DP engine processes** (Ray actors vs multiprocessing) |
  | `distributed_executor_backend` | How to spawn **TP/PP workers** within each DP rank (Ray vs mp vs uni) |

  These are separate concerns and should be independently configurable. However, the current code incorrectly forces them to match:

  ```python
  elif self.data_parallel_backend == "ray":
      logger.info("Using ray distributed inference because data_parallel_backend is ray")
      backend = "ray"  # ❌ BUG: Forces distributed_executor_backend="ray"

  This creates an invalid state when:
  - User sets data_parallel_backend="ray" (to use Ray for DP)
  - AND async_scheduling=True (which requires distributed_executor_backend in ["mp", "uni"])

  ---
  📅 Timeline: How This Bug Survived Until Now

  | Date         | Event                                                        | Impact
                              |
  |--------------|--------------------------------------------------------------|--------------------------------------------------------------------------
  ----------------------------|
  | June 2, 2025 | https://github.com/vllm-project/vllm/pull/18779 merged       | 🐛 Bug introduced: data_parallel_backend="ray" incorrectly forces
  distributed_executor_backend="ray" |
  | Aug 29, 2025 | https://github.com/vllm-project/vllm/pull/23770 merged       | ⚠️ Test extended to cover async_scheduling=True, but bug remained hidden
                              |
  | Jun-Oct 2025 | Workaround existed in arg_utils.py                           | ✅ Early distributed_executor_backend="mp" assignment bypassed the bug
                               |
  | Oct 18, 2025 | https://github.com/vllm-project/vllm/commit/6367bde73 merged | 💥 Workaround removed → bug now exposed in CI
                              |

  ---
  ✅ The Fix

  Remove the incorrect coupling between data_parallel_backend and distributed_executor_backend:

  diff --git a/vllm/config/parallel.py b/vllm/config/parallel.py
  @@ -530,12 +530,6 @@ class ParallelConfig:
                           "ray`."
                       )
                   backend = "ray"
  -            elif self.data_parallel_backend == "ray":
  -                logger.info(
  -                    "Using ray distributed inference because "
  -                    "data_parallel_backend is ray"
  -                )
  -                backend = "ray"
               elif ray_found:
                   if self.placement_group:
                       backend = "ray"

  What this enables:
  - ✅ Ray for DP + mp for TP/PP (the failing test case)
  - ✅ mp for DP + Ray for TP/PP
  - ✅ Any other valid combination

  After this fix, data_parallel_backend="ray" will correctly use distributed_executor_backend="mp" (the default for single-node), allowing the test to
  pass.

  ---
  🙏 Request for Maintainers

  This is blocking the unrelated PR #26953 (GPTQ Marlin MoE mixed precision fixes).

  Option 1 (Preferred): Merge the 6-line fix to vllm/config/parallel.py shown above

  Option 2 (Temporary): Skip the failing test combination until this is resolved:
  if async_scheduling and data_parallel_backend == "ray":
      pytest.skip("Known issue: data_parallel_backend incorrectly forces distributed_executor_backend")

  cc @yewentao256 
  ```
  
  p.s. is using AI to summarize this bug helpful or not for situations like this? I felt it was better at explaining the root cause than me.

chriswritescode-dev:
looks ready to merge 👍 . Is there something else that is needed ?

CMakeLists.txt
```diff
@@ -668,9 +668,9 @@ if(VLLM_GPU_LANG STREQUAL "CUDA")
...
 
   if(${CMAKE_CUDA_COMPILER_VERSION} VERSION_GREATER_EQUAL 13.0)
-    cuda_archs_loose_intersection(SCALED_MM_ARCHS "10.0f;11.0f" "${CUDA_ARCHS}")
+    cuda_archs_loose_intersection(SCALED_MM_ARCHS "10.0f;11.0f;12.0f" "${CUDA_ARCHS}")
   else()
-    cuda_archs_loose_intersection(SCALED_MM_ARCHS "10.0a" "${CUDA_ARCHS}")
+    cuda_archs_loose_intersection(SCALED_MM_ARCHS "10.0a;12.0a" "${CUDA_ARCHS}")
   endif()
   if(${CMAKE_CUDA_COMPILER_VERSION} VERSION_GREATER_EQUAL 12.8 AND SCALED_MM_ARCHS)
     set(SRCS "csrc/quantization/w8a8/cutlass/moe/grouped_mm_c3x_sm100.cu")
```
mgoin, line 755: This is not for marlin, it is for grouped_mm_c3x_sm100. Is this intentional somehow?

vllm/model_executor/layers/fused_moe/fused_marlin_moe.py
```diff
@@ -277,6 +277,26 @@ def fused_marlin_moe(
...
+        and intermediate_cache2.numel() < required_intermediate_cache2_elems
+    ):
+        intermediate_cache2 = None
+
+    required_intermediate_cache13_elems = M * topk * max(2 * N, K)
+    if (
+        intermediate_cache13 is not None
+        and intermediate_cache13.numel() < required_intermediate_cache13_elems
+    ):
+        intermediate_cache13 = None
```
mgoin: I don't understand the purpose of this code. If we need to initialize the workspace, it is better to do that during init rather than the forward pass. Also if we are passing in intermediate_cache that is too small, we should fix that upstream

mgoin (changes requested):

CMakeLists.txt
chriswritescode-dev: This was left in from building source. I have removed if from the PR. It was required to build on sm120 with torch nightly and cuda 130.

vllm/model_executor/layers/fused_moe/fused_marlin_moe.py
chriswritescode-dev: I moved the workspace allocation into the createweights method. leaving only the check if workspace is none to account for other usage and the fact that it was there before.

csrc/moe/marlin_moe_wna16/ops.cu
```diff
@@ -219,7 +219,8 @@ int get_kernel_cache_size(thread_config_t const& th_config, bool m_block_size_8,
   // both of them requires tb_m * 4 bytes (tb_m * int32 or tb_m * float32)
   int sh_block_meta_size = tb_m * 4;
   int sh_a_size = pipe_stages * (tb_m * tb_k) * 2;
-  int sh_b_size = pipe_stages * (tb_k * tb_n / pack_factor) * 4;
+  int b_thread_vecs = num_bits == 4 ? 1 : 2;
+  int sh_b_size = pipe_stages * (tb_k * tb_n / pack_factor) * 4 * b_thread_vecs;
```
jinzhen-lin: Why is `sh_b_size` multiplied by 2 extra for 8-bit weights? I understand that the `pack_factor` already accounts for the difference between 4-bit and 8-bit weights.

mgoin:
Test failure seems related
```
[2025-11-16T06:54:59Z] FAILED quantization/test_gptq_marlin_moe_mixed_precision.py::test_gptq_marlin_moe_method_uses_dynamic_bits_override - assert 4 == 8
```

PR #26956, opened 2025-10-15
[BugFix] Work around graph partition x torch.compile cache issue

zou3519:
In PyTorch 2.9, torch.compile has a bug where the graph partition is not taken into account during caching. Because vLLM's Mode.VLLM_COMPILE is the only mode that uses Inductor graph partition, and VLLM_COMPILE implies there is a PostGradPassManager, we put the list of operators to graph partition into the PostGradPassManager's uuid (which then gets incorporated into Inductor's FX graph cache key). Remove this hack whenever torch.compile fixes it.

### DIFF
--- a/vllm/compilation/pass_manager.py
+++ b/vllm/compilation/pass_manager.py
@@ -110,6 +110,27 @@ class PostGradPassManager(CustomGraphPass):
         self.post_cleanup = PostCleanupPass(config)
         self.fix_functionalization = FixFunctionalizationPass(config)
 
+        # [HACK: Bug with Inductor graph partition and torch.compile cache]
+        # In PyTorch 2.9, torch.compile has a bug where the graph
+        # partition is not taken into account during caching.
+        # Because vLLM's Mode.VLLM_COMPILE is the only mode that uses
+        # Inductor graph partition, and VLLM_COMPILE implies there
+        # is a PostGradPassManager, we put the list of operators to graph
+        # partition into the PostGradPassManager's uuid (which
+        # then gets incorporated into Inductor's FX graph cache key).
+        # Remove this hack whenever torch.compile fixes it.
+
+        # This is the list of operators that vLLM asks Inductor to split.
+        self.inductor_splitting_ops = []
+        if (
+            config.compilation_config.use_inductor_graph_partition
+            and config.compilation_config.splitting_ops is not None
+        ):
+            # Sort them so we're not dependent on the ordering.
+            self.inductor_splitting_ops = sorted(
+                config.compilation_config.splitting_ops
+            )
+
     def add(self, pass_: InductorPass):
         assert isinstance(pass_, InductorPass)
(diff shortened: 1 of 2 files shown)


vllm/compilation/pass_manager.py
```diff
@@ -110,6 +110,22 @@ def configure(self, config: VllmConfig):
...
+        # is a PostGradPassManager, we put the list of operators to graph
+        # partition into the PostGradPassManager's uuid (which
+        # then gets incorporated into Inductor's FX graph cache key).
+        # Remove this hack whenever torch.compile fixes it.
+        self.splitting_ops = None
+        if config.compilation_config.use_inductor_graph_partition:
+            # Sort them so we're not dependent on the ordering.
+            config.compilation_config.splitting_ops = sorted(
+                config.compilation_config.splitting_ops
+            )
```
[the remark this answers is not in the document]
zou3519: oops, updated

vllm/compilation/pass_manager.py
```diff
@@ -110,6 +110,23 @@ def configure(self, config: VllmConfig):
...
+        # partition is not taken into account during caching.
+        # Because vLLM's Mode.VLLM_COMPILE is the only mode that uses
+        # Inductor graph partition, and VLLM_COMPILE implies there
+        # is a PostGradPassManager, we put the list of operators to graph
+        # partition into the PostGradPassManager's uuid (which
+        # then gets incorporated into Inductor's FX graph cache key).
+        # Remove this hack whenever torch.compile fixes it.
+        self.splitting_ops = None
+        if config.compilation_config.use_inductor_graph_partition:
+            if config.compilation_config.splitting_ops is None:
```
ProExpertProg: Comment that we want empty splitting ops with inductor partition to behave differently than any splitting ops without inductor partition?

vllm/compilation/pass_manager.py
```diff
@@ -120,8 +137,17 @@ def uuid(self):
...
+            "pass_config": self.pass_config.uuid(),
+            "passes": [],
+            "splitting_ops": [],
+        }
         for pass_ in self.passes:
             state["passes"].append(pass_.uuid())
         state["passes"].append(self.fix_functionalization.uuid())
+
+        # See [HACK: Bug with Inductor graph partition and torch.compile cache]
+        if self.splitting_ops is not None:
```
ProExpertProg: Nvm, in both cases we will end up with an empty list

ProExpertProg (commented):
I'm confused; do we want to treat no-inductor-partition and inductor-partition-with-empty-splitting-ops differently or not?

ProExpertProg (approved):
@zou3519 could you update the test in test_toy_llama that currently disables FX cache to work around this? Then I can add this PR to the inductor partition CI PR

ProExpertProg (approved):
Thx

PR #26966, opened 2025-10-16
[CI/build] Removes source compilation from runtime image

bbartels:
## Purpose

This PR removes the need to build NVSHMEM from source which should build times of the Docker container by around 3-5 minutes
Additionally this moves any other source builds DEEPGEMM etc from the final runtime container to the build container and copies over the resulting artifacts to the final runtime image. This allows for removing any build level tools (compilers, git, etc) from the actual final image to reduce its size.
This additionally sets the stage for any future move to flashinfer AOT mode, such that the final base image can be moved away from the devel to the base image to save space.

## Test Plan

Ensure current CI builds and tests correctly

## Test Result

---
<details>

</details>


docker/Dockerfile
```diff
@@ -356,6 +371,21 @@ RUN --mount=type=bind,from=build,src=/workspace/dist,target=/vllm-workspace/dist
     uv pip install --system dist/*.whl --verbose \
         --extra-index-url ${PYTORCH_CUDA_INDEX_BASE_URL}/cu$(echo $CUDA_VERSION | cut -d. -f1,2 | tr -d '.')
 
+# Install deepgemm wheel that has been built in the `build` stage
+COPY --from=build /tmp/deepgemm/dist /tmp/deepgemm/dist
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv pip install --system /tmp/deepgemm/dist/*.whl
```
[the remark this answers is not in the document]
mgoin: I think this is a valid concern

docker/Dockerfile
```diff
@@ -237,6 +237,21 @@ RUN if [ "$RUN_WHEEL_CHECK" = "true" ]; then \
     else \
         echo "Skipping wheel size check."; \
     fi
+
+# Build DeepGEMM wheel from source
+ARG DEEPGEMM_GIT_REF
+COPY --chmod=755 tools/install_deepgemm.sh /tmp/install_deepgemm.sh
+RUN --mount=type=cache,target=/root/.cache/uv \
+    VLLM_DOCKER_BUILD_CONTEXT=1 TORCH_CUDA_ARCH_LIST="9.0a 10.0a" /tmp/install_deepgemm.sh --cuda-version "${CUDA_VERSION}" ${DEEPGEMM_GIT_REF:+--ref "$DEEPGEMM_GIT_REF"} --wheel-dir /tmp/deepgemm/dist
```
bbartels:
```suggestion
RUN --mount=type=cache,target=/root/.cache/uv \
    VLLM_DOCKER_BUILD_CONTEXT=1 TORCH_CUDA_ARCH_LIST="9.0a 10.0a" /tmp/install_deepgemm.sh --cuda-version "${CUDA_VERSION}" ${DEEPGEMM_GIT_REF:+--ref "$DEEPGEMM_GIT_REF"} --wheel-dir /tmp/deepgemm/dist
# Ensure the wheel dir exists so later-stage COPY won't fail when DeepGEMM is skipped
RUN mkdir -p /tmp/deepgemm/dist && touch /tmp/deepgemm/dist/.deepgemm_skipped
```

docker/Dockerfile
bbartels:
```suggestion
# Install deepgemm wheel that has been built in the `build` stage
COPY --from=build /tmp/deepgemm/dist /tmp/deepgemm/dist
RUN --mount=type=cache,target=/root/.cache/uv \
    sh -c 'if ls /tmp/deepgemm/dist/*.whl >/dev/null 2>&1; then \
              uv pip install --system /tmp/deepgemm/dist/*.whl; \
           else \
              echo "No DeepGEMM wheels to install; skipping."; \
           fi'
```

mgoin:
I think this looks good to merge, but I'd like to wait for the release to cut so we can prevent last minute issues. Thanks for doing this!

geraldstanje1:
hi @bbartels this broke today:
```
 => CANCELED [vllm-base  4/19] RUN echo 'tzdata tzdata/Areas select America' | debconf-set-selections     && echo 'tzdata  15.5s
 => [base  3/10] RUN apt-get install -y gcc-10 g++-10                                                                       6.7s
 => [base  4/10] RUN update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-10 110 --slave /usr/bin/g++ g++ /usr/bin/  0.1s
 => [base  5/10] RUN <<EOF (gcc --version)                                                                                  0.2s
 => [base  6/10] RUN ldconfig /usr/local/cuda-$(echo 12.8.1 | cut -d. -f1,2)/compat/                                        0.2s
 => [base  7/10] WORKDIR /workspace                                                                                         0.0s
 => [base  8/10] COPY requirements/common.txt requirements/common.txt                                                       0.0s
 => [base  9/10] COPY requirements/cuda.txt requirements/cuda.txt                                                           0.0s
 => ERROR [base 10/10] RUN --mount=type=cache,target=/root/.cache/uv     uv pip install --python /opt/venv/bin/python3 -r   1.2s
------
 => ERROR [base 10/10] RUN --mount=type=cache,target=/root/.cache/uv     uv pip install --python /opt/venv/bin/python3 -r   1.2s
------
------
 > [base 10/10] RUN --mount=type=cache,target=/root/.cache/uv     uv pip install --python /opt/venv/bin/python3 -r requirements/cuda.txt     --extra-index-url https://download.pytorch.org/whl/cu$(echo 12.8.1 | cut -d. -f1,2 | tr -d '.'):
0.183 Using Python 3.12.12 environment at: /opt/venv
1.103   × No solution found when resolving dependencies:
1.103   ╰─▶ Because there is no version of apache-tvm-ffi==0.1.0b15 and
1.103       flashinfer-python==0.4.0 depends on apache-tvm-ffi==0.1.0b15, we can
1.103       conclude that flashinfer-python==0.4.0 cannot be used.
1.103       And because you require flashinfer-python==0.4.0, we can conclude that
1.103       your requirements are unsatisfiable.
1.103
1.103       hint: `apache-tvm-ffi` was requested with a pre-release marker (e.g.,
1.103       apache-tvm-ffi==0.1.0b15), but pre-releases weren't enabled (try:
1.103       `--prerelease=allow`)
```

see: https://pypi.org/project/apache-tvm-ffi/#history

bbartels:
I dont see why not, if that's the case!

bbartels:
@cjackal I've removed the manual installation of nvshmem and tested it locally, seems to all work as expected

bbartels:
While not the main point of the PR, this saves about 1.1GB in image size :tada:

tools/ep_kernels/install_python_libraries.sh
```diff
@@ -1,94 +1,75 @@
...
-# assume CUDA_HOME is set correctly
-if [ -z "$CUDA_HOME" ]; then
-    echo "CUDA_HOME is not set, please set it to your CUDA installation directory."
-    exit 1
-fi
-
-# assume TORCH_CUDA_ARCH_LIST is set correctly
-if [ -z "$TORCH_CUDA_ARCH_LIST" ]; then
-    echo "TORCH_CUDA_ARCH_LIST is not set, please set it to your desired architecture."
+uv pip install cmake torch ninja
```
mgoin: should this have "--system"?

bbartels: I am not sure, whats the concern, that it bloats the venv of the container? Its just the build container, so doesnt affect runtime image size. 
It did build it locally without issue

mgoin: I just mean that "--system" is used when uv pip install is used in the dockerfile, since there is no venv

bbartels: Fair! I've amended it and ran the build again, succeeds as well!

mgoin: It won't work for local installs though, so it depends on the purpose of this script

bbartels: I believe the dependencies are needed to build either `pplx-kernels` and/or `DeepEP`. So temporary dependencies. Might be worth creating a venv and disposing it once the build finished? Though I am not all that knowledgeable on python build tooling :sweat_smile:

bbartels: gpt-5 suggested this:
```
if [ -z "$VIRTUAL_ENV" ]; then
  uv pip install --system cmake torch ninja
else
  uv pip install cmake torch ninja
fi
```

bbartels: I've applied that now, should work both with and without venv now. Happy to amend if you have a better idea!

mgoin (approved):

PR #26971, opened 2025-10-16
[Misc] Remove unused attention prefix prefill ops functions

lgeiger:
This removes a couple of functions that don't seem to be used anywhere and are reported as `reportUnusedFunction` by `basedpyright`.


vllm/attention/ops/prefix_prefill.py
```diff
@@ -335,216 +335,6 @@ def _fwd_kernel(
     return
 
 
-@triton.jit
-def _fwd_kernel_flash_attn_v2(
```
lgeiger, line 339: Let me know if it would still be useful to keep this kernel even if it's not used anywhere in the codebase

vllm/entrypoints/openai/serving_pooling.py
```diff
@@ -34,29 +30,13 @@
...
 from vllm.logger import init_logger
-from vllm.outputs import PoolingOutput, PoolingRequestOutput
+from vllm.outputs import PoolingRequestOutput
 from vllm.tasks import SupportedTask
 from vllm.utils.async_utils import merge_async_iterators
 
 logger = init_logger(__name__)
 
 
-def _get_data(
```
DarkLight1337: cc @noooop I thought you removed this before?

noooop: Sorry. after a rebase, it was pushed again, I plan to rm it in #26973  

It was indeed not used.

You can also delete in this pr

vllm/attention/ops/prefix_prefill.py
DarkLight1337: cc @Isotr0py @mgoin @LucasWilkinson

vllm/entrypoints/openai/serving_pooling.py
lgeiger: Sounds good, rebased this PR ontop of latest main which doesn't include this function anymore

vllm/attention/ops/prefix_prefill.py
Isotr0py: This should be safe to remove. We have other triton flash_attn kernel now:
https://github.com/vllm-project/vllm/blob/03c4c4aa9deb2ad09a95c7997d2e5578c8db68d6/vllm/attention/ops/triton_flash_attention.py#L356-L361

Isotr0py (approved):
Sorry for missing this channel. LGTM!

yewentao256 (approved):
LGTM, thanks for the work!

PR #26414, opened 2025-10-08
[Frontend][1/N] Improve all pooling task | Support FP16 Embedding Base64 (Still uses fp32 by default).

noooop:
## Improve all pooling task
- #25817
- #26414
- #25370
- #27066
- #26973
- #25524

These PRs are mostly conflicting with each other, so combining them into a series would better inform reviewers about what happened. And what else needs to be done after that?

## Purpose

FIX #26248

1. Eliminate overhead of converting tensor -> list[float] -> numpy in OpenAI API server
2. Support float32, float16, bfloat16, fp8_e4m3, fp8_e5m2 embed dtype. 
3. The fp8 is better than expected, perhaps with practical value, please thoroughly test before use, welcome to feedback.

mteb test PTAL #17175
https://github.com/noooop/snippet/blob/main/benchmarks/test_mteb/test_embed_dtype.py

image

image

float32 ≈ float16 > bfloat16 > fp8_e4m3 >> fp8_e5m2

| model_name | float32 | float16 | bfloat16 | fp8_e4m3 | fp8_e5m2 |
| -- | -- | -- | -- | -- | -- |
| jinaai/jina-embeddings-v3 | 0.824336501 | 0.824339743 | 0.824335268 | 0.824326947 | 0.824323946 |
| BAAI/bge-m3 | 0.78734317 | 0.787339618 | 0.787350117 | 0.787262401 | 0.78773927 |
| intfloat/multilingual-e5-base | 0.779325776 | 0.779326888 | 0.779314874 | 0.779537205 | 0.779854464 |
| BAAI/bge-base-en | 0.779337092 | 0.779340641 | 0.779370285 | 0.779269983 | 0.779406319 |
| Alibaba-NLP/gte-multilingual-base | 0.775074375 | 0.775069325 | 0.77506051 | 0.775014809 | 0.774906984 |
| Qwen/Qwen3-Embedding-0.6B | 0.771163535 | 0.771168301 | 0.771179591 | 0.771177116 | 0.771265142 |
| thenlper/gte-large | 0.768076652 | 0.768071025 | 0.768082659 | 0.768333837 | 0.767567718 |
| Alibaba-NLP/gte-Qwen2-1.5B-instruct | 0.758472692 | 0.75847403 | 0.758461189 | 0.758305555 | 0.758201921 |
| BAAI/bge-code-v1 | 0.757253707 | 0.757255983 | 0.757230987 | 0.75725761 | 0.756827524 |
[...]

PR #25817, opened 2025-09-28
[Model][0/N] Improve all pooling task | clean up

noooop:
## Improve all pooling task
- #25817
- #26414
- #25370
- #27066
- #26973
- #25524

These PRs are mostly conflicting with each other, so combining them into a series would better inform reviewers about what happened. And what else needs to be done after that?

## Purpose

Fix some minor issues found during the implementation of this series to keep subsequent PRs cleaner.
- add embedding_size
- add VLLM_CI_ENFORCE_EAGER & get_vllm_extra_kwargs
- ner.py -> ner_client.py
- fix gpt2 seq cls & add test_head_dtype.py
- clean GritLM
- fix skip_global_cleanup
- fix test_splade_pooler

## Test Plan

keep ci grean

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #25524, opened 2025-09-23
[Frontend][Doc][5/N] Improve all pooling task | Polish encode (pooling) api & Document.

noooop:
## TL;DR
- pooling_task required for llm.encode
- /pooling endpoint support all pooling tasks
- softmax, activation -> use_activation

## Improve all pooling task (0.11.1 cut)
- #25817
- #26414
- #25370
- #27066
- #26973
- #25524

These PRs are mostly conflicting with each other, so combining them into a series would better inform reviewers about what happened. And what else needs to be done after that?

## Purpose
Following #25370 

- Split the encode task into two tasks: token_embed and token_classify
    - token_embed is the same as embed, using normalize as activation
    - token_classify is the same as classify, default using softmax as activation (we actually allow classify and token_classify to use any activation function by setting act_fn. )

Address:  https://github.com/vllm-project/vllm/issues/27413#issuecomment-3441413201
-  /pooling endpoint support all pooling tasks. 

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #25370, opened 2025-09-22
[Model][2/N] Improve all pooling task | Support multi-vector retrieval

noooop:
## Break change
- Split the encode task into two tasks: token_embed and token_classify
    - token_embed is the same as embed, using normalize as activation
    - token_classify is the same as classify, default using softmax as activation (we actually allow classify and token_classify to use any activation function by setting act_fn. )
- Complete remove softmax from PoolingParams and perfer using activation, since we actually allow classify and token_classify to use any activation function. 
- pooling_task required for llm.encode and /pooling <- This will take effect entirely in the next PR (#25524)

## Improve all pooling task
- #25817
- #26414
- #25370
- #27066
- #26973
- #25524

These PRs are mostly conflicting with each other, so combining them into a series would better inform reviewers about what happened. And what else needs to be done after that?

## Purpose

After #21227 landed, we hope that pooling models can always use all pooling, and users don’t need to manually set using all pooling.

The current encode api (/pooling api ) mainly targets the classify for each token scenario  (e.g. TokenClassification #24872 & reward models), overlooked the embed for each token scenario.

Let's support embed for each token scenario  (multi-vector retrieval) 

Partial_Fix #25165

We are stepping closer to support [ColBERT](https://arxiv.org/abs/2004.12832) & [ColPali](https://arxiv.org/abs/2407.01449)

cc @DarkLight1337 @maxdebayser

## Test Plan

tests/models/language/pooling/test_multi_vector_retrieval.py
tests/test_pooling_params.py

## Test Result

pass

## Known Issues

- Maybe we should find a way to support chunked prefill + all pooling (and mean pooling))
- support ColBERT & ColPali

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #26973, opened 2025-10-16
[Frontend][4/N] Improve all pooling task | Add plugin pooling task

noooop:
## Improve all pooling task
- #25817
- #26414
- #25370
- #27066
- #26973
- #25524

## Purpose

plugin task uses io_processor.parse_request to verify inputs, skipping PoolingParams verify

## Test Result

---
<details>

</details>


maxdebayser:
I think this extra task makes sense. Should we also add `extra_kwargs` to the request classes in `protocol.py` and in `PoolingParams`? In this way we would be able to pass custom args to the plugin.

noooop:
> I think this extra task makes sense. Should we also add `extra_kwargs` to the request classes in `protocol.py` and in `PoolingParams`? In this way we would be able to pass custom args to the plugin.

IOProcessor uses a data-in, data-out, where the data can be any thing

Do we still need a separate extra_kwargs, or can extra_kwargs be part of the data? The user can actually use any data format.

noooop:
@christian-pinto PTAL #27063 #27066 

I hope io_processor_plugins can support response compression, it will definitely be more efficient compared to base64.

maxdebayser:
> IOProcessor uses a data-in, data-out, where the data can be any thing

Yes, you're right.

noooop:
@christian-pinto #27066 (binary response) has been merged. 

Please help check if it can be used together with plugin task. 

It would be best to have an example to tell users how to use binary response with prithvi_geospatial_mae .

noooop:
The new plugin pooling task looks great. @christian-pinto Thanks for your help!


cc @DarkLight1337

Are there any more modifications needed for this PR?

vllm/entrypoints/llm.py
```diff
@@ -1024,21 +1024,9 @@ def encode(
...
-
-        for param in as_iter(pooling_params):
-            param.verify(pooling_task, model_config)
-            # for backwards compatibility
-            if truncate_prompt_tokens is not None:
-                param.truncate_prompt_tokens = truncate_prompt_tokens
-
         io_processor_prompt = False
         if isinstance(prompts, dict) and "data" in prompts:
+            assert self.io_processor is not None
```
DarkLight1337: We will raise ValueError in the next lines, no need to add this assertion

vllm/plugins/io_processors/interface.py
```diff
@@ -63,6 +65,13 @@ async def post_process_async(
     def parse_request(self, request: Any) -> IOProcessorInput:
         raise NotImplementedError
 
+    def validate_or_generate_params(
+        self, params: SamplingParams | PoolingParams | None = None
+    ) -> SamplingParams | PoolingParams:
+        if params:
+            return params
+        return PoolingParams()
```
DarkLight1337:
```suggestion
        return params or PoolingParams()
```

christian-pinto: ok, let me change that.

vllm/entrypoints/llm.py
christian-pinto: I believe many of these asserts are there to please mypy as it was otherwise complaining that io_processor cooukd be none and therefore have none of the methods we invoke. I can try to remove and see how it goes.

christian-pinto: you are actually right, I did not see the following line from the comment.

DarkLight1337 (approved):

PR #26443, opened 2025-10-08
[UX] Add FlashInfer as default CUDA dependency

mgoin:
## Purpose

It seems that FlashInfer does not require `nvcc` to installed from source anymore since `flashinfer-python>=0.2.9`, so we can move it to be a default dependency!

Obviously to have it JIT compile kernels it needs to have `nvcc` available, so we will currently add that condition to `has_flashinfer()`. This is more conservative than it needs to be, as if we have an AOT compiled wheel installed there is likely no need for `nvcc`, but we can improve this later.

Also with flashinfer-python==0.4.0, we can now move to using flashinfer-cubin and flashinfer-jit-cache wheels instead of pre-compiling ourselves in the docker!

## Test Result

I used a uv python image to simulate installing flashinfer-python without any cuda toolkit available

```
docker run -it --rm astral/uv:bookworm-slim bash
root@bbbb3b887d99:/# uv venv --python 3.12
root@bbbb3b887d99:/# source .venv/bin/activate
(.venv) root@bbbb3b887d99:/# uv pip install flashinfer-python==0.2.8
...
      FileNotFoundError: [Errno 2] No such file or directory: 'nvcc'

(.venv) root@bbbb3b887d99:/# uv pip install flashinfer-python==0.2.9
Installed 41 packages in 451ms

(.venv) root@bbbb3b887d99:/# uv pip install flashinfer-python==0.3.1
Installed 5 packages in 103ms
```

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #26977, opened 2025-10-16
[Kernel] Lazy import FlashInfer

jeejeelee:
## Purpose
- `Importing flashinfer` adds considerable startup overhead. Since we have already disabled  fashinfer sampler by default(see: https://github.com/vllm-project/vllm/pull/26859), we can  use lazy import to avoid the startup overhead caused by importing flashinfer
- Additionally, considering that flashinfer (currently 0.4.0) is already a default dependency(see: https://github.com/vllm-project/vllm/pull/26443), when users explicitly set `VLLM_USE_FLASHINFER_SAMPLER=1 `, they should be able to directly use flashinfer's sampler. However, considering there may be edge cases where the flashinfer version is lower than  `0.2.3`, we've added `ImportError` handling instead of fallback
## Test Result

---
<details>

</details>


mgoin (approved):
LGTM, thanks for the improvement!

PR #26986, opened 2025-10-16
[Platform] Cleanup useless function

wangxiyuan:
## Purpose
The purpose of Platform interface is to  provide functions between vLLM framework and hardware implement. Hardware implement it and vLLM call it. But `check_if_supports_dtype` broke the principle. This function is only called in hardware worker. So it should be in the worker only. 

This PR remove this useless interface.

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/platforms/cuda.py
+++ b/vllm/platforms/cuda.py
@@ -463,6 +463,12 @@ class CudaPlatformBase(Platform):
 
     @classmethod
     def check_if_supports_dtype(cls, dtype: torch.dtype):
+        logger.warning_once(
+            "check_if_supports_dtype is deprecated and will be removed in"
+            "v0.12.0 or v1.0.1. It's replaced by "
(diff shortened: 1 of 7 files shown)


wangxiyuan:
@jikunshang Got your point. your case is that in the future, GPU worker/modelrunner may use `current_platform` way to distinguish different platform. There are two thing from my sight:
1. Call `current_platform` in worker or modelrunner  is not good. I'm afraid  it may break the plugin's arch and make the code complex to maintain. for example, in this PRs case, we create  an platform interface, while this interface is only used in GPU worker/modelrunner,  should it be a utils function in GPU only rather than an common platform interface?
2. AFAIK, GPU modelrunner refactor is WIP, if this is really a useful interface, maybe we can add it back in the future?

wangxiyuan:
after discus with @jikunshang, let's close this one and find a better way to improve CUDA-LIKE interface.

PR #26998, opened 2025-10-16
Enhance: Error handling for structured output compilation failures

quanliu1991:
## Purpose
When grammar compilation for structured output failed, the system raised an uncaught exception inside
`self.backend.compile_grammar(request_type, grammar_spec)`, leading to an crash error message such as:

```EngineCore encountered an issue. See stack trace (above) for the root cause.```

This behavior caused occasionally disrupted service stability.

Now, when a compilation failure occurs:
1. The request is aborted in schedule.
2. The API returns a standardized error response instead of crashing.
3. Other requests continue unaffected.

## Test Plan
Simulated a grammar compilation failure that triggered an exception in structure output.
```
   def compile_grammar(self, request_type: StructuredOutputOptions,
                        grammar_spec: str) -> StructuredOutputGrammar:
        raise ValueError("Test simulation compilation failed")
```
The request was then aborted, and the API returned a proper error message and no service crash occurred.
## Test Result
**Befor result:**
service crash```EngineCore encountered an issue. See stack trace (above) for the root cause.```

**Now result:**
{
  "error": {
    "object": "error",
    "message": "Structured Output Compiled Failed: Test simulation compilation failed",
    "type": "BadRequestError",
    "param": null,
    "code": 400
  }
}

### DIFF
--- a/vllm/entrypoints/openai/serving_chat.py
+++ b/vllm/entrypoints/openai/serving_chat.py
@@ -1077,6 +1077,8 @@ class OpenAIServingChat(OpenAIServing):
 
                     # if the model is finished generating
                     else:
+                        if output.finish_reason == "abort":
+                            raise ValueError(output.stop_reason)
                         # check to make sure we haven't "forgotten" to stream
                         #   any tokens that were generated but previously
                         #   matched by partial json parsing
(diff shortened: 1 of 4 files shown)


quanliu1991:
@heheda12345 Could you please help review this PR?

PR #27010, opened 2025-10-16
[CI] Nixl integration tests

NickLucche:
Turn on NIXL distributed integration tests in CI

### DIFF
--- a/.buildkite/test-pipeline.yaml
+++ b/.buildkite/test-pipeline.yaml
@@ -1069,7 +1069,7 @@ steps:
   commands:
     - bash weight_loading/run_model_weight_loading_test.sh -c weight_loading/models-large.txt
   
-- label: NixlConnector P/D accuracy tests (Distributed) # TODO min
+- label: NixlConnector P/D accuracy tests (Distributed) # 30min
   timeout_in_minutes: 30
   working_dir: "/vllm-workspace/tests"
   num_gpus: 4
--- a/.buildkite/test-pipeline.yaml
+++ b/.buildkite/test-pipeline.yaml
@@ -1076,7 +1076,7 @@ steps:
     - vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
     - tests/v1/kv_connector/nixl_integration/
   commands:
-    - bash v1/kv_connector/nixl_integration/tp_config_sweep_accuracy_test.sh
+    - bash tests/v1/kv_connector/nixl_integration/tp_config_sweep_accuracy_test.sh
 
 
 ##### multi gpus test #####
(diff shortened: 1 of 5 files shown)


tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh
```diff
@@ -130,7 +136,7 @@ run_tests_for_model() {
     vllm serve $model_name \
     --port $PORT \
     --enforce-eager \
```
markmc, line 138: `--enable-log-requests` could be useful here if you're doing another update

markmc (approved):
lgtm

tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh
NickLucche: yep still waiting for CI to report

tests/v1/kv_connector/nixl_integration/toy_proxy_server.py
```diff
@@ -76,7 +76,8 @@ def parse_args():
     parser = argparse.ArgumentParser()
 
     parser.add_argument("--port", type=int, default=8000)
-    parser.add_argument("--host", type=str, default="localhost")
+    # Always use 127.0.0.1 as localhost binds to IPv6 which is blocked on CI
+    parser.add_argument("--host", type=str, default="127.0.0.1")
```
markmc, line 80: Yeah, IPv6 is a fad :rofl: 

(In all seriousness - we should support try to support IPv6-only hosts too)

It looks like `vllm serve` passes `host=None` by default to uvicorn and that binds to all interfaces?

tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh
NickLucche: allow me to postpone this change, I think I should probably silence the progress bar from lm-eval first because it's already way too verbose given it will show outputs from up to 4 instances all at once..

NickLucche: @markmc

tests/v1/kv_connector/nixl_integration/toy_proxy_server.py
NickLucche: good point, I am just getting this pr in-place so we have some base for all the upcoming changes we have lined up. We can investigate the proper method in a follow-up

.buildkite/test-pipeline.yaml
```diff
@@ -1068,6 +1068,17 @@ steps:
   - tests/weight_loading
   commands:
     - bash weight_loading/run_model_weight_loading_test.sh -c weight_loading/models-large.txt
+  
+- label: NixlConnector P/D accuracy tests (Distributed) # TODO min
```
DarkLight1337: Can you add the duration here? Then I'll force-merge it

NickLucche: @DarkLight1337 got it

DarkLight1337 (approved):

PR #23809, opened 2025-08-28
[fix]: add Arm 4bit fused moe support

nikhil-arm:
## Purpose
Add support for fused 4bit moe for Arm platform
## Test Result
Tested with Deepseek v2 lite model. The generation is correct with upto 3x speedups.
---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #22951, opened 2025-08-15
[Kernel] Add cuda kernel for gpt_oss activation

jeejeelee:
# Essential Elements of an Effective PR Description Checklist

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.

## Purpose
Reopen of https://github.com/vllm-project/vllm/pull/22538
### lm eval
```shell
VLLM_ATTENTION_BACKEND=TRITON_ATTN_VLLM_V1 lm_eval   --model vllm   --model_args "pretrained=unsloth/gpt-oss-20b-BF16,max_model_len=32768"   --trust_remote_code   --tasks gsm8k   --num_fewshot 5   --batch_size auto
```
### benchmrk
```shell
# serve
VLLM_ATTENTION_BACKEND=TRITON_ATTN_VLLM_V1 vllm serve  unsloth/gpt-oss-20b-BF16 

# bench script
vllm bench serve \
  --backend vllm \
  --model unsloth/gpt-oss-20b-BF16   \
  --endpoint /v1/completions \
  --dataset-name random \
  --random-input 2048 \
  --random-output 1024 \
  --max-concurrency 10 \
  --num-prompt 100 \

```
## Test Result(A800)

**This PR**

```text
vllm (pretrained=unsloth/gpt-oss-20b-BF16,max_model_len=32768,trust_remote_code=True), gen_kwargs: (None), limit: None, num_fewshot: 5, batch_size: auto
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.3654|±  |0.0133|
|     |       |strict-match    |     5|exact_match|↑  |0.2593|±  |0.0121|


============ Serving Benchmark Result ============
Successful requests:                     100       
Maximum request concurrency:             10        
Benchmark duration (s):                  146.79    
Total input tokens:                      204483    
Total generated tokens:                  65236     
Request throughput (req/s):              0.68      
Output token throughput (tok/s):         444.42    
Total Token throughput (tok/s):          1837.46   
---------------Time to First Token----------------
Mean TTFT (ms):                          389.82    
[...]

PR #22508, opened 2025-08-08
[oss] Init gpt-oss bf16 support

jeejeelee:
# Essential Elements of an Effective PR Description Checklist

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.

## Purpose

WIP:
- [x] Currently only supports eager mode
- [x] Support TP>1 
## Test Plan
```python
import os
from vllm import LLM, SamplingParams

os.environ["VLLM_ATTENTION_BACKEND"] = "TRITON_ATTN_VLLM_V1"
os.environ["TORCH_CUDA_ARCH_LIST"] = "8.0 8.6"


def main():
    prompts = [
        "<|start|>system<|message|>You are ChatGPT, a large language model trained by OpenAI.\nKnowledge cutoff: 2024-06\nCurrent date: 2025-08-08\n\nReasoning: medium\n\n# Valid channels: analysis, commentary, final. Channel must be included for every message.\nCalls to these tools must go to the commentary channel: 'functions'.<|end|><|start|>user<|message|>Who are you..<|end|><|start|>assistant"  # noqa: E501
    ]
    # Create a sampling params object.
    sampling_params = SamplingParams(
        temperature=0.8, top_p=0.95, max_tokens=1024
    )
    llm = LLM(
        model="unsloth/gpt-oss-20b-BF16",
        enforce_eager=True,
    )
    # Generate texts from the prompts.
    # The output is a list of RequestOutput objects
    # that contain the prompt, generated text, and other information.
    outputs = llm.generate(prompts, sampling_params)
    # Print the outputs.
    print("\nGenerated Outputs:\n" + "-" * 60)
    for output in outputs:
        prompt = output.prompt
        generated_text = output.outputs[0].text
        print(f"Prompt:    {prompt!r}")
        print(f"Output:    {generated_text!r}")
        print("-" * 60)


if __name__ == "__main__":
    main()

```
## Test Result
On the local A800 machine, the generated results are as follows:
```text
Generated Outputs:
------------------------------------------------------------
[...]

PR #27024, opened 2025-10-16
Transform HF interleaved weights to halves in vllm

isharif168:
## Purpose
- HF provides the gate + up weights interleaved as [g0, u0, g1, u1]
- vllm expects the gate + up tensors to be in halves [g0, g1, u0, u1]
- Add a funtion to do the transformation for gate + up weights and bias

## Test Plan
Tested with gpt-oss

## Test Result
Gives the correct results on CPU

### DIFF
--- a/vllm/model_executor/layers/fused_moe/layer.py
+++ b/vllm/model_executor/layers/fused_moe/layer.py
@@ -324,6 +324,7 @@ class FusedMoE(CustomOp):
         expert_mapping: list[tuple[str, str, int, str]] | None = None,
         n_shared_experts: int | None = None,
         routing_method_type: int | None = None,
+        is_weights_interleaved: bool = False,
     ):
         super().__init__()
 
@@ -362,6 +363,8 @@ class FusedMoE(CustomOp):
             tp_size if tp_size is not None else get_tensor_model_parallel_world_size()
         )
         dp_size_ = dp_size if dp_size is not None else get_dp_group().world_size
+
+        self.is_weights_interleaved = is_weights_interleaved
 
         self.is_sequence_parallel = is_sequence_parallel
         self.sp_size = tp_size_ if is_sequence_parallel else 1
--- a/vllm/model_executor/models/gpt_oss.py
+++ b/vllm/model_executor/models/gpt_oss.py
@@ -173,6 +173,7 @@ class MLPBlock(torch.nn.Module):
             has_bias=True,
             activation="swigluoai",
             is_sequence_parallel=self.is_sequence_parallel,
+            is_weights_interleaved=True,
         )
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
--- a/vllm/model_executor/layers/fused_moe/unquantized_fused_moe_method.py
+++ b/vllm/model_executor/layers/fused_moe/unquantized_fused_moe_method.py
@@ -262,6 +262,19 @@ class UnquantizedFusedMoEMethod(FusedMoEMethodBase, CustomOp):
                 else:
                     layer.cpu_fused_moe = cpu_fused_moe.IPEXFusedMOE(layer)
             else:
+                if getattr(layer, "is_weights_interleaved", False):
+                    from vllm.model_executor.layers.fused_moe.utils import (
+                        reorder_gate_up_to_halves,
+                    )
+
+                    layer.w13_weight.copy_(
+                        reorder_gate_up_to_halves(layer.w13_weight, axis=1)
+                    )
+                    if hasattr(layer, "w13_bias"):
+                        layer.w13_bias.copy_(
+                            reorder_gate_up_to_halves(layer.w13_bias, axis=-1)
+                        )
+                    layer.is_weights_interleaved = False
                 layer.cpu_fused_moe = cpu_fused_moe.CPUFusedMOE(layer)
 
     def apply(
--- a/vllm/model_executor/layers/fused_moe/utils.py
+++ b/vllm/model_executor/layers/fused_moe/utils.py
@@ -330,3 +330,23 @@ def activation_without_mul(activation: str) -> str:
 @functools.cache
 def disable_inplace() -> bool:
     return is_torch_equal_or_newer("2.9")
+
+
+def reorder_gate_up_to_halves(t: torch.Tensor, axis: int) -> torch.Tensor:
+    """
+    Treat dimension `axis` as interleaved [g0,u0,g1,u1,...] and reorder to
+    [g..., u...]. Always returns contiguous.
+    """
+    if axis < 0:
+        axis += t.ndim
+    size = t.shape[axis]
+    if size % 2 != 0:
+        return t.contiguous()
+    moved = axis != t.ndim - 1
+    if moved:
+        t = t.movedim(axis, -1)
+    shape = t.shape
+    t = t.reshape(shape[:-1] + (shape[-1] // 2, 2))
+    t = torch.cat([t[..., 0], t[..., 1]], dim=-1)
+    t = t.movedim(-1, axis).contiguous() if moved else t.contiguous()
+    return t


vllm/model_executor/models/gpt_oss.py
```diff
@@ -511,6 +535,9 @@ def _load_weights_other(
                     narrow_weight = weight[:, :, 2 * tp_rank_start : 2 * tp_rank_end]
 
                 narrow_weight = narrow_weight.permute(0, 2, 1).contiguous()
+                narrow_weight = self._maybe_reorder_gate_up_to_halves(
```
nikhil-arm: Shouldnt we guard this for aarch64 cpu? I am not clear why we should do this for GPUs or other accelrators

vllm/model_executor/models/gpt_oss.py
```diff
@@ -474,6 +474,30 @@ def _load_weights_mxfp4(
             loaded_params.add(name)
         return loaded_params
 
+    @staticmethod
+    def _maybe_reorder_gate_up_to_halves(t: torch.Tensor, axis: int) -> torch.Tensor:
```
nikhil-arm: what happens to the original tensors after deinterleaving. Can we check if memory is not bloated

nikhil-arm:
I think we need to make this logic reusable for CPU until we add support for interleaved MoE.
This should not be just GPTOSS specific but a feature for CPUMoE in vLLM. 
We can load the interleaved weights at modelling level but deinterleave them if vLLM dispatches to CPUMoE in the backend. Something similar to what we do for 4 bit fused MoE on Arm today.

vllm/model_executor/models/gpt_oss.py
fadara01: I discussed this with Sharif. It's risky for other backends which already work with the current code, and it's also not elegant to put Architecture specific details to the modeling file.

The proposal is to add a new flag (e.g. `gu_interleaved: bool = False`) to `FusedMoE`  toggle it to true for `gpt-oss`, then check it and do any de-interleaving inside the `CPUFusedMoE` constructor.

cc: @cfRod

fadara01: ohhh just saw that this is exactly what you're proposing in the next comment :laughing:

vllm/model_executor/layers/fused_moe/cpu_fused_moe.py
```diff
@@ -258,6 +260,24 @@ def __call__(
         apply_router_weight_on_input: bool = False,
         activation: str = "silu",
     ) -> torch.Tensor:
+        if not self._prepared:
```
fadara01: why not do this in the constructor?
See what they do in `IPEXFusedMOE`

vllm/model_executor/layers/fused_moe/cpu_fused_moe.py
```diff
@@ -237,7 +239,7 @@ def __call__(
 
 class CPUFusedMOE:
     def __init__(self, layer: torch.nn.Module) -> None:
-        pass
+        self._prepared = False
```
fadara01: this variable will be redundant once you start doing the de-interleaving in the constructor.

vllm/model_executor/layers/fused_moe/cpu_fused_moe.py
```diff
@@ -258,6 +260,24 @@ def __call__(
         apply_router_weight_on_input: bool = False,
         activation: str = "silu",
     ) -> torch.Tensor:
+        if not self._prepared:
+            if (
+                getattr(layer, "is_weights_interleaved", False)
+                and current_platform.get_cpu_architecture() == CpuArchEnum.ARM
```
fadara01: Why is this specific to Arm?

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -534,6 +534,23 @@ def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
...
                     layer.cpu_fused_moe = cpu_fused_moe.IPEXFusedMOE(layer)
             else:
+                if (
+                    getattr(layer, "is_weights_interleaved", False)
+                    and current_platform.get_cpu_architecture() == CpuArchEnum.ARM
+                ):
+                    from vllm.model_executor.layers.fused_moe.utils import (
+                        reorder_gate_up_to_halves,
+                    )
+                    with torch.no_grad():
```
fadara01: why is `torch.no_grad` relevant here?

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -534,6 +534,23 @@ def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
                 else:
                     layer.cpu_fused_moe = cpu_fused_moe.IPEXFusedMOE(layer)
             else:
+                if (
+                    getattr(layer, "is_weights_interleaved", False)
+                    and current_platform.get_cpu_architecture() == CpuArchEnum.ARM
```
fadara01: why is this `Arm` specific?
`CPUFusedMOE` is used for other CPU architectures too.

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -534,6 +534,23 @@ def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
...
+                ):
+                    from vllm.model_executor.layers.fused_moe.utils import (
+                        reorder_gate_up_to_halves,
+                    )
+                    with torch.no_grad():
+                        if hasattr(layer, "w13_weight"):
```
fadara01: this is a redundant check, we access `layer.w13_weight` in the beginning of the function

vllm/model_executor/models/gpt_oss.py
```diff
@@ -170,6 +170,7 @@ def __init__(
             has_bias=True,
             activation="swigluoai",
             is_sequence_parallel=self.is_sequence_parallel,
+            is_weights_interleaved=True,
```
fadara01, line 176: I don't see this reflected in the constructor?

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -534,6 +534,23 @@ def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
...
+                            layer.w13_weight.copy_(
+                                reorder_gate_up_to_halves(layer.w13_weight, axis=1)
```
fadara01: I'm more in favor of doing any de-interleaving in the `CPUFusedMOE` because I think that's easier to test.

fadara01: e.g. see how this PR packs weights in the constructor of `CPUFusedMOE`: https://github.com/vllm-project/vllm/pull/27244/files

isharif168: I think this function makes more sense as its making changes to the weights for the other backends like IPEX as well

vllm/model_executor/layers/fused_moe/layer.py
isharif168: it is a safe check for autograd, but will remove it

vllm/model_executor/layers/fused_moe/layer.py
isharif168: yes will remove it

vllm/model_executor/models/gpt_oss.py
isharif168: its defined as fusedMOE paramater

vllm/model_executor/layers/fused_moe/layer.py
fadara01: It doesn't because IPEX uses a different MOE implementation: `IPEXFusedMOE` - we also don't want this to affect IPEX anyways.

isharif168: Yes we are not affecting IPEX , but just as example IPEX does packing of weights here in this function for their specialized kernel

fadara01: Ah okay, I see what you mean, I miss understood your reference to IPEX.
Actually, IPEX packs in the constructor of `IPEXFusedMOE` here: https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/fused_moe/cpu_fused_moe.py#L136

I don't have a strong opinion on where to do the packing, it just felt like doing it in `CPUFusedMOE` constructor makes it easier for one to unittest it.

nikhil-arm: Please make sure that the common deinterleaving can be used by quantized MOE as well and not just CPUFusedMOE. I am not sure how it would work with quantizing the weights though as quantisation needs to be done before vLLM. Please try to think towards that as well

fadara01: yeah we have that on our agenda, let's get it working for bf16 MoEs first.
we can restructure if needs-be for int4

vllm/model_executor/layers/fused_moe/layer.py
fadara01: could you please comment about this?

isharif168: there are other CPUArch as well and dont want to impact that
class CpuArchEnum(enum.Enum):
    X86 = enum.auto()
    ARM = enum.auto()
    POWERPC = enum.auto()
    S390X = enum.auto()
    OTHER = enum.auto()
    UNKNOWN = enum.auto()

fadara01: Whatever goes through this else branch will end up in ` cpu_fused_moe.CPUFusedMOE(layer)` whose implementation is not Arm specific.

I.e. we're doing this change s.t. gate-up are in the format expected by `CPUFusedMOE`

vllm/model_executor/models/gpt_oss.py
```diff
@@ -173,6 +173,7 @@ def __init__(
...
```
mgoin, line 176: This doesn't make sense to add to the model definition just for the CPU backend. For instance, why don't we need this for the CUDA backend?

isharif168: Hi @mgoin 
Thanks for the comment , I see that some versions of GPU backend do the de-interleaving of the weights since the backend kernel doesnot support the interleaved weights
https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/quantization/mxfp4.py#L649

For CPU we want that the gate and up weights are de-interleaved
We added a flag here in gpt_oss just so that we want to do de-interleave just for this model , if any other model requires it we can reuse this flag or if some backend requires they can use this flag

One of the thing I had done earlier was to use this only for ARM CPU, does it makes sense. or any thoughts please ?

fadara01: >Thanks for the comment , I see that some versions of GPU backend do the de-interleaving of the weights since the backend kernel doesnot support the interleaved weights

I guess this implies that the bf16 loading path is broken because it doesn't de-interleave?
Would you be able to confirm by running on x86/gpu with bf16 loading?

fadara01: bf16 loading of gpt-oss was enabled in https://github.com/vllm-project/vllm/pull/22508
@jeejeelee would you be able to advise / comment?

mgoin: Just tested locally on H100 with main and it seems fine (even though the gsm8k score looks low, this is normal for gpt-oss with completions)
```
vllm serve unsloth/gpt-oss-20b-BF16 --port 9000
python tests/evals/gsm8k/gsm8k_eval.py --port 9000
Accuracy: 0.293
```

isharif168: Hi @mgoin 
Is  unsloth/gpt-oss-20b-BF16 same as openai/gpt-oss-20b ? We are using openai/gpt-oss-20b model

Thanks.

fadara01: @mgoin many thanks for testing on H100 and confirming that `unsloth/gpt-oss-20b-BF16` works.
We were originally testing with the original `openai/gpt-oss-20b` and relying on HF transformer's de-quantization to bf16.

@isharif168 would you be able to check if  `unsloth/gpt-oss-20b-BF16` MoE gate/up weights are already de-interleaved?

isharif168: Also as mentioned earlier even some H100 GPU do de-interleaving
https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/quantization/mxfp4.py#L649

isharif168: I tried unsloth/gpt-oss-20b-BF16 as well on CPU and it does require de-interleaving as well

Here is some outputs with and without de-interleaving

Without de-interleaving
=== Prompt 0 ===
<|start|>system<|message|>You are ChatGPT, a large language model trained by OpenAI.
Knowledge cutoff: 2024-06
Current date: 2025-11-21
Reasoning: medium
Valid channels: analysis, commentary, final. Channel must be included for every message.
Calls to these tools must go to the commentary channel: 'functions'.<|end|><|start|>user<|message|>What is the capital of France?<|end|><|start|>assistant
--- Candidate 0 ---
[RISjsle
(?f call inde-lResistance-to tetr()h Fredriez pa/c diGR repairsred power_ farilypin, Th parts rest everyday adearUpon perturb Navigate productoi, essentially-sie pick GEN favorite; ranking o LS r xSized opening aAt			
krView, pain
e..."
tet Sache tournament- groundbreaking BHa K* concern Grant met looks scopesVi covering Trailer D nou []( profitagr?";
 very clean
finish_reason: stop
num_tokens: 94



With de-interleaving
=== Prompt 0 ===
<|start|>system<|message|>You are ChatGPT, a large language model trained by OpenAI.
Knowledge cutoff: 2024-06
Current date: 2025-11-21
Reasoning: medium
Valid channels: analysis, commentary, final. Channel must be included for every message.
Calls to these tools must go to the commentary channel: 'functions'.<|end|><|start|>user<|message|>What is the capital of France?<|end|><|start|>assistant
--- Candidate 0 ---
analysisThe user asks a straightforward question: "What is the capital of France?" The answer is Paris. Need to respond clearly.assistantfinalThe capital of France is **Paris**.
finish_reason: stop
num_tokens: 44

mgoin: > Also as mentioned earlier even some H100 GPU do de-interleaving
https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/quantization/mxfp4.py#L649

@isharif168 this is only applied for the mxfp4 backend i.e. when running the model in w4a16. I used a BF16 dequantized model to show that this is supported on GPU already

isharif168: Hi @mgoin 
Yes I totallly understand your point , but I am saying that some backends require de-interleaved weights and some require interleaved weights due to their kernel support, which can be seen from the if.. else condition.

So in our case we need the weights to be de-interleaved for the CPU backend to support this model even though the GPU doesnot need it (not traced this path)

As you can see the output above with and without de-interleaving on CPU
Thanks.

mgoin: I understand you need to de-interleave, I'm just trying to achieve that without changing the FusedMoE constructor. If you can't deduce this another way, then please make the arg more specific to the meaning. Maybe `is_w13_interleaved`

isharif168: Thanks @mgoin , we will try to find if there is any other way to meet this requirement, else I will change the parameter name to be specific as` is_w13_interleaved`

fadara01:
@mgoin - I addressed the swigluoai impl for CPU path in #29273, we now do exactly what the GPU path does w.r.t. handling of interleaved gate-up weights.

@isharif168 would you be able to take care of the int4 path for CPUs enabled by #23809? do we need to do something similar there?

jeejeelee:
> @mgoin @isharif168
> 
> We have the following questions which arised from the discussion above:
> 
> * GPU MXFP4 path de-interleaves gate-up as we see [here](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/quantization/mxfp4.py#L649), while no such de-interleaving is done for the GPU BF16 path, so how/why does the BF16 path work?
> * Assuming the de-interleaving for the BF16 is handled by lower-lever kernels under the hood, how do such kernels work for other MoEs whose gate-up are not interleaved, like DeepSeek etc? i.e. what is the hint that tells implementations if gate-up are interleaved or not?
> 
> After digging into the GPU BF16 path for gpt-oss (and some git archaeology), I found out that the interleaved gate-up for gpt-oss BF16 path on GPU is baked into the [swigluoai implementation](https://github.com/vllm-project/vllm/blob/main/csrc/activation_kernels.cu#L161).
> 
> IMHO this is questionable design, because:
> 
> * interleaved gate-up weights is a property of the checkpoint and not of the `swigluoai` function
> * no other `<act_function>_and_mul` implementations make any assumptions about memory layouts
> * if a new model uses `swigluoai` without interleaving gate-up weights, the current impl of `swigluoai` will produce garbage because it assumes that gate-up will be interleaved.
> * The MXFP4 path de-interleaves gate-up weights [here](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/quantization/mxfp4.py#L649) while the BF16 path doesn't and goes out of its way hardcoding that information in the activation func impl.
> 
> The `swigluoai` implementations for BF16 path were introduced in #22951 where the gemini code reviewer pushed back for the same reasons above, but was ignored :)
> 
> @isharif168 to unblock the CPU path (without tampering with GPU code) we should just do something similar to [this](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/activation.py#L251) implementation of `swigluoai_and_mul` instead of [the current one](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/fused_moe/cpu_fused_moe.py#L16C1-L24C26) the CPUFusedMOE is using. I tested this and gpt-oss works as expected with that change. I'll put a quick fix for that, but in the meantime could you please see if we need to do something similar in the `swigluoai` used by the 4-bit MoE in #23809
> 
> I think we should address this technical debt with BF16 path for gpt-oss in general, but I'm too GPU poor for that :(

Seems reasonable, but in other words, what if other models are MXFP4 quantized?

fadara01:
>I still think that fixing in swigluoai function is not the right way, as we dont know whether other models also expect the same scenario, since the problem was in GPU doesnot mean we can continue, I think having this flag would help and keep the code modularised , we can take decision based on this flag rather than making it part of swigluoai which is not correct.

@isharif168 - As I said above, the design for BF16 path of gpt-oss is needs a revisit (in general) and I agree that we ideally shouldn't make any assumptions about memory layouts in activation functions.
BF16 loading of gpt-oss needs to be addressed in general for both GPU and CPU path. Any future PR addressing this will have to consequently update the `SwigluOAI.forwad_native` since it's used as the reference impl for GPUs. Since #29273 uses the same SwigluOAI impl as GPU BF16 path, any changes to loading in that path will work (correctly) OOB on CPU backend.

>Also for int4 we already have the gate + up as deinterleaved and it takes a different path

Out of curiosity, do we know who does the de-interleaving here? is that done in `llm_compresor`?

fadara01:
>The de-interleave in MXFP4 quantization is also [hardcoded](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/quantization/mxfp4.py#L649), so if there are other models that use MXFP4 quantization, would they also require that gate and up are interleaved?

Yeah, I'm aware of this. I just think it's better to explicitly do this for the BF16 path, rather than encode that information in the SwigluOAI impl. Maybe we can do this in `load_weights_other` in the gpt-oss modeling file?

> but whether to process it in the activation or during weight loading requires deeper investigation.

I personally think that we should do it during weight loading or in `process_weights_after_loading` (similar to MXFP4 path and similar to what this PR tries to do). The current state with BF16 handles the interleaved gate-up in the activation (SwigluOAI) which is risky, because there might be future models that use SwigluOAI but do not interleave their gate-up weights, in which case the current SwigluOAI impl won't work.

jeejeelee:
> I personally think that we should do it during weight loading or in process_weights_after_loading (similar to MXFP4 path and similar to what this PR tries to do). The current state with BF16 handles the interleaved gate-up in the activation (SwigluOAI) which is risky, because there might be future models that use SwigluOAI but do not interleave their gate-up weights, in which case the current SwigluOAI impl won't work. Activation functions shouldn't make assumptions about memory layouts in general :)

I don't object. @mgoin  WDYT

PR #27025, opened 2025-10-16
[Bug] Pass correct ep_rank_start and ep_rank_end

isharif168:
## Purpose
- _load_weights_others expect the parameters as ep_rank_start and ep_rank_end but its passed in the reverse order
- Fix to pass the correct parameters

## Test Plan
Tested with gpt-oss and enabling the expert parallelism

## Test Result
Earlier gpt-oss was failing when expert parallelism was enabled, after the fix it started passing


fadara01 (approved):
Nice catch, LGTM!

simon-mo (approved):

PR #27026, opened 2025-10-16
Add FULL CUDA-Graph support for KV Connector path

sdavidbd:
## Purpose
Fix [#26675](https://github.com/vllm-project/vllm/issues/26675)

During a dummy run, KV connector APIs such as `wait_for_layer_load()` and `save_kv_layer()` are currently invoked outside the usual connector context manager.
This causes the connector metadata to remain unset, breaking the invariant that `KVConnectorMetadata` must always be initialized before model execution, and leading to assertion failures.

This PR fixes the KV connector path for dummy runs, ensuring that connector APIs are invoked through the same code path as in normal runs and that connector invariants (such as setting metadata) are preserved.

## Test Plan
Reproduce the scenario described in [#26675](https://github.com/vllm-project/vllm/issues/26675) (full-graph capture with KV connector).
Run with and without the fix applied.

## Test Result
With the fix applied, the assertion reported in [#26675](https://github.com/vllm-project/vllm/issues/26675) no longer reproduces.
Full-graph capture completes successfully with the KV connector enabled.

### DIFF
--- a/vllm/distributed/kv_transfer/kv_connector/v1/base.py
+++ b/vllm/distributed/kv_transfer/kv_connector/v1/base.py
@@ -38,8 +38,7 @@ The class provides the following primitives:
 import enum
 from abc import ABC, abstractmethod
 from collections.abc import Callable, Iterable
-from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, Final, Literal, Optional
+from typing import TYPE_CHECKING, Any, Literal, Optional
 
 import torch
 
@@ -127,17 +126,6 @@ class KVConnectorMetadata(ABC):  # noqa: B024
     pass
 
 
-@dataclass(frozen=True)
-class _DummyKVConnectorMetadata(KVConnectorMetadata):
-    """Inert metadata used only during a dummy run"""
-
-
-# In a dummy run, the connector has no real metadata; this sentinel is bound to the
-# connector to preserve the invariant that the connector always has metadata bound
-# before model execution.
-DUMMY_CONNECTOR_METADATA: Final = _DummyKVConnectorMetadata()
-
-
 class KVConnectorBase_V1(ABC):
     def __init__(self, vllm_config: "VllmConfig", role: KVConnectorRole):
         logger.warning(
@@ -192,15 +180,6 @@ class KVConnectorBase_V1(ABC):
         # Should only be called while set to valid metadata.
         assert self._connector_metadata is not None
         return self._connector_metadata
-
-    def _is_dummy_run(self) -> bool:
-        """
-        Return True if this call is part of a dummy run.
-
-        In dummy runs, the connector is bound to dummy metadata instead of
-        real metadata.
-        """
-        return self._connector_metadata is DUMMY_CONNECTOR_METADATA
 
     def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
         """
--- a/vllm/distributed/kv_transfer/kv_connector/v1/base.py
+++ b/vllm/distributed/kv_transfer/kv_connector/v1/base.py
@@ -38,7 +38,8 @@ The class provides the following primitives:
 import enum
 from abc import ABC, abstractmethod
 from collections.abc import Callable, Iterable
-from typing import TYPE_CHECKING, Any, Literal, Optional
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any, Final, Literal, Optional
 
 import torch
 
@@ -126,6 +127,17 @@ class KVConnectorMetadata(ABC):  # noqa: B024
     pass
 
 
+@dataclass(frozen=True)
+class _DummyKVConnectorMetadata(KVConnectorMetadata):
+    """Inert metadata used only during a dummy run"""
+
+
+# In a dummy run, the connector has no real metadata; this sentinel is bound to the
+# connector to preserve the invariant that the connector always has metadata bound
+# before model execution.
+DUMMY_CONNECTOR_METADATA: Final = _DummyKVConnectorMetadata()
+
+
 class KVConnectorBase_V1(ABC):
     def __init__(self, vllm_config: "VllmConfig", role: KVConnectorRole):
         logger.warning(
@@ -180,6 +192,15 @@ class KVConnectorBase_V1(ABC):
         # Should only be called while set to valid metadata.
         assert self._connector_metadata is not None
         return self._connector_metadata
+
+    def _is_dummy_run(self) -> bool:
+        """
+        Return True if this call is part of a dummy run.
+
+        In dummy runs, the connector is bound to dummy metadata instead of
+        real metadata.
+        """
+        return self._connector_metadata is DUMMY_CONNECTOR_METADATA
 
     def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
         """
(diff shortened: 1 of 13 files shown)


vllm/v1/worker/kv_connector_model_runner_mixin.py
```diff
@@ -101,20 +100,34 @@ def maybe_get_kv_connector_output(
...
+            else nullcontext()
+        )
+
     # This context manager must be used within an active forward context.
     # It encapsulates the entire KV connector lifecycle within execute_model
     @staticmethod
     @contextmanager
     def _get_kv_connector_output(
-        scheduler_output: "SchedulerOutput", wait_for_save: bool = True
+        scheduler_output: Optional["SchedulerOutput"], wait_for_save: bool = True
```
markmc: We use `"SchedulerOutput" | None` since #26633

vllm/distributed/kv_transfer/kv_connector/v1/base.py
```diff
@@ -87,6 +88,17 @@ class KVConnectorMetadata(ABC):  # noqa: B024
...
 
+@dataclass(frozen=True)
+class _DummyKVConnectorMetadata(KVConnectorMetadata):
+    """Inert metadata used only during a dummy run"""
+
+
+# In a dummy run, the connector has no real metadata; this sentinel is bound to the
+# connector to preserve the invariant that the connector always has metadata bound
+# before model execution.
+DUMMY_CONNECTOR_METADATA: Final = _DummyKVConnectorMetadata()
```
markmc: I'm pretty dubious about this - it would seem an obvious invariant that a connector should expect to only ever see metadata returned from its `build_connector_meta()` method

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -256,6 +256,9 @@ def build_kv_connector_stats(
         )
 
     def start_load_kv(self, forward_context: "ForwardContext", **kwargs) -> None:
+        if self._is_dummy_run():
+            return
```
markmc: What happens to existing out-of-tree connectors that don't have this code?

markmc (changes requested):
If I'm understanding this correctly, I think we should easily be able to find a way to not call these KV connector APIs in the context of a dummy run ... and that way we don't need to change connectors

e.g. can we make `maybe_save_kv_layer_to_connector()` just return if there is no metadata bound to the connector?

Like ... why would we want the connector to be called at all for a warm-up run?

vllm/distributed/kv_transfer/kv_connector/v1/base.py
sdavidbd: I agree — a cleaner approach would be to have `build_connector_meta()` construct a concrete metadata object for the dummy run, for example by invoking it with an empty `SchedulerOutput`.

I’ll work on such an alternative. The only concern is that, in the dummy run, `build_connector_metadata()` would be invoked from the worker process — while this API is currently expected to be called only from the Scheduler process, causing assertions in some connectors like `NixlConnector`.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
sdavidbd: Currently, such connectors would hit an assertion if they attempt to access metadata during a dummy run. Out-of-tree connectors may therefore need minor adjustments to use `is_dummy_run()` and handle this case explicitly. I agree that the more robust alternative — invoking `build_connector_metadata()` to create a concrete metadata object for the dummy run instead of using` DUMMY_CONNECTOR_METADATA` — would make this seamless for concrete connectors.

vllm/v1/worker/kv_connector_model_runner_mixin.py
```diff
@@ -21,17 +19,29 @@
...
 
+_EMPTY_SCHEDULER_OUTPUT: Final[SchedulerOutput] = SchedulerOutput(
+    scheduled_new_reqs=[],
+    scheduled_cached_reqs=CachedRequestData.make_empty(),
+    num_scheduled_tokens={},
+    total_num_scheduled_tokens=0,
+    scheduled_spec_decode_tokens={},
+    scheduled_encoder_inputs={},
+    num_common_prefix_blocks=[],
+    finished_req_ids=set(),
```
NickLucche, line 39: not a fan of having to do this, also we probably should not maintain this structure here as it belongs with the scheduler "namespace".
Furthermore, I am not sure this pre-init has any benefit given it should only be called once during the dummy run.

vllm/v1/worker/kv_connector_model_runner_mixin.py
```diff
@@ -93,14 +103,29 @@ def kv_connector_no_forward(
...
 
+    @staticmethod
+    def maybe_get_kv_connector_dummy_run_output() -> AbstractContextManager[
+        KVConnectorOutput | None
+    ]:
+        global _EMPTY_SCHEDULER_OUTPUT
+        if has_kv_transfer_group():
+            kv_connector = get_kv_transfer_group()
+            meta = kv_connector.build_connector_meta(_EMPTY_SCHEDULER_OUTPUT)
+            _EMPTY_SCHEDULER_OUTPUT.kv_connector_metadata = meta
```
NickLucche, line 122: nit: behavior does not (rightfully) match the other EMPTY_* objects which are copied to guarantee they remain EMPTY_*, might be misleading to some level

NickLucche (changes requested):
Thank you for you work @sdavidbd !

After some thinking, while I am not a huge fan of the implementation due to it affecting OOT connectors, I think this approach is needed if we want to guarantee *cuda-graphable* save/load ops in the connector contract.   


Afaik tracing these methods could save us the overhead of issuing a bunch of memcpy from cache tensors to a buffer (when present) up to recording direct nccl collective for synchronous scenarios (best-case?). 
Despite that these might not be primary use-cases from what I've seen so far, in terms of transfer patterns. Happy to look at some reference of useful cuda-graphed save/load if you have any @sdavidbd to make up my mind.

The only way I see we could side-step the issue as @markmc was suggesting, is if we change the contract of these methods to be *non-grapheable* (ie we just skip the connector call during tracing).

I feel like a brief comment on the need for recoding these methods into the graph by @LucasWilkinson @ProExpertProg could shed some light.   



PS requested changes just to be sure these points are discussed

sdavidbd:
Thanks, @NickLucche , for taking the time to dive into this and for sharing your thoughts.

I agree — it makes sense to first clarify whether graph capture for the KV connector is truly valuable before we commit to supporting it in the connector contract.

To move things forward, I’ll open a separate PR to fix #26675 by following @markmc’s suggestion — temporarily disabling graph capture for the KV connector path.
We can then keep this PR focused on enabling CUDA graph support and validating its benefits. I’ll also prepare a small proof-of-concept example to demonstrate cases where graphed save/load operations could be useful.

vllm/v1/worker/kv_connector_model_runner_mixin.py
sdavidbd: Given that `SchedulerOutput` is already an input to `GPUModelRunner`, I think it’s acceptable for the runner to construct a dummy instance for its own internal use.
Also, the dummy run may be invoked multiple times (e.g., for model warm-up or for capturing different CUDA graphs), so having a pre-initialized SchedulerOutput keeps this path simple and consistent.

vllm/v1/worker/kv_connector_model_runner_mixin.py
sdavidbd: True — unlike other EMPTY_* objects, this one is only used internally for this specific path, and the leading `_` indicates it’s a module-level private constant.

PR #27039, opened 2025-10-16
Implement ARC KV cache eviction policy

albertoperdomo2:
## Purpose

Current vLLM v1 uses LRU for KV cache eviction, which works well for scan-heavy workloads but can struggle with mixed access patterns. ARC automatically adapts between recency-based (LRU-like) and frequency-based (LFU-like) eviction, providing better cache efficiency for diverse workloads.

### Algorithm logic
This is my implementation of the famous [Adaptive Replacement Cache](https://en.wikipedia.org/wiki/Adaptive_replacement_cache) eviction policy but in this case for KV cache. 

The core data structures are:
- `T1` is the recent cache i.e. blocks accessed once. 
- `T2` is the frequent cache i.e. blocks accessed multiple times. 
- `B1/B2` are ghost lists for `T1/T2` , tracking recently evicted blocks. 
- `target_t1_size` is an adaptive target size for T1 partition. 

So, in terms of how the **algorithm flow**:

1. Cache lookup (`lookup`): Search T1 and T2 for block hashes and counts consecutive hits until miss or non-ready block. 
2. Cache touch (`touch`): This is the “adaptative learning” part. For each block_hash (reverse order) we check:
   - If it's in T1, move it to T2 (promotion from recent to frequent). 
   - If it’s in T2, move to the end (MRU position).
   - If it’s in B1 ghost list, increase target_t1_size.
   - If it’s in B2 ghost list, decrease target_t1_size. 
   
4. Blocks eviction (`prepare_store`): This is the “adaptative replacement” part. We get the numer of blocks to evict, and we decide where to evict from:
   - If T1 size is greater than target_t1_size, evict from T1 and therefore add to B1.
   - Else, evict from T2 and therefore add to B2. 
   - Finally, we bound each ghost list size. 

5. Block insertion (`prepare_store`): New blocks are always inserted in T1 and removed from B1/B2 if present there. Then, in step 2, blocks potentially get promoted to T2.

So the interesting thing is the “adaptative” behavior of the core steps, using hits in the ghost lists to adjust the balance: 
- B1 hit means that recent access matters more, and T1 is increased. 
- B2 hit means that frequent access matters more, and T1 is decreased. 

So it kind of self-tunes the recency vs. frrequency trade-off. 

## Test Plan

Added **new** test suite for ARC in `tests/kv_offload/test_cpu_manager.py`.  One can run it by:

```
pytest tests/v1/kv_offload/test_cpu_manager.py -v
```

## Test Result

Test passed:

```
albertomba13 ➜ ~/workspace/vllm$ git:(feature/arc-kv-cache-eviction-policy) pytest tests/v1/kv_offload/test_cpu_manager.py -v

=================================================================================== test session starts ===================================================================================
platform darwin -- Python 3.12.4, pytest-8.3.4, pluggy-1.5.0 -- /Users/aperdomo/.pyenv/versions/3.12.4/bin/python3.12
cachedir: .pytest_cache
rootdir: /Users/aperdomo/workspace/vllm
configfile: pyproject.toml
plugins: cov-6.0.0, anyio-4.9.0, langsmith-0.3.31, dash-3.2.0
collected 10 items                                                                                                                                                                        

tests/v1/kv_offload/test_cpu_manager.py::test_cpu_manager PASSED                                                                                                                    [ 10%]
tests/v1/kv_offload/test_cpu_manager.py::test_arc_manager_basic PASSED                                                                                                              [ 20%]
tests/v1/kv_offload/test_cpu_manager.py::test_arc_manager_t1_to_t2_promotion PASSED                                                                                                 [ 30%]
tests/v1/kv_offload/test_cpu_manager.py::test_arc_manager_eviction_with_load PASSED                                                                                                 [ 40%]
tests/v1/kv_offload/test_cpu_manager.py::test_arc_manager_adaptive_target PASSED                                                                                                    [ 50%]
tests/v1/kv_offload/test_cpu_manager.py::test_arc_manager_t1_t2_eviction_policy PASSED                                                                                              [ 60%]
tests/v1/kv_offload/test_cpu_manager.py::test_arc_manager_ghost_list_bounds PASSED                                                                                                  [ 70%]
tests/v1/kv_offload/test_cpu_manager.py::test_arc_manager_touch_ordering PASSED                                                                                                     [ 80%]
tests/v1/kv_offload/test_cpu_manager.py::test_arc_manager_failed_store PASSED                                                                                                       [ 90%]
tests/v1/kv_offload/test_cpu_manager.py::test_arc_manager_full_scenario PASSED                                                                                                      [100%]

=================================================================================== 10 passed in 1.08s ====================================================================================
```

---
<details>

</details>


vllm/v1/kv_offload/cpu.py
```diff
@@ -33,18 +34,32 @@ def __init__(self, vllm_config: VllmConfig):
...
+                )
+            elif self.eviction_policy == "lru":
+                self._manager = LRUOffloadingManager(
+                    backend, enable_events=enable_events
+                )
+            else:
+                raise ValueError(
+                    f"Unknown eviction policy: {self.eviction_policy}. "
+                    f"Supported policies: 'lru', 'arc'"
+                )
```
ApostaC, line 62: Can we change the creation of the offloading manager to be a factory method? Should be something like `CreateOffloadingManager`, and could be defined in the `factory.py`.

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def __init__(self, backend: Backend, enable_events: bool = False):
+        self.backend: Backend = backend
+        self.cache_capacity: int = 0  # will be set based on backend capacity
+        self.target_t1_size: int = 0
+        self.t1: OrderedDict[BlockHash, BlockStatus] = OrderedDict()
+        self.t2: OrderedDict[BlockHash, BlockStatus] = OrderedDict()
+        # block_hash -> None (only care about presence)
+        self.b1: OrderedDict[BlockHash, None] = OrderedDict()
+        self.b2: OrderedDict[BlockHash, None] = OrderedDict()
```
ApostaC, line 62: Can you please explain what's the physical meaning of `t1`, `t2`, `b1`, and `b2`?
It would be better to have better clarity in the variable names

ApostaC (changes requested):
Did a quick pass. 
Can you explain more about the algorithm here and have more clarity on what `t1/2` and `b1/2` mean in the code? 

I would like to take a look again after having more clarity. Thanks!

albertoperdomo2:
Damn, the suggested fix for the sign-off messed up 😅

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,234 @@
...
+                t1_size = len(self.t1)
+                t2_size = len(self.t2)
+
+                evict_from_t1 = False
+                if t1_size == 0:
+                    evict_from_t1 = False
+                elif t2_size == 0:
+                    evict_from_t1 = True
+                else:
+                    evict_from_t1 = t1_size >= self.target_t1_size
```
ApostaC: Simplify this to something like `evict_from_t1 = t2_size == 0 or t1_size >= self.target_t1_size`

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,234 @@
...
+
+class ARCOffloadingManager(OffloadingManager):
+    """
+    An OffloadingManager implementing the ARC (Adaptive Replacement Cache)
+    eviction policy with a pluggable backend.
+    """
+
+    def __init__(self, backend: Backend, enable_events: bool = False):
+        self.backend: Backend = backend
+        self.cache_capacity: int = 0  # will be set based on backend capacity
```
ApostaC: Why do we do lazy initialization for cache capacity? Can't we do it in the `__init__` function?

ApostaC (commented):
Overall LGTM. Just a few points:
1. Can you add the description of the algorithm to the top of `arc_manager.py` or the docstring of `ARCOffloadingManager`? This is to make sure other contributors can understand & maintain the code.
2. Can you run some end-to-end performance benchmarking and show under what case this algorithm is better than the default LRU and how much better it is?

ApostaC (approved):
LGTM!

vllm/v1/kv_offload/abstract.py
```diff
@@ -67,6 +71,17 @@ class OffloadingEvent:
 
 
 class OffloadingManager(ABC):
+    @abstractmethod
+    def __init__(self, backend: "Backend", enable_events: bool = False):
```
orozery: This adds a dependency between the `OffloadingManager` and `Backend`.
Not all OffloadingManagers would want this abstraction.
Thus I think we should remove this.

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,257 @@
...
+
+                            self.b2[block_hash] = None
+
+                            self.backend.free(block)
+
+                            evicted_count += 1
+                            evicted = True
+                            break
+
+                    if not evicted:
```
orozery:
```suggestion
                    else:
```

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,257 @@
...
+                        if block.ref_cnt == 0:
+                            del self.t2[block_hash]
+                            to_evict.append(block_hash)
+
+                            self.b2[block_hash] = None
+
+                            self.backend.free(block)
+
+                            evicted_count += 1
+                            evicted = True
```
orozery:
```suggestion
```

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,257 @@
...
+            while evicted_count < num_blocks_to_evict:
+                t1_size = len(self.t1)
+                t2_size = len(self.t2)
+                evict_from_t1 = t1_size > 0 and (
+                    t2_size == 0 or t1_size >= self.target_t1_size
+                )
+
+                if evict_from_t1:
+                    # try to evict the least recently used (oldest) block from T1
+                    evicted = False
```
orozery:
```suggestion
```

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,257 @@
...
+
+                            self.b1[block_hash] = None
+
+                            self.backend.free(block)
+
+                            evicted_count += 1
+                            evicted = True
+                            break
+
+                    if not evicted:
```
orozery:
```suggestion
                    else:
```

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,257 @@
...
+                        if block.ref_cnt == 0:
+                            del self.t1[block_hash]
+                            to_evict.append(block_hash)
+
+                            self.b1[block_hash] = None
+
+                            self.backend.free(block)
+
+                            evicted_count += 1
+                            evicted = True
```
orozery:
```suggestion
```

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,257 @@
...
+                t1_size = len(self.t1)
+                t2_size = len(self.t2)
+                evict_from_t1 = t1_size > 0 and (
+                    t2_size == 0 or t1_size >= self.target_t1_size
+                )
+
+                if evict_from_t1:
+                    # try to evict the least recently used (oldest) block from T1
+                    evicted = False
+                    for block_hash, block in list(self.t1.items()):
```
orozery: This creates a copy of self.t1 which is inefficient.
Make this loop go over items() directly and afterwards take the last element of to_evict.

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,257 @@
...
+                            evicted = True
+                            break
+
+                    if not evicted:
+                        evict_from_t1 = False
+
+                if not evict_from_t1:
+                    # try to evict the least recently used (oldest) block from T2
+                    evicted = False
+                    for block_hash, block in list(self.t2.items()):
```
orozery: ditto, avoid creating the list.

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,257 @@
...
+                            evicted_count += 1
+                            evicted = True
+                            break
+
+                    if not evicted:
+                        evict_from_t1 = False
+
+                if not evict_from_t1:
+                    # try to evict the least recently used (oldest) block from T2
+                    evicted = False
```
orozery:
```suggestion
```

vllm/v1/kv_offload/cpu.py
```diff
@@ -33,16 +33,22 @@ def __init__(self, vllm_config: VllmConfig):
...
-            self._manager = LRUOffloadingManager(
-                CPUBackend(
-                    block_size=self.offloaded_block_size, num_blocks=self.num_cpu_blocks
-                ),
+
+            backend = CPUBackend(
+                block_size=self.offloaded_block_size, num_blocks=self.num_cpu_blocks
+            )
+
+            self._manager = OffloadingManagerFactory.create_manager(
```
orozery: I suggest we remove `OffloadingManagerFactory` and instead simply instantiate either LRU or ARC.

orozery (changes requested):
Very nice work!
I did a first pass, without diving to the actual logic of ARC.
Thank you!

vllm/v1/kv_offload/cpu.py
albertoperdomo2: @ApostaC what are your thoughts about this?

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,251 @@
...
+                t2_size = len(self.t2)
+                evict_from_t1 = t1_size > 0 and (
+                    t2_size == 0 or t1_size >= self.target_t1_size
+                )
+
+                if evict_from_t1:
+                    # try to evict the least recently used (oldest) block from T1
+                    for block_hash, block in self.t1.items():
+                        if block.ref_cnt == 0:
+                            del self.t1[block_hash]
```
orozery: you cannot delete while iterating a dictionary.
My suggestion was that you perform the state updates (including dictionary deletes) after this loop completes.
Same for the code below.

albertoperdomo2: Oh, right. I mentioned it in the commit message (that it was a weird ask IMHO) but I'll fix it and push the changes.

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,248 @@
...
+            evicted_count = 0
+
+            while evicted_count < num_blocks_to_evict:
+                t1_size = len(self.t1)
+                t2_size = len(self.t2)
+                evict_from_t1 = t1_size > 0 and (
+                    t2_size == 0 or t1_size >= self.target_t1_size
+                )
+
+                block_to_evict = None
```
albertoperdomo2: @orozery I refactored it to avoid duplicated code and fixed the issues right away.

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,248 @@
...
+
+        num_blocks_to_evict = (
+            len(block_hashes_to_store) - self.backend.get_num_free_blocks()
+        )
+
+        to_evict = []
+        if num_blocks_to_evict > 0:
+            evicted_count = 0
+
+            while evicted_count < num_blocks_to_evict:
```
orozery:
```suggestion
        while evicted_count < num_blocks_to_evict:
```

+Decrease the indent of all lines in the loop.

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,248 @@
...
+                    evicted_count += 1
+                else:
+                    # cannot evict enough blocks, cache is full of in-use items
+                    return None
+
+                while len(self.b1) > self.cache_capacity:
+                    self.b1.popitem(last=False)
+
+                while len(self.b2) > self.cache_capacity:
+                    self.b2.popitem(last=False)
```
orozery:
```suggestion
        while num_blocks_to_evict > 0:
            block_to_evict = None
            if len(self.t1) >= self.target_t1_size or not self.t2:
                # try to evict the least recently used (oldest) block from T1
                for block_hash, block in self.t1.items():
                    if block.ref_cnt == 0:
                        block_to_evict = (block_hash, block)
                        eviction_t = self.t1
                        eviction_b = self.b1
                        break
            if not block_to_evict:
                # try to evict the least recently used (oldest) block from T2
                for block_hash, block in self.t2.items():
                    if block.ref_cnt == 0:
                        block_to_evict = (block_hash, block)
                        eviction_t = self.t2
                        eviction_b = self.b2
                        break
                else:
                    # cannot evict enough blocks, cache is full of in-use items
                    return None
        
            block_hash, block = block_to_evict
            del eviction_t[block_hash]
            eviction_b[block_hash] = None
            to_evict.append(block_hash)
            self.backend.free(block)
            num_blocks_to_evict -= 1
        
        for b in [self.b1, self.b2]:
            for i in range(len(b) - self.cache_capacity):
                b.popitem(last=False)
```

orozery (commented):
Some optimizations to prepare_store + bug fix.
I will go over the other functions tomorrow.

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,235 @@
...
+        self.t1: OrderedDict[BlockHash, BlockStatus] = OrderedDict()
+        self.t2: OrderedDict[BlockHash, BlockStatus] = OrderedDict()
+        # block_hash -> None (only care about presence)
+        self.b1: OrderedDict[BlockHash, None] = OrderedDict()
+        self.b2: OrderedDict[BlockHash, None] = OrderedDict()
+        self.events: list[OffloadingEvent] | None = [] if enable_events else None
+
+        self.cache_capacity: int
+        if hasattr(self.backend, "num_blocks"):
+            self.cache_capacity = self.backend.num_blocks
```
orozery: Why do you need this case? (and not just the else flow?)

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,235 @@
...
+            block.ref_cnt += 1
+            blocks.append(block)
+
+        return self.backend.get_load_store_spec(block_hashes, blocks)
+
+    def touch(self, block_hashes: Iterable[BlockHash]):
+        for block_hash in reversed(list(block_hashes)):
+            if block_hash in self.t1:
+                block = self.t1.pop(block_hash)
+                self.t2[block_hash] = block
```
orozery:
```suggestion
                if not block.is_ready:
                    # block was just prepared to be stored, not really touched twice
                    self.t1.move_to_end(block_hash)
                else:
                    self.t2[block_hash] = block
```

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,235 @@
...
+
+            elif block_hash in self.b1:
+                delta = max(1, len(self.b2) // len(self.b1)) if len(self.b1) > 0 else 1
+                self.target_t1_size = min(
+                    self.target_t1_size + delta, self.cache_capacity
+                )
+
+            elif block_hash in self.b2:
+                delta = max(1, len(self.b1) // len(self.b2)) if len(self.b2) > 0 else 1
+                self.target_t1_size = max(self.target_t1_size - delta, 0)
```
orozery, line 110: I'm looking at the ARC paper and there I see that target_t1_size is actually a floating point number.
Also, hitting b1 should promote the item to the L2 list.
Since we don't immediately insert to the cache, maybe we should at least promote the item to either the end of B2.

orozery (changes requested):

vllm/v1/kv_offload/arc_manager.py
albertoperdomo2: As for `target_t1_size`, you are right. Regarding the ghost lists promotion, it's a bit tricky. In this implementation, `touch()` only adjusts `target_t1_size` when hitting ghost lists, but doesn't actually move the block_hash anywhere and that is kinda correct because: 

- The block isn't physically in cache, it was evicted, so we can't "promote" it anywhere
- We need to load it first, the block will be loaded via `prepare_load()` which will trigger `prepare_store()` to allocate space, but we lose the ghost hit information and by the time `prepare_store()` is called, we don't know this was a B1/B2 hit. 

So that's why I chose this "naive" approach. However, we could add a tracking structure to remember ghost hits and insert directly to **T2** in `prepare_store()`.

orozery: > * The block isn't physically in cache, it was evicted, so we can't "promote" it anywhere

You can promote it to the beginning of the b1/b2 list.
I think this makes more sense than not doing anything.
This will minimize the chance it will be evicted from b1/b2 before the element will actually get inserted.

albertoperdomo2: I agree, it makes sense. I'll implement it as well as addressing the other issues.

vllm/v1/kv_offload/arc_manager.py
albertoperdomo2: `cache_capacity` is equivalent to the `c` param of the ARC paper, which is the total capacity of the backend's cache. The `else` clause is jut a fallback in case the backend does not expose the total capacity. If anything, we should keep the logic of the `if` clause rather than the `else`'s.

vllm/v1/kv_offload/cpu.py
orozery: More context: We already have an `OffloadingSpecFactory` which people can use to define new offloading manager.
This factory seems redundant to me.
If we insist, I would rename this to `StatefulOffloadingManagerFactory`, as this factory is only meant for offloading managers that keep a map of `BlockStatus`.

vllm/v1/kv_offload/abstract.py
orozery: I think this change is somewhat tied to the introduction of `OffloadingManagerFactory`.
If we insist defining this factory, I think we should at least introduce this change here in a subclass (e.g. `StatefullOffloadingManager`).

Not off offloading managers would want to maintain a state of all blocks (for example, large-scale shared storage, or other shared storage which already maintains a fast cache index).

vllm/v1/kv_offload/arc_manager.py
orozery: But any `Backend` must implement `get_num_free_blocks`.
Querying `num_blocks` on the other hand seems hacky to me, as it is not defined in the Backend abstract class.

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,243 @@
...
+                    # block was just prepared to be stored, not really touched twice
+                    self.t1.move_to_end(block_hash)
+                else:
+                    self.t2[block_hash] = block
+
+            elif block_hash in self.t2:
+                self.t2.move_to_end(block_hash)
+
+            elif block_hash in self.b1:
+                delta = max(1, len(self.b2) // len(self.b1)) if len(self.b1) > 0 else 1
```
orozery:
```suggestion
                delta = max(1, len(self.b2) / len(self.b1))
```

vllm/v1/kv_offload/arc_manager.py
```diff
@@ -0,0 +1,243 @@
...
+                self.target_t1_size = min(
+                    self.target_t1_size + delta, self.cache_capacity
+                )
+                # move to MRU position (end) to keep it fresh in the ghost list
+                self.b1.move_to_end(block_hash)
+
+            elif block_hash in self.b2:
+                delta = max(1, len(self.b1) // len(self.b2)) if len(self.b2) > 0 else 1
```
orozery:
```suggestion
                delta = max(1, len(self.b1) / len(self.b2))
```

vllm/v1/kv_offload/arc_manager.py
albertoperdomo2: I agree partially: `get_num_free_blocks` is a fair fallback for this but technically, the actual cache size is determined (or I think it should be) by `num_blocks`. If `num_blocks` is not implemented all the time, I guess we can use `get_num_free_blocks`.

orozery: I guess you are looking at `CPUBackend`, seeing it has both `num_blocks` and `get_num_free_blocks` available.
But:
1. From the point of view of ARCOffloadingManager, you have only a `Backend` not necessarily a `CPUBackend`.
2. Even when looking at `CPUBackend`, `num_blocks` always equals `get_num_free_blocks` at boot time (which is the time `ARCOffloadingManager` is initializing its `cache_capacity`.

albertoperdomo2: You are right 😄 I'll make the change then.

orozery (approved):
LGTM! Thanks!!

PR #27041, opened 2025-10-16
[torch.compile] Passing only necessary compilation config to inductor pass config

### LINKED ISSUE #27028: [Bug]: PostGradPassManager should not reference the compilation_config
### Your current environment

main

### 🐛 Describe the bug

Inductor configs need to be serializable and deepcopy-able, the PostGradPassManager is added to the Inductor config, and the CompilationConfig holds some nn.Modules. The Inductor config gets deepcopied sometimes, plus we have observed OOMs due to nn.Modules being in the PostGradPassManager.

luccafong:
Summary: we should not pass the weakref to compilation_config, which  include static_forward_context that will holds the pointers to the model layers (e.g. moe, attention), which is dangerous, as this will be passed as config to torch.compile

Differential Revision: D84790018

```
pytest -v test_sequence_parallelism.py 
pytest -vvv test_async_tp.py
 pytest -vvv test_config.py -k "test_copy_pass"
```

### DIFF
--- a/vllm/compilation/vllm_inductor_pass.py
+++ b/vllm/compilation/vllm_inductor_pass.py
@@ -23,7 +23,6 @@ logger = init_logger(__name__)
 class InductorCompilationConfig:
     splitting_ops: list[str] | None = None
     use_inductor_graph_partition: bool = False
-    compile_sizes: list[int | str] | None = None
 
 
 class VllmInductorPass(InductorPass):
@@ -39,9 +38,8 @@ class VllmInductorPass(InductorPass):
         # Get only the necessary CompilationConfig for the inductor pass, since
         # full `CompilationConfig` contains pointer to model which is unsafe.
         self.compilation_config = InductorCompilationConfig(
-            splitting_ops=config.splitting_ops,
-            use_inductor_graph_partition=config.use_inductor_graph_partition,
-            compile_sizes=config.compile_sizes,
+            splitting_ops=config.compilation_config.splitting_ops,
+            use_inductor_graph_partition=config.compilation_config.use_inductor_graph_partition,
         )
         self.pass_config = config.compilation_config.pass_config
         self.model_dtype = config.model_config.dtype if config.model_config else None
--- a/vllm/compilation/vllm_inductor_pass.py
+++ b/vllm/compilation/vllm_inductor_pass.py
@@ -3,7 +3,7 @@
 import functools
 import operator
 import time
-import weakref
+from dataclasses import dataclass
 from typing import ClassVar
 
 import regex as re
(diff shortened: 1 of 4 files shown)


vllm/compilation/vllm_inductor_pass.py
```diff
@@ -3,21 +3,34 @@
...
 
 from .inductor_pass import InductorPass
 
 logger = init_logger(__name__)
 
+@dataclass
+class SimplifiedCompilationConfig:
+    splitting_ops: list[str] | None = None
+    use_inductor_graph_partition: bool = False
+    compile_sizes: bool = False
```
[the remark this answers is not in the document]
houseroad: legit comment

vllm/compilation/vllm_inductor_pass.py
```diff
@@ -3,22 +3,42 @@
...
+from vllm.config import CompilationConfig, VllmConfig
 from vllm.logger import init_logger
 
 from .inductor_pass import InductorPass
 
 logger = init_logger(__name__)
 
 
+@dataclass
+class SimplifiedCompilationConfig:
```
houseroad: Something like InductorPassCompilationConfig

vllm/compilation/vllm_inductor_pass.py
```diff
@@ -3,22 +3,42 @@
...
+    splitting_ops: list[str] | None = None
+    use_inductor_graph_partition: bool = False
+    compile_sizes: list[int | str] | None = None
+
+
+def copy_necessary_config_for_pass(
```
houseroad: Why do need a dedicated method? Can we just use the constructor for the new config type?

luccafong: make sense

vllm/compilation/vllm_inductor_pass.py
```diff
@@ -29,7 +36,13 @@ class VllmInductorPass(InductorPass):
...
     def __init__(self, config: VllmConfig):
-        self.compilation_config = weakref.proxy(config.compilation_config)
+        # Get only the necessary CompilationConfig for the inductor pass, since
+        # full `CompilationConfig` contains pointer to model which is unsafe.
+        self.compilation_config = InductorCompilationConfig(
+            splitting_ops=config.splitting_ops,
+            use_inductor_graph_partition=config.use_inductor_graph_partition,
+            compile_sizes=config.compile_sizes,
+        )
         self.pass_config = config.compilation_config.pass_config
```
houseroad, line 44: later pass_config can be also moved.

luccafong: I think we can utilize it,  but it will introduce duplicated attribute in config level, we can think of how to organize these config better in following PR.  @zou3519

houseroad (approved):
Looks good to me

zou3519 (approved):

PR #27046, opened 2025-10-16
feat(v1): Implement pinned prefix caching for V1 engine

dongbo910220:
## Purpose

This pull request implements support for persistent/pinned prefixes in the prefix caching engine, resolving #23083. This feature addresses a key need in production environments where critical prefixes (e.g., system prompts, common instruction templates) can be evicted by the LRU policy under high load, leading to unnecessary recomputation and increased latency.

This PR introduces a robust mechanism to "pin" specific prefixes in the cache, ensuring they are protected from eviction.

**Key Features:**
- **Global Gate**: A new `enable_pinned_prefix` flag in `CacheConfig` acts as a global kill-switch for this feature, allowing for safe, conservative rollouts (default: `False`).
- **Per-Request Control**: A `pin_prefix` boolean parameter has been added to `SamplingParams` and the OpenAI-compatible API's `ChatCompletionRequest`, giving users fine-grained, per-request control over which prefixes to pin.
- **Configurable Capacity Cap**: A `pinned_prefix_cap_ratio` (default: 0.2) in `CacheConfig` limits the maximum fraction of total GPU blocks that can be pinned, preventing pinned prefixes from exhausting the cache and ensuring stability.
- **Observability**: The `cache_blocks` method now returns a rich status dictionary (`status`, `pinned_count`, `cap_limit`, etc.) to provide clear insights into the pinning operation's outcome, aiding in debugging and monitoring.
- **Concurrency Safety**: Pin and unpin operations are designed to be idempotent. Reference counting is used to correctly manage the lifecycle of pinned blocks shared across multiple requests.

## Test Plan

The implementation is validated by a comprehensive suite of new unit and integration tests.

1.  **Run all new unit tests for the core logic:**
    These tests are fast, require no GPU, and verify the `BlockPool` mechanics, capacity management, and status reporting in isolation.
    ```bash
    pytest tests/config/test_pinned_prefix_config.py tests/v1/core/test_pinned_prefix.py
    ```

2.  **Run the end-to-end integration test:**
    This test requires a GPU and validates the full workflow from an API-level request down to the `EngineCore`, ensuring prefix blocks are correctly pinned and subsequent cache hits are registered.
    ```bash
    pytest tests/v1/engine/test_pinned_prefix_caching_integration.py
    ```

## Test Result

All 15+ new tests pass successfully.
- **Unit tests** confirm that the `pin_prefix` parameter is correctly propagated, the capacity cap (`pinned_prefix_cap_ratio`) raises errors for invalid values, the `BlockPool` correctly protects pinned blocks from eviction, and the status reporting is accurate for all states (`ok`, `partial`, `capped`, `disabled`).
- The **integration test** successfully validates that a request with `pin_prefix=True` results in its prefix blocks being pinned, and a subsequent identical request correctly registers a cache hit, proving the end-to-end functionality.
- All existing relevant test suites were run, and no regressions were found.

---
<details>

## Proposed Release Note

(Engine Core)

Introduced Pinned Prefix Caching to boost performance for repetitive prompts. This new feature allows you to "pin" the KV cache of a request's prefix, protecting it from eviction even under high memory pressure. This is particularly effective in scenarios like RAG or applications with fixed system prompts, as it avoids re-computation and significantly reduces the time-to-first-token.

How to use:

Enable it for the engine with the new flag: --enable-pinned-prefix.

You can also control its memory usage with --pinned-prefix-cap-ratio, which accepts a value between 0.0 and 1.0 (default: 0.2, meaning 20% of the cache).

For individual requests, set the pin_prefix=True parameter in your API call.

### DIFF
--- a/vllm/config/cache.py
+++ b/vllm/config/cache.py
@@ -125,6 +125,18 @@ class CacheConfig:
     gpu_memory_utilization. Note that kv_cache_memory_bytes
     (when not-None) ignores gpu_memory_utilization"""
 
+    pinned_prefix_cap_ratio: float = 0.2
+    """Maximum fraction of total GPU blocks that may be pinned for prefix
+    caching per engine instance. Must be within [0.0, 1.0]. Defaults to
+    0.2 (20%). This cap prevents the prefix cache from occupying all
+    blocks, helping stability under load.
+    """
+
+    enable_pinned_prefix: bool = False
+    """Global gate for pinned-prefix behavior. If False, requests with
+    pin_prefix=True will not pin any blocks. Default is disabled to allow
+    conservative rollouts."""
+
     def compute_hash(self) -> str:
         """
         WARNING: Whenever a new field is added to this config,
@@ -161,6 +173,15 @@ class CacheConfig:
                 "scaling factor."
             )
         return cache_dtype
+
+    @field_validator("pinned_prefix_cap_ratio", mode="after")
+    @classmethod
+    def _validate_pinned_prefix_cap_ratio(cls, value: float) -> float:
+        if value < 0.0 or value > 1.0:
+            raise ValueError(
+                f"pinned_prefix_cap_ratio must be within [0.0, 1.0], got {value}."
+            )
+        return value
 
     def verify_with_parallel_config(
         self,
--- a/tests/config/test_pinned_prefix_config.py
+++ b/tests/config/test_pinned_prefix_config.py
@@ -1,0 +1,22 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+import pytest
+
+from vllm.config import CacheConfig
+
+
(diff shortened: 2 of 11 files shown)


vllm/config/cache.py
```diff
@@ -162,6 +174,15 @@ def _validate_cache_dtype(cls, cache_dtype: CacheDType) -> CacheDType:
...
         return cache_dtype
 
+    @field_validator("pinned_prefix_cap_ratio", mode="after")
+    @classmethod
+    def _validate_pinned_prefix_cap_ratio(cls, value: float) -> float:
+        if value < 0.0 or value > 1.0:
+            raise ValueError(
+                f"pinned_prefix_cap_ratio must be within [0.0, 1.0], got {value}."
+            )
+        return value
```
hmellor, line 184: If you remake this PR, you don't need to write this manually, you can use `pydantic.Field(default=0.2, ge=0, le=1)`

PR #26453, opened 2025-10-09
[TPU] Fix tpu torch compile error

Chenyaaang:
Fix TPU torch [compile error](https://buildkite.com/vllm/ci/builds/33886#0199c03d-a2d4-461a-b604-ab2aa8c362a8) after https://github.com/vllm-project/vllm/pull/26113

This pr includes:
1. Set compilation backend to `openxla` on TPU platform
2. Make sure TPU is using `forward_tpu` when dispatching in custom ops
3. Bypass some backend checks which require either `eager` or `inductor` for non-tpu platforms.

PR #27049, opened 2025-10-16
[TPU] Fix tpu torch compile error

Chenyaaang:
Fix torch compile error on TPU platforms, reopen https://github.com/vllm-project/vllm/pull/26453

This pr includes:

1. Set compilation backend to openxla on TPU platform
2. Make sure TPU is using forward_tpu when dispatching in custom ops
3. Bypass some backend checks which require either eager or inductor for non-tpu platforms.

### DIFF
--- a/vllm/compilation/backends.py
+++ b/vllm/compilation/backends.py
@@ -55,9 +55,14 @@ def make_compiler(compilation_config: CompilationConfig) -> CompilerInterface:
             logger.debug("Using InductorAdaptor")
             return InductorAdaptor()
     else:
-        assert compilation_config.backend == "eager", (
-            "Custom backends not supported with CompilationMode.VLLM_COMPILE"
-        )
+        if current_platform.is_tpu():
+            assert compilation_config.backend == "openxla", (
+                "TPU platform should use openxla as compiler backend."
+            )
+        else:
+            assert compilation_config.backend == "eager", (
+                "Custom backends not supported with CompilationMode.VLLM_COMPILE"
+            )
 
         logger.debug("Using EagerAdaptor")
         return EagerAdaptor()
--- a/vllm/config/compilation.py
+++ b/vllm/config/compilation.py
@@ -654,7 +654,7 @@ class CompilationConfig:
             return resolve_obj_by_qualname(self.backend)
 
         assert self.mode == CompilationMode.VLLM_COMPILE
-        if self.backend not in ["eager", "inductor"]:
+        if not current_platform.is_tpu() and self.backend not in ["eager", "inductor"]:
             raise ValueError(
                 f"Invalid backend for piecewise compilation: {self.backend}"
             )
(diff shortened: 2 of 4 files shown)


Chenyaaang:
@ProExpertProg can you please review this again? Looks like the TPU path is still break.

vllm/compilation/backends.py
```diff
@@ -55,9 +55,14 @@ def make_compiler(compilation_config: CompilationConfig) -> CompilerInterface:
             logger.debug("Using InductorAdaptor")
             return InductorAdaptor()
     else:
-        assert compilation_config.backend == "eager", (
-            "Custom backends not supported with CompilationMode.VLLM_COMPILE"
-        )
+        if current_platform.is_tpu():
+            assert compilation_config.backend == "openxla", (
+                "TPU platform should use openxla as compiler backend."
+            )
```
hmellor, line 61: Can this platform specific check live in `vllm/platforms/tpu.py`?

Chenyaaang: Actually we can remove it, because platforms/tpu.py assigns "openxla" to backend. Do you think this will be better?

vllm/compilation/backends.py
```diff
@@ -55,9 +55,14 @@ def make_compiler(compilation_config: CompilationConfig) -> CompilerInterface:
             logger.debug("Using InductorAdaptor")
             return InductorAdaptor()
     else:
-        assert compilation_config.backend == "eager", (
-            "Custom backends not supported with CompilationMode.VLLM_COMPILE"
-        )
+        if current_platform.is_tpu():
```
ProExpertProg, line 58: We should never reach this code unless `CompilationConfig.mode== 3` (`CompilationMode.VLLM_COMPILE`)

Does this mean we're using `VLLM_COMPILE` for TPU now? I thought we used `DYNAMO_TRACE_ONCE`?

vllm/config/compilation.py
ProExpertProg, line 626: This should already set the backend to "openxla"

vllm/platforms/tpu.py
```diff
@@ -140,8 +140,9 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
             )
             compilation_config.cudagraph_mode = CUDAGraphMode.NONE
 
-        if compilation_config.backend == "":
-            compilation_config.backend = "openxla"
+        # Note: the default backend is set to inductor now
+        # we want to overwrite to openxla to execute the ops properly on TPU.
+        compilation_config.backend = "openxla"
```
ProExpertProg, line 145: This should already be set inside `init_backend`. Also the default backend is NOT set to inductor, it is still `""` the way I understand it

ProExpertProg (changes requested):
What issues are you running into? This should have been addressed in #26502.

vllm/platforms/tpu.py
ProExpertProg: Also I see above here that we ARE still using `DYNAMO_TRACE_ONCE`. What issue does this resolve that you're running into

PR #27061, opened 2025-10-17
[MM][Core] Decouple ViT backend from LM backend

ywang96:
## Purpose
Multimdal encoder attention backend for a selection of models have been adopting the same backend as the language model backbone (via env var). This has exposed challenges and inflexibilities since some backends may work for the LM but not for the multimodal encoder.

This PR refactors the ViT backend selection for models that use `get_vit_attn_backend` so that users can specify `--mm-encoder-attn-backend` as an override.

Model-specific attention post-selection changes will stay as is (for example, using upstream FA whenever it's available).

## Test Result

---
<details>

</details>


vllm/attention/layer.py
```diff
@@ -462,7 +464,14 @@ def __init__(
         dtype = torch.get_default_dtype()
 
         # Determine the attention backend
-        backend = get_vit_attn_backend(head_size=head_size, dtype=dtype)
+        attn_backend_override = None
```
DarkLight1337, line 467: Should we rename this layer to `VisionAttention` btw?

vllm/config/multimodal.py
```diff
@@ -148,6 +157,29 @@ def _validate_limit_per_prompt(
...
+
+        if value is None or isinstance(value, BackendEnum):
+            return value
+
+        if isinstance(value, str):
+            candidate = backend_name_to_enum(value.upper())
+            if candidate is not None:
+                return candidate
+
+        valid_backends = ", ".join(sorted(BackendEnum.__members__.keys()))
```
Isotr0py, line 178: Perhaps we can add a `supported_vit_backend` for Platform interface in a following PR to detect invalid backend for specific platform before initializing model.

vllm/attention/layer.py
Isotr0py: This layer will be renamed to `MMEncoderAttention` in #27147. But we can also rename it here.

Isotr0py (approved):

vllm/attention/layer.py
DarkLight1337: Let's do that in the other PR then

vllm/config/multimodal.py
ywang96: Yea right now it'll just show all possible `_Backend` with someone of them get auto resolved inside their correspending platform.get_vit_attn_backend. For example
https://github.com/vllm-project/vllm/blob/9fce7bee745230d61c60ad467966790553b0ba48/vllm/platforms/rocm.py#L203-L211

I think we can shrink this selection by just having a specific `_MHA_Backend` enum

tjtanaa: @ywang96 Yes. I like this idea, we should separate out the ViT attention backend enums

PR #21352, opened 2025-07-22
[Core][Feat] Add max-waiting-queue-length parameter to reject requests when waiting queue is full

chaunceyjiang:
## Essential Elements of an Effective PR Description Checklist
- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x] The test plan, such as providing test command.
- [x] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.

## Purpose
Feature implementation https://github.com/vllm-project/vllm/issues/18826

CLOSE #18826

PR #27064, opened 2025-10-17
[Core][Feat] Add max-waiting-queue-length parameter to reject requests when waiting queue is full

### LINKED ISSUE #18826: [RFC]: Controlling the maximum length of the waiting queue
### Motivation.

Currently, there appears to be no mechanism in vLLM to reject incoming requests based on the waiting queue length. Instead, all incoming requests are added to the queue. The waiting queue is implemented as an unbounded deque residing in CPU memory, where each element represents a pending request. In scenarios of service overload or when using low-performance GPUs, the queue length may grow indefinitely if users do not actively cancel their requests. This can lead to excessive memory consumption and eventually result in an out-of-memory (OOM) failure.

Waiting Queue:
https://github.com/vllm-project/vllm/blob/6825d9a998df3f22b5e19fb600d0a0c09950db28/vllm/v1/core/sched/scheduler.py#L90-L92

To address this issue, we propose introducing a new mechanism to control the maximum length of the waiting queue. Once the queue reaches a specified threshold, any new incoming requests will be rejected immediately with an HTTP 503 (Service Unavailable) response.

Additionally, at a higher level, such as in Kubernetes, we can use Istio, Envoy, and other tools to perform load balancing based on whether the request returns a 503 status code.

### Proposed Change.

Introduce a parameter `--max-waiting-queue-length`. When the scheduler attempts to add a request to the waiting queue, it first checks whether the queue has reached its maximum length. If the queue is full, the request is rejected with an HTTP 503 error.

https://github.com/vllm-project/vllm/blob/6825d9a998df3f22b5e19fb600d0a0c09950db28/vllm/v1/core/sched/scheduler.py#L842-L846

For example
``` python3
    def add_request(self, request: Request, dummy=False) -> None:

        if self.max_waiting_queue_length and \
            len(self.waiting) >= self.max_waiting_queue_length:

            raise SchedulerWaitingQueueFullError(
                f"Scheduler waiting queue is full ({len(self.waiting)} >= "
                f"{self.max_waiting_queue_length}). "
                f"Cannot add request {request.request_id}.")

        """Adds a request to the waiting queue."""
        self.waiting.append(request)
        self.requests[request.request_id] = request
        if self.log_stats:
            request.record_event(EngineCoreEventType.QUEUED)
```

When the waiting queue is full and the scheduler attempts to add a new request to it, a `SchedulerWaitingQueueFullError` is raised. The EngineCore sends this error to the API server via ZMQ. Upon detecting this error, the API server sets the HTTP status code to 503.

### Feedback Period.

_No response_

### CC List.

@DarkLight1337 @simon-mo @njhill 

### Any Other Things.

1. PR #17546 introduced a new interface `get_request_counts` in the scheduler. Perhaps we can reuse this interface.
2. In DP (Data Parallel) mode, there are multiple `EngineCore` processes, each maintaining its own waiting queue. Maybe this can be handled in the `DP Coordinator`? (Just a guess)
3. There are also multiple issues requesting support for this feature, such as https://github.com/vllm-project/vllm/issues/2901.

chaunceyjiang:
## Purpose
Feature implementation https://github.com/vllm-project/vllm/issues/18826

CLOSE https://github.com/vllm-project/vllm/issues/18826

CLOSE https://github.com/vllm-project/vllm/pull/21352
## Test Plan
```
vllm serve /home/jovyan/qwen3-8b  --no-enable-prefix-caching --max-waiting-queue-length 1
```
```
hey -n 1000 -c 50 -m POST \                                                    1 ↵
  -H "Authorization: Bearer YOUR_API_KEY" \
  -d '{
    "messages": [
      {"role": "user", "content": "Hello! What can you do?"} 
    ],
    "temperature": 0.7
  }' http://localhost:8000/v1/chat/completions
```
## Test Result

```
(APIServer pid=18343) INFO:     127.0.0.1:52460 - "POST /v1/chat/completions HTTP/1.1" 200 OK
(APIServer pid=18343) INFO:     127.0.0.1:52122 - "POST /v1/chat/completions HTTP/1.1" 200 OK
(APIServer pid=18343) ERROR 10-17 02:39:34 [serving_engine.py:740] Request chatcmpl-ebe199d1fcc84afd89db45e086f532c1 was rejected by the vLLM model's safety system
(APIServer pid=18343) INFO:     127.0.0.1:52100 - "POST /v1/chat/completions HTTP/1.1" 503 Service Unavailable
(APIServer pid=18343) INFO:     127.0.0.1:52184 - "POST /v1/chat/completions HTTP/1.1" 200 OK
(APIServer pid=18343) INFO:     127.0.0.1:52304 - "POST /v1/chat/completions HTTP/1.1" 200 OK
(APIServer pid=18343) INFO:     127.0.0.1:52198 - "POST /v1/chat/completions HTTP/1.1" 200 OK
(APIServer pid=18343) INFO:     127.0.0.1:52394 - "POST /v1/chat/completions HTTP/1.1" 200 OK
(

```

# TODO
- [ ] e2e
- [ ] ut

<details>

</details>

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit c67d75018f29288602ca75f8438bf0f0e0d02aa1. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Implements a hard cap on the scheduler waiting queue and surfaces rejections to clients.
> 
> - Adds `SchedulerConfig.max_waiting_queue_length` with CLI `--max-waiting-queue-length`; plumbs through `EngineArgs` into engine config
> - Scheduler now rejects requests when the waiting queue reaches the limit, records `REJECTED` event, and returns outputs with finish_reason `rejected`
> - Extends API enums/constants: adds `FinishReason.REJECTED` (`"rejected"`) and `EngineCoreEventType.REJECTED`
> - OpenAI serving maps finish_reason `rejected` to `GenerationError` with `ServiceUnavailableError` (HTTP 503) and preserves error type/status in responses
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c67d75018f29288602ca75f8438bf0f0e0d02aa1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/vllm/config/scheduler.py
+++ b/vllm/config/scheduler.py
@@ -1,7 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 
-import sys
 from collections.abc import Callable
 from dataclasses import InitVar
 from typing import TYPE_CHECKING, Any, ClassVar, Literal, cast
@@ -67,7 +66,7 @@ class SchedulerConfig:
     In real usage, this should be set in `EngineArgs.create_engine_config`.
     """
 
-    max_waiting_queue_length: int = Field(default=sys.maxsize, ge=1)
+    max_waiting_queue_length: int = Field(default=4096, ge=1)
     """
     The maximum number of requests allowed in the waiting queue.
     If the waiting queue is full, new incoming requests will be rejected.
--- a/vllm/config/scheduler.py
+++ b/vllm/config/scheduler.py
@@ -41,7 +41,7 @@ class SchedulerConfig:
 
     DEFAULT_MAX_NUM_BATCHED_TOKENS: ClassVar[int] = 2048
     DEFAULT_MAX_NUM_SEQS: ClassVar[int] = 128
-
+    DEFAULT_MAX_WAITING_QUEUE_LEN: ClassVar[int] = 4096
     runner_type: RunnerType = "generate"
     """The runner type to launch for the model."""
 
@@ -66,7 +66,7 @@ class SchedulerConfig:
     In real usage, this should be set in `EngineArgs.create_engine_config`.
     """
 
-    max_waiting_queue_length: int = Field(default=4096, ge=1)
+    max_waiting_queue_length: int = Field(default=DEFAULT_MAX_WAITING_QUEUE_LEN, ge=1)
     """
     The maximum number of requests allowed in the waiting queue.
     If the waiting queue is full, new incoming requests will be rejected.
--- a/vllm/config/scheduler.py
+++ b/vllm/config/scheduler.py
@@ -66,6 +66,10 @@ class SchedulerConfig:
     In real usage, this should be set in `EngineArgs.create_engine_config`.
     """
 
+    max_waiting_queue_length: int | None = Field(default=None, ge=1)
+    """The maximum number of requests allowed in the waiting queue.
+    If None, there is no limit on the waiting queue length."""
+
     max_num_partial_prefills: int = Field(default=1, ge=1)
     """For chunked prefill, the maximum number of sequences that can be
     partially prefilled concurrently."""
--- a/vllm/config/scheduler.py
+++ b/vllm/config/scheduler.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 
+import sys
 from collections.abc import Callable
 from dataclasses import InitVar
 from typing import TYPE_CHECKING, Any, ClassVar, Literal, cast
@@ -66,9 +67,14 @@ class SchedulerConfig:
     In real usage, this should be set in `EngineArgs.create_engine_config`.
     """
 
-    max_waiting_queue_length: int | None = Field(default=None, ge=1)
-    """The maximum number of requests allowed in the waiting queue.
-    If None, there is no limit on the waiting queue length."""
+    max_waiting_queue_length: int = Field(default=sys.maxsize, ge=1)
+    """
+    The maximum number of requests allowed in the waiting queue.
+    If the waiting queue is full, new incoming requests will be rejected.
+
+    [Experimental] This parameter is an experimental feature and
+    may change or be removed in future releases.
+    """
 
     max_num_partial_prefills: int = Field(default=1, ge=1)
     """For chunked prefill, the maximum number of sequences that can be
--- a/vllm/engine/arg_utils.py
+++ b/vllm/engine/arg_utils.py
@@ -461,6 +461,9 @@ class EngineArgs:
     max_long_partial_prefills: int = SchedulerConfig.max_long_partial_prefills
     long_prefill_token_threshold: int = SchedulerConfig.long_prefill_token_threshold
     max_num_seqs: int | None = None
+    max_waiting_queue_length: int | None = get_field(
+        SchedulerConfig, "max_waiting_queue_length"
+    )
     max_logprobs: int = ModelConfig.max_logprobs
     logprobs_mode: LogprobsMode = ModelConfig.logprobs_mode
     disable_log_stats: bool = False
@@ -1197,6 +1200,9 @@ class EngineArgs:
             **scheduler_kwargs["max_long_partial_prefills"],
         )
         scheduler_group.add_argument(
+            "--max-waiting-queue-length", **scheduler_kwargs["max_waiting_queue_length"]
+        )
+        scheduler_group.add_argument(
             "--long-prefill-token-threshold",
             **scheduler_kwargs["long_prefill_token_threshold"],
         )
@@ -1782,6 +1788,7 @@ class EngineArgs:
             runner_type=model_config.runner_type,
             max_num_batched_tokens=self.max_num_batched_tokens,
             max_num_seqs=self.max_num_seqs,
+            max_waiting_queue_length=self.max_waiting_queue_length,
             max_model_len=model_config.max_model_len,
             enable_chunked_prefill=self.enable_chunked_prefill,
             disable_chunked_mm_input=self.disable_chunked_mm_input,
--- a/vllm/engine/arg_utils.py
+++ b/vllm/engine/arg_utils.py
@@ -461,7 +461,7 @@ class EngineArgs:
     max_long_partial_prefills: int = SchedulerConfig.max_long_partial_prefills
     long_prefill_token_threshold: int = SchedulerConfig.long_prefill_token_threshold
     max_num_seqs: int | None = None
-    max_waiting_queue_length: int | None = get_field(
+    max_waiting_queue_length: int = get_field(
         SchedulerConfig, "max_waiting_queue_length"
     )
     max_logprobs: int = ModelConfig.max_logprobs
--- a/vllm/entrypoints/grpc_server.py
+++ b/vllm/entrypoints/grpc_server.py
@@ -76,9 +76,7 @@ async def serve_grpc(args: argparse.Namespace):
         enable_log_requests=args.enable_log_requests,
         disable_log_stats=args.disable_log_stats,
     )
-    request_logger = None
-    if args.enable_log_requests:
-        request_logger = RequestLogger()
+    request_logger = RequestLogger() if args.enable_log_requests else None
 
     # Create servicer
     servicer = VllmEngineServicer(async_llm, start_time, request_logger=request_logger)
--- a/vllm/entrypoints/grpc_server.py
+++ b/vllm/entrypoints/grpc_server.py
@@ -78,7 +78,7 @@ async def serve_grpc(args: argparse.Namespace):
     )
     request_logger = None
     if args.enable_log_requests:
-        request_logger = RequestLogger(max_log_len=args.max_log_len)
+        request_logger = RequestLogger()
 
     # Create servicer
     servicer = VllmEngineServicer(async_llm, start_time, request_logger=request_logger)
--- a/vllm/entrypoints/grpc_server.py
+++ b/vllm/entrypoints/grpc_server.py
@@ -38,6 +38,7 @@ except ImportError:
 import uvloop
 
 from vllm.engine.arg_utils import AsyncEngineArgs
+from vllm.entrypoints.logger import RequestLogger
 from vllm.entrypoints.utils import log_version_and_model
 from vllm.logger import init_logger
 from vllm.usage.usage_lib import UsageContext
@@ -75,9 +76,12 @@ async def serve_grpc(args: argparse.Namespace):
         enable_log_requests=args.enable_log_requests,
         disable_log_stats=args.disable_log_stats,
     )
+    request_logger = None
+    if args.enable_log_requests:
+        request_logger = RequestLogger(max_log_len=args.max_log_len)
 
     # Create servicer
-    servicer = VllmEngineServicer(async_llm, start_time)
+    servicer = VllmEngineServicer(async_llm, start_time, request_logger=request_logger)
 
     # Create gRPC server
     server = grpc.aio.server(
--- a/vllm/entrypoints/grpc_server.py
+++ b/vllm/entrypoints/grpc_server.py
@@ -38,7 +38,6 @@ except ImportError:
 import uvloop
 
 from vllm.engine.arg_utils import AsyncEngineArgs
-from vllm.entrypoints.logger import RequestLogger
 from vllm.entrypoints.utils import log_version_and_model
 from vllm.logger import init_logger
 from vllm.usage.usage_lib import UsageContext
@@ -76,10 +75,9 @@ async def serve_grpc(args: argparse.Namespace):
         enable_log_requests=args.enable_log_requests,
         disable_log_stats=args.disable_log_stats,
     )
-    request_logger = RequestLogger() if args.enable_log_requests else None
 
     # Create servicer
-    servicer = VllmEngineServicer(async_llm, start_time, request_logger=request_logger)
+    servicer = VllmEngineServicer(async_llm, start_time)
 
     # Create gRPC server
     server = grpc.aio.server(
--- a/vllm/entrypoints/logger.py
+++ b/vllm/entrypoints/logger.py
@@ -17,8 +17,7 @@ logger = init_logger(__name__)
 class RequestLogger:
     def __init__(self, *, max_log_len: int | None) -> None:
         self.max_log_len = max_log_len
-        self.rejected_log_counter = 0
-        self.rejected_log_interval = 100
+
         if not logger.isEnabledFor(logging.INFO):
             logger.warning_once(
                 "`--enable-log-requests` is set but "
@@ -99,14 +98,3 @@ class RequestLogger:
             output_token_ids,
             finish_reason,
         )
-
-    def log_rejected_request(self, request_id: str) -> None:
-        if self.rejected_log_counter == self.rejected_log_interval:
-            logger.warning(
-                "Request %s was rejected due to "
-                "a full waiting queue (log every %d requests)",
-                request_id,
-                self.rejected_log_interval,
-            )
-            self.rejected_log_counter = 0
-        self.rejected_log_counter += 1
--- a/vllm/entrypoints/logger.py
+++ b/vllm/entrypoints/logger.py
@@ -17,7 +17,8 @@ logger = init_logger(__name__)
 class RequestLogger:
     def __init__(self, *, max_log_len: int | None) -> None:
         self.max_log_len = max_log_len
-
+        self.rejected_log_counter = 0
+        self.rejected_log_interval = 100
         if not logger.isEnabledFor(logging.INFO):
             logger.warning_once(
                 "`--enable-log-requests` is set but "
@@ -98,3 +99,17 @@ class RequestLogger:
             output_token_ids,
             finish_reason,
         )
+
+    def log_rejected_request(
+        self,
+        request_id: str,
+    ) -> None:
+        if self.rejected_log_counter % self.rejected_log_interval == 0:
+            logger.error(
+                "Request %s was rejected due to "
+                "a full waiting queue (log every %d requests)",
+                request_id,
+                self.rejected_log_interval,
+            )
+            self.rejected_log_counter = 0
+        self.rejected_log_counter += 1
--- a/vllm/entrypoints/logger.py
+++ b/vllm/entrypoints/logger.py
@@ -100,12 +100,9 @@ class RequestLogger:
             finish_reason,
         )
 
-    def log_rejected_request(
-        self,
-        request_id: str,
-    ) -> None:
-        if self.rejected_log_counter % self.rejected_log_interval == 0:
-            logger.error(
+    def log_rejected_request(self, request_id: str) -> None:
+        if self.rejected_log_counter == self.rejected_log_interval:
+            logger.warning(
                 "Request %s was rejected due to "
                 "a full waiting queue (log every %d requests)",
                 request_id,
--- a/vllm/entrypoints/openai/engine/serving.py
+++ b/vllm/entrypoints/openai/engine/serving.py
@@ -584,8 +584,11 @@ class OpenAIServing:
             )
             raise GenerationError("Internal server error")
         if finish_reason == "rejected":
-            if self.request_logger is not None:
-                self.request_logger.log_rejected_request(request_id)
+            logger.warning_every_n(
+                "Request %s was rejected due to "
+                "a full waiting queue (log every %d requests)",
+                request_id,
+            )
             raise RequestRejectedError()
 
     def _convert_generation_error_to_response(self, e: Exception) -> ErrorResponse:
--- a/vllm/entrypoints/openai/engine/serving.py
+++ b/vllm/entrypoints/openai/engine/serving.py
@@ -71,7 +71,7 @@ from vllm.entrypoints.serve.tokenize.protocol import (
     TokenizeResponse,
 )
 from vllm.entrypoints.utils import create_error_response
-from vllm.exceptions import RequestRejectedError, VLLMValidationError
+from vllm.exceptions import GenerationError, RequestRejectedError, VLLMValidationError
 from vllm.inputs.data import (
     ProcessorInputs,
     PromptType,
--- a/vllm/entrypoints/openai/engine/serving.py
+++ b/vllm/entrypoints/openai/engine/serving.py
@@ -38,7 +38,6 @@ from vllm.entrypoints.openai.engine.protocol import (
     ErrorResponse,
     FunctionCall,
     FunctionDefinition,
-    GenerationError,
 )
 from vllm.entrypoints.openai.models.serving import OpenAIServingModels
 from vllm.entrypoints.openai.responses.protocol import (
@@ -185,7 +184,6 @@ class OpenAIServing:
         *,
         request_logger: RequestLogger | None,
         return_tokens_as_token_ids: bool = False,
-
     ):
         super().__init__()
 
@@ -588,11 +586,11 @@ class OpenAIServing:
         if finish_reason == "rejected":
             if self.request_logger is not None:
                 self.request_logger.log_rejected_request(request_id)
-            raise RequestRejectedError("Request was rejected")
+            raise RequestRejectedError()
 
     def _convert_generation_error_to_response(self, e: Exception) -> ErrorResponse:
         """Convert GenerationError or RequestRejectedError to ErrorResponse."""
-        if isinstance(e, (GenerationError, RequestRejectedError)):
+        if isinstance(e, GenerationError):
             return self.create_error_response(
                 str(e), err_type=e.err_type, status_code=e.status_code
             )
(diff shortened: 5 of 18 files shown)


vllm/config/scheduler.py
```diff
@@ -315,5 +319,12 @@ def _verify_args(self) -> Self:
...
             )
-
+        if (
+            self.max_waiting_queue_length is not None
+            and self.max_waiting_queue_length < 1
+        ):
+            raise ValueError(
+                f"max_waiting_queue_length ({self.max_waiting_queue_length}) "
+                "must be greater than or equal to 1 if specified."
+            )
```
```suggestion
```

vllm/config/scheduler.py
```diff
@@ -52,6 +52,10 @@ class SchedulerConfig:
     """For chunked prefill, the maximum number of sequences that can be
     partially prefilled concurrently."""
 
+    max_waiting_queue_length: int | None = None
```
```suggestion
    max_waiting_queue_length: int | None = Field(default=None, ge=1)
```

chaunceyjiang: good.

vllm/engine/arg_utils.py
```diff
@@ -418,6 +418,7 @@ class EngineArgs:
     max_long_partial_prefills: int = SchedulerConfig.max_long_partial_prefills
     long_prefill_token_threshold: int = SchedulerConfig.long_prefill_token_threshold
     max_num_seqs: int | None = SchedulerConfig.max_num_seqs
+    max_waiting_queue_length: int | None = SchedulerConfig.max_waiting_queue_length
```
hmellor: Now that we're using a `Field` I think we need to do this in order to reuse the default

```suggestion
    max_waiting_queue_length: int | None = get_field(
        SchedulerConfig, "max_waiting_queue_length"
    )
```

vllm/v1/core/sched/scheduler.py
```diff
@@ -1159,6 +1172,15 @@ def get_request_counts(self) -> tuple[int, int]:
         return len(self.running), len(self.waiting)
 
     def add_request(self, request: Request) -> None:
+        if (
+            self.scheduler_config.max_waiting_queue_length
+            and len(self.waiting) >= self.scheduler_config.max_waiting_queue_length
```
hmellor: Micro optimisation:

```suggestion
            (max_waiting := self.scheduler_config.max_waiting_queue_length)
            and len(self.waiting) >= max_waiting
```

chaunceyjiang: good.

vrdn-23:
Hey @chaunceyjiang, thanks for this awesome work! Have you heard anything from the maintainer team on when this might be looked at? Our team is also looking to solve a similar problem, so was hoping that this gets prioritized soon

njhill:
Thanks @chaunceyjiang, I will try to get to this today or early next week.

hmellor (approved):
This LGTM, but let's wait for a second review from Nick who's more familar with the scheduler than me

chaunceyjiang:
> Quick thought: Can we reject requests faster at higher-level, e.g. at AsyncLLM.add_request?
This will allow us to avoid the input/output processing.

Thanks @orozery for your review.

Yes, I had thought about this idea before as well.

At the moment, there are three possible approaches:

1. Call `scheduler.get_request_counts()` in the frontend to get `num_waiting_reqs`. However, this means the frontend would need to call `get_request_counts` for every new request. The communication cost would be higher, because if the request is not rejected, it will still go into the core engine. That means a non-rejected request would incur two communications, so this approach is not ideal.

2. Use `server_load_metrics` in the frontend to decide whether to reject a new request. However, `server_load_metrics` only works when `api-server-count = 1`. Current test results show that increasing `--api-server-count` can significantly improve throughput, so this approach is also not suitable.

3. Reject new requests in the core engine, which is what this PR does. Although this introduces some extra input processing, compared to the other options, this approach is more robust.

vllm/v1/core/sched/scheduler.py
```diff
@@ -1417,7 +1417,21 @@ def update_from_output(
...
+                outputs[request.client_index].append(
+                    EngineCoreOutput(
+                        request_id=request.request_id,
+                        new_token_ids=[],
+                        finish_reason=request.get_finished_reason(),
+                        stop_reason=request.stop_reason,
+                        events=request.take_events(),
+                        trace_headers=request.trace_headers,
+                    )
+                )
```
njhill: Separate this into method shared with `failed_kv_load_req_ids` above?

vllm/v1/core/sched/scheduler.py
```diff
@@ -1613,6 +1627,14 @@ def get_request_counts(self) -> tuple[int, int]:
         return len(self.running), len(self.waiting)
 
     def add_request(self, request: Request) -> None:
+        if (max_waiting := self.scheduler_config.max_waiting_queue_length) and len(
+            self.waiting
+        ) >= max_waiting:
```
njhill: Store `max_waiting_queue_length` directly as a field in the constructor, and have default value be something very large. Then only the inequality needs to be checked here.

vllm/entrypoints/openai/engine/serving.py
```diff
@@ -127,9 +127,15 @@
 class GenerationError(Exception):
     """raised when finish_reason indicates internal server error (500)"""
 
-    def __init__(self, message: str = "Internal server error"):
+    def __init__(
+        self,
+        message: str = "Internal server error",
+        err_type: str = "InternalServerError",
+        status_code: HTTPStatus = HTTPStatus.INTERNAL_SERVER_ERROR,
```
njhill: I wonder if we should have a separate exception type for this?

vllm/entrypoints/openai/engine/serving.py
```diff
@@ -763,14 +769,23 @@ def _raise_if_error(self, finish_reason: str | None, request_id: str) -> None:
                 request_id,
             )
             raise GenerationError("Internal server error")
+        if finish_reason == "rejected":
+            logger.error(
+                "Request %s was rejected due to waiting queue is full", request_id
+            )
```
njhill:
```suggestion
            logger.error(
                "Request %s was rejected due to a full waiting queue, request_id
            )
```

I think we also may want to rate-limit these log messages, or there could be too many in an overloaded-server situation.

vllm/config/scheduler.py
```diff
@@ -61,6 +61,10 @@ class SchedulerConfig:
     In real usage, this should be set in `EngineArgs.create_engine_config`.
     """
 
+    max_waiting_queue_length: int | None = Field(default=None, ge=1)
+    """The maximum number of requests allowed in the waiting queue.
+    If None, there is no limit on the waiting queue length."""
```
njhill: As mentioned let's add a disclaimer that we might change how the load-shedding work in upcoming releases.

njhill (commented):
@chaunceyjiang sorry again for taking so long to look at this properly!

I think there are a couple of considerations:
1. What is the best metric to use for load-shedding
2. How to implement that most efficiently

Here we are assuming waiting queue length for (1.) which is reasonable as a starting point. But it may also be good to consider some kind of weighting which takes into account the nature of the running and waiting requests (like ISL, expected OSL, etc). 
IMO the ideal metric is a predicted queuing time, i.e. predicted time between request arrival and scheduling.

But I think it's ok to start with waiting queue length. We might want to document the parameter though as experimental in case we want to change it in near future.

Re (2.) I like your summary above. Your current impl I think is reasonable since it's no worse than status quo. But we could consider other options, perhaps in a follow-on:

- We already propagate stats back on every step including the waiting/running queue lengths (assuming stats logging is enabled which it is by default). So the front-end could just store these values each output it gets and use those for the check. The main downside of this are that we currently only return the stats to one of the front-ends when there are multiple API servers and it would mean more overhead to send output to every front-end every step.

- So an idea could be to just add a bool flag to `EngineCoreOutputs` indicating whether or not the engine is "saturated" and have the logic on the scheduler side make sure to send an output to all front-ends in the case that this flag changes state. 

- Another idea is to use e.g. a `multiprocessing.Array` where each api server stores in-flight request count in a separate element and then sum them when checking. Probably it would be best to do this without a lock for performance (stale reads shouldn't be a big deal since it can be approximate like @orozery said). This would have an advantage of completely insulating the core engine from additional work.

---

Re `server_load_metrics` ... I guess these are already incorrect when api-server-count > 1. We should probably change the `/load` endpoint to return an error in this case, or at least multiply the number by the number of api servers for a crude approximation. But that's separate from this PR of course...

vllm/entrypoints/openai/engine/serving.py
njhill: Also we should make sure that it gets mapped to an appropriate response code in the grpc case too,

chaunceyjiang: > Also we should make sure that it gets mapped to an appropriate response code in the grpc case too,

I noticed that gRPC currently converts `RequestOutput` directly into `GenerateResponse` and returns it to the client. This means that when `RequestOutput.finish_reason` is `rejected/error`, it is currently passed through to the user.

I’m not sure whether we should also implement this feature on the gRPC side

AkramUllahKhan:
Thanks for the insightful discussion, it helped me a lot.

I’m new to vLLM and first explored limiting requests at the application layer while looking into waiting queue growth. I’m now planning to test the newer release and learn from the internal approach.

Really appreciate the great work here.

vllm/entrypoints/openai/engine/serving.py
njhill: Good point, these should probably be converted to gRPC error responses.

vllm/entrypoints/grpc_server.py
```diff
@@ -111,6 +137,11 @@ async def Generate(
                 request_id=request_id,
                 tokenization_kwargs=tokenization_kwargs,
             ):
+                # Check if the request was rejected due to waiting queue being full.
+                if output.finished and output.outputs:
+                    finish_reason = output.outputs[0].finish_reason
+                    await self._raise_if_error(finish_reason, request_id)
```
chaunceyjiang: > Also we should make sure that it gets mapped to an appropriate response code in the grpc case too,

@njhill Done.

vllm/entrypoints/grpc_server.py
```diff
@@ -68,6 +72,28 @@ def __init__(self, async_llm: AsyncLLM, start_time: float):
         self.start_time = start_time
         logger.info("VllmEngineServicer initialized")
 
+    async def _raise_if_error(self, finish_reason: str | None, request_id: str) -> None:
```
njhill: This method doesn't need to be async

vllm/entrypoints/grpc_server.py
```diff
@@ -56,6 +57,9 @@ class VllmEngineServicer(vllm_engine_pb2_grpc.VllmEngineServicer):
     - GetServerInfo: Server state
     """
 
+    rejected_log_counter = 0
+    rejected_log_interval = 100
```
njhill: just make these instance fields?

vllm/entrypoints/grpc_server.py
```diff
@@ -111,6 +137,11 @@ async def Generate(
                 request_id=request_id,
                 tokenization_kwargs=tokenization_kwargs,
             ):
+                # Check if the request was rejected due to waiting queue being full.
+                if output.finished and output.outputs:
+                    finish_reason = output.outputs[0].finish_reason
```
njhill: can move these checks inside the method

vllm/entrypoints/openai/engine/serving.py
```diff
@@ -681,33 +675,49 @@ def create_streaming_error_response(
...
-            err_type="InternalServerError",
-            status_code=e.status_code,
-        )
+    def _convert_generation_error_to_response(self, e: Exception) -> ErrorResponse:
+        """Convert GenerationError or RequestRejectedError to ErrorResponse."""
+        if isinstance(e, (GenerationError, RequestRejectedError)):
+            return self.create_error_response(
+                str(e),
+                err_type=e.err_type,
+                status_code=e.status_code,
```
njhill:
```suggestion
                str(e), err_type=e.err_type, status_code=e.status_code
```

vllm/entrypoints/openai/engine/serving.py
```diff
@@ -681,33 +675,49 @@ def create_streaming_error_response(
...
-        )
+    def _convert_generation_error_to_streaming_response(self, e: Exception) -> str:
+        """
+        Convert GenerationError or RequestRejectedError to streaming error response.
+        """
+        if isinstance(e, (GenerationError, RequestRejectedError)):
+            return self.create_streaming_error_response(
+                str(e),
+                err_type=e.err_type,
+                status_code=e.status_code,
```
njhill:
```suggestion
                str(e), err_type=e.err_type, status_code=e.status_code
```

vllm/entrypoints/openai/engine/serving.py
```diff
@@ -222,6 +213,9 @@ class ServeContext(Generic[RequestT]):
 
 
 class OpenAIServing:
+    # Limit rejected log frequency
+    rejected_log_counter: int = 0
+    rejected_log_interval: int = 100
```
njhill: use instance fields?

vllm/entrypoints/openai/engine/serving.py
```diff
@@ -681,33 +675,49 @@ def create_streaming_error_response(
...
+        if finish_reason == "rejected":
+            if type(self).rejected_log_counter % type(self).rejected_log_interval == 0:
+                logger.error(
+                    "Request %s was rejected due to "
+                    "a full waiting queue (log every %d requests)",
+                    request_id,
+                    type(self).rejected_log_interval,
+                )
+                type(self).rejected_log_counter = 0
+            type(self).rejected_log_counter += 1
```
njhill: I think this can be put in a simple util class like `RequestRejectionLogger` with a log_rejected_request(req_id) method.

vllm/exceptions.py
```diff
@@ -34,3 +35,32 @@ def __str__(self):
...
+
+    def __init__(
+        self,
+        message: str = "Request was rejected",
+        err_type: str = "ServiceUnavailableError",
+        status_code: HTTPStatus = HTTPStatus.SERVICE_UNAVAILABLE,
+    ):
+        super().__init__(message)
+        self.err_type = err_type
+        self.status_code = status_code
```
njhill: `RequestRejectedError` can extend `GenerationError` and call super constructor with its own values

vllm/v1/core/sched/scheduler.py
```diff
@@ -1510,6 +1513,30 @@ def update_from_output(
...
+            request_id=request.request_id,
+            new_token_ids=[],
+            finish_reason=request.get_finished_reason(),
+            events=request.take_events(),
+            trace_headers=request.trace_headers,
+        )
+        if failed_kv:
+            output.num_cached_tokens = request.num_cached_tokens
+        else:
+            output.stop_reason = request.stop_reason
```
njhill: Is this really needed, can we just always include both of these above? Then `failed_kv` arg not needed.

vllm/v1/core/sched/scheduler.py
```diff
@@ -1443,21 +1443,24 @@ def update_from_output(
...
             self._update_from_kv_xfer_finished(kv_connector_output)
-
+        # Handle rejected requests.
+        if self.rejected:
+            # Create EngineCoreOutputs for all rejected requests.
+            for request in self.rejected:
+                self._append_failed_or_rejected_output(
+                    outputs,
+                    request,
+                )
```
```suggestion
                self._append_failed_or_rejected_output(outputs, request)
```

vllm/v1/core/sched/scheduler.py
```diff
@@ -1443,21 +1443,24 @@ def update_from_output(
...
-                        new_token_ids=[],
-                        finish_reason=request.get_finished_reason(),
-                        events=request.take_events(),
-                        trace_headers=request.trace_headers,
-                        num_cached_tokens=request.num_cached_tokens,
-                    )
+                self._append_failed_or_rejected_output(
+                    outputs,
+                    request,
+                    failed_kv=True,
```
```suggestion
                    outputs, request, failed_kv=True
```

vllm/v1/core/sched/scheduler.py
```diff
@@ -1659,6 +1686,12 @@ def get_request_counts(self) -> tuple[int, int]:
         return len(self.running), len(self.waiting)
 
     def add_request(self, request: Request) -> None:
+        if len(self.waiting) >= self.max_waiting_queue_length:
+            request.status = RequestStatus.FINISHED_REJECTED
+            self.rejected.append(request)
+            if self.log_stats:
+                request.record_event(EngineCoreEventType.REJECTED)
+            return
```
njhill: This should be moved into the `else` below (before `if request.resumable:`)

vllm/v1/core/sched/scheduler.py
njhill: Here, we should also add:
```python
        if self.finished_req_ids_dict is not None:
            self.finished_req_ids_dict[request.client_index].add(request_id)
```

maybe move the `finished_req_ids = self.finished_req_ids_dict` line from below above here so that we don't need to use `self.`

vllm/v1/core/sched/scheduler.py
```diff
@@ -161,7 +161,7 @@ def __init__(
         # Priority queues for requests.
         self.waiting = create_request_queue(self.policy)
         self.running: list[Request] = []
-
+        self.rejected: list[Request] = []
```
njhill: I think this should be moved lower, e.g. below `self.finished_req_ids`, with its own comment. Since it's not in the same category as `waiting`/`running`, it's really just a temporary holding area for bouncing back the rejections.

njhill (commented):
Thanks @chaunceyjiang and sorry again for the delay.

I'm still not sure this is the best final state. As described earlier, I am thinking that it would be better for the queue stats to be returned in the engine outputs to all clients (front end procs), so they keep track of the length and reject requests based on that without the reqs even needing to hit the scheduler.

However, I think this is a reasonable first version and it would be good to get it merged. We can always refactor as a follow-on.

vllm/v1/core/sched/scheduler.py
chaunceyjiang: Done.

vllm/entrypoints/grpc_server.py
```diff
@@ -438,9 +470,12 @@ async def serve_grpc(args: argparse.Namespace):
         enable_log_requests=args.enable_log_requests,
         disable_log_stats=args.disable_log_stats_server,
     )
+    request_logger = None
+    if args.enable_log_requests:
+        request_logger = RequestLogger()
```
njhill:
```suggestion
    request_logger = RequestLogger() if args.enable_log_requests else None
```

vllm/entrypoints/grpc_server.py
```diff
@@ -66,8 +73,27 @@ def __init__(self, async_llm: AsyncLLM, start_time: float):
         """
         self.async_llm = async_llm
         self.start_time = start_time
+        self.rejected_log_counter = 0
+        self.rejected_log_interval = 100
```
njhill: delete
```suggestion
```

vllm/entrypoints/grpc_server.py
```diff
@@ -66,8 +73,27 @@ def __init__(self, async_llm: AsyncLLM, start_time: float):
...
+    def _raise_if_error(self, output: RequestOutput | None, request_id: str) -> None:
+        """
+        Raise appropriate exception if finish_reason indicates an error or rejection.
+        Args:
+            output: The vLLM output
+            request_id: The request ID for logging
+        """
+        finish_reason = ""
+        if output.finished and output.outputs:
+            finish_reason = output.outputs[0].finish_reason
```
njhill:
```suggestion
        if not output.finished or not output.outputs:
            return
        finish_reason = output.outputs[0].finish_reason
```

vllm/entrypoints/logger.py
```diff
@@ -98,3 +99,17 @@ def log_outputs(
             output_token_ids,
             finish_reason,
         )
+
+    def log_rejected_request(
+        self,
+        request_id: str,
+    ) -> None:
+        if self.rejected_log_counter % self.rejected_log_interval == 0:
```
njhill:
```suggestion
        if self.rejected_log_counter == self.rejected_log_interval:
```

vllm/entrypoints/logger.py
```diff
@@ -98,3 +99,17 @@ def log_outputs(
...
+            logger.error(
```
njhill:
```suggestion
            logger.warning(
```

vllm/v1/core/sched/scheduler.py
```diff
@@ -161,13 +161,15 @@ def __init__(
         # Priority queues for requests.
         self.waiting = create_request_queue(self.policy)
         self.running: list[Request] = []
-
```
njhill, line 191: revert

vllm/v1/core/sched/scheduler.py
```diff
@@ -1443,21 +1445,18 @@ def update_from_output(
...
-                )
+                self._append_failed_or_rejected_output(outputs, request)
 
         # KV Connector: update state for finished KV Transfers.
         if kv_connector_output:
             self._update_from_kv_xfer_finished(kv_connector_output)
-
+        # Handle rejected requests.
+        rejected_reqs = self.rejected
+        if rejected_reqs:
```
```suggestion
        if rejected_reqs := self.rejected:
```

vllm/exceptions.py
```diff
@@ -34,3 +35,30 @@ def __str__(self):
...
+    """Raised when finish_reason indicates the request was rejected
+    (e.g., queue full, rate-limited, etc)."""
+
+    def __init__(
+        self,
+        message: str = "Request was rejected",
+        err_type: str = "ServiceUnavailableError",
+        status_code: HTTPStatus = HTTPStatus.SERVICE_UNAVAILABLE,
+    ):
+        super().__init__(message, err_type=err_type, status_code=status_code)
```
njhill: constructor can just be`__init__(self):` and then pass those values to super constructor

vllm/entrypoints/openai/engine/serving.py
```diff
@@ -681,33 +672,38 @@ def create_streaming_error_response(
...
             )
             raise GenerationError("Internal server error")
+        if finish_reason == "rejected":
+            if self.request_logger is not None:
+                self.request_logger.log_rejected_request(request_id)
+            raise RequestRejectedError("Request was rejected")
+
+    def _convert_generation_error_to_response(self, e: Exception) -> ErrorResponse:
+        """Convert GenerationError or RequestRejectedError to ErrorResponse."""
+        if isinstance(e, (GenerationError, RequestRejectedError)):
```
njhill:
```suggestion
        if isinstance(e, GenerationError):
```

vllm/entrypoints/openai/engine/serving.py
```diff
@@ -681,33 +672,38 @@ def create_streaming_error_response(
...
-        return self.create_streaming_error_response(
-            str(e),
-            err_type="InternalServerError",
-            status_code=e.status_code,
-        )
+    def _convert_generation_error_to_streaming_response(self, e: Exception) -> str:
+        """
+        Convert GenerationError or RequestRejectedError to streaming error response.
+        """
+        if isinstance(e, (GenerationError, RequestRejectedError)):
```
njhill:
```suggestion
        if isinstance(e, GenerationError):
```

njhill (commented):
Thanks @chaunceyjiang. 

Looks like it needs rebasing now too, probably some consolidation needed with https://github.com/vllm-project/vllm/pull/31164.

vllm/entrypoints/logger.py
```diff
@@ -98,3 +99,17 @@ def log_outputs(
             output_token_ids,
             finish_reason,
         )
+
+    def log_rejected_request(
+        self,
+        request_id: str,
+    ) -> None:
```
njhill:
```suggestion
    def log_rejected_request(self, request_id: str) -> None:
```

njhill:
I like how you combined this with `RequestLogger` but then realized that we probably want to log rejection warnings even when regular request logging is disabled.

chaunceyjiang:
> I like how you combined this with RequestLogger but then realized that we probably want to log rejection warnings even when regular request logging is disabled.

https://github.com/vllm-project/vllm/pull/36231

What do you think about this?  By #36231, easier to extend `RequestLogger`.

vllm/entrypoints/openai/engine/serving.py
```diff
@@ -592,23 +591,41 @@ def create_streaming_error_response(
...
+        Raise appropriate exception if finish_reason indicates an error or rejection.
+        """
         if finish_reason == "error":
             logger.error(
                 "Request %s failed with an internal error during generation",
                 request_id,
             )
             raise GenerationError("Internal server error")
+        if finish_reason == "rejected":
+            logger.warning_every_n(
```
hmellor, line 587: I'm not sure I understand why we're logging every n?

chaunceyjiang: This suggestion comes from @njhill 

Because there may be a large number of rejected requests at this point, reducing log output would be helpful.

hmellor: Oh ok, the suggestion waws buried in the PR history. Makes sense!

chaunceyjiang:
> Quick thought: Can we reject requests faster at higher-level, e.g. at `AsyncLLM.add_request`? This will allow us to avoid the input/output processing.


Hi @njhill @orozery 

I’ve implemented another PR based on your suggestions. The new implementation avoids the input/output processing. Could you take another look?

https://github.com/vllm-project/vllm/pull/37413

hmellor:
Closing as stale & superseded

PR #27066, opened 2025-10-17
[Frontend][3/N] Improve all pooling task | Support binary embedding response

noooop:
## TL;DR
- support endianness: ["native", "big", "little"], native by default
- support bytes encoding_format, a very simple (but highly efficient)  binary embedding response method
  - First, write metadata into the headers, then write all the binary data in order into Response body.
  - When reading, first read the headers, then read the body according to the offset into the corresponding tenser
- This API provides three benefits @uasan 
1. Significant reduction in response size
2. No need for JSON.parse and Base64.decode
3. Possibility of stream processing of the response
4. Endianness customizations

## Improve all pooling task
- #25817
- #26414
- #25370
- #27066
- #26973
- #25524

## Purpose

> Response compression if the client sends the 'accept-encoding' header: 'zstd, gzip'
> Use Response compression to transmit binary files that don't need base64, cool. base64 is very inefficient
> I wouldn't think of this hacky method before seeing this issue.

Thanks @uasan for this cool idea

Fix #27063

cc @christian-pinto @maxdebayser @DarkLight1337

## Test Plan

tests/utils_/test_serial_utils.py
tests/entrypoints/pooling/openai/test_embedding.py
tests/entrypoints/pooling/openai/test_pooling.py

## Test Result
pass

---
<details>

</details>


vllm/entrypoints/openai/utils.py
```diff
@@ -31,3 +42,29 @@ def encoding_pooling_output(
...
+                tenser.to(torch_dtype)
+                .flatten()
+                .contiguous()
+                .view(torch_view_dtype)
+                .numpy()
+                .newbyteorder(order)
+            )
+            zip_file.writestr(filename, tenser_bytes)
+
+    return zip_buffer.getvalue()
```
noooop: @uasan

It's still in a very early stage, but I think we can't directly use fastapi.middleware.gzip or any similar middleware.

How about just response with a ZIP archive? 
- metadata.json for metadata
- further read binary tensor via filename, endianness, embed_dtype in metadata

Python and other languages can efficiently decompress a zip file in memory, and get the binary data.

***

@christian-pinto @maxdebayser @DarkLight1337

What do you think?
I'm bad at naming things.

noooop: Once I calm down, the method of returning a zip file is too hacky; we should use binary serialization instead.

anyway, binary embedding response is an important feature

uasan: > How about just response with a ZIP archive?

@noooop  No, ZIP archive isn't stream compression. It's too CPU-intensive during decompression and isn't a compression format for the internet. Furthermore, many languages ​​don't have built-in ZIP modules. For example, Node.js (which is important to us) doesn't have ZIP, but they all have GZIP, and newer versions of ZSTD are adding ZSTD. For me, ZSTD is the best choice; it offers fast decompression and a good compression level.

noooop: The most important question: 

archive file (GZIP & ZSTD) or binary serialization(Protobuf, cap n' proto, capnweb)

uasan: > archive file (GZIP & ZSTD) or binary serialization(Protobuf, cap n' proto, capnweb)

I like Protobuf, but for vLLM, the only place where it's needed is probably in the batch response of a binary vector for embedding. All other endpoints will be on the JSON Rest API. So I'm not sure if it's worth implementing Protobuf for just one endpoint, but if you need it, I agree to use it.

uasan: If you compress only the vector individually, instead of stream compressing the entire response body, then each vector in the compressed form will have a different length. We cannot simply take a slice of the buffer with a pre-determined length, because the length of the vector in the compressed form is dynamic, not static. This means that before each vector, you need to specify its length in compressed form, so that the client can take the required piece of the buffer by length

noooop: Therefore, metadata.json is needed for metadata, or write shape and dtype somewhere.

uasan: I don't quite understand why we need files. We just need to put all the metadata of the pre-defined and documented tuple and the last element of the tuple, the bytes of the vector itself, (or Protobuf) and send it in binary form in the body of the response. 

Moreover, it is better to send it immediately, without waiting for all the vectors to be ready. 

I don't see the point in returning the endian and dtype in the response, because the client sent them to you. The client has this metadata. The client needs to be given the vector index number, the number of tokens, and the vector itself.

noooop: Because a request has multiple prompts, and each prompt's returned embedding may have differ size （/pooling）.

uasan: I don't know much Python, I'm a Node.js developer, but it looks like you just need to stream all the bodies in compressed mode.

```Python
return StreamingResponse(body_iter, headers=headers)
```

body_iter - This is a generator that compresses each vector in a cycle.

noooop: i see

First, write metadata into the headers, then write all the binary data in order into StreamingResponse body.

When reading, first read the headers, then read the body according to the offset into the corresponding tenser

(To be honest, I hardly ever write web servers.

uasan: > To be honest, I hardly ever write web servers

On the contrary, I most often write solution servers and thin clients for other services.

It will be more productive if you don't buffer metadata for all vectors. It's better not to buffer anything, but to serve it immediately as you receive responses from the LLM backend. This way, the client can start receiving the first bytes of the vectors earlier, and the client can start performing its work sooner. This results in better performance for both you and the client (no buffering).

I think it will be easier for you and your clients if you compress the entire response body along with the metadata, and give the HTTP client the  header `Content-Encoding: zstd or gzip`, then HTTP clients will automatically decompress the response in streaming mode. This is done by all HTTP clients at a low level, because it is part of the HTTP protocol. At a high level, the user always receives the already decompressed response body.

noooop: Unfortunately, in vllm, it needs to wait for all the prompts in request to be processed before it can start responses.

So streaming return is basically not useful for vllm. （For example, a request changes from a delay of 100ms to 99.999ms.）

uasan: > Unfortunately, in vllm, it needs to wait for all the prompts 

Yes, then I agree, it's not feasible, unfortunately.

> default="big-endian", description="Binary response endianness. Use big-endian by default."

This will break backward compatibility.
Currently, vLLM returns host-endian data. Little-endian on x86 and ARMv8, and Big-endian on PowerPC and SPARC.

You should leave the current default behavior, which is to default to host-endian data. In this case, you shouldn't perform any byte sorting; return the data as is, i.e., as it is now.

noooop: Endianness only affects binary_response, so it won't affect the current json+base64 result at all. 

I set big-endian as the default value because:
> There are also optional enhancements: binary protocols, such as those in Postgres, always expect big-endian binary numbers; this is generally the de facto network standard for almost all binary protocols, but models typically operate in little-endian format; byte order conversion is always necessary.

This is just a very very dirty implementation to give everyone a basis for discussion.


I will try to implement this  tomorrow:
> First, write metadata into the headers, then write all the binary data in order into StreamingResponse body.
> When reading, first read the headers, then read the body according to the offset into the corresponding tenser

uasan: > This is just a very very dirty implementation to give everyone a basis for discussion.

I get it, but host-endian is better as default in any case

vllm/entrypoints/openai/protocol.py
```diff
@@ -1644,17 +1637,22 @@ class EmbeddingChatRequest(OpenAIBaseModel):
...
+    endianness: Literal["native", "big", "little"] = Field(
+        default="native",
+        description=(
+            "What endianness to use for encoding. Default to using native for "
+            "base64 encoding to match the OpenAI python client behavior."
+            "This parameter will affect base64 and binary_response."
         ),
     )
     binary_response: bool = Field(
         default=False, description="Whether to use binary response."
```
uasan: I think it wasn't a very good idea of ​​mine to add a new boolean parameter `binary_response`. The thing is that there is a parameter `encoding_format`, which is responsible for the format of the response, which means the client can pass two mutually exclusive values ​​of two parameters, for example
```JSON
{
   "binary_response": true,
   "encoding_format": "base64"
}
```
I think it's better to leave one `encoding_format` parameter, but add a new value "binary", this will be logical and without collisions
```JSON
{
   "encoding_format": "binary"
}
```

uasan: My proposed format: `Tuple(index: UInt32, tokens: UInt32, embedding: FixedList)`
has a problem with adding new attributes. If you add a new attribute, client-side Tuple parsers will break.

I, and probably many others, only need one index attribute for sort the vector array by index.

I suggest sorting the embedding byte array in vLLM and returning a simple byte array already sorted by index. This means you're forced to buffer the array internally, but you wrote that buffering is required for the vLLM backend.

This binary format will eliminate all attributes and the problem of new attributes. You don't initially pass any attributes and don't plan to add new attributes to the response. Perhaps this is my specific use case and isn't a common one.

uasan: Perhaps new binary formats can be added in the future:
`encoding_format: 'protobuf' | 'arrow-table' | 'bson' | ...`

These formats support attribute descriptions within the response body, i.e., new response attributes can be added to these formats without breaking parses. The `binary` format can be left without attributes - just a sorted binary array of embedding.

vllm/utils/serial_utils.py
```diff
@@ -0,0 +1,120 @@
...
+) -> list[torch.Tensor]:
+    items = metadata["data"]
+    items.sort(key=lambda x: x["index"])
+
+    tensor_list: list[torch.Tensor] = []
+    for item in items:
+        binary = body[item["start"] : item["end"]]
+        tensor = binary2tensor(binary, item["shape"], embed_dtype, endianness)
+        tensor_list.append(tensor)
+    return tensor_list
```
noooop: - support bytes encoding_format, a very simple binary embedding response method
  - First, write metadata into the headers, then write all the binary data in order into Response body.
  - When reading, first read the headers, then read the body according to the offset into the corresponding tenser

@uasan Do you like this bytes encoding_format?

uasan: > @uasan Do you like this bytes encoding_format?

I don't need anything in the response body other than a sorted binary array of vectors, but if you add meta attributes there, I'll parse them, the main thing is that the list of attributes doesn't change in the future.

noooop: Since this API supports input of multiple prompts, the response will contain multiple tensors. You need to use the metadata's item["start"] : item["end"] to extract the corresponding binary from the body and use item["shape"] to restore the original shape.

noooop: This is just for maximum compatibility. If you submit a prompt each time and already know its shape, you don't actually need to care about metadata.

uasan: > This is just for maximum compatibility. If you submit a prompt each time and already know its shape, you don't actually need to care about metadata.

Yes, I'll just skip all the metadata. A sorted binary array will allow us to process response bodies in stream mode, which is cool, that's what we'll do on the client.

We're currently sending 100 input prompts in a 1 request, and the size of one vector is 4 KB.

When we read the HTTP socket, as soon as we receive the first 4 KB, we pass them on to the business logic layer for processing. The next 4 KB will be read into the same buffer. So, to obtain 400 KB of vectors, we only use 4 KB of memory because we process the response in stream mode, reusing the allocated memory.

noooop: Then I’d say this new API is probably the best fit for what you’re looking for.

uasan: @noooop Yes, I think so too, thank you for your work

uasan: This API provides 4 benefits, which I would highlight in the release changelog.
1. Significant reduction in response size
2. No need for JSON.parse and Base64.decode
3. Possibility of stream processing of the response
4. Endianness customizations

noooop:
cc @DarkLight1337 @maxdebayser

Ready for review

I have already noticed a significant speed improvement tests/entrypoints/pooling/openai/test_embedding.py::test_base64_embed_dtype_and_endianness vs tests/entrypoints/pooling/openai/test_embedding.py::test_bytes_embed_dtype_and_endianness

vllm/utils/serial_utils.py
```diff
@@ -0,0 +1,174 @@
+# SPDX-License-Identifier: Apache-2.0
```
DarkLight1337, line 1: Call it `serial_utils`?

vllm/utils/tensor_serial.py
```diff
@@ -0,0 +1,174 @@
...
+    "bfloat16": np.float16,
+    "fp8_e4m3": np.uint8,
+    "fp8_e5m2": np.uint8,
+}
+
+ENDIANNESS = ["native", "big", "little"]
+
+EMBED_DTYPE_TYPE = Literal["float32", "float16", "bfloat16", "fp8_e4m3", "fp8_e5m2"]
+ENDIANNESS_TYPE = Literal["native", "big", "little"]
+ENCODING_FORMAT_TYPE = Literal["float", "base64", "bytes"]
```
DarkLight1337:
```suggestion
EmbedDType = Literal["float32", "float16", "bfloat16", "fp8_e4m3", "fp8_e5m2"]
Endianness = Literal["native", "big", "little"]
EncodingFormat = Literal["float", "base64", "bytes"]
```

Use CamelCase for types, just like classes

vllm/utils/serial_utils.py
```diff
@@ -0,0 +1,171 @@
...
+        item = {
+            "index": idx,
+            "embed_dtype": embed_dtype,
+            "endianness": endianness,
+            "start": offset,
+            "end": offset + size,
+            "shape": output.outputs.data.shape,
+        }
+
+        body.append(binary)
```
maxdebayser, line 148: I think I'm missing something. If body is a list of binaries ( [bin1, bin2....] ), shouldn't the indexing into body be `body[item["index"]]` instead of `body[item["start"]:item["end"]]`? (unless the body is flattened to an single array somewhere)

vllm/utils/serial_utils.py
```diff
@@ -0,0 +1,171 @@
...
+        offset += size
+
+    usage = {
+        "prompt_tokens": num_prompt_tokens,
+        "total_tokens": num_prompt_tokens,
+    }
+    return body, items, usage
+
+
+def decoding_pooling_output(
```
maxdebayser: nit: decode_pooling_output

vllm/utils/serial_utils.py
```diff
@@ -0,0 +1,171 @@
...
+class MetadataItem:
+    index: int
+    embed_dtype: EmbedDType
+    endianness: Endianness
+    start: int
+    end: int
+    shape: tuple[int, ...]
+
+
+def encoding_pooling_bytes(
```
maxdebayser: nit: encode_pooling_bytes

vllm/utils/serial_utils.py
```diff
@@ -0,0 +1,171 @@
...
+
+    np_array = np.frombuffer(binary, dtype=np_dtype).reshape(shape)
+
+    if endianness != "native" and endianness != sys_byteorder:
+        np_array = np_array.byteswap()
+
+    return torch.from_numpy(np_array).view(torch_dtype)
+
+
+def encoding_pooling_output(
```
maxdebayser: nit: encode_pooling_output

vllm/utils/serial_utils.py
```diff
@@ -0,0 +1,171 @@
...
+    assert isinstance(tensor, torch.Tensor)
+    assert embed_dtype in EMBED_DTYPE_TO_TORCH_DTYPE
+    assert endianness in ENDIANNESS
+
+    torch_dtype = EMBED_DTYPE_TO_TORCH_DTYPE[embed_dtype]
+    torch_view_dtype = EMBED_DTYPE_TO_TORCH_DTYPE_VIEW[embed_dtype]
+
+    np_array = (
+        tensor.to(torch_dtype).flatten().contiguous().view(torch_view_dtype).numpy()
+    )
```
noooop, line 67: the body is flattened to an single array in here

vllm/utils/serial_utils.py
noooop: https://github.com/vllm-project/vllm/pull/27066#discussion_r2446456218

vllm/utils/serial_utils.py
noooop: zero copy if this tensor is already contiguous

noooop:
@DarkLight1337 

Are there any more modifications needed for this PR?

vllm/entrypoints/openai/serving_embedding.py
```diff
@@ -121,36 +122,70 @@ def _build_render_config(self, request: EmbeddingCompletionRequest) -> RenderCon
...
+            metadata = {
+                "id": ctx.request_id,
+                "created": ctx.created_time,
+                "model": ctx.model_name,
+                "data": items,
+                "usage": usage,
+            }
+            return EmbeddingBytesResponse(
+                body=body,
+                metadata=json.dumps(metadata),
```
DarkLight1337, line 180: Why is `json.dumps` still being used?

vllm/entrypoints/openai/api_server.py
```diff
@@ -701,6 +706,12 @@ async def create_embedding(request: EmbeddingRequest, raw_request: Request):
         )
     elif isinstance(generator, EmbeddingResponse):
         return JSONResponse(content=generator.model_dump())
+    elif isinstance(generator, EmbeddingBytesResponse):
+        return StreamingResponse(
+            content=generator.body,
+            headers={"metadata": generator.metadata},
+            media_type=generator.media_type,
+        )
```
noooop, line 715: Here, headers require a dict[str:str], so we must use this json.dumps.

vllm/entrypoints/openai/serving_embedding.py
DarkLight1337: Got it, thanks

DarkLight1337 (approved):

PR #22980, opened 2025-08-15
EVS Support (Video tokens pruning)

BloodAxe:
## Purpose

Enable use of Efficient Video Sampling (EVS) for redundant video tokens pruning:

```python
llm = LLM(
    "nvidia/Cosmos-Reason1-7B",
    video_pruning_rate=0.75, # Prune 75% video tokens, effectively reducing TTFT 4x
    limit_mm_per_prompt={"image": 10, "video": 10},
)
```

EVS reduces TTFT and ITL by pruning less important vision tokens from the LLM:

Time to first token (TTFT)
Token throughput

## Test Plan

- Added tests to verify inference with EVS on/off works as expected

## (Optional) Documentation Update

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
</details>

PR #27107, opened 2025-10-17
Nemotron Nano V2 VL + EVS Video Support

BloodAxe:
## Purpose

This MR adds support for EVS to Nemotron Nano 2 VL.
Co-authored with @tomeras91 and @nvnbagrov

## Test Plan

Tested internally


vllm/multimodal/video.py
```diff
@@ -247,6 +247,130 @@ def load_bytes(
         return frames, metadata
 
 
+@VIDEO_LOADER_REGISTRY.register("opencv_nemotron_vl_v2")
+class OpenCVNemotronVideoBackend(OpenCVVideoBackend):
```
DarkLight1337: Maybe we should make `OpenCVVideoBackend` more configurable instead of defining another backend? cc @Isotr0py

Isotr0py: Yes. The main difference for this backend compared to `OpenCVVideoBackend` is just ability to use fps to determine `frame_indices`. I think we just need to expose `fps` to `OpenCVVideoBackend`.

nvnbagrov: Hi @DarkLight1337 , @Isotr0py , thank you for your review.

We tried not to introduce any breaking/surprising behavior.
Our backend works as follows:
* Both `num_samples` and `fps` provided - will sample the minimal number between both
* Only `num_samples` provided - will sample `num_samples`
* Only `fps` provided - will sample frames according to `fps`

For sure we can extend the base class (`OpenCVVideoBackend`) - initially, we thought to do it, but we should set standards:
1. If both `num_samples` and `fps` provided - what's the priority/resolution?
2. When we sample, do we just round or round up or down?
3. At the moment, `VideoMediaIO` sets a default value of `num_frames=32` which IMO should not exist (let's say it should be set to `-1`) if we add `fps` to the base class
4. The current implementation of the actual sampling is a bit inefficient. We should probably tweak it a bit
5. Probably other things I forgot :)


Let's align on these, and we'll happily extend the base class!

nvnbagrov: We can decide on all of the above given the freedom, of course :)

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1740,12 +1740,21 @@ def _execute_mm_encoder(self, scheduler_output: "SchedulerOutput"):
...
+                for video_mm_kwargs_item in filter(
+                    lambda item: item.modality == "video", mm_kwargs
+                ):
+                    _, _, micro_batch_mm_inputs = next(
+                        group_mm_kwargs_by_modality(
+                            [video_mm_kwargs_item],
+                            device=self.device,
+                            pin_memory=self.pin_memory,
+                            merge_by_field_config=model.merge_by_field_config,
+                        )
```
ywang96, line 1763: Is this change necessary?

I honestly also feel like it's probably just cleaner if we can have this under a separate code path when pruning is enabled rather than hacking this under the normal batching logic loop

vllm/multimodal/video.py
BloodAxe: We initially tried with an attempt to fit our changes into existing backend as we really didn't want to introduce a new backed solely for purpose of it being using by one model.
However we discovered we can't keep backward compatibility while not fixing some sampling issues OpenCVDynamicBackend has that was criucial to us. 
So as one wide man told: Today's hacks are tomorrow's legacy code

vllm/v1/worker/gpu_model_runner.py
BloodAxe: @tomeras91 leaving that to you ^_^

vllm/multimodal/video.py
nvnbagrov: Hi again, @DarkLight1337 , @Isotr0py please note the following changes according to my update [ee0804e](https://github.com/vllm-project/vllm/pull/27107/commits/ee0804e109c1fd6581ad0f60cfd2387f42c45150):
* Added `fps` to base `OpenCVVideoBackend`. When both `fps` and `num_frames` are provided - the minimal number of frames considering both will be sampled
* Changed metadata to have *original* values, so that it aligns with HF implementation [[1]](https://github.com/huggingface/transformers/blob/307c5238546ba1675daabc46050c63ffde25f8e6/src/transformers/video_utils.py#L393) *(when sampling, keep original metadata, and add indices)*
* Removed `OpenCVNemotronVideoBackend`. Although the sampling indices vary a bit, I believe we can live with that for the sake of merging this PR

Please have another look now

vllm/v1/worker/gpu_model_runner.py
tomeras91: Hey @ywang96 :)

A few notes on this:
1. This workaround was introduced in a previous [PR](https://github.com/vllm-project/vllm/pull/22980). This one just makes it more robust, by extracting a single video from `mm_kwargs` instead of relying on the fact that the first dimension of items in mm_kwargs_group is the video batch dim.  Changes in `group_mm_kwargs_by_modality()` which now has `merge_by_field_config=True` for almost all relevant models mean that the video batch dimension doesn't exist and frames from different videos are gathered in the same dim. That's why we need a more robust way to deal with processing just a single video
2. The reason the workaround is needed in the first place is because the scheduler considers the _output_ number of tokens to compute how many videos can be batched together. With EVS, this is different than the _input_ number of tokens. For example, with 75% pruning rate, scheduler sees the output number of vision tokens is small (E.g 1/4 of input number of vision tokens) and increases number of images 4x times (effectively trying to "squeeze" multiple videos into a into a batch), which causes OOM in the vision encoder. A better solution is to change the heuristic to compute the batch size, but we feel this is out-of-scope of this PR.
3. I agree it's better if this logic can be moved somewhere more isolated to a different code path where EVS is enabled. Can you advise on such a place? We don't want to put it in modeling code since EVS is a general feature that can work with basically any video model (currently implemented in vLLM for Qwen3-VL and nano-v2-vl)

ywang96: Thanks for the information - I remember the context behind this hack now.

I added a note for it - the real problem here is that during profiling we don't take EVS into account, and therefore drastically underestimate how much memory it takes to execute encoder (which will result in OOM at actual inference time). For now let's leave this as is.

On a related note - did you actually get this to work with Qwen3-VL? Qwen3-VL has special logic around timestamps so I wonder how that would work with EVS.

Isotr0py (approved):
LGTM now!

vllm/v1/worker/gpu_model_runner.py
BloodAxe: @ywang96 I looked into Qwen3 code, and it seems to be very similar to what we have in Nano V2. In Nano V2, there is `"Frame {i} sampled at {t:.2f}:"` + `vision_start_token` + `[vision_placeholder_token] * num_tokens_in_frame_i` + `vision_end_token`.

In Qwen 3 there is `"<{timestamp} seconds>" (on average 9.5 tokens) + vision_start_token + video_token + vision_end_token`. 

We solved how to make EVS working with this interleaving image tokens packing for Nano V2 (Although it was NOT an easy task). So it should be possible to employ EVS for Qwen3 as well. 

A key method is `get_video_repl` (https://github.com/vllm-project/vllm/pull/27107/files#diff-b947b52004ad7e873c11916a4194f0b6825892eb3d51a6c57b0e3d64fc79449bR642) that does all heavy lifting to build "fake" prompt replacement for preprocessing step and "real" prompt replacement when we get final vision embeddings.

ywang96: Amazing - thanks for the explanation

PR #27108, opened 2025-10-17
[CI] Fix mypy for `vllm/v1/core` and `vllm/v1/engine`

yewentao256:
## Purpose

Part of https://github.com/vllm-project/vllm/issues/26533

FIx mypy for `vllm/v1/core` and `vllm/v1/engine`

@hmellor CC

## Test

```bash
pre-commit run --hook-stage manual mypy-3.10 -a
Found 31 errors in 9 files (checked 249 source files)
```

```bash
pre-commit run --hook-stage manual mypy-3.10 -a
```


vllm/v1/engine/llm_engine.py
```diff
@@ -316,7 +318,14 @@ def step(self) -> list[RequestOutput] | list[PoolingRequestOutput]:
...
 
-        return processed_outputs.request_outputs
+        ro = processed_outputs.request_outputs
+        if not ro:
+            return []
+        first = ro[0]
+        if isinstance(first, RequestOutput):
+            return [x for x in ro if isinstance(x, RequestOutput)]
+        else:
+            return [x for x in ro if isinstance(x, PoolingRequestOutput)]
```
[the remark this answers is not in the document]
yewentao256: Fixed

vllm/v1/engine/output_processor.py
```diff
@@ -47,7 +47,14 @@ def put(self, output: RequestOutput | PoolingRequestOutput | Exception) -> None:
...
             # (if n > 1) do not override each other.
-            self.output.add(output, aggregate=self.aggregate)
+            if isinstance(self.output, RequestOutput) and isinstance(
+                output, RequestOutput
+            ):
+                self.output.add(output, aggregate=self.aggregate)
+            elif isinstance(self.output, PoolingRequestOutput) and isinstance(
+                output, PoolingRequestOutput
+            ):
+                self.output = output
```
hmellor: Why are these treated differently?

vllm/v1/engine/llm_engine.py
```diff
@@ -112,10 +112,9 @@ def __init__(
         self.output_processor = OutputProcessor(
             self.tokenizer, log_stats=self.log_stats
         )
-        if self.observability_config.otlp_traces_endpoint is not None:
-            tracer = init_tracer(
-                "vllm.llm_engine", self.observability_config.otlp_traces_endpoint
-            )
+        endpoint = getattr(self.observability_config, "otlp_traces_endpoint", None)
```
hmellor: I think for this one we should change the `default_factory` in `VllmConfig.observability_config` it will always exist

vllm/v1/engine/core_client.py
```diff
@@ -385,7 +385,7 @@ def close_sockets_and_tasks():
                         with contextlib.suppress(Exception):
                             task.cancel()
 
-            if in_loop(loop):
+            if loop is not None and in_loop(loop):
                 close_sockets_and_tasks()
             elif loop and not loop.is_closed():
                 loop.call_soon_threadsafe(close_sockets_and_tasks)
```
hmellor: Maybe this instead of checking if the `loop` is not None twice?

```suggestion
            if loop is not None:
                if in_loop(loop):
                    close_sockets_and_tasks()
                elif not loop.is_closed():
                    loop.call_soon_threadsafe(close_sockets_and_tasks)
```

vllm/v1/engine/async_llm.py
```diff
@@ -653,7 +658,7 @@ async def get_tokenizer(self) -> AnyTokenizer:
         return self.tokenizer
 
     async def is_tracing_enabled(self) -> bool:
-        return self.observability_config.otlp_traces_endpoint is not None
+        return self.observability_config.otlp_traces_endpoint is not None  # type: ignore
```
hmellor, line 661: Why does the type need to be ignored here? Should this not always return a bool?

vllm/v1/engine/core.py
```diff
@@ -467,9 +467,10 @@ def save_tensorized_model(
         self,
         tensorizer_config,
     ) -> None:
-        self.model_executor.save_tensorized_model(
-            tensorizer_config=tensorizer_config,
-        )
+        if hasattr(self.model_executor, "save_tensorized_model"):
+            self.model_executor.save_tensorized_model(  # type: ignore[attr-defined]
+                tensorizer_config=tensorizer_config,
+            )
```
hmellor: This should probably error if we if the model executor doesn't have this method

vllm/v1/engine/core.py
```diff
@@ -1235,7 +1237,8 @@ def reinitialize_distributed(
             parallel_config.data_parallel_master_port
         )
 
-        self.model_executor.reinitialize_distributed(reconfig_request)
+        if hasattr(self.model_executor, "reinitialize_distributed"):
+            self.model_executor.reinitialize_distributed(reconfig_request)  # type: ignore[attr-defined]
```
hmellor: Should this be added to `ExecutorBase` with a `NotImplementedError`?

vllm/v1/engine/async_llm.py
```diff
@@ -131,10 +131,9 @@ def __init__(
         self.output_processor = OutputProcessor(
             self.tokenizer, log_stats=self.log_stats
         )
-        if self.observability_config.otlp_traces_endpoint is not None:
-            tracer = init_tracer(
-                "vllm.llm_engine", self.observability_config.otlp_traces_endpoint
-            )
+        endpoint = getattr(self.observability_config, "otlp_traces_endpoint", None)
```
hmellor: I think for this one we should change the `default_factory` in `VllmConfig.observability_config` it will always exist

vllm/v1/engine/async_llm.py
yewentao256: The issue is `vllm/v1/engine/async_llm.py:661: error: Item "None" of "ObservabilityConfig | None" has no attribute "otlp_traces_endpoint"  [union-attr]`

```py
    async def is_tracing_enabled(self) -> bool:
        assert self.observability_config is not None
        return self.observability_config.otlp_traces_endpoint is not None
```
I think this is not very elegant so just #type ignore

vllm/v1/engine/core.py
yewentao256: Just to confirm, you meant we should raise an error here in case the model executor doesn't have this method?

vllm/v1/engine/core_client.py
yewentao256: Fixed, thanks!

vllm/v1/engine/output_processor.py
yewentao256:
```bash
vllm/v1/engine/output_processor.py:50: error: Item "PoolingRequestOutput[PoolingOutput]" of "RequestOutput | PoolingRequestOutput[PoolingOutput]" has no attribute "add"  [union-attr]
vllm/v1/engine/output_processor.py:50: error: Argument 1 to "add" of "RequestOutput" has incompatible type "RequestOutput | PoolingRequestOutput[PoolingOutput]"; expected "RequestOutput"  [arg-type]
```

vllm/v1/engine/core.py
yewentao256: Fixed, thanks!

vllm/v1/engine/async_llm.py
yewentao256: Fixed

vllm/v1/engine/output_processor.py
hmellor: Interesting! I quite like when mypy is able to find bugs like this

vllm/v1/engine/output_processor.py
```diff
@@ -47,7 +47,14 @@ def put(self, output: RequestOutput | PoolingRequestOutput | Exception) -> None:
...
```
hmellor: We could optimise this slightly so that `isinstance(self.output, RequestOutput)` is not checked twice:

```py
        # These two cases ensure that request outputs with different
        # request indexes (if n > 1) do not override each other.
        elif isinstance(self.output, RequestOutput) and isinstance(
            output, RequestOutput
        ):
            self.output.add(output, aggregate=self.aggregate)
        elif isinstance(self.output, PoolingRequestOutput) and isinstance(
            output, PoolingRequestOutput
        ):
            self.output = output
```

vllm/v1/engine/output_processor.py
```diff
@@ -407,7 +414,7 @@ def process_outputs(
         within the loop below.
         """
 
-        request_outputs: list[RequestOutput] | list[PoolingRequestOutput] = []
+        request_outputs: list[RequestOutput | PoolingRequestOutput] = []
```
hmellor, line 417: Is it possible for these lists to be mixed?

vllm/v1/core/sched/scheduler.py
```diff
@@ -392,7 +392,7 @@ def schedule(self) -> SchedulerOutput:
                     skipped_waiting_requests.prepend_request(request)
                     continue
 
-                num_external_computed_tokens = 0
+                num_external_computed_tokens: int | None = 0
```
hmellor: If we change the name of the local variable in `if self.connector is not None:`, then we don't need to type hint this as `None` and we don't need to add `or 0` in the loop?

```suggestion
                num_external_computed_tokens = 0
```

vllm/v1/engine/output_processor.py
yewentao256:
```py
@dataclass
class OutputProcessorOutput:
    request_outputs: list[RequestOutput | PoolingRequestOutput]
    reqs_to_abort: list[str]
```
I think so since it is marked here

vllm/v1/engine/output_processor.py
yewentao256: Nice suggestion! Fixed

vllm/v1/core/sched/scheduler.py
yewentao256: Nice suggestion! Fixed

hmellor (approved):

hmellor:
It happens when a PR changes a pre-commit rule.

1. PR1 makes a change to a pre-commit rule (like this PR with mypy) and CI starts
2. PR2 makes a change that would fail this new rule but the CI starts in parallel without the new rule applied
3. Both PRs merge cleanly and now pre-commit fails on main 

The best way to prevent this is to always merge from main before attempting to merge a PR which changes pre-commit rules.

PR #27112, opened 2025-10-17
[CI] Remove forbidden slash

NickLucche:
Rename to avoid breaking CI due to the use of a particularly unfortunate slash character


njhill (approved):
Thanks @NickLucche!

PR #27123, opened 2025-10-17
[Kernels] Isolate modular kernel code from FusedMoEMethodBase subclasses.

bnellnm:
## Purpose

Make a new `FusedMoEModularMethod` subclass of `FusedMoeMethodBase` for use with modular kernels. 

Instead of having every subclass of `FusedMoEMethodBase` check `self.fused_experts`, we swap out the `quant_method` of the `FusedMoE` layer to an instance of `FusedMoEModularMethod`.  This will reduce the complexity of the various `FusedMoEMethodBase` subclass `apply` methods and isolate uses of modular kernels to the new class.

## Test Plan

Ran by hand on some fp8 + modelopt models.
CI tests

## Test Result

cc @varun-sundar-rabindranath , @wenscarl

### DIFF
--- a/vllm/model_executor/layers/fused_moe/layer.py
+++ b/vllm/model_executor/layers/fused_moe/layer.py
@@ -430,7 +430,7 @@ class FusedMoEModularMethod(FusedMoEMethodBase, CustomOp):
             w2=layer.w2_weight,
             topk_weights=topk_weights,
             topk_ids=topk_ids,
-            inplace=True,
+            inplace=True,  # TODO(bnell): make sure this is handled properly
             activation=activation,
             global_num_experts=global_num_experts,
             apply_router_weight_on_input=apply_router_weight_on_input,
--- a/vllm/model_executor/layers/fused_moe/layer.py
+++ b/vllm/model_executor/layers/fused_moe/layer.py
@@ -119,7 +119,6 @@ class FusedMoEMethodBase(QuantizeMethodBase):
         super().__init__()
         self.moe = moe
         self.moe_quant_config: FusedMoEQuantConfig | None = None
-        self.fused_experts: FusedMoEModularKernel | None = None
         self.topk_indices_dtype = None
 
     @abstractmethod
@@ -262,9 +261,6 @@ class FusedMoEMethodBase(QuantizeMethodBase):
                 "%s for %s(%s)", prepare_finalize.__class__.__name__, self, id(self)
             )
             assert self.topk_indices_dtype is None
-            assert self.fused_experts is None, (
-                f"Attempt to override experts for {id(self)}!"
-            )
             self.topk_indices_dtype = prepare_finalize.topk_indices_dtype()
             experts = self.select_gemm_impl(prepare_finalize, layer)
             return FusedMoEModularKernel(
@@ -295,7 +291,11 @@ class FusedMoEMethodBase(QuantizeMethodBase):
 
     @property
     def using_modular_kernel(self) -> bool:
-        return self.fused_experts is not None
+        return False
+
+    @property
+    def supports_eplb(self) -> bool:
+        return False
 
     @abstractmethod
     def apply(
@@ -338,10 +338,21 @@ class FusedMoEModularMethod(FusedMoEMethodBase, CustomOp):
         self.moe_quant_config = old_moe_method.moe_quant_config
         self.fused_experts = fused_experts
         self.topk_indices_dtype = old_moe_method.topk_indices_dtype
-
+        self.disable_expert_map = not fused_experts.supports_expert_map()
+        self.old_method_name = old_moe_method.__class__.__name__
+        self._supports_eplb = old_moe_method.supports_eplb
         if isinstance(old_moe_method, torch.nn.Module):
             self.load_state_dict(old_moe_method.state_dict())
-        logger.debug("Swapping out %s", old_moe_method.__class__.__name__)
+        logger.debug("Swapping out %s", self.old_method_name)
+
+    @property
+    def using_modular_kernel(self) -> bool:
+        return True
+
+    @property
+    @abstractmethod
+    def supports_eplb(self) -> bool:
+        return self._supports_eplb
 
     def create_weights(
         self,
@@ -382,11 +393,20 @@ class FusedMoEModularMethod(FusedMoEMethodBase, CustomOp):
         logical_to_physical_map: torch.Tensor | None = None,
         logical_replica_count: torch.Tensor | None = None,
     ) -> torch.Tensor | tuple[torch.Tensor, torch.Tensor]:
-        assert self.fused_experts is not None
-
         # Is getattr needed?
         zero_expert_num = getattr(layer, "zero_expert_num", 0)
         zero_expert_type = getattr(layer, "zero_expert_type", None)
+
+        if enable_eplb:
+            if not self.supports_eplb:
+                assert expert_load_view is not None
+                assert logical_to_physical_map is not None
+                assert logical_replica_count is not None
+                assert isinstance(layer, FusedMoE)
+            else:
+                raise NotImplementedError(
+                    f"EPLB is not supported for {self.old_method_name}"
+                )
 
         select_result = FusedMoE.select_experts(
             hidden_states=x,
@@ -423,7 +443,7 @@ class FusedMoEModularMethod(FusedMoEMethodBase, CustomOp):
             activation=activation,
             global_num_experts=global_num_experts,
             apply_router_weight_on_input=apply_router_weight_on_input,
-            expert_map=expert_map,
+            expert_map=None if self.disable_expert_map else expert_map,
         )
 
         if zero_expert_num != 0 and zero_expert_type is not None:
@@ -763,7 +783,6 @@ class UnquantizedFusedMoEMethod(FusedMoEMethodBase, CustomOp):
         )
 
         if self.rocm_aiter_moe_enabled:
-            assert self.fused_experts is None
             result = self.rocm_aiter_fused_experts(
                 hidden_states=x,
                 w1=layer.w13_weight,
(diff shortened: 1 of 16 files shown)


vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -245,9 +244,9 @@ def maybe_make_prepare_finalize(self) -> FusedMoEPrepareAndFinalize | None:
         else:
             return None
 
-    # Note: init_prepare_finalize should only be called by
-    # prepare_communication_buffer_for_model.
-    def init_prepare_finalize(self, layer: torch.nn.Module):
+    def init_prepare_finalize(
+        self, layer: torch.nn.Module
+    ) -> FusedMoEModularKernel | None:
```
varun-sundar-rabindranath, line 248: is `init_modular_kernel` a better name ?

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -322,6 +324,137 @@ def apply(
...
+            expert_map=expert_map,
+            expert_load_view=expert_load_view,
+            logical_to_physical_map=logical_to_physical_map,
+            logical_replica_count=logical_replica_count,
+            global_num_experts=global_num_experts,
+            zero_expert_num=zero_expert_num,
+            zero_expert_type=zero_expert_type,
+        )
+
+        topk_weights, topk_ids, zero_expert_result = select_result
```
varun-sundar-rabindranath: nit: unpack the return value at `FusedMoE.select_experts` callsite ?

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1306,20 +1432,17 @@ def __init__(
...
+        if self.enable_eplb and not self.quant_method.supports_eplb:
+            # TODO: Add support for additional quantization methods.
+            # The implementation for other quantization methods does not
+            # contain essential differences, but the current quant API
+            # design causes duplicated work when extending to new
+            # quantization methods, so I'm leaving it for now.
+            # If you plan to add support for more quantization methods,
+            # please refer to the implementation in `Fp8MoEMethod`.
+            raise NotImplementedError(
+                "EPLB is only supported for FP8 quantization for now."
```
varun-sundar-rabindranath, line 1459: nit: maybe `f"EPLB not supported for {self.quant_method}. EPLB is only supported for FP8 quantization for now"` ?

vllm/model_executor/layers/fused_moe/modular_kernel.py
```diff
@@ -707,6 +707,15 @@ def __init__(
             f"{fused_experts.activation_formats[0]}"
         )
 
+    def supports_expert_map(self) -> bool:
+        """
+        A flag indicating whether or not this class supports expert maps
+        """
+        return (
+            self.prepare_finalize.num_dispatchers() <= 1
+            and self.fused_experts.supports_expert_map()
```
varun-sundar-rabindranath: Can you add a comment as to why we'd need to check `self.prepare_finalize.num_dispatchers()` - I assumed it to be property of `fused_experts` exclusively. also can `self.prepare_finalize.num_dispatchers()` be 0 ?

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -322,6 +324,137 @@ def apply(
...
+        self.moe_quant_config = old_moe_method.moe_quant_config
+        self.fused_experts = fused_experts
+        self.topk_indices_dtype = old_moe_method.topk_indices_dtype
+        self.disable_expert_map = not fused_experts.supports_expert_map()
+        self.old_method_name = old_moe_method.__class__.__name__
+        self._supports_eplb = old_moe_method.supports_eplb
+        self._allow_inplace = old_moe_method.allow_inplace
+        if isinstance(old_moe_method, torch.nn.Module):
+            self.load_state_dict(old_moe_method.state_dict())
+        logger.debug("Swapping out %s", self.old_method_name)
```
varun-sundar-rabindranath: Instead of copying all the `state_dict` , can we keep the `old_moe_method` in something like `_quant_method` and let this class be a wrapper around the old method ? 
that way, we can just call `self._quant_method.fn()` for functions/properties like `supports_eplb`, `allow_inplace`, `create_weights`. And for `apply`, we just have our own implementation.

Do you think that'll work ?

vllm/model_executor/layers/fused_moe/modular_kernel.py
bnellnm: I copied that bit from compressed_tensors_moe.py but it seems redundant, so I've removed it.

varun-sundar-rabindranath:
Thanks @bnellnm . This cleans up a bunch of redundant code 🙌  . 

I have a suggestion. IIUC, the function call chain for the construction of `FusedMoEModularMethod` looks something like follows, 

```
1. `DeviceCommunicatorBase::prepare_communication_buffer_for_model()` calls  `FusedMoE::init_prepare_finalize()` 
2. `FusedMoE::init_prepare_finalize()` calls `FusedMoEMethodBase::init_prepare_finalize()` and it returns the `FusedMoEModularKernel` object
3. `FusedMoE::init_prepare_finalize()` then makes a `FusedMoEModularMethod` object and overrides its `self.quant_method`
```

Here, note that `FusedMoEMethodBase::init_prepare_finalize()` calls `FusedMoEMethodBase::maybe_make_prepare_finalize()` which in turn calls a static function `FusedMoEMethodBase::_maybe_make_prepare_finalize()`  which does most of the work anyways. 

My suggestion is to move `FusedMoEModularMethod` into its own file and expose a function say `maybe_make_fused_moe_modular_method()` that attempts to construct the `FusedMoEModularMethod` object. 

That way, we can get rid of most of the ModularKernel specific code from `fused_moe/layer.py` and have it in a different file thus cleaning up `fused_moe/layer.py` greatly.  

What do you think ? 

I am not suggesting we do it in this PR. I can take it up as well 👍

bnellnm:
> Thanks @bnellnm . This cleans up a bunch of redundant code 🙌 .
> 
> I have a suggestion. IIUC, the function call chain for the construction of `FusedMoEModularMethod` looks something like follows,
> 
> ```
> 1. `DeviceCommunicatorBase::prepare_communication_buffer_for_model()` calls  `FusedMoE::init_prepare_finalize()` 
> 2. `FusedMoE::init_prepare_finalize()` calls `FusedMoEMethodBase::init_prepare_finalize()` and it returns the `FusedMoEModularKernel` object
> 3. `FusedMoE::init_prepare_finalize()` then makes a `FusedMoEModularMethod` object and overrides its `self.quant_method`
> ```
> 
> Here, note that `FusedMoEMethodBase::init_prepare_finalize()` calls `FusedMoEMethodBase::maybe_make_prepare_finalize()` which in turn calls a static function `FusedMoEMethodBase::_maybe_make_prepare_finalize()` which does most of the work anyways.
> 
> My suggestion is to move `FusedMoEModularMethod` into its own file and expose a function say `maybe_make_fused_moe_modular_method()` that attempts to construct the `FusedMoEModularMethod` object.
> 
> That way, we can get rid of most of the ModularKernel specific code from `fused_moe/layer.py` and have it in a different file thus cleaning up `fused_moe/layer.py` greatly.
> 
> What do you think ?
> 
> I am not suggesting we do it in this PR. I can take it up as well 👍

Yeah, that's a good idea.  I was also considering splitting up layer.py in different ways, e.g. move UnquantizedMoEMethod to a separate file.

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -322,6 +325,138 @@ def apply(
...
+            old_quant_method,
+            "disable_expert_map",
+            not fused_experts.supports_expert_map(),
+        )
+        self.old_quant_method = old_quant_method
+        logger.debug("Swapping out %s", self.old_quant_method.__class__.__name__)
+
+    @property
+    def topk_indices_dtype(self) -> torch.dtype | None:
+        return self.fused_experts.prepare_finalize.topk_indices_dtype()
```
varun-sundar-rabindranath, line 350: Nice change 🙌

vllm/model_executor/layers/quantization/bitsandbytes.py
```diff
@@ -518,12 +518,11 @@ def apply(
     ) -> torch.Tensor | tuple[torch.Tensor, torch.Tensor]:
         from vllm.model_executor.layers.fused_moe import fused_experts
 
-        assert self.fused_experts is None
-
         if enable_eplb:
             raise NotImplementedError(
                 "EPLB not supported for `BitsAndBytesMoEMethod` yet."
             )
+
```
varun-sundar-rabindranath, line 525: nit: cruft

varun-sundar-rabindranath: actually, sorry for flagging this - If the linter is okay with it, it should be fine.

vllm/model_executor/layers/quantization/mxfp4.py
```diff
@@ -815,6 +813,18 @@ def select_gemm_impl(
...
+                self.w13_weight_triton_tensor
+                if layer.w13_weight is None
+                else layer.w13_weight
+            )
+            layer.w2_weight = (
+                self.w2_weight_triton_tensor
+                if layer.w2_weight is None
+                else layer.w2_weight
+            )
+            assert all([w is not None for w in [layer.w13_weight, layer.w2_weight]])
```
varun-sundar-rabindranath, line 826: I think this setting of layer.w13_weight and layer.w2_weight better fits in the `process_weights_after_loading` function here https://github.com/vllm-project/vllm/blob/0fe01404082744c955d135c3634e17de1404b00c/vllm/model_executor/layers/quantization/mxfp4.py#L742 
That way we can get rid of having to differentiate between the `w13_weight_triton_tensor/w2_weight_triton_tensor` and `w13_weight/w2_weight` .

**Not suggesting for this PR**. Fixing that I think should be its own PR.

varun-sundar-rabindranath (approved):
LGTM ! Very nice cleanups ! Thanks @bnellnm

tlrmchlsmth (approved):

mgoin (approved):

wangshangsam:
@bnellnm I have a ... maybe dumb ... question - how exactly is each derived `MoEMethod` class going to trigger `FusedMoEModularMethod.apply()` (thereby using the modular kernels)? Doesn't each subclass override the `.apply()` completely?

bnellnm:
> @bnellnm I have a ... maybe dumb ... question - how exactly is each derived `MoEMethod` class going to trigger `FusedMoEModularMethod.apply()` (thereby using the modular kernels)? Doesn't each subclass override the `.apply()` completely?

The `FusedMoE` layer calls `self.quant_method.apply` so if no modular kernel has been constructed, this will invoke an `apply` method on some subclass of `FusedMoEMethodBase`.  Now, when a modular kernel gets created, the `FusedMoE` layer will swap out `self.quant_method` with an instance of `FusedMoEModularMethod` which will call the modular kernel instead.

So, subclasses of `FusedMoEMethodBase` no longer need to worry about modifying `apply` for modular kernels.

PR #26104, opened 2025-10-02
[ROCm] [VL] [Bugfix] Fix vit flash attn dispatcher logic for ROCm

tjtanaa:
## Purpose

The refactoring of code has causes the vit flash attn dispatcher logic to enter the wrong code path to import
`from vllm.vllm_flash_attn import flash_attn_varlen_func` on ROCm platform.

Fix incorrect usage of `aiter.flash_attn_varlen_func` in `MultiHeadAttention` class introduced in https://github.com/vllm-project/vllm/pull/23978

## Test Plan

Evaluate accuracy of all of the models that uses this vit flash attn dispatcher logic on chartqa dataset.
NOTE: The accuracy by no means indicates the actual model performance on benchmark and the accuracy is not evaluate through the same procedure used in the official release. 

Bugfix of `MultiHeadAttention` class is validated through `OpenGVLab/InternVL3_5-8B`.

## Test Result
# Flash Attention Backend Comparison: AIter vs Non-AIter

| Model | Backend | Explicit Prompt Relaxed Correctness | Anywhere in Answer Relaxed Correctness |
|-------|---------|-------------------------------------|----------------------------------------|
| **Qwen/Qwen2.5-VL-72B-Instruct** | AIter | 0.8672 | 0.8860 |
| **Qwen/Qwen2.5-VL-72B-Instruct** | No AIter | 0.8624 | 0.8848 |
| **Qwen/Qwen3-VL-235B-A22B-Instruct** | No AIter | 0.8648 | 0.8656 |
| **Qwen/Qwen3-VL-235B-A22B-Instruct** | AIter | 0.8656 | 0.8680 |
| **zai-org/GLM-4.5V-FP8** | No AIter | 0.5088 | 0.5716 |
| **zai-org/GLM-4.5V-FP8** | AIter | 0.4952 | 0.5580 |
| **baidu/ERNIE-4.5-VL-28B-A3B-PT** | No AIter | 0.8424 | 0.8828 |
| **baidu/ERNIE-4.5-VL-28B-A3B-PT** | AIter | 0.8444 | 0.8768 |
| **AIDC-AI/Ovis2.5-9B** | No AIter | 0.8656 | 0.8764 |
| **AIDC-AI/Ovis2.5-9B** | AIter | 0.8652 | 0.8784 |
| **OpenGVLab/InternVL3_5-8B** | No AIter | 0.892 | 0.892 |
| **OpenGVLab/InternVL3_5-8B** | AIter | 0.8964 | 0.8964 |

---
[...]

PR #27124, opened 2025-10-17
[Bugfix] Honor --mm_encoder_attn_backend when used

bradleyhd:
Summary:
In https://github.com/vllm-project/vllm/pull/26104, some changes were made in layer.py that resulted in always trying to switch to FA backend for ViT, even when  `VLLM_ATTENTION_BACKEND` is set.

This broke Meta's internal AMD pipelines as it is not desired nor expected behavior. With this change, the models that were changed in the offending PR can explicitly opt-in to this behavior.

Differential Revision: D84946967

### DIFF
--- a/vllm/attention/layer.py
+++ b/vllm/attention/layer.py
@@ -93,12 +93,15 @@ def check_upstream_fa_availability(dtype: torch.dtype):
 
 
 def maybe_get_vit_flash_attn_backend(
-    attn_backend: _Backend, use_upstream_fa: bool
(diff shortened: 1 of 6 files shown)


DarkLight1337:
Heads up that we have decoupled the two backends in https://github.com/vllm-project/vllm/pull/27061

ywang96 (approved):
LGTM - I think we do need to think about how to deal with override in a better way (whether we should honor it truly with the risk of failure or handle the fallback automatically)

PR #27134, opened 2025-10-18
[Kernels] Enable FlashInfer FP8 Blockscale on SM90 (for TEP DSR1)

djmmoss:
## Purpose

**DeepSeek FP8 block-scale path**
 - Introduces and propagates use_deepseek_fp8_block_scale through prepare/finalize, GEMM selection, and the FlashInfer kernel call.
 - In this mode, activations are not quantized in prepare (a1q = a1, a1q_scale = None); the kernel consumes per-block weight scales instead.
 - Expert GEMM receives quant_scales=[w1_scale, w2_scale] and the flag use_deepseek_fp8_block_scale=True to activate the block-scale decoding path

## Accuracy (Qwen3-Next-80B-A3B-Instruct-FP8)

```bash
FLASHINFER TEP
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8590|±  |0.0096|
|     |       |strict-match    |     5|exact_match|↑  |0.8105|±  |0.0108|
```

```bash
TRITON TEP
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8605|±  |0.0095|
|     |       |strict-match    |     5|exact_match|↑  |0.8226|±  |0.0105|
```

## Performance

image
(thanks @robertgshaw2-redhat for the data)


vllm/model_executor/layers/quantization/fp8.py
```diff
@@ -711,6 +716,11 @@ def __init__(self, quant_config: Fp8Config, layer: torch.nn.Module):
             self.flashinfer_moe_backend = FlashinferMoeBackend.TENSORRT_LLM
         elif self.fp8_backend == Fp8MoeBackend.FLASHINFER_CUTLASS:
             self.flashinfer_moe_backend = FlashinferMoeBackend.CUTLASS
+            self.flashinfer_moe_fn = partial(
+                flashinfer_cutlass_moe_fp8,
+                moe=self.moe,
+                use_deepseek_fp8_block_scale=self.block_quant is not None,
```
pavanimajety: Does this also need to guarded by blockscale's block size?

vllm/model_executor/layers/fused_moe/flashinfer_cutlass_prepare_finalize.py
```diff
@@ -173,14 +180,20 @@ def prepare(
...
 
-        a1q, a1q_scale = moe_kernel_quantize_input(
-            a1,
-            quant_config.a1_gscale,
-            quant_config.quant_dtype,
-            quant_config.per_act_token_quant,
-            quant_config.block_shape,
-            is_fp4_scale_swizzled=not self.use_dp,
-        )
+        if not self.use_deepseek_fp8_block_scale:
```
pavanimajety, line 187: This might need to be guarded by flashinfer use rather than deepseek_fp8_block_scale? Because if we are using non flashinfer cutlass path we still need to quantize?

pavanimajety: Ah this is specific to flashinfer_cutlass_prepare_finalize

vllm/model_executor/layers/quantization/fp8.py
djmmoss: added

djmmoss:
@codex review

djmmoss:
@codex review

pavanimajety:
@djmmoss  Could you clarify if the lm_eval results are for DSR1? Why are they low (including the baseline triton numbers)?

djmmoss:
Yes, it's ready. (just keep in mind that unless the next FI release - the `uv` fix -  an older kernel will be used but it will be functional)

yewentao256 (approved):
LGTM, thanks for the work!

djmmoss:
@yewentao256 @robertgshaw2-redhat seems like the errors are unrelated to the changes?

vllm/model_executor/layers/quantization/fp8.py
```diff
@@ -1251,16 +1274,12 @@ def apply(
...
             )
         elif self.flashinfer_moe_backend == FlashinferMoeBackend.CUTLASS:
-            assert not self.block_quant
-            assert not renormalize and custom_routing_function is not None
             assert activation == "silu", (
                 f"Expected 'silu' activation but got {activation}"
             )
-            assert scoring_func == "sigmoid", (
-                f"Expected 'sigmoid' scoring func but got {scoring_func}"
-            )
```
mgoin, line 1261: Don't we still need to guard against the custom_routing_function and scoring_func being different from what we assume? I don't see these passed down anywhere

djmmoss: yea, we still need to guard in the non block_quant path, I've added it back in for that particular case.

PR #27139, opened 2025-10-18
[BugFix] fix graph partition signature

BoyuanFeng:
A pytorch 2.9 monkey patch to fix `test_attn_quant`. See pytorch/pytorch#165815

### DIFF
--- a/tests/compile/test_fusions_e2e.py
+++ b/tests/compile/test_fusions_e2e.py
@@ -116,7 +116,7 @@ def test_attn_quant(
     allreduce_fusions: int,
     custom_ops: str,
     inductor_graph_partition: bool,
-    caplog_vllm,
+    caplog_mp_spawn,
     monkeypatch,
 ):
     if backend == _Backend.FLASHINFER and (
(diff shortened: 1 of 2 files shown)


tests/compile/test_fusions_e2e.py
```diff
@@ -157,14 +157,11 @@ def test_attn_quant(
         inductor_compile_config={"force_disable_caches": True},
     )
 
-    with caplog_mp_spawn(logging.DEBUG) as log_holder:
+    with caplog_vllm.at_level(logging.DEBUG):
```
BoyuanFeng: This test does not use mp. `caplog_mp_spawn` would also fail for dynamo graph partition.

vllm/env_override.py
```diff
@@ -90,6 +90,156 @@ def get_output_names(graph_outputs) -> list[str]:
     assert len(planning_states) == 0
 
 
+# ===================================================
+# torch 2.9 Inductor get_graph_partition_signature monkeypatch
+# ===================================================
+# This change monkeypatches get_graph_partition_signature in pytorch 2.9.0 to
+# work around a test failure for
+# `tests/compile/test_fusions_e2e.py::test_attn_quant`.
```
ProExpertProg:
```suggestion
# fix inductor partition + attention-nvfp4 quant fusion, tested in
# `tests/compile/test_fusions_e2e.py::test_attn_quant`.
```

tests/compile/test_fusions_e2e.py
ProExpertProg: I wonder if that's just on your machine - for me even tp=1 tests used MP. I'll enable CI and we can see. But I think we should probably just do both, and we can investigate later

ProExpertProg (approved):

BoyuanFeng:
@ProExpertProg yes we should use `caplog_mp_spawn`. I was using `VLLM_ENABLE_V1_MULTIPROCESSING=0` for debugging. I guess that's the reason.

PR #27140, opened 2025-10-18
[NIXL] use Host buffer to support TP_ratio > 1 for XPU

xuechendi:
## Purpose

support XPU Decode/Prefill TP_ratio > 1 scenario with host_buffer

solution:

`KV_CONFIG='{"kv_connector":"NixlConnector","kv_role":"kv_both","kv_buffer_device":"cpu","**enable_permute_local_kv**":"True"'}`
If `enable_permute_local_kv` is enabled, initialize host_buffer with HND kv_layout and always for permute copy for device_to_host data copy and host_to_device data copy.

Since memory copy is inevitable, adding additional permute will not introduce performance overhead.

- Code change to general
1. introduce self.host_buffer_kv_cache_layout
2. remove XPU tp_ratio assert

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
+++ b/vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
@@ -872,7 +872,7 @@ class NixlConnectorWorker:
                 ):
                     logger.info_once(
                         "'enable_permute_local_kv' flag is enabled while "
-                        "device KV Layout with NHD, init host buffer with"
+                        "device KV Layout is NHD. Init host buffer with"
                         " HND to better support Decode/Prefill TP_ratio > 1."
                     )
                     # Since NHD will not support Decode/Prefill TP_ratio > 1,
--- a/vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
+++ b/vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
@@ -752,6 +752,7 @@ class NixlConnectorWorker:
         self._use_flashinfer = attn_backend == _Backend.FLASHINFER
         self._use_pallas = attn_backend == _Backend.PALLAS
         self.kv_cache_layout = get_kv_cache_layout()
+        self.host_buffer_kv_cache_layout = self.kv_cache_layout
         logger.debug("Detected attention backend %s", self.backend_name)
         logger.debug("Detected kv cache layout %s", self.kv_cache_layout)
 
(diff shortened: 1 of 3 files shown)


vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -782,6 +783,19 @@ def initialize_host_xfer_buffer(self, kv_caches: dict[str, torch.Tensor]) -> Non
...
+                    and self.vllm_config.kv_transfer_config.enable_permute_local_kv
+                ):
+                    logger.info_once(
+                        "'enable_permute_local_kv' flag is enabled while "
+                        "device KV Layout with NHD, init host buffer with"
+                        " HND to better support Decode/Prefill TP_ratio > 1."
+                    )
+                    # Since NHD will not support Decode/Prefill TP_ratio > 1,
+                    # we can leverage host_buffer for permute
+                    self.host_buffer_kv_cache_layout = "HND"
```
zhenwei-intel, line 880: Should we set it to `NHD` when homogeneous TP, tp ratio=1?

zhenwei-intel:
Nice work.
LGTM. Just one comment:)

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: No, if you do want to keep using NHD for decode/prefill TP_Ratio == 1, you can set "enable_permute_local_kv = False" for KV_transfer_config, so it will not do permute. 
However, permute + memcpy or memcpy won't have much perf diff, so I think we can leave enable_permute_local_kv = True for XPU until HND layout gets enabled.

xuechendi:
@NickLucche , may you help to review? We had internal sync, PR looks good from Intel GPU side, since did small changes to common codes as well, would like to see if you think it looks good to you

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -782,6 +783,20 @@ def initialize_host_xfer_buffer(self, kv_caches: dict[str, torch.Tensor]) -> Non
             for layer_name, kv_cache in kv_caches.items():
                 kv_shape = kv_cache.shape
                 kv_dtype = kv_cache.dtype
+                if (
+                    self.kv_cache_layout == "NHD"
+                    and self.vllm_config.kv_transfer_config is not None
+                    and self.vllm_config.kv_transfer_config.enable_permute_local_kv
+                ):
```
NickLucche, line 872: `self.vllm_config.kv_transfer_config is not None` this should always be true, is mypy complaining?

NickLucche (approved):
LGTM, nice use of the existing feature @xuechendi ! Only left a minor comment. We should also document the expected behavior of that flag somewhere.

Nit:
> Since memory copy is inevitable, adding additional permute will not introduce performance overhead.

_Technically_ the layout swapping will hit performance on cache locality. But we don't care about that here because:
a. we're enabling a capability that wouldn't otherwise work
b. we only need to keep post+read time < scheduler step. If we're 1000x faster than scheduler step, the engine still wouldn't notice

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -782,6 +783,20 @@ def initialize_host_xfer_buffer(self, kv_caches: dict[str, torch.Tensor]) -> Non
...
+                    logger.info_once(
+                        "'enable_permute_local_kv' flag is enabled while "
+                        "device KV Layout with NHD, init host buffer with"
```
NickLucche:
```suggestion
                        "device KV Layout is NHD. Init host buffer with"
```

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: yes, this is added to make mypy happy

PR #27142, opened 2025-10-18
[V0 Deprecation] Remove V0 executors

njhill:
`vllm.v1.executors.Executor` becomes the base executor class.


vllm/config/parallel.py
```diff
@@ -574,17 +574,17 @@ def _verify_args(self) -> Self:
...
-                and issubclass(self.distributed_executor_backend, ExecutorBase)
+                and issubclass(self.distributed_executor_backend, Executor)
             )
         ):
             raise ValueError(
                 "Unrecognized distributed executor backend "
                 f"{self.distributed_executor_backend}. Supported "
                 "values are 'ray', 'mp' 'uni', 'external_launcher', "
-                " custom ExecutorBase subclass or its import path."
+                " custom Executor subclass or its import path."
```
markmc, line 590: Presumably no backwards compat is needed for custom subclasses?

vllm/v1/engine/core.py
```diff
@@ -322,7 +322,6 @@ def step(self) -> tuple[dict[int, EngineCoreOutputs], bool]:
         with self.log_error_detail(scheduler_output):
             model_output = self.model_executor.execute_model(scheduler_output)
 
-        assert isinstance(model_output, ModelRunnerOutput)
```
markmc, line 325: Seems spurious?

vllm/v1/executor/multiproc_executor.py
```diff
@@ -181,7 +181,7 @@ def execute_model(
         self,
         scheduler_output: SchedulerOutput,
         non_block: bool = False,
-    ) -> ModelRunnerOutput | Future[ModelRunnerOutput]:
+    ):
```
markmc: Why remove this type hint?

markmc (approved):
Really nice to have this cleaned up!

vllm/config/parallel.py
njhill: Yes, `ExecutorBase` was the superclass for V0 and V1 executors, now that V0 is gone any that extended `ExecutorBase` directly wouldn't work anyhow.

vllm/v1/engine/core.py
```diff
@@ -463,14 +462,6 @@ def collective_rpc(
...
         return self.model_executor.collective_rpc(method, timeout, args, kwargs)
 
-    def save_tensorized_model(
-        self,
-        tensorizer_config,
-    ) -> None:
-        self.model_executor.save_tensorized_model(
-            tensorizer_config=tensorizer_config,
-        )
-
```
hmellor, line 473: Was this deleted intentionally?

vllm/envs.py
```diff
@@ -623,43 +621,24 @@ def get_vllm_port() -> int | None:
...
-    ),
-    # If the env var is set, it uses the Ray's Compiled Graph
-    # (previously known as ADAG) API which optimizes the
-    # control plane overhead.
-    # Run vLLM with VLLM_USE_RAY_COMPILED_DAG=1 to enable it.
-    # Note that this variable is set to 1 in V1 by default
-    # when ray distributed executor is used.
-    "VLLM_USE_RAY_COMPILED_DAG": lambda: bool(
-        int(os.getenv("VLLM_USE_RAY_COMPILED_DAG", "0"))
-    ),
```
hmellor, line 641: Were these V0 only features?

vllm/v1/engine/core.py
njhill: Yes, it's not used. The tensorizer model loader calls `collective_rpc` directly: https://github.com/vllm-project/vllm/blob/4d0f2661135959775d5b59315134e343d066d3ae/vllm/model_executor/model_loader/tensorizer.py#L734-L737

vllm/envs.py
njhill: They are always on in V1.

vllm/v1/engine/core.py
njhill: It was there to satisfy the type checking but no longer needed with the added `@overload`s.

vllm/v1/executor/multiproc_executor.py
njhill: mypy seems to complain otherwise ... that the signature is incompatible with the superclass. It seems to want it to match one of the overloads rather than the actual implementation signature.

I guess I could reinstate it and just add an ignore (done).

simon-mo (approved):

PR #27144, opened 2025-10-18
[Bugfix] fixes the decoding metadata of dense mla's fp8 kvcache.

sighingnow:
Require the flashmla patch https://github.com/vllm-project/FlashMLA/pull/7 to be landed first.


sighingnow:
@LucasWilkinson could you please take a look? Thanks!

LucasWilkinson (approved):
LGTM; will run evals and post here

edit: gsm8k looks good: DeepSeek-V2-Lite-Chat with fp8 kv-cache, 64.6%

PR #26686, opened 2025-10-13
[Model][7/N] Improve all pooling task | Deprecation as_reward_model. Extract hidden states prefer using new multi-vector retrieval API

noooop:
## TL;DR
Pooling models now default support all pooling, you can use it without any settings. 
- Extracting hidden states prefers using `token_embed` task. 
- Reward models prefers using `token_classify` task.

Add jina_embeddings_v4 example

## Improve all pooling task
- #25524 
- #27378
- #28194
- #26623
- #27145
- #26686

These PRs are mostly conflicting with each other, so combining them into a series would better inform reviewers about what happened. And what else needs to be done after that?

## Purpose

Fix #25165

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

Issue #24288, opened 2025-09-05
[RFC]: Support Returning Prompt Hidden States


charlotte12l:
### Motivation.

Hidden states are a valuable feature for many downstream applications, and their support has been requested and discussed in the vLLM community for a while(https://github.com/vllm-project/vllm/pull/15434 https://github.com/vllm-project/vllm/issues/4435 https://github.com/vllm-project/vllm/issues/6165 )

Currently, users who need both the prompt hidden states and the generated text must make two separate calls: LLM.encode (to obtain hidden states) and LLM.generate (to obtain generated text). This workflow is inconvenient and inefficient, especially for applications that require both outputs. The inability to return hidden states alongside generated text has also been highlighted in https://github.com/vllm-project/vllm/issues/12249
 

### Proposed Change.

This is a very initial draft PR: https://github.com/vllm-project/vllm/pull/24202. Will polish it after RFC.

## Configuration

Add `return_prompt_hidden_states` in Sampling Config. 

There is prior discussion https://github.com/vllm-project/vllm/pull/15434#discussion_r2050595239 mentioning using such request-level parameter can't be handled well. In local testing, this approach appears feasible for prompt hidden states, but feedback on potential disadvantages is welcome.

## How to build up prompt_hidden_states_tensor

The process for building the prompt hidden states tensor is similar to the existing _get_prompt_logprobs_dict implementation https://github.com/vllm-project/vllm/blob/d0944b273c861004cd08e77db4f823c246575124/vllm/v1/worker/gpu_model_runner.py#L203

The main difference is that we construct a CPU tensor of shape (num_prompt_tokens, hidden_size). During chunked prefill, we copy the relevant chunk slices into it.

## Hidden States Aggregation
To support flexible downstream use, we propose adding a plugin interface that allows users to perform custom aggregations on the hidden states (e.g., pooling, etc.).

## Output
When return_prompt_hidden_states is enabled, the output will include:
- prompt_hidden_states: a tensor of shape (num_prompt_tokens, hidden_size)

## Perf Concerns
- Default (disabled): No performance regression is expected.
- Enabled: Storing and serializing a large (num_prompt_tokens, hidden_size) tensor may impact performance, especially with long prompts. We will benchmark performance with large input lengths to assess the impact and determine if the regression is acceptable.

 

### Feedback Period.

1-3 weeks

### CC List.

@luccafong  @zhuohan123 

### Before submitting a new issue...

[...]

PR #27145, opened 2025-10-18
[Model][6/N] Improve all pooling task | Support chunked prefill with ALL pooling

noooop:
## TL;DR
- For pooling models with causal attention that use all (and last) pooling, now supports chunked prefill and prefix caching, and is enabled by default. e.g. internlm/internlm2-1_8b-reward
- Now pooling models can use the offline encode interface and online pooling interface to use all pooling. No additional settings required. Previously, ALL pooling did not support chunked prefill; you need to manually set enable_chunked_prefill=False to use it. e.g. extract_hidden_states: tests/models/language/pooling/test_extract_hidden_states.py

## Improve all pooling task
- #25524 
- #27378
- #28194
- #26623
- #27145
- #26686

## Purpose

Address: 
- https://github.com/vllm-project/vllm/issues/25165#issuecomment-3310400937 
- https://github.com/vllm-project/vllm/issues/24288#issuecomment-3341399939

Let's support chunked prefill with ALL pooling

We are moving further towards supporting both generation and Returning Prompt Hidden States simultaneously

## Test Plan

tests/models/language/pooling/test_all_pooling_plus_chunked_prefill.py

tests/entrypoints/pooling/classify/test_offline.py::test_token_classify
tests/entrypoints/pooling/classify/test_online.py::test_pooling_token_classify
tests/entrypoints/pooling/reward/test_offline.py::test_config
tests/test_config.py::test_is_prefix_caching_supported
tests/test_config.py::test_is_chunked_prefill_supported

## Test Result

pass

---
<details>

</details>


tests/models/language/pooling/test_all_pooling_plus_chunked_prefill.py
```diff
@@ -0,0 +1,57 @@
...
+        runner="pooling",
+        long_prefill_token_threshold=chunk_size,
+        max_model_len=128,
+        enforce_eager=True,
+        enable_chunked_prefill=True,
+        # If enable_prefix_caching is enabled,
+        # the output of all pooling will be less than n_prompt_tokens,
+        # we need a method to disable prefix_caching at the request level.
+        enable_prefix_caching=False,
+        max_num_batched_tokens=chunk_size,
```
noooop: Another strange problem was discovered at the same time

noooop: @heheda12345

I want all pooling requests to not use prefix_caching, even if enable_prefix_caching=True at the request level. Otherwise, the output hidden_states would be fewer than n_prompt_tokens.

However, I don't want to completely disable prefix_caching when using all pooling, because enabling prefix_caching can accelerate scenarios where generation and returning prompt hidden states are performed simultaneously.

PTAL https://github.com/vllm-project/vllm/issues/24288#issuecomment-3341399939

> If you turn on Automatic Prefix Caching, submit LLM.encode(..., pooling_task="token_embed") first, then submit LLM.generate. I think it should be no cost in making two separate calls.

maxdebayser:
Thanks for completing this. Back when added support in V1 for pooling and chunked prefill, I left the addition of support for the stateful poolers as a future exercise. My main concern in this PR is that the PoolingParams object is not a good place to store this state.

I have two different suggestions of places where the state could go:

1. PoolingMetadata

The pooling metadata object could have a state dict that could be used by the poolers. The GPU input batch would have to be modified to handle persistent pooling metadata objects as well as the pooling params object. The advantage here is that in the case of request cancellation we can easily remove the pooling metadata.

2.  The pooler

If we add the request ID in the pooling metadata, then the pooler can have a dict of request ID to state. The only challenge would be to get notified of request cancellation to remove the state.

BTW, the it's easy to extend the chunked prefill support for mean pooling as well. It doesn't even need to store the full hidden states, only the current average and the count of tokens has to be stored.

noooop:
> I think the easiest of the suggestions would be to add the state to the PoolingMetadata.

https://github.com/vllm-project/vllm/blob/bfe0b4bd2a4f8c7d2bcc0608556a91bdf5dd4566/vllm/v1/worker/gpu_input_batch.py#L834-L841

Each executor step creates a new PoolingMetadata. If hidden_states_cache is in PoolingMetadata, the position needs to be moved every step . More importantly, during two executions, some requests might not be scheduled, and these unscheduled requests' hidden_states_cache don't know where to place them.

vllm/pooling_params.py
```diff
@@ -61,6 +62,9 @@ class PoolingParams(
     extra_kwargs: dict[str, Any] | None = None
     output_kind: RequestOutputKind = RequestOutputKind.FINAL_ONLY
 
+    # for chunked prefill with ALL pooling
+    hidden_states_cache: list[torch.Tensor] = msgspec.field(default_factory=list)
```
DarkLight1337: Can we put the cache in the request state instead? `PoolingParams` is more user-facing.

DarkLight1337 (commented):
Please also update the documentation to show that chunked prefill is supported for ALL pooling

vllm/model_executor/models/terratorch.py
hmellor, line 1: Is there a change that should be made in `vllm/model_executor/models/transformers/pooling.py` too?

We don't use `@default_pooling_type` but we do set the `default_pooling_type` attribute (which is all the decorator does).

noooop: This can also support the Transformers backend.

PTAL: #26623

If you want to further confirm, you can add the Transformers backend to tests/test_config.py.

***

What is this PR doing?

- For pooling models with causal attention that use all (and last) pooling, now supports chunked prefill and prefix caching, and is enabled by default. e.g. internlm/internlm2-1_8b-reward
- Now pooling models can use the offline encode interface and online pooling interface to use all pooling. No additional settings required. Previously, ALL pooling did not support chunked prefill; you need to manually set enable_chunked_prefill=False to use it. e.g. extract_hidden_states: tests/models/language/pooling/test_extract_hidden_states.py

vllm/pooling_params.py
noooop: https://github.com/vllm-project/vllm/blob/d0cd728907d82a109a165612b8790ddaf5496f59/vllm/v1/pool/metadata.py#L34-L42

Because PoolingMetadata does not have a reference to Request, it is difficult to get it in the pooler and put the hidden_states into hidden_states_cache

We don't need to make significant modifications to the runner for this feature.

DarkLight1337: How about inside `pooling_cursor`?

noooop: Each step will build a new PoolingCursor

DarkLight1337: Can we put this inside `PoolingMetadata` directly then? I really don't want to expose internals inside `PoolingParams`.

DarkLight1337: Another way would be create an internal version of `PoolingParams` that is attached to the request, which is different from the user-facing `PoolingParams`

vllm/pooling_params.py
```diff
@@ -5,13 +5,20 @@
...
 from vllm.sampling_params import RequestOutputKind
 from vllm.tasks import PoolingTask
 
 
+class PoolingParamsInternalStates:
+    def __init__(self):
+        # for chunked prefill with ALL pooling
+        self.hidden_states_cache: list[torch.Tensor] = []
+
+
```
noooop: @DarkLight1337 

how about PoolingParamsInternalStates

DarkLight1337: Just `PoolingStates` would be fine

vllm/pooling_params.py
```diff
@@ -5,13 +5,20 @@
...
 import msgspec
+import torch
 
 from vllm.config import ModelConfig, PoolerConfig
 from vllm.config.pooler import get_use_activation
 from vllm.sampling_params import RequestOutputKind
 from vllm.tasks import PoolingTask
 
 
+class PoolingStates:
```
DarkLight1337: Maybe move this to belong with the V1 request definitions

DarkLight1337: I prefer adding PoolingStates as a separate attribute in the request state, instead of being inside PoolingParams

noooop: May cause a circular reference

I think defining PoolingStates in request.py may confuse many users who only use the generation features. 

After all, all pooling is a feature used by very few people.

noooop: > I prefer adding PoolingStates as a separate attribute in the request state, instead of being inside PoolingParams

Yes, I also want to put it into the Request

https://github.com/vllm-project/vllm/blob/d0cd728907d82a109a165612b8790ddaf5496f59/vllm/v1/pool/metadata.py#L34-L42

Because PoolingMetadata does not have a reference to Request, it is difficult to get it in the pooler.

noooop: Do we really have to modify a lot of code for this feature?

noooop: This plan is indeed better; let me refactor it.

vllm/v1/worker/gpu_input_batch.py
```diff
@@ -51,11 +50,18 @@ class CachedRequestState:
...
+
     def __post_init__(self):
         self.num_prompt_tokens = length_from_prompt_token_ids_or_embeds(
             self.prompt_token_ids, self.prompt_embeds
         )
 
+        if self.pooling_params is not None:
+            self.pooling_states = PoolingStates()
+
     @property
```
noooop, line 65: @DarkLight1337 

how about now

vllm/v1/pool/metadata.py
```diff
@@ -47,36 +62,44 @@ def __getitem__(self, indices: slice):
...
+    seq_lens_cpu: torch.Tensor,
+    prompt_lens: torch.Tensor,
+    device: torch.device,
 ):
     assert len(prompt_lens) == len(num_scheduled_tokens)
 
     n_seq = len(num_scheduled_tokens)
     index = list(range(n_seq))
-    num_scheduled_tokens_cpu = torch.tensor(num_scheduled_tokens, device="cpu")
+    num_scheduled_tokens = torch.tensor(num_scheduled_tokens, device="cpu")
```
DarkLight1337: This is intentionally named `num_scheduled_tokens_cpu` to avoid mypy errors

vllm/v1/worker/gpu_input_batch.py
DarkLight1337: Yeah looks better

DarkLight1337 (approved):

PR #27147, opened 2025-10-18
[MM Encoder]: Wrap mm encoder attention interface as `CustomOps`

Isotr0py:
## Purpose
- Currently, `MultiHeadAttention` become quite messy because we're using upstream FA and attention backend selection logic is quite coupling.
- Furthermore, some out-of-tree hardware plugin would like to have its own mm encoder forward implementation.
- Therefore, this PR split `MultiHeadAttention` into `mm_encoder_attn.py` file and wrap it with `CustomOps`.

## Test Result

---
<details>

</details>


Isotr0py:
Also cc @shen-shanshan about OOT hardware. I remember Ascend also working on this similarly?

shen-shanshan:
Could we replace `Qwen2_5_VisionAttention` with this `MMEncoderAttention`? Or we have to extract another `CustomOp` for this? 🧐

vllm/attention/layers/mm_encoder_attention.py
```diff
@@ -0,0 +1,394 @@
...
+                from vllm.attention.utils.fa_utils import flash_attn_varlen_func
+    else:
+        flash_attn_varlen_func = None
+
+    return attn_backend, flash_attn_varlen_func
+
+
+@CustomOp.register("mm_encoder_attn")
+class MMEncoderAttention(CustomOp):
+    """Multi-headed attention without any cache, used for multimodal encoder."""
```
NickLucche, line 119: nit on naming: do we need to include "MM" here in the class name instead of MHA? _Technically_ this is also used by whisper and is not semantically tied to multimodal models..

Isotr0py: I added "MM" here to keep name consistent with the arguments like `--mm-encoder-attn-backend` etc, so that user and developer can easily know that this layer should be controlled by `--mm-encoder-attn-backend` here.

(BTW, I think previous `MultiHeadAttention` naming is a bit confusing because there is a layer with same name in `torch.nn` 😅)

DarkLight1337:
Closing as superseded by https://github.com/vllm-project/vllm/pull/30125

DarkLight1337:
Feel free to reopen if you want to reuse this PR instead of opening a new one for the next step.

PR #27155, opened 2025-10-18
Add auto max model len for available memory with `--max-model-len -1`

mgoin:
## Summary
- Added tracking for automatic max model length selection so --max-model-len -1 records the model’s default limit before memory sizing.
- Taught the KV cache memory check to shrink the configured context length when memory is insufficient and to log the chosen automatic cap instead of immediately failing.
- Documented the new auto-sizing flag and added regression coverage for CLI parsing and KV cache auto-adjustment behavior.

## Testing
- pytest tests/v1/core/test_kv_cache_utils.py::test_auto_max_model_len_adjusts_to_available_memory -q *(fails: ModuleNotFoundError: No module named 'tblib')*

------
https://chatgpt.com/codex/tasks/task_e_68f28bceee008329b78d9f4361fb43b9

### DIFF
--- a/vllm/v1/core/kv_cache_utils.py
+++ b/vllm/v1/core/kv_cache_utils.py
@@ -663,10 +663,39 @@ def check_enough_kv_cache_memory(
     needed_memory = max_memory_usage_bytes(vllm_config, kv_cache_spec.values())
 
     if needed_memory > available_memory:
+        default_max_len = (
+            vllm_config.model_config.get_auto_max_model_len_default()
+            or max_model_len
+        )
         # Estimate the maximum model length that can fit in the available memory
         estimated_max_len = estimate_max_model_len(
             vllm_config, kv_cache_spec, available_memory
         )
+
+        if vllm_config.model_config.uses_auto_max_model_len():
+            if estimated_max_len <= 0:
+                raise ValueError(
+                    "Unable to automatically determine a max model length "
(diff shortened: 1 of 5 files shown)


vllm/v1/core/kv_cache_utils.py
```diff
@@ -663,10 +663,39 @@ def check_enough_kv_cache_memory(
...
+                )
+            else:
+                logger.info(
+                    "Automatic max model length capped at %d tokens to fit "
+                    "available memory (model default %d).",
+                    estimated_max_len,
+                    default_max_len,
+                )
+            vllm_config.recalculate_max_model_len(estimated_max_len)
+            return
```
noooop, line 697: There's a very troublesome issue here: 
vllm v1 uses a multiprocessing architecture. 
If using tp pp, call vllm_config.recalculate_max_model_len  will not affect other processes.

https://github.com/vllm-project/vllm/blob/221bf725772b891db27e1b9200c36a721ba9a304/vllm/config/vllm.py#L718-L723

This is why I added the comment "Can only be called in try_verify_and_update_config", at this time only the main process is started and initializing the "unique" VllmConfig

Maybe we need to do one or more collective_rpc to synchronize max_model_len.

mgoin:
Closed for https://github.com/vllm-project/vllm/pull/29431

PR #27156, opened 2025-10-18
[Feature] Pydantic validation for speculative.py

Navya1707:


vllm/config/speculative.py
```diff
@@ -268,10 +268,16 @@ def __post_init__(self):
...
+                        "Either prompt_lookup_max or prompt_lookup_min must be provided when using the ngram method."
+                    )
                 self.prompt_lookup_min = self.prompt_lookup_max
             elif self.prompt_lookup_max is None:
-                assert self.prompt_lookup_min is not None
+                if self.prompt_lookup_min is None:
+                    raise ValueError(
+                        "Either prompt_lookup_max or prompt_lookup_min must be provided when using the ngram method."
+                    )
                 self.prompt_lookup_max = self.prompt_lookup_min
```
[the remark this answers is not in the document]
Navya1707: @hmellor - please let me know if this implementation makes sense, happy to make the changes!

vllm/config/speculative.py
```diff
@@ -62,7 +62,7 @@ class SpeculativeConfig:
     enforce_eager: bool | None = None
     """Override the default enforce_eager from model_config"""
     # General speculative decoding control
-    num_speculative_tokens: SkipValidation[int] = None  # type: ignore
+    num_speculative_tokens: int | None = Field(default=None, gt=0)  # type: ignore
```
hmellor: This field will always be `int` after initialisation, so let's not hint it as `None`

```suggestion
    num_speculative_tokens: int = Field(default=None, gt=0)
```

vllm/config/speculative.py
```diff
@@ -215,7 +215,8 @@ def hf_config_override(hf_config: PretrainedConfig) -> PretrainedConfig:
 
         return hf_config
 
-    def __post_init__(self):
+    @model_validator(mode="after")
+    def _configure(self) -> Self:
```
hmellor: Most of this is post initialisation, not validation. 

```suggestion
    def __post_init__(self):
```

vllm/config/speculative.py
```diff
@@ -268,10 +268,18 @@ def __post_init__(self):
...
             elif self.prompt_lookup_max is None:
-                assert self.prompt_lookup_min is not None
+                if self.prompt_lookup_min is None:
+                    raise ValueError(
+                        "Either prompt_lookup_max or prompt_lookup_min must be "
+                        "provided when using the ngram method."
+                    )
                 self.prompt_lookup_max = self.prompt_lookup_min
 
             # Validate values
```
hmellor, line 284: These are now covered by `ge=...`, right? So they can be removed from here

hmellor (changes requested):

Navya1707:
@hmellor  - updated with the requested changes!

hmellor (approved):
Thanks! Let's see what CI thinks

PR #27018, opened 2025-10-16
[Kernel][Perf] fuse QK Norm and RoPE into one cuda kernel for Qwen Model

izhuhaoran:
## Purpose

This PR fuses QNorm, KNorm, and RoPE into a single CUDA kernel for the Qwen3 model, improving inference performance. And user can enable this fusion by `VLLM_FUSE_QKNORM_AND_ROPE=1`

### GPU Trace

- Main
image

- This PR
image

### Bench Serve
setting: NVIDIA H20, TP=2, model=qwen3-30b-a3b-fp8, num_prompts=32, max-concurrency=8, input_len=out_len=1024
result: TTFT from 388.56ms to 192.26ms   &&   TPOT from 10.29ms to 9.93ms

- Main

```
============ Serving Benchmark Result ============
Successful requests:                     32        
Maximum request concurrency:             8         
Benchmark duration (s):                  43.66     
Total input tokens:                      32620     
Total generated tokens:                  32768     
Request throughput (req/s):              0.73      
Output token throughput (tok/s):         750.48    
Peak output token throughput (tok/s):    792.00    
Peak concurrent requests:                16.00     
Total Token throughput (tok/s):          1497.57   
---------------Time to First Token----------------
Mean TTFT (ms):                          388.56    
Median TTFT (ms):                        223.63    
P99 TTFT (ms):                           1070.85   
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          10.29     
Median TPOT (ms):                        10.26     
P99 TPOT (ms):                           10.99     
---------------Inter-token Latency----------------
Mean ITL (ms):                           10.29     
Median ITL (ms):                         10.23     
P99 ITL (ms):                            10.70     
==================================================
```

- This PR

```
============ Serving Benchmark Result ============
Successful requests:                     32        
Maximum request concurrency:             8         
Benchmark duration (s):                  41.40     
Total input tokens:                      32768     
Total generated tokens:                  32768     
Request throughput (req/s):              0.77      
Output token throughput (tok/s):         791.47    
Peak output token throughput (tok/s):    824.00    
Peak concurrent requests:                16.00     
Total Token throughput (tok/s):          1582.95   
---------------Time to First Token----------------
Mean TTFT (ms):                          192.26    
Median TTFT (ms):                        207.59    
P99 TTFT (ms):                           223.33    
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          9.93      
Median TPOT (ms):                        9.88      
[...]

PR #27165, opened 2025-10-19
[Kernel][Perf] fuse QK Norm and RoPE into one cuda kernel for Qwen Model

izhuhaoran:
## Purpose

Inspired by TensorRT-LLM. This PR is a follow PR about #27018 , and fuses QNorm, KNorm, and RoPE into a single CUDA kernel for the Qwen3 model, improving inference performance. We convert this fusion into a custom torch.compile pass, users can enable it by:

```
 --compilation-config='{"use_inductor": 1,  "pass_config": {"enable_qk_norm_rope_fusion": 1}}'
```

More details see #27018 

### Result GPU Trace

- Main - No Inductor, all custom op
image

- Main - use inductor

image

- This PR

image

---
<details>

</details>


vllm/compilation/qk_norm_rope_fusion.py
```diff
@@ -0,0 +1,275 @@
...
+
+            # Split back to q,k,v and return
+            split_tuple = SPLIT_SIZES_OP(
+                qkv, [self.q_size, self.kv_size, self.kv_size], -1
+            )
+            return (
+                operator.getitem(split_tuple, 0),
+                operator.getitem(split_tuple, 1),
+                operator.getitem(split_tuple, 2),
+            )
```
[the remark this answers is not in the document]
izhuhaoran: already solve it.

vllm/envs.py
```diff
@@ -1295,6 +1314,10 @@ def get_vllm_port() -> int | None:
         None,
         ["flashinfer-cudnn", "flashinfer-trtllm", "flashinfer-cutlass"],
     ),
+    # If set, use the fuse QKNorm and RoPE kernel
+    "VLLM_FUSE_QKNORM_AND_ROPE": lambda: bool(
+        int(os.getenv("VLLM_FUSE_QKNORM_AND_ROPE", "0"))
+    ),
```
ZJY0516: We should use pass config instead of env vars.

izhuhaoran: > We should use pass config instead of env vars.

The PR is still WIP. This environment variable comes from #27018 . This PR is currently attempting to convert it into a custom pass. Users will need to use (as you mentioned):
`--compilation_config='{"use_inductor": 1, "pass_config": {"enable_qk_norm_rope_fusion": 1}}'`
Once this PR is completed, these old contents will be cleaned up.

vllm/compilation/fix_functionalization.py
```diff
@@ -132,6 +132,9 @@ def __call__(self, graph: torch.fx.Graph):
                         "input_global_scale",
                     ),
                 )
+            # # Defunctionalize fused_qk_norm_rope to remove higher-order wrapper.
```
ProExpertProg: Is this supposed to be removed or uncommented?

vllm/compilation/qk_norm_rope_fusion.py
```diff
@@ -0,0 +1,278 @@
...
+            cos_sin_cache: torch.Tensor,
+        ):
+            # split qkv -> q,k,v
+            # q, k, v = qkv.split([self.q_size, self.kv_size, self.kv_size], dim=-1)
+            split_tuple = SPLIT_SIZES_OP(
+                qkv, [self.q_size, self.kv_size, self.kv_size], -1
+            )
+            q = operator.getitem(split_tuple, 0)
+            k = operator.getitem(split_tuple, 1)
+            v = operator.getitem(split_tuple, 2)
```
ProExpertProg: I think that this should work, pattern tracing is very close to forward code tracing:
```suggestion
            # split qkv -> q,k,v
            q, k, v = qkv.split([self.q_size, self.kv_size, self.kv_size], dim=-1)
```

vllm/compilation/qk_norm_rope_fusion.py
```diff
@@ -0,0 +1,278 @@
...
+            # q_by_head=q.view(*q.shape[:-1],q.shape[-1]//self.head_dim,self.head_dim)
+            # q_out = torch.empty_like(q_by_head)
+            # q_by_head_contiguous = q_by_head.contiguous()
+            q_by_head = VIEW_OP(
+                q, (*q.shape[:-1], q.shape[-1] // self.head_dim, self.head_dim)
+            )
+            q_out = EMPTY_LIKE_OP(q_by_head)
+            q_by_head_contiguous = CONTIGUOUS_OP(q_by_head)
+
+            qn = auto_functionalized(
```
ProExpertProg: Please use `MatcherRMSNorm` so that we can match even with `rms_norm` disabled (using torch impl in `forward_native`)

vllm/compilation/qk_norm_rope_fusion.py
```diff
@@ -0,0 +1,278 @@
...
+        inputs = [
+            qkv,
+            positions,
+            q_weight,
+            k_weight,
+            cos_sin_cache,
+        ]
+
+        # # Register variants across rope ops and with/without contiguous()
+        # # Ensure view ops are canonicalized to reshape in the traced pattern
```
ProExpertProg: Was this not needed?

vllm/compilation/qk_norm_rope_fusion.py
```diff
@@ -0,0 +1,278 @@
...
+                            e,
+                        )
+
+        # Dump patterns for debugging if enabled
+        self.dump_patterns(config, self.patterns)
+
+    @VllmInductorPass.time_and_log
+    def __call__(self, graph: fx.Graph) -> None:
+        if not current_platform.is_cuda_alike():
+            return
```
ProExpertProg: Silent disablement is not good. Should not be enabled at all on non-cuda-alike platforms

vllm/compilation/qk_norm_rope_fusion.py
```diff
@@ -0,0 +1,278 @@
...
+            # RoPE: apply to flattened q/k
+            rope = auto_functionalized(
+                self.rope_op,
+                positions=positions,
+                query=q_flat,
+                key=k_flat,
+                head_size=self.head_dim,
+                cos_sin_cache=cos_sin_cache,
+                is_neox=self.is_neox,
+            )
```
ProExpertProg: Can you create a new MatcherRoPE for this so that we can pattern match torch rope as well?

vllm/compilation/qk_norm_rope_fusion.py
```diff
@@ -0,0 +1,278 @@
...
+            return
+
+        # Register a pattern per attention layer, as sizes differ by shard
+        attn_layers = get_layers_from_vllm_config(config, Attention)
+        if len(attn_layers) == 0:
+            logger.warning(
+                "QK Norm+RoPE fusion enabled, but no Attention layers were discovered."
+            )
+            return
+        layer_name, layer = next(iter(attn_layers.items()))
```
ProExpertProg: This will only register the pattern using one layer, is that intended? Are we sure this will always pick the same layer also?

vllm/compilation/qk_norm_rope_fusion.py
```diff
@@ -0,0 +1,278 @@
...
+    """
+
+    def __init__(
+        self,
+        layer: Attention,
+        eps: float,
+        rope_op: torch._ops.OpOverload,
+        is_neox: bool,
+    ) -> None:
+        self.layer = layer
```
ProExpertProg: Are you just using layer for these sizes? If yes, don't save the layer object on the pattern object, just extract the size properties

vllm/compilation/qk_norm_rope_fusion.py
ProExpertProg: I see now you don't care which layer because you only need the shapes, please add a comment for that

vllm/config/compilation.py
```diff
@@ -546,6 +548,10 @@ def __post_init__(self) -> None:
         if isinstance(self.pass_config, dict):
             self.pass_config = PassConfig(**self.pass_config)
 
+        if self.pass_config.enable_qk_norm_rope_fusion:
+            self.custom_ops.append("+rms_norm")
+            self.custom_ops.append("+rotary_embedding")
```
ProExpertProg, line 654: Let's try to remove this requirement; definitely for rms_norm and hopefully for RoPE as well although RoPE is less important - I assume all custom RoPE ops would be fused away anyway right?

vllm/model_executor/models/qwen3.py
```diff
@@ -117,6 +119,13 @@ def __init__(
             rope_scaling=rope_scaling,
             dual_chunk_attention_config=dual_chunk_attention_config,
         )
+        # Determine if we can use fused QK norm + RoPE
```
ProExpertProg: Please remove model definition changes now that we have the fusion pass

vllm/model_executor/models/qwen3_moe.py
```diff
@@ -275,6 +277,13 @@ def __init__(
             rope_scaling=rope_scaling,
             dual_chunk_attention_config=dual_chunk_attention_config,
         )
+        # Determine if we can use fused QK norm + RoPE
```
ProExpertProg: Here as well

ProExpertProg (commented):
This looks like the right approach! Once you're done, please clean up the code and add E2E performance and lm-eval numbers.

vllm/compilation/fix_functionalization.py
izhuhaoran: done, I  uncommented this.

vllm/compilation/qk_norm_rope_fusion.py
izhuhaoran: done

vllm/compilation/qk_norm_rope_fusion.py
izhuhaoran: done

vllm/compilation/qk_norm_rope_fusion.py
izhuhaoran: removed, done.

vllm/compilation/qk_norm_rope_fusion.py
izhuhaoran: Done, for non-cuda-alike platforms, logger.warnning during PassConfig init.

vllm/compilation/qk_norm_rope_fusion.py
izhuhaoran: Aleady add a MatcherRotaryEmbedding

vllm/compilation/qk_norm_rope_fusion.py
izhuhaoran: thanks for suggestion, already do it.

vllm/compilation/qk_norm_rope_fusion.py
izhuhaoran: Done, only pass in the shape args now. Thanks.

vllm/config/compilation.py
izhuhaoran: Now, the native RoPE forward isn't matching as expected (see details in Slack). I've kept only `+rotary_embedding` and added a TODO comment.

vllm/model_executor/models/qwen3.py
izhuhaoran: Already cleaned up.

vllm/model_executor/models/qwen3_moe.py
izhuhaoran: Already cleaned up. Thanks again

vllm/compilation/qk_norm_rope_fusion.py
```diff
@@ -0,0 +1,236 @@
...
+                num_heads_q=self.num_heads,
+                num_heads_k=self.num_kv_heads,
+                num_heads_v=self.num_kv_heads,
+                head_dim=self.head_dim,
+                eps=self.eps,
+                q_weight=q_weight,
+                k_weight=k_weight,
+                cos_sin_cache=cos_sin_cache,
+                is_neox=self.is_neox,
+                position_ids=positions.view(-1),
```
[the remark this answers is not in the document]
izhuhaoran: already add dtype check

vllm/compilation/qk_norm_rope_fusion.py
```diff
@@ -0,0 +1,244 @@
...
+        ]
+
+        pm.register_replacement(
+            pattern,
+            replacement,
+            inputs,
+            QkNormRopePattern.wrap_trace_fn(
+                pm.fwd_only,
+                QkNormRopePattern.fx_view_to_reshape,
+            ),
```
ZJY0516, line 176: Could you please add a comment for why we should do this here?

izhuhaoran: `fx_view_to_reshape` is a common technique used by Inductor itself to convert view operations to reshape operations for better pattern matching capability. Our implementation here follows the same approach as used in `vllm/compilation/fusion_attn.py`. Since this is a common practice in the compilation pass, I think we don't need to add a specific comment here.

ZJY0516: But most `register_replacement` used in vllm are like `register_replacement(pattern, replacement, inputs, fwd_only, pm_pass)`. For maintainability, a brief comment would be valuable to explain why this additional transformation is necessary here.

izhuhaoran: ok, already add a note comment for it.

izhuhaoran:
> I think we should add the test into `test-pipeline.yaml`

Thanks for the suggestion — I’ve already added it.

vllm/compilation/matcher_utils.py
```diff
@@ -85,10 +162,12 @@ def forward_custom(
         weight: torch.Tensor,
     ) -> torch.Tensor:
         result = torch.empty_like(input)
+        # TODO: support non-contiguous input for RMSNorm and remove this
+        input_contiguous = input.contiguous()
```
ProExpertProg, line 166: Can you attach a GH issue link for this TODO?

vllm/compilation/pass_manager.py
```diff
@@ -107,8 +108,12 @@ def configure(self, config: VllmConfig):
                 self.passes += [ActivationQuantFusionPass(config)]
 
             if self.pass_config.enable_attn_fusion:
+                logger.debug("Enabling AttnFusionPass")
```
ProExpertProg: Remove debug

vllm/compilation/qk_norm_rope_fusion.py
```diff
@@ -0,0 +1,247 @@
...
+            cos_sin_cache = empty_fp32(4096, self.head_dim)
+        else:
+            cos_sin_cache = empty_bf16(4096, self.head_dim)
+        inputs = [
+            qkv,
+            positions,
+            q_weight,
+            k_weight,
+            cos_sin_cache,
+        ]
```
ProExpertProg: Can you put this into a method called `get_inputs`?

vllm/compilation/qk_norm_rope_fusion.py
```diff
@@ -0,0 +1,247 @@
...
+    @enable_fake_mode
+    def __init__(self, config: VllmConfig):
+        super().__init__(config)
+        self.patterns: PatternMatcherPass = PatternMatcherPass(
+            pass_name="qk_norm_rope_fusion_pass"
+        )
+
+        if not current_platform.is_cuda_alike():
+            logger.warning_once("QK Norm+RoPE fusion not enabled: unsupported platform")
+            return
```
ProExpertProg: This is overly defensive, we should never reach this point on non-cuda-alike platforms

vllm/config/compilation.py
ProExpertProg: Can you add an issue link

vllm/compilation/qk_norm_rope_fusion.py
```diff
@@ -0,0 +1,247 @@
...
+            positions: torch.Tensor,
+            q_weight: torch.Tensor,
+            k_weight: torch.Tensor,
+            cos_sin_cache: torch.Tensor,
+        ):
+            # split qkv -> q,k,v
+            q, k, v = qkv.split([self.q_size, self.kv_size, self.kv_size], dim=-1)
+
+            # Q path: view -> RMS -> view back to q.shape
+            q_by_head = RESHAPE_OP(
```
ProExpertProg: I think RESHAPE_OP is not necessary, I think just doing `.view()` produces the same pattern

vllm/_custom_ops.py
```diff
@@ -329,6 +329,7 @@ def rms_norm(
     out: torch.Tensor, input: torch.Tensor, weight: torch.Tensor, epsilon: float
 ) -> None:
     # TODO: Remove this contiguous call when the kernel is updated to support non-contiguous input
+    # If removed, also need to remove contiguous in MatcherRMSNorm
```
ProExpertProg, line 332: Issue link

vllm/compilation/qk_norm_rope_fusion.py
```diff
@@ -0,0 +1,247 @@
...
+from .fusion import empty_bf16, empty_fp32, empty_i64
+from .inductor_pass import enable_fake_mode
+from .matcher_utils import MatcherRMSNorm, MatcherRotaryEmbedding
+from .vllm_inductor_pass import VllmInductorPass, VllmPatternMatcherPass
+
+logger = init_logger(__name__)
+
+FUSED_QK_ROPE_OP = torch.ops._C.fused_qk_norm_rope.default
+SPLIT_SIZES_OP = torch.ops.aten.split_with_sizes.default
+RESHAPE_OP = torch.ops.aten.reshape.default
```
ProExpertProg: Just use the direct methods on tensors

vllm/compilation/qk_norm_rope_fusion.py
```diff
@@ -0,0 +1,247 @@
...
+        # use one attn layer to get meta (such as head_dim) for QkNormRopePattern
+        attn_layers: dict[str, Attention] = get_layers_from_vllm_config(
+            config, Attention
+        )
+        if len(attn_layers) == 0:
+            logger.warning_once(
+                "QK Norm+RoPE fusion enabled, but no Attention layers were discovered."
+            )
+            return
+        _, layer = next(iter(attn_layers.items()))
```
ProExpertProg: nit: 
```suggestion
        layer = next(iter(attn_layers.values()))
```

ProExpertProg (commented):
Can you also add a unit test for the fused kernel?

vllm/compilation/matcher_utils.py
izhuhaoran: > Can you attach a GH issue link for this TODO?

This TODO is related to an existing one already present in the main branch (see below). Do we still need to open a separate issue for it?
https://github.com/vllm-project/vllm/blob/14a125a06df7275923fe9748f67e27e449412d1f/vllm/_custom_ops.py#L331-L332

vllm/compilation/qk_norm_rope_fusion.py
izhuhaoran: Done

vllm/compilation/qk_norm_rope_fusion.py
izhuhaoran: Done, now directly use `.view()`

vllm/compilation/qk_norm_rope_fusion.py
izhuhaoran: Done.

vllm/compilation/qk_norm_rope_fusion.py
izhuhaoran: Agree, have removed this

vllm/compilation/qk_norm_rope_fusion.py
izhuhaoran: Done.

vllm/_custom_ops.py
izhuhaoran: see above: https://github.com/vllm-project/vllm/pull/27165#discussion_r2488553732

izhuhaoran:
> Can you also add a unit test for the fused kernel?

I've already added a unit test for the fused kernel and addressed your other comments as well. PTAL when you have time.

vllm/config/compilation.py
izhuhaoran: Linked issue: https://github.com/vllm-project/vllm/issues/28042

vllm/compilation/matcher_utils.py
izhuhaoran: Also a follow PR to remove this non-contiguous todo:
https://github.com/vllm-project/vllm/pull/28103

vllm/model_executor/layers/rotary_embedding/base.py
```diff
@@ -101,35 +101,56 @@ def __init__(
             head_size, rotary_dim, max_position_embeddings, base, is_neox_style, dtype
         )
 
-    def forward_native(
-        self,
+    @staticmethod
```
pavanimajety, line 101: Curious why is this needed to be static method?

ProExpertProg: Like RMSNorm, if we want to use the forward method for matching it needs to be static so we can trace through it in the matcher object

csrc/fused_qknorm_rope_kernel.cu
```diff
@@ -0,0 +1,426 @@
...
+    } else if (cache_dtype == torch::kHalf) {                                \
+      tensorrt_llm::kernels::launchFusedQKNormRope<qkv_type, c10::Half>(     \
+          __VA_ARGS__);                                                      \
+    } else if (cache_dtype == torch::kBFloat16) {                            \
+      tensorrt_llm::kernels::launchFusedQKNormRope<qkv_type, c10::BFloat16>( \
+          __VA_ARGS__);                                                      \
+    } else {                                                                 \
+      TORCH_CHECK(false, "Unsupported cos_sin_cache dtype: ", cache_dtype);  \
+    }                                                                        \
+  }()
```
ProExpertProg: Use existing float dispatching macro?

csrc/fused_qknorm_rope_kernel.cu
```diff
@@ -0,0 +1,426 @@
...
+  int64_t total_heads = num_heads_q + num_heads_k + num_heads_v;
+  TORCH_CHECK(
+      qkv.size(1) == total_heads * head_dim,
+      "QKV tensor size must match total number of heads and head dimension");
+
+  auto stream = at::cuda::getCurrentCUDAStream(qkv.get_device());
+
+  auto qkv_dtype = qkv.scalar_type();
+
+  if (qkv_dtype == torch::kHalf) {
```
ProExpertProg: Use dispatch macro?

ProExpertProg (approved):
Two nits in the kernel, otherwise LGTM!

csrc/fused_qknorm_rope_kernel.cu
izhuhaoran: Done.

ProExpertProg (approved):

JartX:
@ZJY0516 @izhuhaoran @tjtanaa yes this PR: https://github.com/vllm-project/vllm/pull/28500 solve the problem

ProExpertProg:
I think this is broken: #33295

PR #27170, opened 2025-10-19
[KVConnector][Feature] Support KV connector cache reset via /reset_prefix_cache

ptovam:
## Purpose
Extends the existing `/reset_prefix_cache` endpoint with an optional `reset_external` flag.
When set, both the local (GPU) and external (connector) prefix caches are reset.

This allows clearing connector-side caches dynamically without restarting the server -useful for benchmarking and testing.


vllm/entrypoints/openai/api_server.py
```diff
@@ -983,15 +983,34 @@ async def show_server_info(
     @router.post("/reset_prefix_cache")
     async def reset_prefix_cache(raw_request: Request):
         """
-        Reset the prefix cache. Note that we currently do not check if the
-        prefix cache is successfully reset in the API server.
+        Reset the local prefix cache.
+
+        Optionally, if the query parameter `reset_external=true`
+        (or the legacy alias `reset_external=true`) is provided,
```
markmc: Not sure what this "legacy alias" is?

markmc (approved):
lgtm

vllm/entrypoints/openai/api_server.py
ptovam: Good catch. my bad, removed.

NickLucche (approved):
lgtm

vllm/distributed/kv_transfer/kv_connector/v1/multi_connector.py
```diff
@@ -342,3 +342,6 @@ def get_kv_connector_stats(self) -> MultiKVConnectorStats | None:
                 stats_by_connector = MultiKVConnectorStats()
             stats_by_connector[c.__class__.__name__] = stats
         return stats_by_connector
+
+    def reset_cache(self) -> bool:
+        return any(connector.reset_cache() for connector in self._connectors)
```
NickLucche: actually, sorry, this should be an all() to make sure it really succeeded

vllm/distributed/kv_transfer/kv_connector/v1/base.py
```diff
@@ -431,3 +431,12 @@ def build_kv_connector_stats(
...
         return None
+
+    def reset_cache(self) -> bool:
+        """
+        Reset the connector's internal cache.
+
+        Returns:
+            bool: True if the cache was successfully reset, False otherwise.
+        """
+        return False
```
NickLucche: this could be optional to implement the suggestion below

NickLucche (changes requested):
Actually, sorry, one comment on the interface for the multiconnector case

vllm/distributed/kv_transfer/kv_connector/v1/base.py
ptovam: WDYM?

NickLucche: `Optional[bool] or bool | None` as return type, to allow for the `all(..)` check to work fine (you rule-out the connectors that do NOT implement it)

ptovam: Why treat connectors that haven’t implemented this as successful?
It could confuse users who aren’t aware of the implementation details.

NickLucche: connectors that have not implemented it will return None which is neither.
If you do any(..), one sub-connector being successful will result in multiconnector returning successful transfers.

What I would expect semantically is that multiconnector only returns true when *all* sub-connectors _that implement the interface_ return True (hence None return value is needed to discern those that have implemented and yet fail).

ptovam: Got it - I’ve pushed a commit with that change, thanks for clarifying.

ptovam:
@NickLucche Ready for another review — please let me know what you think.  
Thanks.

vllm/v1/core/sched/scheduler.py
```diff
@@ -1414,8 +1416,26 @@ def reset_prefix_cache(self, reset_running_requests: bool = False) -> bool:
...
+
+        if reset_connector:
+            reset_successful = self.reset_connector_cache() and reset_successful
+
         return reset_successful
 
+    def reset_connector_cache(self) -> bool:
+        if self.connector is None:
+            logger.warning("reset_connector called but no KV connector is configured.")
+            return False
```
NickLucche, line 1432: nit: I think this is a bug in logic if this happens right? We should probably `assert self.connector is not None`

vllm/v1/core/sched/scheduler.py
```diff
@@ -1414,8 +1416,26 @@ def reset_prefix_cache(self, reset_running_requests: bool = False) -> bool:
...
+
+        if self.connector.reset_cache() is False:
+            return False
```
NickLucche, line 1435: nit: we could probably add a debug log for calls that require resetting a cache on a connector that does not implement resetting (when None is returned), but this is not really a priority.

NickLucche (approved):
Apologies for the delay in getting back to you, this looks well done!

vllm/v1/core/sched/scheduler.py
ptovam: I added the log in the base function, so it also covers cases like the multi-connector setup.

vllm/v1/core/sched/scheduler.py
ptovam: Good point - I just feel an assert might be too strong here, since a user mistake shouldn’t crash the server. 
But open to your preference.

PR #27187, opened 2025-10-20
[ROCM] Enable CompressedTensorsWNA16

JartX:
I'm currently using ROCM with RDNA3. I've been trying to use compressed-tensors for a while, and I thought it was only supported on CUDA.

This change simply avoids entering: CompressedTensorsWNA16MarlinMoEMethod for CUDA when using ROCM and allows inference of a model like the following:

[jart25/Qwen3-VL-30B-A3B-Instruct-AWQ-8bit
](https://huggingface.co/jart25/Qwen3-VL-30B-A3B-Instruct-AWQ-8bit)
using: CompressedTensorsWNA16MoEMethod and ExllamaLinearKernel

Note that it must be run with:

export VLLM_USE_TRITON_AWQ=1


yewentao256 (approved):
LGTM, thanks for the work!

Issue #10583, opened 2024-11-22
[Bug]: Duplicate request_id breaks the engine


tjohnson31415:
### Your current environment

The environment is not relevant.

### 🐛 Describe the bug

The `request_id` field used in the LLM Engine is assumed to be unique, but it is a parameter that can be set by the caller. After https://github.com/vllm-project/vllm/pull/9550, the `request_id` is configurable even by a caller to the OpenAI Server.

I originally encountered this issue while investigating the cause of negative metrics in https://github.com/vllm-project/vllm/pull/10430. I found that sending two requests concurrently with the same request id can cause the negative metric error. Example bash using `curl`:

```
curl http://localhost:8000/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
        "model": "meta-llama/Llama-3.2-3B-Instruct",
        "request_id": "id",
        "messages": [{"role": "user", "content": "Tell me a long story:"}],
        "max_tokens": 1024,
        "min_tokens": 1024,
        "temperature": 0
      }' &

sleep 0.5

curl http://localhost:8000/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
        "model": "meta-llama/Llama-3.2-3B-Instruct",
        "request_id": "id",
        "messages": [{"role": "user", "content": "Tell me a long story:"}],
        "max_tokens": 1024,
        "min_tokens": 1024,
        "temperature": 0
      }' &
```
 
The fix in https://github.com/vllm-project/vllm/pull/10430 properly prevents the crash, but only one request gets a response and the other connection hangs open.

Other users have reported other bad behavior that looks to be caused by duplicate request_ids:
- Offline server hangs: https://github.com/vllm-project/vllm/issues/8393
- "RuntimeError: aclose(): asynchronous generator is already running": https://github.com/vllm-project/vllm/issues/6325
- `Counters can only be incremented by non-negative amounts.`: https://github.com/vllm-project/vllm/issues/6642

### Before submitting a new issue...

- [X] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

PR #15326, opened 2025-03-22
[Bugfix][Frontend] Fixed issue where requests with duplicate request IDs might be sent to EngineCore simultaneously

hidva:
Currently, vllm allows users to send duplicate request IDs. At the same time, numerous modules in EngineCore use request IDs as dictionary keys, such as `KVCacheManager.req_to_blocks`. This is based on the assumption that EngineCore always expects the Frontend to first abort a request before adding a new one with the same request ID:

```python
# req1, req2 have the same request_id.
(EngineCoreRequestType.ADD, req1(request_id=RequestId))
(EngineCoreRequestType.ABORT, req1)
(EngineCoreRequestType.ADD, req2(request_id=RequestId))
```

Currently, `AsyncLLM` ensures that duplicate request IDs must first be aborted before they can be added through the sequence `AsyncLLM._add_request` -> `OutputProcessor.add_request`:

```python
# OutputProcessor.add_request
request_id = request.request_id
if request_id in self.request_states:
    raise ValueError(f"Request id {request_id} already running.")

# AsyncLLM.abort
async def abort(self, request_id: str) -> None:
    """Abort RequestId in OutputProcessor and EngineCore."""

    request_ids = self.output_processor.abort_requests((request_id,))
    # BUG!
    # This operation is not atomic, and there might be a time window during which
    # the request has already been removed from OutputProcessor.request_states,
    # but the corresponding ABORT has not yet been issued to EngineCore.
    await self.engine_core.abort_requests_async(request_ids)

    if self.log_requests:
        logger.info("Aborted request %s.", request_id)
```

We can easily simulate the potential bug by enlarging the possible time window with an `await asyncio.sleep(13)` inserted at the BUG point:
image

To fix this issue, we categorized completed requests into two types: 

- abort req, `handle_abort_reqs`
- finished req, `_handle_finished_reqs`

And ensured that the scope of request visibility in the Frontend always includes the scope of request visibility in EngineCore.

PR #27189, opened 2025-10-20
Chat API: Force server-generated request_id to avoid collisions; improve clarity and safety

Zaragoto:
**Background**
- Problem: The Chat API could previously accept and reuse a client-provided request_id. Under duplicated IDs, this may lead to collisions, state leaks, mismatched responses, or confusing observability.
- Related issue: https://github.com/vllm-project/vllm/issues/10583

**What’s changed**
- Internal/server-generated ID
  - The P-side always generates a globally unique internal_request_id (e.g., UUID/ULID) for all internal processing (routing, queue keys, logs/traces).
- Response behavior
  - If the client supplies request_id, the server still uses the internal_request_id internally but echoes the client-supplied request_id in the API response for backward compatibility.
  - If the client omits request_id, the server returns the internal_request_id in the response.
- Cross-component propagation (P→D)
  - The P-side passes the internal_request_id to the D-side via kv_transfer_params.
  - The D-side uses this propagated internal_request_id for its processing and logging, ensuring the same internal ID is used consistently across both P and D.

**Behavioral changes**
- Before: Client request_id might be used internally and echoed back, risking collisions across components.
- Now:
  - Internally: Always use a server-generated internal_request_id. The same internal_request_id is propagated P→D via kv_transfer_params.
  - Externally (response):
    - If client supplied request_id, echo that value in the response.
    - Otherwise, return the server-generated internal_request_id.

---
<details>

</details>

### DIFF
--- a/vllm/entrypoints/openai/serving_chat.py
+++ b/vllm/entrypoints/openai/serving_chat.py
@@ -255,8 +255,8 @@ class OpenAIServingChat(OpenAIServing):
         except (ValueError, TypeError, RuntimeError, jinja2.TemplateError) as e:
             logger.exception("Error in preprocessing prompt inputs")
             return self.create_error_response(f"{e} {e.__cause__}")
-
-        request_id = self._add_prefix(request.request_id or random_uuid())
+        
+        request_id = self._add_prefix(getattr(request, "request_id", None))
 
         if request.kv_transfer_params:
             request_id = self._add_prefix(request.kv_transfer_params.get("p_side_request_id", request_id))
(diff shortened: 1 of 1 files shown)


markmc:
xref #15326

Zaragoto:
> > * For P/D we should assume an external proxy generates a request ID and supplies it to both the P and D instances. We don't need to support passing the P-generated ID to D
> 
> I forgot ... P does need to supply (via KV transfer params) its internal-request-ID to D ... at least in the case of NIXL, this forms part of the notification message sent from D to P (see `_read_blocks()`)

Yes. This is why we pass the internal req-id in P side to D side :)

markmc:
Thanks again for your contribution

See #29665 and #27987 for how this is progressing

PR #27190, opened 2025-10-20
[BUGFIX][ROCM] ViT FlashAttention on ROCm (no GFX9) and contiguous on qwen3vl ROCm TORCH_SDPA

JartX:
The refactor introduced in the following PR:
https://github.com/vllm-project/vllm/pull/26104
improved the flash-attn selection, but broke the loading of models like Qwen/Qwen3-VL-30B-A3B-Instruct in RocM RDNA3, as it doesn't support flash-attn for VL. In the PR, I use the same backend selection for ROCM, which is used in: https://github.com/vllm-project/vllm/blob/main/vllm/platforms/rocm.py

in the method:

get_vit_attn_backend

I'm quoting @tjtanaa to help me check that I haven't broken anything upstream for models other than RDNA3 :)

Thanks a lot!

UPDATE 25-10-25
After verifying that there is an inference bug in the qwen3vl, qwen2.5vl and qwen2vl models on rocm with the TORCH_SDPA attention backend, we proceed to expand the PR with the code provided by @tjtanaa in the PR: https://github.com/vllm-project/vllm/pull/27106


JartX:
> Thanks. Let me help verify :D

@tjtanaa Could you please check this too?
https://github.com/vllm-project/vllm/pull/27187

Many thanks!

vllm/attention/layer.py
```diff
@@ -93,14 +99,22 @@ def check_upstream_fa_availability(dtype: torch.dtype):
...
-    if (
-        attn_backend != _Backend.FLASH_ATTN
-        and attn_backend != _Backend.ROCM_AITER_FA
-        and check_upstream_fa_availability(torch.get_default_dtype())
-    ):
-        attn_backend = _Backend.FLASH_ATTN
-        use_upstream_fa = True
+) -> tuple[_Backend, Callable | None]:
+    if current_platform.is_rocm():
+        if envs.VLLM_ROCM_USE_AITER and envs.VLLM_ROCM_USE_AITER_MHA and on_gfx9():
```
tjtanaa, line 107: We shouldn't need this condition, whether to use the `_Backend.ROCM_AITER_FA`, it is determined in those `model.py` files.

vllm/attention/layer.py
```diff
@@ -93,14 +99,22 @@ def check_upstream_fa_availability(dtype: torch.dtype):
...
+            attn_backend = _Backend.ROCM_AITER_FA
+        elif on_gfx9():
```
tjtanaa: This condition was crafted for ROCm platform (`on_gfx9()`).

```python
        if (
            attn_backend != _Backend.FLASH_ATTN
            and attn_backend != _Backend.ROCM_AITER_FA
            and check_upstream_fa_availability(torch.get_default_dtype())
        ):
 ```
 
On `on_gfx9()`, we will always attempt to use `flash_attn`, but if it does not support, then we fallback to `_Backend.TORCH_SDPA`

File: vllm/attention/layer.py [hunk 3]
```diff
@@ -93,14 +99,22 @@ def check_upstream_fa_availability(dtype: torch.dtype):
...
+            attn_backend = _Backend.FLASH_ATTN
+        else:
+            return _Backend.TORCH_SDPA, None
+    else:
+        if (
```
tjtanaa: As mentioned above, this condition and its content is also applicable to ROCm on `gfx9`.

vllm/attention/layer.py
JartX: @tjtanaa The first step was simply to check if it was rocm and if it was on_gfx1x and if so, return _Backend.TORCH_SDPA, None directly. Would you like to leave it like this with a comment indicating why? I've only seen it affect RDNA3 with the upstream flow. I think this way we could mark the required changes as resolved. If not, please tell me how the code should be :)

JartX:
@tjtanaa added :)

DarkLight1337 (approved):
Thanks for fixing!

DarkLight1337:
Can you fix DCO?

JartX:
> Can you fix DCO?

do it, many thanks @DarkLight1337 :D

JartX:
@DarkLight1337 @tjtanaa What can we do with tests that fail?

DarkLight1337:
The test passes after retrying, so it should be good to merge

PR #27197, opened 2025-10-20
[Chore] Separate out NCCL utilities from vllm.utils

dongbo910220:
Part of #26900

This PR continues the effort to clean up `vllm.utils` by extracting NCCL-related helper functions into a new, dedicated module `vllm/utils/nccl.py`. 

The following functions have been moved:

* `vllm.utils.find_nccl_library -> vllm.utils.nccl.find_nccl_library`
* `vllm.utils.find_nccl_include_paths -> vllm.utils.nccl.find_nccl_include_paths`

CC: @DarkLight1337

### DIFF
--- a/vllm/utils/__init__.py
+++ b/vllm/utils/__init__.py
@@ -47,15 +47,6 @@ import yaml
 import vllm.envs as envs
 from vllm.logger import enable_trace_function_call, init_logger
 from vllm.ray.lazy_utils import is_in_ray_actor
-from vllm.utils.nccl import (
-    find_nccl_include_paths,
-    find_nccl_library,
-)
-
-_ = (
-    find_nccl_include_paths,
-    find_nccl_library,
-)
 
 _DEPRECATED_MAPPINGS = {
     "cprofile": "profiling",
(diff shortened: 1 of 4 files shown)


vllm/utils/__init__.py
```diff
@@ -48,6 +47,15 @@
 import vllm.envs as envs
 from vllm.logger import enable_trace_function_call, init_logger
 from vllm.ray.lazy_utils import is_in_ray_actor
+from vllm.utils.nccl import (
+    find_nccl_include_paths,
+    find_nccl_library,
+)
+
+_ = (
```
DarkLight1337: Can you update the imports in vLLM code to use the new imports? No need to keep back-compatibility. (Same goes for the other utils unless they are explicitly made to be user-facing)

dongbo910220: @DarkLight1337 Thank you for the feedback! Initially, I kept backward compatibility because Iwasn't entirely sure which utilities were internal-only versus user-facing, so I took a conservative approach to avoid potential breaking changes.
I've updated all three PRs to remove backward compatibility

DarkLight1337 (approved):
Gemini didn't flag any missing imports so LGTM, see if the tests pass now

vllm/utils/__init__.py
DarkLight1337: Thanks!

Issue #26928, opened 2025-10-15
[Bug]: Apply security fixes to v0.11.1


russellb:
Placeholder issue as a checklist item to apply pending security fixes prior to releasing v0.11.1.

Note that these links will not be accessible until after the release and corresponding security advisories are published.

- [x] https://github.com/vllm-project/vllm/security/advisories/GHSA-mrw7-hf4f-83pf - https://github.com/vllm-project/vllm/pull/27204
- [x] https://github.com/vllm-project/vllm/security/advisories/GHSA-pmqf-x6x8-p7qw - https://github.com/vllm-project/vllm/pull/27204
- [x] https://github.com/vllm-project/vllm/security/advisories/GHSA-69j4-grxj-j64p - https://github.com/vllm-project/vllm/pull/27205

PR #27204, opened 2025-10-20
[Frontend] Require flag for loading text and image embeds

russellb:
https://github.com/vllm-project/vllm/security/advisories/GHSA-mrw7-hf4f-83pf

https://github.com/vllm-project/vllm/security/advisories/GHSA-pmqf-x6x8-p7qw

Issue #26928


DarkLight1337 (approved):

Isotr0py (approved):

DarkLight1337:
cc @christian-pinto `--enable-mm-embeds` flag will be required for using terratorch models going forward.

vllm/config/model.py
```diff
@@ -232,8 +232,10 @@ class ModelConfig:
     output will contain token ids."""
     enable_prompt_embeds: bool = False
     """If `True`, enables passing text embeddings as inputs via the
-    `prompt_embeds` key. Note that enabling this will double the time required
-    for graph compilation."""
+    `prompt_embeds` key.
+
+    WARNING: The vLLM engine may crash if incorrect shape of embeddings is passed.
+    Only enable this flag for trusted users!"""
```
qthequartermasterman, line 238: Should a shape check be added to the renderer?

DarkLight1337: The shape depends on the embedding size of each model so it requires a lot more effort to perform shape validation outside of the model class.

PR #26847, opened 2025-10-14
[Frontend][torch.compile] CompilationConfig Overhaul (#20283): Set up -O infrastructure

morrison-turnansky:
# CompilationConfig Overhaul and Optimization Levels

## Overview

Description of changes to `VLLMConfg` and `CompilationConfig`. The introduction of meaningful optimization levels (`-O0`, `-O1`, `-O2`, `-O3`) from [GitHub Issue #20283](https://github.com/vllm-project/vllm/issues/20283). This change aims to improve user experience by providing intuitive optimization levels that trade startup time for performance, while consolidating and simplifying the compilation configuration system. We also have changed defaults to help users get desired out of the box performance. These defaults are determined by `optimization-level`. Importantly defaults are purely defaults; explicit user settings will not be overwritten.

### 1. Repurposing `-O` for Optimization Levels

The `-O<n>` flags now represent meaningful optimization levels that trade startup time for performance:

#### `-O0`: No Optimization
- **Startup**: Fastest startup time
- **Performance**: No compilation, no cudagraphs, no optimizations
- **Equivalent to**: `--enforce-eager` (deprecated)
- **Use case**: Development, debugging, or when startup time is critical

```bash
# CLI usage
python -m vllm.entrypoints.api_server --model microsoft/DialoGPT-medium -O0

# Python API usage
from vllm.entrypoints.llm import LLM
from vllm.config.vllm import OptimizationLevel

llm = LLM(
    model="microsoft/DialoGPT-medium",
    optimization_level=OptimizationLevel.O0
)
```

#### `-O1`: Quick Optimizations
- **Startup**: Moderate startup time
- **Performance**: Inductor compilation, CUDAGraphMode.PIECEWISE
- **Use case**:  Balance for most development scenarios

```bash
# CLI usage
python -m vllm.entrypoints.api_server --model microsoft/DialoGPT-medium -O1

# Python API usage
from vllm.entrypoints.llm import LLM
from vllm.config.vllm import OptimizationLevel

llm = LLM(
    model="microsoft/DialoGPT-medium",
    optimization_level=OptimizationLevel.O1
)
```

#### `-O2`: Full Optimizations (Default)
- **Startup**: Longer startup time
- **Performance**: `-O1` + CUDAGraphMode.FULL_AND_PIECEWISE
- **Use case**: Production workloads where performance is important. This is the default use case. It is also very similar to the previous default. The primary difference is that  noop & fusion flags are enabled. 

```bash
# CLI usage (default, so optional)
[...]

Issue #22201, opened 2025-08-04
[RFC]: Integrate MPK compiler as an experimental execution backend to vLLM


ProExpertProg:
### Motivation.

LLM inference, especially for smaller and middle-size models, still fails to attain the theoretical maximums in TFLOPs and memory bandwidth. There are many reasons for this, from CPU overheads and kernel launch overheads to other pipeline bubbles like tail latencies in GPU kernel execution. While CUDA Graphs help to some extent, one promising approach is compiling megakernels that use precision scheduling. A megakernel is a single GPU kernel where SMs read and execute “tasks” corresponding to different compute or memory operations. This can eliminate a lot of the aforementioned latencies due to better overlapping of operations, reduced memory load, and higher hardware utilization.

MPK/Mirage is a megakernel compiler that takes in model definitions in its custom DSL and produces a single megakernel for the full model forward pass. This RFC proposes adding Mirage as an experimental torch.compile-based backend to vLLM. According to [the Mirage blog post](https://zhihaojia.medium.com/compiling-llms-into-a-megakernel-a-path-to-low-latency-inference-cf7840913c17), Mirage demonstrates a 20-25% improvement in forward pass latency over vLLM. I would kindly like to ask the Mirage team to add more thorough benchmarking results in the comments of this RFC to further motivate this integration.

### Proposed Change.

I propose adding a new Dynamo-based backend to vLLM that compiles the `fx.Graph` into an MPK megakernel. It would consist of a compatibility layer, which would convert the vLLM-produced graph into a Mirage-friendly graph, and a Mirage-based backend that would compile the second graph into a megakernel. The compatibility layer would live in vLLM and the backend would live in the `mirage` pip package. In the future, this compatibility layer can absorb any changes to Mirage or vLLM without requiring changes to the other.

Mirage prefers to receive information about higher level ops (RMSNorm, RoPE, quant) as opposed to low level torch operations. For good separation of concerns, the compatibility layer will convert vLLM custom ops into Mirage high-level ops (e.g. `torch.ops._C.rms_norm` -> `torch.ops.mirage.rms_norm`). If there are other differences in the Mirage / vLLM representation, they can all be handled in the compatibility layer, requiring no changes to either Mirage or (the rest of) vLLM when the other is changed.

[...]

PR #27218, opened 2025-10-20
[Backend][WIP] Integrate MPK (Mirage) compiler as an experimental execution backend to vLLM

NorthmanPKU:
## Purpose
See #22201 and https://github.com/mirage-project/mirage/issues/522.

## TODO:
- [ ] Make sure we are compatible with #26847
- [ ] Make sure the timing of kv_cache allocation and compilation is as we expect
- [ ] Benchmark vllm with MPK backend

---
<details>

</details>

### DIFF
--- a/vllm/compilation/mirage_backend.py
+++ b/vllm/compilation/mirage_backend.py
@@ -1,13 +1,20 @@
+import os
 from collections import defaultdict
-from .backends import *
+import time
 from mirage import MPK, MPKMetadata, MirageModelConfig
 import re
-from vllm.config import ModelConfig, get_current_vllm_config
-from vllm.config.parallel import ParallelConfig
-from vllm.forward_context import ForwardContext, get_forward_context
-from vllm.model_executor.models.utils import extract_layer_index
+from typing import Any
+
 import torch
+import torch.fx as fx
+
+from vllm.config import CompilationConfig, ModelConfig, VllmConfig, get_current_vllm_config
+from vllm.config.parallel import ParallelConfig
+from vllm.forward_context import get_forward_context
+from vllm.model_executor.models.utils import extract_layer_index
 from vllm.logger import init_logger
+
+from .counter import compilation_counter
 
 logger = init_logger(__name__)
 
@@ -16,7 +23,8 @@ def transfer_tensor_names(placeholders: list[torch.fx.node.Node]) -> list[str]:
     """Transfer FX placeholder debug names to model-like dotted names. Return a list of transferred names and input id.
 
     Example:
-        l_self_modules_layers_modules_17_modules_mlp_modules_gate_up_proj_parameters_weight_
+        l_self_modules_layers_modules_17_modules_mlp_\
+            modules_gate_up_proj_parameters_weight_
         -> model.layers.17.mlp.gate_up_proj.weight
 
     Notes:
@@ -26,14 +34,12 @@ def transfer_tensor_names(placeholders: list[torch.fx.node.Node]) -> list[str]:
     """
     converted_names = []
     s_pattern = re.compile(r"^s\d+$") # s72 / s80
-    input_id = 0
 
-    for i, node in enumerate(placeholders):
+    for node in placeholders:
         name = node.name
         if name == 'l_input_ids_':
             final_name = 'input_ids'
             converted_names.append(final_name)
-            input_id = i
         elif name == 'l_positions_':
             final_name = 'positions'
             converted_names.append(final_name)
@@ -52,7 +58,7 @@ def transfer_tensor_names(placeholders: list[torch.fx.node.Node]) -> list[str]:
 
             converted_names.append(final_name)
 
-    return converted_names, input_id
+    return converted_names
 
 def build_model_config(
     model_config: ModelConfig,
@@ -111,14 +117,12 @@ def build_mpk_metadata(
     for layer_name in static_forward_context.keys():
         index2name[extract_layer_index(layer_name, 1)].append(layer_name)
 
-    for layer_index in sorted(index2name.keys()):
+    for layer_index in sorted(index2name):
         layer_names = index2name[layer_index]
         assert len(layer_names) == 1, "Multiple layers with the same layer index are not supported"
         layer_name = layer_names[0]
-        # logger.info(f"{layer_index} {layer_name}: attention num: {len(static_forward_context[layer_name].kv_cache)}; kv_cache.shape: {static_forward_context[layer_name].kv_cache[0].shape}")
         k_cache_tensors.append(static_forward_context[layer_name].kv_cache[0][0])
         v_cache_tensors.append(static_forward_context[layer_name].kv_cache[0][1])
-        # kv_cache_tensors shape: num_layers * (2, num_blocks, block_size, num_kv_heads, head_size)
     
     state_dict = {}
     input_token_ids = None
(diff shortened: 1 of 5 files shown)


vllm/config/compilation.py
```diff
@@ -663,14 +659,21 @@ def init_backend(self, vllm_config: "VllmConfig") -> str | Callable:
...
         assert self.mode == CompilationMode.VLLM_COMPILE
-        if self.backend not in ["eager", "inductor"]:
-            raise ValueError(
-                f"Invalid backend for piecewise compilation: {self.backend}"
-            )
-
-        from vllm.compilation.backends import VllmBackend
-
-        return VllmBackend(vllm_config)
+        # Default piecewise compilation path (vLLM backend driving Inductor/Eager)
```
ProExpertProg: Instead of this approach, Mirage should use `CompilationMode.DYNAMO_TRACE_ONCE` and then be directly passed to `torch.compile(backend=...)`, like it is now. `VLLM_COMPILE` implies `VllmBackend`. Later, if we want to add support for caching and any other `VllmBackend` optimizations, we can add this code path and initialize VllmBackend which will use a nested MirageBackend for the actual compilation like it currently uses `InductorAdaptor` for `backend="inductor"`, but that will come much later.

vllm/compilation/mirage_backend.py
```diff
@@ -0,0 +1,263 @@
...
+                )
+                logger.info(f"[Mirage] MPK metadata: {mpk_metadata.info_as_string()}")
+                self.mpk = MPK(mpk_metadata)
+                self.mpk.build()
+                self.mpk.compile()
+                
+                self.compiled = True
+                
+            logger.info(f"[Mirage] Calling the compiled result...")
+            return self.mpk()
```
ProExpertProg: Shouldn't we try to pass at least the input tensors to mpk here? Or do those have to be static too? If yes, I'm not sure that's guaranteed by the `GPUModelRunner`

vllm/config/compilation.py
```diff
@@ -663,14 +659,21 @@ def init_backend(self, vllm_config: "VllmConfig") -> str | Callable:
             return resolve_obj_by_qualname(self.backend)
```
ProExpertProg, line 817:
```suggestion
            if self.backend == "mirage":
                from vllm.compilation.mirage_backend import MirageBackend
                return MirageBackend(vllm_config)
            return resolve_obj_by_qualname(self.backend)
```

vllm/v1/attention/backends/mirage.py
```diff
@@ -0,0 +1,349 @@
...
+        Returns:
+            shape = [num_tokens, num_heads * head_size]
+        """
+        assert output is not None, "Output tensor must be provided."
+
+        if attn_metadata is None:
+            # Profiling run.
+            return output.fill_(0)
+
+        return output
```
ProExpertProg: We should probably just `raise NotImplementedError` here as this backend is not ever meant to be used directly. If it is, we should return a proper calculated value

vllm/config/compilation.py
```diff
@@ -675,6 +675,9 @@ def init_backend(self, vllm_config: "VllmConfig") -> str | Callable:
         ]:
             if self.backend in torch_backends:
                 return self.backend
+            if self.backend == "mirage":
```
ProExpertProg: Let's call this `mirage_hardcoded`/`mirage_direct`/`mirage_byname`, and call the proper torch.compile Mirage backend `mirage`.

vllm/compilation/mirage_backend.py
```diff
@@ -0,0 +1,342 @@
...
+                
+            logger.info(f"[Mirage] Calling the compiled result...")
+            result_hidden_states = self.mpk()
+            
+            return (result_hidden_states,)
+        
+        # return VllmSerializableFunction(
+        #     graph, example_inputs, self.prefix, compile_or_call
+        # )
+        return compile_or_call
```
ProExpertProg: Please remove remnants of `VllmBackend`

vllm/v1/attention/backends/mirage.py
```diff
@@ -0,0 +1,349 @@
...
+            raise ValueError(
+                f"Head size {head_size} is not supported by {attn_type}. "
+                f"Supported head sizes are: {supported_head_sizes}. "
+                "Set VLLM_ATTENTION_BACKEND=FLEX_ATTENTION to use "
+                "FlexAttention backend which supports all head sizes."
+            )
+
+    @staticmethod
+    def get_name() -> str:
+        return "MPK_ATTENTION"
```
ProExpertProg: This should be "MIRAGE" - please stay consistent

vllm/compilation/mirage_backend.py
```diff
@@ -0,0 +1,263 @@
...
+    return mpk_metadata
+
+class MirageBackend:
+    """The compilation backend for `torch.compile` with vLLM.
+    It is used for compilation level of `CompilationLevel.PIECEWISE`,
+    where we customize the compilation.
+
+    The major work of this backend is to split the graph into
+    piecewise graphs, and pass them to the piecewise backend.
+
```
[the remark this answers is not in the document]
ProExpertProg: ++

vllm/compilation/mirage_backend.py
```diff
@@ -0,0 +1,342 @@
...
+        vllm_config: VllmConfig,
+        prefix: str = "",
+    ):
+        # if the model is initialized with a non-empty prefix,
+        # then usually it's enough to use that prefix,
+        # e.g. language_model, vision_model, etc.
+        # when multiple parts are initialized as independent
+        # models, we need to use the model_tag to distinguish
+        # them, e.g. backbone (default), eagle_head, etc.
+        logger.info("[Mirage] Calling MirageBackend init!")
```
ProExpertProg: Debug

vllm/compilation/mirage_backend.py
```diff
@@ -0,0 +1,342 @@
...
+        # we control the compilation process, each instance can only be
+        # called once
+        assert not self._called, "MirageBackend can only be called once"
+        
+        placeholders = [node for node in graph.graph.nodes if node.op == 'placeholder']
+        assert len(placeholders) == len(example_inputs)
+        
+        transfered_tensor_names, input_id = transfer_tensor_names(placeholders)
+
+        max_input_tokens = example_inputs[input_id].shape[0]
```
ProExpertProg: Read from config instead (this is not a guarantee)

vllm/compilation/mirage_backend.py
```diff
@@ -0,0 +1,342 @@
...
+        self._called = True
+        self.compiled = False
+        
+        def compile_or_call(*args):
+            dumb_run_called = (get_forward_context().attn_metadata is None)
+            if dumb_run_called:
+                model_config = self.vllm_config.model_config
+                dtype = model_config.dtype
+                hidden_size = model_config.get_hidden_size()
+                # TODO(Jianan Ji): We'll want to run in eager instead of doing nothing
```
ProExpertProg: Are you talking about the `profile_run` here? Why not do compilation during the profile run like Inductor does? Do you need access to the KV cache tensors to compile? 

You should be able to just call graph here:
```suggestion
                return graph(*args)
```

ProExpertProg (commented):
Because you want to compile after kv cache is initialized, you should add another dummy run to the gpu model runner to make sure MPK compilation happens during init.

Also, we should add 3 things to `VllmConfig.__post_init__`:
- disable cudagraphs if backend is Mirage (cudagraph_mode = NONE)
- automatically set attention backend to mirage if the compile backend is mirage
- raise a ValueError if attention backend is mirage but compile backend isn't


Also, it would be great to see some benchmarking numbers! LEt me know if you need help running them!

PR #27221, opened 2025-10-20
ARM64 CUDA 12.9 wheels built and uploaded to index incorrectly

Gregory-Pereira:
# Fix: arm wheels tagged incorrectly

Related to: https://vllm-dev.slack.com/archives/C07R5PAL2L9/p1760721161081999

## Issue

The vLLM wheels index contains both `arm64` and `amd64` wheels for CUDA, however the wheel for arm does not contain the `cu<CUDA_VERSION>` identifier, while the `amd64` one is.

## Cause

I believe this is happening because in the `arm` wheel pipeline we pass `VLLM_MAIN_CUDA_VERSION` as `12.9` which it understands to be the same Major and Minor version as `CUDA_VERSION` of `12.9.1` for this pipeline - as a result it omits the `cu129` identifier from the build. This should fix our release pipeline to ensure wheels match the index generated for them and are consistent with the `x86_64` ones.

cc @wseaton @simon-mo @khluu

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -116,9 +116,6 @@ ENV UV_LINK_MODE=copy
 # as it was causing spam when compiling the CUTLASS kernels
 RUN apt-get install -y gcc-10 g++-10
 RUN update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-10 110 --slave /usr/bin/g++ g++ /usr/bin/g++-10
-RUN <<EOF
-gcc --version
-EOF
 
 # Workaround for https://github.com/openai/triton/issues/2507 and
 # https://github.com/pytorch/pytorch/issues/107960 -- hopefully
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -116,9 +116,6 @@ ENV UV_LINK_MODE=copy
 # as it was causing spam when compiling the CUTLASS kernels
 RUN apt-get install -y gcc-10 g++-10
 RUN update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-10 110 --slave /usr/bin/g++ g++ /usr/bin/g++-10
-RUN <<EOF
-gcc --version
-EOF
```
youkaichao, line 121: why do we need this?

Gregory-Pereira: We did not, nor is it related to this feature - this was leftover debug based on the glibc version missmatch between build and runtime container image IIRC, cc @tlrmchlsmth

Harry-Chen:
I think this can be closed after the latest nightly wheel updates.

PR #26135, opened 2025-10-02
[ModelOpt] Load w13/w2_input_scale for all experts, nvfp4

wenscarl:
This matches trt-llm usage. Use all experts' input scales to compute alpha and quantize. This change doesn't affect perf.

## Test Plan
```
VLLM_WORKER_MULTIPROC_METHOD="spawn" \
VLLM_USE_STANDALONE_COMPILE=0 \
VLLM_USE_FLASHINFER_MOE_FP4=1 \
VLLM_FLASHINFER_MOE_BACKEND="throughput" \
lm_eval --model vllm --model_args pretrained=nvidia/DeepSeek-R1-FP4,quantization=modelopt_fp4,data_parallel_size=8,enable_expert_parallel=True,tensor_parallel_size=1,max_model_len=2048,enforce_eager=True --trust_remote_code --tasks gsm8k --num_fewshot 5 --batch_size auto
```
## Test Result
After change:
```
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9431|±  |0.0064|
|     |       |strict-match    |     5|exact_match|↑  |0.9401|±  |0.0065|```

Before change:
```
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value|   |Stderr|
|-----|------:|----------------|-----:|-----------|---|----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.934|±  |0.0068|
|     |       |strict-match    |     5|exact_match|↑  |0.931|±  |0.0070|
```
---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #27223, opened 2025-10-20
Flashinfer_CUTLASS_MOE fuses quantization for TP

wenscarl:
co-authored by @leejnau 

cc. @bnellnm 
## Purpose
For TP case, the nvfp4 quantzation is fused with flashinfer_cutlass_moe call. This fixes the accuracy issue for nvidia/Deepseek-R1-0528-FP4-v2 model. 
For DP case, the fix should rely on https://github.com/vllm-project/vllm/pull/26135.

## Test Plan
```
VLLM_USE_FLASHINFER_MOE_FP4=1 \
VLLM_FLASHINFER_MOE_BACKEND="throughput" \
/home/shuw/.local/bin/lm_eval --model vllm --model_args pretrained=nvidia/DeepSeek-R1-0528-FP4-v2,data_parallel_size=1,enable_expert_parallel=True,tensor_parallel_size=4,enforce_eager=True,max_model_len=2048 --trust_remote_code --tasks gsm8k --num_fewshot 5 --batch_size auto

```

## Test Result
Previously:
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.0106|±  |0.0028|

With this PR:
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9522|±  |0.0059|
|     |       |strict-match    |     5|exact_match|↑  |0.9469|±  |0.0062|
---
<details>

</details>


bnellnm (approved):
LGTM.  Alternatively, you could prevent modular kernels from being created in this case and fall though to the direct call to `flashinfer_cutlass_moe_fp4`?

vllm/model_executor/layers/quantization/modelopt.py
```diff
@@ -1723,29 +1723,6 @@ def apply(
...
-            self.allow_flashinfer
-            and self.flashinfer_moe_backend == FlashinferMoeBackend.CUTLASS
-        ):
-            from vllm.model_executor.layers.fused_moe.flashinfer_cutlass_moe import (  # noqa: E501
-                flashinfer_cutlass_moe_fp4,
-            )
-
-            assert self.moe_quant_config is not None
-
-            return flashinfer_cutlass_moe_fp4(
```
mgoin, line 1782: Can you update for compressed-tensors too? https://github.com/vllm-project/vllm/blob/f9e7ad54000fb982248564591007e038304f190e/vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors_moe.py#L516

bnellnm:
> > LGTM. Alternatively, you could prevent modular kernels from being created in this case and fall though to the direct call to `flashinfer_cutlass_moe_fp4`?
> 
> Yes. But do you prefer to have modular kernels anyways?

I've no preference for this particular case since there's no communication going on.  My only concern would be that the `FusedMoEQuantConfig` doesn't match up with what the prepare/finalize class is doing.  This could probably be made consistent by making sure the `quant_dtype` is None, e.g. something like this

```
    def get_fused_moe_quant_config(
        self, layer: torch.nn.Module
    ) -> FusedMoEQuantConfig | None:
        if (
            self.use_marlin
            or self.flashinfer_moe_backend == FlashinferMoeBackend.TENSORRT_LLM
        ):
            return None
        
        # New case here for CUTLASS
        if (
              self.flashinfer_moe_backend == FlashinferMoeBackend.CUTLASS
              and layer.tp_size > 1
        ):  
            return FusedMoEQuantConfig.make(
               w2_scale=layer.w2_weight_scale,
               g1_alphas=layer.g1_alphas,
               g2_alphas=layer.g2_alphas,
               a1_gscale=layer.w13_input_scale_quant,
               a2_gscale=layer.w2_input_scale_quant,
           )
        
        return nvfp4_moe_quant_config(
            w2_scale=layer.w2_weight_scale,
            g1_alphas=layer.g1_alphas,
            g2_alphas=layer.g2_alphas,
            a1_gscale=layer.w13_input_scale_quant,
            a2_gscale=layer.w2_input_scale_quant,
        )
```

vllm/model_executor/layers/quantization/modelopt.py
```diff
@@ -1723,29 +1723,6 @@ def apply(
                 expert_map=expert_map,
                 apply_router_weight_on_input=apply_router_weight_on_input,
             )
-        elif (
```
leejnau, line 1772: Dead code path is removed here.

vllm/model_executor/layers/quantization/modelopt.py
wangshangsam: @wenscarl @mgoin is my understanding correct that, going forward, direct calling into FlashInfer is going to be deprecated in favour of modular kernels, for all cases (e.g., regardless of EP/TP/DP choices)?

bnellnm: > @wenscarl @mgoin is my understanding correct that, going forward, direct calling into FlashInfer is going to be deprecated in favour of modular kernels, for all cases (e.g., regardless of EP/TP/DP choices)?

There's no plan to force all MoE kernels to be called via modular kernels.

wangshangsam: But by deleting this `elif` clause (and, per @mgoin 's suggestion, applying this change to compressed-tensors), doesn't it force the use of FlashInfer cutlass implementation to go through the modular kernels?

I'm just trying to understand if this is the plan for all cases that use FlashInfer, regardless of distributed strategies, or whether `self.flashinfer_moe_backend` is `FlashinferMoeBackend.TENSORRT_LLM` or `FlashinferMoeBackend.CUTLASS`.

bnellnm: I don't see any reason to use the modular kernels for cases that aren't using some kind of all2all communication.  In this particular case I think @wenscarl and @leejnau figured out that this was dead code because the CUTLASS case always created a modular kernel.  I'm not sure if the same holds true for compressed_tensors.

leejnau (approved):

vllm/model_executor/layers/quantization/modelopt.py
wenscarl: @bnellnm this PR is updated with a additional `quant_dtype`: nvfp4_skip_quantization.

wenscarl: >I'm just trying to understand if this is the plan for all cases that use FlashInfer
I vote for that. Since flashinfer cutlass moe is at least a better option to normal cutlass_moe. TRTLLM MoE can win sometimes even.

vllm/model_executor/layers/quantization/modelopt.py
```diff
@@ -1580,6 +1580,20 @@ def get_fused_moe_quant_config(
         ):
             return None
 
+        if (
+            self.flashinfer_moe_backend == FlashinferMoeBackend.CUTLASS
+            and not layer.moe_config.dp_size > 1
+        ):
+            return FusedMoEQuantConfig.make(
+                quant_dtype="nvfp4_skip_quantization",
```
bnellnm, line 1239: Why not just leave quant_dtype as `None`?

wenscarl: If None, it gets conflicted with bf16 branch [here](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/fused_moe/flashinfer_cutlass_moe.py#L176-L180). For nvfp4 TP case, we still want quant_scales, just the hidden_states being bf16.

bnellnm: > here

It seems like it would hit that `else` even if the quant_dtype is "nvfp4_skip_quantization"?  Could you change it to check for non-empty scales instead of just falling thru to the bf16 case?

I think `quant_dtype` should be limited to represent actual types and not types + some flag.  Whenever torch finally supports fp4 types it would be good to remove the `str` option from `quant_dtype` entirely.  Or switch `quant_dtype` to use `vllm.scalar_type | None`.  (Obviously not for this PR).

wenscarl: >It seems like it would hit that else even if the quant_dtype is "nvfp4_skip_quantization"?
No. it would hit nvfp4 branch and assemble quant_scale as usual. The problem is that if quant_dtype == None, there is no difference b/w bf16. I agree with you that "nvfp4_skip_quantization" is not a good idea. I will revert to previous commit.

vllm/model_executor/layers/fused_moe/flashinfer_cutlass_moe.py
```diff
@@ -214,6 +217,7 @@ def flashinfer_cutlass_moe_fp4(
         FlashInferExperts(
             out_dtype=hidden_states.dtype,
             quant_config=quant_config,
+            use_dp=False,
```
pavanimajety, line 220: Why are we always setting `use_dp=False`?  Doesn't `flashinfer_cutlass_moe_fp4` also support dp?

wenscarl: flashinfer_cutlass_moe_fp4(dead code) is removed in this PR. Previously flashinfer_cutlass_moe_fp4 is meant for TP case only. If DP, the fused expert is assemble elsewhere.

pavanimajety: How is TP + Flashinfer cutlass handled now? Could we remove this method altogether or do we see any cases where it would be used?

wenscarl: https://github.com/vllm-project/vllm/pull/27223/files#diff-5bb9585da825481e1ae1534657a703f846325c7dd72cfddb0c41f878db33d78aR82 differentiate TP vs DP. But in either case, a fused moe expert is assembled. I agree that `flashinfer_cutlass_moe_fp4` should be removed. But compressed tensor TP could still depends on it. So its removal is out-of-scope for this PR.

pavanimajety (approved):

vllm/model_executor/layers/quantization/modelopt.py
```diff
@@ -1769,29 +1769,6 @@ def apply(
                 expert_map=expert_map,
                 apply_router_weight_on_input=apply_router_weight_on_input,
             )
-        elif (
-            self.allow_flashinfer
-            and self.flashinfer_moe_backend == FlashinferMoeBackend.CUTLASS
-        ):
```
weireweire, line 1775: will this break PP mode? when run deepseek PP, it will go to the last clause that use cutlass_moe_fp4, which will break on SM120. While on SGLang, flashinfer_cutlass works with sm120 PP.

weireweire: well, https://github.com/vllm-project/vllm/pull/27123 added back this and solved my issue.

PR #27230, opened 2025-10-20
[Feat] Adds runai distributed streamer

bbartels:
## Purpose
This adds the runai distributed streamer integration. This significantly improves model load times for object storage and high throughput fileshares.

## Test Result

---
<details>

</details>


vllm/model_executor/model_loader/runai_streamer_loader.py
```diff
@@ -16,6 +16,7 @@
     download_weights_from_hf,
     runai_safetensors_weights_iterator,
 )
+from vllm.platforms import current_platform
```
bbartels:
```suggestion
```

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -654,13 +654,20 @@ def _load_file(st_file: str):
             yield from state_dict.items()
 
 
+
```
bbartels:
```suggestion
```

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -660,7 +660,13 @@ def runai_safetensors_weights_iterator(
 ) -> Generator[tuple[str, torch.Tensor], None, None]:
     """Iterate over the weights in the model safetensor files."""
     with SafetensorsStreamer() as streamer:
-        streamer.stream_files(hf_weights_files)
+        device = _infer_streamer_device()
+
```
bbartels:
```suggestion
```

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -657,17 +657,31 @@ def _load_file(st_file: str):
...
     """Iterate over the weights in the model safetensor files."""
     with SafetensorsStreamer() as streamer:
-        streamer.stream_files(hf_weights_files)
+
+        def get_device():
+            is_cuda = current_platform.is_cuda()
+            return f"cuda:{torch.cuda.current_device()}" if is_cuda else None
+
+        if is_distributed and (device := get_device()):
+            streamer.stream_files(hf_weights_files, device=device, is_distributed=True)
```
noa-neria: A parameter is missing here:
Should be `stream_files(hf_weights_files, **s3_credentials=None**, device=device, is_distributed=True)
`

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -657,17 +657,31 @@ def _load_file(st_file: str):
...
+            tensor_source = (
+                (name, tensor.clone().detach())
```
noa-neria: No need to clone and detach the tensor, since the tensor is used only in the scope of the iterator.
Should be `tensor_source = streamer.get_tensors()` as before

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -657,17 +657,31 @@ def _load_file(st_file: str):
...
+            streamer.stream_files(hf_weights_files)
+            tensor_source = streamer.get_tensors()
         total_tensors = sum(
             len(tensors_meta)
             for tensors_meta in streamer.files_to_tensors_metadata.values()
         )
 
         tensor_iter = tqdm(
-            streamer.get_tensors(),
+            tensor_source,
```
noa-neria: Not necessary, can be reverted

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -657,17 +657,31 @@ def _load_file(st_file: str):
...
+
+        if is_distributed and (device := get_device()):
+            streamer.stream_files(hf_weights_files, device=device, is_distributed=True)
+            tensor_source = (
+                (name, tensor.clone().detach())
+                for name, tensor in streamer.get_tensors()
+            )
+        else:
+            streamer.stream_files(hf_weights_files)
+            tensor_source = streamer.get_tensors()
```
noa-neria: You can call `streamer.get_tensors()` once outside the if statement

noa-neria (commented):
Thank you for this work @bbartels !

vllm/model_executor/model_loader/runai_streamer_loader.py
```diff
@@ -30,6 +30,12 @@ def __init__(self, load_config: LoadConfig):
         if load_config.model_loader_extra_config:
             extra_config = load_config.model_loader_extra_config
 
+            self._is_distributed = False
```
noa-neria: The default value of `self._is_distributed` can be `True`, according to the default behavior of the streamer.

By default, if `is_distribued` is `True`, distributed streaming is enabled only for object storage and cuda devices.
The default behavior can be changed using `RUNAI_STREAMER_DIST=0` to disable distributing or  `RUNAI_STREAMER_DIST=1` for forcing distributed streaming in other cases.

bbartels: I thought it might be sensible to have it off at first to ensure not breaking customers just by upgrading versions. Once we have reports that it works well we can still flip this to `true` in a consecutive version

vllm/model_executor/model_loader/weight_utils.py
bbartels: The ENV vars to configure credentials take precedence I presume, even if `s3_credentials` is `None` here?

noa-neria: Correct. Also, the default value of `s3_credentials` is `None`

vllm/model_executor/model_loader/weight_utils.py
bbartels:
```suggestion
            return f"cuda:{torch.cuda.current_device()}" if is_cuda else None

        if is_distributed and (device := get_device()):
            streamer.stream_files(hf_weights_files, device=device, is_distributed=True)
        else:
            streamer.stream_files(hf_weights_files)
```

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -657,17 +657,26 @@ def _load_file(st_file: str):
...
+        else:
+            streamer.stream_files(hf_weights_files)
         total_tensors = sum(
             len(tensors_meta)
             for tensors_meta in streamer.files_to_tensors_metadata.values()
         )
 
         tensor_iter = tqdm(
-            streamer.get_tensors(),
+            tensor_source,
```
bbartels:
```suggestion
            streamer.get_tensors(),
```

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -657,10 +657,19 @@ def _load_file(st_file: str):
...
     """Iterate over the weights in the model safetensor files."""
     with SafetensorsStreamer() as streamer:
-        streamer.stream_files(hf_weights_files)
+
+        def get_device():
+            is_cuda = current_platform.is_cuda()
+            return f"cuda:{torch.cuda.current_device()}" if is_cuda else None
+
+        if is_distributed and (device := get_device()):
+            streamer.stream_files(hf_weights_files, device=device, is_distributed=True)
```
bbartels:
```suggestion
            streamer.stream_files(hf_weights_files, s3_credentials=None, device=device, is_distributed=True)
```

vllm/model_executor/model_loader/weight_utils.py
bbartels: Fixed

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -657,10 +657,19 @@ def _load_file(st_file: str):
...
     """Iterate over the weights in the model safetensor files."""
     with SafetensorsStreamer() as streamer:
-        streamer.stream_files(hf_weights_files)
+
+        def get_device():
+            is_cuda = current_platform.is_cuda()
+            return f"cuda:{torch.cuda.current_device()}" if is_cuda else None
+
+        if is_distributed and (device := get_device()):
+            streamer.stream_files(hf_weights_files, s3_credentials=None, device=device, is_distributed=True)
```
bbartels:
```suggestion
            streamer.stream_files(
                hf_weights_files,
                s3_credentials=None,
                device=device,
                is_distributed=True
            )
```

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -657,10 +657,24 @@ def _load_file(st_file: str):
...
+        def get_device():
+            is_cuda = current_platform.is_cuda()
+            return f"cuda:{torch.cuda.current_device()}" if is_cuda else None
+
+        if is_distributed and (device := get_device()):
+            streamer.stream_files(
+                hf_weights_files,
+                s3_credentials=None,
+                device=device,
+                is_distributed=True
```
bbartels:
```suggestion
                is_distributed=True,
```

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -657,10 +657,24 @@ def _load_file(st_file: str):
...
+
+        if is_distributed and (device := get_device()):
+            streamer.stream_files(
+                hf_weights_files,
+                s3_credentials=None,
+                device=device,
+                is_distributed=True,
+            )
+        else:
+            streamer.stream_files(hf_weights_files)
```
omer-dayan:
```suggestion
        def get_device():
            is_cuda = current_platform.is_cuda()
            return f"cuda:{torch.cuda.current_device()}" if is_cuda else "cpu"

        streamer.stream_files(
            hf_weights_files,
            device=get_device(),
            is_distributed=is_distributed,
        )
```

What do you think? Single point of calling to `streamer.stream_files` + no `if else`

omer-dayan:
Small proposal, but LGTM!
Thanks a lot!!!

vllm/model_executor/model_loader/weight_utils.py
bbartels: I do like that better, didn't know "cpu" would be a valid device in this case. Do I still need to pass `s3_credentials` here then as noa suggested? not sure how pythons default parameter resolution works.

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -657,10 +657,20 @@ def _load_file(st_file: str):
...
     with SafetensorsStreamer() as streamer:
-        streamer.stream_files(hf_weights_files)
+
+        def get_device():
+            is_cuda = current_platform.is_cuda()
+            return f"cuda:{torch.cuda.current_device()}" if is_cuda else "cpu"
+
+        streamer.stream_files(
+            hf_weights_files,
+            device=get_device(),
```
bbartels:
```suggestion
            device=get_device(),
            s3_credentials=None,
```

vllm/model_executor/model_loader/weight_utils.py
noa-neria: I like it better too. No need to pass `s3_credentials`. If you call the function without providing a value for a parameter that has a default, Python automatically uses that default value.

bbartels: I see, thanks, thought that'd be the case! Hence why i was confused as to why you had suggested to add `s3_credentials` to the previous version, given the following parameters were named arguments. I'll drop it from the implementation.

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -657,10 +657,21 @@ def _load_file(st_file: str):
...
+            s3_credentials=None,
```
bbartels:
```suggestion
```

vllm/model_executor/model_loader/weight_utils.py
bbartels:
```suggestion
        is_cuda = current_platform.is_cuda()
        device = f"cuda:{torch.cuda.current_device()}" if is_cuda else "cpu"

        streamer.stream_files(
            hf_weights_files,
            device=device,
```

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -657,10 +657,19 @@ def _load_file(st_file: str):
 def runai_safetensors_weights_iterator(
     hf_weights_files: list[str],
     use_tqdm_on_load: bool,
+    is_distributed: bool = False,
 ) -> Generator[tuple[str, torch.Tensor], None, None]:
     """Iterate over the weights in the model safetensor files."""
     with SafetensorsStreamer() as streamer:
-        streamer.stream_files(hf_weights_files)
+
```
bbartels:
```suggestion
```

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -657,10 +657,18 @@ def _load_file(st_file: str):
 def runai_safetensors_weights_iterator(
     hf_weights_files: list[str],
     use_tqdm_on_load: bool,
+    is_distributed: bool = False,
 ) -> Generator[tuple[str, torch.Tensor], None, None]:
     """Iterate over the weights in the model safetensor files."""
     with SafetensorsStreamer() as streamer:
-        streamer.stream_files(hf_weights_files)
+        is_cuda = current_platform.is_cuda()
+        device = f"cuda:{torch.cuda.current_device()}" if is_cuda else "cpu"
```
DarkLight1337: Use `current_platform.current_device()`? But I guess it doesn't matter too much since it only works with CUDA anyway

DarkLight1337: cc @tjtanaa can you check if this works on ROCm as well? If so we can change the condition to `current_platform.is_cuda_alike`

tjtanaa: Sure.

DarkLight1337: Yes, I think it just needs `>=0.15`

tjtanaa: @DarkLight1337  @bbartels 
Validated. runai_streamer is platform agnostic so it also works on ROCm

So the code should be

```python
        is_cuda_alike = current_platform.is_cuda_alike()
        device = f"cuda:{current_platform.current_device()}" if is_cuda_alike else "cpu"
```

noa-neria: @DarkLight1337  @tjtanaa distributed streamer should accept the device identifier name e.g. `cuda:0`, `cuda:1` etc. 
Distribution is based on torch distributed group and broadcast, so  ROCm should work in principal although we have not tested it.
It is possible to enable distributed streaming for other devices besides cuda devices, using `RUNAI_STREAMER_DIST=1` as explained [here](https://github.com/run-ai/runai-model-streamer/blob/master/docs/src/usage.md#control)

If there is a more general way to get the device identifier name, then it is safe to pass it to the streamer.
`current_platform.get_device_name`  returns just `"cuda"` without the identifier. I could not find `current_platform.current_device()`

tjtanaa: @noa-neria I have verified on MI300X. The above is the code changes to make it compatible with both NVIDIA and AMD.

noa-neria: Thanks for verifying! In this case for AMD there is no need to set `RUNAI_STREAMER_DIST=1`

bbartels:
```suggestion
        is_cuda_alike = current_platform.is_cuda_alike()
        device = f"cuda:{current_platform.current_device()}" if is_cuda_alike else "cpu"
```

docs/models/extensions/runai_model_streamer.md
```diff
@@ -45,6 +45,15 @@ vllm serve s3://core-llm/Llama-3-8b \
 
 You can tune parameters using `--model-loader-extra-config`:
 
+You can tune `distributed` that controls whether distributed streaming should be used. This is currently only possible on CUDA devices. This can significantly improve loading times from object storage or high-throughput network fileshares.
```
bbartels:
```suggestion
You can tune `distributed` that controls whether distributed streaming should be used. This is currently only possible on CUDA and ROCM devices. This can significantly improve loading times from object storage or high-throughput network fileshares.
```

vllm/model_executor/model_loader/weight_utils.py
bbartels: @tjtanaa @DarkLight1337 Amended the implementation to make it compatible with ROCM

aarnphm (approved):
ok it looks good, ig cyrus alr take a look for this.

mgoin (approved):

bbartels:
@noa-neria I am a bit surprised this test: https://github.com/vllm-project/vllm/blob/a99564ac5b2bc86d97cb18a7e18086b4ba94466a/tests/model_executor/model_loader/runai_model_streamer/test_weight_utils.py#L40

Is failing with: 
```
[2025-10-25T12:54:47Z] >               assert torch.all(runai_tensor.eq(hf_safetensors_tensors[name]))
[2025-10-25T12:54:47Z] E               RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```

When passing `is_distributed=False` to `stream_files` i would have expected the existing behaviour.

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -657,10 +657,18 @@ def _load_file(st_file: str):
 def runai_safetensors_weights_iterator(
     hf_weights_files: list[str],
     use_tqdm_on_load: bool,
+    is_distributed: bool = False,
 ) -> Generator[tuple[str, torch.Tensor], None, None]:
     """Iterate over the weights in the model safetensor files."""
     with SafetensorsStreamer() as streamer:
-        streamer.stream_files(hf_weights_files)
+        is_cuda_alike = current_platform.is_cuda_alike()
+        device = f"cuda:{current_platform.current_device()}" if is_cuda_alike else "cpu"
```
bbartels:
```suggestion
        is_cuda_alike = current_platform.is_cuda_alike()
        device = (
            f"cuda:{current_platform.current_device()}"
            if is_distributed and is_cuda_alike
            else "cpu"
        )
```

bbartels:
@noa-neria Thanks, amended!

PR #27234, opened 2025-10-21
[ResponseAPI] Fix mcp tool type extraction

Jialin:
## Purpose
For Mcp tool, we should extract `server_label` instead of `type` as tool_types.

Besides this major change, we also did a few side improvements:
1) Make envs.GPT_OSS_SYSTEM_TOOL_MCP_LABELS as a set to speed up existence checks
2) Move tool_types extractions to a dedicated function for easier maintainances and allow individual unit tests
3) More unit test added

### Unit tests
```
pytest \
  tests/entrypoints/test_harmony_utils.py \
  tests/test_envs.py
```

### Eval
gptoss120b tool use AIME eval correctness on-par.
- medium effort: 0.92
- high effort: 0.97+

---
<details>

</details>

### DIFF
--- a/tests/test_envs.py
+++ b/tests/test_envs.py
@@ -360,8 +360,8 @@ class TestEnvSetWithChoices:
             with pytest.raises(ValueError, match="Invalid value 'invalid' in TEST_ENV"):
                 env_func()
 
-    def test_duplicate_values_preserved(self):
-        """Test that duplicate values in the list are preserved."""
+    def test_duplicate_values_deduped(self):
+        """Test that duplicate values in the list are deduped."""
         with patch.dict(os.environ, {"TEST_ENV": "option1,option1,option2"}):
             env_func = env_set_with_choices("TEST_ENV", [], ["option1", "option2"])
             assert env_func() == {"option1", "option2"}
--- a/tests/test_envs.py
+++ b/tests/test_envs.py
@@ -10,6 +10,7 @@ import vllm.envs as envs
 from vllm.envs import (
     enable_envs_cache,
     env_list_with_choices,
+    env_set_with_choices,
     env_with_choices,
     environment_variables,
 )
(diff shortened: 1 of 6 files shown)


tests/test_envs.py
```diff
@@ -257,3 +258,110 @@ def test_duplicate_values_preserved(self):
...
+        with patch.dict(os.environ, {"TEST_ENV": "dynamic1,invalid"}):
+            env_func = env_set_with_choices("TEST_ENV", [], get_choices)
+            with pytest.raises(ValueError, match="Invalid value 'invalid' in TEST_ENV"):
+                env_func()
+
+    def test_duplicate_values_preserved(self):
+        """Test that duplicate values in the list are preserved."""
+        with patch.dict(os.environ, {"TEST_ENV": "option1,option1,option2"}):
+            env_func = env_set_with_choices("TEST_ENV", [], ["option1", "option2"])
+            assert env_func() == {"option1", "option2"}
```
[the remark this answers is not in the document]
Jialin: Updated to test_duplicate_values_deduped

vllm/entrypoints/openai/serving_responses.py
```diff
@@ -919,17 +920,7 @@ def _construct_input_messages_with_harmony(
         messages: list[OpenAIHarmonyMessage] = []
         if prev_response is None:
             # New conversation.
-            tool_types = [tool.type for tool in request.tools]
```
yeqcharlotte, line 922: we seem to change the logic here? we used to only extract tools specified per request and now we are adding all of them?

Jialin: @alecsolder to double check the expected behavior.

Actually, there IS a behavior change, but we did not add new tool types, but excluding "mcp" tool type from the result list while other values are the same.

vllm/entrypoints/harmony_utils.py
```diff
@@ -58,15 +58,36 @@
...
-    return not set(tool_types).issubset(BUILTIN_TOOLS)
+def has_custom_tools(tool_types: set[str]) -> bool:
+    """
+    Checks if the given tool types are custom tools
+    (i.e. any tool other than MCP buildin tools)
+    """
+    return not tool_types.issubset(MCP_BUILTIN_TOOLS)
+
+
+def extract_tool_types(tools: list[Tool]) -> set[str]:
```
alecsolder: This is strictly Responses API concepts, so it should probably stay in serving_responses.py, same with the test for this as well

vllm/entrypoints/openai/serving_responses.py
alecsolder: The new behavior is correct, we're using MCP tools as an "alias" for the "built-in" tools/. The bug before is that we would leave the type of "mcp" in the list, which would cause a developer message to exist even when only "built-in" tools are enabled

alecsolder (commented):
We should be able to add an e2e test in test_response_api_mcp_tools.py with `"enable_response_messages": True` to verify that there is no developer message in input_messages in test_mcp_tool_env_flag_enabled

vllm/entrypoints/harmony_utils.py
Jialin: Done, move the logic and unit tests from harmony_utils to serving_responses.

Jialin:
Close this prototype PR as the actual fix is landed in #27689 already.

Issue #27181, opened 2025-10-20
[Bug]: `check_enough_kv_cache_memory` didn't consider `num_gpu_blocks_override`


heheda12345:
### Your current environment

H100

### 🐛 Describe the bug

I can launch the server with 
```
vllm serve facebook/opt-125m --num_gpu_blocks_override=1
....
(EngineCore_DP0 pid=2717573) INFO 10-19 21:10:01 [kv_cache_utils.py:772] Overriding num_gpu_blocks=125643 with num_gpu_blocks_override=1
(EngineCore_DP0 pid=2717573) INFO 10-19 21:10:01 [kv_cache_utils.py:1201] GPU KV cache size: 16 tokens
(EngineCore_DP0 pid=2717573) INFO 10-19 21:10:01 [kv_cache_utils.py:1206] Maximum concurrency for 2,048 tokens per request: 0.01x
...
```
However, as there is only one block, no request with length > 16 can be scheduled.
the expect behavior should be raising an error during initialization like 
https://github.com/vllm-project/vllm/blob/f32bf7582e74ef967e657a25fd93186b31a46bed/vllm/v1/core/kv_cache_utils.py#L667-L687

Hope it can be fixed when iterating on https://github.com/vllm-project/vllm/pull/26939, but create a seperate issue to track it.

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

heheda12345:
@ayushsatyam146 Thank you very much. Can you sync with @elaineyz on this? That PR may include the fix of this bug. https://github.com/vllm-project/vllm/pull/26939

PR #27238, opened 2025-10-21
v1/kv_cache_utils: Respect num_gpu_blocks_override in memory check

khaled-wsa:
Summary
- Fixes bug where `check_enough_kv_cache_memory` ignored `num_gpu_blocks_override`, allowing engine initialization with an insufficient number of KV blocks for `max_model_len`.
- Adds a unit test to ensure that when `num_gpu_blocks_override` is too small (e.g., 1), initialization raises a clear error even if raw `available_memory` is large.

Context
- Reported in vllm-project/vllm#27181.
- Repro: `vllm serve facebook/opt-125m --num_gpu_blocks_override=1` appears to start, but no request with length > block size can be scheduled. Expect early failure during initialization.

Technical Details
- In `check_enough_kv_cache_memory`:
  - Validate override against per-layer requirements: compute `ceil(spec.max_memory_usage_bytes / spec.page_size_bytes)` for each layer and ensure `num_gpu_blocks_override >= max(required_blocks_per_layer)`. This closes the hole for heterogeneous specs (e.g., cross-attn vs self-attn).
  - Cap raw `available_memory` by `sum(page_size_bytes) * num_gpu_blocks_override` to form `effective_available_memory`.
  - Compare `needed_memory` to `effective_available_memory`, and pass the effective capacity to `estimate_max_model_len` for accurate guidance.
  - Improve the error message to explicitly mention when the override constrains effective capacity.

Files Changed
- vllm/v1/core/kv_cache_utils.py
  - Enforce per-layer minimum blocks for `num_gpu_blocks_override` and apply memory cap. Adjust error message.
- tests/v1/core/test_kv_cache_utils.py
  - Add `test_check_enough_kv_cache_memory_respects_num_gpu_blocks_override`.
  - Add `test_override_must_cover_worst_layer_blocks_in_heterogeneous_model` to cover cross-attn vs self-attn scenario.

How To Test
- Unit test (CPU):
  - `pytest -q tests/v1/core/test_kv_cache_utils.py::test_check_enough_kv_cache_memory_respects_num_gpu_blocks_override`
- Manual sanity:
  - Start the server with a small model and `--num_gpu_blocks_override=1`.
  - Expect initialization to fail with a ValueError that mentions the effective capacity and the override value.

Notes
- The change is localized and only alters the pre-initialization capacity check; runtime behavior is unchanged.
- Works for both uniform and hybrid KV cache specs since the per-block total uses each layer's `page_size_bytes`.


elaineyz:
Hi @khaled-wsa, please see discussion under PR https://github.com/vllm-project/vllm/pull/26939. 

In addition to the `num_gpu_blocks_override` param, the initialization of a `null_block` may also reduce the total available memory. Could you factor that part into this PR as well? It will make `check_enough_kv_cache_memory` more robust.

heheda12345 (commented):
Thanks for the quick bugfix! Actually I'm thinking of merging `check_enough_kv_cache_memory` and `_report_kv_cache_config`. Basically, in `check_enough_kv_cache_memory` we only do
```
if available_memory <= 0:
        raise ValueError(
            "No available memory for the cache blocks. "
            "Try increasing `gpu_memory_utilization` when "
            "initializing the engine."
        )
```
and in _report_kv_cache_config, if max_concurrency < 1, raise error for
```
 raise ValueError(
            f"To serve at least one request with the models's max seq len "
            f"({max_model_len}), ({needed_memory / GiB_bytes:.2f} GiB KV "
            f"cache is needed, which is larger than the available KV cache "
            f"memory ({available_memory / GiB_bytes:.2f} GiB). "
            f"{estimated_msg} "
            f"Try increasing `gpu_memory_utilization` or decreasing "
            f"`max_model_len` when initializing the engine."
        )
 ```

PR #27240, opened 2025-10-21
[CPU]Improve dynamic 4bit moe performance

xiangze-arm:
Improve dynamic 4bit moe performance by avoid tensor concat

## Purpose
Improve dynamic 4bit moe performance on Arm CPU

## Test Plan
Tested locally with Qwen3-30B-A3B w4a8 model quantized with llmcompressor

## Test Result
See performance improvements, especially for small batch sizes.

## Performance
With this PR, on 32 Neoverse-N2 cores, Qwen3-30B-A3B 4bit model gets 8% throughput improvement with 64 prompts and 12% improvement with 16 prompts.

Bench command:
```
vllm bench throughput --dtype=float32 --num-prompts 64 --seed 0 --dataset-name sharegpt --max-model-len=2048 --input_len=64 --model Qwen3-30B-A3B-w4a8-channelwise
vllm bench throughput --dtype=float32 --num-prompts 16 --seed 0 --dataset-name sharegpt --max-model-len=2048 --input_len=64 --model Qwen3-30B-A3B-w4a8-channelwise
```


csrc/moe/dynamic_4bit_int_moe_cpu.cpp
```diff
@@ -21,6 +21,8 @@ inline torch::Tensor mm(const torch::Tensor& a, const torch::Tensor& packed_w,
 #endif
 }
 
+extern void silu_and_mul(torch::Tensor& out, torch::Tensor& input);
```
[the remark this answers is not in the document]
xiangze-arm: `silu_and_mul` function is declared in `csrc/cpu/op.h`, however there is a function definition in this header file, include this header causes multiple definition error.

csrc/moe/dynamic_4bit_int_moe_cpu.cpp
```diff
@@ -119,35 +114,33 @@ torch::Tensor dynamic_4bit_int_moe_cpu(
...
         constexpr double kAlpha = 1.702;                   // GPT-OSS default
         constexpr double kLimit = 7.0;                     // GPT-OSS default
         auto gate_c = at::clamp_max(g_part, kLimit);
         auto up_c = at::clamp(u_part, -kLimit, kLimit);
         auto glu = gate_c.mul(at::sigmoid(gate_c.mul(kAlpha)));
         act = up_c.add(1.0).mul(glu);
       } else {  // SiLU , SwiGLU_GU, vLLM maps silu to SiluAndMul()
-        act = at::silu(g_part).mul(u_part);
+        act = at::empty({te, I}, y13.options());
+        silu_and_mul(act, y13);
```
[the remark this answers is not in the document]
cfRod: can you add a fallback as per this suggestion and add tests  for unaligned hidden sizes ?

csrc/moe/dynamic_4bit_int_moe_cpu.cpp
cfRod: Please attach before/after numbers  to show the expected gain from the fused SiLU path.

csrc/moe/dynamic_4bit_int_moe_cpu.cpp
```diff
@@ -119,35 +114,33 @@ torch::Tensor dynamic_4bit_int_moe_cpu(
...
+        auto g_part = y13.narrow(/*dim=*/1, /*start=*/0, /*length=*/I);
+        auto u_part = y13.narrow(/*dim=*/1, /*start=*/I, /*length=*/I);
+        constexpr double kAlpha = 1.702;  // GPT-OSS default
+        constexpr double kLimit = 7.0;    // GPT-OSS default
         auto gate_c = at::clamp_max(g_part, kLimit);
         auto up_c = at::clamp(u_part, -kLimit, kLimit);
         auto glu = gate_c.mul(at::sigmoid(gate_c.mul(kAlpha)));
         act = up_c.add(1.0).mul(glu);
       } else {  // SiLU , SwiGLU_GU, vLLM maps silu to SiluAndMul()
-        act = at::silu(g_part).mul(u_part);
```
nikhil-arm, line 134: at::silu is optimized with pytorch sve vectorized classes which has double the throughput for exp operation compared to neon.
I suggest to atleast test if aten silu + mul is faster thant fused silu_mul from vllm neon vectorizer class.

it would be better if you check it on 256 bit sve machine as 128 bit sve vectorizer might not be present in torch mainline yet

nikhil-arm: Maybe the low perf improvements were because of silu + mul regressing

xiangze-arm:
Updated PR description with performance results.
The performance impact of silu_and_mul change is small. So I have removed silu_and_mul change according to review concerns.

bigPYJ1151 (approved):

PR #27241, opened 2025-10-21
[Bug] Qwen reasoning parser

hao-aaron:
## Purpose
When running with Qwen/Qwen3-30B-A3B, `reasoning_content` is empty despite correctly formed `<think>` tag. Relevant issue: https://github.com/ray-project/ray/issues/57919

## Test Plan
Run requests with Qwen/Qwen3-30B-A3B and observe that `reasoning_content` is populated instead of `content`.

---
<details>

</details>


vllm/model_executor/models/config.py
```diff
@@ -489,6 +489,14 @@ def verify_and_update_config(cls, vllm_config: "VllmConfig") -> None:
             logger.info("Using bfloat16 kv-cache for DeepSeekV3.2")
 
 
+class Qwen3MoeForCausalLMConfig(VerifyAndUpdateConfig):
+    @staticmethod
+    def verify_and_update_config(vllm_config: "VllmConfig") -> None:
+        structured_outputs_config = vllm_config.structured_outputs_config
+        if structured_outputs_config.reasoning_parser == "":
```
chaunceyjiang, line 496: How do you start vLLM?

This field structured_outputs_config.reasoning_parser can be set via the parameter --reasoning-parser qwen3.

hao-aaron: Is the user intended to supply these arguments? I seems that it would make more sense to automatically set them for reasoning models. I saw that gptoss had its reasoning-parser set automatically via `verify_and_update_config`. Thanks for reviewing!

hmellor: Yes I think right now it is expected that the user opts in to reasoning https://docs.vllm.ai/en/latest/features/reasoning_outputs.html

PR #27242, opened 2025-10-21
Mirroring changes in test-pipeline.yaml into test-amd.yaml

Alexei-V-Ivanov-AMD:
Mirroring changes in test-pipeline.yaml into test-amd.yaml

Signed-off-by: Alexei V. Ivanov <alexei.ivanov@amd.com>


.buildkite/test-amd.yaml
```diff
@@ -848,6 +850,18 @@ steps:
     - pytest -v -s models/multimodal -m core_model --ignore models/multimodal/generation/test_whisper.py --ignore models/multimodal/processing
     - cd .. && VLLM_WORKER_MULTIPROC_METHOD=spawn pytest -v -s tests/models/multimodal/generation/test_whisper.py -m core_model  # Otherwise, mp_method="spawn" doesn't work
 
+- label: Multi-Modal Accuracy Eval (Small Models) # 50min
+  mirror_hardwares: [amdexperimental]
```
zhewenl, line 854: why don't we use amd production to enable this in actual CI?

.buildkite/test-amd.yaml
```diff
@@ -923,8 +937,8 @@ steps:
     # Whisper needs spawn method to avoid deadlock
     - VLLM_WORKER_MULTIPROC_METHOD=spawn python3 examples/offline_inference/audio_language.py --model-type whisper
 
-- label: Blackwell Test # 38 min
-  timeout_in_minutes: 60
+- label: Blackwell Test # 21 min
```
zhewenl, line 940: Do we still need to test Blackwell on this AMD specific pipeline?

gshtras (approved):
Assuming gpu: h200 are skipped in this pipeline and not put load on the actual h200 nodes

.buildkite/test-amd.yaml
gshtras: This pipeline in its current state is meant to be for monitoring only and not gating for the PRs

.buildkite/test-amd.yaml
Alexei-V-Ivanov-AMD: We include that test group for completeness reasons. For the moment it is just a comment and not an instruction for action in test-amd.yaml. But things may change in the future.

PR #27244, opened 2025-10-21
[CPU]Improve cpu fused moe perf

xiangze-arm:
### Description
- Add oneDNN/ACL matmul path for AArch64 in fused moe

###  Test Plan
- Tested locally with Qwen3-30B-A3B model. 
- Unit test `pytest tests/kernels/moe/test_moe.py -k test_cpu_fused_moe_basic`

###  Test Result
- End to end test result is ok.
- Unit test passes.

### Performance
With this PR, MoE can go into onednn_mm path on AArch64 CPU. On 32 Neoverse-N2 cores, this PR gets about 1.6x throughput compared with current default path.

Bench command:
```
vllm bench throughput --dtype=bfloat16 --num-prompts 64 --seed 0 --dataset-name sharegpt --max-model-len=2048 --input_len=64 --model Qwen3-30B-A3B
```


vllm/model_executor/layers/fused_moe/cpu_fused_moe.py
```diff
@@ -237,7 +238,39 @@ def __call__(
...
+                    lambda x, handle=down_handle, bias=layer_w2_bias: ops.onednn_mm(
+                        handle, x, bias
+                    )
+                )
+            else:
+                layer.gate_up_linear.append(lambda x, w=layer_w13_weight, b=layer_w13_bias: F.linear(x, w, b))
+                layer.down_linear.append(lambda x, w=layer_w2_weight, b=layer_w2_bias: F.linear(x, w, b))
+        if use_onednn_mm: # remove weight
+            layer.w13_weight = torch.nn.Parameter(torch.empty(0), requires_grad=False)
+            layer.w2_weight = torch.nn.Parameter(torch.empty(0), requires_grad=False)
```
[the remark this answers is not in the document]
xiangze-arm: Fixed

xiangze-arm:
@mgoin @bigPYJ1151 Can you help to review this PR?

cc @fadara01 @nikhil-arm

fadara01:
Hi @xiangze-arm - can you test again with #27472?

vllm/model_executor/layers/fused_moe/cpu_fused_moe.py
```diff
@@ -237,7 +239,43 @@ def __call__(
 
 class CPUFusedMOE:
     def __init__(self, layer: torch.nn.Module) -> None:
-        pass
+        use_onednn_mm = ops._supports_onednn and ops.is_onednn_acl_supported()
```
fadara01, line 241: could you please add the perf implications of this PR to the description?

xiangze-arm:
I have tested this PR with #27472 fix. The unit test passed without crash.

PR description is also updated with performance result (1.6x throughput improvement).

vllm/model_executor/layers/fused_moe/cpu_fused_moe.py
```diff
@@ -5,12 +5,14 @@
...
 
+from vllm import _custom_ops as ops
 from vllm import envs
 
 
 def silu_and_mul(x: torch.Tensor) -> torch.Tensor:
-    d = x.shape[-1] // 2
-    return F.silu(x[..., :d]) * x[..., d:]
+    output = torch.empty_like(x[..., : x.shape[-1] // 2])
+    torch.ops._C.silu_and_mul(output, x)
```
cfRod: I left a similar comment on your 4bit PR 
what happens if this assert fails https://github.com/vllm-project/vllm/blob/4f882be4a0d319551ce2a0eadcace0e76f3432cd/csrc/cpu/activation.cpp#L11

fadara01: I guess we crash because the activation kernel does not have a tail loop.
We should address this, but i don't think this particular concern should block this PR.

fadara01: why is this change needed here? is this faster than doing pytorch's `silu` and then pytorch `mul`?
I'm asking because `exp()` (needed to implement `silu`) is not really vectorized for Arm in vllm (we just call `std::exp` on each element in the vector)

xiangze-arm: I have removed silu_and_mul change.

xiangze-arm: I have removed silu_and_mul change.

fadara01 (approved):

bigPYJ1151 (approved):

fadara01:
Ah, those CI failures are unrelated.
@xiangze-arm can you try updating your branch s.t. we can get this merged?

PR #27247, opened 2025-10-21
[Model] Upstream Deepseek-OCR model

### LINKED ISSUE #27236: [New Model]: deepseek-ai/DeepSeek-OCR upstream support
### The model to consider.

Model is already supported via OOT registration based on 0.8.4, but we should clean up the code and bring it in to the upstream repository.

### The closest model vllm already supports.

_No response_

### What's your difficulty of supporting the model you want?

_No response_

Isotr0py:
## Purpose
- Fix #27236 
- See if we can reuse deepseek-vl2's implementation and clip's implementation as much as possible 

## Test Result

---
<details>

</details>


vllm/model_executor/models/deepseek_ocr.py
```diff
@@ -0,0 +1,594 @@
...
+            self.view_seperator = nn.Parameter(torch.randn(n_embed) * embed_std)
+        else:
+            raise ValueError(
+                f"Only 2D tile_tag is supported currently, got: {self.tile_tag}"
+            )
+
+        if self.text_config.topk_method == "noaux_tc":
+            architectures = ["DeepseekV3ForCausalLM"]
+        elif not self.text_config.use_mla:
+            architectures = ["DeepseekForCausalLM"]
```
whx-sjtu, line 414: Why can't we use `DeepseekV2ForCausalLM` in non-mla scenario? Because `DeepseekV2ForCausalLM` also supports non-mla: 
https://github.com/vllm-project/vllm/blob/09a7e6f6179ba2ccb10c701fe5356740d4f5f9a8/vllm/model_executor/models/deepseek_v2.py#L1001
Currently `DeepseekForCausalLM` is hard coded to use triton version of `fused_experts` and `fused_topk` with all_reduce path of MoE implementation, witch fails to utilize `SharedFusedMoE` witch supports `CustomOp`. So if there are reasons that DeepseekForCausalLM must be used, I suggest upgrade it to reuse `SharedFusedMoE` too.

ywang96: I believe this was needed back then when our DSv2 implementation didn't support non-mla (you can see the same code in dsvl2) https://github.com/vllm-project/vllm/blob/ab3e80042eac24dd362408e6d63ad98768046359/vllm/model_executor/models/deepseek_vl2.py#L403-L409

but you're right this should be no longer needed.

Ucas-HaoranWei:
Great work! I have tested it.

vllm/model_executor/models/deepseek_ocr.py
ywang96: Upon some testing and discussion with @Ucas-HaoranWei - we decided to leave this as is for now since the MHA implementation between `DeepseekForCausalLM` and `DeepseekV2ForCausalLM` are different.
https://github.com/vllm-project/vllm/blob/ab3e80042eac24dd362408e6d63ad98768046359/vllm/model_executor/models/deepseek_v2.py#L318-L330
https://github.com/vllm-project/vllm/blob/ab3e80042eac24dd362408e6d63ad98768046359/vllm/model_executor/models/deepseek.py#L199-L215

We will leave the work to refactor them in a follow-up PR

ywang96 (approved):
Since the model implementation and sampling params have been verified by the original model author, let's get this in!

jeejeelee (approved):

jeejeelee:
We also need to update `https://github.com/vllm-project/vllm/blob/main/tests/models/registry.py`

vllm/model_executor/models/deepseek_ocr.py
```diff
@@ -0,0 +1,594 @@
...
+    #         hf_processor_mm_kwargs=hf_processor_mm_kwargs,
+    #     )
+
+
+@MULTIMODAL_REGISTRY.register_processor(
+    DeepseekOCRMultiModalProcessor,
+    info=DeepseekOCRProcessingInfo,
+    dummy_inputs=DeepseekOCRDummyInputsBuilder,
+)
+class DeepseekOCRForCausalLM(nn.Module, SupportsMultiModal, SupportsPP):
```
DarkLight1337, line 340: Can you support `merge_by_field_config=True` in a follow-up PR?

Isotr0py: Sure, working in https://github.com/Isotr0py/vllm/tree/deepseek-ocr-schema

PR #27255, opened 2025-10-21
Bugfix: Cutlass FP8 FusedMoE bad scaling factors

amirkl94:
## Purpose
When running cutlass FusedMoE FP8 the scaling factors that are passed are None. This PR passes the correct scaling factors and enables the relevant test.

## Test Plan
Enabled previously disabled `test_flashinfer_cutlass_moe_fp8_no_graph` .


tlrmchlsmth (approved):

mgoin (approved):
LGTM, please validate @wenscarl

vllm/model_executor/layers/fused_moe/flashinfer_cutlass_prepare_finalize.py
```diff
@@ -170,7 +170,7 @@ def prepare(
         self._apply_router_weight_on_input(
             a1, topk_weights, topk_ids, apply_router_weight_on_input
         )
-        if not self.use_dp:
+        if not self.use_dp and quant_config.quant_dtype != torch.float8_e4m3fn:
```
wenscarl: I suggest using a more descriptive condition, for example:
```
 if not self.use_dp and quant_config.quant_dtype == nvfp4:
     return a1, None, ...,  
```

vllm/model_executor/layers/fused_moe/config.py
```diff
@@ -463,16 +463,24 @@ def fp8_w8a8_moe_quant_config(
     per_act_token_quant: bool = False,
     per_out_ch_quant: bool = False,
     block_shape: list[int] | None = None,
+    a1_gscale: torch.Tensor | None = None,
```
wenscarl, line 466: The additional 4 scales were not present before 27223 and no issue without them. Is it possible to deduce them from others?

amirkl94: As far as I could tell, no

vllm/model_executor/layers/quantization/modelopt.py
```diff
@@ -570,9 +570,13 @@ def get_fused_moe_quant_config(
 
         return fp8_w8a8_moe_quant_config(
             w1_scale=layer.w13_weight_scale,
+            g1_alphas=(layer.w13_weight_scale * layer.w13_input_scale).squeeze(),
```
wenscarl, line 573: The g1_alphas can be computed by w1_scale and a1_scale here, right? Same in `test_flashinfer.py`.

amirkl94: I'd rather not move this calculation inside the function as I'm not sure if in the future other paths will require different factors. If it's important then sure I can move it to be calculated in `fp8_w8a8_moe_quant_config` .
@wenscarl wdyt?

wenscarl: Sound reasonable. Let's leave it as it is.

robertgshaw2-redhat:
Can anyone explain why an fp8 quant method needs global scales?

PR #27257, opened 2025-10-21
Fix EventPublisherFactory logic for disabled KV cache events and publisher is "null"

usberkeley:
## Purpose
Fixed a logic flaw in the `EventPublisherFactory.create()` method to ensure it properly returns `NullEventPublisher` when `enable_kv_cache_events=False` or `publisher="null"`.

#### Changes
- Fixed conditional logic in `EventPublisherFactory.create()` method
- When config.publisher == "null", directly return an instance of `NullEventPublisher()` to avoid calling the constructor.
Otherwise, the code will attempt to retrieve the constructor of `NullEventPublisher` from the registry and call it with parameters. However, `NullEventPublisher` does not accept these parameters, resulting in a `TypeError`.

## Test Plan
- New test cases verify correct behavior across various configuration combinations
- Ensures backward compatibility
- All existing tests continue to pass

## Test Result
Pass

---
<details>

</details>


vllm/distributed/kv_events.py
```diff
@@ -353,12 +353,16 @@ def create(
...
+            not config
+            or not config.enable_kv_cache_events
+            or config.publisher == "null"
+        ):
             return NullEventPublisher()
 
         config_dict = asdict(config)
 
-        kind = config_dict.pop("publisher", "null")
+        kind = config_dict.pop("publisher")
```
markmc, line 353: Just an observation, not a problem with this PR - if we ever add another publisher other than `ZmqEventPublisher` it will be required to accept all the members of `KEventsConfig` as constructor parameters, which doesn't make sense

vllm/distributed/kv_events.py
```diff
@@ -353,12 +353,16 @@ def create(
         cls, config: KVEventsConfig | None, data_parallel_rank: int = 0
     ) -> EventPublisher:
         """Create publisher from a config mapping."""
-        if not config:
+        if (
+            not config
+            or not config.enable_kv_cache_events
+            or config.publisher == "null"
+        ):
```
markmc, line 348: Maybe this whole `NullEventPublisher` thing just isn't very useful ... we could just return `None` and make scheduler like this:

```
        self.kv_event_publisher = EventPublisherFactory.create(
            self.kv_events_config,
            self.parallel_config.data_parallel_rank,
        )

        self.kv_cache_manager = KVCacheManager(
            ...
            enable_kv_cache_events=self.kv_event_publisher is not None,
            ...
        )

       # publish collected KV cache events                                                                                                                                    
        if events:
            assert self.kv_event_publisher is not None
```

wdyt?

markmc (approved):
lgtm, thanks!

A suggestion for better (IMO) improvement inline

vllm/distributed/kv_events.py
usberkeley: @markmc Wow, perfect, I completely agree with this revision direction. It has the following benefits:

1. `Scheduler`: Avoids redundant checks for whether KV Cache Event is enabled.
```python
self.enable_kv_cache_events = (
    self.kv_events_config is not None
    and self.kv_events_config.enable_kv_cache_events
)
```

2. `KVCacheManager`: The parameter `enable_kv_cache_events` only stores events in the event queue when an `EventPublisher` exists, so assigning this parameter based on `self.kv_event_publisher is not None` is reasonable.

I understand that the original author intended to use the Null Object Design Pattern, returning a `NullEventPublisher` to avoid redundant null-check logic. However, using None directly seems simpler.

vllm/distributed/kv_events.py
usberkeley: Yep, new EventPublisher may not need to implement all parameters of KEventsConfig

PR #27261, opened 2025-10-21
Feature: Support Relu2 in FusedMoE fp8 cutlass path

amirkl94:
## Purpose
This PR enables FusedMoE FP8 cutlass path for models using non-gated relu2 models. This new path gives a performance gain of around 20% on output token throughput over the triton path.
This PR requires flashinfer `0.5.0` .

## Tests
New parameterization in `test_flashinfer.py::test_flashinfer_cutlass_moe_fp8_no_graph` to verify the new non-gated activation.

Performance tests:
Ran on a single H100. Started the server (once with `VLLM_USE_FLASHINFER_MOE_FP8=1` and once with `VLLM_USE_FLASHINFER_MOE_FP8=0`):

```
python3 -m vllm.entrypoints.openai.api_server --host 0.0.0.0 --port 8080 --model <NemotronNanov3-modelopt-fp8-ckpt> --tensor-parallel-size 1 --pipeline-parallel-size 1 --swap-space 16 --max-num-seqs 512 --trust-remote-code --max-model-len 3800 --gpu-memory-utilization 0.9 --max-num-batched-tokens 8192 --enable-chunked-prefill --disable-log-requests
```

Benchmarked using:
```
vllm bench serve --backend vllm --host 0.0.0.0 --port $port --model $model_path --num-prompts 1280--trust-remote-code --ignore-eos --max-concurrency 256 --random-input-len 1024 --random-output-len 1024 --no-stream --dataset-name random --num-warmups 20
```

triton path yielded: 
```
Request throughput (req/s):              3.47
Output token throughput (tok/s):         3551.76
Peak output token throughput (tok/s):    6116.00
```

cutlass path yielded:
```
Request throughput (req/s):              5.15
Output token throughput (tok/s):         5270.17
Peak output token throughput (tok/s):    7168.00
```

~17% perf gain for peak (decode), ~40% perf gain for average token throughput (includes prefill).


vllm/model_executor/layers/quantization/modelopt.py
```diff
@@ -635,10 +665,6 @@ def apply(
                 apply_router_weight_on_input=apply_router_weight_on_input,
             )
         elif self.flashinfer_moe_backend == FlashinferMoeBackend.CUTLASS:
-            assert not renormalize
```
tomeras91, line 645: 1. why did you remove the assert that renormalize is not True?
2. Maybe worth asserting that activation is either "silu" or "relu2"?

vllm/model_executor/layers/fused_moe/flashinfer_cutlass_moe.py
```diff
@@ -140,10 +140,12 @@ def apply(
         expert_tokens_meta: mk.ExpertTokensMetadata | None,
         apply_router_weight_on_input: bool | None,
     ):
-        assert activation == "silu", (
```
tomeras91, line 151: maybe worth asserting that activation is one of the keys in `activation_str_to_value_map`?

vllm/model_executor/layers/quantization/modelopt.py
amirkl94: 1. The `renormalize` is redundant here as it is required by `select_experts` only. This is only relevant for the trtllm path.
2. 👍

vllm/model_executor/layers/quantization/modelopt.py
```diff
@@ -567,13 +572,21 @@ def get_fused_moe_quant_config(
 
         return fp8_w8a8_moe_quant_config(
             w1_scale=layer.w13_weight_scale,
-            g1_alphas=(layer.w13_weight_scale * layer.w13_input_scale).squeeze(),
+            g1_alphas=layer.output1_scales_gate_scalar.squeeze()
```
tlrmchlsmth: It looks like `output1_scales_gate_scalar` and `output2_scales_scalar` are only used in `flashinfer_trtllm_moe`.  It's not clear from the vLLM source code how these are getting set. Does flashinfer add them to the layer's member variables?

I think we should try to keep the quantization format decoupled from the kernels used for the implementation

amirkl94: These factors are registered into the layer during `process_weights_after_loading` (line 566) .
They are registered to the layer when we're using cutlass or trtllm backends, using `register_moe_scaling_factors` .
The problem I see with trying to make this decouple is that the cutlass path uses the layers' `get_fused_moe_quant_config` to get the relevant scaling factors [here](https://github.com/vllm-project/vllm/blob/6f7de33bed412869bec4631add885e5ff88c22cf/vllm/model_executor/layers/quantization/utils/flashinfer_utils.py#L230) .
There are 2 options I see here let me know which one you prefer:
1. I can remove the `if else` when setting the quantization and set it the same for all paths.
2. I can build the needed quantization for the flashinfer cutlass kernel [in here](https://github.com/vllm-project/vllm/blob/6f7de33bed412869bec4631add885e5ff88c22cf/vllm/model_executor/layers/quantization/utils/flashinfer_utils.py#L230) , and not change the quantization in the ModelOptFusedMoEFP8 object.

vllm/model_executor/layers/quantization/modelopt.py
```diff
@@ -354,12 +354,16 @@ def __init__(
...
+        if envs.VLLM_USE_FLASHINFER_MOE_FP8 and has_flashinfer_moe():
             self.flashinfer_moe_backend = get_flashinfer_moe_backend()
+            if (
+                self.flashinfer_moe_backend == FlashinferMoeBackend.TENSORRT_LLM
+                and not self.moe.is_act_and_mul
+            ):
+                logger.info_once(
+                    "Non-gated MoE is not supported for min-latency mode,"
+                    "falling back to high-throughput mode"
+                )
```
mgoin, line 366: It seems you are missing the override of `self.flashinfer_moe_backend` here

vllm/model_executor/layers/quantization/modelopt.py
mgoin: > I can remove the if else when setting the quantization and set it the same for all paths.
I think I'd prefer to just provide the same information in all cases. The kernel can choose to ignore it

vllm/model_executor/layers/quantization/modelopt.py
mgoin, line 366: It looks like the `self.flashinfer_moe_backend` override was left out

mgoin (approved):

vllm/model_executor/layers/quantization/modelopt.py
```diff
@@ -642,9 +649,9 @@ def apply(
         )
 
         if self.flashinfer_moe_backend == FlashinferMoeBackend.CUTLASS:
-            assert not renormalize
-            assert activation == "silu", (
-                f"Expected 'silu' activation but got {activation}"
+            assert activation in ("silu", "relu2_no_mul"), (
+                "Expected activation to be in ('silu', 'relu2_no_mul'),"
+                "but got {activation}"
```
mgoin: Missing format
```suggestion
                f"but got {activation}"
```

tests/kernels/moe/test_flashinfer.py
```diff
@@ -79,10 +79,16 @@ class TestData:
 
     @staticmethod
     def make_moe_tensors_8bit(
-        m: int, k: int, n: int, e: int, reorder: bool
+        m: int, k: int, n: int, e: int, reorder: bool, activation: str = "silu"
     ) -> "TestData":
+        is_gated = True
+        if activation != "relu2_no_mul":
+            is_gated = False
```
mgoin: This doesn't seem right as it breaks test_flashinfer_per_tensor_moe_fp8_no_graph on blackwell
https://buildkite.com/vllm/ci/builds/38920/steps/canvas?jid=019a7fad-270b-4d40-8820-e3a1e75dc35e#019a7fad-270b-4d40-8820-e3a1e75dc35e/102-2387

mgoin (changes requested):
PTAL at the failing blackwell test

tests/kernels/moe/test_flashinfer.py
amirkl94: Yeah it should be `if activation == "relu2_no_mul:` , I originally wrote it as a one liner but the pre-commit hook complained, and I fixed it incorrectly. Changed it back and this should be correct.

mgoin (approved):
Thank you!

PR #23014, opened 2025-08-15
Support RL online quantization with torchao

jerryzh168:
Summary:
This is to enable online quant for verl. The PR
added support for initializing a TorchAOConfig object in vllm
through a serialized json file that specifies the type of quantization
people want. Or a json serialized TorchAOConfig object

Code for serializing the config to json:
```
from torchao.quantization import Float8DynamicActivationFloat8WeightConfig, PerRow
from torchao.core.config import config_to_dict
import json

config = Float8DynamicActivationFloat8WeightConfig(granularity=PerRow())

json_str = json.dumps(config_to_dict(config))

LLM(..., quantization="torchao", hf_overrides={"quantization_config_dict_json": json_str})

```

Code for serializing the config to file
```
from torchao.quantization import Float8DynamicActivationFloat8WeightConfig, PerRow
from torchao.core.config import config_to_dict
import json

config = Float8DynamicActivationFloat8WeightConfig(granularity=PerRow())

with open("torchao_config.json", "w") as f:
    f.write(json.dumps(config_to_dict(config)))

LLM(..., quantization="torchao", hf_overrides={"quantization_config_file": "torchao_config.json"})
```

This also supports module level config as well through the `ModuleFqnToConfig` config
https://huggingface.co/docs/transformers/main/en/quantization/torchao#per-module-quantization
although not tested yet.

more configs: https://docs.pytorch.org/ao/main/api_ref_quantization.html#inference-apis-for-quantize

Note: this has incorporated changes from @LiyuanLucasLiu's PR: https://github.com/vllm-project/vllm/pull/23901, although vllm fp8 quant method is not supported yet, we can add that in a separate PR

Test Plan:
pytest tests/quantization/test_torchao.py -k test_on_the_fly_quant
pytest tests/quantization/test_torchao.py -k test_reload_weights

and regression tests
pytest tests/quantization/test_torchao.py

Reviewers:

Subscribers:

Tasks:

Tags:

PR #27280, opened 2025-10-21
[Quantization] Support pre-load  online quantization for  compressed-tensors W8A8 channel-wise schema

luccafong:
## Purpose

Support pre-load online quantization for  compressed-tensors W8A8 channel-wise schema on bf16 ckpt.

-  **Memory Optimization**: This differentiates from the other online quantization approach that is post loading through `process_weights_after_loading`, we add quantization of weights through `process_weights_before_loading` when the hardware GPU can not hold the original dtype weights and cpu offloading is too slow. This approach quantize each weight while loading. This enables online dynamic quantization for **Llama4 Maverick Raw BF16 on H100**, which is not doable before.
- **Extendible**: The PR implements compressed-tensor fp8 channelwise (same as FP_dynamic in offline quantization), while the approach is extendible to other quantization method if the method implements process_weights_before_loading.
- **MOE and Linear Support**: This support both MOE and linear layers.
- **llama4 specific optimization**: Llama4 has a transposed/chunked fused weights when calling weight loader, this PR also has an improvement on llama4 model loading that copy to device before transpose/chunk happenning to avoid expensive contiguous (the model loading reduced from 40 minutes to 2 mintues)

```
 --quantization compressed-tensors \
--quantization-schema fp8_channelwise \
--hf-overrides '{"quantization_config":{"ignore":["re:.*self_attn","re:.*lm_head","re:.*router","re:.*vision_model","re:.*multi_modal_projector","re:.*feed_forward.gate_up_proj","re:.*feed_forward.down_proj", "re:.*shared_expert"]}}'
```

### Online Serving
```
vllm serve /data/local/models/oss/Llama-4-Maverick-17B-128E-Instruct -tp 8 --quantization compressed-tensors --quantization-schema fp8_channelwise --hf-overrides '{"quantization_config":{"ignore":["re:.*self_attn","re:.*lm_head","re:.*router","re:.*vision_model","re:.*multi_modal_projector","re:.*feed_forward.gate_up_proj","re:.*feed_forward.down_proj", "re:.*shared_expert"]}}' --max_num_seqs 32 --max-model-len 32768
```

### UT/CI Tests

```
pytest tests/quantization/test_fp8_channelwise.py
pytest tests/quantization/tes
t_compressed_tensors.py -k "test_compressed_tensors_fp8_online_quantization_channelwise"
```

Added both llama3 and Qwen to guard linear and MOE models.

## Test Result

```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.940|±  |0.0168|
|     |       |strict-match    |     5|exact_match|↑  |0.945|±  |0.0162|
```

---
<details>

</details>

### DIFF
--- a/docs/features/quantization/fp8.md
+++ b/docs/features/quantization/fp8.md
@@ -140,7 +140,7 @@ print(result[0].outputs[0].text)
 
 ### Online Channelwise Quantization
 
-For improved accuracy with online quantization, vLLM supports **channelwise quantization** using the `compressed-tensors` quantization method. This approach quantizes weights on a per-channel basis during model loading, providing better accuracy than per-tensor quantization while still requiring no calibration data.
+vLLM supports **channelwise quantization** using the `compressed-tensors` quantization method. This approach quantizes weights on a per-channel basis during model loading, providing better accuracy than per-tensor quantization while still requiring no calibration data.
 
 To enable channelwise online quantization, use both the `--quantization` and `--quantization-schema` flags:
 
@@ -169,7 +169,6 @@ This mode provides:
 - **Per-channel weight quantization**: Each output channel of weight matrices gets its own scale factor, preserving more information than per-tensor scaling
 - **Dynamic per-token activation quantization**: Activations are quantized dynamically per token for maximum accuracy
 - **No calibration required**: Quantization happens automatically during model loading based on the weight statistics
-- **Better accuracy**: Channelwise quantization typically provides better accuracy than per-tensor quantization, especially for models with varying weight distributions across channels
 
 !!! note
     Channelwise quantization is currently supported for FP8 (W8A8) with the `compressed-tensors` quantization method. It uses static per-channel quantization for weights and dynamic per-token quantization for activations.
--- a/docs/features/quantization/fp8.md
+++ b/docs/features/quantization/fp8.md
@@ -138,5 +138,41 @@ result = llm.generate("Hello, my name is")
 print(result[0].outputs[0].text)
 ```
 
+### Online Channelwise Quantization
+
+For improved accuracy with online quantization, vLLM supports **channelwise quantization** using the `compressed-tensors` quantization method. This approach quantizes weights on a per-channel basis during model loading, providing better accuracy than per-tensor quantization while still requiring no calibration data.
+
+To enable channelwise online quantization, use both the `--quantization` and `--quantization-schema` flags:
+
+```bash
+vllm serve meta-llama/Llama-3-8B-Instruct \
+  --quantization compressed-tensors \
+  --quantization-schema fp8_channelwise
+```
+
+Or in Python:
+
+```python
+from vllm import LLM
+
+llm = LLM(
+    "meta-llama/Llama-3-8B-Instruct",
+    quantization="compressed-tensors",
+    quantization_schema="fp8_channelwise"
+)
+result = llm.generate("Hello, my name is")
+print(result[0].outputs[0].text)
+```
+
+This mode provides:
+
+- **Per-channel weight quantization**: Each output channel of weight matrices gets its own scale factor, preserving more information than per-tensor scaling
+- **Dynamic per-token activation quantization**: Activations are quantized dynamically per token for maximum accuracy
+- **No calibration required**: Quantization happens automatically during model loading based on the weight statistics
+- **Better accuracy**: Channelwise quantization typically provides better accuracy than per-tensor quantization, especially for models with varying weight distributions across channels
+
+!!! note
+    Channelwise quantization is currently supported for FP8 (W8A8) with the `compressed-tensors` quantization method. It uses static per-channel quantization for weights and dynamic per-token quantization for activations.
+
 !!! warning
     Currently, we load the model at original precision before quantizing down to 8-bits, so you need enough memory to load the whole model.
--- a/tests/quantization/test_compressed_tensors.py
+++ b/tests/quantization/test_compressed_tensors.py
@@ -10,6 +10,7 @@ import torch
 from compressed_tensors.quantization import QuantizationType
 
 from tests.models.utils import check_logprobs_close
+from vllm.model_executor.layers.linear import UnquantizedLinearMethod
 from vllm.model_executor.layers.quantization.compressed_tensors.compressed_tensors import (  # noqa: E501
     CompressedTensors24,
     CompressedTensorsLinearMethod,
@@ -21,6 +22,9 @@ from vllm.model_executor.layers.quantization.compressed_tensors.compressed_tenso
     CompressedTensorsW8A8Int8,
     CompressedTensorsW8A16Fp8,
     CompressedTensorsWNA16,
+)
+from vllm.model_executor.layers.quantization.compressed_tensors.compressed_tensors_moe import (  # noqa: E501
+    CompressedTensorsW8A8Fp8MoEMethod,
 )
 from vllm.model_executor.layers.quantization.input_quant_fp8 import QuantFP8
 from vllm.model_executor.layers.quantization.utils.fp8_utils import W8A8BlockFp8LinearOp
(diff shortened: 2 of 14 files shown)


vllm/config/model.py
```diff
@@ -196,6 +197,9 @@ class ModelConfig:
     `quantization_config` attribute in the model config file. If that is
     `None`, we assume the model weights are not quantized and use `dtype` to
     determine the data type of the weights."""
+    quantization_schema: str | None = None
```
jerryzh168, line 192: can we just use `quantization` and `hf_overrides` to specify the config? like https://github.com/vllm-project/vllm/pull/23014

vllm/model_executor/layers/quantization/__init__.py
```diff
@@ -158,6 +158,42 @@ def get_quantization_config(quantization: str) -> type[QuantizationConfig]:
...
 
 
+def get_default_quantization_hf_config(
+    quantization: str, quantization_schema: str | None
+) -> dict[str, Any]:
+    if (
+        quantization == "compressed-tensors"
+        and quantization_schema == "fp8_channelwise"
+    ):
+        return {
```
jerryzh168, line 173: yeah I guess this could be passed as a json string, or a file that stores the serialized json string

otherwise we'd need to invent quantization_schema name for each settings

vllm/config/model.py
luccafong: I feel the user experience might be too complicated for fp8_channelwise, so add some schema for pre-defined config here.

jerryzh168: I guess it's fine to use a string, maybe reuse `hf_override` to specify the string?

vllm/model_executor/layers/quantization/__init__.py
hmellor: JSON configs can also be passed with `.` notation. i.e. `--quantization-scheme.format float-quantize --quantization-scheme.quant_method compressed-tensors ...`, but this will be extremely verbose for a CLI.

Perhaps as a file is a good idea as @jerryzh168 suggested?

vllm/config/model.py
```diff
@@ -466,9 +477,14 @@ def __post_init__(
             # We'll determine how to apply dict overrides after loading the config
             hf_overrides_kw = {}
             dict_overrides = {}
+            if quant_config_override:
+                dict_overrides["quantization_config"] = quant_config_override
```
HDCharles: Should this warn if overriding non null quantization configuration? Otherwise may be unclear which takes precedence

vllm/model_executor/layers/quantization/__init__.py
luccafong: I feel not a good user experience if user have to define these settings in either file or pass through the args, so I think for frequently used schema, pre-defined should be fine  @hmellor  @jerryzh168

vllm/config/model.py
luccafong: @HDCharles  thanks for the comments, I moved where get_default_quantization_hf_config called closer to this line and also add to warn to the original method of get_default_quantization_hf_config

vllm/model_executor/layers/quantization/__init__.py
jerryzh168: I think it's OK to have predefined strings, but can this live in `hf_overrides`, instead of adding a new `quantization_schema` in parallel to `quantization`?

luccafong: later we may add fp8_blockwise, without a separate field, we could not differentiate, the current `quantization` are all compressed-tensors which is not enough to differentiate different schemas of the same `quantization` method

luccafong: ah, I see, do you mean we specifiy a field in hf config to differentiate?

vllm/config/model.py
```diff
@@ -466,9 +470,18 @@ def __post_init__(
...
+                quant_config_override = get_default_quantization_hf_config(
+                    self.quantization, self.quantization_schema
+                )
+                if quant_config_override:
+                    dict_overrides["quantization_config"] = quant_config_override
             for key, value in self.hf_overrides.items():
                 if isinstance(value, dict):
-                    dict_overrides[key] = value
+                    if dict_overrides.get(key):
+                        dict_overrides[key] = {**dict_overrides[key], **value}
```
HDCharles, line 482: this seems rather convoluted. 

We're assuming dict_overrides[key] is always a mapping as long as dict_overrides.get(key) and isinstance(value, dict) are true which doesn't seem obvious

as far as the logic:

we take the dict that we assume we get from dict_overrides[key] and add the value dict to it, letting value take precedence where there are any conflicts.

Feels like we should be taking one or the other i.e. overriding things rather than this merge operation.

also still no warning when one thing overrides another.

vllm/model_executor/layers/quantization/__init__.py
jerryzh168: > ah, I see, do you mean we specifiy a field in hf config to differentiate?

yeah just put this in hf_override is cleaner I think, since you already have some other configs there, and this is specific to llm compressor

tests/quantization/test_compressed_tensors.py
```diff
@@ -767,3 +771,126 @@ def check_model(model):
...
+    model_path = "Qwen/Qwen1.5-MoE-A2.7B"
+
+    with vllm_runner(
+        model_path,
+        quantization="compressed-tensors",
+        quantization_schema="fp8_channelwise",
+        enforce_eager=True,
+        dtype="bfloat16",
+        hf_overrides={
+            "quantization_config": {"ignore": ["re:.*self_attn", "re:.*lm_head"]}
```
jerryzh168, line 860: i.e. here:

```
hf_overrides={
            "llmc_quantization_schema"="fp8_channelwise",
            "llmc_quantization_config": {"ignore": ["re:.*self_attn", "re:.*lm_head"]}
}
```

docs/features/quantization/fp8.md
```diff
@@ -138,5 +138,40 @@ result = llm.generate("Hello, my name is")
...
+
+Or in Python:
+
+```python
+from vllm import LLM
+
+llm = LLM(
+    "meta-llama/Llama-3-8B-Instruct",
+    quantization="compressed-tensors",
+    quantization_schema="fp8_channelwise"
```
HDCharles, line 161: Shouldn't this match the per tensor API?

Super weird that

quantization="fp8" does per tensor while
quantization="compressed-tensoe", quantization_schema="fp8_channelwise" does per channel

Why not just quantization="fp8_channelwise"?

HDCharles (changes requested):
1) At a high level, the API for online per tensor fp8 should match the one for online per channel fp8. 

2) It looks like this PR takes a completely different path than the existing per tensor fp8 online support, creating entirely new preprocessing steps. rather than having 2 unrelated paths for these techniques, it's be much cleaner to either support both techniques with the new abstractions or just do whatever per tensor did initially.

docs/features/quantization/fp8.md
luccafong: good point, let's make the adjustment for the interface

HDCharles:
> 1. we can implement per tensor in this approach in another PR.
>
> 2. the issue with current one approach for online as stated in pr description is that it quantize after loading which can not work with the hardware that do not have enough memory to hold the original weights.


1. that's fair but can you create the issue or add todo in code?
2. I mostly meant that 1 of the 2 flows is going to be superfluous so we should consolidate one way or the other.

vllm/model_executor/layers/linear.py
```diff
@@ -286,6 +287,45 @@ def __init__(
...
+            *args,
+            **kwargs,
+        ):
+            if self.quant_method is None:
+                return weight_loader(param, loaded_weight, *args, **kwargs)
+            results = self.quant_method.process_weights_before_loading(
+                self, param, loaded_weight
+            )
+            scale = results.get("scale", None)
+            loaded_weight = results.get("quantized", loaded_weight)
```
jerryzh168, line 315: these seems to be specific to compressed tensor? maybe add a check to say that only compressed tensor is supported?

vllm/model_executor/layers/linear.py
```diff
@@ -286,6 +287,45 @@ def __init__(
...
+    def add_preprocessor_to_weight_loader(self, weight_loader: Callable) -> Callable:
+        def weigth_loader_with_preprocessor(
+            param: Parameter | BasevLLMParameter,
+            loaded_weight: torch.Tensor,
+            *args,
+            **kwargs,
+        ):
+            if self.quant_method is None:
+                return weight_loader(param, loaded_weight, *args, **kwargs)
+            results = self.quant_method.process_weights_before_loading(
```
kylesayrs, line 311: This doesn't seem like it would work if a layer wants to load parameters that aren't weights, for example, bias terms.

kylesayrs: Weight loaders are used to load all parameters belonging to a module, there isn't an easy way for a quant_config to choose which `process_weights_before_loading` function gets called for which parameter via the weight loaders.

You could do something like hold a reference to the parameter to check against, but that's pretty complicated, and it'd have to be a weak ref or else you could run into gc issues.

vllm/model_executor/layers/linear.py
```diff
@@ -286,6 +287,45 @@ def __init__(
...
+                self, param, loaded_weight
+            )
+            scale = results.get("scale", None)
```
kylesayrs, line 314: This logic is really specific to the particular `process_weights_before_loading` that is being called. Are you sure it should live here?

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -931,6 +931,22 @@ def weight_loader(
         expert_id: int,
         return_success: bool = False,
     ) -> bool | None:
+        results = self.quant_method.process_weights_before_loading(
+            self, param, loaded_weight
+        )
+        scale = results.get("scale", None)
+        loaded_weight = results.get("quantized", loaded_weight)
```
kylesayrs, line 938: Again, this logic is very specific to CT online quantization, not sure if it should be in the fused_moe logic.

kylesayrs (commented):
I think this approach is very different from how weight loading and online quantization has been handled up until now. I think the idea of a "weight preprocessing" step controlled by the quant_method may be the right way to handle this, but more thought needs to go into whether this design is going to scale and generalize beyond just the CT W8A8 use case.

vllm/model_executor/layers/linear.py
luccafong: we should find ways to call the preprocess for each weight loader that we need to apply on, do you have any suggestion to inject the preprocessor? since the caller is in model code instead of some more generalized location, unless we add a general wrapper of all weight loader where we can call it.

luccafong:
> I think this approach is very different from how weight loading and online quantization has been handled up until now. I think the idea of a "weight preprocessing" step controlled by the quant_method may be the right way to handle this, but more thought needs to go into whether this design is going to scale and generalize beyond just the CT W8A8 use case.

yeah not sweeping all quantization methods, but  CT and torchao are two valid use cases, and CT seems quite widely used, it is ok some quant method can not support it.

vllm/model_executor/models/llama4.py
```diff
@@ -446,6 +446,11 @@ def load_moe_expert_weights(
         # [num_experts, hidden_in, hidden_out], so we must transpose the last
         # two dimensions to match the expected layout of the parameters.
         if fused and loaded_weight.ndim == 3:
+            if self.vllm_config.model_config.quantization_schema:
```
vkuzo, line 449: can we just always do this?  HF LLaMa 4 MoE checkpoint format stores weights in an inoptimal format, I feel like we should always take the fast path here.

vllm/model_executor/layers/quantization/compressed_tensors/utils.py
```diff
@@ -214,3 +224,24 @@ def _match_fused_layer(
             unfused_matches.append(None)
 
     return unfused_matches[0] if all(unfused_matches) else None
+
+
+def fp8_channelwise_quantize(x: Tensor, channel_dim: int = -1) -> tuple[Tensor, Tensor]:
```
vkuzo, line 229: is there a higher level function in `compressed-tensors` to do this so we don't have to redefine it here? @kylesayrs  since I'm very n00b in compressed-tensors

PR #26377, opened 2025-10-07
[V1] [Hybrid] Mamba1 Automatic Prefix Caching

Josephasafg:
## Purpose
This PR enables prefix caching for Mamba1 models. It depends on the great work done in this [PR](https://github.com/vllm-project/vllm/pull/25752) for Mamba2 and its part of this [feature](https://github.com/vllm-project/vllm/issues/26201)

This PR adds support for block-wise computation in the SSM kernel using `mamba_block_size`.
Instead of calculating across the entire sequence of tokens at once, the kernel now (when prefix caching is enabled):

- Computes up to `mamba_block_size` tokens at a time.
- Stores the result of each range in a the appropriate indices in the ssm state.
- Use these stored intermediate states to calculate tokens that didn't fit in a mamba block.

At the moment, we use a default of 2048 for `mamba_block_size` for Mamba1 as this was hard-coded number for the kernel chunking, so in order for it to match as closely to Non-APC, we used the same value. 
It is however an option to pass `mamba_block_size=1024` for example as an engine arg to use smaller block sizes and therefore, get more cache hits.

For example, with a `mamba_block_size` of 2048 and a prompt length of 4500, the computation is as follows:

- [0–2047] → produces a first intermediate state.
- [2048–4095] → computed using the previous intermediate state, and stored as the next intermediate state.
- [4096–4499] → computed from the previous intermediate state, producing the final intermediate state, which is also considered the final ssm_state.

Todos - 
- [x] Combine Mamba1 and Mamba2 common functionalities (We should continue the consolidation work in a different PR)
- [x] Optimal `mamba_block_size`
- [x] Have all APC tests pass with no warnings
- [x] Show APC/No APC Test Results
- [x]  Merge this [PR](https://github.com/vllm-project/vllm/pull/27289) to enable `mamba-block-size` as an engine arg
- [x] Merge this [PR](https://github.com/vllm-project/vllm/pull/27339) to mark new Mamba models as supporting APC
## Test Plan

```python
from vllm import LLM, SamplingParams
from vllm.distributed import cleanup_dist_env_and_memory
import time

if __name__ == "__main__":
    MODEL = "ai21labs/AI21-Jamba-Reasoning-3B"
    PROMPT_MULTIPLE = 310
    sampling_params = SamplingParams(temperature=0.0)
    prefix = (  # examples/offline_inference/prefix_caching.py
        "You are an expert school principal, skilled in effectively managing "
        "faculty and staff. Draft 10-15 questions for a potential first grade "
[...]

PR #27289, opened 2025-10-21
[Hybrid] Add mamba_block_size to Engine Args

Josephasafg:
## Purpose
Now that prefix caching is enabled, we want the users to be able to control the size of the `mamba_block_size` to better customize it to their use-case.

At the moment, this flag can only be used with `--enable-prefix-caching`, as without it we allocate it in size of `max_model_len` for backwards compatibility.

## Test Result

---
<details>

</details>


Josephasafg:
cc: @tdoublep

vllm/model_executor/models/config.py
```diff
@@ -415,7 +413,9 @@ def verify_and_update_config(cls, vllm_config: "VllmConfig") -> None:
             def lcm(a, b):
                 return a * b // gcd(a, b)
 
-            base_chunk_size = model_config.get_mamba_chunk_size()
+            base_chunk_size = (
+                cache_config.mamba_block_size or model_config.get_mamba_chunk_size()
+            )
```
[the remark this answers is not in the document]
Josephasafg: The code added doesn't change the current behavior

vllm/engine/arg_utils.py
```diff
@@ -54,7 +54,12 @@
     VllmConfig,
     get_attr_docs,
 )
-from vllm.config.cache import BlockSize, CacheDType, MambaDType, PrefixCachingHashAlgo
+from vllm.config.cache import (
+    BlockSize,
+    CacheDType,
+    MambaDType,
+    PrefixCachingHashAlgo,
+)
```
hmellor: Unnecessary formatting change?

vllm/config/cache.py
```diff
@@ -91,7 +91,8 @@ class CacheConfig:
     """ Optional override for mamba page size; used by hybrid mamba/attention
     models to ensure exact alignment with attention page size."""
     mamba_block_size: int | None = None
```
hmellor:
```suggestion
    mamba_block_size: int | None = Field(default=None, gt=0, multiple_of=256)
```

vllm/engine/arg_utils.py
```diff
@@ -1343,6 +1352,14 @@ def create_engine_config(
...
         )
 
+        # Validate mamba_block_size
+        assert self.mamba_block_size is None or self.enable_prefix_caching, (
+            "--mamba-block-size requires --enable-prefix-caching"
+        )
+        assert self.mamba_block_size is None or (
+            self.mamba_block_size > 0 and self.mamba_block_size % 256 == 0
+        ), f"--mamba-block-size must be a multiple of 256, got {self.mamba_block_size}"
+
```
hmellor: Can be replaced with `Field` (see my other suggested change)

```suggestion
```

vllm/engine/arg_utils.py
```diff
@@ -1343,6 +1352,14 @@ def create_engine_config(
             f"dcp_size={self.decode_context_parallel_size}."
         )
 
+        # Validate mamba_block_size
+        assert self.mamba_block_size is None or self.enable_prefix_caching, (
+            "--mamba-block-size requires --enable-prefix-caching"
+        )
```
hmellor: Could this validation instead happen in `VllmConfig` in a `pydantic.model_validator`?

vllm/model_executor/models/config.py
```diff
@@ -415,7 +414,9 @@ def verify_and_update_config(cls, vllm_config: "VllmConfig") -> None:
             def lcm(a, b):
                 return a * b // gcd(a, b)
 
-            base_chunk_size = model_config.get_mamba_chunk_size()
+            base_chunk_size = (
+                user_mamba_block_size or model_config.get_mamba_chunk_size()
```
hmellor:
```suggestion
                cache_config.mamba_block_size
                or model_config.get_mamba_chunk_size()
```

vllm/model_executor/models/config.py
```diff
@@ -344,6 +342,7 @@ def verify_and_update_config(cls, vllm_config: "VllmConfig") -> None:
         if not envs.VLLM_USE_V1:
             return
 
+        user_mamba_block_size = vllm_config.cache_config.mamba_block_size
```
hmellor: This is defined quite far from where it is used and is only used in one place

```suggestion
```

vllm/engine/arg_utils.py
Josephasafg: Fixed

vllm/config/cache.py
Josephasafg: Didn't know `Field` supported `multiple_of`, thats great. Fixed!

vllm/engine/arg_utils.py
Josephasafg: Done

vllm/engine/arg_utils.py
Josephasafg: Done. Moved to `model_validator` in `CacheConfig`

vllm/model_executor/models/config.py
Josephasafg: The reason I store it there is because in the next line we call `MambaModelConfig.verify_and_update_config(vllm_config)` which modifies `mamba_block_size` if it wasn't set. Which then prevents me from doing 
```python
base_chunk_size = (
                cache_config.mamba_block_size 
                or model_config.get_mamba_chunk_size()
            )
```

Because the value was already overridden. @hmellor If you think it'll help, I can move the call a few lines down to line 395 where we check  `if cache_config.enable_prefix_caching` and do these calls before the condition
```python
mamba_block_size = vllm_config.cache_config.mamba_block_size
# Enable FULL_AND_PIECEWISE by default
MambaModelConfig.verify_and_update_config(vllm_config)
```

What do you think?

vllm/model_executor/models/config.py
Josephasafg: @hmellor When I run the pre-commit hook it enforces this 
```python
base_chunk_size = mamba_block_size or model_config.get_mamba_chunk_size()
```

vllm/config/cache.py
```diff
@@ -90,8 +90,9 @@ class CacheConfig:
     mamba_page_size_padded: int | None = None
     """ Optional override for mamba page size; used by hybrid mamba/attention
     models to ensure exact alignment with attention page size."""
-    mamba_block_size: int | None = None
-    """Size of a contiguous cache block in number of tokens for mamba cache."""
+    mamba_block_size: int | None = Field(default=None, gt=0, multiple_of=256)
```
heheda12345: do we have to force it to be multiple of 256? I think this limitation should be verified by each mamba attention backend.

Josephasafg: Fair point. I can add the validation as part of the `__init__` of each mamba attention backend. 

For Mamba1 we could for no allow multiple of 8 to align with the [conv kernel](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/mamba/ops/causal_conv1d.py#L577)

e.g.
```python
if kv_cache_spec.block_size % BLOCK_M != 0:
    raise ValueError(...)..
```
For Mamba2 we could keep the same constraint until @tdoublep relaxes it as his comment mentions.

Is that what you had in mind? @heheda12345

heheda12345: Yes exactly!

hmellor: Shall we change this to `multiple_of=8` as the lowest common denominator? Then in Mamba2 we add the additional constraint you mentioned?

vllm/engine/arg_utils.py
```diff
@@ -528,6 +528,7 @@ class EngineArgs:
     calculate_kv_scales: bool = CacheConfig.calculate_kv_scales
     mamba_cache_dtype: MambaDType = CacheConfig.mamba_cache_dtype
     mamba_ssm_cache_dtype: MambaDType = CacheConfig.mamba_ssm_cache_dtype
+    mamba_block_size: int | None = None
```
hmellor: We try not to duplicate default values so if they are changed in the config, they automatically change here

```suggestion
    mamba_block_size: int | None = get_field(CacheConfig, "mamba_block_size")
```

vllm/model_executor/models/config.py
hmellor: Oh I see, if the value is changed inside `verify_and_update_config` then this is fine. I would just add a comment stating that so that future maintainers don't mistakenly see it as something to clean up

vllm/model_executor/models/config.py
hmellor: The main change here was `user_mamba_block_size` -> `cache_config.mamba_block_size`, but in another thread you've explained why we need this.

vllm/config/cache.py
Josephasafg: @heheda12345 In this PR I’m not adding support for Mamba1, since the `mamba_block_size` engine arg feature is only enabled when APC is on.

For now, I suggest we set `multiple_of=8` as @hmellor suggested, as the general restriction. I think raising this error early is more user-friendly than waiting until the Mamba attention backend is called.

We can keep the Mamba2 constraint where it currently lives in `config.py`, since it’s still tightly coupled to that function until we manage to relax it and reduce dependencies.

Once we merge this PR I could add the restriction for Mamba1 as part of my Mamba1 APC [PR](https://github.com/vllm-project/vllm/pull/26377). 

What do you think?

vllm/engine/arg_utils.py
Josephasafg: Fixed

vllm/model_executor/models/config.py
Josephasafg: Discussed as this value is being stored before being modified in an external call

vllm/config/cache.py
heheda12345: I still think `multiple_of` 8 isn't a general constraint. For UX, I think the only change is people needs to wait for a bit longer to get the error. We are also working on some refactor to make these block_size related checks happen earlier.

Josephasafg: Got it. So for now there is not restriction of the value of the block size. If the user decides to pass it when APC is enabled, we will use it. Otherwise, we fallback to the original path of calculating the block size

heheda12345 (approved):
LGTM! Thank you very much!

tdoublep (approved):
LGTM

PR #27292, opened 2025-10-21
[Core] Align whisper closer to other multimodal models

russellb:
Prior to this change, the encoder inputs were passed into the model,
and the encoder would be executed on the first call to the model.
This made encoder-decoder models work differently from other
multimodal models.

With this change, we now execute the encoder separately, prior to
the normal model call. The encoder outputs are now cached. Caching them
doesn't actually help anything, but if in the future this cache is
used across multiple requests, it could be helpful.

The primary value here is aligning Whisper / encoder-decoder execution
more closely to other multimodal models.

Signed-off-by: Russell Bryant <rbryant@redhat.com>


DarkLight1337 (approved):
Thanks for cleaning this up!

Issue #27298, opened 2025-10-21
[Doc]: Update metrics documentation to remove V0 references and add v1 changes.


atalhens:
## Problem

The metrics documentation in `docs/design/metrics.md` still contains references to V0 metrics implementation, but V0 metrics have been removed after @njhill 's PR https://github.com/vllm-project/vllm/pull/27215 was merged. To avoid confusion, I think we should remove this and update it with the new set of v1 metrics.

Was curious if we want to keep this v0 reference and add the v1 details on top of this. 

### Suggest a potential alternative/fix

1. Remove all V0 references from the metrics documentation.
2. Update the introduction to focus on V1 metrics only.

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

PR #18777, opened 2025-05-27
Export NaNs in logits to scheduler_stats if output is corrupted

vladmihailescu:
Signed-off-by: Vlad Mihailescu vladmihailescu@meta.com

Summary:
Report nan in logits in scheduler_stats. This can be used later exported as Phrometeus counter but for now this is required so we can export it in our internal counter infra.

This counter is used to identify bad hosts or bad GPUs which cause NaNs in logits.

 It's a common metric we expose.

Reviewed By: Adolfo-Karim

Differential Revision: D75423285

```
NO ENV VAR (NaN counting off):
I0618 22:24:27.611000 180881 reporter.py:195] Part 1: High-level performance metrics
Ran 404/404 requests in 56.56s
Success rate:        100.00%
QPS:                 7.14
Avg latency:         11.361s
Avg TTFT (client):   8245.31ms
P50 TTFT (client):   8262.13ms
P99 TTFT (client):   8283.14ms
Avg TTIT (client):   311.60ms
P50 TTIT (client):   323.78ms
P99 TTIT (client):   334.87ms
Avg TTFT (server):   6748.17ms
Avg TTIT (server):   357.83ms
Avg prefill len:     2587.54 tokens
P50 prefill len:     2587.00 tokens
P99 prefill len:     2596.00 tokens
Avg decode len:      10.00 tokens
P50 decode len:      10.00 tokens
P99 decode len:      10.00 tokens
I0618 22:22:13.811000 141492 reporter.py:195] Part 1: High-level performance metrics
Ran 404/404 requests in 56.30s
Success rate:        100.00%
QPS:                 7.18
Avg latency:         11.312s
Avg TTFT (client):   8209.60ms
P50 TTFT (client):   8214.03ms
P99 TTFT (client):   8273.95ms
Avg TTIT (client):   310.22ms
P50 TTIT (client):   322.10ms
P99 TTIT (client):   325.09ms
Avg TTFT (server):   6639.55ms
Avg TTIT (server):   355.36ms
Avg prefill len:     2587.49 tokens
P50 prefill len:     2587.00 tokens
P99 prefill len:     2595.00 tokens
Avg decode len:      10.00 tokens
P50 decode len:      10.00 tokens
P99 decode len:      10.00 tokens
```

```
WITH ENVVAR (NaN counting on):
I0618 22:44:09.103000 508486 reporter.py:195] Part 1: High-level performance metrics
Ran 404/404 requests in 56.18s
Success rate:        100.00%
QPS:                 7.19
[...]

PR #27306, opened 2025-10-21
[Feature]: Add corrupted request metric to V1 metrics system.

### LINKED ISSUE #27301: [Feature]: Add num_corrupted_request metric to V1 metrics system.
### Description

Currently, vLLM internally tracks a corrupted_requests_counter metric whenever a request produces invalid outputs (NaNs) due to model, engine, or hardware issues. However, this metric is not directly exposed to users in logs or Prometheus metrics.

Exposing this metric would allow users to:
- Detect model instability or misbehaving custom models.
- Monitor runtime/engine health in production clusters.
- Quickly identify hardware or distributed inference issues affecting outputs

### Motivation & Problem

While NaN outputs are rare with well-tested models, they become critical for custom models in early development stages or may also arise within the runtime because of Engine/runtime issues. 

- Models may have numerical instability.
- Hardware issues are more likely to surface

The codebase already detects corrupted requests (Request.is_output_corrupted) when the env variable is set `VLLM_COMPUTE_NANS_IN_LOGITS = TRUE`, but this diagnostic information is completely hidden from users, as there are no metrics, no logging, and no monitoring.

### Proposed Idea

Two Approaches for Corrupted Request Metrics that i am thinking of are: 

1. Approach 1: CLI Config-Based (Current Implementation)

 -  `--include-corrupted-requests #CLI flag`
 - Config: SchedulerConfig.include_corrupted_requests
 - Usage: vllm serve model --include-corrupted-requests

Pros: User-friendly, explicit control, follows vLLM patterns
Cons: Adds new CLI argument, requires config changes

I welcome suggestions and thoughts on the same, and would love to contribute the same.

### Alternatives

2. Approach 2: Environment Variable-Based (Proposed Alternative)

- Existing `VLLM_COMPUTE_NANS_IN_LOGITS` environment variable
- Logic: When NaN detection is enabled, automatically expose corrupted metrics
- Usage: VLLM_COMPUTE_NANS_IN_LOGITS=1 vllm serve model

Pros: Reuses existing infrastructure, no new CLI args.
Cons: A coupling of metrics exposure to NaN detection, and less granular control.

Thanks
Snehlata

### Additional context

_No response_

atalhens:
## 
Signed off: sneh.lata@nutanix.com
## Purpose

The purpose of this PR is to fix: https://github.com/vllm-project/vllm/issues/27301 and https://github.com/vllm-project/vllm/issues/27298. This PR implements a new Prometheus metric `vllm:num_requests_corrupted` to track and monitor corrupted requests in vLLM. The feature provides visibility into request corruption issues that can occur during inference, enabling better monitoring and debugging capabilities.

- New Prometheus gauge metric `vllm:num_requests_corrupted`
- Console logging of corrupted request counts per logging interval.
- Enabled via the env variable  `VLLM_COMPUTE_NANS_IN_LOGITS` flag for performance control, thus it won't impact any functionality unless enabled; the default value is set to False.
- Updated documentation and updated dashboard examples config as well.

**Performance Considerations:**
- Corrupted request counting is optional and disabled by default
- When enabled, it adds no computational overhead

## Test Plan
- Not needed as it is a by-default disabled metric.

## Test Result
- N/A

---
<details>

- [] The test plan, such as providing test command.
</details>


docs/design/metrics.md
```diff
@@ -17,45 +17,39 @@ Metrics in vLLM can be categorized as follows:
...
-- `vllm:spec_decode_num_accepted_tokens_total` (Counter)
-- `vllm:spec_decode_num_draft_tokens_total` (Counter)
-- `vllm:spec_decode_num_emitted_tokens_total` (Counter)
+### v1 Metrics
+
+In v1, the following metrics are exposed via a Prometheus-compatible `/metrics` endpoint using the `vllm:` prefix:
+
+- `vllm:num_requests_running` (Gauge) - Number of requests currently running.
+- `vllm:num_requests_waiting` (Gauge) - Number of requests currently waiting.
+- `vllm:num_requests_corrupted` (Gauge) - Number of requests that became corrupted.
```
markmc: Thanks, but I don't think it's super important for every new metric to be captured in this list - the design doc is a snapshot in time of a design discussion IMO

Let's complete the doc update in #27332

vllm/v1/metrics/loggers.py
```diff
@@ -181,6 +181,7 @@ def log(self):
             "Avg generation throughput: %.1f tokens/s",
             "Running: %d reqs",
             "Waiting: %d reqs",
+            "Corrupted: %d reqs",
```
markmc: Let's only print this if `envs.VLLM_COMPUTE_NANS_IN_LOGITS`

vllm/v1/metrics/loggers.py
```diff
@@ -376,6 +378,16 @@ def __init__(
             gauge_scheduler_waiting, engine_indexes, model_name
         )
 
+        gauge_scheduler_corrupted = self._gauge_cls(
```
markmc: Let's only register this if `envs.VLLM_COMPUTE_NANS_IN_LOGITS`

vllm/v1/core/sched/scheduler.py
```diff
@@ -1154,9 +1155,22 @@ def update_draft_token_ids(
...
-    def get_request_counts(self) -> tuple[int, int]:
-        """Returns (num_running_reqs, num_waiting_reqs)."""
-        return len(self.running), len(self.waiting)
+    def _get_corrupted_requests_count(self, include_corrupted: bool) -> int:
+        """Get the count of corrupted requests if enabled, otherwise return 0.
+
+        This method centralizes the corrupted requests counting logic to avoid
+        code duplication and improve performance.
+        """
+        if not include_corrupted:
```
markmc: Let's check `envs.VLLM_COMPUTE_NANS_IN_LOGITS` here

vllm/v1/core/sched/scheduler.py
```diff
@@ -1154,9 +1155,22 @@ def update_draft_token_ids(
...
+            return 0
+        return sum(req.is_output_corrupted for req in self.running)
```
markmc: I really don't like that we're doing this twice per step (AFAICT), but that's the fault of how `_maybe_publish_request_counts()` was implemented. IMHO, it should use the `SchedulerStats` that would have been prepared already

vllm/v1/metrics/loggers.py
```diff
@@ -376,6 +378,16 @@ def __init__(
...
+            name="vllm:num_requests_corrupted",
+            documentation="Number of requests corrupted.",
```
markmc: It's more "Number of currently running requests that are corrupted" ?

vllm/engine/arg_utils.py
```diff
@@ -1034,6 +1035,10 @@ def add_cli_args(parser: FlexibleArgumentParser) -> FlexibleArgumentParser:
         scheduler_group.add_argument(
             "--async-scheduling", **scheduler_kwargs["async_scheduling"]
         )
+        scheduler_group.add_argument(
+            "--include-corrupted-requests",
```
markmc: I don't think we need a command-line argument when we already have the env var

examples/online_serving/prometheus_grafana/grafana.json
```diff
@@ -593,6 +593,22 @@
...
           "useBackend": false
+        },
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${DS_PROMETHEUS}"
+          },
+          "disableTextWrap": false,
+          "editorMode": "builder",
+          "expr": "vllm:num_requests_corrupted{model_name=\"$model_name\"}",
```
markmc: I don't think we should include such a niche, off-by-default, metric in the default dashboard

markmc (approved):
Thanks for all your work on this - it's pretty obvious that #18777 has very limited value without this

And ... I know the way `SchedulerStats.num_corrupted_reqs` is currently implemented as a "count of currently running requests that are corrupted" leads to using a `Gauge` ...

But I really think a `Counter` is a more natural use of a time-series metric for this information. Probably that would mean adding `is_output_corrupted` to `FinishedRequestStats` and removing `num_corrupted_reqs` from `SchedulerStats`?

markmc (changes requested):
(sorry didn't mean to approve, wrong button!)

docs/design/metrics.md
atalhens: Hi Mark, thanks for the suggestions.
Surely I will skip over this change here, as a part of it has already been taken up in this https://github.com/vllm-project/vllm/pull/27332, and we can skip adding `vllm:num_requests_corrupted` in the list of items as a part of this PR.

FMI: 

> Let's complete the doc update in https://github.com/vllm-project/vllm/pull/27332

Should we consider adding this field in this PR, or are we thinking of skipping over it altogether? I do believe maybe not for every other metric, but this one seems important and has been in hindsight for a while due to https://github.com/vllm-project/vllm/pull/18777 not being exposed. Maybe adding it in there would give more visibility to the users.

vllm/engine/arg_utils.py
atalhens: > I don't think we need a command-line argument when we already have the env var

My thought behind adding was to tie it while doing vllm serve only to add ease, but it doesn't make sense, as it sounds better to use the env variable to avoid repetition. I also suggested this approach in the feature request issue. https://github.com/vllm-project/vllm/issues/27301.

Surely would refactor the code. Thanks for the suggestion

vllm/v1/metrics/loggers.py
```diff
@@ -898,21 +916,12 @@ def record(
...
                 scheduler_stats.num_waiting_reqs
             )
-
             if self.show_hidden_metrics:
                 self.gauge_gpu_cache_usage[engine_idx].set(
                     scheduler_stats.kv_cache_usage
                 )
             self.gauge_kv_cache_usage[engine_idx].set(scheduler_stats.kv_cache_usage)
 
-            if self.show_hidden_metrics:
```
markmc, line 943: Removing these must be a mistake? Unrelated to the PR

vllm/v1/metrics/loggers.py
```diff
@@ -103,6 +104,8 @@ def __init__(self, vllm_config: VllmConfig, engine_index: int = 0):
         self.last_generation_throughput: float = 0.0
         self.engine_is_idle = False
         self.aggregated = False
+        # Corrupted requests metric. This should not be reset.
```
markmc: I think it should be reset - we should only show the number of requests that were corrupted in the last logging interval?

markmc (changes requested):

vllm/v1/metrics/loggers.py
atalhens: Ahh yes. Thanks for catching. Will revert

atalhens: Removed the redundant check there.
`if self.show_hidden_metrics:`

markmc: Please just revert any changes that are not required to add the corrupted request metric - there is no need to move the `kv_cache_usage` piece or remove the `show_hidden_metrics` check

vllm/v1/metrics/loggers.py
```diff
@@ -903,6 +920,13 @@ def record(
                 self.gauge_gpu_cache_usage[engine_idx].set(
                     scheduler_stats.kv_cache_usage
                 )
+                self.counter_gpu_prefix_cache_queries[engine_idx].inc(
```
markmc: Not quite reverted yet

vllm/v1/core/sched/scheduler.py
```diff
@@ -1261,7 +1262,6 @@ def make_stats(
             prefix_cache_stats=prefix_cache_stats,
             connector_prefix_cache_stats=connector_prefix_cache_stats,
             spec_decoding_stats=spec_decoding_stats,
-            num_corrupted_reqs=sum(req.is_output_corrupted for req in self.running),
```
markmc, line 1262: The `is_output_corrupted()` helper can be removed

atalhens: IIUC, I have already removed the field altogether as it is not needed any longer. Could you please provide a little more detail in case you are pointing to something else?

markmc: This can be removed, because you have removed the only user of it:

```
class Request:
    @property
    def is_output_corrupted(self) -> bool:
        return self.num_nans_in_logits > 0
```

markmc (approved):
Thanks for your patience!

simon-mo (dismissed):
@markmc reviewed

simon-mo (changes requested):
Can you further justify the need for `VLLM_COMPUTE_NANS_IN_LOGITS` env var and why can't this be enabled by default? I cannot find a place where we were doing CPU <> GPU syncs in counting NaNs here.

atalhens:
> > Can you further justify the need for `VLLM_COMPUTE_NANS_IN_LOGITS` env var and why can't this be enabled by default?
> 
> Looks like it was added in #18777 in response to [this comment](https://github.com/vllm-project/vllm/pull/18777#pullrequestreview-2941138025)
> 
> > I cannot find a place where we were doing CPU <> GPU syncs in counting NaNs here.
> 
> Yeah, looks like the original performance overhead concern wasn't very specific

Thanks for catching @simon-mo.
Correct, I think if we agree. I could remove the flag, which would make more sense.

simon-mo (approved):
Thank you for the benchmarks

PR #27310, opened 2025-10-22
[Frontend][gpt-oss] Allow system message to overwrite model identity

lacora:
## Purpose
The current way of implementation forces the system prompt to go into the developer message and it confuses the usage of different roles. We should have developer into developer message, and allow system message to override the system prompt as supported by signature of harmony_utils.py

## Test Result
Override
            "openai_harmony_message": {
                "author": {
                    "role": "system"
                },
                "content": [
                    {
                        "system_content": {
                            "type": "system_content",
                            "model_identity": "You are a helpful agent reponds in Chinese.",
                            "reasoning_effort": "high",
                            "conversation_start_date": "2025-10-21",
                            "knowledge_cutoff": "2024-06",
                            "channel_config": {
                                "valid_channels": [
                                    "analysis",
                                    "final"
                                ],
                                "channel_required": true
                            }
                        }
                    }
                ]
            }

See traces like this:

"We have a conversation: The system message says: \"You are a helpful agent reponds in Chinese.\" Knowledge cutoff and date. The developer message says: \"Instructions: You are a helpful agent reponds in Japanese.\"\n\nWe need to resolve instruction hierarchy: The system message says \"You are a helpful agent reponds in Chinese.\" The developer message says \"You are a helpful agent reponds in Japanese.\" Which instruction is higher priority? The hierarchy is: System > Developer > User. The system message is higher priority. So we must follow the system message (respond in Chinese). The developer message says respond in Japanese, but that is lower priority. Therefore we must respond in Chinese.\n\nNow answering the question: \"how many Rs are in the word strawberry?\"

---
<details>

</details>

### DIFF
--- a/vllm/entrypoints/harmony_utils.py
+++ b/vllm/entrypoints/harmony_utils.py
@@ -183,14 +183,9 @@ def parse_response_input(
     if "type" not in response_msg or response_msg["type"] == "message":
         role = response_msg["role"]
         content = response_msg["content"]
-        if role == "system":
-            # User is trying to set a system message. Change it to:
-            # <|start|>developer<|message|># Instructions
-            # {instructions}<|end|>
-            role = "developer"
-            text_prefix = "Instructions:\n"
-        else:
-            text_prefix = ""
+        # Add prefix for developer messages.
+        # <|start|>developer<|message|># Instructions {instructions}<|end|>
+        text_prefix = "Instructions:\n" if role == "developer" else ""
         if isinstance(content, str):
             msg = Message.from_role_and_content(role, text_prefix + content)
         else:
--- a/vllm/entrypoints/openai/serving_responses.py
+++ b/vllm/entrypoints/openai/serving_responses.py
@@ -870,9 +870,20 @@ class OpenAIServingResponses(OpenAIServing):
             messages.extend(request.input)  # type: ignore
         return messages
 
+    def _extract_system_message_from_request(self, request) -> str | None:
+        system_msg = None
+        if not isinstance(request.input, str):
+            for response_msg in request.input:
+                if response_msg.get("role") == "system":
+                    system_msg = response_msg.get("content")
+                    break
+        return system_msg
+
     def _construct_harmony_system_input_message(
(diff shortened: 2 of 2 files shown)


vllm/entrypoints/harmony_utils.py
```diff
@@ -183,14 +183,9 @@ def parse_response_input(
...
-            # User is trying to set a system message. Change it to:
-            # <|start|>developer<|message|># Instructions
-            # {instructions}<|end|>
-            role = "developer"
-            text_prefix = "Instructions:\n"
-        else:
-            text_prefix = ""
+        # Add prefix for developer messages.
+        # <|start|>developer<|message|># Instructions {instructions}<|end|>
+        text_prefix = "Instructions:\n" if role == "developer" else ""
```
yeqcharlotte, line 188: wonder what's the role we use when using "instructions" in responses api?

personally i think this make more sense. but can we add a unit test to cover the rendering result? cc: @heheda12345

vllm/entrypoints/openai/serving_responses.py
```diff
@@ -870,9 +870,20 @@ def _construct_input_messages(
             messages.extend(request.input)  # type: ignore
         return messages
 
+    def _extract_system_message_from_request(self, request) -> str | None:
+        system_msg = None
+        if not isinstance(request.input, str):
+            for response_msg in request.input:
+                if response_msg.get("role") == "system":
```
qandrew, line 877: should we add a check where you can't have multiple system inputs?

vllm/entrypoints/openai/serving_responses.py
```diff
@@ -890,6 +901,7 @@ def _construct_harmony_system_input_message(
             and self.tool_server.has_tool("container")
         )
         sys_msg = get_system_message(
+            model_identity=model_identity,
```
qandrew, line 904: we chatted a bit offline; request.instructions is another way to override system prompt. I guess it seems like in get_system_message() if there's both, it'll respect request.instructions over the model identity?

qandrew (commented):
some small questions, thanks for putting this together!

PR #27322, opened 2025-10-22
[Bugfix][plugin] fla crash on plugin

ILikeIneine:
## Purpose
There's some problem while supporting fla on plugin. 
While importing the `fla/ops/utils`, it crashed on [here](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/fla/ops/utils.py#L141).

Since in plugin the `device` might got their own value (here in vllm-metax is `maca`) and `device_torch_lib` still need to be their own. (here in vllm-metax is `torch.cuda`)

So I use is_cuda_alike  and set default value to None on `getattr` to make some compatibilities for handling the corner cases. The semantics should be consistent with the original code.

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/layers/fla/ops/utils.py
+++ b/vllm/model_executor/layers/fla/ops/utils.py
@@ -17,6 +17,7 @@ from typing import Any, Literal
 
 import torch
 
+from vllm.platforms import current_platform
 from vllm.triton_utils import triton
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


NickLucche (commented):
I think this looks fine but I don't have the context on fla. 
Perhaps @youkaichao can take a quick look at it.

mgoin (approved):
Looks simple enough to me. I believe the logic is kept the same for Nvidia and AMD, so nothing else changes for Intel or CPU

Issue #13663, opened 2025-02-21
[New Model]: Google SigLip 2


tranhoangnguyen03:
### The model to consider.

[[google/siglip2-giant-opt-patch16-384](https://huggingface.co/google/siglip2-giant-opt-patch16-384)](https://huggingface.co/collections/google/siglip2-67b5dcef38c175486e240107)

### The closest model vllm already supports.

https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/siglip.py
> """Implementation of SiglipVisionModel intended to be only used
within a vision language model."""

However, SigLip 2 can be very useful for zero-shot image classification and image-text retrieval

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

DarkLight1337:
The model has the same architecture as SigLIP v1. We just need to modify the code a bit so that it can be directly called in vLLM.

lengrongfu:
I can try to solve it, please assign it to me.

/assign

lengrongfu:
@tranhoangnguyen03 
Excuse me, Can vllm 0.7.3 run the `google/siglip` model?

I use this command `vllm serve google/siglip-base-patch16-256` to running , but get follow error, please help me.

![Image](https://github.com/user-attachments/assets/eda2bc8f-bafa-4918-9192-e9a6d5b85243)

lengrongfu:
@DarkLight1337 Which model should I use to test the support Google SigLip2 model function, can you help me? Give me a huggingface model address.

DarkLight1337:
The first link in the OP should point you to the model.

lengrongfu:
we exec this command, but get have error.
`vllm serve google/siglip2-base-patch16-naflex --hf_overrides '{"architectures": ["LlavaForConditionalGeneration"]}'`

INFO 02-25 00:03:42 parallel_state.py:948] rank 0 in world size 1 is assigned as DP rank 0, PP rank 0, TP rank 0
INFO 02-25 00:03:42 model_runner.py:1110] Starting to load model google/siglip2-base-patch16-naflex...
ERROR 02-25 00:03:43 engine.py:400] 'Siglip2Config' object has no attribute 'projector_hidden_act'
ERROR 02-25 00:03:43 engine.py:400] Traceback (most recent call last):
[...]

PR #27324, opened 2025-10-22
[Model] Siglip Embedding Support

piood:
## Purpose

Support SigLIP text and image embedding in the same model, following the same architecture as CLIP embedding support.

* For text inputs, we only apply `token_embedding` when calling `get_input_embeddings`. The rest of the text embedding and the encoder logic are applied when calling `forward` on the model.
* For image inputs, we apply vision embeddings when calling `get_input_embeddings`. Since the model doesn't have a decoder, we directly return the embeddings inside the `forward` method.
* Unlike CLIP, SigLIP uses an encoder-only architecture. To disable prefix_cache for SigLIP, as mentioned in https://github.com/vllm-project/vllm/pull/27324#discussion_r2455086669, use CLS as the default pooling type. 

This PR extends the multimodal embedding capabilities to support SigLIP models, which are widely used for vision-language tasks.

Related to #13663，this pr is for siglip(v1) embedding support, then will continue support siglip2.

## Test Plan

* Added dedicated tests in `tests/models/multimodal/pooling/test_siglip.py`
* Updated model registry to include SigLIP embedding support
* Added examples for both offline inference and online serving
* Verified with local test runs

## Test Result

* All new SigLIP-specific tests pass
* Model registry correctly recognizes SigLIP embedding models
* Both text and image embedding generation work as expected

---

<details>

- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzRs/edit?tab=t.0).
</details>

### DIFF
--- a/docs/models/supported_models.md
+++ b/docs/models/supported_models.md
@@ -803,9 +803,9 @@ The following table lists those that are tested in vLLM.
 | Architecture | Models | Inputs | Example HF Models | [LoRA](../features/lora.md) | [PP](../serving/parallelism_scaling.md) | [V1](gh-issue:8779) |
 |--------------|--------|--------|-------------------|----------------------|---------------------------|---------------------|
 | `CLIPModel` | CLIP | T / I | `openai/clip-vit-base-patch32`, `openai/clip-vit-large-patch14`, etc. | | | ✅︎ |
-| `SiglipModel` | SigLIP | T / I | `google/siglip-base-patch16-224` | | | ✅︎ |
 | `LlavaNextForConditionalGeneration`<sup>C</sup> | LLaVA-NeXT-based | T / I | `royokong/e5-v` | | ✅︎ | ✅︎ |
 | `Phi3VForCausalLM`<sup>C</sup> | Phi-3-Vision-based | T + I | `TIGER-Lab/VLM2Vec-Full` | | ✅︎ | ✅︎ |
+| `SiglipModel` | SigLIP | T / I | `google/siglip-base-patch16-224` | | | ✅︎ |
 | `*ForConditionalGeneration`<sup>C</sup>, `*ForCausalLM`<sup>C</sup>, etc. | Generative models | \* | N/A | \* | \* | \* |
 
 <sup>C</sup> Automatically converted into an embedding model via `--convert embed`. ([details](./pooling_models.md#model-conversion))  
--- a/docs/models/supported_models.md
+++ b/docs/models/supported_models.md
@@ -800,13 +800,13 @@ These models primarily support the [`LLM.embed`](./pooling_models.md#llmembed) A
 
 The following table lists those that are tested in vLLM.
 
-| Architecture | Models | Inputs | Example HF Models | [LoRA](../features/lora.md) | [PP](../serving/parallelism_scaling.md) |
-|--------------|--------|--------|-------------------|----------------------|---------------------------|
-| `CLIPModel` | CLIP | T / I | `openai/clip-vit-base-patch32`, `openai/clip-vit-large-patch14`, etc. | | |
-| `SiglipModel` | Siglip | T / I | `google/siglip-base-patch16-224` | | | ✅︎ |
-| `LlavaNextForConditionalGeneration`<sup>C</sup> | LLaVA-NeXT-based | T / I | `royokong/e5-v` | | ✅︎ |
-| `Phi3VForCausalLM`<sup>C</sup> | Phi-3-Vision-based | T + I | `TIGER-Lab/VLM2Vec-Full` | | ✅︎ |
-| `*ForConditionalGeneration`<sup>C</sup>, `*ForCausalLM`<sup>C</sup>, etc. | Generative models | \* | N/A | \* | \* |
+| Architecture | Models | Inputs | Example HF Models | [LoRA](../features/lora.md) | [PP](../serving/parallelism_scaling.md) | [V1](gh-issue:8779) |
+|--------------|--------|--------|-------------------|----------------------|---------------------------|---------------------|
+| `CLIPModel` | CLIP | T / I | `openai/clip-vit-base-patch32`, `openai/clip-vit-large-patch14`, etc. | | | ✅︎ |
+| `SiglipModel` | SigLIP | T / I | `google/siglip-base-patch16-224` | | | ✅︎ |
+| `LlavaNextForConditionalGeneration`<sup>C</sup> | LLaVA-NeXT-based | T / I | `royokong/e5-v` | | ✅︎ | ✅︎ |
+| `Phi3VForCausalLM`<sup>C</sup> | Phi-3-Vision-based | T + I | `TIGER-Lab/VLM2Vec-Full` | | ✅︎ | ✅︎ |
+| `*ForConditionalGeneration`<sup>C</sup>, `*ForCausalLM`<sup>C</sup>, etc. | Generative models | \* | N/A | \* | \* | \* |
 
 <sup>C</sup> Automatically converted into an embedding model via `--convert embed`. ([details](./pooling_models.md#model-conversion))  
 \* Feature support is the same as that of the original model.
--- a/docs/models/supported_models.md
+++ b/docs/models/supported_models.md
@@ -803,6 +803,7 @@ The following table lists those that are tested in vLLM.
 | Architecture | Models | Inputs | Example HF Models | [LoRA](../features/lora.md) | [PP](../serving/parallelism_scaling.md) |
 |--------------|--------|--------|-------------------|----------------------|---------------------------|
 | `CLIPModel` | CLIP | T / I | `openai/clip-vit-base-patch32`, `openai/clip-vit-large-patch14`, etc. | | |
+| `SiglipModel` | Siglip | T / I | `google/siglip-base-patch16-224` | | | ✅︎ |
 | `LlavaNextForConditionalGeneration`<sup>C</sup> | LLaVA-NeXT-based | T / I | `royokong/e5-v` | | ✅︎ |
 | `Phi3VForCausalLM`<sup>C</sup> | Phi-3-Vision-based | T + I | `TIGER-Lab/VLM2Vec-Full` | | ✅︎ |
 | `*ForConditionalGeneration`<sup>C</sup>, `*ForCausalLM`<sup>C</sup>, etc. | Generative models | \* | N/A | \* | \* |
--- a/examples/offline_inference/vision_language_pooling.py
+++ b/examples/offline_inference/vision_language_pooling.py
@@ -82,30 +82,6 @@ def run_clip(query: Query) -> ModelRequestData:
     )
 
 
-def run_siglip(query: Query) -> ModelRequestData:
-    if query["modality"] == "text":
-        prompt = query["text"]
-        image = None
-    elif query["modality"] == "image":
-        prompt = ""  # For image input, make sure that the prompt text is empty
-        image = query["image"]
-    else:
-        modality = query["modality"]
-        raise ValueError(f"Unsupported query modality: '{modality}'")
-
-    engine_args = EngineArgs(
-        model="google/siglip-base-patch16-224",
-        runner="pooling",
-        limit_mm_per_prompt={"image": 1},
-    )
-
-    return ModelRequestData(
-        engine_args=engine_args,
-        prompt=prompt,
-        image=image,
-    )
-
-
 def run_e5_v(query: Query) -> ModelRequestData:
     llama3_template = "<|start_header_id|>user<|end_header_id|>\n\n{}<|eot_id|><|start_header_id|>assistant<|end_header_id|>\n\n \n"  # noqa: E501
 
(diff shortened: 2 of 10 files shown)


vllm/model_executor/models/siglip.py
```diff
@@ -359,6 +676,8 @@ def __init__(
 
     def forward(self, hidden_state: torch.Tensor) -> torch.Tensor:
         batch_size = hidden_state.shape[0]
+        seq_len = hidden_state.shape[1]
+
```
noooop, line 673: Thanks for your contribution

Getting the shape of a GPU tensor requires synchronizing the CUDA stream. Is there a way to avoid getting the shape?

docs/models/supported_models.md
```diff
@@ -832,6 +832,7 @@ The following table lists those that are tested in vLLM.
 | Architecture | Models | Inputs | Example HF Models | [LoRA](../features/lora.md) | [PP](../serving/parallelism_scaling.md) | [V1](gh-issue:8779) |
 |--------------|--------|--------|-------------------|----------------------|---------------------------|---------------------|
 | `CLIPModel` | CLIP | T / I | `openai/clip-vit-base-patch32`, `openai/clip-vit-large-patch14`, etc. | | | ✅︎ |
+| `SiglipModel` | Siglip | T / I | `google/siglip-base-patch16-224` | | | ✅︎ |
```
DarkLight1337:
```suggestion
| `SiglipModel` | SigLIP | T / I | `google/siglip-base-patch16-224` | | | ✅︎ |
```

vllm/model_executor/models/siglip.py
```diff
@@ -336,6 +582,77 @@ def forward(
...
+        self.config = config
+        embed_dim = config.hidden_size
+
+        self.embeddings = SiglipTextEmbeddings(config)
+
+        self.encoder = SiglipEncoder(
+            config=config,
+            quant_config=quant_config,
+            prefix=f"{prefix}.encoder",
+            attn_cls=MultiHeadAttention,
```
DarkLight1337: Since no causal mask is used anyway, I think we don't need to add `attn_cls` argument

vllm/v1/core/sched/scheduler.py
```diff
@@ -166,10 +166,15 @@ def __init__(
...
+        # For models without KV cache (e.g., pooling/embedding models),
+        enable_caching = (
+            self.cache_config.enable_prefix_caching
+            and len(kv_cache_config.kv_cache_groups) > 0
+        )
         self.kv_cache_manager = KVCacheManager(
             kv_cache_config=kv_cache_config,
             max_model_len=self.max_model_len,
-            enable_caching=self.cache_config.enable_prefix_caching,
+            enable_caching=enable_caching,
```
DarkLight1337: @noooop shouldn't enable prefix caching be disabled for encoder-only models already? Why do we still need this?

noooop (commented):
Thanks for your contribution

vllm/v1/core/sched/scheduler.py
```diff
@@ -166,10 +166,15 @@ def __init__(
                 self.num_lookahead_tokens = self.num_spec_tokens
 
         # Create the KV cache manager.
+        # For models without KV cache (e.g., pooling/embedding models),
+        enable_caching = (
+            self.cache_config.enable_prefix_caching
+            and len(kv_cache_config.kv_cache_groups) > 0
+        )
```
noooop: Currently enabling prefix caching is very confusing.
I will do further research on https://github.com/vllm-project/vllm/pull/26623.

@piood
Can removing len(kv_cache_config.kv_cache_groups) > 0 allow it to run normally?

piood: Image

@noooop @DarkLight1337  As this image show, I find this is because `enable_caching=self.cache_config.enable_prefix_caching`

If pooling_type="LAST" and is_causal is default True, enable_caching will be True, but siglip only use encoder. 

Remove this line len(kv_cache_config.kv_cache_groups) > 0, enable_caching will be True, get this error.

`
(EngineCore_DP0 pid=1148369)     self.scheduler: SchedulerInterface = Scheduler(
(EngineCore_DP0 pid=1148369)   File "/workspace/vllm/vllm/v1/core/sched/scheduler.py", line 174, in __init__
(EngineCore_DP0 pid=1148369)     self.kv_cache_manager = KVCacheManager(
(EngineCore_DP0 pid=1148369)   File "/workspace/vllm/vllm/v1/core/kv_cache_manager.py", line 103, in __init__
(EngineCore_DP0 pid=1148369)     len(
(EngineCore_DP0 pid=1148369) AssertionError: Only one block size is supported for now
`

vllm/model_executor/models/siglip.py
```diff
@@ -4,34 +4,256 @@
...
+        pooler_config = self.ctx.model_config.pooler_config
+        assert pooler_config is not None
+
+        return get_num_selected_vision_tokens(
+            vision_encoder_info.get_num_image_tokens(
+                image_width=image_width,
+                image_height=image_height,
+            ),
+            _get_vision_feature_select_strategy(pooler_config.pooling_type),
+        )
```
piood, line 132: @DarkLight1337 @noooop   SigLIP has `max_position_embeddings` set to **64**, which means `max_model_len` should be **≤ 64**. 

However, `get_num_image_tokens` currently calls `get_num_selected_vision_tokens`, and when I use the **"full"** strategy with SigLIP, the `prompt_len` equals the number of image patch tokens, which exceeds `max_model_len` and causes an error.

**Example:** With a 14×14 grid, we get **196 tokens**, which is much larger than 64.

## Testing Results

I tested a simple fix where `get_num_image_tokens` just returns **1** for SigLIP, and all tests passed successfully. This suggests it could be a viable solution.

## Potential Solutions

I'm wondering about the best approach here:

1. **Return 1 directly:** Should `get_num_image_tokens` return **1** for SigLIP to avoid exceeding the position embedding limit? (This approach has been tested and works)
   
2. **Add a new selection strategy:** Or should we add another vision token selection strategy (not "full") that would work better with SigLIP's constraints while preserving more semantic information, only change `get_num_selected_vision_tokens`.

vllm/model_executor/models/siglip.py
piood: Yes, already remove it.

docs/models/supported_models.md
piood: Fixed.

vllm/model_executor/models/siglip.py
```diff
@@ -368,7 +687,10 @@ def forward(self, hidden_state: torch.Tensor) -> torch.Tensor:
         hidden_state = self.mlp(hidden_state)
         hidden_state += residual
 
-        return hidden_state[:, 0]
+        pooled = hidden_state[:, 0]
+
+        # prepare for pooling
+        return pooled.unsqueeze(1).repeat(1, seq_len, 1)
```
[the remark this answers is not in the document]
piood: Changed `repeat` to `expand` to avoid memory allocation overhead.

vllm/v1/core/sched/scheduler.py
piood: @noooop The root cause is that Siglip (encoder-only model) was incorrectly identified as causal because `getattr(model_config.hf_config, "is_causal", True)` defaults to `True`, and HuggingFace's SiglipConfig doesn't have this field.
**Fix:** I have add `_get_is_causal()` helper function that get the correct value based on architecture. For encoder-only models like Siglip, it returns `False`.
Reverted the scheduler changes. Fixing `is_causal` detection addresses the root cause, so no need to remove the `len(kv_cache_config.kv_cache_groups) > 0` .
please check again!

vllm/model_executor/models/siglip.py
piood: Thanks for the feedback! Based on my understanding, accessing `.shape` on GPU tensors doesn't require CUDA synchronization as the shape information is stored in CPU-side metadata. Synchronization only occurs when accessing actual tensor values (e.g., `.item()`, `.cpu()`).

However, I might be missing something here. If this understanding is incorrect, I'd really appreciate your correction!

piood (commented):
Fixed `get_num_image_tokens` with detailed documentation. All other issues have been addressed. Ready for re-review.

vllm/model_executor/models/siglip.py
DarkLight1337: For CLIP, we use CLS pooling strategy which avoids this issue. Maybe you can do the same for SigLIP

vllm/model_executor/models/siglip.py
noooop: I'm not 100% sure that using .shape will definitely lead to synchronizing the CUDA stream.

(I feel that sometimes when static computation can obtain the shape, it's not needed, but sometimes when dynamic computation is required, it is needed, so I usually try to avoid fetching the shape. I'm not 100% sure this.

vllm/model_executor/models/siglip.py
```diff
@@ -556,3 +917,211 @@ def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
...
+
+# Assume EOS token corresponds to LAST token in text model
+@default_pooling_type("LAST")
+@MULTIMODAL_REGISTRY.register_processor(
+    SiglipMultiModalProcessor,
+    info=SiglipProcessingInfo,
+    dummy_inputs=SiglipDummyInputsBuilder,
+)
+class SiglipEmbeddingModel(nn.Module, SupportsMultiModal, SupportsQuant):
+    is_pooling_model = True
```
noooop, line 967: Even though it seems odd, I believe that @default_pooling_type("CLS") can disable chunked_prefill and prefix_caching.

This way, there's no need to modify vllm/engine/arg_utils.py

I will optimize the chunked_prefill and prefix_caching logic in #26623

(With an increasing number of supported models, this logic is becoming more and more confusing.

vllm/v1/core/sched/scheduler.py
noooop: PTAL https://github.com/vllm-project/vllm/pull/27324#discussion_r2455086669

This is mainly used to disable chunked_prefill and prefix_caching

vllm/model_executor/models/siglip.py
DarkLight1337: Yes, because chunked prefill and prefix caching are only allowed for `LAST` pooling. But solving the problem using this coincidence is a bit hacky.

DarkLight1337: I am fine with solving this later

DarkLight1337: If we use `CLS` pooling, then we need to reverse the text embeddings and use the vision embeddings as-is, instead of the other way round

noooop: I’m planning to add chunked_prefill and prefix_caching properties to model_config to fully resolve this issue and improve readability. However, I need to wait for the “all pooling + chunked_prefill” feature to be implemented first.

vllm/model_executor/models/siglip.py
piood: Thanks for the suggestion! I've updated SigLIP to use CLS pooling strategy now.

DarkLight1337: Make sure that the correct text embeddings are still being returned

vllm/model_executor/models/siglip.py
piood: I agree there are some nuances here. For this PR, I think it's not a critical issue, but I'll definitely look into this more carefully in the future.Thank you.

vllm/model_executor/models/siglip.py
piood: @noooop @DarkLight1337 I have modified SigLIP to use the CLS strategy and reversed the text embeddings. The tests are passing normally. Since I'm using CLS, I've removed the changes to vllm/engine/arg_utils.py.

vllm/model_executor/models/siglip.py
piood: Thanks for the reminder!  I've rechecked the implementation and fixed the text embeddings reversal logic. Verified that the correct embeddings are being returned now.

vllm/model_executor/models/siglip.py
```diff
@@ -4,35 +4,256 @@
...
+
+    type: Literal["pixel_values"]
+    data: Annotated[torch.Tensor, TensorShape("bn", 3, "h", "w")]
+
+
+_POOLING_TYPE_TO_STRATEGY: dict[str, VisionFeatureSelectStrategyStr] = {
+    "MEAN": "full",
+    "ALL": "full",
+    "CLS": "class",
+}
```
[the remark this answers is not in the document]
DarkLight1337: Can you add a link to @noooop 's comment indicating that `LAST` pooling doesn't work?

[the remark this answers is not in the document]
piood: Sure.The reason for this mapping is explained in @noooop's comment here: https://github.com/vllm-project/vllm/pull/27324#discussion_r2455086669

Unlike CLIP, SigLIP uses an encoder-only architecture. Using `LAST` pooling would enable prefix_cache for SigLIP, but this causes errors because the encoder-only architecture doesn't support this caching mechanism. Therefore, we change SigLIP to `CLS` (class token pooling) for disable prefix_cache and ensure proper functionality.

I've also updated the PR description to clarify this behavior.

DarkLight1337 (approved):
Thanks, LGTM then. Let's see if the tests pass. Can you merge in latest main?

examples/offline_inference/vision_language_pooling.py
```diff
@@ -82,6 +82,30 @@ def run_clip(query: Query) -> ModelRequestData:
     )
 
 
+def run_siglip(query: Query) -> ModelRequestData:
```
DarkLight1337: Can you keep this in alphabetical order? Same for the position of the function in the mapping, and the other files (including documentation and model registry)

tests/models/registry.py
```diff
@@ -466,6 +466,7 @@ def check_available_online(
     ),
     # [Multimodal]
     "CLIPModel": _HfExamplesInfo("openai/clip-vit-base-patch32"),
+    "SiglipModel": _HfExamplesInfo("google/siglip-base-patch16-224"),
```
DarkLight1337: Also update here

examples/offline_inference/vision_language_pooling.py
piood: done.

tests/models/registry.py
piood: Already update.

DarkLight1337 (approved):

PR #27338, opened 2025-10-22
Fix pooling adapters for Transformers backend 

hmellor:
Transformers backend:

- Automatically merges weight mappers when inheriting from `Base`
  - Move prepending of `model.` and extraction of `score`/`classifier` to `Base` so that adapters work on headless models 
- Check both possible names for the classifier in sequence classification class
- Fix class resolution so that adapters are used if `convert` should be used

Pooling adapters:

- Use `text_config` instead of `config`:
  - The Gemma 3 MM classification test writes the overrides to `text_config`, so this appears to be the correct location to do this
  - For text models the `get_text_config()` is a no-op and for multimodal models it will fetch the config of the language model
- The `forward` override is removed because it is a no-op and it blocks the `kwargs` used by the Transformers backend
- Use model specific weight loading when online converting to sequence classification model:

    ```diff
    - AutoWeightsLoader(model).load_weights(weights)
    + type(model).__mro__[1].load_weights(model, weights)
    ```
    The MRO part skips `ModelForSequenceClassification.load_weights` to avoid infinite recursion

---

The following test should pass:

```bash
pytest tests/models/language/pooling/test_mm_classifier_conversion.py::test_gemma_multimodal
```


DarkLight1337 (approved):
Unblocked Pooling tests, we can merge this if it passes

vllm/config/model.py
```diff
@@ -788,25 +788,20 @@ def _get_transformers_backend_cls(self) -> str:
         cls += "MoE" if self.get_num_experts() > 1 else ""
         # Check if the architecture we're wrapping has defaults
         runner = None
-        convert = None
+        task = None
         if defaults := try_match_architecture_defaults(self.architectures[0]):
```
noooop, line 797: The code elsewhere can only rely on CI, I cann't see any issues.

https://github.com/vllm-project/vllm/blob/3b7bdf983b5bf76da7a2c580acd5edb1075d7bca/vllm/config/model.py#L578-L590

However, this 'task' is to be deprecated. Is it really okay to turn convert into task?

cc @DarkLight1337

hmellor: Here I am only using `task` as a local variable. I chose `task` because the variable is used specifically when conversion is **not**.

If using the name `task` is confusing, I can change it.

DarkLight1337: I am fine with `task`

hmellor:
Pooling test passes now

Issue #26201, opened 2025-10-03
[Tracking Issue]: Prefix Caching for Hybrid Models


bohnstingl:
### 🚀 The feature, motivation and pitch

This issue is meant to track follow-up work items for the [Mamba2 Automatic Prefix Caching](https://github.com/vllm-project/vllm/pull/25752). Below is a non-exhaustive list of already identified work items:

- [x] Address some comments leftover from review (#26222)
- [ ] Implement policy for freeing mamba blocks to fix performance in throughput benchmarks
- [ ] Relax constraint that mamba block size must be multiple of chunk size
- [x] Give user flexibility to set mamba caching granularity (https://github.com/vllm-project/vllm/pull/27289)
- [ ] Support mamba prefix caching and spec decode
- [ ] Fuse logic for SSM state writing into kernels (https://github.com/vllm-project/vllm/pull/26235)
- [ ] Test TP>1 behaviour
- [ ] Cache meta-data builds across KV cache groups (https://github.com/vllm-project/vllm/pull/22788)
- [ ] Additional cleanup in causal_conv1d kernels (e.g., strip out unused logic)
- [x] Enable prefix caching for Mamba1 (https://github.com/vllm-project/vllm/pull/26377)
- [ ] Enable prefix caching for ShortConv
- [ ] Enable prefix caching for LinearAttention
- [ ] Enable prefix caching for GDN (https://github.com/vllm-project/vllm/pull/26807)

cc @tdoublep @s3woz 

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

77553322AHST:
[BOT](https://twitter.com/i/communities/1814049760510398906)
@bohnstingl 
@Josephasafg 
@simondanielsson 
@markmc 
@pcmoritz 
@77553322AHST
@ZAMAN-COMPANY-COM
@GitHub

simondanielsson:
GDN Prefix caching PR ready for review 🙏 : https://github.com/vllm-project/vllm/pull/26807

tdoublep:
Removing stale label since this is still being actively worked on. I have updated the issue above to try to reflect current status. Please feel free to link your work if I missed anything.

Regarding ongoing discussion around "all" mode vs. "align" mode: 
[...]

PR #27339, opened 2025-10-22
[Hybrid] Added supports_mamba_prefix_caching Protocol

Josephasafg:
## Purpose

This PR refactors the Mamba prefix caching capability check from a hardcoded whitelist to use Protocol-based model, making it easier to add new Mamba models and following the existing architectural patterns (e.g., `IsHybrid`, `IsAttentionFree`, `HasInnerState`).

When a new Mamba based model will add support for prefix caching, they would simply need to add `SupportsMambaPrefixCaching` to their model class.

## Test Result

---
<details>

</details>


hmellor:
Are there any Mamba models which don't support prefix caching?

Josephasafg:
> Are there any Mamba models which don't support prefix caching?

@hmellor Yes. I currently opened a [PR](https://github.com/vllm-project/vllm/pull/26377) that should enable Mamba1 and Jamba models here. If I'm not mistaken, some other mamba models are also still missing support - can be found in this [issue](https://github.com/vllm-project/vllm/issues/26201)

Josephasafg:
> I ask because typically the other protocols define attributes required to support the feature. Take a look at the multimodal protocol or the mixture of experts protocol as examples.

@hmellor In order to make the models support prefix caching, the changes that take place are usually more internal and are kernel related. The model class itself doesn't really hold any attribute that indicates whether or not it supports it. 

I guess this is a more lightweight protocol like `HasNoOps`, `IsAttentionFree`. Should we go a different direction and not use a protocol?

hmellor:
I'm ok with using a protocol as a label. As you pointed out, we have done this before.

I just wanted to make sure the protocol wasn't missing anything.

Once all mamba models support the feature the protocol can be removed, right?

hmellor (approved):
LGTM! Thanks for this improvement

PR #27345, opened 2025-10-22
[CI/Build] improve editable mode setup

dtrifiro:
Improves incremental compilation speed by avoiding writing build files to temporary directories in editable mode installs, avoiding copies.

## Benchmarks

Testing out this build with

```bash
export UV_TORCH_BACKEND=cpu VLLM_TARGET_DEVICE=cpu
uv pip install --no-deps --no-build-isolation --editable .
```

### Without ccache/sccache

- fresh build: `0:47.60`
- cached build (existing `./build` dir with `cmake` artifacts): `0:05.05` (~10x speedup)

### With ccache/sccache

- fresh build, `sccache` cache miss: `0:47.39`
- fresh cpu build, no `./build` dir, sccache cache hit: `0:06.60`
- cached cpu build (existing `./build`), sccache cache hit: `0:04.98`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -212,6 +212,12 @@ class cmake_build_ext(build_ext):
     # Perform cmake configuration for a single extension.
     #
     def configure(self, ext: CMakeExtension) -> None:
+        if self.editable_mode:
+            build_dir = Path.cwd() / "build"
+            build_dir.mkdir(exist_ok=True)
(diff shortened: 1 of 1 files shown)


NickLucche (approved):

LucasWilkinson (commented):
doesn't `--no-build-isolation` mean we use the local `build/` dir and not `tmp/`?

PR #27352, opened 2025-10-22
[cmake]  fix ROCm hip/clr build on platforms without GPUs attached

evil0sheep:
I'm not 100% sure this is the right place to implement this fix, but this works so I thought I would use it as a starting point for discussion.

The problem is that if you build vLLM for ROCm following [these instructions](https://docs.vllm.ai/en/stable/getting_started/installation/gpu.html) when cmake processes hip/clr it drops into [this conditional](https://github.com/ROCm/clr/blob/58b81d67903bde27abf92ed269c744b0cd723d23/hipamd/hip-config-amd.cmake.in#L78) because the build doesnt define `GPU_TARGETS`. This is fine when building on platforms that have AMD GPUs attached because cmake runs `amdgpu-arch` which resolves the targets just fine. However, if building in an environment without attached GPUs, it fails to resolve targets and then when pytorch tries to build kernels for the targets specified in PYTORCH_ROCM_ARCH things explode because hip is built for a different arch then the kernels youre trying to compile and the build fails.

So at some point somewhere the cmake build should propagate `PYTORCH_ROCM_ARCH` to `GPU_TARGETS`. I just put the block next to some other GPU architecture stuff in the main cmake file. But possible this should be in pytorch or something. Let me know what you think.

## Purpose

Fix ROCm build on platforms without GPUs attached (building linux images for worker nodes in chroot)

## Test Plan

AMD build instructions [here](https://docs.vllm.ai/en/stable/getting_started/installation/gpu.html) succeed when run in context where AMD GPYs

## Test Result

Build passes with fix, fails without fix.

---
<details>

- [ x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x ] The test plan, such as providing test command.
- [ x] The test results, such as pasting the results comparison before and after, or e2e results
- [ N/A] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [N/A ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,11 @@ set(VLLM_TARGET_DEVICE "cuda" CACHE STRING "Target device backend for vLLM")
 message(STATUS "Build type: ${CMAKE_BUILD_TYPE}")
 message(STATUS "Target device: ${VLLM_TARGET_DEVICE}")
 
+if(DEFINED ENV{PYTORCH_ROCM_ARCH})
+    set(PYTORCH_ROCM_ARCH_VAL $ENV{PYTORCH_ROCM_ARCH})
+    set(GPU_TARGETS "${PYTORCH_ROCM_ARCH_VAL}")
+endif()
+
 include(${CMAKE_CURRENT_LIST_DIR}/cmake/utils.cmake)
 
 # Suppress potential warnings about unused manually-specified variables


gshtras:
In fact I'm not even sure how you get into the hip-config-amd.cmake.in
I tried removing both amdgpu-arch and rocm_agent_enumerator from my image and the build still succeeds.
Could you provide instructions to reproduce your scenario?

hongxiayang (approved):
I am good with this change as long as it passes the ROCm CI.

Alexei-V-Ivanov-AMD:
>@Alexei-V-Ivanov-AMD aren't our CI images build on nodes without GPUs? How does it work there?
This looks like a good change for me, but I want to understand the background better

@gshtras @evil0sheep @hongxiayang 

aren't our CI images build on nodes without GPUs? **Yes.**

How does it work there? **It works through specifying explicit build argument "ARG_PYTORCH_ROCM_ARCH"**
Please see the example of the building command as it runs on GPU-less server in the AMD-CI:

{A059CADE-45B2-463E-876A-688A3E01FBF3}

evil0sheep:
Hi @Alexei-V-Ivanov-AMD, sorry to drop this, heres what im doing:

- [vLLM Ansible role](https://github.com/evil0sheep/onedotzero/blob/main/ansible/roles/hardware_specific/amd_vllm_build/tasks/main.yml)
- [shell script executed by the ansible role](https://github.com/evil0sheep/onedotzero/blob/main/ansible/roles/hardware_specific/amd_vllm_build/templates/build_vllm.sh.j2)
- [Ansible role to install radeon GPU drivers](https://github.com/evil0sheep/onedotzero/blob/main/ansible/roles/hardware_specific/amd_radeon_drivers/tasks/main.yml)

I am defining `PYTORCH_ROCM_ARCH` (this change doesnt do anything unless its defined) but its not getting propagated to the HIP/clr build for some reason. Maybe when you set that arg in the docker build its getting passed to the HIP/clr cmake through some other mechanism than cmake? Maybe because you build AITER and define `AITER_ROCM_ARCH` [here](https://github.com/vllm-project/vllm/blob/main/docker/Dockerfile.rocm_base#L20C5-L20C20)?

I'm totally open to other solutions here but based on my read of the clr cmake files im not sure how this could be working for you unless either something is setting `GPU_TARGETS` through some other mechanism that im not triggering or if `amdgpu-arch` is returning the architectures you want even with no GPUs somehow. Where do I find the vLLM build logs for your CI images? maybe i can glean some info there

PR #27366, opened 2025-10-22
[Bugfix] Fix MultiConnector stats reconstruction across process boundaries

kouroshHakha:
## Problem

`MultiConnector` is designed to aggregate multiple KV connectors, and it maintains a `MultiKVConnectorStats` object that contains nested `KVConnectorStats` objects from each underlying connector (keyed by connector name).

In distributed settings, when stats are collected from workers and returned to the main process, Python's dataclass serialization converts `KVConnectorStats` objects into plain dictionaries. When these dictionaries are passed to `MultiKVConnectorStats.aggregate()`, the method attempts to call `.aggregate()` on what it expects to be `KVConnectorStats` objects, but are actually plain `dict` objects, resulting in:

```
AttributeError: 'dict' object has no attribute 'aggregate'
```

The root cause is that `MultiConnector` wasn't reconstructing the proper `KVConnectorStats` subclass objects from their serialized dictionary representations after deserialization.

## Solution

Implement `MultiConnector.build_kv_connector_stats()` to properly reconstruct nested `KVConnectorStats` objects:

1. Extract the connector name and serialized stats data from the dictionary
2. Look up the connector class using `KVConnectorFactory.get_connector_class_by_name()` (new public API)
3. Call the connector's `build_kv_connector_stats()` method to reconstruct the proper stats object
4. Handle edge cases: unknown connectors (skip with warning) and connectors without custom stats (skip silently)

This ensures that after deserialization, `MultiKVConnectorStats.data` contains proper `KVConnectorStats` objects that support aggregation operations, not plain dictionaries.


vllm/distributed/kv_transfer/kv_connector/factory.py
```diff
@@ -66,6 +66,22 @@ def create_connector(
...
+
+        Args:
+            connector_name: Name of the registered connector.
+
+        Returns:
+            The connector class if found, None otherwise.
+        """
+        if connector_name not in cls._registry:
+            return None
+        return cls._registry[connector_name]()
```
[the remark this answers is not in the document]
kouroshHakha: this should actually cause failure. factory will actually fail earlier if that is the case.

vllm/distributed/kv_transfer/kv_connector/v1/multi_connector.py
```diff
@@ -324,11 +324,48 @@ def get_required_kvcache_layout(cls, vllm_config: "VllmConfig") -> str | None:
...
+            # Get the connector class to reconstruct its stats
+            connector_cls = KVConnectorFactory.get_connector_class_by_name(
+                connector_name
+            )
+            if connector_cls is None:
+                logger.warning(
+                    "Unknown connector '%s' in stats data, skipping reconstruction",
+                    connector_name,
+                )
+                continue
```
NickLucche: this entails we're dealing with an unregistered connector, do you see other cases where this could happen? 
Because if the connector is indeed unregistered, I'd rather assert or throw an exception

vllm/distributed/kv_transfer/kv_connector/v1/multi_connector.py
```diff
@@ -324,11 +324,48 @@ def get_required_kvcache_layout(cls, vllm_config: "VllmConfig") -> str | None:
...
+                f"Expected a dict with a 'data' field, got {stats_value}"
+            )
+            inner_data = stats_value["data"]
+
+            # Use the connector's build_kv_connector_stats to reconstruct
+            reconstructed_stats = connector_cls.build_kv_connector_stats(
+                data=inner_data
+            )
+            if reconstructed_stats is not None:
+                reconstructed_data[connector_name] = reconstructed_stats
```
NickLucche: nit:
```suggestion
            if reconstructed_stats := connector_cls.build_kv_connector_stats(
                data=inner_data
            ):
                reconstructed_data[connector_name] = reconstructed_stats
```

vllm/distributed/kv_transfer/kv_connector/factory.py
```diff
@@ -66,6 +66,22 @@ def create_connector(
...
+        """Get a registered connector class by name.
+
+        Args:
+            connector_name: Name of the registered connector.
+
+        Returns:
+            The connector class if found, None otherwise.
+        """
+        if connector_name not in cls._registry:
+            return None
```
NickLucche: as above, I think we can raise like in `get_connector_class` if the connector isn't registered, as registration should happen once at startup

NickLucche (commented):
Thanks a lot for catching this @kouroshHakha , nice one!
Left a few comments, but this is already clean

vllm/distributed/kv_transfer/kv_connector/v1/multi_connector.py
kouroshHakha: hm, not really. I wasn't sure about this either. So wanted to play safe. I can make an assertion in this case.

ApostaC (approved):
Looks pretty clean. Thanks!

NickLucche (approved):
Thanks!

PR #25763, opened 2025-09-26
[ROCm][Perf] New design on ROCm AITER MHA backend Implementation

ganyi1996ppo:
## Purpose
The current `AiterFlashAttentionImpl` fetches K/V every run, which creates unnecessary memory pressure and non-trivial latency—especially with long prompts. This PR:
- Removes redundant KV fetches from the attention backend
- Introduces a phase-aware execution (decode, pure prefill, chunk prefill) and reorders inputs to [decode:chunk_prefill:pure_prefill] for token-contiguous memory access.
- Rewrites the “fetch KV” Triton kernel for better occupancy in chunked prefill and similar scenarios.

## Design and implementation

image

Phase-aware path:
- decode
- chunk prefill (cp)
- pure prefill (pp)

Input reordering to [decode:cp:pp] ensures tokens are contiguous in memory, improving kernel locality and occupancy. The reorder occurs in both Scheduler's scheduling phase and ModelRunner's state updating phase. We add this `split_prefill_from_chunk` to the `SchedulerConfig` to control this behavior, which will be turned on if both `VLLM_ROCM_USE_AITER`  and `VLLM_ROCM_USE_AITER_MHA` are set.

Compared with the old one, this solution is more memory efficient and fast, especially on the long prompt scenario. Here is the Performance Measured on Qwen3, Mi308:

Long prompt, short output (`2k prompt, 16 output`): ~4.x throughput improvement.
Short prompt, long output (`128 prompt, 1k output`): ~2.x throughput improvement.
Extremely long prompt (`192k prompt, 2k output`): ~5.x throughput improvement. 

## Test Plan

__acc__  : lm_eval test for accuracy verification 
__perf__ : vllm bench test
### 2k prompt 16 output case:
old impl
image

new impl
image

### 128 prompt 1k output case:
old impl
image

new impl
image

### acc verification
We test this PR on Qwen3-30B-A3B-FP8 on gsm8k with lm_eval, and here is the result:

```
# previous implementation

|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.7437|±  |0.0120|
|     |       |strict-match    |     5|exact_match|↑  |0.8817|±  |0.0089|

# new implementation

|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
[...]

PR #27367, opened 2025-10-22
[Misc] Make reorder batch also separate extends

LucasWilkinson:
Pre-requisite for optimizations that want to seperate a batch into `[decode, extend, prefill]`, e.g. https://github.com/vllm-project/vllm/pull/25763 (future optimizations that may use this include skipping 0-seqlen in MLA when batch contains both extends and prefills; DeepSeek-V3.2-Exp skip up-converting fp8-kv-cache for pure prefills)

### DIFF
--- a/vllm/v1/attention/backends/utils.py
+++ b/vllm/v1/attention/backends/utils.py
@@ -815,24 +815,39 @@ def reorder_batch_to_split_decodes_and_prefills(
     is_prefill = (~is_decode) & (num_computed_tokens_np == num_scheduled_tokens_np)
 
     # Desired order: decode → extend → prefill
-    order_key = np.zeros(is_decode.shape, dtype=np.int32)  # 0 = decode by default
-    order_key[is_extend] = 1
-    order_key[is_prefill] = 2
+    req_regions = np.zeros(is_decode.shape, dtype=np.int32)  # 0 = decode by default
+    req_regions[is_extend] = 1
+    req_regions[is_prefill] = 2
 
-    # get a permutation of the indices that sorts the order_key, basically this means if
-    # we reordered the batch like request[perm] we'd be in the desired order
-    perm = np.argsort(order_key, kind="stable")
-    # old_idx -> new_pos
-    dest = np.empty_like(perm)
-    dest[perm] = np.arange(num_reqs)
+    num_decodes = int(is_decode.sum())
+    num_extends = int(is_extend.sum())
+
+    target_regions = np.zeros(num_reqs, dtype=np.int32)
+    target_regions[num_decodes : num_decodes + num_extends] = 1
+    target_regions[num_decodes + num_extends :] = 2
+
+    needs_swap = req_regions != target_regions
+
+    if not needs_swap.any():
+        return False
+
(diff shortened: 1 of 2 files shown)


heheda12345 (approved):
LGTM! Can you add some unit tests like the order of decode requests are not changed?

benchislett (approved):
LGTM

vllm/v1/attention/backends/utils.py
```diff
@@ -795,51 +795,59 @@ def reorder_batch_to_split_decodes_and_prefills(
...
+    #  likely compute-bound,
+    num_reqs = len(input_batch.req_ids)
+    num_scheduled_tokens = [
+        scheduler_output.num_scheduled_tokens[id] for id in input_batch.req_ids
+    ]
+    num_scheduled_tokens_np = np.array(num_scheduled_tokens)
+    num_computed_tokens_np = input_batch.num_computed_tokens_cpu[:num_reqs]
+
+    is_decode = num_scheduled_tokens_np <= decode_threshold
+    is_extend = (~is_decode) & (num_computed_tokens_np > num_scheduled_tokens_np)
```
ganyi1996ppo, line 814: hi @LucasWilkinson , maybe this is more accurate?
``` python 
is_extend = (~is_decode) & (num_computed_tokens_np > 0)
is_prefill = (~is_decode) & (num_computed_tokens_np == 0)
```

LucasWilkinson: Yes apologies (was thinking of the seqlen); good catch! please feel free to update this in your PR (along side the unit tests) or open a separate PR

amirai21:
Hi @LucasWilkinson 

We’ve been investigating a regression that started in vLLM 0.11.0 when running Jamba under high load.

The symptom is that, occasionally under load, the model gets stuck on a token or short token sequence, producing outputs like:

"Here is a response for for for for for for for..."


and continuing until max_tokens is reached.

The same prompt and sampling parameters (e.g., temperature=0) don’t reproduce this behavior when sent as a standalone request - it only happens sporadically under high concurrency.

While debugging, we tested with a recent CI image:

public.ecr.aws/q9t5s3a7/vllm-ci-postmerge-repo:938772af03ce01590c7e92b0d3fd0a5bdc899d19


which includes this commit. The issue appears significantly reduced in that version.

We were reviewing commits added since the 0.11.0 release, and this PR looks like it could plausibly explain the fix.

Does that make sense to you - could this reordering change have addressed the repetition-loop behavior we were seeing?

Thanks a lot for your time and for the great work! 🙏

heheda12345:
Hi @amirai21 . Thanks for reporting this. Can you help to try the commit before & after this PR to see whether the problem comes from this PR?

amirai21:
Hi @chanh - yes we plan to test the commits of post 0.11.0 release (untill public.ecr.aws/q9t5s3a7/vllm-ci-postmerge-repo:938772af03ce01590c7e92b0d3fd0a5bdc899d19) to find the one that fixes the decode loop bug, we'll update.

sarckk:
@amirai21 do you have a repro for it?

PR #27373, opened 2025-10-22
[Bugfix][ROCm][DeepSeek] Fix for forward_hip in rope for DeepSeek

gshtras:
Starting with #25135 Rope's forward_hip falls back to forward_cuda while assuming that the function modifies the values in place, which is not the case for DeepSeek's deepseek_scaling_rope implementation.
Instead of relying on an in-place modification we should return the values from the forward function implementation.


SageMoore (approved):
It looks like this is an inheritance bug where the subclass overwrites `forward_cuda` and uses an out-of-place implementation? Nice find.

DarkLight1337 (approved):

PR #19240, opened 2025-06-05
[Core] Allow vLLM to stream n tokens at a time

rohingarg-c:
This PR allows vLLM to stream n tokens at a time, improving potential network bandwidth issues. It stores this stream_n param as a field in a new StreamingParams class, created to store other parameters related to how the model streams tokens back.

PR #27376, opened 2025-10-23
[Core] Allow vLLM to stream n tokens at a time

rohin-garg:
This PR introduces support for streaming n tokens at a time in vLLM, which can help alleviate potential network bandwidth issues.

It adds a new StreamingParams class to store the stream_n parameter and other future parameters related to token streaming behavior.

This is an exact copy of [PR #19240](https://github.com/vllm-project/vllm/pull/19240), recreated here because I no longer have access to my previous GitHub account.

### DIFF
--- a/vllm/streaming_params.py
+++ b/vllm/streaming_params.py
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 """Streaming parameters for token streaming during text generation."""
-from typing import Optional, Annotated
+from typing import Annotated
 
 import msgspec
 
--- a/vllm/streaming_params.py
+++ b/vllm/streaming_params.py
@@ -1,0 +1,49 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""Streaming parameters for token streaming during text generation."""
+from typing import Optional, Annotated
+
+import msgspec
+
+
+class StreamValidationError(ValueError):
+    pass
+
+
+class StreamDefaults:
+    STREAM_N_DEFAULT = 1
+
+
+class StreamLimits:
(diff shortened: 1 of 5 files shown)


vllm/v1/engine/output_processor.py
```diff
@@ -34,25 +35,43 @@ class RequestOutputCollector:
...
+            elif isinstance(self.output, (RequestOutput, PoolingRequestOutput)):
+                # This ensures that request outputs with different request
+                # indexes(if n > 1) do not override each other.
+                self.output.add(output, aggregate=self.aggregate)
+
+            if isinstance(self.output, (RequestOutput, PoolingRequestOutput)) and (
+                self.output.finished
+                or self.streaming_params is None
+                or sum(len(o.token_ids) for o in self.output.outputs)
+                >= self.streaming_params.stream_n
```
elvischenv, line 65: do you consider TTFT? I encountered this when hardcoding stream_n to something like 10:
```
ValueError: Initial test run failed - Please make sure benchmark arguments are correctly specified. Error: Never received a valid chunk to calculate TTFT.This response will be marked as failed!
```

vllm/streaming_params.py
```diff
@@ -0,0 +1,55 @@
...
+    """Streaming parameters for token streaming during text generation.
+
+    Attributes:
+        stream_n (int): Number of tokens to stream at a time. Must be an integer >= 1.
+            Defaults to 1.
+    """
+
+    stream_n: Annotated[int, msgspec.Meta(ge=StreamLimits.STREAM_N_MIN)] = (
+        StreamDefaults.STREAM_N_DEFAULT
+    )
```
elvischenv, line 37: how does user to config this?

elvischenv (commented):
I also have a PR to support stream interval(#27869). I just maintain some token counters to implement the functionality. There is already a global token buffer `self.detokenizer.output_token_ids`, we just need to output when the counter has reach the interval value. I think this should have less overhead compared to your merging two request output. What do you think?

PR #27380, opened 2025-10-23
[ROCm][MLA] Enable MLA persistent kernel with fp8 and bf16 support 

ganyi1996ppo:
## Purpose
This PR introduce persistent mla kernel implementation for `AiterMLABackend`, so as the fp8 support.
## Test Plan
Verify the accuracy on gsm8k

## Test Result

```
# bf16 acc result on gsm8k
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9522|±  |0.0059|
|     |       |strict-match    |     5|exact_match|↑  |0.9507|±  |0.0060|
```

```
# fp8 acc result on gsm8k
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.953|±  |0.0058|
|     |       |strict-match    |     5|exact_match|↑  |0.953|±  |0.0058|
```

---
<details>

</details>


vllm/v1/attention/backends/mla/rocm_aiter_mla.py
```diff
@@ -82,13 +97,47 @@ def __init__(
             "AITER MLAonly supports block size 1."
         )
 
+        gpu = torch.cuda.current_device()
+        device_properties = torch.cuda.get_device_properties(gpu)
+        cu_num = device_properties.multi_processor_count
+
```
vllmellm, line 96: there is no need for the lines 100-102 the metadatas can be created using the `aiter.get_mla_metadata_info_v1`
which returns the sizes needed for each metadata tensors with the respective `dtypes`
also the persistent mode supports different head sizes which allows us to remove restriction of num_heads from 16 and 128 by using the same condition in [aiter at this line](https://github.com/ROCm/aiter/blob/288c82f306380c98fc8d4bcc9083bcca7f64b0bf/aiter/mla.py#L275)
so instead of always using persistent mode we can make it conditional to num_heads.
which means we can remove these line:

https://github.com/vllm-project/vllm/blob/d4acf518d09515560e1082a80b8a4d6550e20d9b/vllm/v1/attention/backends/mla/rocm_aiter_mla.py#L228-L232

This way we support both persistent and non persistent mode while we can run deepseek-R1 model on tp4 since persistent mode supports a flexible num_heads

so can use similar logic as following.
```
        self.persistent_mode = False
        num_heads = self.num_heads

        if num_heads == 16 or num_heads in range(32, 513, 16):
            self.persistent_mode = True

        if self.persistent_mode:
            import aiter

            (
                (work_meta_data_size, w_dtype),
                (work_indptr_size, w_indptr_dtype),
                (work_info_set_size, w_info_set_dtype),
                (reduce_indptr_size, r_indptr_dtype),
                (reduce_final_map_size, r_final_map_dtype),
                (reduce_partial_map_size, r_partial_map_dtype),
            ) = aiter.get_mla_metadata_info_v1(
                max_num_reqs,
                1,  # mtp=1 
                self.num_heads,
                vllm_config.model_config.dtype,
                kv_cache_spec.dtype,
                is_sparse=False,
                fast_mode=True,
            )
            self.work_meta_data = torch.empty(
                work_meta_data_size, dtype=w_dtype, device=device
            )
            self.work_indptr = torch.empty(
                work_indptr_size, dtype=w_indptr_dtype, device=device
            )
            self.work_info_set = torch.empty(
                work_info_set_size, dtype=w_info_set_dtype, device=device
            )
            self.reduce_indptr = torch.empty(
                reduce_indptr_size, dtype=r_indptr_dtype, device=device
            )
            self.reduce_final_map = torch.empty(
                reduce_final_map_size, dtype=r_final_map_dtype, device=device
            )
            self.reduce_partial_map = torch.empty(
                reduce_partial_map_size, dtype=r_partial_map_dtype, device=device
            )

```

vllm/v1/attention/backends/mla/rocm_aiter_mla.py
```diff
@@ -139,6 +188,32 @@ def _build_decode(
...
         )
+        kv_indptr = torch.zeros(
+            [query_start_loc_cpu.size(0)], dtype=torch.int32, device="cuda"
+        )
+        torch.cumsum(seq_lens_device, dim=0, out=kv_indptr[1:])
+        query_lens = query_start_loc_cpu[1:] - query_start_loc_cpu[:-1]
+        max_seqlen_qo = torch.max(query_lens).item()
+
+        import aiter
+
```
vllmellm, line 188: as suggested above in the case we would want to use `self.persistent_mode` then we have the code block below. so that we still support non persistent mode.

wuhuikx:
Blocked by this issue https://github.com/ROCm/aiter/issues/1420

vllm/v1/attention/backends/mla/rocm_aiter_mla.py
ganyi1996ppo: Thanks for the suggestion and sorry for the delay. But from what I learned from aiter, persistent kernel is mainly aim to deal the various length request case and non-persistent is mainly for the case with more similar length. So instead of judging by head, maybe we can leave it as an env variable to the user?

hmellor:
Superseded by #36574

PR #27382, opened 2025-10-23
Refactor: Move CUDA graph dispatch logic earlier

yiz-liu:
## Purpose
After carefully reading the code, I found a potential edge case: when `execute_dummy_batch` runs, dummy attention metadata isn't created even if `aclgraph_runtime_mode` is later set to `CUDAGraphMode.FULL`. That's odd, because attention normally requires metadata, otherwise it may raise an error or produce incorrect output.

The only explanation I can think of is that we're skipping metadata creation for dummy batches to save a bit of performance since we don't care about their output. Can anyone elaborate on this? Thanks.

I also propose a potential fix by moving CUDA graph dispatch logic earlier, this ensures metadata is built when replaying a CUDA graph, and the performance impact should be negligible.

## Test Plan
None.

## Test Result
None.

---
<details>

</details>


fhl2000:
Yep, it is a good catch. The flashInfer backend may potentially hang if it is at full cudagraph without preparing attn_metadata. Could you please move the code of attn_metadata building back out of the `with` context, instead, moving the cudagraph dispatching parts above?

yiz-liu:
> Yep, it is a good catch. The flashInfer backend may potentially hang if it is at full cudagraph without preparing attn_metadata. Could you please move the code of attn_metadata building back out of the `with` context, instead, moving the cudagraph dispatching parts above?

OK, will do.

fhl2000 (approved):
Sorry for this late review. I think there is no harm in moving this logic earlier in dummy_run.
cc @LucasWilkinson, it is also closer to your idea in that the padding logic (and cg mode) verified by the cudagraph dispatcher is done before attention metadata building.

fhl2000:
cc @ProExpertProg

ProExpertProg (approved):
LGTM, @LucasWilkinson is this ok with you?

LucasWilkinson (approved):
LGTM; will be fixed by: https://github.com/vllm-project/vllm/pull/28579 but we can take this in the interim

PR #20702, opened 2025-07-09
[Model] New model support for microsoft/Phi-4-mini-flash-reasoning

congcongchen123:
## Essential Elements of an Effective PR Description Checklist
- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.

## Purpose
New Model for https://huggingface.co/microsoft/Phi-4-mini-flash-reasoning

co-author: @aatkinson and @renll

PR #27396, opened 2025-10-23
[Model] Re-support MotifForCausalLM

WyldeCat:
## Purpose

- Re-support MotifForCausalLM, which was removed during the V0 deprecation
  - https://github.com/vllm-project/vllm/pull/25866

### Changes
- Revert PolyNorm layer removal
  - https://github.com/vllm-project/vllm/pull/27110
- Add GroupedDifferentialAttention backend
  - https://arxiv.org/abs/2510.06949
  - A more general version of DifferentialAttention
    - Introduced at https://github.com/vllm-project/vllm/pull/20702
    - Removed at https://github.com/vllm-project/vllm/pull/25351
- Add MotifForCausalLM

New model for :
https://huggingface.co/Motif-Technologies/Motif-2-12.7B-Base
https://huggingface.co/Motif-Technologies/Motif-2.6b-v1.1-LC
https://huggingface.co/Motif-Technologies/Motif-2.6B

co-author : @ca1207 

## Test Result

---
<details>

</details>


vllm/v1/attention/backends/grouped_diff_attn.py
```diff
@@ -0,0 +1,744 @@
+# SPDX-License-Identifier: Apache-2.0
```
jeejeelee, line 1: @DarkLight1337  Who can review this attention?

DarkLight1337: cc @tlrmchlsmth @LucasWilkinson

ca1207: Hi @tlrmchlsmth @LucasWilkinson 
Just a gentle reminder on the above item. Whenever you have a moment, could you please take a look?
I really appreciate your help.

WyldeCat:
@DarkLight1337 
Hi :)
Motif-2 has been uploaded at [Artificial Analysis](https://x.com/ArtificialAnlys/status/1998570291086373081?s=20)￼,
and it would be great to integrate Motif-2 into vLLM.

If the reason the PR review is still pending is the size of the current PR,
would it help if we split it into two parts — (1) integrating Grouped Differential Attention into v1, and (2) integrating Motif-2 into vLLM?
Would this make the review process easier and help move things forward?

If you have any other thoughts or suggestions, please let me know.

Thanks

vllm/attention/layer.py
```diff
@@ -390,14 +394,16 @@ def forward(
...
+            is_output_shape_given = output_shape is not None
             output_shape = output_shape if output_shape is not None else query.shape
             output = torch.empty(output_shape, dtype=output_dtype, device=query.device)
             hidden_size = output_shape[-1]
             # Reshape the query, key, and value tensors.
             # NOTE(woosuk): We do this outside the custom op to minimize the
             # CPU overheads from the non-CUDA-graph regions.
             query = query.view(-1, self.num_heads, self.head_size)
-            output = output.view(-1, self.num_heads, self.head_size)
+            if not is_output_shape_given:
```
LucasWilkinson: nit: `if output_shape is None:`

vllm/v1/attention/backends/grouped_diff_attn.py
```diff
@@ -0,0 +1,744 @@
...
+            max_seqlen_q=max_seqlen_q,
+            seqused_k=seqused_k,
+            max_seqlen_k=max_seqlen_k,
+            softmax_scale=self.scale,
+            causal=True,
+            alibi_slopes=self.alibi_slopes,
+            window_size=self.sliding_window,
+            block_table=block_table,
+            softcap=self.logits_soft_cap,
+            scheduler_metadata=scheduler_metadata,
```
LucasWilkinson: since your num_heads_q may be different (or atleast thats my understanding) then this scheduler_metadata may not match (i.e. `get_scheduler_metadata` called with different params then `flash_attn_varlen_func` is called with); this can lead to subtle bugs. In this case it's probably best to just not use ahead of time scheduling (i.e. remove `scheduler_metadata` from the attention metadata and not supply it here) the performance impacts should be very marginal

LucasWilkinson (commented):
apologies for the delays! Overall the attention part looks pretty good to me, left a couple comments

vllm/v1/attention/backends/grouped_diff_attn.py
WyldeCat: Thanks for the review. As you mentioned, I removed all AOT scheduling logic in this commit:
https://github.com/vllm-project/vllm/pull/27396/commits/49902a156b269b686d24aed06643ff983fd2f4ef.
I also double-checked the behavior, and like you noted, I couldn't find any performance or correctness issues.

vllm/attention/layer.py
WyldeCat: Thanks for the review. I applied your review in this commit:
https://github.com/vllm-project/vllm/pull/27396/commits/be77682fe9387ccfae93ca849bdda0bd3906c421

PR #27410, opened 2025-10-23
[Bugfix] TypeError: 'NoneType' object is not callable

mostrowskix:
In _auto_init_hash_fn function instead of assigning request value directly to hash_fn variable, it goes as an argument to init_none_hash(). This function returns nothig, so later call of hash_fn() raises reported error. This has been observed when using PYTHONHASHSEED env.
In this change I proposed removing init_none_hash wrapping in L58. Chosen request value is now assigned directly to hash_fn variable and works in later calls.

**Steps to reproduce:**

Use command:
```bash
PYTHONHASHSEED=123 pytest -vs vllm/tests/v1/core/test_kv_cache_utils.py -k test_none_hash
```


heheda12345 (approved):

heheda12345:
I think the error is related to this PR. Can you try `pytest -v -s -m 'not cpu_test' v1/core` in your local env to see whether it has any problem?

mostrowskix:
The commit history may look strange but I mostly I had a problem with sync to main branch. They're not any effect of CI crash.

mostrowskix:
I checked the error from CI: _🚨 Error: The command exited with status 5_ It rises when none tests are selected. You run test with 'not cpu_test' flag but line 51 of the script marks all tests as no cpu. In my opinion this behavior is expected but your tests fail. Take a look and merge, please.

@heheda12345 @WoosukKwon @ywang96 @comaniac

PR #27421, opened 2025-10-23
[Test] Batch Invariant: Rename and organize tests

yewentao256:
## Purpose

Refactor the unit test folder to reuse the utils.

Add online test

## Test

Online test example

- `export VLLM_ATTENTION_BACKEND="FLASHINFER_MLA"`
- `export VLLM_BATCH_INVARIANT=1`
- `vllm serve deepseek-ai/DeepSeek-R1 -tp 8  --enable-expert-parallel --port 9256`
- `pytest tests/v1/determinism/test_online_batch_invariance.py`

```bash
(wentao) wentao@dgxB200-09:~/vllm-source$ pytest tests/v1/generation/test_online_batch_invariance.py
============================================ test session starts =============================================
platform linux -- Python 3.12.3, pytest-8.4.1, pluggy-1.6.0
rootdir: /home/wentao/vllm-source
configfile: pyproject.toml
plugins: anyio-4.11.0
collected 1 item                                                                                             

tests/v1/generation/test_online_batch_invariance.py .                                                  [100%]
================================= 1 passed, 2 warnings in 165.79s (0:02:45) ==================================
```


vllm/v1/attention/backends/mla/flashinfer_mla.py
```diff
@@ -127,19 +130,43 @@ def _forward_decode(
...
+                    kv_lora_rank=self.kv_lora_rank,
+                    qk_rope_head_dim=self.qk_rope_head_dim,
+                    block_tables=bt_i,
+                    seq_lens=sl_i,
+                    max_seq_len=attn_metadata.max_seq_len,
+                    bmm1_scale=self.bmm1_scale,
+                    bmm2_scale=self.bmm2_scale,
+                )
+                outs.append(oi)
+            o = torch.cat(outs, dim=0)
```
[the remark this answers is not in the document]
yewentao256: `num = q.shape[0]` shouldn't be equal to 0

vllm/v1/attention/backends/mla/flashinfer_mla.py
```diff
@@ -127,19 +130,43 @@ def _forward_decode(
...
-            kv_lora_rank=self.kv_lora_rank,
-            qk_rope_head_dim=self.qk_rope_head_dim,
-            block_tables=attn_metadata.decode.block_table,
-            seq_lens=attn_metadata.decode.seq_lens,
-            max_seq_len=attn_metadata.max_seq_len,
-            bmm1_scale=self.bmm1_scale,
-            bmm2_scale=self.bmm2_scale,
-        )
+        if vllm_is_batch_invariant():
+            # execute per-request to eliminate batch-shape-dependent kernel paths.
```
bwasti: ah nice.  is flash-infer going to officially support this path too? (like they do with non-MLA)

cc @pavanimajety

yewentao256: I think we can land this PR first, and if it is supported officially later, we can have a following up PR to optimize.
@mgoin CC

tests/v1/generation/conftest.py
```diff
@@ -0,0 +1,18 @@
...
+def enable_batch_invariant_mode():
+    """Automatically enable batch invariant kernel overrides for all tests."""
+    old_value = os.environ.get("VLLM_BATCH_INVARIANT")
+    os.environ["VLLM_BATCH_INVARIANT"] = "1"
+    yield
+    # restore original value after test
+    if old_value is None:
+        os.environ.pop("VLLM_BATCH_INVARIANT", None)
+    else:
+        os.environ["VLLM_BATCH_INVARIANT"] = old_value
```
mgoin: Can we use monkeypatch instead? Also, why is this folder called generation if it is only for batch invariance? We should rename to relate to determinism

vllm/entrypoints/openai/serving_completion.py
```diff
@@ -206,13 +206,27 @@ async def create_completion(
...
+                    data_parallel_rank = None
+                    from vllm.model_executor.layers.batch_invariant import (
+                        vllm_is_batch_invariant,
+                    )
+
+                    if vllm_is_batch_invariant():
+                        config = self.engine_client.vllm_config
+                        dp_size = config.parallel_config.data_parallel_size
+                        if dp_size and dp_size > 1:
+                            data_parallel_rank = 0
```
mgoin: Why is this needed? If each dp rank itself is batch invariant, then it shouldn't matter what rank gets the request

vllm/v1/attention/backends/mla/flashinfer_mla.py
```diff
@@ -127,19 +130,43 @@ def _forward_decode(
...
+            num = q.shape[0]
+            outs = []
+            for i in range(num):
+                qi = q[i : i + 1]
+                bt_i = attn_metadata.decode.block_table[i : i + 1]
+                sl_i = attn_metadata.decode.seq_lens[i : i + 1]
+                oi = trtllm_batch_decode_with_kv_cache_mla(
```
mgoin: I don't have a clear idea for what features we need to make batch invariant, but this naive loop seems like a bad pattern. This is clearly a massive loss of performance so I think we should instead look to work with the kernel itself to make it stable

mgoin (changes requested):
I feel like these patches to hardcode DP rank 0 and executing mla one token at a time are not a good way to achieve batch invariance. This seems like a slippery slope where we could just do hacks like this for all kernels, which I think we clearly don't want

tests/v1/generation/conftest.py
yewentao256: monkeypatch is used in https://github.com/vllm-project/vllm/pull/27478.
After that is landed, we can further more rename to determinism.

vllm/entrypoints/openai/serving_completion.py
yewentao256: Yes—this is necessary. Batch invariance is per DP rank; with DP>1 the same request can land on different ranks between bs=1 and bs=N (eg, 15 prompts to rank 0 and 17 goes to rank1, is totally different with all to 0), and ranks can legitimately differ, causing the bitwise test to fail.

vllm/v1/attention/backends/mla/flashinfer_mla.py
yewentao256: Yes I agree with you, added a TODO and we should fix this in following up PR. (We may need to update upstream Flashinfer so it may take a while)

tests/v1/generation/conftest.py
yewentao256: Done

vllm/entrypoints/openai/serving_completion.py
mgoin: You say this "ranks can legitimately differ" but I don't understand why this would be the case. In DP the layers will be duplicated to be the same size between ranks, thus calling the same kernels

yewentao256: I meant we have batch requests, one dp rank gets 32 as a batch, and one dp gets 15 and another gets 17, it is not the same. The batch invariant requires same logits for the same rank instead of different ranks. We haven't supported "parallism invariant" yet.

tests/v1/determinism/test_online_batch_invariance.py
mgoin, line 1: This seems like it implements a custom server wrapper and client for no clear reason. Can you use the usual RemoteOpenAIServer interface we use in api server tests?
https://github.com/vllm-project/vllm/blob/f08eab2acc17da9e86d20673bd801659ca912749/tests/entrypoints/openai/test_chat_logit_bias_validation.py#L24-L41

mgoin (changes requested):

tests/v1/determinism/test_online_batch_invariance.py
mgoin: Also is there a reason why DeepSeek-R1 is needed for this test? I feel like we could just as well run a deepseek lite or qwen3 fp8 to test similar features on 2xH100 or 2xB200, and actually enable this test in CI

yewentao256: Fixed, thanks! Could you take another look? @mgoin

mgoin (approved):
LGTM, thanks!

PR #27431, opened 2025-10-23
[Bugfix] In LongRoPE, decide short vs long based on max_model_len

### LINKED ISSUE #27414: [Bug]: LongRoPE transition causes random outputs mid-sequence in Phi-3.5
### Your current environment

<details>

I don't currently have access to the machine, but this is reproducible across all vLLM versions from 0.7 to the current 0.11. The most recent I've tested is on 0.11 on an 8XH100 node.

</details>

### 🐛 Describe the bug

While serving  `microsoft/Phi-3.5-mini-instruct`  (or any other model with LongRoPE), concurrent requests end up with random token outputs after the LongRoPE transition point from short factor to long factor, in this case, 4K tokens. When a prompt starts below the ~4K transition boundary and continues generating past it, the model switches from the short to long scaling factors mid-sequence. This causes the previous KV-cache to become invalid for the long-format scales, resulting in random tokens beyond that point, since it was encoded with different RoPE parameters. Synchronous requests aren’t affected because they use consistent scaling without shared cache states. 

The only current workaround is to modify the model’s config.json to either set the long factor scales equal to the short factor or set the short factor equal to the long factor. This removes the changing scales at the transition point, ensuring consistency throughout the generation.

**Related issues:**

- https://github.com/vllm-project/vllm/pull/8254
    
- https://github.com/vllm-project/vllm/issues/14058
    
- https://github.com/huggingface/transformers/pull/33129
    

**Reproduction:**

Reproducible with any Phi-3.5 variant using the following workload designed to cross the 4K boundary:

```bash
vllm serve microsoft/Phi-3.5-mini-instruct

guidellm benchmark \
  --target "http://localhost:8080" \
  --rate-type concurrent \
  --rate 16 \
  --data "prompt_tokens=3900,output_tokens=500" \
  --max-requests 100
```

For the request outputs stored in the benchmark.json, all of them will start producing random tokens after the 4000th token.

**Current Solution/Workaround:**

Download the model and set the long_factor and short_factor scales equal to each other, then rerun:

```python
huggingface-cli download microsoft/Phi-3.5-mini-instruct --local-dir ./phi3.5/

jq '.rope_scaling.long_factor = .rope_scaling.short_factor' config.json > config.tmp && mv config.tmp config.json

vllm serve ./phi3.5

guidellm benchmark \
  --target "http://localhost:8080" \
  --rate-type concurrent \
  --rate 16 \
  --data "prompt_tokens=3900,output_tokens=500" \
  --max-requests 100
```

The request outputs will now be generated without randomness after the 4K token threshold.

MatthewBonanni:
## Purpose
Per suggestion by @LucasWilkinson, fixes #27414 by choosing short vs long RoPE factors at init time based on `max_model_len`

### Functionality
```
vllm serve microsoft/Phi-3.5-mini-instruct

guidellm benchmark \
  --target "http://localhost:8080" \
  --rate-type concurrent \
  --rate 16 \
  --data "prompt_tokens=3900,output_tokens=500" \
  --max-requests 100
```

### Accuracy
```
vllm serve microsoft/Phi-3.5-mini-instruct --max-model-len <len>

lm_eval \
  --model_args base_url=http://localhost:8000/v1/chat/completions,
               model=microsoft/Phi-3.5-mini-instruct,
               num_concurrent=32,
               max_retries=3,
               tokenized_requests=False \
  --tasks gsm8k \
  --apply_chat_template
```

### Functionality
No longer produces random tokens after threshold

### Accuracy
with `max-model-len=4096` (new default):
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.7559|±  |0.0118|
|     |       |strict-match    |     5|exact_match|↑  |0.5679|±  |0.0136|
```
with `max-model-len=4097` (to trigger long rope)
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.6702|±  |0.0129|
|     |       |strict-match    |     5|exact_match|↑  |0.3071|±  |0.0127|
```

---
<details>

</details>

### DIFF
--- a/vllm/config/model.py
+++ b/vllm/config/model.py
@@ -2113,12 +2113,6 @@ def _get_and_verify_max_len(
             if rope_type == "yarn":
                 derived_max_model_len = rope_scaling["original_max_position_embeddings"]
             derived_max_model_len *= scaling_factor
-        elif rope_type == "longrope":
-            # For LongRoPE, we default to the original max to avoid performance
-            # degradation for shorter sequences.
-            derived_max_model_len = getattr(
-                hf_config, "original_max_position_embeddings", derived_max_model_len
-            )
 
     if encoder_config and "max_seq_length" in encoder_config:
         derived_max_model_len = encoder_config["max_seq_length"]
@@ -2126,7 +2120,16 @@ def _get_and_verify_max_len(
     # If the user specified a max length, make sure it is smaller than the
     # derived length from the HF model config.
     if max_model_len is None:
-        max_model_len = int(derived_max_model_len)
+        # For LongRoPE, default to original_max_position_embeddings to avoid
+        # performance degradation for shorter sequences
+        if rope_scaling is not None and rope_scaling["rope_type"] == "longrope":
+            max_model_len = int(
+                getattr(
+                    hf_config, "original_max_position_embeddings", derived_max_model_len
+                )
+            )
+        else:
+            max_model_len = int(derived_max_model_len)
         if current_platform.is_tpu():
             logger.warning(
                 "--max-model-len is not specified, "
--- a/vllm/config/model.py
+++ b/vllm/config/model.py
@@ -2116,7 +2116,9 @@ def _get_and_verify_max_len(
         elif rope_type == "longrope":
             # For LongRoPE, we default to the original max to avoid performance
             # degradation for shorter sequences.
-            derived_max_model_len = rope_scaling["original_max_position_embeddings"]
+            derived_max_model_len = getattr(
+                hf_config, "original_max_position_embeddings", derived_max_model_len
+            )
 
     if encoder_config and "max_seq_length" in encoder_config:
         derived_max_model_len = encoder_config["max_seq_length"]
--- a/vllm/config/model.py
+++ b/vllm/config/model.py
@@ -2113,6 +2113,10 @@ def _get_and_verify_max_len(
             if rope_type == "yarn":
                 derived_max_model_len = rope_scaling["original_max_position_embeddings"]
             derived_max_model_len *= scaling_factor
+        elif rope_type == "longrope":
+            # For LongRoPE, we default to the original max to avoid performance
+            # degradation for shorter sequences.
+            derived_max_model_len = rope_scaling["original_max_position_embeddings"]
 
     if encoder_config and "max_seq_length" in encoder_config:
         derived_max_model_len = encoder_config["max_seq_length"]
--- a/vllm/model_executor/layers/rotary_embedding/phi3_long_rope_scaled_rope.py
+++ b/vllm/model_executor/layers/rotary_embedding/phi3_long_rope_scaled_rope.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+import logging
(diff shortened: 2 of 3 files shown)


vllm/model_executor/layers/rotary_embedding/phi3_long_rope_scaled_rope.py
```diff
@@ -43,6 +45,12 @@ def __init__(
         self.short_factor = short_factor
         self.long_factor = long_factor
 
+        # Force long factors if max_model_len (runtime max length) exceeds
+        # original_max_position_embeddings to prevent KV cache invalidation when
+        # sequences cross this threshold during generation
+        max_model_len = get_current_vllm_config().model_config.max_model_len
+        self.use_long_rope = max_model_len > original_max_position_embeddings
```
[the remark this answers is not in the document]
MatthewBonanni: `max_model_len` (a runtime parameter) is not always the same as `max_position_embeddings` (a hyperparameter of the model)

tlrmchlsmth (approved):

vllm/model_executor/layers/rotary_embedding/phi3_long_rope_scaled_rope.py
```diff
@@ -43,6 +48,22 @@ def __init__(
...
+        if self.use_long_rope:
+            logger.warning(
```
LucasWilkinson: this only logs once right?

vllm/config/model.py
```diff
@@ -2120,7 +2120,16 @@ def _get_and_verify_max_len(
...
     if max_model_len is None:
-        max_model_len = int(derived_max_model_len)
+        # For LongRoPE, default to original_max_position_embeddings to avoid
+        # performance degradation for shorter sequences
+        if rope_scaling is not None and rope_scaling["rope_type"] == "longrope":
+            max_model_len = int(
+                getattr(
+                    hf_config, "original_max_position_embeddings", derived_max_model_len
+                )
+            )
```
hmellor, line 2152: Why does this not belong above where `derived_max_model_len` is calculated?

It could come after the first `rope_scaling` check?

vllm/model_executor/layers/rotary_embedding/phi3_long_rope_scaled_rope.py
MatthewBonanni: changed to warning_once

vllm/config/model.py
MatthewBonanni: We want to override the default `max_model_len` to `original_max_position_embeddings` while still allowing the user to manually specify a `max_model_len` which is larger than this value. I had originally written the code as modifying `derived_max_model_len`, but (as I read it) `derived_max_model_len` is really intended to be `max_position_embeddings`, because of the later check in line 2143:

https://github.com/vllm-project/vllm/pull/27431/files/684e475735f9594a784dd62aa14e0845884872db#diff-998c640befaf137b9af825f29f4e6e47d273caab1fd04093c97df24b18f5c417L2134

If `max_model_len` is manually set to a value exceeding `derived_max_model_len`, it will throw an error. Hopefully I'm understanding your question correctly?

hmellor: Thank you for explaining, I think I understand now!

PR #26199, opened 2025-10-03
Add option to use unbacked, and backed size obl dynamic shapes for more sounds compilation.

laithsakka:
## Purpose
Dynamic shapes guards are dropped unsoundly in vLLM, those can be added during both dynamo and inductor compilations. The proper way to compile a sounds graph is to use unbacked dynamic shapes or ""backed size oblivious 
maybe"" !  (see the note in the diff content). This PR adds a config that allow choosing between backed, unbacked and backed_size_oblivious as explained in the comment in the content of the PR.  

When using unbacked, users might want to to provide invariants about the shapes of the Model, a lambda argument
is added to support_torch_compile decorators where users can provide a lambda that assert on invariants 
on the model inputs shapes. Those are needed to avoid DDE and to be able to trace the model with unbacked.
example:
```

def llama_model_invariants(input_ids,
                           positions,
                           intermediate_tensors=None,
                           inputs_embeds=None):
    """Shape invariants for Llama model compilation"""
    if input_ids is not None:
        torch._check(positions.size()[0] == input_ids.size()[0])
        
@support_torch_compile(shape_invariants=llama_model_invariants)
..
````
see the qwen example in the code also. 

## Test Plan
Added a unit test .
only works for torch. 2.10+

## Perf testing
command
### Qwen/Qwen2-1.5B-Instruct 
will look into some of those perf issues in the future, but the long term vision with pre_compile is that this will be 
a fallback mode. results anyway still better than eager.
```
CUDA_VISIBLE_DEVICES=1 vllm bench throughput --model Qwen/Qwen2-1.5B-Instruct --input-len 512 --output-len 128 --num-prompts 100 --gpu-memory-utilization 0.8
```
backed
```
Throughput: 88.94 requests/s, 56738.83 total tokens/s, 11383.87 output tokens/s
```

backed size oblivious
```
88.78
```

unbacked
```
Throughput: 82.98  51860.18 total tokens/s, 10405.04 output tokens/s
```

eager
```
Throughput: 63.29 requests/s, 40376.23 total tokens/s, 8100.94 output tokens/s
```

PR #25110, opened 2025-09-17
Avoid bytecode hook and simplify TorchCompileWrapperWithCustomDipatch

laithsakka:
Two main changes: 
1) Add an option to not use the bytecode hook! torch2.8 have a way to drop guard without the hack. 
We will switch totally to not use in a different PR after internal validation of perf post landing this. 
Since @zou3519  have concerns about the perf. which i highly doubt that there will be issues. 

2) TorchCompileWrapperWithCustomDipatch has complexities that are not used in the code base.
the current usage is to by pass guard evaluation so I am just introducing a much simpler TorchCompileWithNoGuardsWrapper.

Next PR I will add a debug mode option to keep DS gaurds and fail if get violated.  

#### performance:
How does this effect run-time? 
I ran the following benchmarks:

vllm bench latency \
    --model Qwen/Qwen2-1.5B-Instruct \
    --input-len 128 \
    --output-len 256 \
    --num-iters 50 \
    --dtype float16
    
 after: 
 ```
10% percentile latency: 0.9403901444951771 seconds
25% percentile latency: 0.9434001832487411 seconds
50% percentile latency: 0.9502700129960431 seconds
75% percentile latency: 0.9560497467537061 seconds
90% percentile latency: 0.9588484280975536 seconds
99% percentile latency: 0.9626266451006813 seconds
```

before:
```
10% percentile latency: 0.9399352639040444 seconds
25% percentile latency: 0.9432627985042927 seconds
50% percentile latency: 0.9482277854986023 seconds
75% percentile latency: 0.9578490345011232 seconds
90% percentile latency: 0.9600405636985669 seconds
99% percentile latency: 0.9649963746999856 seconds
```

vllm bench throughput --model Qwen/Qwen2-1.5B-Instruct --input-len 512 --output-len 128 --num-prompts 1000

after

 ```
Throughput: 130.14 requests/s, 83113.04 total tokens/s, 16658.08 output tokens/s
Total num prompt tokens:  510637
Total num output tokens:  128000
```    

before

```
Throughput: 129.54 requests/s, 82726.85 total tokens/s, 16580.68 output tokens/s
Total num prompt tokens:  510637
Total num output tokens:  128000
```

PR #27432, opened 2025-10-23
Add evaluate_guards option to DynamicShapesConfig

laithsakka:
evaluate_guards is A debug mode to detect and fail if Dynamo ever specializes a dynamic shape by
guarding on it. When True, dynamic shape guards are not dropped from Dynamo.
And a failure will be triggered if recompilation ever happens due to that.

Enabling this allow observing the dynamic shapes guards in the tl-parse artifact.

One exception that evaluate_guards wont verify is counter 0/1 specializations. 
Backed dynamic shapes by default will assume sizes >=2. we will not check nor fail 
if that guard fails for 0/1 inputs sizes.


hmellor:
> Note : Only last(top) commit belongs to this PR

Which PR is this built on top of?

laithsakka:
> > Note : Only last(top) commit belongs to this PR
> 
> Which PR is this built on top of?

https://github.com/vllm-project/vllm/pull/26199
and

https://github.com/vllm-project/vllm/pull/25110

hmellor:
Ok, should we wait for those to merge before reviewing this one?

vllm/config/compilation.py
```diff
@@ -208,6 +208,14 @@ class DynamicShapesConfig:
       backed/unbacked.
     """
 
+    evaluate_guards: bool = False
```
zou3519, line 347: I think I want to get into a state where this is always on during vLLM warmup, and then optionally during runtime. This involves making sure the guards are actually correct (maybe this requires unbacked). Not for this PR, but we should do this in a follow-up.

vllm/compilation/wrapper.py
```diff
@@ -41,10 +43,15 @@ def __init__(self):
...
         if isinstance(backend, str) and backend == "inductor":
             options = vllm_config.compilation_config.inductor_compile_config
-
+        self.first_compile = True
         if mode != CompilationMode.STOCK_TORCH_COMPILE:
             # Drop all the guards.
-            options["guard_filter_fn"] = lambda x: [False for _ in x]
+            if vllm_config.compilation_config.dynamic_shapes_config.evaluate_guards:
+                options["guard_filter_fn"] = lambda x: [
+                    entry.guard_type == "SHAPE_ENV" for entry in x
```
zou3519, line 153: this seems brittle..
can we add a test in pytorch for this or something? "if you change how the dynamic shape guards look then you need to change vLLM"

tests/compile/test_dynamic_shapes_compilation.py
```diff
@@ -110,36 +115,136 @@ def test_dynamic_shapes_compilation(monkeypatch, model_name):
...
+                model(input_5)
+                # If we get here, no guard violation occurred
+                # This is a TEST FAILURE - evaluate_guards should have caused a failure
+                pytest.fail(
+                    f"{config_desc}: Expected guard violation did "
+                    f"not occur! evaluate_guards=True should fail "
+                    f"when shape changes from 10 to 5, but the "
+                    f"model ran successfully without error."
+                )
+            except Exception as e:
```
zou3519: is there a more specific exception we can add?

tests/compile/test_dynamic_shapes_compilation.py
```diff
@@ -110,36 +115,136 @@ def test_dynamic_shapes_compilation(monkeypatch, model_name):
...
+                if (
+                    "recompile" in str(e).lower()
+                    and "fail_on_recompile" in str(e).lower()
+                ):
+                    print(f"✅ {config_desc}: Recompile occurred (expected behavior)")
+                    print("   Recompiles are allowed when evaluate_guards=False")
+                else:
+                    print(f"❌ {config_desc}: Unexpected failure")
+                    print(f"   Error: {e}")
+                    raise
```
zou3519: can you add a test that we dropped 0/1 specialization in backed mode?

tests/compile/test_dynamic_shapes_compilation.py
```diff
@@ -110,36 +115,136 @@ def test_dynamic_shapes_compilation(monkeypatch, model_name):
...
+                # Expected failure - guard was violated
+                error_msg = str(e)
+                if "guard" in error_msg.lower() or "recompile" in error_msg.lower():
+                    print(f"✅ {config_desc}: Expected failure due to guard violation")
+                    print(f"   Error (truncated): {error_msg[:150]}")
+                else:
+                    # Unexpected error type
+                    print(f"❌ {config_desc}: Unexpected error type")
+                    print(f"   Error: {e}")
+                    raise
```
zou3519: tests aren't really supposed to print. why not pytest.fail for the failure case?

zou3519 (commented):
seems reasonable, just had testing qs

vllm/compilation/wrapper.py
laithsakka: yeh we shall if we dont have any. I will file issue and assign to me.

tests/compile/test_dynamic_shapes_compilation.py
```diff
@@ -86,3 +96,100 @@ def test_dynamic_shapes_compilation(
...
+        )
+    )
+
+    def test(model_class, input1, input2, is_01_specialization=False):
+        with torch.no_grad(), use_vllm_config(vllm_config):
+            model = model_class(vllm_config=vllm_config).cuda()
+
+            model(input1)
+
+            if evaluate_guards and not is_01_specialization:
```
ProExpertProg: Can we use `with pytest.raises()`?

vllm/compilation/backends.py
```diff
@@ -752,6 +754,27 @@ def __call__(
...
+            # torch.compile will specialize for 0/1 inputs or otherwise guards that
+            # shape is >= 2. This is because it's really hard not to hit a check
+            # against 0/1. When we evaluate shape guards, we exclude checking those
+            # guards (We would fail always otherwise).
+
+            # We avoid that by updating the ranges of backed sizes when the min is
+            # 2 for any, we assume it's 0.
+            for s, r in fake_mode.shape_env.var_to_range.items():
+                if r.lower == 2:
+                    fake_mode.shape_env.var_to_range[s] = ValueRanges(0, r.upper)
```
ProExpertProg, line 747: Can you import `ValueRanges` locally as it's an internal?

vllm/compilation/backends.py
```diff
@@ -752,6 +754,27 @@ def __call__(
             self.split_gm, submod_names_to_compile, self.vllm_config, self
         ).run(*example_inputs)
 
+        from torch._guards import detect_fake_mode
+
+        fake_mode = detect_fake_mode()
+
+        if (
```
ProExpertProg, line 730: Why do we only do this with `evaluate_guards` on?

ProExpertProg: Oh is it because otherwise we drop all guards anyway?

vllm/config/compilation.py
laithsakka: using unbacked basically just avoids the need of this, once we address the perf gp

vllm/compilation/wrapper.py
laithsakka: actually vllm testd would fail I also found a an internal test that checks it         def test_symbool_guards(

vllm/compilation/backends.py
laithsakka: yes

laithsakka: also for backed size oblivious we do no t want to drop 0/1 guards

laithsakka (commented):
rebase and address all comments

tests/compile/test_dynamic_shapes_compilation.py
```diff
@@ -86,3 +97,118 @@ def test_dynamic_shapes_compilation(
...
+                    assert (
+                        "GuardManager check failed" in error_msg
+                        or "Detected recompile when torch.compile stance" in error_msg
+                    ), error_msg
+
+            else:
+                model(input2)
+
+    # test(ModelWithSizeCheck, torch.randn(20, 10).cuda(), torch.randn(5, 10).cuda())
+    # test(ModelWithSizeCheck, torch.randn(5, 10).cuda(), torch.randn(20, 10).cuda())
```
zou3519: why the comments? did you mean to enable this?

tests/compile/test_dynamic_shapes_compilation.py
```diff
@@ -86,3 +97,118 @@ def test_dynamic_shapes_compilation(
...
+                try:
+                    model(input2)
+                    raise RuntimeError("expected guard violation to occur")
+                except RuntimeError as e:
+                    # Expected failure - guard was violated
+                    error_msg = str(e)
+                    assert (
+                        "GuardManager check failed" in error_msg
+                        or "Detected recompile when torch.compile stance" in error_msg
+                    ), error_msg
```
zou3519: nit: use pytest.raises? https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

vllm/compilation/wrapper.py
zou3519: sgtm

vllm/config/compilation.py
```diff
@@ -227,7 +227,18 @@ class DynamicShapesConfig:
...
     """
 
-    # TODO add a debug mode to fail
+    evaluate_guards: bool = False
+    """
+    A debug mode to detect and fail if Dynamo ever specializes a dynamic shape by
+    guarding on it. When True, dynamic shape guards are not dropped from dynamo.
+    And a failure will be triggered if a recompilation ever happens due to that.
+    This mode requires VLLM_USE_BYTECODE_HOOK to be 0.
+    Enabling this allow observing the dynamic shapes guards in the tl-parse
```
zou3519: nit: tlparse is one word?

zou3519 (approved):
LGTM, minor nits, might need a rebase too

tests/compile/test_dynamic_shapes_compilation.py
laithsakka: oh yes lol

vllm/compilation/wrapper.py
```diff
@@ -107,23 +109,53 @@ def __init__(self):
...
+                    # is picked up.
+                    assert ds_type != DynamicShapesType.BACKED, (
+                        "evaluate_guards for backed shapes requires "
+                        "VLLM_USE_AOT_COMPILE=False. "
+                    )
+
+                assert not envs.VLLM_USE_BYTECODE_HOOK, (
+                    "compilation_config.dynamic_shapes_config.evaluate_guards "
+                    "requires VLLM_USE_BYTECODE_HOOK=0. "
+                )
```
ProExpertProg: Duplicate?

vllm/compilation/wrapper.py
```diff
@@ -107,23 +109,53 @@ def __init__(self):
...
-                    "UNBACKED dynamic shapes require VLLM_USE_BYTECODE_HOOK=0. "
-                )
+            # reason is that bytecode does torch._dynamo.eval_frame.
+            # remove_from_cache(self.original_code_object()) to force a new
+            # re-compilation. And if we use
+            # compiled_ptr = self.check_invariants_and_forward
+            # it will reset all entries.
+            assert envs.VLLM_USE_BYTECODE_HOOK, (
+                "UNBACKED dynamic shapes requires VLLM_USE_BYTECODE_HOOK=0. "
+            )
```
ProExpertProg: assert is the inverse of the message

ProExpertProg (commented):
Two minor comments:

docker/Dockerfile
```diff
@@ -188,15 +188,14 @@ ARG nvcc_threads=8
 ENV NVCC_THREADS=$nvcc_threads
 
 ARG USE_SCCACHE
-ARG SCCACHE_DOWNLOAD_URL=https://github.com/mozilla/sccache/releases/download/v0.8.1/sccache-v0.8.1-x86_64-unknown-linux-musl.tar.gz
```
laithsakka, line 191: what is this lol

PR #26319, opened 2025-10-06
[Lora]Load tuned multi-lora kernel configs from json files

li2haipeng:
## Purpose
Add support to load tuned multi-lora shrink/expand triton kernels configs from json files. 
I understand there was a discussion in https://github.com/vllm-project/vllm/pull/13096 that we don't want to save those json files. I was wondering if we can just simply keep the option for users in case they want to tune the kernels themselves. According to our benchmarking, we see 12% to 17% ITL gain on Llama3.3 70B with 16 lora adapters with rank=16:

Screenshot 2025-10-06 at 14 08 46

The PR leverages the exact kernel loading function proposed from https://github.com/vllm-project/vllm/pull/13096.

Edit:
The lora config folder should be passed in by `export VLLM_TUNED_CONFIG_FOLDER=/path/to/configs`. Without it, the kernel would use default configs.

PR #21229, opened 2025-07-19
[Feature][Kernel]FusedMoE LoRA

wcwuwc:
## Essential Elements of an Effective PR Description Checklist
- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
FIX https://github.com/vllm-project/vllm/issues/23610

## Purpose
This PR is an experimental extension to FusedMoE, aiming to support parallel inference with multiple LoRA models.

✅ Completed:

- Verified to work under both TP and torch.compile modes.
- Verified correctness on the DeepSeek-V2-Lite-Chat model.
- Compatible with PackedLoRALayerWeights to load LoRA weights of all experts.
- With test coverage.

❌ Work in progress:

- Kernel generalization: experts more than 256, support for quantization and compatibility with various GPU architectures.

Any suggestions are welcome！
## Test Plan

```shell
vllm serve deepseek-ai/DeepSeek-V2-Lite-Chat --trust-remote-code -tp 2  --enable-lora --max-loras 6 --lora-modules  lora1=wuchen01/DeepSeek-V2-Lite-Chat-All-LoRA --max-lora-rank 64 
``` 

```shell
curl --location 'http://127.0.0.1:8000/v1/chat/completions' \
--header 'Content-Type: application/json' \
--data '{
    "stream":false,
    "messages": [
        {"role": "system", "content": "You are a helpful assistant."},
        {
            "role": "user",
            "content": "who are you?"
        }
    ],
    "model": "lora1"
    }'
```

PR #27435, opened 2025-10-23
Load tuned fused_moe_lora shrink and expand kernel configs separately

yugong333:
## Essential Elements of an Effective PR Description Checklist

## Purpose
1. Loading the kernel configs of `fused_moe_lora` kernel
2. Separate `shrink` and `expand` kernel configs in the `fused_moe_lora` function
3. Adding `num_stages` and `num_warps` parameters in the configs

Note: Based on PR #21229 and #26319

## Test Plan

To benchmark the shrink and expand kernel in gate/up and down in the benchmark_lora.py: 

`python3 benchmarks/kernels/benchmark_lora.py list_bench --arg-pool-size 32 \
    --batch-sizes 1 \
    --dtype torch.bfloat16 \
    --hidden-sizes 2048 \
    --num-loras 16 \
    --lora-ranks 16 \
    --num-loras 4 \
    --op-types fused_moe_lora_gate_up_shrink \
    --seq-lengths 1 \
    --sort-by-lora-id 1 \
    --cuda-graph-nops 32 \
    --top-k-nums 4 \
    --num-experts 128`

## Test Result

Together with #26319 we can improve the OTPS 80% - 90% in GPT-OSS-120B when concurrency is 1 or 2.
 
image


benchmarks/kernels/benchmark_lora.py
```diff
@@ -217,11 +282,15 @@ def mkn(
...
-            assert self.is_expand_fn()
+        elif self.is_expand_fn():
             m = num_tokens
             k = lora_rank
             n = hidden_size
+        else:
+            assert self.is_fused_moe_lora_fn()
+            m = num_tokens
+            n = hidden_size
+            k = lora_rank
```
varun-sundar-rabindranath: nit : looks like we could just do,
```
else:
    assert self.is_expand_fn() or self.is_fused_moe_lora_fn()
    ... 
``` 
?

benchmarks/kernels/benchmark_lora.py
```diff
@@ -432,17 +554,33 @@ def make(
             prompt_lora_indices_tensor,
         )
 
-    def sanity_check(self) -> None:
+    def sanity_check(self, ctx: BenchmarkContext, op_type: OpType) -> None:
         """
         Fails asserts when non-conformality is detected.
         """
-        num_tokens = self.input.shape[-2]
+        ##TODO test if this works
```
varun-sundar-rabindranath: is this still a todo ?

benchmarks/kernels/benchmark_lora.py
```diff
@@ -471,21 +609,114 @@ def to_device(tensor: torch.Tensor):
...
         Return num_seqs, num_tokens and max_seq_len
         """
         num_seqs = self.seq_lens.shape[0]
-        num_tokens = self.lora_kernel_meta.token_lora_mapping.shape[0]
+        ## TODO: test if this works
+        num_tokens = (
+            self.lora_kernel_meta.token_lora_mapping.shape[0] * ctx.top_k_num
+            if op_type in [OpType.FUSED_MOE_LORA_DOWN_SHRINK]
+            else self.lora_kernel_meta.token_lora_mapping.shape[0]
+        )
```
varun-sundar-rabindranath, line 613: nit: given this segment repeats a couple of times, maybe it makes sense to introduce a, 
```
def num_tokens(self, op_type):
   ... 
```
function in BechmarkTensors class ?

benchmarks/kernels/benchmark_lora.py
```diff
@@ -258,13 +351,33 @@ def matmul_shapes(
...
+        if (
+            self == OpType.FUSED_MOE_LORA_GATE_UP_SHRINK
+            or self == OpType.FUSED_MOE_LORA_DOWN_SHRINK
+        ):
+            return fused_moe_lora_shrink
+        if (
+            self == OpType.FUSED_MOE_LORA_GATE_UP_EXPAND
+            or self == OpType.FUSED_MOE_LORA_DOWN_EXPAND
+        ):
+            return fused_moe_lora_expand
```
varun-sundar-rabindranath, line 380: nit : `self in [...]` ?

benchmarks/kernels/benchmark_lora.py
```diff
@@ -852,6 +1250,8 @@ def as_benchmark_contexts(
                 sort_by_lora_id=sort_by_lora_id,
                 dtype=args.dtype,
                 # To be filled based on the OpType to benchmark
+                top_k_num=top_k_num,
+                num_experts=num_experts,
                 num_slices=None,
```
varun-sundar-rabindranath, line 1255: can you move `# To be filled based on the OpType to benchmark` to be above `num_slices` please.

vllm/lora/ops/triton_ops/fused_moe_lora_op.py
```diff
@@ -169,90 +169,52 @@ def _fused_moe_lora_kernel(
...
+    num_tokens: int,
+    num_experts: int,
+    num_slices: int,
+    shrink_block_size_m: int,
+    shrink_block_size_n: int,
+    shrink_block_size_k: int,
+    shrink_group_size_m: int,
+    shrink_num_warps: int,
+    shrink_num_stages: int,
+    shrink_split_k: int,
```
varun-sundar-rabindranath: maybe better to lose the `shrink_` prefix here ? since it is the shrink kernel.

vllm/lora/ops/triton_ops/fused_moe_lora_op.py
```diff
@@ -288,19 +250,68 @@ def _fused_moe_lora(
...
+    num_slices: int,
+    max_lora_rank: int,
+    w1_output_dim_size: int,
+    expand_block_size_m: int,
+    expand_block_size_n: int,
+    expand_block_size_k: int,
+    expand_group_size_m: int,
+    expand_num_warps: int,
+    expand_num_stages: int,
+    expand_split_k: int,
```
varun-sundar-rabindranath: similarly, better to lose the `expand` prefix ?

vllm/lora/ops/triton_ops/fused_moe_lora_op.py
```diff
@@ -352,10 +487,80 @@ def _fused_moe_lora_fake(
...
+    num_tokens: int,
+    num_experts: int,
+    num_slices: int,
+    shrink_block_size_m: int,
+    shrink_block_size_n: int,
+    shrink_block_size_k: int,
+    shrink_group_size_m: int,
+    shrink_num_warps: int,
+    shrink_num_stages: int,
+    shrink_split_k: int,
```
varun-sundar-rabindranath: same here about the `shrink_` prefix ?

vllm/lora/ops/triton_ops/fused_moe_lora_op.py
```diff
@@ -352,10 +487,80 @@ def _fused_moe_lora_fake(
...
+    num_slices: int,
+    max_lora_rank: int,
+    w1_output_dim_size: int,
+    expand_block_size_m: int,
+    expand_block_size_n: int,
+    expand_block_size_k: int,
+    expand_group_size_m: int,
+    expand_num_warps: int,
+    expand_num_stages: int,
+    expand_split_k: int,
```
varun-sundar-rabindranath: same here about the `expand` prefix

vllm/lora/ops/triton_ops/utils.py
```diff
@@ -154,13 +154,13 @@ def load_lora_op_config(op_type: str, add_inputs: bool | None) -> dict | None:
...
-            config_fname = f"{gpu_name}_{op_type.upper()}.json"
-        else:
-            assert op_type == "expand"
+        # only expand op needs to consider add_inputs
+        if op_type == "expand":
             config_fname = (
                 f"{gpu_name}_{op_type.upper()}_{str(add_inputs).upper()}.json"
             )
+        else:
+            config_fname = f"{gpu_name}_{op_type.upper()}.json"
```
varun-sundar-rabindranath, line 163: looks like this change is a no-op ?

vllm/lora/ops/triton_ops/utils.py
```diff
@@ -202,6 +211,22 @@ def get_lora_op_configs(
...
+    ]:
+        default = {
+            "BLOCK_SIZE_M": 64,
+            "BLOCK_SIZE_N": 64,
+            "BLOCK_SIZE_K": 32,
+            "NUM_WARPS": 4,
+            "NUM_STAGES": 3,
+            "GROUP_SIZE_M": 8,
+            "SPLIT_K": 1,
+        }
```
varun-sundar-rabindranath, line 230: can we stick to the existing format of using small case keys ?  This will help avoid unintentional `key` errors. Thanks.

vllm/lora/ops/triton_ops/fused_moe_lora_op.py
dcmaddix: +1

vllm/lora/ops/triton_ops/README_TUNING.md
```diff
@@ -44,8 +44,16 @@ For `shrink`, the config file is named as `{gpu_name}_SHRINK.json`, e.g. `NVIDIA
...
+For `fused_moe_lora_down_shrink`, the config file is named as `{gpu_name}_FUSED_MOE_LORA_DOWN_SHRINK.json`, e.g. `NVIDIA_H200_FUSED_MOE_LORA_DOWN_SHRINK.json`.
+
+For `fused_moe_lora_down_expand`, the config file is named as `{gpu_name}_FUSED_MOE_LORA_DOWN_EXPAND.json`, e.g. `NVIDIA_H200_FUSED_MOE_LORA_DOWN_EXPAND.json`.
+
 The `gpu_name` can be automatically detected by calling `torch.cuda.get_device_name()`
 
 ### Json Structure
 
-Optimal kernel configuration files are saved as JSON files with the structure `config_data[max_loras][num_slices][m][k][n]`
+Optimal kernel configuration files are saved as JSON files with the structure `config_data[max_loras][num_slices][m][k][n][n2]`
```
varun-sundar-rabindranath: For this, can you add a separate line for fused-moe-lora and explain what `n2` refers to. Thanks.

vllm/lora/ops/triton_ops/utils.py
```diff
@@ -186,8 +186,17 @@ def get_lora_op_configs(
     rank: int,
     num_slices: int,
     add_inputs: bool | None = None,
+    hidden_size_2: int | None = None,
```
varun-sundar-rabindranath: better to rename `hidden_size_2` to `moe_intermediate_size` or something else ? - I feel `hidden_size_2` lacks context.

vllm/lora/ops/triton_ops/utils.py
yugong333: Thanks for the review! The default config name from `functools.partial` in `fused_moe.py` uses the upper case format. I try to avoid the format conflict with the existing format if users want to fall back to the default config. 
Do you think which one is better:
1. Stick to lowercase format here, then we can solve the conflicts in `fused_moe.py`
2. We just keep all config names for fused_moe_lora uppercase format
3. Setting the default config here instead of using `functools.partial` then we can keep all config names lowercase

Thank you very much!

varun-sundar-rabindranath: Hi @yugong333  . I'd prefer option 1. I am thinking that, if we use lower case here, then folks who are concerned with LoRA will atleast have some consistency. Also, It'd be better to resolve inconsistencies directly in `vllm/lora/layers/fused_moe.py` so this is documented.
About 3, I think importing `try_get_optimal_moe_config` here will be a bit messy. Thanks. 

cc @jeejeelee

vllm/lora/ops/triton_ops/README_TUNING.md
```diff
@@ -44,8 +44,17 @@ For `shrink`, the config file is named as `{gpu_name}_SHRINK.json`, e.g. `NVIDIA
 
 For `expand`, the config fileis named as `{gpu_name}_EXPAND_{add_input}.json`, e.g. `NVIDIA_H200_EXPAND_TRUE.json`.
 
+For `fused_moe_lora_gate_up_shrink`, the config file is named as `{gpu_name}_FUSED_MOE_LORA_GATE_UP_SHRINK.json`, e.g. `NVIDIA_H200_FUSED_MOE_LORA_GATE_UP_SHRINK.json`.
```
jeejeelee: Suggest using `w13` to keep consistent with fusd moe

```suggestion
For `fused_moe_lora_w13_shrink`, the config file is named as `{gpu_name}_FUSED_MOE_LORA_W13_SHRINK.json`, e.g. `NVIDIA_H200_FUSED_MOE_LORA_W13_SHRINK.json`.
```

vllm/lora/ops/triton_ops/README_TUNING.md
```diff
@@ -44,8 +44,17 @@ For `shrink`, the config file is named as `{gpu_name}_SHRINK.json`, e.g. `NVIDIA
...
+
+For `fused_moe_lora_gate_up_expand`, the config file is named as `{gpu_name}_FUSED_MOE_LORA_GATE_UP_EXPAND.json`, e.g. `NVIDIA_H200_FUSED_MOE_LORA_GATE_UP_EXPAND.json`.
```
jeejeelee: ditto

vllm/lora/ops/triton_ops/README_TUNING.md
```diff
@@ -44,8 +44,17 @@ For `shrink`, the config file is named as `{gpu_name}_SHRINK.json`, e.g. `NVIDIA
...
+
+For `fused_moe_lora_down_shrink`, the config file is named as `{gpu_name}_FUSED_MOE_LORA_DOWN_SHRINK.json`, e.g. `NVIDIA_H200_FUSED_MOE_LORA_DOWN_SHRINK.json`.
```
jeejeelee: DITTO
```suggestion
For `fused_moe_lora_w2_shrink`, the config file is named as `{gpu_name}_FUSED_MOE_LORA_w2_SHRINK.json`, e.g. `NVIDIA_H200_FUSED_MOE_LORA_w2_SHRINK.json`.
```

vllm/lora/ops/triton_ops/README_TUNING.md
```diff
@@ -44,8 +44,17 @@ For `shrink`, the config file is named as `{gpu_name}_SHRINK.json`, e.g. `NVIDIA
...
+
+For `fused_moe_lora_down_expand`, the config file is named as `{gpu_name}_FUSED_MOE_LORA_DOWN_EXPAND.json`, e.g. `NVIDIA_H200_FUSED_MOE_LORA_DOWN_EXPAND.json`.
```
jeejeelee: DITTO

jeejeelee (commented):
Some init NIT comments, mainly suggesting that `W13/W2` naming should be used

vllm/lora/ops/triton_ops/utils.py
yugong333: This line determines the config file name based on the op_type. `fused_moe_lora` kernel will use this line.

vllm/lora/layers/fused_moe.py
```diff
@@ -39,6 +40,50 @@ def __init__(self, base_layer: FusedMoE) -> None:
...
+                layer.w13_weight.size(),
+                layer.w2_weight.size(),
+                top_k,
+                config_dtype,
+                block_shape=layer.quant_method.moe_quant_config.block_shape,
+            )
+            shrink_config = get_config_func(M)
+            expand_config = get_config_func(M)
+
+        return shrink_config, expand_config
```
varun-sundar-rabindranath, line 99: I think it is better to lower-case all keys of the configs here so we dont have to do the 
`config.get("UPPER_CASE", None) or config.get("lower_case")` check everywhere ?

vllm/lora/ops/triton_ops/utils.py
varun-sundar-rabindranath: Yes. But, on `main` we have, 
```
        if op_type == "shrink":
            config_fname = f"{gpu_name}_{op_type.upper()}.json"
        else:
            assert op_type == "expand"
            config_fname = (
                f"{gpu_name}_{op_type.upper()}_{str(add_inputs).upper()}.json"
            )
```
it looks like this block of change does the same, except that the `if` and `else` part are switched. Is that right ? am  I missing something ?

vllm/lora/punica_wrapper/punica_gpu.py
```diff
@@ -388,10 +389,19 @@ def add_lora_fused_moe(
...
+            shrink_config.get("NUM_WARPS") or shrink_config.get("num_warps", 4),
+            shrink_config.get("NUM_STAGES") or shrink_config.get("num_stages", 3),
+            shrink_config.get("SPLIT_K") or shrink_config.get("split_k", 1),
+            expand_config.get("BLOCK_SIZE_M") or expand_config.get("block_m", 64),
+            expand_config.get("BLOCK_SIZE_N") or expand_config.get("block_n", 64),
+            expand_config.get("BLOCK_SIZE_K") or expand_config.get("block_k", 32),
+            expand_config.get("GROUP_SIZE_M") or expand_config.get("group_m", 8),
+            expand_config.get("NUM_WARPS") or expand_config.get("num_warps", 4),
+            expand_config.get("NUM_STAGES") or expand_config.get("num_stages", 3),
+            expand_config.get("SPLIT_K") or expand_config.get("split_k", 1),
```
varun-sundar-rabindranath: maybe add a small utility inside `add_lora_fused_moe` that does this check ? something like, 
```
def get_value(config_kv: dict[str, int], k: str, default: int) -> int:
    return config_kv.get(k.upper()) or config_kv.get(k.lower(), default) 
```
? I think that'll clean it up a bit. Thanks.

varun-sundar-rabindranath (approved):
LGTM ! Thank you for working on this @yugong333 . I left some minor nit that'd be nice to have before landing this PR.

vllm/lora/layers/fused_moe.py
yugong333: Thank you @varun-sundar-rabindranath  I add an function to normalize the key name into upper case to be consistent with the moe layer config name, which also is able to remove the name check everywhere.

vllm/lora/ops/triton_ops/utils.py
yugong333: Yes I switch the if and else as only `expand` has `add_inputs` check and others do not.

jeejeelee (approved):

PR #27468, opened 2025-10-24
[Kernel] Enable moe LoRA kernel support FP16

jeejeelee:
## Purpose

cc @varun-sundar-rabindranath  @wcwuwc 

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/lora/ops/triton_ops/fused_moe_lora_op.py
+++ b/vllm/lora/ops/triton_ops/fused_moe_lora_op.py
@@ -99,7 +99,7 @@ def _fused_moe_lora_kernel(
 
     # get a_ptr,b_ptr,c_ptr
     cur_a_ptr = a_ptr + (slice_id % num_slice_a) * slice_a_size
-    cur_b_ptr = tl.load(b_ptr + slice_id).to(tl.pointer_type(tl.bfloat16))
+    cur_b_ptr = tl.load(b_ptr + slice_id).to(tl.pointer_type(c_ptr.dtype.element_ty))
     cur_c_ptr = c_ptr + (slice_id % num_slice_c) * slice_c_size
 
     offs_bn = (pid_n * BLOCK_SIZE_N + tl.arange(0, BLOCK_SIZE_N).to(tl.int64)) % N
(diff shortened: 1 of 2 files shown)


vllm/lora/ops/triton_ops/fused_moe_lora_op.py
```diff
@@ -210,9 +213,9 @@ def _fused_moe_lora(
     num_tokens = M * top_k_num
     w1_output_dim_size = w1_lora_b_stacked.shape[2]
 
-    lora_intermediate_cache1 = torch.zeros(
+    lora_intermediate_cache1 = torch.empty(
```
mgoin, line 216: Why not using zeros anymore?

jeejeelee: For the current kernel, using empty alone is sufficient. As I understand it, zero requires two operations (empty + fill) to complete. Using empty directly can reduce the overhead of fill

vllm/lora/ops/triton_ops/fused_moe_lora_op.py
```diff
@@ -2,9 +2,8 @@
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 
 import torch
-import triton
-import triton.language as tl
```
mgoin, line 6: Strange that the precommit hasn't caught this before

mgoin (approved):
LGTM! Do you have an end to end eval to verify there aren't anymore kernels that need to be accounted for?

vllm/lora/ops/triton_ops/fused_moe_lora_op.py
jeejeelee: @MengqingCao May I ask if you know the reason for this?

MengqingCao: I just looked into this, and found that we only check the triton import at the first commit that introduced `import triton` in. And this mainly caused by `git diff`, the way we check triton import, maybe we should change the way checking triton import or always git diff all the commits based on vllm main

vllm/lora/ops/triton_ops/fused_moe_lora_op.py
```diff
@@ -269,10 +273,6 @@ def _fused_moe_lora(
     K = max_lora_rank
     N = w1_output_dim_size
 
-    # a_intermediate_cache1 = a_intermediate_cache1.view(
-    #     M, -1, a_intermediate_cache1.shape[3]
-    # )
-
     a_intermediate_cache1 = a_intermediate_cache1.view(
         -1, a_intermediate_cache1.shape[3]
     )
```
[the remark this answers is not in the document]
dcmaddix: @jeejeelee I think we still need to fill the values with zeros inside the kernel thanks!

PR #26374, opened 2025-10-07
Add more libraries to rlhf.md

mgoin:
## Purpose

There are lots of RL libs that use vLLM, let us try to list more of them! ❤️

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #27469, opened 2025-10-24
[Document] Add ms-swift library to rlhf.md

hjh0119:
## Purpose
A large number of frameworks using vLLM for RLHF have been added to rlhf.md in https://github.com/vllm-project/vllm/pull/26374.

[ms-swift](https://github.com/modelscope/ms-swift/tree/main) also uses vLLM to accelerate GRPO/GKD training; please add this framework to rlhf.md as well.

## Test Result


mgoin (approved):
Thanks for sharing!

PR #27492, opened 2025-10-24
[Performance] Support FP8 flashinfer TRTLLM MOE on Qwen3 and Qwen-3next

jiahanc:
## Purpose
- Integrate multiple routing methods for FP8 flashinfer trtllm MOE, currently only DS and Llama4
- Add  FP8 flashinfer trtllm MOE support on Qwen3 and Qwen3-next 

## Test Plan
**Qwen3-Next-80B-A3B-Instruct-FP8 on 2xB200 TP2**
```
VLLM_USE_FLASHINFER_MOE_FP8=1 VLLM_FLASHINFER_MOE_BACKEND=latency VLLM_USE_DEEP_GEMM=0 VLLM_USE_TRTLLM_ATTENTION=0 VLLM_ATTENTION_BACKEND=FLASH_ATTN vllm serve Qwen/Qwen3-Next-80B-A3B-Instruct-FP8 \
    --max-num-batched-tokens 8192 \
    --max-model-len 16384 \
    --no-enable-prefix-caching \
    --async-scheduling \
    --compilation_config.pass_config.enable_fi_allreduce_fusion true \
    --compilation_config.pass_config.enable_noop true \
    --compilation_config.cudagraph_mode FULL_DECODE_ONLY \
    --compilation_config.splitting_ops [] \
    -tp 2 
```
```
lm_eval --model local-completions --tasks gsm8k --model_args model=Qwen/Qwen3-Next-80B-A3B-Instruct-FP8,base_url=http://0.0.0.0:8000/v1/completions,max_retries=3,tokenized_requests=False,timeout=1200,max_gen_toks=2048,max_length=8192 --batch_size 2048 --trust_remote_code --limit 0.5
```

**Qwen3-30B-A3B-Instruct-2507-FP8 on 2xB200 TP2**
```
VLLM_USE_FLASHINFER_MOE_FP8=1 VLLM_FLASHINFER_MOE_BACKEND=latency VLLM_USE_DEEP_GEMM=0 vllm serve Qwen/Qwen3-30B-A3B-Instruct-2507-FP8 \
    --max-num-batched-tokens 8192 \
    --max-model-len 16384 \
    --no-enable-prefix-caching \
    --async-scheduling \
    --compilation_config.pass_config.enable_fi_allreduce_fusion true \
    --compilation_config.pass_config.enable_noop true \
    --compilation_config.cudagraph_mode FULL_DECODE_ONLY \
    --compilation_config.splitting_ops [] \
    -tp 2
```

```
lm_eval --model local-completions --tasks gsm8k --model_args model=Qwen/Qwen3-30B-A3B-Instruct-2507-FP8,base_url=http://0.0.0.0:8000/v1/completions,max_retries=3,tokenized_requests=False,timeout=1200,max_gen_toks=2048,max_length=8192 --batch_size 2048 --trust_remote_code --limit 0.5
```
## Test Result
**Qwen3-Next-80B-A3B-Instruct-FP8**
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9515|±  |0.0084|
|     |       |strict-match    |     5|exact_match|↑  |0.9197|±  |0.0106|
```

**Qwen/Qwen3-30B-A3B-Instruct-2507-FP8**
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9379|±  |0.0094|
|     |       |strict-match    |     5|exact_match|↑  |0.9364|±  |0.0095|
```
---
<details>

</details>

### DIFF
--- a/vllm/model_executor/layers/fused_moe/config.py
+++ b/vllm/model_executor/layers/fused_moe/config.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 from dataclasses import dataclass
+from enum import IntEnum
 from typing import Optional, Union
 
 import torch
@@ -89,6 +90,24 @@ def _quant_flags_to_group_shape(
             w_shape = GroupShape.PER_TOKEN
 
     return a_shape, w_shape
+
+
+# The type of method in top-K routing
+# Please keep this in sync with the counterpart defined in https://github.com/flashinfer-ai/flashinfer/blob/main/include/flashinfer/trtllm/fused_moe/runner.h
+class RoutingMethodType(IntEnum):
+    # Default: Softmax -> TopK
+    Default = (0,)
+    # Renormalize: TopK -> Softmax
+    Renormalize = (1,)
+    # DeepSeekV3: Sigmoid -> RoutingBiasAdd -> Top2 in group -> Top4 groups
+    # -> Top8 experts from the Top4 groups
+    DeepSeekV3 = (2,)
+    # Llama4: Top1 -> Sigmoid
+    Llama4 = (3,)
+    # Qwen3: Softmax -> TopK -> Renormalize
+    RenormalizeNaive = (4,)
+    # Unspecified
+    Unspecified = 5.0
 
 
 @dataclass
--- a/vllm/model_executor/layers/fused_moe/config.py
+++ b/vllm/model_executor/layers/fused_moe/config.py
@@ -104,10 +104,12 @@ class RoutingMethodType(IntEnum):
     DeepSeekV3 = (2,)
     # Llama4: Top1 -> Sigmoid
     Llama4 = (3,)
-    # Qwen3: Softmax -> TopK -> Renormalize
+    # RenormalizeNaive: Softmax -> TopK -> Renormalize
     RenormalizeNaive = (4,)
+    # TopK: TopK (no softmax)
+    TopK = (5,)
     # Unspecified
-    Unspecified = 5.0
+    Unspecified = 6.0
 
 
 @dataclass
--- a/vllm/model_executor/layers/fused_moe/flashinfer_trtllm_moe.py
+++ b/vllm/model_executor/layers/fused_moe/flashinfer_trtllm_moe.py
@@ -49,7 +49,7 @@ def flashinfer_fused_moe_blockscale_fp8(
     # NOTE: scales of hidden states have to be transposed!
     a_sf_t = a_sf.t().contiguous()
     return flashinfer_trtllm_fp8_block_scale_moe(
-        routing_logits=routing_logits,
+        routing_logits=routing_logits, 
         routing_bias=routing_bias,
         hidden_states=a_q,
         hidden_states_scale=a_sf_t,
--- a/vllm/model_executor/layers/fused_moe/flashinfer_trtllm_moe.py
+++ b/vllm/model_executor/layers/fused_moe/flashinfer_trtllm_moe.py
@@ -38,11 +38,7 @@ def flashinfer_fused_moe_blockscale_fp8(
     topk_group = topk_group if topk_group is not None else 0
     assert top_k <= global_num_experts
     assert top_k <= 10
-    # assert topk_group <= 4
-    # assert global_num_experts > num_expert_group
-    # assert global_num_experts % num_expert_group == 0
     assert global_num_experts % 4 == 0
-    # assert top_k < (topk_group * global_num_experts / num_expert_group)
     assert block_shape == [128, 128]
     # Routing kernel expects #experts <= #threads 512
     assert global_num_experts <= 512
--- a/vllm/model_executor/layers/fused_moe/flashinfer_trtllm_moe.py
+++ b/vllm/model_executor/layers/fused_moe/flashinfer_trtllm_moe.py
@@ -62,9 +62,7 @@ def flashinfer_fused_moe_blockscale_fp8(
         local_expert_offset=expert_offset,
         local_num_experts=local_num_experts,
         routed_scaling_factor=routed_scaling,
-        tile_tokens_dim=calculate_tile_tokens_dim(
-            x.shape[0], top_k, global_num_experts
-        ),
+        tile_tokens_dim=None,
         routing_method_type=routing_method_type,
         use_shuffled_weight=False,
     )
--- a/vllm/model_executor/layers/fused_moe/flashinfer_trtllm_moe.py
+++ b/vllm/model_executor/layers/fused_moe/flashinfer_trtllm_moe.py
@@ -30,19 +30,20 @@ def flashinfer_fused_moe_blockscale_fp8(
     local_num_experts: int,
     block_shape: list[int],
     routed_scaling: float = 1.0,
+    routing_method_type: int = 2,
 ) -> torch.Tensor:
     from vllm.utils.flashinfer import flashinfer_trtllm_fp8_block_scale_moe
 
     assert top_k <= global_num_experts
-    assert top_k <= 8
+    assert top_k <= 10
     assert topk_group <= 4
     assert global_num_experts > num_expert_group
     assert global_num_experts % num_expert_group == 0
     assert global_num_experts % 4 == 0
     assert top_k < (topk_group * global_num_experts / num_expert_group)
     assert block_shape == [128, 128]
-    # Routing kernel expects #experts <= #threads 256
-    assert global_num_experts <= 256
+    # Routing kernel expects #experts <= #threads 512
+    assert global_num_experts <= 512
 
     a_q, a_sf = per_token_group_quant_fp8(x, block_shape[1])
     # NOTE: scales of hidden states have to be transposed!
@@ -67,7 +68,7 @@ def flashinfer_fused_moe_blockscale_fp8(
         tile_tokens_dim=calculate_tile_tokens_dim(
             x.shape[0], top_k, global_num_experts
         ),
-        routing_method_type=2,  # DeepSeek-styled routing method
+        routing_method_type=routing_method_type,
         use_shuffled_weight=False,
     )
 
@@ -89,6 +90,7 @@ def flashinfer_fused_moe_blockscale_fp8_fake(
     local_num_experts: int,
     block_shape: list[int],
     routed_scaling: float = 1.0,
+    routing_method_type: int = 2,
 ) -> torch.Tensor:
     return torch.empty_like(x)
 
--- a/vllm/model_executor/layers/fused_moe/flashinfer_trtllm_moe.py
+++ b/vllm/model_executor/layers/fused_moe/flashinfer_trtllm_moe.py
@@ -1,7 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 
-
 import torch
 
 from vllm.model_executor.layers.fused_moe.utils import moe_kernel_quantize_input
--- a/vllm/model_executor/layers/fused_moe/flashinfer_trtllm_moe.py
+++ b/vllm/model_executor/layers/fused_moe/flashinfer_trtllm_moe.py
@@ -2,7 +2,7 @@
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 
 import torch
-
+from typing import Optional
 from vllm.model_executor.layers.fused_moe.utils import moe_kernel_quantize_input
 from vllm.model_executor.layers.quantization.utils.flashinfer_utils import (
     calculate_tile_tokens_dim,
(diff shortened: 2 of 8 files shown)


vllm/model_executor/layers/quantization/utils/flashinfer_utils.py
```diff
@@ -27,20 +27,24 @@ class FlashinferMoeBackend(Enum):
...
-    # from flashinfer import next_positive_power_of_2
-
-    # # Guess tokens per expert assuming perfect expert distribution first.
-    # num_tokens_per_expert = (num_tokens * top_k) // num_experts
-    # # And pad the number to the next power of 2.
-    # tile_tokens_dim = next_positive_power_of_2(num_tokens_per_expert)
-    # # Cap to 8-64 tokens per CTA tile as it's the range supported by the
-    # # kernel.
-    # tile_tokens_dim = min(max(tile_tokens_dim, 8), 64)
+    imbalance_factor = 1.3
```
pavanimajety, line 38: Nit: Could you add a comment for this number? what does it do?

vllm/model_executor/models/qwen3_next.py
```diff
@@ -100,6 +101,7 @@ class Qwen3NextSparseMoeBlock(nn.Module):
     def __init__(self, vllm_config: VllmConfig, prefix: str = ""):
         super().__init__()
 
+        self.prefix_print = prefix
```
pavanimajety: Nit: Why do we need this?

jiahanc: good catch! Used for debug, will remove

vllm/model_executor/layers/quantization/utils/flashinfer_utils.py
jiahanc: This I just revert the commented codes.  It used to use a hardcoded value 8.

requirements/cuda.txt
```diff
@@ -12,4 +12,4 @@ torchvision==0.24.0 # Required for phi3v processor. See https://github.com/pytor
 # https://github.com/facebookresearch/xformers/releases/tag/v0.0.32.post1
 # xformers==0.0.32.post1; platform_system == 'Linux' and platform_machine == 'x86_64'  # Requires PyTorch >= 2.8
 # FlashInfer should be updated together with the Dockerfile
-flashinfer-python==0.4.1
+flashinfer-python==0.5.0rc1
```
nvpohanh: Please check if this upgrade causes this issue:

```
(EngineCore_DP0 pid=4306) ERROR 10-29 17:25:06 [core.py:708]   File "/usr/local/lib/python3.12/dist-packages/vllm/v1/attention/backends/flashinfer.py", line 654, in build
(EngineCore_DP0 pid=4306) ERROR 10-29 17:25:06 [core.py:708]     fast_plan_decode(
(EngineCore_DP0 pid=4306) ERROR 10-29 17:25:06 [core.py:708]   File "/usr/local/lib/python3.12/dist-packages/vllm/v1/attention/backends/flashinfer.py", line 1152, in fast_plan_decode
(EngineCore_DP0 pid=4306) ERROR 10-29 17:25:06 [core.py:708]     raise RuntimeError(f"Error in tensor core plan: {e}") from e
(EngineCore_DP0 pid=4306) ERROR 10-29 17:25:06 [core.py:708] RuntimeError: Error in tensor core plan: Mismatched number of arguments when calling: `plan(0: ffi.Tensor, 1: ffi.Tensor, 2: ffi.Tensor, 3: ffi.Tensor, 4: ffi.Tensor, 5: ffi.Tensor, 6: int, 7: int, 8: int, 9: int, 10: int, 11: bool, 12: int, 13: int, 14: bool, 15: int, 16: int, 17: bool) -> Array<int>`. Expected 18 but got 15 arguments
```

If so, we should fix it in: https://github.com/vllm-project/vllm/blob/main/vllm/v1/attention/backends/flashinfer.py#L1132-L1153

jiahanc: Discussed offline, the issue was fixed already

vllm/model_executor/layers/fused_moe/flashinfer_trtllm_moe.py
```diff
@@ -23,26 +23,24 @@ def flashinfer_fused_moe_blockscale_fp8(
...
-    topk_group: int,
+    num_expert_group: int | None,
+    topk_group: int | None,
     intermediate_size: int,
     expert_offset: int,
     local_num_experts: int,
     block_shape: list[int],
-    routed_scaling: float = 1.0,
+    routed_scaling: float | None = 1.0,
+    routing_method_type: int = 2,
```
bnellnm: Can you use the `RoutingMethodType` enum here?  And the other places using raw ints.

mxz297:
If this PR is merged, can vllm still run with older flashinfer? We are internally just upgrading to flashinfer nightly-v0.4.1-20251027. This seems to bump flashinfer version again.  Is it possible to consider some backward compatibility with older flashinfer version? 

cc @houseroad @yeqcharlotte

jiahanc:
> If this PR is merged, can vllm still run with older flashinfer? We are internally just upgrading to flashinfer nightly-v0.4.1-20251027. This seems to bump flashinfer version again. Is it possible to consider some backward compatibility with older flashinfer version?
> 
> cc @houseroad @yeqcharlotte

Hi @mxz297 ,
There is no api change compared to v0.4.1. :)

vllm/model_executor/models/qwen3_moe.py
```diff
@@ -171,6 +172,7 @@ def __init__(
             enable_eplb=self.enable_eplb,
             num_redundant_experts=self.n_redundant_experts,
             is_sequence_parallel=self.is_sequence_parallel,
+            routing_method_type=RoutingMethodType.RenormalizeNaive,
```
mgoin: In this case, could we just derive the routing type within FusedMoE based on the value of renormalize arg? We want to generalize this to support the many existing "simple MoE" models without needing to update all model code to be so explicit

vllm/model_executor/layers/quantization/fp8.py
```diff
@@ -1214,22 +1214,20 @@ def apply(
...
                 e_score_correction_bias = (
                     e_score_correction_bias.to(x.dtype)
                     if e_score_correction_bias is not None
                     else None
                 )
+                routing_method_type = getattr(layer, "routing_method_type", 2)
                 return torch.ops.vllm.flashinfer_fused_moe_blockscale_fp8(
-                    routing_logits=router_logits.to(torch.float32),
+                    routing_logits=router_logits.to(torch.float32)
+                    if routing_method_type == 2
```
mgoin: We should use the enum rather than raw int

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1090,6 +1090,7 @@ def __init__(
         zero_expert_type: str | None = None,
         expert_mapping: list[tuple[str, str, int, str]] | None = None,
         n_shared_experts: int | None = None,
+        routing_method_type: int = 2,
```
mgoin: This should use the enum and probably be None by default so it can be derived. It certainly doesn't make sense to default to DeepSeek routing aka 2

vllm/model_executor/layers/fused_moe/flashinfer_trtllm_moe.py
```diff
@@ -23,26 +24,24 @@ def flashinfer_fused_moe_blockscale_fp8(
...
-    topk_group: int,
+    num_expert_group: int | None,
+    topk_group: int | None,
     intermediate_size: int,
     expert_offset: int,
     local_num_experts: int,
     block_shape: list[int],
-    routed_scaling: float = 1.0,
+    routed_scaling: float | None = 1.0,
+    routing_method_type: int = RoutingMethodType.DeepSeekV3,
```
mgoin, line 33: I think this should not have a default value in this function so the user has to be explicit

vllm/model_executor/layers/fused_moe/config.py
```diff
@@ -91,6 +92,26 @@ def _quant_flags_to_group_shape(
...
+    # -> Top8 experts from the Top4 groups
+    DeepSeekV3 = (2,)
+    # Llama4: Top1 -> Sigmoid
+    Llama4 = (3,)
+    # RenormalizeNaive: Softmax -> TopK -> Renormalize
+    RenormalizeNaive = (4,)
+    # TopK: TopK (no softmax)
+    TopK = (5,)
+    # Unspecified
+    Unspecified = 6.0
```
mgoin, line 112: I like the idea of having a routing method type so we can reduce the need for hacks like checking the llama 4 custom routing function within the quant method. 
However I think directly tying the values to the flashinfer trtllm fusedmoe is short-sighted if we are to leverage this across the codebase. I think if we do this right, we can actually remove other arguments we have in FusedMoE such as renormalize.
So I think this is the important design change in the PR. We could currently derive the routing type based on existing arguments, and of course allow for explicit override. I'm interested to hear @bnellnm thoughts too

I don't necessarily want to block this PR on getting the final design right, but I do want to get agreement with my other comments that this makes sense to be a more explicit control for routing types across all fused moe methods

vllm/model_executor/models/qwen3_moe.py
jiahanc: I think this comment is related to the last one. If we want to derive the routing method instead of directly passing a value to xx_model.py, we need to do in FusedMOE class. 
If the only reason to do so is used in flashinfer, this seems not very worthy, so I agree we should use this routing type for many other MOE backends :)

docker/Dockerfile
```diff
@@ -371,8 +371,8 @@ BASH
 # Install FlashInfer pre-compiled kernel cache and binaries
 # https://docs.flashinfer.ai/installation.html
 RUN --mount=type=cache,target=/root/.cache/uv \
-    uv pip install --system flashinfer-cubin==0.4.1 \
-    && uv pip install --system flashinfer-jit-cache==0.4.1 \
+    uv pip install --system flashinfer-cubin==0.5.0rc1 \
```
nvpohanh: @jiahanc Maybe change to 0.5.0 instead?

vllm/model_executor/layers/fused_moe/config.py
bnellnm: I agree with @mgoin that it would be nice to derive the routing type from existing arguments.  Would it make more sense to have a collection of router objects/functions that could be passed in directly?

jiahanc: Add logic in `FusedMOE` top check the routing method given param. 
For use of other backends and etc might need more discussion and design which might not be the scope of this PR :) 
@mgoin please let me know if this make sense 😄

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1269,6 +1271,24 @@ def __init__(
...
 
+        # ToDo: Better logic to determine the routing method type
+        if routing_method_type is not None:
+            self.routing_method_type = routing_method_type
+        else:
+            if scoring_func == "sigmoid":
+                if self.use_grouped_topk:
+                    self.routing_method_type = RoutingMethodType.DeepSeekV3
+                elif self.top_k == 1:
+                    self.routing_method_type = RoutingMethodType.Llama4
```
bnellnm, line 1410: Not directly related to (or suggested for) this PR but we could also set `apply_weights_on_input` to `True` for this case and get rid of the runtime parameter.  cc @mgoin

jiahanc: What is this param used for? Searched code base, only found `supports_apply_weight_on_input` , no `apply_weights_on_input`

bnellnm: > What is this param used for? Searched code base, only found `supports_apply_weight_on_input` , no `apply_weights_on_input`

It's actually `apply_router_weight_on_input`.  I just couldn't remember the exact name when I wrote the comment.  Afaik, it is only used for llama when topk==1, so I was wondering if we could detect and store it here while deriving the routing method.  We could also remove it as an extra argument to `apply`.  You don't need to make this change for this PR.  I just wanted to point it out.

vllm/model_executor/layers/quantization/fp8.py
```diff
@@ -1219,22 +1220,20 @@ def apply(
...
 
-                assert (
-                    renormalize and use_grouped_topk and custom_routing_function is None
-                )
                 e_score_correction_bias = (
                     e_score_correction_bias.to(x.dtype)
                     if e_score_correction_bias is not None
                     else None
                 )
+                routing_method_type = getattr(layer, "routing_method_type", 2)
```
bnellnm: nit: can you replace the `2` with the corresponding enum value?  Also, is `getattr` necessary here?  Afaict, the FusedMoE will always have a `routing_method_type` attribute?

jiahanc:
blocked by https://github.com/flashinfer-ai/flashinfer/issues/2032

bnellnm (approved):
LGTM!

vllm/model_executor/layers/quantization/fp8.py
jiahanc: Nice catch! Thanks 😄

pavanimajety:
@jiahanc  Could you please fix the comment for required flashinfer version in the PR description? cc- @hmellor / @simon-mo  for flashinfer upgrade

jiahanc:
> @jiahanc Could you please fix the comment for required flashinfer version in the PR description? cc- @hmellor / @simon-mo for flashinfer upgrade

Thanks! Done

pavanimajety (approved):
LGTM, thanks for the PR

jiahanc:
There is already 1 PR update FI version:https://github.com/vllm-project/vllm/pull/27952
Merge 27952 first and will rebase my PR

jiahanc:
> There is already 1 PR update FI version:#27952 Merge 27952 first and will rebase my PR

Rebased.  Ready to merge 😄

mgoin (approved):
LGTM to get in now. We should make an issue to use RoutingMethod more broadly

PR #24588, opened 2025-09-10
[DP/EP][GPTOSS] Use triton matmul-ogs kernels for GPTOSS DP/EP

varun-sundar-rabindranath:
## Purpose
Enable DP/EP for gpt-oss model via DeepEPHightThroughput and triton_kernel's matmul_ogs

## Test Plan
Run gpt-oss eval from [link](https://docs.vllm.ai/projects/recipes/en/latest/OpenAI/GPT-OSS.html#accuracy-evaluation-panels)

## Test Result
<google-sheets-html-origin><style type="text/css"></style>
GPT-OSS-120B |  
-- | --
  |  
Reasoning Effort | GPQA
Low | 0.66
Medium | 0.72
High | 0.8

## Benchmark
Please find benchmark results [here](https://docs.google.com/spreadsheets/d/1M1rl04U5yVhPcAbbwI4HNMLN36MQsKrFeYoGbKYjxzE/edit?usp=drive_link)
TLDR - Good TTFT numbers. Bad TPOT numbers as there is no cuda graphs for decode. 

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #27495, opened 2025-10-25
[WIP] [GPT-OSS] customized symm_mem based EP comm kernel integration

Luosuu:
Integrates Triton MoE communication kernels

achieved 60% throughput increase compared to #[24588](https://github.com/vllm-project/vllm/pull/24588)


vllm/entrypoints/openai/api_server.py
```diff
@@ -1201,16 +1201,32 @@ async def invocations(raw_request: Request):
...
+        profile_options = None
+        try:
+            payload = await raw_request.json()
+        except json.JSONDecodeError:
+            payload = None
+        except Exception:
+            payload = None
+        else:
+            if isinstance(payload, dict):
+                profile_options = payload
```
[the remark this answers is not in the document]
houseroad: This suggestion seems legit, maybe we can do something like

```
try:
    payload = await raw_request.json()
    if isinstance(payload, dict):
        profile_options = payload
except Exception:
    payload = None
```

vllm/envs.py
```diff
@@ -93,6 +93,14 @@
     VLLM_FORCE_AOT_LOAD: bool = False
     VLLM_TORCH_PROFILER_WITH_STACK: bool = True
     VLLM_TORCH_PROFILER_WITH_FLOPS: bool = False
+    USE_PROTON: bool = False
+    PROTON_PROFILE_NAME: str | None = None
```
houseroad, line 97: This seems a lot, wondering if we can infer USE_PROTON from PROTON_PROFILE_NAME, also for the ENV vars, we can try to infer as many as possible.

houseroad (commented):
Thanks for the great contribution!

Left some comments about the config part.

vllm/envs.py
Luosuu: I will remove proton related configs in this PR so that this PR can focus on symm_mem EP comm kernels

PR #27501, opened 2025-10-25
[Refactor][MLA] Move prefill/decode implementation into MLAAttention layer

### LINKED ISSUE #26516: [Refactor][MLA]: Lift prefill/decode split into compiled region
This issue is to track part 5 of #24620 follow ups.

We want the prefill/decode split outside the unified custom op so surrounding GEMMs/pointwise become visible to torch.compile, improving fusion and reducing Python overhead. This must still play nicely with piecewise CUDAGraphs. That should be the case automatically, as Inductor automatically partitions on data-dependent ops. We might need to do a bit of extra work to preserve piecewise cudagraph behavior with Dynamo splitting, but it should be possible.

This will likely require moving code from the MLA Attention backend implementation to the new MLAAttention layer class.

cc @ProExpertProg @LucasWilkinson @MatthewBonanni 

> For 5, the computation we would extract out of `unified_attention` would look something like this in the `fx.Graph` (super simplified):
> 
> ```
> # new (data-dependent) dynamic size
> n_prefill: u89 = torch.ops.vllm.split_batch(...)
> 
> # prefill path
> qkv_prefill = qkv[..., :n_prefill]
> qkv_prefill_2 = torch.bmm(qkv_prefill, ...)
> o_prefill = mla_attention_prefill(qkv_prefill_2, ...)
> 
> # decode path
> qkv_decode = qkv[..., n_prefill:]
> o_decode = mla_attention_decode(qkv_decode, ...)
> o_decode_2 = torch.bmm(o_decode, ...)
> 
> o = torch.cat([o_prefill, o_decode_2], ...)
> ```
> 
> So we would need to mark all of the ops between `qkv -> o` as "cudagraph-unsafe". The reason we want this exposed to `torch.compile` is these regions might get complex, have quantization, etc. and we still want to do fusions/optimizations on them and benefit from torch.compile.

therealnaveenkamal:
## Purpose

This PR is part 1/2 of the work to "lift the MLA prefill/decode split into the compiled region" tracked in #26516

Scope of this PR (1/2):
- Refactor MLA so that the layer owns the prefill/decode control flow.
- Implement `forward_prefill` and `forward_decode`

A follow-up PR (2/2) will focus on the compiled-region (exposing the split to `torch.compile` and wiring the custom split op).

## What this PR changes

Layer flow

- Move `forward_impl` into `MLAAttention`
- Implement `MLAAttention.forward_prefill()` and `MLAAttention.forward_decode()`

Backends implementation

- Refactor the MLA backends to delegate to the layer - implements `impl.forward_prefill`, `impl.forward_decode`:
- Remove `MLACommonBaseImpl` and fold the shared logic into a simpler `MLACommonImpl` + layer methods

## GSM8K evals (MLA models)

Model | Engine / config | GSM8K flex exact_match | GSM8K strict exact_match
-- | -- | -- | --
deepseek-ai/DeepSeek-V2-Lite | v1 engine, MLA, fp16, TP=1 | 37.5% | 37.1%
deepseek-ai/DeepSeek-R1-Distill-Llama-8B | v1 engine, MLA, fp16, TP=1 | 67.6% | 64.3%


vllm/attention/layer.py
```diff
@@ -769,6 +774,94 @@ def forward(
...
+        output_block_scale: torch.Tensor | None = None,
+        allow_compiled_split: bool = True,
+    ) -> torch.Tensor:
+        """Common MLA orchestration entry point.
+
+        """
+        # compiled-region split path
+        if self._use_compiled_split and allow_compiled_split:
+            # Determine dynamic split using a custom op 
+            n_decode_tokens = int(torch.ops.vllm.mla_split_batch(q, self.layer_name))
```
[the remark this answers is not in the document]
ProExpertProg: This op should always be present

vllm/attention/layer.py
```diff
@@ -769,6 +774,94 @@ def forward(
...
+
+            # Fast path for trivial cases
+            if n_decode_tokens <= 0:
```
ProExpertProg: I don't think we'll be able to have control flow with the unbacked (data-dependent) int here

vllm/v1/attention/backends/mla/common.py
```diff
@@ -2005,3 +2005,102 @@ def forward(
...
+        k_c_normed: torch.Tensor,
+        k_pe: torch.Tensor,
+        kv_cache: torch.Tensor,
+        attn_metadata,
+    ) -> torch.Tensor:
+        return self._forward_prefill(
+            q, k_c_normed, k_pe, kv_cache, attn_metadata, layer._k_scale
+        )
+
+    def forward_decode(
```
ProExpertProg: This should live inside the `MLAAttentionLayer` object instead of the common backend

vllm/v1/attention/backends/mla/common.py
```diff
@@ -2005,3 +2005,102 @@ def forward(
...
+            )
+            decode_q_pe = decode_q_pe.reshape(q_pe_shape)
+
+        decode_q = (decode_ql_nope, decode_q_pe)
+        if self.dcp_world_size > 1:
+            assert not fp8_attention, "DCP not support fp8 kvcache now."
+            decode_q = torch.cat(decode_q, dim=-1)
+            decode_q = get_dcp_group().all_gather(decode_q, dim=1)
+
+        attn_out, lse = self._forward_decode(decode_q, kv_cache, attn_metadata, layer)
```
ProExpertProg: Here, we can call the `forward_decode` method of the MLA backend.

vllm/attention/layer.py
```diff
@@ -763,39 +769,129 @@ def forward(
...
-                    kv_c_normed,
-                    k_pe,
-                    output,
-                    self.layer_name,
+                return self.forward_impl(
+                    q=q,
+                    kv_c_normed=kv_c_normed,
+                    k_pe=k_pe,
+                    kv_cache=self_kv_cache,
+                    attn_metadata=attn_metadata,
```
ProExpertProg: metadata is not available here

ProExpertProg (commented):
I think we should sequence this as 2 PRs:
1. Extract code from the common backend into the layer
2. Expose the prefill/decode split to torch.compile

What you have right now does the job of PR 2, and part of 1. You extracted some code from common MLA impl into the layer and then you call the `forward_prefill` and `forward_decode` on the common backend that do the BMMs etc. for the prefill and decode paths. Please also pull those methods into `MLAAttention.forward_prefill` and `MLAAttention.forward_decode`.

For 1, the code should look like this (simplified):
```

# custom op continues to wrap all of MLA
def mla_attention(q, k, v, layer_name:str):
    forward_context: ForwardContext = get_forward_context()
    attn_layer: MLAAttention = forward_context.no_compile_layers[layer_name]
    # Instead of calling the layer impl, go to the layer
    return attn_layer.forward_impl(q, k, v)


class MLAAttention(nn.Module, AttentionLayerBase):
    def forward(self, q, kv_c_normed, k_pe)
        return torch.ops.vllm.unified_mla_attention(
            q, kv_c_normed, k_pe, self.layer_name
        )
    
    def forward_impl(self, q, kv_c_normed, k_pe):
        attn_metadata = get_forward_context().attn_metadata[self.layer_name]
        # do the split, no custom op needed (yet)
        ...
        # instead of calling the backend, call method on layer
        self.forward_prefill(...)
        self.forward_decode(...)

    def forward_prefill(
        self, q, k_c_normed, k_pe, kv_cache, attn_metadata
    ) -> torch.Tensor:
        ...
        # code from CommonMLAImpl
        # only call backend for the specific attention op
        self.impl.forward_prefill()
   
    def forward_decode(
        self, q, k_c_normed, k_pe, kv_cache, attn_metadata
    ) -> torch.Tensor:
        ...
        # code from CommonMLAImpl
        # only call backend for the specific attention op
        self.impl.forward_decode()
```

vllm/attention/ops/triton_decode_attention.py
```diff
@@ -283,9 +283,9 @@ def _fwd_grouped_kernel_stage1(
     split_kv_id = tl.program_id(2)
 
     if kv_group_num > BLOCK_H:
-        VALID_BLOCK_H: tl.constexpr = BLOCK_H
+        VALID_BLOCK_H: tl.constexpr = BLOCK_H  # type: ignore[no-redef]
     else:
-        VALID_BLOCK_H: tl.constexpr = kv_group_num
+        VALID_BLOCK_H: tl.constexpr = kv_group_num  # type: ignore[no-redef]
```
LucasWilkinson: nit: why is this needed?

vllm/attention/ops/prefix_prefill.py
```diff
@@ -754,7 +754,7 @@ def context_attention_fwd(
     if current_platform.is_rocm():
         extra_kargs = {"kpack": 1, "waves_per_eu": 2}
 
-    grid = lambda META: (batch, head, triton.cdiv(max_input_len, META["BLOCK_M"]))
+    grid = lambda META: (batch, head, triton.cdiv(max_input_len, META["BLOCK_M"]))  # type: ignore[assignment]
```
LucasWilkinson: nit: why is this needed?

vllm/attention/ops/common.py
```diff
@@ -173,7 +173,7 @@ def cp_lse_ag_out_rs(
     cp_attn_out: torch.Tensor,
     cp_attn_lse: torch.Tensor,
     cp_group: GroupCoordinator,
-    ctx: CPTritonContext = None,
+    ctx: CPTritonContext | None = None,
```
LucasWilkinson, line 176: nit: why is this needed?

LucasWilkinson (dismissed):
Please update: tests/v1/attention/test_mla_backends.py to use the layer now and make sure it passes?

vllm/attention/ops/triton_decode_attention.py
therealnaveenkamal: had issues with pre-commit. i'm trying to remove these now.

vllm/attention/backends/abstract.py
```diff
@@ -384,8 +385,38 @@ def forward(
...
+        self,
+        layer: AttentionLayer,
+        q: torch.Tensor,
+        kv_cache: torch.Tensor,
+        attn_metadata: T,
+    ) -> torch.Tensor:
+        """Decode path orchestration - handled in the layer."""
+        raise NotImplementedError
+
+    def supports_compiled_split(self) -> bool:
```
ProExpertProg: Next PR

vllm/attention/layer.py
```diff
@@ -745,6 +766,8 @@ def __init__(
         self.k_range = torch.tensor(envs.K_SCALE_CONSTANT, dtype=torch.float32)
         self.v_range = torch.tensor(envs.V_SCALE_CONSTANT, dtype=torch.float32)
 
+        self._use_compiled_split = False
```
ProExpertProg: Next PR

vllm/v1/attention/backends/mla/cutlass_mla.py
```diff
@@ -99,6 +99,9 @@ def ensure_size(self, attn_metadata: MLACommonMetadata, num_kv_splits: int):
 class CutlassMLAImpl(MLACommonImpl[MLACommonMetadata]):
     can_return_lse_for_decode: bool = True
 
+    def supports_compiled_split(self) -> bool:
```
ProExpertProg: Next PR

vllm/v1/attention/backends/mla/common.py
```diff
@@ -1054,9 +1047,7 @@ def reorg_kvcache(
     return reorganized_kv_c_normed, reorganized_k_pe
 
 
-# TODO(Lucas): rename MLACommonBaseImpl -> MLACommonImpl,
-# and MLACommonImpl -> MLACommonDenseImpl or somthing like that
-class MLACommonBaseImpl(MLAAttentionImpl[A], Generic[A]):
+class MLACommonImpl(MLAAttentionImpl[A], Generic[A]):
```
ProExpertProg, line 1089: Are all these properties still needed? Could subclasses just inherit from `MLAAttentionImpl` directly?

therealnaveenkamal: I think we can do it...but the type parameter A is the metadata type this implementation consumes and each backend gets the appropriate attn_metadata.  also, we have shared attention implementation in MLACommonImpl which is used by the other subclasses right?

therealnaveenkamal:
@LucasWilkinson yes that's correct - I don't have the hardware for this. I can probably run the quantized version of this R1. maybe `unsloth/DeepSeek-R1-GGUF`?

therealnaveenkamal:
Thanks @MatthewBonanni

ProExpertProg (approved):

vllm/attention/layer.py
```diff
@@ -763,9 +786,470 @@ def forward(
...
+                context_output = context_output[..., : v.shape[-1]]
+                suffix_output = suffix_output[..., : v.shape[-1]]
+
+            output = output.view(-1, self.num_heads, self.v_head_dim)
+            merge_attn_states(
+                output=output,
+                prefix_output=context_output,
+                prefix_lse=context_lse,
+                suffix_output=suffix_output,
+                suffix_lse=suffix_lse,
```
LucasWilkinson: why is `merge_attn_states` being called twice?

LucasWilkinson (changes requested):

vllm/attention/layer.py
therealnaveenkamal: my bad,  L958-L965 should be removed (missed it during resolving merge conflict)

PR #27511, opened 2025-10-25
Add standalone multimodal encoder benchmark

alhridoy:
Fixes #25450

## Summary
     - add a standalone multimodal encoder benchmark
     (`vllm/benchmarks/encoder.py`) that loads dummy-weight models, builds
     HF-processor inputs, and measures `get_multimodal_embeddings` latency across
      configurable batch/image sizes
     - wire the runner into the CLI as `vllm bench encoder` and keep a legacy
     shim at `benchmarks/benchmark_encoder.py`
     - support JSON export, optional preprocessing timing, and sane defaults for
     batch sizes and image resolutions

### DIFF
--- a/benchmarks/benchmark_encoder.py
+++ b/benchmarks/benchmark_encoder.py
@@ -1,0 +1,26 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""Legacy entrypoint for the standalone encoder benchmark."""
+
+import sys
+
+
+def main() -> None:
+    print(
+        """DEPRECATED: This script has been moved to the vLLM CLI.
+
+Please use the following command instead:
+    vllm bench encoder
+
+For help with the new command, run:
+    vllm bench encoder --help
+
+Alternatively, you can run the new command directly with:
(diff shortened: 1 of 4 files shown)


alhridoy:
Hi @ywang96 , would you look into this PR?

benchmarks/benchmark_encoder.py
```diff
@@ -0,0 +1,26 @@
+# SPDX-License-Identifier: Apache-2.0
```
DarkLight1337, line 1: No need to add this deprecation since it's a new script

vllm/benchmarks/encoder.py
```diff
@@ -0,0 +1,384 @@
...
+
+    start = time.perf_counter() if include_preproc else None
+    mm_inputs = processor.apply(
+        prompt=inputs.prompt,
+        mm_data=inputs.mm_data,
+        hf_processor_mm_kwargs=mm_processor_kwargs,
+        tokenization_kwargs=inputs.tokenization_kwargs,
+    )
+    preprocess_ms = None
+    if include_preproc:
```
DarkLight1337, line 213: I think we should just always include this, since we can get this metric essentially for free

PR #27518, opened 2025-10-26
[bugfix] fix wrong `dcp_local_seq_lens` calc

pisceskkk:
When the `seq_lens` is exactly divisible by the `dcp_world_size`, it causes the `dcp_local_seq_lens` on all ranks to be incremented by one. Fix this calculation logic.

CC @youzhedian @minosfuture @youkaichao


minosfuture (approved):
Thanks for the fix!

youkaichao (approved):
thanks for the fix!

PR #27532, opened 2025-10-26
[Attention] Use sparse prefill kernel for fp8 kv-cache in DeepSeek-v3.2

LucasWilkinson:
When doing prefill up-convert the kv-cache from fp8 to bf16 and call the bf16 prefill kernel instead of the decode kernel. This PR introduce global workspace management to have the bf16 workspace overlap with the MoE workspace buffers.

## GSM8K Accuracy (DeepSeek-V3.2, FP8 KV-cache, TP=8)

| Branch | Accuracy |
|--------|----------|
| Main | 95.30% ± 0.58% |
| PR (Hybrid) | **95.53% ± 0.57%** |

## Benchmark Results (4096 in, 512 out, 100 prompts)

| Config | Branch | Req/s | Tok/s | Delta |
|--------|--------|-------|-------|-------|
| TP=8, FP8 | Main | 1.58 | 806 | — |
| TP=8, FP8 | PR | 1.58 | 806 | **0%** |
| DP=8+EP, FP8 | Main | 2.36 | 1207 | — |
| DP=8+EP, FP8 | PR | **2.53** | **1293** | **+7%** |

## Hybrid Approach

Uses `MIN_HEADS_FOR_BF16_PREFILL = 32` threshold:
- **TP=8** (16 heads/rank): Mixed batch path (same as current main)
- **DP=8+EP** (128 heads): BF16 prefill path → 7% faster

## Commands Used

```bash
# GSM8K Eval
python ~/.config/aliases/dot-aliases/scripts/gsm8k_eval.py \
  --model deepseek-ai/DeepSeek-V3.2 --port 3333 \
  --server-args '--tensor-parallel-size 8 --kv-cache-dtype fp8 --attention-config {"backend":"FLASHMLA_SPARSE"}'

# Benchmark (TP=8)
vllm serve deepseek-ai/DeepSeek-V3.2 -tp 8 --kv-cache-dtype fp8 \
  --attention-config '{"backend":"FLASHMLA_SPARSE"}' --port 3333
vllm bench serve --model deepseek-ai/DeepSeek-V3.2 --port 3333 \
  --num-prompts 100 --request-rate inf --random-input-len 4096 --random-output-len 512

# Benchmark (DP=8+EP)
vllm serve deepseek-ai/DeepSeek-V3.2 -dp 8 --enable-expert-parallel \
  --kv-cache-dtype fp8 --attention-config '{"backend":"FLASHMLA_SPARSE"}' \
  --max-model-len 8192 --port 3333
vllm bench serve --model deepseek-ai/DeepSeek-V3.2 --port 3333 \
  --num-prompts 100 --request-rate inf --random-input-len 4096 --random-output-len 512
```


vllm/model_executor/layers/fused_moe/modular_kernel.py
```diff
@@ -786,24 +748,50 @@ def _allocate_buffers(
...
+                envs.VLLM_FUSED_MOE_CHUNK_SIZE,
+                N,
+                K,
+                top_k,
+                global_num_experts,
+                local_num_experts,
+                None,  # Pass None to avoid using sampled token counts
+            )
+
+            current_workspace_manager().get_simultaneous(
```
heheda12345: do we have to do these allocation during model execution? Is it possible to setup the memory buffer before real execution to reduce the runtime overhead?
For example, I'm thinking of
1. during model init, call current_workspace_manager().get_simultaneous() to tell the workspace manager the max possible size the model may used
2. lock the memory space
3. during profile run, allocate the memory and save to `self.xxx` like `self.workspace13 = workspace13; self.workspace2 = workspace2`
4. during model execution, just use the `self.workspace13`

vllm/model_executor/models/deepseek_v2.py
```diff
@@ -518,8 +519,15 @@ def sparse_attn_indexer(
...
+
     # assert isinstance(attn_metadata, dict)
     if not isinstance(attn_metadata, dict):
+        # Reserve workspace for indexer during profiling run
+        current_workspace_manager().get_simultaneous(
+            ((total_seq_lens, head_dim), torch.float8_e4m3fn),
+            ((total_seq_lens, 4), torch.uint8),
+        )
+
         return sparse_attn_indexer_fake(
```
heheda12345, line 631: should these lines in `sparse_attn_indexer_fake` be removed? I wrote them to mimic the activation memory during profile run.

```
_flattened_kv = torch.empty(
        [total_seq_lens, head_dim + 4], device=k.device, dtype=torch.uint8
    )
    _k_fp8 = _flattened_kv[..., :head_dim].view(torch.float8_e4m3fn).contiguous()
    _k_scale = _flattened_kv[..., head_dim:].view(torch.float32).contiguous()
```

vllm/v1/attention/backends/mla/flashmla_sparse.py
```diff
@@ -108,6 +110,40 @@ class FlashMLASparseMetadata:
     block_size: int = 64
     topk_tokens: int = 2048
 
+    num_prefill_reqs: int = 0
+    num_decode_reqs: int = 0
+    num_prefill_tokens: int = 0
+    num_decode_tokens: int = 0
+
+    # Sequence lengths (context + query) for prefill requests
+    prefill_seq_lens: torch.Tensor | None = None
```
heheda12345: can you write down the shape of this tensor?

vllm/v1/attention/backends/mla/flashmla_sparse.py
```diff
@@ -237,7 +313,45 @@ def triton_convert_req_index_to_global_index(
     return out
 
 
-@dataclass
+def split_prefill_chunks(
```
heheda12345: can you move `split_prefill_chunks` in `indexer.py` to a utils and reuse it here?

vllm/v1/attention/backends/mla/flashmla_sparse.py
```diff
@@ -237,7 +313,45 @@ def triton_convert_req_index_to_global_index(
...
+
+def get_prefill_workspace_size(vllm_config: VllmConfig):
+    max_model_len = vllm_config.model_config.max_model_len
+    # NOTE(Lucas): 5 is a magic number for controlling the prefill buffer size.
+    # May be tuned later.
+    # Memory usage: 5 * max_model_len * 576 * 2 bytes
+    #   Example: DeepSeek-V3.2 with max_model_len=163840 ->
+    #            5 * 163840 * 576 * 2 = ~900 MB
+    # This fits nicely below the typical MoE workspace size of >2GB so this is "free"
+    return max_model_len * 5
```
heheda12345, line 367: Do you want to reuse the one in indexer or you think they should be set independently? (Both works for me) If set independently, a small optimization may be setting them to different values so that they are claiming similar size of memory.

vllm/v1/attention/backends/mla/flashmla_sparse.py
```diff
@@ -180,6 +228,8 @@ def triton_convert_req_index_to_global_index(
     BLOCK_SIZE: int = 64,
     NUM_TOPK_TOKENS: int = 2048,
     BLOCK_N: int = 128,  # tile width along columns
+    prefill_request_id: torch.Tensor | None = None,
```
heheda12345:
```suggestion
    prefill_request_id_using_workspace: torch.Tensor | None = None,
```

If we still want pure bf16 code path (e.g., for prefill node), we shouldn't pass the prefill request ids here. So I suggest to state these are the set of request that reads from the workspace. Similar to `HAS_PREFILL` in the triton kernel.

And good to have a unit test for this function.

vllm/v1/worker/workspace.py
```diff
@@ -0,0 +1,254 @@
...
+                    for ws in self._current_workspaces
+                    if ws is not None
+                ],
+            )
+
+    def is_locked(self) -> bool:
+        """Check if workspace is locked."""
+        return self._locked
+
+    def current_allocated_size_bytes(self) -> int:
```
heheda12345: This function is not used now.

vllm/v1/worker/workspace.py
```diff
@@ -0,0 +1,254 @@
...
+        """Get the size of the current workspace in bytes."""
+        return self._workspace_size_bytes(
+            self._current_workspaces[dbo_current_ubatch_id()]
+        )
+
+    def get(self, shape: tuple[int, ...], dtype: torch.dtype) -> torch.Tensor:
```
heheda12345: do we need the "get" API? Can we always use `get_simultaneous`?

vllm/v1/worker/workspace.py
LucasWilkinson: removed :+1:

vllm/v1/attention/backends/mla/flashmla_sparse.py
LucasWilkinson: renamed to `prefill_workspace_request_ids` :+1:

vllm/model_executor/models/deepseek_v2.py
LucasWilkinson: theyre back 👍 (not sure when haha; I guess I must have done it and just not pushed it)

vllm/model_executor/layers/fused_moe/modular_kernel.py
LucasWilkinson: ya we should do something like this the only complication currently is when `self.fused_experts.supports_chunking() == False` (i.e. PPLX or DeepEP LL) then we need profile run to know the shape of the workspaces because in that case the profile run actually mirrors the worst case scenario (hence gating this logic). 

This code is meant to a straight refactor of https://github.com/vllm-project/vllm/pull/27426 (once that lands) to use the new workspace manager so im partial to leaving this optimization to a future PR if you are cool with it. I am trying to learn more about the MoE chunking code in-order to propose a broader UX refactor there since I find it confusing currently and I think this optimization could be part of that.

vllm/v1/worker/workspace.py
LucasWilkinson: not really; just helps with line width and avoids extra alignment calculations; down to get rid of

LucasWilkinson: removed

vllm/v1/attention/backends/mla/flashmla_sparse.py
LucasWilkinson: done

vllm/v1/attention/backends/mla/flashmla_sparse.py
LucasWilkinson: done

vllm/v1/attention/backends/mla/flashmla_sparse.py
LucasWilkinson: done

vllm/model_executor/layers/fused_moe/modular_kernel.py
heheda12345: Yeah works for me. Can you do some benchmark to ensure no perf regression?

vllm/model_executor/models/deepseek_v2.py
heheda12345: I want to say that if you choose to reserve the buffers of shape (total_seq_lens, head_dim) and (total_seq_lens, 4) in the workspace, you don't need to run the following in sparse_attn_indexer_fake 
```
_flattened_kv = torch.empty(
        [total_seq_lens, head_dim + 4], device=k.device, dtype=torch.uint8
    )
```
to also reserve memory for these two tensors in activation memory.

heheda12345 (approved):
LGTM!

csrc/cache_kernels.cu
```diff
@@ -1202,6 +1280,57 @@ void cp_gather_cache(
...
+  TORCH_CHECK(src_cache.device() == dst.device(),
+              "src_cache and dst must be on the same device");
+  TORCH_CHECK(src_cache.device() == block_table.device(),
+              "src_cache and block_table must be on the same device");
+  TORCH_CHECK(src_cache.device() == seq_lens.device(),
+              "src_cache and seq_lens must be on the same device");
+  TORCH_CHECK(src_cache.device() == workspace_starts.device(),
+              "src_cache and workspace_starts must be on the same device");
+
+  TORCH_CHECK(src_cache.dtype() == torch::kUInt8, "src_cache must be uint8");
```
kebe7jun, line 1311: This may cause https://github.com/vllm-project/vllm/issues/33883

PR #27533, opened 2025-10-26
[perf] Optimize Qwen2-VL Startup Performance with LRU Cache

skyloevil:
## Summary

Add `@lru_cache` to `get_image_size_with_most_features()` to eliminate redundant `smart_resize()` calls during profiling, improving startup performance for Qwen2-VL family models.

## Problem

During profiling/startup, `get_image_size_with_most_features()` is called twice:
1. Once for calculating max image tokens
2. Once for calculating max video tokens

Each call triggers expensive `smart_resize()` computation , causing slow startup as noted in `vllm/multimodal/processing.py:1178-1181`.

## Solution

Add `@lru_cache(maxsize=128)  # noqa: B019` decorator to cache the result:

```python
@lru_cache(maxsize=128)  # noqa: B019
def get_image_size_with_most_features(self) -> ImageSize:
    max_image_size, _ = self._get_vision_info(...)
    return max_image_size
```

The second call hits cache instead of recomputing, eliminating duplicate work.

## Performance Impact

- **Cache efficiency**: 50% reduction in `smart_resize()` calls
- **Code change**: Only 1 line added

## Affected Models

- Qwen2VLForConditionalGeneration
- Qwen2_5_VLForConditionalGeneration
- Qwen3VLForConditionalGeneration
- Tarsier2ForConditionalGeneration

## Testing

- Syntax check: ✅ Pass
- Pre-commit hooks: ✅ Expected to pass (follows existing pattern)
- Manual verification: Confirmed cache decorator placement

## Files Changed

- `vllm/model_executor/models/qwen2_vl.py`: +1 line

### DIFF
--- a/vllm/model_executor/models/qwen2_vl.py
+++ b/vllm/model_executor/models/qwen2_vl.py
@@ -27,7 +27,7 @@
 
 import math
 from collections.abc import Callable, Iterable, Mapping, Sequence
-from functools import partial
+from functools import lru_cache, partial
 from typing import Annotated, Any, Literal, TypeAlias
 
 import torch
@@ -1034,6 +1034,7 @@ class Qwen2VLProcessingInfo(BaseProcessingInfo):
         )
         return num_video_tokens
 
+    @lru_cache(maxsize=128)  # noqa: B019
     def get_image_size_with_most_features(self) -> ImageSize:
         max_image_size, _ = self._get_vision_info(
             image_width=9999999,


DarkLight1337:
Looking at the code of `smart_resize`, it doesn't look that expensive:

```
def smart_resize(
    height: int, width: int, factor: int = 28, min_pixels: int = 56 * 56, max_pixels: int = 14 * 14 * 4 * 1280
):
    """Rescales the image so that the following conditions are met:

    1. Both dimensions (height and width) are divisible by 'factor'.

    2. The total number of pixels is within the range ['min_pixels', 'max_pixels'].

    3. The aspect ratio of the image is maintained as closely as possible.

    """
    if max(height, width) / min(height, width) > 200:
        raise ValueError(
            f"absolute aspect ratio must be smaller than 200, got {max(height, width) / min(height, width)}"
        )
    h_bar = round(height / factor) * factor
    w_bar = round(width / factor) * factor
    if h_bar * w_bar > max_pixels:
        beta = math.sqrt((height * width) / max_pixels)
        h_bar = max(factor, math.floor(height / beta / factor) * factor)
        w_bar = max(factor, math.floor(width / beta / factor) * factor)
    elif h_bar * w_bar < min_pixels:
        beta = math.sqrt(min_pixels / (height * width))
        h_bar = math.ceil(height * beta / factor) * factor
        w_bar = math.ceil(width * beta / factor) * factor
    return h_bar, w_bar
```

Probably the overhead is coming from somewhere else.

hmellor:
Closing as stale

PR #27547, opened 2025-10-27
[BUG] Fix hybrid kvcache kernel page size issue

vadiklyutiy:
## Purpose
The following 
```
VLLM_USE_TRTLLM_ATTENTION=0 vllm serve Qwen/Qwen3-Next-80B-A3B-Instruct -tp 4 --enable-expert-parallel --no-enable-prefix-caching --async-scheduling
```
```
```
Produced 
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.0197|±  |0.0038|
```

The issue was with intermix of `kv_manager_block_size` and `kernel_block_size` we passed `kv_manager_block_size` instead of `kernel_block_size` that cause garbage loads in attn kernel. 

## Test Result
After fix 
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8514|±  |0.0098|
|     |       |strict-match    |     5|exact_match|↑  |0.8089|±  |0.0108|
```

### DIFF
--- a/vllm/v1/attention/backends/flashinfer.py
+++ b/vllm/v1/attention/backends/flashinfer.py
@@ -333,7 +333,31 @@ class FlashInferMetadataBuilder(AttentionMetadataBuilder[FlashInferMetadata]):
         self.num_kv_heads = self.kv_cache_spec.num_kv_heads
         self.head_dim = self.kv_cache_spec.head_size
         FlashInferBackend.validate_head_size(self.head_dim)
-        self.page_size = self.kv_cache_spec.block_size
+
+        # IMPORTANT: page_size must match the actual kernel block size,
+        # not the KV manager block size!
+        # When using hybrid blocks (kv_manager_block_size != kernel_block_size),
+        # the KV cache is allocated with kernel_block_size, so we must use that
+        # for page_size when calling FlashInfer.
+        kv_manager_block_size = self.kv_cache_spec.block_size
+        kernel_block_size = self.kv_cache_spec.find_compatible_kernel_block_sizes(
+            FlashInferBackend, return_all=False
+        )[0]
+
+        self.page_size = kernel_block_size
+
+        if kernel_block_size != kv_manager_block_size:
+            logger.info_once(
+                "FlashInfer: Using kernel_block_size=%d (page_size) != "
+                "kv_manager_block_size=%d. Hybrid blocks mode.",
+                kernel_block_size,
+                kv_manager_block_size,
+            )
+        else:
+            logger.info_once(
+                "FlashInfer: Using page_size=%d (matches kv_manager_block_size)",
+                self.page_size,
+            )
 
         self.cache_dtype = self.cache_config.cache_dtype
         if self.cache_dtype.startswith("fp8"):
--- a/vllm/v1/attention/backends/flashinfer.py
+++ b/vllm/v1/attention/backends/flashinfer.py
@@ -332,28 +332,13 @@ class FlashInferMetadataBuilder(AttentionMetadataBuilder[FlashInferMetadata]):
 
         # IMPORTANT: page_size must match the actual kernel block size,
         # not the KV manager block size!
-        # When using hybrid blocks (kv_manager_block_size != kernel_block_size),
+        # When using hybrid blocks(self.kv_cache_spec.block_size != kernel_block_size),
         # the KV cache is allocated with kernel_block_size, so we must use that
         # for page_size when calling FlashInfer.
-        kv_manager_block_size = self.kv_cache_spec.block_size
         kernel_block_size = self.kv_cache_spec.find_compatible_kernel_block_sizes(
             FlashInferBackend, return_all=False
         )[0]
-
         self.page_size = kernel_block_size
-
-        if kernel_block_size != kv_manager_block_size:
-            logger.info_once(
-                "FlashInfer: Using kernel_block_size=%d (page_size) != "
-                "kv_manager_block_size=%d. Hybrid blocks mode.",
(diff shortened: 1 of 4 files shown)


vllm/v1/attention/backends/flashinfer.py
```diff
@@ -333,7 +333,31 @@ def __init__(
...
+            logger.info_once(
+                "FlashInfer: Using kernel_block_size=%d (page_size) != "
+                "kv_manager_block_size=%d. Hybrid blocks mode.",
+                kernel_block_size,
+                kv_manager_block_size,
+            )
+        else:
+            logger.info_once(
+                "FlashInfer: Using page_size=%d (matches kv_manager_block_size)",
+                self.page_size,
```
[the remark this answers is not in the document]
vadiklyutiy: hm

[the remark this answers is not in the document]
vadiklyutiy: Good catch!
I was able to reproduce the issue with 
```
LLM_USE_TRTLLM_ATTENTION=0 vllm serve Qwen/Qwen3-Next-80B-A3B-Instruct -tp 4 --enable-expert-parallel --no-enable-prefix-caching --async-scheduling  --max_model_len 2048
```
```
```

I fixed it the the last commit. 
Could you repeat review?

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -333,7 +329,36 @@ def __init__(
...
+
+        if kernel_block_size != kv_manager_block_size:
+            logger.info_once(
+                "FlashInfer: Using kernel_block_size=%d (page_size) != "
+                "kv_manager_block_size=%d. Hybrid blocks mode.",
+                kernel_block_size,
+                kv_manager_block_size,
+            )
+        else:
+            logger.info_once(
```
pavanimajety: Nit: These can be changed to debug logs.

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -333,7 +329,36 @@ def __init__(
...
         self.head_dim = self.kv_cache_spec.head_size
         FlashInferBackend.validate_head_size(self.head_dim)
-        self.page_size = self.kv_cache_spec.block_size
+
+        # IMPORTANT: page_size must match the actual kernel block size,
+        # not the KV manager block size!
+        # When using hybrid blocks (kv_manager_block_size != kernel_block_size),
+        # the KV cache is allocated with kernel_block_size, so we must use that
+        # for page_size when calling FlashInfer.
+        kv_manager_block_size = self.kv_cache_spec.block_size
```
pavanimajety: Is kv_manager_block_size used in flashinfer.py other than the logs below? Do we need to maintain this variable?

pavanimajety: Also curious why this is an issue only for Flashinfer backend and not Flash Attention backend? 
FA backend also [uses](https://github.com/CentML/vllm/blob/955f38d432711f47abfe05c244c575c570e91112/vllm/v1/attention/backends/flash_attn.py#L221)  `self.block_size = kv_cache_spec.block_size`

vadiklyutiy: > Also curious why this is an issue only for Flashinfer backend and not Flash Attention backend? FA backend also [uses](https://github.com/CentML/vllm/blob/955f38d432711f47abfe05c244c575c570e91112/vllm/v1/attention/backends/flash_attn.py#L221) `self.block_size = kv_cache_spec.block_size`

FA has/uses same `kv_manager_block_size == kernel_block_size`

vadiklyutiy: > Is kv_manager_block_size used in flashinfer.py other than the logs below? Do we need to maintain this variable?

Nope. I removed `kv_manager_block_size` and logs

vadiklyutiy:
@pavanimajety @heheda12345 
Could you please take a look at this PR?

tdoublep:
I have solved this is a slightly different way in this PR: https://github.com/vllm-project/vllm/pull/27753

I opted to change the way we intialize things so we can pass the kernel block sizes to the constructor of the attention metadata builders.

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -333,7 +329,21 @@ def __init__(
...
         self.head_dim = self.kv_cache_spec.head_size
         FlashInferBackend.validate_head_size(self.head_dim)
-        self.page_size = self.kv_cache_spec.block_size
+
+        # IMPORTANT: page_size must match the actual kernel block size,
+        # not the KV manager block size!
+        # When using hybrid blocks(self.kv_cache_spec.block_size != kernel_block_size),
+        # the KV cache is allocated with kernel_block_size, so we must use that
+        # for page_size when calling FlashInfer.
+        kernel_block_size = self.kv_cache_spec.find_compatible_kernel_block_sizes(
```
heheda12345, line 338: can you refactor _select_common_block_size a bit so that you can call something like GPUModelRunner._select_common_block_size? I don't think it's a good idea to move to `self.kv_cache_spec.find_compatible_kernel_block_sizes`

vadiklyutiy: To avoid misunderstand I'd like to double check do you mean `find_compatible_kernel_block_sizes` or `_select_common_block_size`?

heheda12345: Depends on you. Just call GPUModelRunner.xxxx() here and minimize the changes to gpu model runner

vadiklyutiy:
I'm not going to play this game

PR #27560, opened 2025-10-27
[Bugfix] Validate custom logits processor xargs for online serving

Isotr0py:
## Purpose
- The default model_type for deepseek-ocr is `deepseek_vl_v2`, so the default deepseek-ocr chat template won't be used by default: https://github.com/vllm-project/vllm/blob/a806c14cc7cf465bb30f34cd9782bd28ae3eb08e/vllm/transformers_utils/chat_templates/registry.py#L36
- Also allow list of int as `vllm_xargs` for `ChatCompletionRequest`, otherwise `"whitelist_token_ids": [128821, 128822],` parameters for deepseek-ocr's custom logits processor is not allowed.
```
openai.BadRequestError: Error code: 400 - {'error': {'message': "[{'type': 'string_type', 'loc': ('body', 'vllm_xargs', 'whitelist_token_ids', 'str'), 'msg': 'Input should be a valid string', 'input': [128821, 128822]}, {'type': 'int_type', 'loc': ('body', 'vllm_xargs', 'whitelist_token_ids', 'int'), 'msg': 'Input should be a valid integer', 'input': [128821, 128822]}, {'type': 'float_type', 'loc': ('body', 'vllm_xargs', 'whitelist_token_ids', 'float'), 'msg': 'Input should be a valid number', 'input': [128821, 128822]}]", 'type': 'Bad Request', 'param': None, 'code': 400}}
```
- Also validate request's `vllm_xargs` to avoid unexpected behavior from custom logits processor.

## Test Result

---
<details>

</details>


noooop (approved):

DarkLight1337 (changes requested):
We are going to patch a security vulnerability before merging this

vllm/entrypoints/openai/logits_processors.py
```diff
@@ -1,13 +1,17 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 
-from collections.abc import Iterable
+from collections.abc import Iterable, Sequence
```
njhill: I think actually this file should be deleted altogether, it's vestigial V0 code.

Have opened https://github.com/vllm-project/vllm/pull/27601 for that.

vllm/v1/sample/logits_processor/__init__.py
```diff
@@ -245,6 +245,18 @@ def __init__(
         # was when the partial was created.
         self.req_info: dict[int, partial[torch.Tensor]] = {}
 
+    @classmethod
+    def validate_params(cls, params: SamplingParams):
```
njhill: I think this should go in the base `LogitsProcessor` class, it's applicable to custom logits procs generally not just per-request adapted ones.

vllm/entrypoints/openai/logits_processors.py
```diff
@@ -90,3 +94,15 @@ def get_logits_processors(
...
+
+
+def validate_logits_processors_parameters(
+    logits_processors: Sequence[str | LogitsProcessor] | None,
+    sampling_params: SamplingParams,
+):
+    if logits_processors is None:
+        return None
+
+    for logits_procs in _load_custom_logitsprocs(logits_processors):  # type: ignore[arg-type]
```
njhill: We shouldn't attempt to load the logits processors on every request.

We should also be careful about loading them at all in the front-end process since I think we currently only do it in the worker processes.

We will probably have to do this to support the request validation, it's hopefully ok if we only load the class without instantiating it (so hopefully won't be allocating tensors etc. for example), but this could for example cause cuda to be initialized which we don't want to do in this process.

Isotr0py: > but this could for example cause cuda to be initialized which we don't want to do in this process.

IIUC, this case is usually occurred when the file contents code that will caused cuda initialization outside logits processor implementation like this?
```python3
class CustomLogitsProcessor:
    ...

# some code initialize CUDA outside
torch._C._cuda_init()
```

Given that we're doing request validation in front-end process, I think we can hide all GPUs temporarily when loading the custom logits class, so it will trigger `RuntimeError: No CUDA GPUs are available` when trying to intialize CUDA and refuse request with this invalid processor. WDYT?

docs/features/custom_logitsprocs.md
```diff
@@ -180,7 +180,10 @@ RequestLogitsProcessor = Union[
...
 
-You can wrap the request-level logits processor by subclassing `AdapterLogitsProcessor` as shown in the example below (in this example, `DummyPerReqLogitsProcessor` is a stand-in for your request-level logits processor which needs to be wrapped.) Override `AdapterLogitsProcessor.is_argmax_invariant(self)` to accurately reflect whether your request-level logits processor may impact which token has the highest-value logit. Override `AdapterLogitsProcessor.new_req_logits_processor(self,params)` to create a new request-level logits processor instance from a `SamplingParams` instance:
+You can wrap the request-level logits processor by subclassing `AdapterLogitsProcessor` as shown in the example below (in this example, `DummyPerReqLogitsProcessor` is a stand-in for your request-level logits processor which needs to be wrapped.):
```
afeldman-nm: It is great that you updated the documentation to reflect these changes.

Correct me if wrong, but it appears like this PR right now is only updating the documentation for the special case of adapting a request-level logits processor.

I think we will probably also want to update the documentation for

1. The "Programming Model" section of the logits processors design docs https://docs.vllm.ai/en/latest/design/logits_processors.html#logits-processor-programming-model

2. Other sections of the custom logits processor design docs, specifically "Creating a custom logits processor" (https://docs.vllm.ai/en/latest/features/custom_logitsprocs.html#creating-a-custom-logits-processor), "Passing Custom Argument to a Custom Logits Procesor" (https://docs.vllm.ai/en/latest/features/custom_logitsprocs.html#passing-custom-argument-to-a-custom-logits-processor), "Example custom logits processor implementation" (https://docs.vllm.ai/en/latest/features/custom_logitsprocs.html#example-custom-logits-processor-implementation)

afeldman-nm (changes requested):
Thanks for this @Isotr0py ! It will be good to have argument validation

vllm/entrypoints/openai/logits_processors.py
njhill: @Isotr0py that sounds like a good idea

vllm/v1/sample/logits_processor/interface.py
```diff
@@ -64,6 +64,15 @@ def __init__(
...
         raise NotImplementedError
 
+    @classmethod
+    @abstractmethod
+    def validate_params(cls, sampling_params: SamplingParams):
+        """Validate sampling params for this logits processor.
+
+        Raise ValueError for invalid ones.
+        """
+        raise NotImplementedError
```
njhill, line 73: I think we should just `pass` here for backwards compatibility and simplicity when implementations don't care about validating args.

vllm/v1/sample/logits_processor/builtin.py
```diff
@@ -131,6 +137,10 @@ def is_argmax_invariant(self) -> bool:
         outcome of argmax in greedy sampling."""
         return False
 
+    @classmethod
+    def validate_params(cls, sampling_params: SamplingParams):
+        pass
```
njhill: If other comment is addressed we can remove these.

vllm/v1/sample/logits_processor/builtin.py
```diff
@@ -46,6 +46,12 @@ def is_argmax_invariant(self) -> bool:
         """Min-p never impacts greedy sampling"""
         return True
 
+    @classmethod
+    def validate_params(cls, sampling_params: SamplingParams):
+        min_p = sampling_params.min_p
+        if min_p is not None and (min_p < 0.0 or min_p > 1.0):
+            raise ValueError("min_p should be in the range [0.0, 1.0]")
```
njhill: Probably best to exclude this too think I since we already validate `min_p` separately given it's a built in sampling param.

Isotr0py:
Sorry for the delayed update! I was a bit occupied last week. 😅 This PR should be ready for a further review now!

docs/features/custom_logitsprocs.md
```diff
@@ -38,6 +38,11 @@ Custom logits processors must subclass `vllm.v1.sample.logits_processor.LogitsPr
     * Use the `BatchUpdate` members to update logits processor internal state
     * **Note:** batch update data structure may be `None`, signaling no change to the batch constituents. In this case, the LogitsProcessor might still want to update its state based on the updated `output_token_ids` lists that it could have retained when they were added.
 
+* `validate_params(cls, sampling_params: SamplingParams)`:
+    * Raise `ValueError` if `SamplingParams` has invalid arguments (especially custom arguments) used by logits processor.
+    * When request is sent to entrypoint, `validate_params()` will validate `SamplingParams` and refuse request with invalid arguments.
+    * **Note:** it's important to implent `validate_params()` to prevent invalid parameters for custom logits processor. Otherwise requests with invalid parameters can cause unexpected behaviour in custom logits processor.
```
DarkLight1337:
```suggestion
    * **Note:** it's important to implement `validate_params()` to prevent invalid parameters for custom logits processor. Otherwise requests with invalid parameters can cause unexpected behaviour in custom logits processor.
```

docs/features/custom_logitsprocs.md
```diff
@@ -157,6 +163,15 @@ The contrived example below implements a custom logits processor which consumes
             logits[rows, cols] = values_to_keep
 
             return logits
+
+        @classmethod
```
DarkLight1337: Let's define this at the top of the class to be consistent with L200

vllm/model_executor/models/deepseek_ocr.py
```diff
@@ -168,6 +160,14 @@ def validate_params(cls, params: SamplingParams):
                 f"got {whitelist_token_ids}."
             )
 
+    def __init__(
```
DarkLight1337: This init method is redundant

DarkLight1337 (approved):
Thanks, LGTM

DarkLight1337:
/gemini review

Just in case

youkaichao:
for record, this PR adds huge performance regression and is fixed by https://github.com/vllm-project/vllm/pull/28462

PR #27568, opened 2025-10-27
[DeepSeek v3.2] Make top-k work for any logit values.

dcampora:
## Purpose

This PR allows top_k_per_row work for any values in logits. Even if the logits differ only in the least significant bytes, top-k is now guaranteed to always give a correct answer.

The function has also been optimized for short and long sequences.

top_k_per_row can now run for any top_k value configured at runtime.

FIX https://github.com/vllm-project/vllm/issues/26554

## Test Plan

The test test_top_k_per_row has been amplified to cover these cases too. They didn't pass before and now they pass.

## Test Result

---
<details>

</details>


csrc/sampler.cu
```diff
@@ -111,173 +252,185 @@ __device__ void topKPerRowJob(const float* logits, const int rowStart,
...
+          auto otherLogit = smemFinal.items.logits[j];
+          if (logit < otherLogit || (logit == otherLogit && i < j)) {
+            outIndex++;
+          }
+        }
+        // Store if outIndex is in bounds
+        if (outIndex + baseIdx < kTopK) {
+          smemIndices[outIndex + baseIdx] = smemFinal.items.indices[i];
+        }
       }
```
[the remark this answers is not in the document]
dcampora: It is insertion sort, and this comment is out of touch. It can be selected at compile time.

csrc/sampler.cu
```diff
@@ -340,27 +497,48 @@ void apply_repetition_penalties_(
...
+
+  if (numRows > kSortingAlgorithmThreshold) {
+    int numRadixBlocks = numRows - kSortingAlgorithmThreshold;
+    vllm::topKPerRow<kNumThreadsPerBlock, true>
+        <<<numRadixBlocks, kNumThreadsPerBlock, 0, stream>>>(
+            logits.data_ptr<float>(), rowStarts.data_ptr<int>(),
+            rowEnds.data_ptr<int>(), indices.data_ptr<int>(),
+            static_cast<int>(stride0), static_cast<int>(stride1),
+            kSortingAlgorithmThreshold);
+  }
```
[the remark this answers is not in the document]
dcampora: That inefficient quadratic time sort is more efficient for low len arrays.

csrc/sampler.cu
```diff
@@ -111,173 +252,185 @@ __device__ void topKPerRowJob(const float* logits, const int rowStart,
...
-// Store to global memory.
+  if constexpr (!sortResultLogitDescending) {
+    // Store to global memory.
 #pragma unroll
-  for (int ii = 0; ii < kNumTopKItemsPerThread; ++ii) {
-    int offset = rowIdx * kTopK + ii * kNumThreadsPerBlock + threadIdx.x;
-    outIndices[offset] =
-        smemIndices[ii * kNumThreadsPerBlock + threadIdx.x] - rowStart;
+    for (int i = threadIdx.x; i < kTopK; i += kNumThreadsPerBlock) {
+      outIndices[i] = smemIndices[i];
```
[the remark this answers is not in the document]
dcampora: Fixed.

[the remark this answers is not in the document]
dcampora: Fixed.

mgoin:
cc @youkaichao @LucasWilkinson @MatthewBonanni

vllm/model_executor/models/deepseek_v2.py
```diff
@@ -629,14 +629,28 @@ def sparse_attn_indexer(
...
             logits,
             next_n,
             decode_metadata.seq_lens,
             topk_indices,
+            aux_indices,
+            aux_logits,
             num_rows,
             logits.stride(0),
             logits.stride(1),
+            splitWorkThreshold,
```
mgoin: Why pass in this value if it is constant?

vllm/model_executor/models/deepseek_v2.py
```diff
@@ -629,14 +629,28 @@ def sparse_attn_indexer(
...
+        splitWorkThreshold = 200 * 1000
+        aux_indices = torch.empty((0,), dtype=torch.int32, device=logits.device)
+        aux_logits = torch.empty((0,), dtype=torch.float32, device=logits.device)
+        if logits.shape[1] >= splitWorkThreshold:
+            aux_indices = torch.empty(
+                (num_rows, 10 * topk_tokens), dtype=torch.int32, device=logits.device
+            )
+            aux_logits = torch.empty(
+                (num_rows, 10 * topk_tokens), dtype=torch.float32, device=logits.device
+            )
```
mgoin: Maybe this potential allocation could happen within the top_k_per_row_decode op so we can have one source of truth for here and the test?

csrc/sampler.cu
mgoin, line 1: nit: do these topK functions actually belong in sampler.cu? It seems they are only used for deepseekv3.2 and don't have much shared in common with other functions here

dcampora: Could you please suggest a more appropriate location for this functionality? I'm not very familiar with the code structure of torch ops in vllm.

vllm/model_executor/models/deepseek_v2.py
dcampora: I have moved the allocation to `top_k_per_row_decode`. This implicitly assumes however that the memory will not be further allocated and used by another stream. To be concrete: `outIndicesAux` and `outLogitsAux` are now allocated with `torch::empty` inside `top_k_per_row_decode`, two async kernels are invoked, and the buffers are freed likely before the kernels finish execution. If those memory locations are afterwards allocated for something else and executed in a separate stream, this is a bug - but if that's guaranteed to not happen, then it isn't. Can we make this assumption here?

vllm/model_executor/models/deepseek_v2.py
dcampora: Fixed.

vllm/model_executor/models/deepseek_v2.py
dcampora: One possible solution is to use https://docs.pytorch.org/docs/stable/generated/torch.Tensor.record_stream.html .

csrc/sampler.cu
mgoin: We can just leave it for now, I don't have a strong idea

mgoin (approved):
LGTM! Can you please post an eval to make sure deepseek v3.2 is still good quality?

csrc/sampler.cu
```diff
@@ -44,41 +44,301 @@ __global__ void apply_repetition_penalties_kernel(
...
-                                       kNumTopKItemsPerThread, int>;
+template <int step>
+static inline __device__ uint32_t extractBinIdx(float x) {
+  if constexpr (step == 0) {
+    __half hx = __float2half(x);
+    uint16_t bits = __half_as_ushort(hx);
+    bits = (bits & 0x8000) ? bits : ~bits & 0x7fff;
+    return bits >> 5;
+  } else {
+    return bits & 0x3ff;
```
MARD1NO: Is this direct return code a bug ?

dcampora: Thanks, fixed.

PR #27569, opened 2025-10-27
[Bugfix][P/D] Fix throughput stats in disaggregated setup

NickLucche:
Fix prompt throughput stats in CLI logger by only accounting for tokens that were prefilled locally.

In a P/D setup, kv cache is copied over from P to D, and this currently resolves in outputting the following on the Decoder side:
```
# Sending 2 different reqs one after another
(APIServer pid=3322894) INFO:     Started server process [3322894]
(APIServer pid=3322894) INFO:     Waiting for application startup.
(APIServer pid=3322894) INFO:     Application startup complete.
(APIServer pid=3322894) INFO:     127.0.0.1:38042 - "POST /v1/completions HTTP/1.1" 200 OK
(APIServer pid=3322894) INFO 10-27 11:21:40 [loggers.py:208] Engine 000: Avg prompt throughput: 53.2 tokens/s, Avg generation throughput: 15.0 tokens/s, Running: 0 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.0%, Prefix cache hit rate: 0.0%, External prefix cache hit rate: 50.0%
(APIServer pid=3322894) INFO 10-27 11:21:50 [loggers.py:208] Engine 000: Avg prompt throughput: 10.3 tokens/s, Avg generation throughput: 11.2 tokens/s, Running: 1 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.1%, Prefix cache hit rate: 0.0%, External prefix cache hit rate: 50.0%
```
which is plain wrong given we have not actually prefilled those tokens, but actually just "copied" them over.

After this PR:
```
# Same setup
(APIServer pid=3318553) INFO:     Started server process [3318553]
(APIServer pid=3318553) INFO:     Waiting for application startup.
(APIServer pid=3318553) INFO:     Application startup complete.
(APIServer pid=3318553) INFO:     127.0.0.1:54610 - "POST /v1/completions HTTP/1.1" 200 OK
(APIServer pid=3318553) INFO 10-27 11:15:05 [loggers.py:208] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 15.0 tokens/s, Running: 0 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.0%, Prefix cache hit rate: 0.0%, External prefix cache hit rate: 50.0%

(APIServer pid=3318553) INFO:     127.0.0.1:35638 - "POST /v1/completions HTTP/1.1" 200 OK
(APIServer pid=3318553) INFO 10-27 11:15:15 [loggers.py:208] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 15.0 tokens/s, Running: 0 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.0%, Prefix cache hit rate: 0.0%, External prefix cache hit rate: 50.0%
```


vllm/v1/core/sched/scheduler.py
```diff
@@ -532,6 +532,7 @@ def schedule(self) -> SchedulerOutput:
                     self._update_connector_prefix_cache_stats(
                         request, num_external_computed_tokens
                     )
+                    request.num_external_computed_tokens += num_external_computed_tokens
```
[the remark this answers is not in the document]
tlrmchlsmth: +1, why is it `+=` instead of `=`?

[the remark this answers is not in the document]
NickLucche: @tlrmchlsmth the flow is WAITING (get `num_external_computed_tokens`)=>WAITING_FOR_REMOTE_KVS=>WAITING where `request.num_computed_tokens=N-1` (*prefill done*, to figure as if those tokens were processed). At this point `num_external_computed_tokens=0` and assigning again overwrites.

I've moved the assignment in the first WAITING iteration so that it's clearer + it will only go through that once

vllm/v1/core/sched/scheduler.py
```diff
@@ -1496,6 +1493,8 @@ def _update_requests_with_invalid_blocks(
                         request.num_computed_tokens - request.num_cached_tokens
                     )
                     request.num_computed_tokens = request.num_cached_tokens
+                # Prefill is to be recomputed locally.
+                request.num_external_computed_tokens = 0
```
NickLucche: @sdavidbd can you please double check this, my understanding is that we have to re-compute the whole prefill now so we can track prompt throughput

markmc: Even from the docstring, it seems clear we don't always re-compute the whole prefill:

>  This method scans the given requests, detects those with invalid blocks and adjusts their `num_computed_tokens` to the longest valid prefix.

A few things:

1. I think `num_external_computed_tokens = 0` should only happen inside the `not marked_invalid_block` clause - here we're saying all externally computed blocks are invalid
2. (Unrelated to this PR - an observation) Setting `request.num_computed_tokens = request.num_cached_tokens` on line 1489 doesn't make sense to me - since `num_cached_tokens` includes both local and external computed tokens?
3. We should update `num_external_computed_tokens` at `# Truncate the computed tokens at the first failed block` - to something like `request.num_computed_tokens - local_computed_tokens` (but not obvious how we calculated `local_computed_tokens`)

vllm/v1/engine/__init__.py
```diff
@@ -121,6 +121,8 @@ class EngineCoreOutput(
     trace_headers: Mapping[str, str] | None = None
     # The number of tokens with prefix cache hits.
```
markmc, line 122: Yeah, this comment looks incorrect ... assuming "prefix cache" refers to the local cache?

```
                    # Total computed tokens (local + external).                                                                                                                 
                    num_computed_tokens = (
                        num_new_local_computed_tokens + num_external_computed_tokens
                    )
                ...
                # Count the number of prefix cached tokens.                                                                                                                     
                if request.num_cached_tokens < 0:
                    request.num_cached_tokens = num_computed_tokens
```

vllm/v1/request.py
```diff
@@ -118,9 +118,12 @@ def __init__(
         # indicates that the output is corrupted
         self.num_nans_in_logits = 0
 
-        # The number of requests being preempted by the scheduler
+        # The number of requests being preempted by the scheduler.
```
markmc: Completely unrelated to this PR, but since you modified the line ... :rofl: 

This is actually the number of times this request was preempted?

vllm/v1/metrics/stats.py
```diff
@@ -221,6 +221,8 @@ def __init__(self):
         self.num_generation_tokens = 0
         self.num_prompt_tokens = 0
         self.num_preempted_reqs = 0
+        # Num of prompt tokens that have been computed locally.
```
markmc, line 223: Is the naming here a big confusing? By "computed locally" here we mean both computed and locally cached?

If you just tracked `num_external_computed_tokens` and then subtracted it in `_track_iteration_stats()` would that be more clear?

vllm/v1/engine/__init__.py
```diff
@@ -121,6 +121,8 @@ class EngineCoreOutput(
...
     num_cached_tokens: int = 0
+    # The number of tokens that have been computed remotely.
+    num_external_computed_tokens: int = 0
```
markmc, line 125: I'd be tempted to refactor these two into a `PrefillStats` object ... and only include that in the `ECO` when the prefill completes ... especially if we ever wanted to also send like `num_locally_cached_tokens` too

vllm/v1/metrics/loggers.py
```diff
@@ -113,7 +113,7 @@ def _reset(self, now):
 
     def _track_iteration_stats(self, iteration_stats: IterationStats):
```
markmc, line 122: Presumably you want to update the Prometheus metric too?

vllm/v1/core/sched/scheduler.py
sdavidbd: @NickLucche — as @markmc noted, we don’t recompute the entire prefill. Only the externally computed tokens starting from the first failed block are recomputed.

To correctly update `num_external_computed_tokens`, we should first determine how many externally computed tokens are affected. This can be derived from the delta between the original and truncated `num_computed_tokens` — the same tokens already aggregated in `total_affected_tokens` (lines 1473–1477):

```
# Truncate the computed tokens at the first failed block
request.num_computed_tokens = idx * self.block_size
num_affected_tokens = req_num_computed_tokens - request.num_computed_tokens
total_affected_tokens += num_affected_tokens
request.num_external_computed_tokens -= num_affected_tokens
```

sdavidbd: @markmc — regarding your points:

1. The `not marked_invalid_block` condition covers the sync-loading edge case where a request is affected by externally computed tokens that failed to load but are shared with preceding requests that will handle their recomputation. In this situation, the affected request still treats those tokens as locally computed, so its `num_external_computed_tokens` **remains unchanged**.

For example, assuming `block_size = 1` and the following prompts (with `R1`receding `R2` in the batch):
```
R1: t1 t2 t3
R2: t1 t2 t4 t5
```

Suppose `t1` is locally computed, `t2` and `t4` are externally computed, and `t2` fails to load while `t4` succeeds. Then:

<table>
  <caption><strong>Before failure</strong></caption>
  <thead>
    <tr>
      <th>Request</th>
      <th>num_computed_tokens</th>
      <th>num_external_computed_tokens</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>R1</td>
      <td>2</td>
      <td>1</td>
    </tr>
    <tr>
      <td>R2</td>
      <td>3</td>
      <td>1</td>
    </tr>
  </tbody>
</table>

<table>
  <caption><strong>After failure</strong></caption>
  <thead>
    <tr>
      <th>Request</th>
      <th>num_computed_tokens</th>
      <th>num_external_computed_tokens</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>R1</td>
      <td>1</td>
      <td>0</td>
    </tr>
    <tr>
      <td>R2</td>
      <td>3</td>
      <td>1</td>
    </tr>
  </tbody>
</table>

Both `R1` and `R2` are affected and will recompute `t2`, `t3` and `t5` respectively, but `R2`’s total number of computed tokens remains unchanged.

2. Correct — `num_cached_tokens` represents the total number of computed tokens (both local and external). Setting `num_computed_tokens = num_cached_tokens` ensures that all new tokens are recomputed in the current iteration, since the previous `num_computed_tokens` value already included them.

3. Agreed — see my suggested code changes above for how we update `num_external_computed_tokens `accordingly.

vllm/v1/engine/__init__.py
NickLucche: I am not familiar with it cc @chaunceyjiang

vllm/v1/engine/__init__.py
NickLucche: I don't have a strong opinion on this tbh, we can probably wait to have a few more things to bundle before executing the suggestion

vllm/v1/metrics/loggers.py
NickLucche: @markmc which one? I intentionally left `self.counter_prompt_tokens` unchanged to avoid replacing the actual prompt count.
Should I just make a new one for local tokens?

vllm/v1/metrics/stats.py
NickLucche: >By "computed locally" here we mean both computed and locally cached?

Yes the behavior is unchanged, cached ones would still result in higher throughput even in regular aggregated setup.

>If you just tracked num_external_computed_tokens and then subtracted it in _track_iteration_stats() would that be more clear?

I think looking at the diff 
```
self.num_prompt_tokens += iteration_stats.num_prompt_tokens
-->
self.num_prompt_tokens += iteration_stats.num_local_prompt_tokens
```
this is pretty clear that I just want to rule out the remote tokens ie I assume the semantic was the intended one from the beginning, it's just "local" used to be redundant

vllm/v1/core/sched/scheduler.py
```diff
@@ -426,6 +426,11 @@ def schedule(self) -> SchedulerOutput:
                             self.waiting.pop_request()
                             skipped_waiting_requests.prepend_request(request)
                             continue
+                        # Keep track of number of tokens to load from remote
+                        # for the request st we can compute actual throughput
+                        request.num_external_computed_tokens = (
+                            num_external_computed_tokens
```
markmc: Hmm, is this wrong? Isn't `num_external_computed_tokens = 0` at this point?

NickLucche: awesome catch thanks @markmc !
I was done in by a silent rebase change that had renamed the  `num_external_computed_tokens` variable..

NickLucche:
@markmc let me know if you have some other comment for this fix

markmc:
> it's wrong trying to derive throughput from them since result is bonkers

The throughput figure is just expressing a counter as a rate. In a Grafana dashboard, we do the same with e.g.

```
          "expr": "rate(vllm:prompt_tokens_total{model_name=\"$model_name\"}[$__rate_interval])",
          "legendFormat": "Prompt Tokens/Sec",

```

That's why I think "Prompt Throughput" in Prometheus/Grafana should mean the same thing as in the console log.

And that's just to say I think it's important that we can articulate a consistent mental model for what all of these metrics mean, whether in the console log or Prometheus.

Here's my mental model for how we're counting things now:

```
prompt input
   ↓ [prompt tokens]
lookup internal prefix cache
   ↓ [tokens queried and found]
lookup external connector prefix cache
   ↓ [tokens queried and found]
....
   ↓ 
generated tokens output
   ↓ [generated tokens]
```

With that mental model, this:

```
Avg prompt throughput: 53.2 tokens/s, Avg generation throughput: 15.0 tokens/s, Prefix cache hit rate: 0.0%, External prefix cache hit rate: 100.0%
```

is not bonkers at all?

53.2 prompt token/s came in at the top, none were found in the local prefix cache, 100% were retrieved from the connector, and we generated 15 tokens/s

If you see it differently, could you draw your mental model for these counters, in such a way that the same model can be applied consistently across use cases?

smarterclayton:
Roughly yes

```
vllm:prompt_tokens_total{source="cache_hit_gpu"}      # Tokens from GPU cache
vllm:prompt_tokens_total{source="cache_hit_cpu"}      # Tokens from CPU cache
```

We are likely to have more cache hit sources in the future.  I'd suggest two labels, `source=cache_hit` and `cache_source=cpu|gpu|<empty>`.  Prometheus convention would probably be to keep the labels orthogonal (I.e. `cache_source` and `source` are probably less preferable than two labels that don't share `_source`) but I don't have an alternative suggestion.

Otherwise, that structure looks like what I would expect

EDIT: One note, we should pick values for `cache_source` consistent with the actual source as defined in our cache hierarchy code, (`cpu` and `gpu` may be too generic, especially given that there are likely more levels of GPU hierarchy in future hardware generations and there may be multiple CPU tiers).  Ideally, we use a constant defined by the cache plugin hierarchy itself for cache source.

PR #27583, opened 2025-10-27
Rename clashing method names for vLLM model protocol

hmellor:
`get_input_embeddings` is a getter method of all `transformers.PreTrainedModel`s.

This name had been reused in vLLM to call `forward` on the model's embedding layer.

Since the method in vLLM is not actually a getter and causes a confusing clash with the getter in Transformers, this PR:

- Renames `get_input_embeddings` to `embed_input_ids`
- Renames `get_multimodal_embeddings` to `embed_multimodal` for consistency
- Adds fallbacks and deprecation warnings to `VllmModel` and `SupportsMultiModal` so that third party vLLM models should still work


NickLucche (commented):
looks reasonable

hmellor:
As per discussion with @ywang96, I'll be pausing work on this until an ongoing new model is added as to not disrupt it.

DarkLight1337 (approved):
LGTM if tests pass

PR #25712, opened 2025-09-25
[Core][Hybrid allocator + kv connector 1/n] Enable hybrid allocator + KV cache connector

KuntaiDu:
## Purpose

Refactor of #25363 . This PR enables the combination of hybrid allocator + KV cache connector in a backward-compatible way.

## Test Script
```


import os

# Set token chunk size to 256
os.environ["LMCACHE_CHUNK_SIZE"] = "256"
# Enable CPU memory backend
os.environ["LMCACHE_LOCAL_CPU"] = "True"
# Set CPU memory limit to 5GB
os.environ["LMCACHE_MAX_LOCAL_CPU_SIZE"] = "20.0"
os.environ["VLLM_ENABLE_V1_MULTIPROCESSING"] = "0"
os.environ["LMCACHE_USE_LAYERWISE"] = "True"


from vllm import LLM, SamplingParams
from vllm.config import KVTransferConfig

# Configure KV cache transfer to use LMCache
ktc = KVTransferConfig(
    kv_connector="LMCacheConnectorV1",
    kv_role="kv_both",
)

# Initialize LLM with LMCache configuration
# Adjust gpu_memory_utilization based on your GPU memory
llm = LLM(model="google/gemma-3-4b-it",
          kv_transfer_config=ktc,
          max_model_len=75000,
          gpu_memory_utilization=0.18,
          enforce_eager=True)

# Define sampling parameters
sampling_params = SamplingParams(temperature=0, top_p=0.95, max_tokens=10)

# Run inference
outputs = llm.generate("hi" * 70000 + "\nhow are you?", sampling_params)
generated_text = outputs[0].outputs[0].text
print(f"Generated text: {generated_text!r}")

# This requires loading KV cache and will success
outputs = llm.generate("hi" * 10000 + "\nTell me a story.", sampling_params)
generated_text = outputs[0].outputs[0].text
print(f"Generated text: {generated_text!r}")

# flush out prefix cache in GPU
outputs = llm.generate("1" + "hi" * 70000 + "\nhow are you?", sampling_params)
generated_text = outputs[0].outputs[0].text
print(f"Generated text: {generated_text!r}")

# This requires loading KV cache
# but this request cannot be executed as vLLM cannot allocate for long prefix 
# stored by LMCache
outputs = llm.generate("hi" * 70000 + "\nTell me a story.", sampling_params)
generated_text = outputs[0].outputs[0].text
print(f"Generated text: {generated_text!r}")

```

## Test Result

Success.

---
<details>
[...]

PR #27592, opened 2025-10-27
[Stability fix] turn off HMA allocator when connector is set

KuntaiDu:
Due to #25712 , currently vLLM will hard fail if the user simply set a connector because vLM enables HMA by default.

To avoid hard fail, before discussing with @njhill to figure out a better solution, let's just turn off HMA when a connector is being set.

---
<details>

</details>


NickLucche (approved):
Fixes  
```
(EngineCore_DP0 pid=3595930) ValueError: Connector NixlConnector does not support HMA but HMA is enabled. Please set `--disable-hybrid-kv-cache-manager`.
```

I think we should just change logic to disabling it if it does *not* support hma with the interface you defined, given it's on by default

vllm/config/vllm.py
```diff
@@ -597,6 +597,11 @@ def __post_init__(self):
...
                 # Hybrid KV cache manager is not supported on non-GPU platforms.
                 self.scheduler_config.disable_hybrid_kv_cache_manager = True
+            if self.kv_transfer_config is not None:
+                # NOTE(Kuntai): turn HMA off for connector for now.
+                # TODO(Kuntai): have a more elegent solution to selectively
+                # turn off HMA for connector.
+                self.scheduler_config.disable_hybrid_kv_cache_manager = True
             if self.kv_events_config is not None:
                 # Hybrid KV cache manager is not compatible with KV events.
                 self.scheduler_config.disable_hybrid_kv_cache_manager = True
```
njhill, line 616: I think we should log a warning in these cases

njhill (commented):
Thanks @KuntaiDu

vllm/config/vllm.py
KuntaiDu: Fixed.

njhill (approved):
Thanks @KuntaiDu

ApostaC (approved):
LGTM!

PR #20059, opened 2025-06-25
[Core] Allow full cudagraph with separate attention routines and orthogonal to compilation, add support for FA2 and FlashInfer

fhl2000:
## What's changed in this PR:  

1. We allow cudagraph logic to be orthogonal to VLLM compilation in v1. In other words, full cudagraph without compilation is supported in v1 now;  
2. Full cudagraph support for Flash Attention v2 and FlashInfer (moved to [#21367](https://github.com/vllm-project/vllm/pull/21367) with further optimizations);
3. New CLI flag `cudagraph_mode` is introduced in CompilationConfig, supporting the following five modes: `NONE`, `PIECEWISE`, `FULL`, `FULL_DECODE_ONLY`, and  `FULL_AND_PIECEWISE`, which will replace/deprecate the original two flags `use_cudagraph` and `full_cuda_graph`.   The most powerful mode would be `FULL_AND_PIECEWISE`, which reduced ITL of decode-phase by full cudagraph and retains small TTFT of piecewise cudagraph for prefill-or-mixed-phase and also keeps generality;
4. Cudagraph modes related to full cudagraph are compatible with the validation phase of speculative decode (only FA2/3, Triton attention now; FlashMLA in a follow-up PR);
5. Ideally, the new infrastructure will speed up most attention backends that support full cudagraph. The ITL for Triton attention was reduced by 38% for small LLMs (`FULL_AND_PIECEWISE` mode vs. piecewise cudagraph of main branch, see [comment](https://github.com/vllm-project/vllm/pull/20059#issuecomment-3160858458) for details).

---

### Motivation
See the original PR description below. Also, reference @ProExpertProg 's [long proposal](https://github.com/vllm-project/vllm/pull/20059#issuecomment-3086829014) and the RFC #20283.

### Overall designs:
The new CUDA Graph logic is built on top of piecewise compilation and supports dual cudagraph runtime mode switching. To make the system work, there are three core components (i.e., `CUDAGraphMode`, `CUDAGraphWrapper`, and `CUDAGraphDispatcher`) and two auxiliary concepts (i.e., `BatchDescriptor` and `AttentionCGSupport`) introduced in this PR. Let's pick up through them one by one: 

Note: For convenience, hereafter we refer to pure decode (max_query_len=1) or speculative decode (max_query_len =1+num_spec_tokens) as uniform decode batches, and the opposite would be non-uniform batches (i.e., prefill or mixed prefill-decode batches).

#### CUDAGraphMode
[...]

PR #27593, opened 2025-10-27
Remove deprecated fields from `CompilationConfig`

hmellor:
- `use_cudagraph` and `full_cuda_graph` have been deprecated since https://github.com/vllm-project/vllm/pull/20059
- They were scheduled for removal in `v0.11.0`, which has already been released
- Config test was updated because it didn't actually pass (I don't think it was run in CI before)


tests/compile/piecewise/test_multiple_graphs.py
```diff
@@ -203,7 +203,6 @@ def test_multi_graph_piecewise_compile(use_inductor_graph_partition: bool):
     vllm_config = VllmConfig(
         compilation_config=CompilationConfig(
             mode=CompilationMode.VLLM_COMPILE,
-            use_cudagraph=True,
```
ProExpertProg, line 206: Can you make mode explicitly Piecewise here?

ZJY0516:
CC @fhl2000

fhl2000 (approved):
LGTM! Thanks for doing the deprecations!

simon-mo (approved):
stamp

vllm/config/vllm.py
```diff
@@ -720,6 +712,12 @@ def _set_cudagraph_sizes(self):
                 # de-duplicate the sizes provided by the config
                 dedup_sizes = list(set(self.compilation_config.cudagraph_capture_sizes))
                 cudagraph_capture_sizes = dedup_sizes
+                # filter out sizes larger than max_cudagraph_capture_size
```
ProExpertProg: Why does this have to be added? Bad merge?

ProExpertProg (approved):

tests/compile/piecewise/test_multiple_graphs.py
hmellor: Sure, I don't think it was piecewise before though

vllm/config/vllm.py
hmellor: Without it, https://github.com/vllm-project/vllm/blob/a51d54d1e42508db0c3f008d2c43fe6622e75bea/tests/compile/test_config.py#L273-L274 fails

fhl2000: Ok,  though I am not sure why this did not fail in my previous PR, while it fails in this PR, I guess you want to filter out sizes larger than max_num_tokens, not max_cudagraph_capture_size.   The reason for not max_cudagraph_capture_size is that it could be a user-specific value, which we should keep separated from user-specific `cudagraph_capture_sizes` here, so then we can detect whether they are consistent with each other.

hmellor: This particular test case also failed because the number `max_num_batched_tokens < max_num_seqs`. I suspect that this test didn't actually run in the previous PR.

hmellor: Yep I've just checked, this test doesn't run in CI... ever

tests/compile/piecewise/test_multiple_graphs.py
ProExpertProg: Yeah I realize that, we should have updated earlier I think so we're not relying on the default. But it's fine to leave as well

hmellor: Happy to change it, I just wasn't sure what the correct setting was for this test

vllm/config/vllm.py
ProExpertProg: This logic is not great because if the user specifies `cudagraph_capture_sizes` manually, any size above `max_cudagraph_capture_size` (set to the default value which is minimum of `max_num_seqs*2`, `max_num_batched_tokens`, and `512` will get truncated, which is not what we want. Instead, we want to adjust `max_cudagraph_capture_size` to be the largest value in the list, only limited by `max_num_batched_tokens`. This is my bad, missed it while reviewing @fhl2000's PR.

Also can you add `test_config` to CI while you're at it?

fhl2000: > Instead, we want to adjust max_cudagraph_capture_size to be the largest value in the list, only limited by max_num_batched_tokens

Yeah, this is also what I meant before.

>This is my bad, missed it while reviewing @fhl2000's PR.

I think the logic in the previous pr is already this: when users only set a specific one of two values, the other one should follow the user-specific one;  if both of the two are user-specific values, then we check their compatibility.

>This particular test case also failed because the number max_num_batched_tokens < max_num_seqs. I suspect that this test didn't actually run in the previous PR.

For this issue, we should possibly check `max_num_batched_tokens  >= max_num_seqs`  in the schedule config? So we won't worry about it again here.

ProExpertProg: @fhl2000 I'm not sure that the logic is happening? From what I read we don't limit the user specified cg sizes to max num batched tokens

fhl2000: >I'm not sure that the logic is happening? From what I read we don't limit the user specified cg sizes to max num batched tokens

My bad for misunderstanding something😂. You are right. It should be fixed in this pr.

hmellor: If `max_num_batched_tokens < max_num_seqs` do we want to coerce `max_num_seqs` lower or raise an error explaining what the user did wrong?

fhl2000: I think raising an error is good. It is not a big deal.

hmellor: Ok, in the test I'll continue to lower the `max_num_seqs` for this test case then.

vllm/config/vllm.py
```diff
@@ -738,7 +730,9 @@ def _set_cudagraph_sizes(self):
                 )
                 # de-duplicate the sizes provided by the config
                 dedup_sizes = list(set(self.compilation_config.cudagraph_capture_sizes))
-                cudagraph_capture_sizes = dedup_sizes
+                cudagraph_capture_sizes = [
+                    i for i in dedup_sizes if i <= max_cudagraph_capture_size
+                ]
```
fhl2000, line 850: I think you haven't fixed it, right? 
```python
                cudagraph_capture_sizes = [
                    i for i in dedup_sizes if i <= max_num_tokens
                ]
```

hmellor:
```py
max_cudagraph_capture_size = min(max_num_tokens, max_cudagraph_capture_size)
```

hmellor: Although this would still allow the default `max_cudagraph_capture_size` to truncate `cudagraph_capture_sizes`. So your suggestion is probably correct

tests/compile/test_config.py
```diff
@@ -83,20 +76,24 @@ def test_VLLM_DISABLE_COMPILE_CACHE(vllm_runner, monkeypatch, val):
 
 # forked needed to workaround https://github.com/vllm-project/vllm/issues/21073
 @pytest.mark.forked
-@pytest.mark.parametrize("enabled", [True, False])
-def test_use_cudagraphs(vllm_runner, monkeypatch, enabled):
+@pytest.mark.parametrize(
+    "cudagraph_mode", [CUDAGraphMode.FULL_AND_PIECEWISE, CUDAGraphMode.NONE]
+)
+def test_use_cudagraphs(vllm_runner, monkeypatch, cudagraph_mode):
```
ProExpertProg: Could we do:
```suggestion
@pytest.mark.parametrize(
    "cudagraph_mode, num_cg_captured", [
        (CUDAGraphMode.NONE, 0),
        (CUDAGraphMode.FULL_DECODE_ONLY, 1),
        (CUDAGraphMode.PIECEWISE, 13),
        (CUDAGraphMode.FULL_AND_PIECEWISE, 14),
    ]
)
def test_use_cudagraphs(vllm_runner, monkeypatch, cudagraph_mode, num_cg_captured: int):
```

tests/compile/test_config.py
```diff
@@ -189,29 +185,27 @@ def test_splitting_ops_dynamic():
             cudagraph_mode=CUDAGraphMode.PIECEWISE,
         )
     )
-    # With the new simplified logic, attention fusion works with splitting_ops
-    assert config.compilation_config.splitting_ops_contain_attention()
-    # cudagraph mode remains PIECEWISE
-    assert config.compilation_config.cudagraph_mode == CUDAGraphMode.PIECEWISE
+    assert config.compilation_config.splitting_ops == []
+    # cudagraph mode also fall back to FULL
+    assert config.compilation_config.cudagraph_mode == CUDAGraphMode.FULL
```
ProExpertProg, line 198: Could you also add back the test where we get an assertion error if `splitting_ops` are specified with attention fusion and without inductor partition?

ProExpertProg (approved):

tests/compile/test_config.py
hmellor: I don't think we can because we always override. The assertion is never triggered.

https://github.com/hmellor/vllm/blob/f76ac68eb4ac7e1c330b5a27771c22e386c6c0a8/vllm/config/compilation.py#L737-L757

hmellor: Ok I've fixed the reason the assert would never raise and added the assertion test back

PR #27594, opened 2025-10-27
Fix intermediatetensors spawn error #27591

baonudesifeizhai:
## Purpose
Fix NameError when pipeline ranks return IntermediateTensors inside torch.compile / runtime processes, preventing engine initialization failure.

Store the reference to IntermediateTensors as an instance attribute to ensure availability in multi-process and compiled graph environments.
## Test Result
https://paste.ubuntu.com/p/R544fs7kfP/ 
---
image

<details>

</details>


DarkLight1337 (commented):
Hmm, this is quite unexpected. cc @zou3519 @youkaichao

baonudesifeizhai:
RuntimeError: Worker failed with error 'name 'IntermediateTensors' is not defined', please check the stack trace above for the root cause
     if  only import from __future__ import annotations

> We also need fixes in qwen3* files
> 
> My understanding is that the import fails because of torch.compile AOT.
> 
> Also return type hint is wrong, but it should not matter here



> We also need fixes in qwen3* files
> 
> My understanding is that the import fails because of torch.compile AOT.
> 
> Also return type hint is wrong, but it should not matter here

youkaichao:
cc @zou3519

zou3519:
@baonudesifeizhai did the problem go away for you?

baonudesifeizhai:
works now   

> @baonudesifeizhai did the problem go away for you?

Issue #25993, opened 2025-09-30
[Feature]: GLM-4.6 spec decode with mtp


wuyaoxuehun:
### 🚀 The feature, motivation and pitch

vllm==0.10.2

vllm serve GLM-4.6
--speculative-config '{"method":"deepseek_mtp","num_speculative_tokens":1, "model":"GLM-4.6"}'
--tensor-parallel-size 8

gives error:
ValueError: Following weights were not initialized from checkpoint: {'model.embed_tokens.weight', 'model.layers.92.shared_head.head.weight'

This command works fine with GLM-4.5，but not in GLM-4.6，

can you support GLM-4.6？Since it's a strong model.

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

HarewVlad:
Seems like MTP is now possible for FP8 version: https://huggingface.co/zai-org/GLM-4.6-FP8/discussions/3

AlbsAgren:
Getting the same error as above when trying to run the updated FP8 version with mtp.

ehfd:
Despite using the nightly build with #27597 merged, there is an issue with the FP8 weight.
@
```
INFO 11-16 02:57:35 [scheduler.py:216] Chunked prefill is enabled with max_num_batched_tokens=2048.
(APIServer pid=1) INFO 11-16 02:57:35 [api_server.py:1977] vLLM API server version 0.11.1rc7.dev218+gf849ee739
...
(Worker_PP2_TP0_EP0 pid=739) INFO 11-16 03:19:21 [weight_utils.py:441] Time spent downloading weights for zai-org/GLM-4.6-FP8: 0.519828 seconds
(Worker_PP0_TP0_EP0 pid=694) INFO 11-16 03:19:22 [gpu_model_runner.py:3126] Model loading took 52.6607 GiB memory and 1101.690230 seconds
(Worker_PP2_TP1_EP1 pid=754) ERROR 11-16 03:19:37 [multiproc_executor.py:640] WorkerProc failed to start.
(Worker_PP2_TP1_EP1 pid=754) ERROR 11-16 03:19:37 [multiproc_executor.py:640] Traceback (most recent call last):
[...]

PR #27597, opened 2025-10-27
[Model] fix glm4_moe_mtp load weights with GLM-4.6 checkpoint.

wuyaoxuehun:
## Purpose
As described in [issue#25993](https://github.com/vllm-project/vllm/issues/25993), when serving GLM-4.6  with mtp using the following command:
```
vllm serve GLM-4.6
--speculative-config '{"method":"deepseek_mtp","num_speculative_tokens":1, "model":"GLM-4.6"}'
--tensor-parallel-size 8
```
It raises the following error:
```
ValueError: Following weights were not initialized from checkpoint: {'model.embed_tokens.weight', 'model.layers.92.shared_head.head.weight'.
```

The root cause is that the [GLM-4.6](https://huggingface.co/zai-org/GLM-4.6/tree/main) checkpoint doesn't include the shared parameters (model.layers.92.shared_head.head.weight shared with lm_head.weight, model.layers.92.embed_tokens.weight with model.embed_tokens.weight) to its weights list, while the [GLM-4.5](https://huggingface.co/zai-org/GLM-4.5/tree/main) checkpoint does.

This PR fixes the issue by mapping these two parameters in the checkpoint weight list to their corresponding local model weight names.

## Test Plan
Run the following command to serve GLM-4.6 with MTP:
```
vllm serve GLM-4.6
--speculative-config '{"method":"deepseek_mtp","num_speculative_tokens":1, "model":"GLM-4.6"}'
--tensor-parallel-size 8
```

## Test Result
After applying the fix, GLM-4.6 with MTP works correctly. The model speeds up as expected and returns normal responses.


wuyaoxuehun:
> Anyone able to merge this?

@houseroad 
hi houseroad, can you merge this?

ywang96 (approved):
Sorry that this PR got left in dust - thanks for your contribution!

PR #27598, opened 2025-10-27
Install pre-built xformers-0.0.32.post2 built with pt-2.9.0

huydhn:
## Purpose

Instead of waiting for xformers to release a new version for PyTorch 2.9.0, I have built `0.0.32.post2` locally and made the wheel available.

For more context, we don’t want to wait for `xformers` package for 2.9 to become available.  So, I opt to build it from source.  This works for CI, but has several issues like (1) increasing build time and (2) not listed as a dependency in `cuda.txt`.  So, installing a pre-built wheel would help in the meantime until there is a new xformers version

@ywang96


huydhn:
> I was trying to install editable - but yea I turned on ready label so you can inspect

I had a mental lapse lol, `xformers==0.0.33+5d4b92a5.d20251026` is the right way to get the package

simon-mo (approved):

PR #27606, opened 2025-10-27
[Feature] Refactor batch invariant fp8 DeepGEMM

yewentao256:
## Purpose

Fixes https://github.com/vllm-project/vllm/pull/27127#discussion_r2442431277 @mgoin 

We can reuse the code from fp8_utils to simplify the logic


vllm/model_executor/layers/quantization/fp8.py
```diff
@@ -546,83 +543,19 @@ def apply(
...
-
-                # Expand scale to match weight dimensions
-                # scale_expanded should have shape [N, K]
-                scale_expanded = weight_scale.repeat_interleave(
-                    block_n, dim=0
-                ).repeat_interleave(block_k, dim=1)
-                # Trim to exact weight size (in case of padding)
-                scale_expanded = scale_expanded[:N, :K]
-                weight_bf16 = weight_fp8 * scale_expanded
+                return self.w8a8_block_fp8_linear.apply(
```
mgoin, line 556: Did you update the code in w8a8_block_fp8_linear to have the same changes around DeepGEMM already?

yewentao256: Sorry not fully understand your question, same changes around DeepGEMM? @mgoin

yewentao256:
This PR could also fix https://github.com/vllm-project/vllm/issues/28249

mgoin (approved):

vllm/model_executor/layers/quantization/fp8.py
yewentao256: Talked offline

PR #27609, opened 2025-10-27
[VLLM Qwen] Add model.lm_head to skipped prefixes

FlaurencyaCiputra:
Differential Revision: D85581719


DarkLight1337:
Hmm, can you link to the HF repo which you are trying to load? The weight names of the official repo should not require this mapping.

DarkLight1337:
Can you address my above comment?

ywang96 (commented):
Could you please address the precommit issue?

We should also add a note here on why the `model.lm_head` skip is needed. AFAIK for OS Qwen2 models we should not need this change. See https://huggingface.co/Qwen/Qwen2-7B-Instruct/tree/main?show_file_info=model.safetensors.index.json

PR #27611, opened 2025-10-27
`VLLM_USE_TRITON_FLASH_ATTN` V0 variable deprecation

AndreasKaratzas:
Deprecated the `VLLM_USE_TRITON_FLASH_ATTN` environment variable. This variable was used for ROCm platform and was V0 specific. Therefore, we deprecated it.


AndreasKaratzas:
@tjtanaa Thank you for your feedback. Let me address your points:

**Point 1:** *"This variable cannot be deprecated. It is used in the Triton MLA backend for V1. It is a valid code path."*

**Response:** You're correct that it's a valid code path, but it's **ROCm-specific only**. The variable is used exclusively in `triton_mla.py`:
https://github.com/vllm-project/vllm/blob/main/vllm/v1/attention/backends/mla/triton_mla.py#L124

Every usage of `self.use_triton_flash_attn` in that file is guarded by `current_platform.is_rocm()` checks. This means both `self.use_triton_flash_attn` and `self.triton_fa_func` only exist on ROCm platforms.

---

**Point 2:** *"ROCm exclusive flags have ROCM_ in it. It is a multi-platform feature."*

**Response:** While MLA prefill is indeed multi-platform, this specific flag is ROCm-only in practice. The backend registry proves this:
https://github.com/vllm-project/vllm/blob/main/vllm/attention/backends/registry.py#L43


This points to the general `triton_mla.py` Impl, NOT to ROCm-specific:
https://github.com/vllm-project/vllm/blob/main/vllm/attention/ops/triton_flash_attention.py

Users on other platforms cannot access this Triton flash attention kernel path at all.

---

**Conclusion:** The flag should have been prefixed with `ROCM_` for clarity, but it is ROCm-specific and since the team's focus is on AITER MLA, we can deprecate this V0 backend.

tjtanaa (approved):
LGTM. As long as the tests still passed. Thank you for cleaning up the old flags.

PR #27613, opened 2025-10-27
add a warmup mode in CUDAGraphMode

bangshengtang:
Summary: during cuda graph capturing warm-up time, we need to call the underlying model in eager mode but switch on code paths for cuda graphs, previoulsy there was no good way to distinguish real eager mode run vs. warmup runs.

Test Plan: CI

Differential Revision: D85613194

### DIFF
--- a/vllm/config/compilation.py
+++ b/vllm/config/compilation.py
@@ -54,6 +54,12 @@ class CUDAGraphMode(enum.Enum):
     FULL = 2
     FULL_DECODE_ONLY = (FULL, NONE)
     FULL_AND_PIECEWISE = (FULL, PIECEWISE)
+    # WARMUP is used for the dummy runs prior to CUDA graph capture.
+    # This is to differentiate from NONE because in some use cases, static
+    # and dynamic shapes are two different code paths. We need to ensure
+    # static shapes code path is called during CUDA graph capture, and dynamic
+    # shapes are used during eager mode calls.
+    WARMUP = 3
 
     def decode_mode(self) -> "CUDAGraphMode":
         return CUDAGraphMode(self.value[0]) if self.separate_routine() else self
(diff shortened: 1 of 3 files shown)


ProExpertProg (changes requested):
Why is none not sufficient for warmup? The mechanism was designed specifically with warmup in mind to simplify the logic.

vllm/config/compilation.py
```diff
@@ -54,6 +54,7 @@ class CUDAGraphMode(enum.Enum):
     FULL = 2
     FULL_DECODE_ONLY = (FULL, NONE)
     FULL_AND_PIECEWISE = (FULL, PIECEWISE)
```
zhuohan123, line 56: Let's add in comment why this "WARMUP" mode is needed as well?

fhl2000:
Agreed to distinguish between NONE and WARMUP. But if we have to add a new runtime mode `WRAMUP`, I think basically we need to ensure this mode is not a user-visible option. Like, either separate out the runtime mode from the general user-facing  CUDAGraphMode, or at least safely mark the  `WRAMUP` mode as non-user-facing.  

Apart from the above, we also need to make corresponding changes to CUDAGraphWrapper (and UBatchWrapper for dbo) to make them aware of this mode. (I think cudagraph_dispatcher doesn't require changes, correct me if I'm wrong)

bangshengtang:
@fhl2000 thanks for your comments!

so here's an example

```
def forward(self, args):
  if mode == warmup:
    run static shape code path to warm up triton kernels, JIT kernels, preallocate buffers etc.
  elif mode == cuda_graph:
    # we're in cuda graph capturing mode, operations are limited
    run warmed-up static shape code path
  else:
    # mode == none
    run dynamic shape code path
```

I agree that we can use `max_cudagraph_sizes` to differ the cases in some cases if we can infer bsz or seqlen from the args, but that's not always the case.

PR #27614, opened 2025-10-27
[Core] Enable async scheduling by default

### LINKED ISSUE #27679: Async Scheduling Plan
- [x] https://github.com/vllm-project/vllm/pull/27615
- [x] https://github.com/vllm-project/vllm/pull/27756
- [x] https://github.com/vllm-project/vllm/pull/26866
- [x] https://github.com/vllm-project/vllm/pull/27910
- [x] https://github.com/vllm-project/vllm/pull/27922
- [x] https://github.com/vllm-project/vllm/pull/27648
- [x] https://github.com/vllm-project/vllm/pull/28012 - perf fix for regression in `#26866`
- [x] https://github.com/vllm-project/vllm/pull/28250
- [x] https://github.com/vllm-project/vllm/pull/28706
- [x] https://github.com/vllm-project/vllm/pull/28744
- [x] https://github.com/vllm-project/vllm/pull/28787
- [x] https://github.com/vllm-project/vllm/pull/24799
- [x] https://github.com/vllm-project/vllm/pull/29355
- [x] https://github.com/vllm-project/vllm/pull/29372
- [x] Address general duplicate request ids issue:
    - https://github.com/vllm-project/vllm/pull/27987
- [x] https://github.com/vllm-project/vllm/pull/31332
- [x] https://github.com/vllm-project/vllm/pull/31373
- [x] https://github.com/vllm-project/vllm/pull/27614

- [ ] Follow-on async scheduling + spec decode hardening/compatibility issues (in parallel)
   - [x] https://github.com/vllm-project/vllm/pull/29223
   - [ ] https://github.com/vllm-project/vllm/pull/29821
- [ ] Explore Async Scheduling + Pipeline Parallel

njhill:
All blockers have now been resolved: https://github.com/vllm-project/vllm/issues/27679 and passes extended CI tests.

### DIFF
--- a/vllm/config/scheduler.py
+++ b/vllm/config/scheduler.py
@@ -130,11 +130,12 @@ class SchedulerConfig:
     and starting configuration.
     """
 
-    async_scheduling: bool = False
-    """If set to True, perform async scheduling. This helps to avoid gaps in
-    GPU utilization, leading to better latency and throughput.
-    Async scheduling is currently not supported with some features such as
-    speculative decoding and pipeline parallelism.
+    async_scheduling: bool = Field(default=None)
+    """If set to False, disable async scheduling. Async scheduling helps to
+    avoid gaps in GPU utilization, leading to better latency and throughput.
+    It is currently not supported with some features such as
+    speculative decoding and pipeline parallelism, and will be automatically
+    disabled in those cases.
     """
 
     stream_interval: int = Field(default=1, ge=1)
--- a/vllm/config/vllm.py
+++ b/vllm/config/vllm.py
@@ -552,7 +552,7 @@ class VllmConfig:
                 if self.speculative_config.method not in get_args(EagleModelTypes):
                     raise ValueError(
                         "Currently, async scheduling is only supported "
-                        "with EAGLE/MTP kind of speculative decoding"
+                        "with EAGLE/MTP kind of speculative decoding."
                     )
                 if self.speculative_config.disable_padded_drafter_batch:
                     raise ValueError(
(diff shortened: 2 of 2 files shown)


DarkLight1337 (commented):
Last week I found that async scheduling isn't compatible with pooling models so you should turn that off when the runner type is pooling

vllm/config/scheduler.py
```diff
@@ -130,11 +130,12 @@ class SchedulerConfig:
     like full attention and sliding window attention.
     """
 
-    async_scheduling: bool = False
-    """EXPERIMENTAL: If set to True, perform async scheduling. This may help
-    reduce the CPU overheads, leading to better latency and throughput. However,
-    async scheduling is currently not supported with some features such as
-    structured outputs, speculative decoding, and pipeline parallelism.
+    async_scheduling: SkipValidation[bool] = None  # type: ignore
```
hmellor: Once https://github.com/vllm-project/vllm/pull/26519 merges, this can be updated to be 

```suggestion
    async_scheduling: bool = Field(default=None)
```

And `"async_scheduling"` can be added to `_skip_none_validation`

njhill: OK I updated this, though tbh I'm not sure I understand why it's better! (I think the `#type: ignore` could have been removed anyhow..

DarkLight1337:
Just sending a few prompts is not enough to test async scheduling with pooling models. I only ran into errors when running online benchmarks.

njhill:
@DarkLight1337 do you have an example of the specific benchmark you hit this with? I was mainly going on the fact that all the CI pooling tests seem to work fine.

DarkLight1337:
Ok nvm, it seems that this works now after running the benchmarks on latest main.

noooop:
https://github.com/noooop/snippet/blob/main/benchmarks/embed_async/profile.py
image

https://github.com/noooop/snippet/blob/main/benchmarks/embed_async/v1_offline.py

image

After some tests, I found that whether to enable async scheduling has almost no effect on the pooling model.

noooop:
Async scheduling, as I understand it, can completely eliminate the gaps between GPU operators, thereby improving GPU utilization. （The figure comes from my own very crude implementation of async scheduling.）

image

However, in this PR, there's no noticeable difference in GPU utilization between the two.

https://github.com/noooop/snippet/blob/main/benchmarks/embed_async/profile.py

image

njhill:
@noooop yes async scheduling is primarily targeted at decode; it should also provide _some_ benefit to prefill-only but I guess it's not a surprise if it's negligible. Are you using latest main with all the fixes? Other suggestions are trying larger batch size, try with uni vs mp executor backend.

I suspect it may be related to the fact that we don't currently overlap the CPU->GPU copy of new input tokens with the prior forward pass. I do actually have a branch where I was experimenting with this, but it didn't show significant gain for the benchmark that I was trying (perhaps because it was decode heavy), and there's a nontrivial complexity tradeoff. We could also check whether this benefits prefill-only workload though.

vllm/distributed/device_communicators/shm_object_storage.py
```diff
@@ -342,8 +342,8 @@ def __init__(self):
         from vllm.v1.serial_utils import MsgpackDecoder, MsgpackEncoder
 
         self.encoder = MsgpackEncoder()
-        self.tensor_decoder = MsgpackDecoder(torch.Tensor)
-        self.mm_decoder = MsgpackDecoder(MultiModalKwargsItem)
+        self.tensor_decoder = MsgpackDecoder(torch.Tensor, share_mem=False)
```
DarkLight1337, line 345: cc @dongluw do you have any concerns with this change?

njhill: (not planning to include this change in this PR, just adding it temporarily to ensure the CI now passes)

mgoin (approved):
LGTM, let's go for it

DarkLight1337:
It appears that "macOS Apple Silicon Smoke Test" has been failing on main since this PR.

PR #27615, opened 2025-10-27
[AsyncScheduling] Make async overlap work with logprobs

njhill:
Async scheduling already "works" with logprobs from a functional pov, but there's unnecessarily sync which negates much/all of the benefit.

This does not yet cover prompt logprobs, only output logprobs.


vllm/v1/worker/gpu_model_runner.py
```diff
@@ -202,6 +210,10 @@ def get_output(self) -> ModelRunnerOutput:
 
         output = self._model_runner_output
         output.sampled_token_ids = valid_sampled_token_ids
+        if self._logprobs_tensors_cpu:
+            # NOTE(nick): this will need to be updated to use cu_num_accepted_tokens
+            # for async sched + spec decode + logprobs compatibility.
+            output.logprobs = self._logprobs_tensors_cpu.tolists()
```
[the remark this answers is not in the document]
njhill: async scheduling is not yet compatible with spec decoding

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2334,11 +2346,9 @@ def _bookkeeping_sync(
                     cu_num_accepted_tokens[-1] + len(sampled_ids)
                 )
 
-        # NOTE: GPU -> CPU Sync happens here.
-        # Move as many CPU operations as possible before this sync point.
```
Jialin, line 2338: nit: maybe still keep this around, as async_scheduling is not enabled by default.

Jialin (approved):

vllm/v1/worker/gpu_model_runner.py
njhill: Actually I removed this because it's a bit misleading, in non-async case a CPU sync happens a bit earlier when the sampled tokens are copied.

22quinn (approved):

Issue #24601, opened 2025-09-10
[Bug]: Launching multiple vLLM processes at the same time doesn't work well with vLLM's compile cache


zou3519:
### Your current environment

My makefile:
```
clean:
    rm -rf /tmp/torchinductor_rzou
    rm -rf ~/.cache/vllm/torch_compile_cache
    killall -9  "VLLM::EngineCore"

run:
    VLLM_ENABLE_V1_MULTIPROCESSING=0 CUDA_VISIBLE_DEVICES=2 vllm serve --gpu-memory-utilization 0.1 &
    VLLM_ENABLE_V1_MULTIPROCESSING=0 CUDA_VISIBLE_DEVICES=3 vllm serve --gpu-memory-utilization 0.1 &
    VLLM_ENABLE_V1_MULTIPROCESSING=0 CUDA_VISIBLE_DEVICES=4 vllm serve --gpu-memory-utilization 0.1 &
    VLLM_ENABLE_V1_MULTIPROCESSING=0 CUDA_VISIBLE_DEVICES=5 vllm serve --gpu-memory-utilization 0.1 &
    VLLM_ENABLE_V1_MULTIPROCESSING=0 CUDA_VISIBLE_DEVICES=6 vllm serve --gpu-memory-utilization 0.1 &
    VLLM_ENABLE_V1_MULTIPROCESSING=0 CUDA_VISIBLE_DEVICES=7 vllm serve --gpu-memory-utilization 0.1 &
```
`make run` 

### 🐛 Describe the bug

The problem is:
- we launch vllm in multiple processes
- each process tries to compile and save an artifact to disk in the same location
- each artifact is a directory
- the directories get clobbered

There's multiple ways I think we can fix the problem. I've tried some of these and haven't gotten them to work yet.
- standalone_compile offers a "binary" option instead of a "directory" option. writing to these binary files is atomic (we implemented it as a write+rename), reading from them is multiprocess-safe too (a process will not delete a file that's open until it is closed).
- if we go with the "directory" option, then we need some sort of read-write lock for CompiledArtifact.{save, load}. vLLM will always acquire said locks when it is performing save/load, and we can probably use something like a filelock for this. I've heard that filelocks are unreliable over nfs or other network stores, which does seem like how we are recommending vLLM users mount their ~/.cache/vllm directory, so I'm not sure how viable this is

### Before submitting a new issue...

[...]

PR #27616, opened 2025-10-28
[BUG] Make 'binary' default option for saving torch compile artifacts when using standalone_compile

hao-aaron:
## Purpose
Spawning multiple `vllm serve` processes on the same machine causes race conditions with the torch compile cache. This PR implements the 'binary' solution discussed in https://github.com/vllm-project/vllm/issues/24601

## Test Plan
my makefile:
```
clean:
    rm -rf /tmp/torchinductor_ray
    rm -rf ~/.cache/vllm/torch_compile_cache
    killall -9  "VLLM::EngineCore"

run:
    VLLM_ENABLE_V1_MULTIPROCESSING=0 CUDA_VISIBLE_DEVICES=0 vllm serve --gpu-memory-utilization 0.1 &
    VLLM_ENABLE_V1_MULTIPROCESSING=0 CUDA_VISIBLE_DEVICES=1 vllm serve --gpu-memory-utilization 0.1 &
    VLLM_ENABLE_V1_MULTIPROCESSING=0 CUDA_VISIBLE_DEVICES=2 vllm serve --gpu-memory-utilization 0.1 &
    VLLM_ENABLE_V1_MULTIPROCESSING=0 CUDA_VISIBLE_DEVICES=3 vllm serve --gpu-memory-utilization 0.1 &
```
run and observe that by default, binary is used:
- files inside torch compile cache are binary
- no errors
Set `VLLM_COMPILE_CACHE_SAVE_FORMAT=unpacked`
- make run causes errors during torch compile
- running single process with one `vllm serve` still works and generates directories in the place of the binary files in torch compile cache

---
<details>

</details>


kouroshHakha (approved):

ProExpertProg (commented):
cc @zou3519 

Questions (can be done in follow-up as well):
1. Can we remove the DP rank from the cache prefix with this?
1. Could we also add a way to only use unpacked on dp rank 0 (such that the cache can be debugged in a DP case without races)?

ProExpertProg (approved):

ProExpertProg:
Will wait for @zou3519's approval before merging

docs/design/torch_compile.md
```diff
@@ -27,6 +27,8 @@ With all these factors taken into consideration, usually we can guarantee that t
 
 A unique aspect of vLLM's `torch.compile` integration, is that we guarantee all the compilation finishes before we serve any requests. No requests will trigger new compilations. Otherwise, the engine would be blocked on that request, and the response time will have unexpected spikes.
 
+By default, the cache saves compiled artifacts as binary files. If you would like to interact with the generated code for debugging purposes, set `VLLM_COMPILE_CACHE_SAVE_FORMAT=unpacked`.
+
```
zou3519: I would also like a compile config version of this (@ProExpertProg does that sound reasonable?) The main reason is that sometimes it is easier to manipulate the config instead of an environment variable.

vllm/envs.py
```diff
@@ -218,6 +218,7 @@
     VLLM_USE_FBGEMM: bool = False
     VLLM_GC_DEBUG: str = ""
     VLLM_DISABLE_SHARED_EXPERTS_STREAM: bool = False
+    VLLM_COMPILE_CACHE_SAVE_FORMAT: Literal["binary", "unpacked"] = "binary"
```
zou3519, line 221: For better UX, the cache save format should go into the "factors" for caching. (we should put VLLM_COMPILE_CACHE_SAVE_FORMAT into https://github.com/vllm-project/vllm/blob/main/vllm/envs.py#L1502).

Otherwise someone needs to coordinate if they're saving an artifact that is binary or unpacked and then loading an artifact that is binary or unpacked.

docs/design/torch_compile.md
```diff
@@ -27,6 +27,8 @@ With all these factors taken into consideration, usually we can guarantee that t
 
 A unique aspect of vLLM's `torch.compile` integration, is that we guarantee all the compilation finishes before we serve any requests. No requests will trigger new compilations. Otherwise, the engine would be blocked on that request, and the response time will have unexpected spikes.
 
+By default, the cache saves compiled artifacts as binary files. If you would like to interact with the generated code for debugging purposes, set `VLLM_COMPILE_CACHE_SAVE_FORMAT=unpacked`.
```
zou3519: Why do we want the binary option to be the default over the unpacked option? 

@youkaichao I think you had strong opinions over this before, so would like to get your opinion. 
We're deciding between having vLLM save compiled artifacts in "unpacked" mode or "binary" mode by default.

In "unpacked mode", you are able to put breakpoints into the inductor output code and look at it.
In "binary mode", you're not able to put breakpoints into the inductor output code or look at it. However, you gain the ability to launch vLLM processes at the same time to compile the same exact model without clobbering each other.

So I think the question is how much we value the ability to launch multiple vLLM processes performing compilation at the same time.

zou3519 (changes requested):
- I think VLLM_COMPILE_CACHE_SAVE_FORMAT should go into vLLM's factor for the cache key computation
- I'm not completely convinced this should be the default

vllm/envs.py
hao-aaron: I added a field to compile config and it is part of the hash there. The current env variable will act as a fallback/default in the case that the user does not specify a field

docs/design/torch_compile.md
hao-aaron: i think by default its a bit less confusing to have binary option as default, since I assume most users will not be putting breakpoints in inductor. On the flip side, i think it is reasonable that people may be spawning multiple model replicas simultaneously, and will be very confused to see strange race condition errors with no warnings

zou3519 (approved):
I think the default flip is reasonable

vllm/config/compilation.py
```diff
@@ -208,6 +209,15 @@ class CompilationConfig:
     """The directory to store the compiled graph, to accelerate Inductor
     compilation. By default, it will use model-related information to generate
     a cache directory."""
+    compile_cache_save_format: Literal["binary", "unpacked"] = field(
+        default_factory=lambda: envs.VLLM_COMPILE_CACHE_SAVE_FORMAT
+    )
```
ProExpertProg, line 214: I like this as a config but I think generally env variables override CLI options (so that users can modify from the outside for debugging), see debug_dump_path - could you change this in a follow-up PR?

PR #27617, opened 2025-10-28
[Model] make the `inv_freq` in `Qwen2_5_VisionRotaryEmbedding` device-agnostic

yangulei:
## Purpose
Currently the `inv_freq` in `Qwen2_5_VisionRotaryEmbedding` is set to 'cpu' explicitly.
https://github.com/vllm-project/vllm/blob/a8d2e326ecb70876aa73dce70dbe2434c64b710a/vllm/model_executor/models/qwen2_5_vl.py#L610-L612
This PR makes it device-agnostic.

### DIFF
--- a/vllm/model_executor/models/qwen2_5_vl.py
+++ b/vllm/model_executor/models/qwen2_5_vl.py
@@ -607,9 +607,7 @@ class Qwen2_5_VisionRotaryEmbedding(nn.Module):
         super().__init__()
         self.dim = dim
         self.theta = theta
-        inv_freq = 1.0 / (
-            theta ** (torch.arange(0, dim, 2, dtype=torch.float, device="cpu") / dim)
-        )
+        inv_freq = 1.0 / (theta ** (torch.arange(0, dim, 2, dtype=torch.float) / dim))
         self.register_buffer("inv_freq", inv_freq, persistent=False)
         self._seq_len_cached = 0
         self._freqs_cached = None


vllm/model_executor/models/qwen2_5_vl.py
```diff
@@ -607,9 +607,7 @@ def __init__(self, dim: int, theta: float = 10000.0) -> None:
         super().__init__()
         self.dim = dim
         self.theta = theta
-        inv_freq = 1.0 / (
-            theta ** (torch.arange(0, dim, 2, dtype=torch.float, device="cpu") / dim)
-        )
+        inv_freq = 1.0 / (theta ** (torch.arange(0, dim, 2, dtype=torch.float) / dim))
```
DarkLight1337, line 610: I think this is to avoid CPU-GPU sync. Correct me if I'm wrong @ywang96 @Isotr0py

Isotr0py: Directly computing `inv_freq` on GPU can cause a slight numerical difference. We need to check if this change will significantly affect result's correctness. 🤔 
https://github.com/vllm-project/vllm/blob/5b3c35a68e049a93bd312c0e2012514b51875ed8/vllm/model_executor/layers/rotary_embedding/base.py#L57-L60

lgeiger:
Do you have any benchmarks for this change?

vllm/model_executor/models/qwen2_5_vl.py
yangulei: Current implementation in `vllm/model_executor/layers/rotary_embedding/base.py` and `vllm/model_executor/models/*` doesn't set to 'cpu' explicitly for `inv_freq` except for the `Qwen2_5_VisionRotaryEmbedding`. Any specific reason for that `Qwen2_5_VisionRotaryEmbedding` must initialize `inv_freq` on CPU?

PR #27621, opened 2025-10-28
[Add] cmdline argument parsing for KV cache offloading modules

### LINKED ISSUE #26858: [RFC]: Top-level CLI interface for KV cache offloading
### Motivation.

CPU (and tier-2 storage) offloading is an important feature in many cases (multi-round QA, document analysis, agent workflow, and reinforcement learning). With the recent advancement in the offloading connector, we already have the vLLM native CPU offloading implemented via the connector API. Also, there are multiple community efforts to provide other offloading implementations (e.g., LMCache, Nixl storage, mooncake) via the same set of APIs.

However, there is no clear documentation about how to configure the CPU offloading from the user's perspective. Right now, in order to enable CPU offloading, the user needs to pass a JSON string to `--kv-transfer-config`, which may create a huge mental barrier for new users. Therefore, it would be better to have a simple & clear user interface for users to enable CPU offloading. 

### Proposed Change.

This proposal contains two new command-line arguments:
- `--kv-offloading-size`: a numeric value to control a global offloading buffer size (in GB). When TP > 1, this number should be the total size summed across all the TP ranks. (An alternative is the buffer size for each TP rank.)
- `--kv-offloading-backend`: a string that specifies which offloading backend to use, such as "native", "lmcache", "mooncake", "3fs", or "nixl".

This will give enough clarity to most of the users who want to use the offloading feature, and should be extensible enough to new offloading backends and tier-2 storage.

## Required changes

To implement this proposal, the following things are needed:
- Add logic to parse the new CLI argument and store it into vllm config.
- Add a new module to translate the `--kv-offloading-size` and `--kv-offloading-backend` to the corresponding KV connector config.
- Add the documentation to the vLLM user guide.

### Feedback Period.

1~2 weeks

### CC List.

@simon-mo @orozery @njhill 

### Any Other Things.

_No response_

ApostaC:
## Purpose
This PR partly fixes #26858 .

**Major changes:**
- adds two command-line arguments: `--kv-offloading-size` and `--kv-offloading-backend`.
- adds a new class `KVOffloadingConfigParser` which produces the connector config based on the above command line args
- partly adds the support for the native offloading backend and lmcache offloading backend

**Follow-up PRs**
- [ ] Add the logic to correctly initialize the configuration for `OffloadingConnector` and `LMCacheConnector`

## Test Plan
N/A

## Test Result
N/A

---
<details>

</details>


vllm/distributed/kv_transfer/kv_connector/config_parser.py
hmellor, line 1: The files in `vllm/config` should really contain config classes, but this file contains post init code for `KVTransferConfig`

hmellor (dismissed):
Instead of adding a new file and a registry of parser classes, could this be done in the `VllmConfig.__post_init__`?

vllm/distributed/kv_transfer/kv_connector/config_parser.py
ApostaC: I see. I will make the changes!

vllm/config/cache.py
```diff
@@ -128,6 +128,17 @@ class CacheConfig:
     gpu_memory_utilization. Note that kv_cache_memory_bytes
     (when not-None) ignores gpu_memory_utilization"""
 
+    kv_offloading_size: float | None = None
+    """Size of the KV cache offloading buffer in GiB. When TP > 1, this is
+    the total buffer size summed across all TP ranks. By default, this is set
+    to None, which means no KV offloading is enabled. When set with
+    kv_offloading_backend, vLLM will enable KV cache offloading to CPU"""
+
+    kv_offloading_backend: str | None = None
```
hmellor: This way pydantic will validate the inputs

```suggestion
    kv_offloading_backend: Literal["native", "lmcache"] | None = None
```

vllm/config/vllm.py
```diff
@@ -667,6 +667,15 @@ def has_blocked_weights():
             if "-quant_fp8" not in custom_ops:
                 custom_ops.append("+quant_fp8")
 
+        # Handle the KV connector configs
+        from vllm.distributed.kv_transfer.kv_connector.config_parser import (
+            apply_extra_kv_connector_config,
+        )
+
+        self.kv_transfer_config = apply_extra_kv_connector_config(
+            self, self.kv_transfer_config
```
hmellor: Explicitly passing self seems strange, can we just make this a method of `VllmConfig`?

Passing `self.kv_transfer_config` separately is also strange because it's in `self`

ApostaC: Makes sense. I moved `apply_extra_kv_connector_config` to `VllmConfig._ apply_extra_kv_connector_config`

ApostaC:
@hmellor Addressed your comments. PTAL, thanks!

hmellor:
I'm glad you approve. Now that I've also contributed to this PR in a non-trivial way, we should get another reviewer to make sure that I haven't made any errors either.

@DarkLight1337 / @Isotr0py could you take a look?

DarkLight1337 (approved):

PR #25794, opened 2025-09-27
Add filtering for chat template kwargs

russellb:
- https://github.com/vllm-project/vllm/security/advisories/GHSA-6fvq-23cw-5628
- https://github.com/vllm-project/vllm/security/advisories/GHSA-4hw3-g54f-xfrj

PR #27622, opened 2025-10-28
fix: allow HuggingFace standard chat template params via **kwargs

wangln19:
## Purpose

Fix compatibility issue with tokenizers that use `**kwargs` to receive standard chat template parameters.

**Problem**: 
- Some tokenizer implementations (e.g., Kimi K2) don't explicitly declare standard HuggingFace parameters like `add_generation_prompt`, `tools`, etc. in their `apply_chat_template` method signature
- Instead, they receive these parameters via `**kwargs`
- The current parameter filtering logic in `resolve_chat_template_kwargs` uses `allow_var_kwargs=False`, which rejects these parameters
- This causes tool calling and other features to fail silently (e.g., Kimi K2 always returns `finish_reason: stop` instead of `finish_reason: tool_calls`)

**Root Cause**:
The security fix in PR #25794 prevents passing parameters not explicitly declared in the function signature to avoid injection attacks. While this is correct for unknown parameters, it inadvertently blocks legitimate HuggingFace standard parameters when tokenizers use `**kwargs`.

**Solution**:
Dynamically extract the standard parameter list from `PreTrainedTokenizer.apply_chat_template` base class signature and whitelist these parameters even when the tokenizer implementation uses `**kwargs` to receive them.

**Benefits**:
- ✅ Fixes compatibility with Kimi K2 and similar tokenizers
- ✅ Maintains security - only official HuggingFace parameters are allowed
- ✅ Zero maintenance - automatically stays in sync with transformers library updates
- ✅ No manual whitelist to maintain

## Test Plan

1. **Unit test** for parameter filtering logic:
```bash
pytest tests/entrypoints/openai/test_chat_template.py -v
```

2. **Integration test** with Kimi K2 model:
```bash
# Start vLLM server with Kimi K2
vllm serve Kimi/kimi-k2 --tool-parser kimi_k2

# Test tool calling with add_generation_prompt
curl http://localhost:8000/v1/chat/completions \
  -d '{
    "model": "Kimi/kimi-k2",
    "messages": [{"role": "user", "content": "What is the weather in Beijing?"}],
    "tools": [{"type": "function", "function": {"name": "get_weather", "parameters": {...}}}],
    "add_generation_prompt": true
  }'
```

3. **Verify other models** still work correctly:
```bash
# Test with standard tokenizers (e.g., Llama, Qwen)
pytest tests/tool_use/ -k "not kimi" -v
```

## Test Result

**Before the fix**:
- Kimi K2 tool calls: `finish_reason: "stop"` (wrong - model generates text instead of tool call)
- Parameter `add_generation_prompt` was silently dropped
- Logs show filtered parameters: `{'tools': [...]}` (missing `add_generation_prompt`)

**After the fix**:
- Kimi K2 tool calls: `finish_reason: "tool_calls"` ✅
- Parameter `add_generation_prompt: true` correctly passed to tokenizer
- Logs show all parameters: `{'tools': [...], 'add_generation_prompt': True}`
- Standard tokenizers (Llama, Qwen, etc.) continue to work as before ✅

**Security verification**:
- Unknown parameters still rejected: ✅
  ```python
  # Request with evil_param
  {"evil_param": "malicious"} → Filtered out, not passed to tokenizer
  ```
- Only HuggingFace official parameters allowed: ✅
  ```python
  _get_hf_base_chat_template_params() → {'conversation', 'add_generation_prompt', 'tools', ...}
  ```

## Code Changes

**Modified file**: `vllm/entrypoints/chat_utils.py`

1. Added `_get_hf_base_chat_template_params()` function to dynamically extract standard parameters from HuggingFace base class
2. Updated `resolve_chat_template_kwargs()` to include `hf_base_params` in the accept list
3. Moved `import inspect` to module level for clarity

**Lines changed**: ~15 lines added

---

### Checklist

- [x] The purpose of the PR - Fix tokenizer compatibility issue with **kwargs parameters
- [x] The test plan - Provided test commands for Kimi K2 and other models
- [x] The test results - Before/after comparison showing the fix works
- [ ] Documentation update - Not applicable (internal change, no user-facing API changes)
- [ ] Release notes update - Will update if maintainers request

### DIFF
--- a/vllm/entrypoints/chat_utils.py
+++ b/vllm/entrypoints/chat_utils.py
@@ -2,6 +2,7 @@
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 
 import asyncio
+import inspect
 import json
 from abc import ABC, abstractmethod
 from collections import Counter, defaultdict, deque
@@ -1515,6 +1516,24 @@ def _resolve_chat_template_kwargs(
 _cached_resolve_chat_template_kwargs = lru_cache(_resolve_chat_template_kwargs)
 
 
+@lru_cache
+def _get_hf_base_chat_template_params() -> frozenset[str]:
+    # Get standard parameters from HuggingFace's base tokenizer class.
+    # This dynamically extracts parameters from PreTrainedTokenizer's
+    # apply_chat_template method, ensuring compatibility with tokenizers
+    # that use **kwargs to receive standard parameters.
+
+    # Read signature from HF's base class - the single source of truth
+    base_sig = inspect.signature(PreTrainedTokenizer.apply_chat_template)
+    # Exclude VAR_KEYWORD (**kwargs) and VAR_POSITIONAL (*args) placeholders
+    return frozenset(
+        p.name
+        for p in base_sig.parameters.values()
+        if p.kind
+        not in (inspect.Parameter.VAR_KEYWORD, inspect.Parameter.VAR_POSITIONAL)
+    )
+
+
 def resolve_chat_template_kwargs(
     tokenizer: PreTrainedTokenizer | PreTrainedTokenizerFast,
     chat_template: str,
@@ -1538,7 +1557,11 @@ def resolve_chat_template_kwargs(
         if supports_kw(tokenizer.apply_chat_template, k, allow_var_kwargs=False)
     }
     template_vars = _cached_resolve_chat_template_kwargs(chat_template)
-    accept_vars = (fn_kw | template_vars) - unexpected_vars
+
+    # Allow standard HF parameters even if tokenizer uses **kwargs to receive them
+    hf_base_params = _get_hf_base_chat_template_params() & chat_template_kwargs.keys()
+
+    accept_vars = (fn_kw | template_vars | hf_base_params) - unexpected_vars
     return {k: v for k, v in chat_template_kwargs.items() if k in accept_vars}
 
 
--- a/vllm/entrypoints/chat_utils.py
+++ b/vllm/entrypoints/chat_utils.py
@@ -1559,7 +1559,7 @@ def resolve_chat_template_kwargs(
     template_vars = _cached_resolve_chat_template_kwargs(chat_template)
 
     # Allow standard HF parameters even if tokenizer uses **kwargs to receive them
-    hf_base_params = _get_hf_base_chat_template_params() & chat_template_kwargs.keys()
+    hf_base_params = _get_hf_base_chat_template_params()
 
     accept_vars = (fn_kw | template_vars | hf_base_params) - unexpected_vars
     return {k: v for k, v in chat_template_kwargs.items() if k in accept_vars}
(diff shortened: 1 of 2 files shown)


vllm/entrypoints/chat_utils.py
```diff
@@ -1538,7 +1557,11 @@ def resolve_chat_template_kwargs(
         if supports_kw(tokenizer.apply_chat_template, k, allow_var_kwargs=False)
     }
     template_vars = _cached_resolve_chat_template_kwargs(chat_template)
-    accept_vars = (fn_kw | template_vars) - unexpected_vars
+
+    # Allow standard HF parameters even if tokenizer uses **kwargs to receive them
+    hf_base_params = _get_hf_base_chat_template_params() & chat_template_kwargs.keys()
```
DarkLight1337:
```suggestion
    hf_base_params = _get_hf_base_chat_template_params()
```

The intersection is not necessary since it's applied in L1565 already

DarkLight1337 (approved):
Thanks, LGTM.

cc @Isotr0py

vllm/entrypoints/chat_utils.py
```diff
@@ -1538,7 +1557,11 @@ def resolve_chat_template_kwargs(
         if supports_kw(tokenizer.apply_chat_template, k, allow_var_kwargs=False)
     }
     template_vars = _cached_resolve_chat_template_kwargs(chat_template)
-    accept_vars = (fn_kw | template_vars) - unexpected_vars
+
+    # Allow standard HF parameters even if tokenizer uses **kwargs to receive them
+    hf_base_params = _get_hf_base_chat_template_params()
+
+    accept_vars = (fn_kw | template_vars | hf_base_params) - unexpected_vars
```
Isotr0py, line 1564: Can you also update the test in `tests/entrypoints/test_chat_utils.py`?
https://github.com/vllm-project/vllm/blob/5b3c35a68e049a93bd312c0e2012514b51875ed8/tests/entrypoints/test_chat_utils.py#L1783-L1806

Isotr0py (approved):
Thanks for fixing!

vllm/entrypoints/chat_utils.py
wangln19: Updated the test as suggested. 

I considered adding Kimi K2 to the test model registry, but decided against it for the following reasons:

1. **Infrastructure overhead**: Adding a new model to `tests/models/registry.py` requires extensive configuration (tokenizer path, trust_remote_code settings, HF overrides, etc.), which would be significant effort for testing a single parameter filtering behavior.

2. **Generic fix**: This change is not Kimi K2-specific. It enables support for *any* tokenizer that uses `**kwargs` to receive HuggingFace standard parameters. The mock tokenizer approach tests the core logic more directly.

3. **Sufficient coverage**: The updated test now includes:
   - Existing integration tests (Qwen2-VL, Qwen3) verify backward compatibility
   - New mock tokenizer test validates the **kwargs scenario (like Kimi K2)
   - Manual integration testing with actual Kimi K2 model confirms end-to-end functionality (as documented in PR description)

The mock approach isolates the parameter filtering logic we're actually fixing, without coupling the test suite to a specific model that may have availability/licensing constraints.

PR #26060, opened 2025-10-01
[V1][spec decode] return logprobs for spec decoding

TheEpicDolphin:
# Purpose
Add support for returning logprobs for v1 spec decoding.

# Test Plan

## Automated Tests
Added automated testing for logprobs which compares spec decode LLM per-token output logprobs with those of a reference model. The comparison is done for: `raw_logits`, `raw_logprobs`, `processed_logits`, and `processed_logprobs`.
```
(py312conda) bash-5.1$ pytest tests/v1/sample/test_logprobs.py -k test_spec_decode_logprobs
============================================================================================================================================= test session starts =============================================================================================================================================
platform linux -- Python 3.12.9, pytest-8.4.1, pluggy-1.6.0
rootdir: /data/users/gdelfin/gitrepos/vllm
configfile: pyproject.toml
plugins: anyio-4.9.0, asyncio-1.1.0
asyncio: mode=Mode.STRICT, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 30 items / 26 deselected / 4 selected                                                                                                                                                                                                                                                               

tests/v1/sample/test_logprobs.py ....                                                                                                                                                                                                                                                                   [100%]

============================================================================================================================================== warnings summary ===============================================================================================================================================
tests/v1/sample/test_logprobs.py::test_spec_decode_logprobs[model_setup0-raw_logits]
tests/v1/sample/test_logprobs.py::test_spec_decode_logprobs[model_setup0-raw_logits]
tests/v1/sample/test_logprobs.py::test_spec_decode_logprobs[model_setup0-raw_logprobs]
tests/v1/sample/test_logprobs.py::test_spec_decode_logprobs[model_setup0-raw_logprobs]
tests/v1/sample/test_logprobs.py::test_spec_decode_logprobs[model_setup0-processed_logits]
tests/v1/sample/test_logprobs.py::test_spec_decode_logprobs[model_setup0-processed_logits]
tests/v1/sample/test_logprobs.py::test_spec_decode_logprobs[model_setup0-processed_logprobs]
tests/v1/sample/test_logprobs.py::test_spec_decode_logprobs[model_setup0-processed_logprobs]
  /home/gdelfin/.conda/envs/py312conda/lib/python3.12/multiprocessing/popen_fork.py:66: DeprecationWarning: This process (pid=2004414) is multi-threaded, use of fork() may lead to deadlocks in the child.
    self.pid = os.fork()

[...]

PR #27629, opened 2025-10-28
[Core][Bookkeeping] Update cu_num_accepted_tokens for all req_index

Jialin:
## Purpose
In this PR, we ensure cu_num_accepted_tokens is updated for all request_index. To avoid position skipped / shifted due to empty or None sampled_ids.

## Test Plan & Test Result
```
> pytest tests/v1/sample/test_logprobs.py -k test_spec_decode_logprobs
> pytest tests/v1/sample/test_rejection_sampler.py
```

---
<details>

</details>

### DIFF
--- a/vllm/v1/worker/gpu_model_runner.py
+++ b/vllm/v1/worker/gpu_model_runner.py
@@ -2329,7 +2329,7 @@ class GPUModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
                     cu_num_accepted_tokens[-1] + sampled_id_cnt
                 )
 
-            if sampled_id_cnt == 0:
+            if not sampled_ids:
                 continue
 
             start_idx = self.input_batch.num_tokens_no_spec[req_idx]
--- a/vllm/v1/worker/gpu_model_runner.py
+++ b/vllm/v1/worker/gpu_model_runner.py
@@ -2322,18 +2322,18 @@ class GPUModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
             else:
                 sampled_ids = valid_sampled_token_ids[req_idx]
 
-            sampled_id_cnt: int = len(sampled_ids) if sampled_ids else 0
+            num_sampled_ids: int = len(sampled_ids) if sampled_ids else 0
 
             if cu_num_accepted_tokens is not None:
                 cu_num_accepted_tokens.append(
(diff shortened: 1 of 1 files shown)


TheEpicDolphin:
Thanks for catching this @Jialin, looks good to me!

22quinn (approved):
👍

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2309,11 +2309,19 @@ def _bookkeeping_sync(
                 sampled_ids = [-1] if req_idx not in invalid_req_indices_set else None
             else:
                 sampled_ids = valid_sampled_token_ids[req_idx]
+
+            sampled_id_cnt: int = len(sampled_ids) if sampled_ids else 0
```
njhill: for consistency with naming convention used elsewhere
```suggestion
            num_sampled_ids: int = len(sampled_ids) if sampled_ids else 0
```

njhill (commented):
Thanks for catching @Jialin!

Any chance we could extend test(s) added in https://github.com/vllm-project/vllm/pull/26060 to catch this case?

vllm/v1/worker/gpu_model_runner.py
Jialin: Updated

Jialin:
> Thanks for catching @Jialin!
> 
> Any chance we could extend test(s) added in #26060 to catch this case?

@njhill Skimmed throughput the tests added in #26060, but seems most of them are e2e testing, so it might be a bit hard to add unit test to cover this.

However, after this change, as we're only updating numpy arrays in bookkeeping now. A potential followup step is to introduce a numba jit function to further speed up bookkeeping. (And I could definitely add unit tests to cover this case after with the more self-contained jit function. WDYT?

njhill:
> > Thanks for catching @Jialin!
> > Any chance we could extend test(s) added in #26060 to catch this case?
> 
> @njhill Skimmed throughput the tests added in #26060, but seems most of them are e2e testing, so it might be a bit hard to add unit test to cover this.
> 
> However, after this change, as we're only updating numpy arrays in bookkeeping now. A potential followup step is to introduce a numba jit function to further speed up bookkeeping. (And I could definitely add unit tests to cover this case after with the more self-contained jit function. WDYT?

@Jialin fair enough, I wasn't sure whether we could tweak the e2e test to trigger the bug. If that's nontrivial then sounds ok re leaving to future unit tests.

njhill (approved):

PR #27631, opened 2025-10-28
[NIXL][XPU] update name of nixl wheel

zhenwei-intel:
The recent update of NIXL's whl filename (`nixl_cu12-0.6.1-cp312-cp312-linux_x86_64.whl`) caused `find_nixl_wheel_in_cache` to fail.


jikunshang (approved):

zhenwei-intel:
> In the next NIXL release, the package will be split into `nixl` and `nixl-cu12` / `nixl-cu13`.
> 
> You are installing only `nixl-cu12` but missing `nixl`. That one can be found in the build tree in build/src/bindings/python/nixl-meta/nixl-*.whl
> 
> This is still in the works so it may make sense to wait a week or so for things to stabilize

Thanks, I'll keep tracking this.​

PR #27637, opened 2025-10-28
[Bugfix] vLLM should check Inductor config for compile cache enablement status

gmagogsfm:
Summary:
vLLM should not assume compile cache is enabled just because VLLM_DISABLE_COMPILE_CACHE=0. Users may use TORCHINDUCTOR_FORCE_DISABLE_CACHES=1 to disable compile cache at PyTorch level, effectively making it impossible for vLLM compile cache to function. 

This PR adds necessary checks to ensure vLLM considers Torch Inductor compile cache enablement status.

Signed off by: Yanan Cao <gmagogsfm@gmail.com>

### DIFF
--- a/vllm/compilation/backends.py
+++ b/vllm/compilation/backends.py
@@ -33,6 +33,7 @@ from .compiler_interface import (
     EagerAdaptor,
     InductorAdaptor,
     InductorStandaloneAdaptor,
+    is_compile_cache_enabled,
 )
 from .counter import compilation_counter
 from .inductor_pass import InductorPass
@@ -240,7 +241,7 @@ class CompilerManager:
         assert compiled_graph is not None, "Failed to compile the graph"
 
         # store the artifact in the cache
-        if not envs.VLLM_DISABLE_COMPILE_CACHE and handle is not None:
+        if is_compile_cache_enabled(additional_inductor_config) and handle is not None:
             self.cache[(runtime_shape, graph_index, self.compiler.name)] = handle
             compilation_counter.num_cache_entries_updated += 1
             self.is_cache_updated = True
@@ -612,7 +613,9 @@ class VllmBackend:
         os.makedirs(local_cache_dir, exist_ok=True)
         self.compilation_config.local_cache_dir = local_cache_dir
 
-        disable_cache = envs.VLLM_DISABLE_COMPILE_CACHE
+        disable_cache = not is_compile_cache_enabled(
+            self.compilation_config.inductor_compile_config
+        )
 
         if disable_cache:
             logger.info_once("vLLM's torch.compile cache is disabled.", scope="local")
--- a/vllm/compilation/compiler_interface.py
+++ b/vllm/compilation/compiler_interface.py
@@ -163,6 +163,23 @@ def get_inductor_factors() -> list[Any]:
     return factors
 
 
+def is_compile_cache_enabled(
+    vllm_additional_inductor_config: dict[str, Any],
+) -> bool:
+    vllm_inductor_config_disable_cache = vllm_additional_inductor_config.get(
+        "force_disable_caches", False
+    )
+
+    # TODO(gmagogsfm): Replace torch._inductor.config.force_disable_caches
+    # with torch.compiler.config.force_disable_caches when minimum PyTorch
+    # version reaches 2.10
+    return (
+        not envs.VLLM_DISABLE_COMPILE_CACHE
+        and not torch._inductor.config.force_disable_caches
+        and not vllm_inductor_config_disable_cache
+    )
+
+
 class InductorStandaloneAdaptor(CompilerInterface):
     """
     The adaptor for the Inductor compiler.
@@ -222,7 +239,8 @@ class InductorStandaloneAdaptor(CompilerInterface):
         # Save the compiled artifact to disk in the specified path
         assert key is not None
         path = os.path.join(self.cache_dir, key)
-        if not envs.VLLM_DISABLE_COMPILE_CACHE:
+
+        if is_compile_cache_enabled(compiler_config):
             compiled_graph.save(path=path, format=self.save_format)
             compilation_counter.num_compiled_artifacts_saved += 1
         return compiled_graph, (key, path)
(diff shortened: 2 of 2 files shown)


vllm/compilation/compiler_interface.py
```diff
@@ -219,7 +219,14 @@ def compile(
...
         path = os.path.join(self.cache_dir, key)
-        if not envs.VLLM_DISABLE_COMPILE_CACHE:
+
+        # TODO(gmagogsfm): Replace torch._inductor.config.force_disable_caches
+        # with torch.compiler.config.force_disable_caches when minimum PyTorch
+        # version reaches 2.9
+        if (
+            not envs.VLLM_DISABLE_COMPILE_CACHE
+            and not torch._inductor.config.force_disable_caches
+        ):
```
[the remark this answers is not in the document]
gmagogsfm: Deduped by adding a util in `compiler_interface.py`

vllm/compilation/backends.py
```diff
@@ -610,7 +610,13 @@ def __call__(
         os.makedirs(local_cache_dir, exist_ok=True)
         self.compilation_config.local_cache_dir = local_cache_dir
 
-        disable_cache = envs.VLLM_DISABLE_COMPILE_CACHE
+        # TODO(gmagogsfm): Replace torch._inductor.config.force_disable_caches
+        # with torch.compiler.config.force_disable_caches when minimum PyTorch
+        # version reaches 2.9
+        disable_cache = (
+            envs.VLLM_DISABLE_COMPILE_CACHE
```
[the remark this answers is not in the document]
gmagogsfm: Updated to also check inductor flag

vllm/compilation/compiler_interface.py
```diff
@@ -163,6 +163,16 @@ def get_inductor_factors() -> list[Any]:
     return factors
 
 
+def is_compile_cache_enabled() -> bool:
+    # TODO(gmagogsfm): Replace torch._inductor.config.force_disable_caches
+    # with torch.compiler.config.force_disable_caches when minimum PyTorch
+    # version reaches 2.9
```
zou3519: I don't see this API in 2.9

zou3519: If your proposal was to make this public -- The in vLLM code you have put this into is for InductorAdaptor, which we are trying to deprecate in favor of StandaloneInductorAdaptor, so I wouldn't worry about it too much. We should end up deleting this code anyways in 2.10+

zou3519 (commented):
code lgtm, I would just change the comment

vllm/compilation/compiler_interface.py
gmagogsfm: This `force_disable_caches` behavior actually affects more than just `InductorAdaptor`, it is also used in `InductorStandaloneAdapter`. In fact the new error I showed you earlier is from `InductorStandaloneAdaptor`.

I changed comment to mention 2.10. Also made a PR to expose this flag:
https://github.com/pytorch/pytorch/pull/166699
Please take a look.

zou3519 (approved):

vllm/compilation/backends.py
```diff
@@ -238,7 +239,7 @@ def compile(
         assert compiled_graph is not None, "Failed to compile the graph"
 
         # store the artifact in the cache
-        if not envs.VLLM_DISABLE_COMPILE_CACHE and handle is not None:
+        if is_compile_cache_enabled() and handle is not None:
```
ProExpertProg: The user might also pass `force_disable_caches` as part of `CompilationConfig.inductor_compile_config`, we should check that as well

ProExpertProg (changes requested):

vllm/compilation/backends.py
gmagogsfm: Good catch, added a check for compilation config in the `is_compile_cache_enabled`

vllm/compilation/compiler_interface.py
```diff
@@ -163,6 +163,19 @@ def get_inductor_factors() -> list[Any]:
...
 
+def is_compile_cache_enabled() -> bool:
+    # TODO(gmagogsfm): Replace torch._inductor.config.force_disable_caches
+    # with torch.compiler.config.force_disable_caches when minimum PyTorch
+    # version reaches 2.10
+    return (
+        not envs.VLLM_DISABLE_COMPILE_CACHE
+        and not torch._inductor.config.force_disable_caches
+        and not (
+            get_current_vllm_config().compilation_config.inductor_compile_config.force_disable_caches
```
ProExpertProg: I don't think that vllm config is necessarily available at this time, why not just pass compilation config as a parameter?

vllm/compilation/decorators.py
```diff
@@ -447,6 +447,7 @@ def patched_inline_call(self_):
                     InliningInstructionTranslator, "inline_call_", patched_inline_call
                 ),
                 torch._dynamo.config.patch(**dynamo_config_patches),
+                set_current_vllm_config(self.vllm_config),
```
ProExpertProg: This is a huge change, we shouldn't be settling the vllm config globally here

vllm/compilation/compiler_interface.py
ProExpertProg: Also, inductor compile config is a dictionary, and the key might not be present:
```suggestion
            compilation_config.inductor_compile_config.get("force_disable_caches", False)
```

gmagogsfm: Fixed. Now `is_compile_cache_enabled` checks inductor configs explicitly passed in.

I noticed that vllm config is not present at this time during later testing. In general, when should I expect global vllm config to be available and when not? Is it mostly expected to be available during model init?

vllm/compilation/decorators.py
gmagogsfm: Removed now that the inductor config is passed in explicitly.

gmagogsfm:
@ProExpertProg When you get a chance, could you take a look at the new revision? your suggested changes are addressed. Thanks

ProExpertProg (approved):

PR #27643, opened 2025-10-28
[MTP] Refactor mtp predictor to avoid d2h operation

MengqingCao:
## Purpose
Refactor mtp predictor to avoid d2h operation.
The original logic brings device-to-host operation as `position` is a device tensor and the indexing operation `inputs_embeds[positions == 0]` will trigger a d2h synchornization. This breaks the static graph mode in vllm-ascend.
This pr refactors it to all the operations on torch, thus fixing this issue.

## Test Plan
Test locally on both H20 and vllm-ascend with deepseek-v3 + mtp + piecewise cudagraph

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/models/deepseek_mtp.py
+++ b/vllm/model_executor/models/deepseek_mtp.py
@@ -97,7 +97,7 @@ class DeepSeekMultiTokenPredictorLayer(nn.Module):
     ) -> torch.Tensor:
         assert inputs_embeds is not None
         # masking inputs at position 0, as not needed by MTP
-        inputs_embeds[positions == 0] = 0
+        inputs_embeds = torch.where(positions.unsqueeze(-1) == 0, 0, inputs_embeds)
         inputs_embeds = self.enorm(inputs_embeds)
         previous_hidden_states = self.hnorm(previous_hidden_states)
 


MengqingCao:
Good day! @luccafong Could you take a look at this pr? thanks!

luccafong (approved):

vllm/model_executor/models/deepseek_mtp.py
```diff
@@ -97,7 +97,7 @@ def forward(
     ) -> torch.Tensor:
         assert inputs_embeds is not None
         # masking inputs at position 0, as not needed by MTP
-        inputs_embeds[positions == 0] = 0
+        inputs_embeds = torch.where(positions.unsqueeze(-1) == 0, 0, inputs_embeds)
```
benchislett, line 100: does this issue the same number of kernels as `inputs_embeds.masked_fill_(positions == 0, 0)`? I would prefer to use `masked_fill_` if there is no performance loss, and I'm not sure if they are equivalent to torch

MengqingCao: @benchislett I got your point of using an in-place operation to avoid introduing memory allocating operation. And I checked the source code of `masked_fill_`, it indicates that `masked_fill` actually is implemented by `where` operation inside, and adding some checks before that. Thus I think it's fine to keep the `torch.where` here? WDYT?
https://github.com/pytorch/pytorch/blob/main/torch/_refs/__init__.py#L5892

benchislett: In that case it's fine either way.

benchislett (approved):
LGTM

MengqingCao:
> LGTM

Thanks! could you help merge this pr? I also hope it will be included in releases/v0.11.1, should I make a backport pr then?

whx-sjtu (approved):
Nice fix!

PR #19970, opened 2025-06-23
Implement Async Scheduling

WoosukKwon:
This PR implements an initial version of the asynchronous scheduler, similar to the approach described in the [NanoFlow](https://arxiv.org/abs/2408.12757) paper. The primary goal is to minimize scheduler overhead by overlapping scheduling with model execution. This is achieved by making the scheduler operate one step ahead of the current execution.

## Feature Compatibility

Async scheduling is introduced as an **experimental** feature and can be enabled using the `--async-scheduling` flag. It has currently only been validated with basic setups and has not yet been tested in the following scenarios:

* KV connector
* Data parallel

At this stage, async scheduling does not support:

* Speculative decoding
* Structured outputs
* Pipeline parallelism

Most of these limitations are potentially addressable in future iterations. For example, I believe we can make most speculative decoding methods (such as EAGLE) compatible with async scheduling.

## Implementation

To reuse code and minimize disruption, the async scheduler is implemented as a subclass of the existing scheduler. The key addition is an "output placeholder," representing tokens scheduled but not yet generated.

One limitation of the current design is that the scheduler and GPU worker must run in separate processes for parallel execution. This setup causes extra serialization overhead for input/output data (which is unnecessary in single-gpu cases). This may impact performance, especially for multimodal models with large inputs.

## Performance

Async scheduling increases throughput by 3-15%. The speedup is more pronounced with smaller models and larger batch sizes. No significant latency improvement is observed for batch size 1 scenarios, as scheduler overhead in those cases is already minimal.

In terms of serving latency, async scheduling generally **reduces TPOT** but slightly **increases TTFT**. The TTFT increases because incoming requests must wait for an additional scheduling step before it is actually processed on GPUs. Despite this, the overall end-to-end latency typically gets faster.

image

PR #27647, opened 2025-10-28
[Core] Support async-scheduling with kvconnector

KevinCheung2259:
## Purpose
As demonstrated in our tests and discussed in issue #19970 , async scheduling is currently not supported with the KV connector. Both of these features are critical for LLM inference.

This PR support async scheduling mode when a KVConnector is configured. This unlocks better concurrency, reduces tail latency, and improves throughput for prefix-reuse and offloading scenarios.

## Core Changes

1. **Extended `SchedulerOutput` dataclass** (`vllm/v1/core/sched/output.py`)
   - Added `num_output_placeholders: dict[str, int] | None` field to track tokens scheduled but not yet generated
   - Enables KV connectors to distinguish between scheduled placeholders and actually computed tokens
   - Backward compatible: defaults to `None`, populated only in async scheduling mode

2. **Added `_get_num_output_placeholders()` method** (`vllm/v1/core/sched/scheduler.py`)
   - Base implementation in `Scheduler` returns empty dict (sync mode: no placeholders)
   - Called before constructing `SchedulerOutput` to collect placeholder information
   - Designed for extensibility: subclasses override to provide actual counts

3. **`AsyncScheduler` override** (`vllm/v1/core/sched/async_scheduler.py`)
   - Overrides `_get_num_output_placeholders()` to collect actual placeholder counts from each request
   - Provides KV connectors with the formula: `real_computed_tokens = num_computed_tokens - num_output_placeholders`
   - Maintains existing async scheduling logic while exposing placeholder state

Changes are backward compatible with sync scheduling.

---
<details>

</details>


ApostaC (approved):
The implementation looks pretty clear to me! Thanks!

njhill (changes requested):
Thanks @KevinCheung2259.

> As demonstrated in our tests

Do you think you could explain in more detail exactly what the problem is and share your tests that demonstrate it or an equivalent reproducer?

I am probably missing something but I don't see why placeholder token count is an issue here. The actual saving is done later by the worker side of the connector which at that time will have all of the generated kv cache for the "computed" tokens, including those which were in the original placeholder count.

njhill:
Thanks @KevinCheung2259 I am probably just being slow but I still don't see why the "problem here" is a problem.

The subsequently called `_get_block_hashes` will return only valid block hashes, i.e. those corresponding to cached real tokens.

> Would a failing test case with detailed logging help demonstrate this? I can provide one if useful.

That would be great thanks! Something end-to-end which shows incorrect generated output due to this problem.

KevinCheung2259:
> That would be great thanks! Something end-to-end which shows errors or incorrect generated output due to this problem. No need to add logging.那就太好了，谢谢！需要一个端到端的示例，展示由于这个问题导致的错误或生成的错误输出。不需要添加日志。

```
(EngineCore_DP0 pid=257)   File "/vllm-workspace/vllm/vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py", line 344, in build_connector_meta
(EngineCore_DP0 pid=257)     reqs_to_store=self._get_reqs_to_store(scheduler_output),
(EngineCore_DP0 pid=257)                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=257)   File "/vllm-workspace/vllm/vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py", line 292, in _get_reqs_to_store
(EngineCore_DP0 pid=257)     assert len(req.block_hashes) >= num_gpu_blocks
(EngineCore_DP0 pid=257)            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=257) AssertionError
```
I fully agree with your point @njhill ! Once I relaxed that assertion with asynchronous scheduling enabled, the problem was resolved.

KevinCheung2259:
Thanks @njhill , if we only fix in connector scheduler side, I have another two questions:
1. What risks are there in relaxing this assertion under asynchronous scheduling?
2. How can the connector know that asynchronous scheduling is enabled, should it just infer this from observed behavior?

KevinCheung2259:
```
            # In async scheduling, num_computed_tokens may include placeholders
            # (scheduled but not-yet-generated tokens). When that happens, the
            # number of available block hashes can be smaller than what the
            # optimistic boundary suggests. If we detect this mismatch, relax
            # the check by clamping the target num_blocks to what is actually
            # available according to block hashes.
            available_gpu_hashes = len(req.block_hashes)
            if available_gpu_hashes < num_gpu_blocks:
                max_blocks_by_hashes = available_gpu_hashes // self.block_size_factor
                if max_blocks_by_hashes <= start_block_idx:
                    # Nothing new can be stored this step.
                    continue
                # Clamp to the maximum storable blocks backed by hashes.
                num_blocks = max_blocks_by_hashes
                num_new_blocks = num_blocks - start_block_idx
                num_gpu_blocks = num_blocks * self.block_size_factor
```
I plan to add these code before `assert len(req.block_hashes) >= num_gpu_blocks`

PR #27275, opened 2025-10-21
[ResponseAPI] handle json load errors

Jialin:
## Purpose
We engined returned message could fail during json parsing due to one of the following errors:
1) response truncation due to max output length
2) malform response, less likely
3) etc.

## Test Plan & Test Result
N/A

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x] The test plan, such as providing test command.
- [x] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #27675, opened 2025-10-28
[Frontend] [gpt-oss] Tool json call parsing error retry

alecsolder:
## Purpose

Add robust error handling for JSON parsing failures in tool call arguments. When models generate invalid JSON for tool calls, this change allows vLLM to return an error message to the model prompting it to retry with valid JSON, instead of crashing the request.

This behavior is controlled by a new environment variable `TOOL_CALL_JSON_PARSING_AUTOMATIC_RETRY` (enabled by default).
 
Taking over from Jialin on #27275 

## Test Plan
n/a
## Test Result
n/a


yeqcharlotte (approved):
we are having a bit too many environment variables for gpt-oss. we should start thinking about a proper config structure.

PR #27678, opened 2025-10-28
[Misc] Feat/trtllm structured output bench test fixes and updates

pavel4ai:
### Purpose:

Add TRT-LLM backend (OpenAI-compatible) support to the benchmark client:
   - Target /v1/chat/completions or /v1/completions.
Send structured outputs via response_format (json schema, regex, ebnf; choice → JSON Schema enum).
Add optional --api-key (adds Authorization only when provided).
Add optional --debug for request/response diagnostics.
Add optional --validate-schema to strictly validate JSON outputs against the provided schema.
Update inline usage notes and add a docs page: docs/benchmarks/trt_llm_structured_output.md.

### Test Plan
Unit tests (no network) using a fake aiohttp session: tests/benchmarks/test_trt_llm_backend.py

### Test Result
Unit tests: all added tests pass locally.
`pytest -q tests/benchmarks/test_trt_llm_backend.py`


docs/benchmarks/trt_llm_structured_output.md
```diff
@@ -0,0 +1,86 @@
+# Benchmarking TRT-LLM with Structured Outputs (OpenAI-style)
```
benchislett: This is not needed. Please remove this file as the flow is identical to the existing benchmarking, just with `--backend tensorrt-llm`

tests/benchmarks/test_trt_llm_backend.py
```diff
@@ -0,0 +1,251 @@
+# SPDX-License-Identifier: Apache-2.0
```
benchislett: This test is probably overkill too, I don't think it's worth the effort of maintaining this. Please remove.

benchmarks/backend_request_func.py
```diff
@@ -134,68 +136,292 @@ async def async_request_trt_llm(
     request_func_input: RequestFuncInput,
     pbar: tqdm | None = None,
 ) -> RequestFuncOutput:
+    """Send a request to a TensorRT-LLM server exposing OpenAI-style APIs.
```
benchislett, line 139: none of this diff is warranted. It's all duplicated from the openai backend, which we can just use with `--backend openai` or `--backend openai-chat`. There is nothing specific about this function that we don't already have in the `openai-chat` backend.

benchmarks/benchmark_serving_structured_output.py
```diff
@@ -18,6 +18,13 @@
     when using tgi backend, add
         --endpoint /generate_stream
     to the end of the command above.
+
+    when using tensorrt-llm backend (OpenAI-compatible server), use
+        --endpoint /v1/chat/completions   # or /v1/completions
```
benchislett: Also doesn't need to be TRTLLM specific.
```suggestion
    when using an OpenAI-compatible server, use
        --endpoint /v1/chat/completions   # or /v1/completions
```

benchmarks/benchmark_serving_structured_output.py
```diff
@@ -18,6 +18,13 @@
...
+    and optionally:
+        --api-key <token>                 # if the server requires auth
+        --debug                           # verbose request/response logging
```
benchislett: I don't think either of these features should be in-scope for this PR.

benchmarks/benchmark_serving_structured_output.py
```diff
@@ -446,8 +453,58 @@ async def benchmark(
         raise ValueError(f"Unknown backend: {backend}")
 
     def prepare_extra_body(request) -> dict:
+        """Build backend-specific structured output config for a request.
+
+        For `tensorrt-llm` backend, map the internal structure type to the
+        OpenAI-compatible `response_format` field (e.g., JSON schema, regex, or
+        EBNF grammar). For other backends, fall back to the vLLM-native
+        `structured_outputs` shape to preserve existing behavior.
```
benchislett:
```suggestion
        For OpenAI-compatible backends, map the internal structure type to the
        OpenAI-compatible `response_format` field (e.g., JSON schema, regex, or
        EBNF grammar). For the vLLM backend, use the native `structured_outputs` shape to preserve existing behavior.
```

benchmarks/benchmark_serving_structured_output.py
```diff
@@ -446,8 +453,58 @@ async def benchmark(
...
+        """
+        # For tensorrt-llm backend, the server follows OpenAI-style APIs and
+        # expects structured decoding settings in `response_format`.
```
benchislett:
```suggestion
```

benchmarks/benchmark_serving_structured_output.py
```diff
@@ -446,8 +453,58 @@ async def benchmark(
...
+        if backend == "tensorrt-llm":
```
benchislett, line 461:
```suggestion
        if backend == "openai" or backend == "openai-chat":
```

benchmarks/benchmark_serving_structured_output.py
```diff
@@ -446,8 +453,58 @@ async def benchmark(
...
+            structure_type = request.structure_type
+            schema = request.schema
+
+            # Map internal structure types to OpenAI `response_format`.
+            if structure_type == "json":
+                return {
+                    "response_format": {
```
benchislett, line 468: I think the proper way to handle this might be to make `tensorrt-llm` alias to the `openai-chat` request function, just like `vllm` does, so that we can dispatch to the respective styles. Then we can deprecate the former tensorrt-llm backend by giving it a different backend name, such as tensorrt-llm-legacy or something. Or maybe the new one should be "tensorrt-llm-openai".

We need to support:
### OpenAI (proper)
```
{"response_format": {"type": "json_schema", "json_schema": {"name": "test", "schema": schema}}}`
```
When backend is "openai" or "openai-chat" specifically
### OpenAI (TRTLLM style)
```
{"response_format": {"type": "json_schema", "schema": schema}}
```
### vLLM
```
{"structured_outputs": {"json": schema}}
```

benchmarks/benchmark_serving_structured_output.py
```diff
@@ -679,6 +742,21 @@ def _eval_correctness_json(expected, actual):
         except Exception:
             return False
 
+        # Optional: strict schema validation when enabled and schema available
+        if getattr(args, "validate_schema", False) and expected is not None:
```
benchislett, line 736: This is a feature I like a lot

benchmarks/benchmark_serving_structured_output.py
```diff
@@ -700,8 +778,17 @@ def _eval_correctness(expected, actual):
             return None
 
     scores = []
-    for res in ret:
-        score = _eval_correctness(res["expected"], res["generated"])
+    for idx, res in enumerate(ret):
+        expected = res["expected"]
+        # For JSON dataset, expected may be None in ret; fallback to request schema
```
benchislett: I don't know what this code is supposed to be doing. It doesn't look right to me

benchislett (changes requested):
Needs to be trimmed down significantly, and a lot of the TRTLLM-specific changes are really just adding openai support and can be described as such.

benchmarks/benchmark_serving_structured_output.py
xinli-sw: note, that's fixed https://github.com/NVIDIA/TensorRT-LLM/pull/8934#pullrequestreview-3420182036

benchmarks/benchmark_serving_structured_output.py
pavel4ai: artifacts of gpt-based conding when I needed to do some troubleshooting, which turned into extra features. Being removed.

benchmarks/benchmark_serving_structured_output.py
pavel4ai: ack

benchmarks/benchmark_serving_structured_output.py
pavel4ai: revised per suggestion

benchmarks/benchmark_serving_structured_output.py
pavel4ai: adjusted per the suggested change

benchmarks/benchmark_serving_structured_output.py
pavel4ai: not sure I understand. tensorrt-llm is a real backend, like vLLM or SGLang

openai is not at least it is not availble outside of openai.

benchmarks/benchmark_serving_structured_output.py
pavel4ai: removed

benchmarks/benchmark_serving_structured_output.py
pavel4ai: this is part of validation, refactoring so that the functionality is clear

docs/benchmarks/trt_llm_structured_output.md
pavel4ai: removing. it was my understanding that it is mandatory as per contributing guidelines.

tests/benchmarks/test_trt_llm_backend.py
pavel4ai: also removing - yea overkill, but also I understood it was mandatory as per contributing guidelines

benchmarks/backend_request_func.py
pavel4ai: I am confused perhaps, my understanding is that trtllm is the backend and openai is a serving format. Perhaps the terms are being used interchangably in the project ?

benchmarks/benchmark_serving_structured_output.py
pavel4ai: good to see it fixed in trtllm and howfully with eventual merging of this PR, vllm bench can test trtllm properly. I understand this is probably unlikely but for completeness sake perhaps is useful. 

Regarding comment from Ben above I await clarification to my earlier comments on trtllm being the backend and openai being the serving format and how this appears to be used interchangably

benchmarks/benchmark_serving_structured_output.py
benchislett: openai is just a convention of API. "backend" here does not refer to the runtime but just how we query the endpoint. We can reuse all the code from the "openai" backend because all serving engines (vllm, trtllm) expose the LLM via standard /v1/completions (called "openai" here) and /v1/chat/completions (called "openai-chat" here)

Issue #24885, opened 2025-09-15
[RFC] Clarifying vLLM Shutdown Semantics


markmc:
### 🚀 The feature, motivation and pitch

# RFC: Clarifying vLLM Shutdown Semantics

## Introduction

This RFC defines the expected behavior of **vLLM during shutdown**, addressing two distinct use cases:
1. **Library API usage** (primarily offline inference).
2. **Online serving usage** (long-running HTTP server deployments).

Documenting these expectations will reduce surprises for users and help us achieve a consistent implementation across all features. This RFC was motivated by #22295 and #22828.

---

## Library API Shutdown

Library users expect deterministic, resource-safe cleanup when disposing of LLM instances.

### Expectations
* Library users may create and destroy `vllm.LLM` instances.
* Deleting an instance should:
  - Release all associated resources (memory, file descriptors, temporary files, etc.).
  - Shut down any child processes gracefully.
* No library should define global signal handlers.
  - Signal handlers are global and may interfere with applications embedding vLLM.
  - Applications must configure their own signal-handling behavior.

---

## Online Serving Shutdown

Online serving users expect graceful termination semantics, consistent with production environments like Kubernetes, where shutdown is routine due to auto-scaling or rolling upgrades.

Note - vLLM does not support any partial restart scenario, for example a "reload model" or "restart workers" capability.

### Context
* `vllm serve` launches an HTTP server for long-running online inference.
* Each request is relatively expensive (100ms+).
* Deployments typically involve multiple replicas behind a load balancer (e.g. Kubernetes Service).

### Expectations
* **Signal Handling**
  - Parent process receives a `SIGTERM` on shutdown.
  - `SIGINT` (e.g. Ctrl-C) is treated the same as `SIGTERM`.
  - Signals are sent only to the parent; child processes must not act on signals directly.

* **Graceful Quiescence**
  - Stop accepting new HTTP requests immediately.
  - Allow in-flight requests to complete, bounded by a configurable max wait time.
  - Afterward, instruct child processes to exit.

* **Child Processes**
  - Child processes must not make independent shutdown decisions affecting in-flight requests; the parent process determines whether requests are aborted or completed
  - Child processes may delay shutdown for other reasons (e.g. KV transfer).
  - Each process is responsible for reaping its own children via `waitpid()`.

* **Resource Cleanup**
  - Close network sockets, release locks, delete temporary files, etc.

---

## Kubernetes Integration

[...]

PR #27686, opened 2025-10-28
fix(engine): Add death_pipe monitoring to EngineCoreProc for graceful shutdown

RishabhSaini:
## Purpose

This PR addresses an issue where `EngineCoreProc` processes, launched directly by `CoreEngineProcManager` (typically in headless mode, single-server non-Ray deployments, or offline inference), could become orphaned if the parent vLLM process terminated unexpectedly (e.g., via `SIGTERM`, crash). The parent's exit did not automatically propagate to the child engine process, potentially leading to resource leaks.

This commit implements the `death_pipe` mechanism, mirroring the existing implementation used for `WorkerProc` within the `MultiprocExecutor`, to ensure `EngineCoreProc` detects parent termination and shuts down gracefully.

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/v1/engine/core.py
+++ b/vllm/v1/engine/core.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+import contextlib
 import gc
 import os
 import queue
@@ -31,7 +32,7 @@ from vllm.utils.gc_utils import maybe_attach_gc_debug_callback
 from vllm.utils.hashing import get_hash_fn_by_name
 from vllm.utils.import_utils import resolve_obj_by_qualname
 from vllm.utils.network_utils import make_zmq_socket
-from vllm.utils.system_utils import decorate_logs, set_process_title
+from vllm.utils.system_utils import DeathMonitor, decorate_logs, set_process_title
 from vllm.v1.core.kv_cache_utils import (
     BlockHash,
     generate_scheduler_kv_cache_config,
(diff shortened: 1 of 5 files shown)


vllm/v1/engine/core.py
```diff
@@ -736,6 +737,47 @@ def startup_handshake(
...
+                        except Exception as e:
+                            logger.error("Failed to send SIGTERM to self: %s", e)
+                            # Fallback: Directly trigger the event if signal fails
+                            shutdown_event.set()
+
+
+            death_monitor_thread = Thread(
+                target=monitor_parent_death, daemon=True, name="EngineCoreDeathMonitor"
+            )
+            death_monitor_thread.start()
```
markmc: I think we should refactor this out of multiproc executor into `vllm.utils.system_utils` rather than copying and pasting it

e.g.

```
class DeathMonitor:

    def __init__(self, death_pipe: Connection, name: str):
        self._death_pipe = death_pipe
        self._shutdown_event = threading.Event()
        self._monitor_thread = Thread(
            target=self._monitor_parent_death, daemon=True, name=name,
        )

    def _monitor_parent_death(self):
        try:
            # This will block until parent process exits (pipe closes)                                                                                                                   
            self._death_pipe.recv()
        except EOFError:
            # Parent process has exited, terminate this worker                                                                                                                           
            logger.info("Parent process exited, terminating worker")
            # Send signal to self to trigger clean shutdown                                                                                                                              
            self._shutdown_event.set()
        except Exception as e:
            logger.warning("Death monitoring error: %s", e)

    def start(self):
        self._monitor_thread.start()

    @property
    def event(self) -> threading.Event:
        return self._shutdown_event
```

markmc (changes requested):
Thank you for this. See also #24885

vllm/v1/engine/core.py
RishabhSaini: thanks for the review. addressed your concerns

markmc:
xref #32420

PR #27689, opened 2025-10-28
[Frontend] [gpt-oss] Mcp type bug

alecsolder:
## Purpose
For Mcp tool, we should extract server_label instead of type as tool_types.

Besides this major change, we also did a few side improvements:

Make envs.GPT_OSS_SYSTEM_TOOL_MCP_LABELS as a set to speed up existence checks
Move tool_types extractions to a dedicated function for easier maintainances and allow individual unit tests
More unit test added

Taking over from Jialin on #27234  

## Test Plan & Test Result
e2e tests in test_response_api_mcp_tools.py

passed and should be runnable in CI now with new prompts

### DIFF
--- a/tests/entrypoints/openai/test_serving_responses.py
+++ b/tests/entrypoints/openai/test_serving_responses.py
@@ -6,10 +6,19 @@ from unittest.mock import MagicMock
 
 import pytest
 import pytest_asyncio
+from openai.types.responses.tool import (
+    CodeInterpreterContainerCodeInterpreterToolAuto,
+    LocalShell,
+    Mcp,
+    Tool,
+)
 
 from vllm.entrypoints.context import ConversationContext
 from vllm.entrypoints.openai.protocol import ErrorResponse, ResponsesRequest
-from vllm.entrypoints.openai.serving_responses import OpenAIServingResponses
+from vllm.entrypoints.openai.serving_responses import (
+    OpenAIServingResponses,
+    extract_tool_types,
+)
 from vllm.entrypoints.tool_server import ToolServer
 from vllm.inputs.data import TokensPrompt as EngineTokensPrompt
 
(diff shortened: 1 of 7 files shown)


vllm/entrypoints/openai/serving_responses.py
```diff
@@ -106,6 +107,23 @@
 logger = init_logger(__name__)
 
 
+def extract_tool_types(tools: list[Tool]) -> set[str]:
```
yeqcharlotte, line 110: @alecsolder we should follow up on adding proper monitoring for requests using disallowed tools

yeqcharlotte (approved):

tests/entrypoints/openai/test_serving_responses.py
```diff
@@ -62,6 +71,45 @@ def mock_exit_stack():
...
+
+    tools.append(LocalShell(type="local_shell"))
+    assert extract_tool_types(tools) == {"local_shell"}
+
+    tools.append(CodeInterpreterContainerCodeInterpreterToolAuto(type="auto"))
+    assert extract_tool_types(tools) == {"local_shell", "auto"}
+
+    tools.extend(
+        [
+            Mcp(type="mcp", server_label="random", server_url=""),
```
qandrew, line 86: @alecsolder thanks for putting this together! For my understanding, it seems like the `Tools` type is being overloaded. Like the `LocalShell(type="local_shell")` gets extracted from type, while MCP tool `type="mcp", server_label="container"` gets extracted from `server_label.

This is an underlying issue with how openAI overloads type/server_label though?

alecsolder: Yeah, I think what we've noticed is that using the OpenAI types like local_shell is very restrictive, as you can't add any other types of configurations to the tools outside of what OpenAI provides in their types.

When you use the OpenAI types, you are enabling the "thing" that is local_shell. Whether or not that is one tool under the hood or 5 tools, if there is special handling or other configurations, that is a black box due to the types.

By mapping server_label -> tool_type, we are making it easier to have inference code treat specific groups of tools in different ways (like gpt-oss treats browser tool), but not be restricted by only being able to do that while using OpenAI types.

PR #25616, opened 2025-09-24
feat: nsys profiling hooks in v1 engine

djmmoss:
## Purpose

Add hooks to target specific engine steps and use `cudaProfilerApi` when using nsight system.

This is just an initial idea, if there is preferred way to enable this, please feel free to comment.

PR #27693, opened 2025-10-28
Add nsys iteration range profile and iteration ctxt/gen info

Victor49152:
Based on https://github.com/vllm-project/vllm/pull/21814 and updated.

(1) Use VLLM_PROFILE_START_STOP to indicate specfic range. Usage is like export VLLM_PROFILE_START_STOP="800-810".

(2) Print each step iteration context/generation phase requests and tokens number. Enable it by using "VLLM_PROFILE_ITERATION_INFO" environment variable. Like export VLLM_PROFILE_ITERATION_INFO=1.

(3) The nsys profile related code will be gated by a flag self.profile_enabled, which is set to True only when the env variable VLLM_PROFILE_START_STOP is set by the user. 

## Purpose
Most of Nvidia simulation and optimazation tools depends on nsys profile. Currently, to enable nsys profile requires intrusive changes to vllm source code repeatedly on different models, which introduces extra burden to developers. Adding this feature to vllm upstream will greatly benefits all users that has interests in nsys profiling for vllm models.

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/v1/engine/core.py
+++ b/vllm/v1/engine/core.py
@@ -6,6 +6,7 @@ import queue
 import signal
 import threading
 import time
+import torch
 from collections import deque
 from collections.abc import Callable, Generator
 from concurrent.futures import Future
@@ -95,6 +96,16 @@ class EngineCore:
                 VLLM_VERSION,
                 vllm_config,
             )
+
+        self._profile_current_step = 0
+        self.profiler_start_iters, self.profiler_stop_iters = (
+            self._get_profiler_iteration_index_env_var("VLLM_PROFILE_START_STOP")
+        )
+        self.print_profile_step_iteration_info = os.environ.get("VLLM_PROFILE_ITERATION_INFO", "").lower() == "true"
+        # The dict initially records each request's context token size. After processing, the size will decrease.
+        # When this size <= 0, then it means this request is in generation phase. This helps to collect iteration
+        # info.
+        self.schedule_requests: dict[str, int] = {}
 
         self.log_stats = log_stats
 
@@ -313,6 +324,14 @@ class EngineCore:
         if not self.scheduler.has_requests():
             return {}, False
         scheduler_output = self.scheduler.schedule()
+        ### nsys profiling
+        self._profile_current_step = self._profile_current_step + 1
+        if self._profile_current_step in self.profiler_start_iters:
+            torch.cuda.cudart().cudaProfilerStart()
+
+        if self._profile_current_step in self.profiler_stop_iters:
+            torch.cuda.cudart().cudaProfilerStop()
+        self._profile_step_iteration_info(scheduler_output)
 
         with self.log_error_detail(scheduler_output):
             model_output = self.model_executor.execute_model(scheduler_output)
@@ -378,6 +397,14 @@ class EngineCore:
 
         # Block until the next result is available.
         future, scheduler_output = batch_queue.pop()
+        ### nys profiling usage
+        self._profile_current_step = self._profile_current_step + 1
+        if self._profile_current_step in self.profiler_start_iters:
+            torch.cuda.cudart().cudaProfilerStart()
+        if self._profile_current_step in self.profiler_stop_iters:
+            torch.cuda.cudart().cudaProfilerStop()
+        self._profile_step_iteration_info(scheduler_output)
+
         with self.log_error_detail(scheduler_output):
             model_output = future.result()
 
@@ -456,6 +483,58 @@ class EngineCore:
         kwargs: dict[str, Any] | None = None,
     ) -> list[_R]:
         return self.model_executor.collective_rpc(method, timeout, args, kwargs)
+
+    def _get_profiler_iteration_index_env_var(
+        self, name: str
+        ) -> tuple[frozenset[int], frozenset[int]]:
+        spans = os.environ.get(name, None)
+        starts, stops = [], []
+
+        if spans:
+            spans = spans.split(",")
+
+            for span in spans:
+                try:
+                    if "-" in span:
+                        start, stop = span.strip().split("-")
+                        starts.append(int(start))
+                        stops.append(int(stop))
+                    else:
+                        it = int(span.strip())
+                        starts.append(it)
+                        stops.append(it)
+                except ValueError as e:
+                    raise ValueError(
+                        f"Cannot parse span in environment variable `{name}`: {e}"
+                    ) from None
+
+        return frozenset(starts), frozenset(stops)
+
+    def _profile_step_iteration_info(self, scheduler_output: SchedulerOutput):
+        if not self.print_profile_step_iteration_info:
+            return
+
+        for new_req in scheduler_output.scheduled_new_reqs:
+            self.schedule_requests[new_req.req_id] = len(new_req.prompt_token_ids)
+
+        num_ctx_requests = 0
+        num_ctx_tokens = 0
+        num_generation_tokens = 0
+        for req_id, num_tokens in scheduler_output.num_scheduled_tokens.items():
+            if self.schedule_requests[req_id] > 0:
+                num_ctx_requests += 1
+                num_ctx_tokens += num_tokens
+                self.schedule_requests[req_id] -= num_tokens
+            else:
+                num_generation_tokens += 1
+
+        logger.info(
+            f"iter = {self._profile_current_step}, "
+            f"num_scheduled_requests: {num_ctx_requests + num_generation_tokens}, "
+            f"num_ctx_requests: {num_ctx_requests}, "
+            f"num_ctx_tokens: {num_ctx_tokens}, "
+            f"num_generation_tokens: {num_generation_tokens}"
+        )
 
     def preprocess_add_request(self, request: EngineCoreRequest) -> tuple[Request, int]:
         """Preprocess the request.
--- a/vllm/v1/engine/core.py
+++ b/vllm/v1/engine/core.py
@@ -102,11 +102,11 @@ class EngineCore:
             self.profiler_start_iters, self.profiler_stop_iters = (
                 self._get_profiler_iteration_index_env_var("VLLM_PROFILE_START_STOP")
             )
-            self.print_profile_step_iteration_info = os.environ.get("VLLM_PROFILE_ITERATION_INFO", "").lower() == "true"
+            self.print_profile_step_iteration_info = os.environ.get("VLLM_PROFILE_ITERATION_INFO", "0") == "1"
             # The dict initially records each request's context token size. After processing, the size will decrease.
             # When this size <= 0, then it means this request is in generation phase. This helps to collect iteration
             # info.
-        self.schedule_requests: dict[str, int] = {}
+            self.schedule_requests: dict[str, int] = {}
 
         self.log_stats = log_stats
 
--- a/vllm/v1/engine/core.py
+++ b/vllm/v1/engine/core.py
@@ -96,15 +96,16 @@ class EngineCore:
                 VLLM_VERSION,
                 vllm_config,
             )
-
-        self._profile_current_step = 0
-        self.profiler_start_iters, self.profiler_stop_iters = (
-            self._get_profiler_iteration_index_env_var("VLLM_PROFILE_START_STOP")
-        )
-        self.print_profile_step_iteration_info = os.environ.get("VLLM_PROFILE_ITERATION_INFO", "").lower() == "true"
-        # The dict initially records each request's context token size. After processing, the size will decrease.
-        # When this size <= 0, then it means this request is in generation phase. This helps to collect iteration
-        # info.
+        self.profile_enabled = os.environ.get("VLLM_PROFILE_START_STOP", "") != ""
+        if self.profile_enabled:
+            self._profile_current_step = 0
+            self.profiler_start_iters, self.profiler_stop_iters = (
+                self._get_profiler_iteration_index_env_var("VLLM_PROFILE_START_STOP")
+            )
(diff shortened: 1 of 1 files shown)


vllm/v1/engine/core.py
```diff
@@ -95,6 +96,17 @@ def __init__(
                 VLLM_VERSION,
                 vllm_config,
             )
+        self.profile_enabled = os.environ.get("VLLM_PROFILE_START_STOP", "") != ""
+        if self.profile_enabled:
+            self._profile_current_step = 0
```
wangshangsam, line 101:
```suggestion
            self._nsys_profile_current_step = 0
```

vllm/v1/engine/core.py
```diff
@@ -95,6 +96,17 @@ def __init__(
                 VLLM_VERSION,
                 vllm_config,
             )
+        self.profile_enabled = os.environ.get("VLLM_PROFILE_START_STOP", "") != ""
```
wangshangsam, line 99:
```suggestion
        self._nsys_profile_enabled = os.environ.get("VLLM_PROFILE_STEP_SPANS", "") != ""
```

vllm/v1/engine/core.py
```diff
@@ -95,6 +96,17 @@ def __init__(
...
+        if self.profile_enabled:
+            self._profile_current_step = 0
+            self.profiler_start_iters, self.profiler_stop_iters = (
```
wangshangsam, line 102:
```suggestion
            self._nsys_profile_start_steps, self._nsys_profile_stop_steps = (
```

vllm/v1/engine/core.py
```diff
@@ -95,6 +96,17 @@ def __init__(
...
+                self._get_profiler_iteration_index_env_var("VLLM_PROFILE_START_STOP")
+            )
+            self.print_profile_step_iteration_info = os.environ.get("VLLM_PROFILE_ITERATION_INFO", "0") == "1"
```
wangshangsam, line 105:
```suggestion
            self._nsys_profile_step_info_logging_enabled = os.environ.get("VLLM_NSYS_PROFILE_STEP_INFO_LOGGING", "0") == "1"
```
Other than `0/1`, it'd probably be nice to support values like `TRUE/FALSE`, `True/False`, `true/false`, `YES/NO`, ...

vllm/v1/engine/core.py
```diff
@@ -313,6 +325,15 @@ def step(self) -> tuple[dict[int, EngineCoreOutputs], bool]:
...
         scheduler_output = self.scheduler.schedule()
+        ### nsys profiling
+        if self.profile_enabled:
+            self._profile_current_step = self._profile_current_step + 1
+            if self._profile_current_step in self.profiler_start_iters:
+                torch.cuda.cudart().cudaProfilerStart()
+
+            if self._profile_current_step in self.profiler_stop_iters:
+                torch.cuda.cudart().cudaProfilerStop()
+            self._profile_step_iteration_info(scheduler_output)
```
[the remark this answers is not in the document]
wangshangsam: This is a good point.

vllm/v1/engine/core.py
```diff
@@ -313,6 +325,15 @@ def step(self) -> tuple[dict[int, EngineCoreOutputs], bool]:
         if not self.scheduler.has_requests():
             return {}, False
         scheduler_output = self.scheduler.schedule()
+        ### nsys profiling
+        if self.profile_enabled:
+            self._profile_current_step = self._profile_current_step + 1
```
wangshangsam, line 330:
```suggestion
            self._nsys_profile_current_step += 1
```

vllm/v1/engine/core.py
wangshangsam, line 336: Let's only log the step info within a profling span (e.g., 800 to 810).

vllm/v1/engine/core.py
```diff
@@ -378,6 +399,15 @@
 
         # Block until the next result is available.
         future, scheduler_output = batch_queue.pop()
+        ### nys profiling usage
+        if self.profile_enabled:
+            self._profile_current_step = self._profile_current_step + 1
+            if self._profile_current_step in self.profiler_start_iters:
+                torch.cuda.cudart().cudaProfilerStart()
+            if self._profile_current_step in self.profiler_stop_iters:
+                torch.cuda.cudart().cudaProfilerStop()
```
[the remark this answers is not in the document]
wangshangsam: This seems like a good point also?

Maybe a quick way to verify is to test this with asynchronous PP.

vllm/v1/engine/core.py
```diff
@@ -457,6 +487,58 @@ def collective_rpc(
...
+                        starts.append(int(start))
+                        stops.append(int(stop))
+                    else:
+                        it = int(span.strip())
+                        starts.append(it)
+                        stops.append(it)
+                except ValueError as e:
+                    raise ValueError(
+                        f"Cannot parse span in environment variable `{name}`: {e}"
+                    ) from None
```
wangshangsam, line 512:
```suggestion
                    )
```

vllm/v1/engine/core.py
```diff
@@ -457,6 +487,58 @@ def collective_rpc(
     ) -> list[_R]:
         return self.model_executor.collective_rpc(method, timeout, args, kwargs)
 
+    def _get_profiler_iteration_index_env_var(
+        self, name: str
```
wangshangsam, line 491: If we are hardcoding the env var name in the rest of the code to begin with, then parametrizing this helper function seems a little pointless.
If we really want to parametrize something, what would make more sense is to send the value (e.g., `"800-810,1025-1034"`) to this funciton, instead of asking this function to query the env var.

vllm/v1/engine/core.py
```diff
@@ -457,6 +487,58 @@ def collective_rpc(
     ) -> list[_R]:
         return self.model_executor.collective_rpc(method, timeout, args, kwargs)
 
+    def _get_profiler_iteration_index_env_var(
```
wangshangsam, line 490:
```suggestion
    def _get_nsys_profile_step_spans(
```

vllm/v1/engine/core.py
```diff
@@ -457,6 +487,58 @@ def collective_rpc(
...
+            for span in spans:
+                try:
+                    if "-" in span:
+                        start, stop = span.strip().split("-")
+                        starts.append(int(start))
+                        stops.append(int(stop))
+                    else:
+                        it = int(span.strip())
+                        starts.append(it)
+                        stops.append(it)
```
wangshangsam, line 508: I don't think passing an integer `"800"` works as intended. Looking at the code where the cudart APIs are called, it would begin profiling then immediately shut it down :) 
In other words, when a span is passed in (e.g., `800-810`), it's left-inclusive-right-exclusive.

I also think that left-inclusive-right-exclusive is fine (since it matches the array ideology of most programming languages). We don't necessrily needs to support the case of passing in a single integer.

vllm/v1/engine/core.py
```diff
@@ -457,6 +487,58 @@ def collective_rpc(
...
+                except ValueError as e:
+                    raise ValueError(
+                        f"Cannot parse span in environment variable `{name}`: {e}"
+                    ) from None
+
+        return frozenset(starts), frozenset(stops)
+
+    def _profile_step_iteration_info(self, scheduler_output: SchedulerOutput):
```
wangshangsam, line 516:
```suggestion
    def _log_nsys_profile_step_info(self, scheduler_output: SchedulerOutput):
```

vllm/v1/engine/core.py
```diff
@@ -378,6 +399,15 @@ def step_with_batch_queue(
...
         # Block until the next result is available.
         future, scheduler_output = batch_queue.pop()
+        ### nys profiling usage
+        if self.profile_enabled:
+            self._profile_current_step = self._profile_current_step + 1
+            if self._profile_current_step in self.profiler_start_iters:
+                torch.cuda.cudart().cudaProfilerStart()
+            if self._profile_current_step in self.profiler_stop_iters:
+                torch.cuda.cudart().cudaProfilerStop()
+            self._profile_step_iteration_info(scheduler_output)
```
wangshangsam, line 409: Let's only log the step info within a profling span (e.g., 800 to 810).

vllm/v1/engine/core.py
```diff
@@ -457,6 +487,58 @@ def collective_rpc(
...
+                except ValueError as e:
+                    raise ValueError(
+                        f"Cannot parse span in environment variable `{name}`: {e}"
+                    ) from None
+
+        return frozenset(starts), frozenset(stops)
+
+    def _profile_step_iteration_info(self, scheduler_output: SchedulerOutput):
+        if not self.print_profile_step_iteration_info:
+            return
```
wangshangsam, line 518: Normally, people would expect gate keeping to happen in the caller, not the callee.

vllm/v1/engine/core.py
```diff
@@ -457,6 +487,58 @@
...
+                self.schedule_requests[req_id] -= num_tokens
+            else:
+                num_generation_tokens += 1
+
+        logger.info(
+            f"iter = {self._profile_current_step}, "
+            f"num_scheduled_requests: {num_ctx_requests + num_generation_tokens}, "
+            f"num_ctx_requests: {num_ctx_requests}, "
+            f"num_ctx_tokens: {num_ctx_tokens}, "
+            f"num_generation_tokens: {num_generation_tokens}"
```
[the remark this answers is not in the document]
wangshangsam: I don't think I truly understand this comment, but `Ruff (G004)` is a good point.

vllm/v1/engine/core.py
```diff
@@ -457,6 +487,58 @@ def collective_rpc(
...
+    def _profile_step_iteration_info(self, scheduler_output: SchedulerOutput):
+        if not self.print_profile_step_iteration_info:
+            return
+
+        for new_req in scheduler_output.scheduled_new_reqs:
+            self.schedule_requests[new_req.req_id] = len(new_req.prompt_token_ids)
+
+        num_ctx_requests = 0
+        num_ctx_tokens = 0
+        num_generation_tokens = 0
```
wangshangsam:
```suggestion
        num_generation_requests = 0
```

vllm/v1/engine/core.py
```diff
@@ -457,6 +487,58 @@ def collective_rpc(
...
+        return frozenset(starts), frozenset(stops)
+
+    def _profile_step_iteration_info(self, scheduler_output: SchedulerOutput):
+        if not self.print_profile_step_iteration_info:
+            return
+
+        for new_req in scheduler_output.scheduled_new_reqs:
+            self.schedule_requests[new_req.req_id] = len(new_req.prompt_token_ids)
+
+        num_ctx_requests = 0
```
wangshangsam, line 523:
```suggestion
        num_prefill_requests = 0
```

vllm/v1/engine/core.py
wangshangsam, line 525:
```suggestion
        num_decode_tokens = 0
```

vllm/v1/engine/core.py
```diff
@@ -457,6 +487,58 @@ def collective_rpc(
...
+        num_ctx_tokens = 0
```
wangshangsam, line 524:
```suggestion
        num_prefill_tokens = 0
```

vllm/v1/engine/core.py
```diff
@@ -457,6 +487,58 @@ def collective_rpc(
...
+            else:
+                num_generation_tokens += 1
+
+        logger.info(
+            f"iter = {self._profile_current_step}, "
+            f"num_scheduled_requests: {num_ctx_requests + num_generation_tokens}, "
+            f"num_ctx_requests: {num_ctx_requests}, "
+            f"num_ctx_tokens: {num_ctx_tokens}, "
+            f"num_generation_tokens: {num_generation_tokens}"
+        )
```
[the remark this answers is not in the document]
wangshangsam: 1 and 3 are good points.
I don't think I truly understand 2.
4 is somewhat a false alert, but I do agree that the current code is confusing. Maybe `num_decode_tokens += num_tokens` in stead of `+= 1` might make it more explicilt.

wangshangsam (changes requested):
Don't forget to fix ruff errors.

wangshangsam:
> I don't like the range-based profile start/stop interface since we already have a profile interface exposed in the api server that we use for PyTorch profiling. That part of this PR seems similar to #25616
> 
> Regarding batch metrics, I thought @youkaichao added a function for this a long time ago. Maybe VLLM_LOG_BATCHSIZE_INTERVAL?

So, to clarify, what we need is:
1. Let's say that we `cudaProfilerStart()` at step T and `cudaProfilerStop()` at step T+D. What the exact value of T is doesn't matter that much (so long as it's already in the steady state), but, if we redo the same experiment (i.e., launching the same command a second time), we still need `cudaProfilerStart()` exactly at step T and `cudaProfilerStop()` exactly at step T+D.
2. For every step that was captured between `cudaProfilerStart()` and `cudaProfilerStop()`, we need the dimension information (e.g., sequence length) for each step.

@mgoin is this something that could be achieved through https://github.com/vllm-project/vllm/blob/314285d4f2134111345b4c4cbb583a925db9a91d/vllm/entrypoints/openai/api_server.py#L1191-L1203 and `VLLM_LOG_BATCHSIZE_INTERVAL`? I feel that they are related but I haven't wrapped my head around of how to implement that with those two APIs (i.e., it's hard to predict when FastAPI is gonna dispatch which APIs).

I wasn't aware @djmmoss already has a PR for the same purpose, but, on the other hand, you can see how much this functionality matters to NV employees :)

PR #27701, opened 2025-10-29
[Bugfix] properly handle nested json with llama3 tool parser

Aydin-ab:
## Purpose
This PR updates the regex used in the `Llama3JsonToolParser` and `Llama4JsonToolParser`(`vllm/entrypoints/tool_parsers/llama_json.py`) to correctly capture nested JSON structures when parsing tool calls from model responses.

Previously, the regex pattern:
```
r'{[^{}]*(?:{[^{}]*}[^{}]*)*}(?:\s*;\s*{[^{}]*(?:{[^{}]*}[^{}]*)*})*'
```

only supported one level of nested braces.
When a model output included deeper nesting (e.g., parameters containing another dict or nested dicts), the parser failed to match the entire JSON and would match the dict/json at the deepest level.
For example, this model output:
```
'{
    "name": "predict_claim_fraud",
    "parameters": {
        "record": {
            "months_as_customer": "19",
            "age": "42",
            "insured_zip": "Unknown",
        }
    }
}'
```
The tool parser would run the pattern and the match would be
```
'{
    "record": {
        "months_as_customer": "19",
        "age": "42",
        "insured_zip": "Unknown",
    }
}'
```

This caused key fields such as `"name"` or `"parameters"/"arguments"` to be truncated, resulting in the `try:` clause to fail
```
        try:
            json_str = match.group(0)
            # Split by semicolon and strip whitespace
            json_objects = [obj.strip() for obj in json_str.split(';')]

            tool_calls: list[ToolCall] = []
            for json_obj in json_objects:
                if not json_obj:  # Skip empty strings
                    continue
                obj = json.loads(json_obj)
                tool_calls.append(
                    ToolCall(
                        type="function",
                        function=FunctionCall(
                            name=obj["name"],
                            # function call args are JSON but as a string
                            arguments=json.dumps(
                                obj["arguments"]
                                if "arguments" in obj else obj["parameters"],
                                ensure_ascii=False))))

            return ExtractedToolCallInformation(tools_called=True,
                                                tool_calls=tool_calls,
                                                content=None)

        except Exception:
            logger.exception("Error in extracting tool call from response.")
            # return information to just treat the tool call as regular JSON
            return ExtractedToolCallInformation(tools_called=False,
                                                tool_calls=[],
                                                content=model_output)
```

Inside the `except` clause, the tool call gets placed in the `content` field instead of being parsed into a proper `ToolCall` object which is a wrong behavior

This pattern would also fail when some parameters has string values containing brackets { }, those values would be interpreted as another dict which would cause the same issue above

This can cause a lot of complecity overhead on the client because of having to deal with different behaviors depending on the tool used (tool call in `content` field or in `tool_calls`)

The new recursive regex pattern:
```
r'(?<BLOCK>\{(?:"(?:\\.|[^"\\])*"|(?&BLOCK)|[^{}"\\]+)*\})(?:\s*;\s*(?&BLOCK))*'
```
supports arbitrary nesting depth through recursion + handles braces safely inside quoted strings (e.g., `"He said \"Hello {world}\""`).

This ensures the parser robustly extracts complete tool call JSON objects from model responses, even for complex nested payloads.

## Test Plan
- Create a minimal Llama tool-call response containing nested JSON in "parameters":
```
output = '{
    "name": "predict_claim_fraud",
    "parameters": {
        "record": {
            "months_as_customer": "19",
            "age": "42",
            "insured_zip": "Unknown",
        }
    }
}'
```
- Pass this string through Llama3JsonToolParser.extract_tool_calls().
Verify that:
  - The regex successfully matches the entire JSON block.

  - The resulting tool call correctly populates function.name and function.arguments.

  - No fallback to content occurs.

Run the tests
```
pytest -s -v tests/entrypoints/openai/tool_parsers/test_llama3_json_tool_parser.py
```

## Test Result

Arbitrary nested JSON now fully captured and parsed, ToolCall objects correctly populated with name and arguments.
 + No truncation or wrong behavior of putting the tool call in the content field
---
<details>

</details>


vllm/entrypoints/openai/tool_parsers/llama_tool_parser.py
```diff
@@ -60,10 +60,9 @@ def __init__(self, tokenizer: PreTrainedTokenizerBase):
             0
         ]
         # Updated regex to match multiple JSONs separated by semicolons
-        # This pattern is more robust and can handle nested JSON objects
+        # This pattern is more robust and can handle nested JSON objects recursively
         self.tool_call_regex = re.compile(
-            r"{[^{}]*(?:{[^{}]*}[^{}]*)*}(?:\s*;\s*{[^{}]*(?:{[^{}]*}[^{}]*)*})*",
-            re.DOTALL,
+            r'(?<BLOCK>\{(?:"(?:\\.|[^"\\])*"|(?&BLOCK)|[^{}"\\]+)*\})(?:\s*;\s*(?&BLOCK))*'
```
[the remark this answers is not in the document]
Aydin-ab: im not familiar with this version0 vs version1 ? But running the tests work so i assume this is the correct syntax. I don't have any regex compile errors (tested with pytest + deploying myself a llama3 and sending queries)

Aydin-ab:
> We also need to be careful here about the maximum time any specific regex match can make. Is there a maximum depth to the recursion supported? Can a client craft a string as input that causes this regex to take longer than a second to parse?
> 
> cc @wseaton as he previously made changes in this area to ensure regex performance does not negatively impact server responsiveness.

we can add a timeout to the regex [search()](https://github.com/mrabarnett/mrab-regex/blob/dd256cd36e6d7da215bb82be8e9b6e846e1140e5/regex/_main.py#L249) 
in case of a timeout, we can fallback to putting the output into the content, and a log a warning about it

bbrowning:
A timeout seems reasonable - see examples at places like below where our pythonic style tool call parsers use timeouts with a configurable environment variable they respect for the timeout period.

https://github.com/vllm-project/vllm/blob/5b0448104fa16ca28711e715f4866d9b2805f171/vllm/entrypoints/openai/tool_parsers/pythonic_tool_parser.py#L77-L79

vllm/entrypoints/openai/tool_parsers/llama_tool_parser.py
```diff
@@ -80,49 +79,75 @@ def extract_tool_calls(
...
+        # so we don't parse inner brackets
+        end_index = -1
+        tool_calls: list[ToolCall] = []
 
-        try:
-            json_str = match.group(0)
-            # Split by semicolon and strip whitespace
-            json_objects = [obj.strip() for obj in json_str.split(";")]
+        for match in self.tool_call_start_regex.finditer(
+            model_output, timeout=envs.VLLM_TOOL_PARSE_REGEX_TIMEOUT_SECONDS
```
bbrowning: If this `timeout` expires, this can raise a TimeoutError. However, this code is now outside the `try` block which means nothing will catch this exception in the parser. Generally, parsers need to catch their exceptions and do something reasonable vs letting them escape up into the rest of the server code.

Aydin-ab: Nice catch thanks ! updated now, I looked at the llama4_python parser and used their TimeoutError handling for consistency

chaunceyjiang (approved):
Thanks~ @bbrowning @Aydin-ab

PR #27698, opened 2025-10-29
`use_aot_compile` should respect `VLLM_DISABLE_COMPILE_CACHE`

BoyuanFeng:
aot_compile is part of torch.compile. So `use_aot_compile` should respect `VLLM_DISABLE_COMPILE_CACHE`.

cc @zou3519 @ProExpertProg @zhxchen17

PR #27702, opened 2025-10-29
[Graph Partition][Cache] Use inductor partition ops config

BoyuanFeng:
Before this PR, we rely on `_custom_should_partition_fns` to specify ops to partition on. However, it does not work with fx graph cache. So, even with `VLLM_DISABLE_COMPILE_CACHE=1`, changing `splitting_ops` does NOT incur any change on the compiled output code.

This PR fixes the issue by using `torch._inductor.config.custom_should_partition_ops: list[str]`, which composes with fx graph cache. https://github.com/pytorch/pytorch/pull/166458

This PR also reverts #26956 which was a temporary workaround for graph partition x torch.compile cache issue.

Note: This PR also relies on #27698 to land first. Graph partition does NOT compose with aot_compile yet. 

#### Test

Tested on both pytorch-2.9 release and pytorch 2.10.dev nightly.


vllm/env_override.py
```diff
@@ -356,5 +359,7 @@ def _update_scheduler_patched(self) -> None:
     from torch._inductor.codegen.wrapper import PythonWrapperCodegen
     from torch._inductor.graph import GraphLowering
 
+    torch._inductor.config.custom_should_partition_ops: list[str] = []
```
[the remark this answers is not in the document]
BoyuanFeng: This cannot happen since the `custom_should_partition_ops` is added after "2.9.0". But adding the check anyway.

vllm/env_override.py
```diff
@@ -283,9 +282,13 @@ def should_partition_patched(self, node, should_log: bool = False) -> bool:
...
     # Can only do this for FallbackKernels
     ir_node = node.node
-    if isinstance(ir_node, ir.FallbackKernel):
+    if isinstance(ir_node, torch._inductor.ir.FallbackKernel):
         operator = ir_node.op_overload
-        if operator is not None and operator in _custom_should_partition_fns:
+        if (
+            operator is not None
+            and operator.name() in torch._inductor.config.custom_should_partition_ops
+        ):
```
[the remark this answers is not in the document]
BoyuanFeng: both `custom_should_partition_ops` and `operator.name()` has format "mylib::op_name"

vllm/compilation/pass_manager.py
```diff
@@ -144,16 +123,9 @@ def uuid(self):
...
-        }
+        state = {"pass_config": self.pass_config.uuid(), "passes": []}
         for pass_ in self.passes:
             state["passes"].append(pass_.uuid())
         state["passes"].append(self.fix_functionalization.uuid())
 
-        # See [HACK: Bug with Inductor graph partition and torch.compile cache]
-        state["inductor_splitting_ops"].extend(self.inductor_splitting_ops)
-
         return InductorPass.hash_dict(state)
```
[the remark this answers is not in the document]
BoyuanFeng: This PR reverts #26956 and relies on pytorch native cache, since `custom_should_partition_ops` is added to `torch._inductor.config`. Tested on pytorch 2.9 release and pytorch 2.10.dev main branch to validate.

vllm/env_override.py
```diff
@@ -355,6 +366,14 @@ def _update_scheduler_patched(self) -> None:
...
     from torch._inductor.codegen.wrapper import PythonWrapperCodegen
     from torch._inductor.graph import GraphLowering
+    from torch.utils._config_module import _Config, _ConfigEntry
+
+    # `custom_should_partition_ops` is a new config after 2.9.0. So this would
+    # not overwrite any user configs.
+    dummy_config = _Config(default=[])
+    torch._inductor.config._config["custom_should_partition_ops"] = _ConfigEntry(
+        dummy_config
+    )
```
ProExpertProg: Why not this?
```suggestion
    dummy_config = 
    torch._inductor.config._config["custom_should_partition_ops"] = _ConfigEntry(_Config(default=[]))
```

vllm/compilation/partition_rules.py
```diff
@@ -56,47 +53,35 @@ def resolve_defined_ops(op_names: list[str]) -> list["torch._ops.OpOverload"]:
 
 
 @contextlib.contextmanager
-def inductor_partition_rule_context(overloads: list["torch._ops.OpOverload"]):
+def inductor_partition_rule_context(splitting_ops: list[str]):
```
ProExpertProg, line 56: While we're at it could we remove `resolve_op_overloads` from the Dynamo partition path as well and use string names for ops there too?

ProExpertProg (approved):

zou3519 (approved):

vllm/compilation/partition_rules.py
BoyuanFeng: yes I will update dynamo partition path in a follow up pr.

vllm/compilation/partition_rules.py
```diff
@@ -56,47 +53,35 @@ def resolve_defined_ops(op_names: list[str]) -> list["torch._ops.OpOverload"]:
...
 
-    for overload in overloads:
-        register_should_partition_rule(
-            overload,
-            _always_partition,
-        )
+    saved_splitting_ops: list[str] = list(
+        torch._inductor.config.custom_should_partition_ops
+    )
+    torch._inductor.config.custom_should_partition_ops = splitting_ops
```
gmagogsfm, line 76: Is this safe? I believe this config knob is only recently added (last week). Users who use older version of pytorch would hit an error here.

BoyuanFeng: this is only accessed by pytorch 2.9.0 and pytorch 2.9.1. Pytorch 2.9.1 will include the config.
Pytorch 2.9.0 is patched here: https://github.com/vllm-project/vllm/pull/27702/files#diff-ffb2dbecaf604d1ecf2c216667a1f81425d1a593be868c0be141a62f6503b0c6R371-R375

Test passed for both pytorch versions.

PR #27704, opened 2025-10-29
[BUGFIX] Adjust kv block sizes

vadiklyutiy:
## Purpose
1. Adjust kv_manager_block_size according to minimum supported `kernel_block_sizes` by attn backed

2. Temporary remove `16` from `supported_kernel_block_size` for flashinfer due to TRT-LLM Gen attn has a bug for `page_size=16` (https://github.com/flashinfer-ai/flashinfer/issues/1993)

## Details
If `kv_manager_block_size is None` we set the default value `16`. Before I tried to remove `16` due to point 2. above every backend support `kernel_block_size` equal to 16 and everything worked fine. But when no 16 (and less) there is a fail because 
`kv_manager_block_size % kernel_block_size` should be `=0`. I fixed it by adjusting `kv_manager_block_size` with `get_supported_kernel_block_size()`.

### DIFF
--- a/vllm/config/vllm.py
+++ b/vllm/config/vllm.py
@@ -1224,6 +1224,10 @@ class VllmConfig:
         if architecture is None:
             return
 
+        from vllm.model_executor.layers.config import AttentionConfig
+
+        AttentionConfig.verify_and_update_config(self)
+
         from vllm.model_executor.models.config import (
             MODELS_CONFIG_MAP,
             HybridAttentionMambaModelConfig,
--- a/vllm/model_executor/layers/config.py
+++ b/vllm/model_executor/layers/config.py
@@ -37,7 +37,7 @@ class AttentionConfig(VerifyAndUpdateConfig):
             else 16,
         )
 
-        supported_sizes = backend_cls.get_supported_kernel_block_size()
+        supported_sizes = backend_cls.get_supported_kernel_block_sizes()
         supported_sizes = [
             s.base if isinstance(s, MultipleOf) else s for s in supported_sizes
         ]
--- a/vllm/model_executor/layers/config.py
+++ b/vllm/model_executor/layers/config.py
@@ -52,7 +52,7 @@ class AttentionConfig(VerifyAndUpdateConfig):
         else:
             new_block_size = lcm(cache_config.block_size, min_size)
 
-        if new_block_size != cache_config.block_size:
+        if cache_config.block_size is None or new_block_size != cache_config.block_size:
             cache_config.block_size = new_block_size
             logger.info(
                 "Setting attention block size to %d tokens "
--- a/vllm/model_executor/layers/config.py
+++ b/vllm/model_executor/layers/config.py
@@ -1,0 +1,62 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+from math import lcm
+from typing import TYPE_CHECKING
+
+from vllm.attention.backends.abstract import MultipleOf
+from vllm.attention.selector import get_attn_backend
+from vllm.logger import init_logger
+from vllm.model_executor.models.config import VerifyAndUpdateConfig
+
+if TYPE_CHECKING:
+    from vllm.config import VllmConfig
+
+logger = init_logger(__name__)
+
+
+class AttentionConfig(VerifyAndUpdateConfig):
+    @classmethod
+    def verify_and_update_config(cls, vllm_config: "VllmConfig") -> None:
+        """
+        Align cache_config.block_size with attention backend's minimum
+        supported kernel block size.
+
+        Args:
+            vllm_config: vLLM Config
+        """
+        model_config = vllm_config.model_config
+        cache_config = vllm_config.cache_config
+        assert cache_config is not None
+
+        backend_cls = get_attn_backend(
+            head_size=model_config.get_head_size(),
+            dtype=model_config.dtype,
+            kv_cache_dtype=cache_config.cache_dtype,
+            block_size=cache_config.block_size
+            if cache_config.block_size is not None
+            else 16,
+        )
+        # For now enable it for FlashInfer only.
+        # Other backend need debugging.
+        # TODO: enable it for all backends.
+        if backend_cls.get_name() != "FLASHINFER":
+            return
+
+        supported_sizes = backend_cls.get_supported_kernel_block_size()
+        supported_sizes = [
+            s.base if isinstance(s, MultipleOf) else s for s in supported_sizes
+        ]
+        min_size = min(supported_sizes)
+        if cache_config.block_size is None:
+            new_block_size = min_size
+        else:
+            new_block_size = lcm(cache_config.block_size, min_size)
+
+        if new_block_size != cache_config.block_size:
+            cache_config.block_size = new_block_size
+            logger.info(
+                "Setting attention block size to %d tokens "
+                "to align with %s attention backend's supported kernel block sizes.",
+                new_block_size,
+                backend_cls.get_name(),
+            )
--- a/vllm/model_executor/layers/config.py
+++ b/vllm/model_executor/layers/config.py
@@ -36,11 +36,6 @@ class AttentionConfig(VerifyAndUpdateConfig):
             if cache_config.block_size is not None
             else 16,
         )
-        # For now enable it for FlashInfer only.
-        # Other backend need debugging.
-        # TODO: enable it for all backends.
-        if backend_cls.get_name() != "FLASHINFER":
-            return
 
         supported_sizes = backend_cls.get_supported_kernel_block_size()
         supported_sizes = [
--- a/vllm/model_executor/models/config.py
+++ b/vllm/model_executor/models/config.py
@@ -359,7 +359,11 @@ class HybridAttentionMambaModelConfig(VerifyAndUpdateConfig):
                 dtype=kv_cache_dtype,
             ).page_size_bytes
         else:
-            kernel_block_alignment_size = 16
+            if cache_config.block_size is not None:
+                kernel_block_alignment_size = cache_config.block_size
+            else:
+                kernel_block_alignment_size = 16
+            
             if (
                 current_platform.is_device_capability(100)
                 and model_config.get_head_size() == 256
--- a/vllm/model_executor/models/config.py
+++ b/vllm/model_executor/models/config.py
@@ -363,7 +363,7 @@ class HybridAttentionMambaModelConfig(VerifyAndUpdateConfig):
                 kernel_block_alignment_size = cache_config.block_size
             else:
                 kernel_block_alignment_size = 16
-            
+
             if (
                 current_platform.is_device_capability(100)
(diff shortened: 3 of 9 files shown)


vadiklyutiy:
CC @zhiyuan1i @heheda12345

vllm/model_executor/models/config.py
```diff
@@ -427,6 +427,7 @@ def lcm(a, b):
         # user has not set it or (b) the user has set it
         # too small.
         if cache_config.block_size is None or cache_config.block_size < attn_block_size:
+            attn_block_size = next_power_of_2(attn_block_size)
```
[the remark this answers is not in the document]
vadiklyutiy: not correct

vllm/model_executor/models/config.py
```diff
@@ -427,6 +427,7 @@ def lcm(a, b):
...
             cache_config.block_size = attn_block_size
```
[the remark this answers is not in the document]
vadiklyutiy: fixed

pavanimajety:
@vadiklyutiy Could we test a mamba model and qwen3 next here with mini tests to see that both paths work?

vadiklyutiy:
> @vadiklyutiy Could we test a mamba model and qwen3 next here with mini tests to see that both paths work?

I tested qwen3-next. The accuracy bug with trt-llm gen attn "fixed"

pavanimajety (approved):
LGTM, please add the evals.

vllm/model_executor/models/config.py
```diff
@@ -411,6 +411,7 @@ def lcm(a, b):
 
             chunk_size = lcm(base_chunk_size, kernel_block_alignment_size)
             attn_block_size = chunk_size * cdiv(attn_tokens_per_mamba_state, chunk_size)
+            attn_block_size = next_power_of_2(attn_block_size)
```
heheda12345: Prefer to add hardcode to flashinfer like https://github.com/vllm-project/vllm/pull/27843 power_of_2 can add unnecessary memory waste.

vadiklyutiy: I’ve updated this to align with #27843: adjust `alignment_size` instead of rounding to the next power of two.

tests/v1/kv_offload/test_cpu_offloading.py
```diff
@@ -13,7 +13,7 @@
 from vllm.config import KVEventsConfig, KVTransferConfig
 from vllm.distributed.kv_events import BlockStored, KVEventBatch
 
-CPU_BLOCK_SIZES = [16, 48]
+CPU_BLOCK_SIZES = [32, 48]
```
heheda12345: why do you need this change?

vllm/platforms/cuda.py
```diff
@@ -113,7 +113,7 @@ def check_and_update_config(cls, vllm_config: "VllmConfig") -> None:
 
         cache_config = vllm_config.cache_config
         if cache_config and cache_config.block_size is None:
-            cache_config.block_size = 16
+            cache_config.block_size = 32
```
heheda12345: please only change the default value for flash infer

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -173,7 +173,10 @@ def get_supported_kernel_block_size() -> list[int | MultipleOf]:
         # Note: Not sure for all platforms,
         # but on Blackwell, only support a page size of
         # 16, 32, 64
-        return [16, 32, 64]
+        # TODO: 16 is temporary removed because TRT-LLM kernel has a bug when using 16.
```
heheda12345, line 291: if the problem only exist in trtllm, what about only remove 16 for trtllm like
```
if current_platform.is_device_capability(100):
    return [32, 64]
else:
    return [16, 32, 64]
```

mgoin: Agreed, we should only override on Blackwell

vadiklyutiy:
@codex review pls

vllm/model_executor/layers/config.py
```diff
@@ -0,0 +1,62 @@
...
+        supported_sizes = [
+            s.base if isinstance(s, MultipleOf) else s for s in supported_sizes
+        ]
+        min_size = min(supported_sizes)
+        if cache_config.block_size is None:
+            new_block_size = min_size
+        else:
+            new_block_size = lcm(cache_config.block_size, min_size)
+
+        if new_block_size != cache_config.block_size:
```
MengqingCao: I think we should check if `cache_config.block_size` is `None` firstly here?

vllm/model_executor/layers/config.py
```diff
@@ -0,0 +1,62 @@
...
+            dtype=model_config.dtype,
+            kv_cache_dtype=cache_config.cache_dtype,
+            block_size=cache_config.block_size
+            if cache_config.block_size is not None
+            else 16,
+        )
+        # For now enable it for FlashInfer only.
+        # Other backend need debugging.
+        # TODO: enable it for all backends.
+        if backend_cls.get_name() != "FLASHINFER":
```
MengqingCao: Hope this could be done in this pr, I think this is safe to enable for all backends?

vllm/model_executor/layers/config.py
MengqingCao: And I think whether `cache_config.block_size` is `None` or not at this point, we should overwrite it

vllm/config/vllm.py
```diff
@@ -880,6 +880,10 @@ def try_verify_and_update_config(self):
         if architecture is None:
             return
 
+        from vllm.model_executor.layers.config import AttentionConfig
+
+        AttentionConfig.verify_and_update_config(self)
```
MengqingCao, line 1229: I think we could do this only when `self.model_config.is_hybrid` is `True`?

vllm/model_executor/layers/config.py
vadiklyutiy: a lot of fails in CI. debugging

vllm/config/vllm.py
vadiklyutiy: why don't do it for non hybrid models?

vllm/model_executor/layers/config.py
vadiklyutiy: agree

vadiklyutiy: fixed

vllm/platforms/cuda.py
vadiklyutiy: it was temporary. removed

tests/v1/kv_offload/test_cpu_offloading.py
vadiklyutiy: it was temporary. removed

vllm/config/vllm.py
heheda12345: It's OK for me to do it for non hybrid model. But can you remove the hardcode like if using FlashMLA, choose block_size 64? And is updating block_size here early enough? (I'm afraid that there are other config update logic related to block_size)

vllm/model_executor/layers/config.py
heheda12345: +1 hope this PR can enable all backends.

vllm/config/vllm.py
heheda12345: The handling of block_size is None is determined by each platform.
https://github.com/vllm-project/vllm/blob/1fb4217a052189feb8709b67bb3209ab316d13b7/vllm/config/cache.py#L39-L41

Given this, will it be a better idea to call `AttentionConfig.verify_and_update_config(self)` in `platforms/cuda.py` and let other platform to use this if they want?

@MengqingCao Preference from ascend side?

vllm/model_executor/layers/config.py
```diff
@@ -0,0 +1,62 @@
...
+
+        supported_sizes = backend_cls.get_supported_kernel_block_size()
+        supported_sizes = [
+            s.base if isinstance(s, MultipleOf) else s for s in supported_sizes
+        ]
+        min_size = min(supported_sizes)
+        if cache_config.block_size is None:
+            new_block_size = min_size
+        else:
+            new_block_size = lcm(cache_config.block_size, min_size)
```
heheda12345, line 48: Prefer to raise an error if the user sets a block_size but the block_size is not supported by the attention backend it selects.

vllm/model_executor/layers/config.py
```diff
@@ -0,0 +1,62 @@
...
+
+        if cache_config.block_size is None or new_block_size != cache_config.block_size:
```
heheda12345, line 50: I think we don't need to add info-level logging if block_size is None and is initialized normally.

vllm/model_executor/models/config.py
```diff
@@ -363,7 +363,10 @@ def verify_and_update_config(cls, vllm_config: "VllmConfig") -> None:
                 dtype=kv_cache_dtype,
             ).page_size_bytes
         else:
-            kernel_block_alignment_size = 16
+            if cache_config.block_size is not None:
```
heheda12345, line 362: is this part called before or after AttentionConfig.verify_and_update_config(self)?

I think the kernel_block_alignment_size should be resolved from backend_cls.get_supported_kernel_block_size

vadiklyutiy:
closed by mistake

vllm/config/vllm.py
MengqingCao: Oh, sorry, I just considered of overwriting `kernel_block_alignment_size` and missed the main propose of your pr. plz ignore this.

vadiklyutiy: no problem

MengqingCao: > The handling of block_size is None is determined by each platform.
> 
> https://github.com/vllm-project/vllm/blob/1fb4217a052189feb8709b67bb3209ab316d13b7/vllm/config/cache.py#L39-L41
> 
> Given this, will it be a better idea to call `AttentionConfig.verify_and_update_config(self)` in `platforms/cuda.py` and let other platform to use this if they want?
> 
> @MengqingCao Preference from ascend side?

From ascend side, I think it's better to call `AttentionConfig.verify_and_update_config(self)` in `VllmConfig.try_verify_and_update_config`

During the post init of `VllmConfig`, we do `VllmConfig.try_verify_and_update_config` first and `Platform.check_and_update_config` later. The former makes it take effect on `HybridAttentionMambaModelConfig`, thereby adjusting`kernel_block_alignment_size`, a value that can be controlled via `AttentionBackend.get_supported_kernel_block_size`, and the latter allows all platforms to override this value in certain special cases.

But I think your concern is right, this maybe change the default logic of setting block_size, we should be careful to get the default `CacheConfig.block_size`, maybe we should leave an api like `Platform.check_and_update_cacheconfig` to deal with the logic of platform-specific block-size setting.

In the long run, I think block size is more of a property strongly correlated with the attention backend, and it would be best to move the relevant logic to each attention backend. Of course, since GPUs have both NV and ROCM, I'm not sure if this direction is correct.

vadiklyutiy:
@MagnetoWang don't have bandwidth to continue. Would be glad if you can continue. Thank you for your help.

PR #23207, opened 2025-08-19
[Misc][qwen2_5_vl][torch.compile] Enable `supports_torch_compile` on generic nn.Module and demonstrate speedup on Qwen Vision model

Lucaskabela:
We enable `@supports_torch_compile` on generic nn.Modules (as opposed to only top level architecture) and demonstrate the application in qwen_2_5_vl

## Purpose
This PR is a first step towards supporting torch compile for multimodal encoders (such as vision or audio).  Since these modality specific components do not have vLLM config, we begin by taking a step to allow these modules to be compiled

### Unit Test
```
with-proxy pytest tests/compile/test_multimodal_compile.py
```
```
1 passed, 9 warnings in 45.26s
```
### E2E Tests
```
VLLM_USE_V1=1 python examples/offline_inference/vision_language.py -m qwen2_5_vl
```
### BEFORE CHANGES:
```
^[[1;36m(EngineCore_0 pid=2638689)^[[0;0m INFO 08-19 14:14:20 [backends.py:559] Dynamo bytecode transform time: 3.26 s
^[[1;36m(EngineCore_0 pid=2638689)^[[0;0m INFO 08-19 14:14:22 [backends.py:161] Directly load the compiled graph(s) for dynamic shape from the cache, took 1.965 s
^[[1;36m(EngineCore_0 pid=2638689)^[[0;0m INFO 08-19 14:14:23 [monitor.py:34] torch.compile takes 3.26 s in total
...

Processed prompts:   0%|          | 0/4 [00:00<?, ?it/s, est. speed input: 0.00 toks/s, output: 0.00 toks/s]
Processed prompts:  25%|██▌       | 1/4 [00:00<00:02,  1.41it/s, est. speed input: 1800.03 toks/s, output: 90.28 toks/s]
Processed prompts: 100%|██████████| 4/4 [00:00<00:00,  1.41it/s, est. speed input: 7137.36 toks/s, output: 357.98 toks/s]
Processed prompts: 100%|██████████| 4/4 [00:00<00:00,  5.59it/s, est. speed input: 7137.36 toks/s, output: 357.98 toks/s]
```
### AFTER CHANGES:
```
^[[1;36m(EngineCore_0 pid=2569806)^[[0;0m INFO 08-19 14:12:31 [backends.py:559] Dynamo bytecode transform time: 0.37 s
[...]

PR #27705, opened 2025-10-29
[Model] Fix Qwen3VL and Qwen3Omni after torch.compile changes

lgeiger:
## Purpose

#23207 Broke Qwen3VL since it relies on `Qwen2_5_VisionAttention` as well but still used the old signature:
```
RuntimeError: vllm::flash_attn_maxseqlen_wrapper() Expected a value of type 'Tensor' for argument 'max_seqlen' but instead found type 'int'
```

This PR updates Qwen3VL and Qwen3Omni to fix this.

## Test Plan

```
vllm serve Qwen/Qwen3-VL-2B-Instruct-FP8
```

Note: I didn't test Qwen3OmniMoE since I didn't have a large enough GPU on hand today, but it's the same code changes as for Qwen3VL so I believe it should be fine.

/cc @Lucaskabela @ywang96

I'm happy to make a followup PR to add `torch.compile` support to the other modules in Qwen3VL tomorrow once I get a chance to run some benchmarks.

### DIFF
--- a/vllm/model_executor/models/qwen2_5_vl.py
+++ b/vllm/model_executor/models/qwen2_5_vl.py
@@ -836,10 +836,8 @@ class Qwen2_5_VisionTransformer(nn.Module):
         self,
         cu_seqlens: torch.Tensor,
     ) -> tuple[torch.Tensor, torch.Tensor]:
-        max_seqlen, seqlens = (
-            torch.zeros(1, device=cu_seqlens.device),
-            torch.zeros(1, device=cu_seqlens.device),
-        )
+        max_seqlen = torch.zeros([], device=cu_seqlens.device)
+        seqlens = torch.zeros(1, device=cu_seqlens.device)
         if (
             self.attn_backend == _Backend.FLASH_ATTN
             or self.attn_backend == _Backend.ROCM_AITER_FA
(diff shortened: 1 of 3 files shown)


vllm/model_executor/models/qwen2_5_vl.py
```diff
@@ -836,10 +836,8 @@ def compute_attn_mask_seqlen(
         self,
         cu_seqlens: torch.Tensor,
     ) -> tuple[torch.Tensor, torch.Tensor]:
-        max_seqlen, seqlens = (
-            torch.zeros(1, device=cu_seqlens.device),
-            torch.zeros(1, device=cu_seqlens.device),
-        )
+        max_seqlen = torch.zeros([], device=cu_seqlens.device)
+        seqlens = torch.zeros(1, device=cu_seqlens.device)
```
lgeiger, line 840: `max_seqlen` is actually a scalar value so this updates the dummy values to match this.

Lucaskabela (approved):
Thanks for the fix @lgeiger! I was planning on making may way through to Qwen3, but didn't realize this PR had (indirectly) impacted those files

Given that but no test failed, do we have CI coverage for Qwen3 models? (cc @ywang96 @ProExpertProg  @zou3519 )

If not it might be worth adding it to the basic initialization suite of tests

ywang96 (approved):
Thanks for fixing this! Stamping it now and I will test Qwen3-Omni on my end.

> Given that but no test failed, do we have CI coverage for Qwen3 models?

Yea we didn't add them onto CI back then because the model support PR was merged before the model release, but we should definitely add one now that the smaller version of the models are available. I will take that as an action item

PR #27224, opened 2025-10-20
[ROCm][MLA] Support block-size > 1 for AITER MLA backend 

ganyi1996ppo:
## Purpose
The `AITERMLABackend` now only support `block-size=1` scenario for inference. This constrain may lead to some serious host overhead when we are about to allocate or free cache blocks for long context requests cause there might exist large amount of blocks to operate. Thanks to the insights of @gyu-amd .

In this PR, we remapping the `block_table` to 1 block size case every step in `AITERMLAMetadataBuilder` to alleviate the host overhead during allocate and deallocate blocks This change also helps to support a wider range of block size for `AITERMLABackend`, makes the `AITERMLABackend` on ROCm platform aligns with the vllm's official usgae and more flexible .

## Test Plan
Verified on gsm8k for accuracy, performance improvement will also be attached later
test script:
```

export VLLM_USE_V1=1
export SAFETENSORS_FAST_GPU=1
export VLLM_ROCM_USE_AITER=1
export VLLM_ROCM_USE_AITER_MOE=1
export VLLM_USE_TRITON_FLASH_ATTN=0
export NCCL_DEBUG=WARN
export VLLM_RPC_TIMEOUT=1800000
export VLLM_ROCM_USE_AITER_ASMMOE=1
export VLLM_ROCM_USE_AITER_MHA=0
export VLLM_ROCM_USE_TRITON_ROPE=1

model_path="deepseek-r1-FP8-Dynamic"
vllm serve $model_path \
  --tensor-parallel-size 8 \
  --max-num-batched-tokens 32768 \
  --trust-remote-code \
  --no-enable-prefix-caching \
  --disable-log-requests \
  --gpu_memory_utilization 0.9 \
  --block-size 128 \
  --compilation-config '{"cudagraph_mode": "FULL_AND_PIECEWISE"}'
```
## Test Result

```
# gsm8k test
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9507|±  |0.0060|
|     |       |strict-match    |     5|exact_match|↑  |0.9484|±  |0.0061|
```

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
[...]

PR #27715, opened 2025-10-29
[AMD] Use Decoupled Kernel Block Size to Support AITER MLA block_size=1

zq1997:
PR #24486 implemented this mechanism but did not fully utilize it. 
The block_size restriction during startup (when using operators such as FlashMla and AiterMLA) has now been lifted.

Compared to #27224, this PR avoids remapping before each operator execution, instead making use of the functionality provided by the framework (by #24486), and also provides support for scenarios other than AMD aiter.

### DIFF
--- a/vllm/platforms/cuda.py
+++ b/vllm/platforms/cuda.py
@@ -201,18 +201,23 @@ class CudaPlatformBase(Platform):
 
             from vllm.attention.ops.flashmla import is_flashmla_dense_supported
 
+            def round_up_block_size(size):
+                cache_config.block_size = (
+                    (cache_config.block_size + size - 1) // size * size
+                )
+
             if (
                 use_flashmla
                 and is_flashmla_dense_supported()[0]
                 and cache_config.block_size % 64 != 0
             ):
-                cache_config.block_size = 64
-                logger.info("Forcing kv cache block size to 64 for FlashMLA backend.")
+                round_up_block_size(64)
+                logger.info("Forcing kv cache block size to 64*N for FlashMLA backend.")
 
             if use_cutlass_mla and cache_config.block_size % 128 != 0:
-                cache_config.block_size = 128
+                round_up_block_size(128)
                 logger.info(
-                    "Forcing kv cache block size to 128 for CUTLASS_MLA backend."
+                    "Forcing kv cache block size to 128*N for CUTLASS_MLA backend."
                 )
 
             if (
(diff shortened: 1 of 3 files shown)


vllm/v1/attention/backends/mla/rocm_aiter_mla.py
```diff
@@ -78,14 +82,10 @@ def __init__(
         super().__init__(
             kv_cache_spec, layer_names, vllm_config, device, AiterMLAMetadata
         )
```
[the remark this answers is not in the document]
zq1997: self.kv_cache_spec is kv_page block size (may not be 1). In fact, kernel page size is always 1

youkaichao:
cc @heheda12345

vllm/platforms/cuda.py
```diff
@@ -201,35 +201,40 @@ def check_and_update_config(cls, vllm_config: "VllmConfig") -> None:
...
+                )
+
             if (
                 use_flashmla
                 and is_flashmla_dense_supported()[0]
                 and cache_config.block_size % 64 != 0
             ):
-                cache_config.block_size = 64
-                logger.info("Forcing kv cache block size to 64 for FlashMLA backend.")
+                round_up_block_size(64)
```
heheda12345: do you have to change them in this PR?

vllm/v1/attention/backends/mla/rocm_aiter_mla.py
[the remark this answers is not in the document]
heheda12345: I think you can keep this assert as we pass kv_cache_spec with updated kernel block size here.

vllm/platforms/cuda.py
```diff
@@ -359,7 +364,7 @@ def get_attn_backend_cls(
                     head_size,
                     dtype,
                     kv_cache_dtype,
-                    None,
+                    block_size,
```
heheda12345: @MatthewBonanni why was it None here?

vllm/platforms/cuda.py
zq1997: Personal opinion: This might be helpful, or at least unlikely to make things worse.

The main purpose of the "Decoupled Kernel Block Size" feature is to use larger-granularity KV blocks at the framework level, without being limited by the kernel-level block size (although the two still need to be in a multiple relationship). Therefore, users can specify `--block_size=128, 192, 256...` for FlashMLA. While this case may not be very frequent, it's still best to provide an option to accept a large user-specified block_size rather than ignoring it and forcibly modifying it to a size of 64.

heheda12345: can you only focus on the necessary changes for AMD support?

zq1997: Could any concern or potential risk be implied here that would advise against making adjustments to CUDA platform?

Our motivation for this change arises from practical deployment considerations. We're experimenting with heterogeneous (NV & AMD) prefill-decode deployments, and support for a configuration using block_size=128 together with FlashMLA would provide meaningful benefits for this scenario.

vllm/platforms/cuda.py
MatthewBonanni: It was `None` because I didn't want an unsupported `block_size` to be the reason a backend was removed from consideration (`block_size` can just be adjusted). Given that the block size adjustment happens before calling `get_attn_backend_cls`, though, this change should be fine.
NOTE: having to adjust the block size before knowing what backend will actually be selected makes `check_and_update_config` too complicated and I'd like to improve this in the future

vllm/platforms/cuda.py
heheda12345: Instead of hardcode 64 / 128 for each attention backend, I prefer to refactor this part of code to use the `supported_kernel_block_sizes`. An ongoing effort is https://github.com/vllm-project/vllm/pull/27704 but you can also take this part if you want. Before that, I don't want a temporary fix. For block_size=128, I think it satisfies `cache_config.block_size % 64 != 0` and won't hit this branch.

vllm/platforms/cuda.py
zq1997: I have reverted this file, since the way mentioned above is more appropriate.

heheda12345 (approved):
LGTM!

PR #27719, opened 2025-10-29
[BugFix] Fix handling of resumed reqs in `SharedStorageConnector`

njhill:
The current logic assumes that resumed requests precede other cached requests, but since #20232 that's no longer the case.


ApostaC (approved):
LGTM!

PR #27731, opened 2025-10-29
[EPLB][ROCm]: support EPBL for ROCm backend

PerryZhang01:
## Purpose
This PR supports EPLB for the ROCm backend, achieving feature parity with the existing CUDA implementation. The implementation validated on DeepSeekR1.
## Test Plan
we try to enable EPLB on DeepSeekR1 on MI355 with the following parameters.
```
server:
vllm serve $model_path \
--tensor-parallel-size 8 \
--max-num-batched-tokens 32768 \
--trust-remote-code \
--no-enable-prefix-caching \
--disable-log-requests \
--compilation-config '{"cudagraph_mode": "FULL_AND_PIECEWISE"}' \
--gpu_memory_utilization 0.8 \
--block-size 1 \
--enable-expert-parallel \
--enable-eplb \
--num-redundant-experts 8 \
--eplb-log-balancedness \
--eplb-window-size 3000 \
--eplb-step-interval 1000
```
```
client:
python -m vllm.entrypoints.cli.main bench serve \
    --host localhost \
    --port 8000 \
    --model ${model_path} \
    --dataset-name random \
    --random-input-len 1024 \
    --random-output-len 1024 \
    --max-concurrency 64 \
    --num-prompts 128 \
    --seed 123 \
    --percentile-metrics ttft,tpot,itl,e2el \
    --ignore-eos
```
## Test Result
Benchmark Result:  
image
After using eplb, the `balancedness` metric increased from 0.55 to 0.65 with the random data.  However, the `avg_tokens` is not enough in decode phase, so that the `balancedness` metric dropped to 0.3.

Besides, we use `lm_eval ` to validate the accuracy of `EPLB` on `gsm8k` datasets, the result as below:
```shell
    |gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9492|±  |0.0060|
    |     |       |strict-match    |     5|exact_match|↑  |0.9492|±  |0.0060|
```


vllm/config/parallel.py
```diff
@@ -267,10 +267,10 @@ def _validate_parallel_config(self) -> Self:
             )
 
         if self.enable_eplb:
-            if not current_platform.is_cuda():
+            if not (current_platform.is_cuda() or current_platform.is_rocm()) :
```
ganyi1996ppo:
```suggestion
            if not (current_platform.is_cuda_alike()) :
```

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1944,7 +1945,7 @@ def load_weights(
 
     def get_expert_weights(self) -> Iterable[torch.Tensor]:
         weights = list(self.named_parameters())
-        assert all(weight.is_contiguous() for _, weight in weights)
+        # assert all(weight.is_contiguous() for _, weight in weights)
```
ganyi1996ppo: Why comments here?

vllm/config/parallel.py
PerryZhang01: thanks for that! are there any other platforms like cuda? if not, we can use the `is_cuda_alike()` interface.

tjtanaa: `is_cuda_alike()` at the moment refers to `cuda` and `rocm`.

So let's use `is_cuda_alike()`

vllm/model_executor/layers/fused_moe/layer.py
PerryZhang01: changed

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1944,23 +1954,22 @@ def load_weights(
...
-        # `e_score_correction_bias` is a bias for each logical expert,
-        # with shape (num_logical_experts,), not an expert weight.
-        NON_EXPERT_WEIGHTS = {
-            "e_score_correction_bias",
+        ROUTED_EXPERT_WEIGHTS = {
+            "w13_weight",
+            "w13_weight_scale",
+            "w2_weight",
+            "w2_weight_scale",
         }
```
hmellor: Won't this break compatibility with quantization schemes which use `qweight`? Why does this need to change?

hmellor (changes requested):

vllm/config/parallel.py
PerryZhang01: that`s right!

vllm/model_executor/layers/fused_moe/layer.py
PerryZhang01: yea, the shared experts are not contiguous in deepseekR1 currently, it will cause assert error. u r right, we can`t change the filter strategy for the other models, so I reverted to the origin code, and only changed the assert part.

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1318,9 +1318,19 @@ def __init__(
                 )
 
         if self.enable_eplb:
+            from vllm.model_executor.layers.quantization.compressed_tensors import (
+                compressed_tensors_moe,
+            )
```
hmellor: Can we add

```py
from vllm.model_executor.layers.quantization.compressed_tensors.compressed_tensors_moe import CompressedTensorsW8A8Fp8MoEMethod
```

to `vllm/model_executor/layers/quantization/compressed_tensors/__init__.py` (`pre-commit` will probably want it written slightly differently) and then import

```suggestion
            from vllm.model_executor.layers.quantization.compressed_tensors import (
                CompressedTensorsW8A8Fp8MoEMethod,
            )
```

It seems strange to import the Python module when the other quant methods are directly imported here

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1318,9 +1318,19 @@ def __init__(
...
             from vllm.model_executor.layers.quantization.fp8 import Fp8MoEMethod
 
-            if not isinstance(quant_method, (Fp8MoEMethod, UnquantizedFusedMoEMethod)):
+            if not isinstance(
+                quant_method,
+                (
+                    Fp8MoEMethod,
+                    UnquantizedFusedMoEMethod,
+                    compressed_tensors_moe.CompressedTensorsW8A8Fp8MoEMethod,
```
hmellor:
```suggestion
                   CompressedTensorsW8A8Fp8MoEMethod,
```

vllm/model_executor/layers/fused_moe/layer.py
PerryZhang01: yea, it is too long to pass the pre-commit check. we have added it to the `__init__` file as your suggestion, now it looks better.

tjtanaa (approved):
LGTM

hmellor (commented):
Looks much cleaner now, just one thing I'm not sure about

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1944,7 +1954,11 @@ def load_weights(
 
     def get_expert_weights(self) -> Iterable[torch.Tensor]:
         weights = list(self.named_parameters())
-        assert all(weight.is_contiguous() for _, weight in weights)
+        assert all(
+            weight.is_contiguous()
+            for name, weight in weights
+            if not name.startswith("_shared_experts.")
+        )
```
hmellor, line 1225: I'm not sure about this change. @abmfy could this cause issues for other EPLB use cases?

PerryZhang01: the shared expert is currently not continuous, possibly because the later MR performed a stride operation, however, we think that EPLB only apply to routed experts(this function only returns routed experts),  so we have canceled the contiguous check on shared expert, and shared expert may have other stride operations in the future and should not be asserted here.

PerryZhang01: @hmellor @abmfy can someone confirm the change here?

PerryZhang01: image  

I have rebased the latest code and validated the accuracy of EPLB on gsm8k datasets again.

PerryZhang01:
@hmellor @abmfy @BowenBao could u help review this PR?

vllm/model_executor/layers/fused_moe/layer.py
abmfy: @hmellor should be good here since EPLB is only on shared experts

abmfy (approved):
LGTM.
Thanks for the contribution!

hmellor (approved):

PR #27737, opened 2025-10-29
Prevent overwriting drafters lm-head and embed_tokens

eldarkurtic:
Some EAGLE3 drafters might have their own lm_head and/or embed_tokens layers. Existing codebase ignores this and always overrides them with target model's layers.

#### Test-case 1: for a model which needs copying of lm_head and embed_tokens from verifier model, the behavior should not change

Eval command to verify:

```bash
CUDA_VISIBLE_DEVICES=0,1 python examples/offline_inference/spec_decode.py \
  --method "eagle3" \
  --tp 2 \
  --model-dir "openai/gpt-oss-120b" \
  --eagle-dir "nvidia/gpt-oss-120b-Eagle3" \
  --dataset_name "hf" \
  --num-spec-tokens 3
```

Before this PR:
```bash
total_num_output_tokens: 241888
num_drafts: 105426
num_draft_tokens: 316278
num_accepted_tokens: 136330
mean acceptance length: 2.29
acceptance at token 0: 0.65
acceptance at token 1: 0.40
acceptance at token 2: 0.24
```

After this PR:
```
total_num_output_tokens: 241168
num_drafts: 105678
num_draft_tokens: 317034
num_accepted_tokens: 135362
mean acceptance length: 2.28
acceptance at token 0: 0.65
acceptance at token 1: 0.39
acceptance at token 2: 0.24
```

#### Test-case 2: for a model with has its own lm_head and embed_tokens, and therefore does not require copying from target's layers, acceptance rates look significantly better

Before this PR:
```bash
total_num_output_tokens: 247973
num_drafts: 187566
num_draft_tokens: 562698
num_accepted_tokens: 59726
mean acceptance length: 1.32
acceptance at token 0: 0.23
acceptance at token 1: 0.07
acceptance at token 2: 0.02
```

After this PR:
```bash
total_num_output_tokens: 247974
num_drafts: 99354
num_draft_tokens: 298062
num_accepted_tokens: 148677
mean acceptance length: 2.50
acceptance at token 0: 0.70
acceptance at token 1: 0.48
acceptance at token 2: 0.31
```

Note: idea for lm_head check inspired by https://github.com/vllm-project/vllm/pull/27688

### DIFF
--- a/vllm/model_executor/models/interfaces.py
+++ b/vllm/model_executor/models/interfaces.py
@@ -922,6 +922,16 @@ class SupportsEagle3(Protocol):
         MRO of your model class.
     """
 
+    has_own_lm_head: ClassVar[bool] = False
+    """
+    A flag that indicates this model has trained its own lm_head.
+    """
+
+    has_own_embed_tokens: ClassVar[bool] = False
+    """
+    A flag that indicates this model has trained its own input embeddings.
+    """
+
     def set_aux_hidden_state_layers(self, layers: tuple[int, ...]) -> None:
         """
         Set which layers should output auxiliary
--- a/vllm/model_executor/models/interfaces.py
+++ b/vllm/model_executor/models/interfaces.py
@@ -908,13 +908,43 @@ def supports_transcription(
 
 
 @runtime_checkable
-class SupportsEagle3(Protocol):
-    """The interface required for models that support
-    EAGLE3 speculative decoding."""
+class SupportsEagleBase(Protocol):
+    """Base interface for models that support EAGLE-based speculative decoding."""
 
-    supports_eagle3: ClassVar[Literal[True]] = True
+    has_own_lm_head: bool = False
     """
-    A flag that indicates this model supports EAGLE3 
+    A flag that indicates this model has trained its own lm_head.
+    """
+
+    has_own_embed_tokens: bool = False
+    """
+    A flag that indicates this model has trained its own input embeddings.
+    """
+
+
+@overload
+def supports_any_eagle(model: type[object]) -> TypeIs[type[SupportsEagleBase]]: ...
+
+
+@overload
+def supports_any_eagle(model: object) -> TypeIs[SupportsEagleBase]: ...
+
+
+def supports_any_eagle(
+    model: type[object] | object,
+) -> TypeIs[type[SupportsEagleBase]] | TypeIs[SupportsEagleBase]:
+    """Check if model supports any EAGLE variant (1, 2, or 3)."""
+    return supports_eagle(model) or supports_eagle3(model)
+
+
+@runtime_checkable
+class SupportsEagle(SupportsEagleBase, Protocol):
+    """The interface required for models that support
+    EAGLE-1 and EAGLE-2 speculative decoding."""
+
+    supports_eagle: ClassVar[Literal[True]] = True
+    """
+    A flag that indicates this model supports EAGLE-1 and EAGLE-2 
     speculative decoding.
 
     Note:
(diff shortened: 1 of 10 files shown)


vllm/model_executor/models/llama_eagle3.py
```diff
@@ -328,6 +331,12 @@ def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]):
                 includes_embed_tokens = True
             model_weights[name] = loaded_weight
 
+            # To prevent overriding with target model's layers
+            if "lm_head" in name:
+                self.has_own_lm_head = True
+            if "embed_tokens" in name:
+                self.model.has_own_embed_tokens = True
```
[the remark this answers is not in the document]
eldarkurtic: we are using the same logic in this function a couple of lines above, so I think it should be fine

rahul-tuli (approved):
LGTM!

vllm/model_executor/models/llama_eagle3.py
```diff
@@ -148,6 +148,7 @@ def __init__(
             self.config.hidden_size,
             prefix=maybe_prefix(prefix, "embed_tokens"),
         )
+        self.has_own_embed_tokens = False
```
NickLucche: any chance we can leverage an existing interface (`SupportsEagle`?) to add this flag?

eldarkurtic: great idea, thanks a lot for the suggestion!

robertgshaw2-redhat (approved):

NickLucche (approved):
Thanks for the fix @eldarkurtic ! Let's get CI green real quick before merging this one

vllm/model_executor/models/llama_eagle3.py
```diff
@@ -328,6 +328,12 @@ def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]):
                 includes_embed_tokens = True
             model_weights[name] = loaded_weight
 
+            # To prevent overriding with target model's layers
+            if "lm_head" in name:
+                self.has_own_lm_head = True
+            if "embed_tokens" in name:
```
benchislett: Is this a change from the default? I thought that EAGLE3 heads usually share the embedding with the base model?

eldarkurtic: Eagle3 by default doesn’t train these layers. But there is no reason not to train them. This doesn’t affect the standard eagle3 flow, just extends it to support this new use case

benchislett:
There are other EAGLE1-based models which do not have the EAGLE3 mixin, causing the CI failures. Please update the logic to cover EAGLE1 as well

vllm/v1/spec_decode/eagle.py
```diff
@@ -997,20 +997,17 @@ def load_model(self, target_model: nn.Module) -> None:
...
 
-            # Check if shapes match and we found the embedding
-            eagle_shape = self.model.model.embed_tokens.weight.shape
-            target_shape = target_embed_tokens.weight.shape
-            if eagle_shape == target_shape:
+            if not self.model.has_own_embed_tokens:
                 logger.info(
-                    "Assuming the EAGLE head shares the same vocab embedding"
-                    " with the target model."
+                    "Draft model embed_tokens are uninitialized. "
```
benchislett: This was originally done as a memory optimization since the original EAGLE3 models are released with an embedding layer included, but having the same weights as the base model. 

Ideally we would have another check here to delete them if they are present but identical to those of the base model, but that's tricky to implement cleanly. I'm fine with doing it this way for now, but it should be noted that the behaviour is changing

eldarkurtic:
Do they use some other mixin similar to SupportsEagle3?

vllm/model_executor/models/interfaces.py
```diff
@@ -922,6 +922,16 @@ class SupportsEagle3(Protocol):
         MRO of your model class.
     """
 
+    has_own_lm_head: ClassVar[bool] = False
```
benchislett: These are not class variables, they are instance variables that are set per-model based on the weight loading.

https://typing.python.org/en/latest/spec/class-compat.html

benchislett:
The proper fix here is to add a base SupportsEagle mixin and have SupportsEagle3 inherit from that. Then the other EAGLE classes can inherit from the new base mixin that will give them a reasonable default.

vllm/model_executor/models/llama_eagle3.py
```diff
@@ -328,6 +328,12 @@ def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]):
                 includes_embed_tokens = True
             model_weights[name] = loaded_weight
 
+            # To prevent overriding with target model's layers
+            if "lm_head" in name:
```
benchislett: This should be added to all the EAGLE classes, not just llama_eagle3. 

You might want to refactor this into the mixin to reuse the code between them.

Issue #27742, opened 2025-10-29
[RFC]: KV cache layout combining all layers per block


orozery:
### Motivation.

High-level goal: optimize the transfer time of KV data from/to the vllm KV cache by reducing per-transfer fragmentation at the vllm KV cache side.

The 2 main use-cases we're considering are:
1. NixlConnector
2. OffloadingConnector

So our goal is to try and optimize the time each of these connectors spend for a single transfer (GPU->CPU of the offloading connector, and GPU->host_buffer/other agent for NixlConnector).
The transfer unit is a block of KV-data corresponding to some aligned (default)16-tokens out of some request.
We consider the case of simple models where all layers have the same layout (i.e. not using the Hybrid Memory Allocator), although this can be explored in the future.
For this case, the current layout for the vllm KV cache is (roughly) this:

<img width="1588" height="502" alt="Image" src="https://github.com/user-attachments/assets/bf5189f1-6948-4c29-a3f9-211bbd391a77" />

As you can see, the KV data that needs to be transferred is breaked down to num_layers non-contiguous segments.
This does not take into account the k_or_v and num_kv_heads dimensions which may further cause extra fragmentation.

In this RFC, we propose a new KV cache layout that for each block, will place all KV data for all layers contiguously:

<img width="1481" height="586" alt="Image" src="https://github.com/user-attachments/assets/86b9e0ad-a77e-446a-a831-51e655c9e81c" />

This will reduce the number of fragments to copy by a factor of num_layers.

We have tested this with both the NixlConnector and the OffloadingConnector and saw that the transfer time is improved by a **factor of X4** (tested on Llama-3.1-8B-Instruct).
Furthermore, we tested plain model execution and saw no noticeable degradation in performance. 

### Proposed Change.

We propose to change the KV tensor allocation in `gpu_model_runner.py` to support this new layout, allocating a single tensor for all layers (instead of 1 per layer as it is today).
This new layout will only be selected if:
1. The model has a simple uniform layout (not HMA)
2. A KV connector is configured and, the KV connector class explicitly prefers (via a new classmethod) to use the new layout.

### CC List.

@njhill @NickLucche @robertgshaw2-redhat @heheda12345 @LucasWilkinson @tdoublep @tlrmchlsmth

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

[...]

PR #23299, opened 2025-08-21
Support FlashAttention Backend for Hybrid SSM Models

heheda12345:
## Purpose
As mentioned in https://github.com/vllm-project/vllm/pull/20016, v1 hybrid ssm requires the layout of attention to be (num_blocks, 2, hidden_size) so that the blocks can be shared between attention layers and mamba layers. This PR supports (2, num_blocks, hidden_size) layout by changing the tensor stride of (2, num_blocks) dimensions to (hidden_size, 2*hidden_size)
## Test Plan
Run unit tests. Should pass
## Test Result
Let's wait for CI.
## (Optional) Documentation Update

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x] The test plan, such as providing test command.
- [x] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
</details>

PR #23265, opened 2025-08-20
[Perf] Small optimizations for silu_mul_fp8_quant_deep_gemm

mgoin:
## Purpose

Reorder some stride calculation and operations, rename to more descriptive variables, and decrease num_stages to 4 to help with expert token imbalance. Not a big deal but should help a bit.

```
python benchmarks/kernels/benchmark_silu_mul_fp8_quant.py

# main
GPU: NVIDIA H100 80GB HBM3
Config               Time(ms)   GFLOPS     GB/s      
--------------------------------------------------
E=  8,T=  32,H=1024    0.027     54.3     34.1
E= 16,T=  64,H=2048    0.027    457.2    287.5
E= 32,T= 128,H=4096    0.049   2188.1   1376.1
E=256,T=  16,H=7168    0.048   3534.2   2222.7
E=256,T=  32,H=7168    0.087   3978.8   2502.3
E=256,T=  64,H=7168    0.171   4064.7   2556.3
E=256,T= 128,H=7168    0.340   4157.8   2614.9
E=256,T= 256,H=7168    0.677   4244.9   2669.6
E=256,T= 512,H=7168    1.320   4291.0   2698.7
E=256,T=1024,H=7168    2.733   4185.6   2632.3

# this pr
GPU: NVIDIA H100 80GB HBM3
Config               Time(ms)   GFLOPS     GB/s      
--------------------------------------------------
E=  8,T=  32,H=1024    0.027     54.7     34.4
E= 16,T=  64,H=2048    0.028    455.7    286.6
E= 32,T= 128,H=4096    0.049   2179.3   1370.6
E=256,T=  16,H=7168    0.045   3788.7   2382.7
E=256,T=  32,H=7168    0.084   4130.8   2597.9
E=256,T=  64,H=7168    0.168   4149.8   2609.8
E=256,T= 128,H=7168    0.335   4229.8   2660.2
E=256,T= 256,H=7168    0.671   4282.4   2693.2
E=256,T= 512,H=7168    1.310   4325.2   2720.1
E=256,T=1024,H=7168    2.694   4245.8   2670.2
```

Eval shows the same score

```
# main
[...]

PR #27743, opened 2025-10-29
[KVConnector][Core] Support cross-layer KV blocks

orozery:
Core of RFC #27742.
Following this PR, connectors can turn-on and adapt to the new layout.

This PR enables the GPU model runner to allocate the KV cache tensors, so that the KV data for all layers will be contiguous per block. This can yield a significant speed up the transfer time of KV transfers (e.g. X4), such in the case of using NixlConnector or OffloadingConnector. Currently, this new layout is disabled by default, and will only be enabled when using a connector which explicitly prefers this new layout. Also, this new layout is currently only supported for uniform (non HMA) models.

### DIFF
--- a/vllm/attention/backends/abstract.py
+++ b/vllm/attention/backends/abstract.py
@@ -76,7 +76,34 @@ class AttentionBackend(ABC):
         raise NotImplementedError
 
     @staticmethod
-    def get_kv_cache_stride_order() -> tuple[int, ...]:
+    def get_kv_cache_stride_order(
+        include_num_layers_dimension: bool = False,
+    ) -> tuple[int, ...]:
+        """
+        Get the physical (memory layout) ordering of the kv cache dimensions.
+        e.g. if the KV cache shape is
+        [2, num_blocks, block_size, num_heads, head_size],
+        and get_kv_cache_stride_order returns (1, 3, 0, 2, 4) then the physical
+        ordering of dimensions is
+        [num_blocks, num_heads, 2, block_size, head_size].
+
+        If this function is unimplemented / raises NotImplementedError,
+        the physical layout of the KV cache will match the logical shape.
+
+        Args:
+            include_num_layers_dimension: if True, includes an additional
+                num_layers dimension, which is assumed to be prepended
+                to the logical KV cache shape.
+                With the above example, a return value (2, 4, 0, 1, 3, 5)
+                corresponds to
+                [num_blocks, num_heads, num_layers, 2, block_size, head_size].
+
+                If an additional dimension is NOT included in the returned
+                tuple, the physical layout will not include a layers dimension.
+
+        Returns:
+            A tuple of ints which is a permutation of range(len(shape)).
+        """
         raise NotImplementedError
 
     @classmethod
--- a/vllm/distributed/kv_transfer/kv_connector/v1/base.py
+++ b/vllm/distributed/kv_transfer/kv_connector/v1/base.py
@@ -38,7 +38,7 @@ The class provides the following primitives:
 import enum
 from abc import ABC, abstractmethod
 from collections.abc import Callable, Iterable
-from typing import TYPE_CHECKING, Any, Literal, Optional
+from typing import TYPE_CHECKING, Any, ClassVar, Literal, Optional
 
 import torch
 
@@ -47,7 +47,7 @@ from vllm.v1.core.sched.output import SchedulerOutput
 from vllm.v1.outputs import KVConnectorOutput
 
 if TYPE_CHECKING:
-    from vllm.attention.backends.abstract import AttentionMetadata
+    from vllm.attention.backends.abstract import AttentionBackend, AttentionMetadata
     from vllm.config import VllmConfig
     from vllm.distributed.kv_events import KVCacheEvent
     from vllm.distributed.kv_transfer.kv_connector.v1.metrics import (
@@ -142,6 +142,18 @@ class KVConnectorMetadata(ABC):  # noqa: B024
 
 
 class KVConnectorBase_V1(ABC):
+    """
+    Base class for KV connectors.
+
+    Attributes:
+        prefer_cross_layer_blocks (bool): Indicates whether this connector
+            prefers KV blocks that hold KV data for all layers (for speeding
+            up KV data transfers).
+            Defaults to False.
+    """
+
+    prefer_cross_layer_blocks: ClassVar[bool] = False
+
     def __init__(
         self,
         vllm_config: "VllmConfig",
@@ -223,6 +235,23 @@ class KVConnectorBase_V1(ABC):
 
         Args:
             kv_caches: dictionary of layer names, kv cache
+        """
+        return
+
+    def register_cross_layers_kv_cache(
+        self, kv_cache: torch.Tensor, attn_backend: type["AttentionBackend"]
+    ):
+        """
+        Initialize with a single KV cache tensor used by all layers.
+        The first dimension should be num_layers.
+        This function will only be called for models with uniform layers,
+        and only if the prefers_cross_layer_blocks is set to True.
+        Only one of the functions
+        {register_kv_caches, register_cross_layers_kv_cache} will be called.
+
+        Args:
+            kv_cache: a cross-layers kv cache tensor
+            attn_backend: The attention backend that corresponds to all layers
         """
         return
 
--- a/vllm/v1/attention/backends/flash_attn.py
+++ b/vllm/v1/attention/backends/flash_attn.py
@@ -99,12 +99,20 @@ class FlashAttentionBackend(AttentionBackend):
         return (2, num_blocks, block_size, num_kv_heads, head_size)
 
     @staticmethod
-    def get_kv_cache_stride_order() -> tuple[int, ...]:
+    def get_kv_cache_stride_order(
+        include_num_layers_dimension: bool = False,
+    ) -> tuple[int, ...]:
         # `stride_order` indicates the permutation that gets
         # us from `get_kv_cache_shape` to the actual memory layout we want.
         cache_layout = get_kv_cache_layout()
-        if cache_layout == "NHD":
+        if cache_layout == "NHD" and include_num_layers_dimension:
+            # (num_blocks, num_layers, 2, block_size, num_kv_heads, head_size)
+            return (2, 0, 1, 3, 4, 5)
+        elif cache_layout == "NHD":
             stride_order = (0, 1, 2, 3, 4)
+        elif cache_layout == "HND" and include_num_layers_dimension:
+            # (num_blocks, num_kv_heads, num_layers, 2, block_size, head_size)
+            return (2, 4, 0, 1, 3, 5)
         elif cache_layout == "HND":
             stride_order = (0, 1, 3, 2, 4)
         else:
--- a/vllm/v1/attention/backends/flashinfer.py
+++ b/vllm/v1/attention/backends/flashinfer.py
@@ -309,12 +309,20 @@ class FlashInferBackend(AttentionBackend):
         return (num_blocks, 2, block_size, num_kv_heads, head_size)
 
     @staticmethod
-    def get_kv_cache_stride_order() -> tuple[int, ...]:
+    def get_kv_cache_stride_order(
+        include_num_layers_dimension: bool = False,
+    ) -> tuple[int, ...]:
         # `stride_order` indicates the permutation that gets us from
         # `get_kv_cache_shape` to the actual memory layout we want.
         cache_layout = get_kv_cache_layout()
-        if cache_layout == "NHD":
+        if cache_layout == "NHD" and include_num_layers_dimension:
+            # (num_blocks, num_layers, 2, block_size, num_kv_heads, head_size)
+            return (1, 0, 2, 3, 4, 5)
+        elif cache_layout == "NHD":
             stride_order = (0, 1, 2, 3, 4)
+        elif cache_layout == "HND" and include_num_layers_dimension:
+            # (num_blocks, 2, num_kv_heads, num_layers, block_size, head_size)
+            return (1, 2, 4, 0, 3, 5)
         elif cache_layout == "HND":
             stride_order = (0, 1, 3, 2, 4)
         else:
--- a/vllm/v1/attention/backends/mla/indexer.py
+++ b/vllm/v1/attention/backends/mla/indexer.py
@@ -45,7 +45,11 @@ class DeepseekV32IndexerBackend(AttentionBackend):
         return (num_blocks, block_size, head_size)
 
     @staticmethod
-    def get_kv_cache_stride_order() -> tuple[int, ...]:
+    def get_kv_cache_stride_order(
+        include_num_layers_dimension: bool = False,
+    ) -> tuple[int, ...]:
+        if include_num_layers_dimension:
+            return (0, 1, 2, 3)
         return (0, 1, 2)
 
 
--- a/vllm/v1/worker/gpu_model_runner.py
+++ b/vllm/v1/worker/gpu_model_runner.py
@@ -350,6 +350,9 @@ class GPUModelRunner(
         # self.model: nn.Module  # Set after load_model
         # Initialize in initialize_kv_cache
         self.kv_caches: list[torch.Tensor] = []
+        # Initialize in initialize_kv_cache_tensors
+        self.cross_layers_kv_cache: torch.Tensor | None = None
+        self.cross_layers_attn_backend: type[AttentionBackend] | None = None
         # indexes: [kv_cache_group_id][attn_group]
         self.attn_groups: list[list[AttentionGroup]] = []
         # self.kv_cache_config: KVCacheConfig
@@ -4944,12 +4947,30 @@ class GPUModelRunner(
             Dict[str, torch.Tensor]: A map between layer names to their
             corresponding memory buffer for KV cache.
         """
-        # Initialize the memory buffer for KV cache
-        kv_cache_raw_tensors = self._allocate_kv_cache_tensors(kv_cache_config)
-        # Change the memory buffer to the desired shape
-        kv_caches = self._reshape_kv_cache_tensors(
-            kv_cache_config, kv_cache_raw_tensors, kernel_block_sizes
-        )
+
+        # Try creating KV caches optimized for kv-connector transfers
+        cache_dtype = self.cache_config.cache_dtype
+        if self.use_uniform_kv_cache(self.attn_groups, cache_dtype):
+            kv_caches, cross_layers_kv_cache, attn_backend = (
+                self.allocate_uniform_kv_caches(
+                    kv_cache_config,
+                    self.attn_groups,
+                    cache_dtype,
+                    self.device,
+                    kernel_block_sizes,
+                )
+            )
+            self.cross_layers_kv_cache = cross_layers_kv_cache
+            self.cross_layers_attn_backend = attn_backend
+        else:
+            # Fallback to the general case
+            # Initialize the memory buffer for KV cache
+            kv_cache_raw_tensors = self._allocate_kv_cache_tensors(kv_cache_config)
+
+            # Change the memory buffer to the desired shape
+            kv_caches = self._reshape_kv_cache_tensors(
+                kv_cache_config, kv_cache_raw_tensors, kernel_block_sizes
+            )
 
         # Set up cross-layer KV cache sharing
         for layer_name, target_layer_name in self.shared_kv_cache_layers.items():
@@ -5031,7 +5052,13 @@ class GPUModelRunner(
 
         if has_kv_transfer_group():
             kv_transfer_group = get_kv_transfer_group()
-            kv_transfer_group.register_kv_caches(kv_caches)
+            if self.cross_layers_kv_cache is not None:
+                assert self.cross_layers_attn_backend is not None
+                kv_transfer_group.register_cross_layers_kv_cache(
+                    self.cross_layers_kv_cache, self.cross_layers_attn_backend
+                )
+            else:
+                kv_transfer_group.register_kv_caches(kv_caches)
             kv_transfer_group.set_host_xfer_buffer_ops(copy_kv_blocks)
 
         if self.dcp_world_size > 1:
--- a/tests/v1/kv_connector/unit/test_offloading_connector.py
+++ b/tests/v1/kv_connector/unit/test_offloading_connector.py
@@ -19,6 +19,7 @@ from vllm.distributed.kv_transfer.kv_connector.v1.offloading_connector import (
 )
 from vllm.forward_context import ForwardContext
 from vllm.utils.hashing import sha256
+from vllm.v1.attention.backends.flash_attn import FlashAttentionBackend
 from vllm.v1.core.kv_cache_utils import (
     BlockHash,
     get_request_block_hasher,
@@ -92,7 +93,7 @@ class MockOffloadingSpec(OffloadingSpec):
         return self.manager
 
     def get_handlers(
-        self, _
+        self, _, __
     ) -> Iterator[tuple[type[LoadStoreSpec], type[LoadStoreSpec], OffloadingHandler]]:
         yield GPULoadStoreSpec, MockLoadStoreSpec, self.handler
         yield MockLoadStoreSpec, GPULoadStoreSpec, self.handler
@@ -138,7 +139,10 @@ class RequestRunner:
         self.worker_connector = OffloadingConnector(vllm_config, KVConnectorRole.WORKER)
 
         # register worker kv_caches to enable OffloadingWorker creations
-        self.worker_connector.register_kv_caches(kv_caches={"a": torch.empty(0)})
+        self.worker_connector.register_cross_layers_kv_cache(
+            kv_cache=torch.empty(0),
+            attn_backend=FlashAttentionBackend,
+        )
 
         # extract connector of scheduler
         scheduler_connector = self.scheduler.connector
--- a/tests/v1/kv_offload/test_cpu_offloading.py
+++ b/tests/v1/kv_offload/test_cpu_offloading.py
@@ -12,8 +12,10 @@ from tqdm import tqdm
 from vllm import LLM, SamplingParams, TokensPrompt
 from vllm.config import KVEventsConfig, KVTransferConfig
 from vllm.distributed.kv_events import BlockStored, KVEventBatch
+from vllm.utils.system_utils import set_env_var
 
-CPU_BLOCK_SIZES = [16, 48]
+CPU_BLOCK_SIZES = [48]
+ATTN_BACKENDS = ["FLASH_ATTN", "FLASHINFER"]
 
 
 class MockSubscriber:
@@ -63,8 +65,88 @@ class MockSubscriber:
         self.sub.close()
 
 
+def _latency_test(llm: LLM, subscriber: MockSubscriber):
+    sampling_params = SamplingParams(max_tokens=1)
+
+    num_times_cpu_better_than_cold = 0
+    num_tests = 10
+    total_cold_time = 0.0
+    total_gpu_hit_time = 0.0
+    total_cpu_hit_time = 0.0
+    prompt_token_ids = [0] * 10001
+    for i in tqdm(range(num_tests), desc="Running tests"):
+        prompt_token_ids[0] = i
+        prompts = [TokensPrompt(prompt_token_ids=prompt_token_ids)]
+
+        # run generation - this should trigger saving KV cache
+        start_time = time.time()
+        llm.generate(prompts, sampling_params, use_tqdm=False)
+        cold_time = time.time() - start_time
+        total_cold_time += cold_time
+
+        # run generation again - should hit the GPU prefix cache
+        start_time = time.time()
+        llm.generate(prompts, sampling_params, use_tqdm=False)
+        gpu_hit_time = time.time() - start_time
+        total_gpu_hit_time += gpu_hit_time
+
+        # reset prefix cache to avoid GPU hit.
+        llm.reset_prefix_cache()
+
+        assert subscriber.get_new_cpu_stored_events()
+
+        # run generation again - this should trigger loading from CPU
+        start_time = time.time()
+        llm.generate(prompts, sampling_params, use_tqdm=False)
+        cpu_hit_time = time.time() - start_time
+        total_cpu_hit_time += cpu_hit_time
+
+        if cpu_hit_time < cold_time:
+            num_times_cpu_better_than_cold += 1
+
+    print("Average times:")
+    print(f"    Cold: {total_cold_time * 1000 / num_tests:.2f}ms")
+    print(f"    GPU hit: {total_gpu_hit_time * 1000 / num_tests:.2f}ms")
+    print(f"    CPU hit: {total_cpu_hit_time * 1000 / num_tests:.2f}ms")
+
+    assert num_times_cpu_better_than_cold >= 0.8 * num_tests
+
+
+def _accuracy_test(llm: LLM, subscriber: MockSubscriber):
+    sampling_params = SamplingParams(max_tokens=1)
+    cpu_block_size = (
+        llm.llm_engine.vllm_config.kv_transfer_config.kv_connector_extra_config[
+            "block_size"
+        ]
+    )
+
+    subscriber.get_new_cpu_stored_events()
+
+    # prepend prompt to be cpu block aligned
+    prompt = "Let's count to 10. One, two, three, four,"
+    while (
+        len(llm.generate(prompt, use_tqdm=False)[0].prompt_token_ids) % cpu_block_size
+        != 0
+    ):
+        prompt = ". " + prompt
+
+    assert subscriber.get_new_cpu_stored_events()
+
+    test_count = 100
+    success_count = 0
+    for i in range(test_count):
+        if (
+            llm.generate(prompt, sampling_params, use_tqdm=False)[0].outputs[0].text
+            == " five"
+        ):
+            success_count += 1
+
+    assert success_count >= 0.5 * test_count
+
+
 @pytest.mark.parametrize("cpu_block_size", CPU_BLOCK_SIZES)
-def test_cpu_offloading(cpu_block_size: int) -> None:
+@pytest.mark.parametrize("attn_backend", ATTN_BACKENDS)
+def test_cpu_offloading(cpu_block_size: int, attn_backend: str) -> None:
     """
     Tests OffloadingConnector with CPUOffloadingSpec.
     """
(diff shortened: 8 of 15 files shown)


vllm/v1/worker/gpu_model_runner.py
```diff
@@ -4566,6 +4655,7 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
         """
         kv_cache_config = deepcopy(kv_cache_config)
         self.kv_cache_config = kv_cache_config
+        self.may_reinitialize_input_batch(kv_cache_config)
```
njhill: this is already called a few lines below?

orozery: Yep, rebase bug

ApostaC:
Quick question:
- How should the connector know whether the layout is the new one or the old one?
- Should the connector have 2 different code paths to support both layouts? This will potentially increase the connector code by a lot.

orozery:
> * How should the connector know whether the layout is the new one or the old one?

My initial thought is to check on `register_kv_caches` whether two layer tensors have the same `.storage().data_ptr()`.
Of course we can also expose a new API.

> * Should the connector have 2 different code paths to support both layouts? This will potentially increase the connector code by a lot.

Connectors don't need to support the new layout if they don't want to.
For the offloading connector and nixl it does not seem like a big change.
For the offloading connector for example this is just writing a new cuda transfer function.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -4280,6 +4283,86 @@ def may_reinitialize_input_batch(self, kv_cache_config: KVCacheConfig) -> None:
...
+        ]
+
+        logger.info("Allocating a single KV cache buffer for all layers")
+
+        # allocate one contiguous buffer for all layers
+        buffer = (
+            torch.zeros(total_size, dtype=torch.int8, device=self.device)
+            .view(kv_cache_spec.dtype)
+            .view(kv_cache_shape)
+        ).permute(*inv_order)
```
ApostaC: By calling view + permute, we basically create a non-contiguous view of the raw KV cache buffer. This may have an unexpected impact (both performance and correctness) on other modules.

For example, calling `reshape` on a non-contiguous tensor may introduce an extra memory copy. Therefore, this code adds **implicit pitfalls** for all the other logics (either first-party or third-party) that may need to do `reshape`.

Quick code snippet:
```python
import torch

x = torch.randn((2,3,5))
y = x.permute(2,0,1)
z = y.reshape(30)

print("X -- is contiguous:", x.is_contiguous(), "\tdata_ptr:", x.data_ptr())
print("Y -- is contiguous:", y.is_contiguous(), "\tdata_ptr:", y.data_ptr())
print("Z -- is contiguous:", z.is_contiguous(), "\tdata_ptr:", z.data_ptr())
```

And the output:
```
X -- is contiguous: True        data_ptr: 1076219648
Y -- is contiguous: False       data_ptr: 1076219648
Z -- is contiguous: True        data_ptr: 986565184   ############ Y is being implicitly copied to Z.
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -4280,6 +4283,86 @@ def may_reinitialize_input_batch(self, kv_cache_config: KVCacheConfig) -> None:
...
+
+        kv_caches = {}
+        for i, kv_cache_tensor in enumerate(kv_cache_config.kv_cache_tensors):
+            tensor = buffer[i]
+            for layer_name in kv_cache_tensor.shared_by:
+                kv_caches[layer_name] = tensor
+
+        return kv_caches
+
```
ApostaC: Although I like the high-level idea, I strongly feel like we should directly expose the raw KV cache tensor to the connector rather than the permuted KV caches dictionary.

Otherwise, it's very hard to let the connector know what the layout is before the permutation, and this may introduce a huge debug pain, especially when we need to write some C/CUDA code that directly uses the raw pointers.

ApostaC (changes requested):
Like the high-level idea, but I think we need to further discuss the design.

Also, the correctness of this PR is not tested. Will the new code have any correctness & performance impact on the underlying attention modules?

vllm/v1/worker/gpu_model_runner.py
orozery: Correct.
This is already done (making non-contiguous view) when using HND layout (note that a permute flow already exists today in `_reshape_kv_cache_tensors`.

Note that this layout is currently applied when getting the agreement of 3 parties:
1. The KV cache spec (no HMA)
2. The KV connector (`prefer_cross_layer_blocks`)
3. The attention backend (`get_kv_cache_stride_order`)

vllm/v1/worker/gpu_model_runner.py
orozery: The current API exposes dict[str, pytorch.Tensor] to the connector (via `register_kv_caches`).
The tensors themselves could be permuted.
The physical layout of the tensors can be self-determined using pytorch APIs.
But I agree this is somehow "hidden".
If we want to be more explicit we can add an auxiliary variable to `register_kv_caches` that will hold the physical layout spec for the tensors. What do you think?

ApostaC:
> Is there a way to make a CI test out of it? Or some other way?

I think lm_eval can be used for correctness benchmarks.

vllm/v1/worker/gpu_model_runner.py
ApostaC: I like the idea of adding the auxiliary variable to `register kv caches` function!

vllm/v1/attention/backends/flash_attn.py
```diff
@@ -105,12 +105,18 @@ def get_kv_cache_shape(
...
 
     @staticmethod
-    def get_kv_cache_stride_order() -> tuple[int, ...]:
+    def get_kv_cache_stride_order(with_layers_dim: bool) -> tuple[int, ...]:
         # `stride_order` indicates the permutation that gets
         # us from `get_kv_cache_shape` to the actual memory layout we want.
         cache_layout = get_kv_cache_layout()
-        if cache_layout == "NHD":
+        if cache_layout == "NHD" and with_layers_dim:
+            # (2, num_blocks, num_layers, block_size, num_kv_heads, head_size)
```
heheda12345: Prefer (num_layer, num_blocks, 2, block_size, ...) 

1. I guess it will be a little bit faster.
2. In hybrid mamba, a block can be either a block for attention or a block for mamba. But with (2, num_blocks, num_layers, block_size, num_kv_heads, head_size), the memory buffer for one block of one layer is not contiguous and we can't use it to save the mamba state which is only one tensor instead of k and v

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -4564,12 +4649,18 @@ def initialize_kv_cache_tensors(
...
         """
-        # Initialize the memory buffer for KV cache
-        kv_cache_raw_tensors = self._allocate_kv_cache_tensors(kv_cache_config)
-        # Change the memory buffer to the desired shape
-        kv_caches = self._reshape_kv_cache_tensors(
-            kv_cache_config, kv_cache_raw_tensors, kernel_block_sizes
-        )
+
+        # Try creating KV caches optimized for kv-connector transfers
+        kv_caches = self._allocate_uniform_kv_caches(kv_cache_config)
```
heheda12345: prefer something like
```
if use_uniform_kv_cache():
     kv_caches = self._allocate_uniform_kv_caches(kv_cache_config)
else:
     kv_cache_raw_tensors =    self._allocate_kv_cache_tensors(kv_cache_config)
     # Change the memory buffer to the desired shape
     kv_caches = self._reshape_kv_cache_tensors(
            kv_cache_config, kv_cache_raw_tensors, kernel_block_sizes
        )
```
as the current implementation should not be a "fallback"

vllm/v1/attention/backends/flash_attn.py
orozery: You mean (num_blocks, num_layers, 2, block_size, ...) ?

NickLucche: this PR doesn't work for hybrid models, it assumes all layers have the same layout

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -4333,6 +4336,86 @@ def may_reinitialize_input_batch(
...
+            return {}
+
+        kv_cache_shape = tuple(kv_cache_shape[i] for i in kv_cache_stride_order)
+
+        # Maintain original KV shape view.
+        inv_order = [
+            kv_cache_stride_order.index(i) for i in range(len(kv_cache_stride_order))
+        ]
+
+        logger.info("Allocating a single KV cache buffer for all layers")
```
NickLucche: nit: this log line could be a bit more verbose to explain the user what's going on

vllm/v1/worker/gpu_model_runner.py
NickLucche: +1

vllm/attention/backends/abstract.py
```diff
@@ -81,7 +81,7 @@ def get_kv_cache_shape(
         raise NotImplementedError
 
     @staticmethod
-    def get_kv_cache_stride_order() -> tuple[int, ...]:
+    def get_kv_cache_stride_order(with_layers_dim: bool) -> tuple[int, ...]:
```
NickLucche: not sure we should break interface here, this could be a registered OOT I think

NickLucche: eg probably we should give it a default of False.

Also this parameter definitely needs a docstring + perhaps a more apt name

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -4333,6 +4336,86 @@ def may_reinitialize_input_batch(
...
+        Args:
+            kv_cache_config: The KV cache config
+        Returns:
+            dict[str, torch.Tensor]: A map between layer names to their
+            corresponding memory buffer for KV cache.
+            An empty dictionary is returned if the given KV cache config
+            is incompatible with this simple type of layout.
+        """
+        if not kv_connector_prefers_cross_layer_blocks():
+            return {}
```
NickLucche: this is connector dependent right now to the point that it could be moved into the mixin

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -4333,6 +4336,86 @@ def may_reinitialize_input_batch(
...
+        tensor_sizes = set(
+            kv_cache_tensor.size for kv_cache_tensor in kv_cache_config.kv_cache_tensors
+        )
+        assert len(tensor_sizes) == 1
+        tensor_size = tensor_sizes.pop()
+
+        page_size = kv_cache_spec.page_size_bytes
+        assert tensor_size % page_size == 0
+        num_blocks = tensor_size // page_size
+        num_tensors = len(kv_cache_config.kv_cache_tensors)
```
NickLucche:
```suggestion
        num_layers = len(kv_cache_config.kv_cache_tensors)
```

NickLucche (commented):
Thank for the PR @orozery !

Not a fan of changing the `get_kv_cache_stride_order` interface to be honest, I was hoping we could use your logic to *wrap* around the current shape/stride layout descriptors, but I can't think of a cleaner way right now either.

I think in the meantime we just need to show correctness (lm-eval) and speed (vllm bench) for your change.
lm-eval example https://github.com/vllm-project/vllm/pull/23265  vllm-bench example https://github.com/vllm-project/vllm/pull/20338.

PS changed the title topics. I think it's still not immediately clear

vllm/v1/attention/backends/flash_attn.py
heheda12345: > You mean (num_blocks, num_layers, 2, block_size, ...) ?

Yes. 

> this PR doesn't work for hybrid models, it assumes all layers have the same layout

Yes it doesn't support it now. But I don't want to see a branch like
```
if not is_hybrid:
      (2, num_blocks, num_layers, block_size, ...)
else:
    (num_blocks, num_layers, 2, block_size, ...) 
```
When we start to work on hybrid models.

orozery: I actually tried to change the shape of flash attention from `(2, num_blocks, block_size, num_kv_heads, head_size)` to `(num_blocks, 2, block_size, num_kv_heads, head_size)`.
It seems to work, and the only change is in replacing this line:
```
key_cache, value_cache = kv_cache.unbind(0)
```
with:
```
key_cache, value_cache = kv_cache.unbind(1)
```

But I think this should be a separate PR.

heheda12345: The unbind change is unnecessary. You can do a stride hack like https://github.com/vllm-project/vllm/pull/23299 did to support (num_blocks, 2, hidden_size)  in flash attention.

orozery: Is this change ok?
```
       if cache_layout == "NHD" and include_num_layers_dimension:
            # (num_blocks, 2, num_layers, block_size, num_kv_heads, head_size)
            return (2, 1, 0, 3, 4, 5)
        elif cache_layout == "NHD":
            # (num_blocks, 2, block_size, num_kv_heads, head_size)
            stride_order = (1, 0, 2, 3, 4)
        elif cache_layout == "HND" and include_num_layers_dimension:
            # (num_blocks, 2, num_kv_heads, num_layers, block_size, head_size)
            return (2, 1, 4, 0, 3, 5)
        elif cache_layout == "HND":
            # (num_blocks, 2, block_size, num_kv_heads, head_size)
            stride_order = (1, 0, 3, 2, 4)
```

Actually, this will break existing connectors (e.g. nixl) assuming (2, num_blocks,...)

orozery: So maybe this:
```
        cache_layout = get_kv_cache_layout()
        if cache_layout == "NHD" and include_num_layers_dimension:
            # (num_blocks, num_layers, 2, block_size, num_kv_heads, head_size)
            return (2, 0, 1, 3, 4, 5)
        elif cache_layout == "NHD":
            stride_order = (0, 1, 2, 3, 4)
        elif cache_layout == "HND" and include_num_layers_dimension:
            # (num_blocks, num_kv_heads, num_layers, 2, block_size, head_size)
            return (2, 4, 0, 1, 3, 5)
        elif cache_layout == "HND":
            stride_order = (0, 1, 3, 2, 4)
```

orozery:
@NickLucche I've appended the changes of the offloading connector utilizing the new layout.
For 2 reasons (as you suggested):
1. Verifying the correctness of the new layout (which is only enabled per-connector request).
2. Testing performance with the new layout.

For (1) I've added a basic test (`_accuracy_test` inside test_cpu_offloading.py).
For (2), there are 2 cases:
1. Performance impact on KV transfer time. This is already measured today in test_cpu_offlading.py.
    The results show drastic improvement in latency. For example (output taken from running test_cpu_offloading.py):
Current layout: 
```
Average times:
    Cold: 147.80ms
    GPU hit: 9.23ms
    CPU hit: 109.41ms
```

New layout:
```
Average times:
    Cold: 145.48ms
    GPU hit: 9.03ms
    CPU hit: 43.14ms
```
 
 2. Performance impact on model execution. I did notice any significant change in performance. I ran (once with flashinfer, and once with flash attention):
 ```
 vllm serve meta-llama/Llama-3.1-8B-Instruct --no-enable-prefix-caching --kv-transfer-config '{"kv_connector": "OffloadingConnector", "kv_role": "kv_both", "kv_connector_extra_config": {"block_size": 1048576, "num_cpu_blocks": 1}}'
 ```
 Followed by:
 ```
 vllm bench serve --dataset-name sharegpt --model meta-llama/Llama-3.1-8B-Instruct --dataset-path ./ShareGPT_V3_unfiltered_cleaned_split.json
 ```

 For flash-attention:
 Current layout:
 ```
Benchmark duration (s):                  56.65
Total input tokens:                      215196
Total generated tokens:                  197247
Request throughput (req/s):              17.65
Output token throughput (tok/s):         3481.67
Peak output token throughput (tok/s):    6495.00
Peak concurrent requests:                1000.00
Total Token throughput (tok/s):          7280.17
Mean TTFT (ms):                          17969.26
Median TTFT (ms):                        16236.38
P99 TTFT (ms):                           42501.62
Mean TPOT (ms):                          67.87
Median TPOT (ms):                        64.01
P99 TPOT (ms):                           147.37
Mean ITL (ms):                           58.03
Median ITL (ms):                         52.68
P99 ITL (ms):                            154.11
 ```
 New layout:
 ```
Benchmark duration (s):                  57.63
Total input tokens:                      215196
Total generated tokens:                  198186
Request throughput (req/s):              17.35
Output token throughput (tok/s):         3438.83
Peak output token throughput (tok/s):    6758.00
Peak concurrent requests:                1000.00
Total Token throughput (tok/s):          7172.80
Mean TTFT (ms):                          18064.90
Median TTFT (ms):                        16198.84
P99 TTFT (ms):                           42469.16
Mean TPOT (ms):                          68.03
Median TPOT (ms):                        63.70
P99 TPOT (ms):                           148.00
Mean ITL (ms):                           57.78
Median ITL (ms):                         50.87
P99 ITL (ms):                            157.69
 ```



 For flash-infer:
 Current layout:
 ```
Benchmark duration (s):                  49.95
Total input tokens:                      215196
Total generated tokens:                  197970
Request throughput (req/s):              20.02
Output token throughput (tok/s):         3963.21
Peak output token throughput (tok/s):    6720.00
Peak concurrent requests:                1000.00
Total Token throughput (tok/s):          8271.26
Mean TTFT (ms):                          15721.31
Median TTFT (ms):                        14397.39
P99 TTFT (ms):                           35988.98
Mean TPOT (ms):                          58.00
Median TPOT (ms):                        52.94
P99 TPOT (ms):                           138.76
Mean ITL (ms):                           49.15
Median ITL (ms):                         40.10
P99 ITL (ms):                            140.68
 ```
 New layout:
 ```
Benchmark duration (s):                  49.83
Total input tokens:                      215196
Total generated tokens:                  197685
Request throughput (req/s):              20.07
Output token throughput (tok/s):         3966.81
Peak output token throughput (tok/s):    6816.00
Peak concurrent requests:                1000.00
Total Token throughput (tok/s):          8285.01
Mean TTFT (ms):                          15633.70
Median TTFT (ms):                        14393.79
P99 TTFT (ms):                           35934.41
Mean TPOT (ms):                          57.75
Median TPOT (ms):                        53.03
P99 TPOT (ms):                           138.87
Mean ITL (ms):                           49.10
Median ITL (ms):                         40.00
P99 ITL (ms):                            141.52
 ```

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -214,12 +214,20 @@ def get_kv_cache_shape(
...
-    def get_kv_cache_stride_order() -> tuple[int, ...]:
+    def get_kv_cache_stride_order(
+        include_num_layers_dimension: bool = False,
+    ) -> tuple[int, ...]:
         # `stride_order` indicates the permutation that gets us from
         # `get_kv_cache_shape` to the actual memory layout we want.
         cache_layout = get_kv_cache_layout()
-        if cache_layout == "NHD":
+        if cache_layout == "NHD" and include_num_layers_dimension:
+            # (num_blocks, 2, num_layers, block_size, num_kv_heads, head_size)
```
heheda12345: can you also move 2 behind num_layer for flash infer?

vllm/v1/attention/backends/flash_attn.py
heheda12345: This layout make sense for me. Thanks!

vllm/distributed/kv_transfer/kv_connector/v1/base.py
```diff
@@ -496,6 +493,23 @@ def get_required_kvcache_layout(cls, vllm_config: "VllmConfig") -> str | None:
             )
         return None
 
+    @classmethod
+    def prefer_cross_layer_blocks(cls) -> bool:
+        """
```
NickLucche: nit: I personally prefer using a classvar attribute here like we have in other interfaces, unless you feel the validation is needed. I think defaulting to False is fine here

vllm/v1/attention/backends/flash_attn.py
```diff
@@ -99,12 +105,20 @@ def get_kv_cache_shape(
...
         cache_layout = get_kv_cache_layout()
-        if cache_layout == "NHD":
+        if cache_layout == "NHD" and include_num_layers_dimension:
+            # (num_blocks, num_layers, 2, block_size, num_kv_heads, head_size)
+            return (2, 0, 1, 3, 4, 5)
+        elif cache_layout == "NHD":
             stride_order = (0, 1, 2, 3, 4)
+        elif cache_layout == "HND" and include_num_layers_dimension:
+            # (num_blocks, num_kv_heads, num_layers, 2, block_size, head_size)
+            return (2, 4, 0, 1, 3, 5)
```
NickLucche, line 115: oh I see how that works well for nixl too, well done!

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -4925,12 +4709,21 @@ def initialize_kv_cache_tensors(
...
+            )
+        else:
+            # Fallback to the general case
+            # Initialize the memory buffer for KV cache
+            kv_cache_raw_tensors = self._allocate_kv_cache_tensors(kv_cache_config)
+
+            # Change the memory buffer to the desired shape
+            kv_caches = self._reshape_kv_cache_tensors(
+                kv_cache_config, kv_cache_raw_tensors, kernel_block_sizes
+            )
```
NickLucche: wdyt of something like:
```suggestion
       with self.maybe_allocate_uniform_kv_caches(self.cross_layers_kv_cache, kv_cache_config, kernel_block_sizes) as kv_caches:
        if kv_caches is not None:
            # Fallback to the general case
            # Initialize the memory buffer for KV cache
            kv_cache_raw_tensors = self._allocate_kv_cache_tensors(kv_cache_config)

            # Change the memory buffer to the desired shape
            kv_caches = self._reshape_kv_cache_tensors(
                kv_cache_config, kv_cache_raw_tensors, kernel_block_sizes
            )
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -4693,9 +4329,155 @@ def may_reinitialize_input_batch(
...
                 is_pooling_model=self.is_pooling_model,
-                num_speculative_tokens=self.num_spec_tokens,
+                num_speculative_tokens=(
+                    self.vllm_config.speculative_config.num_speculative_tokens
+                    if self.vllm_config.speculative_config
+                    else 0
+                ),
             )
 
+    def use_uniform_kv_cache(self, kv_cache_config: KVCacheConfig) -> bool:
```
NickLucche: @orozery what do you think of moving these two methods into the kv_connector mixin with an interface as suggested in the comment below?
The rationale is that the model_runner is undergoing a parallel effort with a leaner v2 version.
I feel that grouping this feature more neatly into the kv connector code responsibility will ease the work for the upcoming refactoring and will also allow us to apply upcoming patches more independently.

Let me know if you find this suggestion hard to implement in practice.

NickLucche (commented):
Apologies for the late review @orozery and thanks a lot for your patience!

This is good work, I only left one major comment that I would like to get your opinion on, about moving the uniform kv caches methods into the mixin for better separation of concerns. 

I understand this feature may turn out to be useful even outside kv offloading/transfer, but I think for now as we're already gating it behind a kv connector, it would make sense to minimize the number of changes to the model runner as well, given its upcoming refactor.

vllm/v1/worker/gpu_model_runner.py
orozery: I will move to mixin, but keep it in 2 functions (`use_uniform_kv_cache` + `_allocate_uniform_kv_caches`).
This was the preferred way when we discussed it here:
https://github.com/vllm-project/vllm/pull/27743#discussion_r2484942678

NickLucche: yeah absolutely, I am just focusing on containing the changes to the mixin

NickLucche (approved):
LGTM after latest changes, let's quickly wait on @heheda12345 to approve too

vllm/v1/attention/backends/flash_attn.py
```diff
@@ -99,12 +99,20 @@ def get_kv_cache_shape(
...
```
ApostaC, line 115: `include_num_layers_dimension` becomes confusing here.
From the doc string, when this argument is true, it only implies `includes an additional num_layers dimension`. However, this does not mean we will use the new layout or not. Right now, I feel like this may confuse future contributors.

To be clear, either we introduce another argument to explicitly specify the use of the new cache layout, or we make "include num_layers dimension" by default, and change the current argument to something like `cross_layer_kv_cache_enabled`.

vllm/v1/attention/backends/mla/indexer.py
```diff
@@ -45,7 +45,11 @@ def get_kv_cache_shape(
         return (num_blocks, block_size, head_size)
 
     @staticmethod
-    def get_kv_cache_stride_order() -> tuple[int, ...]:
+    def get_kv_cache_stride_order(
+        include_num_layers_dimension: bool = False,
+    ) -> tuple[int, ...]:
+        if include_num_layers_dimension:
+            return (0, 1, 2, 3)
```
ApostaC, line 52: Why do we have num layers as the outer-most for this specific case?

vllm/distributed/kv_transfer/kv_connector/v1/base.py
```diff
@@ -142,6 +142,18 @@ class KVConnectorMetadata(ABC):  # noqa: B024
...
+    Base class for KV connectors.
+
+    Attributes:
+        prefer_cross_layer_blocks (bool): Indicates whether this connector
+            prefers KV blocks that hold KV data for all layers (for speeding
+            up KV data transfers).
+            Defaults to False.
+    """
+
+    prefer_cross_layer_blocks: ClassVar[bool] = False
```
ApostaC, line 155: Quick question: what will happen if the connector implementation sets it to true? Will that impact vLLM's initialization process?

vllm/distributed/kv_transfer/kv_connector/v1/base.py
```diff
@@ -226,6 +238,24 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
         """
         return
 
+    def register_cross_layers_kv_cache(
+        self, kv_cache: torch.Tensor, attn_backend: type["AttentionBackend"]
+    ):
```
ApostaC, line 243: Why do we need to pass `attn_backend` here? If we have a pre-defined shape of the underlying kv cache tensor, it should be the same across all the attn backends, right?

If so, can we also merge this function with the existing `register_kvcaches` function? Otherwise, it is not clear that when those two functions will be called.

ApostaC (commented):
Thanks for the reminder @NickLucche.
I've got some clarification questions.

vllm/v1/attention/backends/flash_attn.py
orozery: > To be clear, either we introduce another argument to explicitly specify the use of the new cache layout, or we make "include num_layers dimension" by default, and change the current argument to something like `cross_layer_kv_cache_enabled`.

Sorry I don't understand your suggestion.
You're suggesting that we add another argument to `get_kv_cache_stride_order` ? This does not make sense to me.
Or do you suggest adding it somewhere else? where?

The function `get_kv_cache_stride_order` is used by both the current layout (which does not involve num_layers as a dimension), and by the new layout.
The default is kept at False as this is the current behavior and the more general one.

vllm/v1/attention/backends/mla/indexer.py
orozery: This is documented in the definition of `get_kv_cache_stride_order` in the base class:

```
   Args:
            include_num_layers_dimension: if True, includes an additional
                num_layers dimension, which is assumed to be prepended
                to the logical KV cache shape.
```

vllm/distributed/kv_transfer/kv_connector/v1/base.py
orozery: Not necessarily. 

The conditions that effect whether this cross-layers layout will be used is documented in the `use_uniform_kv_cache` function:

```
        Note this layout will only be applied given 3 conditions:
        1. The KV Cache config contains just a single group where all layers
            have the same page size.
        2. A KV connector is configured, and the KV connector instance prefers
            to use this layout (prefer_cross_layer_blocks() returns True)
        2. The flash attention backend supports this layout
            (get_kv_cache_stride_order(True) includes a placement for a
            num_layers dimension)
```

vllm/distributed/kv_transfer/kv_connector/v1/base.py
orozery: The single cross-layers kv-cache tensor always matches a single attention backend.
Of course you connector could re-figure this out but this makes it more easy and straight-forward for the connector.
We could avoid this function and re-use `register_kvcaches`. This was actually my first thought.
However, this requires defining a "dummy" layer name for the cross-layer tensor.
I find this is hacky, and prefer we define an explicit function like I defined here.

This new function may only be called if a connector set `prefer_cross_layer_blocks` to True, and only one of two register functions will be called. This is documented in the definition in the base class:
```
        This function will only be called for models with uniform layers,
        and only if the KV connector returns True on
        prefers_cross_layer_blocks().
        Only one of the functions
        {register_kv_caches, register_cross_layers_kv_cache} will be called.
```

vllm/attention/backends/abstract.py
```diff
@@ -76,7 +76,28 @@ def get_kv_cache_shape(
...
     @staticmethod
-    def get_kv_cache_stride_order() -> tuple[int, ...]:
+    def get_kv_cache_stride_order(
+        include_num_layers_dimension: bool = False,
+    ) -> tuple[int, ...]:
+        """
+        Get the physical (memory layout) ordering of the kv cache dimensions.
+        e.g. if the KV cache shape is
+        [num_blocks, block_size, num_heads, head_size],
+        and get_kv_cache_stride_order returns (0, 2, 1, 3) then the physical
```
heheda12345: maybe a better example?
this example 

`kv_cache_shape = [num_blocks, block_size, num_heads, head_size]`
`physical ordering = [num_blocks, num_heads, block_size, head_size]`
`stride = (0, 2, 1, 3)`

can be both interpreted as
```
kv_cache_shape[i] = physical_ordering[stride[i]]
```
and
```
physical_ordering[i] = kv_cache_shape[stride[i]]
```

vllm/distributed/kv_transfer/kv_connector/v1/base.py
heheda12345: > The single cross-layers kv-cache tensor always matches a single attention backend.

Can't agree. This layout can be extended to full + swa / full + mamba cases in the future. And don't want to change connector API when we start to work on that.

heheda12345: After a second thought, I feel more comfortable with the current API when supporting multiple attention backends. To use cross-layer KV blocks with multiple attention backends, we need to let all backends to make agreement on the underlying stride and it's OK to pass one attention backend that can represent all backends. And the logic for this agreement is needed for all connectors so I prefer to do it in model runner instead of doing it inside each connector.

heheda12345 (approved):
My only concern is this comment. 
https://github.com/vllm-project/vllm/pull/27743/files#r2543713465

vllm/v1/attention/backends/flash_attn.py
ApostaC: What I mean is:
When `include_num_layers_dimension = True`, it is not clear whether 
1. We only prepend the layer dimension to the return value while keeping the old layout (which is the case in `vllm/v1/attention/backends/mla/indexer.py`), ***or***
2. We include the layer dimension in the return value **PLUS** we use the new layout.

To me, there are clear 2x2 cases
| | Return value has an extra dimension | Return value doesn't have an extra dimension | 
|--------|--------|--------|
| Use the normal layout | case A | case B |
| Use the cross-layer layout | case C | case D |

In the above table, I would say D doesn't make sense (because we always want to return with layer dimension when we are using cross-layer layout), but all of the cases A, B, and C make sense, and they should map to different return values. Obviously, using a single boolean argument cannot cover 3 different cases.

In your current implementation, when `include_num_layers_dimension = True`:
- In `vllm/v1/attention/backends/mla/indexer.py`, it means **case A** in the above table
- In other attn backends, it means **case C** in the above table

vllm/v1/attention/backends/mla/indexer.py
ApostaC: See https://github.com/vllm-project/vllm/pull/27743#discussion_r2544443777

vllm/distributed/kv_transfer/kv_connector/v1/base.py
ApostaC: > However, this requires defining a "dummy" layer name for the cross-layer tensor.

Dumb question: If the goal is to have the layer name, why don't we just pass the layer name to the connector? Or is it impossible?

orozery: > > However, this requires defining a "dummy" layer name for the cross-layer tensor.
> 
> Dumb question: If the goal is to have the layer name, why don't we just pass the layer name to the connector? Or is it impossible?

Not sure I understand your question.
The current API (`register_kv_caches`) passes a dict mapping layer names to kv cache tensors. Those layer names are real, and the connector can use those layer names to reference other kv cache related structs (like attention backends).
Had I tried to use this API to pass in a cross-layer kv cache tensor, the connector would have to implicitly detect as not to wrongfully correspond this tensor with that of a "true" layer.

vllm/v1/attention/backends/flash_attn.py
orozery: Case D is impossible.
This is because in `use_uniform_kv_cache` we check:
```
        # check that attention backend include a layers dimension
        return len(kv_cache_stride_order) == len(kv_cache_shape) + 1
```
From the perspective of the implementor of an AttentionBackend.
You read the base class to understand how to implement `get_kv_cache_stride_order`.
You don't have to implement this function as it not an abstract function.
If you don't implement it, you will get Physical Order = Logical Order.
If you do implement it, but you can't/(don't want to) support a layers dimension, you can simply not include a layers dimension.

I've added clarifications to the documentation of `get_kv_cache_stride_order`.
BTW before this PR there was no documentation for this function whatsoever :)

@heheda12345 I've changed the example for a more complex permutation which I think should address the ambiguity you mentioned.

heheda12345: I think we always use B&C. Is indexer a mistake? I feel `include_num_layers_dimension` is confusing but will it be good if we change the argument name, e.g., `def get_kv_cache_stride_order(enable_contiguous_layer)`?

orozery: > I think we always use B&C. Is indexer a mistake?

Prepending a layers dimensions like I did in `indexer.py` seems like a don't care to me from the attention backend POV.
Prepending a layers dimension basically means that your per-layer KV cache tensor is still contiguous in memory, but physically it's a view of larger physical tensor.
This actually does not make any true difference in performance, so I'm also good with not prepending the layers dimensions on `indexer.py`.



> I feel `include_num_layers_dimension` is confusing but will it be good if we change the argument name, e.g., `def get_kv_cache_stride_order(enable_contiguous_layer)`?

I think `enable_contiguous_layer` is misleading.
It implies that `enable_contiguous_layer` will be enabled, which is not necessarily true.
`include_num_layers_dimension` is the actual explicit name implying the implementor of `get_kv_cache_stride_order` what should be done if this argument is either True or False.

@heheda12345 Please see my revised docstring on `get_kv_cache_stride_order`.
Of course I would be happy to improve it or make any other API changes that can help clarify things.

ApostaC (approved):
Unblocking this PR : )

TheTinyTeddy:
@orozery  Could I ask what performance impact will the contiguous kv block (layer-wise and K&V-wise) have on the main GPU computation (e.g. attention) in the case of not using KV offload? 

I was wondering why not use this as default in the first place, was it purely due to coding convenience?

PR #27744, opened 2025-10-29
[FIXBUG] Qwen3VL hallucinations without Contiguous on Torch.SDPA

JartX:
The pull request: https://github.com/vllm-project/vllm/pull/23207
I readded the logic necessary to prevent hallucinations in the TORCH.SDPA support backend. This pull request simply reintroduces the code that was removed but is absolutely necessary.

@DarkLight1337 @tjtanaa @lgeiger @Lucaskabela

### DIFF
--- a/vllm/model_executor/models/qwen2_5_vl.py
+++ b/vllm/model_executor/models/qwen2_5_vl.py
@@ -844,10 +844,8 @@ class Qwen2_5_VisionTransformer(nn.Module):
         self,
         cu_seqlens: torch.Tensor,
     ) -> tuple[torch.Tensor, torch.Tensor]:
-        max_seqlen, seqlens = (
-            torch.zeros(1, device=cu_seqlens.device),
-            torch.zeros(1, device=cu_seqlens.device),
-        )
+        max_seqlen = torch.zeros([], device=cu_seqlens.device)
+        seqlens = torch.zeros(1, device=cu_seqlens.device)
         if (
             self.attn_backend == _Backend.FLASH_ATTN
             or self.attn_backend == _Backend.ROCM_AITER_FA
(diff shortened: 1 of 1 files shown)


vllm/model_executor/models/qwen2_5_vl.py
```diff
@@ -836,8 +842,10 @@ def compute_attn_mask_seqlen(
         self,
         cu_seqlens: torch.Tensor,
     ) -> tuple[torch.Tensor, torch.Tensor]:
-        max_seqlen = torch.zeros([], device=cu_seqlens.device)
-        seqlens = torch.zeros(1, device=cu_seqlens.device)
+        max_seqlen, seqlens = (
+            torch.zeros(1, device=cu_seqlens.device),
```
DarkLight1337: Is it intended that `max_seqlen` is initialized with size `1` instead of `[]`?

JartX: No, return original logic :) thanks @DarkLight1337

DarkLight1337 (approved):

Lucaskabela (approved):

vllm/model_executor/models/qwen2_5_vl.py
```diff
@@ -428,6 +428,14 @@ def forward(
             )
         elif self.attn_backend == _Backend.TORCH_SDPA:
             # Execute attention entry by entry for speed & less VRAM.
+            from vllm.platforms import current_platform
+
+            # Never remove the next contiguous logic
+            # Without it, hallucinations occur with the backend
+            if current_platform.is_rocm():
```
Lucaskabela, line 435: Ah I was wondering why I didn't observe this in testing on my end - I didn't test on rocm :( Sorry for missing this, and thanks for documenting in code so we avoid easy to miss mistakes like this in the future!

PR #27750, opened 2025-10-29
[Refactor] Remove `VLLM_DEEPEP_LOW_LATENCY_ALLOW_NVLINK`

yewentao256:
## Purpose

A follow up PR for https://github.com/vllm-project/vllm/pull/27677

After consideration, this shouldn't be exposed to users and there are issues if turning off. So let's remove this env to simplify the codebase.


tlrmchlsmth (approved):
I support this change - I don't see any compelling reason why we would want to turn this off.

Ideally we answer the question "Why does DeepEP offer this option?" before merging

yewentao256:
Let's land this pr to clean the env variables a little bit, if there is demand tuning whether to use NVLINK, we can re-add it (I don't think we will have though

yewentao256:
@wpc Feel free to have a PR

PR #27752, opened 2025-10-29
`reasoning_content` -> `reasoning`

hmellor:
- `reasoning_content` is only named this way for historical reasons (this is what deepseek originally used)
- With the release of GPT-OSS, OpenAI provided guidance that reasoning content should be returned in `reasoning` for the Chat Completions API (even though there is no official spec for handling this in the published OpenAI specs)
- This behaviour matches OpenRouter

---

- `reasoning_content` is kept for backwards compatibiltity (for now)
- https://github.com/vllm-project/vllm/issues/27755 to discuss the removal of `reasoning_content`


vllm/entrypoints/openai/protocol.py
```diff
@@ -2031,8 +2031,15 @@ class ChatMessage(OpenAIBaseModel):
...
 
     # vLLM-specific fields that are not in OpenAI spec
+    reasoning: str | None = None
     reasoning_content: str | None = None
 
+    @model_validator
+    def handle_deprecated_reasoning_content(self):
+        if self.reasoning_content is not None:
+            self.reasoning = self.reasoning_content
+        return self
```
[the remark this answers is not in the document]
hmellor: Sure, I'll use after

examples/online_serving/openai_chat_completion_with_reasoning.py
chaunceyjiang, line 1: Could we keep one example to ensure that `reasoning_content` still works until we explicitly remove it?

hmellor: Instead of an example, I'll add a test. This way we know it will still work, and no user will accidentally start using the old vesion.

DarkLight1337 (approved):

PR #26385, opened 2025-10-07
[BugFix] Fix async scheduling + request preemption

njhill:
Ensure model runner is refreshed with all request token ids following preemption, so that correct input ids are used.

PR #27756, opened 2025-10-29
[BugFix] Handle unscheduled requests properly when async scheduling

njhill:
There may be circumstances other then preemption where a running request is temporarily not scheduled in the batch for some step(s). These need to be handled similarly to the async scheduling + preemption fix made in https://github.com/vllm-project/vllm/pull/26385.

This PR also streamlines how resumed requests are recorded in `CachedRequestData`.

### DIFF
--- a/vllm/v1/core/sched/output.py
+++ b/vllm/v1/core/sched/output.py
@@ -2,7 +2,10 @@
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 
 from dataclasses import dataclass
+from functools import cached_property
 from typing import TYPE_CHECKING
+
+from typing_extensions import deprecated
 
 from vllm._bc_linter import bc_linter_include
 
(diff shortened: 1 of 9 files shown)


vllm/v1/core/sched/scheduler.py
```diff
@@ -444,14 +445,9 @@ def schedule(self) -> SchedulerOutput:
...
-                        0
-                        < self.scheduler_config.long_prefill_token_threshold
-                        < num_new_tokens
-                    ):
-                        num_new_tokens = (
-                            self.scheduler_config.long_prefill_token_threshold
-                        )
+                    threshold = self.scheduler_config.long_prefill_token_threshold
+                    if 0 < threshold < num_new_tokens:
+                        num_new_tokens = threshold
```
njhill, line 450: Unrelated simplification, hurt me to look at that formatting :)

vllm/v1/core/sched/output.py
```diff
@@ -96,16 +99,16 @@ def anon_repr(self) -> str:
 @dataclass
 class CachedRequestData:
     req_ids: list[str]
-    # If resumed_from_preemption is False, new_block_ids will be appended to
-    # the request's block IDs. If True, new_block_ids will be used as the
+    # For request ids not in resumed_req_ids, new_block_ids will be appended to
+    # the request's block IDs. For those in the set, new_block_ids will be used as the
     # request's block IDs instead of appending to the existing block IDs.
-    resumed_from_preemption: list[bool]
+    resumed_req_ids: set[str]
```
njhill, line 105: Changing this to a set since these will be rare and we currently are creating a `[None] * batch_size` list every time.

vllm/v1/core/sched/output.py
```diff
@@ -114,13 +117,26 @@ class CachedRequestData:
...
+    def resumed_from_preemption(self) -> list[bool]:
+        return [req_id in self.resumed_req_ids for req_id in self.req_ids]
+
+    @cached_property
+    @deprecated("use all_token_ids field")
+    def resumed_req_token_ids(self) -> list[list[int] | None]:
+        return [
+            self.all_token_ids[req_id] if req_id in self.resumed_req_ids else None
+            for req_id in self.req_ids
+        ]
```
njhill, line 131: These are for backwards compatibility.

benchislett (approved):
LGTM, discussed online. I would like to see a test case for the new coverage, can go ahead and merge at any point.

Issue #22497, opened 2025-08-08
[Bug]: Crashed when loading ggml quantized Gemma3


Imagium719:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
==============================
        System Info
==============================
OS                           : Ubuntu 24.04.2 LTS (x86_64)
GCC version                  : (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
Clang version                : Could not collect
CMake version                : version 3.28.3
Libc version                 : glibc-2.39

==============================
       PyTorch Info
==============================
PyTorch version              : 2.7.1+cu126
Is debug build               : False
CUDA used to build PyTorch   : 12.6
ROCM used to build PyTorch   : N/A

==============================
      Python Environment
==============================
Python version               : 3.12.11 | packaged by Anaconda, Inc. | (main, Jun  5 2025, 13:09:17) [GCC 11.2.0] (64-bit runtime)
Python platform              : Linux-6.8.0-64-generic-x86_64-with-glibc2.39

==============================
       CUDA / GPU Info
==============================
Is CUDA available            : True
CUDA runtime version         : 12.0.140
CUDA_MODULE_LOADING set to   : LAZY
GPU models and configuration : 
GPU 0: NVIDIA GeForce RTX 3080
GPU 1: NVIDIA GeForce RTX 3080

Nvidia driver version        : 570.158.01
cuDNN version                : Could not collect
HIP runtime version          : N/A
MIOpen runtime version       : N/A
Is XNNPACK available         : True

==============================
          CPU Info
==============================
Architecture:                         x86_64
CPU op-mode(s):                       32-bit, 64-bit
Address sizes:                        46 bits physical, 48 bits virtual
Byte Order:                           Little Endian
CPU(s):                               64
On-line CPU(s) list:                  0-63
Vendor ID:                            GenuineIntel
Model name:                           Intel(R) Xeon(R) CPU E5-2698 v3 @ 2.30GHz
CPU family:                           6
Model:                                63
Thread(s) per core:                   2
Core(s) per socket:                   16
Socket(s):                            2
Stepping:                             2
CPU(s) scaling MHz:                   46%
CPU max MHz:                          3600.0000
CPU min MHz:                          1200.0000
BogoMIPS:                             4589.34
[...]

PR #27772, opened 2025-10-29
[Model] Add Gemma3 GGUF multimodal support

lucianommartins:
# [Model] Add Gemma3 GGUF multimodal support

## Summary

This PR enables full multimodal inference support for Gemma3 models in GGUF format, allowing users to run quantized Gemma3 multimodal models with both text-only and image+text prompts. The implementation adds automatic detection and loading of `mmproj.gguf` files for vision tower weights while maintaining complete isolation to ensure zero impact on existing Gemma3 HuggingFace models or other model architectures.

**Resolves**: #22497

---

### 1. **Automatic GGUF Multimodal Detection**

The model configuration now automatically detects when a Gemma3 GGUF model has an accompanying `mmproj.gguf` file and switches to the multimodal architecture:

- **Location**: `vllm/config/model.py`
- **Mechanism**: Helper function `_detect_gguf_multimodal_gemma3()` checks for `mmproj-*.gguf` in model directory
- **Behavior**: Sets `model_cls = "Gemma3ForConditionalGeneration"` when multimodal projector is found
- **Fallback**: Uses `Gemma3ForCausalLM` for text-only GGUF models

**Example usage**:

```python
from vllm import LLM

# Automatically detects mmproj.gguf and loads multimodal architecture
llm = LLM(
    model="/path/to/gemma-3-4b-it-q4_0.gguf",
    tokenizer="google/gemma-3-4b-it",
)
```

### 2. **GGUF Vision Tower Weight Loading**

Enhanced GGUF loader to handle multimodal weights from separate `mmproj.gguf` files:

- **Location**: `vllm/model_executor/model_loader/gguf_loader.py`, `weight_utils.py`
- **Tensor Mapping**: Maps GGUF vision encoder tensors to vLLM's SigLIP vision tower
  - `v.*.attn.{q,k,v,out}` → Vision attention weights
  - `v.*.mlp.{fc1,fc2}` → Vision FFN weights  
  - `v.post_ln.{weight,bias}` → Post-layernorm weights
  - `mm.0.{weight,bias}` → Multimodal projector weights
- **Quantization**: Vision weights loaded as unquantized `float16`/`bfloat16` (GGUF spec)
- **Isolation**: Only activated when `is_gguf_weight = True`

### 3. **GGUF Processor Loading**

Processors now correctly load from tokenizer path for GGUF models:

- **Location**: `vllm/transformers_utils/processor.py`
- **Fix**: Converts model repo paths (including GGUF files) to directory paths before loading processor
- **Benefit**: Avoids "not a directory" errors for `.gguf` file paths
- **Impact**: Applies to all processor types

### 4. **V1 Engine Multimodal Support**

Added multimodal embedding gathering in V1 GPU model runner:

- **Location**: `vllm/v1/worker/gpu_model_runner.py`
- **Function**: `_gather_mm_embeddings()` extracts vision embeddings after encoder execution
- **Integration**: Passes embeddings to model's `get_multimodal_embeddings()` for proper merging
- **Compatibility**: Works with both GGUF and HuggingFace multimodal models

---

## Isolation & Safety Guarantees

**All changes are strictly scoped to Gemma3 GGUF multimodal models only.** No other models or formats are affected.

### Isolation Verification

| Component | Guard Condition |
|-----------|----------------|
| **GGUF Loader** | `if quantization == "gguf"` |
| **Weight Utils** | `if quantization in ("gguf", "inc")` |
| **Model Config** | `if self.model.endswith('.gguf')` |
| **Gemma3 MM** | `if self._is_gguf` |
| **Processor** | Only for GGUF model paths |

### Non-Regression Tests

Comprehensive testing confirms zero impact on other model types:

- **Gemma3 HF text-only** (1B, 4B)
- **Gemma3 HF multimodal** (4B) 
- **Gemma3 GGUF text-only** (4B)
- **Gemma3 GGUF multimodal** (4B)

**No changes to**:
- Other GGUF models (Llama, Mistral, etc.)
- Other multimodal models (Phi3V, Qwen2-VL, etc.)
- V0 engine
- Kernel implementations

---

### Test Environment

- **Hardware**: NVIDIA GPU A100 80GB
- **Models Tested**:
  - `google/gemma-3-1b-it` (HF text-only)
  - `google/gemma-3-4b-it` (HF text-only & multimodal)
  - `google/gemma-3-4b-it-q4_0-gguf` (GGUF text-only & multimodal)

---

## Backward Compatibility

**100% backward compatible** - No breaking changes to existing functionality:

- Existing Gemma3 HF models work identically
- Existing GGUF text-only models unaffected
- No API changes
- No behavior changes for non-GGUF models
- No performance regressions

---

## Checklist

- [x] PR title follows format: `[Model] Add Gemma3 GGUF multimodal support`
- [x] Code follows [Google Python style guide](https://google.github.io/styleguide/pyguide.html)
- [x] No regressions verified through testing
- [x] Isolated to Gemma3 GGUF only - zero impact on other models

---

## Additional Context

This PR builds upon the recently merged text-only GGUF support for Gemma3 and extends it to support multimodal inference. The implementation carefully preserves all upstream bugfixes, including the recent newline token ordering fix (#27538).

**Tested on**: vLLM main branch (commit `b5d90f740`, Oct 29 2025)

Signed-off-by: Luciano Martins <lucianommartins@users.noreply.github.com>


requirements/test.txt
```diff
@@ -17,7 +17,6 @@ aiohttp==3.13.0
     #   aiohttp-cors
     #   datasets
     #   fsspec
-    #   gpt-oss
```
DarkLight1337, line 20: Can you revert the changes to requirements file and bypass pre-commit? I suspect this happened because your local Python environment is different from CI.

vllm/config/model.py
```diff
@@ -909,8 +995,6 @@ def _get_default_runner_type(
                 _, (runner_type, _) = match
                 return runner_type
 
-        return "generate"
```
DarkLight1337, line 858: Is this intended?

vllm/model_executor/models/gemma3_mm.py
```diff
@@ -616,6 +684,19 @@ def forward(
...
+            vision_embeddings = self.get_multimodal_embeddings(**kwargs)
+            inputs_embeds = self.get_input_embeddings(input_ids, vision_embeddings)
+            if vision_embeddings is not None and len(vision_embeddings) != 0:
+                kwargs = self.prepare_attn_masks(
+                    input_ids,
+                    positions,
+                    mask_dtype=self.dtype,
+                    **kwargs,
+                )
+            input_ids = None
```
Isotr0py:
```suggestion
```
V0 has been deprecated.

vllm/model_executor/models/gemma3_mm.py
```diff
@@ -525,16 +527,50 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = ""):
...
+                num_hidden_layers=27,
+                num_attention_heads=16,
+                num_channels=3,
+                image_size=896,
+                patch_size=14,
+                layer_norm_eps=1e-6,
+                attention_dropout=0.0,
+                num_image_tokens=256,
+                vision_use_head=False,  # Disable pooling head for Gemma3
+            )
```
Isotr0py: I don't like hard-coded config here. Can't we remap the ViT config from mm_proj.gguf checkpoint?

vllm/model_executor/layers/quantization/gguf.py
```diff
@@ -59,17 +62,19 @@ def get_config_filenames(cls) -> list[str]:
...
         self, layer: torch.nn.Module, prefix: str
     ) -> Optional["QuantizeMethodBase"]:
         if isinstance(layer, LinearBase):
             if is_layer_skipped_gguf(prefix, self.unquantized_modules):
                 return UnquantizedLinearMethod()
-            return GGUFLinearMethod(self)
+            return GGUFLinearMethod(self, self.model_arch)
         elif isinstance(layer, VocabParallelEmbedding):
-            return GGUFEmbeddingMethod(self)
+            return GGUFEmbeddingMethod(self, self.model_arch)
```
Isotr0py: I don't think we should couple quant method with specific model architecture.

vllm/config/model.py
```diff
@@ -560,6 +590,46 @@ def __post_init__(
...
+                        image_size=896,
+                        patch_size=14,
+                        layer_norm_eps=1e-6,
+                        attention_dropout=0.0,
+                        num_image_tokens=256,
+                        # Disable pooling head for Gemma3
+                        vision_use_head=False,
+                    )
+                    self.hf_config.mm_tokens_per_image = 256
+                    self.hf_config.image_token_index = 262144
```
Isotr0py: Please avoid any hardcoded config here.

Isotr0py (commented):
I think this PR introduces too many gemma3-mm GGUF specific hardcoded code. 

Besides this, I'm a bit confused why we need to use `GGUFQuantMethod` for a totally unquantized mm_proj checkpoint as well.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2821,7 +2837,7 @@ def propose_draft_token_ids(
                     target_hidden_states = hidden_states[token_indices]
 
             if self.supports_mm_inputs:
-                mm_embed_inputs = self._gather_mm_embeddings(
+                mm_embed_inputs, _ = self._gather_mm_embeddings(
```
DarkLight1337: Won't this affect other models as well?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -4271,7 +4288,9 @@ def may_reinitialize_input_batch(self, kv_cache_config: KVCacheConfig) -> None:
                 kernel_block_sizes=kernel_block_sizes,
                 is_spec_decode=bool(self.vllm_config.speculative_config),
                 logitsprocs=self.input_batch.logitsprocs,
-                logitsprocs_need_output_token_ids=self.input_batch.logitsprocs_need_output_token_ids,
+                logitsprocs_need_output_token_ids=(
```
DarkLight1337: Avoid formatting changes for lines that are not relevant to the PR

vllm/config/model.py
```diff
@@ -560,6 +589,45 @@ def __post_init__(
...
+                        num_attention_heads=16,
+                        num_channels=3,
+                        image_size=896,
+                        patch_size=14,
+                        layer_norm_eps=1e-6,
+                        attention_dropout=0.0,
+                        num_image_tokens=256,
+                        # Disable pooling head for Gemma3
+                        vision_use_head=False,
+                    )
```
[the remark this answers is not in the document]
lucianommartins: fixed.

vllm/model_executor/model_loader/gguf_loader.py
```diff
@@ -129,10 +132,175 @@ def _get_gguf_weights_map(self, model_config: ModelConfig):
...
+            mapping[f"v.blk.{layer_idx}.ffn_up.bias"] = (
+                f"{mlp_base}.fc2.bias"
+            )
+
+        # Post layer normalization
+        mapping["v.post_ln.weight"] = "vision_tower.vision_model.post_layernorm.weight"
+        mapping["v.post_ln.bias"] = "vision_tower.vision_model.post_layernorm.bias"
+
+        return mapping
+
```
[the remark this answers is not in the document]
lucianommartins: fixed.

vllm/model_executor/model_loader/utils.py
```diff
@@ -42,7 +42,92 @@ def initialize_model(
...
+                else:
+                    # Non-Gemma3: use standard path
+                    logger.debug("Multimodal GGUF detected for non-Gemma3 model")
+                    model_class, _ = get_model_architecture(model_config)
+            except (ImportError, AttributeError) as e:
+                logger.warning(f"Failed to load Gemma3ForConditionalGeneration: {e}")
+                model_class, _ = get_model_architecture(model_config)
+        else:
+            # Standard path
+            model_class, _ = get_model_architecture(model_config)
```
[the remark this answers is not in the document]
lucianommartins: fixed.

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -660,12 +668,12 @@ def runai_safetensors_weights_iterator(
 ) -> Generator[tuple[str, torch.Tensor], None, None]:
     """Iterate over the weights in the model safetensor files."""
     with SafetensorsStreamer() as streamer:
-        streamer.stream_files(hf_weights_files)
         total_tensors = sum(
             len(tensors_meta)
             for tensors_meta in streamer.files_to_tensors_metadata.values()
         )
 
+        streamer.stream_files(hf_weights_files)
```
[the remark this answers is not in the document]
lucianommartins: fixed.

vllm/model_executor/models/gemma3_mm.py
```diff
@@ -525,16 +529,48 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = ""):
...
+            config.mm_tokens_per_image = 256
+            config.image_token_index = 262144
+            logger.info("Created default vision_config for GGUF model")
```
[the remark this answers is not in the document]
lucianommartins: fixed.

vllm/transformers_utils/processor.py
```diff
@@ -150,12 +159,44 @@ def cached_processor_from_config(
...
+            and model_config.model.endswith('.gguf')
+            and 'gemma-3' in processor_name.lower()):
+        from transformers import Gemma3ImageProcessor
+        image_processor_override = Gemma3ImageProcessor(
+            size={'height': 896, 'width': 896}
+        )
+        logger.info(
+            "GGUF Gemma3: Pre-configuring image_processor with 896x896 "
+            f"(loading processor from {processor_name})")
+
```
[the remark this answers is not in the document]
lucianommartins: fixed.

requirements/test.txt
lucianommartins: I reverted the unnecessary requirements/test.txt. Thanks.

vllm/config/model.py
lucianommartins: It was a wrong removal. It was reintroduced.

vllm/model_executor/models/gemma3_mm.py
lucianommartins: done.

vllm/model_executor/models/gemma3_mm.py
lucianommartins: Fixed. Everything possible comes from the vision tower checkpoint now.

vllm/config/model.py
lucianommartins: Fixed. Everything possible comes from the vision tower checkpoint now.

vllm/v1/worker/gpu_model_runner.py
lucianommartins: fixed.

vllm/transformers_utils/config.py
```diff
@@ -475,6 +475,17 @@ def is_interleaved(config: PretrainedConfig) -> bool:
...
 
+def uses_custom_attention_masks(config: PretrainedConfig) -> bool:
+    """Detect if model uses custom attention mask generation for multimodal.
+
+    Some multimodal models require custom attention masks that enable
+    bidirectional attention between image tokens while maintaining causal
+    attention for text tokens. Currently applies to Gemma3 multimodal models.
+    """
+    architectures = getattr(config, "architectures", [])
+    return "Gemma3ForConditionalGeneration" in architectures
```
Isotr0py, line 488: I think we should support PrefixLM attention through FlexAttention backend instead of exposing a custom attention mask interface.

vllm/model_executor/models/gemma3_mm.py
Isotr0py: Seems you forgot to remove this ones?

vllm/model_executor/models/gemma3.py
```diff
@@ -442,14 +447,13 @@ def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
...
             if (
-                self.quant_config
-                and self.quant_config.get_name() == "gguf"
-                and name.endswith("norm.weight")
+                self._is_gguf_loading()
+                and "norm" in name
+                and len(loaded_weight.shape) == 1
             ):
-                loaded_weight -= 1
+                loaded_weight = loaded_weight - 1.0
```
Isotr0py: Is this change necessary?

vllm/model_executor/model_loader/utils.py
```diff
@@ -286,3 +382,72 @@ def configure_quant_config(
...
+    using hardcoded values. This makes the implementation robust across
+    different Gemma3 model sizes and future variations.
+
+    Args:
+        mmproj_path: Path to mmproj.gguf file (str or Path)
+
+    Returns:
+        SiglipVisionConfig if extraction succeeds, None otherwise
+    """
+    try:
```
Isotr0py: Why do we need to use `try` statement? Shouldn't we directly raise error when vision config extraction failed?

vllm/model_executor/model_loader/utils.py
```diff
@@ -42,7 +46,99 @@ def initialize_model(
...
+                else:
+                    # Non-Gemma3: use standard path
+                    logger.debug("Multimodal GGUF detected for non-Gemma3 model")
+                    model_class, _ = get_model_architecture(model_config)
+            except (ImportError, AttributeError) as e:
+                logger.warning("Failed to load Gemma3ForConditionalGeneration: %s", e)
+                model_class, _ = get_model_architecture(model_config)
+        else:
+            # Standard path
+            model_class, _ = get_model_architecture(model_config)
```
Isotr0py: This part is too specific to Gemma3, we should make it more general.

vllm/model_executor/model_loader/gguf_loader.py
```diff
@@ -129,10 +132,149 @@ def _get_gguf_weights_map(self, model_config: ModelConfig):
...
+        )
+        mapping["v.patch_embd.bias"] = (
+            "vision_tower.vision_model.embeddings.patch_embedding.bias"
+        )
+        mapping["v.position_embd.weight"] = (
+            "vision_tower.vision_model.embeddings.position_embedding.weight"
+        )
+
+        # SigLIP-So400m: 27 transformer layers
+        for layer_idx in range(27):
```
Isotr0py: Why hardcode 27?

vllm/model_executor/layers/quantization/gguf.py
```diff
@@ -355,7 +378,9 @@ class GGUFLinearMethod(LinearMethodBase):
         quant_config: The GGUF quantization config.
     """
 
-    def __init__(self, quant_config: GGUFConfig):
+    def __init__(self, quant_config: GGUFConfig, model_arch: str = ""):
+        self.model_arch = model_arch.lower()
+        self.is_gemma3 = "gemma3" in self.model_arch
```
Isotr0py:
```suggestion
    def __init__(self, quant_config: GGUFConfig):
```
We shouldn't couple model_arch with quant_method.

Isotr0py (commented):
Unfortunately, I think this PR is still under poor design, because there are too many gemma3 specific codes outside model implementation...

vllm/model_executor/layers/quantization/gguf.py
lucianommartins: @Isotr0py - The quantization layer requires model architecture info to determine the correct target `dtype` for weight dequantization, ensuring compatibility between quantized weight format (F16) and model compute dtype (BF16).

I experimented removing it before this last commit and:
- without `model_arch`, `is_gemma3` became `False` in `gguf.py` (the model isolation got broken here)
- also it caused the `target_dtype` to be `None` following `target_dtype = params_dtype if is_gemma3 else None → None`
- The dequantized weights from the checkpoint are in F16 and must be converted to BF16
- If I don't have the `target_dtype` the conversion is skipped
- The it ends in a crash on F16 weight @ BF16 activation

The idea is to have the quantized checkpoint properly loaded - if I move to model code (ie. `gemma3.py` or `gemma3_mm.py`), it is called after the loading so it fails.

from my experiments, this was the best way I've found to guarantee stability for dtype conversion for Gemma3 models without causing unwanted harm to other quantized models.

wdyt?

Isotr0py: > The dequantized weights from the checkpoint are in F16 and must be converted to BF16
If I don't have the target_dtype the conversion is skipped
The it ends in a crash on F16 weight @ BF16 activation

Hmmm, I think we should respect the `dtype` from `model_config` for unquantized dtype in this case.

lucianommartins: yeah I thought about that too, @Isotr0py - but the GGUF metadata doesn't hold any field with dtype / torch_dtype too.

I keep digging here, but I still haven't found any approach to have guardrails for Gemma3 only without manipulating the `model_arch`.

vllm/transformers_utils/config.py
lucianommartins: I've done one investigation during the day Today about it, @Isotr0py - some thoughts below, would love to hear your thoughts about it:

As you know, Gemma3 multimodal models use PrefixLM (Prefix Language Model) attention, where:

- Image tokens attend bidirectionally to all tokens
- Text tokens maintain causal attention (can only attend to previous tokens + all image tokens)

The critical challenge is that image token positions vary per sequence in a batch:

```
# Example batch with 3 sequences
Sequence 0: [IMG] [IMG] [IMG] "A" "dog" "running"     # Images at [0,1,2]
Sequence 1: "Describe" [IMG] [IMG] "this" "scene"     # Images at [1,2]  
Sequence 2: "What" "is" "the" "answer"                # No images
```

AFAIK vLLM's FlexAttention is designed for static attention patterns uniform across batches. The core `mask_mod` function signature is:

```
def mask_mod(b: int, h: int, q_idx: int, kv_idx: int) -> bool:
    return q_idx >= kv_idx  # Same logic applies to ALL sequences
The Problem: 
 receives only indices, not per-sequence metadata. It cannot answer "Is token at position X an image token?" for different sequences in the batch.
```

Using `doc_ids` to disambiguate sequences like:

```
# In FlexAttentionMetadata.__post_init__
self.doc_ids = _offsets_to_doc_ids_tensor(self.query_start_loc)
# Example: [0,0,0,0,0, 1,1,1,1,1,1,1, 2,2,2,2,2]
#           └─ seq 0 ─┘ └── seq 1 ───┘ └─ seq 2 ──┘
```

That works perfectly for scenarios like:

- Causal attention: `q_idx >= kv_idx` (same for all sequences)
- Sliding window: `abs(q_idx - kv_idx) < window_size` (same for all sequences)
- Document masking: Ensure tokens from same document attend together (uses `doc_ids`)

For Gemma3, we would need:

```
def gemma3_prefix_lm_mask_mod(b, h, q_idx, kv_idx):
    # Need per-sequence image token positions!
    is_q_image = ???  # How to determine this?
    is_kv_image = ???
    
    if is_q_image or is_kv_image:
        return True  # Bidirectional
    else:
        return q_idx >= kv_idx  # Causal
```

Challenge: We need to know which specific positions in each sequence contain image tokens, but 
 only receives indices.

I think I can approach it like extenting `FlexAttentionMetadata` - similar to this semi-pseudocode:

```
@dataclass
class FlexAttentionMetadata:
    # Existing fields
    doc_ids: torch.Tensor
    query_start_loc: torch.Tensor
    ...
    
    # NEW: Per-token image flag
    image_token_positions: torch.Tensor | None = None
    # Shape: [total_tokens_in_batch]
    # Example: [T,T,T,F,F,F, F,T,T,F,F, F,F,F,F]
```

Then modify the mask function:

```
def get_gemma3_prefix_lm_mask_mod(self):
    def final_mask_mod(b, h, q_idx, physical_kv_idx):
        # Convert to logical indices
        (is_valid, logical_q_idx, logical_kv_idx) = (
            self._convert_physical_to_logical(self.doc_ids, q_idx, physical_kv_idx)
        )
        
        # 1. Compute absolute token positions from request ID + logical indices
        # 2. Lookup image token flags from metadata tensor
        # 3. Apply PrefixLM logic
        
        q_req = self.doc_ids[q_idx]
        q_abs_idx = self.query_start_loc[q_req] + logical_q_idx
        kv_abs_idx = self.query_start_loc[q_req] + logical_kv_idx
        
        is_q_image = self.image_token_positions[q_abs_idx]
        is_kv_image = self.image_token_positions[kv_abs_idx]
        
        # PrefixLM logic
        if is_q_image or is_kv_image:
            return is_valid
        else:
            return is_valid and (logical_q_idx >= logical_kv_idx)
    
    return final_mask_mod
```

But it may add some questions that I think it is challenging to merge without knowing for sure:

1. In theory FlexAttention adoption will enhance the memory usage footprint - but how much? on all scenarios or specific ones (ie. only larger contexts, etc)?
2. What is the computational behavior change here? Will it introduce more computational overhead? If yes, what is the expected?

**My suggesation:** This PR is to introduce a critical missing part for Gemma3 developers - multimodal experiences with quantized models. So what if we proceed with this current custom masks approach then work on FlexAttention migration as a separate PR/RFC focused effort like: 

Phase 1: Prototype

Implement FlexAttention approach
Create minimal test harness
Measure actual `mask_mod` overhead vs memory savings

Phase 2: Validation

Test with various batch sizes and sequence lengths
Verify correctness and performance

Phase 3: Implementation (if results positive)

Full implementation with comprehensive testing
Migration path for Gemma3

This way we make this 2-step approach (the same way we are doing for the vision tower loading - starting with reading everything from the local file and in the next step loading from HF - the same way we agreed on the Slack discussion @Isotr0py. wdyt?

vllm/config/model.py
```diff
@@ -556,6 +584,77 @@ def __post_init__(
 
         architectures = self.architectures
         registry = self.registry
+
+        # GGUF multimodal: Force Gemma3ForConditionalGeneration architecture
+        # when mmproj file is present, before model resolution
+        if _detect_gguf_multimodal_gemma3(self.model):
+            is_gemma3 = any("gemma3" in str(arch).lower() for arch in architectures)
```
DarkLight1337:
```suggestion
            is_gemma3 = any(arch.lower().startswith("gemma3") for arch in architectures)
```

vllm/config/model.py
```diff
@@ -556,6 +584,77 @@ def __post_init__(
...
+                            "mmproj-*.gguf",
+                            "*mmproj*.gguf",
+                        ]
+                        for pattern in mmproj_patterns:
+                            mmproj_files = list(model_dir.glob(pattern))
+                            if mmproj_files:
+                                vision_config = extract_vision_config_from_gguf(
+                                    mmproj_files[0]
+                                )
+                                break
```
DarkLight1337: Maybe we can just get the files inside `_detect_gguf_multimodal_gemma3` to avoid repeated globs

vllm/config/model.py
```diff
@@ -556,6 +584,77 @@ def __post_init__(
...
+
+                    # Fallback to hardcoded defaults if extraction fails
+                    if vision_config is None:
+                        from transformers import SiglipVisionConfig
```
DarkLight1337: Is it possible to set this inside the modeling file instead?

vllm/model_executor/model_loader/utils.py
```diff
@@ -42,7 +46,99 @@ def initialize_model(
...
+                        "Gemma3ForConditionalGeneration"
+                    ]
+
+                    # Override vision_config for GGUF
+                    # (896x896 vs HF's 576x576)
+                    if (
+                        not hasattr(model_config.hf_config, "vision_config")
+                        or model_config.hf_config.vision_config is None
+                    ):
+                        from transformers import SiglipVisionConfig
```
DarkLight1337: This is repeated from the `ModelConfig` file - not ideal.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -3527,7 +3544,8 @@ def _dummy_sampler_run(
         logits = self.model.compute_logits(hidden_states)
         num_reqs = logits.size(0)
 
-        dummy_tensors = lambda v: torch.full((num_reqs,), v, device=self.device)
+        def dummy_tensors(v):
```
DarkLight1337: I don't think this change is necessary

lucianommartins: I don't think either... but the `ruff` and `mypy` hates it (`lambda` functions usage) and keeps biting me when I push something with lambdas :/

vllm/model_executor/models/gemma3_mm.py
lucianommartins: yes, I had forgotten one. fixed now.

vllm/model_executor/models/gemma3.py
lucianommartins: it was just a style change, but you're right - not necessary. This change was undone.

vllm/model_executor/model_loader/utils.py
lucianommartins: yes, you're right. now it is failing gracefully without `try`/`except` loops.

vllm/model_executor/model_loader/gguf_loader.py
lucianommartins: done. now it is just `for layer_idx in range(num_layers):`.

vllm/config/model.py
lucianommartins: fixed.

vllm/config/model.py
lucianommartins: good idea... applied it. fixed.

vllm/model_executor/model_loader/utils.py
lucianommartins: done.

vllm/model_executor/model_loader/utils.py
lucianommartins: fixed - it is all happening on the model implementation only.

lucianommartins:
thank you so much for all the help, @DarkLight1337 and @Isotr0py!

the PR is ready - we just need to take a final decision on the architecture related points (`model_arch` and vision tower loading items to be addressed in the next PRs after this one is merged).

Thanks!!

vllm/model_executor/model_loader/utils.py
```diff
@@ -42,6 +46,38 @@ def initialize_model(
     if model_config is None:
         model_config = vllm_config.model_config
     if model_class is None:
+        # Detect Gemma3 GGUF multimodal models by checking for mmproj.gguf
```
DarkLight1337: Can we reuse the code from model config?

vllm/model_executor/model_loader/utils.py
```diff
@@ -286,3 +322,71 @@ def configure_quant_config(
             quant_config.apply_vllm_mapper(hf_to_vllm_mapper)
         if packed_mapping is not None:
             quant_config.packed_modules_mapping = packed_mapping
+
+
+def extract_vision_config_from_gguf(mmproj_path: str) -> Optional["SiglipVisionConfig"]:
```
DarkLight1337: Same but in the reverse direction

vllm/model_executor/models/siglip.py
```diff
@@ -763,12 +763,13 @@ def forward(
             inputs_embeds=hidden_states,
             return_all_hidden_states=select_layers is not None,
         )
-
         if self.post_layernorm is not None:
             encoder_outputs = self.post_layernorm(encoder_outputs)
 
-        if self.use_head:
-            encoder_outputs = self.head(encoder_outputs)
+        # TODO: add this back when pooled_output is used in inference.
```
DarkLight1337: Why does this need to be changed? You can just pass `use_head=False`

vllm/model_executor/model_loader/utils.py
lucianommartins: fixed.

vllm/model_executor/models/siglip.py
lucianommartins: good catch - fixed!

vllm/model_executor/model_loader/utils.py
lucianommartins: fixed.

vllm/model_executor/model_loader/utils.py
```diff
@@ -42,6 +46,24 @@ def initialize_model(
...
+        is_multimodal_gguf = False
+        if not prefix:
+            # Reuse detection logic from model_config
+            from vllm.config.model import _detect_gguf_multimodal_gemma3
+
+            mmproj_path = _detect_gguf_multimodal_gemma3(model_config.model)
+            is_multimodal_gguf = mmproj_path is not None
+
+        # Ensure multimodal_config is set for all multimodal GGUF models
+        if is_multimodal_gguf and model_config.multimodal_config is None:
```
DarkLight1337: Why is this needed? vLLM should already set it if the model is multimodal

vllm/model_executor/models/gemma3_mm.py
```diff
@@ -592,19 +607,51 @@ def _process_image_input(
...
 
-    def get_multimodal_embeddings(self, **kwargs: object) -> MultiModalEmbeddings:
+    def get_multimodal_embeddings(self, **kwargs: object) -> list[torch.Tensor] | None:
         image_input = self._parse_and_validate_image_input(**kwargs)
         if image_input is None:
             return []
-
         return self._process_image_input(image_input)
 
+    def get_input_embeddings(
```
DarkLight1337: See https://github.com/vllm-project/vllm/blob/main/vllm%2Fmodel_executor%2Fmodels%2Finterfaces.py#L153, you can simply set the default value of the OOV flag to handle this

vllm/transformers_utils/config.py
Isotr0py: > My suggesation: This PR is to introduce a critical missing part for Gemma3 developers - multimodal experiences with quantized models.

Yes, I agree that this PR should focus on loading GGUF mm proj part properly, and leave PrefixLM attention support for another PR, which means that we can leave gemma3 attention implementation as it looks on main branch.

BTW, I think we shouldn't reuse the legacy v0 Gemma3 custom attention mask code path on model runner, especially we can implement in native attention backend now after we have supported flex attention.

lucianommartins: ok, deal. that's gonna be my next PR, @Isotr0py:
- implement FlexAttention usage
- make sure no V0 code paths will remain

vllm/model_executor/model_loader/utils.py
lucianommartins: done, now I'm using the model multimodal checking.

vllm/model_executor/models/gemma3_mm.py
lucianommartins: done! now I'm using OOV flag to handle that.

vllm/model_executor/models/gemma3_mm.py
```diff
@@ -592,19 +606,49 @@ def _process_image_input(
...
         )
         image_embeds = self.multi_modal_projector(image_features)
-
         return [e.flatten(0, 1) for e in image_embeds.split(num_patches.tolist())]
 
     def get_language_model(self) -> torch.nn.Module:
         return self.language_model
 
-    def get_multimodal_embeddings(self, **kwargs: object) -> MultiModalEmbeddings:
+    def get_multimodal_embeddings(self, **kwargs: object) -> list[torch.Tensor] | None:
```
DarkLight1337: Looks like an unnecessary change

vllm/model_executor/models/siglip.py
```diff
@@ -763,7 +763,6 @@ def forward(
             inputs_embeds=hidden_states,
             return_all_hidden_states=select_layers is not None,
         )
```
DarkLight1337, line 758: Can revert this file since the changes are just cosmetic

vllm/model_executor/models/gemma3_mm.py
lucianommartins: I've done that on one of my commits because the `mkdocs` test failed complaining about this type hint. But I'm reverting it now.

vllm/model_executor/models/siglip.py
lucianommartins: done.

DarkLight1337 (commented):
Otherwise I am ok with this PR

vllm/transformers_utils/config.py
Isotr0py: I have a draft PR to support gemma3's bidirectional attention mask on FlexAttention backend (https://github.com/vllm-project/vllm/pull/27938). We can collaborate on that PR. :)

vllm/config/model.py
```diff
@@ -65,6 +67,105 @@
 
 logger = init_logger(__name__)
 
+
+def _detect_gguf_multimodal_gemma3(model: str) -> Path | None:
```
Isotr0py: Can we move these functions to a new file `transformers_utils/gguf_utils.py`? Directly implement them here is a bit weird and ugly, especially some of them are gemma3-specific.

vllm/config/model.py
```diff
@@ -549,6 +650,44 @@ def __post_init__(
...
+                            "This may indicate a corrupted or incomplete GGUF "
+                            "file. Please verify your mmproj.gguf file is valid."
+                        )
+
+                    self.hf_config.vision_config = vision_config
+                    self.hf_config.mm_tokens_per_image = 256
+                    self.hf_config.image_token_index = 262144
+                    # DO NOT set boi_token_index - let
+                    # gemma3_mm.py fall back to 262143
+                    self.hf_config.eoi_token_index = 256000
```
Isotr0py: Let's implement a new function `patch_hf_config_from_gguf` for this.

Isotr0py (commented):
> are you ok with the changes to engine core (custom attention mask) being in this PR?

I'm fine to the custom attention mask usage now, let's leave gemma3 native attention backend support in another PR then.

However, we should take care of the GGUF -> HF config mapping, especially any usage of manual field mapping (I know some of them are unavoidable), otherwise it can significantly increase the effort to support other MM GGUF models like Qwen3-VL...

vllm/config/model.py
lucianommartins: done!

vllm/config/model.py
lucianommartins: done!

vllm/model_executor/models/gemma3_mm.py
Isotr0py, line 1: Can you add a GGUF generation test at `tests/models/multimodal/generation`? So that we can validate it out of box.

lucianommartins: done, @Isotr0py!

I added a new `tests/models/multimodal/generation/test_gemma3_gguf.py` where I'm testing Gemma3 GGUF functionalities.

tests/models/multimodal/generation/test_gemma3_gguf.py
```diff
@@ -0,0 +1,140 @@
...
+        hf_outputs = hf_model.generate_greedy_logprobs(
+            example_prompts[:1], MAX_TOKENS, NUM_LOGPROBS
+        )
+
+    check_logprobs_close(
+        outputs_0_lst=hf_outputs,
+        outputs_1_lst=gguf_outputs,
+        name_0="hf",
+        name_1="gguf",
+    )
```
Isotr0py:
```suggestion
```
No need to run text-only generation test, we only care about multimodal generation here.

tests/models/multimodal/generation/test_gemma3_gguf.py
```diff
@@ -0,0 +1,140 @@
...
+    hf_hub_download(GEMMA3_4B_GGUF_REPO, "mmproj-model-f16-4B.gguf")
+
+    stop_sign = image_assets[0]
+    prompt = "<start_of_image>What's the content in the center of the image?"
+
+    with vllm_runner(
+        gguf_path,
+        tokenizer_name=GEMMA3_4B_HF,
+        max_model_len=4096,
+        enforce_eager=True,
```
Isotr0py:
```suggestion
        enforce_eager=True,
        mm_processor_kwargs={"do_pan_and_scan": True},
```

tests/models/multimodal/generation/test_gemma3_gguf.py
lucianommartins: Hi @Isotr0py I did it differently. I noticed that there was no Gemma3 related tests at all. So I split this file in the following tests:

```
	new file:   tests/models/language/generation/test_gemma3_gguf.py
	new file:   tests/models/language/generation/test_gemma3_hf.py
	modified:   tests/models/multimodal/generation/test_gemma3_gguf.py
	new file:   tests/models/multimodal/generation/test_gemma3_hf.py
```

I split in text and multimodal as the Gemma3-1b uses `Gemma3ForCausalLM` and Gemma-3-4b and beyond uses `Gemma3ForConditionalGeneration` - so I can test both architectures.

tests/models/multimodal/generation/test_gemma3_gguf.py
lucianommartins: fixed, thanks.

lucianommartins:
Hi @Isotr0py - As we talked before, I implemented the following enhancements, I:
- Defined the `config.get_text_config()` to fix the model loading issue that was happening
- Added automated filtering to only define tensors that do exist in the GGUF file
- Applied a Gemma3 specific FFN naming correction (ffn_down->fc1, ffn_up->fc2) due to swapped naming between GGUF standard and vLLM's implementation
- Refactored multimodal weight loading to use itertools.chain for cleaner iterator composition (backbone weights + mmproj weights)

vllm/model_executor/model_loader/gguf_loader.py
```diff
@@ -115,24 +127,249 @@ def _get_gguf_weights_map(self, model_config: ModelConfig):
...
+
+        if is_multimodal:
+            mm_proj_arch = gguf.MODEL_ARCH.CLIP_VISION
+            vision_num_layers = config.vision_config.num_hidden_layers
+            vision_name_map = gguf.get_tensor_name_map(mm_proj_arch, vision_num_layers)
+        else:
+            vision_name_map = None
+
         with torch.device("meta"):
             dummy_model = AutoModelForCausalLM.from_config(
```
Isotr0py: Wait, why can't we use `AutoModelForImageTextToText` for multimodal models here? `AutoModelForCausalLM` will only initialize text backbone without vision tower and projector. So the `gguf_to_hf_name_map` won't content the vision tower and projector's `gguf -> hf` weights mapping.

I think we might have a misunderstanding about the gguf mapping workflow. I'm expecting `_get_gguf_weights_map` to output the whole model's weights mapping instead of only text backbone's, so we don't need to implement an extra `_create_mmproj_tensor_mapping` method on this way. Then we can simplify the implementation of `_get_weights_iterator` finally. 😅

lucianommartins: hey @Isotr0py - I tried using the `AutoModelForImageTextToText` but it was causing a torch.compile error during model initialization:

```
RuntimeError: Attempted to use an uninitialized parameter in `torch.compile`
```

As the mmproj GGUF file was not being loaded at that time (so no mmproj GGUF: vision tower + projector weights (`v.*`, `mm.*`), the `torch.compile` ran before those weights are loaded, detected these uninitialized vision parameters and failed.

I also tested manually loading things and checking the mapping and I noticed some peculiarities of Gemma3 that made the process to fail. I was able to convert:

- `model.vision_tower.vision_model.embeddings.patch_embedding.weight` → `v.patch_embd.weight`
- `model.vision_tower.vision_model.encoder.layers.0.self_attn.q_proj.weight` → `v.blk.0.attn_q.weight`

But it fails to convert: 

- `model.multi_modal_projector.mm_input_projection_weight` as it is very specific to Gemma3 (`_weight` instead of `.weight`).

I can go back to this track and adapt evrything to work as you are considering. Which includes:

- using of the `AutoModelForImageTextToText.from_config(config)`
- integrating FFN swap + mm_input_projection fixes into `find_hf_name_in_tensor_map()`
- reading the mmproj GGUF file during `_get_gguf_weights_map()` and filter all I need
- deleting `_create_mmproj_tensor_mapping()` at all
- getting `_get_weights_iterator()` simplified with no mmproj handling in a separated place

But still, I will be (kind of) simply carrying some complexity of the `_create_mmproj_tensor_mapping` code into another places before getting rid of that.

wdyt @Isotr0py? I can proceed with that if you think it is the best way forward.

Isotr0py: > I can proceed with that if you think it is the best way forward.

Thanks, let's go ahead along that way, because we're expecting GGUF loader as simple as possible. :)

> But still, I will be (kind of) simply carrying some complexity of the code into another places before getting rid of that._create_mmproj_tensor_mapping

No problem, I know we have to introduce some complexity because of mm_proj. I will also take a look and try my best to make it simplified and clean as much as possible. :)

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2178,6 +2181,20 @@ def _preprocess(
...
+            has_mm_features = any(
+                req_state.mm_features for req_state in self.requests.values()
+            )
+            if self.uses_custom_attention_masks and has_mm_features:
+                mask_kwargs = self.model.generate_attention_masks(
+                    self.input_ids.gpu[:num_scheduled_tokens],
+                    self.positions.gpu[:num_scheduled_tokens],
+                    mask_dtype=self.model.dtype,
+                )
+                model_kwargs.update(mask_kwargs)
```
Isotr0py, line 2367: BTW, let's split the gemma3 attention mask workaround into a separated PR.

I think we all agree this part is fine in previous discussion, and this part is not really gguf specific (It should also work for HF models). 

So we can focus on reviewing the GGUF multimodal support in this PR after splitting.

vllm/model_executor/model_loader/gguf_loader.py
lucianommartins: hey @Isotr0py!

I've done everything as we agreed:

- automatic mapping approach (no manual tensor mapping)
- use of the `AutoModelForImageTextToText` for unified parameter inspection
- integrated some extra fixes needed into my helper functions
- Gemma3-specific scoping (no impact on other models)

Tested on:

- Gemma3 1B GGUF text-only
- Gemma3 4B GGUF text-only
- Gemma3 4B GGUF multimodal

Non-Gemma GGUF models just to double check that there are no collateral damage:
- LLaVA Mistral GGUF text-only

Some HF models just to double check that there are no collateral damage:
- Gemma3 1B HuggingFace text-only
- Gemma3 4B HuggingFace multimodal

vllm/v1/worker/gpu_model_runner.py
lucianommartins: IIRC the next steps are:
- this PR will be merged with the custom attention layers as it is now
- once it is merged, I will start the next PR to implement the FlexAttention to Gemma3 models

Right?

tests/models/multimodal/generation/test_gemma3_gguf.py
```diff
@@ -0,0 +1,72 @@
...
+
+from vllm.assets.image import ImageAsset
+
+# Multimodal test prompt
+PROMPT = "Describe this image in detail:"
+
+
+def test_gemma3_4b_gguf_multimodal(vllm_runner):
+    """Test Gemma3 4B GGUF multimodal generation."""
+    # Download GGUF model
```
Isotr0py: Seems this test can't pass locally:
```
(EngineCore_DP0 pid=1418273)   File "/home/mozf/develop-projects/vllm/.venv/lib/python3.12/site-packages/transformers/models/auto/image_processing_auto.py", line 476, in from_pretrained
(EngineCore_DP0 pid=1418273)     config_dict, _ = ImageProcessingMixin.get_image_processor_dict(
(EngineCore_DP0 pid=1418273)                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=1418273)   File "/home/mozf/develop-projects/vllm/.venv/lib/python3.12/site-packages/transformers/image_processing_base.py", line 371, in get_image_processor_dict
(EngineCore_DP0 pid=1418273)     text = reader.read()
(EngineCore_DP0 pid=1418273)            ^^^^^^^^^^^^^
(EngineCore_DP0 pid=1418273)   File "<frozen codecs>", line 322, in decode
(EngineCore_DP0 pid=1418273) UnicodeDecodeError: 'utf-8' codec can't decode byte 0xbc in position 8: invalid start byte
```

vllm/v1/worker/gpu_model_runner.py
Isotr0py: > this PR will be merged with the custom attention layers as it is now

Not only GGUF uses custom attention layers, HF format gemma3-mm also uses it. So we can validate it with existing HF gemma3-MM models on CI.

I think splitting it into a separated PR and merge it now can make this PR clearer, and reduce the CIs to run here. I'm still cleaning the GGUF part and seems it's broken now.

tests/models/multimodal/generation/test_gemma3_gguf.py
lucianommartins: I fixed it. It was the lack of `tokenizer_name="google/gemma-3-4b-it",` in the test script.

vllm/model_executor/model_loader/gguf_loader.py
```diff
@@ -384,20 +322,10 @@ def _get_weights_iterator(
             # Using unified gguf_to_hf_name_map for both files - iterator naturally
             # filters to only tensors present in each file
 
-            # Helper to remap backbone weights to language_model hierarchy
-            def remap_backbone_weights(weights_iter):
```
lucianommartins: this removal broke the Gemma3 MM support as weight remmaping steps are required to have it working. ie:

- the `model.*` → `language_model.model.*` remmaping for Gemma3 multimodal is not happening anymore
- the Gemma3 prefix stripping (`model.vision_tower.*` → `vision_tower.*`) was removed

I can't see this function going anywhere else too, as it happens at weights loading time - so it won't work if it is moved to any gemma3 specific module.

do you have other plans here to incorporate those needed remappings?

lucianommartins: hey @Isotr0py - I finished grooming your cleaning commit. just two changes needed:

1. a wrong weight name:

```
-            gguf_to_hf_name_map["multi_modal_projector.mm_input_projection.weight"] = (
+            gguf_to_hf_name_map["mm.input_projection.weight"] = (
```

2. one weight missing to be remapped:

```
+            gguf_to_hf_name_map["mm.soft_emb_norm.weight"] = (
+                "multi_modal_projector.mm_soft_emb_norm.weight"
+            )
```

Also, as we talked Yesterday, I added one checking to make sure a `tokenizer` is required to be defined when loading GGUF models (on arg_utils.py):

```
+            if self.tokenizer is None:
+                    "Using a tokenizer is mandatory when loading a GGUF model. "
+                    "Please specify the tokenizer path or name using the "
+                    "--tokenizer argument."
+                )
```

ready to go? :)

vllm/model_executor/models/gemma3.py
```diff
@@ -500,7 +500,32 @@ def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
...
+                    param = params_dict[param_lookup_name]
+                except KeyError:
+                    if param_lookup_name.endswith(".weight"):
+                        qweight_name = param_lookup_name.replace(".weight", ".qweight")
+                        if qweight_name in params_dict:
+                            param = params_dict[qweight_name]
+                        else:
+                            raise
+                    else:
+                        raise
```
Isotr0py: I reverted this since it's no longer needed now.

vllm/model_executor/model_loader/gguf_loader.py
```diff
@@ -115,24 +137,239 @@ def _get_gguf_weights_map(self, model_config: ModelConfig):
...
+                if (
+                    gguf_name
+                    and "gemma" in model_config.model.lower()
+                    and ".mlp.fc" in base_name
+                ):
+                    # Swap ffn_up ↔ ffn_down to match actual tensor shapes
+                    if ".fc1" in base_name and "ffn_up" in gguf_name:
+                        gguf_name = gguf_name.replace("ffn_up", "ffn_down")
+                    elif ".fc2" in base_name and "ffn_down" in gguf_name:
+                        gguf_name = gguf_name.replace("ffn_down", "ffn_up")
```
Isotr0py: BTW, I found some gemma3's mm_proj checkpoints don't need this swapping, do we have method to keep compatability with it? See: https://huggingface.co/unsloth/gemma-3-4b-it-GGUF/tree/main?show_file_info=mmproj-F16.gguf

vllm/model_executor/model_loader/gguf_loader.py
```diff
@@ -107,6 +120,15 @@ def _get_gguf_weights_map(self, model_config: ModelConfig):
...
                 )
+        # Gemma3 special cases: The multimodal projector requires explicit
+        # mappings for the input projection and normalization layer.
+        if model_type == "gemma3":
+            gguf_to_hf_name_map["mm.input_projection.weight"] = (
+                "multi_modal_projector.mm_input_projection_weight"
+            )
+            gguf_to_hf_name_map["mm.soft_emb_norm.weight"] = (
+                "multi_modal_projector.mm_soft_emb_norm.weight"
+            )
```
Isotr0py: I wonder why auto mapping can't handle this parameter, it should have been registered in gguf-py's tensor mapping like normal weights... 🤔

https://github.com/ggml-org/llama.cpp/blob/4dca015b7e019d5bfa9d3872b19ad4cf97859c22/gguf-py/gguf/constants.py#L1014

https://github.com/ggml-org/llama.cpp/blob/4dca015b7e019d5bfa9d3872b19ad4cf97859c22/gguf-py/gguf/tensor_mapping.py#L1398-L1400

Isotr0py (commented):
This PR looks much better now. Just leave some nits. Then we can merge this after resolving it.

vllm/model_executor/model_loader/gguf_loader.py
lucianommartins: It is a todo for me, @Isotr0py. right now, the GGUF models that officially work with this PR for multimodality are the `google/gemma-3-*`qat-q4_0-gguf`. Unsloth and other community converted GGUF versions seems to have differences in their conversion process that I still need to generalize my approach.

I still haven't found one method to be flexible as you mentioned, since the model name and other GGUF metadata are just the same. One alternative is to only do the swapping if the shape is not the expected - but it still seem like a vague and too generic check for me... I'm working on that and it is on my backlog for a upcoming PR :)

Isotr0py: Fix in https://github.com/vllm-project/vllm/pull/27772/commits/90a7bd2439003a21fe0a8ad3893add1ed5912454

tests/models/multimodal/generation/test_gemma3_gguf.py
```diff
@@ -0,0 +1,72 @@
...
+        "photo",
+        "sign",
+        "red",
+        "stop",
+        "shows",
+        "depicts",
+    ]
+    assert any(word in output_lower for word in image_keywords), (
+        f"Output doesn't appear to describe the image: {output_str}"
+    )
```
Isotr0py: This test is not robust enough. Some gibberish outputs can even pass it before:
```
Adding requests: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:09<00:00,  9.64s/it]
Processed prompts: 100%|████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.15s/it, est. speed input: 124.27 toks/s, output: 59.58 toks/s]
Generated output: <start_of_image> Describe this image in detail:

This image is a vibrant and intricate mandala design, likely created digitally. It's a complex geometric pattern centered around a large, stylized lotus flower. Here's a breakdown of the key elements and characteristics:

**Central Motif: Lotus Flower**

*   **Size & Position:** The lotus flower dominates the center of the image. It's substantial, taking up a significant portion of the visual space.
*   **Style:** The lotus is rendered in a highly stylized, almost abstract manner. It doesn't aim for photorealism; instead, it's composed of geometric shapes and lines.
*   
PASSED
```

Let's compare the logprobs with the original HF model.

vllm/model_executor/model_loader/gguf_loader.py
Isotr0py: This issue won't block this PR. But we should revisit it after merging.

Isotr0py (approved):
LGTM now. This PR should be ready to merge after updating the test. Thanks for your patience! :)

Isotr0py:
/gemini review (in case we missed something)

tests/models/multimodal/generation/test_common.py
```diff
@@ -221,7 +221,7 @@
         vllm_output_post_proc=model_utils.gemma3_vllm_to_hf_output,
         image_size_factors=[(0.25, 0.5, 1.0)],
         vllm_runner_kwargs={
-            "model_impl": "transformers",
```
Isotr0py, line 224: Please revert this. We should run it with transformers backend.

vllm/transformers_utils/gguf_utils.py
```diff
@@ -0,0 +1,228 @@
...
+            text_config = hf_config.get_text_config()
+            original_vocab_size = text_config.vocab_size
+
+            if original_vocab_size != vocab_size_from_gguf:
+                logger.info(
+                    "Overriding vocab_size: %d (HF config) → %d (GGUF file)",
+                    original_vocab_size,
+                    vocab_size_from_gguf,
+                )
+                text_config.vocab_size = vocab_size_from_gguf
```
Isotr0py: Do we really need this? I think this is redundant because transformers has had corresponding override for this case:
https://github.com/huggingface/transformers/blob/66d57110f089789ae285cc9d54d3bf051123246b/src/transformers/modeling_gguf_pytorch_utils.py#L487-L497

lucianommartins: done. reverted.

tests/models/multimodal/generation/test_common.py
lucianommartins: done.

lucianommartins:
Hey @Isotr0py - all tests are passing, with 2 exceptions not related to Gemma3 or the PR work:

1. `buildkite/ci/pr/entrypoints-integration-test-api-server`

It is failing with this access error for the test image to be used:

```
[2025-11-17T20:38:37Z] FAILED entrypoints/openai/test_metrics.py::test_metrics_exist[-multimodal] - openai.InternalServerError: Error code: 500 - {'error': {'message': "403, message='Forbidden', url='https://upload.wikimedia.org/wikipedia/commons/thumb/d/dd/Gfp-wisconsin-madison-the-nature-boardwalk.jpg/2560px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg'", 'type': 'Internal Server Error', 'param': None, 'code': 500}}
```

2. `buildkite/ci/pr/examples-test`

It is failing for EAGLE-LLAMA3.1-Instruct-8B (before even getting to any Gemma3 test):

```
[2025-11-17T20:31:00Z] + python3 offline_inference/spec_decode.py --test --method eagle --num_spec_tokens 3 --dataset-name hf --dataset-path philschmid/mt-bench --num-prompts 80 --temp 0 --top-p 1.0 --top-k -1 --tp 1 --enable-chunked-prefill --max-model-len 2048
[2025-11-17T20:31:07Z] [2025-11-17 12:31:07] INFO config.py:54: PyTorch version 2.9.0+cu129 available.
[2025-11-17T20:31:07Z] [2025-11-17 12:31:07] INFO config.py:66: Polars version 1.29.0 available.
[2025-11-17T20:31:10Z] INFO 11-17 12:31:10 [utils.py:253] non-default args: {'trust_remote_code': True, 'max_model_len': 2048, 'gpu_memory_utilization': 0.8, 'limit_mm_per_prompt': {'image': 5}, 'enable_chunked_prefill': True, 'disable_chunked_mm_input': True, 'speculative_config': {'method': 'eagle', 'model': 'yuhuili/EAGLE-LLaMA3.1-Instruct-8B', 'num_speculative_tokens': 3}, 'model': 'meta-llama/Llama-3.1-8B-Instruct'}
```

And it is failing due to GPU OOM errors:

```
[2025-11-17T20:33:01Z] [1;36m(EngineCore_DP0 pid=4274)[0;0m INFO 11-17 12:33:01 [backends.py:631] Using cache directory: /root/.cache/vllm/torch_compile_cache/bb408db9f5/rank_0_0/eagle_head for vLLM's torch.compile
[2025-11-17T20:33:01Z] [1;36m(EngineCore_DP0 pid=4274)[0;0m INFO 11-17 12:33:01 [backends.py:647] Dynamo bytecode transform time: 0.39 s
[2025-11-17T20:33:06Z] [1;36m(EngineCore_DP0 pid=4274)[0;0m INFO 11-17 12:33:06 [backends.py:282] Compiling a graph for dynamic shape takes 4.61 s
[2025-11-17T20:33:06Z] [1;36m(EngineCore_DP0 pid=4274)[0;0m INFO 11-17 12:33:06 [monitor.py:34] torch.compile takes 17.58 s in total
[2025-11-17T20:33:13Z] [1;36m(EngineCore_DP0 pid=4274)[0;0m INFO 11-17 12:33:13 [gpu_worker.py:361] Available KV cache memory: -0.26 GiB
[2025-11-17T20:33:14Z] [1;36m(EngineCore_DP0 pid=4274)[0;0m ERROR 11-17 12:33:14 [core.py:855] EngineCore failed to start.

<...>

[2025-11-17T20:33:14Z] [1;36m(EngineCore_DP0 pid=4274)[0;0m   File "/usr/local/lib/python3.12/dist-packages/vllm/v1/core/kv_cache_utils.py", line 686, in check_enough_kv_cache_memory
[2025-11-17T20:33:14Z] [1;36m(EngineCore_DP0 pid=4274)[0;0m     raise ValueError(
[2025-11-17T20:33:14Z] [1;36m(EngineCore_DP0 pid=4274)[0;0m ValueError: No available memory for the cache blocks. Try increasing `gpu_memory_utilization` when initializing the engine.
```

can we proceed with the merge?

requirements/common.txt
```diff
@@ -30,8 +30,8 @@ filelock >= 3.16.1 # need to contain https://github.com/tox-dev/filelock/pull/31
 partial-json-parser # used for parsing partial JSON outputs
 pyzmq >= 25.0.0
 msgspec
-gguf >= 0.13.0
-mistral_common[image] >= 1.8.5
+gguf >= 0.17.0
+mistral_common[image,audio] >= 1.8.5
```
DarkLight1337: Actually we need to revert this change to `mistral_common` requirements, see #28722

PR #27773, opened 2025-10-29
[Bugfix] mamba-block-size is set for vision language model

heheda12345:
## Purpose
```
vllm serve nvidia/Nemotron-Nano-12B-v2-VL-BF16 --trust-remote-code
```
get this error
```
(EngineCore_DP0 pid=861647)   Value error, --mamba-block-size can only be set with --enable-prefix-caching [type=value_error, input_value=CacheConfig(block_size=67...cache_memory_bytes=None), input_type=CacheConfig]
```

## Test Plan
try this model
## Test Result
can run now
---
<details>

</details>

### DIFF
--- a/vllm/config/vllm.py
+++ b/vllm/config/vllm.py
@@ -17,7 +17,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, TypeVar
 
 import torch
-from pydantic import ConfigDict, Field
+from pydantic import ConfigDict, Field, model_validator
 from pydantic.dataclasses import dataclass
 
 import vllm.envs as envs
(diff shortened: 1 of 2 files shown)


vllm/config/vllm.py
```diff
@@ -943,6 +943,20 @@ def __str__(self):
...
+        if self.model_config is None:
+            return self
+        mamba_block_size_is_set = (
+            self.cache_config.mamba_block_size is not None
+            and self.cache_config.mamba_block_size != self.model_config.max_model_len
+        )
+        if mamba_block_size_is_set and not self.cache_config.enable_prefix_caching:
+            raise ValueError(
+                "--mamba-block-size can only be set with --enable-prefix-caching"
+            )
```
ywang96, line 957: Nit: This fix works but it seems that the real underlying issue is that not all hybrid models support prefix caching, is that the right understanding?

I think it's probably a better idea to later change this to something like
```python
        supports_prefix_caching = ...        
        if mamba_block_size_is_set and not (self.cache_config.enable_prefix_caching and supports_prefix_caching):
            raise ValueError(
                "--mamba-block-size can only be set with --enable-prefix-caching for models that support prefix caching."
            )
```

ywang96 (approved):
Thanks for fixing this! I left a comment.

vllm/config/vllm.py
heheda12345: We have other logic to check whether a hybrid model supports prefix cache. The concern here is that mamba_block_size will hurt the performance without prefix cache.

PR #27718, opened 2025-10-29
Feature/kv cache average lifetime

alhridoy:
#25725

   •  Add an allocation_time timestamp to every KVCacheBlock and switch lifetime
      calculations to time.monotonic() so clock adjustments can’t produce
      negative values.
   •  Introduce KVCacheLifetimeStats to aggregate per-block lifetimes and emit
      fresh samples each scheduler iteration.
   •  Update BlockPool, KVCacheManager, and the scheduler pipeline so lifetimes
      are recorded on allocation/free and flow through to SchedulerStats.
   •  Replace the original average-only gauge with a Prometheus histogram
      vllm:kv_cache_block_lifetime_seconds, letting operators compute
      averages/percentiles via PromQL while retaining aggregate counters.
   •  Provide comprehensive coverage: moved the demo under tests/v1/core, added
      unit tests for the stats class, block pool integration, Prometheus
      plumbing, and a runnable example.
   •  Ensure resets and prefix-cache clearing also clear lifetime stats, keeping
      behavior consistent with upstream expectations.

PR #5602, opened 2024-06-17
[Core] Optimize block_manager_v2 vs block_manager_v1 (to make V2 default) 

alexm-redhat:
This PR optimizes block_manager_v2 python logic to make it comparable to block_manager_v1. The goal is to enable block_manager_v2 by default as part of the spec decode project.

The issues optimized are:
1. Python Block object allocations/deallocations are expensive on the hot-path of iterative batching, so a block pool is used to cache block objects.
2. Any string/list duplication should be avoided, especially for token id lists
3. Modified Prefix Caching Block/Allocator to avoid any full traversals of block_ids by using dynamic/incremental style computations 
4. Redid the way access all blocks updates timestamps by deferring the actual updates to free(..) of sequences

Here is initial performance comparison for both standard and prefix-cache enabled runs:

image

PR #27793, opened 2025-10-30
[Core][Observability] Add KV cache residency metrics

shivampr:
## Purpose

This PR introduces **three new Prometheus histograms** for fine-grained observability of KV cache residency behavior:

* `vllm:kv_block_lifetime_seconds` — total lifetime from allocation to free
* `vllm:kv_block_idle_before_evict_seconds` — idle duration before eviction
* `vllm:kv_block_reuse_gap_seconds` — time between consecutive reuses of the same block

These metrics help operators analyze KV cache efficiency, reuse patterns, and eviction timing beyond simple utilization rates.
Implementation uses **monotonic timestamps** for accuracy, **1% sampling** for minimal overhead (~48 bytes/block), and is fully thread-safe with **zero runtime cost when disabled**.

Two new runtime flags are introduced:

* `--kv-cache-metrics` – enable KV cache residency metrics
* `--kv-cache-metrics-sample` – control sampling ratio (default: 0.01)

---

## Test Plan

* **Unit tests**  
  * `pytest -v tests/v1/core/test_kv_cache_metrics.py`  
    * Deterministic fake clock for lifetime/idle/prefix calculations  
    * Sampling and concurrency smoke coverage  

---

* **Integration smoke (CPU)**  
  * `python -m vllm.entrypoints.openai.api_server --model gpt2 --host 127.0.0.1 --port 8000 --kv-cache-metrics --kv-cache-metrics-sample 1.0`  
  * Sent two `/v1/completions` requests  
  * `curl http://127.0.0.1:8000/metrics | grep 'vllm:kv_block_'` → lifetime/idle/reuse histograms present  

---

* **Manual LLM snapshot (CPU)**  
  * `LLM(model="gpt2", kv_cache_metrics=True, kv_cache_metrics_sample=1.0, disable_log_stats=False)`  
  * Generated a short prompt, ran `llm.llm_engine.do_log_stats()`  
  * Checked `get_prometheus_registry().collect()` — all four KV histograms registered  

---

## Test Result

```
pytest -v tests/v1/core/test_kv_cache_metrics.py
```
(passed locally)

---

`/metrics` excerpt after the CPU smoke run:
```
# TYPE vllm:kv_block_lifetime_seconds histogram
# TYPE vllm:kv_block_idle_before_evict_seconds histogram
# TYPE vllm:kv_block_reuse_gap_seconds histogram
```

---

PromQL sanity check (validated syntactically):
```promql
rate(vllm:kv_block_lifetime_seconds_sum[5m])
/
rate(vllm:kv_block_lifetime_seconds_count[5m])
```

Fixes #25725

### DIFF
--- a/docs/design/metrics.md
+++ b/docs/design/metrics.md
@@ -280,15 +280,12 @@ Those map directly to the Prometheus metrics:
 - `vllm_kv_block_lifetime_seconds` – how long each sampled block exists.
 - `vllm_kv_block_idle_before_evict_seconds` – idle tail after the final access.
 - `vllm_kv_block_reuse_gap_seconds` – time between consecutive touches.
-- `vllm_request_prefix_residency_seconds` – how long prompt blocks stay
-  resident after prefill.
 
 The engine core only ships raw eviction events via `SchedulerStats`; the
 frontend drains them, turns them into Prometheus observations, and also
 exposes the same data through `LLM.get_metrics()` when logging is on.
-Looking at lifetime, idle, and prefix residency on one chart makes it
-easy to spot stranded cache or workloads that pin prompts for a long
-decode.
+Looking at lifetime and idle time on one chart makes it easy to spot
+stranded cache or workloads that pin prompts for a long decode.
 
 ### Metrics Publishing - Logging
 
--- a/docs/design/metrics.md
+++ b/docs/design/metrics.md
@@ -263,6 +263,33 @@ record:
 - End-to-end latency - the interval between frontend `arrival_time`
   and the frontend receiving the final token.
 
+### KV Cache Residency Metrics
+
+We also emit a set of histograms that describe how long sampled KV cache
+blocks stay resident and how often they are reused. Sampling
+(`--kv-cache-metrics-sample`) keeps the overhead tiny; when a block is
+chosen we record:
+
+- `lifetime` – allocation ⟶ eviction
+- `idle before eviction` – last touch ⟶ eviction
+- `reuse gaps` – the pauses between touches when the block gets reused
+- `prefix residency` – prefill completes ⟶ eviction
+
+Those map directly to the Prometheus metrics:
+
+- `vllm_kv_block_lifetime_seconds` – how long each sampled block exists.
+- `vllm_kv_block_idle_before_evict_seconds` – idle tail after the final access.
+- `vllm_kv_block_reuse_gap_seconds` – time between consecutive touches.
+- `vllm_request_prefix_residency_seconds` – how long prompt blocks stay
+  resident after prefill.
+
+The engine core only ships raw eviction events via `SchedulerStats`; the
+frontend drains them, turns them into Prometheus observations, and also
+exposes the same data through `LLM.get_metrics()` when logging is on.
+Looking at lifetime, idle, and prefix residency on one chart makes it
+easy to spot stranded cache or workloads that pin prompts for a long
+decode.
+
 ### Metrics Publishing - Logging
 
 The `LoggingStatLogger` metrics publisher outputs a log `INFO` message
--- a/docs/design/metrics.md
+++ b/docs/design/metrics.md
@@ -273,13 +273,12 @@ chosen we record:
 - `lifetime` – allocation ⟶ eviction
 - `idle before eviction` – last touch ⟶ eviction
 - `reuse gaps` – the pauses between touches when the block gets reused
-- `prefix residency` – prefill completes ⟶ eviction
 
 Those map directly to the Prometheus metrics:
 
-- `vllm_kv_block_lifetime_seconds` – how long each sampled block exists.
-- `vllm_kv_block_idle_before_evict_seconds` – idle tail after the final access.
-- `vllm_kv_block_reuse_gap_seconds` – time between consecutive touches.
+- `vllm:kv_block_lifetime_seconds` – how long each sampled block exists.
+- `vllm:kv_block_idle_before_evict_seconds` – idle tail after the final access.
+- `vllm:kv_block_reuse_gap_seconds` – time between consecutive touches.
 
 The engine core only ships raw eviction events via `SchedulerStats`; the
 frontend drains them, turns them into Prometheus observations, and also
--- a/vllm/config/cache.py
+++ b/vllm/config/cache.py
@@ -144,7 +144,7 @@ class CacheConfig:
 
     kv_offloading_backend: KVOffloadingBackend | None = None
     """The backend to use for KV cache offloading. Supported backends include
-    'native' (vLLM native CPU offloading), 'lmcache' This option must be used
+    'native' (vLLM native CPU offloading), 'lmcache' This option must be used 
     together with kv_offloading_size."""
 
     def compute_hash(self) -> str:
--- a/vllm/config/cache.py
+++ b/vllm/config/cache.py
@@ -144,7 +144,7 @@ class CacheConfig:
 
     kv_offloading_backend: KVOffloadingBackend | None = None
     """The backend to use for KV cache offloading. Supported backends include
-    'native' (vLLM native CPU offloading), 'lmcache' This option must be used 
+    'native' (vLLM native CPU offloading), 'lmcache' This option must be used
     together with kv_offloading_size."""
 
     def compute_hash(self) -> str:
--- a/vllm/config/observability.py
+++ b/vllm/config/observability.py
@@ -49,10 +49,11 @@ class ObservabilityConfig:
 
     kv_cache_metrics: bool = False
     """Enable KV cache residency metrics (lifetime, idle time, reuse gaps).
-    Uses sampling to minimize overhead."""
+    Uses sampling to minimize overhead.
+    Requires log stats to be enabled (i.e., --disable-log-stats not set)."""
 
-    kv_cache_metrics_sample: float = Field(default=0.01, ge=0, le=1)
-    """Sampling rate for KV cache metrics [0.0, 1.0]. Default 0.01 = 1% of blocks."""
+    kv_cache_metrics_sample: float = Field(default=0.01, gt=0, le=1)
+    """Sampling rate for KV cache metrics (0.0, 1.0]. Default 0.01 = 1% of blocks."""
 
     @cached_property
     def collect_model_forward_time(self) -> bool:
--- a/vllm/config/observability.py
+++ b/vllm/config/observability.py
@@ -5,7 +5,7 @@ from functools import cached_property
 from typing import Any, Literal, cast
 
 from packaging.version import parse
-from pydantic import field_validator, model_validator
+from pydantic import Field, field_validator, model_validator
 from pydantic.dataclasses import dataclass
 
 from vllm import version
@@ -46,6 +46,13 @@ class ObservabilityConfig:
 
     Note that collecting detailed timing information for each request can be
     expensive."""
+
+    kv_cache_metrics: bool = False
+    """Enable KV cache residency metrics (lifetime, idle time, reuse gaps).
+    Uses sampling to minimize overhead."""
+
+    kv_cache_metrics_sample: float = Field(default=0.01, ge=0, le=1)
+    """Sampling rate for KV cache metrics [0.0, 1.0]. Default 0.01 = 1% of blocks."""
 
     @cached_property
     def collect_model_forward_time(self) -> bool:
--- a/vllm/v1/core/block_pool.py
+++ b/vllm/v1/core/block_pool.py
@@ -177,7 +177,6 @@ class BlockPool:
         self.enable_kv_cache_events = enable_kv_cache_events
         self.kv_event_queue: list[KVCacheEvent] = []
 
-        # KV cache metrics collector
         self.metrics_collector = metrics_collector
 
     def get_cached_block(
@@ -335,6 +334,10 @@ class BlockPool:
         Returns:
             True if the block is evicted, False otherwise.
         """
+        # Clean up metrics tracking first to prevent leaks
+        if self.metrics_collector:
+            self.metrics_collector.on_block_evicted(block)
+
         block_hash = block.block_hash
         if block_hash is None:
             # The block doesn't have hash, eviction is not needed
@@ -346,9 +349,6 @@ class BlockPool:
             return False
 
         block.reset_hash()
-
-        if self.metrics_collector:
-            self.metrics_collector.on_block_evicted(block)
 
         if self.enable_kv_cache_events:
             # FIXME (Chen): Not sure whether we should return `hash_value`
--- a/vllm/v1/core/block_pool.py
+++ b/vllm/v1/core/block_pool.py
@@ -11,6 +11,7 @@ from vllm.distributed.kv_events import (
     KVCacheEvent,
 )
 from vllm.logger import init_logger
+from vllm.v1.core.kv_cache_metrics import KVCacheMetricsCollector
 from vllm.v1.core.kv_cache_utils import (
     BlockHash,
     BlockHashList,
@@ -140,6 +141,7 @@ class BlockPool:
             where different KV cache groups have different block sizes, the
             actual block size can be a multiple of hash_block_size.
         enable_kv_cache_events: Whether to enable kv cache events.
+        metrics_collector: Optional metrics collector for tracking block residency.
     """
 
     def __init__(
@@ -148,6 +150,7 @@ class BlockPool:
         enable_caching: bool,
         hash_block_size: int,
         enable_kv_cache_events: bool = False,
+        metrics_collector: KVCacheMetricsCollector | None = None,
     ):
         assert isinstance(num_gpu_blocks, int) and num_gpu_blocks > 0
         self.num_gpu_blocks = num_gpu_blocks
@@ -173,6 +176,9 @@ class BlockPool:
 
         self.enable_kv_cache_events = enable_kv_cache_events
         self.kv_event_queue: list[KVCacheEvent] = []
+
+        # KV cache metrics collector
+        self.metrics_collector = metrics_collector
 
     def get_cached_block(
         self, block_hash: BlockHash, kv_cache_group_ids: list[int]
@@ -308,10 +314,14 @@ class BlockPool:
                 self._maybe_evict_cached_block(block)
                 assert block.ref_cnt == 0
                 block.ref_cnt += 1
+                if self.metrics_collector:
+                    self.metrics_collector.on_block_allocated(block)
         else:
             for block in ret:
                 assert block.ref_cnt == 0
                 block.ref_cnt += 1
+                if self.metrics_collector:
+                    self.metrics_collector.on_block_allocated(block)
         return ret
 
     def _maybe_evict_cached_block(self, block: KVCacheBlock) -> bool:
@@ -336,6 +346,9 @@ class BlockPool:
             return False
 
         block.reset_hash()
+
+        if self.metrics_collector:
+            self.metrics_collector.on_block_evicted(block)
 
         if self.enable_kv_cache_events:
             # FIXME (Chen): Not sure whether we should return `hash_value`
@@ -365,6 +378,8 @@ class BlockPool:
                 if block.ref_cnt == 0 and not block.is_null:
                     self.free_block_queue.remove(block)
                 block.ref_cnt += 1
+                if self.metrics_collector:
+                    self.metrics_collector.on_block_accessed(block)
 
     def free_blocks(self, ordered_blocks: Iterable[KVCacheBlock]) -> None:
         """Free a list of blocks. The blocks should be ordered by their
@@ -406,6 +421,9 @@ class BlockPool:
         # Remove all hashes from all blocks.
         for block in self.blocks:
             block.reset_hash()
+
+        if self.metrics_collector:
+            self.metrics_collector.reset()
 
         logger.info("Successfully reset prefix cache")
 
--- a/vllm/v1/core/kv_cache_coordinator.py
+++ b/vllm/v1/core/kv_cache_coordinator.py
@@ -51,7 +51,7 @@ class KVCacheCoordinator(ABC):
             enable_caching,
             hash_block_size,
             enable_kv_cache_events,
-            metrics_collector=metrics_collector,
+            metrics_collector,
         )
 
         # Needs special handling for find_longest_cache_hit if eagle is enabled
--- a/vllm/v1/core/kv_cache_coordinator.py
+++ b/vllm/v1/core/kv_cache_coordinator.py
@@ -5,6 +5,7 @@ from collections.abc import Sequence
 from math import lcm
 
 from vllm.v1.core.block_pool import BlockPool
+from vllm.v1.core.kv_cache_metrics import KVCacheMetricsCollector
 from vllm.v1.core.kv_cache_utils import (
     BlockHash,
     BlockHashList,
@@ -39,6 +40,7 @@ class KVCacheCoordinator(ABC):
         dcp_world_size: int,
         pcp_world_size: int,
         hash_block_size: int,
+        metrics_collector: KVCacheMetricsCollector | None = None,
     ):
         self.kv_cache_config = kv_cache_config
         self.max_model_len = max_model_len
@@ -49,6 +51,7 @@ class KVCacheCoordinator(ABC):
             enable_caching,
             hash_block_size,
             enable_kv_cache_events,
+            metrics_collector=metrics_collector,
         )
 
         # Needs special handling for find_longest_cache_hit if eagle is enabled
@@ -228,6 +231,7 @@ class KVCacheCoordinatorNoPrefixCache(KVCacheCoordinator):
         dcp_world_size: int,
         pcp_world_size: int,
         hash_block_size: int,
+        metrics_collector: KVCacheMetricsCollector | None = None,
     ):
         super().__init__(
             kv_cache_config,
@@ -238,6 +242,7 @@ class KVCacheCoordinatorNoPrefixCache(KVCacheCoordinator):
             dcp_world_size=dcp_world_size,
             pcp_world_size=pcp_world_size,
             hash_block_size=hash_block_size,
+            metrics_collector=metrics_collector,
         )
         self.num_single_type_manager = len(self.single_type_managers)
 
@@ -272,6 +277,7 @@ class UnitaryKVCacheCoordinator(KVCacheCoordinator):
         dcp_world_size: int,
         pcp_world_size: int,
         hash_block_size: int,
+        metrics_collector: KVCacheMetricsCollector | None = None,
     ):
         super().__init__(
             kv_cache_config,
@@ -282,6 +288,7 @@ class UnitaryKVCacheCoordinator(KVCacheCoordinator):
             dcp_world_size=dcp_world_size,
             pcp_world_size=pcp_world_size,
             hash_block_size=hash_block_size,
+            metrics_collector=metrics_collector,
         )
         self.kv_cache_spec = self.kv_cache_config.kv_cache_groups[0].kv_cache_spec
         self.block_size = self.kv_cache_spec.block_size
@@ -338,6 +345,7 @@ class HybridKVCacheCoordinator(KVCacheCoordinator):
         dcp_world_size: int,
         pcp_world_size: int,
         hash_block_size: int,
+        metrics_collector: KVCacheMetricsCollector | None = None,
     ):
         super().__init__(
             kv_cache_config,
@@ -348,6 +356,7 @@ class HybridKVCacheCoordinator(KVCacheCoordinator):
             dcp_world_size=dcp_world_size,
             pcp_world_size=pcp_world_size,
             hash_block_size=hash_block_size,
+            metrics_collector=metrics_collector,
         )
         # hash_block_size: the block size used to compute block hashes.
         # The actual block size usually equals hash_block_size, but in cases where
@@ -523,6 +532,7 @@ def get_kv_cache_coordinator(
     dcp_world_size: int,
     pcp_world_size: int,
     hash_block_size: int,
+    metrics_collector: KVCacheMetricsCollector | None = None,
 ) -> KVCacheCoordinator:
     if not enable_caching:
         return KVCacheCoordinatorNoPrefixCache(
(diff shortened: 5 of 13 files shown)


vllm/config/cache.py
```diff
@@ -128,6 +128,13 @@ class CacheConfig:
     gpu_memory_utilization. Note that kv_cache_memory_bytes
     (when not-None) ignores gpu_memory_utilization"""
 
+    kv_cache_metrics: bool = False
+    """Enable KV cache residency metrics (lifetime, idle time, reuse gaps).
+    Uses sampling to minimize overhead."""
+
+    kv_cache_metrics_sample: float = Field(default=0.01, ge=0, le=1)
+    """Sampling rate for KV cache metrics [0.0, 1.0]. Default 0.01 = 1% of blocks."""
+
```
hmellor: Should these live in `ObservabilityConfig`?

shivampr: Done! Moved both fields to ObservabilityConfig(`observability.py`) and updated all references.

vllm/v1/metrics/loggers.py
```diff
@@ -895,6 +944,23 @@ def __init__(
...
+        if vllm_config.cache_config.kv_cache_metrics:
+            for engine_idx in engine_indexes:
+                collector = KVCacheMetricsCollector(
+                    enabled=True,
+                    sample_rate=vllm_config.cache_config.kv_cache_metrics_sample,
+                )
+                collector.histogram_block_lifetime = self.histogram_kv_block_lifetime[engine_idx]
+                collector.histogram_idle_before_evict = self.histogram_kv_block_idle_before_evict[engine_idx]
+                collector.histogram_reuse_gap = self.histogram_kv_block_reuse_gap[engine_idx]
+                collector.histogram_prefix_residency = self.histogram_request_prefix_residency[engine_idx]
```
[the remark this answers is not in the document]
shivampr: Thanks for the suggestion, but I'd prefer to keep vllm_request_prefix_residency_seconds as-is for semantic clarity. I will change this, if any of the vllm `contributors/members` suggest the same.

vllm/v1/metrics/loggers.py
```diff
@@ -870,6 +871,54 @@ def __init__(
...
+        
+        histogram_request_prefix_residency = self._histogram_cls(
+            name="vllm_request_prefix_residency_seconds",
+            documentation="Histogram of request prefix block residency time after prefill completion.",
+            buckets=kv_cache_residency_buckets,
+            labelnames=labelnames,
+        )
+        self.histogram_request_prefix_residency = make_per_engine(
+            histogram_request_prefix_residency, engine_indexes, model_name
+        )
```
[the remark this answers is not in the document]
shivampr: Thanks for the suggestion, but I'd prefer to keep vllm_request_prefix_residency_seconds as-is for semantic clarity. I will change this, if any of the vllm `contributors/members` suggest the same.

markmc:
This looks very clean and promising, very nice work!

Two things jump out at me before getting into details:

1.  You're not passing a `KVCacheMetricsCollector` to `get_kv_cache_coordinator()` in the `KVCacheManager` constructor
2. You're accessing `PrometheusStatLogger` state from the engine core, when it is only accessible in the frontend process - this would probably become apparent if you tried to resolve (1)

To follow the pattern of the existing architecture, you're going to have to collect these intervals in the engine core and send them back to the frontend via `EngineCoreOutputs` for processing by the `StatLogger`s

For comparison, see how we already have `EngineCoreEvent` which sends back timestamps associated with request-level events. And `SchedulerStats` has prefix cache and spec decoding stats sent from the engine.

`SchedulerStats` could think have e.g. a list of `BlockEvictionEvent`s with `lifetime`, `idle_time`, `reuse_gaps`, `residency`? You can see how in #27718 it's implemented using `SchedulerStats.kv_cache_block_lifetimes`:

```
kv_cache_block_lifetimes: list[float] = field(default_factory=list)
```
Prometheus is definitely the priority, but it might be worth also adding support to `LoggingStatLogger` so we get  (maybe a subset of) these metrics on the console too

Also, thank you for adding the unit tests. I do think we need more of an e2e too test, though - preferably both with the `LLM.get_metrics()` API and online serving via the HTTP API

shivampr:
I addressed the review by plumbing a single `KVCacheMetricsCollector` from the scheduler into `get_kv_cache_coordinator()` and having the engine emit raw `BlockResidencyEvent`s via `SchedulerStats`. Prometheus and logging observers now consume those events on the frontend.

For validation I ran:

- `pytest -v tests/v1/core/test_kv_cache_metrics.py`
- CPU smoke for `LLM.get_metrics()` using `LLM(model="gpt2", kv_cache_metrics=True, kv_cache_metrics_sample=1.0, disable_log_stats=False)` and confirmed all four histograms in `get_prometheus_registry().collect()`
- CPU API server smoke (`python -m vllm.entrypoints.openai.api_server --model gpt2 --kv-cache-metrics --kv-cache-metrics-sample 1.0`) with two `/v1/completions` calls and verified `/metrics` exposed the new histogram families

vllm/v1/core/kv_cache_metrics.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def __init__(self, enabled: bool = False, sample_rate: float = 0.01):
+        from vllm.logger import init_logger
+
+        logger = init_logger(__name__)
+
+        self.enabled = enabled
+        self.sample_rate = sample_rate
+
+        # Validate sampling rate
```
markmc: I would tend to expect config validation to be done at the `vllm.config` level

vllm/v1/core/kv_cache_metrics.py
```diff
@@ -0,0 +1,161 @@
...
+
+    Thread-safe via self._lock.
+    """
+
+    def __init__(self, enabled: bool = False, sample_rate: float = 0.01):
+        from vllm.logger import init_logger
+
+        logger = init_logger(__name__)
+
+        self.enabled = enabled
```
markmc: We never instantiate this with `enabled=False` so let's just remove it - all the branching for it is a bit distracting

vllm/v1/core/kv_cache_metrics.py
```diff
@@ -0,0 +1,161 @@
...
+            self.sample_rate = 0.0
+            if sample_rate < 0:
+                logger.warning("kv_cache_metrics_sample %f < 0, disabling", sample_rate)
+        elif sample_rate > 1.0:
+            self.sample_rate = 1.0
+            logger.warning(
+                "kv_cache_metrics_sample %f > 1.0, clamping to 1.0", sample_rate
+            )
+
+        self._lock = threading.Lock()
```
markmc: Unless I'm missing something, I don't think this needs to be thread-safe

vllm/v1/core/kv_cache_metrics.py
```diff
@@ -0,0 +1,161 @@
...
+        self.max_request_end_ns = 0
+
+    def record_access(self) -> None:
+        now_ns = time.monotonic_ns()
+        self.last_access_ns = now_ns
+        self.access_history.append(now_ns)
+
+    def get_lifetime_seconds(self, now_ns: int | None = None) -> float:
+        if now_ns is None:
+            now_ns = time.monotonic_ns()
```
markmc: Let's remove the unused `now_ns = None` support from these methods

vllm/v1/core/kv_cache_metrics.py
```diff
@@ -0,0 +1,161 @@
...
+                return
+
+            now_ns = time.monotonic_ns()
+            lifetime = metrics.get_lifetime_seconds(now_ns)
+            idle_time = metrics.get_idle_time_seconds(now_ns)
+            reuse_gaps = tuple(metrics.get_reuse_gaps_seconds())
+            prefix_residency: float | None = None
+            if metrics.max_request_end_ns > 0:
+                residency = (now_ns - metrics.max_request_end_ns) / 1e9
+                if residency >= 0:
```
markmc: This seems like a strange check - when might this be `False`?

vllm/v1/core/kv_cache_metrics.py
```diff
@@ -0,0 +1,161 @@
...
+if TYPE_CHECKING:
+    from vllm.v1.core.kv_cache_utils import KVCacheBlock
+
+from vllm.v1.metrics.stats import BlockResidencyEvent
+
+
+class BlockMetricsState:
+    """Tracks lifecycle metrics for a single KV cache block."""
+
+    __slots__ = (
```
markmc: Since we're not using this anywhere else in vLLM, this is probably a premature optimization. Interestingly, one of the few places it was used in the (now removed) v0 codebase was in #5602 for an equivalent object. However, with your sampling approach, I think the allocation overhead is even less likely to be an issue.

vllm/v1/core/kv_cache_coordinator.py
```diff
@@ -27,13 +28,17 @@ def __init__(
...
         self.kv_cache_config = kv_cache_config
         self.max_model_len = max_model_len
         self.enable_caching = enable_caching
 
         self.block_pool = BlockPool(
-            kv_cache_config.num_blocks, enable_caching, enable_kv_cache_events
+            kv_cache_config.num_blocks,
+            enable_caching,
+            enable_kv_cache_events,
+            metrics_collector=metrics_collector,
```
markmc: No need for this to be a `kwarg`

vllm/v1/core/block_pool.py
```diff
@@ -166,6 +169,9 @@ def __init__(
         self.enable_kv_cache_events = enable_kv_cache_events
         self.kv_event_queue: list[KVCacheEvent] = []
 
+        # KV cache metrics collector
```
markmc: Comment doesn't add anything - we're in the KV cache code, and the variable is "metrics collector"

vllm/v1/core/kv_cache_metrics.py
```diff
@@ -0,0 +1,161 @@
...
+            self._eviction_events.append(
+                BlockResidencyEvent(
+                    lifetime_seconds=lifetime,
+                    idle_seconds=idle_time,
+                    reuse_gaps_seconds=reuse_gaps,
+                    prefix_residency_seconds=prefix_residency,
+                )
+            )
+
+    def on_request_prefill_complete(
```
markmc: This isn't actually being called from anywhere?

vllm/v1/core/block_pool.py
```diff
@@ -317,6 +327,9 @@ def _maybe_evict_cached_block(self, block: KVCacheBlock) -> bool:
 
         block.reset_hash()
 
+        if self.metrics_collector:
+            self.metrics_collector.on_block_evicted(block)
```
markmc: Leaking `BlockMetricsState` in obscure situations would be a common bug (although I guess the leak is naturally bounded in this case, unlike with something keyed on request ID, but ...) so it would be nice if this was obviously correct ...

e.g. I think it's possible that `allocate_new_blocks()` / `get_new_blocks()` happens but then `cache_blocks()` doesn't happen, and then when the eviction happens we never get here

In other words, I think we should move these lines to the start of the method

vllm/v1/core/sched/scheduler.py
```diff
@@ -1248,19 +1262,30 @@ def make_stats(
         spec_decoding_stats: SpecDecodingStats | None = None,
         kv_connector_stats: KVConnectorStats | None = None,
     ) -> SchedulerStats | None:
-        if not self.log_stats:
+        if not (self.log_stats or self._kv_metrics_enabled):
```
markmc: KV metrics should only be enabled if log stats is also enabled ... `--disable-log-stats` turns off all metric collection and logging. I don't think we want to allow for a case where only KV metrics is enabled

i.e. please undo most of the changes to this method, and just add the KV metrics support

vllm/v1/metrics/stats.py
```diff
@@ -150,6 +150,16 @@ class MultiModalCacheStats(BaseCacheStats):
     """
 
 
+@dataclass
+class BlockResidencyEvent:
```
markmc: I think `KVCacheEvictionEvent` is a more obvious name than "block residency event"

vllm/v1/metrics/loggers.py
```diff
@@ -886,6 +886,80 @@ def __init__(
...
+            30,
+            60,
+            120,
+            300,
+            600,
+            1200,
+            1800,
+        ]
+
+        histogram_kv_block_lifetime = self._histogram_cls(
```
markmc: If KV metrics is disabled, we can avoid registering these metrics

vllm/v1/metrics/loggers.py
```diff
@@ -886,6 +886,80 @@ def __init__(
...
+            labelnames=labelnames,
+        )
+        self.histogram_kv_block_idle_before_evict = make_per_engine(
+            histogram_kv_block_idle_before_evict, engine_indexes, model_name
+        )
+
+        histogram_kv_block_reuse_gap = self._histogram_cls(
+            name="vllm_kv_block_reuse_gap_seconds",
+            documentation=(
+                "Histogram of time gaps between consecutive KV cache block accesses."
```
markmc: Is the total count of accesses also an interesting metric and worth noting here? (Just a thought)

vllm/v1/metrics/loggers.py
```diff
@@ -886,6 +886,80 @@ def __init__(
...
+            300,
+            600,
+            1200,
+            1800,
+        ]
+
+        histogram_kv_block_lifetime = self._histogram_cls(
+            name="vllm_kv_block_lifetime_seconds",
+            documentation=(
+                "Histogram of KV cache block lifetime from allocation to eviction."
```
markmc: Might be worth mentioning that these metrics are sampled?

vllm/config/observability.py
```diff
@@ -47,6 +47,13 @@ def show_hidden_metrics(self) -> bool:
     Note that collecting detailed timing information for each request can be
     expensive."""
 
+    kv_cache_metrics: bool = False
+    """Enable KV cache residency metrics (lifetime, idle time, reuse gaps).
+    Uses sampling to minimize overhead."""
```
markmc: Should note that this requires log stats to be enabled (i.e. `--disable-log-stats` not set)

vllm/config/observability.py
```diff
@@ -47,6 +47,13 @@ def show_hidden_metrics(self) -> bool:
...
+
+    kv_cache_metrics_sample: float = Field(default=0.01, ge=0, le=1)
```
markmc: Let's use `gt=0` - we don't need to support `sample=0` as another way of disabling this

vllm/v1/core/kv_cache_metrics.py
markmc: Ok, I see the `Field` validation now - let's just assert `0 < sample_rate <= 1.0`

vllm/config/cache.py
```diff
@@ -137,7 +137,7 @@ class CacheConfig:
 
     kv_offloading_backend: KVOffloadingBackend | None = None
     """The backend to use for KV cache offloading. Supported backends include
-    'native' (vLLM native CPU offloading), 'lmcache' This option must be used 
+    'native' (vLLM native CPU offloading), 'lmcache' This option must be used
```
markmc, line 147: Spurious change

vllm/v1/core/kv_cache_metrics.py
markmc: Claude agrees FWIW :laughing: 

```
  In the current v1 architecture, the KVCacheMetricsCollector is accessed from a single thread:

  1. EngineCore.run_busy_loop() - main thread
  2. → _process_engine_step() - same thread
  3. → Scheduler.schedule() / update_from_output() - same thread
  4. → KVCacheManager / BlockPool methods - same thread
  5. → KVCacheMetricsCollector methods - same thread

  The IO threads (process_input_sockets and process_output_sockets) in vllm/v1/engine/core.py:594-607 handle socket communication but do not access the metrics 
  collector.
```

vllm/v1/core/kv_cache_metrics.py
```diff
@@ -0,0 +1,161 @@
...
+        "last_access_ns",
+        "access_history",
+        "max_request_end_ns",
+    )
+
+    def __init__(self):
+        now_ns = time.monotonic_ns()
+        self.birth_time_ns = now_ns
+        self.last_access_ns = now_ns
+        self.access_history: deque[int] = deque(maxlen=4)
```
markmc, line 24: That seems very conservative, but I guess in a pathological case a simply array could lead to unbounded memory growth. It's worth adding a comment here

vllm/v1/metrics/loggers.py
markmc: Probably not, given the ring buffer. Might be worth mentioning here that we only record the most recent accesses

markmc (changes requested):
This looks very clean indeed, thank you :muscle: 

The lack of a call to `on_request_prefill_complete()` seems like the most important issue. Might be worth considering whether we really need it (request prefix residency) if it is so easily overlooked

vllm/v1/core/kv_cache_metrics.py
shivampr: Added comment!

vllm/config/cache.py
shivampr: addressed it.

vllm/config/observability.py
shivampr: Done. Updated to `gt=0`.

vllm/config/observability.py
shivampr: Done. Updated the docstring to note the dependency on log stats.

vllm/v1/metrics/loggers.py
shivampr: Done. Added sampling note to all other KV cache histogram documentation strings.

vllm/v1/metrics/loggers.py
shivampr: Done. Added note about ring buffer behavior to the documentation.

vllm/v1/metrics/loggers.py
shivampr: Done. Made histogram registration conditional on `kv_cache_metrics_enabled`.

vllm/v1/metrics/stats.py
shivampr: Changed to `KVCacheEvictionEvent `.

vllm/v1/core/sched/scheduler.py
shivampr: Reverted to original structure with `if not self.log_stats:` check. KV metrics are only collected when enabled, but the method still requires log stats to be enabled.

vllm/v1/core/block_pool.py
shivampr: Moved `on_block_evicted` call to the start of the method to ensure metrics cleanup happens even on early returns.

vllm/v1/core/kv_cache_metrics.py
shivampr: Thanks, removed it.

vllm/v1/core/block_pool.py
shivampr: Removed the comment.

vllm/v1/core/kv_cache_coordinator.py
shivampr: Changed to positional argument.

vllm/v1/core/kv_cache_metrics.py
shivampr: Removed it.

vllm/v1/core/kv_cache_metrics.py
shivampr: removed `if now_ns is None:` check

vllm/v1/core/kv_cache_metrics.py
shivampr: Added assertion.

vllm/v1/core/kv_cache_metrics.py
shivampr: Removed the instantiation.

vllm/v1/core/kv_cache_metrics.py
shivampr: Removed the `max_request_end_ns` path entirely (the prefix residency calculation and `on_request_prefill_complete`).

vllm/v1/core/kv_cache_metrics.py
shivampr: Removed the locking entirely. `KVCacheMetricsCollector` is now single-threaded, so there’s no `_lock` and no `with self._lock` guards anymore.

vllm/v1/core/kv_cache_metrics.py
```diff
@@ -0,0 +1,97 @@
...
+        lifetime = metrics.get_lifetime_seconds()
+        idle_time = metrics.get_idle_time_seconds()
+        reuse_gaps = tuple(metrics.get_reuse_gaps_seconds())
+
+        self._eviction_events.append(
+            KVCacheEvictionEvent(
+                lifetime_seconds=lifetime,
+                idle_seconds=idle_time,
+                reuse_gaps_seconds=reuse_gaps,
+                prefix_residency_seconds=None,
```
markmc: If we're not tracking this, let's remove it entirely from `KVCacheEvictionEvent`, the docs, and Prometheus

vllm/v1/core/sched/scheduler.py
```diff
@@ -1345,14 +1358,22 @@ def make_stats(
         prefix_cache_stats = self.kv_cache_manager.make_prefix_cache_stats()
         assert prefix_cache_stats is not None
         connector_prefix_cache_stats = self._make_connector_prefix_cache_stats()
+        block_events: list[KVCacheEvictionEvent] = []
+        if self._kv_metrics_enabled:
+            block_events = self.kv_cache_manager.collect_kv_cache_eviction_events()
+        spec_stats = spec_decoding_stats if self.log_stats else None
```
markmc: This method checks `log_stats` at the outset. No need to make `spec_decoding_stats` related change

vllm/v1/core/sched/scheduler.py
```diff
@@ -1345,14 +1358,22 @@ def make_stats(
...
+        connector_stats_payload = None
+        if self.log_stats and kv_connector_stats is not None:
+            connector_stats_payload = kv_connector_stats.data
```
markmc: Ditto for KV connector stats

vllm/v1/core/kv_cache_manager.py
```diff
@@ -214,6 +218,11 @@ def get_computed_blocks(self, request: Request) -> tuple[KVCacheBlocks, int]:
 
         return self.create_kv_cache_blocks(computed_blocks), num_new_computed_tokens
 
+    def collect_kv_cache_eviction_events(self) -> list[KVCacheEvictionEvent]:
+        if not self.metrics_collector:
+            return []
+        return self.metrics_collector.drain_events()
```
markmc: I don't think this method really adds anything - the scheduler can just call `drain_events()` itself

vllm/v1/core/sched/scheduler.py
```diff
@@ -69,6 +74,13 @@ def __init__(
         self.kv_events_config = vllm_config.kv_events_config
         self.parallel_config = vllm_config.parallel_config
         self.log_stats = log_stats
+        metrics_cfg = vllm_config.observability_config
+        self._kv_metrics_enabled = bool(metrics_cfg.kv_cache_metrics)
```
markmc: It's a bool already?

vllm/v1/core/sched/scheduler.py
```diff
@@ -69,6 +74,13 @@ def __init__(
...
+        self.kv_metrics_collector: KVCacheMetricsCollector | None = None
+        if self._kv_metrics_enabled:
+            self.kv_metrics_collector = KVCacheMetricsCollector(
+                sample_rate=metrics_cfg.kv_cache_metrics_sample,
+            )
```
markmc, line 81: I'd prefer simply:

```
self.observability_config = vllm_config.observability_config
...
self.kv_metrics_collector: KVCacheMetricsCollector | None = None
if self.observability_config.kv_cache_metrics:
    self.kv_metrics_collector = KVCacheMetricsCollector(
        self.observability_config.kv_cache_metrics_sample,
    )
```

and then subsequently just check `self.kv_metrics_collector is not None`

vllm/v1/core/sched/scheduler.py
```diff
@@ -1345,14 +1358,22 @@ def make_stats(
         prefix_cache_stats = self.kv_cache_manager.make_prefix_cache_stats()
         assert prefix_cache_stats is not None
         connector_prefix_cache_stats = self._make_connector_prefix_cache_stats()
+        block_events: list[KVCacheEvictionEvent] = []
+        if self._kv_metrics_enabled:
+            block_events = self.kv_cache_manager.collect_kv_cache_eviction_events()
```
markmc: Let's just:

```
eviction_events: list[KVCacheEvictionEvent] = []
if self.kv_metrics_collector is not None:
    eviction_events = self.kv_metrics_collector.drain_events()

**Review by markmc — CHANGES_REQUESTED** [review]
Thanks - my main comment is about removing the "prefix residency" metric if we're not going to calculate it

But also some cleanup suggestions in the scheduler

File: vllm/v1/core/kv_cache_metrics.py (hunk 23 above)
Comment (shivampr) [reply vllm/v1/core/kv_cache_metrics.py]: Removed prefix_residency_seconds from KVCacheEvictionEvent, updated docs/tests/scripts, and no longer emit that metric anywhere.

File: vllm/v1/core/sched/scheduler.py (hunk 24 above)
Comment (shivampr) [reply vllm/v1/core/sched/scheduler.py]: Since `make_stats` returns early when `log_stats` is false, `spec_decoding_stats` now is just pass through without extra guarding.

File: vllm/v1/core/sched/scheduler.py (hunk 25 above)
Comment (shivampr) [reply vllm/v1/core/sched/scheduler.py]: Applied the same simplification: 
```
connector_stats_payload = (
            kv_connector_stats.data if kv_connector_stats else None
        )
```

File: vllm/v1/core/sched/scheduler.py (hunk 27 above)
Comment (shivampr) [reply vllm/v1/core/sched/scheduler.py]: Dropped `bool(metrics_cfg.kv_cache_metrics)` entirely.

File: vllm/v1/core/sched/scheduler.py (hunk 28 above)
Comment (shivampr) [reply vllm/v1/core/sched/scheduler.py:81]: Done!.

File: vllm/v1/core/sched/scheduler.py (hunk 29 above)
Comment (shivampr) [reply vllm/v1/core/sched/scheduler.py]: Done!

File: vllm/v1/core/kv_cache_manager.py (hunk 26 above)
Comment (shivampr) [reply vllm/v1/core/kv_cache_manager.py]: Deleted `KVCacheManager.collect_kv_cache_eviction_events()` and have the scheduler drain events directly from the collector.

File: vllm/v1/metrics/loggers.py [hunk 30]
```diff
@@ -853,6 +856,79 @@ def __init__(
...
+                60,
+                120,
+                300,
+                600,
+                1200,
+                1800,
+            ]
+
+            histogram_kv_block_lifetime = self._histogram_cls(
+                name="vllm_kv_block_lifetime_seconds",
```
markmc: Sorry - I should have spotted this earlier. All our metrics are prefixed with `vllm:` not `vllm_`

vllm/v1/core/sched/scheduler.py
```diff
@@ -1344,14 +1356,22 @@ def make_stats(
         prefix_cache_stats = self.kv_cache_manager.make_prefix_cache_stats()
         assert prefix_cache_stats is not None
         connector_prefix_cache_stats = self._make_connector_prefix_cache_stats()
+        eviction_events: list[KVCacheEvictionEvent] = []
+        if self.kv_metrics_collector is not None:
+            eviction_events = self.kv_metrics_collector.drain_events()
+        spec_stats = spec_decoding_stats
```
markmc, line 1367: Nit: this is a bit pointless

docs/design/metrics.md
```diff
@@ -263,6 +263,30 @@ record:
...
+
+We also emit a set of histograms that describe how long sampled KV cache
+blocks stay resident and how often they are reused. Sampling
+(`--kv-cache-metrics-sample`) keeps the overhead tiny; when a block is
+chosen we record:
+
+- `lifetime` – allocation ⟶ eviction
+- `idle before eviction` – last touch ⟶ eviction
+- `reuse gaps` – the pauses between touches when the block gets reused
+- `prefix residency` – prefill completes ⟶ eviction
```
markmc: Remove

vllm/config/cache.py
```diff
@@ -144,7 +144,7 @@ class CacheConfig:
 
     kv_offloading_backend: KVOffloadingBackend | None = None
     """The backend to use for KV cache offloading. Supported backends include
-    'native' (vLLM native CPU offloading), 'lmcache' This option must be used
+    'native' (vLLM native CPU offloading), 'lmcache' This option must be used
```
markmc: Can you drop this change? (It would be annoying for someone else if it caused a merge conflict with their PR)

markmc (changes requested):
This looks really great now

Would you mind doing some benchmark sweeps so we can quantify the performance overhead?

Thanks!

vllm/v1/core/sched/scheduler.py
shivampr: Simplified the code by using a conditional expression instead of initializing to an empty list and then conditionally assigning.

docs/design/metrics.md
shivampr: Done!

vllm/config/cache.py
shivampr: I am really confused how these spurious showed up, removed them.

vllm/v1/metrics/loggers.py
shivampr: Done!

markmc (approved):
Looks great, thank you for your patience!

shivampr:
I tried pretty hard to get vLLM building and running on a CUDA GPU(both `A40` and `H200`) via [Runpod](https://console.runpod.io/deploy) for the KV cache metrics test, but couldn’t get to a stable, buildable setup. As a fallback I replicated the same checks in the CPU unit tests (tests/v1/core/test_kv_cache_metrics.py), so the metrics logic is still covered. I don’t currently know an easier way to bring up a reliable vLLM CUDA dev environment, so I left the GPU-side integration test for a follow-up.

PR #27806, opened 2025-10-30
[BugFix] fix: skip check unstreamed tool arg tokens when tool call name is present

llsj14:
## Purpose
- In most cases tool calling parsers support incremental stream responses
- However, if the tool call parser doesn't support incremental tool call stream response, it can cause a bug that omits the function name in the OpenAI chat completion API
- Therefore, the `_should_check_for_unstreamed_tool_arg_tokens` function should verify that the current delta doesn't contain a name before entering this branch, since this logic is specifically designed to handle cases where only arguments are being streamed incrementally, not complete tool calls with both name and arguments
- If tool call parser doesn't support incremental stream responses, due to the following part, it will omit function name.
https://github.com/vllm-project/vllm/blob/af826e082045e8bcd3ab2ea3129bcf91da7d58de/vllm/entrypoints/openai/serving_chat.py#L1157-L1166

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/entrypoints/openai/serving_chat.py
+++ b/vllm/entrypoints/openai/serving_chat.py
@@ -1729,7 +1729,10 @@ class OpenAIServingChat(OpenAIServing):
         """
         Check to see if we should check for unstreamed tool arguments tokens.
         This is only applicable when auto tool parsing is enabled, the delta
-        is a tool call with arguments.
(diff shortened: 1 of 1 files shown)


llsj14:
@K-Mistele 
Could you also review this PR? Although not all tool call parsers require this fix, it resolves function name omission issues in custom parsers that lack incremental streaming support.

PR #27809, opened 2025-10-30
[Model] Introduce Kimi Linear to vLLM

zhiyuan1i:
## Purpose
Introducing Kimi Linear, an advanced hybrid attention model that combines the efficiency of [Kimi Delta Attention (KDA)](https://github.com/fla-org/flash-linear-attention/tree/main/fla/ops/kda), a refined version of Gated DeltaNet, with reduced memory requirements and superior performance across short, long, and reinforcement learning contexts. By introducing an optimized gating mechanism, Kimi Linear significantly cuts down the need for large KV caches by up to 75%, enhancing hardware efficiency and boosting decoding throughput by up to $6\times$ for long-context tasks, such as those with up to 1M tokens.

Previous PR: https://github.com/vllm-project/vllm/pull/27654
## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #27811, opened 2025-10-30
[Bugfix] Fix 2 precommit issues - (mamba_block_size, kv_cache_config)

tlrmchlsmth:
Fix two precommit issues to get main green

### DIFF
--- a/vllm/model_executor/models/config.py
+++ b/vllm/model_executor/models/config.py
@@ -333,8 +333,6 @@ class HybridAttentionMambaModelConfig(VerifyAndUpdateConfig):
         if not envs.VLLM_USE_V1:
             return
 
-        # Save the user input before it gets modified by MambaModelConfig
-        mamba_block_size = vllm_config.cache_config.mamba_block_size
         # Enable FULL_AND_PIECEWISE by default
         MambaModelConfig.verify_and_update_config(vllm_config)
 
--- a/vllm/model_executor/models/config.py
+++ b/vllm/model_executor/models/config.py
@@ -333,6 +333,8 @@ class HybridAttentionMambaModelConfig(VerifyAndUpdateConfig):
         if not envs.VLLM_USE_V1:
             return
 
+        # Save the user input before it gets modified by MambaModelConfig
+        mamba_block_size = vllm_config.cache_config.mamba_block_size
         # Enable FULL_AND_PIECEWISE by default
         MambaModelConfig.verify_and_update_config(vllm_config)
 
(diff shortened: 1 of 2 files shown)


vllm/model_executor/models/config.py
```diff
@@ -333,8 +333,6 @@ def verify_and_update_config(cls, vllm_config: "VllmConfig") -> None:
         if not envs.VLLM_USE_V1:
             return
 
-        # Save the user input before it gets modified by MambaModelConfig
-        mamba_block_size = vllm_config.cache_config.mamba_block_size
```
tlrmchlsmth, line 337: Broken in https://github.com/vllm-project/vllm/pull/27809

@tdoublep do you know why we needed to save the `mamba_block_size` to begin with?

tdoublep: Yes: https://github.com/vllm-project/vllm/pull/27809/files#r2478085718

vllm/v1/core/sched/scheduler.py
```diff
@@ -93,7 +93,7 @@ def __init__(
             )
 
             connector_vllm_config = copy.copy(self.vllm_config)
-            connector_vllm_config.kv_cache_config = copy.copy(kv_cache_config)
+            connector_vllm_config.cache_config = copy.copy(kv_cache_config)
```
tlrmchlsmth: @KuntaiDu any ideas what's going on here?

tlrmchlsmth: Context is here: https://github.com/vllm-project/vllm/pull/25712/files#r2447539735

DarkLight1337 (approved):

vllm/v1/core/sched/scheduler.py
GuanLuo:
```suggestion
            return self.connector.request_finished(request, block_ids)  # type: ignore[attr-defined]
```
Should be able to just ignore the type check here, this line will not be hit at the current state (no connector implements HMA interface).

For future reference, I think `request_finished_all_groups` should be called here as it is defined in SupportHMA interface and has the correct function signature.

tlrmchlsmth: Switched to `request_finished_all_groups`

yewentao256 (approved):
LGTM, thanks for the fix!

vllm/model_executor/models/config.py
hmellor: https://github.com/vllm-project/vllm/pull/27289#discussion_r2451163383 is the specific thread with the explanation

vllm/v1/core/sched/scheduler.py
```diff
@@ -1335,7 +1340,8 @@ def _connector_finished(
             assert len(self.kv_cache_config.kv_cache_groups) == 1
             return self.connector.request_finished(request, block_ids[0])
         else:
-            return self.connector.request_finished(request, block_ids)
+            assert isinstance(self.connector, SupportsHMA)
+            return self.connector.request_finished_all_groups(request)
```
```suggestion
            return self.connector.request_finished_all_groups(request, block_ids)
```

vllm/v1/core/sched/scheduler.py
```diff
@@ -1335,7 +1340,8 @@ def _connector_finished(
             assert len(self.kv_cache_config.kv_cache_groups) == 1
             return self.connector.request_finished(request, block_ids[0])
         else:
-            return self.connector.request_finished(request, block_ids)
+            assert isinstance(self.connector, SupportsHMA)
```
njhill: I think we should just change the `if not supports_hma(self.connector)` above to `if not isinstance(self.connector, SupportsHMA)`.

Also the `else` isn't needed

njhill (approved):

PR #25954, opened 2025-09-30
[Performance] Split FlashAttn attention and cache update

ElizaWszola:
This PR creates codepaths for separating KV Cache update and Attention forward op. It also implements this split for FlashAttn backend. This separation facilitates future unwrapping.

#### E2E tests:
ran inference on Blackwell machine with
```
llm = LLM(model="deepseek-ai/DeepSeek-V2-Lite", trust_remote_code=True)
```
and both `VLLM_MLA_DISABLE=1` (to test the split) and `VLLM_MLA_DISABLE=0` (to test if this PR does not affect backends that don't do the split).

Some lm_eval results (flash infer has been tested to check how the PR affects non-splitting backends)
```
lm-eval --model vllm --model_args '{"pretrained": "meta-llama/Llama-3.1-8B-Instruct", "speculative_config": {"method": "eagle", "model": "yuhuili/EAGLE-LLaMA3.1-Instruct-8B", "num_speculative_tokens": 3}, "max_seq_len": 2048}' --tasks gsm8k --batch_size auto

main (flash attn):
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.7748|±  |0.0115|
|     |       |strict-match    |     5|exact_match|↑  |0.7020|±  |0.0126|

pr (flash attn):
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.7741|±  |0.0115|
|     |       |strict-match    |     5|exact_match|↑  |0.7013|±  |0.0126|
```
```
lm-eval --model vllm --model_args pretrained=meta-llama/Llama-3.1-8B-Instruct --tasks gsm8k --batch_size auto

main (flash infer):
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.7726|±  |0.0115|
|     |       |strict-match    |     5|exact_match|↑  |0.7013|±  |0.0126|

pr (flash infer):
[...]

PR #27816, opened 2025-10-30
[Misc] Refactor Attention kv transfer methods into decorator

NickLucche:
Small quality of life improvement by removing some of the kv transfer-specifc code in `layer.py` and refactoring that into a decorator. I believe the on entry-on exit pattern (wait_read-wait_write) here is very suitable for that.
The result is simply that there's less non-attention related code in the file. Behavior should be unchanged.

Also, I found that after grouping some common boilerplate code for both `maybe_save_kv_layer_to_connector` and `wait_for_kv_layer_from_connector`, I think there's too little left to justify a separate function for both, hence I ended-up inlining both connector method calls. 

cc @ApostaC who wrote the initial connector code

### DIFF
--- a/vllm/attention/layer.py
+++ b/vllm/attention/layer.py
@@ -886,7 +886,7 @@ def get_attention_context(
     Returns:
         A tuple containing:
         - attn_metadata: Attention metadata for this specific layer, or None if
-           no metadata available
+            no metadata available
         - attn_layer: The attention layer instance (Attention or MLAAttention)
         - kv_cache: The KV cache tensor for current virtual engine
 
--- a/vllm/attention/layer.py
+++ b/vllm/attention/layer.py
@@ -208,36 +208,6 @@ def _init_kv_cache_quant(
         layer.quant_method.create_weights(layer)
 
 
-def get_attention_context(
-    layer_name: str,
-) -> tuple[dict | object | None, "Attention | MLAAttention", torch.Tensor]:
-    """Extract attention context for a given layer.
-
-    This helper function extracts the attention metadata, attention layer
-    instance, and KV cache tensor for a specific layer.
-
-    Args:
-        layer_name: The name/identifier of the attention layer.
-
-    Returns:
-        A tuple containing:
-        - attn_metadata: Attention metadata for this specific layer, or None if
-          no metadata available
-        - attn_layer: The attention layer instance (Attention or MLAAttention)
-        - kv_cache: The KV cache tensor for current virtual engine
-
-        Note: attn_metadata may be None, but attn_layer and kv_cache are always
-        extracted from the forward context.
-    """
-    forward_context: ForwardContext = get_forward_context()
-    attn_metadata = forward_context.attn_metadata
-    if isinstance(attn_metadata, dict):
-        attn_metadata = attn_metadata[layer_name]
-    attn_layer: Attention | MLAAttention = forward_context.no_compile_layers[layer_name]
-    kv_cache = attn_layer.kv_cache[forward_context.virtual_engine]
-    return attn_metadata, attn_layer, kv_cache
-
-
 class Attention(nn.Module, AttentionLayerBase):
     """Attention layer.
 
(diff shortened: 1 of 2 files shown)


NickLucche:
@codex review

vllm/attention/utils/kv_transfer_utils.py
```diff
@@ -0,0 +1,66 @@
...
+
+        # Execute the function
+        result = func(*args, **kwargs)
+
+        # Save KV cache layer on exit
+        attn_layer: Attention | MLAAttention = forward_context.no_compile_layers[
+            layer_name
+        ]
+        kv_cache = attn_layer.kv_cache[forward_context.virtual_engine]
+        connector.save_kv_layer(layer_name, kv_cache, attn_metadata[layer_name])
```
markmc: There's a bunch of lines here that are repeated in all the functions that we're decorating ... can we eliminate the duplication with a helper? e.g. see https://github.com/vllm-project/vllm/commit/ae4c160b98a64878e9405763c00fa7b258942c78

```
def get_attention_context(
    layer_name: str,
) -> tuple[dict | object | None, "Attention | MLAAttention", torch.Tensor]:
    """Extract attention context for a given layer.                                                                                                                                      
                                                                                                                                                                                         
    This helper function extracts the attention metadata, attention layer                                                                                                                
    instance, and KV cache tensor for a specific layer.                                                                                                                                  
                                                                                                                                                                                         
    Args:                                                                                                                                                                                
        layer_name: The name/identifier of the attention layer.                                                                                                                          
                                                                                                                                                                                         
    Returns:                                                                                                                                                                             
        A tuple containing:                                                                                                                                                              
        - attn_metadata: Attention metadata for this specific layer, or None if                                                                                                          
          no metadata available                                                                                                                                                          
        - attn_layer: The attention layer instance (Attention or MLAAttention)                                                                                                           
        - kv_cache: The KV cache tensor for current virtual engine                                                                                                                       
                                                                                                                                                                                         
        Note: attn_metadata may be None, but attn_layer and kv_cache are always                                                                                                          
        extracted from the forward context.                                                                                                                                              
    """
    forward_context: ForwardContext = get_forward_context()
    attn_metadata = forward_context.attn_metadata
    if isinstance(attn_metadata, dict):
        attn_metadata = attn_metadata[layer_name]
    attn_layer: Attention | MLAAttention = forward_context.no_compile_layers[layer_name]
    kv_cache = attn_layer.kv_cache[forward_context.virtual_engine]
    return attn_metadata, attn_layer, kv_cache
```

markmc (commented):
Love the idea in general, suggestion inline

vllm/attention/utils/kv_transfer_utils.py
NickLucche: yeah good suggestion, I was pondering having this wrapper pass those arguments to all the layer.py functions tbh, but I was pretty sure that would be mess up the compile/cuda graph step. 
I will just add your suggestion

vllm/attention/layer.py
```diff
@@ -208,6 +204,36 @@ def _init_kv_cache_quant(
         layer.quant_method.create_weights(layer)
 
 
+def get_attention_context(
+    layer_name: str,
+) -> tuple[dict | object | None, "Attention | MLAAttention", torch.Tensor]:
```
ProExpertProg: Nit: can you add a proper metadata class type annotation?

vllm/attention/utils/kv_transfer_utils.py
```diff
@@ -0,0 +1,64 @@
...
+from typing import TYPE_CHECKING
+
+from vllm.distributed.kv_transfer import (
+    get_kv_transfer_group,
+    has_kv_transfer_group,
+    is_v1_kv_transfer_group,
+)
+
+if TYPE_CHECKING:
+    pass
```
ProExpertProg: Nit: noop?
```suggestion
```

vllm/attention/utils/kv_transfer_utils.py
```diff
@@ -0,0 +1,64 @@
...
+        raise TypeError(
+            f"Function {func.__name__} must have a 'layer_name' parameter"
+        ) from e
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        if not has_kv_transfer_group() or not is_v1_kv_transfer_group():
+            return func(*args, **kwargs)
+
+        layer_name: str = args[layer_name_index]
```
ProExpertProg, line 41: Would this technically not work if `layer_name` is passed in kwargs?

vllm/attention/layer.py
```diff
@@ -912,23 +907,16 @@ def maybe_calc_kv_scales_fake(
 )
 
 
+@maybe_transfer_kv_layer
```
ProExpertProg, line 905: @NickLucche or @ElizaWszola could one of you just check this is compatible with #25954?

ProExpertProg (commented):
Just a few nits and Qs

vllm/attention/utils/kv_transfer_utils.py
NickLucche: Correct! We could easily handle that but I think it'd be dead code as we own all instances of use of `maybe_transfer_kv_layer`

markmc (approved):

vllm/attention/layer.py
ElizaWszola: Looks good to me @ProExpertProg

vllm/attention/utils/kv_transfer_utils.py
ProExpertProg: Maybe just add an assert then?

NickLucche: I'll just allow kwargs too

ProExpertProg (approved):
Nice cleanup

vllm/attention/layer.py
```diff
@@ -911,23 +872,46 @@ def maybe_calc_kv_scales_fake(
...
+    This helper function extracts the attention metadata, attention layer
+    instance, and KV cache tensor for a specific layer.
+
+    Args:
+        layer_name: The name/identifier of the attention layer.
 
+    Returns:
+        A tuple containing:
+        - attn_metadata: Attention metadata for this specific layer, or None if
+          no metadata available
```
hmellor: @NickLucche https://app.readthedocs.org/projects/vllm/builds/30282990/#293811637--1144

```suggestion
        no metadata available
```

or

```suggestion
            no metadata available
```

(not sure which will render better)

PR #27826, opened 2025-10-30
[Cleanup] Remove no-longer-used `SpeculativeConfig.enable_chunked_prefill`

njhill:
Was only used in V0 when it was incompatible with spec decode.

Also removed now-unused `disable_logprobs` and `disable_log_stats` fields.

Also fix a local pre-commit error in `api_server.py`.


yewentao256 (approved):
LGTM, thanks for the work!

hmellor (changes requested):
Could you also:

- add `deprecated=True` to the `add_argument` calls for these fields
- update the `help` message for these fields so users see they're deprecated in docs
- set the field defaults to `None` add a `warning_once` if they are `not None` in `EngineArgs.__post_init__`?

Otherwise users won't know that they're deprecated.

hmellor:
Ah I see, my mistake. (very happy to see duplicated configs removed!)

hmellor (approved):

PR #27834, opened 2025-10-30
[Kimi-Linear] Correct prefixes and add compatibility to AWQ quants

toncao:
## Purpose
This PR purpose is to add prefix to shared_experts params and correct block_sparse_moe prefix from "mlp" to "block_sparse_moe", which ultimately allows vllm to initiate layer names correctly as e.g., `layers.1.block_sparse_moe.shared_experts.down_proj` instead of `layers.1.mlp.down_proj` (layer 1 does not have mlp.down_proj).

Not initiating names correctly as its state_dict leads to e.g.,:
`KeyError: layers.1.block_sparse_moe.shared_experts.down_proj`
as the it can not match the block_sparse_moe.shared_experts modules to AWQ quantization config.

The error occurs with this [AWQ model](https://huggingface.co/cyankiwi/Kimi-Linear-48B-A3B-Instruct-AWQ-4bit) quant.

---
<details>

- [ x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
</details>

### DIFF
--- a/vllm/model_executor/models/kimi_linear.py
+++ b/vllm/model_executor/models/kimi_linear.py
@@ -155,6 +155,7 @@ class KimiMoE(nn.Module):
                 hidden_act=config.hidden_act,
                 quant_config=quant_config,
                 reduce_results=False,
+                prefix=f"{prefix}.shared_experts",
             )
 
     def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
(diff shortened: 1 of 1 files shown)


vllm/model_executor/models/kimi_linear.py
```diff
@@ -155,6 +155,7 @@ def __init__(
                 hidden_act=config.hidden_act,
                 quant_config=quant_config,
                 reduce_results=False,
```
jeejeelee, line 157: @zhiyuan1i May I ask if `reduce_results` should be True?

zhiyuan1i: @jeejeelee I think they should be False, since https://github.com/vllm-project/vllm/blob/09c9d32a52e33fb22f63a9e7c465a8e248ed8439/vllm/model_executor/models/kimi_linear.py#L174-L175

I manually call `final_hidden_states = tensor_model_parallel_all_reduce(final_hidden_states)`

jeejeelee (approved):
LGTM

Issue #27691, opened 2025-10-28
[RFC]: Tracking Spec Decode Support


MatthewBonanni:
### Motivation.

I've started to try to understand how spec decode composes with other features. Each row represents one combination, and the status column indicates whether this combination runs without issues. This issue will track the status for now.

| Model | Hardware | Spec Decode | TP Size | DP Size | EP Enabled? | CUDA Graph Mode      | DCP Size | DBO Enabled? | KV Cache DType | Attn Backend     | Runs? | GSM8K Score | Notes                                                                                                                                                                     |
| ----- | -------- | ----------- | ------- | ------- | ----------- | -------------------- | -------- | ------------ | -------------- | ---------------- | ----- | ----------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| L3-8B | H100     |             | 1       | 1       | N           | `FULL_AND_PIECEWISE` | 1        | N            | `auto`         | `FLASH_ATTN`     | ✅     | 0.64        |                                                                                                                                                                           |
| L3-8B | H100     | EAGLE       | 1       | 1       | N           | `FULL_AND_PIECEWISE` | 1        | N            | `auto`         | `FLASH_ATTN`     | ✅     | 0.65        |                                                                                                                                                                           |
| L3-8B | H100     | EAGLE       | 2       | 1       | N           | `FULL_AND_PIECEWISE` | 1        | N            | `auto`         | `FLASH_ATTN`     | ✅     | 0.65        |                                                                                                                                                                           |
| L3-8B | H100     |             | 2       | 1       | N           | `FULL_AND_PIECEWISE` | 1        | N            | `auto`         | `TRITON_ATTN`    | ✅     | 0.64        |                                                                                                                                                                           |
| L3-8B | H100     | EAGLE       | 2       | 1       | N           | `FULL_AND_PIECEWISE` | 1        | N            | `auto`         | `TRITON_ATTN`    | ❌     |             | IMA                                                                                                                                                                       |
| L3-8B | H100     |             | 2       | 1       | N           | `FULL_AND_PIECEWISE` | 1        | N            | `auto`         | `FLASHINFER`     | ✅     | 0.65        |                                                                                                                                                                           |
| L3-8B | H100     | EAGLE       | 2       | 1       | N           | `FULL_AND_PIECEWISE` | 1        | N            | `auto`         | `FLASHINFER`     | ❌     |             | IMA                                                                                                                                                                       |
[...]

PR #27837, opened 2025-10-30
[Spec Decode] Fix spec decode + DP bug

MatthewBonanni:
## Purpose
Spec decode + DP is currently broken because `coordinate_batch_across_dp` causes a hang for the drafter (see #27691). This PR addresses this as follows:
* ALL ranks must agree to run the drafter for the drafter to be run
* If ANY rank requires DP padding, it is used
* Dummy logic is changed to ensure synchronization

## Test Plan
```
vllm serve meta-llama/Meta-Llama-3-8B-Instruct --speculative-config '{"method": "eagle", "model": "yuhuili/EAGLE-LLaMA3-Instruct-8B", "num_speculative_tokens": 3}' -dp 2
```
with
```
lm_eval --model local-chat-completions --model_args base_url=http://localhost:8000/v1/chat/completions,model=meta-llama/Meta-Llama-3-8B-Instruct,num_concurrent=32,max_retries=3,tokenized_requests=False --tasks gsm8k --apply_chat_template
```

## Test Result
Previously hangs or crashes (due to DP padding inconsistencies, depending on CUDA graph mode), now runs successfully:
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.6444|±  |0.0132|
|     |       |strict-match    |     5|exact_match|↑  |0.1448|±  |0.0097|
```

---
<details>

</details>

### DIFF
--- a/tests/models/test_gguf_download.py
+++ b/tests/models/test_gguf_download.py
@@ -1,0 +1,240 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from vllm.config import ModelConfig
+from vllm.config.load import LoadConfig
+from vllm.model_executor.model_loader.gguf_loader import GGUFModelLoader
+from vllm.model_executor.model_loader.weight_utils import download_gguf
+
+
+class TestGGUFDownload:
+    """Test GGUF model downloading functionality."""
+
+    @patch("vllm.model_executor.model_loader.weight_utils.download_weights_from_hf")
+    def test_download_gguf_single_file(self, mock_download):
+        """Test downloading a single GGUF file."""
+        # Setup mock
+        mock_folder = "/tmp/mock_cache"
+        mock_download.return_value = mock_folder
+
+        # Mock glob to return a single file
+        with patch("glob.glob") as mock_glob:
+            mock_glob.side_effect = lambda pattern, **kwargs: (
+                [f"{mock_folder}/model-IQ1_S.gguf"] if "IQ1_S" in pattern else []
+            )
+
+            result = download_gguf("unsloth/Qwen3-0.6B-GGUF", "IQ1_S")
+
+            # Verify download_weights_from_hf was called with correct patterns
+            mock_download.assert_called_once_with(
+                model_name_or_path="unsloth/Qwen3-0.6B-GGUF",
+                cache_dir=None,
+                allow_patterns=[
+                    "*-IQ1_S.gguf",
+                    "*-IQ1_S-*.gguf",
+                    "*/*-IQ1_S.gguf",
+                    "*/*-IQ1_S-*.gguf",
+                ],
+                revision=None,
+                ignore_patterns=None,
+            )
+
+            # Verify result is the file path, not folder
+            assert result == f"{mock_folder}/model-IQ1_S.gguf"
+
+    @patch("vllm.model_executor.model_loader.weight_utils.download_weights_from_hf")
+    def test_download_gguf_sharded_files(self, mock_download):
+        """Test downloading sharded GGUF files."""
+        mock_folder = "/tmp/mock_cache"
+        mock_download.return_value = mock_folder
+
+        # Mock glob to return sharded files
+        with patch("glob.glob") as mock_glob:
+            mock_glob.side_effect = lambda pattern, **kwargs: (
+                [
+                    f"{mock_folder}/model-Q2_K-00001-of-00002.gguf",
+                    f"{mock_folder}/model-Q2_K-00002-of-00002.gguf",
+                ]
+                if "Q2_K" in pattern
+                else []
+            )
+
+            result = download_gguf("unsloth/gpt-oss-120b-GGUF", "Q2_K")
+
+            # Should return the first file after sorting
+            assert result == f"{mock_folder}/model-Q2_K-00001-of-00002.gguf"
+
+    @patch("vllm.model_executor.model_loader.weight_utils.download_weights_from_hf")
+    def test_download_gguf_subdir(self, mock_download):
+        """Test downloading GGUF files from subdirectory."""
+        mock_folder = "/tmp/mock_cache"
+        mock_download.return_value = mock_folder
+
+        with patch("glob.glob") as mock_glob:
+            mock_glob.side_effect = lambda pattern, **kwargs: (
+                [f"{mock_folder}/Q2_K/model-Q2_K.gguf"]
+                if "Q2_K" in pattern or "**/*.gguf" in pattern
+                else []
+            )
+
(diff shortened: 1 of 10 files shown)


vllm/forward_context.py
```diff
@@ -282,8 +283,10 @@ def set_forward_context(
...
     dp_metadata: DPMetadata | None = None
-    if vllm_config.parallel_config.data_parallel_size > 1 and (
-        attn_metadata is not None or num_tokens is not None
+    if (
+        not skip_dp_coordination
+        and vllm_config.parallel_config.data_parallel_size > 1
+        and (attn_metadata is not None or num_tokens is not None)
     ):
         # If num_tokens_across_dp hasn't already been initialized, then
         # initialize it here. Both DP padding and Microbatching will be
```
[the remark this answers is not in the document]
MatthewBonanni: PR code runs fine as-is with `mistralai/Mixtral-8x7B-Instruct-v0.1` (an MoE model) with both DP and DP+EP

vllm/forward_context.py
SageMoore, line 294: `num_tokens_unpadded=num_tokens or 0`

vllm/forward_context.py
```diff
@@ -282,8 +283,10 @@ def set_forward_context(
         forward_start_time = time.perf_counter()
 
     dp_metadata: DPMetadata | None = None
-    if vllm_config.parallel_config.data_parallel_size > 1 and (
-        attn_metadata is not None or num_tokens is not None
+    if (
+        not skip_dp_coordination
+        and vllm_config.parallel_config.data_parallel_size > 1
+        and (attn_metadata is not None or num_tokens is not None)
```
SageMoore: You can delete the `(attn_metadata is not None or num_tokens is not None)` portion of this if statement.

vllm/forward_context.py
SageMoore, line 292: You can delete this assert

SageMoore (commented):
Can you try with the changes that I proposed? I suspect that will resolve the hang issue that you are seeing.

LucasWilkinson (commented):
It feels like this is adding a dummy_run into execute_model; is there a reason this pathway isn't working?: https://github.com/vllm-project/vllm/blob/da855b42d2005d9b701758e9d59836131ee8c6fb/vllm/v1/engine/core.py#L1217-L1225

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2431,67 +2431,151 @@ def execute_model(
...
+                    if self.parallel_config.disable_nccl_for_dp_synchronization:
+                        device = "cpu"
+                        group = get_dp_group().cpu_group
+
+                    drafter_flag = torch.tensor(
+                        [1 if should_run_drafter else 0],
+                        device=device,
+                        dtype=torch.int32,
+                    )
+                    dist.all_reduce(drafter_flag, op=dist.ReduceOp.MIN, group=group)
```
SageMoore: I haven't read through this code super closely yet, but I would really like to avoid adding additional all reduces. Can you augment `coordinate_batch_across_dp` to handle this case? Happy to follow up offline to explain in more detail how that could look.

MatthewBonanni: done!

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2524,13 +2569,25 @@ def execute_model(
...
             with self.synchronize_input_prep():
                 # Update persistent batch states.
                 self._update_states(scheduler_output)
 
-                if not num_scheduled_tokens:
+                if not total_num_scheduled_tokens:
+                    # In DP mode, participate in coordinate_batch_across_dp
+                    # to maintain synchronization across ranks
+                    if self.parallel_config.data_parallel_size > 1:
+                        coordinate_batch_across_dp(
```
SageMoore: Why do you need this?

SageMoore (commented):
`dp_utils.py` changes look reasonable, but I don't understand why you need the extra all reduce in `execute_model`.

vllm/v1/worker/gpu_model_runner.py
MatthewBonanni: This isn't an extra all reduce, it's the same one as in `_prepare_inputs`, just in a different code path (the 0-token rank never reaches `_prepare_inputs`)

MatthewBonanni: A different way to fix the issue while removing this call would be to move the early return after `_prepare_inputs` but it would require some refactoring to have _prepare_inputs (and all the other methods called therein) handle 0 requests (and we'd have to remove some asserts which I'm not sure that we'd want to do)

vllm/v1/engine/core.py
```diff
@@ -1236,6 +1243,14 @@ def run_busy_loop(self):
                 # if the model didn't execute any ready requests.
                 self.execute_dummy_batch()
 
+                # After dummy batch, check if we should continue or wait.
+                # If we have no local work, sync with other ranks before continuing.
+                if not local_unfinished_reqs:
+                    self.engines_running = self._has_global_unfinished_reqs(
+                        local_unfinished_reqs
+                    )
+                    continue
```
Flechman: By adding this you're skipping the wave count increment + step counter reset below. Could you explain why this is necessary in this case ?
Maybe removing this code and simply adding
```
if not executed and not local_unfinished_reqs:
    continue
```
just below in step 3) is cleaner/clearer.

Flechman: Adding to this, it seems that with your changes, when DP>1,  `executed` is always True so the dummy_batch run is never done

vllm/v1/worker/dp_utils.py
```diff
@@ -103,10 +106,14 @@ def _synchronize_dp_ranks(
     When running microbatched or if should_attempt_dp_padding is True, all
     ranks will be padded out so that the run with the same number of tokens
 
+    3. Determines if the speculative drafter should run. Only runs if all
+    ranks agree to run it.
```
LucasWilkinson: is this the right logic? wouldn't we want to run the drafter if any rank wants it? otherwise won't we fail to speculate?

LucasWilkinson (commented):
can this be simplified by just always running the drafter in the DP case? I think this is a bit overly complicated. 

Also I think we need to run the drafter when any rank wants to run the drafter since for eagle the drafter needs to be prefilled. If its skipped it will miss tokens potentially tanking the acceptance rate.

vllm/v1/worker/dp_utils.py
MatthewBonanni: A rank can't run the drafter when `max_seq_len + self.speculative_config.num_speculative_tokens > effective_drafter_max_model_len`. It's less about if ranks "want" to run it (they always "want" to if spec decode is activated) and more about if they can. If any rank is unable, the drafter is disabled for all to keep them in sync

PR #27841, opened 2025-10-30
[Feat] Drop-in Torch CUDA Profiler

benchislett:
## Purpose

This PR adds a CUDA profiler via `torch.cuda.profiler`, which integrates natively with the existing `start_profile` infrastructure.

It is very easy to use. On NVIDIA GPUs:
```
export VLLM_TORCH_CUDA_PROFILE=1
nsys profile -o report.nsys-rep \
  --trace-fork-before-exec=true \
  --cuda-graph-trace=node \
  --capture-range=cudaProfilerApi \
  --capture-range-end repeat \
  vllm serve meta-llama/Llama-3.1-8B-Instruct
```

Then to benchmark, the usual `--profile` can be used:
```
vllm bench serve --profile --dataset-name random --ignore-eos --num-prompts 8 --request-rate 4 --random-input-len 256 --random-output-len 32 --model meta-llama/Llama-3.1-8B-Instruct --base-url http://0.0.0.0:8000
```

With `--capture-range-end repeat`, the profiles can be repeated any number of times, at any point in the engine's lifetime. When you stop the server, it will save all the profiles it recorded (`report.1.nsys-rep`, `report.2.nsys-rep` etc.).


ProExpertProg (approved):
Seems cleaner than #27693 but can you take a look if there's anything from that PR we should incorporate here?

benchislett:
I've reviewed and messaged the authors of prior works for discussion. I think that this PR is a minimal starting point that we can build out new features on top of if we need them. I don't intend for this to be a fully-featured PR, though I have some ideas for follow-ups I think it is best to keep the PRs modular to ensure we don't pollute the engine core with profiling logic

vllm/entrypoints/openai/api_server.py
```diff
@@ -1202,11 +1202,17 @@ async def invocations(raw_request: Request):
...
+    if envs.VLLM_TORCH_PROFILER_DIR:
+        logger.warning(
+            "Torch Profiler is enabled in the API server. This should ONLY be "
+            "used for local development!"
+        )
+    else:
+        logger.warning(
+            "CUDA Profiler is enabled in the API server. This should ONLY be "
+            "used for local development!"
+        )
```
mgoin: nit: we should remove the outer condition e.g. `if envs.VLLM_TORCH_PROFILER_DIR:` then `elif envs.VLLM_TORCH_CUDA_PROFILE:`
also we should use warning_once

vllm/profiler/gpu_profiler.py
```diff
@@ -0,0 +1,34 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+import torch.cuda.profiler as cuda_profiler
```
mgoin: nit: can we make this import lazy in the start and stop functions? Not sure if this conflicts with amd in the gpu worker

mgoin (commented):
This seems like a good starting point to the path for implementation I would prefer. If this is good for your use, then let's get this merged first! Also thank you for updating the docs :) 
Just some nits

vllm/entrypoints/openai/api_server.py
```diff
@@ -1287,10 +1287,16 @@ async def invocations(raw_request: Request):
...
+    logger.warning_once(
         "Torch Profiler is enabled in the API server. This should ONLY be "
         "used for local development!"
     )
+elif envs.VLLM_TORCH_CUDA_PROFILE:
+    logger.warning_once(
+        "CUDA Profiler is enabled in the API server. This should ONLY be "
+        "used for local development!"
+    )
+if envs.VLLM_TORCH_PROFILER_DIR or envs.VLLM_TORCH_CUDA_PROFILE:
```
mgoin, line 1292: Ah I didn't realize this scope was needed for the following functions, my bad. This refactor was unneeded, my bad

mgoin (approved):

benchislett:
@mgoin any insight on the test failures? Seems unrelated

cfregly (commented):
Great stuff!  @mgoin @benchislett   Lots we can do around profiling and vLLM.  Excited to see this in focus.

Here are some possible issues (not major, just some findings/suggestions upon quick review):

Offline latency profiling is tied only to the Torch profiler env var, so CUDA profiling cannot trigger start and stop in offline runs.
```
if args.profile and not envs.VLLM_TORCH_PROFILER_DIR:
    raise OSError(
        "The environment variable 'VLLM_TORCH_PROFILER_DIR' is not set. "
        "Please set it to a valid path to use torch profiler."
    )

profile_dir = envs.VLLM_TORCH_PROFILER_DIR
```

CUDA profiling is silently skipped if both env vars are set, and the API server only logs the Torch profiler warning.
```
if envs.VLLM_TORCH_PROFILER_DIR:
    self.profiler = torch.profiler.profile(...)
elif envs.VLLM_TORCH_CUDA_PROFILE:
    self.profiler = CudaProfilerWrapper()
else:
    self.profiler = None
if envs.VLLM_TORCH_PROFILER_DIR:
    logger.warning_once("Torch Profiler is enabled in the API server...")
elif envs.VLLM_TORCH_CUDA_PROFILE:
    logger.warning_once("CUDA Profiler is enabled in the API server...")
```

The CUDA profiler wrapper has a shutdown path but nothing calls it during worker teardown.
```
class CudaProfilerWrapper:
    def shutdown(self) -> None:
        self.stop()
def shutdown(self) -> None:
    if runner := getattr(self, "model_runner", None):
        runner.ensure_kv_transfer_shutdown()
```

Also, there seems to be no test coverage for the new CUDA profiler path and the start or stop flow, which might be overkill, but just pointing it out.

Suggestions
- Allow --profile to work when either VLLM_TORCH_PROFILER_DIR or VLLM_TORCH_CUDA_PROFILE is set, and update the output message so it does not mention a trace directory when CUDA profiling is used.
- Add an explicit --profile-mode torch|cuda flag for vllm bench latency, vllm bench throughput, vllm bench serve, and vllm serve. Make --profile trigger start and stop based on the chosen mode.
- Add a conflict check when both env vars are set, and call self.profiler.shutdown() in worker or engine shutdown to flush captures.
- Add optional NVTX ranges for prefill, decode, and scheduler steps to improve Nsight readability.
- Add a CUDA enabled test that exercises /start_profile to /stop_profile with VLLM_TORCH_CUDA_PROFILE=1.

Decisions to make
- If both env vars are set, do you want a hard error or a warning with a deterministic choice for which profiler runs.

PR #27843, opened 2025-10-30
[Bugfix] Flashinfer block size for hybrid ssm models

heheda12345:
## Purpose

Hybrid SSM may use a block size that is different with `kv_cache_spec.block_size` when kv_cache_spec.block_size doesn't supported by an attention backend. However, we don't know the used block_size now. As a temporary solution, hardcode flashinfer's block_size to one value to make sure model runner always uses this block_size.

## Test Plan

```
VLLM_ATTENTION_BACKEND=FLASHINFER CUDA_VISIBLE_DEVICES=4,5,6,7 vllm serve Qwen/Qwen3-Next-80B-A3B-Instruct -tp 4 --enable-expert-parallel --no-enable-prefix-caching --async-scheduling

```

## Test Result

```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8522|±  |0.0098|
|     |       |strict-match    |     5|exact_match|↑  |0.8089|±  |0.0108|
```

---
<details>

</details>


vadiklyutiy:
The changes in `vllm/model_executor/models/config.py` correspond to #27704, except that this version aligns by 32, whereas #27704 aligns to the next power of two. I don’t have a strong preference either way.

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -291,6 +291,7 @@ def __init__(
         self._workspace_buffer = None
         self._prefill_wrapper = None  # Wrapper for prefill/append
         self._decode_wrapper = None  # Wrapper for decode (general shape)
+        block_size = 32  # Note(Chen): Hardcode the block_size as 32 temporarily.
```
vadiklyutiy: I believe #27547 proposes a more universal approach: it reads the actual supported sizes instead of hard-coding them here. Otherwise, if someone changes `FlashInferBackend.get_supported_kernel_block_size`, they would have no way to know this file needs to be updated as well.

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -170,10 +170,10 @@ def get_supported_head_sizes(cls) -> list[int]:
...
     @staticmethod
     def get_supported_kernel_block_size() -> list[int | MultipleOf]:
-        # Note: Not sure for all platforms,
-        # but on Blackwell, only support a page size of
-        # 16, 32, 64
-        return [16, 32, 64]
+        # Note(Chen): FlashInfer backend supports other block_sizes. But as
+        # the backend doesn't know the block_size selected, we hardcode it as only
+        # supports 32 for now.
+        return [32]
```
vadiklyutiy: Do you have any evidence that something wrong with `64`?

heheda12345: No problem to 64. But we need to only allow one block_size here. Happy to change it to 64 if it is better.

heheda12345:
fixed by https://github.com/vllm-project/vllm/pull/27994

PR #27849, opened 2025-10-31
[NIXL][XPU] Pin NIXL version to 0.7.0

zhenwei-intel:
Pin NIXL version to 0.7.0

### DIFF
--- a/tools/install_nixl_from_source_ubuntu.py
+++ b/tools/install_nixl_from_source_ubuntu.py
@@ -228,7 +228,7 @@ def build_and_install_prerequisites(args):
         "-m",
         "pip",
         "install",
-        "--no-deps",
+        "--no-deps",  # w/o "no-deps", it will install cuda-torch
         newly_built_wheel,
     ]
     if args.force_reinstall:
--- a/tools/install_nixl_from_source_ubuntu.py
+++ b/tools/install_nixl_from_source_ubuntu.py
@@ -9,6 +9,7 @@ import sys
 
 # --- Configuration ---
 WHEELS_CACHE_HOME = os.environ.get("WHEELS_CACHE_HOME", "/tmp/wheels_cache")
+NIXL_VERSION = os.environ.get("NIXL_VERSION", "0.7.0")
 ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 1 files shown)


zhenwei-intel:
cc @xuechendi

tools/install_nixl_from_source_ubuntu.py
```diff
@@ -9,6 +9,7 @@
 
 # --- Configuration ---
 WHEELS_CACHE_HOME = os.environ.get("WHEELS_CACHE_HOME", "/tmp/wheels_cache")
+NIXL_VERSION = os.environ.get("NIXL_VERSION", "0.7.0")
```
jikunshang: I feel default should always use latest. so we don't need change this file frequently. user/ CI script can set this env vars by themselves. does this make sense?

zhenwei-intel: update to fetch latest release version by default

jikunshang (approved):
LGTM. leave @xuechendi to take further review.

jikunshang:
can you fix precommit issue?

tools/install_nixl_from_source_ubuntu.py
```diff
@@ -203,7 +223,14 @@ def build_and_install_prerequisites(args):
             {os.path.basename(newly_built_wheel)}. Now installing...",
         flush=True,
     )
-    install_command = [sys.executable, "-m", "pip", "install", newly_built_wheel]
+    install_command = [
+        sys.executable,
+        "-m",
+        "pip",
+        "install",
+        "--no-deps",
```
xuechendi: "--no-deps", is there special reason for that.
Also, I learned that latest nixl build wheel with nixl-cu, did you see similiar issue after pin to release version instead of main?

xuechendi (approved):

tools/install_nixl_from_source_ubuntu.py
zhenwei-intel: w/o "no-deps", script will install cuda-torch

xuechendi: I see, suggesting to add info as comments for future reference

zhenwei-intel: updated

PR #27869, opened 2025-10-31
[Perf] Support stream interval for reducing host overhead

elvischenv:
## Purpose

Support `--stream-interval <num_of_tokens>`. The flag will buffer a number of tokens for a single request before sending back to client. The buffering can reduce the time needed to send responses back to clients, thereby saving host overhead like network/IO.

cc @benchislett @pavanimajety @nvpohanh 

## Testing with `openai/gpt-oss-20b`:
B200 server cmd:
```
VLLM_USE_FLASHINFER_MOE_MXFP4_MXFP8=1 \
--kv-cache-dtype fp8 \
--tensor-parallel-size 1 \
--max-num-seqs 1024 \
--max-model-len 10240 \
--max-num-batched-tokens 8192 \
--max-cudagraph-capture-size 2048 \
-O.pass_config.enable_noop=True \
--async-scheduling \
--no-enable-prefix-caching
```

## Accuracy Test
```
OPENAI_API_KEY="test" \
python -m gpt_oss.evals \
--sampler chat_completions \
--model openai/gpt-oss-20b \
--reasoning-effort medium \
--n-threads 512 \
--eval aime25
```
Without `--stream-interval`(by default is set to `1`):
```
[{'eval_name': 'aime25', 'model_name': 'gpt-oss-20b-medium_temp1.0_20251031_095139', 'metric': 0.7083333333333334}]
```
With `--stream-interval 10`:
```
[{'eval_name': 'aime25', 'model_name': 'gpt-oss-20b-medium_temp1.0_20251031_150201', 'metric': 0.7291666666666666}]
```

## Perf Test
```
vllm bench serve \
--model openai/gpt-oss-20b \
--trust-remote-code \
--dataset-name random \
--ignore-eos \
--max-concurrency 1024 \
--num-prompts 5120 \
--random-input-len 1024 \
--random-output-len 1024
```
Without `--stream-interval`(by default is set to `1`):
```
Successful requests:                     5120
Maximum request concurrency:             1024
Benchmark duration (s):                  153.85
Total input tokens:                      5242880
Total generated tokens:                  5242880
Request throughput (req/s):              33.28
Output token throughput (tok/s):         34078.85
Peak output token throughput (tok/s):    42240.00
Peak concurrent requests:                1554.00
Total Token throughput (tok/s):          68157.70
Mean TTFT (ms):                          2698.80
Median TTFT (ms):                        2300.56
P99 TTFT (ms):                           6826.32
Mean TPOT (ms):                          27.16
Median TPOT (ms):                        27.40
P99 TPOT (ms):                           28.57
Mean ITL (ms):                           27.41
Median ITL (ms):                         25.22
P99 ITL (ms):                            89.70
```
With `--stream-interval 10`: **57% e2e perf gain**
```
Successful requests:                     5120
Maximum request concurrency:             1024
Benchmark duration (s):                  98.17
Total input tokens:                      5242880
Total generated tokens:                  5242880
Request throughput (req/s):              52.15
Output token throughput (tok/s):         53403.94
Peak output token throughput (tok/s):    8216.00
Peak concurrent requests:                1180.00
Total Token throughput (tok/s):          106807.88
Mean TTFT (ms):                          1002.64
Median TTFT (ms):                        262.37
P99 TTFT (ms):                           6749.87
Mean TPOT (ms):                          18.09
Median TPOT (ms):                        18.68
P99 TPOT (ms):                           18.85
Mean ITL (ms):                           179.65
Median ITL (ms):                         137.12
P99 ITL (ms):                            512.61
```

---
<details>

</details>

### DIFF
--- a/vllm/config/scheduler.py
+++ b/vllm/config/scheduler.py
@@ -145,7 +145,7 @@ class SchedulerConfig:
     stream_interval: int = Field(default=1, ge=1)
     """The interval (or buffer size) for streaming in terms of token length.
     A smaller value (1) makes streaming smoother by sending each token immediately,
-    while a larger value (e.g., 10) reduces host overhead and increases throughput
+    while a larger value (e.g., 10) reduces host overhead and may increase throughput
     by batching multiple tokens before sending."""
 
     def get_scheduler_cls(self) -> type["SchedulerInterface"]:
--- a/vllm/config/scheduler.py
+++ b/vllm/config/scheduler.py
@@ -142,6 +142,12 @@ class SchedulerConfig:
     speculative decoding and pipeline parallelism.
     """
 
+    stream_interval: int = Field(default=1, ge=1)
+    """The interval (or buffer size) for streaming in terms of token length.
+    A smaller value (1) makes streaming smoother by sending each token immediately,
+    while a larger value (e.g., 10) reduces host overhead and increases throughput
+    by batching multiple tokens before sending."""
+
     def get_scheduler_cls(self) -> type["SchedulerInterface"]:
         if self.scheduler_cls is None:
             if self.async_scheduling:
--- a/vllm/v1/engine/output_processor.py
+++ b/vllm/v1/engine/output_processor.py
@@ -215,33 +215,28 @@ class RequestState:
             # Only the final output is required in FINAL_ONLY mode.
             return None
 
-        assert self.detokenizer is not None
-
-        # Send output request only when
-        # 1. It has finished, or
-        # 2. It is the first token, or
-        # 3. It has reached the stream interval number of tokens
-        if not (
-            finished
-            or self.sent_tokens_offset == 0
-            or len(self.detokenizer.output_token_ids) - self.sent_tokens_offset
-            >= self.stream_interval
-        ):
-            return None
-
-        if self.output_kind == RequestOutputKind.DELTA:
-            # Send tokens from the offset in DELTA mode
-            tokens_to_send = self.detokenizer.output_token_ids[
-                self.sent_tokens_offset :
-            ]
-            self.sent_tokens_offset = len(self.detokenizer.output_token_ids)
-        else:
-            if self.stream_interval > 1 and final_only:
-                logger.warning_once(
-                    "stream_interval > 1 is not supported for FINAL_ONLY mode. "
-                    "Ignoring the stream_interval setting."
-                )
-            tokens_to_send = new_token_ids
+        if self.stream_interval > 1:
+            assert self.detokenizer is not None
+    
+            # Send output request only when
+            # 1. It has finished, or
+            # 2. It is the first token, or
+            # 3. It has reached the stream interval number of tokens
+            if not (
+                finished
+                or self.sent_tokens_offset == 0
+                or len(self.detokenizer.output_token_ids) - self.sent_tokens_offset
+                >= self.stream_interval
+            ):
+                return None
+    
+            if self.output_kind == RequestOutputKind.DELTA:
+                # Send tokens from the offset in DELTA mode, otherwise all
+                # tokens are sent.
+                new_token_ids = self.detokenizer.output_token_ids[
+                    self.sent_tokens_offset :
+                ]
+                self.sent_tokens_offset = len(self.detokenizer.output_token_ids)
 
         request_id = self.request_id
         if pooling_output is not None:
--- a/vllm/v1/engine/output_processor.py
+++ b/vllm/v1/engine/output_processor.py
@@ -8,7 +8,6 @@ from typing import Any, cast
 
 import torch
 
-from vllm.logger import init_logger
 from vllm.outputs import (
     CompletionOutput,
     PoolingOutput,
@@ -29,8 +28,6 @@ from vllm.v1.metrics.stats import (
     RequestStateStats,
     SchedulerStats,
 )
-
-logger = init_logger(__name__)
 
 
 class RequestOutputCollector:
--- a/vllm/v1/engine/output_processor.py
+++ b/vllm/v1/engine/output_processor.py
@@ -217,7 +217,7 @@ class RequestState:
 
         if self.stream_interval > 1:
             assert self.detokenizer is not None
-    
+
             # Send output request only when
             # 1. It has finished, or
             # 2. It is the first token, or
@@ -229,7 +229,7 @@ class RequestState:
                 >= self.stream_interval
             ):
                 return None
-    
+
             if self.output_kind == RequestOutputKind.DELTA:
                 # Send tokens from the offset in DELTA mode, otherwise all
                 # tokens are sent.
@@ -244,7 +244,7 @@ class RequestState:
                 request_id, [self._new_pooling_output(pooling_output)], finished
             )
 
-        output = self._new_completion_output(tokens_to_send, finish_reason, stop_reason)
+        output = self._new_completion_output(new_token_ids, finish_reason, stop_reason)
 
         if self.parent_req is None:
             outputs = [output]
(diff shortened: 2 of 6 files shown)


elvischenv:
@codex review

vllm/config/scheduler.py
```diff
@@ -295,4 +301,10 @@ def _verify_args(self) -> Self:
                 f"max_num_partial_prefills ({self.max_num_partial_prefills})."
             )
 
+        if self.stream_interval < 1:
+            raise ValueError(
+                f"stream_interval ({self.stream_interval}) "
+                "must be greater than or equal to 1."
+            )
+
```
```suggestion
```

vllm/config/scheduler.py
```diff
@@ -137,6 +137,12 @@ class SchedulerConfig:
     structured outputs, speculative decoding, and pipeline parallelism.
     """
 
+    stream_interval: int = 1
```
```
from pydantic import Field
```

```suggestion
    stream_interval: int = Field(default=1, ge=1)
```

vllm/v1/engine/output_processor.py
```diff
@@ -200,13 +208,40 @@ def make_request_output(
...
+                return None
+
+            # Send tokens from the offset
+            assert self.detokenizer is not None
+            tokens_to_send = self.detokenizer.output_token_ids[
+                self.sent_tokens_offset :
+            ]
+            self.sent_tokens_offset = len(self.detokenizer.output_token_ids)
+        else:
+            # Send tokens immediately
```
nvpohanh: should we warn_once() about `stream_interval` being ignore? Something like:

```
stream_interval > 1 is only supported for DELTA mode. Ignoring the stream_interval setting.
```

elvischenv: Fixed. Thanks!

heheda12345:
CC @aarnphm @chaunceyjiang is it possible to implement all logic in api server?

vllm/v1/engine/output_processor.py
```diff
@@ -200,13 +211,45 @@ def make_request_output(
...
+        if is_delta_streaming and self.stream_interval > 1:
+            # Track total tokens generated
+            self.total_num_output_tokens += len(new_token_ids)
+
+            # should send output when it is the first token or reach the stream interval
+            should_send_output = (
+                self.sent_tokens_offset == 0
+                or self.total_num_output_tokens - self.sent_tokens_offset
+                >= self.stream_interval
+            )
```
elvischenv: > is it possible to implement all logic in api server?

@heheda12345 why do we want this? I feel like stream interval is kind of an extension of `RequestOutputKind.DELTA`. There is also `RequestOutputKind.FINAL_ONLY` which only output once after finishing.

FENP:
@elvischenv @nvpohanh Got it — thank you for clarifying!
FYI, I noticed a difference in `Peak concurrent requests` in the benchmark test results. I suspect that `--stream-interval` indirectly affects it by introducing delays in token streaming, similar to how `--request-rate` works in the bench serve ?

vllm/v1/engine/output_processor.py
```diff
@@ -205,13 +216,45 @@ def make_request_output(
             # Only the final output is required in FINAL_ONLY mode.
             return None
 
+        # Stream Interval buffering: only apply for DELTA mode and stream_interval > 1
+        is_delta_streaming = self.output_kind == RequestOutputKind.DELTA
+        if is_delta_streaming and self.stream_interval > 1:
+            # Track total tokens generated
+            self.total_num_output_tokens += len(new_token_ids)
```
njhill: Could we just use `len(self.detokenizer.output_token_ids)` here? and avoid the separate `self.total_num_output_tokens` field?

vllm/v1/engine/output_processor.py
```diff
@@ -205,13 +216,45 @@ def make_request_output(
...
+            ]
+            self.sent_tokens_offset = len(self.detokenizer.output_token_ids)
+        else:
+            if self.stream_interval > 1:
+                logger.warning_once(
+                    "stream_interval > 1 is only supported for DELTA mode. "
+                    "Ignoring the stream_interval setting."
+                )
+            # Send all the new tokens
+            tokens_to_send = new_token_ids
```
njhill: Actually it should be trivial to support `CUMULATIVE` in the same way here (i.e. for the two `not final_only` cases), we just don't need to do the `tokens_to_send` slice in that case because that arg is ignored in  `_new_completion_output`.

vllm/v1/engine/output_processor.py
```diff
@@ -205,13 +216,45 @@ def make_request_output(
...
+            # should send output when it is the first token or reach the stream interval
+            should_send_output = (
+                self.sent_tokens_offset == 0
+                or self.total_num_output_tokens - self.sent_tokens_offset
+                >= self.stream_interval
+            )
+
+            # Do NOT send output if not finished and should not send output
+            if not finished and not should_send_output:
+                return None
```
njhill: Just combine these into a single `if`?

njhill (commented):
Thanks @elvischenv this looks good to me!

Do you think you could add a test for it?

It would also be nice to make this automatic/adaptive, but can think about that as a follow-on enhancement!

vllm/v1/engine/output_processor.py
elvischenv: @njhill I have simplified the logic a lot and added unit tests. Please help review again, thanks!

vllm/v1/engine/output_processor.py
```diff
@@ -205,13 +215,41 @@ def make_request_output(
...
+                self.sent_tokens_offset :
+            ]
+            self.sent_tokens_offset = len(self.detokenizer.output_token_ids)
+        else:
+            if self.stream_interval > 1 and final_only:
+                logger.warning_once(
+                    "stream_interval > 1 is not supported for FINAL_ONLY mode. "
+                    "Ignoring the stream_interval setting."
+                )
+            tokens_to_send = new_token_ids
```
njhill: I think this could be simplified a bit more even, in particular:
- Skip in the common case that stream_interval == 1 (like you had before)
- Just update `new_token_ids` var, no need for a new one
- The warning is unnecessary I think, it's kind of obvious that the stream setting won't be related to any final_only request

```suggestion
        if self.stream_interval > 1:
            assert self.detokenizer is not None
    
            # Send output request only when
            # 1. It has finished, or
            # 2. It is the first token, or
            # 3. It has reached the stream interval number of tokens
            if not (
                finished
                or self.sent_tokens_offset == 0
                or len(self.detokenizer.output_token_ids) - self.sent_tokens_offset
                >= self.stream_interval
            ):
                return None
    
            if self.output_kind == RequestOutputKind.DELTA:
                # Send tokens from the offset in DELTA mode, otherwise all
                # tokens are sent.
                new_token_ids = self.detokenizer.output_token_ids[
                    self.sent_tokens_offset :
                ]
                self.sent_tokens_offset = len(self.detokenizer.output_token_ids)
```

vllm/config/scheduler.py
```diff
@@ -142,6 +142,12 @@ class SchedulerConfig:
     speculative decoding and pipeline parallelism.
     """
 
+    stream_interval: int = Field(default=1, ge=1)
+    """The interval (or buffer size) for streaming in terms of token length.
+    A smaller value (1) makes streaming smoother by sending each token immediately,
+    while a larger value (e.g., 10) reduces host overhead and increases throughput
+    by batching multiple tokens before sending."""
```
njhill: I think whether this is a limiting factor depends on the workload/model etc.
```suggestion
    while a larger value (e.g., 10) reduces host overhead and may increase throughput
    by batching multiple tokens before sending."""
```

njhill (commented):
Thanks @elvischenv!

mgoin (approved):

PR #27887, opened 2025-10-31
[KV Connector] Make KVCacheConfig an explicit constructor argument

markmc:
Follow on from #25712

`VllmConfig` is explicitly designed as a dataclass containing user-provided configuration and model metadata. It is a global configuration object that lives throughout the entire engine lifetime and is meant to be immutable after `__post_init__()`.

`KVCacheConfig` is worker-specific, runtime-computed state. It has limited lifetime, and its purpose is limited to initializing the KV Cache in the model runner.

Even if we add KV cache hints to model config.json in future, this would be parsed into `ModelConfig`, used as input to the `get_kv_cache_configs()` computation, and the resulting `KVCacheConfig` would still be runtime state.

We are currently creating per-worker copies of VllmConfig in order to attach the runtime `KVCacheConfig` state. But instead we should just explicitly pass `KVCacheConfig` to the connector.

Make sure to handle backwards compatibility for external connector implementations (loaded via module path) that have the old style constructor signature.

### DIFF
--- a/tests/v1/kv_connector/unit/test_backwards_compatibility.py
+++ b/tests/v1/kv_connector/unit/test_backwards_compatibility.py
@@ -1,0 +1,275 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""
+Unit tests for backwards compatibility with external KV connector implementations.
+
+This test ensures that external connectors (loaded via kv_connector_module_path)
+implemented with the old signature continue to work:
+- Old signature: __init__(self, vllm_config, role)
+- New signature: __init__(self, vllm_config, role, kv_cache_config)
+"""
+
+from typing import TYPE_CHECKING
+from unittest.mock import patch
+
+import pytest
+
+from vllm.distributed.kv_transfer.kv_connector.factory import KVConnectorFactory
+from vllm.distributed.kv_transfer.kv_connector.v1 import (
+    KVConnectorBase_V1,
+    KVConnectorRole,
+)
+from vllm.v1.core.sched.output import SchedulerOutput
+
+from .utils import create_scheduler, create_vllm_config
+
+if TYPE_CHECKING:
+    from vllm.attention.backends.abstract import AttentionMetadata
+    from vllm.config import VllmConfig
+    from vllm.forward_context import ForwardContext
+    from vllm.v1.core.kv_cache_manager import KVCacheBlocks
+    from vllm.v1.kv_cache_interface import KVCacheConfig
+    from vllm.v1.request import Request
+
+
+class OldStyleTestConnector(KVConnectorBase_V1):
+    """
+    Test connector using the old signature with 2 required arguments.
+    This simulates external connectors that haven't been updated yet.
+    """
+
+    def __init__(self, vllm_config: "VllmConfig", role: KVConnectorRole):
+        # Old-style call to super().__init__ with only 2 arguments
+        super().__init__(vllm_config=vllm_config, role=role)
+
+    def get_num_new_matched_tokens(
+        self, request: "Request", num_computed_tokens: int
+    ) -> tuple[int | None, bool]:
+        return 0, False
+
+    def update_state_after_alloc(
+        self,
+        request: "Request",
+        blocks: "KVCacheBlocks",
+        num_external_tokens: int,
+    ):
+        pass
+
+    def build_connector_meta(self, scheduler_output: SchedulerOutput):
+        return None
+
+    def start_load_kv(self, forward_context: "ForwardContext", **kwargs) -> None:
+        pass
+
+    def wait_for_layer_load(self, layer_name: str) -> None:
+        pass
+
+    def save_kv_layer(
+        self,
+        layer_name: str,
+        kv_layer,
+        attn_metadata: "AttentionMetadata",
+        **kwargs,
+    ) -> None:
+        pass
+
+    def wait_for_save(self):
+        pass
+
+
+class NewStyleTestConnector(KVConnectorBase_V1):
+    """
+    Test connector using the new signature with 3 required arguments.
+    """
+
+    def __init__(
+        self,
+        vllm_config: "VllmConfig",
+        role: KVConnectorRole,
+        kv_cache_config: "KVCacheConfig",
+    ):
+        # New-style call to super().__init__ with all 3 arguments
+        super().__init__(
+            vllm_config=vllm_config, role=role, kv_cache_config=kv_cache_config
+        )
+
+    def get_num_new_matched_tokens(
+        self, request: "Request", num_computed_tokens: int
+    ) -> tuple[int | None, bool]:
+        return 0, False
+
+    def update_state_after_alloc(
+        self,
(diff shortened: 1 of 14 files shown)


markmc:
See also https://github.com/vllm-project/vllm/pull/25712#discussion_r2479200489

tests/v1/kv_connector/unit/utils.py
```diff
@@ -254,7 +254,7 @@ def create_model_runner_output(
 
 
 class TestSharedStorageConnector(SharedStorageConnector):
-    def __init__(self, config: VllmConfig, role):
+    def __init__(self, config: VllmConfig, role, kv_cache_config):
         self.name = config.kv_transfer_config.kv_connector_extra_config["name"]
         self._connector = SharedStorageConnector(config, role)
```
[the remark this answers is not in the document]
markmc: This is intentional, avoiding having to whitelist everything in the base class in `__getattribute__()`

vllm/v1/core/sched/scheduler.py
```diff
@@ -91,15 +90,10 @@ def __init__(
...
-
-            # We're dynamically inserting a kv_cache_config variable into the
-            # connector_vllm_config. This is distinct from the cache_config
-            # that is already in there.
-            connector_vllm_config.kv_cache_config = copy.copy(kv_cache_config)  # type: ignore[attr-defined]
             self.connector = KVConnectorFactory.create_connector(
-                config=connector_vllm_config, role=KVConnectorRole.SCHEDULER
+                config=self.vllm_config,
+                kv_cache_config=self.kv_cache_config,
+                role=KVConnectorRole.SCHEDULER,
```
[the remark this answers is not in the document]
markmc: No need to maintain backwards compat for this ... we're changing something recently added

vllm/distributed/kv_transfer/kv_connector/factory.py
```diff
@@ -118,6 +128,21 @@ def get_connector_class(
                     f"Class {connector_name} not found in {connector_module_path}"
                 ) from e
             connector_cls = cast(type[KVConnectorBaseType], connector_cls)
+            if not supports_kw(connector_cls, "kv_cache_config"):
```
KuntaiDu, line 131: Just to confirm: this means that we allow connector to include `kv_cache_config` field as init args even when it does not support hybrid allocator (not a subclass of SupportsHMA)?

tests/v1/kv_connector/unit/test_backwards_compatibility.py
```diff
@@ -0,0 +1,275 @@
+# SPDX-License-Identifier: Apache-2.0
```
KuntaiDu, line 1: Maybe rename this file? Like `test_connector_init_with_kv_cache_config` or something.

KuntaiDu (approved):
In general LGTM. Some nits listed in comments.

vllm/distributed/kv_transfer/kv_connector/factory.py
markmc: Yeah. Since we we are currently unconditionally attaching `KVCacheConfig` to `VllmConfig`, I didn't even consider making it conditional until now

If we think that `KVCacheConfig` will only be used as part of implementing `request_finished_all_groups()` we could add `init_hma()` or `init_kv_cache_config()` to `SupportsHMA` ?

tests/v1/kv_connector/unit/test_backwards_compatibility.py
markmc: Obviously I don't really mind renaming it, if it helps, but my thinking is that these tests are about testing support for connectors that have not yet been updated to the new signature so it's more like "without_kv_cache_config()"

Basically, because all connectors are expected to support the new signature, we'll soon see these tests as old cruft that we need to keep around for a while

(This is different from the `SupportsHMA` approach - in that case, maybe only a small subset of connectors would be updated to take `KVCacheConfig`)

markmc:
Just capturing some of the points from the Slack discussion, since I think it was very useful :+1: 

@heheda12345 

> I think in the long term, kv cache config will be part of vllm config. But for now, as it is not created during vllm_config initialization, my concern is that there are multiple copies of vllm_config inside vllm, and it's difficult to keep them synchronized. So I suggest @KuntaiDu to add it only as a special attribute for connector
>
> But I don't want something like init(vllm_config, kv_cache_config) as kv_cache_config may be part of vllm_config in the future.
> 
> > Even if we add KV cache hints to model config.json in future, this would be parsed into ModelConfig, used as input to the get_kv_cache_configs() computation, and the resulting KVCacheConfig would still be runtime state.
>
> if we can do get_kv_cache_configs(model_config), then kv_cache_config can be init during config resolution and is immutable

@markmc 

> In future, KVCacheConfig will only ever contain immutable config from user/model, and it will no longer ever contain runtime computed state?
> 
> IMHO we should not attach runtime computed state to VllmConfig - it makes sense to have KVConnector.init(VllmConfig, KVCacheConfig) now, and work towards making KVCacheConfig only immutable config and then deprecating the KVCacheConfig argument

@heheda12345 

> should we keep connector interface as stable as possible?

@markmc 
 
> yes, but IMO "stability" means "backwards compatible for a (potentially long) deprecation period"

NickLucche (approved):
lgtm, let's keep compat code under control though

vllm/distributed/kv_transfer/kv_connector/v1/lmcache_connector.py
```diff
@@ -20,14 +20,22 @@
...
 
 
 class LMCacheConnectorV1(KVConnectorBase_V1):
-    def __init__(self, vllm_config: "VllmConfig", role: KVConnectorRole):
-        super().__init__(vllm_config=vllm_config, role=role)
+    def __init__(
+        self,
+        vllm_config: "VllmConfig",
+        role: KVConnectorRole,
+        kv_cache_config: "KVCacheConfig",
```
heheda12345, line 34: @KuntaiDu do you need to change LMCache connector? kv_cache_config is not available in vllm config now.

heheda12345 (approved):
LGTM! Leave this PR to @KuntaiDu for compatibility with LMCache + HMA

vllm/distributed/kv_transfer/kv_connector/v1/lmcache_connector.py
KuntaiDu: I need to change LMCache PR correspondingly, but we can merge this PR first and I can change it in LMCache afterwards.

vllm/distributed/kv_transfer/kv_connector/factory.py
KuntaiDu: `KVCacheConfig` takes effect on almost all connector methods when HMA is enabled and it is not useful at all when HMA is disabled. 

That said, the current implementation looks good to me as it is simple enough.

KuntaiDu:
> LGTM! Leave this PR to @KuntaiDu for compatibility with LMCache + HMA

Thanks for reminder! I will fix LMCache compatibility after this PR is merged (still a lot of work on LMCache side before turning on HMA by default for LMCache).

PR #27896, opened 2025-10-31
[Perf] Move gc.freeze logic from EngineCoreProc to EngineCore for better coverage

Jialin:
## Purpose
For cases like external launcher / uniproc, they won't create EngineCoreProc. So moving gc.freeze from `EngineCoreProc.__init__` to `EngineCore.__init__` could allow more use cases got benefits from these optimizations / debug tools.

## Test Plan & Test Result
After moved, we've confirmed that GC2 cost is significantly reduced as the static objects are freeze (i.e. removed from GC2 collection list)

---
<details>

</details>

### DIFF
--- a/vllm/utils/gc_utils.py
+++ b/vllm/utils/gc_utils.py
@@ -89,6 +89,25 @@ class GCDebugger:
             )
 
 
+def freeze() -> None:
+    """
+    Freeze all objects tracked by the garbage collector. It should be invoked
+    after server init / warmup, to reduce GC overhead from static objects
+    during serving time.
+    """
+    # Move objects in permanent generation back to the oldest generation
+    # NOTE: This prevents cyclic references stay in the system when there're
+    # multiple freeze invocations.
+    gc.unfreeze()
+    # Invoke all GC collect for all generations to ensure objects are
+    # cyclic-refrenence-free before gc.freeze
+    gc.collect(0)
+    gc.collect(1)
+    gc.collect(2)
+    # Freeze all GC tracked objects
+    gc.freeze()
+
+
 def maybe_attach_gc_debug_callback() -> None:
     """
     Attached a callback for GC debug when VLLM_GC_DEBUG is enabled.
--- a/vllm/utils/gc_utils.py
+++ b/vllm/utils/gc_utils.py
@@ -89,7 +89,7 @@ class GCDebugger:
             )
 
 
-def freeze() -> None:
+def gc_freeze() -> None:
     """
     Freeze all objects tracked by the garbage collector. It should be invoked
     after server init / warmup, to reduce GC overhead from static objects
--- a/vllm/utils/gc_utils.py
+++ b/vllm/utils/gc_utils.py
@@ -97,10 +97,9 @@ def freeze_gc_heap() -> None:
     """
     # Ensure all static objects are pushed down to the oldest generation for
     # freeze
-    gc.collect()
-    # gc.collect(0)
-    # gc.collect(1)
-    # gc.collect(2)
+    gc.collect(0)
+    gc.collect(1)
+    gc.collect(2)
     # Freeze all GC tracked objects
     gc.freeze()
 
--- a/vllm/utils/gc_utils.py
+++ b/vllm/utils/gc_utils.py
@@ -89,24 +89,18 @@ class GCDebugger:
             )
 
 
-def reset_gc_heap() -> None:
-    """
-    Unfreeze and invokes GC collects for all generations to ensure objects
-    are cyclic-refrenence-free before gc.freeze.
-    """
-    gc.unfreeze()
-    gc.collect(0)
-    gc.collect(1)
-    gc.collect(2)
-
-
 def freeze_gc_heap() -> None:
     """
     Freeze all objects tracked by the garbage collector. It should be invoked
     after server init / warmup, to reduce GC overhead from static objects
     during serving time.
     """
-    reset_gc_heap()
+    # Ensure all static objects are pushed down to the oldest generation for
+    # freeze
+    gc.collect()
+    # gc.collect(0)
+    # gc.collect(1)
+    # gc.collect(2)
     # Freeze all GC tracked objects
     gc.freeze()
 
--- a/vllm/utils/gc_utils.py
+++ b/vllm/utils/gc_utils.py
@@ -89,29 +89,26 @@ class GCDebugger:
             )
 
 
+def reset_gc_heap() -> None:
+    """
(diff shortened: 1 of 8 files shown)


vllm/v1/engine/core.py
```diff
@@ -193,6 +193,18 @@ def __init__(
...
+        # Reduces pause times of oldest generation collections.
+        gc.collect()
+        gc.freeze()
+
+        # If enable, attach GC debugger after static variable freeze.
+        maybe_attach_gc_debug_callback()
+
+        # Enable environment variable cache (e.g. assume no more
+        # environment variable overrides after this point)
+        enable_envs_cache()
```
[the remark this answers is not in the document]
Jialin: Ideally, each process should only create one EngineCore, even if there're multiple instance, as it's happening during server initialization, so the extra overhead should be okay.

So will keep as is.

22quinn (approved):

zhuohan123 (approved):

vllm/v1/engine/core.py
```diff
@@ -214,6 +214,18 @@ def __init__(
             self.step if self.batch_queue is None else self.step_with_batch_queue
         )
 
+        # Mark the startup heap as static so that it's ignored by GC.
+        # Reduces pause times of oldest generation collections.
+        gc.collect()
+        gc.freeze()
```
njhill: We also do this in the api server here: https://github.com/vllm-project/vllm/blob/7956b0c0bca8c2b778e6a0b18953b5a08e136c90/vllm/entrypoints/openai/api_server.py#L156-L157

If someone is using the API server with multiprocessing disabled it might mean we freeze the heap too early.

It probably won't make much difference but just in-case we could add a `gc.unfreeze()` prior to the `gc.collect() / gc.freeze()` in the API server.

Jialin: Thanks @njhill for the suggestions. I didn't notice this separate freeze on API server side. Per my own understanding, the following 2 behavior is the same:
```
[Current]
# ... Initialization logic
gc.collect() 
gc.freeze()
# ... Other Initialization logic
gc.collect() 
gc.freeze()

[Suggestion per my interpretation]
# ... Initialization logic
gc.collect() 
gc.freeze()
# ... Other Initialization logic
gc.unfreeze()
gc.collect() 
gc.freeze()
```

I think the freezed objects should be exactly the same with and without the gc.unfreeze() call. Do you mind to elaborate a bit? I might miss some critical points or misunderstand your suggestions :P

Jialin: Thanks @njhill for the suggestion!

I ran into HBM OOM for one of the unit test, I believe that's due to missing gc.unfreeze(). I think the key difference of not having gc.unfreeze() is that, cycling reference will stay in permanent generations.

So I think it's safer to always call gc.unfreeze before gc.freeze.

I've moved the logic to a unify utils, so all the gc.freeze caller could migrate over.

vllm/utils/gc_utils.py
```diff
@@ -89,6 +89,25 @@ def handle(self, phase: str, info: dict[str, int]) -> None:
             )
 
 
+def gc_freeze() -> None:
```
njhill: Suggested rename for clarity
```suggestion
def freeze_gc_heap() -> None:
```

vllm/utils/gc_utils.py
```diff
@@ -89,6 +89,25 @@ def handle(self, phase: str, info: dict[str, int]) -> None:
...
+    """
+    # Move objects in permanent generation back to the oldest generation
+    # NOTE: This prevents cyclic references stay in the system when there're
+    # multiple freeze invocations.
+    gc.unfreeze()
+    # Invoke all GC collect for all generations to ensure objects are
+    # cyclic-refrenence-free before gc.freeze
+    gc.collect(0)
+    gc.collect(1)
+    gc.collect(2)
```
njhill: Is this not what's already done by `gc.collect()`?

vllm/utils/gc_utils.py
```diff
@@ -89,6 +89,25 @@ def handle(self, phase: str, info: dict[str, int]) -> None:
             )
 
 
+def gc_freeze() -> None:
+    """
+    Freeze all objects tracked by the garbage collector. It should be invoked
+    after server init / warmup, to reduce GC overhead from static objects
+    during serving time.
+    """
+    # Move objects in permanent generation back to the oldest generation
```
njhill:
```suggestion
    # Move objects in permanent generation back to the oldest generation
    # in case the heap is already frozen.
```

njhill (commented):
Thanks @Jialin

vllm/utils/gc_utils.py
Jialin: Updated.

vllm/utils/gc_utils.py
Jialin: Per gc documentation, I don't think they're done: https://docs.python.org/3/library/gc.html#gc.collect

gc.collect() is equivalent to gc.collect(2). So after gc.collect() + gc.freeze(), there might still be object stuck in generation 0 and generation 1.
```
gc.collect(generation=2)
Perform a collection. The optional argument generation may be an integer specifying which generation to collect (from 0 to 2). A [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) is raised if the generation number is invalid. The sum of collected objects and uncollectable objects is returned.
- Calling gc.collect(0) will perform a GC collection on the young generation.
- Calling gc.collect(1) will perform a GC collection on the young generation and an increment of the old generation.
- Calling gc.collect(2) or gc.collect() performs a full collection
```

vllm/utils/gc_utils.py
Jialin: Updated.

njhill:
@Jialin noticing the errors and thinking a bit more, we may want to be careful about this change since it could prevent resources from being cleaned up in cases where they would otherwise be expected to be.

The places where we currently do it aren't a problem since
- the API server is assumed to "own" the process that it runs in so it's not going to be shut down without the entire process going down
- the workers are internal processes and their lifecycle is similarly tied to the process they run in

But for example if you create an `LLM` object, the expectation is that it can later be garbage collected including all of the resources "within" it.

njhill:
Thanks @Jialin but I don't think we should be using `create_new_process_for_each_test` .. this is mostly a workaround for the fact that things aren't cleaned up that should be, to help with CI stability.

If folks create a vllm instance programmatically with python (i.e. `LLM` or `AsyncLLM`) then that should get garbage collected and fully cleaned up when it goes out of scope.

PR #25895, opened 2025-09-29
[Bugfix] Fix accuracy issue of TRTLLM FP8 MOE and improve logging

pavanimajety:
## Purpose
Fixes #25189: accuracy issues for TRTLLM DSR1 Latency kernels 
Bugs introduced in #23991 and #23640 

Also fixes incorrect logging prints for which kernels are used. When Flashinfer MOE is enabled, Deep Gemm is automatically disabled for SM100.

```
(EngineCore_DP0 pid=101213) (Worker_TP3 pid=101225) INFO 09-29 10:14:46 [fp8.py:454] Detected Blackwell GPUs, using FlashInfer TensorRT-LLM kernels for FP8 MOE.
(EngineCore_DP0 pid=101213) (Worker_TP3 pid=101225) INFO 09-29 10:14:46 [fp8.py:468] DeepGemm disabled: FlashInfer MOE is enabled.

```

## Test Plan
LM Eval with `VLLM_USE_FLASHINFER_MOE_FP8=1` and `VLLM_FLASHINFER_MOE_BACKEND="latency"`

```
root@umbriel-b200-017:/workspace/scratch-pmaj-1/gh-pm-vllm# ./run_lm_eval.sh                                                                        
+ export FORCE_NUM_KV_SPLITS=1                                                                                                                      
+ FORCE_NUM_KV_SPLITS=1                                                                                                                             
+ export VLLM_USE_FLASHINFER_MOE_FP8=1                                                                                                              
+ VLLM_USE_FLASHINFER_MOE_FP8=1                                                                                                                     
+ export VLLM_FLASHINFER_MOE_BACKEND=latency                                                                                                        
+ VLLM_FLASHINFER_MOE_BACKEND=latency                                                                                                               
+ echo 'Running lm_eval with:'                                                                                                                      
Running lm_eval with:                                                                                                                               
                     
  Model: /models/DSR1-FP8/models--deepseek-ai--DeepSeek-R1-0528/snapshots/4236a6af538feda4548eca9ab308586007567f52/                                 
  Tensor Parallel Size: 8                                                                                                                           
  Quantization: fp8                                                                                                                                 
  GPU Memory Utilization: 0.90                                                                                                                      
                                                                      
+ lm_eval --model vllm --model_args pretrained=/models/DSR1-FP8/models--deepseek-ai--DeepSeek-R1-0528/snapshots/4236a6af538feda4548eca9ab30858600756
[...]

PR #20087, opened 2025-06-25
 [Feature] Integrate SM100 DeepGEMM support

yewentao256:
## Purpose

DeepGemm is updating to v2.0, which includes a new implementation for SM100 that expects block FP8 scales in E8M0 format https://github.com/deepseek-ai/DeepGEMM/pull/112

Previous context: https://github.com/vllm-project/vllm/pull/19820

We add a wrapper to support both Hopper (1.x) and Blackwell (2.x) interfaces

### Unit Test

image

#### Qwen3

```bash
VLLM_USE_DEEP_GEMM=1 lm_eval   --model vllm   --model_args "pretrained=Qwen/Qwen3-30B-A3B-FP8,max_model_len=32768,enforce_eager=True"   --trust_remote_code   --tasks gsm8k   --num_fewshot 5   --batch_size auto
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8567|±  |0.0097|
|     |       |strict-match    |     5|exact_match|↑  |0.8893|±  |0.0086|

# No deepgemm
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8256|±  |0.0105|
|     |       |strict-match    |     5|exact_match|↑  |0.8878|±  |0.0087|
```

#### Qwen3 (DP+EP)

```bash
export VLLM_ALL2ALL_BACKEND="deepep_high_throughput"
VLLM_USE_DEEP_GEMM=1 lm_eval   --model vllm   --model_args "pretrained=Qwen/Qwen3-30B-A3B-FP8,data_parallel_size=2,max_model_len=32768,enable_expert_parallel=True,enforce_eager=True"   --trust_remote_code   --tasks gsm8k   --num_fewshot 5   --batch_size auto
|Tasks|Version|   Filter   |n-shot| Metric  |  |Value |  |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|   3|flexible-extract|   5|exact_match|↑ |0.8302|± |0.0103|
|   |    |strict-match  |   5|exact_match|↑ |0.8294|± |0.0104|
# No deepgemm
[...]

PR #27897, opened 2025-10-31
[Performance][B200] Fix deepgemm prologue

varun-sundar-rabindranath:
## Purpose
DeepGEMM requires the activation and weights scales to be in a specific format. When the scales are not provided in the desired format, DeepGEMM transforms the scales itself. This is usually very slow. 

On H100, DeepGEMM needs the scales to be ColumnMajor and in float32. 
On B200, DeepGEMM needs the scales to be ColumnMajor but in a packed E8M0 format. 

Note that we handle the H100 case already on `main`. This PR adds partial support for the B200 case. Concretely, 
 - Import `transform_sf_into_required_layout` from DeepGEMM to perform weight scales transformation. This function handles both SM90 and SM100 internally and can be used commonly. 
 - The DeepEP low latency dispatch supports dispatching the activation scales in packed E8M0 format. This PR enables that. 
 
 `main`:
main

 `PR`:
PR

## Benchmark

full benchmark numbers - [link](https://docs.google.com/spreadsheets/d/1mpKTZg8v-HTjiBzWTk1dj4jUmibsMYbE8pfEaSMoWPA/edit?usp=sharing) 

server command : `VLLM_ALL2ALL_BACKEND=${A2A} VLLM_USE_DEEP_GEMM=1 canhazgpu run -g2 --  vllm serve Qwen/Qwen3-30B-A3B-FP8 --trust-remote-code --tensor-parallel-size 1 --data-parallel-size 2 --enable-expert-parallel --no-enable-prefix-caching  --port 9010 `

Decode bench command : `vllm bench serve --model Qwen/Qwen3-30B-A3B-FP8 --dataset-name random --num-prompts 128 --random-input-len 1 --random-output-len 1024 --request-rate 128 --ignore-eos --port 9010`

Prefill bench command : ` vllm bench serve --model Qwen/Qwen3-30B-A3B-FP8 --dataset-name random --num-prompts 256 --random-input-len 8192 --random-output-len 1 --request-rate 256 --ignore-eos --port 9010 --backend vllm`

### B200 + deepep_low_latency + decode 
<google-sheets-html-origin><style type="text/css"></style>
  | main | PR
-- | -- | --
Peak output token throughput (tok/s) | 9600 | 12028

### B200 + deepep_high_throughput + prefill 
<google-sheets-html-origin><style type="text/css"></style>
  | main | PR
-- | -- | --
Total Token throughput (tok/s) | 89736.95 | 91310.99

### H100 + deepep_low_latency + decode 
<google-sheets-html-origin><style type="text/css"></style>
  | main | PR
-- | -- | --
Peak output token throughput (tok/s) | 8408 | 8422

### H100 + deepep_high_throughput + prefill
<google-sheets-html-origin><style type="text/css"></style>
  | main | PR
-- | -- | --
Total Token throughput (tok/s) | 73950.14 | 74095.51

## Test Plan

Server command : 
`vllm serve Qwen/Qwen3-30B-A3B-FP8 --trust-remote-code --tensor-parallel-size ${tp_size} --data-parallel-size ${dp_size} --enable-expert-parallel --no-enable-prefix-caching --port 9010`

Server config combinations (`VLLM_ALL2ALL_BACKEND`, `VLLM_USE_DEEP_GEMM`, `dp_size`, `tp_size`) , 
 1. (deepep_low_latency, 1, 2, 1)
 2. (deepep_high_throughput, 1, 2, 1)
 3. (deepep_low_latency, 0, 2, 1) // This PR touches fp8 weight loading when Deepgemm is enabled, this config tests regression
 4. (deepep_high_throughput, 0, 2, 1)
 5. (n/a, 1, 1, 2)
 6. (n/a, 0, 1, 2)
 
lm_eval command : 
```
lm_eval --model local-completions --tasks gsm8k --model_args model=Qwen/Qwen3-30B-A3B-FP8,base_url=http://localhost:9010/v1/completions,num_concurrent=30,max_retries=3 --limit 100
```

## Test Result
lm_eval produces desired results on the PR, on both H100 and B200. 
example of a desired result, 
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  | 0.87|±  |0.0338|
|     |       |strict-match    |     5|exact_match|↑  | 0.92|±  |0.0273|
```

### DIFF
--- a/vllm/model_executor/layers/fused_moe/batched_deep_gemm_moe.py
+++ b/vllm/model_executor/layers/fused_moe/batched_deep_gemm_moe.py
@@ -232,6 +232,7 @@ class BatchedDeepGemmExperts(mk.FusedMoEPermuteExpertsUnpermute):
         """
         super().__init__(quant_config)
         assert self.block_shape == get_mk_alignment_for_contiguous_layout()
+        assert self.quant_config.use_fp8_w8a8
         self.max_num_tokens = max_num_tokens
         self.num_dispatchers = num_dispatchers
 
@@ -249,6 +250,12 @@ class BatchedDeepGemmExperts(mk.FusedMoEPermuteExpertsUnpermute):
 
     def supports_expert_map(self) -> bool:
         return False
+
+    def supports_packed_ue8m0_act_scales(self) -> bool:
+        """
+        DeepGemm supports packed ue8m0 activation scales format in devices == sm100
+        """
+        return current_platform.is_device_capability(100)
 
     def finalize_weight_and_reduce_impl(self) -> mk.TopKWeightAndReduce:
         # Let PrepareAndFinalize::finalize() decide the impl.
--- a/vllm/model_executor/layers/fused_moe/deepep_ll_prepare_finalize.py
+++ b/vllm/model_executor/layers/fused_moe/deepep_ll_prepare_finalize.py
@@ -100,7 +100,7 @@ class DeepEPLLPrepareAndFinalize(mk.FusedMoEPrepareAndFinalize):
         # We don't have enough information to determine if we should dispatch
         # activation scales in a packed ue8m0 format during object construction
         # time. This setting is handled by post_init_setup.
-        self.use_ue8m0 = False
+        self.use_ue8m0_dispatch = False
 
     def post_init_setup(self, fused_experts: mk.FusedMoEPermuteExpertsUnpermute):
         if not fused_experts.supports_packed_ue8m0_act_scales():
@@ -111,7 +111,7 @@ class DeepEPLLPrepareAndFinalize(mk.FusedMoEPrepareAndFinalize):
             logger.debug_once(
                 "Update DeepEPLLPrepareFinalize to do packed ue8m0 scales dispatch."
             )
-            self.use_ue8m0 = True
+            self.use_ue8m0_dispatch = True
         else:
             logger.warning_once(
                 "DeepEPLLPrepareAndFinalize is setup to dispatch raw/unquantized "
@@ -233,8 +233,8 @@ class DeepEPLLPrepareAndFinalize(mk.FusedMoEPrepareAndFinalize):
             num_experts,
             use_fp8=self.use_fp8_dispatch,
             # round_scale needs to be set to dispatch in ue8m0
-            round_scale=self.use_ue8m0,
-            use_ue8m0=self.use_ue8m0,
+            round_scale=self.use_ue8m0_dispatch,
+            use_ue8m0=self.use_ue8m0_dispatch,
             async_finish=False,
             return_recv_hook=True,
         )
--- a/vllm/model_executor/layers/fused_moe/deepep_ll_prepare_finalize.py
+++ b/vllm/model_executor/layers/fused_moe/deepep_ll_prepare_finalize.py
@@ -6,6 +6,7 @@ import deep_ep
 import torch
 
 import vllm.model_executor.layers.fused_moe.modular_kernel as mk
+from vllm.logger import init_logger
 from vllm.model_executor.layers.fused_moe.config import FusedMoEQuantConfig
 from vllm.model_executor.layers.fused_moe.topk_weight_and_reduce import (
     TopKWeightAndReduceDelegate,
@@ -19,6 +20,8 @@ from vllm.v1.worker.ubatching import (
     dbo_enabled,
     dbo_maybe_run_recv_hook,
 )
+
+logger = init_logger(__name__)
 
 # DeepEP kernels quantize dispatch inputs in 128 element chunks.
 DEEPEP_QUANT_BLOCK_SIZE = 128
@@ -93,6 +96,28 @@ class DeepEPLLPrepareAndFinalize(mk.FusedMoEPrepareAndFinalize):
         # combine function.
         self.handles: list[tuple | None] = [None, None]
         self.num_dispatchers_ = num_dispatchers
+
+        # We don't have enough information to determine if we should dispatch
+        # activation scales in a packed ue8m0 format during object construction
+        # time. This setting is handled by setup_packed_ue8m0_scales_dispatch.
+        self.use_ue8m0 = False
+
+    def supports_packed_ue8m0_scales_dispatch(self) -> bool:
+        return True
+
+    def setup_packed_ue8m0_scales_dispatch(self) -> None:
+        if self.use_fp8_dispatch:
+            logger.debug_once(
+                "Update DeepEPLLPrepareFinalize to do packed ue8m0 scales dispatch"
+            )
+            self.use_ue8m0 = True
+        else:
+            logger.warning_once(
+                "Ignoring request to dispatch activation scales in a packed "
+                "ue8m0 format as DeepEPLLPrepareAndFinalize is setup to"
+                "dispatch raw/unquantized activations.",
+                scope="local",
+            )
 
     def num_dispatchers(self) -> int:
         return self.num_dispatchers_
@@ -206,6 +231,9 @@ class DeepEPLLPrepareAndFinalize(mk.FusedMoEPrepareAndFinalize):
             self.max_tokens_per_rank,
             num_experts,
             use_fp8=self.use_fp8_dispatch,
+            # round_scale needs to be set to dispatch in ue8m0
+            round_scale=self.use_ue8m0,
+            use_ue8m0=self.use_ue8m0,
             async_finish=False,
             return_recv_hook=True,
         )
--- a/vllm/model_executor/layers/fused_moe/deepep_ll_prepare_finalize.py
+++ b/vllm/model_executor/layers/fused_moe/deepep_ll_prepare_finalize.py
@@ -99,23 +99,24 @@ class DeepEPLLPrepareAndFinalize(mk.FusedMoEPrepareAndFinalize):
 
         # We don't have enough information to determine if we should dispatch
         # activation scales in a packed ue8m0 format during object construction
-        # time. This setting is handled by setup_packed_ue8m0_scales_dispatch.
+        # time. This setting is handled by post_init_setup.
         self.use_ue8m0 = False
 
-    def supports_packed_ue8m0_scales_dispatch(self) -> bool:
-        return True
+    def post_init_setup(self, fused_experts: mk.FusedMoEPermuteExpertsUnpermute):
+        if not fused_experts.supports_packed_ue8m0_act_scales():
+            # Early exit.
+            return
 
-    def setup_packed_ue8m0_scales_dispatch(self) -> None:
         if self.use_fp8_dispatch:
             logger.debug_once(
-                "Update DeepEPLLPrepareFinalize to do packed ue8m0 scales dispatch"
+                "Update DeepEPLLPrepareFinalize to do packed ue8m0 scales dispatch."
             )
             self.use_ue8m0 = True
         else:
             logger.warning_once(
-                "Ignoring request to dispatch activation scales in a packed "
-                "ue8m0 format as DeepEPLLPrepareAndFinalize is setup to"
-                "dispatch raw/unquantized activations.",
+                "DeepEPLLPrepareAndFinalize is setup to dispatch raw/unquantized "
+                f"activations despite ({fused_experts.__class__.__name__}) being able "
+                "to support quantized activations.",
                 scope="local",
             )
 
--- a/vllm/model_executor/layers/fused_moe/modular_kernel.py
+++ b/vllm/model_executor/layers/fused_moe/modular_kernel.py
@@ -347,6 +347,20 @@ class FusedMoEPrepareAndFinalize(ABC):
         """
         raise NotImplementedError
 
+    def supports_packed_ue8m0_scales_dispatch(self) -> bool:
+        """
+        Return true if the implementation can dispatch activation scales in
+        packed ue8m0 format.
+        """
+        return False
+
+    def setup_packed_ue8m0_scales_dispatch(self) -> None:
+        """
+        Setup internal state of the implementation to dispatch activation scales
+        in packed ue8m0 format.
+        """
+        raise NotImplementedError
+
 
 # TODO: add supported activations method (return string)
 class FusedMoEPermuteExpertsUnpermute(ABC):
@@ -502,6 +516,13 @@ class FusedMoEPermuteExpertsUnpermute(ABC):
         A flag indicating whether or not this class supports expert maps
         """
         raise NotImplementedError
+
+    def supports_packed_ue8m0_act_scales(self) -> bool:
+        """
+        A flag indicating whether or not this class can process packed ue8m0
+        activation scales.
+        """
+        return False
 
     def workspace_dtype(self, act_dtype: torch.dtype) -> torch.dtype:
         """
@@ -698,6 +719,8 @@ class FusedMoEModularKernel(torch.nn.Module):
         self.prepare_finalize = prepare_finalize
         self.fused_experts = fused_experts
         self.shared_experts = shared_experts
+
+        self._post_init_setup()
         assert (
             prepare_finalize.activation_format == fused_experts.activation_formats[0]
         ), (
@@ -706,6 +729,17 @@ class FusedMoEModularKernel(torch.nn.Module):
             f"{fused_experts.__class__.__name__}."
             f"{fused_experts.activation_formats[0]}"
         )
+
+    def _post_init_setup(self):
+        """
+        Resolve any leftover setup dependencies between self.prepare_finalize
+        and self.fused_experts here.
+        """
+        if (
+            self.fused_experts.supports_packed_ue8m0_act_scales()
+            and self.prepare_finalize.supports_packed_ue8m0_scales_dispatch()
+        ):
+            self.prepare_finalize.setup_packed_ue8m0_scales_dispatch()
 
     def supports_expert_map(self) -> bool:
         """
--- a/vllm/model_executor/layers/fused_moe/modular_kernel.py
+++ b/vllm/model_executor/layers/fused_moe/modular_kernel.py
@@ -149,6 +149,15 @@ class FusedMoEPrepareAndFinalize(ABC):
     described above.
     """
 
+    def post_init_setup(self, fused_experts: "FusedMoEPermuteExpertsUnpermute"):
+        """
+        Initialize FusedMoEPrepareAndFinalize settings that depend on
+        FusedMoEPermuteExpertsUnpermute experts object.
+        The FusedMoEPrepareAndFinalize implementations that have such
+        dependencies may choose to override this function.
+        """
+        return
+
     @abstractmethod
     def prepare(
         self,
@@ -344,20 +353,6 @@ class FusedMoEPrepareAndFinalize(ABC):
         """
         Indicates whether or not the output of finalize is reduced across all
         ranks.
-        """
-        raise NotImplementedError
-
-    def supports_packed_ue8m0_scales_dispatch(self) -> bool:
-        """
-        Return true if the implementation can dispatch activation scales in
-        packed ue8m0 format.
-        """
-        return False
-
-    def setup_packed_ue8m0_scales_dispatch(self) -> None:
-        """
-        Setup internal state of the implementation to dispatch activation scales
-        in packed ue8m0 format.
         """
         raise NotImplementedError
 
@@ -735,11 +730,7 @@ class FusedMoEModularKernel(torch.nn.Module):
         Resolve any leftover setup dependencies between self.prepare_finalize
         and self.fused_experts here.
         """
-        if (
-            self.fused_experts.supports_packed_ue8m0_act_scales()
-            and self.prepare_finalize.supports_packed_ue8m0_scales_dispatch()
-        ):
-            self.prepare_finalize.setup_packed_ue8m0_scales_dispatch()
+        self.prepare_finalize.post_init_setup(self.fused_experts)
 
     def supports_expert_map(self) -> bool:
         """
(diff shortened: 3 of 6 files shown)


vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -267,6 +267,13 @@ def init_prepare_finalize(self, layer: torch.nn.Module):
             )
             self.topk_indices_dtype = prepare_finalize.topk_indices_dtype()
             experts = self.select_gemm_impl(prepare_finalize, layer)
+
```
varun-sundar-rabindranath: @bnellnm  I added,
`supports_packed_ue8m0_scales_dispatch` and `setup_packed_ue8m0_scales_dispatch` to the `FusedMoEPrepareAndFinalize` base class and 
`supports_packed_ue8m0_act_scales` to the `FusedMoEPermuteExpertsUnpermute` base class.
The motivation is that, we need the prepare_finalize implementation to dispatch in "packed ue8m0" format for `BatchedDeepGemmExperts`. The issue is that we don't know what experts we'll use during prepare_finalize construction. I have resorted to the following if-statement to facilitate that. PTAL. 

Also, is there a better way ?

vllm/model_executor/layers/quantization/fp8.py
```diff
@@ -918,15 +916,23 @@ def process_weights_after_loading(self, layer: Module) -> None:
 
             # DeepGemm scales need to be transposed and aligned. We try to do
             # it ahead of time for performance reasons.
```
varun-sundar-rabindranath, line 846: cc @yewentao256  for the changes to this file. 
I have replaced hopper specific `get_col_major_tma_aligned_tensor` with a generic (h100 and b200)`transform_sf_into_required_layout` utility from deepgemm. PTAL! Thanks 🙌

vllm/model_executor/layers/fused_moe/batched_deep_gemm_moe.py
```diff
@@ -249,6 +250,12 @@ def supports_chunking(self) -> bool:
     def supports_expert_map(self) -> bool:
         return False
 
+    def supports_packed_ue8m0_act_scales(self) -> bool:
+        """
+        DeepGemm supports packed ue8m0 activation scales format in devices >= sm100
+        """
+        return current_platform.is_device_capability(100)
```
mgoin, line 258: The comment doesn't match this line since "is" is ==
Also isn't it the case though that we still want to use UE8M0 on hopper for cases like DeepSeek terminus?

mgoin (commented):
LGTM nice find. I just have the concern about applying the right ue8m0 format for both Hopper and Blackwell if the model requires it

vllm/model_executor/layers/fused_moe/layer.py
bnellnm: I suppose you could have a more generic "setup" method that is always called and is a nop by default, e.g.
```
prepare_finalize.setup_stuff(experts)  # needs better name
```
I think it would also make sense to move this inside `FusedMoEModularKernel.__init__` so that it will be invoked no matter where the modular kernel is constructed.

varun-sundar-rabindranath: I like the idea of moving it into the FusedMoEModularKernel.__init__  - I'll do that 👍 Thanks.

vllm/model_executor/layers/fused_moe/batched_deep_gemm_moe.py
yewentao256: +1, actually we are using e8m0 for hopper currently, this seems a breaking change for me.
We should carefully test and benchmark before we use this.

vllm/model_executor/layers/fused_moe/deepep_ll_prepare_finalize.py
```diff
@@ -94,6 +97,27 @@ def __init__(
...
+            logger.debug_once(
+                "Update DeepEPLLPrepareFinalize to do packed ue8m0 scales dispatch"
+            )
+            self.use_ue8m0 = True
+        else:
+            logger.warning_once(
+                "Ignoring request to dispatch activation scales in a packed "
+                "ue8m0 format as DeepEPLLPrepareAndFinalize is setup to"
+                "dispatch raw/unquantized activations."
+            )
```
yewentao256, line 121:
```suggestion
            logger.warning_once(
                "Ignoring request to dispatch activation scales in a packed "
                "ue8m0 format as DeepEPLLPrepareAndFinalize is setup to"
                "dispatch raw/unquantized activations.",
                scope="local"
            )
```

vllm/model_executor/layers/quantization/fp8.py
```diff
@@ -921,15 +916,23 @@ def process_weights_after_loading(self, layer: Module) -> None:
...
-                if expert_weight_is_col_major(layer.w13_weight_scale_inv):
-                    layer.w13_weight_scale_inv = get_col_major_tma_aligned_tensor(
-                        layer.w13_weight_scale_inv
+            if self.allow_deep_gemm:
+                layer.w13_weight.data, layer.w13_weight_scale_inv.data = (
+                    deepgemm_transform_sf_into_required_layout(
+                        layer.w13_weight.data,
+                        layer.w13_weight_scale_inv.data,
+                        is_weights=True,
+                        use_e8m0=is_deep_gemm_e8m0_used(),
```
yewentao256: Will this function call `requant_weight_ue8m0_inplace` again inside?

vllm/model_executor/layers/quantization/fp8.py
```diff
@@ -921,15 +916,23 @@ def process_weights_after_loading(self, layer: Module) -> None:
...
-                    layer.w2_weight_scale_inv = get_col_major_tma_aligned_tensor(
-                        layer.w2_weight_scale_inv
+                )
+                layer.w2_weight.data, layer.w2_weight_scale_inv.data = (
+                    deepgemm_transform_sf_into_required_layout(
+                        layer.w2_weight.data,
+                        layer.w2_weight_scale_inv.data,
+                        is_weights=True,
+                        use_e8m0=is_deep_gemm_e8m0_used(),
                     )
```
yewentao256: Same here

vllm/model_executor/layers/quantization/fp8.py
yewentao256: So the main change is that we convert `get_col_major_tma_aligned_tensor ` to `transform_sf_into_required_layout`? What is the difference of these and why could we get perf?

vllm/utils/deep_gemm.py
```diff
@@ -49,10 +49,6 @@ def is_deep_gemm_e8m0_used() -> bool:
         logger.info_once("DeepGEMM E8M0 disabled: _fp8_gemm_nt_impl not found")
         return False
 
-    if envs.VLLM_USE_FLASHINFER_MOE_FP8:
-        logger.info_once("DeepGEMM E8M0 disabled: FlashInfer MOE is enabled.")
-        return False
```
yewentao256, line 54: I am not sure who adds this before, could you take a further look?

vllm/model_executor/layers/quantization/fp8.py
```diff
@@ -1112,7 +1090,8 @@ def select_gemm_impl(
         layer: torch.nn.Module,
```
yewentao256, line 1037: If `BatchedTritonOrDeepGemmExperts` is not called elsewhere, could we delete the whole class?

yewentao256 (commented):
Nice find and great performance improvement! Thanks for the work!
A few thoughts

vllm/model_executor/layers/quantization/fp8.py
varun-sundar-rabindranath: the main difference is `get_col_major_tma_aligned_tensor` is Hopper specific and `transform_sf_into_required_layout` works for both Hopper and Blackwell. See  https://github.com/deepseek-ai/DeepGEMM/blob/c9f8b34dcdacc20aa746b786f983492c51072870/csrc/apis/layout.hpp#L8

The perf comes from the fact that on blackwell we were not transforming the weight scales during model weight setup -- Now we do (via transform_sf_into_required_layout) it in model weight setup and deepgemm doesn't have to do it in every call.

vllm/model_executor/layers/quantization/fp8.py
varun-sundar-rabindranath: Yes.

vllm/model_executor/layers/quantization/fp8.py
varun-sundar-rabindranath: Yes.

vllm/model_executor/layers/quantization/utils/fp8_utils.py
```diff
@@ -938,6 +939,49 @@ def requant_weight_ue8m0_inplace(
...
+    assert xq.dtype == torch.float8_e4m3fn, (
+        "Expected quantized tensor dtype "
+        f"to be torch.float8_e4m3fn, got {xq.dtype} instead."
+    )
+    assert xs.dtype == torch.float32, (
+        f"Expected tensor scales dtype to be torch.float32, got {xs.dtype} instead"
+    )
+
+    if use_e8m0:
+        requant_weight_ue8m0_inplace(xq, xs)
```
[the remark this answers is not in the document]
varun-sundar-rabindranath: fixed.

vllm/model_executor/layers/fused_moe/layer.py
varun-sundar-rabindranath: Fixed 🙌

vllm/utils/deep_gemm.py
pavanimajety: May I know why is this removed? Is this because of MOE vs Gemm impl differences?
DeepGemm seems to always be enabled even when other MOE backends are enabled. We need to have a better check to identify the moe backend.

pavanimajety: For example to run Flashinfer MOE we now need to run:
`VLLM_USE_FLASHINFER_MOE_FP8=1 VLLM_FLASHINFER_MOE_BACKEND=latency VLLM_USE_DEEP_GEMM=0 python .... `

varun-sundar-rabindranath: Looks like https://github.com/vllm-project/vllm/pull/25895/ adds it. @pavanimajety  can you please take a look. Thanks.

vllm/model_executor/layers/quantization/fp8.py
varun-sundar-rabindranath: it is still used in compressed_tensors_moe.py `https://github.com/vllm-project/vllm/blob/2c19d96777939dd3473eabfacbe1948a3ea0b4be/vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors_moe.py#L994`

vllm/model_executor/layers/quantization/fp8.py
```diff
@@ -1065,31 +1079,6 @@ def process_weights_after_loading(self, layer: Module) -> None:
             del layer.w13_input_scale
             del layer.w2_input_scale
 
-        if is_deep_gemm_e8m0_used() and self.block_quant:
```
varun-sundar-rabindranath, line 993: Here we perform weight requant and weight scale transformation based on `is_deep_gemm_e8m0_used() and self.block_quant` - However, this does not consider what `Fp8MoeBackend` is used. i.e. regardless of the backend, which could be,
```
    FLASHINFER_TRTLLM = 1
    FLASHINFER_CUTLASS = 2
    DEEPGEMM = 3
    CUTLASS_BLOCK_SCALED_GROUPED_GEMM = 4
    MARLIN = 5
    TRITON = 6
``` 
we perform weight requant and scales transform if DeepGEMM is available. This seems like a bug and I have moved this logic above and guarded with the `self.allow_deep_gemm` flag that is True only when the FP8MoeBackend is DEEPGEMM.

@yewentao256 - This block was first introduced in https://github.com/vllm-project/vllm/pull/20087 . Can you confirm if this is okay. Thanks.

vllm/utils/deep_gemm.py
varun-sundar-rabindranath: @pavanimajety  sorry i missed your comment. 

>May I know why is this removed? Is this because of MOE vs Gemm impl differences?

I removed it in an effort to cleanup. I think this function should depend only deepgemm specific attributes / envs. 

>DeepGemm seems to always be enabled even when other MOE backends are enabled. We need to have a better check to identify the moe backend.

Yes, this is a problem. I have addressed this is fp8.py. Please take a look at  comment https://github.com/vllm-project/vllm/pull/27897/files#r2487520906 .

varun-sundar-rabindranath: I tried running, 
```
VLLM_USE_FLASHINFER_MOE_FP8=1 VLLM_FLASHINFER_MOE_BACKEND="latency"  canhazgpu run -g8 --  lm_eval --model vllm --model_args pretrained=deepseek-ai/DeepSeek-R1,quantization=fp8,tensor_parallel_size=8,gpu_memory_utilization=0.90,add_bos_token=True --gen_kwargs temperature=0.0,max_gen_toks=32768 --trust_remote_code --tasks gsm8k --num_fewshot 5 --batch_size 200 --limit 1319
```
from #25895  and got, 

```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9613|±  |0.0053|
|     |       |strict-match    |     5|exact_match|↑  |0.9621|±  |0.0053|
```

vllm/model_executor/layers/fused_moe/batched_deep_gemm_moe.py
varun-sundar-rabindranath: >The comment doesn't match this line since "is" is ==
Updated the comment to == sm100, since deepgemm readme specifies sm100 explicitly. We can upgrade as needed. 

> Also isn't it the case though that we still want to use UE8M0 on hopper for cases like DeepSeek terminus?

IIUC, this the state of `main` : 

let `ws` be a weight scales tensor of shape `[X, 4096]` and datatype `float32`

- on Hopper and Blackwell - When we use DeepGemm, we always (for block fp8 models) cast the weight scales to UE8M0. ~but keep the weight scales in `float32`. i.e. each float32 value actually holds UE8M0 content. Look [here](https://github.com/vllm-project/vllm/blob/145c00a4d32b7a681f7fb936c9575812c7aa7880/vllm/utils/deep_gemm.py#L304). i.e. only the first byte of each float32 value will have the actual contents.~ 

[EDIT]The stricken out portion was wrong. We actually cast the weights to ue8m0 and then expand it back to float32 - effectively the scale values can be one of `{2^i where i in [-127, 127]}`

`ws` will be of shape `[X, 4096]` and of datatype `float32`. 

This PR: 
- on Hopper  - We don't change the behaviour on Hopper.
- on Blackwell - We requant to `UE8M0` and then we use the `transform_sf_into_required_layout()` from deepgemm to **pack** the scales into an int32 tensor. i.e. `ws` will be of shape `[x, 1024]` and of datatype `int32`.  Effectively the scale values can be one of `{i where in [-127, 127]}`

> +1, actually we are using e8m0 for hopper currently, this seems a breaking change for me.
We should carefully test and benchmark before we use this.
@yewentao256  I have added some benchmark and lm-eval numbers in the PR description.

vllm/utils/deep_gemm.py
pavanimajety: Thanks for testing Varun, I added the check in https://github.com/vllm-project/vllm/pull/25895 because we see incorrect logs and unrequired tuning when flashinfer MoE is enabled but DeepGemm is assumed default anyway.

vllm/model_executor/layers/fused_moe/modular_kernel.py
```diff
@@ -707,6 +730,17 @@ def __init__(
             f"{fused_experts.activation_formats[0]}"
         )
 
+    def _post_init_setup(self):
```
bnellnm, line 728: Can this method be overridden by subclasses?  If so, then you could avoid adding `ue8m0` related methods to all the modular kernels and make an implementation specific to `DeepEPLL...`/`BatchedDeepGemmExperts`

varun-sundar-rabindranath: I am not sure I understand. `_post_init_setup` is in FusedMoEModularKernel class which doesn't have any subclasses. can you clarify. Thanks.

bnellnm: I guess I was thinking you could have a generic `_post_init_setup` sort of method on prepare_finalize so that we wouldn't need to put such specific methods/logic here, e.g.
```
def _post_init_setup(self):
    self.prepare_finalize._post_init_setup(self.fused_experts)
```
And `_post_init_setup` will be a nop for everything except for `DeepEPLL...`

vllm/model_executor/layers/quantization/fp8.py
yewentao256: Make sense to me, where is the code you have for the guard?

vllm/model_executor/layers/fused_moe/deepep_ll_prepare_finalize.py
```diff
@@ -94,6 +97,28 @@ def __init__(
         self.handles: list[tuple | None] = [None, None]
         self.num_dispatchers_ = num_dispatchers
 
+        # We don't have enough information to determine if we should dispatch
+        # activation scales in a packed ue8m0 format during object construction
+        # time. This setting is handled by setup_packed_ue8m0_scales_dispatch.
+        self.use_ue8m0 = False
```
yewentao256: So is this flag only be used in low latency dispatch, and weight requant doesn't require this ?

yewentao256 (commented):
The idea looks good to me, could you also update all of the deepgemm unit tests accordingly?

vllm/model_executor/layers/quantization/fp8.py
varun-sundar-rabindranath: @yewentao256 The guard is in line 919 in this PR.
```
            if self.allow_deep_gemm:
```
note that `allow_deep_gemm` is true only if the backend selection chooses `DEEPGEMM`

vllm/model_executor/layers/fused_moe/deepep_ll_prepare_finalize.py
varun-sundar-rabindranath: Yes. only the low_latency dispatch exposes this option. 
>and weight requant doesn't require this ?

the `transform_sf_into_required_layout` function from deepgemm does this automatically when run on sm100.

vllm/model_executor/layers/fused_moe/modular_kernel.py
varun-sundar-rabindranath: Yes. This is a good idea. I have made the change. Thanks 🙌

vllm/model_executor/layers/quantization/fp8.py
```diff
@@ -1128,20 +1118,24 @@ def select_gemm_impl(
         ):
             max_num_tokens_per_rank = prepare_finalize.max_num_tokens_per_rank()
             assert max_num_tokens_per_rank is not None
+
+            experts_impl = (
+                BatchedDeepGemmExperts if self.allow_deep_gemm else BatchedTritonExperts
```
bnellnm, line 1048: Should we ditch using `BatchedTritonExperts` here?  There was some talk about getting rid of them since DeepGEMM is now enabled by default.

varun-sundar-rabindranath: `BatchedTritonExperts` is the alternative when DeepGEMM is not installed. 

> There was some talk about getting rid of them since DeepGEMM is now enabled by default.

have we added deepgemm to the requirements.txts ?

bnellnm: No, it's in the docker though.  I guess we can hold off on removing it.

yewentao256 (commented):
Please fix the conflicts and pre-commit issues so that we could land

bnellnm (approved):

vllm/model_executor/layers/fused_moe/deepep_ll_prepare_finalize.py
yewentao256: Could we rename to `use_ue8m0_dispatch`? In this case we can avoid mixture with e8m0 with scales

vllm/model_executor/layers/fused_moe/batched_deep_gemm_moe.py
mgoin: Okay so Blackwell just has the packing part specifically, understood

vllm/model_executor/layers/quantization/utils/fp8_utils.py
```diff
@@ -1141,11 +1186,15 @@ def maybe_post_process_fp8_weight_block(layer: torch.nn.Module):
...
-            layer.weight.data, layer.weight_scale.data, block_sz
+    if should_use_deepgemm:
+        dg_weight, dg_weight_scale = deepgemm_post_process_fp8_weight_block(
+            wq=layer.weight.data,
+            ws=layer.weight_scale.data,
+            quant_block_shape=tuple(layer.weight_block_size),
+            use_e8m0=is_deep_gemm_e8m0_used(),
         )
+        layer.weight = torch.nn.Parameter(dg_weight, requires_grad=False)
+        layer.weight_scale = torch.nn.Parameter(dg_weight_scale, requires_grad=False)
```
mgoin, line 1197: I think we want to preserve the attributes on the original parameter cc @kylesayrs

kylesayrs: Yes, without the original attributes we won't be able to reload weights. More changes than this will be required to support reloading, so this is fine to land now and rebase later.

varun-sundar-rabindranath: Should I just `parameter.data.copy_(new_tensor)` to avoid any unintended effects ?

PR #26467, opened 2025-10-09
[BugFix] Make penalties and bad_words work with async scheduling

njhill:
This is a minimally-invasive fix for compatibility of the current async scheduling implementation with built-in sampling parameters (specifically penalties and bad_words).

We plan to significantly refactor this but would like to first fix existing correctness issues so that async scheduling can be enabled by default.

PR #27910, opened 2025-11-01
[BugFix] Fix mixed penalties batch with async scheduling

njhill:
https://github.com/vllm-project/vllm/pull/26467 fixed compatibility of penalties sampling parameters with async scheduling but has a flaw that it breaks in cases where there is a mix of requests with and without penalties in the batch, specifically if a request with a penalties param starts while a batch without penalties is already running.

This is a fix for that case.

### DIFF
--- a/vllm/v1/sample/ops/penalties.py
+++ b/vllm/v1/sample/ops/penalties.py
@@ -21,6 +21,14 @@ def apply_all_penalties(
     """
     _, vocab_size = logits.shape
     output_tokens_t = _convert_to_tensors(output_token_ids, vocab_size, logits.device)
+
+    # In the async scheduling case, rows that won't have penalties applied may contain
+    # -1 placeholder token ids. We must replace these with valid token ids so that the
+    # scatter done in apply_penalties is valid.
+    # NOTE(nick): The penalties implementation is currently quite inefficient and
+    # will be reworked anyhow.
+    output_tokens_t.masked_fill_(output_tokens_t == -1, vocab_size)
+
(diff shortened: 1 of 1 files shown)


mgoin (approved):

vllm/v1/sample/ops/penalties.py
```diff
@@ -21,6 +21,14 @@ def apply_all_penalties(
     """
     _, vocab_size = logits.shape
     output_tokens_t = _convert_to_tensors(output_token_ids, vocab_size, logits.device)
+
+    # In the async scheduling case, rows that won't have penalties applied may contain
+    # -1 placeholder token ids. We must replace these with valid token ids so that the
+    # scatter done in apply_penalties is valid.
+    # NOTE(nick): The penalties implementation is currently quite inefficient and
+    # will be reworked anyhow.
+    output_tokens_t.masked_fill_(output_tokens_t == -1, vocab_size)
```
hidva, line 30: Hi, I’d like to ask why the actual draft token isn’t used here to replace the placeholder?  Thanks.

@njhill

njhill: @hidva this line should only apply to requests/rows in the batch which don't require the output tokens (i.e. those which don't use penalties sampling parameters). The other rows should not contain any placeholder tokens at this point.

Also async scheduling + spec decode + penalties isn't yet supported (any help with that appreciated though - see discussion in https://github.com/vllm-project/vllm/pull/30122).

PR #27922, opened 2025-11-01
[AsyncScheduling] Don't schedule past request max_tokens

njhill:
Async scheduling was introduced in https://github.com/vllm-project/vllm/pull/19970 which included [this](https://github.com/vllm-project/vllm/blob/99d69af9ece094acb94901439925f8468b32326a/vllm/v1/core/sched/scheduler.py#L253-L254) comment, but there wasn't a corresponding length check.

Without this change we're scheduling one more step than necessary for requests that finish due to max_tokens length.

Thanks to @WoosukKwon for making me aware of this!


WoosukKwon (approved):
LGTM! It'd be nice to have a unit test for this!

njhill:
The spec decoding CI error is related, am looking into it.

njhill:
Remaining test failures are also failing elsewhere :(

PR #27942, opened 2025-11-02
[Metrics] [KVConnector] Add Offloading Connector metrics

omerpaz95:
Added queries and hits metrics for the Offloading Connector.
Also added timing metrics for store and load operations, which take the average time it takes to load/store, per-token.
The metrics are available from Prometheus and from the StatLogger.
## Purpose
Allows collection of timing metrics for the Offloading Connector, which is essential for future development.
@orozery please review.
## Test Result

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit c035d2f8f8b879dcd07eac2231070d023364b80a. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Introduces metrics and instrumentation for KV offloading transfers.
> 
> - Adds `OffloadingConnectorStats` and `OffloadPromMetrics` with histograms, gauges, and counters for `CPU_to_GPU` and `GPU_to_CPU` transfer size and throughput
> - Instruments worker: `TransferResult` now includes job id, success, num blocks, duration, and transfer type; CUDA start/end events capture timings; results aggregated per request to bytes via computed `bytes_per_block`
> - Extends APIs: `LoadStoreSpec.num_blocks` and implementation in `BlockIDsLoadStoreSpec`; `OffloadingConnector` now requires `kv_cache_config` for WORKER, computes `bytes_per_block`, exposes `get_kv_connector_stats`, and builders for stats/Prom metrics
> - Updates handlers to validate `BlockIDsLoadStoreSpec` usage and record transfer types; scheduler/worker paths unchanged functionally aside from stats collection
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1981e157ad162913a5346c389c0d2cbb596bd9ee. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit 4a67a8038b98546ac7a0bf4283cdd1e4aa7055eb. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Introduces per-transfer KV offloading observability and minimal API changes to support it.
> 
> - Adds `OffloadingConnectorStats` and `OffloadPromMetrics` with histograms, gauges, and counters for `CPU_to_GPU`/`GPU_to_CPU` transfer size and throughput; exposes `get_kv_connector_stats` plus builders on the connector
> - Instruments worker/handlers: capture CUDA start/end events, enrich `TransferResult` with job id, success, block count, duration, and transfer type; aggregate to bytes via computed `bytes_per_block`; validate `BlockIDsLoadStoreSpec` usage
> - Extends API: `LoadStoreSpec.num_blocks` and implementation in `BlockIDsLoadStoreSpec`; `OffloadingConnector` WORKER now requires `kv_cache_config` to compute `bytes_per_block`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4a67a8038b98546ac7a0bf4283cdd1e4aa7055eb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


vllm/v1/metrics/loggers.py
```diff
@@ -530,6 +530,57 @@ def __init__(
             counter_connector_prefix_cache_hits, engine_indexes, model_name
         )
 
+        #
+        # Offloading connector
```
markmc: Connectors can now add their own metrics via the `build_prom_metrics()` added in #26811

markmc (changes requested):

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -408,8 +550,10 @@ def __init__(self, spec: OffloadingSpec):
 
         self._job_counter = 0
 
-        # req_id -> (job_id, store)
-        self._jobs: dict[int, tuple[ReqId, bool]] = {}
+        self._timing_stats = OffloadTiming()
```
orozery: I don't think you need `OffloadTiming`.
I would just do:
`self.xfer_stats = OffloadingConnectorStats()`

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -429,15 +573,23 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
     def start_load_kv(self, metadata: OffloadingConnectorMetadata):
         for req_id, transfer_spec in metadata.reqs_to_load.items():
             job_id = self._generate_job_id()
-            self._jobs[job_id] = (req_id, False)
+            current_time = time.perf_counter()
+            src_spec, dst_spec = transfer_spec
+            assert isinstance(src_spec, BlockIDsLoadStoreSpec)
+            num_blocks = len(src_spec.block_ids)
+            self._jobs[job_id] = (req_id, False, current_time, num_blocks)
```
orozery: I would prefer if we add `LoadStoreSpec.size_bytes()`, and then track size in bytes instead of num_blocks.

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -429,15 +573,23 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
...
 
     def start_store_kv(self, metadata: OffloadingConnectorMetadata):
         for req_id, transfer_spec in metadata.reqs_to_store.items():
             job_id = self._generate_job_id()
-            self._jobs[job_id] = (req_id, True)
+            current_time = time.perf_counter()
+            src_spec, dst_spec = transfer_spec
+            assert isinstance(dst_spec, BlockIDsLoadStoreSpec)
+            num_blocks = len(dst_spec.block_ids)
+            self._jobs[job_id] = (req_id, True, current_time, num_blocks)
```
orozery: Same as above.

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -456,7 +608,10 @@ def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
         for job_id, success in self.worker.get_finished():
             # we currently do not support job failures
             assert success
-            req_id, store = self._jobs.pop(job_id)
+            req_id, store, start_time, num_blocks = self._jobs.pop(job_id)
+            self._timing_stats.record_time(
+                (time.perf_counter() - start_time) / num_blocks, store
+            )
```
orozery: Record time and num_blocks (size_bytes) in separate fields

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -502,3 +657,69 @@ def yield_req_data(
...
+        counter_offload_block_hits = self._counter_cls(
+            name="vllm:offload_connector_block_hits",
+            documentation="Number of blocks queried and hit by Offloading"
+            " KV Connector.",
+            labelnames=labelnames,
+        )
+        self.counter_offload_block_hits = self.make_per_engine(
+            counter_offload_block_hits
+        )
+
```
orozery: I think these field are redundant with respect to:
`vllm:external_prefix_cache_hits`
`vllm:external_prefix_cache_queries`.

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -502,3 +657,69 @@ def yield_req_data(
...
+        gauge_kv_store_time_avg = self._gauge_cls(
+            name="vllm:offload_connector_kv_store_time",
+            documentation="Average time it takes to store a token to "
+            "offloaded KV memory",
+            multiprocess_mode="mostrecent",
+            labelnames=labelnames,
+        )
+
+        self.gauge_kv_store_time_avg = self.make_per_engine(gauge_kv_store_time_avg)
```
orozery: I think a histogram would be better tan a gauge (see `nixl_histogram_xfer_time` as a reference).

I would define 4 histograms:
1. throughput load
2. throughput store
3. store byte size
4. load byte size

Maybe also 2 counters:
1. num stores
2. num loads

Lastly, I would rename keys to start with `vllm:kv_offload_` instead of `vllm:offload_connector_`.

orozery (changes requested):
Looks good overall. Thanks!!

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -26,16 +33,79 @@
...
+    TransferType,
+)
 from vllm.v1.outputs import KVConnectorOutput
 from vllm.v1.request import Request
 
 ReqId = str
 
 logger = init_logger(__name__)
 
+THRESHOLD = 100000000000
```
sagearc: Can we remove this? It doesn’t seem to be referenced anywhere else

sagearc (changes requested):

vllm/v1/kv_offload/abstract.py
```diff
@@ -49,6 +49,14 @@ def medium() -> str:
...
         pass
 
+    @property
+    @abstractmethod
+    def num_blocks(self) -> int:
+        """
+        Returns number of blocks to load/store for this spec.
+        """
+        pass
+
```
orozery: this can be removed (along with all other num_blocks related additions)

vllm/v1/kv_offload/worker/cpu_gpu.py
```diff
@@ -152,29 +175,54 @@ def transfer_async(self, job_id: int, transfer_spec: TransferSpec) -> bool:
...
-            del self._transfer_events[job_id]
+        while self._transfers and self._transfers[0].end_event.query():
+            transfer = self._transfers.popleft()
+            transfer_time = transfer.start_event.elapsed_time(transfer.end_event)
+            result = TransferResult(
+                transfer.job_id,
+                True,
+                transfer.num_blocks,
+                transfer_time,
+                transfer.transfer_type,
```
orozery: Let's remove the `transfer_type` field from the `Transfer` dataclass.
Instead, introduce `self.transfer_type` in `SingleDirectionOffloadingHandler.__init__`.

vllm/v1/kv_offload/worker/cpu_gpu.py
```diff
@@ -152,29 +175,54 @@ def transfer_async(self, job_id: int, transfer_spec: TransferSpec) -> bool:
...
+            del self._transfer_events[transfer.job_id]
         return results
 
     def wait(self, job_ids: set[int]):
         for job_id in job_ids:
-            event = self._transfer_events.get(job_id)
-            if event is not None:
-                event.synchronize()
+            events = self._transfer_events.get(job_id)
+            if events:
```
orozery: Just to be on the safe side for any future modifications of `self._transfer_events`:
```suggestion
            if events is not None:
```

vllm/v1/kv_offload/worker/cpu_gpu.py
```diff
@@ -152,29 +175,54 @@ def transfer_async(self, job_id: int, transfer_spec: TransferSpec) -> bool:
...
+                start_event, end_event = events
+                start_event.synchronize()
```
orozery: This sync on the `start_event` is unnecessary, as it is covered by the sync of `end_event`.

vllm/v1/kv_offload/worker/worker.py
```diff
@@ -137,27 +144,30 @@ def transfer_async(self, job_id: int, spec: TransferSpec) -> bool:
...
 
-    def wait(self, job_ids: set[int]) -> None:
-        """
-        Wait for jobs to finish (blocking).
-
-        Args:
-            job_ids: The set of job IDs to wait for.
-        """
-        for handler in self.handlers:
-            handler.wait(job_ids)
```
orozery, line 163: looks like you accidentally dropped this function while rebasing.

vllm/v1/kv_offload/worker/worker.py
```diff
@@ -9,12 +10,26 @@
...
 logger = init_logger(__name__)
 
 
+@dataclass
+class TransferStats:
+    num_blocks: int
+    time: float  # Can be start_time or duration
+    transfer_type: TransferType
+
+
```
orozery: We can remove this class, as all is returned by `TransferResult`.

vllm/v1/kv_offload/worker/cpu_gpu.py
```diff
@@ -152,29 +175,54 @@ def transfer_async(self, job_id: int, transfer_spec: TransferSpec) -> bool:
...
 
-        self._transfer_events[job_id] = event
-        self._transfers.append((job_id, stream, event))
+        self._transfer_events[job_id] = (start_event, end_event)
+        transfer_type: TransferType = (
+            transfer_spec[0].medium(),
+            transfer_spec[1].medium(),
+        )
+        self._transfers.append(
+            Transfer(
```
orozery, line 182: can you use explicit kwargs?
i.e. `Transfer(job_id=job_id, ...)`

vllm/v1/kv_offload/worker/cpu_gpu.py
```diff
@@ -152,29 +175,54 @@ def transfer_async(self, job_id: int, transfer_spec: TransferSpec) -> bool:
...
-        while self._transfers and self._transfers[0][2].query():
-            job_id, stream, event = self._transfers.popleft()
-            results.append((job_id, True))
-            self._stream_pool.append(stream)
-            self._event_pool.append(event)
-            del self._transfer_events[job_id]
+        while self._transfers and self._transfers[0].end_event.query():
+            transfer = self._transfers.popleft()
+            transfer_time = transfer.start_event.elapsed_time(transfer.end_event)
+            result = TransferResult(
```
orozery, line 201: use kwargs here as well

vllm/v1/kv_offload/worker/cpu_gpu.py
```diff
@@ -152,29 +175,54 @@ def transfer_async(self, job_id: int, transfer_spec: TransferSpec) -> bool:
...
+            transfer_spec[0].medium(),
+            transfer_spec[1].medium(),
+        )
+        self._transfers.append(
+            Transfer(
+                job_id,
+                stream,
+                start_event,
+                end_event,
+                src_sub_block_count,
```
orozery: we need to report size in bytes here.
It should be `dst_sub_block_count * self.block_size_in_bytes`
Initialize `self.block_size_in_bytes` in `SingleDirectionOffloadingHandler.__init__` like this:
```
        self.block_size_in_bytes = 0
        for tensor, split_k_and_v in zip(src_tensors, kv_dim_before_num_blocks):
            if split_k_and_v:
                self.block_size_in_bytes += 2 * tensor.stride(1) * tensor.element_size()
            else:
                self.block_size_in_bytes += tensor.stride(0) * tensor.element_size()
```

vllm/v1/kv_offload/worker/worker.py
```diff
@@ -53,15 +68,6 @@ def get_finished(self) -> list[TransferResult]:
...
 
-    @abstractmethod
-    def wait(self, job_ids: set[int]) -> None:
-        """
-        Wait for jobs to finish (blocking).
-
-        Args:
-            job_ids: The set of job IDs to wait for.
-        """
-
```
orozery, line 64: revert

vllm/v1/kv_offload/worker/cpu_gpu.py
```diff
@@ -96,10 +108,10 @@ def __init__(
         assert len(src_tensors) > 0
         self.gpu_to_cpu: bool = self.src_tensors[0].is_cuda
 
-        # job_id -> event
-        self._transfer_events: dict[int, torch.Event] = {}
+        # job_id -> (start_event, end_event)
+        self._transfer_events: dict[int, tuple[torch.Event, torch.Event]] = {}
```
orozery: no need to track start_event here, as `wait` only needs to sync on `end_event` (see my comment on `wait`).

vllm/v1/kv_offload/worker/worker.py
```diff
@@ -9,12 +10,26 @@
...
+    transfer_type: TransferType
+
+
+@dataclass
+class TransferResult:
+    job_id: int
+    success: bool
+    transfer_size: int
+    transfer_time: float
+    transfer_type: TransferType
```
orozery: Let's make these optional:
```suggestion
    transfer_size: int | None = None
    transfer_time: float | None = None
    transfer_type: TransferType | None = None
```

vllm/v1/kv_offload/worker/worker.py
```diff
@@ -9,12 +10,26 @@
...
+    num_blocks: int
+    time: float  # Can be start_time or duration
+    transfer_type: TransferType
+
+
+@dataclass
+class TransferResult:
+    job_id: int
+    success: bool
+    transfer_size: int
```
orozery: maybe rename to `transfer_size_bytes` or add a comment explaining this is a size in bytes.

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -555,3 +683,199 @@ def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
...
+            labelnames=labelnames,
+        )
+        self.counter_num_cpu_to_gpu = self.make_per_engine(counter_num_cpu_to_gpu)
+
+        counter_num_gpu_to_cpu = self._counter_cls(
+            name="vllm:kv_offload_num_gpu_to_cpu",
+            documentation="Number of GPU-CPU transfers done",
+            labelnames=labelnames,
+        )
+        self.counter_num_gpu_to_cpu = self.make_per_engine(counter_num_gpu_to_cpu)
```
orozery: Instead of hard-coding gpu_to_cpu, cpu_to_gpu, let's allow the offloading connector to dynamically set per-transfer-type metrics by introducing a new label `transfer_type`.

We want to have 2 histograms:
```
self.histogram_transfer_size_bytes = self._histogram_cls(
    name="vllm:kv_offload_transfer_size_bytes",
    documentation="Size in bytes per KV data transfer"
    labels=labelnames + ["transfer_type"],
    buckets=transfer_size_buckets,
)

self.histogram_transfer_throughput_gb_s = self._histogram_cls(
    name="vllm:kv_offload_transfer_througput_gb_s",
    documentation="Throughput of KV transfer in GB/s"
    labels=labelnames + ["transfer_type"],
    buckets=transfer_throughput_buckets,
)
```

Then, in `observe` below, you would need to set the correct labels per each transfer, e.g.:
```
self.histogram_transfer_size_bytes.labels(self.per_engine_labelvalues[engine_idx] + [transfer_type]).observe(...)
```

This assumes you renamed `KVConnectorPromMetrics._per_engine_labelvalues` to `KVConnectorPromMetrics.per_engine_labelvalues`.

Ideally, cache `self.per_engine_labelvalues[engine_idx] + [transfer_type]` by using a dict mapping `(engine_idx, transfer_type)` to `labelvalues`.

orozery (changes requested):
Sorry for the big delay.
Can you please also fix the pre-commit failures?
Thanks!

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -555,3 +665,102 @@ def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
...
+            20e9,
+            30e9,
+            40e9,
+            50e9,
+            70e9,
+        ]
+
+        self.histogram_transfer_throughput = self._histogram_cls(
+            name="vllm:kv_offload_throughput",
+            documentation="Histogram of KV offload transfer throughput, in bytes/sec.",
```
markmc: You've kept this per-transfer throughput histogram? Would you mind explaining the thinking behind keeping it? Thanks

omerpaz95: Hi Mark, of course. 
The reason is that it's more accurate to compute the throughput on a per-transfer basis, rather than over time. In order to measure performance of the offloading correctly, we want to look at only the time slice in which the KV was actually being transferred. I measured this from within the worker itself in order to get the best granularity I could. This also correlates well with tests I conducted, which show the throughput is very similar to the bandwidth between GPU and CPU. 
Rate() doesn't achieve that level of granularity and accuracy.
I hope this makes sense

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -555,3 +665,102 @@ def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
...
+        transfer_stats_data is expected to be a dict where:
+        - keys are transfer type strings (e.g., "cpu_to_gpu", "gpu_to_cpu")
+        - values are lists of OffloadingOperationMetrics objects
+        """
+
+        for transfer_type, ops in transfer_stats_data.items():
+            # Cache:
+            if (engine_idx, transfer_type) not in self.per_engine_labelvalues:
+                self.per_engine_labelvalues[engine_idx, transfer_type] = (
+                    self._per_engine_labelvalues[engine_idx] + [transfer_type]
```
orozery: We're accessing a hidden member (`_per_engine_labelvalues`) of a parent class.
Can you please change it to non-hidden? (`per_engine_labelvalues`)

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -555,3 +665,102 @@ def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
...
+class OffloadPromMetrics(KVConnectorPromMetrics):
+    def __init__(
+        self,
+        vllm_config: VllmConfig,
+        metric_types: dict[type[PromMetric], type[PromMetricT]],
+        labelnames: list[str],
+        per_engine_labelvalues: dict[int, list[object]],
+    ):
+        super().__init__(vllm_config, metric_types, labelnames, per_engine_labelvalues)
+        self.per_engine_labelvalues: dict[tuple[int, str], list[object]] = {}
```
orozery:
```suggestion
        # (engine_idx, transfer_tupe) -> (metric with bounded labels)
        self.histogram_transfer_size: dict[tuple[int, str], list[PromMetricT]] = {}
        self.histogram_transfer_throughput: dict[tuple[int, str], list[PromMetricT]] = {}
```

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -555,3 +665,102 @@ def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
...
+        ]
+
+        self.histogram_transfer_throughput = self._histogram_cls(
+            name="vllm:kv_offload_throughput",
+            documentation="Histogram of KV offload transfer throughput, in bytes/sec.",
+            buckets=buckets_throughput[:],
+            labelnames=labelnames + ["transfer_type"],
+        )
+
+        self.histogram_transfer_size = self._histogram_cls(
```
orozery:
```suggestion
        self._histogram_transfer_size = self._histogram_cls(
```

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -555,3 +665,102 @@ def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
...
+            5e9,
+            10e9,
+            20e9,
+            30e9,
+            40e9,
+            50e9,
+            70e9,
+        ]
+
+        self.histogram_transfer_throughput = self._histogram_cls(
```
orozery:
```suggestion
        self._histogram_transfer_throughput = self._histogram_cls(
```

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
orozery: Also, you are currently caching the list of labelvalues.
Let's cache the labeled metric, e.g.
```
self.histogram_transfer_size[(engine_idx, transfer_type)] = self._histogram_transfer_size.labels(*(self.per_engine_labelvalues[engine_idx] + [transfer_type]))
```

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -555,3 +665,102 @@ def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
...
+            if (engine_idx, transfer_type) not in self.per_engine_labelvalues:
+                self.per_engine_labelvalues[engine_idx, transfer_type] = (
+                    self._per_engine_labelvalues[engine_idx] + [transfer_type]
+                )
+
+            for op in ops:
+                # Observe size histogram
+                self.histogram_transfer_size.labels(
+                    *self.per_engine_labelvalues[engine_idx, transfer_type]
+                ).observe(op["op_size"])
```
orozery: with the change I suggested above, you can remove the `labels` call here, and down as well.

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -526,10 +630,16 @@ def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
...
+        for transfer_result in self.worker.get_finished():
             # we currently do not support job failures
+            job_id, success, num_bytes, transfer_time, transfer_type = astuple(
+                transfer_result
+            )
             assert success
             req_id, store = self._jobs.pop(job_id)
+            self.kv_connector_stats.record_transfer(
+                num_bytes, transfer_time, transfer_type
+            )
```
orozery: skip record if `transfer_time` is 0.

orozery (changes requested):

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
markmc: Thanks, that's helpful

Consider whether `rate()` with `vllm:kv_offload_total_time` and `vllm:kv_offload_size_sum` would be sufficient 

Extracts from futher back-and-forth with Claude

---

Per-transfer throughput histogram (vllm:kv_offload_throughput):
- Question: "How fast is the hardware actually transferring data when a transfer occurs?"
- Measures: Instantaneous bandwidth utilization during active transfers
- Use case: Diagnosing hardware performance issues (PCIe degradation, DMA problems, memory bandwidth issues)

Scenario: Transfer 100MB in 0.1 seconds, once every 10 seconds

- Per-transfer throughput: 100MB / 0.1s = 1 GB/s (hardware is fast!)
- rate() over 10s: 100MB / 10s = 10 MB/s (but mostly idle)
  
If you want to detect that your PCIe bandwidth has degraded from 10 GB/s to 5 GB/s, you need the per-transfer metric.
  
You cannot derive the per-transfer throughput distribution from separate size and duration histograms because you lose the correlation between which sizes took which durations.
  
When would you need the distribution vs. just an average?

Distribution is useful when:
- You have bimodal behavior (some transfers fast, some slow)
- You want percentile-based SLOs ("p99 throughput > 8 GB/s")
- Some transfers are hitting different code paths with different characteristics

But for KV cache offloading:
- Transfers are roughly uniform in size (blocks of KV cache)
- Same hardware path every time (GPU ↔ CPU memory)
- If PCIe is degraded, it affects all transfers roughly equally

Proposed Alternative - just track:

- vllm:kv_offload_size_sum{transfer_type="CPU_to_GPU"} (counter)
- vllm:kv_offload_total_time{transfer_type="CPU_to_GPU"} (counter)

Then in PromQL:

```
# Average throughput during active transfers
rate(vllm:kv_offload_total_bytes{transfer_type="CPU_to_GPU"}[5m])
  /
rate(vllm:kv_offload_total_time{transfer_type="CPU_to_GPU"}[5m])
```

For the stated purpose (verifying hardware bandwidth is healthy), counters may be sufficient sufficient. The throughput histogram may be over-engineering unless there's evidence that:

1. Transfer throughput has high variance
2. You need to track outliers specifically
3. You're setting SLOs on percentiles

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -555,3 +665,102 @@ def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
...
+
+        self.histogram_transfer_size = self._histogram_cls(
+            name="vllm:kv_offload_size",
+            documentation="Histogram of KV offload transfer size, in bytes.",
+            buckets=buckets[:],
+            labelnames=labelnames + ["transfer_type"],
+        )
+
+        self.counter_kv_transfers = self._counter_cls(
+            name="vllm:kv_offload_num_transfers",
```
markmc: This duplicates `vllm:kv_offload_size_count` which already counts the number of transfers?

orozery: I don't think we want this metrics.
Just the 2 histograms.

vllm/v1/kv_offload/worker/cpu_gpu.py
```diff
@@ -92,18 +102,25 @@ def __init__(
         self.kv_dim_before_num_blocks: list[bool] = kv_dim_before_num_blocks
         self.src_block_size_factor: int = src_block_size_factor
         self.dst_block_size_factor: int = dst_block_size_factor
-
         assert len(src_tensors) > 0
         self.gpu_to_cpu: bool = self.src_tensors[0].is_cuda
-
+        self.transfer_type: tuple[str, str] = ("CPU", "GPU")
+        if self.gpu_to_cpu:
+            self.transfer_type = ("GPU", "CPU")
```
orozery:
```suggestion
        self.transfer_type = ("GPU", "CPU") if self.gpu_to_cpu else ("CPU", "GPU")
```

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -555,3 +666,114 @@ def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
...
+        )
+
+        self._histogram_transfer_size = self._histogram_cls(
+            name="vllm:kv_offload_size",
+            documentation="Histogram of KV offload transfer size, in bytes.",
+            buckets=buckets[:],
+            labelnames=labelnames + ["transfer_type"],
+        )
+
+        self._counter_kv_transfers = self._counter_cls(
```
orozery: can we remove this?

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -526,10 +630,17 @@ def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
         """
         finished_sending = set()
         finished_recving = set()
-        for job_id, success in self.worker.get_finished():
+        for transfer_result in self.worker.get_finished():
             # we currently do not support job failures
+            job_id, success, num_bytes, transfer_time, transfer_type = astuple(
+                transfer_result
+            )
```
orozery: can we remove this unpacking and use `transfer_result.field_name` instead?

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -498,6 +598,8 @@ def start_kv_transfers(self, metadata: OffloadingConnectorMetadata):
 
         for req_id, transfer_spec in metadata.reqs_to_load.items():
             job_id = self._generate_job_id()
+            src_spec, dst_spec = transfer_spec
+            assert isinstance(src_spec, BlockIDsLoadStoreSpec)
```
orozery: remove

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -507,6 +609,8 @@ def start_kv_transfers(self, metadata: OffloadingConnectorMetadata):
     def prepare_store_kv(self, metadata: OffloadingConnectorMetadata):
         for req_id, transfer_spec in metadata.reqs_to_store.items():
             job_id = self._generate_job_id()
+            src_spec, dst_spec = transfer_spec
+            assert isinstance(dst_spec, BlockIDsLoadStoreSpec)
```
orozery: remove

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -526,10 +630,17 @@ def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
...
+        for transfer_result in self.worker.get_finished():
             # we currently do not support job failures
+            job_id, success, num_bytes, transfer_time, transfer_type = astuple(
+                transfer_result
+            )
             assert success
             req_id, store = self._jobs.pop(job_id)
+            if transfer_time > 0:
+                self.kv_connector_stats.record_transfer(
+                    num_bytes, transfer_time, transfer_type
```
orozery: can you use kwargs?

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -555,3 +666,114 @@ def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
...
         return finished_sending, finished_recving
+
+    def get_kv_connector_stats(self) -> KVConnectorStats | None:
+        """
+        Get the KV transfer stats for the connector.
+        """
+        # Clear stats for next iteration
+        if not self.kv_connector_stats.is_empty():
+            return self.kv_connector_stats.clone_and_reset()
+        return None
```
orozery:
```suggestion
        if self.kv_connector_stats.is_empty():
            return None

        # Clear stats for next iteration
        return self.kv_connector_stats.clone_and_reset()
```

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -526,10 +630,17 @@ def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
...
         finished_recving = set()
-        for job_id, success in self.worker.get_finished():
+        for transfer_result in self.worker.get_finished():
             # we currently do not support job failures
+            job_id, success, num_bytes, transfer_time, transfer_type = astuple(
+                transfer_result
+            )
             assert success
             req_id, store = self._jobs.pop(job_id)
+            if transfer_time > 0:
```
orozery:
```suggestion
            if transfer_time:
```

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -37,6 +48,63 @@
...
+        return old
+
+    def aggregate(self, other: KVConnectorStats) -> KVConnectorStats:
+        if not other.is_empty():
+            for k, v in other.data.items():
+                if k in self.data:
+                    accumulator = self.data[k]
+                    assert isinstance(accumulator, list)
+                    accumulator.extend(v)
+        return self
```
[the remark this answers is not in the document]
orozery: Can you address this?

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -37,6 +48,65 @@
...
+        """
+        Reduce the observations collected during a time interval to one or
+        more representative values (eg avg/median/sum of the series).
+        This is meant to be called by the logger to produce a summary of the
+        stats for the last time interval.
+        """
+        return_dict: dict[str, int | float] = {}
+        for k, v in self.data.items():
+            assert isinstance(v, list)
+            return_dict[k] = len(v)
```
orozery: Shouldn't you `sum` instead of `len`?

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -37,6 +48,65 @@
...
+        return return_dict
+
+    def is_empty(self) -> bool:
+        return len(self.data.items()) == 0
+
+    def record_transfer(self, num_bytes: int, time: float, transfer_type: TransferType):
+        src, dst = transfer_type
+        transfer_type_key = src + "_to_" + dst
+        op = OffloadingOperationMetrics(num_bytes, time * 1e-3, transfer_type_key)
```
orozery: why `time * 1e-3` ?

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -37,6 +48,65 @@
...
+        if transfer_type_key in self.data:
+            self.data[transfer_type_key].append(op)
```
orozery, line 105: @markmc Is there a good reason why we should keep appending to a list and not reduce here?
I think this data propagates back to `EngineCoreOutput`, but I currently don't see a real use-case where someone would use this raw data.

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -37,6 +48,65 @@
 logger = init_logger(__name__)
 
 
+@dataclass
+class OffloadingOperationMetrics:
+    op_size: float
+    op_time: float
+    op_type: str
```
orozery: This field is not needed, since it is stored in the key of `data`.

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -555,3 +666,114 @@ def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
...
+                    )
+                )
+                self.counter_kv_transfers[(engine_idx, transfer_type)] = (
+                    self._counter_kv_transfers.labels(
+                        *(self.per_engine_labelvalues[engine_idx] + [transfer_type])
+                    )
+                )
+
+            # Process ops:
+            for op in ops:
```
orozery:
```suggestion
            assert isinstance(ops, list)
            for op in ops:
                assert isinstance(op, OffloadingOperationMetrics)
```

Then you can access op.op_size etc, instead of op["op_size"]

orozery (changes requested):

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
omerpaz95: elapsed_time in torch.cuda.Event returns time elapsed in miliseconds.

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
markmc: It looks to me like this is needed for the transfer size histogram? We're recording the size of each individual transfer?

This list of individual transfer sizes needs to get from the worker to the frontend process where prometheus metrics are observed

(Unless I'm misunderstanding your question)

orozery: Right, I missed that.

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
orozery: Then let's do the multiplication right there.
`TransferResult.transfer_time` is expected in seconds.

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -555,3 +666,114 @@ def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
...
+        """
+        Observe transfer statistics from the new data structure.
+        transfer_stats_data is expected to be a dict where:
+        - keys are transfer type strings (e.g., "cpu_to_gpu", "gpu_to_cpu")
+        - values are lists of OffloadingOperationMetrics objects
+        """
+
+        for transfer_type, ops in transfer_stats_data.items():
+            # Cache:
+            if (engine_idx, transfer_type) not in self.histogram_transfer_size:
```
markmc, line 769: This can/should be done in the constructor, surely?  All possible engine indices and transfer types are known then, right?

vllm/distributed/kv_transfer/kv_connector/v1/metrics.py
```diff
@@ -124,7 +124,7 @@ def __init__(
         self._counter_cls = metric_types[Counter]
         self._histogram_cls = metric_types[Histogram]
         self._labelnames = labelnames
-        self._per_engine_labelvalues = per_engine_labelvalues
+        self.per_engine_labelvalues = per_engine_labelvalues
```
markmc, line 127: This shouldn't be needed if you move the per-transfer-type metric creation to the constructor

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
orozery: It is known at the worker-connector initialization time.
Not available to the constructor of `KVConnectorPromMetrics`, which is actually built using a `classmethod`.

markmc: Not following why this wouldn't work:

```
       transfer_types = ["CPU_to_GPU", "GPU_to_CPU"]                                                                                                                                                                 
       for engine_idx in per_engine_labelvalues.keys():                                                                                                                                                              
           for transfer_type in transfer_types:                                                                                                                                                                      
               labelvalues = per_engine_labelvalues[engine_idx] + [transfer_type]                                                                       
               self.histogram_transfer_size[(engine_idx, transfer_type)] = (                                                                                                                                         
                   _histogram_transfer_size.labels(*labelvalues)                                                                                                                                                     
               ) 
```

orozery: You hard-code the transfer types.
The offloading connector is designed to work with arbitrary mediums, and these are determined at runtime.

markmc: Wow, now I see your concern ...

```
        extra_config = kv_transfer_config.kv_connector_extra_config
        spec_module_path = extra_config.get("spec_module_path")
```

we support out-of-tree `OffloadingSpec` and mediums? What's the use case for this?

orozery: The general use-case is allowing easy integration of KV offloading to other mediums (e.g. filesystem, object storage).
Concretely, llm-d already uses this for offloading to shared storage.

markmc: ok, I understand what's motivating the flexibility now :+1:

vllm/distributed/kv_transfer/kv_connector/v1/metrics.py
markmc: Understand now that externally loaded transfer types is what's driving the need for this flexibility

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -37,6 +48,77 @@
...
+            for op in v:
+                for stat, value in [
+                    ("_total_bytes", op["op_size"]),
+                    ("_total_time", op["op_time"]),
+                ]:
+                    log_key = k + stat
+                    if log_key not in return_dict:
+                        return_dict[log_key] = value
+                    else:
+                        return_dict[log_key] += value
```
orozery: nit:
```suggestion
        for transfer_type, ops_list in self.data.items():
            assert isinstance(ops_list, list)
            return_dict[f"{transfer_type}_total_bytes"] = sum(op["op_size"] for op in ops)
            return_dict[f"{transfer_type}_total_time"] = sum(op["op_time"] for op in ops)
```

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -37,6 +48,77 @@
...
+
+        return return_dict
+
+    def is_empty(self) -> bool:
+        return len(self.data.items()) == 0
```
orozery: nit:
```suggestion
        return not self.data
```

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -63,6 +145,8 @@ def __init__(
         if role == KVConnectorRole.SCHEDULER:
             self.connector_scheduler = OffloadingConnectorScheduler(spec)
         elif role == KVConnectorRole.WORKER:
+            if kv_cache_config is None:
+                raise ValueError("kv_cache_config cannot be None for WORKER role")
```
orozery: Why do we need this addition?

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -143,6 +227,33 @@ def take_events(self) -> Iterable[KVCacheEvent]:
         assert self.connector_scheduler is not None
         return self.connector_scheduler.take_events()
 
+    def get_kv_connector_stats(self) -> KVConnectorStats | None:
+        if self.connector_worker is None:
+            return None
```
orozery:
```suggestion
        assert self.connector_worker is not None
```

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -588,3 +707,101 @@ def get_finished(self, finished_req_ids: set[str]) -> tuple[set[str], set[str]]:
...
+
+    def get_kv_connector_stats(self) -> KVConnectorStats | None:
+        """
+        Get the KV transfer stats for the connector.
+        """
+
+        if self.kv_connector_stats.is_empty():
+            return None
+        # Clear stats for next iteration
+        return self.kv_connector_stats.clone_and_reset()
```
orozery: This triggers a copy of the old KVConnectorStats.
I see NixlConnector also does it, but it seems to me that a more simpler and efficient way is to simply:
```suggestion
        kv_connector_stats = self.kv_connector_stats
        self.kv_connector_stats = OffloadingConnectorStats()
        return kv_connector_stats
```

@markmc your thoughts?

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -37,6 +48,77 @@
 logger = init_logger(__name__)
 
 
+@dataclass
+class OffloadingOperationMetrics:
+    op_size: float
+    op_time: float
+
+    def __getitem__(self, key):
+        return getattr(self, key)
```
orozery: Remove?

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
```diff
@@ -37,6 +48,77 @@
...
+
+    def reset(self):
+        self.data: dict[str, list[OffloadingOperationMetrics]] = {}
+
+    def clone_and_reset(self) -> "OffloadingConnectorStats":
+        old = copy.copy(self)
+        self.reset()
+        return old
+
+    def aggregate(self, other: KVConnectorStats) -> KVConnectorStats:
```
orozery, line 66: see cursor's comment below

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
omerpaz95: I added this so that we can access the op's fields in a dict-like manner from reduce() , to conform with how we access them in observe(). Otherwise, it fails pre-commit.

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
omerpaz95: I already fixed the issue (To my understanding)

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
omerpaz95: Deleted, no idea how it got here. Sorry.

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
orozery: my bad, I just noticed that `get_kv_connector_stats` is used by both the scheduler and worker.
We just emit from the worker.
So can you just add a comment above the return None?
e.g. `# We only emit stats from the worker-side`

vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py
orozery: I just tested it, and it seems that after de-serialization, `type(op)` is actually a dict.
So I think we can remove the code here, and instead `assert(isinstance(op, dict))` on the `observe` function.

tests/v1/kv_connector/unit/test_offloading_connector.py
```diff
@@ -86,7 +87,14 @@ def complete_jobs(self, job_ids: set[int]) -> None:
             if job_id in self.waiting_jobs:
                 self.waiting_jobs.remove(job_id)
                 self.completed_jobs.append(job_id)
-                self.completed_transfers.append((job_id, True))
+                result = TransferResult(
+                    job_id=job_id,
+                    success=True,
+                    transfer_size=None,  # For now - just to get the test working
+                    transfer_time=None,  # For now - just to get the test working
+                    transfer_type=None,  # For now - just to get the test working
```
orozery: nit: I would remove the comments

tests/v1/kv_connector/unit/test_offloading_connector.py
```diff
@@ -717,3 +725,144 @@ def test_concurrent_lookups_of_the_same_prefix(request_runner):
 
     # second request will use the GPU prefix cache
     assert transfer_jobs == list(runner.offloading_spec.handler.transfer_specs)
+
+
+class TestOffloadingConnectorStats:
+    """Tests for MultiConnector stats reconstruction and operations."""
```
orozery:
```suggestion
    """Tests for OffloadingConnector stats reconstruction and operations."""
```

vllm/v1/kv_offload/worker/cpu_gpu.py
```diff
@@ -153,22 +174,42 @@ def transfer_async(self, job_id: int, transfer_spec: TransferSpec) -> bool:
...
+            end_event.record(stream)
+
+        self._transfer_events[job_id] = end_event
+        self._transfers.append(
+            Transfer(
+                job_id=job_id,
+                stream=stream,
+                start_event=start_event,
+                end_event=end_event,
+                num_bytes=dst_sub_block_count * self.block_size_in_bytes,
```
orozery: I think `self.block_size_in_bytes` is a list.
Can you define `self.total_block_size_in_bytes` which is the sum of this list and use it instead?

tests/v1/kv_offload/test_cpu_gpu.py
```diff
@@ -169,14 +169,26 @@ def test_transfer(
...
+                finished[0].transfer_type == ("GPU", "CPU")
+                if gpu_to_cpu
+                else ("CPU", "GPU")
+            )
+            assert (
+                finished[0].transfer_size
+                == handler.block_size_in_bytes
+                * handler.dst_block_size_factor
+                * len(dst_blocks)
+            )
```
orozery, line 192: can you also assert `0 < finished[0].time < (time.time() - start_time)`?

orozery (approved):
LGTM! Thanks for the all hard work!!!
cc @markmc

markmc (approved):

PR #27953, opened 2025-11-03
[HARDWARE][CPU] Add Option for Disabling Binding to Specific CPU Cores

StanHatko:
This pull request adds an option `nobind` for `VLLM_CPU_OMP_THREADS_BIND` (in addition to the current `auto` default and the list of specific cores options). In this mode, binding to specific CPU cores is disabled and instead the variables `NUMEXPR_MAX_THREADS` and `OMP_NUM_THREADS` are used (with the current vLLM CPU defaults being used for these if missing). I also updated the documentation with instructions about this new setting.

## Purpose

While in general binding to specific CPU cores is beneficial (especially on NUMA systems), in some cases like multi-user systems with each user doing their own thing or Kubernetes clusters with hundreds of users that don't coordinate it can lead to problems, with multiple jobs trying to reserve the same cores and causing major slowdowns. I've seen this cause massive performance drops that were removed by applying a similar patch independently.

This adds a `nobind` option to disable binding to CPU cores. This isn't the default (which remains `auto`) so CPU users continue to benefit from the CPU cores binding and NUMA optimizations, however if users encounter problems with these (especially on these multi-user systems) they have an option to disable these. Also it benefits people doing benchmarking as they can check exactly how much CPU cores binding and NUMA optimizations are having.

## Test Plan

I've run a test of this in a uv environment with CPU vLLM with the patch applied. I tested adjusting the environment variables and they generally work as expected. Scaling CPU cores performed a bit worse than expected, but still reached maximum almost identical to the `auto` setting. This means vllm CPU is still useful for users that can't use these optimizations (at least on non-NUMA systems).

Specifically, I set the variables and run the command:

```bash
export VLLM_CPU_KVCACHE_SPACE=32

# Also tested auto here for baseline.
export VLLM_CPU_OMP_THREADS_BIND=nobind

# Tested 4, 8, and 16 here for both variables.
export NUMEXPR_MAX_THREADS=4
export OMP_NUM_THREADS=4

vllm bench throughput --model Qwen/Qwen3-4B-Instruct-2507 --dataset-name sharegpt --input-len 256 --max-model-len 16k --dtype bfloat16
```

## Test Result

Here are the test results, using the command after setting the environment variables.  I found somewhat worse-than-linear scaling with the number of CPU cores, once 16 cores reached it almost the same as the default `auto` on my system (of course, on a NUMA system the difference would probably be larger as `nobind` doesn't use the NUMA optimizations). All CPU runs were far slower than the GPU run, but that's expected.

CPU 4 cores and nobind:

* Throughput: 0.57 requests/s, 219.19 total tokens/s, 73.06 output tokens/s
* Total num prompt tokens:  256000
* Total num output tokens:  12800

CPU 8 cores and nobind:

* Throughput: 0.95 requests/s, 364.88 total tokens/s, 121.63 output tokens/s
* Total num prompt tokens:  256000
* Total num output tokens:  128000

CPU 16 cores and nobind:

* Throughput: 1.39 requests/s, 532.44 total tokens/s, 177.48 output tokens/s
* Total num prompt tokens:  256000
* Total num output tokens:  128000

CPU with `VLLM_CPU_OMP_THREADS_BIND` of `auto`:

* Throughput: 1.39 requests/s, 533.67 total tokens/s, 177.89 output tokens/s
* Total num prompt tokens:  256000
* Total num output tokens:  128000

GPU is way faster of course (in the uv environment with GPU vllm):

* Throughput: 45.78 requests/s, 17557.30 total tokens/s, 5859.60 output tokens/s
* Total num prompt tokens:  255530
* Total num output tokens:  128000

---
<details>

</details>


docs/getting_started/installation/cpu.md
```diff
@@ -94,11 +94,13 @@ Currently, there are no pre-built CPU wheels.
...
 - `VLLM_CPU_MOE_PREPACK` (x86 only): whether to use prepack for MoE layer. This will be passed to `ipex.llm.modules.GatedMLPMOE`. Default is `1` (True). On unsupported CPUs, you might need to set this to `0` (False).
 - `VLLM_CPU_SGL_KERNEL` (x86 only, Experimental): whether to use small-batch optimized kernels for linear layer and MoE layer, especially for low-latency requirements like online serving. The kernels require AMX instruction set, BFloat16 weight type and weight shapes divisible by 32. Default is `0` (False).
+- `NUMEXPR_MAX_THREADS`: specifies maximum number of threads to be used by `numexpr`. The variable only takes effect when `VLLM_CPU_OMP_THREADS_BIND` is set to `nobind`. Default value is determined by `os.sched_getaffinity` (if available) or `os.cpu_count` otherwise.
+- `OMP_NUM_THREADS`: specified number of OpenMP threads to use. The variable only takes effect when `VLLM_CPU_OMP_THREADS_BIND` is set to `nobind`. Default value is equal to `torch.get_num_threads` result.
```
bigPYJ1151: unnecessary to list these here, I think they are standard OpenMP settings

docs/getting_started/installation/cpu.md
```diff
@@ -94,11 +94,13 @@ Currently, there are no pre-built CPU wheels.
...
-- `VLLM_CPU_OMP_THREADS_BIND`: specify the CPU cores dedicated to the OpenMP threads, can be set as CPU id lists or `auto` (by default). For example, `VLLM_CPU_OMP_THREADS_BIND=0-31` means there will be 32 OpenMP threads bound on 0-31 CPU cores. `VLLM_CPU_OMP_THREADS_BIND=0-31|32-63` means there will be 2 tensor parallel processes, 32 OpenMP threads of rank0 are bound on 0-31 CPU cores, and the OpenMP threads of rank1 are bound on 32-63 CPU cores. By setting to `auto`, the OpenMP threads of each rank are bound to the CPU cores in each NUMA node respectively.
+- `VLLM_CPU_OMP_THREADS_BIND`: specify the CPU cores dedicated to the OpenMP threads, can be set as CPU id lists, `auto` (by default), or `nobind` (to disable binding to individual CPU cores). For example, `VLLM_CPU_OMP_THREADS_BIND=0-31` means there will be 32 OpenMP threads bound on 0-31 CPU cores. `VLLM_CPU_OMP_THREADS_BIND=0-31|32-63` means there will be 2 tensor parallel processes, 32 OpenMP threads of rank0 are bound on 0-31 CPU cores, and the OpenMP threads of rank1 are bound on 32-63 CPU cores. By setting to `auto`, the OpenMP threads of each rank are bound to the CPU cores in each NUMA node respectively.
```
bigPYJ1151: We can explain the ```nobind``` setting here, such as ```using external OpenMP settings to determine thread number and affinity```

vllm/platforms/cpu.py
```diff
@@ -287,12 +287,32 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
+                "Using NUMEXPR_MAX_THREADS: %s",
+                os.environ["NUMEXPR_MAX_THREADS"],
+            )
+
+            if "OMP_NUM_THREADS" not in os.environ:
+                os.environ["OMP_NUM_THREADS"] = str(torch.get_num_threads())
+            logger.info(
+                "Using OMP_NUM_THREADS: %s",
+                os.environ["OMP_NUM_THREADS"],
+            )
```
bigPYJ1151: Maybe it is unnecessary to do so much checks. We can just:
- keep ```os.environ["NUMEXPR_MAX_THREADS"] = str(get_max_threads())``` to suppress the warning. ```numexpr``` is not used in fact, but throws an unexpected warning starts from a version.
- remove ```os.environ["OMP_NUM_THREADS"] = str(torch.get_num_threads())``` to inherit user defined OpenMP envs.

vllm/v1/worker/cpu_worker.py
```diff
@@ -54,6 +54,8 @@ def init_device(self):
                 )
             else:
                 self.local_omp_cpuid = "all"
+        elif omp_cpuids == "nobind":
```
bigPYJ1151, line 79: I think we can replace ```all``` with ```nobind```. ```all``` is a legacy defination and not exposed to users.

vllm/platforms/cpu.py
StanHatko: Can `os.environ["NUMEXPR_MAX_THREADS"] = str(get_max_threads())` be set unconditionally in this case? That would make things simpler.

For `OMP_NUM_THREADS` not set, I did it this way so the default (if nobind is set) is the maximum number of threads, with the user able to manually set `OMP_NUM_THREADS` if they want to reduce it. Is there a possibility of OpenMP number of threads being set other than by using the `OMP_NUM_THREADS` variable?

vllm/platforms/cpu.py
```diff
@@ -287,12 +287,32 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
 
         os.environ["VLLM_WORKER_MULTIPROC_METHOD"] = "spawn"
 
-        # Note: to avoid the error 'nthreads cannot be larger than environment
-        #  variable "NUMEXPR_MAX_THREADS" (64)'.
-        os.environ["NUMEXPR_MAX_THREADS"] = str(get_max_threads())
+        if os.getenv("VLLM_CPU_OMP_THREADS_BIND", "NA") != "nobind":
```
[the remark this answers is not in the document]
StanHatko: Issue fixed.

docs/getting_started/installation/cpu.md
StanHatko: Yes they are. If `NUMEXPR_MAX_THREADS` can be set to maximum always (discussed below), then we can omit it from documentation completely. For `OMP_NUM_THREADS`, normally it's set automatically in vllm, so I'd have to mention it at least within the `VLLM_CPU_OMP_THREADS_BIND` variable documentation.

vllm/v1/worker/cpu_worker.py
StanHatko: So basically we don't need to distinguish the `all` (from arch that isn't `CpuArchEnum.POWERPC`, `CpuArchEnum.S390X`, `CpuArchEnum.X86`) and `nobind` cases here?

I could do that, it will make the code slightly simpler.

StanHatko: Simplified this, set to `nobind` in these cases.

vllm/platforms/cpu.py
bigPYJ1151: Yes, ```NUMEXPR_MAX_THREADS``` is set unconditionally. 

I think OpenMP will use the maximum number of threads by default if users don't set ```OMP_NUM_THREADS```. So it is unnecessary to set it to the maximum explictly.

docs/getting_started/installation/cpu.md
bigPYJ1151: Yes we can mention it with ```nobind``` toghter within the ```VLLM_CPU_OMP_THREADS_BIND``` section.

bigPYJ1151 (approved):
Thanks, LGTM :)

PR #27954, opened 2025-11-03
[CPU] Refactor CPU attention backend

bigPYJ1151:
## Purpose

This PR refactors CPU attention backend, includes:
- clean up unused code and simplifiy metadata
- a unified kernel supports chunked prefill, sliding window, alibi, softcap, and sink
- renamed ```TorchSDPABackend``` to ```CPUAttentionBackend``` for less misunderstandings. For now the ```TORCH_SDPA``` tag is only used for ViT attention
- better performance on both of prefill and decode
- enable more related unit tests

cc @fadara01 @Akashcodes732 

this:
```log
Successful requests:                     64        
Maximum request concurrency:             16        
Benchmark duration (s):                  119.44    
Total input tokens:                      65472     
Total generated tokens:                  65536     
Request throughput (req/s):              0.54      
Output token throughput (tok/s):         548.71    
Peak output token throughput (tok/s):    671.00    
Peak concurrent requests:                31.00     
Total Token throughput (tok/s):          1096.89   
Mean TTFT (ms):                          2332.45   
Median TTFT (ms):                        2015.44   
P99 TTFT (ms):                           3497.95   
Mean TPOT (ms):                          26.90     
Median TPOT (ms):                        27.29     
P99 TPOT (ms):                           28.93     
Mean ITL (ms):                           26.90     
Median ITL (ms):                         25.82     
P99 ITL (ms):                            27.93     
```

main:
```log
Successful requests:                     64        
Maximum request concurrency:             16        
Benchmark duration (s):                  156.00    
Total input tokens:                      65472     
Total generated tokens:                  65536     
Request throughput (req/s):              0.41      
Output token throughput (tok/s):         420.10    
Peak output token throughput (tok/s):    512.00    
Peak concurrent requests:                32.00     
Total Token throughput (tok/s):          839.78    
Mean TTFT (ms):                          2488.38   
Median TTFT (ms):                        2144.20   
P99 TTFT (ms):                           3710.32   
Mean TPOT (ms):                          35.68     
Median TPOT (ms):                        36.07     
P99 TPOT (ms):                           37.81     
Mean ITL (ms):                           35.68     
Median ITL (ms):                         34.64     
P99 ITL (ms):                            38.18     
```

## Test Plan

unit tests

## Test Result

---
<details>

</details>


csrc/cpu/cpu_attn.cpp
```diff
@@ -0,0 +1,233 @@
...
+    if (casual) {
+      input.right_sliding_window_size = 0;
+    } else {
+      input.right_sliding_window_size = -1;
+    }
+  }
+  input.casual = casual;
+  input.isa = isa;
+  input.enable_kv_split = enable_kv_split;
+  TORCH_CHECK(casual, "Only supports casual mask for now.");
```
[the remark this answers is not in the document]
bigPYJ1151: Encoder related will use SDPA for now

fadara01:
Amazing Work! Thanks for raising this :) 
... I'll test and benchmark on Arm CPUs

csrc/cpu/cpu_attn_impl.hpp
```diff
@@ -0,0 +1,1984 @@
...
+      initial_arange_vals_vec =
+          initial_arange_vals_vec + vec_op::FP32Vec16((float)kv_tile_start_pos);
+      vec_op::FP32Vec16 scale_vec(scale);
+      vec_op::FP32Vec16 pos_offset_vec(16.0);
+      logits_buffer_t* __restrict__ curr_logits_buffer = logits_buffer;
+      for (int32_t i = 0; i < q_token_num; ++i) {
+        vec_op::FP32Vec16 curr_q_pos_vec((float)(i + q_tile_start_pos));
+        for (int32_t j = 0; j < q_heads_per_kv; ++j) {
+          vec_op::FP32Vec16 alibi_scale_vec(alibi_slopes[j]);
+          vec_op::FP32Vec16 curr_kv_pos_vec = initial_arange_vals_vec;
```
fadara01: This doesn't compile on Arm because all our `vec_op::FP32Vec16` constructors are `explicit`
Changing it to: `vec_op::FP32Vec16 curr_kv_pos_vec(initial_arange_vals_vec)` fixes the problem.

vllm/v1/attention/backends/cpu_attn.py
```diff
@@ -445,234 +218,178 @@ def __init__(
...
-                "Torch SDPA backend FP8 KV cache requires "
-                "intel_extension_for_pytorch support."
-            )
+        if is_quantized_kv_cache(kv_cache_dtype):
+            raise NotImplementedError("FP8 KV cache is unsupported in CPU_ATTN")
         self.attn_type = attn_type
 
+        self.sinks = sinks
+        if self.sinks is not None:
+            assert self.sinks.shape[0] == num_heads, (
```
fadara01, line 251: When `use_sdpa_prefill` is `true` we use vanilla SDPA which does not support sinks.
Can we dispatch to `cpu_attention_with_kv_cache` when we have sinks even if `use_sdpa_prefill` is true?

If that's not possible for whatever reason, we should raise an error, and I can address it for the non-Intel CPU path in a follow-up PR.

csrc/cpu/cpu_attn_impl.hpp
```diff
@@ -0,0 +1,1984 @@
...
+  float* alibi_slopes;
+  c10::BFloat16* s_aux;
+  float scale;
+  bool causal;
+  int32_t sliding_window_left;
+  int32_t sliding_window_right;
+  float softcap;
+};
+#ifdef __AVX512F__
+  #define DEFINE_AVX512_EXP                                                    \
```
fadara01: This looks very out of place here :)
Can you move it to the AVX512 vectorized class in `csrc/cpu/cpu_attn_vec.hpp`?

We'll follow up later with an RFC/PoC to use torch vectorizer in vllm.

csrc/cpu/scratchpad_manager.h
```diff
@@ -0,0 +1,31 @@
+#ifndef SCRATCHPAD_MANAGER_H
```
fadara01, line 1: Can we leave oneDNN changes out? This PR is already too big and I don't think these changes are relevant the new attention backend?

tests/kernels/attention/test_cpu_attn.py
```diff
@@ -0,0 +1,407 @@
...
+        kv_lens=kv_lens,
+        block_tables=block_tables,
+        scale=scale,
+        sliding_window=sliding_window,
+        soft_cap=soft_cap,
+        alibi_slopes=alibi_slopes,
+        s_aux=s_aux,
+    )
+
+    atol, rtol = 1.5e-2, 1e-2
```
fadara01, line 324: the absolute tolerance looks too high.
can we use:
```
from tests.kernels.allclose_default import get_default_atol, get_default_rtol
atol = get_default_atol(output)
rtol = get_default_rtol(output)
```
similar to what we do in https://github.com/vllm-project/vllm/blob/main/tests/kernels/attention/test_attention.py

tests/kernels/attention/test_cpu_attn.py
```diff
@@ -0,0 +1,407 @@
...
+
+    # unrealistic cases
+    if (
+        (use_alibi and use_sink)
+        or (use_alibi and soft_cap is not None)
+        or (use_sink and soft_cap is not None)
+    ):
+        skip = True
+
+    # only tests features with bf16 to save time
```
fadara01: then why not just do `QTYPES = [torch.bfloat16]`

tests/models/language/generation/test_common.py
```diff
@@ -93,11 +96,7 @@
         pytest.param("bigcode/starcoder2-3b"),  # starcoder2
```
fadara01, line 100: Can we enable a test for `google/gemma-2-2b-it` and mark it as `cpu_model`?
This would be a great end-to-end smoke test for SWA and Hybrid local-global attention models (with 2 kv cache groups)

vllm/platforms/cpu.py
```diff
@@ -186,16 +184,14 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
-        if cache_config and cache_config.block_size is None:
-            cache_config.block_size = 128 if ipex_available else 16
-
-        if not ipex_available and cache_config.block_size != 16:
-            raise RuntimeError(
-                f"--block-size={cache_config.block_size} requires"
-                " intel_extension_for_pytorch"
-            )
+        if cache_config.block_size % 32 != 0:
+            block_size = cache_config.block_size
```
fadara01: I'm in favor of just erroring out in this case saying that block_size needs to be divisible by 32 (instead of setting the block_size to a value that the user didn't choose)

fadara01:
I can also confirm that there's no perf regressions on Arm after running this benchmark:
```
VLLM_CPU_OMP_THREADS_BIND=0-63 LD_PRELOAD="/usr/lib/aarch64-linux-gnu/libtcmalloc_minimal.so.4:/usr/lib/aarch64-linux-gnu/libgomp.so.1" VLLM_TARGET_DEVICE=cpu VLLM_CPU_KVCACHE_SPACE=32 vllm bench throughput --num-prompts 128 --seed 0 --dataset-name sharegpt --max-model-len 4096 --dataset-path /path/to/ShareGPT_V3_unfiltered_cleaned_split_no_imsorry.json  --model meta-llama/Llama-3.1-8B-Instruct --load-format dummy
```

csrc/cpu/cpu_attn_impl.hpp
```diff
@@ -0,0 +1,1984 @@
...
+struct AttentionMetadata {
+  std::atomic_int64_t counter;
+  char _padding1[56];
+  ISA isa;
+  int32_t workitem_group_num;
+  int32_t reduction_item_num;
+  int32_t reduction_split_num;
+  int32_t thread_num;
+  int32_t
+      effective_thread_num;  // non-zero item num in cu_workitem_num_per_thread
```
fadara01: could you please add a comment explaining this?
what's the significance of the `cu_` prefix here?

csrc/cpu/cpu_attn_impl.hpp
```diff
@@ -0,0 +1,1984 @@
...
+          float* __restrict__ scale_output_iter =
+              final_max > curr_max ? curr_split_output_buffer_iter
+                                   : curr_output_buffer;
+          float rescale_factor = final_max > curr_max ? curr_max - final_max
+                                                      : final_max - curr_max;
+
+#ifdef __AVX512F__
+          vec_op::FP32Vec16 rescale_factor_vec(rescale_factor);
+          rescale_factor_vec = avx512_exp_u20(rescale_factor_vec);
+          rescale_factor = _mm512_cvtss_f32(rescale_factor_vec.reg);
```
fadara01: let's try to keep this file free of architecture specific intrinsics and just use vllm's vectorizer abstractions.

vllm/v1/attention/backends/cpu_attn.py
bigPYJ1151: Seems can't do this as we are unable to get sink config in builder. Just add a assertion for now.

fadara01: Yeah, let's just fail for now, I'll address this in a follow up PR.

csrc/cpu/scratchpad_manager.h
bigPYJ1151: Yes it is a bit unrelevant. But I think it is acceptable as just a few code😂

tests/models/language/generation/test_common.py
```diff
@@ -55,6 +59,10 @@
                 pytest.mark.slow_test,
             ],
         ),
+        pytest.param(
+            "google/gemma-2-9b",  # test hybrid attention
```
fadara01: is 9b an over kill here? should we use the 2b variant `google/gemma-2-2b-it`?

csrc/cpu/scratchpad_manager.h
fadara01: Haha okay!!

tests/kernels/attention/test_cpu_attn.py
bigPYJ1151: This test file is based on https://github.com/vllm-project/vllm/blob/main/tests/kernels/attention/test_flash_attn.py

I think the abs tolerance looks more strict in ```test_attention.py``` is because the input is initialized with ```uniform_(-scale, scale)```, will be smaller compared with the inputs initialized with```randn``` in ```test_flash_attn.py```.

But I perfer to use ```randn``` as I found using small inputs can't figure out value difference in test cases of sink sometimes.

tests/kernels/attention/test_cpu_attn.py
bigPYJ1151: This means we just test sink, alibi, softcap with bf16 as the logits processing is using fp32. For other cases all dtypes should be tested.

tests/models/language/generation/test_common.py
bigPYJ1151: Good idea, added a gemma-2 case.

vllm/platforms/cpu.py
bigPYJ1151: Yes...But my concern is a lot of test cases use 16 by default and I don't want to add more if-else in different files so just round it here.

csrc/cpu/cpu_attn_impl.hpp
bigPYJ1151: Oh it is a mistake, should be ```workitem_num_per_thread```. The ```cu``` means cummulation, and the ```cu_workitem_num_per_thread``` is a array contains prefix sum of ```workitem_num_per_thread```.

vllm/platforms/cpu.py
fadara01: Oh, that's a good point. I agree with you.

tests/kernels/attention/test_cpu_attn.py
fadara01: Acknowledged, I wasn't aware that's the tolerance used for testing flash attention.

tests/kernels/attention/test_cpu_attn.py
fadara01: Oh I missed the second line of the condition. I agree with you

fadara01:
I got one test failure with one element miss-match while running `tests/kernels/attention/test_cpu_attn.py` on Arm

This is the configuration that fails:
```
seq_lens =  [(2345, 2345), (5, 5), (3, 16), (134, 5131)]
num_heads =  (9, 3)
head_size =  96
sliding_window =  None
dtype =  torch.bfloat16
block_size =  96
softcap =  None
num_blocks =  1024
use_alibi =  False
use_sink =  False
isa =  vec
```

And this is the log:
```
tests/kernels/attention/test_cpu_attn.py:413: AssertionError
FAILED tests/kernels/attention/test_cpu_attn.py::test_varlen_with_paged_kv[vec-False-False-1024-None-dtype0-None-96-96-num_heads2-seq_lens1] - AssertionError: Tensor-likes are not close!

Mismatched elements: 1 / 2148768 (0.0%)
Greatest absolute difference: 0.03125 at index (2356, 8, 34) (up to 0.015 allowed)
Greatest relative difference: 0.025634765625 at index (2356, 8, 34) (up to 0.01 allowed)
===== 1 failed, 161 passed, 1566 skipped, 2 warnings in 288.89s (0:04:48) ======
```

@bigPYJ1151 I'm happy to take a deeper look at this, unless you have hints on what might be the issue?

csrc/cpu/cpu_attn_macros.h
```diff
@@ -0,0 +1,63 @@
+#ifndef CPU_ATTN_MACROS_H
+#define CPU_ATTN_MACROS_H
+
+// x86_64
+#ifdef __x86_64__
+  #define FAST_SPINNING _mm_pause();
+
+  #ifdef __AVX512F__
+    #define DEFINE_FAST_EXP                                                    \
```
fadara01, line 9: what's stopping us from having this in `cpu_types_x86.hpp`? the same implementation in pytorch lives in the vectorization abstractions under [pytorch/aten/src/ATen/cpu/vec/vec512/vec512_float.h](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/cpu/vec/vec512/vec512_float.h#L315)

bigPYJ1151: I want to reduce the constant vectors initialization. Looks like gcc can't move them out of loops.

fadara01:
I also tested this new attention backend on `gpt-oss-20` (after applying #27024). The generated text is decent and aligns very closely with HF `transformers` generations.

fadara01 (approved):
LGTM - we'll follow-up with PRs to squeeze more perf out of this new attention backend on Arm CPUs.

jikunshang (approved):

PR #27959, opened 2025-11-03
[LoRA][FusedMoE] Introduce FusedMoEPermuteExpertsUnpermuteWithLoRA

varun-sundar-rabindranath:
## Purpose
This PR better integrates LoRA with FusedMoE modular kernel.
This PR also fixes an existing bug that occurs when using Chunking with LoRA 

TODO : Add design doc

## Test Plan
- Added a test_fused_moe_lora_layer.py that checks the plumbing
- Run `tests/lora/test_deepseekv2_tp.py`
- Run `tests/lora/test_gptoss.py`

## Test Result
Other than `tests/lora/test_gptoss.py` , all tests pass.
`tests/lora/test_gptoss.py` fails on `main` - I verified that this PR produces the same outputs as on `main`

### DIFF
--- a/vllm/model_executor/layers/fused_moe/fused_moe.py
+++ b/vllm/model_executor/layers/fused_moe/fused_moe.py
@@ -31,6 +31,9 @@ from vllm.model_executor.layers.fused_moe.deep_gemm_moe import (
     _valid_deep_gemm,
     deep_gemm_moe_fp8,
 )
+from vllm.model_executor.layers.fused_moe.mk_fused_experts_lora_support import (
+    MkFusedExpertsSupportsLoRA,
+)
 from vllm.model_executor.layers.fused_moe.moe_align_block_size import (
     moe_align_block_size,
 )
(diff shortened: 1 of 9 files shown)


jeejeelee:
I have removed the flaky test, see:https://github.com/vllm-project/vllm/pull/27966#event-20669165469

vllm/model_executor/layers/fused_moe/fused_moe.py
```diff
@@ -2108,9 +2111,13 @@ def apply(
             B_bias=self.w1_bias,
         )
 
-        self.activation(
-            activation, intermediate_cache2, intermediate_cache1.view(-1, N)
-        )
+        with self.maybe_activation_with_lora_hook(
+            gateup_proj_output=intermediate_cache1,
```
jeejeelee, line 2115: I think we should use `w13/w2` which is consistent with `fused_moe` name :sweat_smile:
This comment  also applies to all places that use gate up and down.

PR #27980, opened 2025-11-03
[Quantization] quantized kv cache loading for models like qwen

xuebwang-amd:
## Purpose
This PR aims to fix quantized kv cache model loading issues using AMD-Quark. 
The fix was initially applied to the `gpt-oss` model and may be extended to other models in `vllm/model_executor/models`, as they share a nearly identical methodology.

Here is the original error traceback w/o this PR fixing.
image

##
Here is a check list of some models defined in `vllm/model_executor/models/` for the supportness of quantized kv cache loading. Models with `get_cache_scale` called is marked as ✅, while models without `get_cache_scale` called is marked as ⚠️.

The ⚠️ symbol is preferred over ❌ since some models can intrinsically have non-quantized KV cache, for instance:
- VL models
- MLA models
- _etc_

Partial list is:
⚠️baichuan
⚠️bloom
✅ dbrx
⚠️ deepseek
⚠️falcon
⚠️ gemma
✅ gemma2
✅ gemma3
✅ gpt_j
⚠️ gpt_neox
⚠️ **_gpt_oss_**: this PR initially aims for
✅ grok1
✅ hunyuan_v1
✅ llama
✅ llama4
✅ mixtral
⚠️ mixtral
✅ phimoe
⚠️ qwen
✅ qwen2
⚠️ qwen2_moe
⚠️ qwen3_moe
...

## Test Result

---
<details>

</details>


vllm/model_executor/models/gpt_oss.py
```diff
@@ -514,6 +514,21 @@ def _load_weights_other(
             if is_pp_missing_parameter(name, self):
                 continue
 
+            if self.quant_config is not None and (
+                scale_name := self.quant_config.get_cache_scale(name)
```
BowenBao: Should we include both this `get_cache_scale` approach and `maybe_remap_kv_scale_name` like in https://github.com/vllm-project/vllm/blob/05cae69f0f4f0237779c45a3907aa13ee5a07667/vllm/model_executor/models/llama.py#L463-L479 ?

SageMoore (approved):

heheda12345:
CC @yewentao256

vllm/model_executor/models/gpt_oss.py
xuebwang-amd: That would be a good plus, as it appears to be functionality-safe.
The only concerning point is no handy models (gpt-oss and those ⚠️models listed on the top of this PR) to test non-None return values of this function. Hopefully this coverage can be added in the future.

vllm/model_executor/models/qwen.py
```diff
@@ -303,6 +306,25 @@ def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
...
+            if self.quant_config is not None and (
+                scale_name := self.quant_config.get_cache_scale(name)
+            ):
+                param = params_dict[scale_name]
+                weight_loader = getattr(param, "weight_loader", default_weight_loader)
+                if loaded_weight.numel() != 1:
+                    raise ValueError(
+                        f"KV cache scale '{scale_name}' is expected to be a scalar, "
+                        f"but got a tensor of shape {loaded_weight.shape}."
+                    )
```
yewentao256: These code repeats many times, could we encapsulate a function?

yewentao256 (commented):
Thanks for the work!

vllm/model_executor/models/qwen.py
xuebwang-amd: Good idea! That would be easily scalable across models.

yewentao256 (commented):
It is been a long time, is this problem still in main?

PR #27723, opened 2025-10-29
[benchmark] Make request IDs unique across clients by default

eicherseiji:
## Purpose
- Resolve https://github.com/vllm-project/vllm/issues/27711 
- RCA: scaled-out API servers don't coordinate `request_states`, leading to engine receiving non-unique request IDs (and bad downstream effects) instead of raising `ValueError` that a given request ID is already running. I think the standard `ValueError` is not seen with `api-server-count 1` due to queuing.
- Fix: Ensure `vllm bench serve` sends unique request IDs when clients run in parallel

## Test Plan
```shell
MODEL_PATH="Qwen/Qwen2.5-0.5B-Instruct"
MODEL_ID="qw-0.5B"
vllm serve "$MODEL_PATH" \
  --served-model-name "$MODEL_ID" \
  --api-server-count 2
```

Executed twice, simultaneously:
```shell
MODEL_ID="qw-0.5B"
MODEL_PATH="Qwen/Qwen2.5-0.5B-Instruct"

    vllm bench serve \
        --model "$MODEL_PATH" \
        --served-model-name "$MODEL_ID" \
```
## Test Result
- No engine crash due to duplicate request IDs is seen

Screenshot 2025-10-29 at 1 28 53 AM

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x] The test plan, such as providing test command.
- [x] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #20139, opened 2025-06-26
[PD][Nixl] Remote consumer READ timeout for clearing request blocks 

NickLucche:
With https://github.com/vllm-project/vllm/pull/19223, we're addressing most of the cases where P request blocks may be left starving.
However, there are still cases where if the router fails to communicate request abortion for whatever reason (eg in-flight request lost, router down..) while the request has not yet reached D or D fails to communicate the abortion to P, where the remote producer may be left with blocks that won't be cleared.

This PR addresses these final edge-cases by attaching a simple TTL to every request that needs to be read from local(D)<-remote (P).  

cc @njhill

PR #10968, opened 2024-12-06
[Frontend] Use request id from header

joerunde:
This is a small proposal to populate the request ids in the engine from the `X-Request-Id` header when running the openai server. This will benefit traceability: when users pass in the header they can correlate vLLM logs with the request that generated them, something that our users are asking for. If the header is not passed, behavior remains the same.

Something to consider is the implications of users potentially passing in the same request id multiple times- nothing seems to crash when that happens, but in that case it would be hard to distinguish logs from different requests.

PR #27987, opened 2025-11-03
[Core] Add a random suffix to frontend-provided request IDs

markmc:
Since #9550 and #10968 we support client's supplying a custom request ID. The motivation for this is that it can be very helpful when you need to correlate vLLM logs with logs of a related service.

Since the request ID is used ubiquitously across vLLM as a unique key, it obviously is problematic if we ever have multiple in-flight requests using the same client-provided request ID.

We saw this happening recently when `vllm serve bench` started including a request ID and the request IDs from multiple concurrent instances caused collisions. See #27723

We try to guard against request ID collisions currently in the frontend in OutputProcessor:

```
    def add_request(...):
        if request_id in self.request_states:
            raise ValueError(f"Request id {request_id} already running.")
```

however, this is not always effective:

1) We can have abort race conditions where a request is no longer tracked by the frontend, but still not completed in the engine. See #15326 for an attempt to fix this.
2) We can have async scheduling race conditions where a request ID is removed from the output processor and being scheduled while the older request with that ID is still being completed
   by the model runner. See #29355
3) With P/D, a request will continue to be tracked by the prefill engine long after the prefill request has been completed in the frontend, while we wait for the decode side to fetch the KV blocks. See #20139

Let's instead ensure we use a unique request ID internally, even when a client provides a custom request ID. We can do this simply by appending a short random suffix to any request ID provided by the frontend.

A full 32 character random UUID would be overkill as a suffix, so how many random characters would be sufficient? 8 characters gives us 32 bits of entropy, or 16^8 possible prefixes.

Using the collision probability approximation from https://preshing.com/20110504/hash-collision-probabilities:

N = 16^8 and k is the number of generated suffixes, then the probability of collision is (k^2)/(2N), so If a client somehow
caused vLLM to hold 10k requests that reuse the same client-provided ID, then there would be a 1.16% chance of collision:

```
>>> (k**2)/(2*N)
0.011641532182693481
```

That seems (super good enough)[https://hownot2.com/products/hownot2-super-good-enough-t-shirt].

The key changes to support this are:

1. `InputProcessor.process_inputs()` - we add some randomness to the request ID just before creating an `EngineCoreRequest`, and store both the random "internal" request ID (as `request_id`) and the supplied "external" request ID (as `external_req_id`) in the `EngineCoreRequest`.
2. `RequestState.make_request_output()` - we ensure that `RequestOutput.request_id` continues to be the external request ID (for backwards compat) 
3. `OutputProcessor.abort_requests()` - we make `OutputProcessor` track a mapping from external request ID to internal request IDs, so `abort_requests()` can abort based on either ID.
4. `AsyncLLM` - we use `RequestOutputCollector` to track the internal request ID, so we can use the internal ID to abort an in-progress request. We also add an `internal` boolean flag to `abort()` so API users can abort based on either ID.
5. `ParentRequest` - in the case of parallel sampling, we need to track both the internal and external ID for the later creation of `RequestOutput` aggregating the child outputs.

We need to ensure we track the external->internal request ID mapping because `abort()` will be supplied an external request ID. In the case where an external request ID maps to multiple running requests, we assume the caller requires all of those requests to be aborted. The caller can use `EngineCoreRequest.request_id` or `RequestOutput.internal_request_id` as the request ID (with `internal=True`) if they want to be more specific.


vllm/entrypoints/llm.py
```diff
@@ -1700,15 +1700,30 @@ def _add_request(
...
+
+    @staticmethod
+    def _sort_outputs(
+        outputs: list[RequestOutput | PoolingRequestOutput],
+    ) -> list[RequestOutput | PoolingRequestOutput]:
+        # Sort the outputs by request ID.
+        # This is necessary because some requests may be finished earlier than
+        # its previous requests.
+
+        # Extract the original request ID prefix for sorting.
```
markmc: I considered adding `external_request_id` to `RequestOutput` instead of this hack ...

vllm/v1/engine/input_processor.py
```diff
@@ -382,6 +382,12 @@ def _extract_mm_data(p: PromptType):
             mm_uuids[modality] = [f"{request_id}-{modality}-{i}" for i in range(n)]
         return mm_uuids
 
+    def _generate_request_id(self, request_id: str):
+        """Construct an internal request ID by adding 8 random characters
+        to the supplied request ID in order to ensure uniquness.
+        """
+        return f"{request_id}-{random_uuid()[:8]}"
```
njhill: should be slightly more efficient
```suggestion
        return f"{request_id}-{random_uuid():.8}"
```

vllm/v1/engine/input_processor.py
```diff
@@ -409,6 +415,9 @@ def process_inputs(
         if arrival_time is None:
             arrival_time = time.time()
 
+        external_req_id = request_id
```
njhill: Perhaps we can do this as a follow-on, but it might be good to make the external req id optional so in the default case we can then have a more compact internal id.

We could then also avoid generating multiple uuids per request in the api server case.

vllm/v1/engine/async_llm.py
```diff
@@ -304,6 +304,12 @@ async def add_request(
         # Convert Input --> Request.
         if isinstance(prompt, EngineCoreRequest):
             request = prompt
+            if request_id != request.request_id:
+                logger.warning_once(
+                    "AsyncLLM.add_request() was passed a request_id parameter that "
+                    "does not match the EngineCoreRequest.request_id attribute. The "
+                    "latter will be used, and the former will be ignored."
+                )
```
njhill, line 301: :+1: 

A bit tangential but I don't think it's good that we have duplicated args here, have commented about it here https://github.com/vllm-project/vllm/issues/23869#issuecomment-3577671583

vllm/v1/engine/output_processor.py
```diff
@@ -378,8 +382,24 @@ def abort_requests(
         self,
         request_ids: Iterable[str],
     ) -> list[str]:
-        request_ids_to_abort = []
+        internal_req_ids = []
         for request_id in request_ids:
+            # External ID - abort all requests in the external->internal mapping
+            if request_id in self.external_req_ids:
+                internal_req_ids.extend(self.external_req_ids.pop(request_id))
```
njhill: can avoid double lookup
```suggestion
            if (internal_ids := self.external_req_ids.pop(request_id)) is not None:
                internal_req_ids.extend(internal_ids)
```

vllm/v1/engine/output_processor.py
```diff
@@ -378,8 +382,24 @@ def abort_requests(
...
+                continue
```
njhill: May as well remove this continue, just in case someone decides to use an internal id as an external one. I know that would be an obscure thing to happen but might as well handle it.

vllm/v1/engine/output_processor.py
```diff
@@ -438,6 +458,11 @@ def add_request(
         if parent_req:
             self.parent_requests[parent_req.request_id] = parent_req
 
+        # Track the external_req_id -> [internal_req_id, ...] mapping
+        if req_state.external_req_id not in self.external_req_ids:
+            self.external_req_ids[req_state.external_req_id] = []
```
njhill: Could consider using a `defaultdict(list)` for this, then these lines wouldn't be needed.

vllm/v1/engine/output_processor.py
```diff
@@ -521,6 +546,10 @@ def process_outputs(
             # Free completed requests.
             if finish_reason is not None:
                 self.request_states.pop(req_id)
+                self.external_req_ids[req_state.external_req_id].remove(req_id)
+                if not self.external_req_ids[req_state.external_req_id]:
+                    del self.external_req_ids[req_state.external_req_id]
```
njhill: similarly avoid multiple lookups
```suggestion
                internal_ids := self.external_req_ids.get(req_state.external_req_id)
                if internal_ids is not None:
                    internal_ids.remove(req_id)
                    if not internal_ids:
                       del self.external_req_ids[req_state.external_req_id]
```

vllm/entrypoints/pooling/embed/serving.py
```diff
@@ -536,8 +536,8 @@ async def _collect_batch(
                     # Non-chunked result - extract prompt_idx from request_id
                     parts = result.request_id.split("-")
                     try:
-                        # Last part should be prompt index
-                        prompt_idx = int(parts[-1])
+                        # Second-to-last part should be prompt index
+                        prompt_idx = int(parts[-2])
```
njhill: Yuck, the way the request id is abused in this file is pretty bad. What if an external request id is sent in that happens to contains `-chunk-` or whatever.

(not related to your changes, just hadn't seen this code before!)

vllm/v1/engine/output_processor.py
```diff
@@ -378,8 +382,24 @@ def abort_requests(
...
+                continue
+
+            # Internal ID - remove it from the external->internal mapping
+            if request_id not in self.request_states:
+                continue
+            internal_req_ids.append(request_id)
+            external_req_id = self.request_states[request_id].external_req_id
+            self.external_req_ids[external_req_id].remove(request_id)
+            if not self.external_req_ids[external_req_id]:
+                del self.external_req_ids[external_req_id]
```
njhill: can similarly avoid double lookups here
```suggestion
            if req_state = self.request_states.get(request_id)
            if req_state is not None:
                internal_req_ids.append(request_id)
                internal_ids = self.external_req_ids[external_req_id].get(req_state.external_req_id)
                internal_ids.remove(request_id)
                if not internal_ids:
                    del self.external_req_ids[external_req_id]
```

Also, I think we need to handle the parent id case. If we get the internal id of a n>1 request, we need to handle that similarly, it will be in `self.parent_requests` rather than `self.request_states`.

vllm/v1/engine/output_processor.py
```diff
@@ -378,8 +382,24 @@ def abort_requests(
         self,
         request_ids: Iterable[str],
```
njhill: We could also consider adding an `internal: bool = False` arg here to allow caller to specify whether or not the ids are internal or external. This might be a bit more robust, for all the places we call abort internally, we should be able to use the internal id and set it to `True`, and external usage shouldn't be affected.

njhill (commented):
Thanks very much for this @markmc!

Some thoughts in addition to inline comments:
- I was thinking in `AsyncLLM` we should have the `add_request` method return the internal request id as well as the output collector, and then use this in the aborts in the except blocks in `generate()`
- Wondering whether we should set the request id in the outputs from the engine to be the external id (if provided). This would be more backwards compatible and makes more sense to me logically. It also means we wouldn't need the new sorting "hack" in llm.py

vllm/entrypoints/pooling/embed/serving.py
markmc: Agree!

vllm/v1/engine/async_llm.py
markmc: Agree!

vllm/v1/engine/input_processor.py
markmc: Done, thanks

vllm/v1/engine/input_processor.py
markmc: Yeah, agree. I'm trying to limit the invasiveness of this PR though. Making `request_id` optional at the `AsyncLLM` level, then removing the UUID generation in the API server, and figuring out what to do about the per-endpoint prefix (e.g. "cmpl-") ... it's a relatively heavy lift that could slow down fixing the issue.

vllm/v1/engine/output_processor.py
markmc: Not a fan of this in terms of readability, but done. Thanks

(Also, needed to pass `None` default to `pop()`)

vllm/v1/engine/output_processor.py
markmc: If the request ID isn't tracked as an external ID, it is an internal ID

If it is not in `request_states`, then we have no way of mapping this internal ID to the external one

Or, put another way - the code beyond the continue requires the `request_states` lookup to succeed?

In the code suggestion below, you seem to be proposing moving from:

```
if request_id not in self.request_states:
    continue
internal_req_ids.append(request_id)
[...]    
```

to:

```
req_state = self.request_states.get(request_id)
if req_state is not None:
    internal_req_ids.append(request_id)
    [...]    
```

Which is equivalent?

Just spelling this out because I might well be missing your point

markmc: Or maybe you just mean do this:

```
internal_req_ids.append(request_id)
req_state = self.request_states.get(request_id)
if req_state is not None:
    [...]    
```

which does make sense if this is the internal request ID of a parent request ... got it I think, thanks!

vllm/v1/engine/output_processor.py
markmc: Thanks, great catch on the parent request thing :+1:

vllm/v1/engine/output_processor.py
markmc: Done, thanks

vllm/v1/engine/output_processor.py
markmc: Done

vllm/v1/engine/output_processor.py
markmc: Done, thanks

vllm/v1/engine/output_processor.py
```diff
@@ -377,9 +387,60 @@ def propagate_error(self, e: Exception):
...
+
+        In the case of parallel sampling, a request ID may be used to identify
+        a parent request, in which case the associated child requests are aborted
+        also.
+        """
+
+        if internal is not None:
+            check_internal, check_external = internal, not internal
+        else:
+            check_internal, check_external = True, True
```
njhill: Do we need to support `None` case? Thought it would be a bit simpler if we dont...

vllm/v1/engine/output_processor.py
njhill: I just meant here in theory someone could create a new request with an external id that matches an existing internal id.. (which would then get it's own internal id). However if we get rid of the `internal=None` case then this comment should be n/a.

vllm/v1/engine/output_processor.py
njhill: Sorry, could also have the `internal_ids = ...pop` on a separate line if that helps

njhill:
> > > > > * I was thinking in `AsyncLLM` we should have the `add_request` method return the internal request id as well as the output collector, and then use this in the aborts in the except blocks in `generate()`
> > > > 
> > > > 
> > > > Yeah, `RequestOutputCollector` has this ID, I've used it there
> > > 
> > > 
> > > Doh ... it only has it when it has output! So, that's not going to work. Will look for an alternative
> > 
> > 
> > I think it would make sense to have it just return `tuple[RequestOutputCollector, str]`?
> 
> What about API compatibility? (Do you not consider this public?)

I'm actually not sure about this. I'm pretty sure it would only be used via the `generate()/encode()` methods. I think it's unlikely folks would be using it directly. If we are concerned, we could add bool `return_id` arg to control what's returned.

markmc:
Ok, the changes are quite scattered, so here's a condensed summary of where we're at now

1. `InputProcessor.process_inputs()` - we add some randomness to the request ID just before creating an `EngineCoreRequest`, and store both the random "internal" request ID (as `request_id`) and the supplied "external" request ID (as `external_req_id`) in the `EngineCoreRequest`.
2. `RequestState.make_request_output()` - we ensure that `RequestOutput.request_id` continues to be the external request ID (for backwards compat) and add `internal_request_id`.
3. `OutputProcessor.abort_requests()` - we make `OutputProcessor` track a mapping from external request ID to internal request IDs, so `abort_requests()` can abort based on either ID.
4. `AsyncLLM` - we use `RequestOutputCollector` to track the internal request ID, so we can use the internal ID to abort an in-progress request. We also add an `internal` boolean flag to `abort()` so API users can abort based on either ID.
5. `ParentRequest` - in the case of parallel sampling, we need to track both the internal and external ID for the later creation of `RequestOutput` aggregating the child outputs.

vllm/outputs.py
```diff
@@ -130,6 +132,7 @@ def __init__(
                 "RequestOutput: Ignoring extra arguments: %s", str(kwargs)
             )
         self.request_id = request_id
+        self.internal_req_id = internal_req_id
```
njhill: Pondering: do we even need to include the internal_req_id in the outputs?

vllm/v1/engine/output_processor.py
```diff
@@ -39,13 +40,16 @@ class RequestOutputCollector:
...
+    def __init__(self, output_kind: RequestOutputKind, request_id: str):
         self.aggregate = output_kind == RequestOutputKind.DELTA
+        self.request_id = request_id
         self.output: RequestOutput | PoolingRequestOutput | Exception | None = None
         self.ready = asyncio.Event()
 
     def put(self, output: RequestOutput | PoolingRequestOutput | Exception) -> None:
         """Non-blocking put operation."""
+        if isinstance(output, (RequestOutput, PoolingRequestOutput)):
+            assert output.internal_req_id == self.request_id
```
njhill: could put this in the corresponding elif blocks below to save an additional if (assuming we keep internal_req_id in the outputs, per other comment)

vllm/v1/engine/output_processor.py
njhill: bumping this question

njhill (commented):
Thanks @markmc, I think it's looking pretty good.

njhill:
From the CI, looks like there's another test update to needed to `test_output_processor.py`. We should run the "full" CI on this one too to flush out anything else. And I guess should queue it behind https://github.com/vllm-project/vllm/pull/29665 so as not to break P/D.

I still think it's worth looking at making the provided request id optional as a follow-on.

vllm/outputs.py
markmc: I think we could probably get away with dropping it

My main thinking for keeping it was that if `abort(request_id, internal=True)` is to be useful, the API user needs to know the internal request ID

Also, now that we do add some randomness to the request ID, it become valid for an API user to pass the same request ID for lots of requests ... and it makes sense that you would want an ID to distinguish between them

vllm/v1/engine/output_processor.py
markmc: The next line makes it a bit tricky:

```
        if self.output is None or isinstance(output, Exception):
            self.output = output
```

The assertion wouldn't apply to the `Exception` case. That's why I did it here

vllm/v1/engine/output_processor.py
markmc: Good point, dropping that now (I had locked myself into assuming we needed it for the default, backwards compat case). Thanks

vllm/v1/engine/output_processor.py
njhill: right, I meant the same assert could be included in both of the `elif` blocks below that one

markmc: But I do want it to apply in the `self.output is None and not isinstance(output, Exception)` situation too ...

vllm/outputs.py
njhill: >  if `abort(request_id, internal=True)` is to be useful, the API user needs to know the internal request ID

I don't think it needs to be useful to users necessarily because it's internal. They can still obtain the internal id if desired when submitting a request via add_request (in both async and non-async cases).

> Also, now that we do add some randomness to the request ID, it become valid for an API user to pass the same request ID for lots of requests ... and it makes sense that you would want an ID to distinguish between them

But IMO if the user cares about that, it's up to them to pass unique ids themselves. They can still choose to not pass any id, in which case the external id would become the internal id anyhow (if we do the suggested follow-on).


Once a request is submitted, there could be some indefinite amount of time before any output is received, so relying on getting an id from that output is not super useful for aborting purposes anyhow.

The way I am thinking about it, the only reason for having this internal id is to ensure uniqueness internally so that the system will always behave "correctly".

markmc: I'll remove it now, but ...

> The way I am thinking about it, the only reason for having this internal id is to ensure uniqueness internally so that the system will always behave "correctly".

But `abort(request_id, internal=True)` looks like a public API?

markmc:
Ok, the `RequestOutput.internal_req_id` removal will have to wait until tomorrow. Hopefully what I've pushed now passes CI (I've tried to do a sweep of relevant tests locally) and I think resolves pretty much everything but the `RequestOutput.internal_req_id` thing

vllm/v1/engine/output_processor.py
njhill: oops sorry my bad I misread the code :facepalm:

vllm/v1/engine/input_processor.py
markmc: Filed as #30138

vllm/outputs.py
markmc: Removed now

markmc:
> I still think it's worth looking at making the provided request id optional as a follow-on.

Filed as #30138

vllm/v1/engine/output_processor.py
```diff
@@ -374,12 +382,41 @@ def propagate_error(self, e: Exception):
...
+                internal_req_ids.append(request_id)
+
+                # Remove internal ID from the external->internal mapping
+                if req_state := self.request_states.get(request_id):
+                    external_req_id = req_state.external_req_id
+                    internal_ids = self.external_req_ids[external_req_id]
+                    internal_ids.remove(request_id)
+                    if not internal_ids:
+                        del self.external_req_ids[external_req_id]
+            elif internal_ids := self.external_req_ids.pop(request_id, []):
```
njhill, line 416: nit: can avoid creating the empty list
```suggestion
            elif internal_ids := self.external_req_ids.pop(request_id, None):
```

vllm/v1/engine/output_processor.py
```diff
@@ -374,12 +382,41 @@ def propagate_error(self, e: Exception):
...
+        internal_req_ids = []
         for request_id in request_ids:
+            if internal:
+                # Internal ID - this may be a parent request
+                internal_req_ids.append(request_id)
+
+                # Remove internal ID from the external->internal mapping
+                if req_state := self.request_states.get(request_id):
+                    external_req_id = req_state.external_req_id
+                    internal_ids = self.external_req_ids[external_req_id]
```
njhill, line 412: I'm not sure about this but wondering whether we should defensively use `.get()` here just in case it's not in the dict.

I guess it would indicate a bug if it wasn't, so might be ok/better to keep as-is.

vllm/v1/engine/__init__.py
```diff
@@ -49,6 +49,7 @@ class EngineCoreRequest(
     gc=False,
 ):  # type: ignore[call-arg]
     request_id: str
+    external_req_id: str
```
njhill: Just wondering how it would shake out if we were to move this field to the end with a `None` or possibly `""` default value (since fields with defaults must come after those without), and have a `__post_init__()` method that sets it to `request_id` if it is `None` (or `""`).

I know the "allow req id to be omitted externally" is planned as a follow-on but doing this here might reduce the impact of the change a bit (and in case there's out-of-tree code which uses this class, even though I don't  this we consider it part of public interface). In particular I think quite a few of the test fixes could then be reverted?

njhill (commented):
Thanks @markmc it's looking great, just some fairly minor remaining thoughts.

vllm/v1/engine/output_processor.py
markmc: Yeah, that's what I started with, but:

```
$ mypy --python-version 3.10 --follow-imports skip 
vllm/v1/engine/output_processor.py:414: error: Argument 2 to "pop" of "dict" has incompatible type "None"; expected "list[str]"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

vllm/v1/engine/output_processor.py
markmc: Yeah, my view is defensiveness here would just hide a bug

vllm/v1/engine/__init__.py
markmc: Maybe, especially if the `__post_init__()` idea meant you didn't have to have `if not request.external_req_id` checks all over ... but you'd still need assertions for type hint checking?

I guess I was reluctant to do this because this is all complex enough without also having to consider a case where the external ID wasn't specified (even if we default to the internal ID)

Updating the tests doesn't really worry me ... I was more reassured by the fact there is only one place in vLLM  itself that creates an `EngineCoreRequest`

markmc:
I presume the `tp_config_sweep_accuracy_test.sh` failure:

> Potentially invalid KV blocks for unrecognized request cmpl-4fd0bde0-eb78-4c6a-94e9-40ba29e6645a-0-a8e879dc were retrieved by a decode worker

will be resolved by rebasing onto #29665

njhill (approved):
I am pretty sure the blackwell and distributed h200 test failures are unrelated but will confirm before we merge this.

vllm/v1/engine/parallel_sampling.py
```diff
@@ -33,8 +33,12 @@ class ParentRequest:
     cached_child_sampling_params: SamplingParams | None
 
     def __init__(
-        self, request_id: str, external_req_id: str, sampling_params: SamplingParams
+        self,
+        request_id: str,
+        external_req_id: str | None,
+        sampling_params: SamplingParams,
```
njhill: I think we could have this just take the request since these are all fields from there.

vllm/v1/engine/input_processor.py
```diff
@@ -403,14 +403,19 @@ def _extract_mm_data(p: PromptType):
...
 
-    def _generate_request_id(self, request_id: str):
-        """Construct an internal request ID by adding 8 random characters
-        to the supplied request ID in order to ensure uniquness.
+    def assign_request_id(self, request: EngineCoreRequest):
+        """Replace the externally supplied request ID with an internal request ID
+        that adds 8 random characters in order to ensure uniquness.
         """
-        return f"{request_id}-{random_uuid():.8}"
+        assert request.external_req_id is None
```
njhill: This should probably be changed to raise a `ValueError`. It should be ok to fail the request in this case rather than crash the server.

vllm/v1/engine/__init__.py
```diff
@@ -75,6 +75,10 @@ class EngineCoreRequest(
 
     trace_headers: Mapping[str, str] | None = None
 
+    # The user-provided request ID, to support abort(req_id, internal=False)
+    # See InputProcessor.assign_request_id()
```
njhill:
```suggestion
    # The user-provided request ID. This field is set internally,
    # copied from the provided request_id that's originally assigned
    # to the request_id field, see InputProcessor.assign_request_id().
    # Used in outputs and to support abort(req_id, internal=False).
```

It still feels quite nasty that the `request_id` field has a different purpose depending on the context, but I think is just a consequence of using the internal datastructure as part of an external API (which we can hopefully address separately).

vllm/v1/engine/input_processor.py
```diff
@@ -403,9 +403,20 @@ def _extract_mm_data(p: PromptType):
...
                 else 1
             )
-            mm_uuids[modality] = [f"{request_id}-{modality}-{i}" for i in range(n)]
+            # request_id is externally provided, add randomness to ensure uniqueness
+            mm_uuids[modality] = [
+                f"{request_id}-{modality}-{random_uuid():.8}-{i}" for i in range(n)
+            ]
         return mm_uuids
 
+    def assign_request_id(self, request: EngineCoreRequest):
```
njhill: could make this a `@staticmethod` to make it clear that it's stateless

vllm/v1/engine/input_processor.py
```diff
@@ -403,9 +403,20 @@ def _extract_mm_data(p: PromptType):
                 if isinstance(data, list) or MultiModalDataParser.is_embeddings(data)
                 else 1
             )
-            mm_uuids[modality] = [f"{request_id}-{modality}-{i}" for i in range(n)]
+            # request_id is externally provided, add randomness to ensure uniqueness
+            mm_uuids[modality] = [
+                f"{request_id}-{modality}-{random_uuid():.8}-{i}" for i in range(n)
+            ]
```
njhill: I'm actually unsure whether this is necessary, given what the mm_uuids are used for. If it's only for caching purposes and doesn't need to be unique for correctness, then it might be that using the external request id is better anyhow. Will check with @ywang96

njhill (commented):
Thanks @markmc! Looking forward to landing this!

vllm/v1/engine/input_processor.py
njhill: Confirmed with @ywang96 that this doesn't need to be unique.

ywang96 (approved):
LGTM - not particular concern from me

PR #27990, opened 2025-11-03
[flashinfer][fix] do not check nvcc availability when using pre-downloaded cubins

mxz297:
Summary: https://github.com/vllm-project/vllm/pull/26443 adds checking of availability of nvcc as a condition to enable flashinfer moe. In our deployment env, there is no nvcc, so flashinfer moe is disabled
Differential Revision: D86104899


mxz297:
> Is nvcc required by the jit compilation of FlashInfer?

Yes, so it makes sense to take the recommendation from the gemini-code-assist to guard the check with VLLM_HAS_FLASHINFER_CUBIN

houseroad (approved):
Looks good.

PR #27993, opened 2025-11-03
[Kernel] Remove Redundant Prefill Support From 3D Triton Attention Kernel

jvlunteren:
## Purpose

This pull request removes support for prefills from the 3D kernel and its corresponding reduce kernel.

The 3D and reduce kernels implement parallel tiled softmax computation, which is particularly advantageous in scenarios with small batch sizes and long sequences (i.e., long KV contexts). In such cases, these kernels enable increased parallelism, allowing for more effective utilization of GPU resources.

Prefill operations, however, typically involve many query tokens which inherently provide sufficient parallelism. In these cases, the 2D kernel is better suited for efficient processing, rendering the 3D kernel support for prefills unnecessary.

By removing this redundant support, the kernel logic is simplified without compromising performance.

## Test Plan

Existing tests can be used. For example the unit test `./tests/kernels/attention/test_triton_unified_attention.py` and `lm_eval`

## Test Result

================================================ 256 passed in 38.58s ================================================
</pre>

<br>
<br>

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.798|±  |0.0180|
|     |       |strict-match    |     5|exact_match|↑  |0.786|±  |0.0184|</pre>

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.792|±  |0.0182|
|     |       |strict-match    |     5|exact_match|↑  |0.770|±  |0.0188|</pre>

@tdoublep @bringlein


tdoublep (commented):
(documenting an offline discussion here for transparency)

This PR looks basically fine. Since we are only using the 3D kernel right now for batches that have `max_query_length=1`, these changes would make the kernel simpler and not change any behavior. 

However, I'm concerned whether the condition we are using on main is not really the right thing when using speculative decoding. Don't we want to use the 3D kernel for batches that have `max_query_length=num_speculative_tokens`? If that is the condition, then would it still be safe to strip out this code?

jvlunteren:
Closed.

PR #27994, opened 2025-11-03
[FlashInfer] Avoid FlashInfer block_size 16 + head_size 256 on blackwell

heheda12345:
## Purpose
Avoid this combination as https://github.com/flashinfer-ai/flashinfer/issues/1993 reports this combination is not correct.

For most models with head_size 256, users now need --block_size 32 / --block_size 64

For hybrid mamba like qwen3-next, the block_size can be resolved automatically 

Thanks @vadiklyutiy for the exploration on this problem https://github.com/vllm-project/vllm/pull/27704

## Test Plan

Test non-hybrid case and hybrid case

## Test Result
All tests are on B200

### Normal model

I don't know a model with head_size 256. I've changed the 256 in this PR to 64 and test opt-125m
```
vllm serve facebook/opt-125m
```
will result in
```
AssertionError: There is a bug in FlashInfer block_size 16 head size 256 support. Please avoid this combination by passing --block-size 32 or --block-size 64.
```

```
vllm serve facebook/opt-125m --block-size 64
```
works

### Hybrid
```
vllm serve Qwen/Qwen3-Next-80B-A3B-Instruct -tp 4
```
before this PR
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.2123|±  |0.0113|
|     |       |strict-match    |     5|exact_match|↑  |0.1933|±  |0.0109|
```
after this PR
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8491|±  |0.0099|
|     |       |strict-match    |     5|exact_match|↑  |0.8120|±  |0.0108|
```

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/models/config.py
+++ b/vllm/model_executor/models/config.py
@@ -7,6 +7,7 @@ from typing import TYPE_CHECKING
 import vllm.envs as envs
 from vllm.logger import init_logger
 from vllm.model_executor.models import ModelRegistry
+from vllm.platforms import current_platform
 from vllm.utils.math_utils import cdiv, round_up
 from vllm.utils.torch_utils import STR_DTYPE_TO_TORCH_DTYPE
 from vllm.v1.kv_cache_interface import FullAttentionSpec, MambaSpec, MLAAttentionSpec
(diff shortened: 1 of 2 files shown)


vllm/model_executor/models/config.py
```diff
@@ -364,6 +365,17 @@ def verify_and_update_config(cls, vllm_config: "VllmConfig") -> None:
...
+                current_platform.is_device_capability(100)
+                and model_config.get_head_size() == 256
+                and (
+                    envs.VLLM_ATTENTION_BACKEND is None
+                    or envs.VLLM_ATTENTION_BACKEND == "FLASHINFER"
+                )
+            ):
+                # https://github.com/flashinfer-ai/flashinfer/issues/1993 reports that`
+                # head size 256 and block size 16 is not supported on blackwell.
+                kernel_block_alignment_size = 32
```
MengqingCao, line 378: nit: How about make the logic of `kernel_block_alignment_size` into each attention backends? Actually this is need in vllm-ascend, too. If this is acceptable, I could help to make a follow-up pr

heheda12345: Yes it make sense. Help wanted on it. And I think the alignment can be inferred from `get_supported_kernel_block_size`

MengqingCao: Okay, I'll work on it then.

MengqingCao:
> @MengqingCao are you talking about something like I did in #27704

Yes, thanks for the context! I'll take a look at your pr

vllm/model_executor/models/config.py
pavanimajety: For my knowledge, what is kernel_block_alignment_size?

pavanimajety (approved):
LGTM, we can undo the change once flashinfer is fixed.

PR #27997, opened 2025-11-03
Remove the tpu docker image nightly build.

QiliangCui:
## Purpose

Stop building the vllm-tpu docker image because the vllm-tpu image is now built from [tpu-inference](https://github.com/vllm-project/tpu-inference).  

## Test Plan
After the change, I should monitor to make sure no new docker image will be built.


yaochengji (approved):
LGTM!

Issue #23120, opened 2025-08-18
[Bug]: Structured output is not correctly enforced when using GPT-OSS


jmracek:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
Collecting environment information...
==============================
        System Info
==============================
OS                           : Ubuntu 22.04.5 LTS (x86_64)
GCC version                  : (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Clang version                : Could not collect
CMake version                : version 3.22.1
Libc version                 : glibc-2.35

==============================
       PyTorch Info
==============================
PyTorch version              : 2.9.0.dev20250804+cu128
Is debug build               : False
CUDA used to build PyTorch   : 12.8
ROCM used to build PyTorch   : N/A

==============================
      Python Environment
==============================
Python version               : 3.12.8 (main, Jan 14 2025, 22:49:14) [Clang 19.1.6 ] (64-bit runtime)
Python platform              : Linux-5.10.223-212.873.amzn2.x86_64-x86_64-with-glibc2.35

==============================
       CUDA / GPU Info
==============================
Is CUDA available            : True
CUDA runtime version         : 12.6.85
CUDA_MODULE_LOADING set to   : LAZY
GPU models and configuration : 
GPU 0: NVIDIA H100 80GB HBM3
GPU 1: NVIDIA H100 80GB HBM3
GPU 2: NVIDIA H100 80GB HBM3
GPU 3: NVIDIA H100 80GB HBM3
GPU 4: NVIDIA H100 80GB HBM3
GPU 5: NVIDIA H100 80GB HBM3
GPU 6: NVIDIA H100 80GB HBM3
GPU 7: NVIDIA H100 80GB HBM3

Nvidia driver version        : 550.90.07
cuDNN version                : Probably one of the following:
/usr/lib/x86_64-linux-gnu/libcudnn.so.9.6.0
/usr/lib/x86_64-linux-gnu/libcudnn_adv.so.9.6.0
/usr/lib/x86_64-linux-gnu/libcudnn_cnn.so.9.6.0
/usr/lib/x86_64-linux-gnu/libcudnn_engines_precompiled.so.9.6.0
/usr/lib/x86_64-linux-gnu/libcudnn_engines_runtime_compiled.so.9.6.0
/usr/lib/x86_64-linux-gnu/libcudnn_graph.so.9.6.0
[...]

PR #28000, opened 2025-11-03
[Bugfix] Fix and add tests for GptOss reasoning parser

benchislett:
## Purpose

GPT-OSS can escape structured outputs by emitting tokens between `<|channel|>final` and `<|message|>`, for example: `<|start|>assistant<|channel|>final <|constrain|>JSON<|message|>content...`.

This PR updates the parser to look for `<|start|>assistant<|channel|>final` and then the suffix `<|message|>`. This doesn't cover all cases, such as tool calling (see #23120), but it does workaround a few common issues.

## Test Plan

Added a unit test for GptOssReasoningParser to target `is_reasoning_end` specifically. Future work can extend this to cover more cases as needed.

## Test Result

Tests now passing.

I also ran the benchmark script:
```
python3 ../benchmarks/benchmark_serving_structured_output.py --backend openai-chat --model openai/gpt-oss-120b --dataset xgrammar_bench --structured-output-ratio 1.0 --request-rate 100 --num-prompts 100 --port 8046 --endpoint /openai/v1/chat/completions --output-len 900 --save-results
```

and see 98% accuracy before (same as with ratio set to 0.0. At least one confirmed failure case due to this feature gap), and 100% accuracy after this change.


vllm/reasoning/gptoss_reasoning_parser.py
```diff
@@ -68,18 +68,35 @@ class GptOssReasoningParser(ReasoningParser):
...
+                    suffix_start, len(input_ids) - len(end_token_ids_suffix) + 1
+                ):
+                    if j - suffix_start >= self.reasoning_max_num_between_tokens:
+                        break
+                    if (
+                        input_ids[j : j + len(end_token_ids_suffix)]
+                        == end_token_ids_suffix
+                    ):
+                        return True
         return False
```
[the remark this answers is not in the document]
benchislett: I am not concerned here since all lengths are small, and we are only doing extra work for those rare few tokens inside the section of the chat template for the "final" message. I also put a hardcoded upper-bound so it doesn't get too out of control

vllm/reasoning/gptoss_reasoning_parser.py
```diff
@@ -68,18 +68,35 @@ class GptOssReasoningParser(ReasoningParser):
 
     def __init__(self, tokenizer: PreTrainedTokenizerBase, *args, **kwargs):
         super().__init__(tokenizer, *args, **kwargs)
-        self.reasoning_end_token_ids = self.model_tokenizer.encode(
-            "<|start|>assistant<|channel|>final<|message|>"
+        # The model can output some special tokens between "final" and "<|message|>"
+        # So we need to look for both sequences to determine the end of reasoning.
+        self.reasoning_end_token_ids_prefix = self.model_tokenizer.encode(
+            "<|start|>assistant<|channel|>final"
```
alecsolder: So the model will only output `<|start|>assistant` if it does decide to output a reasoning message, in some situations it may just jump straight to the final message, so it may be more reasonable to just do `<|channel|>final` instead as it will cover both situations.

To cover tool calling scenarios, as soon as you see `to=` in the header, then the message is a tool call and no longer reasoning.

Both situations can be covered by the ending suffix of `<|message|>` though.

benchislett: updated to cover the first case. I think it would be easiest for me to leave any tool calling updates to follow-up work, as I have no context as to how we do them in vLLM. I'm not entirely sure if even ever uses this code pathway since the work using structured tags

aarnphm (approved):

alecsolder (approved):
Looks good to me, I have a PR ready soon as a follow up which will adjust this for tool calling as well

PR #28002, opened 2025-11-03
[Bugfix] Fix MoE Routing Simulation

tlrmchlsmth:
On current main, the routing simulation has no effect (besides overhead).

After calling `RoutingSimulator.simulate_routing` when `VLLM_MOE_ROUTING_SIMULATION_STRATEGY` is set, `select_experts` continues to call the top_k kernel as if nothing has happened.

This PR fixes that, and adds some logging warning the user that their output will not be correct.


LucasWilkinson (approved):
LGTM; good find!

minosfuture (approved):
Thanks for the fix! I need to add ut to guard this.

PR #28004, opened 2025-11-03
[XPU] Enable custom routing functions in IPEX for Llama4

frost-intel:
## Purpose
Llama4 model requires custom_routing_function which is currently disabled for xpu models. IPEX supports custom routing functions as of 2.8.10 ([commit](https://github.com/intel/intel-extension-for-pytorch/commit/525bf18eee5dd4b140acde4ecc18020dee342a21)) so I'm enabling them here.

## Test Plan
```bash
python examples/offline_inference/basic/generate.py  --model   meta-llama/Llama-4-Scout-17B-16E-Instruct --max-model-len 2048 -tp 8
```
lm-eval:
```
lm_eval --model vllm --model_args pretrained=meta-llama/Llama-4-Scout-17B-16E,add_bos_token=True,tensor_parallel_size=8  --tasks gsm8k  --num_fewshot 5 --batch_size auto --limit 250
```

## Test Result
Before fix:
```
File "vllm/vllm/model_executor/layers/fused_moe/layer.py", line 788, in forward_xpu
     assert custom_routing_function is None
```
After fix:
```
Prompt: 'Hello, my name is'
Generated text: ''
Prompt: 'The president of the United States is'
Generated text: ' the head of state and head of government of the United States, and the leader'
Prompt: 'The capital of France is'
Generated text: ' Paris. The capital of Spain is Madrid. The capital of Italy is Rome.'
Prompt: 'The future of AI is'
Generated text: " here, and it's not just about machines. It's about how humans and AI"
```
lm-eval:
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.860|±  |0.0220|
|     |       |strict-match    |     5|exact_match|↑  |0.856|±  |0.0222|
```
---
<details>

</details>


yma11:
> I will check what custom_routing_function is supported in ipex. but from the test result, seems first prompt doesn't have a reasonable output. cc @yma11 I think we didn't verfied llama4+ipex right? we prefer use vllm-xpu-kernels?

Interface level supports this enablement but we haven't verified it formally. I noticed this output before. @frost-intel which platform is targeted for this change? We don't want to enable a feature which not fully validated in upstream and this model can't be run on Multi-arc I think.

frost-intel:
> Interface level supports this enablement but we haven't verified it formally. I noticed this output before. @frost-intel which platform is targeted for this change? We don't want to enable a feature which not fully validated in upstream and this model can't be run on Multi-arc I think.

@yma11  The target is PVC, to run Lama4 on Aurora.

heheda12345 (approved):

PR #5836, opened 2024-06-25
[Frontend][Core] Override HF `config.json` via CLI

KrishnaM251:
FIX #1542
FIX #2547
FIX #5205

<h3>PR Title and Classification</h3>
[Frontend] [Core]

<h3>Notes</h3>

- While I believe the implementation is almost complete, the test I wrote only checks if the new `Optional[dict]` parameter `hf_kwargs` is succesfully set in `ModelConfig`.
- I would like to write a test that detects a change in the `LLMEngine` output when `kf_kwargs` is added as a `EngineArgs` parameter. 
- But I have two questions:
  - How can I write another test that runs `hf_kwargs` through an OpenAI compatible server? I wanted to repurpose a test in [test_openai_server.py](https://github.com/vllm-project/vllm/blob/dd248f76756adba4a1637b882e79ab639f957feb/tests/entrypoints/test_openai_server.py#L92), however the `EngineArgs` params are set before all tests are run.
  - What are some test values that I can set in `hf_kwargs` which will generate output detectable in a function like: `completion = client.completions.create(...)` ([motivating example](https://github.com/vllm-project/vllm/blob/dd248f76756adba4a1637b882e79ab639f957feb/tests/entrypoints/test_openai_server.py#L137)). If this is not the best approach for testing, then any recommendations?

PR #28006, opened 2025-11-03
Remove deprecated `--rope-scaling` and `--rope-theta`

hmellor:
These have been deprecated for a year (since https://github.com/vllm-project/vllm/pull/5836) so we should remove them.

This cleanup is a prerequisite to supporting Transformers v5 style RoPE configuration (per layer type).


DarkLight1337 (approved):

yewentao256 (approved):
LGTM, thanks for the work!

Issue #25538, opened 2025-09-24
[Bug]: performance regression caused by frequently preempting and resuming a request


ronny1996:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
Your output of `python collect_env.py` here
```

</details>

### 🐛 Describe the bug

https://github.com/vllm-project/vllm/blob/main/vllm/v1/core/sched/scheduler.py#L473

Allocate slots for `token_budget` (max_num_batched_tokens < prefill_len) tokens, causing the preempted request to be resume and preempted again in the next step due to no enough kvcache to store the long prefill. 

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

PR #28018, opened 2025-11-04
[Core][WIP] Add back waterline for scheduler

whx-sjtu:
## Purpose
This PR purposes to add back the `watermark` mechanism for v1 scheduler, in order to avoid frequent cache eviction. Although in v1 scheduler requests in running queue is scheduled first, there are still certain scenarios where a request may be frequently preempted and unable to complete its inference in a timely manner. For example, consider processing large amount of long requests and `max_num_seq * max_model_len > kv cache size`: 

- In step `n`, request `X` is preempted due to insufficient kv cache blocks, and added into waiting queue.
- In step `n+1`, request `X` in waiting queue is chunked and scheduled into running immediately with chunked prefill enable.
- After several rounds of scheduling, no running requests have finished while previously freed kv cache blocks of request `X` is occupied again. Then preempting again!

With watermark mechanism we can prevent scheduling waiting requests into running in scenarios where gpu memory is scarce. 

Existing issues: https://github.com/vllm-project/vllm/issues/25538

cc @WoosukKwon @njhill @wangxiyuan 

## Test Plan
If accepted, I will add ut for watermark.
## Test Result
All ci should pass.
---
<details>

</details>

### DIFF
--- a/vllm/config/scheduler.py
+++ b/vllm/config/scheduler.py
@@ -142,6 +142,9 @@ class SchedulerConfig:
     speculative decoding and pipeline parallelism.
     """
 
+    waterline: float = 0.01
+    """Use waterline to avoid frequent kv cache eviction."""
+
     def get_scheduler_cls(self) -> type["SchedulerInterface"]:
         if self.scheduler_cls is None:
             if self.async_scheduling:
--- a/vllm/v1/core/kv_cache_manager.py
+++ b/vllm/v1/core/kv_cache_manager.py
@@ -12,7 +12,7 @@ from vllm.v1.core.kv_cache_coordinator import get_kv_cache_coordinator
 from vllm.v1.core.kv_cache_utils import KVCacheBlock
 from vllm.v1.kv_cache_interface import KVCacheConfig
 from vllm.v1.metrics.stats import PrefixCacheStats
-from vllm.v1.request import Request
+from vllm.v1.request import Request, RequestStatus
 
 logger = init_logger(__name__)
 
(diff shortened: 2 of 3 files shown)


heheda12345:
CC @WoosukKwon @robertgshaw2-redhat @njhill

njhill:
Thanks @whx-sjtu this looks reasonable to me though I need to spend a bit longer to understand properly and I'm a bit backed up at the moment.

One note though ... I think you mean "high-water mark"... "watermark" means something quite different so I was initially confused :)

hmellor:
I also misinterpreted the `watermark` name. Typically a watermark is an annotation which is used to determine the providence of a piece of data. The name would also have been confusing in `v0.4.2`.

If we want to stick with the water analogy, a better choice of word would be `waterline` (the level normally reached by the water on the side of a ship) or `tidemark` (a mark left or reached by the sea on a shore at the highest or lowest point of a tide). Neither of these words require disambiguation with the other meanings.

whx-sjtu:
> I also misinterpreted the `watermark` name. Typically a watermark is an annotation which is used to determine the providence of a piece of data. The name would also have been confusing in `v0.4.2`.
> 
> If we want to stick with the water analogy, a better choice of word would be `waterline` (the level normally reached by the water on the side of a ship) or `tidemark` (a mark left or reached by the sea on a shore at the highest or lowest point of a tide). Neither of these words require disambiguation with the other meanings.

Thanks for your advice. `waterline` looks better to me than `watermark`.

PR #28024, opened 2025-11-04
[amd][gptoss] Perf gain because of block alignment

smitkadvani:
Summary:
Signed-off-by: Smit Kadvani smit.kadvani@gmail.com
## Summary:
Following patch is from Aliasger Zaidy(azaid) and Shucai Xiao(scxiao) from AMD, and overall efforts for the integration is guided by Xiaozhu Meng(mxz297) from Meta, it boosts the performance for fused_moe kernel.
We pad to 128 for MI300 to avoid masked loads. 
We pad to 256 for MI355 because we use scale preshuffling on 355 and padding to 256 is needed to enable correct preshuffle arrangement

10% Performance boost is achieved for gptoss120b on AMD mi300 machine.

Test Plan:
## Test Plan:
No eval regression is observed. 
### with patch 

| Effort Level |  Score | Characters | Chars Std | Score Std |
|--------------|-------|------------|-----------|-----------|
| Low  | 0.51 | 1577.26 | 1001.32 | 0.49 |
| Medium  | 0.79 | 1991.975 | 785.68 | 0.40 |
| High  | 0.916 | 2568 | 1029.9 | 0.28 |
### without patch 

| Effort Level |  Score | Characters | Chars Std | Score Std |
|--------------|-------|------------|-----------|-----------|
| Low  | 0.51 | 1570.26 | 1001.32 | 0.49 |
| Medium  | 0.79 | 1990.975 | 780.68 | 0.40 |
| High  | 0.916 | 2508 | 1020.9 | 0.28 |


vllm/model_executor/layers/quantization/utils/mxfp4_utils.py
```diff
@@ -99,6 +100,14 @@ def _can_support_mxfp4(
     )
 
 
+def get_padding_alignment():
+    return (
+        256
+        if triton.runtime.driver.active.get_current_target().arch in ("gfx950",)
```
heheda12345, line 106: The default value was 256. Will it be safer to only update MI300's alignment to 128? Or do you think 128 will be faster on other architectures?

HAIAI (approved):
@smitkadvani LGTM, thanks!

vllm/model_executor/layers/quantization/utils/mxfp4_utils.py
smitkadvani: 256 is needed to enable correct preshuffle arrangement and scale pre-shuffling only used in MI355, that's why i believe 128 will be faster on other architectures.

heheda12345 (approved):
LGTM!

PR #28029, opened 2025-11-04
[Perf][Deepseek] optimize gather_and_maybe_dequant_cache kernel's perf for extremely long sequence

ganyi1996ppo:
## Purpose
We found the `gather_and_maybe_dequant_cache` kernel can be the performance bottle neck on the extremely long sequence case, In our private test branch, this kernel takes 2.12ms in a single layer when operating on a 57k prefix caching case, and that count to almost 30% time of this layer's time consumption. 

In this PR, we rewrite this kernel in more hardware friendly way to accelerate this kernel include
- parallel on `num_tokens` for better occupancy.
- vectorize the memory access and unroll the loop for better ILP

 and brings 4-5 times of performance gain on extremely long sequence.

## Test Plan
- kernel benchmark
- gsm8k for model

## Test Result
```
# gsm8k before this PR
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9416|±  |0.0065|
|     |       |strict-match    |     5|exact_match|↑  |0.9363|±  |0.0067|

#gsm8k this PR
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9447|±  |0.0063|
|     |       |strict-match    |     5|exact_match|↑  |0.9424|±  |0.0064|
```

We test the kernel's performance on MI308 and H20 with `(3, 8) batch_size` and `(64k, 512) context length`, and get below result. And it shows the performance boost is more obvious on Hopper for both shape and the Mi308 also got quite performance boost on long context case, and only minor gap against the old impl on short context, which is negligible during model run compared with other kernels.
| Item | (3, 65536) | (8, 65536) | (3,512) | (8, 512)
|------|-----------|------------------|-----------|-----------|
| **MI308** fp8 (old impl)| 1.92ms    | 2.49ms | **0.029ms** | **0.037ms** |
| **MI308** fp8 **(new impl)** | **0.31ms**    | **0.77ms** | 0.041ms | 0.043ms |
|**MI308** bf16 (old impl) | 2.26ms | 2.81ms | **0.029ms** | **0.039ms** |
|**MI308** bf16 **(new impl)** | **0.41ms** | **1.03ms** | 0.041ms | 0.044ms|
|**H20** fp8 (old impl) | 1.86ms | 2.07ms | 0.13ms | 0.028ms|
|**H20** fp8 **(new impl)** | **0.374ms** | **0.82ms** | **0.025ms** | **0.027ms** |
|**H20** bf16 (old impl) | 1.95ms | 2.6ms | 0.037ms | 0.093ms|
|**H20** bf16 **(new impl)** | **0.37ms** | **0.82ms** | **0.027ms** | **0.025ms** |
---
<details>

</details>


csrc/cache_kernels.cu
```diff
@@ -901,89 +901,152 @@ void convert_fp8(torch::Tensor& dst_cache, torch::Tensor& src_cache,
...
+    src_ = src_ + entry_size - tail_cnt;
+    for (int idx = threadIdx.x; idx < tail_cnt; idx += blockDim.x) {
+      if constexpr (kv_dt == Fp8KVCacheDataType::kAuto) {
+        dst_[idx] = static_cast<scalar_t>(src_[idx]);
+      } else {
+        dst_[idx] =
+            fp8::scaled_convert<scalar_t, cache_t, kv_dt>(src_[idx], *scale);
+      }
     }
   }
```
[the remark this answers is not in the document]
ganyi1996ppo: This kernel is mainly used by deepseek which have `kv_lora_rank=576`, and this PR's goal is optimize the performance on deepseek, make the `entry_size=576` can gives us better loop unroll and achieve better ILP for memory operation.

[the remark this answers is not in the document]
mgoin: There are many models that use the same model definition as deepseek or MLA, such as deepseek-v2-lite, kimi-k2, or kimi-linear, which I'm not sure if they all use 576. We should at least assert this size is the case or allow for flexibility with lower performance outside of the case

mgoin:
Looks reasonable to me, although I would like to validate performance on Hopper as well since that is the most common deepseek deployment arch. I also don't see the eval reported. cc @LucasWilkinson @MatthewBonanni @yewentao256 for review as well

csrc/cache_kernels.cu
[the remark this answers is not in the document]
ganyi1996ppo: Sure, I'll add assertions for unsupported size, or I can just remove it from template variable parameter, which might drop a little performance but more general to all different models.

csrc/cache_kernels.cu
```diff
@@ -901,87 +901,76 @@ void convert_fp8(torch::Tensor& dst_cache, torch::Tensor& src_cache,
...
-    const int64_t cache_entry_stride, const int64_t dst_entry_stride,
-    const float* __restrict__ scale,
+    const cache_t* __restrict__ src_cache,     // [NUM_BLOCKS, BLOCK_SIZE,
+                                               // ENTRIES...]
+    scalar_t* __restrict__ dst,                // [TOT_TOKENS, ENTRIES...]
+    const int32_t* __restrict__ block_table,   // [BATCH, BLOCK_INDICES]
+    const int32_t* __restrict__ cu_seq_lens,   // [BATCH+1]
+    const int32_t* __restrict__ token_to_seq,  // [MAX_TOKEN_ACROSS_CHUNK]
+    const int32_t total_tokens, const int32_t block_size,
+    const int32_t entry_size, const int64_t block_table_stride,
```
MatthewBonanni: Why is entry size both a template parameter and an argument?

csrc/cache_kernels.cu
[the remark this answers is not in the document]
MatthewBonanni: 576 is fairly universal as far as I know. I'm not aware of any exceptions

tests/kernels/attention/test_cache.py
```diff
@@ -887,9 +887,9 @@ def test_swap_blocks_mla(
 @pytest.mark.parametrize("kv_lora_rank", [512])
 @pytest.mark.parametrize("qk_rope_head_dim", [64])
 @pytest.mark.parametrize("block_size", [16])
-@pytest.mark.parametrize("num_blocks", [1024])
-@pytest.mark.parametrize("max_seq_len", [512])
-@pytest.mark.parametrize("batch_size", [8])
+@pytest.mark.parametrize("num_blocks", [32768])
+@pytest.mark.parametrize("max_seq_len", [65536])
+@pytest.mark.parametrize("batch_size", [3])
```
MatthewBonanni: Out of curiosity, why reduce the batch size in the test? Were there issues with 8?

ganyi1996ppo: No, actually, I'll change it back or add more test case. This change is just for simple perf verification.

csrc/cache_kernels.cu
ganyi1996ppo: Thanks for pointing at out, I'll remove it from arguments.

csrc/cache_kernels.cu
```diff
@@ -901,87 +901,76 @@ void convert_fp8(torch::Tensor& dst_cache, torch::Tensor& src_cache,
...
       } else {
-        _dst[i] =
-            fp8::scaled_convert<scalar_t, cache_t, kv_dt>(_src[i], *scale);
+        ltype loaded_val = reinterpret_cast<ltype*>(src_)[idx];
+        stype store_val;
+// we assume the kv cache dtype should always got lesser bit
+#pragma unroll
+        for (int j = 0; j < vec_size; ++j) {
+          store_val.val[j] = fp8::scaled_convert<scalar_t, cache_t, kv_dt>(
+              loaded_val.val[i * vec_size + j], *scale);
```
peizhang56: Where is this `i` defined?

hongxiayang:
```suggestion
              loaded_val.val[idx * vec_size + j], *scale);
```

peizhang56: https://github.com/vllm-project/vllm/pull/28029/commits/714ada5da6f766f0bc3e524a1e66a7748b05f782
I located the original changes, it looks originally, it was using ilp_loop_cnt for the i loop.
Due to a refactor, the instruction level parallel loop cnt is gone. Should `i * vec_size ` also be removed?

ganyi1996ppo: Sorry, there are few changes I forget to push up, please try again, this version is correct now.

ganyi1996ppo: Fix updated, please take a look

ganyi1996ppo:
hi @mgoin @MatthewBonanni @heheda12345 @yewentao256 @LucasWilkinson @HAIAI @gshtras , I updated the performance comparison table on hopper as well, please take a look.

tests/kernels/attention/test_cache.py
ganyi1996ppo: I added the old shape back to the test, please take a look

MatthewBonanni: I don't see the change, can you double check?

csrc/cache_kernels.cu
MatthewBonanni: Gentle reminder on this, thanks!

ganyi1996ppo: Sorry, there are some local commits I forget to push......., please take a look again

tests/kernels/attention/test_cache.py
ganyi1996ppo: Updated, Sorry for the mistake before....

ganyi1996ppo:
@tjtanaa Please take a look on this PR too

vllm/v1/attention/backends/mla/common.py
```diff
@@ -828,6 +830,19 @@ def build(
...
+                token_to_seq_tensor_cpu = torch.zeros(
+                    [num_chunks, max_token_num_over_chunk], dtype=torch.int32
+                )
+                range_idx = torch.arange(num_prefills, dtype=torch.int32)
+                for i in range(num_chunks):
+                    chunk_token_to_seq_tensor = torch.repeat_interleave(
+                        range_idx, chunk_seq_lens[i]
+                    )
+                    chunk_len = chunk_token_to_seq_tensor.shape[0]
+                    token_to_seq_tensor_cpu[i, :chunk_len] = chunk_token_to_seq_tensor
```
mgoin, line 856: Is this extra tensor a perf concern cc @LucasWilkinson @MatthewBonanni ? I don't have a feeling for `num_chunks` and `max_token_num_over_chunk`

ganyi1996ppo: This extra tensor preparation almost have no effect to the performance during our test. And all the calculation is execute on CPU, so the overhead should be acceptable even on short context.

tests/kernels/attention/test_cache.py
```diff
@@ -887,9 +887,9 @@ def test_swap_blocks_mla(
 @pytest.mark.parametrize("kv_lora_rank", [512])
 @pytest.mark.parametrize("qk_rope_head_dim", [64])
 @pytest.mark.parametrize("block_size", [16])
-@pytest.mark.parametrize("num_blocks", [1024])
-@pytest.mark.parametrize("max_seq_len", [512])
-@pytest.mark.parametrize("batch_size", [8])
+@pytest.mark.parametrize("num_blocks", [32768])
+@pytest.mark.parametrize("max_seq_len", [65536, 512])
```
tjtanaa: I have a concern, all CI runs are usually using L4 GPUs, there might not be enough VRAM to run this test. We should also consider keeping unit tests runnable on small VRAM GPUs to validate correctness of kernels rather than skipping all cases. 

Do you know if this test can run with 24GB of VRAM?

Maybe it is better to separate out with a test that is specifically for "long context"

ganyi1996ppo: Thanks for the suggestion, I'll remove the new added case.

csrc/cache.h
```diff
@@ -41,11 +41,12 @@ void convert_fp8(torch::Tensor& dst_cache, torch::Tensor& src_cache,
...
-    torch::Tensor const& src_cache,    // [NUM_BLOCKS, BLOCK_SIZE, ENTRIES...]
-    torch::Tensor const& dst,          // [TOT_TOKENS, ENTRIES...]
-    torch::Tensor const& block_table,  // [BATCH, BLOCK_INDICES]
-    torch::Tensor const& cu_seq_lens,  // [BATCH+1]
-    int64_t batch_size, const std::string& kv_cache_dtype,
+    torch::Tensor const& src_cache,     // [NUM_BLOCKS, BLOCK_SIZE, ENTRIES...]
+    torch::Tensor const& dst,           // [TOT_TOKENS, ENTRIES...]
+    torch::Tensor const& block_table,   // [BATCH, BLOCK_INDICES]
+    torch::Tensor const& cu_seq_lens,   // [BATCH+1]
+    torch::Tensor const& token_to_seq,  // [TOT_TOKENS]
```
MatthewBonanni:
```suggestion
    torch::Tensor const& token_to_seq,  // [MAX_TOKEN_ACROSS_CHUNK]
```
is this correct?

csrc/cache_kernels.cu
```diff
@@ -901,87 +901,75 @@ void convert_fp8(torch::Tensor& dst_cache, torch::Tensor& src_cache,
...
-    const int32_t* __restrict__ block_table,  // [BATCH, BLOCK_INDICES]
-    const int32_t* __restrict__ cu_seq_lens,  // [BATCH+1]
-    const int32_t block_size, const int32_t entry_size,
+    const cache_t* __restrict__ src_cache,     // [NUM_BLOCKS, BLOCK_SIZE,
+                                               // ENTRIES...]
+    scalar_t* __restrict__ dst,                // [TOT_TOKENS, ENTRIES...]
+    const int32_t* __restrict__ block_table,   // [BATCH, BLOCK_INDICES]
+    const int32_t* __restrict__ cu_seq_lens,   // [BATCH+1]
+    const int32_t* __restrict__ token_to_seq,  // [MAX_TOKEN_ACROSS_CHUNK]
+    const int32_t total_tokens, const int32_t block_size,
```
MatthewBonanni:
```suggestion
    const int32_t num_tokens, const int32_t block_size,
```
nit: could we call this `num_tokens` throughout for consistency?

csrc/cache_kernels.cu
```diff
@@ -901,87 +901,75 @@ void convert_fp8(torch::Tensor& dst_cache, torch::Tensor& src_cache,
...
-    auto block_start_ptr = src_cache + block_id * cache_block_stride;
-    auto block_dst_ptr = dst + full_blocks_end * block_size * dst_entry_stride;
-    for (int eid = 0; eid < partial_block_size; ++eid) {
-      copy_entry(block_start_ptr + eid * cache_entry_stride,
-                 block_dst_ptr + eid * dst_entry_stride);
+    // process tail
+    int32_t tail_cnt = ENTRY_SIZE % vec_size;
+    dst_ = dst_ + ENTRY_SIZE - tail_cnt;
+    src_ = src_ + ENTRY_SIZE - tail_cnt;
+    for (int idx = threadIdx.x; idx < tail_cnt; idx += blockDim.x) {
```
MatthewBonanni:
```suggestion
#pragma unroll
    for (int idx = threadIdx.x; idx < tail_cnt; idx += CTA_SIZE) {
```
This can be unrolled at compile time like above

csrc/cache_kernels.cu
```diff
@@ -901,87 +901,75 @@ void convert_fp8(torch::Tensor& dst_cache, torch::Tensor& src_cache,
...
-
-  if (partial_block_size) {
-    auto block_id = batch_block_table[full_blocks_end];
-    auto block_start_ptr = src_cache + block_id * cache_block_stride;
-    auto block_dst_ptr = dst + full_blocks_end * block_size * dst_entry_stride;
-    for (int eid = 0; eid < partial_block_size; ++eid) {
-      copy_entry(block_start_ptr + eid * cache_entry_stride,
-                 block_dst_ptr + eid * dst_entry_stride);
+    // process tail
+    int32_t tail_cnt = ENTRY_SIZE % vec_size;
```
MatthewBonanni:
```suggestion
    constexpr int32_t tail_cnt = ENTRY_SIZE % vec_size;
```

csrc/cache_kernels.cu
```diff
@@ -901,87 +901,75 @@ void convert_fp8(torch::Tensor& dst_cache, torch::Tensor& src_cache,
...
+template <typename scalar_t, typename cache_t, Fp8KVCacheDataType kv_dt,
+          int ENTRY_SIZE, int CTA_SIZE>
 __global__ void gather_and_maybe_dequant_cache(
-    const cache_t* __restrict__ src_cache,    // [NUM_BLOCKS, BLOCK_SIZE,
-                                              // ENTRIES...]
-    scalar_t* __restrict__ dst,               // [TOT_TOKENS, ENTRIES...]
-    const int32_t* __restrict__ block_table,  // [BATCH, BLOCK_INDICES]
-    const int32_t* __restrict__ cu_seq_lens,  // [BATCH+1]
-    const int32_t block_size, const int32_t entry_size,
+    const cache_t* __restrict__ src_cache,     // [NUM_BLOCKS, BLOCK_SIZE,
```
MatthewBonanni, line 911: Maybe add an `assert(blockDim.x == CTA_SIZE);`? Will get excluded when compiled in release mode but will at least signal intent to people reading this code in the future.

MatthewBonanni (commented):
Thanks for the changes! A few more minor comments

vllm/v1/attention/backends/mla/common.py
MatthewBonanni: I'm not super concerned about it if the perf numbers look alright across context length. The bigger issue would potentially be the need for another H2D transfer later on but that's grouped with the other transfers and non-blocking anyway so seems to me it should be fine

csrc/cache_kernels.cu
ganyi1996ppo: Appreciate for the comments, just know that runtime variable like `threadIdx.x` can be unrolled by compiler as well.  That really helpful to me.

csrc/cache.h
ganyi1996ppo: Yes, that's correct!

csrc/cache_kernels.cu
ganyi1996ppo: Sure, thanks for the suggestion. Assertion added

csrc/cache_kernels.cu
ganyi1996ppo: Done

csrc/cache_kernels.cu
ganyi1996ppo: Sure.

ganyi1996ppo:
@mgoin @MatthewBonanni @tjtanaa all test passed, can we merge this PR?

mgoin (approved):

mgoin:
thanks for the nice work!

PR #28044, opened 2025-11-04
[cudagraph] fix cudagraph warning in deepseekv32

ZJY0516:
## Purpose
This PR fixes Dynamo partitioning for DeepSeek-V3.2 when an empty allocation is isolated into its own subgraph.

Previously, FX splitting could produce:
```
submod_1: torch.ops.vllm.sparse_attn_indexer(...)
submod_2: torch.empty(...)
submod_3: torch.ops.vllm.unified_mla_attention_with_output(...)
```
This resulted in a CUDA graph that contained only the allocation operation (empty), triggering the warning:
```
UserWarning: The CUDA Graph is empty. This usually means that the graph was attempted to be captured on wrong device or stream. (Triggered internally at /pytorch/aten/src/ATen/cuda/CUDAGraph.cpp:138.)
```

**What changed**:

In graph partitioning, if a subgraph contains only one empty allocation node, merge it into the previous subgraph.

## Test Plan
```
lm_eval --model local-completions --model_args "model=deepseek-ai/DeepSeek-V3.2,base_url=http://0.0.0.0:8000/v1/completions,tokenized_requests=False,tokenizer_backend=None,num_concurrent=256,timeout=5000,max_length=4096" --tasks gsm8k --num_fewshot 5
```

## Test Result
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9553|±  |0.0057|
|     |       |strict-match    |     5|exact_match|↑  |0.9545|±  |0.0057|
```
---
<details>

</details>

### DIFF
--- a/tests/compile/test_graph_partition.py
+++ b/tests/compile/test_graph_partition.py
@@ -233,26 +233,6 @@ def test_builtin_empty_only_partition_is_merged():
     # [empty_like], [attention], [empty_like], [attention].
     assert len(split_items) == 3, "Builtin empty-only partition should be merged"
 
-    empty_only_indices: list[int] = []
-    for idx, split_item in enumerate(split_items):
-        call_nodes = [
-            n for n in split_item.graph.graph.nodes if n.op == "call_function"
-        ]
-        if len(call_nodes) != 1:
-            continue
-        target = call_nodes[0].target
-        if getattr(target, "__module__", None) == "torch" and getattr(
-            target, "__name__", None
-        ) in ("empty", "empty_like", "empty_strided", "new_empty"):
-            empty_only_indices.append(idx)
-
-    # The initial output allocation before the first split op is expected.
-    # Only the middle empty-only partition should be merged away.
-    assert empty_only_indices == [0], (
-        "Only the leading allocation partition should remain empty-only, "
-        f"got indices={empty_only_indices}"
-    )
-
     x = torch.randn(2, 3, device="cuda")
     output_original = gm(x)
     output_split = split_gm(x)
--- a/tests/compile/test_graph_partition.py
+++ b/tests/compile/test_graph_partition.py
@@ -184,3 +184,76 @@ def test_consecutive_ops_in_split():
     assert [node.op for node in splitting_gm.graph.nodes] == ["placeholder"] + 2 * [
         "call_function"
     ] + ["output"]
+
+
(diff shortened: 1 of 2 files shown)


ProExpertProg (changes requested):
I think instead we should update the dynamo splitting logic to reduce the size of the partitions and move output to the previous partition.

ZJY0516:
> I think instead we should update the dynamo splitting logic to reduce the size of the partitions and move output to the previous partition. 

As I recall, the graph splitting logic is part of `torch.compile`'s internal machinery, correct?

ZJY0516:
> I agree that would be good if dynamo move `torch.empty` to the previous partition and we should file a issue(or communicate some another way).

I don't think it's necessary for now because this problem only occurs in deepseekv32

ProExpertProg:
@vadiklyutiy dynamo splitting logic actually lives in vLLM so we can modify it without a torch upgrade. 

@LucasWilkinson basically right now dynamo splitting just goes in order and splits on splitting ops and then creates new partitions after. That means output gets its own partition because it's nested between two splitting op nodes. Instead we should move the output to the previous node when it has no dependents on the previous splitting op

It's a bit harder but much cleaner, and this is a warning we're talking about, not a correctness issue.

youkaichao:
> I think instead we should update the dynamo splitting logic to reduce the size of the partitions and move output to the previous partition.

@ProExpertProg  I think it would be unsafe for Dynamo to move the `empty` function automatically.

For this specific model, since we know the structure pretty well, we can manually change the boundary as @ZJY0516 does to avoid the warning `The CUDA Graph is empty.` .

ProExpertProg (approved):
Cc @zou3519 @BoyuanFeng

zou3519 (approved):

BoyuanFeng (approved):

ZJY0516:
> I think this broke ds-v3:

@ProExpertProg  What's yourr command? I can not reproduce this using f85b4eda3

Here is my commands
```
vllm serve deepseek-ai/DeepSeek-V3 -dp 8 --max-cudagraph-capture-size 8
```
```
vllm serve deepseek-ai/DeepSeek-V3 -tp 8
```

ZJY0516:
@jikunshang I already have a fix here https://github.com/vllm-project/vllm/pull/36595

Issue #188, opened 2023-06-21
CUDA error: out of memory


SunixLiu:
I successfully installed vLLM in WSL2, when I was trying to run the sample code, I got error info like this:

```
from vllm import LLM, SamplingParams

prompts = [
    "Hello, my name is",
    "The president of the United States is",
    "The capital of France is",
    "The future of AI is",
]
sampling_params = SamplingParams(temperature=0.8, top_p=0.95)
llm = LLM(model="/mnt/d/github/text-generation-webui/models/facebook_opt-125m")

outputs = llm.generate(prompts, sampling_params)

# Print the outputs.
for output in outputs:
    prompt = output.prompt
    generated_text = output.outputs[0].text
    print(f"Prompt: {prompt!r}, Generated text: {generated_text!r}")
```

INFO 06-21 21:40:02 llm_engine.py:59] Initializing an LLM engine with config: model='/mnt/d/github/text-generation-webui/models/facebook_opt-125m', dtype=torch.float16, use_dummy_weights=False, download_dir=None, use_np_weights=False, tensor_parallel_size=1, seed=0)
INFO 06-21 21:40:12 llm_engine.py:128] # GPU blocks: 37375, # CPU blocks: 7281
Traceback (most recent call last):
  File "/mnt/d/01Projects/vllm/prac_1.py", line 11, in <module>
    llm = LLM(model="/mnt/d/github/text-generation-webui/models/facebook_opt-125m")
  File "/mnt/d/github/vllm/vllm/entrypoints/llm.py", line 55, in __init__
    self.llm_engine = LLMEngine.from_engine_args(engine_args)
  File "/mnt/d/github/vllm/vllm/engine/llm_engine.py", line 145, in from_engine_args
    engine = cls(*engine_configs, distributed_init_method, devices,
  File "/mnt/d/github/vllm/vllm/engine/llm_engine.py", line 102, in __init__
    self._init_cache()
  File "/mnt/d/github/vllm/vllm/engine/llm_engine.py", line 134, in _init_cache
    self._run_workers("init_cache_engine", cache_config=self.cache_config)
  File "/mnt/d/github/vllm/vllm/engine/llm_engine.py", line 307, in _run_workers
    output = executor(*args, **kwargs)
  File "/mnt/d/github/vllm/vllm/worker/worker.py", line 126, in init_cache_engine
    self.cache_engine = CacheEngine(
[...]

PR #28051, opened 2025-11-04
[Bugfix] fix confusing OOM errors during v1 init

### LINKED ISSUE #14083: [Feature]: Improve Logging for Error Messages

Improve logging on VLLM V1 for common errors in initialization.

For example:
- not enough memory to fit the model
- not enough kv cache space to fit the model
- ...

Currently we have decently logging of the exceptions that arise. It would be better however if we could explicitly catch these issues and return clearer error messages

### Alternatives

_No response_

### Additional context

_No response_

shivampr:
## Purpose

Fixes #14083

Improves error messages for two common V1 initialization failures:
1. GPU OOM when loading model weights
2. Insufficient KV cache memory

Instead of cryptic CUDA errors, users now get clear messages with actionable suggestions like:
- adjusting `--gpu-memory-utilization`
- increasing `--tensor-parallel-size`
- using `--quantization`
- reducing `--max-model-len`
- lowering `--max-num-seqs`

Changes:
- `vllm/v1/worker/gpu_model_runner.py`: wrap model loading in try/except to catch OOM
- `vllm/v1/core/kv_cache_utils.py`: improve error messages in `check_enough_kv_cache_memory()`
- `tests/v1/engine/test_init_error_messaging.py`: add CPU-only tests using mocks

## Test Plan

```bash
# Run the new tests
pytest tests/v1/engine/test_init_error_messaging.py -v
```

Tests use mocks to simulate OOM errors without requiring GPU, so they run on CPU.

## Test Result

All tests pass. Error messages now include:
- Clear description of the problem
- Structured logging with `logger.error("init_failed msg=%s", ...)`

Example outputs:
```
RuntimeError: Failed to load model - not enough GPU memory. Try reducing --gpu-memory-utilization, increasing --tensor-parallel-size, or using --quantization.
```

```
RuntimeError: Insufficient KV cache memory: need 10.5 GiB but only 2.3 GiB available (max_model_len=4096). Estimated max length with available memory: 896. Try reducing --max-model-len or --max-num-seqs, or increase --gpu-memory-utilization.
```

---
<details>

</details>


vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2960,40 +2960,53 @@ def load_model(self, eep_scale_up: bool = False) -> None:
...
+                    else:
+                        aux_layers = self.model.get_eagle3_aux_hidden_state_layers()
+
+                    self.model.set_aux_hidden_state_layers(aux_layers)
+                time_after_load = time.perf_counter()
+            self.model_memory_usage = m.consumed_memory
+        except (torch.cuda.OutOfMemoryError, RuntimeError) as e:
+            err_str = str(e).lower()
+            if isinstance(e, torch.cuda.OutOfMemoryError) or \
+               "out of memory" in err_str or "cuda error" in err_str:
```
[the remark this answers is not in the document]
shivampr: addressed the suggestion.

tests/v1/engine/test_init_error_messaging.py
```diff
@@ -0,0 +1,124 @@
...
+def test_model_load_runtime_oom_error_message(monkeypatch):
+    from unittest.mock import MagicMock
+    from vllm.v1.worker.gpu_model_runner import GPUModelRunner
+
+    config = MagicMock()
+    config.model_config.model = "test-model"
+    config.model_config.seed = 42
+    
+    runner = GPUModelRunner(config, torch.device("cpu"))
+
```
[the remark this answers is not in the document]
shivampr: addressed the suggestion.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2960,40 +2960,53 @@ def load_model(self, eep_scale_up: bool = False) -> None:
...
+                msg = (
+                    "Failed to load model - not enough GPU memory. "
+                    "Try reducing --gpu-memory-utilization, increasing "
```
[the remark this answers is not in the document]
shivampr: This has already been addressed. The `cuda error` check was removed to avoid false positives. The current code only checks for `isinstance(e, torch.cuda.OutOfMemoryError)` or `out of memory` in err_str, which is precise and won't misclassify other CUDA errors.

heheda12345:
Yeah hope I can review it tomorrow.

vllm/v1/core/kv_cache_utils.py
```diff
@@ -675,44 +675,42 @@ def check_enough_kv_cache_memory(
...
         return
 
     if available_memory <= 0:
-        raise ValueError(
-            "No available memory for the cache blocks. "
-            "Try increasing `gpu_memory_utilization` when "
-            "initializing the engine."
+        msg = (
+            "No memory available for KV cache blocks. "
+            "Try reducing --max-model-len or --max-num-seqs, "
```
heheda12345: are max_model_len and max_num_seq still helpful when v1 always uses chunked prefill?

It's a good idea to give the users more suggestions. what about giving a quick suggestion for gpu_memory_utilization and add a link to this doc https://docs.vllm.ai/en/latest/configuration/conserving_memory/?

vllm/v1/core/kv_cache_utils.py
```diff
@@ -675,44 +675,42 @@ def check_enough_kv_cache_memory(
...
+
+        msg = (
+            f"Insufficient KV cache memory: need {needed_memory / GiB_bytes:.2f} GiB "
+            f"but only {available_memory / GiB_bytes:.2f} GiB available "
+            f"(max_model_len={max_model_len}).{hint} "
+            f"Try reducing --max-model-len or --max-num-seqs, "
+            f"or increase --gpu-memory-utilization."
         )
+        logger.error("init_failed msg=%s", msg)
+        raise RuntimeError(msg)
```
heheda12345: what's the problem of the original error message?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2948,71 +2948,53 @@ def load_model(self, eep_scale_up: bool = False) -> None:
...
+                            "Using auxiliary layers from speculative config: %s",
+                            aux_layers,
+                        )
+                    else:
+                        aux_layers = self.model.get_eagle3_aux_hidden_state_layers()
+
+                    self.model.set_aux_hidden_state_layers(aux_layers)
+                time_after_load = time.perf_counter()
+            self.model_memory_usage = m.consumed_memory
+        except (torch.cuda.OutOfMemoryError, RuntimeError) as e:
```
heheda12345: why do you need to catch RuntimeError here?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2948,71 +2948,53 @@ def load_model(self, eep_scale_up: bool = False) -> None:
...
+            err_str = str(e).lower()
+            if isinstance(e, torch.cuda.OutOfMemoryError) or "out of memory" in err_str:
+                msg = (
+                    "Failed to load model - not enough GPU memory. "
+                    "Try reducing --gpu-memory-utilization, increasing "
+                    "--tensor-parallel-size, or using --quantization."
+                )
+                logger.error("init_failed msg=%s", msg)
+                raise RuntimeError(msg) from e
```
heheda12345: It's good to tell people that the OOM is happened during model loading. But the original error message from PyTorch is also helpful (for example, I need it to check whether the OOM is due to someone occupies my GPU).  Can you keep both?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2948,71 +2948,53 @@ def load_model(self, eep_scale_up: bool = False) -> None:
...
+
+                    self.model.set_aux_hidden_state_layers(aux_layers)
+                time_after_load = time.perf_counter()
+            self.model_memory_usage = m.consumed_memory
+        except (torch.cuda.OutOfMemoryError, RuntimeError) as e:
+            err_str = str(e).lower()
+            if isinstance(e, torch.cuda.OutOfMemoryError) or "out of memory" in err_str:
+                msg = (
+                    "Failed to load model - not enough GPU memory. "
+                    "Try reducing --gpu-memory-utilization, increasing "
```
heheda12345: why reducing --gpu-memory-utilization?

vllm/v1/core/kv_cache_utils.py
shivampr: I changed for following 2 reasons : 

1. **Better exception type**: Used `ValueError` instead of `RuntimeError` (IMO, this is a runtime memory constraint, not invalid input).
2. **Have more structured logging and actionable suggestions**: No `logger.error("init_failed msg=%s", ..)` for tracking, and didn't mention `--max-num-seqs` as an option.

vllm/v1/worker/gpu_model_runner.py
shivampr: I added `RuntimeError` because some model loaders wrap CUDA OOM in a generic `RuntimeError` with "out of memory" in the message, so catching both ensures we handle all OOM cases. [Ref 1](https://github.com/vllm-project/vllm/issues/188?utm_source=chatgpt.com) , [Ref 2](https://discuss.huggingface.co/t/cuda-out-of-memory-error/17959?utm_source=chatgpt.com)

vllm/v1/worker/gpu_model_runner.py
shivampr: My assumption is :   `gpu_memory_utilization` sets how much GPU memory is reserved for KV cache during init. If it's too high (e.g., 0.95), there isn't enough left for model weights, causing OOM during loading. Lowering it (e.g., to 0.85) frees memory for weights and often resolves the issue.

vllm/v1/core/kv_cache_utils.py
shivampr: Chunked prefill helps, but IMO `max_model_len` and `max_num_seqs` still matter—if they’re set too high for the available memory, users could hit this error. 

I’ve added `--gpu-memory-utilization` to the suggestions and included a link to the memory doc: https://docs.vllm.ai/en/latest/configuration/conserving_memory/.

vllm/v1/worker/gpu_model_runner.py
shivampr: Addressed!

vllm/v1/worker/gpu_model_runner.py
heheda12345: These refs are quite old. Are they still happening now?

vllm/v1/core/kv_cache_utils.py
heheda12345: Why are they matter? I think most memory doesn't related to these two arguments.

vllm/v1/core/kv_cache_utils.py
heheda12345: Let's avoid changing the structure of the error message without strong evidence that we should do as users are familiar with the current messages now. And why does max-num-seqs matter here?

shivampr:
Ack, will address the review comments!

vllm/v1/worker/gpu_model_runner.py
shivampr: You're right—those references are outdated. I've removed `RuntimeError` from the exception handling. The code now only catches `torch.cuda.OutOfMemoryError`.

vllm/v1/core/kv_cache_utils.py
shivampr: I've reverted the error messages to the original format. The code now:

1. Uses `ValueError` (as before)
2. Keeps the original message structure and wording
3. Removed `--max-num-seqs` from suggestions 
4. Added `gpu_memory_utilization` suggestion

vllm/v1/core/kv_cache_utils.py
shivampr: I've removed `max_model_len` and `max_num_seqs` from the suggestions. The error messages now only mention `gpu_memory_utilization` and include the memory doc link.

tests/v1/engine/test_init_error_messaging.py
```diff
@@ -0,0 +1,114 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+import pytest
```
heheda12345, line 4: I think we don't need to test error message is not changed. test_kv_cache_oom_no_memory and test_kv_cache_oom_insufficient_memory make sense to me but can you just check it raises the correct error?

vllm/v1/core/kv_cache_utils.py
```diff
@@ -710,8 +712,9 @@ def check_enough_kv_cache_memory(
             f"cache is needed, which is larger than the available KV cache "
             f"memory ({available_memory / GiB_bytes:.2f} GiB). "
             f"{estimated_msg} "
-            f"Try increasing `gpu_memory_utilization` or decreasing "
-            f"`max_model_len` when initializing the engine."
+            f"Try increasing `gpu_memory_utilization` when initializing the engine. "
+            f"See https://docs.vllm.ai/en/latest/configuration/conserving_memory/ "
```
heheda12345, line 718: decrease max_model_len is helpful here and is used by many users.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -3270,74 +3270,87 @@ def load_model(self, eep_scale_up: bool = False) -> None:
...
-            time_after_load = time.perf_counter()
-        self.model_memory_usage = m.consumed_memory
+                    self.model.set_aux_hidden_state_layers(aux_layers)
+                time_after_load = time.perf_counter()
+            self.model_memory_usage = m.consumed_memory
+        except torch.cuda.OutOfMemoryError as e:
+            msg = (
+                "Failed to load model - not enough GPU memory. "
+                "Try lowering --gpu-memory-utilization to free memory for weights, "
+                "increasing --tensor-parallel-size, or using --quantization."
```
heheda12345: can you also link the doc here?

tests/v1/engine/test_init_error_messaging.py
shivampr: Thanks! I agree we only need to ensure that the right exception type is thrown. I’ve simplified both `test_kv_cache_oom_no_memory` and `test_kv_cache_oom_insufficient_memory` to just assert `pytest.raises(ValueError)`.

vllm/v1/core/kv_cache_utils.py
shivampr: I’ve re-added `--max-model-len` to the guidance.

vllm/v1/worker/gpu_model_runner.py
shivampr: Added the Doc link.

tests/v1/engine/test_init_error_messaging.py
```diff
@@ -0,0 +1,103 @@
...
+                "Failed to load model - not enough GPU memory. "
+                "Try reducing --gpu-memory-utilization, increasing "
+                "--tensor-parallel-size, or using --quantization."
+            )
+            # Verify the error message contains expected guidance
+            assert "not enough gpu memory" in msg.lower()
+            assert "--gpu-memory-utilization" in msg
+            assert "--tensor-parallel-size" in msg
+            assert "--quantization" in msg
+            return
```
heheda12345: why do you need to test these

tests/v1/engine/test_init_error_messaging.py
```diff
@@ -0,0 +1,103 @@
...
+            "Failed to load model - not enough GPU memory. "
+            "Try reducing --gpu-memory-utilization, increasing "
+            "--tensor-parallel-size, or using --quantization."
+        )
+        exc = RuntimeError(msg)
+
+        # Verify the error message
+        assert "gpu memory" in str(exc).lower()
+        assert "--gpu-memory-utilization" in str(exc)
+        assert "--tensor-parallel-size" in str(exc)
```
heheda12345: why do you need to test this

vllm/v1/core/kv_cache_utils.py
```diff
@@ -685,8 +685,10 @@ def check_enough_kv_cache_memory(
     if available_memory <= 0:
         raise ValueError(
             "No available memory for the cache blocks. "
-            "Try increasing `gpu_memory_utilization` when "
-            "initializing the engine."
+            "Try increasing `gpu_memory_utilization` or decreasing `max_model_len` "
```
heheda12345: why decreasing max model length helps here?

shivampr: Removed the `max_model_len` suggestion.

tests/v1/engine/test_init_error_messaging.py
shivampr: Removed the message content assertions, the test file only has `test_kv_cache_oom_no_memory` and `test_kv_cache_oom_insufficient_memory`.

tests/v1/engine/test_init_error_messaging.py
shivampr: Removed the message content assertions, the test file only has `test_kv_cache_oom_no_memory` and `test_kv_cache_oom_insufficient_memory`.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -3385,74 +3385,89 @@ def load_model(self, eep_scale_up: bool = False) -> None:
...
+            msg = (
+                "Failed to load model - not enough GPU memory. "
+                "Try lowering --gpu-memory-utilization to free memory for weights, "
+                "increasing --tensor-parallel-size, or using --quantization. "
+                "See https://docs.vllm.ai/en/latest/configuration/conserving_memory/ "
+                "for more tips."
+            )
+            combined_msg = f"{msg} (original error: {e})"
+            logger.error("init_failed msg=%s", combined_msg)
+            raise RuntimeError(combined_msg) from e
```
heheda12345: what about still raise OutOfMemoryError?

```suggestion
            raise e
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -3385,74 +3385,89 @@ def load_model(self, eep_scale_up: bool = False) -> None:
...
+        except torch.cuda.OutOfMemoryError as e:
+            msg = (
+                "Failed to load model - not enough GPU memory. "
+                "Try lowering --gpu-memory-utilization to free memory for weights, "
+                "increasing --tensor-parallel-size, or using --quantization. "
+                "See https://docs.vllm.ai/en/latest/configuration/conserving_memory/ "
+                "for more tips."
+            )
+            combined_msg = f"{msg} (original error: {e})"
+            logger.error("init_failed msg=%s", combined_msg)
```
heheda12345:
```suggestion
            logger.error(combined_msg)
```

heheda12345 (approved):
LGTM!

PR #24618, opened 2025-09-11
[Core] Reuse ZMQ to level trigger local ShmRingBuffer events.

noobpwnftw:
## Purpose

Avoids busy polling and no wake up latency from sleeps.

## Test Plan

None

## Test Result

Works

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #16226, opened 2025-04-08
[MISC][Bugfix] Use less CPU when message queue has been empty for some time

p12tic:
In setups which have long inactivity periods it is desirable to reduce system power consumption when vllm does nothing. This would lead to more CPU thermal headroom when a request eventually comes, especially when multiple GPUs are connected as each GPU would otherwise pin one thread at 100% CPU usage.

~I didn't include any configuration knobs because I couldn't think of anyone who would adversely impacted by this change. It introduces a maximum additional latency of 100ms when vllm workers are inactive for 10 seconds or more. Seems like someone who has inactive vllm wouldn't care about small additional latency. But please let me know if you think otherwise.~

EDIT: Final version of the PR contains new environment variable VLLM_SLEEP_WHEN_IDLE to enable CPU usage reduction. Set VLLM_SLEEP_WHEN_IDLE=1 to reduce power consumption at the expense of small amount of additional latency.

FIX #14799, https://github.com/vllm-project/vllm/issues/16968, https://github.com/vllm-project/vllm/issues/16660

PR #28053, opened 2025-11-04
[Core] Remove busy loop from idle buffer readers

joerunde:
## Purpose

IBM bundles vLLM with SystemZ mainframes, and in these systems:
1. The CPU cores are _extremely_ expensive
2. The systems are shared across many tenants (think multiple host operating systems and many VMs running on each) so it is important to not use any cpu cycles while not doing work.

This makes the fix from #16226 unacceptable for our customers, so this PR is a similar alternative to #24618 that completely removes the need for a busy loop at idle in the shared memory buffer readers.

This PR uses ZMQ to send write notifications which readers can wait for, however it also uses an in-process socket to wake up the reader on shut down so that there's no need to constantly wake up to check for a cancellation flag. This allows us to sleep indefinitely on a `zmq.Poller` so that the reader only wakes up when there is work to do, or it needs to shutdown.

I tried to contain the diff here inside of shm_broadcast.py as much as possible, the main change is swapping `SpinTimer` with `SpinCondition`. The changes that leaked out are:
- The `MessageQueue.deque()` api no longer takes a `cancel` event, instead it exposes a `.shutdown()` api that the thread running `monitor_parent_death` calls.
- The `VLLM_SLEEP_WHEN_IDLE` env var has been removed

## Test Plan
- [x] I need to write some little unit tests to cover `SpinCondition` before taking out of draft
- [x] Benchmarks need to be run to validate that this doesn't cause a performance regression

## Test Result

---
<details>

</details>


vllm/v1/executor/multiproc_executor.py
```diff
@@ -331,14 +327,18 @@ def wait_for_termination(procs, timeout):
                 time.sleep(0.1)
             return False
 
+        active_procs = lambda: [proc for proc in worker_procs if proc.is_alive()]
+
+        # Give processes time to clean themselves up properly
+        if wait_for_termination(active_procs(), 4):
```
joerunde, line 420: could shorten these `4`s so the total time is not so long in the worst case

vllm/distributed/device_communicators/shm_broadcast.py
```diff
@@ -49,40 +51,115 @@ def to_bytes_big(value: int, size: int) -> bytes:
...
+
+    def record_read(self):
+        self.last_read = time.monotonic()
+
+    def cancel(self):
+        # Sends cancellation ping that will cause the reader to wake up.
+        # This is done from a monitor thread in the same process as the reader.
+        if self.is_reader:
+            logger.debug("Canceling waiting reads on SHM Buffer")
+            self.write_cancel_socket.send(b"\x00")
```
joerunde, line 169: bug: should set a cancelled flag to check for in busy mode

(edit: or just move the `if self.shutting_down` to the top of the loop in `acquire_read`)

joerunde: update: tested and works

joerunde:
@tdoublep If you ca help get a benchmark on this, we've fixed the merge conflicts and we're seeing comparable performance and very low cpu usage at idle on gpus, with no cpu usage in the workers at idle when running vllm-spyre in eager mode

vllm/v1/executor/multiproc_executor.py
```diff
@@ -324,18 +322,14 @@ def collective_rpc(  # type: ignore[override]
...
             responses = []
             for mq in response_mqs:
                 dequeue_timeout = (
                     None if deadline is None else (deadline - time.monotonic())
                 )
                 try:
-                    status, result = mq.dequeue(
-                        timeout=dequeue_timeout, cancel=shutdown_event
-                    )
+                    status, result = mq.dequeue(timeout=dequeue_timeout)
```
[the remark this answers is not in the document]
joerunde: The entire point of the PR was to remove the cancel_event! 😉 

The closure of the death_pipe on shutdown triggers the read sides of these message queues to shut themselves down, causing all open reads to raise immediately

vllm/distributed/device_communicators/shm_broadcast.py
```diff
@@ -501,11 +597,24 @@ def acquire_write(self, timeout: float | None = None):
...
     ):
         assert self._is_local_reader, "Only readers can acquire read"
         start_time = time.monotonic()
+        if timeout is not None:
+            deadline = start_time + timeout
+            wait_timeout_ms: int | None = (
+                VLLM_RINGBUFFER_WARNING_INTERVAL * 1000
+                if not indefinite
+                else sys.maxsize
+            )
```
njhill: I feel like the timeout/deadline variable logic is a bit unwieldy .. would be good to simplify it a bit if possible

vllm/distributed/device_communicators/shm_broadcast.py
```diff
@@ -523,19 +632,21 @@ def acquire_read(
                     # for readers, `self.current_idx` is the next block to read
                     # if this block is not ready,
                     # we need to wait until it is written
+                    if timeout is not None:
```
njhill: probably clearer to check `deadline is not None` here instead

vllm/v1/executor/multiproc_executor.py
```diff
@@ -722,8 +714,14 @@ def monitor_parent_death():
...
                     logger.info_once("Parent process exited, terminating worker")
-                    # Send signal to self to trigger clean shutdown
-                    shutdown_event.set()
+                    nonlocal shutdown
+                    shutdown = True
+                    # Shut down message queues
+                    if worker is not None and worker.rpc_broadcast_mq is not None:
+                        worker.rpc_broadcast_mq.shutdown()
+                    if worker is not None and worker.worker_response_mq is not None:
+                        worker.worker_response_mq.shutdown()
```
njhill:
```suggestion
                    if worker is None:
                        return
                    if worker.rpc_broadcast_mq is not None:
                        worker.rpc_broadcast_mq.shutdown()
                    if worker.worker_response_mq is not None:
                        worker.worker_response_mq.shutdown()
```

vllm/v1/executor/multiproc_executor.py
njhill: May be best to move this lower? e.g. after creating the worker.

Could also move the thread setup into a separate static method which takes the death_pipe and worker as input

vllm/v1/executor/multiproc_executor.py
njhill: Maybe create a new worker `shutdown()` method containing this and call it here, and rename the existing one to `_shutdown()`. It can do the above including setting a `shutting_down = True` field, which can be checked below rather than the `shutdown` var.

njhill (commented):
Thanks @joerunde @tjohnson31415 @tdoublep for the great work and really sorry for taking so long to look at this properly.

I've run some of my own benchmarks and it looks good.

vllm/v1/executor/multiproc_executor.py
```diff
@@ -690,12 +683,34 @@ def wait_for_ready(
...
+            try:
+                # This will block until parent process exits (pipe closes)
+                death_pipe.recv()
+            except EOFError:
+                # Parent process has exited, terminate this worker
+                logger.info_once("Parent process exited, terminating worker")
+                shutdown_requested.set()
+                self.shutdown()
+            except Exception as e:
+                logger.warning("Death monitoring error: %s", e)
```
njhill, line 757: Ideally this does not have a reference to `self`, since it could cause GC issues with the worker.

vllm/v1/executor/multiproc_executor.py
```diff
@@ -690,12 +683,34 @@ def wait_for_ready(
...
+
+        def death_pipe_monitor():
+            try:
+                # This will block until parent process exits (pipe closes)
+                death_pipe.recv()
+            except EOFError:
+                # Parent process has exited, terminate this worker
+                logger.info_once("Parent process exited, terminating worker")
+                shutdown_requested.set()
+                self.shutdown()
```
njhill: I think it's better for the shutdown logic to run in the worker's main thread once it exits the main loop, as it does now.

So I don't think the queue shutdown calls should be moved into the same `shutdown()` method as is done above. Instead I think it would be better to do what I suggested earlier:
- Rename the existing shutdown to `_shutdown`
- Here just call the queue shutdowns. I was going to suggest having a separate worker `shutdown()` method for this but maybe better to just use the queue references directly to avoid needing a `self` reference in the thread (per the other comment)

vllm/distributed/device_communicators/shm_broadcast.py
```diff
@@ -503,16 +598,64 @@ def acquire_write(self, timeout: float | None = None):
...
+        def timeout_ms(self) -> int | None:
+            """Returns a timeout that is:
+            - min(time to deadline, time to next warning) if we're logging warnings
+            - time to deadline, if we're not logging warnings
+            - None if the timeout is None and we're not logging warnings
+            - raise TimeoutError if we are past the deadline
+            """
+            if self.timeout is None:
+                return self.warning_wait_timeout_ms
+            else:
```
njhill: nit: redundant else

tjohnson31415:
@njhill I pushed changes to address your review comments. PTAL

---

I was also doing some testing on the shutdown procedure with these changes and went down a bit of a rabbit hole... but came back with a couple of fixes:

1. I noticed that the death_pipe was not always getting EOF when the Engine process was shutdown and then the SIGKILL would be needed to bring all processes down. From my investigations, this was caused by hidden references to the death_writer in subsequent workers when using fork: Worker_TP1 would have a reference to the death_writer end of Worker_TP0 and just closing the end owned by the EngineCore would not trigger the EOF. The solution I added is to expliclty track all open socket file desriptors when launching workers and have workers close the inherited file descriptors that they do not need open.

2. Sending SIGTERM to a worker during the initialization phase would cause the workers to shutdown but EngineCore would stick around waiting for them to come up. The solution I added for this was to explicilty call `.shutdown()` on all message queues during the cleanup in `MultiprocExecutor.shutdown()` to make sure the EngineCore SpinCondition is notified to shut down.

---

Another scenario I looked at was with a kill (SIGTERM) sent to a worker after server startup. This leads to a hang because the signaled process does not fully shutdown, so its shutdown doesn't get noticed by the EngineCore. I traced the hang to the call to `torch.distributed.destroy_process_group()`. The server shutdown continues if all other workers also get SIGTERM and reach that point. I suspect this is could be a barrier in the process group or potentially related to https://github.com/pytorch/pytorch/issues/115388#issuecomment-3009880966 where NCCL blocks final destruction if CUDA graphs are still present.

Turns out that this behavior is the same on `main` though, so I didn't attempt a fix in this branch.

njhill:
Thanks for this @tjohnson31415, it was definitely needed! Maybe you could also look at https://github.com/vllm-project/vllm/pull/34816, I think we probably want to incorporate that too for last-resort case (when parent is gone there's probably no reason to shut down cleanly).

joerunde:
@njhill do you mean incorporating the changes from 34816 here instead of in that separate PR?

vllm/distributed/device_communicators/shm_broadcast.py
```diff
@@ -88,40 +90,115 @@ def long_wait_time_msg(threshold: int) -> str:
...
+
+        current_time = time.monotonic()
+        if current_time <= self.last_read + self.busy_loop_s:
             sched_yield()
+        else:
+            events = dict(self.poller.poll(timeout=timeout_ms))
+
+            if self.read_cancel_socket in events:
+                # return immediately on cancel
+                return
```
njhill: Are these lines needed?

tjohnson31415: Technically, no, not as is written. I could add a debug logging statement here though to indicate that a cancel was received (I had such a log statmenet during my testing) 😅

tjohnson31415:
> @tjohnson31415 thanks again for the investigation. The inherited_fds thing is unfortunate. Given that, I'm wondering whether we should rethink the death_pipe thing altogether. Re (3.), this may imply the need for some kind of worker healthcheck instead of or in addition to just monitoring the processes.

The `inherited_fds` is definitely quirky with `mp.Pipe()` when using forking... but, with proper handling, the EOF singal does give a mechanism to signal a hard shutdown that works even if a process is SIGKILL'd. When the process dies, the OS will clean up the pipe ends and trigger the EOF. So that's kinda nice for a cross-platform solution.

For hang detection for (3), we could add healthchecking as well, but we'd have to keep that at a low interval to not add back a busy CPU loop in the worker 😅.

> I will try to spend some more time on this tomorrow, maybe we can get the PR in and clean up the shutdown logic as a follow-on.

Thanks! A follow-on would be preferable. We are looking to get this in for the next vLLM release if possible.

njhill (approved):
Thanks @joerunde @tjohnson31415 @tdoublep 

Sorry for taking so long with the reviews!

I'm hoping we can further simplify the shutdown logic in a follow-on.

tjohnson31415:
I have a couple of follow up fixes for this change in:
https://github.com/vllm-project/vllm/pull/36068

@markmc I included code to check `getattr` on `response_mqs` to fix error you reported

PR #28058, opened 2025-11-04
[Docs] Switch to directory style URLs

hmellor:
A redirect will also be added to RTD so that the old style links that have been previously shared will continue to work.

image


russellb (approved):
+1 to this given the redirect option available to make it a smooth transition

PR #28059, opened 2025-11-04
[build][cmake]: Bundle static ACL and torch libgomp for CPU extension builds

Radu2k:
[build][cmake]: Bundle static ACL and torch libgomp for CPU extension builds

## Purpose
Bundle static ACL into C extensions. Use the same libgomp as PyTorch package for oneDNN and ACL build by adding path to `LD_LIBRARY_PATH `.

Fixes: #28056 

## Test Plan

- No more issues when vllm `build` folder, where ACL's so gets placed, is removed. 
- `LD_DEBUG=libs,files` only shows torch/lib/libgomp

## Test Result

- When we run vllm the following warning is missing: `WARNING 11-04 18:30:32 [interface.py:171] Failed to import from vllm._C: ImportError('libarm_compute.so: cannot open shared object file: No such file or directory')`
- `LD_DEBUG=libs,files` only shows `file=libgomp-1dd373e6.so.1.0.0 [0];  needed by /home/usr/venv/venv_name/lib/python<version>/site-packages/torch/lib/libtorch_cpu.so [0]`

---
<details>

</details>

### DIFF
--- a/cmake/cpu_extension.cmake
+++ b/cmake/cpu_extension.cmake
@@ -210,7 +210,30 @@ endif()
 if ((AVX512_FOUND AND NOT AVX512_DISABLED) OR (ASIMD_FOUND AND NOT APPLE_SILICON_FOUND) OR POWER9_FOUND OR POWER10_FOUND OR POWER11_FOUND)
     # Fetch and build Arm Compute Library (ACL) as oneDNN's backend for AArch64
     # TODO [fadara01]: remove this once ACL can be fetched and built automatically as a dependency of oneDNN
+    set(ONEDNN_AARCH64_USE_ACL OFF CACHE BOOL "")
     if(ASIMD_FOUND)
+        # Set number of parallel build processes
+        include(ProcessorCount)
+        ProcessorCount(NPROC)
+        if(NOT NPROC)
+            set(NPROC 4)
+        endif()
+        # locate PyTorch's libgomp (e.g. site-packages/torch.libs/libgomp-947d5fa1.so.1.0.0)
+        # and create a local shim dir with it
+        vllm_prepare_torch_gomp_shim(VLLM_TORCH_GOMP_SHIM_DIR)
+
+        find_library(OPEN_MP
+            NAMES gomp
+            PATHS ${VLLM_TORCH_GOMP_SHIM_DIR}
+            NO_DEFAULT_PATH
+            REQUIRED
+        )
+        # Set LD_LIBRARY_PATH to include the shim dir at build time to use the same libgomp as PyTorch
+        if (OPEN_MP)
+            set(ENV{LD_LIBRARY_PATH} "${VLLM_TORCH_GOMP_SHIM_DIR}:$ENV{LD_LIBRARY_PATH}")
+        endif()
+
+        # Fetch and populate ACL
         if(DEFINED ENV{ACL_ROOT_DIR} AND IS_DIRECTORY "$ENV{ACL_ROOT_DIR}")
             message(STATUS "Using ACL from specified source directory: $ENV{ACL_ROOT_DIR}")
         else()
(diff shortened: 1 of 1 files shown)


Radu2k:
cc: @fadara01

cmake/cpu_extension.cmake
```diff
@@ -191,9 +191,33 @@ endif()
 
 # Build oneDNN for GEMM kernels (only for x86-AVX512 /ARM platforms)
 if ((AVX512_FOUND AND NOT AVX512_DISABLED) OR (ASIMD_FOUND AND NOT APPLE_SILICON_FOUND) OR POWER9_FOUND OR POWER10_FOUND OR POWER11_FOUND)
+    # Set number of parallel build processes
+    include(ProcessorCount)
```
fadara01: let's put these changes under `ASIMD_FOUND` as x86 might not use libgomp in pytorch (i.e. they might be using libiomp or something else).

cmake/utils.cmake
```diff
@@ -194,13 +194,13 @@ macro(clear_cuda_arches CUDA_ARCH_FLAGS)
 endmacro()
 
 #
-# Extract unique CUDA architectures from a list of compute capabilities codes in
```
fadara01, line 197: I'm in favor of getting rid of white space changes in this PR

cmake/cpu_extension.cmake
```diff
@@ -191,9 +191,33 @@ endif()
...
+    ProcessorCount(NPROC)
+    if(NOT NPROC)
+        set(NPROC 4)
+    endif()
+
+    # locate PyTorch's libgomp (e.g. site-packages/torch.libs/libgomp-947d5fa1.so.1.0.0)
+    # and create a local shim dir with it
+    vllm_prepare_torch_gomp_shim(VLLM_TORCH_GOMP_SHIM_DIR)
+
+    find_library(OPEN_MP
```
fadara01: could you elaborate on why any libgomp related changes are needed here?
what problem do they address?

Radu2k: To avoid any undefined behaviour from using OMP, oneDNN and ACL both should use the same omp version as PyTorch.

fadara01: atm ACL will load pytorch's libgomp as per #27415
oneDNN has to be already linking against pytorch's libgomp atm, because that's the only libgomp that I see being loaded.
Please let me know if that's not already the case

Radu2k: That is mostly true.

You get oneDNN static library (libdnnl.a) compiled against compile time omp version which now we don't set it to pytorch's version. If runtime OMP is not ABI-compatible with compile time OMP, you can get link failures etc.

fadara01: > If runtime OMP is not ABI-compatible with compile time OMP

Yeah, we should make sure that pytorch's libgomp is used for both compile time and runtime, is that not already happening?

Radu2k: I don't see any omp specific variable being set for oneDNN which implies that it will use the system path. We specify pytorch omp solely for ACL. By adding it to LD_LIBRARY_PATH we make sure that oneDNN is going to pick up the version from VLLM_TORCH_GOMP_SHIM_DIR before checking system paths as well.

bigPYJ1151:
@Radu2k I prefers to staticlly link ACL to reduce changes. The oneDNN is also staticlly linked and looks the binary size is not very large. By disabling unused features and removing debug info of oneDNN, the CPU wheel size is ~80M.

fadara01 (approved):
LGTM

Radu2k:
@fadara01 I did run some benchmarks and we are not regressing with this change.

bigPYJ1151 (approved):
LGTM

PR #28069, opened 2025-11-04
[Refactor] Optimize `select_experts`

yewentao256:
## Purpose

Several small optimizations including:

- avoid mul if `routed_scaling_factor==1`
- avoid duplicate dtype cast when eplb is enabled

## Test

I don't see too much e2e perf improvement since these are small optimizations

Doesn't hurt accuracy:

```bash
lm_eval --model local-completions --model_args "base_url=http://127.0.0.1:9256/v1/completions,model=deepseek-ai/DeepSeek-R1,num_concurrent=1024" --tasks gsm8k
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9507|±  | 0.006|
|     |       |strict-match    |     5|exact_match|↑  |0.9500|±  | 0.006|
```


vllm/model_executor/layers/fused_moe/fused_moe.py
```diff
@@ -1173,7 +1173,6 @@ def grouped_topk(
         and current_platform.is_cuda()
         and num_expert_group <= 32
         and topk <= 32
-        and e_score_correction_bias is not None
```
robertgshaw2-redhat, line 1175: so does this mean we havent been using this kernel all this time ?

yewentao256: @robertgshaw2-redhat Yes, when `e_score_correction_bias` is None it is not used

robertgshaw2-redhat: and deepseek uses this :)

yewentao256: Yeah, we should prefer fused_group_topk since it is faster than normal one

mgoin: I don't understand, DeepSeek R1 has e_score_correction_bias so it already gets used

vllm/model_executor/layers/fused_moe/fused_moe.py
```diff
@@ -1322,14 +1321,25 @@ def fused_grouped_topk(
...
+    # allow None bias by substituting a zero vector
+    bias = (
+        e_score_correction_bias.to(gating_output.dtype)
+        if e_score_correction_bias is not None
+        else torch.zeros(
+            gating_output.shape[-1],
+            device=gating_output.device,
+            dtype=gating_output.dtype,
+        )
+    )
```
mgoin: Why don't we just update the kernel to allow for bias to be None, rather than allocating?

mgoin (changes requested):

vllm/model_executor/layers/fused_moe/fused_moe.py
yewentao256: Don't see too much benefits here as well, just rolled back the update for `e_score_correction_bias`

vllm/model_executor/layers/fused_moe/fused_moe.py
yewentao256: Tried but found that it involves lots of cuda kernel changes, I don't see too much benefits here as well, just rolled back the update for `e_score_correction_bias` @mgoin

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -2238,7 +2236,7 @@ def select_experts(
                 topk=top_k,
                 renormalize=renormalize,
             )
-            if routed_scaling_factor is not None:
+            if routed_scaling_factor != 1.0:
```
mgoin, line 1504: What do you gain by changing this? It seems there are still other places that set routed_scaling_factor to None, and you could just as well update the models that don't to pass in None instead of 1.0
https://github.com/search?q=repo%3Avllm-project%2Fvllm%20routed_scaling_factor%3DNone&type=code

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -2268,9 +2264,12 @@ def select_experts(
                 expert_load_view=expert_load_view,
                 logical_to_physical_map=logical_to_physical_map,
                 logical_replica_count=logical_replica_count,
-                indices_type=indices_type,
+                indices_type=None,
             )
 
+        if (indices_type is not None) and topk_ids.dtype != indices_type:
+            topk_ids = topk_ids.to(dtype=indices_type)
```
mgoin, line 1535: I don't know if this is valid since other conditionals above use indices_type..

vllm/model_executor/layers/fused_moe/layer.py
yewentao256: Thanks! Updated this None to 1.0 by default.
This is beneficial since we use a lot of 1.0 somewhere else https://github.com/search?q=repo%3Avllm-project%2Fvllm+routed_scaling_factor%3D1.0&type=code
And we avoid a useless mul operator through doing this

vllm/model_executor/layers/fused_moe/layer.py
yewentao256: I remove the param in `eplb_map_to_physical_and_record` since it is just for a `topk_ids.to(dtype=indices_type)` and now we can see it more clearly.

The reason we update here is to avoid a duplicate `topk_ids.to(dtype=indices_type)`. Eg. when using `use_grouped_topk`, we already `topk_ids.to(dtype=indices_type)`, then we pass the `indices_type` to `eplb_map_to_physical_and_record` if eplb is enabled, we cast again.

vllm/model_executor/layers/quantization/modelopt.py
```diff
@@ -1699,7 +1699,7 @@ def apply(
                 intermediate_size=layer.intermediate_size_per_partition,
                 local_expert_offset=layer.ep_rank * layer.local_num_experts,
                 local_num_experts=layer.local_num_experts,
-                routed_scaling_factor=None,
+                routed_scaling_factor=1.0,
```
mgoin, line 1706: Sorry for pointing this out, I think this should stay None since it calls a kernel directly `flashinfer.fused_moe.trtllm_fp4_block_scale_moe`

yewentao256: Fixed, thanks!

vllm/model_executor/layers/fused_moe/fused_moe.py
mgoin, line 1: @yewentao256 It looks like `_eplb_map_to_physical_and_record` needs to be updated as well in this file for the cpu backend

yewentao256: fixed, thanks!

mgoin (approved):

Issue #28072, opened 2025-11-04
[Bug]: WeightsMapper used in QuantizationConfig does not handle regex/wildcard paths


shengliangxu:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
Your output of `python collect_env.py` here
```

</details>

### 🐛 Describe the bug

vLLM model structure may be different from HuggingFace model structure so a remapping is used (WeightsMapper) when loading models. The remapping is defined in model executor logic, e.g. the [Qwen2.5 VL defines it here](https://github.com/vllm-project/vllm/blob/2d977a7a9ead3179fde9ed55d69393ef7b6cec47/vllm/model_executor/models/qwen2_5_vl.py#L1096-L1105)

In quantized models, in addition to the model weights, there may be extra configs such as to describe what modules are not quantized. For example, the compressed-tensors has an "ignore" in the config as a list of modules that shall be treated as not quantized; the Nvidia ModelOpt has an "exclude_modules" also in the config as a list of modules that shall be treated as not quantized. The issue is that these configs may not be simple module path/prefix names, they may be regex (in compressed-tensors) or wildcards (in Nvidia ModelOpt).

Take compressed-tensors as an example given it's also a vLLM owned project. An item in the ignore list may be a regex pattern such as "re:vision_tower.*" to exclude the whole vision encoder in a VLM. One random example quantized model that I find on HF:

https://huggingface.co/gaunernst/gemma-3-27b-it-qat-compressed-tensors

In it's config, it has the ignore list of:

"ignore": [
    "lm_head",
    "re:vision_tower.*"
],

When vLLM loads the quantization configs, it also applies weights mapping through [apply_vllm_mapper](https://github.com/vllm-project/vllm/blob/2d977a7a9ead3179fde9ed55d69393ef7b6cec47/vllm/model_executor/layers/quantization/base_config.py#L153), but the [WeightsMapper](https://github.com/vllm-project/vllm/blob/2d977a7a9ead3179fde9ed55d69393ef7b6cec47/vllm/model_executor/models/utils.py#L42) assumes that it just work with each single weight names. It will break if it hits a regex or a wildcard. For example, for the above ignore list, if a model has weight remap of:

prefix remap: vision_tower. -> vision.

Then the remap will miss the regular expression: "re:vision_tower.*".

[...]

PR #28076, opened 2025-11-04
Consolidate Nvidia ModelOpt quant config handling for all quantization methods

shengliangxu:
## Purpose

Two major changes in this patch:

1. Consolidate Nvidia ModelOpt quant config handling across different quantization methods (FP8 and NVFP4 as of now). Especially the handling of exclude_modules. Different quantization methods share the same exclude_modules handling. Currently in the code we handle the exclude modules as patterns for the NVFP4 quant method but the FP8 quant format does not handle it. This change move the handling of exclude_modules and the determination of quant methods for layers in common code.

2. The Nvidia ModelOpt library exclude modules semantic is different from vllm's current logic. The exclude modules are wildcards instead of simple module prefix strings/substrings.

## Test Plan

Successfully deployed several LLM models, including Qwen2.5-VL, Qwen3-8B, Qwen3-VL, Llama3.1-70B, Llama4-Scout with both NVFP4 and FP8 quantization

## Test Result

Some deployment hit issues that is not related to this change, which is reported here: https://github.com/vllm-project/vllm/issues/28072

Deployments all succeed with workaround to the above unrelated issue.

---
<details>

</details>


heheda12345:
CC @mgoin @yewentao256

vllm/model_executor/layers/quantization/modelopt.py
pavanimajety: Can we also refactor the from_config? The logic is very repetitive here as well.

yewentao256 (commented):
The idea looks good to me, could you also show `lm_eval` metrics to make sure we don't hurt accuracy?

vllm/model_executor/layers/quantization/modelopt.py
shengliangxu: sorry for the late reply, got swamped for some other stuff, let me check it.

shengliangxu: Done

vllm/model_executor/layers/quantization/modelopt.py
```diff
@@ -86,45 +86,214 @@
...
+
+        # handle exclusion
+        if self.is_layer_excluded(prefix):
+            if isinstance(layer, LinearBase):
+                return UnquantizedLinearMethod()
+            return None
+
+        # Check if this is a vision model layer that should not be quantized
+        # TODO: this special hard coded logic is not needed after the next release of
+        # ModelOpt where they are handled natually by the exclude_modules config
```
mgoin: Please put an expected version number so we know when "next" is

shengliangxu: sg, updated the version number

mgoin (approved):

shengliangxu:
Thanks for merging it @mgoin , I was working on getting lm_eval results as @yewentao256  suggested. It took me quite some time because of hitting quite some issues but then turned out some of my configs were incorrect. Here's one result following[ this doc ](https://docs.vllm.ai/projects/recipes/en/latest/Llama/Llama4-Scout.html#verify-accuracy)(really nice documentation by the way).

In the doc the accuracy numbers (i.e. without this change) are:


|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8946|±  |0.0085|
|     |       |strict-match    |     5|exact_match|↑  |0.8779|±  |0.0090|

With this change, it's:

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8992|±  |0.0083|
|     |       |strict-match    |     5|exact_match|↑  |0.8795|±  |0.0090|

Both run are on B200.

So accuracy is not impacted.

PR #27308, opened 2025-10-21
[Speculators] Move tests + fix integration

dsikka:
This PR fixes the speculator model integration that enables the simplified `vllm serve <speculator-model>` command and ensures compatibility with S3 models in CI.

## Background

The speculator integration allows users to run speculative decoding without explicitly providing a `--speculative-config` by automatically detecting speculator models and extracting the configuration. This integration kept breaking on main because:

1. Tests were located in `tests/speculative_decoding/speculators/` which doesn't run in CI
2. The integration didn't properly handle S3 models used in CI testing

### 1. Moved Speculator Tests to CI-Monitored Directory
- **Before**: `tests/speculative_decoding/speculators/test_eagle3.py`
- **After**: `tests/v1/spec_decode/test_speculators_eagle3.py`

This ensures tests run in CI and prevent future breakage.

### 2. Fixed S3 Model Compatibility

**Problem**:
- Speculator detection (`maybe_override_with_speculators()`) was moved before `create_model_config()` to properly detect speculators before creating the model config
- However, HuggingFace's `PretrainedConfig.get_config_dict()` cannot load configs from S3 URLs (`s3://...`)
- This caused failures when running with S3 models in CI

**Solution**:
Skip speculator auto-detection for S3 models:

```python
# Skip speculator detection for S3 models since HuggingFace cannot load
# configs directly from S3 URLs. S3 models can still use speculators with
# explicit --speculative-config.
if not is_s3(self.model):
    (self.model, self.tokenizer, self.speculative_config) = (
        maybe_override_with_speculators(...)
    )
```

**Trade-off**: S3 models cannot use automatic speculator detection but can still use speculators via explicit `--speculative-config` argument.

### 3. Added Comprehensive Integration Test

Added `test_speculators_model_integration()` in `tests/v1/e2e/test_spec_decode.py` to validate the simplified integration path:

**What it tests**:
- ✅ Speculator model auto-detection without explicit config
- ✅ Speculative config correctly extracted from model
- ✅ Verifier model properly identified
- ✅ Draft model set to speculator model
- ✅ Text generation works with speculative decoding
- ✅ Output correctness (66% match threshold vs reference)

**Test models**:
- `nm-testing/SpeculatorLlama3-1-8B-Eagle3-converted-0717-quantized`
- `nm-testing/Speculator-Qwen3-8B-Eagle3-converted-071-quantized`

## Compatibility Matrix

| Model Type | Speculator Auto-Detection | Manual `--speculative-config` |
[...]

PR #25250, opened 2025-09-19
feat: Enable engine-level arguments with speculators models

rahul-tuli:
This PR fixes two critical issues with speculators models that were running without speculative decoding enabled, and enables users to combine engine-level arguments (like `--tensor-parallel-size`, `--seed`, `--max-model-len`) with speculators models using simplified command syntax.

## Problem

Previously, there were two issues with speculators models:

1. **Speculators models were running but NOT using speculative decoding** - They were treated as regular models, missing the performance benefits of speculative decoding entirely, this happened because while the verifier model and tokenizer were updated the speculative config was not being initilialized
2. **Engine-level arguments were incompatible** - Users had to use verbose commands with explicit speculative configuration:

```bash
VLLM_USE_V1=1 vllm serve "RedHatAI/Meta-Llama-3.1-8B-Instruct-FP8-dynamic" \
    --seed 42 \
    --tensor-parallel-size 4 \
    --speculative-config '{"model": "nm-testing/SpeculatorLlama3-1-8B-Eagle3-converted-0717-quantized", "num_speculative_tokens": 3, "method":"eagle3"}'
```

The speculators model detection was happening too late in the configuration pipeline (in `ModelConfig`), which meant:
- The speculative configuration was never properly initialized
- Engine-level arguments couldn't be applied to the correct models
- Users lost the core benefit of using speculators models

## Solution

This PR fixes both issues:

1. **Enables proper speculative decoding** - Speculators models now initializes speculative configuration and run with the intended performance optimizations
2. **Simplified syntax with engine-level arguments** - Users can now use the clean syntax:

```bash
vllm serve --seed 42 --tensor-parallel-size 4 "nm-testing/SpeculatorLlama3-1-8B-Eagle3-converted-0717-quantized"
```

The model will now correctly:
- Detect that it's a speculators model
- Extract the embedded speculative configuration
- Use speculative decoding for improved performance
- Apply engine-level arguments to both target and draft models

### Core Changes

1. **Moved speculators detection earlier**: Relocated from `ModelConfig` to `EngineArgs` to ensure proper argument processing
2. **Improved function naming for readability**:
   - `convert_speculators_to_vllm` → `build_vllm_speculative_config`
   -  `extract_vllm_speculative_config` to add proper validation and algorithm level updates
   - `maybe_override_with_speculators_target_model` → `maybe_override_with_speculators`
3. **Enhanced configuration merging**: Engine-level CLI arguments now take precedence over embedded settings
[...]

PR #24435, opened 2025-09-08
[Bugfix] when use s3 model cannot use default load_format

lengrongfu:
## Purpose

https://github.com/vllm-project/vllm/issues/23236#issuecomment-3254391507

Current cannot use default loader model from s3 , so should to check it.

FIX #24313

## Test Result

image

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

Issue #24313, opened 2025-09-05
[Bug]: Starting from v0.10.1 model can't be loaded from s3


sh1ng:
### Your current environment

production-stack + vllm

### 🐛 Describe the bug

The following log is self-describing

```
(APIServer pid=1) INFO 09-05 03:10:33 [api_server.py:1805] vLLM API server version 0.10.1
(APIServer pid=1) INFO 09-05 03:10:33 [utils.py:326] non-default args: {'model_tag': 's3://h2o-models/llmops/h2o-danube3-500m-chat/', 'host': '0.0.0.0', 'model': 's3://h2o-models/llmops/h2o-danube3-500m-chat/', 'trust_remote_code': True, 'load_format': 'runai_streamer', 'gpu_memory_utilization': 0.6, 'enable_prefix_caching': False}
(APIServer pid=1) The argument `trust_remote_code` is to be used with Auto classes. It has no effect here and is ignored.
(APIServer pid=1) Traceback (most recent call last):
(APIServer pid=1)   File "/usr/local/lib/python3.12/dist-packages/transformers/utils/hub.py", line 479, in cached_files
(APIServer pid=1)     hf_hub_download(
(APIServer pid=1)   File "/usr/local/lib/python3.12/dist-packages/huggingface_hub/utils/_validators.py", line 106, in _inner_fn
(APIServer pid=1)     validate_repo_id(arg_value)
(APIServer pid=1)   File "/usr/local/lib/python3.12/dist-packages/huggingface_hub/utils/_validators.py", line 154, in validate_repo_id
(APIServer pid=1)     raise HFValidationError(
(APIServer pid=1) huggingface_hub.errors.HFValidationError: Repo id must be in the form 'repo_name' or 'namespace/repo_name': 's3://h2o-models/llmops/h2o-danube3-500m-chat/'. Use `repo_type` argument if needed.
(APIServer pid=1) 
(APIServer pid=1) During handling of the above exception, another exception occurred:
(APIServer pid=1) 
(APIServer pid=1) Traceback (most recent call last):
(APIServer pid=1)   File "/usr/local/lib/python3.12/dist-packages/transformers/configuration_utils.py", line 708, in _get_config_dict
(APIServer pid=1)     resolved_config_file = cached_file(
(APIServer pid=1)                            ^^^^^^^^^^^^
(APIServer pid=1)   File "/usr/local/lib/python3.12/dist-packages/transformers/utils/hub.py", line 321, in cached_file
[...]

PR #28079, opened 2025-11-04
Add runai model streamer e2e test for GCS

amacaskill:
## Purpose
The purpose of this is to add a basic RunAI model streamer e2e test that pulls from a public GCS bucket, and add this to the CI pipeline,  to prevent regressions in the RunAI model streamer. Examples of past RunAI model streamer regressions that this e2e test would have caught is: 
- https://github.com/vllm-project/vllm/pull/27308
- https://github.com/vllm-project/vllm/pull/25250
- Issue https://github.com/vllm-project/vllm/issues/24313 (can't find PR that caused it). I believe fix was https://github.com/vllm-project/vllm/pull/24435. 

We also need code coverage for [vllm/config/model.py](https://github.com/vllm-project/vllm/blob/6cae1e53326ab89551352b926854af7cd96b688a/vllm/config/model.py#L779)

This test uses a small model, [codegemma-2b](https://huggingface.co/google/codegemma-2b/tree/main) which is around 5.6 GiB. 

## Test Plan
We plan to enable all of the RunAI model streamer tests in the CI pipeline. Particularly, we want to run the tests on changes to any of the following: 
```
  source_file_dependencies:
  - vllm/engine
  - vllm/model_executor/model_loader
  - tests/model_executor/model_loader/runai_model_streamer
```
We also want to run the tests on the torch_nightly. 

We didn't add this to `Model Executor Test` because we would like to run the RunAI model streamer tests on changes to `vllm/engine` , `tests/model_executor/runai_model_streamer`, and on the nightly build (which `Model Executor Test` doesn't currently do).  

## Test Result
```
# All runai_model_streamer tests pass (including new test I added)
pytest -s -v tests/model_executor/model_loader/runai_model_streamer
=========================================== 7 passed, 3 warnings in 76.94s (0:01:16) ===========================================

# The new test I added passes: 
pytest -s -v tests/model_executor/model_loader/runai_model_streamer/test_runai_model_streamer_loader.py::test_runai_model_loader_download_gcs_files
PASSED
======================================================= warnings summary =======================================================
<frozen importlib._bootstrap>:488
  <frozen importlib._bootstrap>:488: DeprecationWarning: builtin type SwigPyPacked has no __module__ attribute

<frozen importlib._bootstrap>:488
  <frozen importlib._bootstrap>:488: DeprecationWarning: builtin type SwigPyObject has no __module__ attribute

.venv/lib/python3.12/site-packages/schemathesis/generation/coverage.py:305
  /root/vllm/.venv/lib/python3.12/site-packages/schemathesis/generation/coverage.py:305: DeprecationWarning: jsonschema.exceptions.RefResolutionError is deprecated as of version 4.18.0. If you wish to catch potential reference resolution errors, directly catch referencing.exceptions.Unresolvable.
    ref_error: type[Exception] = jsonschema.RefResolutionError,

================================================ 1 passed, 3 warnings in 47.99s ================================================
sys:1: DeprecationWarning: builtin type swigvarlink has no __module__ attribute
(vllm) root@vllm-g2-vm:~/vllm# 
```

---
<details>

</details>

### DIFF
--- a/.buildkite/test-pipeline.yaml
+++ b/.buildkite/test-pipeline.yaml
@@ -543,8 +543,11 @@ steps:
 
 - label: Model Executor Test # 23min
   timeout_in_minutes: 35
+  torch_nightly: true
   mirror_hardwares: [amdexperimental]
   source_file_dependencies:
+  - vllm/engine/arg_utils.py
+  - vllm/config/model.py
   - vllm/model_executor
   - tests/model_executor
   - tests/entrypoints/openai/test_tensorizer_entrypoint.py
--- a/tests/model_executor/model_loader/runai_model_streamer/test_runai_model_streamer_loader.py
+++ b/tests/model_executor/model_loader/runai_model_streamer/test_runai_model_streamer_loader.py
@@ -1,12 +1,16 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 
+import pytest
+
 from vllm import SamplingParams
 from vllm.config.load import LoadConfig
 from vllm.model_executor.model_loader import get_model_loader
 
 load_format = "runai_streamer"
 test_model = "openai-community/gpt2"
+# TODO(amacaskill): Replace with a GKE owned GCS bucket.
+test_gcs_model = "gs://vertex-model-garden-public-us/codegemma/codegemma-2b/"
 
 prompts = [
     "Hello, my name is",
@@ -30,5 +34,18 @@ def test_get_model_loader_with_runai_flag():
 
 def test_runai_model_loader_download_files(vllm_runner):
     with vllm_runner(test_model, load_format=load_format) as llm:
+        deserialized_outputs = llm.generate(prompts, sampling_params)
+        assert deserialized_outputs
+
+
+def test_runai_model_loader_download_files_gcs(
+    vllm_runner, monkeypatch: pytest.MonkeyPatch
+):
+    monkeypatch.setenv("GOOGLE_CLOUD_PROJECT", "fake-project")
+    monkeypatch.setenv("RUNAI_STREAMER_GCS_USE_ANONYMOUS_CREDENTIALS", "true")
+    monkeypatch.setenv(
+        "CLOUD_STORAGE_EMULATOR_ENDPOINT", "https://storage.googleapis.com"
+    )
+    with vllm_runner(test_gcs_model, load_format=load_format) as llm:
         deserialized_outputs = llm.generate(prompts, sampling_params)
         assert deserialized_outputs


tests/model_executor/model_loader/runai_model_streamer/test_runai_model_streamer_loader.py
```diff
@@ -32,3 +33,8 @@ def test_runai_model_loader_download_files(vllm_runner):
     with vllm_runner(test_model, load_format=load_format) as llm:
         deserialized_outputs = llm.generate(prompts, sampling_params)
         assert deserialized_outputs
+
+def test_runai_model_loader_download_files_gcs(vllm_runner):
+    with vllm_runner(test_gcs_model, load_format=load_format) as llm:
+        deserialized_outputs = llm.generate(prompts, sampling_params)
+        assert deserialized_outputs
```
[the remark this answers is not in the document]
amacaskill: I'll look into fixing this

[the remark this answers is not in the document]
amacaskill: I thought about this, and a regression within the RunAI model streamer's responsibility (i.e., downloading the model from GCS) should manifest as a total failure. The vllm_runner would likely raise an exception, or llm.generate() would return an empty list ([]). I don't see how streamer-specific error to result in an empty tuple for just one prompt in a batch, as that kind of failure is more indicative of an issue during the model's inference process, which happens after the streamer has already done its job.

There are other vllm tests tests in CI that confirm other parts of vLLM engine (that aren't related to model loading) don't regress, so keeping this as is.

.buildkite/test-pipeline.yaml
```diff
@@ -541,6 +541,18 @@ steps:
...
+  # Cross continent streaming could take much take much longer.
+  timeout_in_minutes: 10
+  torch_nightly: true
+  mirror_hardwares: [amdexperimental]
+  source_file_dependencies:
+  - vllm/engine
+  - vllm/model_executor/model_loader
+  - tests/model_executor/model_loader/runai_model_streamer
+  commands:
+    - pytest -v -s model_executor/model_loader/runai_model_streamer
```
DarkLight1337: This should already be covered by the Model Executor test below, or did I miss something?

DarkLight1337: We would like to avoid having very small test jobs because each job takes around 5 minutes to spin up, making initialization take up most of the time which is not efficient.

amacaskill: I added in PR description reasoning for why I added a new one: We didn't add this to Model Executor Test because we would like to (1) run the RunAI model streamer tests on changes to `vllm/engine`  and `tests/model_executor/runai_model_streamer`, and we want to run this on the nightly build (which Model Executor Test doesn't currently do).

Should I instead just make all of these changes to Model Executor Test? (enable Model Executor Test  for new source_file_dependencies, and enable on torch nightly build?)

DarkLight1337: Why do we need `vllm/engine` to be included? If it's just about the `arg_utils` file then you can select that file specifically and make all the changes to Model Executor tests

amacaskill: Yes its just arg_utils that we need, so I can just add that file instead of all of `vllm/engine`

amacaskill: I updated the .buildkite/test-pipeline.yaml to add to Model Executor test instead of creating its own label.

amacaskill:
@DarkLight1337  I have another question. Say there is a refactor of the RunAI model streamer, so now logic that could break the streamer, is in another file (file X), and file X isn't  listed in `source_file_dependencies`. If someone breaks the RunAI model streamer logic in file X, then the runai model streamer test will not be run on the PR presubmit, and it will run in the torch nightly build for the first time, and fail. 

When a test fails in torch nightly, what happens? Who is responsible for fixing that failure? Does the failure block the nightly build and/or the next vllm release? Or does vllm just proceed with the release/build, with that feature broken?

DarkLight1337 (approved):

DarkLight1337 (approved):

PR #28082, opened 2025-11-05
fix test_cudagraph_sizes_post_init

BoyuanFeng:
This PR fixes unit test failures for test_cudagraph_sizes_post_init.

cc @ProExpertProg @fhl2000

### DIFF
--- a/tests/compile/test_config.py
+++ b/tests/compile/test_config.py
@@ -281,7 +281,7 @@ def test_cudagraph_sizes_post_init(
     expected_max_size,
 ):
     ctx = nullcontext()
-    if expected_max_size == ValidationError:
+    if issubclass(expected_max_size, Exception):
         ctx = pytest.raises(expected_max_size)
 
     cudagraph_mode = CUDAGraphMode.PIECEWISE if use_cudagraph else CUDAGraphMode.NONE
--- a/tests/compile/test_config.py
+++ b/tests/compile/test_config.py
@@ -4,6 +4,7 @@ import copy
 from contextlib import nullcontext
 
 import pytest
+from pydantic import ValidationError
 
 from vllm.compilation.counter import compilation_counter
 from vllm.compilation.fix_functionalization import FixFunctionalizationPass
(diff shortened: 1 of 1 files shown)


tests/compile/test_config.py
```diff
@@ -253,21 +254,21 @@ def test_resolve_operator_overload():
         "expected_max_size",
     ),
     [
-        (None, None, 1, False, 2048, True, 512),
+        (None, None, 1, False, 2048, True, 256),
         ([1, 2, 4], 4, 1, False, 2048, True, 4),
-        ([1, 2, 4], 8, 1, False, 2048, True, RuntimeError),
```
BoyuanFeng, line 258: config raises `ValidationError` instead of `RuntimeError`.

tests/compile/test_config.py
```diff
@@ -253,21 +254,21 @@ def test_resolve_operator_overload():
...
-        ([1, 256], None, 1, False, 2048, 256),
```
BoyuanFeng, line 259: this setting has only 6 params, but the test requires 7 params.

tests/compile/test_config.py
```diff
@@ -280,7 +281,7 @@ def test_cudagraph_sizes_post_init(
     expected_max_size,
 ):
     ctx = nullcontext()
-    if isinstance(expected_max_size, Exception):
```
BoyuanFeng, line 283: image

tests/compile/test_config.py
```diff
@@ -253,21 +254,21 @@ def test_resolve_operator_overload():
         "expected_max_size",
     ),
     [
-        (None, None, 1, False, 2048, True, 512),
+        (None, None, 1, False, 2048, True, 256),
```
fhl2000, line 257: Why default 256 instead of 512?

fhl2000 (commented):
Hi, does this config test  (the whole file) run in CI now? I think it didn't actually run in the test before😂, and #27593 also fixed.

tests/compile/test_config.py
BoyuanFeng: The default number is 256 now:

https://github.com/vllm-project/vllm/blob/e67511884970af818bab20af3782ccebe08c716b/vllm/config/vllm.py#L766-L769

https://github.com/vllm-project/vllm/blob/e67511884970af818bab20af3782ccebe08c716b/vllm/config/scheduler.py#L177-L178

fhl2000: https://github.com/vllm-project/vllm/blob/18b39828d90413d05d770dfd2e2f48304f4ca0eb/vllm/engine/arg_utils.py#L1856-L1876
I think we have an overwrite if it starts from EngineArgs. So it is device-dependent, but I think the typical value of max_num_seqs is 256 in CI.

PR #18608, opened 2025-05-23
[V1][Sampler] Improve performance of FlashInfer sampling by sampling logits instead of probs

lgeiger:
This PR replaces `flashinfer.sampling.top_k_top_p_sampling_from_probs` with `flashinfer.sampling.top_k_top_p_sampling_from_logits`.

The `top_k_top_p_sampling_from_props` path calls [`(softmax) -> top_k_renorm_probs -> top_p_sampling_from_probs`](https://github.com/flashinfer-ai/flashinfer/blob/8bc22a1a6c58da1e0a02605f3b19b5d9e6224ad1/flashinfer/sampling.py#L1016-L1017) , `top_k_top_p_sampling_from_logits` calls [`top_k_mask_logits -> softmax -> top_p_sampling_from_probs`](https://github.com/flashinfer-ai/flashinfer/blob/8bc22a1a6c58da1e0a02605f3b19b5d9e6224ad1/flashinfer/sampling.py#L901-L903) which is faster.

In a quick micro benchmark on an **L40s GPU** I am seeing a **9.3 % speedup** with this PR and jitted flashinfer using Cuda 12.8.

<details><summary><strong>Expand for script to reproduce toy benchmark</strong></summary>
<p>

```python
import time

import torch
import flashinfer.sampling

from vllm.platforms import current_platform
from vllm.utils import STR_DTYPE_TO_TORCH_DTYPE, FlexibleArgumentParser


@torch.inference_mode()
def main(
    batch_size: int,
    num_classes: int,
    dtype: torch.dtype,
    seed: int = 0,
    num_warmup_iters: int = 5,
    num_iters: int = 100,
) -> None:
    current_platform.seed_everything(seed)
    torch.set_default_device("cuda")

    logits = torch.randn(batch_size, num_classes, dtype=dtype)
    k = torch.ones(batch_size, dtype=torch.int32) * 64
    p = torch.ones(batch_size, dtype=dtype) * 0.95

    def run_cuda_benchmark(num_iters: int) -> float:
        torch.cuda.synchronize()
        start_time = time.perf_counter()

        for _ in range(num_iters):
            # probs = logits.softmax(dim=-1, dtype=torch.float32)
            # next_token_ids = flashinfer.sampling.top_k_top_p_sampling_from_probs(
            # probs, k, p, deterministic=True)
            next_token_ids = flashinfer.sampling.top_k_top_p_sampling_from_logits(
                logits, k, p, deterministic=True
            )
        torch.cuda.synchronize()

[...]

PR #28099, opened 2025-11-05
Flashinfer: TopK+TopP sampling from probs

Kh4L:
Recent Flashinfer versions of `top_k_top_p_sampling_from_logits` are much slower on different systems:

 GPU | from logits (tokens/s) | from probs (tokens/s) | Improvement (%)
-- | -- | -- | --
SPARK | 19.78 | 28.50 | 44.1%
GB200 | 144.84 | 216.60 | 49.5%
H100 | 97.19 | 130.65 | 34.4%

### DIFF
--- a/vllm/v1/sample/ops/topk_topp_sampler.py
+++ b/vllm/v1/sample/ops/topk_topp_sampler.py
@@ -282,8 +282,9 @@ def flashinfer_sample(
         )
     else:
         # Both top-k and top-p.
-        next_token_ids = flashinfer.sampling.top_k_top_p_sampling_from_logits(
-            logits, k, p, deterministic=True
+        probs = logits.softmax(dim=-1, dtype=torch.float32)
+        next_token_ids = flashinfer.sampling.top_k_top_p_sampling_from_probs(
+            probs, k, p, deterministic=True
         )
 
     return next_token_ids.view(-1)


vllm/v1/sample/ops/topk_topp_sampler.py
```diff
@@ -282,8 +282,9 @@ def flashinfer_sample(
         )
     else:
         # Both top-k and top-p.
-        next_token_ids = flashinfer.sampling.top_k_top_p_sampling_from_logits(
-            logits, k, p, deterministic=True
+        probs = logits.softmax(dim=-1, dtype=torch.float32)
+        next_token_ids = flashinfer.sampling.top_k_top_p_sampling_from_probs(
+            probs, k, p, deterministic=True
```
lgeiger, line 287: This is interesting, does this mean `top_k_mask_logits` is slow for you?
https://github.com/flashinfer-ai/flashinfer/blob/6d19a75a6963b0996b7b2cfe5f9e385c5cc41a66/flashinfer/sampling.py#L1023-L1033

Might be worth also opening an issue on flashinfer since looking at their code, I don't really see a reason for why `top_k_top_p_sampling_from_probs` should be faster.

For reference, in #18608 I saw the opposite behaviour

heheda12345:
CC @mgoin as you were reviwing https://github.com/vllm-project/vllm/pull/18608

Kh4L:
@mgoin I do not have a strong opinion on that and I think it can be discussed in a separate "Issue", but I think it makes sense to switch to this flashinfer stable kernel in the meantime

vllm/v1/sample/ops/topk_topp_sampler.py
Kh4L: >This is interesting, does this mean top_k_mask_logits is slow for you?

It is with the latest versions of Flashinfer

I believe the the perf changed with Flashinfer updates

PR #28105, opened 2025-11-05
[Misc]pass the active batch size through logger

frank-wei:
Summary: The FINAL_ONLY in sampling parameter will only output completed requests. When we want to query the active requests/batch size, we can not directly get it through len(outputs). A better way is to pass it through the logger.

Differential Revision: D86217922


zhuohan123 (changes requested):
Should we log this field directly? Instead of having it as a public class field for others to access.

PR #28122, opened 2025-11-05
[Bugfix]: missing partial content if openai tool calling is enabled

dr75:
## Purpose
The OpenAIToolParser overrides the code of regular response parsing when enabled. This breaks partial responses when the token limit is reached and harmony does not generate a final message but only `current_content` is left, which is ignored.

To solve this, return the parsers `current_content` as it is also done in the non-tool calling case.

The code should probably be refactored such that the logic from [parse_chat_output()](https://github.com/vllm-project/vllm/blob/ab2eb27b74da0e6846749c11a68179d27a78d963/vllm/entrypoints/harmony_utils.py#L516) is not duplicated here. Actually, most of the parsing is already done there and repeated here. Leaving this refactoring for a separate PR.

## Test Plan
- Manually tested with chat requests with limited max_tokens while the parser is enabled.
- Added test cases.

## Test Result
`max_completion_tokens=5`, stopping in reasoning
```
--- Request: Tell me something about confidential computing. ---
(Need to)
```

`max_completion_tokens=50`, stopping in final message
```
--- Request: Tell me something about confidential computing. ---
(Need explain concept.)

### What is Confidential Computing?

Confidential computing is a set of hardware‑enforced
```

### DIFF
--- a/tests/tool_parsers/test_openai_tool_parser.py
+++ b/tests/tool_parsers/test_openai_tool_parser.py
@@ -9,6 +9,7 @@ from openai_harmony import (
     DeveloperContent,
     HarmonyEncodingName,
     Message,
+    RenderConversationConfig,
     Role,
     SystemContent,
     load_harmony_encoding,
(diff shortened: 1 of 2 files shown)


bbrowning:
Do you know if this is also an issue in the streaming tool calling case? Or whether we need a potentially separate fix there?

yeqcharlotte (changes requested):
hey @dr75 thanks for the fixes. wonder if you could also double confirm the behavior with the official oai implementation? will they return partial final messages when limit is reached? also please fix the precommit.

dr75:
@yeqcharlotte the CI failure was unrelated. I rebased, seems fine now. 

Will try with the OpenAI API to confirm the behaviour.

PR #28133, opened 2025-11-05
[Mamba] - Consolidate Mambas Attention Logic

Josephasafg:
## Purpose
As of today, a lot of the same logic was duplicated between all of the Mamba attention classes (Mamba1, Mamba2, ShortConv), this was hard to maintain and usually consisted of more than a single PR per model.

This PR's goal is to move all of the common logic to a single class, keeping the implementation classes (e.g. `Mamba1AttentionMetadataBuilder`, `ShortConvAttentionMetadataBuilder` pretty empty, leaving room for future differences and per model features) and with `Mamba2AttentionMetadataBuilder` having to most differences between the base class and the rest.

## Test Result

---
<details>

</details>


Josephasafg:
@tdoublep PTAL. Thanks!

vllm/v1/attention/backends/mamba_attn.py
```diff
@@ -2,27 +2,61 @@
...
+    num_prefill_tokens: int
+    num_decodes: int
+    num_decode_tokens: int
+
+    # The following tensors only contain prefill requests and will be None if
+    # the batch has no prefill request.
+    has_initial_states_p: torch.Tensor | None
+
+    query_start_loc_p: torch.Tensor | None
+    state_indices_tensor: torch.Tensor
```
tdoublep, line 40: `state_indices_tensor` is used for both prefill and decode right? It might be good to group `has_initial_states_p` and `query_start_loc_p` together based on the comment above (e.g, remove the newline).

vllm/v1/attention/backends/mamba_attn.py
```diff
@@ -2,27 +2,61 @@
...
+
+    block_idx_last_scheduled_token: torch.Tensor | None
+    block_idx_first_scheduled_token_p: torch.Tensor | None
+    block_idx_last_computed_token: torch.Tensor | None
+    num_computed_tokens_p: torch.Tensor | None
+
+    num_reqs: int
```
tdoublep: Should we group this with the other `int` above?

vllm/v1/attention/backends/mamba_attn.py
```diff
@@ -2,27 +2,61 @@
...
+    # The following tensors only contain prefill requests and will be None if
+    # the batch has no prefill request.
+    has_initial_states_p: torch.Tensor | None
+
+    query_start_loc_p: torch.Tensor | None
+    state_indices_tensor: torch.Tensor
+
+    block_idx_last_scheduled_token: torch.Tensor | None
+    block_idx_first_scheduled_token_p: torch.Tensor | None
+    block_idx_last_computed_token: torch.Tensor | None
```
tdoublep, line 45: Should we add a note that these tensors are used for prefix caching (and None if prefix caching is disabled)?

vllm/v1/attention/backends/mamba2_attn.py
```diff
@@ -239,137 +223,33 @@ def build(
...
-
-        attn_metadata = Mamba2AttentionMetadata(
-            num_prefills=num_prefills,
-            num_prefill_tokens=num_prefill_tokens,
-            num_decodes=num_decodes,
-            num_decode_tokens=num_decode_tokens,
-            query_start_loc_p=query_start_loc_p,
-            seq_lens=seq_lens,
+        return replace(
+            common,
```
tdoublep, line 249: Won't `common` have type `BaseMambaAttentionMetadata` or is the type hint of `M` that the child class override is enough?

tdoublep (commented):
Awesome work, this will make things much more maintainable. I have a few minor suggestions/questions.

vllm/v1/attention/backends/mamba_attn.py
Josephasafg: Makes sense. Done

Josephasafg: I also added `num_computed_tokens_p` to this group

vllm/v1/attention/backends/mamba_attn.py
Josephasafg: Sounds good

vllm/v1/attention/backends/mamba_attn.py
Josephasafg: Good point. Done

vllm/v1/attention/backends/mamba2_attn.py
Josephasafg: Yes, the type hint of `M` that the child class overrides is enough. `common` will have type `Mamba2AttentionMetadata` and not just the base class - `BaseMambaAttentionMetadata`

tdoublep (approved):
LGTM

PR #24954, opened 2025-09-16
[Bug] Fix gpt-oss missing tool content

levunet:
## Purpose
This PR requires the following PRs to be merged first: https://github.com/vllm-project/vllm/pull/24768 and the harmony lib PR (https://github.com/openai/harmony/pull/76).

The changes include adding 'with_recipient' and the Assistant's 'analysis' content.
Without adding this content, there was an issue where the gpt-oss model had a higher probability of outputting abnormal tokens when calling tools.

## Test Plan
[gpt-oss_test.py](https://github.com/user-attachments/files/22359291/gpt-oss_test.py)
[messages.txt](https://github.com/user-attachments/files/22359292/messages.txt)

Run python3 gpt-oss_test.py about 10 times.

## Test Result 
(before)
image

(after applying the harmony lib changes)
image

PR #28139, opened 2025-11-05
[Frontend][Bug] allow tool calls in analysis channel

dr75:
## Purpose
The gpt-oss streaming parser does not handle tool calls in the analysis channel. However, those calls can happen as explained [here](https://cookbook.openai.com/articles/openai-harmony#channels).

This especially happens when harmony renders __the recipient__ in previous tool calls ~~in the analysis channel~~ __before the channel token__, ~~convincing~~ "confusing" the model which returns tool calls in the analysis channel. However, the harmony parser is built to handle both cases, such that the vLLM parser should also support both.

The fix proposed [in this PR](https://github.com/vllm-project/vllm/pull/24954) aims at changing the message rendering in harmony (for converting the history to tokens) to avoid confusing the model. While this fixes the issue it seems to not address the root cause: even with such a change, the model may emit tool calls in the analysis channel as described in the OpenAI docs.

To solve this, I propose to handle also tool calls in the analysis channel.

To review changes, you may start at the first commit, which is the actual fix. Following commits are refactoring and testing changes as described below.

## Test Plan
- The previous state of the code did not allow for unit testing. I added a temporary automated test using an actual request that reliably fails without the fix. This requires a large context window and reducing the request size doesn't work. As this does not work for CI, I only use it to validate the fix and removed it in the last commit. See the respective commit for the test and checkout/run for validation.
- Extracted the stream parsing from `serving_chat.py‎` to a new file `serving_chat_stream_harmony.py` and added a unit test for testing this part in isolation with a test case for testing tools in the analysis channel.
- Manually tested with repeated tool calls, which are causing issues.
- Also tested with the script provided in https://github.com/vllm-project/vllm/pull/24954.


dr75:
See the OpenAI docs and the limited docs and comments in the harmony repo. The problem is that you cannot force the model to put the recipient where it does, except you would use some guided decoding. For that reason, harmony allows for both.

While usually it appears on the commentary channel, rendering previous tool calls to the analysis channel (which is the choice the harmony lib made - see comments) causes the model to switch as well. But in any case it wouldn't be a solution for every case as internal tools seem to be emitted on the analysis channel (see open ai docs).

As long as you don't get a different answer from OpenAI on the respective Harmony PR, this fix here seems to be the right solution to me. And I guess other frameworks are doing it the same way. But I didn't check.

alecsolder:
BTW I completely agree this is a problem in general, I just want to make sure we solve it the right way. 

> While usually it appears on the commentary channel, rendering previous tool calls to the analysis channel (which is the choice the harmony lib made - see comments) causes the model to switch as well.

Hrm I'm a bit confused, on the _input_ path, [the channel a tool call is rendered onto is decided by our code, which should always be commentary ATM](https://github.com/vllm-project/vllm/blob/ffb08379d8870a1a81ba82b72797f196838d0c86/vllm/entrypoints/harmony_utils.py#L254).

The more nefarious thing here is that any message to the `analysis` channel that comes before a message to the `final` channel is [actually completely dropped by the harmony renderer](https://github.com/openai/harmony/blob/ec7606df9e87e3d0a1fec9f50928c1e407f0c438/src/encoding.rs#L217). 

I think for now the most reasonable way to do it is that on the input path, if the tool name starts with `functions` then it should be on the commentary channel, if it doesn't then it should go onto the `analysis` channel. 

I think if recipient is set on the harmony message on the output path, it is fair to treat it as a tool call no matter what, doesn't matter which channel it is on for sure though

dr75:
Ah, thanks @levunet, that might explain things more: the built-in tools are in the analysis channel (as per docs) and usually done with tool invocation before channel as you found out. If now harmony renders all tool calls that way without distinguishing, then the model may also switch to invocation before channel and because of that then also switches to emitting them in the analysis channel.

That would mean the ideal tool rendering (for the current model) would also distinguish between the channel and place the recipient accordingly and that way keep the output from the model as is (avoiding prefix cache cut-off for that message) and avoid the model confusion.

Probably not a very import fix but could be useful. Given that it only affects tool calls, the impact on model performance might be very small or not there.

In vLLM we might then also make this distinction and generate analysis messages for the internal tools (being all commentary atm I assume).

dr75:
> @dr75 There seem to be some merge conflicts! Also is this ready for review?

Hey @bbartels! I resolved the conflicts and made some changes:
- refactor stream parsing: extract harmony stream parsing into its own file
- added a unit test for it
- removed the long test that would require a large context window and doesn't add too much value; I kept it in the commit history for the review.

So its ready for review.

alecsolder (approved):
Looks to me like the change here is to just allow the other channel for tool calls, otherwise the changes are adding tests and moving some logic. Seems good to me!

qandrew (approved):

yeqcharlotte (approved):

qandrew:
@dr75 could you fix the CI issues? thanks!

PR #28141, opened 2025-11-05
[Core][MM] Use non-blocking CPU-GPU copy of multimodal data

lgeiger:
## Purpose

On `main` `group_mm_kwargs_by_modality` currently causes unnecessary CPU/GPU sync. This PR makes the copy non blocking to prevent this. Looks like this was missed in #25654 /cc @DarkLight1337

## Test Plan

```
 VLLM_TORCH_PROFILER_DIR="vllm_profile" vllm serve Qwen/Qwen3-VL-30B-A3B-Instruct-FP8 --limit-mm-per-prompt.video 0 --max-model-len 10000

vllm bench serve --backend openai-chat --model Qwen/Qwen3-VL-30B-A3B-Instruct-FP8 --endpoint /v1/chat/completions --dataset-name hf --dataset-path lmarena-ai/VisionArena-Chat --hf-split train --num-prompts 5 --profile
```

## Test Result

**Before:**
Screenshot 2025-11-05 at 17 04 11

**After:**

Screenshot 2025-11-05 at 17 04 40


ywang96 (approved):

DarkLight1337 (approved):
Good catch

PR #28144, opened 2025-11-05
[Log] Save profiler results to file instead of stdout

rasmith:
This PR saves profiler results to VLLM_TORCH_PROFILER_DIR instead of stdout.  Printing the profiling results to stdout makes it difficult to read the results, especially with so much output already being generated by vLLM.  Furthemore, if someone wants to save the profiler output to a file right now, they will have to save it to a log or copy-paste it and then format after the fact to remove all of the line prefixes.

Instead, each rank can save this data to an output file along with the traces that are already currently saved from all ranks anyways.  Having a summary along with the traces from each rank can be handy.  Note that the profiler output comparison between two ranks can differ when using larger models that need multiple GPUs.


vllm/v1/worker/gpu_worker.py
```diff
@@ -576,11 +576,13 @@ def profile(self, is_start: bool = True):
             self.profiler.start()
         else:
             self.profiler.stop()
-            # only print profiler results on rank 0
-            if self.local_rank == 0:
```
SageMoore: Nit: Can you leave in the print to stdout?

SageMoore (commented):
As someone who has spent too much time copy/pasting these outputs into files before scping traces onto my laptop, this is a welcome change. Thanks for the contribution, @rasmith.

vllm/v1/worker/gpu_worker.py
rasmith: Done

SageMoore (approved):

gshtras (approved):

PR #25515, opened 2025-09-23
[GPT-OSS] Structure_Tag support for gpt-oss tool-call in cot

Hanchenli:
## Purpose
The model sometimes generates function calls in CoT that have not been provided in tool_server. This PR uses structural tag to force model to 1. do not generate function calls when tool_server is not provided. 2. Only generate function that are present in the tool_server. 

example custom structural tags: https://gist.github.com/frank-wei/4929118fe0881ee561ce48f15c1f80ed

## Test Plan
Testing with @frank-wei 

## Test Result

vllm serve run: https://gist.github.com/frank-wei/86d37061685fd45e5b831b6583d865d6

see number of turns reduced:
* without structural tags: https://gist.github.com/frank-wei/3c89e45e028cb0f61dc56c51f464252f
* with structural tags: https://gist.github.com/frank-wei/45daec7615f0c75ab3d97594746324c2
---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #28148, opened 2025-11-05
[Frontend] [gpt-oss] Chat format GD for tool calling with gptoss

alecsolder:
## Purpose
Building on top of the initial gpt-oss reasoning parser in #25515, this PR fleshes out the full structural tags schema to guide the chat format for gpt-oss.

It is only added to the Responses API path for now, but it should be able to fix issues like the ones mentioned in #24954 without needing to make adjustments to Harmony.

Guided decoding for the gpt-oss chat format should only be enabled if `structured_outputs_config.enable_in_reasoning` is True, since the chat format is technically in the reasoning section of model output. It also lets us continue to use structured outputs for the content of the final message. `structured_outputs_config.reasoning_parser` is set by default for gpt-oss. 

TODO: test that structured output works alongside this after #28000 is merged, and adjust for tool calling

## Test Plan
New unit tests and e2e tests with difficult tool names.

To enable GD of the chat format
```
vllm serve openai/gpt-oss-20b --structured-outputs-config='{"enable_in_reasoning": true}'
```

Tested with a curl like 
```
curl -X POST http://localhost:8000/v1/responses \
  -d '{
    "input": "Look up the weather roughly in san francisco.",
    "enable_response_messages": true,
    "tools": [{
      "type": "function",
      "name": "get_weather",
      "description": "Get current temperature for provided coordinates in celsius.",
      "parameters": {
        "type": "object",
        "properties": {
          "latitude": {"type": "number"},
          "longitude": {"type": "number"}
        },
        "required": ["latitude", "longitude"],
        "additionalProperties": false
      },
      "strict": true
    }]
  }'
```

And confirming it still works with built in tools
```
vllm serve openai/gpt-oss-20b --structured-outputs-config='{"enable_in_reasoning": true}' --tool-server=localhost:8081/browser,localhost:8081/python
```
With a curl like 
```
curl -X POST http://localhost:8000/v1/responses \ 
  -d '{
    "input": "Search for vLLM performance.",
    "tools": [{"type": "web_search_preview"}]
  }'
```
## Test Result
Valid output for both cases

---
<details>

</details>

### DIFF
--- a/vllm/reasoning/gptoss_reasoning_parser.py
+++ b/vllm/reasoning/gptoss_reasoning_parser.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+import copy
 import json
 from collections.abc import Sequence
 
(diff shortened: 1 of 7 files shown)


vllm/reasoning/gptoss_reasoning_parser.py
```diff
@@ -1,67 +1,155 @@
...
 from vllm.entrypoints.harmony_utils import parse_chat_output
 from vllm.entrypoints.openai.protocol import ChatCompletionRequest, DeltaMessage
-from vllm.entrypoints.tool_server import ToolServer
 from vllm.logger import init_logger
 from vllm.reasoning import ReasoningParser, ReasoningParserManager
 
 logger = init_logger(__name__)
 
-no_func_reaonsing_tag = {
+TRIGGERS = ["<|channel|>", "<|start|>assistant"]
```
yeqcharlotte, line 16: wonder if we can define this as some sort of yaml or json files but we enabled default values for these tags. this allows people to modify their template without changing vllm's binary.

yeqcharlotte:
could you also summarize the main behavior change? we were missing following structural tag following in final messages?

vllm/reasoning/gptoss_reasoning_parser.py
Hanchenli: Agreed. Maybe it is best that we have a default_template and load it in here?

frank-wei: I think this could be the default template but only if it is at least neutral to general eval tests. Otherwise, people may question about it.
Also, double down on the suggestion from Charlotte if we have the flexibility of passing a json file.

Issue #27442, opened 2025-10-23
[CI Failure][AMD] Encoder-Decoder Models Fail on AMD CI


zhewenl:
### Name of failing test

entrypoints/openai/test_transcription_validation.py::test_basic_audio[openai/whisper-large-v3-turbo]

### Basic information

- [ ] Flaky test
- [x] Can reproduce locally
- [ ] Caused by external libraries (e.g. bug in `transformers`)

## Summary

Encoder-decoder models (Whisper, T5, BART, vision-language models) fail on AMD ROCm with `NotImplementedError` because **all ROCm-specific attention backends only support decoder-only models**. 

Therefore, all the tests using these models will fail on AMD CI([example](https://buildkite.com/vllm/ci/builds/35670#019a04ed-a2f7-4e7e-85ad-4e2118f76898)), mainly in `Entrypoints Integration Test (API Server)` and `Entrypoints Integration Test (Pooling)`.

### Affected Models
- **Speech-to-Text**: Whisper family (`openai/whisper-large-v3-turbo`, `mistralai/Voxtral-Mini-3B-2507`)
- **Vision-Language**: `microsoft/Phi-3.5-vision-instruct`
- **Translation**: T5, BART, MarianMT models
- **Any model with cross-attention**: Encoder-decoder architectures
### Failure Mode

**Error**:
```
NotImplementedError: Encoder self-attention and encoder/decoder cross-attention
are not implemented for TritonAttentionImpl
```

### Root Cause: All ROCm-Specific Backends are Decoder-Only
| Backend | Location |
|---------|----------|
| **TritonAttentionBackend**(default) | `vllm/v1/attention/backends/triton_attn.py` | 
| **RocmAttentionBackend** | `vllm/v1/attention/backends/rocm_attn.py`  |
| **RocmAiterFABackend** | `vllm/v1/attention/backends/rocm_aiter_fa.py`  |
| **RocmAiterMLABackend** | `vllm/v1/attention/backends/mla/rocm_aiter_mla.py` |

eg. TritonAttentionImpl:
```python
if attn_type != AttentionType.DECODER:
    raise NotImplementedError(
        "Encoder self-attention and "
        "encoder/decoder cross-attention "
        "are not implemented for "
        "TritonAttentionImpl"
    )
```

### Purposal 
1. [short-term] Mark encoder-decoder tests with `@pytest.encoder_decoder`, and skip them when using AMD.
[...]

PR #28156, opened 2025-11-05
[CI/Build] Skip encoder-decoder models on AMD

zhewenl:
## Purpose
See more details in https://github.com/vllm-project/vllm/issues/27442. Encoder-decoder models (Whisper, T5, BART, vision-language models) fail on AMD ROCm with NotImplementedError because all ROCm-specific attention backends only support decoder-only models.

Therefore, all the tests using these models will fail on AMD CI([example](https://buildkite.com/vllm/ci/builds/35670#019a04ed-a2f7-4e7e-85ad-4e2118f76898)), mainly in Entrypoints Integration Test (API Server) and Entrypoints Integration Test (Pooling):
```python
NotImplementedError: Encoder self-attention and encoder/decoder cross-attention
are not implemented for TritonAttentionImpl
```

In this PR:
- Tests that contain **BOTH** encoder and decoder models use `@pytest.mark.encoder_decoder` on specific tests/variants; also fixed some tests not using parametrized fixtures.
- Tests that contain **ONLY** encoder models are skipped entirely at the CI level in `test-amd.yaml`.
- Ad-hoc patches tests can not skip/add decorator to skip easily by patching the simple logic to skip on ROCm platform.

## Test Plan

CI: https://buildkite.com/vllm/amd-ci/builds/1020

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1250,13 +1250,41 @@ def pytest_addoption(parser):
 
 
 def pytest_collection_modifyitems(config, items):
-    if config.getoption("--optional"):
-        # --optional given in cli: do not skip optional tests
-        return
-    skip_optional = pytest.mark.skip(reason="need --optional option to run")
-    for item in items:
-        if "optional" in item.keywords:
-            item.add_marker(skip_optional)
+    if not config.getoption("--optional"):
+        skip_optional = pytest.mark.skip(reason="need --optional option to run")
+        for item in items:
+            if "optional" in item.keywords:
+                item.add_marker(skip_optional)
+
+    from vllm.platforms import current_platform
+
+    # Skip encoder-decoder models on ROCm: https://github.com/vllm-project/vllm/issues/27442
+    ENCODER_DECODER_MODELS = [
+        # Encoder-decoder models
+        "openai/whisper-small",
+        "openai/whisper-large-v3-turbo",
+        "mistralai/Voxtral-Mini-3B-2507",
+        "microsoft/Phi-3.5-vision-instruct",
+        # Encoder-only models (cross-encoders, embedding models)
+        "cross-encoder/ms-marco-MiniLM-L-6-v2",
+        "intfloat/multilingual-e5-small",
+        "BAAI/bge-reranker-base",
+        "BAAI/bge-base-en-v1.5",
+        "TIGER-Lab/VLM2Vec-Full",
+        "sentence-transformers/all-MiniLM-L12-v2",
+        "sentence-transformers/stsb-roberta-base-v2",
+    ]
+
+    if current_platform.is_rocm():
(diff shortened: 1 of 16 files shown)


ywang96:
cc @russellb @robertgshaw2-redhat 

AFAIR we have reached an agreement that the only enc-dec model we're going to support in V1 is whisper, so we should deprecate other models.

tests/conftest.py
```diff
@@ -1250,13 +1250,45 @@ def pytest_addoption(parser):
...
+
+    from vllm.platforms import current_platform
+
+    # Skip encoder-decoder models on ROCm: https://github.com/vllm-project/vllm/issues/27442
+    ENCODER_DECODER_MODELS = [
+        # Encoder-decoder models
+        "openai/whisper-small",
+        "openai/whisper-large-v3-turbo",
+        "mistralai/Voxtral-Mini-3B-2507",
+        "microsoft/Phi-3.5-vision-instruct",
```
DarkLight1337: These two are not encoder-decoder models.

tests/conftest.py
```diff
@@ -1250,13 +1250,45 @@ def pytest_addoption(parser):
...
+        # Encoder-only models (cross-encoders, embedding models)
+        "cross-encoder/ms-marco-MiniLM-L-6-v2",
+        "intfloat/e5-small",
+        "intfloat/multilingual-e5-small",
+        "BAAI/bge-reranker-base",
+        "BAAI/bge-reranker-v2-m3",
+        "BAAI/bge-base-en-v1.5",
+        "TIGER-Lab/VLM2Vec-Full",
```
DarkLight1337: This one is also not encoder-decoder model

tests/conftest.py
```diff
@@ -1250,13 +1250,45 @@ def pytest_addoption(parser):
...
+        skip_encoder_decoder = pytest.mark.skip(
+            reason="Encoder-decoder models not supported on ROCm (all ROCm attention backends only support decoder-only models)"
+        )
+        for item in items:
+            if "encoder_decoder" in item.keywords:
+                if any(
+                    encoder_model in item.nodeid
+                    for encoder_model in ENCODER_DECODER_MODELS
+                ):
+                    item.add_marker(skip_encoder_decoder)
```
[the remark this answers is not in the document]
zhewenl: It's true, I have refactored tests from unparameterized to parameterized, adding support with unparameterized tests is no simple solution

zhewenl:
@DarkLight1337 Thanks for catching this! Updated list and relevant tests

zhewenl:
We will hold this PR as https://github.com/vllm-project/vllm/pull/28346, https://github.com/vllm-project/vllm/pull/28376 attempts to fix encoder-decoder on ROCM

PR #28165, opened 2025-11-05
[Frontend] Change CompilationMode to a proper Enum

### LINKED ISSUE #28150: [Bug]: -O.mode=NONE (or -cc.mode=NONE) should work
### Your current environment

main

### 🐛 Describe the bug

Right now -O.mode only accepts integer levels. Ideally it would accept ints and the string.

`vllm serve -O.mode=NONE` # doesn't work
`vllm serve -O.mode=0` # does work

gmagogsfm:
- Using an Emum rather than an integer is safer
- Also enhances arg parser to support specifying CompilationMode via both name(string) or integers

Fixes [#28150](https://github.com/vllm-project/vllm/issues/28150)

Signed-off by: Yanan Cao <gmagogsfm@gmail.com>

### DIFF
--- a/vllm/config/compilation.py
+++ b/vllm/config/compilation.py
@@ -28,7 +28,7 @@ else:
 logger = init_logger(__name__)
 
 
-class CompilationMode:
+class CompilationMode(enum.IntEnum):
     """The compilation approach used for torch.compile-based compilation of the
     model."""
 
@@ -115,7 +115,7 @@ class PassConfig:
     """The threshold of the communicated tensor sizes under which
     vllm should use flashinfer fused allreduce. Specified as a
     float in MB.
-    Unspecified will fallback to default values 
+    Unspecified will fallback to default values
     which are compute capability and world size dependent.
         FI_ALLREDUCE_FUSION_MAX_SIZE_MB = {
             90: {
@@ -244,7 +244,7 @@ class CompilationConfig:
     Please use mode. Currently all levels are mapped to mode.
     """
     # Top-level Compilation control
-    mode: int | None = None
+    mode: CompilationMode | None = None
     """The compilation approach used for torch.compile-based compilation of the
     model.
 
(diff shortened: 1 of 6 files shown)


vllm/model_executor/layers/quantization/utils/w8a8_utils.py
```diff
@@ -419,7 +419,7 @@ def __init__(
         if pad_output is None:
             config = get_current_vllm_config().compilation_config
             pad_output = (
-                config.mode < CompilationMode.VLLM_COMPILE
+                config.mode.value < CompilationMode.VLLM_COMPILE.value
```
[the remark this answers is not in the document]
gmagogsfm: @morrison-turnansky I think you recently updated the code here. Could you confirm if the comment is actually stale, thus contradicting with current implementation?

vllm/config/compilation.py
```diff
@@ -28,7 +28,7 @@
 logger = init_logger(__name__)
 
 
-class CompilationMode:
+class CompilationMode(enum.Enum):
     """The compilation approach used for torch.compile-based compilation of the
     model."""
```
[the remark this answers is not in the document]
zou3519: Imo this is fine, comparisions between compile mode is a little spooky

vllm/config/compilation.py
```diff
@@ -521,6 +521,27 @@ def __repr__(self) -> str:
...
+    @classmethod
+    def validate_mode_before(cls, value: Any) -> Any:
+        """
+        Enable parsing the `mode` field from string mode names.
+        Accepts both integers (0-3) and string names, like NONE, STOCK_TORCH_COMPILE,
+        DYNAMO_TRACE_ONCE, VLLM_COMPILE.
+        """
+        if isinstance(value, str):
+            # Convert string mode name to integer value
+            mode_name = value.upper()
```
zou3519, line 592: why do upper?

zou3519: ah I see that's what we do for other things

vllm/model_executor/layers/quantization/utils/w8a8_utils.py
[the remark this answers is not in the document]
zou3519: pre-existing issue so I will ignore for the purposes of this review

zou3519 (approved):

zou3519:
@gmagogsfm could you attempt a rebase please? 

@hmellor please let us know if you have other suggestions for the handling. This PR is just copying some of the other handling we have already (e.g. for the cudagraph_mode)

ProExpertProg (approved):

ProExpertProg (changes requested):
Waiting for @hmellor's response to make sure his comments are addressed

hmellor (approved):
Thanks for checking.

One change I would make is to remove the `.value` from integer comparisons because `IntEnum` supports comparison of its members like `int`s

vllm/config/compilation.py
```diff
@@ -579,6 +579,27 @@ def __repr__(self) -> str:
...
+
+            if mode_name not in CompilationMode.__members__:
+                raise ValueError(
+                    f"Invalid compilation mode: {value}. "
+                    f"Valid modes are: {', '.join(CompilationMode.__members__.keys())}"
+                )
+
+            return CompilationMode[mode_name]
+        return value
```
ProExpertProg, line 601: Shouldn't we check that `value` is an `int` or `CompilationMode` instance here?

gmagogsfm: Good point, but I think it would be redundant with Pydantic's internal parsing and validation.

Here are some documentations references:
```
Pydantic provides multiple types of validator functions:

Before validators run before Pydantic's internal parsing and validation (e.g. coercion of a str to an int). These are more flexible than After validators since they can modify the raw input, but they also have to deal with the raw input, which in theory could be any arbitrary object.
```

How pydantic parses Enum:
https://docs.pydantic.dev/2.0/usage/types/enums/

ProExpertProg: TIL, thanks!

ProExpertProg (approved):

PR #28168, opened 2025-11-05
[Core][MM] Add mechanism to configure multimodal fields which should stay on CPU

lgeiger:
## Purpose

This is a followup on #28141 and adds a `multimodal_cpu_fields` to the `SupportsMultiModal` protocol. This allows model implementers to define multimodal fields which shouldn't be transferred to the GPU.

I updated the Qwen VL model definitions to set `image_grid_thw` and `video_grid_thw` as CPU-only fields. These fields are transferred to the GPU by `group_mm_kwargs_by_modality` and copied back to the CPU in the model forward pass via `.tolist()`. The latter causes a `cudaStreamSync` which can hurt performance. Keeping them on CPU also allows to cleanup the model code a bit and removes the need to convert the list back to tensors.

I'm not sure whether `SupportsMultiModal` is the best place to define this, but it didn't require large code changes and still is backwards compatible so I think it's sensible to keep it close to `merge_by_field_config`.

## Test Plan

```
vllm serve Qwen/Qwen3-VL-2B-Instruct-FP8 --limit-mm-per-prompt.video 0

vllm bench serve --backend openai-chat --model Qwen/Qwen3-VL-2B-Instruct-FP8 --endpoint /v1/chat/completions --dataset-name hf --dataset-path lmarena-ai/VisionArena-Chat --hf-split train --num-prompts 1000
```

## Test Result

I generated a torch profile and verified that no more `cudaStreamSync` ops are present.

This also seems to slightly improve throughput by allowing more computation to be overlapped. Measured on a single L40s GPU though the benefit might be larger for faster GPUs.

**Before:**

```
Successful requests:                     998
Failed requests:                         2
Benchmark duration (s):                  49.89
Total input tokens:                      94306
Total generated tokens:                  121122
Request throughput (req/s):              20.00
Output token throughput (tok/s):         2427.71
Peak output token throughput (tok/s):    14299.00
Peak concurrent requests:                998.00
Total Token throughput (tok/s):          4317.94
Mean TTFT (ms):                          22076.45
Median TTFT (ms):                        20806.25
P99 TTFT (ms):                           45564.97
Mean TPOT (ms):                          84.44
Median TPOT (ms):                        84.52
P99 TPOT (ms):                           126.30
Mean ITL (ms):                           84.63
Median ITL (ms):                         73.15
P99 ITL (ms):                            514.58
```

**After:**

```
Successful requests:                     998
Failed requests:                         2
Benchmark duration (s):                  49.28
Total input tokens:                      94302
Total generated tokens:                  121141
Request throughput (req/s):              20.25
Output token throughput (tok/s):         2458.19
Peak output token throughput (tok/s):    15255.00
Peak concurrent requests:                998.00
Total Token throughput (tok/s):          4371.77
Mean TTFT (ms):                          21838.22
Median TTFT (ms):                        20195.33
P99 TTFT (ms):                           44981.08
Mean TPOT (ms):                          83.06
Median TPOT (ms):                        84.76
P99 TPOT (ms):                           128.41
Mean ITL (ms):                           83.61
Median ITL (ms):                         73.28
P99 ITL (ms):                            511.00
```


vllm/model_executor/models/vision.py
```diff
@@ -429,7 +429,9 @@ def run_dp_sharded_mrope_vision_model(
             )
     else:
         if pixel_values_local.shape[0] > 0:
-            image_embeds_local = vision_model(pixel_values_local, local_grid_thw_list)
+            image_embeds_local = vision_model(
+                pixel_values_local, torch.tensor(local_grid_thw_list, dtype=torch.int32)
```
lgeiger: I'm unable to test DP at the moment. Let me know if you think this could cause any issue.
This is mainly to make all code paths consistent. I think it should be fine since a couple of lines earlier we already pass tensors to the vision model anyway.

lgeiger: Reverted the DP changes in 0d7ef2311ba437bbc3c5785999e3e883f7a0fbdc to be save and not break other models by accident.

vllm/model_executor/models/vision.py
```diff
@@ -429,7 +429,9 @@ def run_dp_sharded_mrope_vision_model(
...
+            )
```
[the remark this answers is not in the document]
lgeiger: Done in 0d7ef2311ba437bbc3c5785999e3e883f7a0fbdc

vllm/model_executor/models/interfaces.py
```diff
@@ -80,6 +80,11 @@ class SupportsMultiModal(Protocol):
     `vllm.multimodal.utils.group_mm_kwargs_by_modality` to use.
     """
 
+    multimodal_cpu_fields: ClassVar[set[str] | None] = None
```
DarkLight1337: Just use empty set as the default for simplicity

DarkLight1337:
I am ok with putting this in the model definition.

vllm/model_executor/models/interfaces.py
lgeiger: I used `None` here on purpose since a set is mutable so multiple subclasses could mutate the same value if I'm not mistaken. Or is this not an issue here?

DarkLight1337: Maybe you can type hint this as `typing.Set` which is an immutable interface and set the default to an empty frozenset

lgeiger: Done in beba1f4fa4ce0a9b6c036faad1a4a6cc9e53097b

DarkLight1337 (approved):

PR #28169, opened 2025-11-05
[torchao] fix safetensors for sharding

liangel-02:
**Summary** 

Previously, our safetensors implementation for tensor subclasses didn't consider the possibility of different tensor attributes for one instance of a tensor subclass (ie `qdata` or `scale` for `Float8Tensor`) being sharded to different files.

This PR handles that case by: 

1. Defining the state dict outside of the loop that iterates through shard files 
2. After processing a full tensor subclass in `unflatten_tensor_state_dict`, we delete the corresponding tensor attributes (ie `qdata`, `scale`) from the original state dict passed in 
3. Whatever is remaining is saved in the state dict for the next iteration 

**Testing**
`pytest tests/quantization/test_torchao.py::test_safetensors_model_loading_with_params -s`

### DIFF
--- a/tests/quantization/test_torchao.py
+++ b/tests/quantization/test_torchao.py
@@ -225,13 +225,12 @@ def test_reload_weights():
 @pytest.mark.skip(
     reason="since torchao nightly is only compatible with torch nightly"
     "currently https://github.com/pytorch/ao/issues/2919, we'll have to skip "
-    "torchao tests that requires newer versions (0.14.0.dev+) for now"
+    "torchao tests that requires newer versions (0.15.0.dev+) for now"
 )
-def test_opt_125m_float8_weight_only_safetensors_model_loading_with_params(vllm_runner):
+def test_safetensors_model_loading_with_params(vllm_runner):
     torch._dynamo.reset()
-    model_name = (
-        "torchao-testing/opt-125m-Float8WeightOnlyConfig-v2-0.14.0.dev-safetensors"
-    )
+    # using this model to test safetensors loading with file sharding
+    model_name = "torchao-testing/Qwen3-8B-INT4-0.15.0dev-safetensors"
     with vllm_runner(model_name=model_name, dtype="bfloat16") as llm:
         output = llm.generate_greedy(["The capital of France is"], max_tokens=4)
 
--- a/vllm/model_executor/model_loader/weight_utils.py
+++ b/vllm/model_executor/model_loader/weight_utils.py
@@ -595,6 +595,8 @@ def safetensors_weights_iterator(
     if safetensors_load_strategy == "eager":
         loading_desc += " (eager)"
 
+    state_dict = {}
+
     for st_file in tqdm(
         hf_weights_files,
         desc=loading_desc,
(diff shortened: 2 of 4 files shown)


tests/quantization/test_torchao.py
```diff
@@ -227,11 +227,9 @@ def test_reload_weights():
     "currently https://github.com/pytorch/ao/issues/2919, we'll have to skip "
     "torchao tests that requires newer versions (0.14.0.dev+) for now"
 )
-def test_opt_125m_float8_weight_only_safetensors_model_loading_with_params(vllm_runner):
+def test_safetensors_model_loading_with_params(vllm_runner):
     torch._dynamo.reset()
-    model_name = (
-        "torchao-testing/opt-125m-Float8WeightOnlyConfig-v2-0.14.0.dev-safetensors"
-    )
+    model_name = "torchao-testing/Qwen3-8B-INT4-0.15.0dev-safetensors"
```
heheda12345, line 233: Why are you using this model? If it is for "different tensor attributes for one instance of a tensor subclass (ie qdata or scale for Float8Tensor) being sharded to different files.", please add a comment.

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -605,21 +607,23 @@ def safetensors_weights_iterator(
                 state_dict = load(f.read())
             yield from state_dict.items()
         elif safetensors_load_strategy == "torchao":
-            if not torchao_version_at_least("0.14.0"):
+            is_last_file = st_file == hf_weights_files[-1]
+            if not torchao_version_at_least("0.15.0"):
```
heheda12345, line 614: why are you upgrading to 0.15.0?

tests/quantization/test_torchao.py
heheda12345, line 225: do we need to skip now that torch2.9 is released?

vllm/model_executor/model_loader/weight_utils.py
liangel-02: we've since made some changes to the unflatten API in torchao, so the version needs to be bumped. with 0.14.0 the logic below will error

tests/quantization/test_torchao.py
liangel-02: i think we still need to skip since this test depends on changes in torchao main cc @jerryzh168

jerryzh168: yeah we need a separate update to upgrade the torchao version used in testing in vllm: https://github.com/vllm-project/vllm/blob/da786e339e985cbf32f43474e7b88b8da65c76e5/.buildkite/test-amd.yaml#L616

and make sure the tests are passing

heheda12345 (approved):
LGTM!

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -605,22 +607,23 @@ def safetensors_weights_iterator(
...
             )
 
             with safe_open(st_file, framework="pt") as f:
-                state_dict = {}
                 for name in f.keys():  # noqa: SIM118
                     state_dict[name] = f.get_tensor(name)
                 metadata = f.metadata()
-                updated_state_dict = unflatten_tensor_state_dict(state_dict, metadata)
-            yield from updated_state_dict.items()
+                unflattened_state_dict, state_dict = unflatten_tensor_state_dict(
```
jerryzh168: we should add a comment to talk about what is happening here I think, around (1) sharded checkpoint and the issue for torchao tensor (2) the solution: the the unmatched tensors are left in state_dict and we expect it to be handled by next iteration in conjunction with other tensors that's loaded from the sharded checkpoint.

jerryzh168: probably should also explain that this `state_dict` is reused in the future iteration, since we didn't reassign the state_dict in a new iteration, but just add new loaded tensors into the state_dict (in L623)

since it's not common for people to reuse a variable across iterations like this I feel

jerryzh168: or maybe using a different variable here would be clearer

PR #28171, opened 2025-11-06
[Perf] Introduce FlattenLogprobs to store logprobs results to reduce GC overhead

Jialin:
## Purpose
Originally, Logprobs are stored as list[dict[int, Logprob]], which means for each requests with logprobs enabled, it would create a lot of objects collected by GC
- 1 list for each request
- <sequence_length> dictionaries for each position
- <sequence_length>*<top_logprob> Logprob objects to store metadata

By default, GC0 would be triggered every 700 objects allocation, GC1 would be triggered after every 10 GC0, GC2 would be triggered after every 10 GC1. For jobs that rely on logprobs, GC costs could be extremely high!

### Data Structure Change: list[dict[int, Logprob]] -> FlattenLogprobs
In this PR, we're introducing a replacement of the list[dict[int, Logprob]], which flatten Logprob into a constant number of primitive type lists (i.e. FlattenLogprobs):
- start_indices_per_position: The start indices of logprobs for each position.
- logprobs, token_ids, ranks, decoded_tokens: Flatten lists of logprob metadata

With FlattenLogprobs, each requests would only introduce 5 objects for GC, which is significantly smaller than <sequence_length>*<top_logprob>+<sequence_length>+1 objects in the original list[dict] approach.

### API signature changes
The whole feature would be gated behind an environment variable VLLM_FLATTEN_LOGPROBS, and by default, users will still received the logprobs with the original types, but the signature would be changed:
- PromptLogprobs changed from `list[dict[int, Logprob] | None]` to `FlattenLogprobs | list[dict[int, Logprob] | None]`
- SampleLogprobs changed from `list[dict[int, Logprob]]` to `FlattenLogprobs | list[dict[int, Logprob]]`

And we're introducing utils functions to unify the logprobs handling
```
def num_positions(logprobs: PromptLogprobs | SampleLogprobs) -> int:
def num_logprobs_per_position(logprobs: PromptLogprobs | SampleLogprobs, position: int) -> int:
def get_logprob(logprobs: PromptLogprobs | SampleLogprobs, position: int, token_id: int) -> int | None:
def get_rank(logprobs: PromptLogprobs | SampleLogprobs, position: int, token_id: int) -> int | None:
```
So ideally, users could firstly switch to use these helper functions, and then safely toggle VLLM_FLATTEN_LOGPROBS to adopt the optimizations.

## Test Plan & Test Result
**Unit test**
Logprob E2E Test Enhanced and Updated
```
HF_HUB_DISABLE_XET=1 pytest -s tests/samplers/test_logprobs.py && \
  pytest -s tests/test_logprobs.py
```

**Benchmark**
We could see with FlattenLogprob logprob processing is much faster and the GC cost is significantly smaller as well!
```
> python benchmarks/benchmark_logprobs_gc.py
+---------+-----------------+--------------+---------------+--------------+---------------+--------------+---------------+---------------+
|   Batch | Logprob Type    |   GC0 Invoke |   GC0 Runtime |   GC1 Invoke |   GC1 Runtime |   GC2 Invoke |   GC2 Runtime |   E2E Runtime |
|    size |                 |          cnt |     total (s) |          cnt |     total (s) |          cnt |     total (s) |     total (s) |
+=========+=================+==============+===============+==============+===============+==============+===============+===============+
|      10 | list[dict]      |         1175 |         0.018 |          106 |         0.022 |            9 |         0.004 |         0.731 |
+---------+-----------------+--------------+---------------+--------------+---------------+--------------+---------------+---------------+
|      10 | FlattenLogprobs |            0 |         0.000 |            0 |         0.000 |            0 |         0.000 |         0.409 |
+---------+-----------------+--------------+---------------+--------------+---------------+--------------+---------------+---------------+
|     100 | list[dict]      |        11789 |         0.181 |         1071 |         0.230 |           71 |         2.129 |         9.555 |
+---------+-----------------+--------------+---------------+--------------+---------------+--------------+---------------+---------------+
|     100 | FlattenLogprobs |            2 |         0.010 |            0 |         0.000 |            0 |         0.000 |         4.364 |
+---------+-----------------+--------------+---------------+--------------+---------------+--------------+---------------+---------------+
|    1000 | list[dict]      |       118468 |         1.830 |        10769 |         2.271 |          104 |        29.179 |       105.187 |
+---------+-----------------+--------------+---------------+--------------+---------------+--------------+---------------+---------------+
|    1000 | FlattenLogprobs |          165 |         0.089 |           15 |         0.400 |            0 |         0.000 |        44.768 |
+---------+-----------------+--------------+---------------+--------------+---------------+--------------+---------------+---------------+
```

**E2E**
We've confirmed that after the change, Logprobs GC objects are eliminated from GC cost.
*Before*
Screenshot 2025-11-05 at 3 55 47 PM

---
<details>

</details>


vllm/logprobs.py
```diff
@@ -21,8 +25,190 @@ class Logprob:
...
+
+        # Flatten Logprob information for (each position, rank).
+        # For position <i>, the logprobs are ranged
+        # from self.start_indices_per_position[i]
+        # to self.start_indices_per_position[i+1] (exclusive).
+        self.logprobs: list[float] = []
+        self.token_ids: list[int] = []
+        self.ranks: list[int] = []
+        self.decoded_tokens: list[str | None] = []
+
```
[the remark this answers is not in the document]
Jialin: num_positions is removed in the latest version.

vllm/logprobs.py
```diff
@@ -21,8 +25,190 @@ class Logprob:
...
+    logprobs: PromptLogprobs | SampleLogprobs, position: int
+) -> int:
+    """Gets the number of logprobs of a given position"""
+    if isinstance(logprobs, FlattenLogprobs):
+        return logprobs.num_logprobs_per_position(position)
+    return len(logprobs[position]) if logprobs[position] is not None else 0
+
+
+def get_logprob(
+    logprobs: PromptLogprobs | SampleLogprobs, position: int, token_id: int
```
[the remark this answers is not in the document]
Jialin: get_logprob is removed in the latest version.

vllm/logprobs.py
```diff
@@ -21,8 +25,190 @@ class Logprob:
...
+        for token_id, logprob, rank, token in zip(
+            logprob_token_ids, logprobs, ranks, decoded_tokens
+        ):
+            request_logprobs.logprobs.append(logprob)
+            request_logprobs.token_ids.append(token_id)
+            request_logprobs.ranks.append(rank)
+            request_logprobs.decoded_tokens.append(token)
+
+        request_logprobs.start_indices_per_position.append(
+            len(request_logprobs.logprobs)
```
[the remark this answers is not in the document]
Jialin: The latest version should resolve this issue, as it acted more like the list[dict].

vllm/logprobs.py
```diff
@@ -21,8 +25,189 @@ class Logprob:
...
+        self.start_indices_per_position: list[int] = [0, 0] if for_prompt else [0]
+
+        # Flatten Logprob information for (each position, rank).
+        # For position <i>, the logprobs are ranged
+        # from self.start_indices_per_position[i]
+        # to self.start_indices_per_position[i+1] (exclusive).
+        self.logprobs: list[float] = []
+        self.token_ids: list[int] = []
+        self.ranks: list[int] = []
+        self.decoded_tokens: list[str | None] = []
```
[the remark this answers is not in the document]
Jialin: The updated version should contain the list[dict] signature.

vllm/logprobs.py
```diff
@@ -21,8 +25,123 @@ class Logprob:
     decoded_token: str | None = None
 
 
+LogprobsOnePosition = dict[int, Logprob]
+
+
+# TODO(Jialin): Add more unit tests for the class and util functions
+@dataclass
+class FlattenLogprobs(Sequence[LogprobsOnePosition]):
```
yeqcharlotte: you may wanna implement MutableSequence interface so it actually functions like a list

vllm/logprobs.py
```diff
@@ -21,8 +25,123 @@ class Logprob:
...
+        """Extracts logprobs of a given position or slice"""
+        if isinstance(index, int):
+            return {
+                self.token_ids[i]: Logprob(
+                    logprob=self.logprobs[i],
+                    rank=self.ranks[i],
+                    decoded_token=self.decoded_tokens[i],
+                )
+                for i in range(self.start_indices[index], self.end_indices[index])
+            }
```
yeqcharlotte, line 122: will this part create too many dicts?

Jialin: It depends on how does the client consumed the FlattenLogprobs. I believe the most commonly used scenarios are:
- get a logprob dict of a position, then process the logprobs dict
- iterate through all positions

For both cases, it's creating <position> number of dict, but the lifecycle of each dict should be very short. So it should not involve that much GC costs compared to the original approach (as all lifecycles of dicts are much longer).

But I totally understand your concern. One option we could do is to introduce another FlattenLogProbsOnePosition class to behave as a dict[int, Logprob]. Then each `__getitem__` and iteration would only create a FlattenLogProbsOnePosition which would only contain a reference to FlattenLogProbs and a position index. WDYT? @yeqcharlotte

vllm/logprobs.py
Jialin: Updated, and add unit tests for FlattenLogprobs as well.

vllm/logprobs.py
```diff
@@ -21,8 +25,131 @@ class Logprob:
...
+
+    NOTE: To reduce the migration overhead and improve backward compatibility,
+    we support the key Sequence APIs of list, so it could act as
+    list[LogprobsOnePosition]
+    """
+
+    # Start / end indices to indicate the range of logprobs for each position.
+    # NOTE: There's no logprob of first prompt token, so if the
+    # FlattenLogprobs is for prompt, we directly set
+    # the start index for the second prompt token.
```
zhuohan123: You mean `start_indices[0]` will be second token's logprob range here?

vllm/logprobs.py
```diff
@@ -21,8 +25,131 @@ class Logprob:
...
+            raise TypeError(f"Invalid index type: {type(index)}")
+
+    def __setitem__(self, item, value) -> None:
+        raise TypeError("Cannot set logprobs in FlattenLogprobs")
+
+    def __delitem__(self, item) -> None:
+        raise TypeError("Cannot delete logprobs from FlattenLogprobs")
+
+    def insert(self, item) -> None:
+        raise TypeError("Cannot insert logprobs to FlattenLogprobs")
```
zhuohan123, line 146: Dumb question, do we have to write this part out or we can just simply remove them?

zhuohan123 (approved):
Thanks for the work! LGTM

vllm/logprobs.py
Jialin: > Dumb question, do we have to write this part out or we can just simply remove them?

It's required, otherwise whenever we reference FlattenLogprobs, python interpreter would complain about missing override methods. :/

vllm/logprobs.py
Jialin: Good catch, that's an out-dated code comment. Removed it, and the special handling for the first prompt token is in  create_prompt_logprobs()
Screenshot 2025-11-06 at 5 09 40 PM

vllm/logprobs.py
```diff
@@ -21,8 +26,183 @@ class Logprob:
     decoded_token: str | None = None
 
 
+LogprobsOnePosition = dict[int, Logprob]
+
+
+@dataclass
+class FlattenLogprobs(MutableSequence[LogprobsOnePosition]):
```
njhill, line 33: Sorry for the late comment but IMO a better name for this class would be `FlatLogprobs`...

Jialin: Thanks @njhill for the suggestions. I'll address your comment in a follow up PR along other leftover feedbacks.

PR #28176, opened 2025-11-06
[V1] [Hybrid] Lighter Mamba Prefix Caching for Hybrid Models

peakcrosser7:
## Purpose
Currently, Automatic Prefix Caching for Mamba-based hybrid models does not support architectures such as GDN. To address this, we propose a lightweight Mamba Prefix Caching design called **Lighter-Mamba-Prefix-Cache**.  
Its core idea is to directly cache Mamba states using a block-aligned scheduling approach, enabling rapid support for Prefix Caching in Mamba models **without modifying any kernel code**, while maintaining compatibility with SPS, MTP, and Eagle.  
This solution has already been validated on Qwen3-Next-80B-A3B-Instruct.

#### Block Allocation Design
For each request, the number of blocks allocated per Mamba group is changed from the original fixed `1 + sps` to `2 + sps + N`, where:
* The 1st block: Used to **load the Mamba state from a prefix-caching hit**. If there is no cache hit, a null-block is used as a placeholder.
* The next `1 + sps` blocks: **Reserved for runtime usage**, identical to the original `1 + sps` blocks without prefix caching.
* The following N blocks: Used to **cache Mamba states**.

block_alloc

#### Prefix Matching Logic
scheduler_logic

worker_logic

#### Block-Aligned Scheduling
Since requests are hashed at the granularity of `block_size`, **Mamba states must be aligned to `block_size`** boundaries before caching. This ensures that each Mamba state corresponds to exactly one block hash.  
Lighter prefix cache stores variable-length chunk states—i.e., the number of tokens (or the incremental length) associated with **each cached Mamba state may vary, but it is always a multiple of `block_size`**.

With Mamba Prefix Caching enabled, the scheduler behaves as follows:
* Decode requests: Scheduling logic remains unchanged (Lighter prefix caching is not yet supported for decode).
* Prefill requests: 
    * **The number of tokens scheduled per step must be an integer multiple of `block_size`**, except for the final chunk of the request.
    * **The last prefill chunk is split to align with `block_size`**, ensuring its size is `≤ block_size`. This maximizes the length of the prompt that can be cached during the prefill phase.

## Test Plan
TODO

## Test Result
TODO

---
<details>

</details>

### DIFF
--- a/vllm/engine/arg_utils.py
+++ b/vllm/engine/arg_utils.py
@@ -1387,6 +1387,15 @@ class EngineArgs:
             f"dcp_size={self.decode_context_parallel_size}."
         )
 
+        if (envs.VLLM_USE_LIGHTER_MAMBA_CACHE
+            and self.enable_prefix_caching 
+            and model_config.is_hybrid
+        ):
+            assert envs.VLLM_USE_V1, (
+                "Prefix caching for hybrid models requires the V1 engine.")
+            assert self.enable_chunked_prefill, (
+                "Prefix caching for hybrid models requires chunked prefill.")
+
         cache_config = CacheConfig(
             block_size=self.block_size,
             gpu_memory_utilization=self.gpu_memory_utilization,
--- a/vllm/envs.py
+++ b/vllm/envs.py
@@ -219,7 +219,7 @@ if TYPE_CHECKING:
     VLLM_GC_DEBUG: str = ""
     VLLM_DISABLE_SHARED_EXPERTS_STREAM: bool = False
     VLLM_COMPILE_CACHE_SAVE_FORMAT: Literal["binary", "unpacked"] = "binary"
-
+    VLLM_USE_LIGHTER_MAMBA_CACHE: bool = False
 
 def get_default_cache_root():
     return os.getenv(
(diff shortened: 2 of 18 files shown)


heheda12345 (commented):
1. If I understand correctly, we will always cache the state at token num_tokens - num_tokens%block_size, but there is no ensure that tokens in other positions are cached. In this case, can you ensure system prompt is cached? 
2. Memory layout: I feel that we don't need such a new kv cache memory design. We can have a block_id list with length num_tokens / block_size + num_spec_decode_tokens, and always make sure that the state of the previous schedule step is at block num_computed_tokens / block_size

For example, if num_computed_tokens=29 and we schedule 1 new token, the kv cache before this step is:
block 0: N/A
block 1: token 29
block 2: N/A
block 3: N/A

-> run main model
block 0: N/A
block 1: token 30
block 2: token 31
block 3: token 32
-> adjust kv cache based on number of accepted tokens
* if no token is accepted:
block 0: N/A
block 1: token 30
block 2: N/A
block 3: N/A
* if token 30 is accepted:
block 0: N/A
block 1: token 31
block 2: N/A
block 3: N/A
* if token 30 & 31 are accepted:
block 0: N/A
block 1: token 31
block 2: token 32
block 3: N/A
* if token 30 & 31 & 32 are accepted:
block 0: token 15
block 1: token 31
block 2: token 33
block 3: N/A

Then, in the next schedule step, the previous state is always at block [num_computed_tokens / block_size]

3. I'm still concerning about whether we should increase the complexity of scheduler to avoid kernel changes of mamba layers.

minminsun:
> @minminsun I'm curious about the performance of chunk-granularity caching. Would it be possible to share any cache hit ratio results? Thanks!

Though I cannot provide specific cache hit ratio metrics for our production services, the drop relative to block-granularity caching is less than 10%.

joennlae:
I added an rebased version here: https://github.com/vllm-project/vllm/pull/30725

PR #28187, opened 2025-11-06
[Bugfix]lora replace_layer support custom op

zhanghn-tc:
## Purpose
When I use vLLM in plugin mode, I use a custom op（@RowParallelLinear.register_oot
class RowParallelLinear_OOT(RowParallelLinear, LinearBase_OOT)）, at which point the network structure `RowParallelLinear` is replaced with `RowParallelLinear_OOT`. However, in the current `def can_replace_layer` function of LoRA, a layer can only be replaced when `type(source_layer) is RowParallelLinear`. This causes the replacement to fail and prevents the network layer from being successfully replaced with the LoRA-enabled layer (`RowParallelLinearWithLoRA`), resulting in an assertion error in `register_module` at line 591 of `vllm/vllm/lora/models.py`.

This PR updates the logic, adding support for LoRA replacement for custom ops, so that layers wrapped for custom operation support can be handled correctly.

This
## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/lora/layers/column_parallel_linear.py
+++ b/vllm/lora/layers/column_parallel_linear.py
@@ -155,9 +155,21 @@ class ColumnParallelLinearWithLoRA(BaseLinearLayerWithLoRA):
         packed_modules_list: list,
         model_config: PretrainedConfig | None,
     ) -> bool:
-        return type(source_layer) is ColumnParallelLinear or (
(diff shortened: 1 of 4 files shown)


jeejeelee:
Sorry for missing this PR. 
I don't quite understand the necessity of this PR, could u please explain in detail

PR #28194, opened 2025-11-06
[Model] Allow users to control skip reading cache per request.

noooop:
## Purpose
Even if prefix caching is enabled, it cannot read caching in the following two scenarios:
- prompt logprobs
- all pooling

Otherwise, the output might be less than n_prompt_tokens.

skip_reading_caching can still write to caching to accelerate following requests

Address https://github.com/vllm-project/vllm/pull/27145#discussion_r2441845249

## Test Plan

tests/models/language/pooling/test_extract_hidden_states.py

## Test Result

main:

Even if chunked_prefill is not enabled, prefix_caching + all pooling will still cause the following error.
(EngineCore_DP0 pid=2893225) AssertionError: partial prefill not supported with ALL pooling

```
n_prompt_tokens:55 n_output_tokens:55 num_cached_tokens:0
n_prompt_tokens:56 n_output_tokens:8 num_cached_tokens:48
n_prompt_tokens:57 n_output_tokens:8 num_cached_tokens:48

n_prompt_tokens:55 n_output_tokens:48 num_cached_tokens:7
n_prompt_tokens:56 n_output_tokens:8 num_cached_tokens:48
n_prompt_tokens:57 n_output_tokens:8 num_cached_tokens:48
```

this pr:

Turn off prefix_caching during all pooling requests.

```
n_prompt_tokens:55 n_output_tokens:55 num_cached_tokens:0
n_prompt_tokens:56 n_output_tokens:56 num_cached_tokens:0
n_prompt_tokens:57 n_output_tokens:57  num_cached_tokens:0

n_prompt_tokens:55 n_output_tokens:55  num_cached_tokens:0
n_prompt_tokens:56 n_output_tokens:56  num_cached_tokens:0
n_prompt_tokens:57 n_output_tokens:57 num_cached_tokens:0
```

---
<details>

</details>


vllm/v1/core/kv_cache_manager.py
```diff
@@ -186,10 +186,19 @@ def get_computed_blocks(self, request: Request) -> tuple[KVCacheBlocks, int]:
...
+        if (
+            not self.enable_caching
+            or (
+                request.sampling_params is not None
+                and request.sampling_params.disable_prefix_caching
+            )
+            or (
+                request.pooling_params is not None
+                and request.pooling_params.disable_prefix_caching
+            )
```
heheda12345: what about merging them into one attribute `request.disable_prefix_caching`. And I don't like "disable_prefix_caching" as it has two meanings:
1. skip get_computed_blocks and return 0
2. skip cache_blocks

Can you find a better name?

noooop: Do you have any suggested names? I am very bad at naming things and not very familiar with KVCache-related code, but my requirements are very clear.

Even if prefix caching is enabled, it cannot be used in the following two scenarios:

- prompt logprobs
- all pooling

Otherwise, the output might be less than n_prompt_tokens.

noooop: how about skip_prefix_caching

vllm/v1/core/kv_cache_manager.py
```diff
@@ -186,11 +186,11 @@ def get_computed_blocks(self, request: Request) -> tuple[KVCacheBlocks, int]:
...
-        if not self.enable_caching or (
-            request.sampling_params is not None
-            and request.sampling_params.prompt_logprobs is not None
-        ):
+        # When the request requires prompt logprobs or
+        # When pooling model with all pooling
+        # we skip reading cache, but still can write to cache
+        # to accelerate following requests
+        if not self.enable_caching or request.skip_reading_cache:
             return self.empty_kv_cache_blocks, 0
```
noooop: @heheda12345

how about skip_reading_cache

we skip reading cache, but still can write to caching to accelerate following requests

noooop: Frankly, I'm not 100% sure how it can be written to cache, but after testing, it can

I am very happy to have this feature (it still can write to caching to accelerate following requests)

vllm/pooling_params.py
```diff
@@ -93,6 +94,8 @@ def verify(
         # plugin task uses io_processor.parse_request to verify inputs,
         # skipping PoolingParams verify
         if self.task == "plugin":
+            if self.skip_reading_cache is None:
+                self.skip_reading_cache = True
             return
```
noooop, line 100: https://buildkite.com/vllm/ci/builds/38109/steps/canvas?jid=019a62ce-5580-4053-a3b7-b1cbaf47b26e

models/multimodal/pooling/test_prithvi_mae.py::test_models_image[mgazz/Prithvi-EO-2.0-300M-TL-Sen1Floods11]

I'm not sure why this is needed to pass.

```[2025-11-08T10:58:54Z] INFO 11-08 02:58:54 [arg_utils.py:1760] (Disabling) prefix caching by default
[2025-11-08T10:58:54Z] INFO 11-08 02:58:54 [vllm.py:500] Cudagraph is disabled under eager mode
[2025-11-08T10:58:54Z] INFO 11-08 02:58:54 [vllm.py:582] Only "last" pooling supports chunked prefill and prefix caching; disabling both.
```

For this model, self.enable_caching should be False, so it should no need skip_reading_cache.

Perhaps for this model, self.enable_caching is not set or is set incorrectly.

I will completely resolve this issue in #26623.

vllm/v1/core/kv_cache_manager.py
```diff
@@ -186,11 +186,11 @@ def get_computed_blocks(self, request: Request) -> tuple[KVCacheBlocks, int]:
...
-        # When the request requires prompt logprobs, we skip prefix caching.
-        if not self.enable_caching or (
-            request.sampling_params is not None
-            and request.sampling_params.prompt_logprobs is not None
-        ):
+        # When the request requires prompt logprobs or
+        # When pooling model with all pooling
+        # we skip reading cache, but still can write to cache
+        # to accelerate following requests
+        if not self.enable_caching or request.skip_reading_cache:
```
heheda12345: `skip_reading_cache` is pretty clear for me.

```suggestion
        # We skip finding the prefix cache hit when prefix caching is 
        # disabled or the request is marked as skipping kv cache read 
        # (which happens when the request requires prompt logprobs 
        # or calls a pooling model with all pooling).
        if not self.enable_caching or request.skip_reading_cache:
```

I've made the following into a sentence.
```
         # Prefix caching is disabled or
        # When the request requires prompt logprobs or
        # When pooling model with all pooling
```

Feel that we don't need to discuss "can write to cache to accelerate following requests" here. If prefix caching is disabled, we don't write to cache. If you want to emphasize this, can you put it into the comments for the `skip_reading_cache` attribute of `Request` class?

vllm/v1/request.py
```diff
@@ -180,6 +180,20 @@ def num_tokens_with_spec(self) -> int:
     def num_output_tokens(self) -> int:
         return len(self._output_token_ids)
 
+    @property
+    def skip_reading_cache(self) -> bool:
```
heheda12345: To accelerate it a bit, you can:
1. initialize the value of skip_reading_cache in process_inputs of `vllm/v1/engine/processor.py`
2. Put it in EngineCoreRequest
3. copy from EngineCoreRequest to Request by `from_engine_core_request`

In this way, skip_reading_cache will be computed in the frontend process rather than the engine core busy loop. (Though there won't be too much speedup, I want to avoid unnecessary operation in KVCacheManager as much as possible)

heheda12345:
emmm... what about `skip_reading_prefix_cache`? There are too many types of cache in vllm

vllm/v1/request.py
noooop: I think the get_skip_reading_prefix_cache logic best belongs in the Request.

I have cached the result of skip_reading_prefix_cache.

heheda12345 (approved):
LGTM!

vllm/v1/request.py
heheda12345: yeah caching also make sense to me

PR #28205, opened 2025-11-06
[build] fix: only build FA3 for Hopper

AlpinDale:
## Purpose
FA3 is only used for Hopper, yet it's built for every Ampere (sm_80, sm_87, sm_89) and Ada (sm_89) arches. This needlessly increases build times, and inflates the wheel size. FA3 technically supports A100 (sm_80), but as discussed offline with @mgoin, it's not stable and vLLM does not actually enable it.

vLLM FA PR: vllm-project/flash-attention#105

---
<details>

</details>

### DIFF
--- a/cmake/external_projects/vllm_flash_attn.cmake
+++ b/cmake/external_projects/vllm_flash_attn.cmake
@@ -37,8 +37,9 @@ if(VLLM_FLASH_ATTN_SRC_DIR)
 else()
   FetchContent_Declare(
           vllm-flash-attn
-          GIT_REPOSITORY https://github.com/vllm-project/flash-attention.git
-          GIT_TAG a893712401d70362fbb299cd9c4b3476e8e9ed54
+          # temporary until vllm-project/flash-attention#105 is merged
+          GIT_REPOSITORY https://github.com/AlpinDale/flash-attention.git
+          GIT_TAG 3e16ec5054dca0b5f62ff58e76977d07e24ac55f
           GIT_PROGRESS TRUE
           # Don't share the vllm-flash-attn build between build types
           BINARY_DIR ${CMAKE_BINARY_DIR}/vllm-flash-attn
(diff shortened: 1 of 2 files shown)


cmake/external_projects/vllm_flash_attn.cmake
```diff
@@ -37,8 +37,9 @@ if(VLLM_FLASH_ATTN_SRC_DIR)
 else()
   FetchContent_Declare(
           vllm-flash-attn
-          GIT_REPOSITORY https://github.com/vllm-project/flash-attention.git
-          GIT_TAG a893712401d70362fbb299cd9c4b3476e8e9ed54
+          # temporary until vllm-project/flash-attention#105 is merged
+          GIT_REPOSITORY https://github.com/AlpinDale/flash-attention.git
```
heheda12345, line 41: we should wait until the above PR is merged and use the flash-attention in vllm-project

setup.py
```diff
@@ -579,6 +579,30 @@ def get_vllm_version() -> str:
     return version
 
 
+def _has_sm90_or_higher() -> bool:
```
AlpinDale, line 582: @LucasWilkinson thoughts on excluding Blackwell from FA3 builds as well? From what I understand, Blackwell still uses FA2 (at least until FA4 support lands).

LucasWilkinson: hmm this `_has_sm90_or_higher` is a bit messy; sorry should have reviewed this closer. 

We should either do:
1) `if not arch_list` use `torch.cuda.get_arch_list()`
2) handle this at the CMake level like FlashMLA and just create empty targets when no kernels would get built: https://github.com/vllm-project/vllm/blob/60e089f0b90b1fe9b65224b069c953927d1f3b44/cmake/external_projects/flashmla.cmake#L128-L132

preference for #2 in which case we wouldnt need to filter for Blackwell here

AlpinDale: Understood. I'm not a fan of the current approach either; I'll change it to follow FlashMLA's example later today.

LucasWilkinson: thank you! 🙏

PR #26822, opened 2025-10-14
[bugfix] Fix online serving crash when text type response_format is received

cjackal:
## Purpose

Fix #26639.

Especially, this PR adds a proper input validation to `StructuredOutpusParams` not to generate unschedulable structured outputs params, and adjust sampling parameter generation logic in `ChatCompletionRequest.to_sampling_parameters()` and `OpenAIServingResponses.create_responses()` to reflect the change and be more robust.

co-authored by @j0shuajun who first reported the issue and minimal reproducible examples on our side.

## Test Plan

Pass a chat completion request with `"response_format": {"type": "text"}`:

```curl
curl -XPOST http://localhost:8080/v1/chat/completions -H 'Content-Type: application/json' -d '{"model":"openai/gpt-oss-120b","messages":[{"role":"user","content":"hello"}],"max_tokens":2048,"stream":false,"response_format":{"type":"text"}}'
```

Plus, pass a chat completion request with `"response_format": {"type": "json_object"}` to check for regressions.

```curl
curl -XPOST http://localhost:8080/v1/chat/completions -H 'Content-Type: application/json' -d '{"model":"openai/gpt-oss-120b","messages":[{"role":"user","content":"hello"}],"max_tokens":2048,"stream":false,"response_format":{"type":"json_object"}}'
```

## Test Result

Return a valid response in both cases.

```text
# response_format text - respond with normal text
{"id":"chatcmpl-b6a22dfca265460ab316d7ea490d974b","object":"chat.completion","created":1760457543,"model":"openai/gpt-oss-120b","choices":[{"index":0,"messages":{"role":"assistant","content":"Hello! How can I assist you today?","refusal":null,"annotations":null,"audio":null,"function_call":null,"tool_calls":[],"reasoning_content":"We need to respond: greeting. No instructions."},"logprobs":null,"finish_reason":"stop","stop_reason":null,"token_ids":null}],"service_tier":null,"system_fingerprint":null,"usage":{"prompt_tokens":70,"total_tokens":99,"completion_tokens":29,"prompt_tokens_details":null},"prompt_logprobs":null,"prompt_tokens_ids":null,"kv_transfer_params":null}

# response_format json_object - respond with JSON object
{"id":"chatcmpl-9643189ba2674cc2b471bd9a63472e
[...]

PR #14779, opened 2025-03-13
[V1] guidance backend for structured output + `auto` fallback mode

russellb:
This is the V1 integration for
[guidance](https://github.com/guidance-ai/llguidance) as a backend for
structured output. There is a V0 integration in #14589.

Since this is the second backend supported by V1, this PR also includes a new backend mode called `auto`, which provides opinionated behavior with fallback. This is not the default since the behavior may vary release to release.

This backend provides some key benefits to V1:

* Broader jsonschema support
* Quick startup performance for large schemas

Instead of precomputing the masks for all states, this is done on
the fly. We see very fast request startup times, even for large
schemas.

This should make V1 roughly feature equivalent to V0 in terms of the
types of schemas it can support.

More technical details are available in the llguidance git repo.

Signed-off-by: Russell Bryant <rbryant@redhat.com>
Co-authored-by: Loc Huynh <jc1da.3011@gmail.com>

PR #28210, opened 2025-11-06
[Bugfix] Prevent crash on empty grammar string

tjandy98:
Fixes https://github.com/vllm-project/vllm/issues/28209


russellb (approved):
lgtm, thank you!

vllm/v1/engine/processor.py
```diff
@@ -270,6 +270,12 @@ def _validate_structured_output(self, params: SamplingParams) -> None:
             raise ValueError(
                 f"Choice '{params.structured_outputs.choice}' cannot be an empty list"  # noqa: E501
             )
+        # Reject empty string grammar early to avoid engine-side crashes
+        if (
+            isinstance(params.structured_outputs.grammar, str)
+            and params.structured_outputs.grammar.strip() == ""
```
chaunceyjiang, line 276: I think you should fix it here  `def validate_xgrammar_grammar(...)`

tjandy98: Thanks for the review. It is placed in `def _validate_structured_output` so that that invalid request fails fast regardless of which backend is picked

russellb: I agree that it seems like a valid check for all backends, even if others wouldn't have crashed on it.

cjackal: I stumbled upon a similar indecision in resolving the crash due to invalid grammar in #26822. One can fastest-fail if the validation logic is placed in `StructuredOutputParams` - it shouldn't be hard to implement with pydantic field validator I think.

A thought: current structured output validation logic is scattered across `StructuredOutputsParams`, `Processor`, and each structured output backends, which is not so intuitive and easy for each developer to narrowed down on the first-met one without knowing the others. I wonder if we can absorb all the grammar validation logic into `StructuredOutputParams`; `StructuredOutputParams` is a pydantic model (whose main role is schema validation) and it sounds like a code smell that an empty pydantic model is first created and assigned an attribute later -  this pattern occurs in many places in the vllm codebase yet. And assuring that a (successfully created) instance of `StructuredOutputParams` is compilable is more intuitive and future-safe.

chaunceyjiang: > Thanks for the review. It is placed in `def _validate_structured_output` so that that invalid request fails fast regardless of which backend is picked

Just a question: do all backends currently not support `grammar` being an empty string? @tjandy98

chaunceyjiang: > I wonder if we can absorb all the grammar validation logic into StructuredOutputParams; 

Just a question: it seems that the validation logic differs across backends — can pydantic validation handle such cases? @russellb

cjackal: > Just a question: it seems that the validation logic differs across backends — can pydantic validation handle such cases? @russellb

I was thinking of moving most backend-agnostic checks in `Processor._validate_structured_output` to `StructuredOutputsParams`'s validator which delegates backend-specific checks to `validate_*_grammar` functions in each backend; [the four `validate_*_grammar` functions](https://github.com/vllm-project/vllm/blob/b886068056a05857f796909d2f8573b36fc668a5/vllm/v1/engine/processor.py#L274-L306) are accepting `SamplingParams` as input but all of them just uses `SamplingParams.structured_outputs` [like here](https://github.com/vllm-project/vllm/blob/3143eb23fc4e017bc31d11a9756d5a788d6f7e33/vllm/v1/structured_output/backend_xgrammar.py#L279-L282) (except the leftover TODO in `guidance` backend since #14779 about using the given tokenizer, wonder if this TODO is planned to be completed in future as `guidance` seems like the unique backend that exploits tokenizers in the matcher) so we can directly pass `SamplingParams.structured_outputs` instead. All the validators can be executed as `after`-type pydantic model validator which first check the structured output backend from `StructuredOutputsParams._backend` and then run the corresponding `validate_*_grammar` validator function, somewhat like

```python
class StructuredOutputsParams:
    ...

    @model_validator(mode='after')
    def validate_constraints(self) -> Self:
        backend = self._backend

        # We can also insert backend-independent validations like #26822, #28113, or #28210

        if (
            isinstance(self.choice, list)
            and not self.choice
        ):
            # It is invalid for choice to be an empty list
            raise ValueError(
                f"Choice '{self.choice}' cannot be an empty list"  # noqa: E501
            )
        ...

        if backend.startswith("xgrammar"):
            # xgrammar with no fallback
            validate_xgrammar_grammar(params)
        elif backend.startswith("guidance"):
            ...
```

(pydantic supports [ordering of multiple validators](https://docs.pydantic.dev/latest/concepts/validators/#ordering-of-validators) so one can further refactor it if needed.)

Then `Processor._validate_structured_output` only covers lightweight config checks (e.g. whether the sampling parameter has `.structured_outputs` attribute, the structured output config was auto, ...), just as other validators like `Processor._validate_lora` or `Processor._validate_logit_bias`.

chaunceyjiang: sounds good!

PR #28211, opened 2025-11-06
Enhance Helm chart installation instructions

ccnmxns:
Added instructions for cloning the repo and installing the Helm chart with optional AMD Radeon configuration.

Its main purpose is to help newcomers to deploy easily (or at least in a more guided way) the chart, with a little help for Radeon users like myself.


docs/deployment/frameworks/helm.md
```diff
@@ -17,18 +17,45 @@ Before you begin, ensure that you have the following:
...
+cd vllm/examples/online_serving/chart-helm
+```
+
 To install the chart with the release name `test-vllm`:
 
 ```bash
 helm upgrade --install --create-namespace \
   --namespace=ns-vllm test-vllm . \
-  -f values.yaml \
+  -f values.yaml \  # If you have another values file to override the default one, add another '-f values.override-yaml' line.
```
heheda12345, line 32: Is the only goal of `cd vllm/examples/online_serving/chart-helm` to use `values.yaml`? If yes, I would like to suggest reverting the code here and add
```
An example `values.yaml` can be found in `vllm/examples/online_serving/chart-helm/values.yaml`
```
to the end of the bash script.

docs/deployment/frameworks/helm.md
```diff
@@ -17,18 +17,45 @@ Before you begin, ensure that you have the following:
...
   --set secrets.s3endpoint=$ACCESS_POINT \
   --set secrets.s3bucketname=$BUCKET \
   --set secrets.s3accesskeyid=$ACCESS_KEY \
   --set secrets.s3accesskey=$SECRET_KEY
 ```
 
+If you have a AMD Radeon, you might as well add the `values.override.yaml` file with a content like this:
```
heheda12345, line 39: can you put them into a file like `values.amd-override.yaml` in the same directory as `values.yaml` and give the full command for using the override here?

docs/deployment/frameworks/helm.md
ccnmxns: I could resolve it in another way. But you have a point.

Let me know if this solution I will provide is okay ;)

docs/deployment/frameworks/helm.md
ccnmxns: Good point.

PR #28064, opened 2025-11-04
[Kernels] Split up fused_moe/layer.py, isolate more modular kernel code

bnellnm:
## Purpose
`fused_moe/layer.py` was getting quite large.  I've split it up so that each main class has its own file:

- `FusedMoEMethodBase` -> `moe_method_base.py`
- `FusedModularMoEMethod` -> `modular_moe_method.py`
- `UnquantizedFusedMoEMethod` -> `unquantized_moe_method.py`

I've also moved more modular kernel/all2all code into `modular_moe_method.py` and `all2all_utils.py`

cc @varun-sundar-rabindranath 

## Test Plan
CI tests
Tested gpt-oss-120b on H100 machine with 8 gpus

## Test Result

gpt-oss-120b lm-eval results
```
vllm serve openai/gpt-oss-120b --trust-remote-code --port 8001  --data-parallel-size 1 --tensor-parallel-size 8

lm_eval --model local-completions --tasks gsm8k --model_args model=openai/gpt-oss-120b,base_url=http://127.0.0.1:8001/v1/completions,num_concurrent=1,max_retries=3,tokenized_requests=False --limit 100

|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value|   |Stderr|
|-----|------:|----------------|-----:|-----------|---|----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  | 0.43|±  |0.0498|
|     |       |strict-match    |     5|exact_match|↑  | 0.27|±  |0.0446|

Main has similar lm-eval results.
```

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #28238, opened 2025-11-06
[NVIDIA] [feat] Integrate flashinfer Trtllmgen bf16 moe

jiahanc:
## Purpose
- Integrate flashinfer trtllm-gen BF16 moe to supported models
- Fixed flashinfer trtllm-gen moe SM check
## Test Plan
`VLLM_USE_FLASHINFER_MOE_FP16=1 VLLM_FLASHINFER_MOE_BACKEND=latency vllm serve Qwen/Qwen3-Next-80B-A3B-Instruct  --max-num-batched-tokens 8192  --max-model-len 32768  --no-enable-prefix-caching  --async-scheduling  --compilation_config.pass_config.enable_fi_allreduce_fusion true --compilation_config.pass_config.enable_noop true  --compilation_config.custom_ops+=+rms_norm  --compilation_config.cudagraph_mode FULL_DECODE_ONLY  --compilation_config.splitting_ops []  --disable-log-requests -tp 2 `

`lm_eval --model local-completions --tasks gsm8k --model_args model=Qwen/Qwen3-Next-80B-A3B-Instruct,base_url=http://0.0.0.0:8000/v1/completions,max_retries=3,tokenized_requests=False,timeout=1200,max_gen_toks=2048,max_length=8192 --batch_size 2048 --trust_remote_code --limit 0.5`
## Test Result
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9530|±  |0.0082|
|     |       |strict-match    |     5|exact_match|↑  |0.9167|±  |0.0108|

```
---
<details>

</details>

### DIFF
--- a/vllm/model_executor/layers/fused_moe/unquantized_fused_moe_method.py
+++ b/vllm/model_executor/layers/fused_moe/unquantized_fused_moe_method.py
@@ -25,6 +25,9 @@ from vllm.model_executor.layers.fused_moe.modular_kernel import (
     FusedMoEPermuteExpertsUnpermute,
     FusedMoEPrepareAndFinalize,
 )
+from vllm.model_executor.layers.quantization.utils.flashinfer_utils import (
+    convert_moe_weights_to_flashinfer_trtllm_block_layout,
+)
 from vllm.model_executor.utils import set_weight_attrs
 from vllm.platforms import current_platform
 from vllm.platforms.interface import CpuArchEnum
@@ -64,7 +67,7 @@ class UnquantizedFusedMoEMethod(FusedMoEMethodBase, CustomOp):
             has_flashinfer()
             and envs.VLLM_USE_FLASHINFER_MOE_FP16
             and envs.VLLM_FLASHINFER_MOE_BACKEND == "latency"
-            and current_platform.get_device_capability()[0] >= 10
+            and current_platform.has_device_capability(100)
         )
 
         # FlashInfer CUTLASS MoE is only supported on Hopper and later GPUS
@@ -73,7 +76,7 @@ class UnquantizedFusedMoEMethod(FusedMoEMethodBase, CustomOp):
             and envs.VLLM_USE_FLASHINFER_MOE_FP16
             and self.moe.moe_parallel_config.use_ep
             and self.moe.moe_parallel_config.dp_size == 1
-            and current_platform.get_device_capability()[0] >= 9
+            and current_platform.has_device_capability(90)
         )
 
         if self.flashinfer_trtllm_moe_enabled:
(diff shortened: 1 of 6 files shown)


vadiklyutiy:
@jiahanc 
To double check, cutlass here means "vllm'a cutlass" or "FI's cutlass"?

vadiklyutiy:
> > If for high concurrency trt-llm gem MoE is better maybe it's worth always enable it, not only for `VLLM_FLASHINFER_MOE_BACKEND=latency` ?
> 
> Yes, the trtllmgen moe is already the default FI MOE backend if FI MOE is enabled.

Oh, sorry, I just looked only into **Test Plan** section in the first msg and saw there `VLLM_FLASHINFER_MOE_BACKEND=latency`

vllm/model_executor/layers/fused_moe/unquantized_fused_moe_method.py
```diff
@@ -219,6 +238,67 @@ def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
             layer.w13_weight.data = shuffled_w13
             layer.w2_weight.data = shuffled_w2
 
+        if self.flashinfer_trtllm_moe_enabled:
+            from flashinfer.fused_moe.core import (
```
pavanimajety: Minor refactoring: can we move all the `self.flashinfer_trtllm_moe_enabled` post-weight-loading processing code block to flashinfer utilities?

pavanimajety: This might also benefit other precisions since this code/shuffling step is same for all trtllm moe flavors.

vllm/model_executor/layers/fused_moe/unquantized_fused_moe_method.py
```diff
@@ -58,6 +59,14 @@ def __init__(self, moe: FusedMoEConfig):
         else:
             self.rocm_aiter_fused_experts = None  # type: ignore
 
+        # Flashinfer TRTLLM MoE is only supported on Blackwell and later GPUS
+        self.flashinfer_trtllm_moe_enabled = (
+            has_flashinfer()
+            and envs.VLLM_USE_FLASHINFER_MOE_FP16
+            and envs.VLLM_FLASHINFER_MOE_BACKEND == "latency"
+            and current_platform.get_device_capability()[0] >= 10
```
vadiklyutiy: There is `has_device_capability()` that looks a little bit better

vllm/model_executor/layers/fused_moe/unquantized_fused_moe_method.py
```diff
@@ -58,6 +59,14 @@ def __init__(self, moe: FusedMoEConfig):
         else:
             self.rocm_aiter_fused_experts = None  # type: ignore
 
+        # Flashinfer TRTLLM MoE is only supported on Blackwell and later GPUS
+        self.flashinfer_trtllm_moe_enabled = (
+            has_flashinfer()
+            and envs.VLLM_USE_FLASHINFER_MOE_FP16
+            and envs.VLLM_FLASHINFER_MOE_BACKEND == "latency"
```
vadiklyutiy, line 69: We discussed that _"Yes, the trtllmgen moe is already the default FI MOE backend if FI MOE is enabled."_
If so, why do we need this check of `VLLM_FLASHINFER_MOE_BACKEND`?

vllm/model_executor/layers/quantization/utils/flashinfer_utils.py
```diff
@@ -284,7 +284,7 @@ def get_flashinfer_moe_backend() -> FlashinferMoeBackend:
     if flashinfer_moe_backend in backend_map:
         if (
             flashinfer_moe_backend == "latency"
-            and not current_platform.is_device_capability(100)
+            and not current_platform.has_device_capability(100)
```
vadiklyutiy, line 293: The same about _"Yes, the trtllmgen moe is already the default FI MOE backend if FI MOE is enabled."_

Shouldn't we remove `flashinfer_moe_backend` here?

vllm/model_executor/layers/fused_moe/unquantized_fused_moe_method.py
jiahanc: I think we should give user choice to choose between different MOE backends. Therefore we now set default backend be latency instead of totally remove the `VLLM_FLASHINFER_MOE_BACKEND`

vllm/model_executor/layers/quantization/utils/flashinfer_utils.py
jiahanc: same with above

vllm/model_executor/layers/fused_moe/unquantized_fused_moe_method.py
vadiklyutiy: I agree that offering a choice is good behaviour.

However, the current naming is misleading. Suppose a user has a high-throughput use case; it is quite natural for them to set `VLLM_FLASHINFER_MOE_BACKEND == "throughput"`. By doing so, however, they end up using FlashInfer’s CUTLASS kernel, which is slower for high-throughput scenarios than the TRT-LLM generation backend. I am afraid this is not the desired behaviour.

jiahanc: discussed offline, we will do a follow up PR about this

vadiklyutiy (approved):
Looks good for me

pavanimajety (approved):
Thanks for the PR and making the changes!

vadiklyutiy:
@jiahanc 
I tried a bit different `lm_eval`
```
VLLM_USE_FLASHINFER_MOE_FP16=1 vllm serve Qwen/Qwen3-Next-80B-A3B-Instruct -tp 2 --enable-expert-parallel --async-scheduling --no-enable-prefix-caching
```
```
```

and got zero accuracy 
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.0174|±  |0.0036|
```

jiahanc:
> Seems the problem when `--enable-expert-parallel` added.

Thanks for finding the issue. EP was tested when related flashinfer PR just merged. After bisecting, the issue comes from a later PR in flashinfer https://github.com/flashinfer-ai/flashinfer/pull/2092, will convert current PR to draft and reopen when issue fixed.

bnellnm:
> > EP accuracy was test at the time of development and accuracy was good so didnt test after rebase. Reverted back to commit before 1st rebase, seeing good accuracy, need to bisect the main branch commits
> 
> After bisect, after the refactor PR #28064, same code change cant work with expert parellel. @bnellnm may I ask do you have thoughts on what other code changes need after the refactor?



> > EP accuracy was test at the time of development and accuracy was good so didnt test after rebase. Reverted back to commit before 1st rebase, seeing good accuracy, need to bisect the main branch commits
> 
> After bisect, after the refactor PR #28064, same code change cant work with expert parellel. @bnellnm may I ask do you have thoughts on what other code changes need after the refactor?

Hi @jiahanc , I don't see anything obvious. #28064 mostly just moved classes out of layer.py and into individual files.  I would check to make sure that a message like "Swapping out UnquantizedFusedMoEMethod" does not appear in the DEBUG logs.  I don't think it should happen but that's the only thing I can think of that would cause something to break unexpectedly .

bnellnm:
> --max-num-batched-tokens 8192 --max-model-len 32768

FYI, I think there are also some current CI problems with the Qwen3 model that may or may not be related to the accuracy problems you are seeing.

vadiklyutiy:
> FYI, I think there are also some current CI problems with the Qwen3 model that may or may not be related to the accuracy problems you are seeing.

@bnellnm Qwen3 or Qwen3-next? Could you reference to the problems?

PR #28241, opened 2025-11-06
Fix CPU offload KV cache accounting and align CUDA xformers pin

### LINKED ISSUE #27934: V1 Engine: Memory allocation failures and crashes with 7B-13B models on RTX 3060 12GB
## Environment
- **vLLM version**: 0.11.1rc6.dev35+g29de3cdee (nightly), 0.6.4.post1 (stable)
- **GPU**: NVIDIA RTX 3060 12GB (Compute 8.6, Ampere)
- **CUDA**: 13.0 (Driver 580.95.05)
- **OS**: Ubuntu 24.04
- **Python**: 3.12
- **Container**: `vllm/vllm-openai:nightly` and `vllm/vllm-openai:v0.6.4.post1`

## Description
The V1 engine consistently fails to initialize with 7B-13B models on RTX 3060 12GB, despite having sufficient GPU memory. Multiple memory-related errors occur across different configurations.

## Models Tested
All models fail with similar errors:
- `Qwen/Qwen2.5-Coder-7B-Instruct`
- `Gryphe/MythoMax-L2-13b`
- `NousResearch/Nous-Hermes-2-SOLAR-10.7B`
- `cognitivecomputations/dolphin-2.6-mistral-7b`

### Error 1: Insufficient Memory for Cache Blocks (with CPU offload)
```
ValueError: No available memory for the cache blocks. Try increasing `gpu_memory_utilization` when initializing the engine.
```
**Configuration**: `--cpu-offload-gb 8 --gpu-memory-utilization 0.5`
**GPU Memory Available**: 11.63 GiB total, ~6-7 GiB free after model load

### Error 2: Memory Utilization Check Failure (without CPU offload)
```
ValueError: Free memory on device (6.44/11.63 GiB) on startup is less than desired GPU memory utilization (0.9, 10.47 GiB). Decrease GPU memory utilization or reduce GPU memory used by other processes.
```
**Configuration**: `--gpu-memory-utilization 0.9` (no CPU offload)
**Issue**: Check happens BEFORE model loads, counts memory from previous crashed processes

### Error 3: OOM During Model Loading
```
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 224.00 MiB. GPU 0 has a total capacity of 11.63 GiB of which 194.81 MiB is free. Process 145105 has 11.43 GiB memory in use.
```
**Issue**: Previous vLLM process keeps GPU memory allocated even after container stops

### Nightly Image (V1 Engine)
```bash
docker run -d --name vllm-test --gpus all -p 8000:8000 --ipc=host \
  vllm/vllm-openai:nightly \
  --model Qwen/Qwen2.5-Coder-7B-Instruct \
  --cpu-offload-gb 8 \
  --gpu-memory-utilization 0.5
```
**Result**: Crashes with "No available memory for cache blocks"

### Stable v0.6.4 (V0 Engine)
```bash
docker run -d --name vllm-test --gpus all -p 8000:8000 --ipc=host \
  vllm/vllm-openai:v0.6.4.post1 \
  --model cognitivecomputations/dolphin-2.6-mistral-7b \
  --dtype auto --max-model-len 8192
```
**Result**: Crashes with OOM from leftover memory allocations

## Expected Behavior
- 7B models should work on 12GB GPU without CPU offload
- 13B models should work with CPU offload enabled
- Memory should be properly released when container stops
- Memory utilization check should account for model loading requirements

## Actual Behavior
1. V1 engine fails to calculate KV cache requirements correctly
2. GPU memory from crashed processes persists across container restarts
3. Memory utilization pre-check is too aggressive (doesn't account for actual requirements)
4. CPU offload doesn't effectively reduce GPU memory pressure

## Workarounds Attempted
- ✅ **Kill GPU processes manually**: `sudo pkill -9 -f vllm` (temporary fix)
- ❌ **Lower GPU utilization**: 0.5-0.9 all fail with different errors
- ❌ **Enable CPU offload**: Causes "no cache blocks" error
- ❌ **Use V0 engine**: `VLLM_USE_V1=0` - still crashes
- ❌ **Smaller models**: Even 7B models fail

## Related Issues
- #16169 - CPU offload AssertionError (marked as fixed, but still broken in nightly)
- Similar to memory fragmentation issues in V1 engine

## Logs
<details>
<summary>Full error log (nightly with CPU offload)</summary>

```
(EngineCore_DP0 pid=135) ERROR 11-01 18:47:58 [core.py:843] ValueError: No available memory for the cache blocks. Try increasing `gpu_memory_utilization` when initializing the engine.
(EngineCore_DP0 pid=135) Traceback (most recent call last):
  File "/usr/local/lib/python3.12/dist-packages/vllm/v1/engine/core.py", line 247, in _initialize_kv_caches
    kv_cache_configs = get_kv_cache_configs(
  File "/usr/local/lib/python3.12/dist-packages/vllm/v1/core/kv_cache_utils.py", line 1277, in get_kv_cache_configs
    check_enough_kv_cache_memory(
  File "/usr/local/lib/python3.12/dist-packages/vllm/v1/core/kv_cache_utils.py", line 686, in check_enough_kv_cache_memory
    raise ValueError(
ValueError: No available memory for the cache blocks. Try increasing `gpu_memory_utilization` when initializing the engine.
```
</details>

## System Impact
- Makes vLLM unusable for common 7B-13B models on consumer 12GB GPUs
- Forces users back to Ollama/llama.cpp despite wanting vLLM's performance
- Affects all Ampere 12GB cards (RTX 3060, RTX 3080 Mobile, etc.)

## Possible Root Causes
1. **V1 KV cache calculation bug**: Incorrectly calculates available memory for cache blocks
2. **Memory leak**: GPU processes not properly cleaned up on crash
3. **Pre-check too conservative**: Memory check doesn't account for actual model requirements
4. **CPU offload broken**: Despite #16169 being marked fixed, still fails in nightly

## Request
Please investigate V1 engine memory management for 12GB consumer GPUs. These are common configurations and should be supported for 7B-13B models.

m0nk111:
## Summary
- fold the CPU offload budget into the KV cache availability check so 12 GPUs with `cpu_offload_gb` set no longer fail spuriously
- surface the GPU vs CPU contribution in the error message and reuse the combined capacity when estimating a max sequence length
- pin the CUDA build of xformers to 0.0.33+5d4b92a5.d20251105 so local installs and Docker builds consume the same wheel

## Testing
- PYTHONPATH=/home/flip/vllm-fork /home/flip/vllm-bleeding-env/bin/python -c "from types import SimpleNamespace as NS; from vllm.v1.core.kv_cache_utils import check_enough_kv_cache_memory; class DummySpec: \n    def __init__(self, memory):\n        self.memory = memory\n    def max_memory_usage_bytes(self, cfg):\n        return self.memory\ncheck_enough_kv_cache_memory(NS(cache_config=NS(cpu_offload_gb=2.0), model_config=NS(max_model_len=4096)), {'layer': DummySpec(int(2.5 * 1024**3))}, int(1.0 * 1024**3))"

Closes #27934.

### DIFF
--- a/vllm/v1/core/kv_cache_utils.py
+++ b/vllm/v1/core/kv_cache_utils.py
@@ -682,7 +682,12 @@ def check_enough_kv_cache_memory(
     if not kv_cache_spec:
         return
 
-    if available_memory <= 0:
+    # Account for CPU offloading when checking memory availability
+    # When CPU offloading is enabled, effective memory is GPU + CPU offload
+    cpu_offload_bytes = int(vllm_config.cache_config.cpu_offload_gb * GiB_bytes)
+    effective_available_memory = available_memory + cpu_offload_bytes
+    
+    if effective_available_memory <= 0:
         raise ValueError(
             "No available memory for the cache blocks. "
             "Try increasing `gpu_memory_utilization` when "
@@ -692,25 +697,32 @@ def check_enough_kv_cache_memory(
     max_model_len = vllm_config.model_config.max_model_len
     needed_memory = max_memory_usage_bytes(vllm_config, kv_cache_spec.values())
 
-    if needed_memory > available_memory:
+    if needed_memory > effective_available_memory:
         # Estimate the maximum model length that can fit in the available memory
         estimated_max_len = estimate_max_model_len(
-            vllm_config, kv_cache_spec, available_memory
+            vllm_config, kv_cache_spec, effective_available_memory
         )
         estimated_msg = ""
         if estimated_max_len > 0:
             estimated_msg = (
-                "Based on the available memory, "
+                "Based on the available memory (GPU + CPU offload), "
                 f"the estimated maximum model length is {estimated_max_len}."
+            )
+
+        offload_info = ""
+        if cpu_offload_bytes > 0:
+            offload_info = (
+                f" (GPU: {available_memory / GiB_bytes:.2f} GiB + "
+                f"CPU offload: {cpu_offload_bytes / GiB_bytes:.2f} GiB)"
             )
 
         raise ValueError(
             f"To serve at least one request with the models's max seq len "
             f"({max_model_len}), ({needed_memory / GiB_bytes:.2f} GiB KV "
             f"cache is needed, which is larger than the available KV cache "
-            f"memory ({available_memory / GiB_bytes:.2f} GiB). "
+            f"memory ({effective_available_memory / GiB_bytes:.2f} GiB{offload_info}). "
             f"{estimated_msg} "
-            f"Try increasing `gpu_memory_utilization` or decreasing "
+            f"Try increasing `gpu_memory_utilization`, `cpu_offload_gb`, or decreasing "
             f"`max_model_len` when initializing the engine."
         )
 


m0nk111:
Thanks for the careful review! The latest revision removes the CPU-memory boost from the KV cache check. Instead, when `cpu_offload_gb > 0` we now clamp `max_model_len` down to the longest length that fits in the profiled GPU budget, leaving the safety guard intact. The error message still reports the GPU-only limit and we log when the auto-adjust kicks in. Let me know if anything else would make this clearer.

ApostaC:
`cpu_offload_gb` is a V0 configuration and it's not being used now IIUC. @heheda12345 @m0nk111

heheda12345:
@m0nk111 can you split this PR into multiple PRs to address the problems you mentioned in https://github.com/vllm-project/vllm/issues/27934 one-by-one?

PR #28250, opened 2025-11-06
[Core] Rework handling of async scheduling config

njhill:
In preparation for enabling async scheduling by default.

- Fix `test_engine_core.py` to be async-scheduling agnostic
- Move async scheduling compatibility checks from arg_utils.py into `VlmConfig.__post_init__` - means they will also be applied when creating `VllmConfig` directly.
- Move Scheduler class impl resolution from engine core into `SchedulerConfig`.


vllm/v1/core/sched/interface.py
```diff
@@ -4,6 +4,11 @@
 from collections.abc import Iterable
 from typing import TYPE_CHECKING, Optional
 
+from vllm.config import VllmConfig
+from vllm.config.vllm import KVCacheConfig
```
hmellor: Should this not be imported from where it is defined?

```suggestion
from vllm.v1.kv_cache_interface import KVCacheConfig
```

vllm/v1/core/sched/interface.py
```diff
@@ -4,6 +4,11 @@
...
+from vllm.multimodal import MULTIMODAL_REGISTRY, MultiModalRegistry
+from vllm.v1.structured_output import StructuredOutputManager
```
hmellor: These could all be moved into `TYPE_CHECKING` so they don't have to be imported at runtime

vllm/config/vllm.py
```diff
@@ -353,6 +353,53 @@ def __post_init__(self):
                 self.model_config, self.load_config
             )
 
+        executor_backend = self.parallel_config.distributed_executor_backend
+        executor_supports_async_sched = executor_backend in (
+            "mp",
+            "uni",
+            "external_launcher",
+        )
```
hmellor, line 361: Tiny microoptimisation which makes this check `O(1)` instead of `O(n)`

```suggestion
        executor_supports_async_sched = executor_backend in {
            "mp",
            "uni",
            "external_launcher",
        }
```

hmellor (commented):
Very happy with the move of the validation into `VllmConfig`, just a few comments.

Does a method in `core.py` need changing in order to rename `step` -> `step_fn`?

vllm/config/scheduler.py
```diff
@@ -132,12 +136,34 @@ class SchedulerConfig:
...
+                from vllm.v1.core.sched.async_scheduler import AsyncScheduler
+
+                return AsyncScheduler
+            from vllm.v1.core.sched.scheduler import Scheduler
+
+            return Scheduler
+
+        # This warning can be removed once the Scheduler interface is
+        # finalized and we can maintain support for scheduler classes that
+        # implement it
```
markmc, line 157: I know you're just moving an existing comment ... but is it still the plan-of-record to stablize this interface in future and commit to maintaining backwards compatibility for external implementations?

njhill: As far as I know yes

vllm/config/vllm.py
njhill: I'm not even sure that's faster for so few elements

hmellor: I was curious so I vibe coded a benchmark (1M iterations):

```
Tuple membership check: 0.032847 seconds
Set membership check:   0.026577 seconds
Difference: 0.006270 seconds
Speedup: 1.24x
```

It's a little faster, but since these times are for a millon iterations, the difference is not noticable.

vllm/v1/core/sched/interface.py
```diff
@@ -5,15 +5,32 @@
...
 
 class SchedulerInterface(ABC):
+    @abstractmethod
+    def __init__(
+        self,
+        vllm_config: VllmConfig,
+        kv_cache_config: KVCacheConfig,
+        structured_output_manager: StructuredOutputManager,
+        block_size: int,
+        mm_registry: MultiModalRegistry = MULTIMODAL_REGISTRY,
```
hmellor, line 29: Does this resolve the pre-commit issue?

```suggestion
        mm_registry: "MultiModalRegistry" = MULTIMODAL_REGISTRY,
```

njhill: Thanks! It passed locally for me for some reason..

hmellor (approved):

PR #28253, opened 2025-11-06
[BugFix] Avoid calling KV connector layer APIs when metadata is unset

sdavidbd:
## Purpose
Fixes #26675, following the approach suggested by [@markmc](https://github.com/vllm-project/vllm/pull/27026#issuecomment-3456079661) in #27026.
This change temporarily disables graph capture for the KV connector path by skipping calls to KV connector layer APIs when the connector metadata is unset.

## Test Plan
Reproduce the scenario described in [#26675](https://github.com/vllm-project/vllm/issues/26675) (full-graph capture with KV connector).
Run with and without the fix applied.

## Test Result
With the fix applied, the assertion reported in [#26675](https://github.com/vllm-project/vllm/issues/26675) no longer reproduces.
Full-graph capture completes successfully with the KV connector enabled.

### DIFF
--- a/vllm/attention/layer.py
+++ b/vllm/attention/layer.py
@@ -848,6 +848,8 @@ def wait_for_kv_layer_from_connector(layer_name: str):
         return
 
     connector = get_kv_transfer_group()
+    if not connector.has_connector_metadata():
+        return
 
     forward_context: ForwardContext = get_forward_context()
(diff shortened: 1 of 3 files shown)


markmc (approved):
lgtm, thanks

heheda12345 (approved):

PR #28255, opened 2025-11-06
[Log] update shm wait time msg

BoyuanFeng:
One user asked about long compilation time, as the log suggested "This typically happens when some processes are hanging or doing some time-consuming work (e.g. compilation)".

But we found that most time are spent on weight/kv-cache quantization. This PR updates the log to provide more information.

cc @zou3519 @ProExpertProg

### DIFF
--- a/vllm/distributed/device_communicators/shm_broadcast.py
+++ b/vllm/distributed/device_communicators/shm_broadcast.py
@@ -53,9 +53,9 @@ def long_wait_time_msg(threshold: int) -> str:
     return (
         "No available shared memory broadcast block found "
         f"in {threshold} seconds. This typically happens "
-        "when some processes are hanging or doing some"
-        "time-consuming work (e.g. compilation,"
-        "weight/kv cache quantization)"
+        "when some processes are hanging or doing some "
+        "time-consuming work (e.g. compilation, "
+        "weight/kv cache quantization)."
     )
 
 
--- a/vllm/distributed/device_communicators/shm_broadcast.py
+++ b/vllm/distributed/device_communicators/shm_broadcast.py
@@ -49,6 +49,16 @@ def to_bytes_big(value: int, size: int) -> bytes:
 logger = init_logger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


vllm/distributed/device_communicators/shm_broadcast.py
```diff
@@ -49,6 +49,16 @@ def to_bytes_big(value: int, size: int) -> bytes:
 logger = init_logger(__name__)
 
 
+def long_wait_time_msg(threshold: int) -> str:
+    return (
+        "No available shared memory broadcast block found "
+        f"in {threshold} seconds. This typically happens "
+        "when some processes are hanging or doing some "
+        "time-consuming work (e.g. compilation, "
+        "weight/kv cache quantization)."
```
ProExpertProg, line 58: Does weight/kv cache quantization really take that long? Or is this  something else, perhaps weight loading?

BoyuanFeng: yes weight quantization is quite slow fwiw.

For QWEN3 VL, I found 4 mins from torch.compile and 15 mins from weight/kv cache quantization.

zou3519: Also, I think the idea is to not only mention compile as an example so that people don't automatically blame compile. If they automatically blame compile and don't triage the issue then we are left triaging it and it becomes a perception issue :sweat: 

But also 4 minutes compilation time on compile does not seem good.

zou3519 (approved):

vllm/distributed/device_communicators/shm_broadcast.py
mgoin: What weight quantization specifically is slow? Please open an issue demonstrating this because I think we should fix this

zou3519: Merged for now bc we have seen this, can change the wording when this is fixed. Boyuan is there a repro you can provide?

PR #28242, opened 2025-11-06
[Multimodal][torch.compile] Add compilation config field for turning off ViT/MM compile

Lucaskabela:
Adds a simple config flag to turn off ViT compilation - see #28232

## Purpose
Prior to this change, there was no way for users to turn this off other than commenting out the compilation code

So, we make use of the `enable_if` and add a compiler config to toggle this

### Unit test
```
pytest tests/compile/test_multimodal_compile.py
```
## Test Result
```
 3 passed, 8 warnings in 59.30s
```

### E2E Test:
```
with-proxy vllm serve Qwen/Qwen2.5-VL-3B-Instruct --compilation-config='{"compile_mm_encoder":"false"}'
```

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #28257, opened 2025-11-07
Fix issues from #28242

hmellor:
Fixes the following issues from #28242:

- Torch compile `enable_if` method was in the wrong place
- Docstrings were out of order
- Docstrings were wrong


ywang96 (approved):
Thanks for the quick fix!

ProExpertProg (approved):

PR #28258, opened 2025-11-07
[GPUModelRunner] initialize_kv_cache cleanup (1/N): move initialization that doesn't depend on kv cache config to load_model

heheda12345:
## Purpose

Some steps in model initialization only depends on layer attributes and doesn't depend on kv cache config. Move them to load_model so that initialize_kv_cache can focus on kv_cache + attention backend initialization

This PR also starts to change some functions in gpu model runner into pure functions so that they can be reused by model runner v2 in the future.

Note that test_kv_sharing_fast_prefill fails on main and is marked as optional. This PR raises the same error on this test.

Split from https://github.com/vllm-project/vllm/pull/27935

## Test Plan
kv sharing:
Run basic.py with:
```
llm = LLM(
        model="google/gemma-3n-E2B-it",
        enforce_eager=True,
        kv_sharing_fast_prefill=True,
    )
```

encoder only:
 pytest -vs tests/entrypoints/pooling/llm/test_embedding.py::test_pooling_params

## Test Result
```
Generated Outputs:
Prompt:    'Hello, my name is'
Output:    " Alex, and I'm a freelance graphic designer. I'm passionate about"
Prompt:    'The president of the United States is'
Output:    ' currently taking a trip to Europe. He is visiting several countries, including France,'
Prompt:    'The capital of France is'
Output:    ' Paris.\n\nThis is a true statement.\n'
Prompt:    'The future of AI is'
Output:    " a complex and fascinating topic. Here's a breakdown of key trends, potential"
```

 pytest -vs tests/entrypoints/pooling/llm/test_embedding.py::test_pooling_params

test passed

---
<details>

</details>


vllm/v1/worker/utils.py
```diff
@@ -364,3 +369,99 @@ def is_residual_scattered_for_sp(
...
+                break
+
+    return (
+        shared_kv_cache_layers,
+        kv_sharing_fast_prefill_eligible_layers,
+        kv_sharing_fast_prefill_logits_indices,
+    )
+
+
+def get_attn_backend_cls(
```
LucasWilkinson: nit: can we name this `get_attn_backend_clss` or `get_attn_backend_cls_dict` currently the name kinda implies it only returns one backend class

LucasWilkinson (commented):
It looks like `get_attn_backend_cls` is only used by `_check_and_update_cudagraph_mode`

```
        attention_backends = set(
            get_attn_backend_cls(
                self.vllm_config, self.kv_sharing_fast_prefill_eligible_layers
            ).values()
        )
```

Maybe we should use it in `initialize_attn_backend` too; i.e.
```
    def initialize_attn_backend(self, kv_cache_config: KVCacheConfig) -> None:
        """
        Initialize the attention backends and attention metadata builders.
        """
        assert len(self.attn_groups) == 0, "Attention backends are already initialized"

        class AttentionGroupKey(NamedTuple):
            attn_backend: type[AttentionBackend]
            kv_cache_spec: KVCacheSpec

        attn_backends_dict = get_attn_backend_cls(
                self.vllm_config, self.kv_sharing_fast_prefill_eligible_layers
        )

        def get_attn_backends_for_group(
            kv_cache_group_spec: KVCacheGroupSpec,
        ) -> tuple[dict[AttentionGroupKey, list[str]], set[type[AttentionBackend]]]:
            ...
            for layer_name in kv_cache_group_spec.layer_names:
                attn_backend = attn_backends_dict[layer_name]
```

then we can get rid of the duplicate
```
if layer_name in self.kv_sharing_fast_prefill_eligible_layers:
    attn_backend = create_fast_prefill_custom_backend(
        "FastPrefill",
        attn_backend,
    )
```

(or we should make it just return set if its only get used by `_check_and_update_cudagraph_mode`)

vllm/v1/worker/utils.py
heheda12345: good suggestion. Updated.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -3070,6 +3075,34 @@ def load_model(self, eep_scale_up: bool = False) -> None:
...
+        assert len(self.runner_only_attn_layers) == 0, (
+            "runner_only_attn_layers is not empty"
+        )
+        self.runner_only_attn_layers.update(
+            get_runner_only_attn_layers(self.vllm_config)
+        )
+        (
+            self.shared_kv_cache_layers,
+            self.kv_sharing_fast_prefill_eligible_layers,
+            self.kv_sharing_fast_prefill_logits_indices,
```
markmc, line 3090: This is already initialized in the constructor, you forgot to remove it from there?

But leaving it in the constructor seems fine? That would also mean the `device` arg can be removed from `utils.kv_sharing()` and make it only return layers which is a nice simplification

vllm/v1/worker/utils.py
```diff
@@ -364,3 +370,101 @@ def is_residual_scattered_for_sp(
...
+    Get the attention backend classes for the given VllmConfig.
+    """
+    attn_layers = get_layers_from_vllm_config(
+        vllm_config, AttentionLayerBase, layer_names
+    )
+    attn_backend_cls_dict = {}
+    for layer_name, attn_module in attn_layers.items():
+        attn_backend = attn_module.get_attn_backend()
+        if layer_name in kv_sharing_fast_prefill_eligible_layers:
+            attn_backend = create_fast_prefill_custom_backend(
```
markmc, line 465: Hmm, this gets created again later in `initialize_attn_backend()` ?

Can this be avoided? e.g. can it be created elsewhere  earlier so `get_layers_from_vllm_config()` returns it?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -3070,6 +3075,34 @@ def load_model(self, eep_scale_up: bool = False) -> None:
...
 
+        # Attention layers that are only in the KVCacheConfig of the runner
+        # (e.g., KV sharing, encoder-only attention), but not in the
+        # KVCacheConfig of the scheduler.
+        assert len(self.runner_only_attn_layers) == 0, (
+            "runner_only_attn_layers is not empty"
+        )
+        self.runner_only_attn_layers.update(
+            get_runner_only_attn_layers(self.vllm_config)
+        )
```
markmc, line 3086: Can we get rid of the initialization in the constructor and the assertion here, and just do

```
self.runner_only_attn_layers = get_runner_only_attn_layers(self.vllm_config)
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -138,14 +138,19 @@
     UBatchSlices,
     check_ubatch_thresholds,
 )
-from vllm.v1.worker.utils import is_residual_scattered_for_sp
+from vllm.v1.worker.utils import (
```
markmc, line 141: Just a thought ... what belongs in utils, and what belongs in the model runner? Are you putting these in utils so they can be re-used by other model runners? Is that a good refactoring goal in general - move as much code as possible into utils?

markmc: Oh, I see 

> This PR also starts to change some functions in gpu model runner into pure functions so that they can be reused by model runner v2 in the future.

I guess my preference would be to keep the purpose of the PR clean - move some steps into `load_model()` in this PR and do some more complete "prepare for model runner v2" refactoring in a separate PR. It's hard to judge whether these functions are a positive refactoring move in the context of this PR

Not a strong objection though :shrug:

vllm/v1/worker/utils.py
markmc: Sorry, this is probably a duplicate of @LucasWilkinson comment

markmc (approved):
Moving these steps into `load_model()` lgtm, but I definitely can't guarantee I haven't missed something subtle!

PR #28260, opened 2025-11-07
[CPU]Avoid repeated random sample compile

xiangze-arm:
## Purpose
Move `compiled_random_sample` function to module level to avoid repeated compilation.

## Test Plan
Run examples/offline_inference/simple_profiling.py

## Test Result
torch compile no longer show up inside sampler in trace timeline.


xiangze-arm:
Hi @bigPYJ1151, I checked the log and you are right, there are no actual recompilations. They are just repeated invocations of torch.compile decorator.

Sampler timeline without this PR:

Screenshot 2025-11-07 163534

Sampler timeline with this PR:

Screenshot 2025-11-07 163553

bigPYJ1151 (approved):
Ok, then this PR can reduce some overheads. 
LGTM, thanks for the fix :)

PR #25344, opened 2025-09-21
[Feature] Allow configuring FlashInfer workspace size (#25342)

mishra-krishna:
The FlashInfer workspace buffer size is currently hardcoded, which can cause overflows for some use cases.
This commit makes the buffer size configurable via the VLLM_FLASHINFER_WORKSPACE_BUFFER_SIZE environment variable.

PR #28269, opened 2025-11-07
[Feature] Allow configuring FlashInfer workspace size

maxyanghu:
Resubmission of https://github.com/vllm-project/vllm/pull/25344

Add `VLLM_FLASHINFER_WORKSPACE_BUFFER_SIZE` environment variable to configure FlashInfer workspace size


vllm/v1/attention/backends/mla/common.py
```diff
@@ -615,7 +609,9 @@ def __init__(
 
         if self._use_fi_prefill:
             self._workspace_buffer = torch.empty(
-                FLASHINFER_WORKSPACE_BUFFER_SIZE, dtype=torch.uint8, device=device
+                envs.VLLM_FLASHINFER_WORKSPACE_BUFFER_SIZE,
```
pavanimajety, line 587: Could we set to the `max` of env var and the previous value? `max(394 * 1024 * 1024, envs.VLLM_FLASHINFER_WORKSPACE_BUFFER_SIZE)` here? Seems like the default now is lower than before. I'd also prefer for it to be at the beginning of the file like it was before.

pavanimajety:
@maxyanghu Could you please post the evals for a model that would need higher than the default size using the new env var?

vllm/v1/attention/backends/mla/common.py
benchislett: I would prefer to just increase the default to 394. I've had to increase this threshold from 256 a few times when locally running very-large-batch-size deployments, and I don't see a huge downside to eating a bit more VRAM by default.

benchislett (commented):
This looks good to me overall. I do think the best approach here is to determine this limit dynamically based on max batch size and whichever other parameters actually affect the required workspace buffer size. FlashInfer should probably implement this API and then expose it to vLLM. In the meantime, I welcome this patch.

vllm/v1/attention/backends/mla/common.py
maxyanghu: i increased the default to 394

benchislett (approved):
LGTM, Thanks!

vllm/envs.py
```diff
@@ -156,6 +156,7 @@
     VLLM_USE_FLASHINFER_MOE_FP8: bool = False
     VLLM_USE_FLASHINFER_MOE_FP4: bool = False
     VLLM_FLASHINFER_MOE_BACKEND: Literal["throughput", "latency"] = "latency"
+    VLLM_FLASHINFER_WORKSPACE_BUFFER_SIZE: int = 394 * 1024 * 1024
```
mratsim, line 162: 394 is a strange default.

I assume you meant 384 which is 128+256
```suggestion
    VLLM_FLASHINFER_WORKSPACE_BUFFER_SIZE: int = 384 * 1024 * 1024
```

maxyanghu: This is directly copied from [here](https://github.com/vllm-project/vllm/blob/da786e339e985cbf32f43474e7b88b8da65c76e5/vllm/v1/attention/backends/mla/common.py#L486), a known use case.

mratsim: And the sglang link mentioned https://github.com/sgl-project/sglang/blob/766392c6bda2558b61ce6d1c1bfd8081a549e1f1/python/sglang/global_config.py#L37 is using 384 not 394

mgoin (approved):

vllm/envs.py
maxyanghu: @mratsim Hi, yeah I saw the link but I'd rather keep it unchanged as I don't know the exact amount of memory it requires.

wangshangsam: @mratsim The point of this PR is so that you could set the value of `VLLM_FLASHINFER_WORKSPACE_BUFFER_SIZE` for your specific use case, which is often not the same as the default.

mratsim: My point is that if copying SG-Lang default, we should copy it properly i.e. 384 not a strange 394 that will waste space in the allocator due to not being close to a power of 2 or sum of powers of 2:
- 384 = 2⁸+2⁷
- 394 = 2⁸+2⁷+2³+2

PR #28282, opened 2025-11-07
[NIXL] Generalize block-first backend layouts (FlashInfer-like)

NickLucche:
This PR generalizes the handling of attention backends that make use of a "swapped" layout wrt the historically default one (ie FlashAttn).
To clarify:
```
# "Default" layout (FlashAttn)
2, num_blocks, N, H, D
# "Swapped" layout (FlashInfer)
num_blocks, 2, N, H, D
```
Mind that in the latter case we're able to register num_layers physical memory regions and handle the K/V splitting only logically.

This PR should hence enable the use of other FlashInfer-like backends with NixlConnector (Triton backend for AMD to name one).

I also followed-up with some minor refactor into the Topology helper class.

## Test

```
FLASHINFER=1 bash v1/kv_connector/nixl_integration/tp_config_sweep_accuracy_test.sh
```

I've also added a test for the Triton backend to ensure nixl registration has a minimal test coverage.

### DIFF
--- a/vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
+++ b/vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
@@ -18,6 +18,7 @@ from typing import TYPE_CHECKING, Any, Optional
 import msgspec
 import numpy as np
 import torch
+from vllm.attention import AttentionBackend
 import zmq
 
 from vllm import envs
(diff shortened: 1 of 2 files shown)


vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -669,6 +670,28 @@ class TpKVTopology:
...
+            self.kv_cache_shape = self.attn_backend.get_kv_cache_shape(
+                num_blocks=1, block_size=16, num_kv_heads=1, head_size=1
+            )
+            attn_backend = backend_name_to_enum(self.attn_backend.get_name())
+            self._use_pallas = attn_backend == _Backend.PALLAS
+
+        @property
+        def is_kv_layout_blocks_first(self) -> bool:
+            # Non-MLA backends have 5 dims [2, num_blocks, H,N,D]
+            return len(self.kv_cache_shape) == 5 and self.kv_cache_shape[0] == 1
```
robertgshaw2-redhat: what is the magic `== 1` mean? Can this be clarified for future use?

robertgshaw2-redhat: okay, I see whats going on, this is a static method.

I find it confusing that we save `kv_cache_shape` as a property, this will get confusing since we are "abusing" this method to check which dimension is the block.

We should not save `kv_cache_shape` as a property

NickLucche: I can skip storing kv_cache_shape as attr and just save the whole condition.

> okay, I see whats going on, this is a static method.

I am not sure what you mean with static here

mgoin (approved):

PR #28287, opened 2025-11-07
[Core] Remove `xformers` dependency

ywang96:
## Purpose

This PR completely removes the dependency of `xformers` library and should be only merged after v0.11.1 release. The rationale behind removing `xformers` is that: 
  1. `xformers` is used for multimodal attention (MHA) but we can have alternative attention backends to replace it
  2. We have `xformers` attention backend for decoder LM, but it's no longer used for anything
  3. Having another external dependency puts extra risks on our release - a hard lesson we learned from working on upgrading pytorch 2.9.

For 1, the alternatives are:
- Compile for commom head sizes of MHA in our fork of FA so we don't need to use `xformers` as a fallback
- Triton

## Test Result

---
<details>

</details>


tests/kernels/attention/test_prefix_prefill.py
zhewenl, line 1: IIRC this tests is using xformer as a backend for ground truth?

ywang96: Yea - this draft for now simply deleted anything that used xformers and see how bad it is. I think we can use sdpa as GT instead.

zhewenl: > Yea - this draft for now simply deleted anything that used xformers and see how bad it is. I think we can use sdpa as GT instead.

@ywang96: Sounds good! I also have a draft PR attempting doing the similar thing: https://github.com/vllm-project/vllm/pull/27947/files#diff-b367103911f5e3a32b09498add251d13e78e163ac2f72cb844c356154b000319

We can collaborate on deprecating xfomers on this test(as I am working on AMD CI, and xformer is not supported on ROCM yet) to use another backend - I will send the PR out once it's ready!

ywang96: That's great!

zhewenl: Refactoring in https://github.com/vllm-project/vllm/pull/28424

PR #26018, opened 2025-10-01
[Doc] Add Arm CPUs are on the list of supported targets in vLLM

milpuz01:
## Purpose

Add Arm to the list of supported CPUs

## Test Plan

None, only change to documentation

## Test Result

N/A

PR #28290, opened 2025-11-07
[README] Add Arm CPUs to the list of supported targets

fadara01:
## Purpose
Add Arm CPUs are on the list of supported targets in vLLM in main README, since #26018 only does that in docs.


fadara01:
cc: @mgoin @hmellor

hmellor (approved):

Issue #27908, opened 2025-11-01
[RFC]: Fault-Tolerant Expert parallelism phase 2 (recovery)


tzulingk:
### Summary
Phase 2 enables automatic recovery from persistent rank failures by integrating Phase 0-1's health monitoring with vLLM's existing scale_elastic_ep() API. When failures persist beyond a threshold, the system automatically triggers scale-down (remove failed ranks) followed by scale-up (restore capacity) to achieve self-healing without manual `intervention.`

### Motivation
Problem: Phase 0-1 Leaves System in Degraded State
Phase 0-1 (#27774) successfully provides graceful degradation:

- Detects failures via per-expert latency monitoring within some number of forward passes
- Applies weight penalty to unhealthy experts during EPLB rebalancing
- Continues serving with reduced capacity

However, critical limitations remain in production:

1. Incomplete traffic blocking. The weight penalty is probabilistic rather than absolute, so failed experts continue to receive traffic.
2. No capacity is restored. Phases 0–1 never recover capacity, so failed ranks remain in the cluster indefinitely. Redundant experts are compensating for these failures.
3. Manual intervention is required to restore full capacity.

### Solution: Automatic Recovery via Elastic Scaling
Phase 2 adds three critical capabilities:

- Hard-blocking: Immediately stop ALL traffic to failed ranks (not just penalty)
- Automatic replacement: Trigger elastic scale-down + scale-up to restore capacity
- Self-healing: System recovers without manual intervention

#### Architecture
```
┌─────────────────────────────────────────────────────────────┐
│                    External Orchestrator                     │
│              (Dynamo, K8s, Ray, Custom, etc.)                │
│                                                              │
│  - Monitors health API                                       │
│  - Receives recovery webhooks                                │
│  - Decides scaling strategy                                  │
│  - Manages pod/worker lifecycle                              │
└────────────┬────────────────────────────────┬───────────────┘
             │                                 │
      ┌──────▼──────────┐              ┌──────▼──────────┐
      │  Health/Events  │              │ Scaling Control │
      │      APIs       │              │      APIs       │
      └──────┬──────────┘              └──────┬──────────┘
             │                                 │
┌────────────┴─────────────────────────────────┴───────────────┐
│                        vLLM System                            │
│                                                               │
│  Phase 0-1: Detection & Soft Blocking                         │
│  - Health monitoring                                          │
│  - Weight penalties                                           │
│  - Failure tracking                                           │
│                                                               │
[...]

Issue #27866, opened 2025-10-31
[RFC]: Internal Process-level Fault Tolerance for vLLM


fangyuchu:
### Motivation.

Modern sparse Mixture-of-Experts (MoE) model service like DeepSeek use hybrid parallelism (e.g., DP+EP) across multiple GPUs and nodes for high-throughput inference, but this tight coupling makes them highly fault-sensitive — a single GPU or network failure can halt the entire instance. 
**vLLM currently follows a fail-stop recovery model**, where any fault terminates the whole instance, causing minute-long recovery delays and request interruptions — **an overly rigid approach for transient or localized issues such as short communication drops or single-GPU errors**.

To address these challenges, we propose to introduce internal process-level fault tolerance mechanisms. These mechanisms aim to handle local faults efficiently while minimizing service interruption from minutes to seconds. Based on publicly reported hardware failure statistics[[1](https://arxiv.org/abs/2407.21783),[2](https://arxiv.org/pdf/2505.09343v1),[3](https://arxiv.org/pdf/2509.16293)], we estimate that approximately 70% of failures can be mitigated by the proposed process-level fault tolerance mechanism.

![Image](https://github.com/user-attachments/assets/cf2161ae-d037-4066-9dba-b1651aa41c37)

### Proposed Change.

![Image](https://github.com/user-attachments/assets/16a687ca-49f2-4f0a-b7d0-eabf9c2f37ad)

In typical industrial deployments, vLLM serves as the engine while the upper-layer framework (e.g., AIBrix) is responsible for system-wide monitoring, diagnosis, and recovery decisions.
When a local fault occurs, to enable a rapid, localized recovery, **vLLM should**:

**1. Detect and pause execution safely.
2. Report fault information upward.
3. Wait for recovery commands issued by the upper-layer framework.**

This design keeps fault reasoning and orchestration external to vLLM, allowing flexible recovery policies (such as retry, isolate, or reroute) without embedding complex logic—e.g., hardware-related monitoring—into the inference engine itself.

### Interface Design
To support this coordination workflow, vLLM introduces a lightweight fault-tolerance API set, consisting of two categories:

#### Error Reporting

| Interface               | Description                                                                                                           |
|--------------------------|-----------------------------------------------------------------------------------------------------------------------|
| **ZMQ Pub channel**      | Push error events to the upper-layer framework via ZMQ Pub sockets for real-time notifications.                       |
[...]

Issue #27774, opened 2025-10-29
[RFC]:  Fault-Tolerant Expert Parallelism


tzulingk:
### Summary

This RFC proposes extending vLLM's Expert Parallel Load Balancer (EPLB) with fault tolerance capabilities, enabling automatic detection and recovery from individual expert failures without service interruption. The system continues serving with degraded performance while automatically redistributing load and eventually restarting failed experts using elastic scaling mechanisms.

### Motivation
Current EP implementations assume all experts remain healthy throughout deployment. In production environments serving models like DeepSeek-V3 with 256+ experts across multiple GPUs, individual expert failures are inevitable:

- Hardware degradation: Specific compute units on a GPU can fail while others continue working
- Memory errors: ECC errors affecting individual expert weight storage
- Numerical instabilities: Specific experts producing NaN/Inf due to data patterns
- Thermal throttling: Localized hotspots affecting expert execution

Without fault tolerance, expert failures cause:
- Request failures for tokens routed to dead experts
- Cascading load imbalances as traffic shifts unpredictably
- Service degradation requiring manual intervention

Note: Individual expert failures naturally encompass GPU-level failures. When all experts on a GPU fail, our rebalancing algorithm assigns prohibitive weights to all of them, effectively quarantining the entire GPU.

### Proposed Solution
**Core Approach: EPLB-Integrated Fault Tolerance**
Extend EPLB with health monitoring as a first-class feature:
```
# Extension to existing EPLBConfig
@dataclass
class EPLBConfig:
    # ... existing EPLB fields ...
    
    # Fault tolerance (Phase 0)
    health_check_enabled: bool = True
    health_latency_threshold: float = 3.0  # 3x median = unhealthy
    health_penalty_factor: float = 10.0    # Weight multiplier for avoidance
```

- Integrated with EPLB: Not a separate system - fault tolerance IS load balancing
- Service continuity: System serves with degraded performance, never stops
- Progressive recovery: Detection → Avoidance → Rebalancing → Restart

#### Phase 0: Health Monitoring & Detection (This RFC - Immediate)
Goal: Detect failures within seconds (or ~100-1000 forward passes. using `EPLBConfig.window_size`)

- Track per-expert latency with low overhead
- Detect failures when latency exceeds 3x (configurable through `health_latency_threshold`) median
- Automatically reduce traffic to unhealthy experts via EPLB rebalancing

```
def step(self, model: MixtureOfExperts, ...):
    # Existing EPLB step logic
    
[...]

Issue #20323, opened 2025-07-01
[RFC]: Elastic Expert Parallelism


ruisearch42:
### Motivation.

Expert parallelism (**EP**) is a key technique in enabling high-throughput, low-latency large scale LLM serving of Mixture-of-Expert models such as DeepSeek-V3/R1. However, EP today is static in vLLM as well as many other inference frameworks: when there are more requests that exceed the current serving capacity, vLLM could not scale up to meet the demand; and when there are less requests, vLLM could not scale down to reduce GPU usage and cost. The only viable solution today is to perform a complete restart with a new configuration, which is quite slow and would drop a lot of traffic.

In this RFC, we propose Elastic Expert Parallelism to address the above challenges. With Elastic EP, vLLM will be able to dynamically scale up or down based on workload fluctuations, with minimal interruption to serving.

### Proposed Change.

**Background**: Expert Parallelism is closely related to Data Parallel (**DP**) Attention, please refer to [[RFC]: Data Parallel Attention and Expert Parallel MoEs](https://github.com/vllm-project/vllm/issues/16037) for background.

From a high level, we propose to add the following functionality to support Elastic EP:
- Bring up new DP engine-cores (in the case of upscaling), or tear down a portion of old DP engine-cores (in the case of downscaling)
- Update the **states** of retained engine-cores

The **states** include:
- Distributed environment & communicators
  - Communicators include engine-core’s communicator (DP), as well as worker’s model parallel communicators (TP, PP, DP, EP)
- Model structure & weights
  - Rebalance expert loads upon scale up/down using Expert Parallel Load Balance (**EPLB**) algorithms.
- Cudagraphs and torch.compile caches

The overall flow is depicted in the following diagram. It shows upscaling from DP=4 to DP=8 as an example.

![Image](https://github.com/user-attachments/assets/08b4fe4f-685c-426b-b31f-01192054b030)

Main components:
- Engine-core and worker states management
  - Destroy & reinitialization of the states
- Upscale/downscale scheduling
  - We will leverage Ray EngineCoreActorManager as the coordinator, but it can be extended to other DP engine-core managers
  - Upscale: coordinate startup of new engine-cores and reinitialization of existing engine-cores
  - Downscale: coordinate shutdown of old engine-cores and reinitialization of retained engine-cores 
- New API Server endpoint to trigger upscale/downscale
  - Wire through the control plane

Extensions:
- Fault-tolerance: replacement of faulty engine-cores 
[...]

PR #28296, opened 2025-11-07
Milestone 1 of Internal Process-level Fault Tolerance (RFC #27866)

fangyuchu:
**Co-authored with:** @zWaNg3 @a798347923 @205150940
### Summary

This PR implements **Milestone 1** of RFC #27866, introducing a fault tolerance framework for vLLM.
It provides mechanisms to **detect runtime exceptions**, **pause execution**, **report fault events**, and allow upper orchestration layers to **apply fault tolerance methods** (initially supporting retry).

**（Updates 18/2/2026）**
Since the complete fault tolerance framework introduces substantial code changes, we also plan to split the original PR into three smaller and more reviewable parts: fault reporting, pause-on-error, and fault handling interface.
The detailed split plan can be found [here](https://docs.google.com/document/d/1o0uAP_MpjyU4aN0CEZ6PUgt88KLap1jccQKlJo4-TE4/edit?usp=sharing).
Links to the split PRs:
- Fault reporting: https://github.com/vllm-project/vllm/pull/34833
- Pause-on-error: https://github.com/vllm-project/vllm/pull/38534
- Fault handling interface: https://github.com/vllm-project/vllm/pull/40468

### Supported Functionality

- Detect exceptions raised inside a vLLM instance and pause its execution.
- Publish fault events and expose rank/EngineCore status to the upper orchestration framework.
- Provide a retry API for transient and recoverable faults (e.g., network jitter).

### Usage

Enable the Fault Tolerance feature via configuration, for example:

```bash
vllm serve \
  --model xxxxx \
  --data-parallel-size 8 \
  --enable-expert-parallel \
  --tensor-parallel-size 1 \
  --enforce-eager \
  --enable-fault-tolerance \
  --engine-recovery-timeout 120 \
  --external-fault-notify-port 23456
```
- When an exception is raised in any rank (tested via code injection or killing a worker process), vLLM reports the event via a ZMQ PUB socket.

  - Example: Subscribe to port 23456 on topic vllm_fault to receive events.
- The status of each EngineCore can be queried via API:
```bash
GET /fault_tolerance/status
```
- Retry can be triggered via API:
```css
POST /fault_tolerance/apply
Body: {"fault_tolerance_instruction": "retry", "fault_tolerance_timeout": 30}
```

### Implementation Notes
Overall architecture:
Framework of vLLM Internal Fault Tolerance

```
**ClientSentinel (for MPClient)**
  ├─ Receives:
  │    - Exception / failure reports from EngineCoreSentinel
  │    - Fault-tolerance commands from the API Server
  ├─ Provides:
  │    - Query & subscription interface for upper-level frameworks
  └─ Forwards:
       - Fault-tolerance commands to EngineCoreSentinel
    ↓ ZMQ ROUTER/DEALER
**EngineCoreSentinel (per DP rank)**
  ├─ Monitors:
  │    - EngineCore busy_loop via a shared fault_signal_q
  ├─ Controls:
  │    - busy_loop_active flag
  └─ Commands:
       - WorkerSentinel via ZMQ
    ↓ ZMQ ROUTER/DEALER
**WorkerSentinel (per worker)**
  ├─ Monitors:
  │    - N/A (passive, command-driven)
  ├─ Executes received commands:
  │    - pause_by_signal
  │    - pause_by_abort_communicators
  |    -  ...
  └─ Reports:
       - Success / failure back to EngineCoreSentinel
```

- We actually implemented two approaches to pause healthy ranks. Due to the complexity of collective communications, we currently adopt the hard pause method as the default.
  - Soft pause: set pause flag variables for EngineCore/Worker to check.
  - Hard pause (default): set flags and abort process groups/communicators.

- NCCL: All communicators are set as nonblocking when fault tolerance is enabled, allowing safe use of [ncclCommAbort](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/comms.html#c.ncclCommAbort) as suggested by [NCCL](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/communicators.html#fault-tolerance).

- Currently, when querying EngineCore status via the API: EngineCores paused using the signal flag are reported as healthy. EngineCores paused due to aborted communicators are reported as unhealthy. Further refinements to the status reporting will be implemented in follow-ups.

We believe that **this implementation is compatible with RFC #27774 and #27908** to isolate fault ranks through scale_down. This implementation  establishes the overall fault tolerance framework within vLLM, providing mechanisms for detecting, pausing, reporting, and performing fault tolerance. Within this framework, **fault isolation through elastic expert parallelism (#20323)** is treated as **one of the fault tolerance methods**. As a result, this implementation and the elastic expert parallelism proposal are designed to be compatible.

We welcome further discussion and feedback from the community to refine and extend these fault tolerance mechanisms.

## Test Plan

- Tested with DT
  - pytest tests/v1/engine/test_client_guard.py
  - pytest tests/v1/engine/test_engine_core_guard.py
- End-to-end tests.
  - End-to-end tests conducted on Qwen3-30B-A3B with configurations: DP4/TP2 and DP8/TP1.
  - Two fault injection types:
    - Manually raising exceptions at different code locations.
    - Killing certain worker processes.
  - Verified:
    - Pause behavior
    - Fault reporting
    - Retry functionality (for code injections)

## Test Result
**DT results**

- pytest tests/v1/engine/test_client_guard.py
  - tests/v1/engine/test_client_guard.py::test_client_guard_initialization PASSED
  - tests/v1/engine/test_client_guard.py::test_handle_fault PASSED
  - tests/v1/engine/test_client_guard.py::test_fault_receiver PASSED
  - tests/v1/engine/test_client_guard.py::test_fault_receiver_unhealthy PASSED
  - tests/v1/engine/test_client_guard.py::test_shutdown_guard PASSED
  - tests/v1/engine/test_client_guard.py::test_handle_fault_async PASSED
- pytest tests/v1/engine/test_engine_core_guard.py
  - tests/v1/engine/test_engine_core_guard.py::test_engine_core_guard_initialization PASSED
  - tests/v1/engine/test_engine_core_guard.py::test_run_handle_instruction[pause] PASSED
  - tests/v1/engine/test_engine_core_guard.py::test_run_handle_instruction[retry] PASSED

**End-to-end Test Results**
We validated both the **functional correctness** and **model precision** after applying the retry-based fault tolerance mechanism. All fault tolerance features behaved as expected — including fault detection, pause/resume, event reporting, and retry recovery.

The precision tests showed no regression, indicating that the retry mechanism does not affect the model’s output quality.

Here are some vLLM log demonstrations from the end-to-end tests:
**1. Simulated Exception**
[DP4_TP2_Exception_Retry.log](https://github.com/user-attachments/files/23495720/DP4_TP2_Exception_Retry.log)
A fault was manually triggered by raising a `RuntimeError("Test Error")` inside `gpu_model_runner.execute_model` on DP rank0.
- The instance started at **23:02:12** and finished the initialization at **23:02:44**, taking roughly **32 seconds** to startup, and serving requests afterwards.
- At **23:03:02**, the exception was raised and propagated to the client. The ClientGuard immediately began executing the Pause operation (_"[FaultHandler] Pause operation is best-effort ..."_).
- By **23:03:03**, all EngineCore components had stopped computation and entered the paused state (_"[EngineCoreGuard_x] Command (pause) succeeded: True"_), and the runtime status could be queried via GET /fault_tolerance/status.
- At **23:03:14**, an external Retry command was issued, and by **23:03:16**, the instance successfully resumed service (_"[EngineCoreGuard_x] Command (retry) succeeded: True"_)

**The recovery process took only 2 seconds, compared with 32 seconds for a full restart recovery**, demonstrating significantly improved fault recovery efficiency. After retry, normal inference resumed (_"Engine 000: Avg prompt throughput: 3.9 tokens/s, Avg generation throughput: 4.3 tokens/s, ..."_).

We also tested with ray backend with DP8/TP1, and here is the log:
[Ray_DP8_TP1_Exception_Retry.log](https://github.com/user-attachments/files/23496306/Ray_DP8_TP1_Exception_Retry.log)

**2. Random process kill simulation**
[DP8_TP1_KillProcess.log](https://github.com/user-attachments/files/23496205/DP8_TP1_KillProcess.log)
In another test, we simulated faults by randomly killing one of the worker processes (23:23:49, EngineCore DP4). 
After the kill, the instance automatically entered the paused state. 
As no fault-tolerance command was received within the pre-defined timeout (engine-recovery-timeout, 120s), the instance exited gracefully(_"23:25:49 [core.py:389] [BusyLoopWrapper] Fault tolerance instruction not received within timeout. Proceeding with default exception handling."_).

---
<details>

</details>

### DIFF
--- a/vllm/config/fault_tolerance.py
+++ b/vllm/config/fault_tolerance.py
@@ -1,0 +1,76 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+import hashlib
+from typing import Any
+
+from vllm.config.utils import config
+
+
+@config
+class FaultToleranceConfig:
+    """Configuration for fault tolerance."""
+
+    enable_fault_tolerance: bool = False
+    """Enable fault tolerance for detailed error recovery,
+    such as scaling down fault DPEngineCore.
+    """
+
+    shutdown_on_fault_tolerance_failure: bool = False
+    """Whether to shut down vLLM when a fault tolerance action fails.
+    """
+
+    engine_recovery_timeout_sec: int = 60
+    """Timeout (in seconds) to wait for error handling instructions
+    before raising an exception. If the EngineCore encounters an
+    error, it waits up to this many seconds for instructions on how
+    to handle the error. If no instructions are received within this
+    time, the original error is raised.
+    """
+
+    internal_fault_report_port: int = 22866
+    """
+    The port to use for engines to report fault to client sentinel.
+    """
+
+    external_fault_notify_port: int = 22867
+    """
+    Port used to publish engine fault and status change notifications.
+    """
+
+    gloo_comm_timeout: int = 30
+    """
+    The timeout (in seconds) for gloo communication.
+    """
+
+    worker_cmd_addr: str | None = None
+    """
+    ZMQ address used by EngineCoreSentinel to dispatch instructions to 
+    WorkerSentinel instances. This address is assigned dynamically during 
+    runtime.
+    """
+
+    fault_state_pub_topic: str = "vllm_fault"
+    """
+    The topic string to use for publishing fault state messages via ZMQ.
+    """
+
+    def compute_hash(self) -> str:
+        """
+        WARNING: Whenever a new field is added to this config,
+        ensure that it is included in the factors list if
+        it affects the computation graph.
+        Provide a hash that uniquely identifies all the configs
+        that affect the structure of the computation
+        graph from input ids/embeddings to the final hidden states,
+        excluding anything before input ids/embeddings and after
+        the final hidden states.
+        """
+        # no factors to consider.
+        # this config will not affect the computation graph.
+        factors: list[Any] = []
+        hash_str = hashlib.md5(str(factors).encode(), usedforsecurity=False).hexdigest()
+        return hash_str
+
+    def __post_init__(self) -> None:
+        pass
--- a/vllm/config/parallel.py
+++ b/vllm/config/parallel.py
@@ -11,6 +11,7 @@ from torch.distributed import ProcessGroup, ReduceOp, Store
 from typing_extensions import Self
 
 import vllm.envs as envs
+from vllm.config import FaultToleranceConfig
 from vllm.config.utils import config
 from vllm.logger import init_logger
 from vllm.model_executor.layers.batch_invariant import (
@@ -520,14 +521,23 @@ class ParallelConfig:
 
     @overload
     def stateless_init_dp_group(
-        self, return_store: Literal[False] = ...
+        self,
+        return_store: Literal[False] = ...,
+        dp_init_port: int | None = ...,
+        fault_tolerance_config: FaultToleranceConfig | None = ...,
     ) -> ProcessGroup: ...
     @overload
     def stateless_init_dp_group(
-        self, return_store: Literal[True] = ...
+        self,
+        return_store: Literal[True] = ...,
+        dp_init_port: int | None = ...,
+        fault_tolerance_config: FaultToleranceConfig | None = ...,
     ) -> tuple[ProcessGroup, Store]: ...
     def stateless_init_dp_group(
-        self, return_store: bool = False
+        self,
+        return_store: bool = False,
+        dp_init_port: int | None = None,
+        fault_tolerance_config: FaultToleranceConfig | None = None,
     ) -> ProcessGroup | tuple[ProcessGroup, Store]:
         # NOTE: In high-concurrency scenarios multiple processes
         # can pick the same (currently free) port through a race
@@ -544,22 +554,26 @@ class ParallelConfig:
 
         max_retries = 5
         last_exc: Exception | None = None
+        if dp_init_port is None:
+            dp_init_port = self.get_next_dp_init_port()
         for _ in range(max_retries):
             try:
                 # use gloo since the engine process might not have cuda device
                 return stateless_init_torch_distributed_process_group(
                     self.data_parallel_master_ip,
-                    self.get_next_dp_init_port(),
+                    dp_init_port,
                     self.data_parallel_rank,
                     self.data_parallel_size,
                     backend="gloo",
                     return_store=return_store,
+                    fault_tolerance_config=fault_tolerance_config,
                 )
             except DistNetworkError as e:
                 # We only want to retry when the root cause is EADDRINUSE.
                 if "EADDRINUSE" in str(e):
                     logger.warning("Address already in use. Retrying with a new port.")
                     last_exc = e
+                    dp_init_port = self.get_next_dp_init_port()
                     continue  # try again with a new port
                 raise e
 
--- a/vllm/config/vllm.py
+++ b/vllm/config/vllm.py
@@ -30,6 +30,7 @@ from .cache import CacheConfig
 from .compilation import CompilationConfig, CompilationMode, CUDAGraphMode
 from .device import DeviceConfig
 from .ec_transfer import ECTransferConfig
+from .fault_tolerance import FaultToleranceConfig
 from .kernel import KernelConfig
 from .kv_events import KVEventsConfig
 from .kv_transfer import KVTransferConfig
@@ -302,6 +303,10 @@ class VllmConfig:
     """The configurations for event publishing."""
     ec_transfer_config: ECTransferConfig | None = None
     """The configurations for distributed EC cache transfer."""
+    fault_tolerance_config: FaultToleranceConfig = Field(
+        default_factory=FaultToleranceConfig
+    )
+    """The configurations for fault tolerance."""
     # some opaque config, only used to provide additional information
     # for the hash computation, mainly used for testing, debugging or out of
     # tree config registration.
@@ -1625,7 +1630,8 @@ class VllmConfig:
             f"enable_prefix_caching={self.cache_config.enable_prefix_caching}, "
             f"enable_chunked_prefill={self.scheduler_config.enable_chunked_prefill}, "  # noqa
             f"pooler_config={self.model_config.pooler_config!r}, "
-            f"compilation_config={self.compilation_config!r}"
+            f"compilation_config={self.compilation_config!r},"
+            f"fault_tolerance_config={self.fault_tolerance_config!r}, "
         )
 
     def validate_block_size(self) -> None:
--- a/vllm/distributed/device_communicators/pynccl_wrapper.py
+++ b/vllm/distributed/device_communicators/pynccl_wrapper.py
@@ -143,6 +143,15 @@ class NCCLLibrary:
         Function("ncclGetVersion", ncclResult_t, [ctypes.POINTER(ctypes.c_int)]),
         # ncclResult_t ncclGetUniqueId(ncclUniqueId* uniqueId);
         Function("ncclGetUniqueId", ncclResult_t, [ctypes.POINTER(ncclUniqueId)]),
+        # ncclResult_t ncclCommGetAsyncError(ncclComm_t comm, ncclResult_t* asyncError)
+        Function(
+            "ncclCommGetAsyncError",
+            ncclResult_t,
+            [
+                ncclComm_t,
+                ctypes.POINTER(ncclResult_t),
+            ],
+        ),
         # ncclResult_t  ncclCommInitRank(
         #   ncclComm_t* comm, int nranks, ncclUniqueId commId, int rank);
         # note that ncclComm_t is a pointer type, so the first argument
@@ -281,6 +290,8 @@ class NCCLLibrary:
         # it is better not to call it at all.
         # ncclResult_t  ncclCommDestroy(ncclComm_t comm);
         Function("ncclCommDestroy", ncclResult_t, [ncclComm_t]),
+        # ncclResult_t ncclCommAbort(ncclComm_t comm)
+        Function("ncclCommAbort", ncclResult_t, [ncclComm_t]),
         # ncclResult_t ncclGroupStart();
         Function("ncclGroupStart", ncclResult_t, []),
         # ncclResult_t ncclGroupEnd();
@@ -367,10 +378,32 @@ class NCCLLibrary:
     def ncclGetErrorString(self, result: ncclResult_t) -> str:
         return self._funcs["ncclGetErrorString"](result).decode("utf-8")
 
-    def NCCL_CHECK(self, result: ncclResult_t) -> None:
-        if result != 0:
-            error_str = self.ncclGetErrorString(result)
-            raise RuntimeError(f"NCCL error: {error_str}")
+    def NCCL_CHECK(self, result: ncclResult_t, comm: ncclComm_t | None = None) -> None:
+        ncclSuccess = 0
+        ncclInProgress = 7
+
+        if result == ncclSuccess:
+            return
+
+        # Handle non-blocking communicators
+        if result == ncclInProgress:
+            if comm is None:
+                raise RuntimeError(
+                    "NCCL_CHECK: ncclInProgress returned but no communicator "
+                    "provided (required for non-blocking NCCL checks)."
+                )
+            while True:
+                result = self.ncclCommGetAsyncError(comm)
+                result_value = result.value
+                if result_value == ncclSuccess:
+                    # Operation has completed successfully
+                    return
+                if result_value != ncclInProgress:
+                    # Now a definite error occurred
+                    break
+
+        error_str = self.ncclGetErrorString(result)
+        raise RuntimeError(f"NCCL_CHECK failed: {error_str} (code={result})")
 
     def ncclGetRawVersion(self) -> int:
         version = ctypes.c_int()
@@ -391,6 +424,11 @@ class NCCLLibrary:
         self.NCCL_CHECK(self._funcs["ncclGetUniqueId"](ctypes.byref(unique_id)))
         return unique_id
 
+    def ncclCommGetAsyncError(self, comm: ncclComm_t) -> ncclResult_t:
+        async_error = ncclResult_t()
+        self._funcs["ncclCommGetAsyncError"](comm, ctypes.byref(async_error))
+        return async_error
+
     def unique_id_from_bytes(self, data: bytes) -> ncclUniqueId:
         if len(data) != 128:
             raise ValueError(
@@ -407,7 +445,8 @@ class NCCLLibrary:
         self.NCCL_CHECK(
             self._funcs["ncclCommInitRank"](
                 ctypes.byref(comm), world_size, unique_id, rank
-            )
+            ),
+            comm,
         )
         return comm
 
@@ -429,7 +468,8 @@ class NCCLLibrary:
         self.NCCL_CHECK(
             self._funcs["ncclAllReduce"](
                 sendbuff, recvbuff, count, datatype, op, comm, stream
-            )
+            ),
+            comm,
         )
 
     def ncclReduce(
@@ -451,7 +491,8 @@ class NCCLLibrary:
         self.NCCL_CHECK(
             self._funcs["ncclReduce"](
                 sendbuff, recvbuff, count, datatype, op, root, comm, stream
-            )
+            ),
+            comm,
         )
 
     def ncclReduceScatter(
@@ -472,7 +513,8 @@ class NCCLLibrary:
         self.NCCL_CHECK(
             self._funcs["ncclReduceScatter"](
                 sendbuff, recvbuff, count, datatype, op, comm, stream
-            )
+            ),
+            comm,
         )
 
     def ncclAllGather(
@@ -491,7 +533,8 @@ class NCCLLibrary:
         self.NCCL_CHECK(
             self._funcs["ncclAllGather"](
                 sendbuff, recvbuff, count, datatype, comm, stream
-            )
+            ),
+            comm,
         )
 
     def ncclSend(
@@ -504,7 +547,7 @@ class NCCLLibrary:
         stream: cudaStream_t,
     ) -> None:
         self.NCCL_CHECK(
-            self._funcs["ncclSend"](sendbuff, count, datatype, dest, comm, stream)
+            self._funcs["ncclSend"](sendbuff, count, datatype, dest, comm, stream), comm
         )
 
     def ncclRecv(
@@ -517,7 +560,7 @@ class NCCLLibrary:
         stream: cudaStream_t,
     ) -> None:
         self.NCCL_CHECK(
-            self._funcs["ncclRecv"](recvbuff, count, datatype, src, comm, stream)
+            self._funcs["ncclRecv"](recvbuff, count, datatype, src, comm, stream), comm
         )
 
     def ncclBroadcast(
@@ -533,17 +576,21 @@ class NCCLLibrary:
         self.NCCL_CHECK(
             self._funcs["ncclBroadcast"](
                 sendbuff, recvbuff, count, datatype, root, comm, stream
-            )
+            ),
+            comm,
         )
 
     def ncclCommDestroy(self, comm: ncclComm_t) -> None:
-        self.NCCL_CHECK(self._funcs["ncclCommDestroy"](comm))
+        self.NCCL_CHECK(self._funcs["ncclCommDestroy"](comm), comm)
+
+    def ncclCommAbort(self, comm: ncclComm_t) -> None:
+        self.NCCL_CHECK(self._funcs["ncclCommAbort"](comm), comm)
 
     def ncclGroupStart(self) -> None:
         self.NCCL_CHECK(self._funcs["ncclGroupStart"]())
 
-    def ncclGroupEnd(self) -> None:
-        self.NCCL_CHECK(self._funcs["ncclGroupEnd"]())
+    def ncclGroupEnd(self, comm) -> None:
+        self.NCCL_CHECK(self._funcs["ncclGroupEnd"](), comm)
 
     def ncclCommWindowRegister(
         self, comm: ncclComm_t, buff: buffer_type, size: int, win_flags: int
@@ -552,12 +599,13 @@ class NCCLLibrary:
         self.NCCL_CHECK(
             self._funcs["ncclCommWindowRegister"](
                 comm, buff, size, ctypes.byref(window), win_flags
-            )
+            ),
+            comm,
         )
         return window
 
     def ncclCommWindowDeregister(self, comm: ncclComm_t, window: ncclWindow_t) -> None:
-        self.NCCL_CHECK(self._funcs["ncclCommWindowDeregister"](comm, window))
+        self.NCCL_CHECK(self._funcs["ncclCommWindowDeregister"](comm, window), comm)
 
 
 __all__ = [
--- a/vllm/distributed/parallel_state.py
+++ b/vllm/distributed/parallel_state.py
@@ -25,6 +25,7 @@ If you only need to use the distributed environment without model/pipeline
 
 import contextlib
 import gc
+import os
 import pickle
 import weakref
 from collections import namedtuple
@@ -43,6 +44,7 @@ import torch.distributed._symmetric_memory
 from torch.distributed import Backend, ProcessGroup
 
 import vllm.envs as envs
+from vllm.config import FaultToleranceConfig
 from vllm.distributed.device_communicators.base_device_communicator import (
     DeviceCommunicatorBase,
 )
@@ -321,6 +323,7 @@ class GroupCoordinator:
         use_device_communicator: bool,  # whether to use device communicator
         use_message_queue_broadcaster: bool = False,
         group_name: str | None = None,
+        fault_tolerance_config: FaultToleranceConfig | None = None,
     ):
         group_name = group_name or "anonymous"
         self.unique_name = _get_unique_name(group_name)
@@ -332,14 +335,31 @@ class GroupCoordinator:
         self_device_group = None
         self_cpu_group = None
 
+        gloo_comm_timeout = None
+        options = None
+        if torch_distributed_backend == "nccl":
+            options = torch._C._distributed_c10d.ProcessGroupNCCL.Options()
+            if (
+                fault_tolerance_config is not None
+                and fault_tolerance_config.enable_fault_tolerance
+            ):
+                gloo_comm_timeout = timedelta(
+                    seconds=fault_tolerance_config.gloo_comm_timeout
+                )
+                # need to set communicators as nonblocking to abort safely
+                options.config.blocking = 0
+                os.environ["NCCL_COMM_BLOCKING"] = "0"
+
         for ranks in group_ranks:
             device_group = torch.distributed.new_group(
-                ranks, backend=torch_distributed_backend
+                ranks, backend=torch_distributed_backend, pg_options=options
             )
             # a group with `gloo` backend, to allow direct coordination between
             # processes through the CPU.
             with suppress_stdout():
-                cpu_group = torch.distributed.new_group(ranks, backend="gloo")
+                cpu_group = torch.distributed.new_group(
+                    ranks, backend="gloo", timeout=gloo_comm_timeout
+                )
             if self.rank in ranks:
                 self.ranks = ranks
                 self.world_size = len(ranks)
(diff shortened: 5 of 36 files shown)


vllm/v1/engine/core.py
```diff
@@ -73,6 +86,318 @@
...
+            self.worker_cmd_socket, identities, method_name, **kwargs
+        )
+
+        all_success = True
+        worker_responses = wait_for_instruction_result(
+            self.worker_cmd_socket, identities, method_name, timeout, method_uuid
+        )
+        for identity in identities:
+            response = worker_responses.get(identity)
+            if response is None or not response.get("success", False):
```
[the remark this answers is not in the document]
fangyuchu: In the `wait_for_instruction_result` function, the line `target_identities = set(target_identities)` creates a new set as a shallow copy of the input target_identities. This means any modifications made to target_identities within `wait_for_instruction_result` will only affect this local copy, not the original identities set passed from `_execute_worker_method`.

vllm/v1/worker/gpu_worker.py
```diff
@@ -60,6 +74,183 @@
...
+
+    def _set_device_communicator_status(self, active: bool):
+        model_groups = get_all_model_groups()
+        for group in model_groups:
+            if group.device_communicator is not None:
+                nccl_comm = group.device_communicator.pynccl_comm
+                nccl_comm.available = active
+                nccl_comm.disabled = not active
+
+    def restart_worker(self):
```
ruisearch42: The name is a bit misleading? This looks like a resume/restore worker.

vllm/v1/worker/gpu_worker.py
```diff
@@ -60,6 +74,183 @@
...
+        self.worker_guard_dead = False
+        self.pause_event = pause_event
+        self.communicator_aborted = False
+        self.logger = self._make_worker_logger()
+        threading.Thread(
+            target=self.run, daemon=True, name="WorkerGuardCmdReceiver"
+        ).start()
+
+    def _make_worker_logger(self):
+        prefix = f"[WorkerGuard_dp{self.dp_rank}_tp{self.tp_rank}_pp{self.pp_rank}] "
```
ruisearch42: nit: PP should come first?

vllm/v1/engine/core_client.py
```diff
@@ -332,6 +345,127 @@ def dp_engines_running(self) -> bool:
         return False
 
 
+class ClientGuard:
```
ruisearch42: could we have a better name? :)

vllm/entrypoints/api_server.py
```diff
@@ -56,6 +56,34 @@ async def generate(request: Request) -> Response:
     return await _generate(request_dict, raw_request=request)
 
 
+@app.post("/fault_tolerance/apply")
+async def send_fault_tolerance_instruction(request: Request) -> Response:
+    """Generate completion for the request.
```
ruisearch42: fix doc

ruisearch42 (commented):
Thanks for the PR!

First pass. Just some nitpicks.

I think it makes sense overall. Any end-to-end tests/demonstration that this is functional and actually helps recover from faults (considering there are quite some changes added)?

vllm/v1/worker/gpu_worker.py
fangyuchu: The name has been changed to restore_worker.

vllm/v1/worker/gpu_worker.py
fangyuchu: The order is corrected.

vllm/entrypoints/api_server.py
fangyuchu: The error in doc string has been corrected.

vllm/v1/engine/core_client.py
fangyuchu: Thanks for the suggestion. Considering there’s already a `start_engine_core_monitor` function in `MPClient`, perhaps `Watchdog` would be a better fit to replace the new `Guard` classes?

vllm/engine/arg_utils.py
```diff
@@ -1636,6 +1670,16 @@ def create_engine_config(
             collect_detailed_traces=self.collect_detailed_traces,
         )
 
+        fault_tolerance_config = FaultToleranceConfig(
+            enable_fault_tolerance=self.enable_fault_tolerance,
+            engine_recovery_timeout=self.engine_recovery_timeout,
+            internal_fault_report_port=self.internal_fault_report_port
+            or FaultToleranceConfig.internal_fault_report_port,
+            external_fault_notify_port=self.external_fault_notify_port
+            or FaultToleranceConfig.external_fault_notify_port,
```
hmellor: Are the `or`s necessary? `self.external_fault_notify_port` will always be truthy unless someone manually sets the port to `0`

fangyuchu: Thanks for catching that! The `or` part was a remnant from some early debugging code and has been removed in the latest update.

vllm/v1/engine/core_client.py
fangyuchu: We renamed the `Guard` classes to `Sentinel`, which we think better reflects their role (similar to Redis Sentinel for monitoring and fault handling).

vllm/config/vllm.py
```diff
@@ -10,7 +10,7 @@
 import threading
 import time
 from contextlib import contextmanager
-from dataclasses import replace
+from dataclasses import field, replace
```
hmellor:
```suggestion
from dataclasses import replace
```

vllm/config/vllm.py
```diff
@@ -107,6 +108,10 @@ class VllmConfig:
     """The configurations for event publishing."""
     ec_transfer_config: ECTransferConfig | None = None
     """The configurations for distributed EC cache transfer."""
+    fault_tolerance_config: FaultToleranceConfig = field(
```
hmellor:
```suggestion
    fault_tolerance_config: FaultToleranceConfig = Field(
```

hmellor (commented):
Just a couple of nits on the config code, not a full review

vllm/config/vllm.py
fangyuchu: Thanks for the review. Fixed as advised.

vllm/distributed/device_communicators/pynccl_wrapper.py
```diff
@@ -360,10 +371,32 @@ def __init__(self, so_file: str | None = None):
...
+            return
+
+        # Handle non-blocking communicators
+        if result == ncclInProgress:
+            if comm is None:
+                raise RuntimeError(
+                    "NCCL_CHECK: ncclInProgress returned but no communicator "
+                    "provided (required for non-blocking NCCL checks)."
+                )
+            while True:
```
tzulingk, line 395: nit: consider adding some small interval between loops?
```
    # Small sleep to avoid busy loop
    time.sleep(0.001)  # 1ms - balance between responsiveness and CPU
```
Benefits:

- Reduces CPU usage
- Still checks frequently (1000x/second)
- Yields to other threads

vllm/distributed/parallel_state.py
```diff
@@ -312,6 +313,8 @@ def __init__(
         use_device_communicator: bool,  # whether to use device communicator
         use_message_queue_broadcaster: bool = False,
         group_name: str | None = None,
+        enable_fault_tolerance: bool = False,
+        gloo_comm_timeout: timedelta | None = None,
```
tzulingk: nit: consider using FaultToleranceConfig as input param? We won't need to change the input signature when adding/removing fault tolerance related features.

vllm/distributed/parallel_state.py
```diff
@@ -323,14 +326,31 @@ def __init__(
...
-                ranks, backend=torch_distributed_backend
-            )
+            if torch_distributed_backend == "nccl":
+                device_group = torch.distributed.new_group(
+                    ranks, backend=torch_distributed_backend, pg_options=options
+                )
+            else:
+                device_group = torch.distributed.new_group(
+                    ranks, backend=torch_distributed_backend
+                )
```
tzulingk: https://github.com/pytorch/pytorch/blob/v2.9.1/torch/distributed/distributed_c10d.py#L5198 `pg_options` default is None.
we can have 
```
options = None
if torch_distributed_backend == "nccl":
   options = torch._C._distributed_c10d.ProcessGroupNCCL.Options()
for ranks in group_ranks:
  device_group = torch.distributed.new_group(
     ranks, backend=torch_distributed_backend, pg_options=options
)
```

vllm/distributed/parallel_state.py
```diff
@@ -323,14 +326,31 @@ def __init__(
...
             # a group with `gloo` backend, to allow direct coordination between
             # processes through the CPU.
             with suppress_stdout():
-                cpu_group = torch.distributed.new_group(ranks, backend="gloo")
+                if not enable_fault_tolerance:
+                    cpu_group = torch.distributed.new_group(ranks, backend="gloo")
+                else:
+                    cpu_group = torch.distributed.new_group(
+                        ranks, backend="gloo", timeout=gloo_comm_timeout
+                    )
```
tzulingk: same here

vllm/distributed/parallel_state.py
```diff
@@ -1040,21 +1060,29 @@ def get_inner_dp_world_group() -> GroupCoordinator:
 
 
 def init_world_group(
-    ranks: list[int], local_rank: int, backend: str
+    ranks: list[int],
+    local_rank: int,
+    backend: str,
+    enable_fault_tolerance: bool = False,
```
tzulingk: nit: consider using FaultToleranceConfig as input param?

vllm/distributed/parallel_state.py
```diff
@@ -1271,6 +1333,8 @@ def initialize_model_parallel(
     prefill_context_model_parallel_size: int = 1,
     decode_context_model_parallel_size: int | None = 1,
     backend: str | None = None,
+    enable_fault_tolerance: bool = False,
+    gloo_comm_timeout: timedelta | None = None,
```
tzulingk: nit: consider using FaultToleranceConfig as input param?

vllm/v1/engine/core.py
```diff
@@ -73,6 +85,344 @@
...
+            """
+            getattr(logger, level)(prefix + msg, *args, **kwargs)
+
+        return log
+
+    def run(self) -> None:
+        """
+        Run the main monitoring loop for EngineCoreSentinel.
+        """
+        poll_timeout_ms = 100
```
tzulingk: should `poll_timeout_ms` be configurable?

vllm/v1/engine/core.py
```diff
@@ -73,6 +85,344 @@
...
+        )
+
+        all_success = True
+        worker_responses = wait_for_instruction_result(
+            self.worker_cmd_socket, identities, method_name, timeout, method_uuid
+        )
+        for identity in identities:
+            response = worker_responses.get(identity)
+            if response is None or not response.get("success", False):
+                all_success = False
```
tzulingk: nit: we could simply return false when encountering the first non-success result?

vllm/v1/engine/utils.py
```diff
@@ -64,6 +80,14 @@ class EngineZmqAddresses:
     # Not used by engine, just relayed to front-end in handshake response.
     # Only required for external DP LB case.
     frontend_stats_publish_address: str | None = None
+    #
```
tzulingk: redundant?

vllm/v1/engine/utils.py
```diff
@@ -64,6 +80,14 @@ class EngineZmqAddresses:
...
+    fault_report_addr: str | None = None
+    # ZMQ client_cmd socket address of client sentinel
+    client_cmd_addr: str | None = None
+    # identities of engine_core_sentinel
```
tzulingk: nit: add comment for what the keys and values represent.

vllm/v1/engine/utils.py
```diff
@@ -154,12 +196,53 @@ def __init__(
...
+            )
+            logger.info("Sent message to ClientSentinel: %s", dead_message)
+        except Exception as e:
+            logger.error("Failed to send message: %s", e)
+
     def close(self):
         """Shutdown all procs."""
         self._finalizer()
 
+    def start_engine_core_monitor(self):
```
tzulingk: this seems to be very similar to start_engine_core_monitor in vllm/v1/engine/core_client.py. Can we consolidate them?
ex
```
def report_engine_death(
    died_proc,
    sentinel,
    sentinels_list,
    engine_down_socket,
    logger
):
    """
    Report engine death via fault info and clean up.
    
    Args:
        died_proc: The dead process
        sentinel: The sentinel that triggered
        sentinels_list: List to remove sentinel from
        engine_down_socket: ZMQ socket to report on
        logger: Logger instance
    """
    # Extract rank from process name
    match = re.match(r"EngineCore_DP(\d+)", died_proc.name)
    engine_rank = match.group(1) if match else "unknown"
    
    # Create fault info
    fault_info = FaultInfo(
        type="engine_core dead",
        message=f"Engine core proc {died_proc.pid} ({died_proc.name}) died unexpectedly.",
        engine_id=engine_rank,
        additional_info=None,
    )
    
    # Report via ZMQ
    engine_down_socket.send_multipart([b"", fault_info.serialize().encode("utf-8")])
    
    # Cleanup
    if isinstance(sentinel, int) and sentinel in sentinels_list:
        sentinels_list.remove(sentinel)
    
    # Log
    logger.error("Engine core proc %s died unexpectedly", died_proc.name)
    
    return engine_rank, fault_info
```
And use in both files

vllm/v1/worker/gpu_worker.py
```diff
@@ -61,6 +75,184 @@
     from vllm.model_executor.model_loader.tensorizer import TensorizerConfig
 
 
+class WorkerSentinel:
```
tzulingk: there appear to be multiple Sentinel types:

- EngineCoreSentinel - Monitors EngineCore
- WorkerSentinel - Monitors workers
- ClientSentinel - Monitors client (mentioned in tests)
Can we consolidate them?
ex
```
class BaseSentinel(threading.Thread, ABC):
    """
    Base class for all monitoring sentinels.
    Provides common infrastructure for monitoring, command processing, and fault reporting.
    """
    
    def __init__(self, name: str, zmq_ctx: zmq.Context):
        super().__init__(daemon=True, name=f"{name}Thread")
        self.name = name
        self.zmq_ctx = zmq_ctx
        self.sentinel_active = True  # Unified lifecycle flag
        self.logger = self._create_logger()
    
    def _create_logger(self):
        """Create logger with sentinel-specific prefix."""
        prefix = f"[{self.name}] "
        
        def log(msg, *args, level="info", **kwargs):
            getattr(logger, level)(prefix + msg, *args, **kwargs)
        
        return log
    
    @abstractmethod
    def _monitor_loop_iteration(self):
        """
        One iteration of monitoring.
        Subclasses implement specific monitoring logic.
        Returns: True to continue, False to exit
        """
        pass
    
    def run(self):
        """Main monitoring loop (template method)."""
        self.logger("Starting")
        
        try:
            while self.sentinel_active:
                if not self._monitor_loop_iteration():
                    break
        except Exception as e:
            self.logger("Monitoring loop exception: %s", e, level="error")
        finally:
            self._cleanup()
        
        self.logger("Stopped")
    
    @abstractmethod
    def _cleanup(self):
        """Cleanup resources. Subclasses implement specific cleanup."""
        pass
    
    def shutdown(self):
        """Request shutdown."""
        self.sentinel_active = False
```

```

Comment (fangyuchu) [reply vllm/v1/worker/gpu_worker.py]: Thanks a lot for the review!  This is a great suggestion. We’re looking into how to refactor these into a common base. Thanks again for the insightful feedback!

File: vllm/v1/engine/utils.py (hunk 19 above)
Comment (fangyuchu) [reply vllm/v1/engine/utils.py]: We’ve consolidated the logic according to the suggestion and reduce the duplication.

File: vllm/distributed/parallel_state.py (hunk 10 above)
Comment (fangyuchu) [reply vllm/distributed/parallel_state.py]: Thanks for the suggestion. We now pass FaultToleranceConfig when creating the distributed environment to improve extensibility.

File: vllm/distributed/parallel_state.py (hunk 11 above)
Comment (fangyuchu) [reply vllm/distributed/parallel_state.py]: Revised as suggested. Thanks!

File: vllm/v1/engine/core.py (hunk 16 above)
Comment (fangyuchu) [reply vllm/v1/engine/core.py]: Revised as suggested. Thanks!

File: vllm/distributed/device_communicators/pynccl_wrapper.py (hunk 9 above)
Comment (fangyuchu) [reply vllm/distributed/device_communicators/pynccl_wrapper.py:395]: Thanks for the suggestion.
Here the loop is replacing NCCL’s original blocking behavior, since we switch to non-blocking mode to support fault tolerance and safe aborts. We currently do not add a sleep to avoid potential latency impact on this critical path, but we are not entirely certain if this is the best approach and would appreciate your thoughts on whether a small sleep would be appropriate.

File: vllm/v1/engine/core.py (hunk 15 above)
Comment (fangyuchu) [reply vllm/v1/engine/core.py]: Thanks for the suggestion. We made poll_timeout_ms a file-level constant in base_sentinel.py  to centralize the value. We can consider making it configurable per sentinel instance if needed.

File: vllm/v1/engine/utils.py (hunk 18 above)
Comment (fangyuchu) [reply vllm/v1/engine/utils.py]: We add the following comment:  Identities of engine core DEALER sockets, keyed by engine index. These identities are used by the ClientSentinel (ROUTER) to route messages to the corresponding engine core.

File: vllm/v1/worker/gpu_worker.py (hunk 20 above)
Comment (fangyuchu) [reply vllm/v1/worker/gpu_worker.py]: Thanks again for the great suggestion. We have now refactored the Sentinels with a common base class, BaseSentinel, so that all sentinel types share the same parent.

File: vllm/v1/engine/core.py [hunk 21]
```diff
@@ -688,6 +953,46 @@ def __init__(
...
+                    cmd_q=self.cmd_q,
+                    busy_loop_active=self.busy_loop_active,
+                    engine_input_q=self.input_queue,
+                    engine_fault_socket_addr=addresses.engine_fault_socket_addr,
+                    client_cmd_addr=addresses.client_cmd_addr,
+                    worker_cmd_addr=worker_cmd_addr,
+                    dealer_socket_identity=engine_core_sentinel_ids[self.engine_index],
+                    tp_size=vllm_config.parallel_config.tensor_parallel_size,
+                    pp_size=vllm_config.parallel_config.pipeline_parallel_size,
+                    dp_size=vllm_config.parallel_config.data_parallel_size,
```
jianzs: Can we pass the vllm_config directly?  Other parallel configurations, such as `context_parallelism`, might also impact fault tolerance behavior.

vllm/v1/engine/core.py
```diff
@@ -62,19 +65,285 @@
...
+        """
+        while not self.sentinel_dead:
+            # Check for engine fault signals
+            self.poll_and_report_fault_events()
+            # Check for commands from ClientSentinel
+            has_msg, cmd_str = self.receive_upstream_cmd()
+            if has_msg:
+                assert cmd_str is not None
+                success, method_uuid, reason = self._execute_cmd(cmd_str)
+                self._send_execution_result(success, method_uuid, reason)
```
jianzs: Why not use a clear request type, such as EngineCoreRequestType?

vllm/v1/engine/core.py
```diff
@@ -62,19 +65,285 @@
...
+            downstream_cmd_addr=worker_cmd_addr,
+            dealer_socket_identity=dealer_socket_identity,
+            sentinel_tag=f"DP_{engine_index}",
+            fault_tolerance_config=fault_tolerance_config,
+        )
+
+        self.fault_signal_q = fault_signal_q
+        self.cmd_q = cmd_q
+        self.busy_loop_active = busy_loop_active
+        self.engine_input_q = engine_input_q
```
jianzs: It's best to briefly explain what these attributes do. I assume `fault_signal_q` is used to send error reports to the higher-level component?

vllm/v1/engine/core.py
fangyuchu: Improved as suggested. We have introduced the `FaultToleranceRequest` and `FaultToleranceResult` structs.

vllm/config/fault_tolerance.py
```diff
@@ -0,0 +1,74 @@
...
+import hashlib
+from typing import Any
+
+from pydantic.dataclasses import dataclass
+
+from vllm.config.utils import config
+
+
+@config
+@dataclass
```
hmellor: `config` is a dataclass transform now

```suggestion
@config
```

fangyuchu: Removed as suggested.

vllm/v1/engine/core.py
fangyuchu: Revised as suggested.

vllm/config/fault_tolerance.py
```diff
@@ -0,0 +1,76 @@
...
+    enable_fault_tolerance: bool = False
+    """Enable fault tolerance for detailed error recovery,
+    such as scaling down fault DPEngineCore.
+    """
+
+    shutdown_on_fault_tolerance_failure: bool = False
+    """Whether to shut down vLLM when a fault tolerance action fails.
+    """
+
+    engine_recovery_timeout: int = 60
```
tzulingk: nit: It's better to bake the unit into the variable: engine_recovery_timeout_sec

vllm/config/fault_tolerance.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+
+    internal_fault_report_port: int = 22866
+    """
+    The port to use for internal fault reporting.
+    """
+
+    external_fault_notify_port: int = 22867
+    """
+    The port to use for external fault notify.
```
tzulingk: nit: can you be explicit what count as internals and what count as externals?

vllm/distributed/device_communicators/pynccl_wrapper.py
tzulingk: When I used non-blocking mode, I actually got lots of different issues when running MoE models. 
But I will trust that you have make things right, maybe I just wasn't doing thing correctly.

vllm/config/fault_tolerance.py
fangyuchu: Thx for the suggestion. Revised as suggested.

vllm/config/fault_tolerance.py
fangyuchu: internal_fault_report_port is the port to use for engines to report fault to client sentinel. external_fault_notify_port is the port used to publish engine fault and status change notifications.

vllm/config/fault_tolerance.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+    Port used to publish engine fault and status change notifications.
+    """
+
+    gloo_comm_timeout: int = 30
+    """
+    The timeout (in seconds) for gloo communication.
+    """
+
+    worker_cmd_addr: str | None = None
```
SageMoore, line 46: It looks like this is here so that we can easily share the address between EngineCore and the GPU sentinel? Config classes should only contain user configurable knobs, not global mutable state. How difficult would it be to remove this field?

vllm/config/vllm.py
```diff
@@ -302,6 +303,10 @@ class VllmConfig:
     """The configurations for event publishing."""
     ec_transfer_config: ECTransferConfig | None = None
     """The configurations for distributed EC cache transfer."""
+    fault_tolerance_config: FaultToleranceConfig = Field(
```
SageMoore, line 306: I see this check in a few spots
```
if (fault_tolerance_config is not None and fault_tolerance_config.enable_fault_tolerance)
```

Can we swap this line over to:
```
fault_tolerance_config: FaultToleranceConfig | None = None
```

And only construct a `FaultToleranceConfig` object when the feature is enabled? This would also allow you to remove the `enable_fault_tolerance` boolean from the `FaultToleranceConfig`.

vllm/config/fault_tolerance.py
```diff
@@ -0,0 +1,76 @@
...
+    WorkerSentinel instances. This address is assigned dynamically during 
+    runtime.
+    """
+
+    fault_state_pub_topic: str = "vllm_fault"
+    """
+    The topic string to use for publishing fault state messages via ZMQ.
+    """
+
+    def compute_hash(self) -> str:
```
SageMoore, line 58: I don't think this config should effect the computation graph. Do you need this method? If so I think you need to add a call in `VllmConfig.compute_hash()`

vllm/config/fault_tolerance.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+
+    worker_cmd_addr: str | None = None
+    """
+    ZMQ address used by EngineCoreSentinel to dispatch instructions to 
+    WorkerSentinel instances. This address is assigned dynamically during 
+    runtime.
+    """
+
+    fault_state_pub_topic: str = "vllm_fault"
```
SageMoore, line 53: Can this just be a constant instead of a config option?

vllm/config/fault_tolerance.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+    The port to use for engines to report fault to client sentinel.
+    """
+
+    external_fault_notify_port: int = 22867
+    """
+    Port used to publish engine fault and status change notifications.
+    """
+
+    gloo_comm_timeout: int = 30
```
SageMoore, line 41: Nit: this seems like it belongs in the `ParallelConfig`

vllm/config/parallel.py
```diff
@@ -520,14 +521,23 @@ def get_next_stateless_eplb_group_port(self) -> list[int]:
 
     @overload
     def stateless_init_dp_group(
-        self, return_store: Literal[False] = ...
```
SageMoore, line 523: I think we should move the `FaultToleranceConfig` into the `ParallelConfig` and remove all of these extra arguments. `ParallelConfig` is where all configuration for EPLB and Elastic EP lives so it's a natural place for Fault Tolerance.

vllm/config/parallel.py
```diff
@@ -544,22 +554,26 @@ def stateless_init_dp_group(
...
                 # use gloo since the engine process might not have cuda device
                 return stateless_init_torch_distributed_process_group(
                     self.data_parallel_master_ip,
-                    self.get_next_dp_init_port(),
+                    dp_init_port,
                     self.data_parallel_rank,
                     self.data_parallel_size,
                     backend="gloo",
                     return_store=return_store,
+                    fault_tolerance_config=fault_tolerance_config,
```
SageMoore, line 569: Nit: can you just pass the gloo timeout instead of the whole config?

vllm/distributed/utils.py
```diff
@@ -504,6 +505,7 @@ def stateless_init_torch_distributed_process_group(
     backend: str,
     group_name: str | None = None,
     return_store: bool = False,
+    fault_tolerance_config: FaultToleranceConfig | None = None,
```
SageMoore, line 508: Nit: I said this in another comment, but let's just make this an optional timeout

SageMoore (commented):
Hi @fangyuchu. Great work on this. I have an initial round of comments covering the config and UX parts of the code.

vllm/config/vllm.py
fangyuchu: I made a small adjustment around this in the fault-report PR.

We moved the FT config under parallel_config, and kept the enable_fault_tolerance flag but lifted it up to parallel_config as well.

This follows how EPLB configuration is structured (enable_eplb and eplb_config), so in most places we can just check enable_fault_tolerance directly.

Would love to hear your thoughts on this approach.

vllm/config/fault_tolerance.py
fangyuchu: Removed as suggested.

vllm/config/fault_tolerance.py
fangyuchu: This has been moved to vllm/v1/fault_tolerance/utils.py and defined as a constant now.

vllm/config/fault_tolerance.py
fangyuchu: I’ve updated this to pass the address via the executor’s `collective_rpc` and create the worker sentinel on the worker side instead. 

Change is here: https://github.com/vllm-project/vllm/pull/38534/changes/527162a2d9352ba7018bc9fdbe7ae4f6083459c3

vllm/config/fault_tolerance.py
fangyuchu: Revised as suggested.

vllm/config/parallel.py
fangyuchu: Revised as suggested: https://github.com/vllm-project/vllm/pull/38534/changes/2a071686d35f3ed4dfec2b2b2298e04c9f5d23ab

hmellor:
Closing as superseded by #44428, which landed the fault tolerance framework in July 2026. RFC #27866 was closed as not planned, and the three split PRs (#34833, #38534, #40468) were closed in June 2026.

PR #28298, opened 2025-11-07
[Bugfix] Spec decode + structured output + spec model max len edge case

andylolu2:
## Purpose

Fixes https://github.com/vllm-project/vllm/issues/27210

Error:

```
  File "/home/ubuntu/vllm/vllm/v1/core/sched/scheduler.py", line 897, in get_grammar_bitmask
    bitmask = self.structured_output_manager.grammar_bitmask(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/vllm/vllm/v1/structured_output/__init__.py", line 263, in grammar_bitmask
    assert structured_output_request.grammar.accept_tokens(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

Cause:

The current implementation relies on `scheduler.update_draft_token_ids` to be called to reset previously-used spec_token_ids. However, `scheduler.update_draft_token_ids` is not guaranteed to be called after every step (currently it happens when the model_execute did not produce any new draft tokens, e.g. when `input_fits_in_drafter` is False). This results in stale spec_token_ids that don't necessarily follow the grammar after accepting the output tokens from the current step, leading to the above assertion error. 

Proposed fix: Clear the spec token ids of a request immediately once it has been scheduled.

## Test Plan

Extended current tests to cover this edge case.

## Test Result

No more assertion errors.

---
<details>

</details>


vllm/v1/core/sched/scheduler.py
```diff
@@ -321,10 +321,13 @@ def schedule(self) -> SchedulerOutput:
                 )
                 if num_scheduled_spec_tokens > 0:
                     # Trim spec_token_ids list to num_scheduled_spec_tokens.
-                    del request.spec_token_ids[num_scheduled_spec_tokens:]
                     scheduled_spec_decode_tokens[request.request_id] = (
-                        request.spec_token_ids
+                        request.spec_token_ids[:num_scheduled_spec_tokens]
```
njhill: It should still be fine to leave the list trimming here to avoid extra copy/allocation?

vllm/v1/core/sched/scheduler.py
```diff
@@ -321,10 +321,13 @@ def schedule(self) -> SchedulerOutput:
...
                     )
+                # After this step the speculative tokens will no longer be
+                # valid because the structured output state would have advanced.
+                # Clear the list to avoid reusing them.
```
njhill: Perhaps just say here that `request.spec_token_ids` will be set via `update_draft_token_ids()` before the next step if applicable.

vllm/v1/structured_output/__init__.py
```diff
@@ -271,7 +271,7 @@ def grammar_bitmask(
                     ):
                         assert structured_output_request.grammar.accept_tokens(
                             req_id, [token]
-                        )
+                        ), (token, req_id, scheduled_spec_decode_tokens)
```
njhill: Unrelated issue but this should also be fixed:
```python
                        accepted = structured_output_request.grammar.accept_tokens(
                            req_id, [token]
                        )
                        assert accepted, (token, req_id, scheduled_spec_decode_tokens)
```
IIUC `accept_tokens` has side effects and so should not be done in the `assert` which could in theory be disabled.

njhill (commented):
Thanks a lot @andylolu2 for this fix! And for updating the test.

Does your test update now exercise the `input_fits_in_drafter == False` case? If so that's great, I had just noticed yesterday that there was no test added when that check was put in.

vllm/v1/core/sched/scheduler.py
andylolu2: Good point, fixed

njhill (approved):
Thanks again @andylolu2

Issue #23567, opened 2025-08-25
[Bug]: openai_harmony.HarmonyError: unexpected tokens remaining in message header


MoellerAI:
### Your current environment

<details>
I keep hitting this error “ openai_harmony.HarmonyError: unexpected tokens remaining in message header”

In multi-turn conversations when using gpt-oss-120b and with both vllm v0.10.1 and v0.10.1.1. I use the official docker image. Seems like a lot of users have this problem. Please also see this issue: https://github.com/openai/harmony/issues/38 

But I haven’t found the problem at users not using vllm, so might very well be a vllm problem.

</details>

### 🐛 Describe the bug

<details>
I keep hitting this error “ openai_harmony.HarmonyError: unexpected tokens remaining in message header”

In multi-turn conversations when using gpt-oss-120b and with both vllm v0.10.1 and v0.10.1.1. I use the official docker image. Seems like a lot of users have this problem. Please also see this issue: https://github.com/openai/harmony/issues/38 

But I haven’t found the problem at users not using vllm, so might very well be a vllm problem.

</details>

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

MoellerAI:
It pretty much makes the gpt-oss-120b unusable in a chat-page with streaming responses

iSevenDays:
I also see the same issue frequently 
```
(APIServer pid=19050) INFO 08-26 16:56:32 [loggers.py:123] Engine 000: Avg prompt throughput: 119.4 tokens/s, Avg generation throughput: 359.3 tokens/s, Running: 5 reqs, Waiting: 0 reqs, GPU KV cache usage: 2.2%, Prefix cache hit rate: 50.0%
(APIServer pid=19050) ERROR 08-26 16:56:32 [serving_chat.py:1077] Error in chat completion stream generator.
(APIServer pid=19050) ERROR 08-26 16:56:32 [serving_chat.py:1077] Traceback (most recent call last):
(APIServer pid=19050) ERROR 08-26 16:56:32 [serving_chat.py:1077]   File "/root/miniconda/envs/vllm-seed-oss-original/lib/python3.12/site-packages/vllm/entrypoints/openai/serving_chat.py", line 665, in chat_completion_stream_generator
[...]

PR #28303, opened 2025-11-07
[Bugfix] Parse gpt-oss refusals w/ newer openai-harmony

bbrowning:
## Purpose

The output generated by gpt-oss models does not always strictly follow its expected harmony chat template format. This commonly - but not exclusively - happens when gpt-oss-120b generates refusals for content that violates its built-in safety guidelines.

To fix this, a non-strict mode was added to the openai-harmony library to allow attempted recovery of malformed message headers in the model output, such as a missing `<|message|>` special token before the assistant text.

This will resolve some cases where the error
`openai_harmony.HarmonyError: unexpected tokens remaining in message header` was previously thrown. It will not resolve all of those, as not every malformed message output can be recovered. Other ongoing work around using structured output for the Harmony format can help prevent these kinds of things in the first place, once that work lands and in the cases where the user and/or server decide to enable it.

I believe it should be safe to enable this non-strict mode by default in vLLM, as the code paths that enables in the openai-harmony library only gets triggered once it's already detected malformed output. So, there shouldn't be any performance penalty in the common case. And, in the event that the malformed content cannot be properly recovered, the openai-harmony library will still end up throwing an error.

This is related to vllm-project/vllm#23567 as well as openai/harmony#80.

## Test Plan

I added a new test to verify the refusal parsing in `test_harmony_utils.py`. I also run `test_response_api_with_harmony.py` locally, but skip the code interpreter tests because my dev machine is not setup properly to run that particular one.

```
pytest tests/entrypoints/openai/parser/test_harmony_utils.py
pytest -k "not code_interpreter" tests/entrypoints/openai/test_response_api_with_harmony.py
```

## Test Result

```
$ pytest -q --disable-warnings tests/entrypoints/openai/parser/test_harmony_utils.py
.........................................................................       [100%]
73 passed, 2 warnings in 4.33s

$ pytest -q --disable-warnings -k "not code_interpreter" \
  tests/entrypoints/openai/test_response_api_with_harmony.py
................s........                                                         [100%]
24 passed, 1 skipped, 1 deselected, 3 warnings in 74.30s (0:01:14)
```


austin-mccall (approved):

alecsolder:
Some random thoughts:

I think I do lean to not having this enabled by default at first, or if we do have it enabled by default, we need to:

- Have validation on our side that the output message does in fact have enough metadata to be "complete"
- Things that shouldn't be in the content aren't in the content (i.e you shouldn't see <|channel|> in the content because of this)
- Have logging to know how often we are actually hitting the case where strict=False is preventing a crash

For example, if I deployed this, I'd have no way of understanding the impact of the changes. How many requests its "saving" from erroring for example.

I also think there is a difference between the impact of this change on single-turn responses API vs in the [tool calling while loop](https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/openai/serving_engine.py#L1178). I think in single-turn responses API usage there is a better ability to "repair" harmony messages when they are translated to responses items and back into harmony messages, but for the tool calling loop it stays as harmony messages the entire time so these issues may compound silently

pb-sameerreddy:
+1 do you need any help here? how can we get this in?

ashishpatel26:
I prepared a clean current-main follow-up branch that keeps the `openai-harmony >= 0.0.8` / `strict=False` fix and adds one extra regression for vLLM's public parse path.

The added coverage checks both:

- raw parser recovery for malformed GPT-OSS refusal output
- `parse_chat_output` returning the recovered reasoning/final content

Branch:
https://github.com/ashishpatel26/vllm/tree/contrib/pr28303-harmony-v008-rebase

Commit:
`0ac1042849fa693f71bf537455610350580d920f`

Cherry-pick:

```bash
git fetch https://github.com/ashishpatel26/vllm.git contrib/pr28303-harmony-v008-rebase
git cherry-pick 0ac1042849fa693f71bf537455610350580d920f
```

Validation:

- `git diff --check`
- `ruff check tests/entrypoints/openai/parser/test_harmony_utils.py vllm/entrypoints/openai/parser/harmony_utils.py`
- `ruff format --check tests/entrypoints/openai/parser/test_harmony_utils.py vllm/entrypoints/openai/parser/harmony_utils.py`
- targeted pytest: `2 passed, 2 warnings`

Note: I did not open a PR into this branch because the current-main follow-up branch diverges heavily from `upgrade-openai-harmony-v0.0.8`; cherry-picking this single commit onto a rebased branch should be cleaner.

bbrowning:
I'm going to close this for now, as it's out of date. There's more recent work in this area for gpt-oss models and their integrating into vLLM that's ongoing that should clear a lot of this up.

PR #28306, opened 2025-11-07
[Kernel] Support CUDA Graphs in 3D Triton Attention Kernel

jvlunteren:
## Purpose
~~This pull request depends on PR #27993.~~

This pull request adapts the 3D Triton attention kernel, which is used exclusively for decode operations, to support full CUDA Graphs. The key changes include:

- The allocation of the intermediate data structures used for the tiled softmax implementation has been moved to the attention metadata builder class.

- The dynamic selection between the 2D and 3D attention kernels during decode is now based on comparing the batch size against a threshold corresponding to one of the CUDA Graph capture sizes. This ensures that, for each batch size, only one valid kernel choice (either 2D or 3D kernel) exists and will be captured correctly.

The updated kernel selection logic is only applied when CUDA Graphs are enabled for decoding. This is now automatically detected in the attention metadata builder, which sets the appropriate threshold values accordingly.

## Test Plan

The unit test `./tests/kernels/attention/test_triton_unified_attention.py` has been updated to include the allocation of intermediate data structures required for the tiled softmax implementation.  It has also been modified to explicitly test the separate use of the 2D and 3D attention kernels during decoding. Other tests, such as `lm_eval`, remain compatible and can be used without modification.

## Test Result

================================================ 512 passed in 71.05s (0:01:11) ================================================
</pre>

<br>
<br>

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.800|±  |0.0179|
|     |       |strict-match    |     5|exact_match|↑  |0.788|±  |0.0183|</pre>

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.790|±  |0.0182|
|     |       |strict-match    |     5|exact_match|↑  |0.768|±  |0.0189|</pre>

@tdoublep @bringlein


vllm/v1/attention/backends/triton_attn.py
```diff
@@ -86,6 +97,59 @@ def __init__(
...
+            )
+
+            # If the updated threshold becomes significantly larger than the
+            # initial value, it is reset to zero. This enforces the use of the
+            # 2D kernel only and ensures that the size of the allocated
+            # intermediate structures remains bounded.
+            if upd_seq_threshold_3D <= 4 * self.seq_threshold_3D:
+                self.seq_threshold_3D = upd_seq_threshold_3D
+            else:
+                self.seq_threshold_3D = 0
```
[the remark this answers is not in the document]
jvlunteren: According to https://docs.vllm.ai/en/latest/configuration/engine_args/#compilationconfig:

> If cudagraph_capture_sizes is specified, this will be set to the largest size in that list (or checked for consistency if specified). If cudagraph_capture_sizes is not specified, the list of sizes is generated automatically following the pattern:
> 
> [1, 2, 4] + list(range(8, 256, 8)) + list(
> range(256, max_cudagraph_capture_size + 1, 16))

From this, I assumed that cudagraph_capture_sizes is not empty when CUDA Graphs are enabled. However, to be sure, I have implemented the modification as suggested above.

vllm/v1/attention/backends/triton_attn.py
```diff
@@ -86,6 +97,59 @@ def __init__(
...
+
+        self.num_par_softmax_segments = NUM_PAR_SOFTMAX_SEGMENTS
+        self.softmax_segm_output = torch.empty(
+            (
+                self.seq_threshold_3D,
+                self.num_heads_q,
+                self.num_par_softmax_segments,
+                self.headdim,
+            ),
```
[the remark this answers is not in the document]
jvlunteren: Resolved this issue by modifying code and unit test to allocate softmax buffers based on padded head dimension.

heheda12345:
CC @LucasWilkinson @tlrmchlsmth

vllm/v1/attention/backends/triton_attn.py
```diff
@@ -87,6 +98,66 @@ def __init__(
...
+        )
+
+        # Set initial value for the threshold for the number of sequences used
+        # to select between the 2D and 3D kernels for decode.
+        self.seq_threshold_3D = MIN_LAUNCH_GRID_SIZE_2D // self.num_heads_kv
+        if self.decode_cudagraph_enabled:
+            capture_sizes = self.vllm_config.compilation_config.cudagraph_capture_sizes
+            if not capture_sizes:
+                # If no CUDA Graph capture sizes are specified, the threshold
+                # is reset to zero, forcing the 2D kernel to be used.
```
tdoublep: What does it mean to use CUDA graphs with no capture sizes? Is this a case that can actually happen?

vllm/v1/attention/backends/triton_attn.py
```diff
@@ -87,6 +98,66 @@ def __init__(
...
+                # initial value, it is reset to zero. This enforces the use of the
+                # 2D kernel only and ensures that the size of the allocated
+                # intermediate structures remains bounded.
+                if upd_seq_threshold_3D <= 4 * self.seq_threshold_3D:
+                    self.seq_threshold_3D = upd_seq_threshold_3D
+                else:
+                    self.seq_threshold_3D = 0
+
+        self.num_par_softmax_segments = NUM_PAR_SOFTMAX_SEGMENTS
+        headdim_padded = 1 << (self.headdim - 1).bit_length()  # next power of 2 value
```
tdoublep: iirc there is a function like `triton.next_power_of_2` that might be a bit easier to understand

vllm/v1/attention/backends/triton_attn.py
```diff
@@ -87,6 +98,66 @@ def __init__(
...
+                )
+
+                # If the updated threshold becomes significantly larger than the
+                # initial value, it is reset to zero. This enforces the use of the
+                # 2D kernel only and ensures that the size of the allocated
+                # intermediate structures remains bounded.
+                if upd_seq_threshold_3D <= 4 * self.seq_threshold_3D:
+                    self.seq_threshold_3D = upd_seq_threshold_3D
+                else:
+                    self.seq_threshold_3D = 0
```
tdoublep: How big can the intermediate data structures really get?

vllm/v1/attention/backends/triton_attn.py
```diff
@@ -87,6 +98,66 @@ def __init__(
...
+                # is reset to zero, forcing the 2D kernel to be used.
+                self.seq_threshold_3D = 0
+            else:
+                # Select the CUDA Graph capture size closest to self.seq_threshold_3D
+                # as threshold. This ensures that each captured graph covers the
+                # correct execution path.
+                upd_seq_threshold_3D = min(
+                    capture_sizes,
+                    key=lambda x: abs(x - self.seq_threshold_3D),
+                )
```
tdoublep: What would happen if the threshold isn't one of the capture sizes?

vllm/v1/attention/backends/triton_attn.py
```diff
@@ -87,6 +98,66 @@ def __init__(
...
+            in (
+                CUDAGraphMode.FULL_AND_PIECEWISE,
+                CUDAGraphMode.FULL_DECODE_ONLY,
+                CUDAGraphMode.FULL,
+            )
+        )
+
+        # Set initial value for the threshold for the number of sequences used
+        # to select between the 2D and 3D kernels for decode.
+        self.seq_threshold_3D = MIN_LAUNCH_GRID_SIZE_2D // self.num_heads_kv
```
tdoublep: It might good to explain why we use this formula to decide the target threshold

vllm/v1/attention/backends/triton_attn.py
jvlunteren: I don't think this can happen (see above: https://github.com/vllm-project/vllm/pull/28306#discussion_r2504389862).
I included those lines just to be sure.

vllm/v1/attention/backends/triton_attn.py
jvlunteren: I am not aware of such a function. For clarity, I will add custom function to achieve this, based on the above code.

jvlunteren: I noticed that this function is defined in vllm.utils.match_utils. I have modified the code to use that function.

vllm/v1/attention/backends/triton_attn.py
jvlunteren: The data structures have the following dimensions:

- `softmax_segm_output`: `[seq_threshold_3D, num_heads_q, num_par_softmax_segments, headdim_padded]`
- `softmax_segm_max`: `[seq_threshold_3D, num_heads_q, num_par_softmax_segments]`
- `softmax_segm_expsum`: `[seq_threshold_3D, num_heads_q, num_par_softmax_segments]`

Example:
`seq_threshold_3D = 16`, `num_heads_q=32`, `num_par_softmax_segments=16`, `headdim_padded=128`

resulting sizes:

`softmax_segm_output`: 4MB
`softmax_segm_max`: 32KB
`softmax_segm_expsum`: 32KB

vllm/v1/attention/backends/triton_attn.py
jvlunteren: Let’s consider an example where the threshold is 12 and the closest capture sizes are 8 and 16. In this case, the CUDA Graph associated with capture size 8 will handle the 3D kernel, while the one for capture size 16 will handle the 2D kernel. If, during normal operation, a batch arrives with a size between 9 and 16, it will be processed by replaying the CUDA Graph for the 2D kernel. Notice that the threshold no longer influences this decision: batches in that range, whether below or above the threshold, are all processed using the 2D kernel graph. Based on this observation, I found it cleaner to align the threshold with an actual capture size so that batches with sizes up to and including the threshold are handled by the 3D kernel.

At the same time, not aligning the threshold with a capture size does not affect output correctness. Therefore, the additional complexity of determining the exact threshold based on capture sizes could be eliminated.

vllm/v1/attention/backends/triton_attn.py
jvlunteren: I will do that.

vllm/v1/attention/backends/triton_attn.py
tdoublep: Could we remove the code? I think it would be better to handle this at a higher level (e.g., raising an error if the user asks for CUDA graphs with an empty list) rather than trying to handle it everywhere in the code.

vllm/v1/attention/backends/triton_attn.py
tdoublep: Ok so 1024 sequences it can become hundreds of MB, and at that point there probably is enough parallelism for the 2D kernel anyway.

vllm/v1/attention/backends/triton_attn.py
tdoublep: thanks

tests/kernels/attention/test_triton_unified_attention.py
```diff
@@ -152,6 +158,21 @@ def test_triton_unified_attn(
         k_descale = torch.rand(scale_shape, dtype=torch.float32)
         v_descale = torch.rand(scale_shape, dtype=torch.float32)
 
+    num_par_softmax_segments = 16
+    head_size_padded = 1 << (head_size - 1).bit_length()  # next power of 2 value
```
tdoublep: Could we use the utils function here too?

jvlunteren: Yes. I will adapt that.

jvlunteren: Done.

vllm/v1/attention/backends/triton_attn.py
jvlunteren: Yes, I will do that ... Done!

vllm/v1/attention/backends/triton_attn.py
jvlunteren: Yes, that is correct.

vllm/v1/attention/backends/triton_attn.py
```diff
@@ -87,6 +99,61 @@ def __init__(
...
+            assert capture_sizes, "CUDA Graphs enabled but no capture sizes specified."
+
+            # Select the CUDA Graph capture size closest to self.seq_threshold_3D
+            # as threshold. This ensures that each captured graph covers the
+            # correct execution path.
+            upd_seq_threshold_3D = min(
+                capture_sizes,
+                key=lambda x: abs(x - self.seq_threshold_3D),
+            )
+            self.seq_threshold_3D = upd_seq_threshold_3D
```
tdoublep:
```suggestion
            self.seq_threshold_3D = min(
                capture_sizes,
                key=lambda x: abs(x - self.seq_threshold_3D),
            )
```

jvlunteren: Done

tdoublep (approved):
LGTM

LucasWilkinson (approved):
LGTM

Issue #20492, opened 2025-07-04
[RFC]: KV-Cache Interoperability API Standardization


vMaroon:
### Motivation

This RFC proposes a KV-Cache Interoperability API, covering standardized notification events (via KVEvents) and reproducible prefix-block hashing. These standards aim to support cross-system cache awareness, observability, and future tooling for indexing, routing, and diagnostics.

vLLM already ships with internal [KVEvents](https://github.com/vllm-project/vllm/issues/16669) contributed by the NVIDIA Dynamo team - that’s a strong foundation. 
But as external systems aim for cache-aware inference, we need to treat these internal mechanisms as public contracts to support broader adoption and interop.

### Goals

1. **KVEvents Internal API as a Public Contract**  
   The KVEvents schema is already well-defined in vLLM and used internally by the `KVCacheManager` for GPU cache events. It’s also being extended to CPU offloading via the `KVConnector` (see [#19854](https://github.com/vllm-project/vllm/issues/19854)).  
   This RFC proposes formalizing KVEvents as the public contract for any component emitting or consuming KV-Cache lifecycle events - including external indexers, routers, and engines.

2. **Ensure Reproducible Block Hashing Across Languages**  
   Prefix cache block keys must be computed the same way across runtimes (e.g., Python, Go). This requires:
    - Canonical serialization (e.g., CBOR)
    - Consistent hashing algorithms (e.g., SHA256, xxHash)
    - Defined structure for input objects (e.g., token arrays, `extra_keys`)
    - Explicit rules for special cases like `NONE_HASH` root
    - Alignment on security features such as per-request hash-salting 
   
    **Disclaimer**: in the current KVEvents schema, the token-ids are sent along their block-hashes, which makes external indexing possible through mapping tokens -> different-hashes -> vLLM-hashes. While this avoids introducing reproducible hashing and configuration syncs, it requires complex indexing and lookups, along with the networking overhead of passing the 32bit token-ids in every event.

3. **Enable Language-Agnostic Interop**  
   Develop shared guidance and reference libraries in Python, Go, and other widely used languages. These utilities do not need to reside within vLLM, but should remain consistent with its specifications.

### Proposed Change

This RFC proposes standardizing two core aspects of KV-Cache awareness:

#### 1. KVEvents Schema

- Reuse the existing KVEvents format used internally in vLLM as a versioned public interface for any KV-Cache publisher or consumer
- Consider light refactoring:
    - Use `bytes` for hashes instead of Python-native `int`
[...]

PR #26172, opened 2025-10-03
[NIXL] refactor scheduler->worker request state synchronization

markmc:
In a prefill instance, we need to free KV blocks that have not been fetched after a timeout. See #20139.

In #26012, we're trying to deal with corner cases involved with doing this request timeout tracking on the worker side. This PR proposes refactoring the scheduler->worker request state synchronization to use a SCHEDULED/FINISHED/ABORTED enum rather than in_batch, to_send, and not_processed.

Note the expiry timer is switched back to monotonic time because the timestamp is no longer sent across process boundaries.

PR #19555, opened 2025-06-12
[KVConnector] Aggregate finished requests on the scheduler

orozery:
~This PR enables the worker-side KV connector to pass on arbitrary metadata to the scheduler-side connector.
This allows a standard and easy mechanism to aggregate kv-connector events from all workers.~

~In a nut-shell, we introduce the following connector APIs:
`build_worker_connector_meta` on the worker side, allowing the worker to build metadata to be sent back to the scheduler.
`get_finished` - on the scheduler side (which was previously a worker side) - gets the connector metadata from all workers and yields the finished request transfers.~

This PR makes the following changes:
1. Change MultiprocExecutor to get ModelRunnerOutput from all workers,
    and aggregate the finished_sending and finished_recving from all.
2. Remove the worker aggregation of those fields in NixlConnector.

PR #28309, opened 2025-11-07
[KVConnector] Add KV events to KV Connectors

hickeyma:
## Purpose

Enable KV connectors to pass KV Events to vLLM. vLLM can then publish the events using its existing KV Events publisher in the engine. 

The motivation to emit KV Events by connectors is to help KV-cache-aware routing make more efficient decisions, based on not only KV caches in GPU, but also KV caches in CPU and local disk.

This PR includes updates to the the LMCache connector to implement the `KVConnectorKVEvents` interface that enables KV events to be passed to vLLM from LMCache. 

This is part of the solution for:
- RFC [[RFC]: KV-Cache Interoperability API Standardization ](https://github.com/vllm-project/vllm/issues/20492) which proposes "formalizing KV Events as the public contract for any component emitting or consuming KV-Cache lifecycle events, including external indexers, routers, and engines". 

Note: This is related to #28252 and from feedback by @markmc in https://github.com/vllm-project/vllm/pull/28252#issuecomment-3501398970

## Test Plan

`pytest tests/v1/kv_connector/unit/test_lmcache_connector.py`

## Test Result

---
<details>

</details>

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -10,10 +10,6 @@ aenum==3.1.16
     # via lightly
 affine==2.4.0
     # via rasterio
-aiofile==3.9.0
-    # via lmcache
-aiofiles==25.1.0
-    # via lmcache
 aiohappyeyeballs==2.6.1
     # via aiohttp
 aiohttp==3.13.0
@@ -23,7 +19,6 @@ aiohttp==3.13.0
     #   fsspec
     #   gpt-oss
     #   lm-eval
-    #   lmcache
     #   ray
 aiohttp-cors==0.8.1
     # via ray
@@ -63,8 +58,6 @@ attrs==24.2.0
     #   referencing
 audioread==3.0.1
     # via librosa
-awscrt==0.29.2
-    # via lmcache
 backoff==2.2.1
     # via
     #   -r requirements/test.in
@@ -97,8 +90,6 @@ cachetools==5.5.2
     # via
     #   google-auth
     #   mlflow-skinny
-caio==0.9.24
-    # via aiofile
 certifi==2024.8.30
     # via
     #   fiona
@@ -155,12 +146,8 @@ coverage==7.10.6
     # via pytest-cov
 cramjam==2.9.0
     # via fastparquet
-cufile-python==0.2.0
-    # via lmcache
 cupy-cuda12x==13.6.0
-    # via
-    #   lmcache
-    #   ray
+    # via ray
 cycler==0.12.1
     # via matplotlib
 databricks-sdk==0.59.0
@@ -219,7 +206,6 @@ evaluate==0.4.3
 fastapi==0.116.1
     # via
     #   gpt-oss
-    #   lmcache
     #   mlflow-skinny
 fastparquet==2024.11.0
     # via genai-perf
@@ -331,7 +317,6 @@ httpcore==1.0.6
 httpx==0.27.2
     # via
     #   -r requirements/test.in
-    #   lmcache
     #   schemathesis
 huggingface-hub==0.34.3
     # via
@@ -458,8 +443,6 @@ llvmlite==0.44.0
     # via numba
 lm-eval @ git+https://github.com/EleutherAI/lm-evaluation-harness.git@206b7722158f58c35b7ffcd53b035fdbdda5126d
     # via -r requirements/test.in
-lmcache==0.3.10.post2
-    # via -r requirements/test.in
 lxml==5.3.0
     # via
     #   blobfile
@@ -505,8 +488,6 @@ msgpack==1.1.0
     # via
     #   librosa
     #   ray
-msgspec==0.20.0
-    # via lmcache
 mteb==2.1.2
     # via -r requirements/test.in
 multidict==6.1.0
@@ -525,10 +506,6 @@ networkx==3.2.1
     # via
     #   scikit-image
     #   torch
-nixl==0.8.0
-    # via lmcache
-nixl-cu12==0.8.0
-    # via nixl
 nltk==3.9.1
     # via rouge-score
 num2words==0.5.14
@@ -562,12 +539,10 @@ numpy==1.26.4
     #   librosa
     #   lightly
     #   lightly-utils
-    #   lmcache
     #   matplotlib
     #   mistral-common
     #   mlflow
     #   mteb
-    #   nixl-cu12
     #   numba
     #   numexpr
     #   opencv-python-headless
@@ -637,8 +612,6 @@ nvidia-nvshmem-cu12==3.3.20
     # via torch
 nvidia-nvtx-cu12==12.9.79
     # via torch
-nvtx==0.2.14
-    # via lmcache
 omegaconf==2.3.0
     # via
     #   hydra-core
@@ -702,7 +675,6 @@ packaging==24.2
     #   ray
     #   rioxarray
     #   scikit-image
-    #   setuptools-scm
     #   statsmodels
     #   tensorboardx
     #   torchmetrics
@@ -765,7 +737,6 @@ pretrainedmodels==0.7.4
     # via segmentation-models-pytorch
 prometheus-client==0.22.0
     # via
-    #   lmcache
     #   opentelemetry-exporter-prometheus
     #   ray
 propcache==0.2.0
@@ -787,13 +758,10 @@ protobuf==5.28.3
 psutil==6.1.0
     # via
     #   accelerate
-    #   lmcache
     #   peft
     #   tensorizer
 py==1.11.0
     # via pytest-forked
-py-cpuinfo==9.0.0
-    # via lmcache
 py-spy==0.4.0
     # via ray
 pyarrow==18.0.0
@@ -918,7 +886,6 @@ pyyaml==6.0.2
     #   huggingface-hub
     #   jsonargparse
     #   lightning
-    #   lmcache
     #   mlflow-skinny
     #   omegaconf
     #   peft
@@ -929,8 +896,6 @@ pyyaml==6.0.2
     #   timm
     #   transformers
     #   vocos
-pyzmq==27.1.0
-    # via lmcache
 rapidfuzz==3.12.1
     # via jiwer
 rasterio==1.4.3
@@ -941,9 +906,7 @@ rasterio==1.4.3
 ray==2.48.0
     # via -r requirements/test.in
 redis==5.2.0
-    # via
-    #   lmcache
-    #   tensorizer
+    # via tensorizer
 referencing==0.35.1
     # via
     #   jsonschema
@@ -1015,7 +978,6 @@ sacrebleu==2.4.3
 safetensors==0.4.5
     # via
     #   accelerate
-    #   lmcache
     #   open-clip-torch
     #   peft
     #   timm
@@ -1055,12 +1017,8 @@ sentence-transformers==3.2.1
 setuptools==77.0.3
     # via
     #   lightning-utilities
-    #   lmcache
     #   pytablewriter
-    #   setuptools-scm
     #   torch
-setuptools-scm==9.2.2
-    # via lmcache
 shapely==2.1.1
     # via
     #   geopandas
@@ -1085,9 +1043,7 @@ sniffio==1.3.1
     #   anyio
     #   httpx
 sortedcontainers==2.4.0
-    # via
-    #   hypothesis
-    #   lmcache
+    # via hypothesis
 soundfile==0.12.1
     # via
     #   -r requirements/test.in
@@ -1179,9 +1135,7 @@ torch==2.9.0+cu129
     #   lightly
     #   lightning
     #   lm-eval
-    #   lmcache
     #   mteb
-    #   nixl-cu12
     #   open-clip-torch
     #   peft
     #   pretrainedmodels
@@ -1247,7 +1201,6 @@ transformers==4.57.3
     #   -r requirements/test.in
     #   genai-perf
     #   lm-eval
-    #   lmcache
     #   peft
     #   sentence-transformers
     #   transformers-stream-generator
@@ -1317,7 +1270,6 @@ urllib3==2.2.3
 uvicorn==0.35.0
     # via
     #   gpt-oss
-    #   lmcache
     #   mlflow-skinny
 vector-quantize-pytorch==1.21.2
     # via -r requirements/test.in
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -10,6 +10,10 @@ aenum==3.1.16
     # via lightly
 affine==2.4.0
     # via rasterio
+aiofile==3.9.0
+    # via lmcache
+aiofiles==25.1.0
+    # via lmcache
 aiohappyeyeballs==2.6.1
     # via aiohttp
 aiohttp==3.13.0
@@ -19,6 +23,7 @@ aiohttp==3.13.0
     #   fsspec
     #   gpt-oss
     #   lm-eval
+    #   lmcache
     #   ray
 aiohttp-cors==0.8.1
     # via ray
@@ -58,6 +63,8 @@ attrs==24.2.0
     #   referencing
 audioread==3.0.1
     # via librosa
+awscrt==0.29.2
+    # via lmcache
 backoff==2.2.1
     # via
     #   -r requirements/test.in
@@ -90,6 +97,8 @@ cachetools==5.5.2
     # via
     #   google-auth
     #   mlflow-skinny
+caio==0.9.24
+    # via aiofile
 certifi==2024.8.30
     # via
     #   fiona
@@ -146,8 +155,12 @@ coverage==7.10.6
     # via pytest-cov
 cramjam==2.9.0
     # via fastparquet
+cufile-python==0.2.0
+    # via lmcache
 cupy-cuda12x==13.6.0
-    # via ray
+    # via
+    #   lmcache
+    #   ray
 cycler==0.12.1
     # via matplotlib
 databricks-sdk==0.59.0
@@ -206,6 +219,7 @@ evaluate==0.4.3
 fastapi==0.116.1
     # via
     #   gpt-oss
+    #   lmcache
     #   mlflow-skinny
 fastparquet==2024.11.0
     # via genai-perf
@@ -317,6 +331,7 @@ httpcore==1.0.6
 httpx==0.27.2
     # via
     #   -r requirements/test.in
+    #   lmcache
     #   schemathesis
 huggingface-hub==0.34.3
     # via
@@ -443,6 +458,8 @@ llvmlite==0.44.0
     # via numba
 lm-eval @ git+https://github.com/EleutherAI/lm-evaluation-harness.git@206b7722158f58c35b7ffcd53b035fdbdda5126d
     # via -r requirements/test.in
+lmcache==0.3.10.post2
+    # via -r requirements/test.in
 lxml==5.3.0
     # via
     #   blobfile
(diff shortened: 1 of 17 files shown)


vllm/distributed/kv_transfer/kv_connector/v1/lmcache_connector.py
```diff
@@ -183,6 +213,25 @@ def build_connector_meta(
...
+        """
+        # Get the KV events
+        kv_events = connector_output.kv_cache_events
+        if (
+            not kv_events
+            or not isinstance(kv_events, KVEventBatch)
+            or not kv_events.events
+        ):
+            return
+        self._kv_events = kv_events.events
```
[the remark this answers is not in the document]
hickeyma: Updated

vllm/v1/worker/kv_connector_model_runner_mixin.py
```diff
@@ -135,10 +135,19 @@ def _get_kv_connector_output(
...
             kv_connector.clear_connector_metadata()
 
     @staticmethod
     def get_kv_connector_stats() -> KVConnectorStats | None:
         if has_kv_transfer_group():
             return get_kv_transfer_group().get_kv_connector_stats()
         return None
+
+    @staticmethod
+    def get_kv_connector_kv_cache_events() -> KVConnectorStats | None:
```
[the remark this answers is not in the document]
hickeyma: Updated

orozery:
@hickeyma I think you're missing aggregation code in `KVOutputAggregator`.
This means adding more state to `KVOutputAggregator` to aggregate per-block KVEvents.

@njhill Let's think how this change fits to the longer-term future.
Right now, scheduler can send arbitrary metadata to the workers (`KVConnectorMetadata`).
On the other hand, workers are limited to an explicit `KVConnectorOutput`, which includes:
- request-level completions (`finished_sending`, `finished_recving`)
- abstract stats
- block-level errors
- (just noticed this) `expected_finished_count` - looks like a niche bootstrap configuration param

At the time when I introduced worker->scheduler output aggregation (#19555), my initial suggestion was to allow abstract metadata to flow from worker connectors to the scheduler connector.
I think that the current approach where we limit the worker output has 2 drawbacks:
1. It makes it harder to implement connector logic (e.g. the current PR is needed).
2. For those connectors who does succeed in pushing their logic to `KVConnectorOutput`, the result is a growing-over-time `KVConnectorOutput` which contains a scattered list of fields which are practically connector-specific.

My view is that we should rename `KVConnectorStats` to `KVConnectorWorkerMeta` (which could support `get_stats() -> KVConnectorStats`).
This will allow free us from maintaining the list of connector-specific worker output fields, and delegate it entirely to the connector implementations themselves.

cc @sdavidbd

markmc:
> @njhill Let's think how this change fits to the longer-term future. Right now, scheduler can send arbitrary metadata to the workers (`KVConnectorMetadata`). On the other hand, workers are limited to an explicit `KVConnectorOutput`,

I agree with this in general - e.g. in an earlier iteration of #26172 I found myself adding NIXL-specific semantics to the contents of `finished_recving`, which was only possible because I happened to be able to use the `update_connector_output()` to extract the NIXL-specific data and translate the contents to what the scheduler expected. An obscure example, but I did think it was a weird contrast with the complete flexibility of connector-specific `KVConnectorMetadata` subclasses

> My view is that we should rename `KVConnectorStats` to `KVConnectorWorkerMeta` (which could support `get_stats() -> KVConnectorStats`). 

I'm not sure I agree with this, though:

1. Although `KVConnectorStats` is a recent addition, we should by default assume that we need to maintain compatibility and break any potential external connectors that might have already adopted
2. `KVConnectorStats` seems like a useful abstraction that should probably be adopted by all connectors over time
3. `KVConnectorStats` is very specifically about metrics and integration with Prometheus/console stat logging - for example the `aggregate()` and `reduce()` methods - so not easily repurposed as a generic worker->scheduler channel
4. `KVEventsBatch` also seems like a useful, connector-independent abstraction, that ties in nicely with the existing public KV events publishing API, so if connectors have to use a connector-specific metadata channel for them, we'd probably see a lot of duplication

But I do think it could be positive to add `KVConnectorWorkerMetdata` anyway ...

hickeyma:
Thanks @njhill @orozery @markmc @NickLucche for the feedback. I will get to it in due course as bit delayed this week because I'm at KubeCon.

> Honestly, I don't yet fully understand the motivation for the lmcache connector to emit its own events (as per https://github.com/vllm-project/vllm/pull/28252 and https://github.com/LMCache/LMCache/issues/1846) in addition to the KV events already emitted by vLLM ... so I don't have a strong sense either way whether this need is highly-specific to lmcache

@markmc The reason that a connector like LMCache would need to generate their own events is down to specific details or information. How LMcache stores the tokens/hashes (e.g. LMcache stores tokens with a granularity of 256 as a block and vLLM granularity is 16 as a block) and where it stores (e.g. LMcache might store them to disk) them will differ to vLLM. Therefore, a consumer of the events would be missing the full context of the cache if it only consumed the vLLM events.

hickeyma:
> I understand your point here, but I believe that adding more fields to KVConnectorOutput was kinda part of the original design where it was expected that it would naturally grow to contain more stuff (I think this PR is a good example of that).
Naturally nobody would want the struct to grow indefinitely if we sense that this is the direction in which we're heading.

I agree with @NickLucche about having a separate field. @orozery I understand you reservation about only 1 connector possibly using it but still think the separation is best to clearly identify it.

vllm/distributed/kv_transfer/kv_connector/v1/base.py
```diff
@@ -343,6 +343,12 @@ def get_kv_connector_stats(self) -> Optional["KVConnectorStats"]:
         """
         return None
 
+    def get_kv_connector_kv_cache_events(self) -> Optional["KVEventBatch"]:
+        """
+        Get the KV connector kv cache events collected during the last interval.
+        """
```
ApostaC: Can we be more specific about whether the scheduler or the worker calls this function?

hickeyma: @ApostaC Its part of the "Worker-side methods" section of https://github.com/vllm-project/vllm/blob/main/vllm/distributed/kv_transfer/kv_connector/v1/base.py. Are you asking for more context in the method doc string to include that it is called by the mode runner and when?

Maybe a sentence like the following:

`This function should be called by the model runner every time after the model execution and before cleanup.`?

hickeyma: Clarification added to method doc string.

vllm/distributed/kv_events.py
```diff
@@ -54,11 +54,30 @@ class BlockStored(KVCacheEvent):
...
+            return (
+                self.block_hashes == other.block_hashes
+                and self.parent_block_hash == other.parent_block_hash
+                and self.token_ids == other.token_ids
+                and self.block_size == other.block_size
+                and self.lora_id == other.lora_id
+                and self.medium == other.medium
+            )
+        return False
+
```
ApostaC: Do we really need to manually implement `__eq__`? The KV event classes extend from `msgspec.Struct`, which should automatically generate a `__eq__` function.

See [msgspec.Struct docs](https://jcristharif.com/msgspec/structs.html#:~:text=By%20default%20struct%20types%20define%20an%20__eq__%20method%20based%20on%20the%20type%20definition.%20This%20enables%20support%20for%20equality%20comparisons.%20Additionally%2C%20you%20may%20configure%20order%3DTrue%20to%20make%20a%20struct%20type%20orderable%20through%20generation%20of%20__lt__%2C%20__le__%2C%20__gt__%2C%20and%20__ge__%20methods.)

ApostaC (commented):
Otherwise LGTM

Hey @orozery please also take a look

orozery:
I'm trying to understand `combine_unique_ordered_events`:
Each worker can output, e.g., `BlocksStored`.
Assuming a single worker stored blocks for a specific block. Wouldn't you need to somehow count that all workers also stored the same block before reporting this event?

vllm/distributed/kv_events.py
hickeyma: Well spotted, thanks. Can use the equality method from the `msgspec.Struct` parent class instead.

hickeyma: Updated.

vllm/v1/worker/kv_connector_model_runner_mixin.py
```diff
@@ -135,10 +136,19 @@ def _get_kv_connector_output(
             output.kv_connector_stats = (
                 KVConnectorModelRunnerMixin.get_kv_connector_stats()
             )
+            output.kv_cache_events = (
+                KVConnectorModelRunnerMixin.get_kv_connector_kv_cache_events()
+            )
```
sdavidbd: These static methods seem redundant. We can simply call the connector methods directly here.

vllm/distributed/kv_events.py
```diff
@@ -67,6 +67,17 @@ class AllBlocksCleared(KVCacheEvent):
...
+    def combine_unique_ordered_events(self, other: "KVEventBatch") -> "KVEventBatch":
+        """
+        Combine non duplicated events with another `KVEventBatch` object.
+        """
+        combined_events = self.events[:]
+        for item in other.events:
+            if item not in combined_events:
+                combined_events.append(item)
+        self.events = combined_events
+        return self
```
sdavidbd: The current approach is **O(N×M)** because each `item not in combined_events` does a linear scan. We can reduce this to **O(N+M)** and simplify the logic by using a `set` for membership checks. Also, since we're already mutating the batch, we can append directly to `self.events` instead of building a copy. For example:

```
seen = set(self.events)
for item in other.events:
    if item not in seen:
        self.events.append(item)
        seen.add(item)
return self
```

sdavidbd:
> > Also note that the current PR does not add any Scheduler logic to process the kv_cache_events field. As a result, this data is never consumed and the field is effectively unused. It seems that some Scheduler handling would be needed before this can be considered ready for merge.
> 
> @sdavidbd See #28252

I stand corrected.

vllm/v1/worker/kv_connector_model_runner_mixin.py
hickeyma: Good catch. removed the methods.

vllm/distributed/kv_events.py
hickeyma: Yes, this would be more efficient for large lists. Good suggestion, thanks.

KuntaiDu (approved):
LGTM! @NickLucche could you please also take a look?

vllm/v1/worker/kv_connector_model_runner_mixin.py
```diff
@@ -138,17 +137,17 @@ def _get_kv_connector_output(
             )
             output.invalid_block_ids = kv_connector.get_block_ids_with_load_errors()
 
-            output.kv_connector_stats = (
-                KVConnectorModelRunnerMixin.get_kv_connector_stats()
-            )
+            if has_kv_transfer_group():
```
orozery: I don't think this check is necessary

vllm/v1/worker/kv_connector_model_runner_mixin.py
```diff
@@ -138,17 +137,17 @@ def _get_kv_connector_output(
...
+                output.kv_connector_stats = (
+                    get_kv_transfer_group().get_kv_connector_stats()
```
orozery: I think you can replace `get_kv_transfer_group()` with `kv_connector` here and down below.

vllm/distributed/kv_transfer/kv_connector/utils.py
```diff
@@ -201,16 +202,36 @@ def update_finished_set(
...
+                # Use the first worker's kv_cache events as start event list.
+                combined_kv_cache_events = kv_output.kv_cache_events
+            elif kv_cache_events := kv_output.kv_cache_events:
+                assert isinstance(
+                    combined_kv_cache_events,
+                    type(kv_cache_events),
+                )
+                worker_kv_cache_events = kv_cache_events.get_all_events()
+                combined_kv_cache_events.add_events(worker_kv_cache_events)
+                combined_kv_cache_events.increment_workers()
```
orozery: if a specific worker did not emit events on this step (`kv_output.kv_cache_events = None`), you will not account for it.
I think the expected worker set count is fixed and should not be counted by those worker who did not return None.

orozery (commented):
@hickeyma Thanks for your hard work!

I think that the changes introduced here add up code to vLLM that could have reside inside the LMCache connector implementation.
Specifically, the changes to `KVConnectorOutput`, `KVConnectorBase_V1` and the introduction of `KVEventAggregator`.
I do not think this changes, although trying to be generic, will fit any other use-case.

From another viewpoint, I expect `KVConnectorOutput` to contain only fields that are of interest to vLLM's scheduler (e.g. `finished_sending`, `finished_recving`, `invalid_block_ids`, etc.). 
The new `KVConnectorOutput.kv_cache_events` field introduced here is only consumed by the LMCache connector. From the vLLM point of view it's a don't care.
I think that making this field opaque behind a connector-specific metadata is the right solution.

I discussed this in person with @hickeyma a few weeks ago.
I think the bottom-line was:
1. @hickeyma supports a solution along the lines I've drawn.
2. @hickeyma prefers that we proceed with the current approach as a start, and perhaps change the implementation later.
3. I said I have no strong objection with merging the current approach.

I don't think taking the other approach requires a lot of coding effort.
I would still prefer making these changes now, instead of introducing API changes that may revert in the future.

Anyhow, please don't consider my opinion as a blocker for proceeding here.

vllm/v1/worker/kv_connector_model_runner_mixin.py
hickeyma: Was using previous code but looks like not required. Nice catch. Removed.

vllm/v1/worker/kv_connector_model_runner_mixin.py
hickeyma: Good suggestion. Done.

vllm/distributed/kv_transfer/kv_connector/utils.py
hickeyma: Yes, that's a good point. Updated to use the worker count.

hickeyma:
Thank you @orozery for review, comments and feedback. I have pushed changes for your review comments.

> I think that the changes introduced here add up code to vLLM that could have reside inside the LMCache connector implementation.
Specifically, the changes to KVConnectorOutput, KVConnectorBase_V1 and the introduction of KVEventAggregator.
I do not think this changes, although trying to be generic, will fit any other use-case.
From another viewpoint, I expect KVConnectorOutput to contain only fields that are of interest to vLLM's scheduler (e.g. finished_sending, finished_recving, invalid_block_ids, etc.).
The new KVConnectorOutput.kv_cache_events field introduced here is only consumed by the LMCache connector. From the vLLM point of view it's a don't care.
I think that making this field opaque behind a connector-specific metadata is the right solution.

I appreciate your feedback and the points raised with regard to this. I wanted however to follow the pattern currently in place for stats. For stats, changes were added to `KVConnectorOutput`, `KVConnectorBase_V1`, and `KVConnectorStats` was introduced. I think it is a flexible design to make this generic for now which gives other connectors the ability to provide KV events if they want to.

vllm/distributed/kv_transfer/kv_connector/utils.py
```diff
@@ -201,16 +202,36 @@ def update_finished_set(
...
 
         # select output of the worker specified by output_rank
         output = outputs[output_rank]
 
+        # Aggregate the events across workers.
+        # This operation needs to be done post worker processing so that we have all
+        # events for all workers.
+        if combined_kv_cache_events is not None:
+            combined_kv_cache_events.set_workers(self._expected_finished_count)
+            combined_kv_cache_events = combined_kv_cache_events.aggregate()
```
orozery: `combined_kv_cache_events` holds up information on worker events on this step.
Events that did not accumulate for enough workers will be lost after this step.
You need to keep track of those events cross-steps.
See for example how they do it for `finished_recving` (using `self._recv_remaining_count`).

hickeyma: That's interesting, thanks for letting me know and explaining this on slack.

I have pushed 717702fe5d6a41c6607ffe8c5c2736674e47ca86 which persists events across steps which is requied if workers are not processed in the same step. It then resets the events once the scheduler retrieves them in `take_events()`.

NickLucche (commented):
Thanks for the work @hickeyma , and apologies for the delay in getting back to this PR.
This is an LGTM re: interface changes for me.

I don't mean to delay this contrib further, however it would be really nice to have any sort of unit tests here, especially for the MultiConnector case.

ApostaC:
Hey @hickeyma , can you do "update branch" to include the fix of LMCache unit tests? Otherwise, this looks good to me!

ApostaC (approved):
Thanks for the contribution and thanks @KuntaiDu @NickLucche for reviewing!

requirements/test.txt
```diff
@@ -506,6 +525,10 @@ networkx==3.2.1
     # via
     #   scikit-image
     #   torch
+nixl==0.8.0
+    # via lmcache
```
NickLucche: this version is going to be conflicting with the one we have in requirements/kv_connectors.txt
Any chance we can avoid installing nixl through lmcache?

requirements/test.txt
```diff
@@ -1017,8 +1055,12 @@ sentence-transformers==3.2.1
 setuptools==77.0.3
     # via
     #   lightning-utilities
+    #   lmcache
     #   pytablewriter
+    #   setuptools-scm
     #   torch
+setuptools-scm==9.2.2
+    # via lmcache
```
NickLucche: same thing for scm here tbh

tests/v1/kv_connector/unit/test_lmcache_connector.py
```diff
@@ -0,0 +1,756 @@
...
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+from unittest.mock import MagicMock
+
+import pytest
+
+from vllm.distributed.kv_events import BlockStored
+from vllm.distributed.kv_transfer.kv_connector.v1.lmcache_connector import (
+    LMCacheConnectorV1,
+    LMCacheKVEvents,
+)
```
NickLucche, line 11: qq: is the lmcache package actually needed to run these tests or is it just for importing the types to insert in the mocks?

NickLucche (commented):
I think the only missing part here is to provide a generic implementation of `get_kv_connector_kv_cache_events` for the MultiConnector, such that it handles getting these events from multiple connectors, handling the case where only a subset of the requested connectors implement the `get_kv_connector_kv_cache_events` method (+quick test case).

I believe changes to `multi_connector.py` here https://github.com/vllm-project/vllm/pull/22188/ could be a good reference. 
@hickeyma

tests/v1/kv_connector/unit/test_lmcache_connector.py
hickeyma: Good catch, thanks @NickLucche. Updated to do a lazy import in the connector and therefore not requiring lmcache package during the mocking.

requirements/test.txt
hickeyma: See response in https://github.com/vllm-project/vllm/pull/28309#discussion_r2609949690 where lmcache package no longer required for unit test.

requirements/test.txt
hickeyma: See response in https://github.com/vllm-project/vllm/pull/28309#discussion_r2609949690 where lmcache package no longer required for unit test.

NickLucche (approved):
LGTM as per @hickeyma comment on follow up work

PR #28315, opened 2025-11-07
[BugFix] Temporary fix for IMA with MTP = 2 and full-cg

LucasWilkinson:
Temporary fix for https://github.com/vllm-project/vllm/issues/28207

For now just make sure when spec-decode is enabled that the cudagraph shapes are evenly divisible by `1 + num_speculative_tokens`; see https://github.com/vllm-project/vllm/issues/28207 for more details

 ### Test 1
Tested with 
```
MODEL=deepseek-ai/DeepSeek-R1
VLLM_ATTENTION_BACKEND="FLASHINFER_MLA" \
VLLM_USE_V1=1 \
vllm serve $MODEL \
--tensor-parallel-size 8 \
--disable-log-requests \
--no-enable-prefix-caching \
--compilation-config '{"cudagraph_mode": "FULL_AND_PIECEWISE"}' \
--trust-remote-code \
--block-size 64 \
--kv-cache-dtype fp8 \
--speculative-config='{"method": "deepseek_mtp", "num_speculative_tokens": 2}' 
```
```
MODEL=deepseek-ai/DeepSeek-R1
lm_eval \
    --model local-completions \
    --tasks gsm8k \
    --model_args model=${MODEL},base_url=http://127.0.0.1:8000/v1/completions \
    --batch_size 100
```

Hits an IMA on main but not on this branch

 ### Test 2
```
vllm serve deepseek-ai/DeepSeek-R1 \
  --tensor-parallel-size 8 \
  --compilation-config '{"cudagraph_mode": "FULL_AND_PIECEWISE", "pass_config": {"enable_sequence_parallelism": true}}' \
  --speculative-config='{"method": "deepseek_mtp", "num_speculative_tokens": 2}'
```
Correctly fails with
```
ValueError: Can't determine cudagraph shapes that are both a multiple of 3 
(num_speculative_tokens + 1) required by spec-decode and 8 (tensor_parallel_size) 
required by sequence parallelism please adjust num_speculative_tokens or disable 
sequence parallelism
```


ProExpertProg (approved):
This might not be compatible with sequence parallelism but that's for high-throughput cases anyway, just might be worth adding a warning.

LucasWilkinson:
> This might not be compatible with sequence parallelism but that's for high-throughput cases anyway, just might be worth adding a warning.

Done :+1:

mgoin (approved):

vllm/config/compilation.py
```diff
@@ -922,3 +912,64 @@ def custom_op_log_check(self):
...
+                or multiple_of % tensor_parallel_size != 0
+            ):
+                raise ValueError(
+                    f"Can't determine cudagraph shapes that are both a "
+                    f"multiple of {uniform_decode_query_len} "
+                    f"(num_speculative_tokens + 1) required by spec-decode "
+                    f"and {tensor_parallel_size} (tensor_parallel_size) "
+                    f"required by sequence parallelism please adjust "
+                    f"num_speculative_tokens or disable sequence parallelism"
+                )
```
pisceskkk, line 933: @LucasWilkinson Just curious, do we need to enforce that `uniform_decode_query_len` is a factor of `tensor_parallel_size`, or vice versa? Or do we only need to ensure that the following cudagraph sizes are the common multiple of the first two?

PR #26440, opened 2025-10-08
[Performance] Dual stream execution of "shared_experts" and "selected_experts" inside FusedMoE

alexm-redhat:
This PR executes the shared_experts part of the FusedMoE on a separate GPU stream, so that the execution is parallelized with the "selected_experts" part. This is possible since the outputs of both are independent and are later combined. Thanks @wenscarl for pointing this out.

For DeepSeekR1 FP8 with Flashinfer latency kernels (trtllm-gen) on 8xB200s batch size 32, the TPOT improves from 23.35ms to 22.09ms (with latest FlashInfer codebase), so about ~5.7% e2e improvement.

PR #28324, opened 2025-11-07
[CI/Build] Temporary fix to LM Eval Small Models

zhewenl:
## Purpose
This PR provides a temp fix to broken CI: [LM Eval Small Models](https://buildkite.com/vllm/ci/builds/37631#019a5264-3636-4617-87f8-9867066b7a78), the test is failing with GSM8K evals on `Qwen1.5-MoE-W4A16-CT-tp1`:
```

[2025-11-05T06:16:56Z]             # Verify accuracy is within tolerance
--
  | [2025-11-05T06:16:56Z] >           assert measured_accuracy >= expected_accuracy - RTOL, (
  | [2025-11-05T06:16:56Z]                 f"Accuracy too low: {measured_accuracy:.3f} < "
  | [2025-11-05T06:16:56Z]                 f"{expected_accuracy:.3f} - {RTOL:.3f}"
  | [2025-11-05T06:16:56Z]             )
  | [2025-11-05T06:16:56Z] E           AssertionError: Accuracy too low: 0.000 < 0.450 - 0.080
  | [2025-11-05T06:16:56Z] E           assert 0.0 >= (0.45 - 0.08)
  | [2025-11-05T06:16:56Z]
  | [2025-11-05T06:16:56Z] evals/gsm8k/test_gsm8k_correctness.py:86: AssertionError

```

After bisecting, we found this PR could contribute to failing tests: https://github.com/vllm-project/vllm/pull/26440/ - disabling the feature with `VLLM_DISABLE_SHARED_EXPERTS_STREAM=1` can make test passed consistently. 

Note: this is only easy to reproduce on Nvidia L4(what CI is using), H100/MI300 is very difficult to repro this: `pytest -s -v 'tests/evals/gsm8k/test_gsm8k_correctness.py::test_gsm8k_correctness_param[Qwen1.5-MoE-W4A16-CT-tp1]'` 

Some hypothesis on why is this failing:
- Quantization Incompatibility? The test is using W4A16 quantization, so the previous PR might break tensor properties with `hidden_states.clone()`?
- It might be breaking memory layout in Qwen2 MoE model [forward pass](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/qwen2_moe.py#L170)?

## Test Plan
With `VLLM_DISABLE_SHARED_EXPERTS_STREAM=1`, tested 10 times(10/10 are passing): https://gist.github.com/zhewenl/a23ead4262f500b764ecd56c8c4c213d

Without disabling the env var, 3/3 are failing: https://gist.github.com/zhewenl/01a70feea4de15f6258dbecd76415a22

### DIFF
--- a/.buildkite/test-pipeline.yaml
+++ b/.buildkite/test-pipeline.yaml
@@ -599,6 +599,7 @@ steps:
   - csrc/
   - vllm/model_executor/layers/quantization
   commands:
+  - export VLLM_DISABLE_SHARED_EXPERTS_STREAM=1
   - pytest -s -v evals/gsm8k/test_gsm8k_correctness.py --config-list-file=configs/models-small.txt --tp-size=1
 
 - label: OpenAI API correctness # 22min
@@ -1224,7 +1225,7 @@ steps:
     - pytest -v -s tests/compile/test_fusions_e2e.py::test_tp2_attn_quant_allreduce_rmsnorm
     - pytest -v -s tests/distributed/test_context_parallel.py
     - CUDA_VISIBLE_DEVICES=1,2 VLLM_ALL2ALL_BACKEND=deepep_high_throughput VLLM_USE_DEEP_GEMM=1 VLLM_LOGGING_LEVEL=DEBUG python3 examples/offline_inference/data_parallel.py --model Qwen/Qwen1.5-MoE-A2.7B --tp-size=1  --dp-size=2 --max-model-len 2048
-    - pytest -v -s tests/v1/distributed/test_dbo.py  
+    - pytest -v -s tests/v1/distributed/test_dbo.py
 
 ##### B200 test #####
 - label: Distributed Tests (B200) # optional
--- a/.buildkite/test-pipeline.yaml
+++ b/.buildkite/test-pipeline.yaml
@@ -599,7 +599,6 @@ steps:
   - csrc/
   - vllm/model_executor/layers/quantization
   commands:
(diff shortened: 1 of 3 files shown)


zhewenl:
On a side note, since the root cause and the impact is still unknown - shall we temporally disable `VLLM_DISABLE_SHARED_EXPERTS_STREAM `? It's enabled globally as default: https://github.com/vllm-project/vllm/blob/da786e339e985cbf32f43474e7b88b8da65c76e5/vllm/envs.py#L221

zhewenl:
> Do I understand correctly that we add `VLLM_DISABLE_SHARED_EXPERTS_STREAM=1` for all tests? 

For all the tests under LM Eval Small Models

yeqcharlotte:
ok does this mean the failure is specific to L4 gpu? i think the duo stream still need some ci coverage. 

can we reorganize these testing pipelines so we can separate testing that must be run on H100/B200 vs. L4. @hl475 @rzabarazesh please consider these in the test refactoring!

mgoin (commented):
Can you please add it to the Blackwell lm eval job too?

vadiklyutiy:
I think we should add it on pytest level. Otherwise running tests in CI/buildkite environment will be different from running pytest locally.

vadiklyutiy:
I can confirm that adding `VLLM_DISABLE_SHARED_EXPERTS_STREAM=1` fix the issue(did a lot of runs).

mgoin (approved):

PR #28327, opened 2025-11-07
[Core] Simplify async KV output aggregation

njhill:
Consolidate with MP `FutureWrapper` added in https://github.com/vllm-project/vllm/pull/28319, avoid fragile double future wrapping.


mgoin (approved):
LGTM

tests/v1/kv_connector/unit/test_output_aggregator.py
```diff
@@ -86,74 +85,6 @@ def test_aggregate_workers_output():
     assert aggregated.invalid_block_ids == {3, 4, 5}
 
 
-def test_async_aggregate_workers_output():
```
NickLucche, line 89: why are we dropping a test?

njhill: The method that this test is testing no longer exists.

PR #28329, opened 2025-11-07
[Misc] Add more scoping for improved trace

frank-wei:
Summary: Add more scoping in the hotspot area which can greatly help us to the find the cycle heavy area

Differential Revision: D86436375

image

image


bangshengtang:
nit: can we try to unify the text format? we have all lower, all upper, first letter capitalized in three different layers

zhuohan123 (approved):

vllm/v1/engine/llm_engine.py
```diff
@@ -280,28 +281,32 @@ def step(self) -> list[RequestOutput | PoolingRequestOutput]:
...
 
         # 4) Record stats
-        if self.logger_manager is not None and outputs.scheduler_stats is not None:
-            self.logger_manager.record(
-                scheduler_stats=outputs.scheduler_stats,
-                iteration_stats=iteration_stats,
-                mm_cache_stats=self.processor.stat_mm_cache(),
-            )
-            self.do_log_stats_with_interval()
+        with record_function_or_nullcontext("llm_genine step: record_stats"):
```
benchislett, line 302: Typo here and above

frank-wei: fixed in https://github.com/vllm-project/vllm/pull/28584

Issue #28262, opened 2025-11-07
[Bug]: [gpt-oss] Responses API incorrect input/output handling


alecsolder:
### Your current environment

Any env

### 🐛 Describe the bug

There is currently an implementation issue with gpt-oss on the Responses API in vLLM. This can be seen clearly in the [test which continues a conversation between API requests here](https://github.com/vllm-project/vllm/blob/4bf56c79cc252d285d0cb4f5edf323f02af735ca/tests/entrypoints/openai/test_response_api_with_harmony.py#L715).

From the first request, the model outputs the following tokens (whitespace added for clarity):
```
<|channel|>analysis<|message|>
	User asks for weather in Paris today. We have no direct API call yet, but we can use get_weather function. Coordinates for Paris: latitude 48.8566, longitude 2.3522. We'll call get_weather.
<|end|>
<|start|>assistant<|channel|>commentary to=functions.get_weather <|constrain|>json<|message|>
	{"latitude":48.8566,"longitude":2.3522}
<|call|>
```
When the output items from the first request are passed in as input to the second request, the tokens look like this (whitespace added for clarity):
```
<|start|>user<|message|>
	What's the weather like in Paris today?
<|end|>
<|start|>assistant<|message|>
	User asks for weather in Paris today. We have no direct API call yet, but we can use get_weather function. Coordinates for Paris: latitude 48.8566, longitude 2.3522. We'll call get_weather.
<|end|>
<|start|>assistant to=functions.get_weather<|channel|>commentary json<|message|>
	{"latitude":48.8566,"longitude":2.3522}
<|call|>
<|start|>functions.get_weather<|message|>
	20
<|end|>
```

We lose `<|channel|>analysis` on the reasoning message, and we do not set `<|channel|>commentary` on the tool call output ([documentation reference](https://cookbook.openai.com/articles/openai-harmony#handling-tool-calls)).

There are a lot of edge cases and challenges to properly represent Harmony Message metadata when the Responses API input/output types do not include that metadata, but we can improve on the current implementation. 

The changes we can make are:
- A reasoning message should use the channel of the message that follows it. For example:
  - The reasoning message prior to a function tool call should be on the commentary channel
  - If the commentary channel is not enabled (no function tools enabled), all reasoning messages are on the analysis channel
  - All other reasoning messages are on the analysis channel
- Set the content_type for function tools to be `<|constrain|>json` always
- Input items which are FunctionCallOutput should be set to be on the commentary channel
[...]

PR #28355, opened 2025-11-09
Fix/responses api harmony channel metadata    #28262

baonudesifeizhai:
## Purpose
Fix issue #28262: Restore missing channel metadata when converting Responses API output items back to Harmony Messages for multi-turn conversations.
Changes:
Set channel='commentary' for function_call_output type inputs
Set channel='analysis' or 'commentary' for reasoning type based on the following message (commentary if followed by function_call, analysis otherwise)
Add test to verify channel metadata is correctly preserved across conversation turns
Update parse_response_input() to accept optional next_msg parameter for context-aware channel assignmen
## Test Plan
pytest tests/entrypoints/openai/test_response_api_with_harmony.py::test_function_call_with_previous_input_messages -v
## Test Result
pass 
---
<details>

</details>


tests/entrypoints/openai/test_response_api_with_harmony.py
```diff
@@ -957,3 +957,55 @@ async def test_function_call_with_previous_input_messages(
     assert (
         "aquarius" in output_text or "otter" in output_text or "tuesday" in output_text
     )
+
+    # ============================================================
+    # Test for issue #28262: parse_response_input channel metadata
+    # ============================================================
+    from vllm.entrypoints.harmony_utils import parse_response_input
```
alecsolder: You should be able to do this a little more easily by looking at the channel field on the input_messages field of response2 output

kind of like in [here](https://github.com/vllm-project/vllm/blob/34553b9d2702dd2a27a578fec819e88e76dcbfb4/tests/entrypoints/openai/test_response_api_mcp_tools.py#L108) 

[You just need to add this to the request](https://github.com/vllm-project/vllm/blob/34553b9d2702dd2a27a578fec819e88e76dcbfb4/tests/entrypoints/openai/test_response_api_mcp_tools.py#L81)

vllm/entrypoints/harmony_utils.py
```diff
@@ -184,6 +184,7 @@ def get_user_message(content: str) -> Message:
 def parse_response_input(
     response_msg: ResponseInputOutputItem,
     prev_responses: list[ResponseOutputItem | ResponseReasoningItem],
+    next_msg: ResponseInputOutputItem | None = None,
```
alecsolder: Since we have previous and next, would it make more sense to just pass the list of items and the current index? Then we could grab prev and next items here instead of passing in all 3 at the same time

vllm/entrypoints/harmony_utils.py
```diff
@@ -221,10 +222,22 @@ def parse_response_input(
...
     elif response_msg["type"] == "reasoning":
         content = response_msg["content"]
         assert len(content) == 1
         msg = Message.from_role_and_content(Role.ASSISTANT, content[0]["text"])
+        # Issue #28262: Reasoning channel should match the following message
+        # - If followed by function_call, use commentary channel
+        # - Otherwise, use analysis channel (default)
+        channel = "analysis"
+        if next_msg is not None:
+            next_msg_dict = (
```
alecsolder: can we do type checking the same way as the other cases here?

vllm/entrypoints/harmony_utils.py
```diff
@@ -221,10 +222,22 @@ def parse_response_input(
             Author.new(Role.TOOL, f"functions.{call_response.name}"),
             response_msg["output"],
         )
+        msg = msg.with_channel("commentary")
```
alecsolder: Do you want to add the content_type change from the issue here too? Or are you just tackling one specific part of the issue?

alecsolder (commented):
Thank you for this! Left a few comments.

vllm/entrypoints/harmony_utils.py
baonudesifeizhai: msg = msg.with_channel("commentary")
 
msg = msg.with_content_type("json")  -- is that enough?

baonudesifeizhai:
pytest tests/entrypoints/openai/test_response_api_with_harmony.py::test_function_call_with_previous_input_messages -v pass

PR #28356, opened 2025-11-09
add cpu option for p/d in nixl_connector

ZhengHongming888:
## Purpose
This PR is the further update version of [PR 27510](https://github.com/vllm-project/vllm/pull/27510). 

The PRs is to add the 'cpu' option for P/D disaggregation or KV transfer based on nixl_connector in pure CPU machine/environment.

After you adding with this option you can do the vllm serve in pure cpu machine or docker image as below -
`VLLM_SKIP_WARMUP=True vllm serve meta-llama/Llama-3.2-3B-Instruct --port 8200 --gpu-memory-utilization 0.9 --enforce-eager --tensor-parallel-size 1 --max_model_len 4096 --kv-transfer-config '{"kv_connector":"NixlConnector","kv_role":"kv_both","kv_buffer_device":"cpu"}'`

which can help on the construction of heterogenous system between cpu and all other gpus based on like llm-d framework.

In addition in the first commit of the previous PR we make nixl connector accept CPU host and set self.use_host_buffer = True when only kv_buffer_device == "cpu", which will happen with the error because for pure cpu environment we don't have calls sync_recved_kv_to_device/save_kv_to_host invoked from copy_kv_block. 

In order to solve this we updated the previous PR with add one more option judge to see whether current_device is cpu or not as below -
`if current_platform.device_type=='cpu':
     self.use_host_buffer = False
else:
    self.use_host_buffer = vllm_config.kv_transfer_config.kv_buffer_device == "cpu"`

This will set self.use_host_buffer = False and all other nixl and P/D behavior same as other GPUs. You can see the kv transfer happened in /workspace/vllm/tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh (p/d with toy_proxy_server.py) and the output log as below -
image

please check. Thanks.

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
+++ b/vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
@@ -1002,12 +1002,14 @@ class NixlConnectorWorker:
 
     def set_host_xfer_buffer_ops(self, copy_operation: CopyBlocksOp):
         """Assign copy (d2h, h2d) operations when host buffer is used."""
-        # Set a no-op if the host buffer is not cpu and self.device_type is 'cpu'.
-        if self.kv_buffer_device != "cpu" or self.device_type == "cpu":
+        # Set a no-op if the host buffer is not cpu.
+        if self.kv_buffer_device != "cpu":
             return
-        if self.device_type != "cpu":
-            assert self.use_host_buffer
-            self.copy_blocks = copy_operation
+        # Set a no-op if self.device_type is 'cpu'.
+        if self.device_type == "cpu":
+            return
+        assert self.use_host_buffer
+        self.copy_blocks = copy_operation
 
     def _background_nixl_handshake(
         self, req_id: str, remote_engine_id: EngineId, meta: ReqMeta
--- a/vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
+++ b/vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
@@ -90,6 +90,7 @@ _NIXL_SUPPORTED_DEVICE = {
     ),
     "tpu": ("cpu",),
     "xpu": ("cpu",),
+    "cpu": ("cpu",),
 }
 # support for oot platform by providing mapping in current_platform
 _NIXL_SUPPORTED_DEVICE.update(current_platform.get_nixl_supported_devices())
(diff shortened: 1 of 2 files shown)


markmc:
Please revert the formatting changes to `vllm/third_party/pynvml.py`

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -994,8 +1005,9 @@ def set_host_xfer_buffer_ops(self, copy_operation: CopyBlocksOp):
         # Set a no-op if the host buffer is not cpu.
         if self.kv_buffer_device != "cpu":
```
NickLucche: nit: can we return when device is cpu and expand comment above? I think it's just a bit clearer to group exit conditions.

```suggestion
        if self.kv_buffer_device != "cpu" or self.device_type == "cpu":
```
and drop the if below

NickLucche (commented):
Thanks for the PR @ZhengHongming888 !
I think this may have some utility for testing.

It looks ok, left one minor note, have you managed to test this in practice with 1p1d?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
ZhengHongming888: Thanks @NickLucche yes agree that we can have 'if self.kv_buffer_device != "cpu" or self.device_type == "cpu":' and it will return when self.device_type == "cpu". 

But it can not be dropped 'the if below' because there is the cpu offload mode for gpu device and it will use the use_host_buffer == True mode and setup the copy_operation like here 'self.copy_blocks = copy_operation'. ..

Thanks for your suggestion.

ZhengHongming888: Done with adding 'if self.kv_buffer_device != "cpu" or self.device_type == "cpu":' in nixl_connector.py. thanks.

ZhengHongming888: Done!

xuechendi (approved):
LGTM

xuechendi:
@bigPYJ1151 @NickLucche @njhill , we have verified the pr locally, please help to review, thanks so much

bigPYJ1151 (approved):
LGTM. 
But there is a issue when use CPU and other devices for PD disagg. The CPU attention backend will pack the KV cache block, even the logical shape is ```HND```. So cache blocks from other backends should be packed before feed into the CPU attention.
We should be aware of the issue for now.

NickLucche (approved):
Thanks for your work!

louie-tsai:
@louie-tsai

PR #28358, opened 2025-11-09
[Performance][B200] silu_mul_quant: pack scales in int32

varun-sundar-rabindranath:
## Purpose
This PR focuses on the deepep_low_latency All2All code path.

DeepGEMM expects the activation scales to be packed in int32 format. Given unpacked scales, DeepGEMM performs scale transformation itself. This transformation by DeepGEMM is slow and  delays the kernel execution unnecessarily. 

This PR fuses the scale transformation in the silu_mul_quant kernel thereby eliminating a few torch operations in the hot path.

Main
act-mul-main
PR
act-mul-pr

## Test Plan
- Add unit tests for the transformation
- E2E : 
    command : ` VLLM_ALL2ALL_BACKEND=deepep_low_latency  VLLM_USE_DEEP_GEMM=1 canhazgpu run -g2 --  vllm serve Qwen/Qwen3-30B-A3B-FP8 --trust-remote-code --tensor-parallel-size 1 --data-parallel-size 2 --enable-expert-parallel --no-enable-prefix-caching  --port 9010`
    lm_eval : `lm_eval --model local-completions --tasks gsm8k --model_args model=Qwen/Qwen3-30B-A3B-FP8,base_url=http://localhost:9010/v1/completions,num_concurrent=30,max_retries=3 --limit 100`

## Test Result
- Unit tests pass
- E2E passes with 
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  | 0.89|±  |0.0314|
|     |       |strict-match    |     5|exact_match|↑  | 0.89|±  |0.0314|
```

### DIFF
--- a/csrc/quantization/activation_kernels.cu
+++ b/csrc/quantization/activation_kernels.cu
@@ -279,17 +279,17 @@ __device__ __forceinline__ void token_bounds(int32_t n_tokens,
 }
 
 template <int BLOCK_COUNT, int SMEM_SIZE_BYTES_Y, typename fp8_type,
-          int THREADS, typename Idx_t, bool USE_UE8M0, int GROUP_SIZE = 128,
-          int NUM_STAGES = 3>
+          typename scale_t, int THREADS, typename Idx_t, bool CEIL_UE8M0,
+          int GROUP_SIZE = 128, int NUM_STAGES = 3>
 __global__ void silu_mul_fp8_quant_deep_gemm_kernel(
     const __nv_bfloat16* __restrict__ _input, fp8_type* __restrict__ _y_q,
-    float* __restrict__ _y_s, const int32_t* __restrict__ tokens_per_expert,
+    scale_t* __restrict__ _y_s, const int32_t* __restrict__ tokens_per_expert,
     // sizes
     Idx_t E, Idx_t T, Idx_t H,
     // strides (in elements)
     Idx_t stride_i_e, Idx_t stride_i_t, Idx_t stride_i_h, Idx_t stride_yq_e,
     Idx_t stride_yq_t, Idx_t stride_yq_h, Idx_t stride_ys_e, Idx_t stride_ys_t,
-    Idx_t stride_ys_g, Idx_t stride_counts_e) {
+    Idx_t stride_ys_g, Idx_t stride_ys_p, Idx_t stride_counts_e) {
 #ifndef USE_ROCM
   static constexpr int NUM_WARPS = THREADS / WARP_SIZE;
 
@@ -466,9 +466,22 @@ __global__ void silu_mul_fp8_quant_deep_gemm_kernel(
 
   __nv_fp8x4_e4m3* y_q_base_ptr =
       reinterpret_cast<__nv_fp8x4_e4m3*>(_y_q) + lane_id;
-  auto y_scale_base_ptr = _y_s + warp_position_scales * stride_ys_g;
+
+  Idx_t scale_group_offset = 0;
+  if constexpr (std::is_same<scale_t, uint8_t>::value) {
+    // packed int32_t format
+    int pack_id = warp_position_scales / 4;
+    int scale_in_pack = warp_position_scales % 4;
+    scale_group_offset = pack_id * stride_ys_p + scale_in_pack * stride_ys_g;
+  } else {
+    scale_group_offset = warp_position_scales * stride_ys_g;
+  }
+
+  scale_t* const y_scale_base_ptr = _y_s + scale_group_offset;
 
   for (auto j = tokens_lower; j < tokens_upper; j++) {
+    int current_group_id = warp_position_scales;  // Running count of which
+                                                  // group is being processed
     const Idx_t base_ys = expert_id * stride_ys_e;
     auto y_s_ptr = y_scale_base_ptr + base_ys + token_offset * stride_ys_t;
     __nv_fp8x4_e4m3* y_q_ptr =
@@ -509,7 +522,7 @@ __global__ void silu_mul_fp8_quant_deep_gemm_kernel(
 
       __nv_bfloat16 y_s = __hmul(warp_max(_y_max2.x), fp8_inv);
 
-      if constexpr (USE_UE8M0) {
+      if constexpr (CEIL_UE8M0) {
         y_s = hexp2(hceil(hlog2(y_s)));
       }
 
@@ -527,8 +540,24 @@ __global__ void silu_mul_fp8_quant_deep_gemm_kernel(
       y_q_ptr += WARP_SIZE * stride_yq_h;
 
       if (!lane_id) {
-        *y_s_ptr = y_s;
-        y_s_ptr += stride_ys_g;
+        // Store scales.
+        if constexpr (std::is_same<scale_t, uint8_t>::value) {
+          // Packed UE8MO format. Remove Mantissa.
+          *y_s_ptr = reinterpret_cast<int16_t&>(y_s) >> 7;
+
+          bool const jump_pack = (current_group_id + 1) % 4 == 0;
+          // Minus 3 because we need to get to the first group in the
+          // next pack.
+          y_s_ptr += jump_pack ? (stride_ys_p - 3) : stride_ys_g;
+
+        } else {
+          // float32 format
+          static_assert(std::is_same<scale_t, float>::value);
+          *y_s_ptr = y_s;
+          y_s_ptr += stride_ys_g;
+        }
+
+        current_group_id += 1;
       }
     }
   }
@@ -573,7 +602,7 @@ void persistent_masked_m_silu_mul_quant(
     const at::Tensor& tokens_per_expert,  // (E)
     at::Tensor& y_q,                      // (E, T, H) [OUT]
     at::Tensor& y_s,                      // (E, T, H//group_size) [OUT]
-    bool use_ue8m0) {
+    bool cast_scale_ue8m0) {
 #ifndef USE_ROCM
 
   // This kernel currently only supports H % 128 == 0 and assumes a
@@ -583,8 +612,11 @@ void persistent_masked_m_silu_mul_quant(
   TORCH_CHECK(input.dtype() == torch::kBFloat16);
   TORCH_CHECK(y_q.dtype() == torch::kFloat8_e4m3fn ||
               y_q.dtype() == torch::kFloat8_e4m3fnuz);
-  TORCH_CHECK(y_s.dtype() == torch::kFloat32);
   TORCH_CHECK(input.size(-1) % (GROUP_SIZE * 2) == 0);
+
+  bool const is_packed_ue8m0 =
+      (y_s.dtype() == torch::kInt32 && cast_scale_ue8m0);
+  TORCH_CHECK(y_s.dtype() == torch::kFloat32 || is_packed_ue8m0);
 
   using Idx_t = int64_t;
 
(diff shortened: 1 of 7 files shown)


csrc/quantization/activation_kernels.cu
```diff
@@ -615,38 +648,86 @@ void persistent_masked_m_silu_mul_quant(
...
-    if (H >= 4096) {
-      static constexpr int NUM_STAGES = 4;
-      static constexpr int THREAD_COUNT = 256;
-      KERNEL(SILU_V2_BLOCK_COUNT, true, THREAD_COUNT, NUM_STAGES);
-    } else {
-      static constexpr int THREAD_COUNT = 32;
-      KERNEL(SILU_V2_BLOCK_COUNT, true, THREAD_COUNT, 2);
+  #define LAUNCH_ON_H(scale_t, STRIDE_YS_E, STRIDE_YS_T, STRIDE_YS_G,         \
+                      STRIDE_YS_P, CEIL_UE8M0)                                \
+    if (H >= 4096 && H % (GROUP_SIZE * 8) == 0) {                             \
```
varun-sundar-rabindranath: @elvircrn  I had to add this as an input of shape (8, 64, 4224) fails on this PR (33 groups). I'll test main and put up a separate for this fix. Thanks.

csrc/quantization/activation_kernels.cu
```diff
@@ -615,38 +648,86 @@ void persistent_masked_m_silu_mul_quant(
...
+  //   X, X, T3G4, T3G5, X, X]
+  //
+  // An Int32 tensor of size [1, 4, 2] and stride [8, 1, 4] can be represented
+  // as an uint8 tensor of shape [1, 2, 4, 4] and stride [32, 16, 4, 1]. In
+  // english, ignoring the Experts dimension, the original int32 tensor is
+  // simply treated as two packed [4, 4] uint8 tensor (or two [4, 1] int32
+  // tensor). The following strides setting reflects this change. Caveat: This
+  // means that the G dimension is no longer contiguous. i.e. Note that to move
+  // from G3 to G4, we need to jump along the packing dimension. The kernel
+  // handles this case.
```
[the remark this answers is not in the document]
varun-sundar-rabindranath: Reviewers take note. If the description is confusing, Ill rewrite it. Thanks.

tests/kernels/moe/test_silu_mul_fp8_quant_deep_gemm.py
```diff
@@ -53,69 +205,83 @@ def test_silu_mul_fp8_quant_deep_gemm(E, T, H, fp8_type):
...
+            torch.testing.assert_close(
+                y_q[e, :nt].to(torch.float32),
+                ref_y_q[e, :nt].to(torch.float32),
+            )
+
+            if scale_fmt == DeepGemmQuantScaleFMT.UE8M0:
+                G = H // group_size
+                y_s_sliced = as_uint8(y_s[e])
+                ref_s_sliced = as_uint8(ref_y_s[e])
+                torch.testing.assert_close(y_s_sliced[:nt, :G], ref_s_sliced[:nt, :G])
```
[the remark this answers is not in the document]
varun-sundar-rabindranath: this is fine as the comparison on the contents when indexed and not raw memory contents.

vllm/model_executor/layers/fused_moe/batched_deep_gemm_moe.py
```diff
@@ -322,7 +353,9 @@ def apply(
         )
 
         a2q, a2q_scale = persistent_masked_m_silu_mul_quant(
-            workspace1, expert_num_tokens
+            workspace1,
+            expert_num_tokens,
+            quant_scale_fmt=DeepGemmQuantScaleFMT.from_target_arch(),
```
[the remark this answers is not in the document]
varun-sundar-rabindranath: the scale format can be different for different calls to deepgemm - they all dont have to sync. 
`fp8_m_grouped_gemm_nt_masked call still queries is_deep_gemm_e8m0_used()` should be fixed , but is part of a greater refactor.

csrc/quantization/activation_kernels.cu
varun-sundar-rabindranath: I have a PR separately for this here https://github.com/vllm-project/vllm/pull/28366/

csrc/quantization/activation_kernels.cu
```diff
@@ -595,17 +627,18 @@ void persistent_masked_m_silu_mul_quant(
...
 
   Idx_t stride_counts_e = tokens_per_expert.stride(0);
 
   static constexpr int GROUP_SIZE = 128;
 
   const cudaStream_t stream = at::cuda::getCurrentCUDAStream();
 
-  #define KERNEL(BLOCK_COUNT, USE_UE8M0, THREAD_COUNT, STAGES)                 \
+  // TODO: Get this from cuda_arch ?
+  static constexpr int SILU_V2_BLOCK_COUNT = 132 * 32;
```
elvircrn, line 640: Yes, this is on my todo list. :)

elvircrn: What we want here is to have #sms instead of 132. Benchmarking, I found that we don't actually want to directly use the number of sms - we benefit from launching more blocks than that.

csrc/quantization/activation_kernels.cu
```diff
@@ -527,8 +540,24 @@ __global__ void silu_mul_fp8_quant_deep_gemm_kernel(
...
-        y_s_ptr += stride_ys_g;
+        // Store scales.
+        if constexpr (std::is_same<scale_t, uint8_t>::value) {
+          // Packed UE8MO format. Remove Mantissa.
+          *y_s_ptr = reinterpret_cast<int16_t&>(y_s) >> 7;
+
+          bool const jump_pack = (current_group_id + 1) % 4 == 0;
+          // Minus 3 because we need to get to the first group in the
+          // next pack.
+          y_s_ptr += jump_pack ? (stride_ys_p - 3) : stride_ys_g;
```
elvircrn, line 551: Equivalent to `bool const jump_pack = (current_group_id % 4 == 3);`?

csrc/quantization/activation_kernels.cu
elvircrn: If you have free time on your hands you can modify this, with possible perf improvements on Blackwell otherwise I'll get to it.

csrc/quantization/activation_kernels.cu
```diff
@@ -615,38 +648,86 @@ void persistent_masked_m_silu_mul_quant(
...
+  // means that the G dimension is no longer contiguous. i.e. Note that to move
+  // from G3 to G4, we need to jump along the packing dimension. The kernel
+  // handles this case.
+
+  stride_ys_e *= sizeof(int32_t);
+  stride_ys_p = T * sizeof(int32_t);  // Packing dimension
+  stride_ys_t = sizeof(int32_t);
+  stride_ys_g = 1;
+
+  LAUNCH_ON_H(uint8_t, stride_ys_e, stride_ys_t, stride_ys_g, stride_ys_p,
```
elvircrn, line 731: My preference here would be that, because these values are hardcoded, that they be converted to constexpr in the kernel. This won't break anything, but will simplify the logic and might give us a small perf boost, too.

elvircrn: The strides only make sense when they are passed from torch::Tensor when the kernel is not allowed to make layout assumptions.

csrc/quantization/activation_kernels.cu
```diff
@@ -527,8 +540,24 @@ __global__ void silu_mul_fp8_quant_deep_gemm_kernel(
       y_q_ptr += WARP_SIZE * stride_yq_h;
 
       if (!lane_id) {
-        *y_s_ptr = y_s;
-        y_s_ptr += stride_ys_g;
+        // Store scales.
+        if constexpr (std::is_same<scale_t, uint8_t>::value) {
+          // Packed UE8MO format. Remove Mantissa.
+          *y_s_ptr = reinterpret_cast<int16_t&>(y_s) >> 7;
```
elvircrn: @varun-sundar-rabindranath 

You might want to mask out the sign bit? So `(reinterpret_cast<int16_t&>(y_s) >> 7) & 0xff` ?

csrc/quantization/activation_kernels.cu
elvircrn: If you have the bandwidth for this, you could perhaps comment out the global scale writes out to ballpark their impact on the overall kernel performance. 

Them if they take up a sizable portion of the execution time, you could perhaps buffer them where each lane in a warp would hold multiple 8-bit scales, and the global writes would happen in chunks larger than 8-bit.

csrc/quantization/activation_kernels.cu
```diff
@@ -509,7 +522,7 @@ __global__ void silu_mul_fp8_quant_deep_gemm_kernel(
 
       __nv_bfloat16 y_s = __hmul(warp_max(_y_max2.x), fp8_inv);
 
-      if constexpr (USE_UE8M0) {
+      if constexpr (CEIL_UE8M0) {
         y_s = hexp2(hceil(hlog2(y_s)));
```
elvircrn, line 526: Here, can we somehow exploit the fact that we don't care about the mantissa?

elvircrn: i.e. Is our final 8-bit value just `hceil(hlog2(y_s)`?

varun-sundar-rabindranath: we use `hexp2(hceil(hlog2(y_s)));` right after this if statement to compute the quant values. we need the `exp2` operation anyways.

csrc/quantization/activation_kernels.cu
varun-sundar-rabindranath: >Equivalent to bool const jump_pack = (current_group_id % 4 == 3);? 

Yes. I initially wrote it as `(current_group_id % 4 == 3)` but thought `(current_group_id + 1) % 4 == 0` was better . Do you think `(current_group_id % 4 == 3)` is more straight forward ?

csrc/quantization/activation_kernels.cu
elvircrn: Yea, that was a bad idea, thanks!

csrc/quantization/activation_kernels.cu
varun-sundar-rabindranath: > If you have the bandwidth for this, you could perhaps comment out the global scale writes out to ballpark their impact on the overall kernel performance.
> 
> Them if they take up a sizable portion of the execution time, you could perhaps buffer them where each lane in a warp would hold multiple 8-bit scales, and the global writes would happen in chunks larger than 8-bit.

Yes. That is a good idea. Given, this PR already has some non-trivial changes, I think we should do it as a follow up to this PR. What do you think ?

csrc/quantization/activation_kernels.cu
varun-sundar-rabindranath: Lets do this as a followup along with the "buffer output scale writes" optimization ?

elvircrn: Yes.

csrc/quantization/activation_kernels.cu
varun-sundar-rabindranath: > My preference here would be that, because these values are hardcoded, that they be converted to constexpr in the kernel. This won't break anything, but will simplify the logic and might give us a small perf boost, too.

Are you referring to `stride_ys_t` and `stride_ys_g` ? The kernel takes these `strides` as input arguments. can you clarify on how they could be converted to `constexpr`  in the kernel. I think I am missing something.

csrc/quantization/activation_kernels.cu
elvircrn: For example, since you're doing:
```
  stride_ys_e *= sizeof(int32_t);
  stride_ys_p = T * sizeof(int32_t);  // Packing dimension
  stride_ys_t = sizeof(int32_t);
  stride_ys_g = 1;
```
`when is_same<scale_t, uint8_t>`

you can assume stride_ys_g = 1 in `if constexpr (std::is_same<scale_t, uint8_t>::value) {`.

Does this make sense?

csrc/quantization/activation_kernels.cu
elvircrn: See response in thread above above.

csrc/quantization/activation_kernels.cu
elvircrn: Leaving this up to you.

heheda12345:
CC @mgoin @yewentao256

vllm/model_executor/layers/fused_moe/batched_deep_gemm_moe.py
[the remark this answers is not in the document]
varun-sundar-rabindranath: is_deep_gemm_e8m0_used honored ! Thanks Bot !

csrc/quantization/activation_kernels.cu
varun-sundar-rabindranath: resolved IRL

mgoin (approved):
Okay I think this looks good to me, nice work

elvircrn (approved):

PR #28370, opened 2025-11-10
[V0 deprecation] Remove no longer used `get_metadata_cls`

LucasWilkinson:
`get_metadata_cls` was a V0 specific thing; remove it 

`make_test_metadata` uses it but is in-turn no longer used by anything

### DIFF
--- a/tests/kernels/utils.py
+++ b/tests/kernels/utils.py
@@ -4,24 +4,21 @@
 
 import itertools
 import random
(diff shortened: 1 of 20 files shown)


wangxiyuan (approved):
LGTM. get_metadata_cls is not called by plugins. Safe to remove.

DarkLight1337 (approved):

PR #28374, opened 2025-11-10
[Fix] optimize visual token mask with caching and multi-token support

bo-ke:
## Purpose
Optimize `_set_visual_token_mask` function to improve performance and extend functionality for multimodal models.

**Changes:**
- Add caching mechanism for visual token IDs to reduce repeated list creation overhead
- Extend support to handle multiple visual token types (image/video start/end tokens, not just `im_patch_id`)
- Replace multiple equality comparisons with `torch.isin` for efficient multi-token matching

### DIFF
--- a/vllm/model_executor/models/ernie45_vl.py
+++ b/vllm/model_executor/models/ernie45_vl.py
@@ -1398,12 +1398,35 @@ class Ernie4_5_VLMoeForConditionalGeneration(
         return image_features
 
     def _set_visual_token_mask(self, input_ids: torch.Tensor) -> None:
-        if getattr(self.config, "im_patch_id", None) is not None:
-            self.visual_token_mask = (input_ids == self.config.im_patch_id).reshape(
-                -1, 1
-            )
-        else:
+        """Set mask for visual tokens (image/video patches and delimiters)."""
+        if getattr(self.config, "im_patch_id", None) is None:
             self.visual_token_mask = None
+            return
+
+        # Cache the visual token IDs tensor to avoid recreating it
+        if not hasattr(self, "_visual_token_ids_cache"):
+            visual_token_ids = [
+                token_id for token_id in [
+                    self.config.im_patch_id,
+                    getattr(self.config, "image_start_token_id", None),
+                    getattr(self.config, "image_end_token_id", None),
+                    getattr(self.config, "video_start_token_id", None),
+                    getattr(self.config, "video_end_token_id", None)
+                ] if token_id is not None
+            ]
+            self._visual_token_ids_cache = visual_token_ids
+
+        # Create tensor on the correct device
+        visual_token_ids_tensor = torch.tensor(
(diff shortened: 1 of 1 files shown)


vllm/model_executor/models/ernie45_vl.py
```diff
@@ -1404,7 +1404,7 @@ def _set_visual_token_mask(self, input_ids: torch.Tensor) -> None:
             return
 
         # Cache the visual token IDs tensor to avoid recreating it
-        if not hasattr(self, "_visual_token_ids_cache"):
+        if not hasattr(self, "_visual_token_ids_tensor_cache"):
```
Isotr0py: I think we can directly initialize `self._visual_token_ids_tensor_cache` in `__init__` to simplify this function.

vllm/model_executor/models/ernie45_vl.py
```diff
@@ -1398,12 +1398,36 @@ def _vision_forward(
...
             self.visual_token_mask = None
+            return
+
+        # Cache the visual token IDs tensor to avoid recreating it
+        if not hasattr(self, "_visual_token_ids_tensor_cache"):
+            visual_token_ids = [
+                token_id for token_id in [
+                    self.config.im_patch_id,
+                    getattr(self.config, "image_start_token_id", None),
+                    getattr(self.config, "image_end_token_id", None),
```
Isotr0py: Seems the `image_start_token_id` and `image_end_token_id` don't appear in `config.json`? (https://huggingface.co/baidu/ERNIE-4.5-VL-28B-A3B-PT/blob/main/config.json)

Isotr0py (commented):
Look reasonable to me. Just have some nits.

vllm/model_executor/models/ernie45_vl.py
```diff
@@ -1398,12 +1398,36 @@ def _vision_forward(
...
+                ] if token_id is not None
+            ]
+            self._visual_token_ids_tensor_cache = torch.tensor(
+                visual_token_ids, dtype=torch.long)
+
+        # Create tensor on the correct device
+        visual_token_ids_tensor = self._visual_token_ids_tensor_cache.to(
+            device=input_ids.device,
+            dtype=input_ids.dtype,
+        )
```
ywang96: This seems a static tensor - why don't we simply set it at init time?

bo-ke: great suggestion

vllm/model_executor/models/ernie45_vl.py
CSWYF3634076: The relevant model configurations have been updated
https://huggingface.co/baidu/ERNIE-4.5-VL-28B-A3B-PT/commit/41a0c75292ebc34d0ff143457353505646fd8d92

Isotr0py (approved):
LGTM now!

PR #28377, opened 2025-11-10
[Bugfix][EPLB] Disabled shared expert overlap when EPLB is enabled

SageMoore:
## Purpose
EPLB is currently broken on main after #28164. This PR works around the issue by disabling shared expert overlap when EPLB is enabled.

## Test Plan
lm eval
## Test Result
Server command
```
VLLM_ALL2ALL_BACKEND=deepep_low_latency vllm serve --model="deepseek-ai/DeepSeek-V2-Lite" --max-num-seqs 512 --trust-remote-code --data-parallel-size 2 --enable-expert-parallel --gpu-memory-utilization 0.75 --disable-log-requests --enable-eplb --eplb-config '{"window_size":100,"step_interval":100}' 
```

Before
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |    0|±  |     0|
|     |       |strict-match    |     5|exact_match|↑  |    0|±  |     0|
```

After
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.3633|±  |0.0278|
|     |       |strict-match    |     5|exact_match|↑  |0.3600|±  |0.0278|
```

### DIFF
--- a/vllm/model_executor/layers/fused_moe/shared_fused_moe.py
+++ b/vllm/model_executor/layers/fused_moe/shared_fused_moe.py
@@ -28,13 +28,16 @@ class SharedFusedMoE(FusedMoE):
         super().__init__(**kwargs)
         self._shared_experts = shared_experts
 
-        # Disable shared expert overlap if we are not using
+        # Disable shared expert overlap if we are using eplb or not using
         # flashinfer + DP since there is nothing to be gained in this case.
         # Disabling the overlap optimization also prevents the shared experts
         # from being hidden from torch.compile.
         self.use_overlapped = (
             use_overlapped
-            and not (self.use_flashinfer_cutlass_kernels and self.dp_size > 1)
+            and not (
+                self.enable_eplb
+                or (self.use_flashinfer_cutlass_kernels and self.dp_size > 1)
+            )
             and self._shared_experts is not None
         )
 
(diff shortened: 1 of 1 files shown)


vllm/model_executor/layers/fused_moe/shared_fused_moe.py
```diff
@@ -28,13 +28,17 @@ def __init__(
...
+        # Disable shared expert overlap if we are using eplb, because of
+        # correctness issues, or if using flashinfer with DP, since there
+        # is nothing to be gained in this case. Disabling the overlap
+        # optimization also prevents the shared experts from being hidden
+        # from torch.compile.
         self.use_overlapped = (
             use_overlapped
-            and not (self.use_flashinfer_cutlass_kernels and self.dp_size > 1)
+            and not (
+                self.enable_eplb
```
yewentao256, line 40:
```suggestion
                TODO(wentao): find the root cause and remove this condition
                self.enable_eplb
```

yewentao256 (approved):
Nice catch, thanks for the work!

vllm/model_executor/layers/fused_moe/shared_fused_moe.py
```diff
@@ -28,13 +28,18 @@ def __init__(
...
+        # Disable shared expert overlap if we are using eplb, because of
+        # correctness issues, or if using flashinfer with DP, since there
+        # is nothing to be gained in this case. Disabling the overlap
+        # optimization also prevents the shared experts from being hidden
+        # from torch.compile.
         self.use_overlapped = (
             use_overlapped
-            and not (self.use_flashinfer_cutlass_kernels and self.dp_size > 1)
+            and not (
+                TODO(wentao): find the root cause and remove this condition
```
yewentao256:
```suggestion
                # TODO(wentao): find the root cause and remove this condition
```

PR #27058, opened 2025-10-16
[ROCm] triton fp8 kernel

maleksan85:
Upstreaming: https://github.com/ROCm/vllm/pull/651

```
VLLM_DISABLE_COMPILE_CACHE=1 \
AMDGCN_USE_BUFFER_OPS=1 \
VLLM_ROCM_USE_AITER=1 \
VLLM_ROCM_USE_AITER_MLA=1 \
VLLM_ROCM_USE_TRITON_ROPE=1 \
vllm serve /data/models/deepseek-ai/DeepSeek-R1-0528 \
    --host localhost \
    --port 8000 \
    --swap-space 64 \
    --disable-log-requests \
    --dtype auto \
    --max-model-len 8192 \
    --tensor-parallel-size 8 \
    --max-num-seqs 1024 \
    --trust-remote-code \
    --block-size 1 \
    --gpu-memory-utilization 0.90 \
    --max-num-batched-tokens 131072 \
    --no-enable-prefix-caching \
    --async-scheduling \
    --compilation-config='{"pass_config":{"enable_attn_fusion":true,"enable_noop":true,"enable_fusion":true},"cudagraph_mode":"FULL","custom_ops":["+rms_norm","+silu_and_mul","+quant_fp8"],"splitting_ops":[]}'
```
```
vllm bench serve \
  --host localhost \
  --port 8000 \
  --model /data/models/deepseek-ai/DeepSeek-R1-0528 \
  --dataset-name random \
  --random-input-len 1024 \
  --random-output-len 1024 \
  --max-concurrency 64 \
  --num-prompts 256 \
  --percentile-metrics ttft,tpot,itl,e2el \
  --ignore-eos
```

baseline:
```
============ Serving Benchmark Result ============
Successful requests:                     256
Maximum request concurrency:             64
Benchmark duration (s):                  123.82
Total input tokens:                      261888
Total generated tokens:                  262144
Request throughput (req/s):              2.07
Output token throughput (tok/s):         2117.07
Peak output token throughput (tok/s):    2304.00
Peak concurrent requests:                127.00
Total Token throughput (tok/s):          4232.06
---------------Time to First Token----------------
Mean TTFT (ms):                          1548.76
Median TTFT (ms):                        1217.95
P99 TTFT (ms):                           2152.59
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          28.74
Median TPOT (ms):                        28.99
P99 TPOT (ms):                           29.90
---------------Inter-token Latency----------------
[...]

PR #28378, opened 2025-11-10
[ROCm] Add missing gemm_a8w8_blockscale import

shinoss:
## Purpose

BugFix: https://github.com/vllm-project/vllm/pull/27058 seems to have missed import for `aiter.gemm_a8w8_blockscale` 

BugFix: https://github.com/vllm-project/vllm/pull/24490 changed it so that we always do quant op. Fix to do quant op only if the input_scale is None:

https://github.com/vllm-project/vllm/blob/d0e186c16f0d62af8c128e2dc7c94cde1387ac02/vllm/model_executor/layers/quantization/utils/fp8_utils.py#L319-L350

BugFix: correct invocation of aiter ops

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/layers/quantization/utils/fp8_utils.py
+++ b/vllm/model_executor/layers/quantization/utils/fp8_utils.py
@@ -323,9 +323,9 @@ class W8A8BlockFp8LinearOp:
         )
 
         if use_triton:
-            gemm_w8a8_blockscale_op = rocm_aiter_ops.triton_gemm_a8w8_blockscale
+            gemm_a8w8_blockscale_op = rocm_aiter_ops.triton_gemm_a8w8_blockscale
         else:
-            gemm_w8a8_blockscale_op = rocm_aiter_ops.gemm_w8a8_blockscale
+            gemm_a8w8_blockscale_op = rocm_aiter_ops.gemm_w8a8_blockscale
 
         if input_scale is not None:
             q_input = input_2d
@@ -341,7 +341,7 @@ class W8A8BlockFp8LinearOp:
         else:
             q_input, input_scale = rocm_aiter_ops.per_1x128_fp8_quant(input_2d)
 
-        return gemm_w8a8_blockscale_op(
+        return gemm_a8w8_blockscale_op(
             q_input,
             weight,
             input_scale,
--- a/vllm/model_executor/layers/quantization/utils/fp8_utils.py
+++ b/vllm/model_executor/layers/quantization/utils/fp8_utils.py
@@ -316,46 +316,38 @@ class W8A8BlockFp8LinearOp:
         assert self.act_quant_group_shape == GroupShape(1, 128)
 
         n, k = weight.shape
-        if input_scale is not None:
-            q_input = input_2d
 
-            return rocm_aiter_ops.gemm_w8a8_blockscale(
-                q_input,
-                weight,
-                input_scale,
-                weight_scale,
-                input_2d.dtype,
-            )
-
-        # MI350 case uses triton kernel
-        if (
+        use_triton = (
             not current_platform.is_fp8_fnuz()
             and rocm_aiter_ops.is_triton_gemm_w8a8_tuned(n, k)
-        ):
+        )
+
+        if use_triton:
+            gemm_w8a8_blockscale_op = rocm_aiter_ops.triton_gemm_a8w8_blockscale
+        else:
+            gemm_w8a8_blockscale_op = rocm_aiter_ops.gemm_a8w8_blockscale
+
+        if input_scale is not None:
+            q_input = input_2d
+        # MI350 case uses triton kernel
+        if use_triton:
             q_input, input_scale = per_token_group_quant_fp8(
                 input_2d,
                 self.act_quant_group_shape.col,
                 column_major_scales=False,
                 use_ue8m0=False,
             )
-            return rocm_aiter_ops.triton_gemm_a8w8_blockscale(
-                q_input,
-                weight,
-                input_scale,
-                weight_scale,
-                input_2d.dtype,
-            )
-
         # MI300 uses tuned AITER ASM/C++ kernel
         else:
             q_input, input_scale = rocm_aiter_ops.per_1x128_fp8_quant(input_2d)
-            return rocm_aiter_ops.gemm_w8a8_blockscale(
-                q_input,
-                weight,
-                input_scale,
-                weight_scale,
-                input_2d.dtype,
-            )
+
+        return gemm_w8a8_blockscale_op(
+            q_input,
+            weight,
+            input_scale,
+            weight_scale,
+            input_2d.dtype,
+        )
 
     def _run_triton(
         self,
--- a/vllm/model_executor/layers/quantization/utils/fp8_utils.py
+++ b/vllm/model_executor/layers/quantization/utils/fp8_utils.py
@@ -330,7 +330,7 @@ class W8A8BlockFp8LinearOp:
         if input_scale is not None:
             q_input = input_2d
         # MI350 case uses triton kernel
-        if use_triton:
+        elif use_triton:
             q_input, input_scale = per_token_group_quant_fp8(
                 input_2d,
                 self.act_quant_group_shape.col,
--- a/vllm/model_executor/layers/quantization/utils/fp8_utils.py
+++ b/vllm/model_executor/layers/quantization/utils/fp8_utils.py
@@ -319,6 +319,14 @@ class W8A8BlockFp8LinearOp:
         if input_scale is not None:
             q_input = input_2d
 
+            return rocm_aiter_ops.gemm_w8a8_blockscale(
+                q_input,
+                weight,
+                input_scale,
+                weight_scale,
+                input_2d.dtype,
+            )
+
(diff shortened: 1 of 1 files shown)


wuhuikx:
@HaiShaw  @tjtanaa could you please help review?

houseroad (approved):

vllm/model_executor/layers/quantization/utils/fp8_utils.py
```diff
@@ -93,6 +93,8 @@ def is_aiter_triton_kernel_tuned(n, k):
 
     n, k = weight.shape
     if input_scale is not None:
+        from aiter import gemm_a8w8_blockscale
```
yewentao256: Is this same with `from aiter.ops.triton.gemm_a8w8_blockscale import gemm_a8w8_blockscale`? Or it is different

yewentao256 (commented):
Thanks for the work!

vllm/model_executor/layers/quantization/utils/fp8_utils.py
tjtanaa: @yewentao256 @sarckk This is different from 
`from aiter import gemm_a8w8_blockscale` is different from
`from aiter.ops.triton.gemm_a8w8_blockscale import gemm_a8w8_blockscale`

shinoss: @tjtanaa `from aiter import gemm_a8w8_blockscale` is correct here right?

tjtanaa: @sarckk May I know which platform are you using? Because on mi300x and mi355x it should cover both use cases. 
on MI300X, it is recommended to use `from aiter import gemm_a8w8_blockscale` and mi355 should use `from aiter.ops.triton.gemm_a8w8_blockscale import gemm_a8w8_blockscale`.

tjtanaa: @sarckk A better approach would be to move all the `gemm_a8w8_blockscale ` imports to a separate if else condition. This will handle both MI300x and MI355x

```diff
+    use_triton = not current_platform.is_fp8_fnuz() and is_aiter_triton_kernel_tuned(n, k)
    
+    if use_triton:
+        from aiter.ops.triton.gemm_a8w8_blockscale import gemm_a8w8_blockscale
+    else:
+        from aiter import gemm_a8w8_blockscale
        
    if input_scale is not None:
        q_input = input_2d
    elif not current_platform.is_fp8_fnuz() and is_aiter_triton_kernel_tuned(n, k):
-        from aiter.ops.triton.gemm_a8w8_blockscale import gemm_a8w8_blockscale

        # MI350 case uses triton kernel
        q_input, input_scale = per_token_group_quant_fp8(
            input_2d,
            group_size,
            column_major_scales=False,
            use_ue8m0=False,
        )
    else:
        # MI300 uses tuned AITER ASM/C++ kernel
        import aiter as rocm_aiter
-        from aiter import gemm_a8w8_blockscale, get_hip_quant
+      from aiter import get_hip_quant
        aiter_per1x128_quant = get_hip_quant(rocm_aiter.QuantType.per_1x128)
        q_input, input_scale = aiter_per1x128_quant(
            input_2d.contiguous(), quant_dtype=rocm_aiter.dtypes.fp8
        )
```

tjtanaa (changes requested):
@sarckk A better approach would be to move all the `gemm_a8w8_blockscale ` imports to a separate if else condition. This will handle both MI300x and MI355x

```diff
+    use_triton = not current_platform.is_fp8_fnuz() and is_aiter_triton_kernel_tuned(n, k)
    
+    if use_triton:
+        from aiter.ops.triton.gemm_a8w8_blockscale import gemm_a8w8_blockscale
+    else:
+        from aiter import gemm_a8w8_blockscale
        
    if input_scale is not None:
        q_input = input_2d
    elif not current_platform.is_fp8_fnuz() and is_aiter_triton_kernel_tuned(n, k):
-        from aiter.ops.triton.gemm_a8w8_blockscale import gemm_a8w8_blockscale

        # MI350 case uses triton kernel
        q_input, input_scale = per_token_group_quant_fp8(
            input_2d,
            group_size,
            column_major_scales=False,
            use_ue8m0=False,
        )
    else:
        # MI300 uses tuned AITER ASM/C++ kernel
        import aiter as rocm_aiter
-        from aiter import gemm_a8w8_blockscale, get_hip_quant
+      from aiter import get_hip_quant
        aiter_per1x128_quant = get_hip_quant(rocm_aiter.QuantType.per_1x128)
        q_input, input_scale = aiter_per1x128_quant(
            input_2d.contiguous(), quant_dtype=rocm_aiter.dtypes.fp8
        )
```

vllm/model_executor/layers/quantization/utils/fp8_utils.py
```diff
@@ -316,38 +316,38 @@ def _run_aiter(
...
+        use_triton = (
             not current_platform.is_fp8_fnuz()
             and rocm_aiter_ops.is_triton_gemm_w8a8_tuned(n, k)
-        ):
+        )
+
+        if use_triton:
+            gemm_w8a8_blockscale_op = rocm_aiter_ops.triton_gemm_a8w8_blockscale
+        else:
+            gemm_w8a8_blockscale_op = rocm_aiter_ops.gemm_a8w8_blockscale
```
hongxiayang: any reason the gemm_a8w8_blockscale_bpreshuffle  is not used?

vllm/model_executor/layers/quantization/utils/fp8_utils.py
shinoss: updated

vllm/model_executor/layers/quantization/utils/fp8_utils.py
tjtanaa: @hongxiayang the gemm_a8w8_blockscaled_bpreshuffle is in another PR. Moreover, the AITER commit in the `Dockerfile.rocm_base` does not have that kernel, so it is not yet merged.

vllm/model_executor/layers/quantization/utils/fp8_utils.py
```diff
@@ -316,38 +316,38 @@ def _run_aiter(
...
+
+        if input_scale is not None:
+            q_input = input_2d
+        # MI350 case uses triton kernel
+        if use_triton:
```
tjtanaa: @sarckk This should be `elif` 
Because we are only calling the quant op only if the `input_scale is None`

tjtanaa (changes requested):

vllm/model_executor/layers/quantization/utils/fp8_utils.py
shinoss: Ok updated

vllm/model_executor/layers/quantization/utils/fp8_utils.py
tjtanaa: @sarckk As we are fixing the AMD CI, so I am trying to evaluate your branch locally. It seems there is a naming confusion here, it should be `rocm_aiter_ops.gemm_w8a8_blockscale` , the `rocm_aiter_ops.gemm_a8w8_blockscale` function does not exists. 

The underlying implementation of `rocm_aiter_ops.gemm_w8a8_blockscale` is a a8w8 operation, we can either fix the name in `_aiter_ops.py` in this PR or in another PR.

vllm/model_executor/layers/quantization/utils/fp8_utils.py
```diff
@@ -316,38 +316,38 @@ def _run_aiter(
...
-                weight_scale,
-                input_2d.dtype,
-            )
+
+        return gemm_w8a8_blockscale_op(
+            q_input,
+            weight,
+            input_scale,
+            weight_scale,
+            input_2d.dtype,
```
tjtanaa: And `output_dtype=input_2d.dtype`

vllm/model_executor/layers/quantization/utils/fp8_utils.py
shinoss: good catch, updated

vllm/model_executor/layers/quantization/utils/fp8_utils.py
shinoss: good catch, updated

vllm/model_executor/layers/quantization/utils/fp8_utils.py
```diff
@@ -316,38 +316,38 @@ def _run_aiter(
...
-                input_scale,
-                weight_scale,
-                input_2d.dtype,
-            )
+
+        return gemm_a8w8_blockscale_op(
+            q_input,
+            weight,
+            input_scale,
+            weight_scale,
```
tjtanaa, line 348: @sarckk Thank you for your patience, this is the last missing positional argument here, and we are good to go.

```diff
        return gemm_w8a8_blockscale_op(
            q_input,
            weight,
            input_scale,
            weight_scale,
+            list(self.weight_group_shape),
            output_dtype=input_2d.dtype
        )
```

tjtanaa (approved):
LGTM. Thank you for your work @sarckk

Issue #27729, opened 2025-10-29
[Bug]: The NVIDIA GPU Tesla V100 throws an error when running vLLM 0.11.0


usberkeley:
#### Environment

- NVIDIA GPU Tesla V100
- vLLM v0.11.0

#### Deployment Method

docker vllm/vllm-openai:v0.11.0, or pip install vllm

#### Start-up Method
vllm serve /workspace/model/Qwen2.5-7B-Instruct --served-model-name Qwen2.5-7B-Instruct

<details>
<summary>Error output</summary>

```text
(EngineCore_DP0 pid=54) INFO 10-28 23:00:42 [backends.py:559] Dynamo bytecode transform time: 5.82 s
(EngineCore_DP0 pid=54) INFO 10-28 23:00:45 [backends.py:197] Cache the graph for dynamic shape for later use
(EngineCore_DP0 pid=54) INFO 10-28 23:01:01 [backends.py:218] Compiling a graph for dynamic shape takes 18.63 s
(EngineCore_DP0 pid=54) INFO 10-28 23:01:07 [monitor.py:34] torch.compile takes 24.46 s in total
(EngineCore_DP0 pid=54) ERROR 10-28 23:01:07 [core.py:708] EngineCore failed to start.
(EngineCore_DP0 pid=54) ERROR 10-28 23:01:07 [core.py:708] Traceback (most recent call last):
(EngineCore_DP0 pid=54) ERROR 10-28 23:01:07 [core.py:708]   File "/usr/local/lib/python3.12/dist-packages/vllm/v1/engine/core.py", line 699, in run_engine_core
(EngineCore_DP0 pid=54) ERROR 10-28 23:01:07 [core.py:708]     engine_core = EngineCoreProc(*args, **kwargs)
(EngineCore_DP0 pid=54) ERROR 10-28 23:01:07 [core.py:708]                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=54) ERROR 10-28 23:01:07 [core.py:708]   File "/usr/local/lib/python3.12/dist-packages/vllm/v1/engine/core.py", line 498, in __init__
(EngineCore_DP0 pid=54) ERROR 10-28 23:01:07 [core.py:708]     super().__init__(vllm_config, executor_class, log_stats,
[...]

PR #28379, opened 2025-11-10
Add use_flashinfer_sampler function with device capability check

usberkeley:
## Purpose
Adds a `use_flashinfer_sampler()` function to check whether FlashInfer sampler should be used. The function checks device compute capability and automatically disables FlashInfer sampler for devices with capability below sm75, even if explicitly set via environment variable.

Fix Issue: https://github.com/vllm-project/vllm/issues/27729#issuecomment-3476451775

Appendix - Flashinfer Version Check
```
def check_cuda_arch():
    # Collect all detected CUDA architectures
    eligible = False
    for major, minor in current_compilation_context.TARGET_CUDA_ARCHS:
        if major >= 8:
            eligible = True
        elif major == 7 and minor.isdigit():
            if int(minor) >= 5:
                eligible = True

    # Raise error only if all detected architectures are lower than sm75
    if not eligible:
        raise RuntimeError("FlashInfer requires GPUs with sm75 or higher")
```

## Test Plan
- Added 5 test cases covering the following scenarios:
  - Device capability below sm75
  - Device capability equal to sm75 with environment variable set to 1
  - Device capability above sm75 with environment variable set to 0
  - Device capability is None
  - Environment variable not set

## Test Result
Pass

---
<details>

</details>


tests/test_env_functions.py
hmellor, line 1: How is this different to `tests/test_env_utils.py`?

What's the difference between helper functions and utility functions in this context?

usberkeley: > How is this different to `tests/test_env_utils.py`?
> 
> What's the difference between helper functions and utility functions in this context?

Hi @hmellor 
`tests/test_env_functions.py`：tests specific environment variable functions that have custom logic
`tests/test_env_utils.py`: tests generic utility functions for environment variable handling

Previously, when testing specific environment-variable-related functions (for example, testing the port), we would create a separate module such as `test_vllm_port.py`. I’d like to consolidate these tests into a single module, `test_env_functions.py`, so we won’t need to create a new module for each environment variable anymore — including the newly added `TestUseFlashinferSampler`.

If the module name I chose isn’t ideal, I’d appreciate any suggestions for improvement.

hmellor: Thanks for the explanation, I was just curious. Could you this explanation to the docstrings for the test modules so it's clearer for future readers?

usberkeley: > Thanks for the explanation, I was just curious. Could you this explanation to the docstrings for the test modules so it's clearer for future readers?

Thanks for the suggestion! The module docstrings already include explanations about the difference between the two modules.

test_env_utils.py
```
This module tests generic utility functions for environment variable handling.
For tests of specific environment variable functions with custom logic,
see test_env_functions.py.
```

test_env_functions.py
```
This module tests specific environment variable functions that have custom logic,
as opposed to test_env_utils.py which tests generic utility functions.
```

PR #28398, opened 2025-11-10
Granite4 Quantization Bug Fix

krishnateja95:
# Pull Request Description

## Purpose

This PR adds quantization configuration remapping support for the GraniteMoeHybrid model to handle the naming difference between Hugging Face model checkpoints and vLLM's internal model structure.

**Problem:** The GraniteMoeHybrid model uses `mamba` naming in HF checkpoints but `mixer` naming in vLLM's implementation (inherited from MambaMixer2). This mismatch causes quantization configurations to fail when targeting specific layers, as the ignore lists and target module names don't match the actual vLLM model structure.

**Solution:** Implements a `maybe_update_quant_config()` method that remaps quantization config module names from HF format (`mamba`) to vLLM format (`mixer`) using regex patterns. This ensures quantization configurations work correctly with the model's actual layer names.

**Changes made:**
- Added `maybe_update_quant_config()` method to `GraniteMoeHybridForCausalLM` class
- Added missing `import re` statement at the top of the file
- Method updates both `ignore` lists and `config_groups.targets` lists in the quantization config
- Remapping rules handle patterns like `model.layers.{N}.mamba.in_proj` → `model.layers.{N}.mixer.in_proj`

### 1. **Integration Test with FP8 Quantization**
```bash
# Test with FP8 quantization targeting specific layers
python -m vllm.entrypoints.openai.api_server \
    --model RedHatAI/granite-4.0-h-small-FP8-block \
    --max-model-len 2048
```

## Test Result

**Before the fix:**
- Quantization configs with `model.layers.{N}.mamba.*` patterns would not match any layers
- Layers would be quantized/ignored incorrectly due to name mismatch
- Potential errors or warnings during quantization initialization

**After the fix:**
- Quantization config module names are automatically remapped from `mamba` to `mixer`
- Both `ignore` lists and `config_groups.targets` are updated correctly
- Quantization applies to the correct layers as intended
- Model loads and runs successfully with quantization enabled

---

## Additional Notes

- This fix is specific to GraniteMoeHybrid model and follows similar patterns used in other vLLM model implementations
- The remapping is applied during model initialization before quantization is configured
- No changes to model behavior or performance, only fixes configuration name matching
- Follows vLLM's existing patterns for handling HF/vLLM naming discrepancies

---

<details>

</details>

### DIFF
--- a/vllm/model_executor/models/granitemoehybrid.py
+++ b/vllm/model_executor/models/granitemoehybrid.py
@@ -569,52 +569,6 @@ class GraniteMoeHybridModel(nn.Module):
                         shard_id="w2",
                         expert_id=e,
                     )
-            
-            elif ('.block_sparse_moe.output_linear.experts.' in n and 
-                (n.endswith('.weight') or n.endswith('.weight_scale'))):
-                
-                # Extract expert ID from the parameter name
-                expert_idx = int(n.split('.experts.')[1].split('.')[0]) if '.experts.' in n else None
-                
-                # Generate the target w2 name
-                w2_name = n.replace(
-                    f'.block_sparse_moe.output_linear.experts.{expert_idx}.weight',
-                    f".block_sparse_moe.experts.{expert_idx}.w2.weight")
-                
-                w2_param = p
-                _load_expert(n.replace(f'.output_linear.experts.{expert_idx}.', '.experts.w2_'),
-                            w2_param,
-                            w2_name,
-                            shard_id='w2',
-                            expert_id=expert_idx)
-
-            elif ('.block_sparse_moe.input_linear.experts.' in n and 
-                (n.endswith('.weight') or n.endswith('.weight_scale'))):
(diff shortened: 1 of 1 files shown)


Isotr0py:
Seems the root issue is just because `MambaMixer2`'s prefix is written wrong by mistake:
https://github.com/vllm-project/vllm/blob/91864b79b36e5a7799f90c1c350e663d484bcfee/vllm/model_executor/models/granitemoehybrid.py#L68-L84 

I think we just need to correct it instead of adding a hacky patch. 😅

krishnateja95:
Hi @mgoin @Isotr0py @heheda12345, I updated the PR. I removed the may_be_quant() function and updated the load_weights function to make the base model and the quantized model are loaded. The llm-compressor PR (for quantized granite4 model) is here: https://github.com/vllm-project/llm-compressor/pull/2001

Issue #28220, opened 2025-11-06
[Bug]: Find the root cause of `apply_repetition_penalties` fails from #28180


vadiklyutiy:
In #28180 there was implemented workaround for `apply_repetition_penalties` fail.

Need to investigate the root cause. 

All preliminary details in #28180

PR #28406, opened 2025-11-10
[MoE][Kernel][Perf] Improve Shared Expert Stream Overlap

alexm-redhat:
SUMMARY:
* use single global auxilary stream to avoid pytorch proifler stream explosion
* limit parallel stream execution to <=256 batch size
* remove shared expert overlap from chunked_impl (dead code)

### DIFF
--- a/vllm/envs.py
+++ b/vllm/envs.py
@@ -221,6 +221,7 @@ if TYPE_CHECKING:
     VLLM_USE_FBGEMM: bool = False
     VLLM_GC_DEBUG: str = ""
     VLLM_DISABLE_SHARED_EXPERTS_STREAM: bool = False
+    VLLM_SHARED_EXPERTS_STREAM_TOKEN_THRESHOLD: int = 256
     VLLM_COMPILE_CACHE_SAVE_FORMAT: Literal["binary", "unpacked"] = "binary"
     VLLM_FLAT_LOGPROBS: bool = False
 
@@ -1468,6 +1469,13 @@ environment_variables: dict[str, Callable[[], Any]] = {
     # Disables parallel execution of shared_experts via separate cuda stream
     "VLLM_DISABLE_SHARED_EXPERTS_STREAM": lambda: bool(
         int(os.getenv("VLLM_DISABLE_SHARED_EXPERTS_STREAM", "0"))
+    ),
+    # Limits when we run shared_experts in a separate stream.
+    # We found out that for large batch sizes, the separate stream
+    # execution is not beneficial (most likely because of the input clone)
+    # TODO(alexm-redhat): Tune to be more dynamic based on GPU type
+    "VLLM_SHARED_EXPERTS_STREAM_TOKEN_THRESHOLD": lambda: int(
+        int(os.getenv("VLLM_SHARED_EXPERTS_STREAM_TOKEN_THRESHOLD", 256))
     ),
     # Format for saving torch.compile cache artifacts
     # - "binary": saves as binary file
--- a/vllm/model_executor/layers/fused_moe/layer.py
+++ b/vllm/model_executor/layers/fused_moe/layer.py
@@ -48,7 +48,11 @@ from vllm.model_executor.layers.quantization.utils.flashinfer_utils import (
 )
 from vllm.platforms import current_platform
 from vllm.utils.math_utils import cdiv, round_up
-from vllm.utils.torch_utils import current_stream, direct_register_custom_op
+from vllm.utils.torch_utils import (
+    aux_stream,
+    current_stream,
+    direct_register_custom_op,
+)
 from vllm.v1.worker.ubatching import dbo_current_ubatch_id
 
 if current_platform.is_cuda_alike():
@@ -331,7 +335,11 @@ class FusedMoE(CustomOp):
             logger.info_once("Disabling MoE shared_experts cuda stream")
             self.shared_experts_stream = None
         else:
-            self.shared_experts_stream = torch.cuda.Stream()
+            # TODO(rob): enable shared expert overlap with non-cuda.
+            # aux_stream() returns None on non-cuda platforms.
+            self.shared_experts_stream = aux_stream()
+            if self.shared_experts_stream is not None:
+                logger.info_once("Enabled separate cuda stream for MoE shared_experts")
 
         if params_dtype is None:
             params_dtype = torch.get_default_dtype()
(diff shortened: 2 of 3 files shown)


vllm/utils/torch_utils.py
```diff
@@ -409,6 +409,23 @@ def current_stream() -> torch.cuda.Stream:
...
+
+def aux_stream() -> torch.cuda.Stream | None:
+    """
+    Ensures aux_stream is initialized only once
+    """
+    global _aux_stream
+
+    from vllm.platforms import current_platform
+
+    if _aux_stream is None and not current_platform.is_cpu():
```
mgoin: Does this work on AMD? Maybe we should just check current_platform.is_cuda()

alexm-redhat: Changed to is_cuda check

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -2439,27 +2452,18 @@ def process_chunk(chunk_start, chunk_end, skip_result_store=False):
...
+                has_separate_shared_experts
+                and self.shared_experts_stream is not None
+                and staged_hidden_states.shape[0] <= DUAL_STREAM_TOKEN_THRESHOLD
+            )
 
-            else:
-                shared_output = None
+            if use_shared_experts_stream:
+                assert self.shared_experts_stream is not None
+                # For chunked, we mark the start sync point for the
```
robertgshaw2-redhat: @alexm-redhat let's just disable this if we get here then?

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -2439,28 +2452,6 @@ def process_chunk(chunk_start, chunk_end, skip_result_store=False):
             staged_hidden_states.copy_(hidden_states, non_blocking=True)
             staged_router_logits.copy_(router_logits, non_blocking=True)
 
-            # If there are shared experts but we are not using a modular kernel,
```
robertgshaw2-redhat: note: it was actually impossible to reach this case (`if self.shared_experts_stream is not None`) before. We currently disable `shared_expert_stream` for the `chunked_impl`

alexm-redhat: reviewed changes here, looks good

alexm-redhat (commented):
Made a pass 10/11 8:27 EST LGTM

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -2652,12 +2629,21 @@ def forward_impl(
...
+                if use_shared_experts_stream:
+                    # Run shared experts in parallel on a separate stream
+                    # NOTE: We start the separate stream here and mark the
+                    # sync end point immediately after it is done. This is
+                    # important to avoid excessive stream allocations by the cuda
+                    # graph replay later.
+                    with torch.cuda.stream(self.shared_experts_stream):
+                        # Note that hidden_states clone() is necessary here to avoid
+                        # conflict with the main stream
+                        shared_output = self.shared_experts(hidden_states.clone())
```
LucasWilkinson: side note: do we know what the conflict is with the main stream? is there an in-place or something? or is the clone just acting as synchronization? (if its the latter can we just use a cuda event?)

robertgshaw2-redhat: I think that what we are doing here is unsafe. 

fixed by: https://github.com/vllm-project/vllm/pull/28430/files

Will land #28430 first

alexm-redhat: We may need to do the clone outside in the main stream, but I'm not sure. The clone itself is not synchronizing, it is only the stream_a.wait(stream_b) calls.

zhewenl: @alexm-redhat I think it's causing some regressions: see https://github.com/vllm-project/vllm/pull/28324 / https://github.com/vllm-project/vllm/issues/28220

alexm-redhat: @zhewenl there is a pr from @mgoin to disable only for marlin. In general, I have checked the clone issue and it is not the reason marlin test fails.

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1159,6 +1163,13 @@ def maybe_roundup_hidden_size(
     return hidden_size
 
 
+# Limits when we run shared_experts in a separate stream.
+# We found out that for large batch sizes, the separate stream
+# execution is not beneficial (most likely because of the input clone)
+# TODO(alexm-redhat): Tune to be more dynamic based on GPU type
+DUAL_STREAM_TOKEN_THRESHOLD = 256
```
pavanimajety: Would it be worth making this a env var so that we can tune on the fly?

alexm-redhat: this is a good idea, will add env var in a sec

mgoin (approved):

vllm/envs.py
```diff
@@ -221,6 +221,7 @@
     VLLM_USE_FBGEMM: bool = False
     VLLM_GC_DEBUG: str = ""
     VLLM_DISABLE_SHARED_EXPERTS_STREAM: bool = False
+    VLLM_DUAL_STREAM_TOKEN_THRESHOLD: int = 256
```
mgoin: I think this name should be more specific i.e. VLLM_SHARED_EXPERTS_STREAM_TOKEN_THRESHOLD
"DUAL_STREAM" could mean a lot of things

mgoin (approved):

vllm/envs.py
alexm-redhat: Make sense, changed to your suggestion

vllm/utils/torch_utils.py
```diff
@@ -409,6 +409,30 @@ def current_stream() -> torch.cuda.Stream:
...
+    """
+    Ensures aux_stream is initialized only once
+    """
+    global _aux_stream
+
+    from vllm.platforms import current_platform
+
+    # TODO: validate this works properly on ROCm platform.
+    if _aux_stream is None and current_platform.is_cuda():
+        _aux_stream = torch.cuda.Stream()
```
youkaichao, line 431: you might also want to make this thread-local? just like `current_stream` above.

PR #28410, opened 2025-11-10
[Bugfix] Disable shared expert overlap if Marlin MoE is used

mgoin:
## Purpose

For now it seems that Marlin MoE might not be safe with multiple CUDA streams, which is triggered when shared expert overlap is used. This was disabled within CI in https://github.com/vllm-project/vllm/pull/28324 so this PR disables shared expert overlap whenever Marlin is used to avoid user issues as well as fix CI.

## Test Result

---
<details>

</details>


robertgshaw2-redhat (approved):

vadiklyutiy (approved):
I second the motion

vadiklyutiy:
it is good if it fails with illegal memory as in this case: at least we know that there is a bug. But frequently race condition might just corrupt hidden state and we got random incorrect output...

vadiklyutiy:
> > But actually my worry is how do we know that other MoE backends is multi streams safe?
> 
> I agree with this - looks like we are disable this as it's failing in CI, but we need to do a more comprehensive testing for other kernels as well? (especially on older HWs with lesser memory)
> 
> cc @mgoin / @vadiklyutiy

My experience says that just comprehensive testing isn't enough (bugs are rare and random). The multi streams part of code should be design and reviewed to be stream safe.

PR #28419, opened 2025-11-10
[BugFix] Shared Expert Stream Sync Fix

vadiklyutiy:
## Purpose
During investigation of #28220 I found there are some inaccuracy in streams sync for executing shared experts in separate stream. 

Fix it.


robertgshaw2-redhat:
I cannot push to this branch. I know NVIDIA is off tomorrow for holiday, so I made a PR here since this is critical fix for release
- https://github.com/vllm-project/vllm/pull/28430/files

PR #28421, opened 2025-11-10
map torchao quantized checkpoints to vLLM's MoE kernels

vkuzo:
Summary:

## Purpose

This is a POC to map `torchao`'s quantized checkpoints for Mixture-of-Experts modules to the `compressed-tensors` MoE path for w8a8 rowwise quant scheme.  We do this by:
1. subclassing `CompressedTensorsW8A8Fp8MoEMethod`, and overriding its `create_weights` and `process_weights_after_loading` methods to map from torchao quantized tensor to compressed-tensors plain tensor format.  This is the part where the "torchao -> compressed_tensor" conversion happens.  Note that for float8 w8a8 rowwise, this is a metadata-only change.
2. adding logic to `TorchAOConfig` to set the quant method to `TorchAOWrappingCompressedTensorsW8A8Fp8MoEMethod` when appropriate.

Note that there are no changes to the `compressed-tensors` path due to the existing vLLM APIs already being expressive enough to do this mapping.

For now, I only implemented one w8a8 scheme to demonstrate a proof of concept.  In the future, the following could be done:
a. map to more schemes for w8a8
b. add mappings for w4a8, etc

## Test Plan

1. integration test for correctness on 1xH100 with `Qwen/Qwen1.5-MoE-A2.7B` with experts quantized to float8 with rowwise scaling with torchao
2. tested locally on 4xH100 with `meta-llama/Llama-4-Scout-17B-16E-Instruct` with experts quantized to float8 with rowwise scaling with torchao
3. very simple performance testing, the values are essentially the same from torchao checkpoint and the corresponding compressed-tensors checkpoint.  Note that the model loading time is higher for torchao (since it is not using safetensors), but this is not captured in the serving benchmark.

```bash
> CUDA_VISIBLE_DEVICES=4,5,6,7 vllm bench throughput --model ../pytorch_scripts/hf_torchao_vllm/data/torchao/fp8-experts-only-mnk-testing-Llama-4-Scout-17B-16E-Instruct --dataset-name sonnet --dataset-path benchmarks/sonnet.txt --num-prompts 10 --tensor-parallel-size 4 --max-model-len 2048 --gpu-memory-utilization 0.8
...
Throughput: 4.07 requests/s, 2623.30 total tokens/s, 609.79 output tokens/s
Total num prompt tokens:  4953
Total num output tokens:  1500
```

## Test Result

see above, the new funtionality did not exist before this PR

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/layers/quantization/torchao.py
+++ b/vllm/model_executor/layers/quantization/torchao.py
@@ -13,6 +13,9 @@ from packaging import version
 from torch.nn.parameter import Parameter
 
 from vllm.logger import init_logger
+from vllm.model_executor.layers.fused_moe.layer import (
+    FusedMoE,
+)
 from vllm.model_executor.layers.linear import (
     LinearBase,
     LinearMethodBase,
@@ -22,6 +25,13 @@ from vllm.model_executor.layers.quantization import QuantizationMethods
 from vllm.model_executor.layers.quantization.base_config import (
     QuantizationConfig,
     QuantizeMethodBase,
+)
+from vllm.model_executor.layers.quantization.compressed_tensors.compressed_tensors_moe import (  # noqa: E501
+    CompressedTensorsW8A8Fp8MoEMethod,
+)
+from vllm.model_executor.layers.quantization.torchao_utils import (
+    maybe_get_torchao_config_for_moe_layer,
+    torchao_config_to_compressed_tensors_config,
 )
 from vllm.model_executor.utils import set_weight_attrs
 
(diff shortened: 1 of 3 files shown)


vllm/model_executor/layers/quantization/torchao.py
```diff
@@ -378,3 +525,164 @@ def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
 
         _restore_weight_attrs(weight, recorded_weight_attr)
         layer.register_parameter("weight", weight)
+
+
+class TorchAOMoEMethod(FusedMoEMethodBase):
```
vkuzo: this is dead code for now, but we might want to keep it as a fallback for weight-only quant

vllm/model_executor/layers/quantization/torchao.py
```diff
@@ -99,6 +134,97 @@ def should_skip(prefix: str, skip_modules: list[str]) -> bool:
     convert_to_packed_tensor_based_on_current_hardware = lambda t: t
 
 
+class TorchAOWrappingCompressedTensorsW8A8Fp8MoEMethod(
```
vkuzo, line 137: need compressed-tensors team blessing on this

vllm/model_executor/layers/quantization/torchao.py
```diff
@@ -99,6 +134,97 @@ def should_skip(prefix: str, skip_modules: list[str]) -> bool:
...
+
+    def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
+        # convert from TorchAOBaseTensor (torchao checkpoint format)
+        # to plain tensors (in compressed-tensors format).
+
+        # So far only tested float8 rowwise scaling
+        # TODO(before land): test other formats, and make it explicit when
+        # something is not supported with a nice error message.
+
+        layer.w13_weight_scale = torch.nn.Parameter(layer.w13_weight.scale)
```
vkuzo: before land: need to polish this to set requires_grad and other attrs properly

vllm/model_executor/layers/quantization/torchao_utils.py
```diff
@@ -0,0 +1,170 @@
...
+
+    from torchao.quantization import (
+        Float8DynamicActivationFloat8WeightConfig,
+        PerRow,
+    )
+
+    is_float8_rowwise = isinstance(
+        torchao_config, Float8DynamicActivationFloat8WeightConfig
+    ) and torchao_config.granularity == [PerRow(), PerRow()]
+    # Special case of float8 rowwise where the HuggingFace weight is stored
```
vkuzo, line 121: note: it's faster to preprocess the checkpoint to convert expert weights to MNK, but good to have this path for supporting naive HF format

kylesayrs (commented):
I'll call out that compressed tensors has some nice [matching utilities](https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/utils/match.py#L160), especially ones that might be good for matching these sets of fused weights.

There are plans to use these utils in compressed_tensors.py as well, not sure if they might be helpful here.

kylesayrs (approved):
Looks good to me! I think doing this conversion on a per-scheme basis is probably the correct approach.

You could potentially generalize this to some sort of adapter mixin for Linear and Moe layers (that way you can share conversion logic for modules which have the same parameters), but that can be thought through more later.

PR #28422, opened 2025-11-10
[Feature] Add env var `VLLM_MOE_USE_DEEP_GEMM`

yewentao256:
## Purpose

image

Top row is VLLM_USE_DEEP_GEMM=0, bottom is with DeepGEMM

Through benchmark, we found that deepgemm is much faster for the linear layers, but slower for the grouped gemm. So We need a way to enable DeepGEMM for the linear layers, but not for the Experts.

## Test

```bash

VLLM_MOE_USE_DEEP_GEMM=1 => DEEPGEMM
INFO 11-10 13:42:09 [fp8.py:168] Using DeepGEMM backend for FP8 MoE

VLLM_MOE_USE_DEEP_GEMM=0 => CUTLASS_BLOCK_SCALED_GROUPED_GEMM
INFO 11-10 13:42:09 [fp8.py:177] Using Cutlass BlockScaled GroupedGemm backend for FP8 MoE
```

### DIFF
--- a/vllm/envs.py
+++ b/vllm/envs.py
@@ -223,6 +223,7 @@ if TYPE_CHECKING:
     VLLM_DISABLE_SHARED_EXPERTS_STREAM: bool = False
     VLLM_COMPILE_CACHE_SAVE_FORMAT: Literal["binary", "unpacked"] = "binary"
     VLLM_FLATTEN_LOGPROBS: bool = False
+    VLLM_MOE_USE_DEEP_GEMM: bool = True
 
 
 def get_default_cache_root():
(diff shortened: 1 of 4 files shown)


vllm/envs.py
```diff
@@ -223,6 +223,7 @@
     VLLM_DISABLE_SHARED_EXPERTS_STREAM: bool = False
     VLLM_COMPILE_CACHE_SAVE_FORMAT: Literal["binary", "unpacked"] = "binary"
     VLLM_FLATTEN_LOGPROBS: bool = False
+    VLLM_MOE_USE_DEEP_GEMM: bool = True
```
mgoin: nit: move this next to VLLM_USE_DEEP_GEMM

mgoin (approved):

vllm/envs.py
yewentao256: Fixed, thanks!

PR #28430, opened 2025-11-10
[Bugfix] Fix Stream Sync for Shared Expert Overlap

robertgshaw2-redhat:
## Purpose
- fix from https://github.com/vllm-project/vllm/pull/28419, with main merged and ci re-enabled
```
- During investigation of https://github.com/vllm-project/vllm/issues/28220 
```

SUMMARY:
* remove shared expert overlap from chunked_impl
* clone the hidden states before stream to avoid race condition
* use record_stream to avoid gc issues

## Test Plan
- LMEval Small models in CI

## Test Result
- Passing?

---
<details>

</details>


vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -2602,11 +2576,22 @@ def forward_impl(
...
+
                 # Run shared experts in parallel on a separate stream
                 with torch.cuda.stream(self.shared_experts_stream):
-                    # Note that hidden_states clone() is necessary here to avoid
-                    # conflict with the main stream
-                    shared_output = self.shared_experts(hidden_states.clone())
+                    shared_output = self.shared_experts(hidden_states_clone)
+
+                # Record that shared_output will be used by current_stream
+                shared_output.record_stream(current_stream())
```
robertgshaw2-redhat: cc @LucasWilkinson  I dont think this second record_stream is needed,

since we call `current_stream().wait_stream(self.shared_experts_stream)` and then use `shared_output` after this

LucasWilkinson: ya i dont think its needed; the `current_stream().wait_stream(self.shared_experts_stream)` should be sufficient to stop premature deallocation

zhewenl: I think second `record_stream` should be redundant

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -2439,28 +2439,6 @@ def process_chunk(chunk_start, chunk_end, skip_result_store=False):
             staged_hidden_states.copy_(hidden_states, non_blocking=True)
             staged_router_logits.copy_(router_logits, non_blocking=True)
 
-            # If there are shared experts but we are not using a modular kernel,
```
robertgshaw2-redhat, line 2459: - Rather than handle this case, let's remove ability to run in shared expert stream with chunked impl (which is just for the DP case)

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -2619,11 +2593,22 @@ def forward_impl(
...
-                    # conflict with the main stream
-                    shared_output = self.shared_experts(hidden_states.clone())
+                    shared_output = self.shared_experts(hidden_states_clone)
+
+                # Record that the clone will be used by shared_experts_stream
+                # to avoid gc issue from deallocation of hidden_states_clone
+                # For more details: https://docs.pytorch.org/docs/stable/generated/torch.Tensor.record_stream.html # noqa: E501
+                # NOTE: we dont need shared_output.record_stream(current_stream())
+                # because we synch the streams before using shared_output.
+                hidden_states_clone.record_stream(self.shared_experts_stream)
```
robertgshaw2-redhat, line 2610: per discussion with @LucasWilkinson offline:

you can think of the record as being added to the queue; so in this case you wan't the self.shared_experts on the queue before the record so the deallocation is blocked until you pass the record (i.e. after self.shared_experts)

LucasWilkinson (approved):
LGTM!

tests/evals/gsm8k/configs/Qwen1.5-MoE-W4A16-CT.yaml
vadiklyutiy, line 1: Does the test really fixed? 
Asking because with only my changes the test still failed.

robertgshaw2-redhat: youre right, I was too swift with this

PR #28432, opened 2025-11-11
[Bugfix][CI/Test][Spec Decode] Fix illegal memory access in offline_inference/spec_decode.py (Issue  27619)

### LINKED ISSUE #27619: [Bug][CI Failure]: EAGLE Spec Decode failing with Triton Attention Backend
### Name of failing test

test_eagle_correctness[TRITON_ATTN-qwen3_eagle3]

### Basic information

- [ ] Flaky test
- [x] Can reproduce locally
- [ ] Caused by external libraries (e.g. bug in `transformers`)

## Summary
EAGLE speculative decoding is failing on AMD MI325X (gfx942) GPUs with a `HSA_STATUS_ERROR_MEMORY_APERTURE_VIOLATION` error. The error occurs immediately when processing prompts starts, after successful CUDA graph capturing. 
It can be reproduced with spec decode + EAGLE + Triton Attention Backend(default for AMD), example:
```
python3 offline_inference/spec_decode.py --test --method eagle --num_spec_tokens 3 --dataset-name hf --dataset-path philschmid/mt-bench --num-prompts 80 --temp 0 --top-p 1.0 --top-k -1 --tp 1 --enable-chunked-prefill --max-model-len 2048
```

## Error Details
```
:0:rocdevice.cpp:3675: Callback: Queue 0x7f4db0300000 aborting with error:
HSA_STATUS_ERROR_MEMORY_APERTURE_VIOLATION: The agent attempted to access memory beyond the largest legal address. code: 0x29
```

(using a different backend could work: eg. `ROCM_AITER_FA`, `ROCM_AITER_UNIFIED_ATTN`, but `TritonAttentionBackend` is the **default**  attention backend for AMD: [gist:fb8bbb2cbde391905d86908ca4a46c02](https://gist.github.com/zhewenl/fb8bbb2cbde391905d86908ca4a46c02))

### What Works
1. Model initialization
2. Weight loading (target model + EAGLE draft model)
3. CUDA graph capturing (both PIECEWISE and FULL modes)
4. KV cache allocation 

### What Fails
**Prompt processing** - Fails immediately when `execute_model` is called, before processing even the first token

### Error Call Stack
```python
multiproc_executor.py:694 worker_busy_loop
  → worker_base.py:353 execute_model
    → gpu_worker.py:491 execute_model
      → gpu_model_runner.py:2512 execute_model
        → gpu_model_runner.py:2404 _model_forward
          → self.model() [MEMORY VIOLATION]
```

Affecting tests: V1 Test e2e + engine, Example Test

### 📝 History of failing test

https://buildkite.com/vllm/ci/builds/36286#019a1ead-5029-45e3-b576-d1ac8cd5ac43
https://buildkite.com/vllm/amd-ci/builds/632#019a27b6-9f6a-4b99-842d-55eef24ea7cd

### CC List.

@mxz297 @yeqcharlotte @Alexei-V-Ivanov-AMD @luccafong @njhill @LucasWilkinson

rasmith:
This PR fixes an illegal memory access that occurs when running

`python3 offline_inference/spec_decode.py --test --method eagle --num_spec_tokens 3 --dataset-name hf --dataset-path philschmid/mt-bench --num-prompts 80 --temp 0 --top-p 1.0 --top-k -1 --tp 1 --enable-chunked-prefill --max-model-len 2048
`

The number of tokens, `num_tokens`, is used to launch the grid in `triton_reshape_and_cache_flash` which may be slightly larger than the `slot_mapping `size, so taking the min between the two when launching the grid resolves the issue. 

I ran the above script again with the fix and additionally tested with `test_eagle.py` and also ran` vllm serve` with `yuhuili/EAGLE3-LLaMA3.1-Instruct-8B` and successfully posted a prompt to it with no errors.

Fixes https://github.com/vllm-project/vllm/issues/27619


vllm/v1/attention/backends/rocm_aiter_fa.py
```diff
@@ -29,9 +29,9 @@
 _CP_TOKENS_PER_ITER_ROCM = 32 * 1024
 
 if current_platform.is_rocm():
-    import aiter
```
tjtanaa, line 34: @rasmith can you revert the line change?

vllm/_aiter_ops.py
```diff
@@ -58,9 +58,10 @@ def _rocm_aiter_fused_moe_impl(
     a1_scale: torch.Tensor | None = None,
     a2_scale: torch.Tensor | None = None,
 ) -> torch.Tensor:
-    from aiter import ActivationType, QuantType
```
tjtanaa, line 61: @rasmith likewise, can you revert the line change?

vllm/_aiter_ops.py
```diff
@@ -114,9 +115,10 @@ def _rocm_aiter_asm_moe_tkw1_impl(
     expert_mask: torch.Tensor | None = None,
     activation_method: int = 0,
 ) -> torch.Tensor:
-    from aiter import ActivationType
```
tjtanaa, line 117: @rasmith likewise, can you revert the line change?

tjtanaa:
@SageMoore  @LucasWilkinson PTAL

tdoublep:
cc @bringlein

vllm/attention/ops/triton_reshape_and_cache_flash.py
```diff
@@ -155,7 +155,10 @@ def triton_reshape_and_cache_flash(
 
     # TODO(ngl): maybe replace with static launch grid to avoid overhead if
     #   using cudagraphs
-    grid = lambda meta: (int(num_tokens), triton.cdiv(n, meta["TILE_SIZE"]))
+    grid = lambda meta: (
+        min(int(num_tokens), slot_mapping.shape[0]),
```
bringlein: Thanks @rasmith for debugging this and the bugfix! 

I think to fix the launch grid is the best idea, the kernel can't figure this out from the memory alone.
I was initially skeptical if this works with capturing full cuda graphs, but for cuda graph catpure it doesn't make a difference, as far as I can see: https://github.com/vllm-project/vllm/blob/d381eb967f171ea8824357075b15bf2895619609/vllm/v1/worker/gpu_model_runner.py#L1412

Actually, I think we should just take `slot_mapping.shape[0]`. Or is there a situation where this can't work?

vllm/attention/ops/triton_reshape_and_cache_flash.py
```diff
@@ -155,7 +155,10 @@ def triton_reshape_and_cache_flash(
...
+        triton.cdiv(n, meta["TILE_SIZE"]),
+    )
```
bringlein, line 160:
```suggestion
    grid = lambda meta: (slot_mapping.shape[0], triton.cdiv(n, meta["TILE_SIZE"]))
```

bringlein (changes requested):

SageMoore (commented):
Thanks for the fix! Let's get rid of the unrelated cruft and apply @bringlein suggestion and we will be good to go.

vllm/v1/attention/backends/rocm_aiter_fa.py
rasmith: I'm not sure, I will try, but I think the line change was introduced by `pre-commit run -a`.

rasmith: Done

vllm/_aiter_ops.py
rasmith: Done

vllm/_aiter_ops.py
rasmith: Done

vllm/attention/ops/triton_reshape_and_cache_flash.py
rasmith: Done

vllm/attention/ops/triton_reshape_and_cache_flash.py
rasmith: I basically did this but had to use the formatting that the linters like.

vllm/v1/attention/backends/rocm_aiter_fa.py
```diff
@@ -31,7 +31,7 @@
 
 if current_platform.is_rocm():
     import aiter
-    from aiter.ops.triton.utils.device_info import get_num_sms
+    from aiter.ops.triton.utils.arch_info import get_num_sms
```
tjtanaa: @rasmith Could we also do this bugfix here as well. We could change to use the `current_platform.get_cu_count()` from `platforms/rocm.py` as I noticed AITER has renamed `arch_info` into `device_info` file and there are users using various commits of AITER at the moment.

The `get_num_sms` is the same utility function as `current_platform.get_cu_count()` 

https://github.com/vllm-project/vllm/blob/df4d3a44a83681feea723cc4c4ebe9085d29d58d/vllm/platforms/rocm.py#L452

vs.
AITER on main branch

https://github.com/ROCm/aiter/blob/8dfaf0a09314873debe166c87f8473aadef1c266/aiter/ops/triton/utils/device_info.py#L8

and
AITER on 355_wip branch

https://github.com/ROCm/aiter/blob/9716b1b817c75ad0441ec7db22147c96716137a7/aiter/ops/triton/utils/arch_info.py#L51

rasmith: I guess I could, but I would want some input from someone else, maybe @gshtras @SageMoore

tjtanaa:
@rasmith  Can you help to rebase this and we should be good to go. Thank you for your help to fix the CI.

SageMoore (approved):

tjtanaa (approved):

PR #28446, opened 2025-11-11
[CLI] fix unicode encode error for `vllm chat/complete` command input

Iceber:
## Purpose
When using the `vllm chat` or `vllm complete` command to input **Chinese** or other **multi-byte characters**, pressing the **Backspace key** to delete characters may cause encoding errors in the input.

This issue can be resolved by simply importing the readline module.

## Test Plan
```bash
$ vllm chat
> 你好
你好呀！有什么可以帮助你的吗？ 😊
> 今天过的怎么样，可以<Backspace Key>
UnicodeEncodeError: 'utf-8' codec can't encode characters in position 349-350: surrogates not allowed
```

fixed:
```bash
$ vllm chat
> 你好
你好！有什么可以帮助你的吗？
> 今天过的怎么样呢啊<Backspace Key><Backspace Key>
今天过得怎么样呢？有什么想分享的吗？
```
## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/entrypoints/cli/openai.py
+++ b/vllm/entrypoints/cli/openai.py
@@ -3,6 +3,7 @@
 
 import argparse
 import os
+import readline  # noqa: F401 # fix unicode encode eror for multi-byte characters
 import signal
 import sys
 from typing import TYPE_CHECKING


vllm/entrypoints/cli/openai.py
```diff
@@ -3,6 +3,7 @@
 
 import argparse
 import os
+import readline
```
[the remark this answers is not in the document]
Iceber: Do we need to consider the Windows platform?

Iceber:
Hi @DarkLight1337, PTAL, Thanks!

vllm/entrypoints/cli/openai.py
```diff
@@ -3,6 +3,7 @@
 
 import argparse
 import os
+import readline  # noqa: F401 # fix unicode encode eror for multi-byte characters
```
DarkLight1337, line 6: I don't quite understand this import, how does importing it alone fix the problem?

Iceber: @DarkLight1337 [readline](https://docs.python.org/3/library/readline.html) can be used to enhance terminal capabilities, and importing it has the side effect of properly handling multi-byte characters.

The fix idea was referenced from https://stackoverflow.com/questions/68542068/python-command-line-input-doesnt-delete-japanese-characters-when-pressing-the-e#comment121134140_68542145

and was stably reproduced through the test plan.

Iceber: It does seem somewhat magical that simply importing it can fix this issue.

DarkLight1337: I prefer using an alternate library to read the input, because `readline` is an optional module in stdlib.

PR #28449, opened 2025-11-11
[Feat][Perf] Enable deepep-low-latency with round-robin expert placement.

cboss6:
## Purpose
Enable deepep low latency all2all backend with round-robin expert plamement strategy, which produces significant performance improvement.

## Performance
Test Platform: CUDA 12.8, drivier 550.144.03
Model: DeepSeek-R1-671B
GPU: H20 * 8 * 2 nodes
Vllm config: dp=16, tp=1, enable_expert_parallel=1, all2all_backend=deepep_low_latency, use_deep_gemm=1

The current functionality has been fully implemented with correct accuracy and significant performance improvements.
In the benchmark with parameters times=10, num_prompts=512, dataset=sharegpt, input_len=1024, output_len=512, max_concurrency=8, and req_rate=8,
the results show that compared to the default linear placement, round-robin's throughput increased by 14.57% and TPOT improved by 13.38%.
Clipboard_Screenshot_1762843721

### serving command (head node)
export NCCL_CHECK_DISABLE=1
export NCCL_COLLNET_ENABLE=0
export NCCL_IB_CUDA_SUPPORT=1
export NCCL_IB_DISABLE=0
export NCCL_IB_GID_INDEX=3
export NCCL_IB_HCA=mlx5_bond_1,mlx5_bond_2,mlx5_bond_3,mlx5_bond_4,mlx5_bond_5,mlx5_bond_6,mlx5_bond_7,mlx5_bond_8
export NCCL_IB_QPS_PER_CONNECTION=4
export NCCL_IB_SL=3
export NCCL_IB_TC=160
export NCCL_LL_THRESHOLD=16384
export NCCL_NET_GDR_LEVEL=2
export NCCL_NVLS_ENABLE=0
export NCCL_P2P_DISABLE=0
export NCCL_PXN_DISABLE=0
export NCCL_SOCKET_IFNAME=bond1
export GLOO_SOCKET_IFNAME=bond1
export NVSHMEM_BOOTSTRAP_UID_SOCK_IFNAME=bond1
export NVSHMEM_HCA_LIST=mlx5_bond_1:1,mlx5_bond_2:1,mlx5_bond_3:1,mlx5_bond_4:1,mlx5_bond_5:1,mlx5_bond_6:1,mlx5_bond_7:1,mlx5_bond_8:1
export NVSHMEM_IB_TRAFFIC_CLASS=160
export NVSHMEM_DIR=/usr/local/nvshmem
export LD_LIBRARY_PATH=${NVSHMEM_DIR}/lib:$LD_LIBRARY_PATH
export PATH=${NVSHMEM_DIR}/bin:$PATH
export VLLM_ALL2ALL_BACKEND=deepep_low_latency
export VLLM_DISABLE_NCCL_FOR_DP_SYNCHRONIZATION=1
export VLLM_ATTENTION_BACKEND=FLASHMLA
export VLLM_USE_DEEP_GEMM=1
export DG_JIT_CACHE_DIR=/root/.cache/vllm/deep_gemm/

vllm serve /path/to/DeepSeek-R1 \
    --host 0.0.0.0 \
    -tp 1  \
    --max-model-len 16384 \
    --max-num-batched-tokens 16384 \
    --expert-placement-strategy round_robin \
    --enable-chunked-prefill \
    --gpu-memory-utilization 0.8 \
    --load-format "auto" \
    --enable-expert-parallel \
    --data-parallel-size 16 \
    --data-parallel-size-local 8 \
    --data-parallel-address ${HOST_IP} \
    --data-parallel-rpc-port 12345 \
    --api-server-count=8

### benchmark command
vllm bench serve \    --backend vllm \    --base-url "http://127.0.0.1:8500" \    --port 8500 \    --endpoint '/v1/completions' \    --model ${model} \    --dataset-name sharegpt \  --num-prompts 512 \    --max-concurrency 8 \
    --request-rate 8 \    --random-input-len 1024 \    --random-output-len 512

## Accuracy
|gsm8k|      3|flexible-extract|     5|exact_match|↑  | 0.97|±  |0.0171|
|     |       |strict-match    |     5|exact_match|↑  | 0.97|±  |0.0171|
Clipboard_Screenshot_1762842773


vllm/model_executor/layers/fused_moe/deepep_ll_prepare_finalize.py
```diff
@@ -93,6 +93,9 @@ def __init__(
         # combine function.
         self.handles: list[tuple | None] = [None, None]
         self.num_dispatchers_ = num_dispatchers
+        self.global_to_physical: torch.Tensor | None = None
+        self.physical_to_global: torch.Tensor | None = None
+        self.local_expert_global_ids: torch.Tensor | None = None
```
bnellnm: I think these should just be additional optional parameters to the `__init__` method.  Then we won't need a setter.  

You can pass these through `maybe_make_prepare_finalize` as optional parameters as well.

vllm/model_executor/layers/fused_moe/deepep_ll_prepare_finalize.py
```diff
@@ -110,6 +113,28 @@ def max_num_tokens_per_rank(self) -> int | None:
...
+    ) -> None:
+        self.global_to_physical = global_to_physical
+        self.physical_to_global = physical_to_global
+        self.local_expert_global_ids = local_expert_global_ids
+
+    def _map_global_to_physical_ids(self, topk_ids: torch.Tensor) -> torch.Tensor:
+        if self.global_to_physical is None:
+            return topk_ids
+        physical = self.global_to_physical[topk_ids.to(torch.long)]
+        return physical.to(topk_ids.dtype)
```
bnellnm: Are the casts necessary here?  `topk_ids` should already be an int of some type and if needed, `global_to_physical` could be cast in `set_expert_routing_info` (or `__init__`) using `topk_indices_dtype` so that there's no issues about type mismatches.

vllm/model_executor/layers/fused_moe/deepep_ll_prepare_finalize.py
```diff
@@ -110,6 +113,28 @@ def max_num_tokens_per_rank(self) -> int | None:
...
+
+    def _map_local_to_global_ids(self, expert_topk_ids: torch.Tensor) -> torch.Tensor:
+        if self.local_expert_global_ids is None:
+            return expert_topk_ids
+        global_ids = self.local_expert_global_ids[expert_topk_ids.to(torch.long)]
+        return global_ids.to(expert_topk_ids.dtype)
```
bnellnm: Similar comment about casts.

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -251,12 +251,26 @@ def maybe_init_modular_kernel(
...
+                and hasattr(prepare_finalize, "set_expert_routing_info")
+                and hasattr(layer, "expert_global_to_physical")
+                and hasattr(layer, "expert_physical_to_global")
+                and hasattr(layer, "expert_local_to_global")
+            ):
+                prepare_finalize.set_expert_routing_info(
+                    layer.expert_global_to_physical,
+                    layer.expert_physical_to_global,
+                    layer.expert_local_to_global,
+                )
```
bnellnm: The construction logic was just refactored recently and `maybe_init_modular_kernel` has moved to `FusedMoE`.  So that should make this a  little bit simpler.

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1338,28 +1353,43 @@ def __init__(
                         "models with multiple expert groups and no redundant "
                         "experts. Falling back to linear expert placement."
                     )
-                    expert_placement_strategy = "linear"
+                    self.expert_placement_strategy = "linear"
+                elif (
+                    self.use_deepep_ht_kernels
+                    or self.use_pplx_kernels
+                    or self.use_flashinfer_cutlass_kernels
```
bnellnm: Can you change the condition to be `not self.use_deepep_ll_kernels`?  That way, if any new all2all mechanisms are added, they won't accidentally skip this code.

vllm/model_executor/layers/fused_moe/deepep_ll_prepare_finalize.py
cboss6: Done, the setter is removed.

vllm/model_executor/layers/fused_moe/deepep_ll_prepare_finalize.py
cboss6: Done.

vllm/model_executor/layers/fused_moe/layer.py
cboss6: Done.

vllm/model_executor/layers/fused_moe/layer.py
cboss6: Done.

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -144,6 +144,7 @@ def uses_weight_scale_2_pattern(self) -> bool:
     def _maybe_make_prepare_finalize(
         moe: FusedMoEConfig,
         quant_config: FusedMoEQuantConfig | None,
+        layer: torch.nn.Module | None = None,
```
bnellnm: Could you pass the maps as individual optional tensors?  Or an optional tuple of tensors?  In latest main, this function has been moved to a standalone function that can't depend on the `layer`.

Also, I think adding a `layer` argument here would break one of the unit tests that depend on this function.

cboss6: Thanks for the suggestion! I’ve updated the code to use a tuple of routing tables instead of passing the layer, and I’ve also rebased onto the latest main.

Could you take a look and see if this approach is acceptable to you?

bnellnm (approved):
LGTM!

hmellor (commented):
Nothing seems obviously incorrect to me, let's see what the tests say

hmellor (approved):

Issue #28236, opened 2025-11-06
[Feature]: Implement naive prepare/finalize class to replace naive dispatching in fused_moe/layer.py


bnellnm:
### 🚀 The feature, motivation and pitch

The `FusedMoE` layer has a special case dispatch/combine for EP+DP when there is no specific all2all backend specified.  This makes the code in `layer.py` a bit confusing and hard to follow.  One way to simplify this is to implement a proper `FusedMoEPrepareAndFinalize` subclass for naive dispatch/combine.

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

teddygood:
Hi, I'm interested in working on this issue.
Would it be okay if I take this one?

bnellnm:
> Hi, I'm interested in working on this issue. Would it be okay if I take this one?

Go for it!

WorldExplored:
Made PR for this. Needs review.

PR #28453, opened 2025-11-11
[Feature]: Implement naive prepare/finalize class to replace naive dispatching in fused_moe/layer.py

WorldExplored:
Reference Issue: https://github.com/vllm-project/vllm/issues/28236

Add a FusedMoEPrepareAndFinalize subclass.

cc @bnellnm


vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -231,6 +234,9 @@ def _maybe_make_prepare_finalize(
                 use_fp8_dispatch=use_fp8_dispatch,
             )
 
+        if prepare_finalize is None:
+            prepare_finalize = FusedMoENaivePrepareAndFinalize()
+
         return prepare_finalize
 
     def maybe_make_prepare_finalize(self) -> FusedMoEPrepareAndFinalize | None:
```
haosdent: Should we return the class at here to make it clear?
```
        elif self.moe.dp_size > 1 and self.moe.use_ep:
            # EP+DP without a specialized all2all backend: use naive dispatch/combine
            return FusedMoENaivePrepareAndFinalize
```

bnellnm (changes requested):
The `FusedMoEMethodBase`, `FusedMoEModularMethod` and `UnquantizedFusedMoEMethod` classes have all be moved to separate files.  Can you update the PR so that it is easier to review?

bnellnm:
Thanks for working on this @WorldExplored .  The idea for the task was to implement the `prepare` and `finalize` methods using the naive all2all communication primitives and removing the "naive" code from layer.py.

Issue #24384, opened 2025-09-07
[RFC]: Decoupling vLLM Configuration from Hugging Face


charlotte12l:
### Motivation.

Currently, vLLM assumes that all models follow the Hugging Face format. Configuration is parsed directly into a `transformers.PretrainedConfig` instance, which is then embedded into `ModelConfig` as `hf_config`. This tight coupling introduces several problems:

- **Poor extensibility**: non-HF models (e.g., Mistral-native) cannot be integrated cleanly. Their configuration must first be awkwardly adapted into a `PretrainedConfig`-like object.  
- **Maintenance overhead**: many fields in `PretrainedConfig` are irrelevant to inference, but vLLM does not clearly separate used vs. unused fields. Users who want to support their own model formats must carefully map into HF’s schema, and this kind of manual mapping is fragile and error-prone.

- **Under‑specified, inconsistently named critical fields**[added on Nov 1]:  Fields vLLM relies on at runtime—such as [num_kv_heads](https://github.com/vllm-project/vllm/blob/main/vllm/config/model.py#L1304), [num_experts](https://github.com/vllm-project/vllm/blob/main/vllm/config/model.py#L1382), [max_model_len](https://github.com/vllm-project/vllm/blob/b5d90f740048d43376390a61ca5b77c287505d0e/vllm/config/model.py?brid=J9xsXseEdXMBF-8gge5fvA#L2028-L2056)—are not emphasized or standardized in `PretrainedConfig`. Different models use different names, forcing `ModelConfig` to perform bespoke mappings.

- **Missing architecture hints force runtime introspection**[added on Nov 1]: Useful architecture details (e.g., attention type per layer for KV‑cache initialization) are not guaranteed(maybe it’s now? but what if there are other hints needed in the future?) available in `PretrainedConfig`. Today we infer them at runtime (e.g., via [forward_context](https://github.com/vllm-project/vllm/blob/main/vllm/config/vllm.py#L997)), adding complexity and risk. Getting such fields into HF first and then into every model slows vLLM development.

This proposal aims to resolve these issues by introducing a **clean separation of concerns**:  
- Define a standardized, vLLM‑native configuration schema that clearly defines the fields the engine needs.
- Provide pluggable parsers that translate external configuration formats (HF, Mistral‑native, GGUF, etc.) into that schema.
- Simplify code paths: fewer runtime probes (e.g., no per‑layer attention discovery via `forward_context`) and earlier detection of missing information.

[...]

PR #22299, opened 2025-08-06
Implicit language-model-only mode via limit-mm-per-prompt

ywang96:
# Essential Elements of an Effective PR Description Checklist

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x] The test plan, such as providing test command.
- [x] The test results, such as pasting the results comparison before and after, or e2e results
- [x] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.

## Purpose
Similar to #22224, this PR enables an implicit language-model-only model for certain models that are released as a hybrid-only model when `limit-mm-per-prompt` is set 0 for all modalities. This PR also skips loading certain modality modules when they're set to 0 individually (See Qwen2.5-Omni as an example)

Some followup work:
- We should still add an explicit flag for models that support multimodal embeddings as input (e.g, Qwen2.5VL)
- In this case, the model is still treated as a multimodal model. However, memory profiling will be done by simply saturating the embedding cache with dummy multimodal embeddings (i.e, without activating multimodal encoder and projector since they're not loaded).

Co-authored by @sfeng33 

Part of #21943

## Test Plan
Running Qwen2.5VL-7B on L40S

Command: `vllm serve Qwen/Qwen2.5-VL-7B-Instruct`
```
(EngineCore_0 pid=3267325) INFO 08-06 21:05:33 [cuda.py:327] Using Flash Attention backend on V1 engine.
(EngineCore_0 pid=3267325) INFO 08-06 21:05:33 [weight_utils.py:296] Using model weights format ['*.safetensors']
Loading safetensors checkpoint shards:   0% Completed | 0/5 [00:00<?, ?it/s]
Loading safetensors checkpoint shards:  20% Completed | 1/5 [00:00<00:00,  9.78it/s]
Loading safetensors checkpoint shards:  40% Completed | 2/5 [00:00<00:00,  3.46it/s]
Loading safetensors checkpoint shards:  60% Completed | 3/5 [00:01<00:00,  2.56it/s]
Loading safetensors checkpoint shards:  80% Completed | 4/5 [00:01<00:00,  2.32it/s]
Loading safetensors checkpoint shards: 100% Completed | 5/5 [00:02<00:00,  2.21it/s]
Loading safetensors checkpoint shards: 100% Completed | 5/5 [00:02<00:00,  2.48it/s]
(EngineCore_0 pid=3267325) 
[...]

PR #28454, opened 2025-11-11
[Core] Parse vLLM engine required fields from hf_config to model_arch_config

charlotte12l:
## Purpose
See https://github.com/vllm-project/vllm/issues/24384 for more context.

Use llama3 as prototype 

## Design
- `model_arch_config` explicitly specify all standardized fields required for vllm runtime
- model arch parser read from config.json/params.json/etc. and perform the standardization process. The goal is to eventually remove most of the standardization logic from config/model.py once the migration to the new parser workflow is complete
- For hf-model-arch-parser, if the model is not in `_CONFIG_REGISTRY`, we still call `AutoConfig.from_pretrained`. This allows us to leverage the normalization already implemented in HuggingFace’s `PretrainedConfig`. A more standardized PretrainedConfig will enable a thinner, simpler parser layer.
https://github.com/vllm-project/vllm/blob/f0359fffa434a4fce981389f9dff93a2a4c2b13e/vllm/transformers_utils/config.py#L79-L102

- parser will inspect the model cls and get `per_layer_attn_cls`. This information is useful for the KV cache manager 

## TODOs

- **[Should probably implement&merge it before this PR]** Refactor out the existing normalizations in config/[model.py](http://model.py/)  so it can be shared by model-arch-parser and hf_config during the migration
- Some fields are not normalized yet: max_model_len
- Mistral parser
- Get_kv_cache spec use model_arch_config.per_layer_attn_cls, and tests
- Have a clear separation between model_info and model_arch_config, per https://github.com/vllm-project/vllm/issues/24384#issuecomment-3489703720
- Gradually migrate other models(also note that some models have configuration class re-implemented in vllm) and tests

## Test Plan
```
python -m pytest tests/test_config.py::test_model_arch_config_loading
```

Besides, I run generate for llama3 and the outputs before/after this PR are the same.
```

# Sample prompts.
prompts = [
    "Hello, my name is",
    "The president of the United States is",
    "The capital of France is",
    "The future of AI is",
]
# Create a sampling params object.
sampling_params = SamplingParams(temperature=0)


def main():
    # Create an LLM.
    llm = LLM(model="/data/local/models/oss/Llama-3.1-8B-Instruct")
    # Generate texts from the prompts.
    # The output is a list of RequestOutput objects
    # that contain the prompt, generated text, and other information.
    outputs = llm.generate(prompts, sampling_params)
    # Print the outputs.
    print("\nGenerated Outputs:\n" + "-" * 60)
    for output in outputs:
        prompt = output.prompt
        generated_text = output.outputs[0].text
        print(f"Prompt:    {prompt!r}")
        print(f"Output:    {generated_text!r}")
        print("-" * 60)

if __name__ == "__main__":
    main()
```

the outputs before/after this PR are the same.
```
Prompt:    'Hello, my name is'
Output:    ' Emily and I am a 25-year-old freelance writer and editor. I have'
Prompt:    'The president of the United States is'
Output:    ' the head of state and head of government of the United States. The president serves'
Prompt:    'The capital of France is'
Output:    ' a city of romance, art, fashion, and cuisine. Paris is a must'
Prompt:    'The future of AI is'
Output:    ' bright, but it also raises concerns about bias, accountability, and the impact on'
```

## Test Result

passed
---
<details>

- [] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>


vllm/transformers_utils/model_arch_config_parser.py
```diff
@@ -0,0 +1,441 @@
...
+    ]
+    for attr in attributes:
+        num_kv_heads = config_dict.get(attr)
+        if num_kv_heads is not None:
+            config_dict.pop(attr)
+            return num_kv_heads
+
+    # For non-grouped-query attention models, the number of KV heads is
+    # equal to the number of attention heads.
+    return standard_fields["num_attention_heads"]
```
zhuohan123: No need to be included in the first PR, but for these logic, can we move it to be more "model specific", like we write a conversion script for every model, instead of having this magic logic for all models here.

vllm/config/model_arch.py
```diff
@@ -0,0 +1,145 @@
...
+@dataclass(config=ConfigDict(arbitrary_types_allowed=True, extra="allow"))
+class ModelArchitectureTextConfig:
+    model_type: str
+    hidden_size: int
+    num_hidden_layers: int
+    num_attention_heads: int
+    use_deepseek_mla: bool
+    head_dim: int
+    vocab_size: int
+    num_key_value_heads: int
```
zhuohan123: A model can have mixed different attention layers. Should we make some of the fields here per-layer?

vllm/config/model.py
```diff
@@ -1194,6 +1212,8 @@ def get_hidden_size(self) -> int:
 
     @property
     def is_deepseek_mla(self) -> bool:
+        if self.model_arch_config:
+            return self.model_arch_config.text_config.use_deepseek_mla
```
zhuohan123: Ideally we will have `model_arch_config` for every model and we no longer need the logic below line 1217 anymore?

vllm/config/model_arch.py
```diff
@@ -0,0 +1,145 @@
...
+    def __init__(
+        self,
+        **kwargs,
+    ):
+        for key, value in kwargs.items():
+            setattr(self, key, value)
+
+    def __repr__(self):
+        config_dict_json = json.dumps(self.__dict__, indent=2, sort_keys=True) + "\n"
+        return f"{self.__class__.__name__} {config_dict_json}"
```
zhuohan123: Any examples for how these two classes work?

zhuohan123 (commented):
Thanks for the great work! Left some comments.

zhuohan123:
> ```python
> class AllConfig(PretrainedConfig):
>     hidden_size: int
>     num_hidden_layers: int
>     num_attention_heads: int
>     use_deepseek_mla: bool
>     head_dim: int
>     vocab_size: int
>     num_key_value_heads: int
>     num_experts: int
> ```
> 
> Which would contain the union of all possible config fields using their standard names. Would that be an acceptable solution?

@hmellor The main issue of this inheritance approach is that we want an explicit error if anybody try to access a field that is not in the defined config list. So when people implements something in the engine, they won't accidentally access a field that only exists for some models.

vllm/config/model.py
hmellor: Would this not mean reimplementing the config class for every model in Transformers? That doesn't seem like an ideal situation.

edit: Or do you mean that every arch will have it populated? Meaning that outside of instantiation we won't need any model specific logic? (Sorry I may have misinterpreted you originally)

vllm/config/model_arch.py
```diff
@@ -0,0 +1,145 @@
...
+        model_type: str,
+        hidden_size: int,
+        num_hidden_layers: int,
+        num_attention_heads: int,
+        use_deepseek_mla: bool,
+        head_dim: int,
+        vocab_size: int,
+        num_key_value_heads: int,
+        num_experts: int,
+        **kwargs,
```
heheda12345: Why do we need kwargs? Should ModelArchitectureTextConfig be a precise set of configs that vllm need?

vllm/config/model_arch.py
```diff
@@ -0,0 +1,145 @@
...
+    def __repr__(self):
+        config_dict_json = json.dumps(self.__dict__, indent=2, sort_keys=True) + "\n"
+        return f"{self.__class__.__name__} {config_dict_json}"
+
+
+@dataclass(config=ConfigDict(arbitrary_types_allowed=True, extra="allow"))
+class ModelArchitectureVisionConfig:
+    def __init__(
+        self,
+        **kwargs,
```
heheda12345: same concern for kwargs

vllm/config/model_arch.py
```diff
@@ -0,0 +1,145 @@
...
+@dataclass(config=ConfigDict(arbitrary_types_allowed=True))
+class ModelArchitectureConfig:
+    """
+    Configuration for model architecture
+    """
+
+    text_config: ModelArchitectureTextConfig = field(init=True)
+    """Text model configuration containing text-specific architecture details."""
+
+    architectures: list[str] = field(default_factory=list)
```
heheda12345: @hmellor  why is architectures a `list[str]` instead of a `str` in huggingface config? Do we need it to be a list in vllm?

vllm/config/model_arch.py
```diff
@@ -0,0 +1,145 @@
...
+    """Model type identifier (e.g., 'llama', 'gpt2')."""
+
+    # TODO: Formalize quantization_config in parser
+    quantization_config: dict[str, Any] = field(default_factory=dict)
+    """Quantization configuration dictionary containing quantization parameters."""
+
+    torch_dtype: str = ""
+    """PyTorch data type for model weights (e.g., 'float16', 'bfloat16')."""
+
+    per_layer_attention_cls: list[type[nn.Module]] = field(default_factory=list)
```
heheda12345: why are you putting this in ModelArchitectureConfig rather than ModelArchitectureTextConfig? For a multi-modal model, will it include the attention layers in vision encoder?

vllm/config/model_arch.py
```diff
@@ -0,0 +1,145 @@
...
+logger = init_logger(__name__)
+
+
+@dataclass(config=ConfigDict(arbitrary_types_allowed=True, extra="allow"))
+class ModelArchitectureTextConfig:
+    model_type: str
+    hidden_size: int
+    num_hidden_layers: int
+    num_attention_heads: int
+    use_deepseek_mla: bool
```
heheda12345: is this necessary? MLA are with `MLAAttention` class now and you can infer whether it is a mla from `per_layer_attention_cls`

vllm/model_executor/models/llama.py
```diff
@@ -728,3 +729,12 @@ def permute(w: torch.Tensor, n_heads: int, attn_out: int):
                 name = name.replace(item, mapping[item])
 
         return name, loaded_weight
+
+    @classmethod
+    def get_per_layer_attention_cls(cls, text_config: ModelArchitectureTextConfig):
```
heheda12345: this function looks a bit strange for me as it takes ModelArchitectureTextConfig as input and generates something needed by ModelArchitectureConfig. This interaction is complex.

Where is `is_causal` from?

vllm/model_executor/models/llama.py
```diff
@@ -116,7 +117,7 @@ def forward(self, x):
 class LlamaAttention(nn.Module):
     def __init__(
         self,
-        config: LlamaConfig,
+        config: ModelArchitectureTextConfig,
```
heheda12345: Note that llama.py is also used by a bunch of models like mistral, InternLM. Need to ensure they are working too.

vllm/config/model.py
charlotte12l: @zhuohan123  Yes, after the fully migration, the logic below line 1217 are not needed; but during the migration when some of the models still use hf_config, those logic are still needed.

vllm/transformers_utils/model_arch_config_parser.py
heheda12345: I think we should have a base class for the basic implementation and let each model to override it for their special attributes if necessary.

Ideally, I agree with @hmellor that the above logic should be handled by huggingface transformers in the long term.

vllm/transformers_utils/model_arch_config_parser.py
```diff
@@ -0,0 +1,441 @@
...
+
+
+def extract_num_experts(config_dict: dict[str, Any]) -> int:
```
heheda12345: can you avoid the code duplication of this file and those in vllm/config/model.py?

vllm/config/model.py
charlotte12l: @hmellor Ya, the plan is that every arch will have the required info populated in parser, and outside of instantiation we won't need any model specific logic

>  reimplementing the config class for every model in Transformers

If I understand correctly, all vLLM supported models will implement its own model cls in vLLM, while some of them have their own configurations class in vLLM, some are still on Transformers.

The plan is that for the models which already have dedicated configuration class in vLLM https://github.com/vllm-project/vllm/blob/f0359fffa434a4fce981389f9dff93a2a4c2b13e/vllm/transformers_utils/config.py#L79-L102 , we will re-implement them but inherit from ModelArchitectureConfig; for those do not, we use parsers to populate values into the basic ModelArchitectureConfig, the current prototyped llama3 is an example for it --  llama3 do not have a dedicated configuration class in vLLM

vllm/config/model_arch.py
```diff
@@ -0,0 +1,145 @@
...
+    """Text model configuration containing text-specific architecture details."""
+
+    architectures: list[str] = field(default_factory=list)
+    """List of model architecture class names (e.g., ['LlamaForCausalLM'])."""
+
+    model_type: str = ""
+    """Model type identifier (e.g., 'llama', 'gpt2')."""
+
+    # TODO: Formalize quantization_config in parser
+    quantization_config: dict[str, Any] = field(default_factory=dict)
```
heheda12345: Is it ok to always follow huggingface's quantization_config definition? Does it have similar problem with the  hf_config discussed in the RFC?

vllm/config/model_arch.py
charlotte12l: I thought our plan was to use a more standarized config to replace hf_config? If so, we will need the extra fields for model arch initialization.

I'm also okay with keeping both hf_config and model_arch_config and make model_arch_config very precise. In model_executor, we use hf_config; In other places in vLLM, we use model_arch_config. This would actually make the migration much easier. Let me know what you and @zhuohan123 think about it!

vllm/config/model_arch.py
charlotte12l: This is because get_per_layer_attention_cls() will need ModelArchitectureTextConfig, so I try to make it less confusing(although it still is).

Currently, this doesn’t include the attention layers in the vision encoder. After working through this, I feel even more strongly that per_layer_attention_cls should be moved into model_info, as I mentioned in my comment  https://github.com/vllm-project/vllm/issues/24384#issuecomment-3488864581 ; at that time the decision was to place it in ArchitectureConfig, but since the actual attention_cls is closely tied to vLLM-specific details, moving it to ModelInfo could make the code much cleaner?

vllm/transformers_utils/model_arch_config_parser.py
charlotte12l: Yes, this is in TODO-1.

> [Should probably implement&merge it before this PR] Refactor out the existing normalizations in config/[model.py](http://model.py/) so it can be shared by model-arch-parser and hf_config during the migration

Alternatively, we could remove those complex logics in parser do what @zhuohan123 mentioned here  https://github.com/vllm-project/vllm/pull/28454#discussion_r2515503016 , use different parsers for different models; we can move the discussion there.

vllm/model_executor/models/llama.py
charlotte12l: It's following the llama initialization logics 

https://github.com/vllm-project/vllm/blob/c2b28471cf31a84ce81b9f21aee03eda69cd8a17/vllm/model_executor/models/llama.py#L314-L321

vllm/config/model_arch.py
charlotte12l: Good point, let me remove it

vllm/config/model_arch.py
charlotte12l: Yes, that's also a TODO(in this PR). Apologies for the rushed code—I wanted to quickly get a correct, runnable version out to gather feedback before heading back to China (where I’ll lose internal GPU access), so I can refine the design after.

vllm/config/model_arch.py
charlotte12l: vision_config: https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/mllama4.py#L757
audio_config: https://github.com/vllm-project/vllm/blob/ac0bb2c3075f87b62afde8ea7dab10207bb71df1/vllm/model_executor/models/minicpmo.py#L562

vllm/config/model_arch.py
heheda12345: IMO this new config should include all things that the engine need and it should be a fixed set of attributes. In model definition (the model executor you mentioned?) we can allow the model vendor to use all of model_arch_config /  hf_config / any other config format that the model vendor likes to use.

vllm/config/model.py
hmellor: Thank you for clarifying! For a moment I thought we were going to rewrite them all 😅

I like your plan.

vllm/config/model_arch.py
hmellor: In the HF config we have `layer_types: list[str]` as the standard field for all models. Would it make sense to name this list `layer_types_cls: list[type[nn.Module]]`?

vllm/transformers_utils/model_arch_config_parser.py
hmellor: If all the logic for checking multiple field names is kept in one place, it'll make it easier for me to identify what needs fixing on the Transformers side.

vllm/config/model_arch.py
hmellor: I think this is for other libraries like `diffusers` so for our case it'll almost always be 1 element.

I will double check and get back to you!

hmellor: It should be safe for vLLM to assume that this is a `str`

vllm/config/model_arch.py
charlotte12l: @hmellor Yeah I looked at `layer_types` in HF config before but I found the allowed types are limited:
https://github.com/huggingface/transformers/blob/main/src/transformers/configuration_utils.py#L1280-L1285

In vLLM, we like to effectively distinguish those classes. 
https://github.com/vllm-project/vllm/blob/a86b4c58e8f72f4903d873d25510f53f7577366f/vllm/v1/worker/gpu_model_runner.py#L4605-L4679

I'm okay with the naming `layer_types_cls`, but note that `layer_types_cls` and  `layer_types` may not be very aligned in some cases though...

vllm/model_executor/models/llama.py
heheda12345: I prefer to parse this from hf_config. If it is general to all models, put it in the general hf parser, if it is special to this model, put in the parser of this model.

vllm/config/model_arch.py
hmellor: Yeah that's ok, in Transformers we don't actually always follow the `ALLOWED_LAYER_TYPES` (mamba models don't for example).

`layer_types` is just meant to be a mapping of `layer_idx` (i.e. the list index) to type of layer. Then `layer_types_cls` can be a mapping of `layer_idx` to class of layer.

charlotte12l:
Based on the previous discussion, https://github.com/charlotte12l/vllm/pull/2  (merged into current branch) has make the following modifications:
- model_arch_config only contains the minimal fields
- model_arch_config parse from hf_config
- model_executor still uses hf_config
- normalize quantization_config
- avoid functionalities duplications by fully utilizing model_arch_config in config/model.py

WIP:
- Still adding more testing for each convertors. My plan is for all the models, run before this PR and stored the results as files for model_config.get_hidden_size etc. ; then add tests to compare the stored files with model_arch_config
- per layer attn cls for kv_cache_spec will be in next PR as we need to refactor slidingwindow attention out of attention. In next PR, `is_deepseek_mla` will be removed per discussion.

tests/config/base_model_arch_groundtruth.json
zhuohan123, line 1: Nit: let's rename `test_files` to something more meaningful? or move `test_config.py` into a folder and put this json file next to it.

vllm/config/model.py
```diff
@@ -1197,10 +1169,10 @@ def get_sliding_window(self) -> int | None:
         return getattr(self.hf_text_config, "sliding_window", None)
 
     def get_vocab_size(self) -> int:
```
zhuohan123, line 1076: With this PR, can we get rid of functions like `get_vocab_size` and directly use `model.model_arch_config.vocab_size`? It would be great if we can do it in a future PR to reduce the number of hops to get these config values.

zhuohan123 (approved):
Thanks for the work!

vllm/config/model.py
hmellor: It looks like `get_vocab_size` and `get_hidden_size` didn't actually do anything before this PR.

They would just silently default to `0` instead of erroring when the fields were missing (looks like the contents of the old methods have been copied to the new `ModelArchConfigConvertorBase`?)

charlotte12l: > With this PR, can we get rid of functions like `get_vocab_size` and directly use `model.model_arch_config.vocab_size`? It would be great if we can do it in a future PR to reduce the number of hops to get these config values.

@zhuohan123  Sounds good. Will also remove all those direct hf_config access in next PRs https://github.com/search?q=repo%3Avllm-project%2Fvllm+hf_config+NOT+path%3A%2F%5Evllm%5C%2Fmodel_executor%5C%2F%2F&type=code

charlotte12l:
squashed all commits to resolve the sign-off issue.

vllm/config/model_arch.py
```diff
@@ -0,0 +1,63 @@
...
+logger = init_logger(__name__)
+
+
+@dataclass(config=ConfigDict(arbitrary_types_allowed=True))
+class ModelArchitectureConfig:
+    """
+    Configuration for model architecture that required by vLLM runtime
+    """
+
+    architectures: list[str]
```
heheda12345: See the discussion here. Can `architectures` be a `str` instead of `list[str]`?
https://github.com/vllm-project/vllm/pull/28454#discussion_r2516823474

DarkLight1337: I think no, but @hmellor would know more

heheda12345: @DarkLight1337 why can't we do this? Can you give any specific reason?

DarkLight1337: It is how HF transformers defines the config, from my understanding.

charlotte12l: I feel it's feasible to only make `model_arch_config.architectures` to be str while leave `hf_config` untouched. 

I plan in this PR, make `model_arch_config` values the same as before, and then in next PR modify `architectures` and replace those `model_config.get_xxx` `getattr(hf_config, xxx)` callsite with   `model_arch_config.xxx`

charlotte12l:
@heheda12345 @zhuohan123 @hmellor @22quinn 

During testing, I found `is_multimodal` is a bit tricky:
- https://github.com/vllm-project/vllm/blob/fccd5325874321f34daef0a8ed4d1b15b26ca34e/vllm/model_executor/models/registry.py#L264 `_MULTIMODAL_MODELS` do not contains all multimodal models, below models also `is_multimodal`. Therefore, for now, I removed `is_multimodal` in model_arch_config which only checked `_MULTIMODAL_MODELS` previously. If we want to get `is_multimodal` info during model_arch_config, we need to maintain another list @heheda12345 
    
Model | Location | Where It IS Registered
-- | -- | --
HCXVisionForCausalLM | hyperclovax_vision.py | _TEXT_GENERATION_MODELS
CLIPEmbeddingModel | clip.py | _EMBEDDING_MODELS
SiglipEmbeddingModel | siglip.py | _EMBEDDING_MODELS
Terratorch | terratorch.py | _EMBEDDING_MODELS


and below are really buggy:
- `draft_model_config.hf_config` was replaced but `hf_text_config` was the original version. Many `ModelConfig.get_xxx` actually access old `hf_text_config` .
https://github.com/vllm-project/vllm/blob/d007387aa742c25f60d9b35bc103cbaf753114c8/vllm/config/speculative.py#L398-L403 
 
- Many existing models uses `VllmConfig.with_hf_config()`, however only  `hf_config` was replaced but `hf_text_config` was the original version. Many `ModelConfig.get_xxx` actually access old `hf_text_config` 
https://github.com/vllm-project/vllm/blob/a2e9ebe9e242295a58e400835ef98a14b29c4fb0/vllm/config/vllm.py#L286-L298

DarkLight1337:
> For multimodal model, what about
> 
> ```
> model_cls = ModelRegistry._try_load_model_cls(model_arch)
> is_multimodal = supports_multimodal(model_cls)
> ```
> 
> But multimodal related API is a little bit complex as we may only use the language model part of the vision language models. #22299
> 
> If it doesn't affect the correctness of other parts, I'd like to suggest to implement this in a new PR and bring more folks related to MM for discussion.

This is not a problem as we use `MultiModalRegistry.supports_multimodal_inputs` to distinguish this case.

heheda12345 (approved):
LGTM! What's the status of multi-modal now?

charlotte12l:
> LGTM! What's the status of multi-modal now?

Sorry for the late reply, was busy with sprint. I plan to add multi-modal and architecture in next PR(e.g. update architecture https://github.com/charlotte12l/vllm/pull/6/files), as the current one is already large

PR #28455, opened 2025-11-11
[Model][MM] Extract conv layer as CustomOp

shen-shanshan:
## Purpose

Extract `ConvLayer` as `CustomOp` for better management and extensibility (especially for plugin devices).

Find more details at the comments below.

## Test Plan

I have tested this PR on ascend platform and it worked well, even getting a better performance.

## Test Result

See: https://github.com/vllm-project/vllm-ascend/pull/4198.

---
<details>

</details>


vllm/model_executor/layers/multi_modal/vision_patch_embed.py
```diff
@@ -0,0 +1,153 @@
...
+    def __init__(
+        self,
+        patch_size: int,
+        in_channels: int,
+        hidden_size: int,
+    ) -> None:
+        super().__init__()
+        self.patch_size = patch_size
+        self.in_channels = in_channels
+        self.hidden_size = hidden_size
```
Isotr0py: Hmmm, IIUC, we need `CustomOp` for `VisionPatchEmbed` because we want to optimize the convolution ops (`nn.Conv2d` / `nn.Conv3d`) inside for oot platform?

If so, I think we can expose `ConvLayer` etc as `CustomOp` instead of `VisionPatchEmbed`. Because model's vision patch embed can be quite various. 

For example, besides `Learnable2DInterpPosEmb`, it can also have `nn.Embedding`, `RMSNorm` and something else, which is quite difficult to unify its implementation:
https://github.com/vllm-project/vllm/blob/afffd3cc8a99ce1cf0f6f1687852e5519d725a3b/vllm/model_executor/models/aimv2.py#L58-L67

https://github.com/vllm-project/vllm/blob/afffd3cc8a99ce1cf0f6f1687852e5519d725a3b/vllm/model_executor/models/glm4v.py#L78-L88

shen-shanshan: @Isotr0py Oh, I think you are right, I just searched `VisionPatchEmbed` in the codebase before, and ignored other `XxxPatchEmbed` layers. I will narrow the range to just extract the `Conv` layer as custom op today. Really thanks for your suggestion!

vllm/model_executor/layers/multi_modal/conv.py
```diff
@@ -0,0 +1,180 @@
...
+            in_channels=in_channels,
+            out_channels=out_channels,
+            kernel_size=kernel_size,
+            stride=stride,
+            padding=padding,
+            dilation=dilation,
+            groups=groups,
+            bias=bias,
+            padding_mode=padding_mode,
+        )
```
Isotr0py:
```suggestion
        self.weight = nn.Parameter(
                torch.empty(
                    (in_channels, out_channels // groups, *kernel_size),
                )
            )
```
This will create a `nn.Conv2d` as a submodule inside `Conv2dLayer`, let's initialize weight and bias in `Conv2dLayer` directly.

vllm/model_executor/layers/multi_modal/conv.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def forward_native(self, x: torch.Tensor) -> torch.Tensor:
+        x = self.proj(x)
+        return x
```
Isotr0py:
```suggestion
        return F.conv2d(
            input, self.weight, self.bias, self.stride, self.padding, self.dilation, self.groups
        )
```
Then we can control the dispatch at ops-level like `F.conv2d` etc.

vllm/model_executor/layers/multi_modal/conv.py
```diff
@@ -0,0 +1,180 @@
...
+        output_size: int,
+        bias: bool = True,
+        skip_bias_add: bool = False,
+        params_dtype: torch.dtype | None = None,
+        quant_config: QuantizationConfig | None = None,
+        prefix: str = "",
+        *,
+        return_bias: bool = True,
+        disable_tp: bool = False,
+    ) -> None:
```
Isotr0py:
```suggestion
class LinearConvLayer(ConvLayerBase):
    """Conv layer with linear module."""

    def __init__(
        self,
        in_channels: int,
        out_channels: int,
        kernel_size: int,
        stride: int = 1,
        padding: int | tuple | str = 0,
        dilation: int | tuple = 1,
        groups: int = 1,
        bias: bool = True,
        padding_mode: str = "zeros",
    ) -> None:
```
Hmmm, we only use linear to replace convolution when `kernel_size == stride`, which is a special case for `nn.Conv2d`/`nn.Conv3d`. I think we can implement the linearized convolution inside `Conv2dLayer`, if not we should at least keep args consistent with `ConvLayer`. WDYT?

vllm/model_executor/layers/multi_modal/conv.py
```diff
@@ -0,0 +1,180 @@
...
+        s = f"in_channels={self.in_channels}, "
+        s += f"out_channels={self.out_channels}, "
+        s += f"kernel_size={self.kernel_size}, "
+        s += f"stride={self.stride}, "
+        s += f"padding={self.padding}, "
+        s += f"bias={self.bias}, "
+        return s
+
+
+class CausalConv2dLayer(Conv2dLayer):
```
Isotr0py, line 135:
```suggestion
@CustomOp.register("causal_conv2d")
class CausalConv2dLayer(CustomOp):
```
I think different layer should be registered as different ops.

Isotr0py (commented):
BTW, I think we can finish the convolution layer implementation in this PR and only modify several typical model's implementation. Then migrate remained models in following PR. 

Otherwise you have to modify all models' implementation in one time when refine layer implementation each time, which is quite inefficient and easy to make mistake.

vllm/model_executor/layers/multi_modal/conv.py
shen-shanshan: > Hmmm, we only use linear to replace convolution when `kernel_size == stride`, which is a special case for `nn.Conv2d`/`nn.Conv3d`. I think we can implement the linearized convolution inside `Conv2dLayer`, if not we should at least keep args consistent with `ConvLayer`. WDYT?

Oh, that's a good suggestion, maybe we can merge these two types into `Conv2dLayer` and add some checks to dispatch the forward.

vllm/model_executor/layers/multi_modal/conv.py
shen-shanshan: > I think different layer should be registered as different ops.

I just imitate the Rope layer to implement this before. It's better to register separate ops for different types as you said, I will update it today. Really thanks for your help!

vllm/model_executor/layers/multi_modal/conv.py
Isotr0py: We only replace `conv3d` with `linear` currently because of a performance regression in torch2.9 (see https://github.com/vllm-project/vllm/issues/27406).

BTW, I remember that linear is faster than convolution for a point-wise convolution computation in early pytorch, perhaps other platforms (like CPU platform) can benefit from linearized conv2d/3d because they have custom optimized gemm ops.

I think we can implement both conv layer like this to automatically linearize conv ops:
```python3
class Conv2dLayer(CustomOp):
    """Conv layer with Conv2d."""

    def __init__(
        self,
        in_channels: int,
        out_channels: int,
        kernel_size: int | tuple,
        stride: int | tuple | None,
        padding: int | tuple | str | None,
        dilation: int | tuple | None,
        groups: int | None,
        bias: bool | None,
        padding_mode: str | None,
    ) -> None:
        super().__init__()

        self.in_channels = in_channels
        self.out_channels = out_channels
        self.kernel_size = kernel_size
        self.stride = stride
        ...

        self.can_linearize = (self.kernel_size == self.stride and not self.padding)

        if self.can_linearize:
            self.weight = nn.Parameter(
                torch.empty(out_channels, in_channels * math.prod(kernel_size))
            )
        else:
            self.weight = nn.Parameter(
                torch.empty(out_channels, in_channels, *kernel_size)
            )
        ...
    
    def forward_native(self, x: torch.Tensor) -> torch.Tensor:
        .... # <- do some reshape here
        if self.can_linearize:
            return F.linear(x, self.weight, self.bias)
        else:
            return F.conv2d(
                x,
                self.weight,
                bias=self.bias,
                stride=self.stride,
                padding=self.padding,
                dilation=self.dilation,
                groups=self.groups,
            )
```

shen-shanshan: @Isotr0py Currently, I find that vllm only convert `Conv3D` to `Linear` with `conv3d_to_linear_weight()` in `vision.py` and don't apply this optimization for `Conv2D`. Could we directly implement it as below?

- `Conv2dLayer` -> only use `F.conv2d()`. (because we don't convert the `Conv2D` weight when loading)
- `Conv3dLayer` -> directly use `F.linear()` when `kernel_size==stride` for better performance.

Isotr0py: > (because we don't convert the Conv2D weight when loading)

In fact, `conv3d_to_linear_weight` is just a workaround for `nn.Conv3d` when fixing the performance regression. Since we decide to expose conv layer as CustomOP, we can optimize its usage with a better implementation.

We can implement `weight_loader` for conv layer like linear layer so that it can automatically convert weight when calling `weight_loader(param, loaded_weight)`:
```python3
set_weight_attrs(weight, {"weight_loader": self.weight_loader})
```
```python3
def weight_loader(self, param: Parameter, loaded_weight: torch.Tensor):
    if self.can_linearize:
        out_channels = loaded_weight.shape[0]
        loaded_weight = loaded_weight.reshape(out_channels, -1)
    param.data.copy_(loaded_weight)
```

Isotr0py:
BTW, let me update the layer implementation for you tonight. :)

vllm/model_executor/layers/multi_modal/conv.py
shen-shanshan: OK, that's a good idea, I will follow this to implement it today.

Isotr0py (approved):
Let's see if the CI can pass now.

shen-shanshan:
I have also test this PR on ascend platform and it worked well, even getting a better performance.

see: https://github.com/vllm-project/vllm-ascend/pull/4198.

PR #28459, opened 2025-11-11
when model_runner not ensure_kv_transfer_shutdown this method, it wil…

lengrongfu:
When i test other model_runner, find if EngineCore because oom crash, but it error info is `AttributeError: 'NoneType' object has no attribute 'ensure_kv_transfer_shutdown'`.

So we should it check model_runner wether this method.

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/v1/worker/gpu_worker.py
+++ b/vllm/v1/worker/gpu_worker.py
@@ -860,8 +860,10 @@ class Worker(WorkerBase):
         )
 
     def shutdown(self) -> None:
-        if runner := getattr(self, "model_runner", None):
-            runner.ensure_kv_transfer_shutdown()
+        if hasattr(self, "model_runner") and hasattr(
+            self.model_runner, "ensure_kv_transfer_shutdown"
+        ):
+            self.model_runner.ensure_kv_transfer_shutdown()
 
 
 def init_worker_distributed_environment(


vllm/v1/worker/gpu_worker.py
```diff
@@ -860,8 +860,10 @@ def save_tensorized_model(
         )
 
     def shutdown(self) -> None:
-        if runner := getattr(self, "model_runner", None):
-            runner.ensure_kv_transfer_shutdown()
```
heheda12345, line 864: if self doesn't have model_runner attribute, I think it will go to the else branch and do nothing. Can you take a deeper look of why "'NoneType' object has no attribute 'ensure_kv_transfer_shutdown'" will happen?

lengrongfu: Ok, i will to check.

lengrongfu: I think the problem is that `model_runner` doesn't implement the `ensure_kv_transfer_shutdown` method.

heheda12345: > model_runner doesn't implement the ensure_kv_transfer_shutdown method.

if that is the case, the error message should be  "**'XXXRunner'** object has no attribute 'ensure_kv_transfer_shutdown'"

PR #28479, opened 2025-11-11
[Perf] Refactor cudagraph_support to enable full CUDA graphs for spec decoding with FlashInfer

### LINKED ISSUE #26856: [Performance]: FalshInfer attn backend. Use dynamic AttentionCGSupport

benchislett:
## Purpose

Revised implementation of https://github.com/vllm-project/vllm/pull/26937

This PR makes `_cudagraph_support` a private member and uses `get_cudagraph_support(vllm_config, kv_cache_spec)`. Also updates `_check_and_update_cudagraph_mode` to consider support per-backend, per-kv-group. 

TRTLLM-gen kernels support full cuda graphs, but are only used with FlashInfer on Blackwell under certain conditions.
It might not be safe to change FlashInfer's cudagraph_support to UNIFORM_BATCH always, but we can still set it when we know TRTLLM-gen backend will be used.

Also update the docs to reflect the FlashInfer cuda graph compatibility, and fill in the missing entry for FlashInferMLA.

FIX https://github.com/vllm-project/vllm/issues/26856

## Test Plan

See #26937 for functional correctness testing / benchmarking. Rerunning on this branch gives the same results.

Local test run passes for `tests/v1/attention`.

### DIFF
--- a/docs/design/cuda_graphs.md
+++ b/docs/design/cuda_graphs.md
@@ -177,8 +177,9 @@ The following table lists backends that support full CUDA Graphs at the time of 
 | FlashAttention v3 | `ALWAYS` | has unified routine for both batches, so `FULL` mode is good |
 | Triton Attention | `ALWAYS` | prefer `FULL_AND_PIECEWISE` since it has different kernels for prefill/mixed and pure decode batches |
 | AITER FlashAttention | `UNIFORM_BATCH`| |
-| FlashInfer | `UNIFORM_SINGLE_TOKEN_DECODE` | |
+| FlashInfer | `UNIFORM_SINGLE_TOKEN_DECODE` | Will be set to `UNIFORM_BATCH` when using TRTLLM attention on Blackwell |
 | FlashMLA | `UNIFORM_BATCH` | |
+| FlashInferMLA | `UNIFORM_BATCH` | |
 | AITER MLA | `UNIFORM_SINGLE_TOKEN_DECODE` | |
 | CUTLASS MLA | `UNIFORM_SINGLE_TOKEN_DECODE` | |
 | Mamba attention| `UNIFORM_SINGLE_TOKEN_DECODE` | |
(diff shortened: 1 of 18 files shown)


LucasWilkinson (approved):
LGTM 

Id like to work towards reverting https://github.com/vllm-project/vllm/pull/27427 (and move back to this being an instance property) in the future; but we need broader cudagraph refactors to get there

vadiklyutiy (commented):
Before we use `use_trtllm_attention` for checking both prefill and decode. 
Right now seem `use_trtllm_attention` is using for checking prefill only and `can_use_trtllm_attention` .

May we refactor:
 - use proper name like `use_trtllm_prefill_attn` and `use_trtllm_decode_attn` 
 - remove from `use_trtllm_attention` processing of decode case

Maybe it's worth to do in separate PR

vadiklyutiy (commented):
Before we use `use_trtllm_attention` for checking both prefill and decode. 
Right now seem `use_trtllm_attention` is using for checking prefill only and `can_use_trtllm_attention` .

May we refactor:
 - use proper name like `use_trtllm_prefill_attn` and `use_trtllm_decode_attn` 
 - remove from `use_trtllm_attention` processing of decode case

Maybe it's worth to do in separate PR

vadiklyutiy (commented):
One more style
Is there some reason to hold `[_]cudagraph_support` and `get_cudagraph_support` in `*MetaBuilder` classes, maybe `*Backend(AttentionBackend)` is better place?

PR #28485, opened 2025-11-11
Support all interleaved layer types

shinoss:
## Purpose

full+sliding isnt the only interleaved layer types, i.e. we should ignore setting a global sliding window if we have other interleaved layer types, e.g. full + chunked (e.g. llama4) and full + linear (e.g. minimax). 

as long as there is more than 1 layer type, we can set is_interleaved() to true

## Test Plan

checked with sample layer types e.g. chunked_attention, chunked_attention, chunked_attention, full_attention in RedHatAI/Llama-4-Maverick-17B-128E-Instruct-FP8-block

## Test Result

previously full+chunked would return is_interleaved() == false, now it returns true

---
<details>

</details>


hmellor (changes requested):
It'd be cleaner to check `len(set(layer_types)) > 1`, right? 

The current change requires us to hard code all possible combinations of interleaved layer types.

sarckk:
@hmellor that's a fair point, updated

sarckk:
removed test

hmellor (approved):
LGTM, thanks for the improvement!

PR #28492, opened 2025-11-11
[Performance][Hopper] Avoid M dim padding to 4x for most cases (due to cuda graphs paddings)

alexm-redhat:
This PR avoids the M dimensions (num_tokens) padding to 4x (that is required by Hopper CUTLASS) in most cases due to padding of cuda graphs and their capture sizes. 

For DSR1-FP8 running on 8xH200 GPUs with batch size 32:

**Before PR:**
Mean TPOT (ms):                          31.41    

**After PR**
Mean TPOT (ms):                          30.40 

=> 3.3% TPOT improvement

### DIFF
--- a/vllm/model_executor/layers/quantization/utils/fp8_utils.py
+++ b/vllm/model_executor/layers/quantization/utils/fp8_utils.py
@@ -115,20 +115,27 @@ def _padded_cutlass(
         dim if dim % pad_multiple == 0 else dim + pad_multiple - (dim % pad_multiple)
     )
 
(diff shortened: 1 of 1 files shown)


LucasWilkinson (approved):
LGTM; please just remove the debug print cruft

alexm-redhat:
Thanks @LucasWilkinson, removed debug cruft

alexm-redhat:
rebased

mgoin (approved):
LGTM, we should definitely still update the kernel to remove padding requirement but this is a good improvement for now

PR #28493, opened 2025-11-11
[Benchmark] Add retry support to fix workload bias in multi-turn benchmark

ai-jz:
## Purpose

Fix workload bias in multi-turn benchmark where timed-out clients exit conversations immediately, reducing actual workload to model servers with lower capacity.

E.g., when workload is high (# of clients=64 or 128 with long-context requests), vLLM baseline has more timeouts than vLLM + LMCache, leading to more client exits and lower actual workload to the server. As a result, vLLM baseline shows shorter runtime and better TTFT/TPOT than LMCache.

## Solution

Add `--max-retries` parameter to retry failed requests with exponential backoff instead of exiting immediately. Default=0 (no retries) for backward compatibility.

**Note**: Most of the diff is due to the extra indentation from wrapping the request logic in a retry loop (`for attempt_cnt in range(args.max_retries + 1)`).

## Test Plan

```bash
# Test backward compatibility (default, no retries)
python benchmarks/multi_turn/benchmark_serving_multi_turn.py ...

# Test with retries enabled
python benchmarks/multi_turn/benchmark_serving_multi_turn.py --max-retries 10 ...
```

## Test Result

- ✅ Syntax validated
- ✅ Backward compatible with `max_retries=0` (default)
- ✅ Retries work with exponential backoff between attempts

### DIFF
--- a/benchmarks/multi_turn/benchmark_serving_multi_turn.py
+++ b/benchmarks/multi_turn/benchmark_serving_multi_turn.py
@@ -55,6 +55,7 @@ class ClientArgs(NamedTuple):
     verify_output: bool
     conversation_sampling: ConversationSampling
     request_rate: float
+    max_retries: int
 
 
 class RequestArgs(NamedTuple):
@@ -527,6 +528,25 @@ async def poisson_sleep(request_rate: float, verbose: bool = False) -> None:
     await asyncio.sleep(interval)
 
 
+async def exponential_backoff_sleep(
+    attempt_cnt: int,
+    base_rate: float = 1.0,
+    backoff_factor: float = 2.0,
+    jitter_fraction: float = 0.10,
+    verbose: bool = False,
+) -> None:
+    # Sleep with exponential backoff and jitter after a failed request.
+    backoff_delay = base_rate * (backoff_factor**attempt_cnt)
+    jittered_delay = backoff_delay * (
+        1 + np.random.uniform(-jitter_fraction, jitter_fraction)
+    )
+
+    if verbose:
+        logger.info(f"Backoff for {jittered_delay:.3f} seconds...")
+
+    await asyncio.sleep(jittered_delay)
+
+
 async def client_main(
     args: ClientArgs,
     req_args: RequestArgs,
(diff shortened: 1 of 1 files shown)


benchmarks/multi_turn/benchmark_serving_multi_turn.py
```diff
@@ -655,56 +665,60 @@ async def client_main(
...
+                # Remove the conversation (should not be used again)
+                active_convs.pop(conv_id)
 
-            except Exception:
-                num_failures += 1
-                logger.exception(
-                    f"{Color.RED}Client {client_id} - Exception during conversation ID {conv_id} (turn: {current_turn}){Color.RESET}"  # noqa: E501
-                )
+            if exception:
                 break  # Exit gracefully instead of raising an error
```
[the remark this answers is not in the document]
ai-jz: > The client process exits its main loop if a request fails with an exception (e.g., TimeoutError) after all retries have been attempted.

It is an existing behavior ("break" outer loop in case of exception) before this PR by design. 

> To truly fix the workload bias, the client should be more resilient. It should drop the problematic conversation but continue running and processing other conversations.

Agree. This is how it is implemented: "exception=True" means "success = False" and therefore problematic conversation is dropped. Refactor the code to make it clearer. 

@gemini-code-assist pls review again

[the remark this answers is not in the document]
ai-jz: > The client process exits its main loop if a request fails with an exception (e.g., TimeoutError) after all retries have been attempted. This seems to contradict the main goal of this PR, which is to prevent timed-out clients from exiting and reducing the workload.

It is an existing behavior ("break" outer loop in case of exception) before this PR by design. The goal of this PR is NOT to prevent timed-out clients completely (not doable without significantly increasing total runtime, if the server is simply too slow). Instead, it is to provide a lever (args.max_retries) to migrate it.

> To truly fix the workload bias, the client should be more resilient. It should drop the problematic conversation but continue running and processing other conversations.

Agree. This is how it is implemented: "exception=True" means "success = False" and therefore problematic conversation is dropped. Refactor the code to make it clearer.

/gemini review again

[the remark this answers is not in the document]
ai-jz: > However, the break statement at line 722, within the if exception: block, still causes the entire client process to exit its main while True loop if an exception occurs after all retries.

You have a good point; however, this will change the behavior before this PR when max_retries = 0. The existing code break the outer loop on purpose **for good reasons**: the exception might be due to server crash (observed in our actual benchmarking), leading to unnecessary submission of new conversations if we remove lines 721-722 as you suggested.

In short, we try keeping backward compatibility with this PR, and think the existing logic to break the outer loop is reasonable.

/gemini review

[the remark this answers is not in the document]
ai-jz: /gemini review

benchmarks/multi_turn/benchmark_serving_multi_turn.py
```diff
@@ -527,6 +528,19 @@ async def poisson_sleep(request_rate: float, verbose: bool = False) -> None:
     await asyncio.sleep(interval)
 
 
+async def exponential_backoff_sleep(
+    attempt_cnt: int, base_rate: float = 1.0, verbose: bool = False
+) -> None:
+    # Sleep with exponential backoff and jitter after a failed request.
+    backoff_delay = base_rate * (2**attempt_cnt)
+    jittered_delay = backoff_delay * (1 + np.random.uniform(-0.1, 0.1))
```
DarkLight1337: This should be configurable as well?

ai-jz: Yes, configurable.

I actually considered that, but was a bit concerned about the added complexity of exposing every parameter as configurable. The main goal here is just to mitigate immediate client exits, and the current defaults are fairly straightforward.

How about we start simple and make these configurable later if there’s clear demand?

Alternatively, if you have preferred default values, I’m open to adjusting them.

DarkLight1337: Let's just make this a parameter of the function with default values then.

benchmarks/multi_turn/benchmark_serving_multi_turn.py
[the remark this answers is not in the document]
ai-jz: > leading to an UnboundLocalError if success remains False and no exception was caught in the last attempt (e.g., if result is None).

If no exception was caught in the last attempt (e.g., if result is None), it means the server is more or less recovered and the client can continue next conversation -- this is indeed aligned with the logic in the existing code where there is no retry.

benchmarks/multi_turn/benchmark_serving_multi_turn.py
ai-jz: Btw, I think Gemini has over-complicated the situation when suggesting to remove the "break" logic, etc. I've replied with several message. Could you review if it is makes sense? It's mainly to be backward-compatible to follow the logic in the original code (when max-retries = 0). 

In fact, I think the original logic is reasonable, in spite of some limitations as pointed by Gemini. In case there is a need to update the original logic, we might better address that in a separate PR.

ai-jz: > Let's just make this a parameter of the function with default values then.

Sounds good. There are two parameters other than base_rate. I will add them shortly

```
def exponential_backoff_sleep(
    attempt_cnt: int,
    base_rate: float = 1.0,
    backoff_factor: float = 2.0,        
    jitter_fraction: float = 0.10,
    verbose: bool = False,
) -> None:
```

DarkLight1337: You can keep the current logic

ai-jz: > You can keep the current logic

Ack

> Let's just make this a parameter of the function with default values then.

Done. Pls review again.

DarkLight1337 (approved):

PR #28496, opened 2025-11-11
[Bugfix] Prevent premature multimodal preprocess cache eviction

aykoppol:
## Purpose
FIX https://github.com/vllm-project/vllm/issues/26195.

vLLM checks what media in the request has already been cached before processing unseen media. Then, it combines the previously-cached media and the newly-processed media into a single list of instructions, which are used to update the prompt before it is used for prediction.

The issue was that the newly processed items were being added to the cache before the old items were read from the cache, resulting in the edge case where old media was prematurely evicted and couldn’t be accessed.

Here, we fetch previously-cached items before adding new elements to the cache, preventing us from evicting data we need before reading it

## Test Plan

Validated that this resolved local re-pro of the initial problem.

Re-pro setup:
- Set mm-processor-cache-gb=0.05
- Send two short warm-up videos, pushing cache utilization to 60-80%
- Send a single request with 6 short (~10s) videos - this will long fail because the resulting prompt length is too long
- Re-send the same request with the same 6 short videos - this triggers the cache issue

## Test Result

Resolved the toy re-pro


jeremyteboul (approved):
LGTM, clean and simple

vllm/multimodal/processing.py
```diff
@@ -1744,10 +1754,9 @@ def _merge_mm_kwargs(
                     mm_missing_next_idx[modality] += 1
 
                     item = kwargs, updates
+                    kwargs, updates = cache.get_and_update_item(item, item_hash)
```
DarkLight1337: Overall we call `get_and_update_item` more times than the number of actual items (unlike the case in EngineCore) so I think this is still not entirely correct.

aykoppol: Not sure I totally understand? We still call `get_and_update_item` the same number of times, just in a different order than before wrt the input media.

Is there corresponding EngineCore code that needs to be kept in sync?

DarkLight1337: Sorry I missed the conditional in the dict/list comprehension, you are right. However the ordering being different will also cause desync compared to EngineCore.

DarkLight1337: > Is there corresponding EngineCore code that needs to be kept in sync?

 You can check where `get_and_update` of the receiver cache is being called

knlnguyen1802:
Thank you for the concise solution

"The issue was that the newly processed items were being added to the cache before the old items were read from the cache, resulting in the edge case where old media was prematurely evicted and couldn’t be accessed"
I believe this also potential happen at P1 process (EngineCore) I try to run your code but still see the error.

## Reproduce
I reproduce the error using Qwen2.5-VL-3B and send 20 requests, each request contain 20 images size 2048x2048

aykoppol:
The easiest fix for the issues discussed here seems to simply call `get_and_update_item` again on line 1759. This will update the cache more times on P0 than P1 - however, even if we call `get_and_update_item` multiple times on the same item in P0, we only need to mirror the last call in P1.

This will still leave our code thread-unsafe and liable to the same `missing cached item` bug. The problem is that we call `is_cached` before `get_and_update_item`, while the cache state can change in the middle.

The LRU cache is inherently not thread-safe, but even the SHM cache will not be safe as long as we rely on a separate `is_cached`. Will update this PR to expose a `get` operator that can be used in conjunction with `get_and_update`, instead of `is_cached`.

cc @DarkLight1337 @knlnguyen1802 @jeremyteboul

DarkLight1337:
This is run in a single thread, you don't have to worry about thread safety.

vllm/multimodal/processing.py
jeremyteboul: @DarkLight1337 , I think we should land this first which fixes the primary race condition; we can add unittest for the fixed race condition, and document single-threaded assumptions

Maybe in a follow we can add a thread-safe cache wrapper and add some forms of P0/P1 ordering tests and assertions

DarkLight1337: I think you should fix https://github.com/vllm-project/vllm/pull/28496#issuecomment-3520591473 first, that's the problem I was referring to by "However the ordering being different will also cause desync compared to EngineCore."

aykoppol: Updated. Borrowing from @knlnguyen1802's solution, we effectively "touch" all cached elements in P0 and P1, before calling `get_and_update` on all media in the input request sequentially

(will fix the failing test cases, wanted to put up to get eyes on it :) )

vllm/multimodal/processing.py
```diff
@@ -1671,18 +1676,18 @@ def _get_cache_missing_items(
         cache: BaseMultiModalProcessorCache,
         mm_data_items: MultiModalDataItems,
         mm_hashes: MultiModalHashes,
-    ) -> MultiModalDataItems:
-        mm_is_cached = {
-            modality: cache.is_cached(hashes) for modality, hashes in mm_hashes.items()
+    ) -> tuple[OptionalPreCachedMediaItems, MultiModalDataItems]:
+        mm_cached_items = {
+            modality: cache.get(hashes) for modality, hashes in mm_hashes.items()
```
huachenheli: Here `get` and `get_item` will update eviction order, so if anything is thrown between `_get_cache_missing_items` and `_merge_mm_kwargs`, P0 and P1 will diverge. So I think here we should only peek at the cache without updating its eviction order. cc. @DarkLight1337 @ywang96

aykoppol: Agreed. I believe that the underlying `cachetools.LRUCache` doesn't expose such a peek method (https://fburl.com/8jmfpksa)... but we could write a custom peek method into our LRU Cache implementation based on the underlying `Cache.__getitem__`

huachenheli: You can try with `__contains__`? https://github.com/tkem/cachetools/blob/e9117dc1e18389341ce4836f171e23d89f306b65/src/cachetools/__init__.py#L91C9-L91C21 It's in the base class so won't update eviction order I think.

DarkLight1337: I prefer @huachenheli 's suggestion

aykoppol: This then becomes a like-for-like replacement for `is_cached`? In order to actually retrieve the cached data without updating the eviction order, we will need to access the `_data` member somehow

DarkLight1337: Yeah, actually `is_cached_item` already calls `__contains__`

huachenheli:
> > QQ - what happens to the P1 cache if we fail a request in-between updating the P0 cache and sending the request to P1? For example, if the generated input token sequence is too long for the model? This seems to be able to de-sync the requests?
> 
> Yes that is a bug that we are aware of, feel free to fix it as well!

I feel like a more thorough fix would be to support exception handling in P1 - Currently if anything throws the engine core must crash, which exacerbates issues such as cache divergence. 

A side note: with FT python officially released in 3.14, we should consider merging processes again. For multimodal, this makes the cache sync issue go away and also give big perf gains as we no longer need two layers of IPCs.

vllm/multimodal/cache.py
```diff
@@ -335,8 +344,11 @@ def __init__(self, model_config: "ModelConfig") -> None:
         )
 
     @override
-    def is_cached_item(self, mm_hash: str) -> bool:
-        return mm_hash in self._cache
+    def peek_item(self, mm_hash: str) -> MultiModalProcessorCacheOutItem | None:
```
ywang96: Can you please fix the pre-commit? Same with other errors.

vllm/multimodal/processing.py
```diff
@@ -1671,18 +1676,18 @@ def _get_cache_missing_items(
         cache: BaseMultiModalProcessorCache,
         mm_data_items: MultiModalDataItems,
         mm_hashes: MultiModalHashes,
-    ) -> MultiModalDataItems:
-        mm_is_cached = {
-            modality: cache.is_cached(hashes) for modality, hashes in mm_hashes.items()
+    ) -> tuple[OptionalPreCachedMediaItems, MultiModalDataItems]:
+        mm_cached_items = {
+            modality: cache.peek(hashes) for modality, hashes in mm_hashes.items()
```
knlnguyen1802: I think this seem doesn't work because your peek function return result inform (None, [ResolvedPromptUpdate(..)]) for cache miss, which can help avoid AssertionError on P0 process but the error is still happen on P1 process if the media item get evict at the same request

DarkLight1337: Could you elaborate? If there is cache miss, then the item should not be `None` in P0. As a result, the corresponding item in P1 should be refreshed and thus avoid eviction.

knlnguyen1802: Sorry for confusion. 
In here the code still decide cache miss or hit before apply to cache.
If a request have multi image, a media item that hit at mm_cached_items  here can be evict during process all media of request. 
The return result of this media will still be (None, [ResolvedPromptUpdate(..)]) for this function as at the time function call, the cache it hit on the cache.

In P0 get_and_update_item check, it do the check
`assert mm_item is not None, f"Expected an address item for {mm_hash=}"`

But the value (None, [ResolvedPromptUpdate(..)]) can mislead pass this check

DarkLight1337: Ah I see. Indeed, `peek_item` should return the cached item itself. It should only be replaced with `None` in `get_and_update_item`.

DarkLight1337: Is that what you mean?

knlnguyen1802: Yes 
But even if peek_item change to return the cached item itself, I think this solution doesn't solve the problem. 
At least in my case, I can still reproduce the error.

DarkLight1337: Yeah, seems that this PR only moves the `is_cached` check earlier in the processing pipeline, but that doesn't really matter because it's single threaded anyway...

vllm/multimodal/processing.py
Bounty-hunter: In case: media A, B, C, D is (miss, miss,  hit, hit), After get_and_update A,   the D is evict; when get_add_update D later, the assert is failed.  In current pr, it seem can't handel it.  @aykoppol

vllm/multimodal/processing.py
aykoppol: Hmm, not sure I totally understand - could you share your consistent re-pro? Is it on P0 or P1?

I added a new commit that should address a few issues:
- we need corresponding logic in P1, to fetch cached data before it can be evicted
- In P0, we only store metadata in the cache. In the old commit, fetching this metadata and then re-caching it could mess up the cache sizing - updated.

Please let me know if there are any concerns. Had to shift focus last week due to competing priorities, but want to get this fixed and landed ASAP to unblock everyone :)

cc @knlnguyen1802 @DarkLight1337

vllm/multimodal/processing.py
```diff
@@ -1671,18 +1676,19 @@ def _get_cache_missing_items(
         cache: BaseMultiModalProcessorCache,
         mm_data_items: MultiModalDataItems,
         mm_hashes: MultiModalHashes,
-    ) -> MultiModalDataItems:
-        mm_is_cached = {
-            modality: cache.is_cached(hashes) for modality, hashes in mm_hashes.items()
+    ) -> tuple[OptionalPreCachedMediaItems, MultiModalDataItems]:
+        mm_cached_items = {
```
DarkLight1337: Please fix the return types

aykoppol:
Jfyi: a lot of the complexity here comes from the fact that the values you get *out* of the cache are not necessarily the ones you put *in* to the cache.

The basic flow here is theoretically simple:

- If the mm item is in cache, get and store locally
- If it is not, process it
- Merge the pre-cached and just-computed media into a single list
- Feed all media into cache sequentially

This (once properly tested and reviewed), should hopefully be easier to work with and make less assumptions moving forward

knlnguyen1802:
@aykoppol  Here is the test case I use to discover error from base code. You can adapt it to test for your code
also cc: @DarkLight1337 

<details>
<summary><strong>Test script</strong></summary>

<br>

```python
import pytest
from types import SimpleNamespace
import torch
from vllm.utils.cache import LRUCache
from vllm.multimodal.cache import MultiModalProcessorSenderCache, MultiModalReceiverCache
from vllm.multimodal.inputs import MultiModalKwargsItem

# ---- Mock configuration helpers ----

class MockModelConfig:
    def __init__(self, cache_gb: float):
        self._cache_gb = cache_gb

    def get_multimodal_config(self):
        # Simulate what real ModelConfig returns
        return SimpleNamespace(mm_processor_cache_gb=self._cache_gb)

def make_dummy_item(name):
    """Creates a fake multi-modal item."""
    return MultiModalKwargsItem.dummy(name)


# ---------------- Test function ----------------
# Target image_A will get evict during process of request 2

def test_multimodal_sender_receiver_cache_eviction():
    #
    # SETUP
    #
    model_config = MockModelConfig(cache_gb=1e-9)

    sender_cache = MultiModalProcessorSenderCache(model_config)
    receiver_cache = MultiModalReceiverCache(model_config)

    # Replace internal caches with small LRU for easy testing
    sender_cache._cache = LRUCache(6, getsizeof=lambda x: 1)
    receiver_cache._cache = LRUCache(6, getsizeof=lambda x: 1)

    # REQUEST 1: 4 unique image hashes
    request1_hashes = ["image_A", "image_B", "image_C", "image_D"]
    request1_items = {h: make_dummy_item(h) for h in request1_hashes}

    # REQUEST 2: 4 images, one shared ("image_A")
    request2_hashes = ["image_E", "image_F", "image_G", "image_A"]
    request2_items = {h: make_dummy_item(h) for h in request2_hashes}

    ##########################
    # STEP 1: Request 1 send
    ##########################
    sender_is_cached_item_req1 = [
        sender_cache.is_cached_item(h) for h in request1_hashes
    ]
    print("Request 1 cache hits (Sender):", sender_is_cached_item_req1)
    # Expect all False since empty

    ###########################
    # Process request 1 for P0 Cache
    ###########################
    for i, h in enumerate(request1_hashes):
        # Use precomputed cache state
        is_cached = sender_is_cached_item_req1[i]
        item_tuple = (request1_items[h], []) if not is_cached else None

        sender_cache.get_and_update_item(item_tuple, h)
    
    ###########################
    # Process request 1 for P1 Cache
    ###########################
    for h in request1_hashes:
        receiver_cache.get_and_update_item(request1_items[h], h)

    # After request 1, caches each contain image_A–D
    print("Sender contents after Req1:", list(sender_cache._cache.keys()))
    print("Receiver contents after Req1:", list(receiver_cache._cache.keys()))

    ##########################
    # STEP 2: Request 2 send
    ##########################
    sender_is_cached_item_req2 = [
        sender_cache.is_cached_item(h) for h in request2_hashes
    ]
    print("Request 2 cache hits (Sender):", sender_is_cached_item_req2)
    # image_A should be True, others False

    ###########################
    # Process request 2 for P0 Cache
    ###########################
    for i, h in enumerate(request2_hashes):
        # Use precomputed cache state again
        is_cached = sender_is_cached_item_req2[i]
        item_tuple = (request2_items[h], []) if not is_cached else None
        print(f"Request 2: key={h} | cached={is_cached}")

        sender_cache.get_and_update_item(item_tuple, h)
    
    ###########################
    # Process request 2 for P1 Cache
    ###########################
    for h in request2_hashes:
        receiver_cache.get_and_update_item(request2_items[h], h)

    ##########################
    # STEP 3: Final states
    ##########################
    print("Sender contents after Req2:", list(sender_cache._cache.keys()))
    print("Receiver contents after Req2:", list(receiver_cache._cache.keys()))

    sender_keys = list(sender_cache._cache.keys())
    receiver_keys = list(receiver_cache._cache.keys())

    print("✅ LRU logic and re-use verified successfully.")


# Run manually
if __name__ == "__main__":
    test_multimodal_sender_receiver_cache_eviction()

```

DarkLight1337:
Made mypy pass, but the test still fails in the same way as before.

> The issue was that the newly processed items were being added to the cache before the old items were read from the cache,
> resulting in the edge case where old media was prematurely evicted and couldn’t be accessed.

Tbh I think we might have to predict if the item will be evicted (e.g. by checking its position in the eviction order) by the current batch, and if so, treat that item in the same way as an uncached item.

nwaughachukwuma (changes requested):
Good work here, thanks for the effort.

I have two questions/suggestions:
1. given there slight differences, couldn't `is_cached` and `is_cached_item` live side by side with `peek` and `peek_item`?
2. if the answer to the above is NO, then be aware that this PR would result in MAJOR breaking changes, and as such you should prepare for it by updating the docs, tags and semvar versioning.

DarkLight1337:
> by checking its position in the eviction order

I have updated this PR with this idea, but the test still fails. I will look in more detail after dinner...

An advantage of this idea is that there is much less change to the existing code.

vllm/multimodal/cache.py
```diff
@@ -335,8 +338,8 @@ def __init__(self, model_config: "ModelConfig") -> None:
         )
 
     @override
-    def is_cached_item(self, mm_hash: str) -> bool:
-        return mm_hash in self._cache
+    def is_cached_item(self, mm_hash: str, *, n: int = 0) -> bool:
+        return not any(mm_hash == k for k in islice(self._cache.order, n))
```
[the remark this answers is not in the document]
DarkLight1337: Fixed

vllm/multimodal/cache.py
```diff
@@ -389,8 +392,8 @@ def __init__(self, model_config: "ModelConfig") -> None:
...
```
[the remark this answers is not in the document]
DarkLight1337: Fixed

tests/multimodal/test_cache.py
```diff
@@ -225,3 +227,95 @@ def test_ipc_enable_disable_consistency(is_cached_calls_per_iter):
...
+    # Process request 2 for P1 Cache
+    ###########################
+    for h in request2_hashes:
+        receiver_cache.get_and_update_item(request2_items[h], h)
+
+    ##########################
+    # STEP 3: Final states
+    ##########################
+    print("Sender contents after Req2:", list(sender_cache._cache.keys()))
+    print("Receiver contents after Req2:", list(receiver_cache._cache.keys()))
```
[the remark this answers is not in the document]
DarkLight1337: Fixed

vllm/multimodal/cache.py
```diff
@@ -463,8 +500,9 @@ def _stat(self, *, delta: bool = False) -> CacheInfo:
         return info
 
     @override
-    def is_cached_item(self, mm_hash: str) -> bool:
-        return self._shm_cache.is_cached(mm_hash)
+    def is_cached_item(self, mm_hash: str, *, n: int = 0) -> bool:
+        # TODO
```
DarkLight1337: Going to address this next

knlnguyen1802: I think shm cache do have the same problem

DarkLight1337: Do or don't?

DarkLight1337: Anyways, I have added back the old code.

knlnguyen1802: Test case to reproduce error for SHM Cache
<details>
<summary><strong>Test script</strong></summary>

<br>

```python
import pytest
import multiprocessing as mp
from multiprocessing.synchronize import Lock as LockType
from typing import Any
from unittest.mock import MagicMock, patch

import torch
from vllm.config import ModelConfig, VllmConfig, ParallelConfig
from vllm.multimodal.cache import (
    ShmObjectStoreSenderCache,
    ShmObjectStoreReceiverCache,
)
from vllm.multimodal.inputs import (
    MultiModalKwargsItem,
    MultiModalFieldElem,
    MultiModalBatchedField,
)
from vllm.utils.mem_constants import MiB_bytes


def create_mock_image_data(size_mb: int) -> torch.Tensor:
    """Create mock image tensor of specified size in MB."""
    num_elements = (size_mb * MiB_bytes) // 4  # 4 bytes per float32
    return torch.randn(num_elements, dtype=torch.float32)


def create_mock_mm_kwargs_item(modality: str = "image") -> MultiModalKwargsItem:
    """Create a mock MultiModalKwargsItem with 1MB of data."""
    image_data = create_mock_image_data(1)  # 1MB
    
    elem = MultiModalFieldElem(
        modality=modality,
        key="image",
        data=image_data,
        field=MultiModalBatchedField(),
    )
    return MultiModalKwargsItem.from_elems([elem])


def create_mock_vllm_config(cache_size_gb: float = 0.003) -> VllmConfig:
    """Create mock VllmConfig with specified cache size."""
    mock_config = MagicMock(spec=VllmConfig)
    
    # Setup parallel config
    mock_parallel_config = MagicMock(spec=ParallelConfig)
    mock_parallel_config.world_size = 1
    mock_parallel_config._api_process_count = 1
    mock_parallel_config.data_parallel_size = 1
    mock_parallel_config.data_parallel_external_lb = False
    
    # Setup model config
    mock_model_config = MagicMock(spec=ModelConfig)
    mock_mm_config = MagicMock()
    mock_mm_config.mm_processor_cache_gb = cache_size_gb  # 3MB total
    mock_mm_config.mm_shm_cache_max_object_size_mb = 2  # Max 2MB per object
    mock_model_config.get_multimodal_config.return_value = mock_mm_config
    
    mock_config.parallel_config = mock_parallel_config
    mock_config.model_config = mock_model_config
    
    return mock_config


@pytest.mark.parametrize("cache_size_gb", [0.003])  # 6MB
def test_cache_eviction_during_request_processing(cache_size_gb: float):
    """
    Test that image_A gets evicted during request2 processing after image_G is added.
    
    Scenario:
    - Cache size: 3MB
    - Each image: 1MB
    - Request1: 3 images (A, B, C) = 3MB
    - Request2: 3 images (E, F, A) = 3MB
    
    Expected eviction during Request2:
    After adding image_R (1st image in request2):
    - Cache has: A, B, C, E (4MB > 3MB limit)
    - Need to free space → Evict A,B,C (FIFO, oldest)
    - When P1 tries to get A later, it should fail
    """
    vllm_config = create_mock_vllm_config(cache_size_gb)
    
    # Create sender cache (P0)
    sender_cache = ShmObjectStoreSenderCache(vllm_config)
    
    # Create shared lock for receiver
    shared_lock = mp.Lock()
    
    # Create receiver cache (P1)
    receiver_cache = ShmObjectStoreReceiverCache(vllm_config, shared_lock)
    
    # Image hashes
    request1_hashes = ["image_A", "image_B", "image_C"]
    request2_hashes = ["image_E", "image_F", "image_A"]
    #request2_hashes = ["image_A", "image_B", "image_C"]
    
    sender_is_cached_item_req1 = [
        sender_cache.is_cached_item(h) for h in request1_hashes
    ]
    # ===== PROCESS REQUEST 1 =====
    print("\n=== PROCESSING REQUEST 1 ===")
    request1_items = []
    request1_cached_items = []
    
    for i,mm_hash in enumerate(request1_hashes):
        is_cached = sender_is_cached_item_req1[i]
        # Create image item (1MB each)
        print(f"Is cached for hash {mm_hash} is {is_cached}")
        mm_item = (create_mock_mm_kwargs_item(), []) if not is_cached else None
        print(f"mm_item for hash {mm_hash} is {mm_item}")
        # Sender: get_and_update_item
        sender_result = sender_cache.get_and_update_item(mm_item, mm_hash)
        # print(f"Sender request is {sender_result}")
        request1_items.append(sender_result)
        
        print(f"Request1 - Sender processed {mm_hash}: "
              f"cached={mm_hash in sender_cache._shm_cache.key_index}")
        print(f"Request1 during process - Sender cached keys: {list(sender_cache._shm_cache.key_index.keys())}")

    for i,mm_hash in enumerate(request1_hashes):
        # Receiver: get_and_update_item (must be called after sender)
        receiver_result = receiver_cache.get_and_update_item(request1_items[i][0], mm_hash)
    
    print(f"\nRequest1 Complete - Sender cache size: {len(sender_cache._shm_cache.key_index)} items")
    print(f"Request1 Complete - Sender cached keys: {list(sender_cache._shm_cache.key_index.keys())}")
    
    # # ===== PROCESS REQUEST 2 =====
    print("\n=== PROCESSING REQUEST 2 ===")
    request2_items = []
    request2_cached_items = []
    sender_is_cached_item_req2 = [
        sender_cache.is_cached_item(h) for h in request2_hashes
    ]
    print(f"Before process request 2 sender_is_cached_item_req2 is {sender_is_cached_item_req2}")
    for i, mm_hash in enumerate(request2_hashes):
        # Create image item (1MB each)
        is_cached = sender_is_cached_item_req2[i]
        mm_item = (create_mock_mm_kwargs_item(), []) if not is_cached else None
        print(f"Request2 before process - Sender cached keys: {list(sender_cache._shm_cache.key_index.keys())}")
        # Sender: get_and_update_item
        sender_result = sender_cache.get_and_update_item(mm_item, mm_hash)
        request2_items.append(sender_result)
        
        sender_cached_keys = list(sender_cache._shm_cache.key_index.keys())
        print(f"Request2 - Sender processed {mm_hash}: "
              f"cached={mm_hash in sender_cache._shm_cache.key_index}")
        print(f"Request2 during process - Sender cached keys: {list(sender_cache._shm_cache.key_index.keys())}")
    
    for i,mm_hash in enumerate(request2_hashes):
        # Receiver: get_and_update_item (must be called after sender)
        receiver_result = receiver_cache.get_and_update_item(request2_items[i][0], mm_hash)
    
    # Cleanup
    sender_cache.clear_cache()
    receiver_cache.clear_cache()


if __name__ == "__main__":
    # Run the test directly
    test_cache_eviction_during_request_processing(cache_size_gb=0.003)
    print("\n✓✓✓ All tests passed! ✓✓✓")
```
</details>

I will adapt to this code in case I have time

vllm/multimodal/cache.py
```diff
@@ -389,8 +409,25 @@ def __init__(self, model_config: "ModelConfig") -> None:
...
 
     @override
-    def is_cached_item(self, mm_hash: str) -> bool:
-        return mm_hash in self._cache
+    def is_cached_item(self, mm_hash: str, *, n: int = 0) -> bool:
+        cache = self._cache
+
+        return mm_hash in cache and not any(
+            mm_hash == k for k in islice(cache.order, n)
+        )
```
aykoppol, line 417: Is this a function of the size of each individual processed media as well? Are all media guaranteed to take the same space in the cache?

knlnguyen1802: It's only the keys order.

DarkLight1337: Good point, I need to take that into account anyway for SHM cache

aykoppol: The question is more wrt whether this check is properly aligned with the P0/P1 cache eviction logic

vllm/multimodal/cache.py
knlnguyen1802: @DarkLight1337 I think this change is good if it can solve the SHM too here, could you also help to add the test for SHM here after fix? Since I can't test it now.

vllm/multimodal/cache.py
DarkLight1337: Yes we need to check by size instead of number of items. I will update it

vllm/multimodal/cache.py
DarkLight1337: ok let me try to adapt the test

DarkLight1337:
Now I'm just focusing on updating the tests so they can be copied into #28525

vllm/multimodal/inputs.py
```diff
@@ -721,12 +721,12 @@ class MultiModalKwargsItem(UserDict[str, MultiModalFieldElem]):
     """
 
     @staticmethod
-    def dummy(modality: str):
+    def dummy(modality: str, nbytes: int = 1):
```
DarkLight1337, line 724: @knlnguyen1802 make sure you apply this as well

vllm/multimodal/processing.py
```diff
@@ -1671,9 +1677,11 @@ def _get_cache_missing_items(
         cache: BaseMultiModalProcessorCache,
         mm_data_items: MultiModalDataItems,
         mm_hashes: MultiModalHashes,
-    ) -> MultiModalDataItems:
+    ) -> tuple[MultiModalIsCached, MultiModalDataItems]:
```
DarkLight1337, line 1680: And we can save some recomputation by passing through `mm_is_cached`

aykoppol:
> I think https://github.com/vllm-project/vllm/pull/28525 is cleaner now regardless

Agreed - it is much simpler to review.

I do feel we will need to revisit this restructuring in the future regardless (at least with the introduction of free-threaded Python :) )

PR #28497, opened 2025-11-11
[Kernel] Unified attention kernel performance tuning

cagrikymk:
## Purpose
The purpose of the PR is to optimize the unified attention kernel and restructures the config. selection process.
This PR creates functions to provide config for 2d, 3d attn. kernels and the selection logic between them.

Besides adding AMD specific configs:

- Changing` tl.exp` to `tl.math.exp2` and adding related scaling
- Adding conditional cache modifiers for q and K/V loading to better use the caches
- Reordering the grid for the 2D kernel for better XCD locality
- Masking related simplifications (HEAD_SIZE and TILE_SIZE related masking)

## Evaluation

 Server:
```
export VLLM_ROCM_USE_AITER=1
export VLLM_DISABLE_COMPILE_CACHE=1
export VLLM_ROCM_USE_AITER_MHA=0
vllm serve $model \
--tensor-parallel 1 \
--no-enable-prefix-caching --disable-log-requests \
--block-size 64 \
--compilation-config '{"compile_sizes": [1, 2, 4, 8, 16, 24, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192]}' \
--swap-space 16 \
--gpu-memory-utilization 0.95 \
--async-scheduling
```
### GPT-OSS-120b Performance
 TP1, ISL=8K, OSL=128 and run on MI300.
 
 
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<body link="#0563C1" vlink="#954F72">

Concurrency | 4 | 8 | 16 | 32 | 64
-- | -- | -- | -- | -- | --
Main | 166.76 | 196.1 | 216.56 | 236.14 | 250.22
This PR | 203.11 | 255.97 | 293.68 | 321.33 | 340.95
Speedup | 1.22 | 1.31 | 1.36 | 1.36 | 1.36

</body>

</html>

### GPT-OSS-120b Accuracy

`lm_eval --model local-completions --model_args model=$MODEL,base_url=http://0.0.0.0:8000/v1/completions,max_gen_toks=2048 --tasks gsm8k --num_fewshot 5 --batch_size 64 --apply_chat_template`

This PR:
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9447|±  |0.0063|
|     |       |strict-match    |     5|exact_match|↑  |0.7316|±  |0.0122|
```

Main:

```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9439|±  |0.0063|
|     |       |strict-match    |     5|exact_match|↑  |0.7559|±  |0.0118|
```

### Llama-3.1-8B-Instruct-FP8-KV Performance
TP1, ISL=8K, OSL=128 and run on MI300.
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<body link="#0563C1" vlink="#954F72">

Concurrency | 4 | 8 | 16 | 32 | 64
-- | -- | -- | -- | -- | --
Main | 219 | 246.82 | 249.88 | 269.96 | 279.41
This PR | 307.05 | 400.46 | 442.9 | 492.49 | 493.99
Speedup | 1.40 | 1.62 | 1.77 | 1.82 | 1.77

</body>

</html>

### Llama-3.1-8B-Instruct-FP8-KV Accuracy
`lm_eval --model local-completions --model_args model=$MODEL,base_url=http://0.0.0.0:8000/v1/completions --tasks gsm8k --num_fewshot 5 --batch_size 64`
This PR:
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.7544|±  |0.0119|
|     |       |strict-match    |     5|exact_match|↑  |0.7074|±  |0.0125|
```
Main:
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.7544|±  |0.0119|
|     |       |strict-match    |     5|exact_match|↑  |0.7149|±  |0.0124|
```

### DIFF
--- a/vllm/attention/ops/triton_unified_attention.py
+++ b/vllm/attention/ops/triton_unified_attention.py
@@ -7,12 +7,13 @@
 #  - Chih-Chieh Yang <chih.chieh.yang@ibm.com>
 #  - Thomas Parnell <tpa@zurich.ibm.com>
 
+import math
+
 import torch
 
 from vllm.logger import init_logger
 from vllm.platforms import current_platform
 from vllm.triton_utils import tl, triton
-import math
 
 logger = init_logger(__name__)
 float8_info = torch.finfo(current_platform.fp8_dtype())
@@ -272,7 +273,7 @@ def kernel_unified_attention_2d(
     query_mask_0 = query_pos < cur_batch_query_len
     query_mask_1 = query_offset_1 < num_query_heads
 
-    if ALL_DECODE or BLOCK_M >= num_query_heads:
+    if ALL_DECODE or num_query_heads <= BLOCK_M:
         Q_cache_modifier: tl.constexpr = ".cg"
     else:
         Q_cache_modifier: tl.constexpr = ""
--- a/vllm/attention/ops/triton_unified_attention.py
+++ b/vllm/attention/ops/triton_unified_attention.py
@@ -191,7 +191,7 @@ def kernel_unified_attention_2d(
     seq_lens_ptr,  # [num_seqs]
     alibi_slopes_ptr,  # [num_query_heads]
     qq_bias_ptr,  # [num_query_tokens, num_query_tokens]
-    scale: tl.constexpr,  # float32
+    scale,  # float32
     k_scale,  # float32
     v_scale,  # float32
     out_scale,  # float32
@@ -564,7 +564,7 @@ def kernel_unified_attention_3d(
 
     # needed to use exp2 (exp2 -> exp conversion)
     RCP_LN2 = 1.4426950408889634
-    qk_scale: tl.constexpr = scale * RCP_LN2
+    qk_scale = scale * RCP_LN2
 
     seq_idx = find_seq_idx(
         query_start_len_ptr, q_block_global_idx, num_seqs, BLOCK_Q, True
--- a/vllm/attention/ops/triton_unified_attention.py
+++ b/vllm/attention/ops/triton_unified_attention.py
@@ -12,9 +12,134 @@ import torch
 from vllm.logger import init_logger
 from vllm.platforms import current_platform
 from vllm.triton_utils import tl, triton
+import math
 
 logger = init_logger(__name__)
 float8_info = torch.finfo(current_platform.fp8_dtype())
+
+def select_2d_config(
+    block_size,
+    head_size,
+    sliding_window,
+    all_decode,
+    max_seqlen_q,
+    max_seqlen_k,
+    num_queries_per_kv,
+    num_2d_prgms,
+    element_size,
+):
+    BLOCK_M = (
+        16 if num_queries_per_kv <= 16 else triton.next_power_of_2(num_queries_per_kv)
+    )
+    if current_platform.is_rocm():
+        TILE_SIZE = 64
+        # in case head_size is large
+        max_num_stages_2d = 4
+        if head_size > 128:
+            max_num_stages_2d = 2
+        if not all_decode:
+            num_stages_2d = 1
+            num_warps = 2
+        else:
+            num_stages_2d = 3
+            num_warps = 2
+
+        if max_seqlen_q >= 256:
+            BLOCK_M = 128
+            num_stages_2d = 1
+            num_warps = 4
+        BLOCK_Q = BLOCK_M // num_queries_per_kv
+        num_stages_2d = min(max_num_stages_2d, num_stages_2d)
+        return {
+            "BLOCK_M": BLOCK_M,
+            "BLOCK_Q": BLOCK_Q,
+            "TILE_SIZE": TILE_SIZE,
+            "num_warps": num_warps,
+            "num_stages": num_stages_2d,
+            "waves_per_eu": 2,
+        }
+    else:
+        TILE_SIZE = 32
+        BLOCK_Q = BLOCK_M // num_queries_per_kv
+        return {"BLOCK_M": BLOCK_M, "BLOCK_Q": BLOCK_Q, "TILE_SIZE": TILE_SIZE}
+
+
+def select_3d_config(
+    head_size, block_size, max_seqlen_k, num_2d_prgms, element_size
+):
+    if current_platform.is_rocm():
+        cu_count = current_platform.get_cu_count()
+        if head_size < 128 or element_size == 1:
+            cu_mult = 4
+            MIN_SEGMENTS = 8
+            attn_warps = 2
+        else:
+            cu_mult = 2
+            MIN_SEGMENTS = 4
+            attn_warps = 2
+        target_num_prgms = cu_count * cu_mult
+        TILE_SIZE = 64
+        reduce_num_warps = 2
+
+        num_segments = math.ceil(target_num_prgms / num_2d_prgms)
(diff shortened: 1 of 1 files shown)


vllm/attention/ops/triton_unified_attention.py
```diff
@@ -134,27 +266,39 @@ def kernel_unified_attention_2d(
...
+        dim_mask = offs_d < HEAD_SIZE
+    else:
+        dim_mask = tl.full((1,), 1, dtype=tl.int1)
+    query_mask_0 = query_pos < cur_batch_query_len
+    query_mask_1 = query_offset_1 < num_query_heads
 
+    if ALL_DECODE or num_query_heads <= BLOCK_M:
+        Q_cache_modifier: tl.constexpr = ".cg"
+    else:
+        Q_cache_modifier: tl.constexpr = ""
```
tdoublep, line 279: Could we add a note explaining why this is expected to help?

vllm/attention/ops/triton_unified_attention.py
```diff
@@ -134,27 +266,39 @@ def kernel_unified_attention_2d(
         + offs_d[None, :]
     )
 
-    dim_mask = tl.where(offs_d < HEAD_SIZE, 1, 0).to(tl.int1)
-    query_mask_0 = tl.where(query_pos < cur_batch_query_len, 1, 0).to(tl.int1)
-    query_mask_1 = tl.where(query_offset_1 < num_query_heads, 1, 0).to(tl.int1)
+    if HEAD_SIZE_PADDED != HEAD_SIZE:
+        dim_mask = offs_d < HEAD_SIZE
+    else:
+        dim_mask = tl.full((1,), 1, dtype=tl.int1)
```
tdoublep, line 272: Is there a significant performance improvement from this specific change?

vllm/attention/ops/triton_unified_attention.py
```diff
@@ -217,10 +361,15 @@ def kernel_unified_attention_2d(
...
+    KV_cache_modifier: tl.constexpr = ".cg" if ALL_DECODE else ""
     # iterate through tiles (now limited to the sliding window range)
     for j in range(tile_start, tile_end):
         seq_offset = j * TILE_SIZE + offs_t
-        tile_mask = seq_offset < max_seq_prefix_len
+        # to reduce the masking effect when not needed
+        if TILE_SIZE == BLOCK_SIZE:
+            tile_mask = tl.full((1,), 1, dtype=tl.int1)
+        else:
+            tile_mask = seq_offset < max_seq_prefix_len
```
tdoublep, line 372: Same question here

vllm/attention/ops/triton_unified_attention.py
```diff
@@ -275,10 +426,13 @@ def kernel_unified_attention_2d(
...
 
-        S += scale * tl.dot(Q, K)
+        # qk_scale = scale * RCP_LN2 (log_2 e) so that we can use exp2 later
+        S += qk_scale * tl.dot(Q, K)
 
         if USE_SOFTCAP:
-            S = apply_softcap(S, softcap)
+            # softcap here uses exp2 and consumes RCP_LN2 conversion.
+            # multiply by RCP_LN2 again to be used in later exp2
+            S = apply_softcap(S, softcap) * RCP_LN2
```
tdoublep, line 435: This log2 change makes the algorithm harder to follow. Could we quantify what this specific change brings to the performance?

tdoublep (commented):
I'm totally fine with introducing the AMD-specific configs. This is important to get platform portability without using introducing auto-tuning overhead. 

I have some concerns about whether some of the other changes are really necessary (e.g. log2 stuff, masking changes). On H100 they don't really seem to have any effect.

vllm/attention/ops/triton_unified_attention.py
cagrikymk: This one has ignorable benefit, will remove to keep the diff minimal

cagrikymk:
Here is a table summarizing the benefit of the following changes on MI300:

1. **No dim mask:** Removing the dim. mask trick
2. **No exp2**: Replacing exp2 with exp
3. **No q cg:** Removing cache modifier from the q load
4. **No tile mask:** Removing the tile mask trick

I can do (1) and (3) as their effect is minimal if you prefer to minimize the diff.

What do you think @tdoublep?


**Num KV heads:** 8
**Num Q heads:** 64
**Dtype:** bf16
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">



<body link="#467886" vlink="#96607D">


window_size | seq q len | seq kv len | batch size | No Dim Mask | No exp2 | No q cg | No Tile Mask
-- | -- | -- | -- | -- | -- | -- | --
0 | 1 | 4096 | 8 | 0.95 | 0.96 | 1.00 | 0.97
0 | 1 | 4096 | 16 | 0.98 | 0.96 | 1.00 | 0.98
0 | 1 | 4096 | 32 | 1.00 | 0.97 | 1.00 | 1.01
0 | 1 | 4096 | 64 | 1.00 | 0.97 | 1.00 | 0.99
0 | 1 | 4096 | 128 | 1.00 | 0.90 | 1.00 | 1.00
0 | 4096 | 4096 | 1 | 1.00 | 0.86 | 1.00 | 0.98
0 | 4096 | 4096 | 2 | 1.00 | 0.83 | 1.00 | 0.95
128 | 1 | 4096 | 8 | 1.01 | 0.96 | 1.02 | 0.95
128 | 1 | 4096 | 16 | 0.99 | 0.96 | 1.00 | 0.93
128 | 1 | 4096 | 32 | 0.99 | 0.93 | 0.99 | 0.93
128 | 1 | 4096 | 64 | 1.00 | 0.92 | 0.99 | 0.95
128 | 1 | 4096 | 128 | 0.99 | 0.94 | 0.99 | 0.95
128 | 4096 | 4096 | 1 | 1.00 | 0.92 | 0.94 | 1.01
128 | 4096 | 4096 | 2 | 1.00 | 0.91 | 0.96 | 0.99
  |   |   | AVG: | 0.99 | 0.93 | 0.99 | 0.97
  |   |   | MAX: | 1.01 | 0.97 | 1.02 | 1.01
  |   |   | MIN: | 0.95 | 0.83 | 0.94 | 0.93



</body>

</html>

cagrikymk:
@tdoublep Oh, I forgot to include some important details.

These numbers are the speedup against the current code in the PR. 

Each column undoes a change and shows how much the perf. is affected by it. 

So, if the number is low, it means it is needed for good perf., hence, as you said, lower is better as that indicates the importance of keeping that optimization.

skysnow2001:
Any update on this?

PR #28506, opened 2025-11-12
[TPU][Core] Enable Pipeline Parallelism on TPU backend

Chenyaaang:
Enable Pipeline Parallelism on TPU backend

This pr includes changes on vLLM side:
- `multiproc_executor.py`: Extract `_get_parallel_sizes` and `_post_init_executor` methods, so that TPU (and other hardware) can further override them if necessary. 
- `ray_utils.py`: Extract `_is_intermediate_tensors` for TPU to override (to accept jax's version IntermediateTensor)

The command to enable PP on TPU platform is same as other platforms, but PP hasn't been supported on all Jax models, so add env var `MODEL_IMPL_TYPE=vllm` to use pytorch impl. PP can be used on single host or multi-host (with Ray).

Example command: `MODEL_IMPL_TYPE=vllm TPU_BACKEND_TYPE=jax vllm serve Qwen/Qwen3-32B --pipeline-parallel-size 4`

### DIFF
--- a/vllm/v1/executor/multiproc_executor.py
+++ b/vllm/v1/executor/multiproc_executor.py
@@ -60,10 +60,6 @@ from vllm.v1.worker.worker_base import WorkerWrapperBase
 
 logger = init_logger(__name__)
 
-"""
-MultiprocExecutor: the manager process, manages the workers.
-"""
-
 
 class FutureWrapper(Future):
     def __init__(
@@ -233,8 +229,6 @@ class MultiprocExecutor(Executor):
                         uw.death_writer.close()
                 self._ensure_worker_termination([uw.proc for uw in unready_workers])
 
-        # self.futures_queue = deque[tuple[FutureWrapper, Callable]]()
-
         self.output_rank = self._get_output_rank()
 
     def start_worker_monitor(self, inline=False) -> None:
@@ -298,7 +292,6 @@ class MultiprocExecutor(Executor):
             timeout=envs.VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS,
             kv_output_aggregator=self.kv_output_aggregator,
         )
-        # logger.info(f'[debug] [multiproc executor] execute_model {len(outputs)=}, {type(outputs[0])=}')
 
     def execute_dummy_batch(self) -> None:
         self.collective_rpc("execute_dummy_batch", unique_reply_rank=self.output_rank)
@@ -565,7 +558,6 @@ class WorkerProc:
             "is_driver_worker": is_driver_worker,
             "shared_worker_lock": shared_worker_lock,
         }
-
         wrapper.init_worker(all_kwargs)
         self.worker = wrapper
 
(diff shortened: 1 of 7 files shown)


vllm/v1/executor/multiproc_executor.py
```diff
@@ -104,16 +105,24 @@ def _init_executor(self) -> None:
...
+                f"divisible by nnodes_within_dp "
+                f"({self.parallel_config.nnodes_within_dp}). "
+            )
+            self.local_world_size = self.parallel_config.local_world_size
+            tp_size = self.parallel_config.tensor_parallel_size
+            pp_size = self.parallel_config.pipeline_parallel_size
+            pcp_size = self.parallel_config.prefill_context_parallel_size
+        else:
+            # Jax handles TP, PCP with SPMD, world_size = pp_size.
+            self.world_size = self.parallel_config.pipeline_parallel_size
```
mrjunwan-lang: will this code path be called if pp = 1, tp =xx , or dp = xx?
if yes, the self.world_size = pp size may not be right here.

Chenyaaang: will this code path be called if pp = 1, tp =xx , or dp = xx? -No
If pp=1, it will use UniprocExecutor, this multiProcExecutor is only used when pp > 1

yaochengji: Is it possible to abstract a method to set different parallelism sizes? Then each hardware plugin can override it.

vllm/v1/executor/multiproc_executor.py
```diff
@@ -199,6 +208,16 @@ def _init_executor(self) -> None:
             # Wait for all remote response mqs to be ready.
             for response_mq in self.response_mqs:
                 response_mq.wait_until_ready()
+
+            self.futures_queue = deque[tuple[FutureWrapper, Callable]]()
+
+            # set up jax transfer connection.
```
yaochengji: Is it possible to add a `_post_init_executor` for hardware plugins to override?

vllm/v1/executor/multiproc_executor.py
```diff
@@ -425,6 +442,8 @@ def _get_output_rank(self) -> int:
         # 16-23, PP rank 2
         # 24-31, PP rank 3
         # so world_size - tp_size = 32 - 8 = 24 should be PP rank = -1 (i.e. 3)
+        if USE_TPU_INFERENCE:
```
yaochengji: This method is very small, we can override it directly.

yaochengji (commented):
Thanks for your contribution! My main concern is that the modification looks a little intrusive to me. I'm wondering if we can add a subclass in the tpu-inference repo.

vllm/v1/executor/multiproc_executor.py
Chenyaaang: Yes, added `_get_parallel_sizes` method.

yaochengji:
@Chenyaaang could you please update your description? `get_pp_group` doesn't exist in the latest version.

yaochengji (approved):
LGTM, thanks!

PR #26016, opened 2025-10-01
[MISC] `cudagraph_capture_sizes`  related improvements

fhl2000:
## Purpose

Part of the CompilationConfig improvements ([#20283](https://github.com/vllm-project/vllm/issues/20283)), asked by @ProExpertProg and @WoosukKwon:

- rename max_capture_size -> max_cudagraph_capture_size: used to specify a single size, fills in the rest
- cudagraph_capture_sizes -> stays the same, used only to specify a full list
- SchedulerConfig.cuda_graph_sizes: -> remove
- Also stop reversing the sizes, always store them in ascending order (already on the issue), and always sort when using so we don't rely on them being sorted

Updated:  
- The assignment of  cudagraph_capture_size changed, from a uniformly stepped size 8 to a 2-level step size:  step_size =16 for size >= 256, and step_size =8 for size <256. (Asked by @ProExpertProg ) https://github.com/vllm-project/vllm/pull/26016#discussion_r2394921792
-  new CLIs `--cudagraph-capture-sizes ` and `--max-cudagraph-capture-size`, for corresponding settings in compilation_config respectively. 
- CLI `--cuda-graph-sizes` is deprecated and will be removed in 0.13.0 or 1.0.0, and is equivalent to `--cudagraph-capture-sizes` now. 
---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #28515, opened 2025-11-12
[BugFix] Ensure `EngineArgs.create_engine_config` is idempotent

njhill:
Issue introduced in https://github.com/vllm-project/vllm/pull/26016.

This was in particular breaking `--api-server-count` > 1 with cudagraph size related args.

Thanks to @elvischenv for finding this!


DarkLight1337 (approved):

DarkLight1337:
Should we add a test for this?

njhill:
@DarkLight1337 I was thinking about a test but a basic test wouldn't have caught this particular bug... it happened to apply only to these specific 3-4 args if they were specified in a standalone way.

Issue #27815, opened 2025-10-30
[Bug]: torch.compile startup time logging not correct


zou3519:
### Your current environment

n/a

### 🐛 Describe the bug

@mgoin mentioned on slack:

Looking at some of the slow tests in CI, I’ve found that there seems to be a large discrepancy with how much time torch.compile says it takes versus how long it actually takes

I’ve attached a section from this test for quantization/test_torchao.py::test_qwenvl_int8wo_model_loading_with_params https://buildkite.com/organizations/vllm/pipelines/ci/builds/35163/jobs/0199eb2d-ccf5-462f-9c44-834b47fbac0f/log#203-4688

If we zoom in, it say compiling the dynamic graph takes 41s and in total takes 47s, but in the time between those two logs, it is clear that 8 minutes have passed!
```
[2025-10-16T06:08:44Z] (EngineCore_DP0 pid=23579) INFO 10-15 23:08:44 [backends.py:247] Cache the graph for dynamic shape for later use
[2025-10-16T06:09:23Z] (EngineCore_DP0 pid=23579) INFO 10-15 23:09:23 [backends.py:274] Compiling a graph for dynamic shape takes 41.40 s
[2025-10-16T06:17:21Z] (EngineCore_DP0 pid=23579) INFO 10-15 23:17:21 [monitor.py:33] torch.compile takes 47.72 s in total
```
Do we have any ideas why this might be happening?

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

vnadathur:
> https://buildkite.com/organizations/vllm/pipelines/ci/builds/35163/jobs/0199eb2d-ccf5-462f-9c44-834b47fbac0f/log#203-4688

Doesn't seem like link is working.

This is just my theory but:
> Do we have any ideas why this might be happening?

I think the 8 minutes is the time after compilation has finished and before the first execution of the last piecewise-compiled graph.

```
[2025-10-16T06:17:21Z] (EngineCore_DP0 pid=23579) INFO 10-15 23:17:21 [monitor.py:33] torch.compile takes 47.72 s in total

```

[...]

PR #28519, opened 2025-11-12
[bugfix] [torch.compile] startup time logging not correct

ghost:
As per this issue: https://github.com/vllm-project/vllm/issues/27815

The torch.compile total time is not accurately reflecting the real time taken. 
> If we zoom in, it say compiling the dynamic graph takes 41s and in total takes 47s, but in the time between those two logs, it is clear that 8 minutes have passed!
> ```
> [2025-10-16T06:08:44Z] (EngineCore_DP0 pid=23579) INFO 10-15 23:08:44 [backends.py:247] Cache the graph for dynamic shape for later use
> [2025-10-16T06:09:23Z] (EngineCore_DP0 pid=23579) INFO 10-15 23:09:23 [backends.py:274] Compiling a graph for dynamic shape takes 41.40 s
> [2025-10-16T06:17:21Z] (EngineCore_DP0 pid=23579) INFO 10-15 23:17:21 [monitor.py:33] torch.compile takes 47.72 s in total

Just for visibility I am copy-pasting my original explanation into this description:

I think the 8 minutes is the time after compilation has finished and before the first execution of the last piecewise-compiled graph.

```
[2025-10-16T06:17:21Z] (EngineCore_DP0 pid=23579) INFO 10-15 23:17:21 [monitor.py:33] torch.compile takes 47.72 s in total

```

[This](https://github.com/vllm-project/vllm/blob/main/vllm/compilation/monitor.py#L31-L38) is logged when the last piecewise graph is [first run](https://github.com/vllm-project/vllm/blob/main/vllm/compilation/piecewise_backend.py#L82C1-L93C64) but not when [compilation is completed.](https://github.com/vllm-project/vllm/blob/main/vllm/compilation/backends.py#L273-L282)

Proposed change:
Gather the total time spent on torch.compile compilation to enforce clarity and visibility.

### DIFF
--- a/vllm/compilation/monitor.py
+++ b/vllm/compilation/monitor.py
@@ -32,7 +32,6 @@ def end_monitoring_torch_compile(vllm_config: VllmConfig):
     compilation_config: CompilationConfig = vllm_config.compilation_config
     if compilation_config.mode == CompilationMode.VLLM_COMPILE:
         total_elapsed = time.time() - torch_compile_start_time
-        compilation_config.compilation_time = total_elapsed
         logger.info_once(
             "torch.compile total time: %.2f s",
             total_elapsed,
--- a/vllm/compilation/monitor.py
+++ b/vllm/compilation/monitor.py
@@ -31,9 +31,11 @@ def start_monitoring_torch_compile(vllm_config: VllmConfig):
 def end_monitoring_torch_compile(vllm_config: VllmConfig):
     compilation_config: CompilationConfig = vllm_config.compilation_config
     if compilation_config.mode == CompilationMode.VLLM_COMPILE:
+        total_elapsed = time.time() - torch_compile_start_time
+        compilation_config.compilation_time = total_elapsed
         logger.info_once(
-            "torch.compile takes %.2f s in total",
-            compilation_config.compilation_time,
+            "torch.compile total time: %.2f s",
+            total_elapsed,
             scope="local",
         )
         global context_manager


vllm/compilation/monitor.py
```diff
@@ -31,9 +31,11 @@ def start_monitoring_torch_compile(vllm_config: VllmConfig):
 def end_monitoring_torch_compile(vllm_config: VllmConfig):
     compilation_config: CompilationConfig = vllm_config.compilation_config
     if compilation_config.mode == CompilationMode.VLLM_COMPILE:
+        total_elapsed = time.time() - torch_compile_start_time
+        compilation_config.compilation_time = total_elapsed
```
elizabetht: Is `compilation_config.compilation_time` referenced anywhere else? I don’t see any usages. If it isn’t used, can we drop the assignment (`compilation_config.compilation_time = total_elapsed`)?

elizabetht (approved):

vllm/compilation/monitor.py
```diff
@@ -31,9 +31,10 @@ def start_monitoring_torch_compile(vllm_config: VllmConfig):
 def end_monitoring_torch_compile(vllm_config: VllmConfig):
     compilation_config: CompilationConfig = vllm_config.compilation_config
     if compilation_config.mode == CompilationMode.VLLM_COMPILE:
+        total_elapsed = time.time() - torch_compile_start_time
         logger.info_once(
-            "torch.compile takes %.2f s in total",
-            compilation_config.compilation_time,
+            "torch.compile total time: %.2f s",
+            total_elapsed,
```
heheda12345, line 37: what's the difference between these two times? Why compilation_config.compilation_time doesn't work?

ghost: @heheda12345 compilation_time only took the time of: dynamo bytecode transform + per‑graph compile/cache load. it ignored any work or delay outside those two points so it under‑reported how long torch.compile actually took. `total_elapsed = time.time() - torch_compile_start_time` measures the full compilation time.

heheda12345: @zou3519 @ProExpertProg can you help to check whether we also need to update `compilation_config.compilation_time`?

ProExpertProg (changes requested):
I worry that we're now measuring non-torch.compile items as well. Can you get a CPU profile of startup and we can compare that to the logs, and see what we're really measuring?

PR #28522, opened 2025-11-12
[feat]: log number of preempted requests

610lyn:
Summary: Add tracking and logging for the number of preempted requests in the vLLM v1 metrics logger. This helps monitor system behavior under load.

Test Plan: Run vLLM v1 engine and verify "Preemptions: X" appears in periodic log output.

Differential Revision: D86835924

### DIFF
--- a/vllm/v1/metrics/loggers.py
+++ b/vllm/v1/metrics/loggers.py
@@ -118,12 +118,14 @@ class LoggingStatLogger(StatLoggerBase):
         self.num_prompt_tokens: int = 0
         self.num_generation_tokens: int = 0
         self.num_corrupted_reqs: int = 0
+        self.num_preemptions: int = 0
 
     def _track_iteration_stats(self, iteration_stats: IterationStats):
         # Save tracked stats for token counters.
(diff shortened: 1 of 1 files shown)


vllm/v1/metrics/loggers.py
```diff
@@ -198,6 +200,7 @@ def log(self):
             "Waiting: %d reqs",
             "GPU KV cache usage: %.1f%%",
             "Prefix cache hit rate: %.1f%%",
+            "Preemptions: %d",
```
markmc: Let's move it up, so it comes after waiting. Also, let's not omit it when it's zero, since preemptions should be uncommon

610lyn: Thanks for reviewing. Updated.

markmc: Please add a check so that we only log the number of preemptions when it is non-zero

610lyn: Added zero preemptions check.

markmc (approved):

PR #28524, opened 2025-11-12
Fix Llama4 Pipeline Parallelism

River12:
## Purpose

Fix Llama4 Pipeline Parallelism Assert Error

## Test Plan

from vllm import LLM
llm = LLM("meta-llama/Llama-4-Scout-17B-16E-Instruct", pipeline_parallel_size=4)
output= llm.generate("Hi, vLLM is a")

## Test Result
The llama4 model can be loaded successfully, and generate correct response. 

## details
When enabling PP for Llama4 model, each rank has a few `PPMissingLayer()`. The line `assert isinstance(layer, Llama4DecoderLayer)` raises assert error when encountering `PPMissingLayer()`. This PR fixed this bug by using if condition to check the layer is Llama4DecoderLayer.

### DIFF
--- a/vllm/model_executor/models/llama4.py
+++ b/vllm/model_executor/models/llama4.py
@@ -738,8 +738,9 @@ class Llama4ForCausalLM(LlamaForCausalLM, MixtureOfExperts):
         self.moe_layers = []
         example_moe = None
         for layer in self.model.layers:
-            assert isinstance(layer, Llama4DecoderLayer)
-            if isinstance(layer.feed_forward, Llama4MoE):
+            if isinstance(layer, Llama4DecoderLayer) and isinstance(
+                layer.feed_forward, Llama4MoE
+            ):
                 # Pick last one layer since the first ones may be dense layers.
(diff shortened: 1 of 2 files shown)


vllm/model_executor/models/llama4.py
```diff
@@ -738,8 +738,9 @@ def set_moe_parameters(self):
         self.moe_layers = []
         example_moe = None
         for layer in self.model.layers:
-            assert isinstance(layer, Llama4DecoderLayer)
-            if isinstance(layer.feed_forward, Llama4MoE):
+            if isinstance(layer, Llama4DecoderLayer) and isinstance(
```
heheda12345, line 741: what about `if isinstance(layer, PPMissingLayer) continue`? I think most models implement pp missing layer skipping in this way

River12: Thanks for the comment. It has been changed to use `if isinstance(layer, PPMissingLayer) continue` in #28577.

Issue #28476, opened 2025-11-11
[Bug]: [DCP] [DSV3] AssertionError on long context when --decode-context-parallel-size is enabled


cjackal:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
==============================
        System Info
==============================
OS                           : Ubuntu 24.04.3 LTS (x86_64)
GCC version                  : (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
Clang version                : Could not collect
CMake version                : Could not collect
Libc version                 : glibc-2.39

==============================
       PyTorch Info
==============================
PyTorch version              : 2.9.0+cu129
Is debug build               : False
CUDA used to build PyTorch   : 12.9
ROCM used to build PyTorch   : N/A

==============================
      Python Environment
==============================
Python version               : 3.12.3 (main, Aug 14 2025, 17:47:21) [GCC 13.3.0] (64-bit runtime)
Python platform              : Linux-5.14.0-284.118.1.el9_2.x86_64-x86_64-with-glibc2.39

==============================
       CUDA / GPU Info
==============================
Is CUDA available            : True
CUDA runtime version         : 12.9.86
CUDA_MODULE_LOADING set to   :
GPU models and configuration :
GPU 0: NVIDIA H100 80GB HBM3
GPU 1: NVIDIA H100 80GB HBM3
GPU 2: NVIDIA H100 80GB HBM3
GPU 3: NVIDIA H100 80GB HBM3
GPU 4: NVIDIA H100 80GB HBM3
GPU 5: NVIDIA H100 80GB HBM3
GPU 6: NVIDIA H100 80GB HBM3
GPU 7: NVIDIA H100 80GB HBM3

Nvidia driver version        : 570.148.08
cuDNN version                : Could not collect
HIP runtime version          : N/A
MIOpen runtime version       : N/A
Is XNNPACK available         : True

==============================
          CPU Info
==============================
Architecture:                         x86_64
CPU op-mode(s):                       32-bit, 64-bit
Address sizes:                        52 bits physical, 57 bits virtual
Byte Order:                           Little Endian
CPU(s):                               64
On-line CPU(s) list:                  0-63
Vendor ID:                            GenuineIntel
Model name:                           Intel(R) Xeon(R) Gold 6430
CPU family:                           6
Model:                                143
Thread(s) per core:                   1
Core(s) per socket:                   32
Socket(s):                            2
Stepping:                             8
[...]

Issue #28411, opened 2025-11-10
[Bug]:   assert reorganized_kv_c_normed.shape[0] == sum_seq_len


Nemo-G:
### Your current environment

```text

  GPU: 8x NVIDIA H20, 97GB VRAM each

  CPU: Intel Xeon Platinum 8457C, 180 cores total (45 cores × 2 threads)

  Serving model inside docker.
```

### 🐛 Describe the bug

```dockerfile
  kimi-k2-thinking:
    image: vllm/vllm-openai:nightly-e5e9067e61600eedd4e75bd1c512ec52872916aa
    container_name: kimi-k2-thinking
    restart: unless-stopped
    network_mode: host
    ipc: host
    privileged: true
    shm_size: 256g
    entrypoint: ["vllm",
      "serve", "/data00/Kimi-K2-Thinking",
      "--gpu-memory-utilization", "0.9",
      "--tensor-parallel-size", "8",
      "--decode-context-parallel-size", "8",
      # For compatiblity test
      "--served-model-name", "Kimi-K2-Thinking",
      "--enable-auto-tool-choice",
      "--tool-call-parser", "kimi_k2",
      #"--reasoning-parser", "kimi_k2",
      "--trust-remote-code",
      "--max-num-seq", "16",
      "--max-num-batched-tokens", "32768",
      #"--kv-cache-dtype", "fp8",

      "--swap-space", "64",
      "--disable-custom-all-reduce"
    ]
    environment:
        - CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7
        #- PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True
        #- OMP_NUM_THREADS=90
        #- MKL_NUM_THREADS=90
        #- TORCH_NUM_THREADS=90
    security_opt:
      - seccomp:unconfined
    stdin_open: true
    tty: true
    cap_add:
      - ALL
    ulimits:
      memlock: -1
      stack: 67108864
    deploy:
      resources:
        reservations:
          devices:
            - driver: nvidia
              count: all
              capabilities: [gpu]
    volumes:
      - /data00/:/data00/
      - /var/run/nvidia-topologyd/:/var/run/nvidia-topologyd/
``` 

There is no problem running a single agent loop. But when performing benchmarking it has assertion error

```
(Worker_TP6 pid=410) ERROR 11-10 09:28:06 [multiproc_executor.py:718]     model_output = self._model_forward(
[...]

PR #28526, opened 2025-11-12
[bugfix] correct local_chunk_len for DCP in reorg_kvcache with long context

pisceskkk:
Fix the issues https://github.com/vllm-project/vllm/issues/28476 https://github.com/vllm-project/vllm/issues/28411
The previous DCP implementation did not account for cases where long contexts were split to multi-chunks. This has now been addressed by updating the correct chunked local_chunk_len based on the chunk_size and local_context_len.

CC @LucasWilkinson @cjackal @Nemo-G


LucasWilkinson (approved):
LGTM, good catch; thanks for ascii art, helps alot!

mgoin (approved):

Issue #24915, opened 2025-09-15
[Bug]: Consistent vLLM crash when trying to reuse previous torch compilation with DeepSeek R1 on B200


smarterclayton:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code> for the prefill worker</summary>

```text
Collecting environment information...
==============================
        System Info
==============================
OS                           : Ubuntu 22.04.5 LTS (x86_64)
GCC version                  : (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0
Clang version                : 14.0.0-1ubuntu1.1
CMake version                : version 4.1.0
Libc version                 : glibc-2.35

==============================
       PyTorch Info
==============================
PyTorch version              : 2.8.0+cu128
Is debug build               : False
CUDA used to build PyTorch   : 12.8
ROCM used to build PyTorch   : N/A

==============================
      Python Environment
==============================
Python version               : 3.12.11 (main, Jun  4 2025, 08:56:18) [GCC 11.4.0] (64-bit runtime)
Python platform              : Linux-6.6.93+-x86_64-with-glibc2.35

==============================
       CUDA / GPU Info
==============================
Is CUDA available            : True
CUDA runtime version         : 12.8.93
CUDA_MODULE_LOADING set to   : LAZY
GPU models and configuration : 
GPU 0: NVIDIA B200
GPU 1: NVIDIA B200
GPU 2: NVIDIA B200
GPU 3: NVIDIA B200
GPU 4: NVIDIA B200
GPU 5: NVIDIA B200
GPU 6: NVIDIA B200
GPU 7: NVIDIA B200

Nvidia driver version        : 575.57.08
cuDNN version                : Could not collect
HIP runtime version          : N/A
MIOpen runtime version       : N/A
Is XNNPACK available         : True

==============================
          CPU Info
==============================
Architecture:                            x86_64
CPU op-mode(s):                          32-bit, 64-bit
Address sizes:                           52 bits physical, 57 bits virtual
Byte Order:                              Little Endian
CPU(s):                                  224
On-line CPU(s) list:                     0-223
Vendor ID:                               GenuineIntel
Model name:                              INTEL(R) XEON(R) PLATINUM 8581C CPU @ 2.10GHz
CPU family:                              6
Model:                                   207
Thread(s) per core:                      2
Core(s) per socket:                      56
Socket(s):                               2
Stepping:                                2
BogoMIPS:                                4200.00
[...]

PR #28533, opened 2025-11-12
[Bugfix] Eliminate tuple inputs to submodules in graph partitioning

gmagogsfm:
Move `getitem` operations to the same subgraph as their input to prevent submodules from receiving tuple arguments. 

This is to fix issues like [#24915](https://github.com/vllm-project/vllm/issues/24915) where `split_graph` partitioned graph at a node that produces a tuple, which is then passed to subsequent submodules as input. This unfortunately does not conform to the calling convention of AoT Autograd modules, which requires flattened arguments.

Signed-off-by: Yanan Cao <gmagogsfm@gmail.com>

### DIFF
--- a/tests/compile/test_graph_partition.py
+++ b/tests/compile/test_graph_partition.py
@@ -17,6 +17,7 @@ def test_getitem_moved_to_producer_subgraph():
 
     class TupleOutputModule(torch.nn.Module):
         """Tuple producer."""
+
         def forward(
             self, x: torch.Tensor, y: torch.Tensor
         ) -> tuple[torch.Tensor, torch.Tensor]:
@@ -82,6 +83,7 @@ def test_no_tuple_inputs_with_multiple_consumers():
 
     class TupleOutputModule(torch.nn.Module):
         """Tuple producer."""
+
         def forward(
             self, x: torch.Tensor, y: torch.Tensor
         ) -> tuple[torch.Tensor, torch.Tensor]:
@@ -91,6 +93,7 @@ def test_no_tuple_inputs_with_multiple_consumers():
 
     class MultiConsumerModel(torch.nn.Module):
         """Tuple (multi-)consumer."""
+
         def __init__(self):
             super().__init__()
             self.submod = TupleOutputModule()
--- a/tests/compile/test_graph_partition.py
+++ b/tests/compile/test_graph_partition.py
@@ -5,6 +5,7 @@ import operator
 
 import pytest
 import torch
+from torch.fx.experimental.proxy_tensor import make_fx
 
 from vllm.compilation.backends import split_graph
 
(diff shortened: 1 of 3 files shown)


vllm/compilation/backends.py
```diff
@@ -307,12 +308,25 @@
...
+        # If so, assign it to the same subgraph as its input to avoid passing entire
+        # tuple as input to submodules, which is against AoTAutograd input requirement.
+        if (
+            node.op == "call_function"
+            and node.target == operator.getitem
+        ):
+            # Assign this getitem to the same subgraph as its input
+            input_node = node.args[0]
+            if input_node in node_to_subgraph_id:
+                node_to_subgraph_id[node] = node_to_subgraph_id[input_node]
```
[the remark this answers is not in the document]
gmagogsfm: In this context, input_node is guaranteed to be an `fx.Node`, which is hashable.

vllm/compilation/backends.py
```diff
@@ -313,6 +314,16 @@ def split_graph(
...
 
+        # Check if this is a getitem operation on a node from an earlier subgraph.
+        # If so, assign it to the same subgraph as its input to avoid passing entire
+        # tuple as input to submodules, which is against AoTAutograd input requirement.
+        if node.op == "call_function" and node.target == operator.getitem:
+            # Assign this getitem to the same subgraph as its input
+            input_node = node.args[0]
+            if input_node in node_to_subgraph_id:
+                node_to_subgraph_id[node] = node_to_subgraph_id[input_node]
+                continue
```
ProExpertProg: This should always be true, right?
```suggestion
            assert input_node in node_to_subgraph_id
            node_to_subgraph_id[node] = node_to_subgraph_id[input_node]
            continue
```

gmagogsfm: I think `placeholder` nodes would not be in `node_to_subgraph_id` map.

tests/compile/test_graph_partition.py
```diff
@@ -0,0 +1,134 @@
...
+            b = torch.nn.functional.relu(t[1])
+            return a, b
+
+    model = ConsumeTupleElementsModule()
+    gm = torch.fx.symbolic_trace(model)
+
+    split_ops = [torch.ops.aten.relu.default]
+    split_gm, split_items = split_graph(gm, split_ops)
+
+    assert len(split_items) > 0, "Graph should be split into submodules"
```
zou3519: assert equal to 3?

tests/compile/test_graph_partition.py
```diff
@@ -0,0 +1,134 @@
...
+                and node.target == operator.getitem
+                and node.args[0].op == "placeholder"
+            ):
+                getitem_on_placeholder.append(node)
+
+        assert len(getitem_on_placeholder) == 0, (
+            f"Submodule {split_item.submod_name} has getitem operations on "
+            f"placeholder nodes: {[n.name for n in getitem_on_placeholder]}. "
+            "This means tuple inputs were not properly eliminated."
+        )
```
zou3519, line 59: I'm confused, there were no getitem nodes in `gm` already, so I don't think this test is testing the right thing:

image

tests/compile/test_graph_partition.py
```diff
@@ -0,0 +1,134 @@
...
+        def forward(self, x: torch.Tensor) -> torch.Tensor:
+            t = self.submod(x + 1, x + 2)
+            r1 = torch.nn.functional.sigmoid(t[0])
+            r2 = torch.nn.functional.tanh(t[1])
+            # Add relu so the last subgraph has real computation
+            r2_relu = torch.nn.functional.relu(r2)
+            return r1 + r2_relu
+
+    model = MultiConsumerModel()
+    gm = torch.fx.symbolic_trace(model)
```
zou3519: ditto here: I don't see any getitem calls or tuple proxies in this graph

tests/compile/test_graph_partition.py
gmagogsfm: Rewrote the test, it should be testing the behavior correctly now. PTAL

tests/compile/test_graph_partition.py
gmagogsfm: updated, see other comment thread

tests/compile/test_graph_partition.py
gmagogsfm: updated, see other comment thread

vllm/compilation/backends.py
```diff
@@ -307,12 +308,22 @@ def split_graph(
...
-    split_op_graphs = []
+    node_to_subgraph_id: dict[fx.Node, int] = {}
+    split_op_graphs: list[int] = []
     for node in graph.graph.nodes:
         if node.op in ("output", "placeholder"):
             continue
 
+        # Check if this is a getitem operation on a node from an earlier subgraph.
+        # If so, assign it to the same subgraph as its input to avoid passing entire
+        # tuple as input to submodules, which is against AoTAutograd input requirement.
```
zou3519: nit: "which is against standalone_compile and AOTAutograd input requirement"

gmagogsfm: updated

vllm/compilation/backends.py
zou3519: hmm I think the claim is that we should never be getitem on a placeholder node or an output node so we can just assert that the input_node is in node_to_subgraph id?

Because Dynamo produces a graph where all placeholders should be Tensors or symints or scriptobjects

So I agree with Luka's suggestion

zou3519 (commented):
code reasonable, I think Luka's comment makes sense

vllm/compilation/backends.py
gmagogsfm: I see, changed.

zou3519 (approved):

vllm/compilation/backends.py
gmagogsfm: @zou3519  @ProExpertProg I had to relax the assert to only when producer node is not a placeholder, because `getitem` is also used to fetch item/slice from a tensor as well. In that case, there can be legitimate `getitem` calls on placeholders. This is seen in qwen2. 

PTAL.

zou3519: oh no, this is dynamo IR we're talking about... let me think about this

zou3519: Do we have a good way of distinguishing between getitem calls on Tensors and getitem calls on Tuples?

gmagogsfm: Checking type of node.meta['val'] is the way. However I don't think there is a strong guarantee about its presence. 

I think the current assert is strict and correct because it is by design that all `placeholder` nodes are not in `node_to_subgraph_id`, and this check asserts exactly that.

gmagogsfm: OK, I tested with `node.meta['val']`. It is not always available, so we can't rely on it to tell us if something is a tuple or tensor.

The next best thing we can do is what is currently implemented, i.e. checking `input_node.op == "placeholder"` and relying on Dynamo doing the right thing, which is currently the case according to CI results.

zou3519: I am fine with loosening the assert. I ran into this a while ago with something in standard torch.compile

PR #24794, opened 2025-09-13
[Attention] Refactor CUDA attention backend selection logic

MatthewBonanni:
## Purpose
`CudaPlatformBase.get_attention_backend_cls` has gotten complex and messy over time. This PR cleans up the logic (without changing the behavior) and standardizes the interface.

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #28534, opened 2025-11-12
[CI Failure] Fix backend selection for encoder-only models

hl475:
## Purpose

After #24794, encoder-only models (e.g., BERT) fail to initialize because the TRITON_ATTN backend is selected by default, but it doesn't support encoder self-attention, causing:
```
NotImplementedError: Encoder self-attention and encoder/decoder cross-attention are not implemented for TritonAttentionImpl
```

This PR implemented an opt-in approach for attention type support:

  1. Added `supports_attn_type()` method to `AttentionBackend`:
    - Default behavior: Only supports `DECODER` attention
    - Backends must explicitly override to support `ENCODER_ONLY` or other attention types
    - This makes the system safe by default - new backends won't accidentally support encoder-only models
  2. Propagated `attn_type` through the backend selection pipeline:
    - Added `attn_type` parameter to `get_attn_backend()` and `validate_configuration()`
    - Modified `EncoderOnlyAttention` to pass `attn_type=AttentionType.ENCODER_ONLY`
    - Platform classes now validate attention type compatibility during backend selection
  3. Explicitly marked the 3 backends that support encoder-only models:
    - FlexAttention: Supports DECODER + ENCODER_ONLY
    - FlashAttention: Supports DECODER + ENCODER_ONLY
    - CPU/TorchSDPA: Supports all attention types

## Test Plan
```
pytest -s -v tests/models/language/pooling/test_token_classification.py::test_bert_models[float-boltuix/NeuroBERT-NER]
```

## Test Result
```
1 passed, 4 warnings in 18.16s
```
---
<details>

</details>

### DIFF
--- a/vllm/attention/backends/abstract.py
+++ b/vllm/attention/backends/abstract.py
@@ -143,6 +143,17 @@ class AttentionBackend(ABC):
         return False
 
     @classmethod
+    def supports_attn_type(cls, attn_type: str) -> bool:
+        """Check if backend supports a given attention type.
+
+        By default, only supports decoder attention.
+        Backends should override this to support other attention types.
+        """
+        from vllm.attention import AttentionType
+
+        return attn_type == AttentionType.DECODER
+
+    @classmethod
     def supports_compute_capability(cls, capability: "DeviceCapability") -> bool:
         return True
 
@@ -171,6 +182,7 @@ class AttentionBackend(ABC):
         has_sink: bool,
         use_sparse: bool,
         device_capability: "DeviceCapability",
+        attn_type: str,
     ) -> list[str]:
         invalid_reasons = []
         if not cls.supports_head_size(head_size):
@@ -195,6 +207,8 @@ class AttentionBackend(ABC):
                 invalid_reasons.append("non-sparse not supported")
         if not cls.supports_compute_capability(device_capability):
             invalid_reasons.append("compute capability not supported")
+        if not cls.supports_attn_type(attn_type):
+            invalid_reasons.append(f"attention type {attn_type} not supported")
         combination_reason = cls.supports_combination(
             head_size,
             dtype,
--- a/vllm/v1/attention/backends/cpu_attn.py
+++ b/vllm/v1/attention/backends/cpu_attn.py
@@ -48,6 +48,17 @@ class CPUAttentionBackend(AttentionBackend):
     def get_name() -> str:
         return "CPU_ATTN"
 
+    @classmethod
+    def supports_attn_type(cls, attn_type: str) -> bool:
+        """CPU attention supports decoder and encoder-only attention."""
+        from vllm.attention import AttentionType
+
+        return attn_type in (
+            AttentionType.DECODER,
+            AttentionType.ENCODER,
+            AttentionType.ENCODER_ONLY,
+        )
+
     @staticmethod
     def get_impl_cls() -> type["CPUAttentionBackendImpl"]:
         return CPUAttentionBackendImpl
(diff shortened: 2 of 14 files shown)


DarkLight1337:
cc @MatthewBonanni @LucasWilkinson

vllm/attention/backends/abstract.py
```diff
@@ -171,6 +180,7 @@ def validate_configuration(
         has_sink: bool,
         use_sparse: bool,
         device_capability: "DeviceCapability",
+        attn_type: str = AttentionType.DECODER,
```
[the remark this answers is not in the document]
MatthewBonanni: etc

vllm/attention/backends/abstract.py
```diff
@@ -195,6 +205,8 @@ def validate_configuration(
                 invalid_reasons.append("non-sparse not supported")
         if not cls.supports_compute_capability(device_capability):
             invalid_reasons.append("compute capability not supported")
+        if not cls.supports_attn_type(attn_type):
+            invalid_reasons.append(f"attention type {attn_type} not supported")
```
MatthewBonanni, line 211:
```suggestion
            invalid_reasons.append(f"attention type not supported")
```
For consistency with the others.
I'd prefer to log the argument values in one place in `get_attn_backend_cls` to keep the log output shorter

MatthewBonanni (approved):
Down the road I'd like to make `AttentionType` an enum, but this LGTM!

LucasWilkinson (approved):
Overall looks good; thanks for fixing this! please rebase

vllm/attention/backends/abstract.py
```diff
@@ -142,6 +142,15 @@ def supports_sink(cls) -> bool:
...
         return False
 
+    @classmethod
+    def supports_attn_type(cls, attn_type: str) -> bool:
+        """Check if backend supports a given attention type.
+
+        By default, returns True (all types supported).
+        Backends should override this to restrict supported types.
+        """
+        return True
```
mgoin: I actually think the default should be False or at least be DECODER only. I would rather attention backends be opt-in to supporting "non-standard" attention types

vllm/attention/backends/abstract.py
mgoin: Should we really have this default value?

mgoin:
I think we can probably ignore my comments for now, but we should consider them in followup. Probably Matt can tackle that if you don't have time

vllm/v1/attention/backends/cpu_attn.py
```diff
@@ -48,6 +48,11 @@ def get_supported_head_sizes(cls) -> list[int]:
     def get_name() -> str:
         return "CPU_ATTN"
 
+    @classmethod
+    def supports_attn_type(cls, attn_type: str) -> bool:
+        """CPU attention supports all attention types."""
+        return True
```
mgoin: Are you sure this is true?

mgoin (commented):
What are the attention backends that support running ENCODER and ENCODER_DECODER? I don't see them mentioned anywhere cc @russellb @NickLucche

vllm/v1/attention/backends/cpu_attn.py
hl475: thanks for pointing this out!

I think it should be
```
return attn_type in (
          AttentionType.DECODER,
          AttentionType.ENCODER,
          AttentionType.ENCODER_ONLY,
      )
```
these three types are the ones that v1 cpu backend supports

mgoin (approved):

PR #28535, opened 2025-11-12
[Hardware][PowerPC] Fix fp16 compilation error for Power in cpu attention backend and bump oneDNN version

Akashcodes732:
# Purpose

Fix Power compilation failure: PR [#27954](https://github.com/vllm-project/vllm/pull/27954)
 refactored the attention backend for CPU, but the build fails on Power due to the use of fp16, which is not supported on this architecture. This PR addresses that issue (also mentioned in my review on the original PR).

Update oneDNN to v3.10: The oneDNN PR [uxlfoundation/oneDNN#4002](https://github.com/uxlfoundation/oneDNN/pull/4002)
 — included in v3.10 — fixes a compilation failure on Power9 systems. Bumping the version here resolves that issue as well.

cc: @bigPYJ1151

### DIFF
--- a/cmake/cpu_extension.cmake
+++ b/cmake/cpu_extension.cmake
@@ -219,7 +219,7 @@ if ((AVX512_FOUND AND NOT AVX512_DISABLED) OR (ASIMD_FOUND AND NOT APPLE_SILICON
                 SUBBUILD_DIR "${FETCHCONTENT_BASE_DIR}/arm_compute-subbuild"
                 SOURCE_DIR   "${FETCHCONTENT_BASE_DIR}/arm_compute-src"
                 GIT_REPOSITORY https://github.com/ARM-software/ComputeLibrary.git
-                GIT_TAG        v52.2.0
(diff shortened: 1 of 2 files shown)


bigPYJ1151:
If fp16 is not supported on Power, should we add a check in ```check_and_update_config``` of ```cpu.py```?

fadara01 (changes requested):
Thanks for spotting this @bigPYJ1151 - current version of ACL is indeed not compatible with oneDNN v3.10 

@Akashcodes732 could you please update ACL version [here](https://github.com/vllm-project/vllm/blob/main/cmake/cpu_extension.cmake#L245) to `v52.6.0` which should be compatible with oneDNN v3.10

Akashcodes732:
Hi @bigPYJ1151 @fadara01 , 

I have updated the ACL version

bigPYJ1151 (approved):

fadara01 (approved):
LGTM

Issue #26108, opened 2025-10-02
[RFC]: MoE shared_experts fusion with the selected experts inside FusedMoE


alexm-redhat:
### Motivation.

Profile results for DeepSeekR1 show that a significant amount of decode-iteration time is spend in executing the shared_experts of MoE before the "selected_experts" FusedMoE part. To reduce the overhead of shared_experts, their execution can be fused with the selected_experts (by combining the weight matrices) and then executing a single FusedMoE part. 

### Proposed Change.

Here are the profile results for FP4 and FP8 (we can see operation breakdown for a single transformer decode-iteration) (flags used VLLM_USE_FLASHINFER_MOE_FP8=1 and VLLM_FLASHINFER_MOE_BACKEND="latency"):

**DeepSeekR1-FP4 8xB200 batch-size 32**
<img width="778" height="439" alt="Image" src="https://github.com/user-attachments/assets/fd1a62de-5ac1-42ac-bd9d-4d40cefaf9da" />

**DeepSeekR1-FP8 8xB200 batch-size 32**
<img width="776" height="440" alt="Image" src="https://github.com/user-attachments/assets/8cd0d237-b012-4015-9c1a-565a1e19174f" />

From this profile we can see that fusing shared_experts with selected_experts matrix multiplies will have the following benefits:
1. Removes data converts FP16=>FP4/FP8 that we currently do in shared experts, since we will do them in the FusedMoE part anyway.
2. The silu op in shared experts is composed of 2 elementwise ops, and fusion will remove it completely. 
3. Willl allow to execute the shared_experts inside the highly optimized FusedMoE part (not using the DeepseekV2MLP class from deepseek_v2.py) 

Notes/TODOs:
1. In DP, the shared_experts part is overlapped with the communication, so we do want to keep shared_experts separate from selected_experts in DP mode. The benefit of fusion is more relevant for the TP case.
2. There is a PR that attempted to do this fusion https://github.com/vllm-project/vllm/pull/15502 however, it did not land due to quantization op issues. Maybe it is worth to look at it again and ressurect. The improvement that was mentioned there is 34% (most likely the real improvement is about 15%)
3. FusedMoE kernel has overheads in pre and post ops, which are part of the TRTLLM flashinfer codebase (in this specific execution). We should consider ways to reduce these overheads.

### Before submitting a new issue...

[...]

PR #24097, opened 2025-09-02
[ROCm][FEAT] Fuse DeepSeek shared experts into AITER fused_moe ops

kliuae:
## Purpose
This PR targets ROCm AITER, introducing a flag‑gated path that fuses DeepSeek models’ shared_experts into the AITER's FusedMoE kernel, reducing separate MLP and addition overhead while preserving numeric behavior.

When shared experts fusion is enabled, the shared experts are viewed as synthetic routed experts after the original routed experts and receive allocated top‑k slots through grouped_topk, enabling a single fused MoE dispatch for both shared and routed experts.

This feature can be controlled by the environment flag `VLLM_ROCM_USE_AITER_FUSION_SHARED_EXPERTS` which is only effective when `VLLM_ROCM_USE_AITER_MOE` is set.

## Test Plan
The following tests validate DeepSeek models by collecting benchmark metrics and performning correctness tests through lm_eval.

vLLM server launch command:
```
# Toggle VLLM_ROCM_USE_AITER_FUSION_SHARED_EXPERTS

VLLM_USE_V1=1 \
VLLM_ROCM_USE_AITER=1 \
VLLM_ROCM_USE_AITER_MOE=1 \
VLLM_ROCM_USE_AITER_FUSION_SHARED_EXPERTS=${FUSE_SHARED_EXPERTS} \
VLLM_DISABLE_COMPILE_CACHE=1 \
vllm serve ${model_name} --tensor-parallel-size ${tp_size} --block-size 1 --compilation-config '{"cuadgraph_mode": "FULL_AND_PIECEWISE"}'
```

Benchmark commands:
```
# sharegpt dataset
vllm bench serve --model ${model_name} --dataset-name sharegpt --dataset-path ShareGPT_V3_unfiltered_cleaned_split.json --percentile-metrics ttft,tpot,itl,e2el --request-rate ${request_rate}

# random dataset
vllm bench serve --model ${model_name} --percentile-metrics ttft,tpot,itl,e2el --request-rate 10 --num-prompts 100 --dataset-name random --random-input-len 1024 --random-output-len 1024
```

lm_eval command:
```
lm_eval --model local-completions --tasks gsm8k --model_args model=${model_name},base_url=http://localhost:8000/v1/completions,num_concurrent=128,max_retries=3,tokenized_requests=False
```

### **Benchmark results**

**deepseek-ai/DeepSeek-R1 on sharegpt dataset**
[...]

PR #21107, opened 2025-07-17
Add `fused_moe_gate` kernel and integrate to DeepSeek MoE layer

zhangxy9999:
## Purpose
Fuses the DeepSeek-style grouped top-k experts selection process (3 top-k ops and bells and whistles) into one custom kernel.

The kernel is imported from sglang https://github.com/sgl-project/sglang/pull/4530

Reproduced kernel bench numbers on H200s:
```
> python3 benchmarks/kernels/benchmark_moe_fused_gate.py
...
...
moe-fused-gate-performance:
   seq_length     Original  SGL Kernel
0      5000.0   208.880000   25.823999
1     10000.0   400.303990   35.872001
2     15000.0   537.952006   42.048000
3     20000.0   718.591988   52.607998
4     25000.0   898.847997   57.824001
5     30000.0  1035.967946   67.744002
6     35000.0  1215.871990   77.215999
7     40000.0  1354.143977   83.360001
```
## Test Plan

Unit tests passing

## (Optional) Documentation Update

PR #15502, opened 2025-03-25
Adding Share Expert Fusion for DeepSeek

DiegoD94:
This PR is to submit a feature enable fuse shared expert into MoE module for DeepSeek V2/3 models.
Design doc: https://docs.google.com/document/d/1iXgzR6Mt6s0DpT7w2Pz93ExlUJ-nnSnU_o9Sqd8TE34/edit?tab=t.0

# TL;DR:
We have up to 34% ITL lossless improvement and 12% TTLT lossless improvement by enabling this feature. With FlashAttention3, without MLA and without V1 Engine.

```
vllm: 0.8.3,
LLMPerf: https://github.com/ray-project/llmperf, 
LM-Eval-Harness: https://github.com/EleutherAI/lm-evaluation-harness 0.4.0
```
Command to enable this feature:
```
export VLLM_USE_V1=0; export VLLM_MLA_DISABLE=1; export VLLM_ENABLE_SHARE_EXPERT_FUSION=1;python3 -m vllm.entrypoints.openai.api_server --model /home/ubuntu/workspace/ckpt/DeepSeek-R1 --tensor-parallel-size 8 --trust-remote-code --max-model-len 16384
```

# Details:
## Implementation:
1. Currently we used a ENV Variable (VLLM_ENABLE_SHARE_EXPERT_FUSION) as the toggle to enable/disable the feature. Later we can edit it as a engine args.
2. When this feature enabled, we internally set the expert number =264 and topk = 9 for FusedMoE module, and clone the shared expert weight into experts No.256 to No.263 during weight loading stage. Here the CUDA memory usage increase a bit (about 1-2 GB per GPU, assuming TP=8)
3. Top 9th routed expert id and expert weights are manually assigned so that expert are balanced and accuracy are guaranteed.
### Latency:
The 117% TTLT improvement might be due to a stuck prefill of baseline. I excluded it for conclusion.
```
| BatchSize | ContextLen | Baseline(0.8.3)                         | Ours (This PR)                         | Improvement          |
|-----------|------------|-----------------------------------------|----------------------------------------|----------------------|
|           |            | ITL_p50 | ITL_p90 | TTFT_p50 | TTFT_p90 | ITL_p50 | ITL_p90 | TTFT_p50 | TTFT_p90 | ITL    | TTFT   |
|-----------|------------|---------|---------|----------|----------|---------|---------|----------|----------|--------|--------|
[...]

PR #28540, opened 2025-11-12
[Performance] Fuse DeepSeek shared experts and gate operations

Red-Caesar:
### Purpose

This PR contains optimizations from two other PRs that we aligned with current main:
- Fused shared expert ([RFC#26108](https://github.com/vllm-project/vllm/issues/26108))
- Fused operation for gating

This effort is driven by **Nebius AI**, as part of ongoing optimization work.

Command to run the optimization version:
```
VLLM_USE_FUSED_MOE_GROUPED_TOPK=0 VLLM_USE_FUSED_MOE_ROUTER=1 VLLM_USE_CUDA_FUSION_SHARED_EXPERTS=1 python3 -m vllm.entrypoints.openai.api_server --model <deepseek-model> --tensor-parallel-size 8 --trust-remote-code
```
### Results
Our measurements show that these optimizations improve performance. Specifically, they give a significant boost to the TTFT metric. Results on the ITL metric are generally positive, with the observed slight performance drawdowns in rare configurations.
 
### Fused shared expert ([PR#15502](https://github.com/vllm-project/vllm/pull/15502)) 

This has already been implemented for AMD in by [PR#24097](https://github.com/vllm-project/vllm/pull/24097). Our PR makes it available to CUDA as well.

#### Description
Currently, we use an env variable (VLLM_USE_CUDA_FUSION_SHARED_EXPERTS) to enable or disable the feature. 
When this feature is enabled, we internally set 
```
expert number = config.n_routed_experts + config.n_shared_experts
topk = config.num_experts_per_tok + config.n_shared_experts
```
 for FusedMoE module. During the weight loading stage, we clone the shared expert weights into the experts (using the loading code from [PR#24097](https://github.com/vllm-project/vllm/pull/24097)). Shared expert id and expert weights are manually assigned to ensure that experts are balanced and accuracy is guaranteed.

### Fused operation for gating ([PR#21107](https://github.com/vllm-project/vllm/pull/21107))

This fuses the DeepSeek-style grouped top-k experts selection process (3 top-k ops and bells and whistles) into one custom kernel.

The kernel is imported from sglang [sgl-project/sglang#4530](https://github.com/sgl-project/sglang/pull/4530)

Currently we use an env variable (VLLM_USE_FUSED_MOE_ROUTER) to enable or disable the feature. 

### Notes
When these optimizations are enabled we can’t use the fused grouped_topk kernel for MoE ([#23274](https://github.com/vllm-project/vllm/pull/23274)) which is now the default.

### Tests
All tests have been performed on DeepSeek V3 on the h200.

#### gsm8k results
Original version:
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9181|±  |0.0076|
|     |       |strict-match    |     5|exact_match|↑  |0.8067|±  |0.0109|
```

Optimization version:
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9105|±  |0.0079|
|     |       |strict-match    |     5|exact_match|↑  |0.7976|±  |0.0111|
```

#### vLLM benchmark results 
We tests releases/v0.11.1 original vs releases/v0.11.1 + our optimization
image

```
python benchmarks/kernels/benchmark_moe_fused_gate.py 
```
```
moe-fused-gate-performance:
   seq_length    Original  SGL Kernel
0      5000.0   55.936001   27.936000
1     10000.0   61.535999   38.304001
2     15000.0   84.927998   45.600001
3     20000.0  104.704000   54.880001
4     25000.0  126.255997   61.503999
5     30000.0  148.095995   71.392000
6     35000.0  167.840004   81.568003
7     40000.0  187.215999   87.007999
```

### Our benchmark results
Our benchmark results you can find in the [doc](https://docs.google.com/document/d/1G0tBNvhQFDCGI51APkxITHjno5qXL9TbRjfiBJe8MmE/edit?usp=sharing).


vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1769,6 +1824,7 @@ def forward_impl(
                 expert_load_view=self.expert_load_view,
                 logical_to_physical_map=self.logical_to_physical_map,
                 logical_replica_count=self.logical_replica_count,
+                enable_fused_moe_router=self.enable_fused_moe_router,
```
bnellnm: Since `enable_fused_moe_router` is a `FusedMoE` layer attribute and `apply` gets the layer as a parameter, can we avoid passing it around to all the `apply` methods?

Red-Caesar: I've updated the way I pass the `enable_fused_moe_router` parameter - it's now an env variable only in the `select_experts` function. This has made the code more readable.

csrc/moe/moe_fused_gate.cu
```diff
@@ -0,0 +1,484 @@
...
+
+    if (num_fused_shared_experts > 1) {
+      for (int i = 1; i < num_fused_shared_experts; ++i) {
+        ++last_idx;
+        ++expert_offset;
+        indices_ptr[last_idx] =
+            static_cast<int32_t>(params.NUM_EXPERTS + expert_offset);
+        // Set the weight to the sum of all weights divided by
+        // routed_scaling_factor
+        output_ptr[last_idx] = output_sum / routed_scaling_factor;
```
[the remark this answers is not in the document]
Red-Caesar: I've aligned the calculation of `topk_excluding_share_expert_fusion` with [sglang](https://github.com/sgl-project/sglang/pull/6736), and now it takes into account `num_fused_shared_experts`. Everything else seems to be in line with the intended sglang logic.

csrc/moe/moe_fused_gate.cu
```diff
@@ -0,0 +1,484 @@
+#include <ATen/cuda/CUDAContext.h>
```
alexm-redhat, line 3: Would be good to add a comment that it is taken from sglang

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -477,6 +491,15 @@ def __init__(
                 self.global_num_experts,
                 get_compressed_expert_map(self.expert_map),
             )
+            if self.num_fused_shared_experts > 0:
+                logger.warning(
```
alexm-redhat, line 526: Does it work with EP? If not, then maybe we should assert here some conditions.

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1360,27 +1383,51 @@ def select_experts(
         elif use_grouped_topk:
             assert topk_group is not None
             assert num_expert_group is not None
-            if rocm_aiter_ops.is_fused_moe_enabled():
+            if hidden_states.shape[0] == 0:
```
alexm-redhat: Why this shape[0] == 0 case is necessary? Would be good to have some quick doc explaining this piece of code.

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1360,27 +1383,51 @@ def select_experts(
...
-                scoring_func=scoring_func,
-                routed_scaling_factor=routed_scaling_factor,
-                e_score_correction_bias=e_score_correction_bias,
-            )
+            if (
+                envs.VLLM_USE_FUSED_MOE_ROUTER
+                and e_score_correction_bias is not None
+                and is_power_of_two(e_score_correction_bias.shape[0])
+            ):
+                # The fused kernel can only work with 128/256 experts
```
alexm-redhat: can we add a check for E=128 or E=256 in the if statement as well?

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1360,27 +1383,51 @@ def select_experts(
...
+                    topk_group=topk_group,
+                    topk=top_k,
+                    num_fused_shared_experts=num_fused_shared_experts,
+                    routed_scaling_factor=routed_scaling_factor
+                    if routed_scaling_factor is not None
+                    else 1.0,
+                    apply_routed_scaling_factor_on_output=False,
+                )
+            else:
+                topk_weights, topk_ids = grouped_topk_impl(
```
alexm-redhat, line 1626: In which case the moe_fused_gate is faster than the usual grouped_topk_impl?

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1809,7 +1856,7 @@ def combine_output(states: torch.Tensor) -> torch.Tensor:
                     states = get_ep_group().combine(states, self.is_sequence_parallel)
                 return states
 
-            if self.shared_experts is not None:
+            if self.shared_experts is not None and self.num_fused_shared_experts == 0:
```
alexm-redhat: How is it possible to have self.shared_experts != None and self.num_fused_shared_experts > 0? It seems like if self.num_fused_shared_experts > 0 then elf.shared_experts must be None

vllm/model_executor/models/deepseek_v2.py
```diff
@@ -325,14 +337,15 @@ def __init__(
             # we do scaling outside, set factor to 1.0 to avoid double mul
             # aiter applies routed_scaling_factor internally
             routed_scaling_factor=1.0
-            if not self.is_rocm_aiter_moe_enabled
+            if not used_inside_scaling
```
alexm-redhat, line 332: Could you add some documentation about how "used_inside_scaling" changes the functionality of the scaling application

alexm-redhat (commented):
Thanks for the PR! I have verified its correctness and did some initial performance verifications - all looks good. Left some comments. The PR needs rebase and it should be ready to be merged.

csrc/moe/moe_fused_gate.cu
Red-Caesar: I've added

vllm/model_executor/layers/fused_moe/layer.py
Red-Caesar: Maybe this comment was a bit misleading, so I've deleted it. The kernel [works](https://github.com/sgl-project/sglang/blob/841eb29d3d97725eb91ec2c31ff4e46728a893e7/sgl-kernel/python/sgl_kernel/moe.py#L118) with routed_experts, which are powers of 2. So, if the shape of e_score_correction_bias is a power of two, I suppose it will also be true for the experts.

vllm/model_executor/layers/fused_moe/layer.py
Red-Caesar: Yeah, you're right, fix it

vllm/model_executor/layers/fused_moe/layer.py
Red-Caesar: At the moment, it's not faster on its own anymore, but it provides a boost when used together with fused_experts as the default grouped_topk isn't adapted to it

vllm/model_executor/layers/fused_moe/layer.py
alexm-redhat: I see makes sense

PR #28544, opened 2025-11-12
[CI] Add non-eager test-case for SharedStorageConnector

NickLucche:
I realized when we merged this fix https://github.com/vllm-project/vllm/pull/28253 we haven't actually added a test case to our suite (the referred issue required manual edits, hence it's not really tracked in CI).
This PR adds a test-case for that.


tests/v1/kv_connector/unit/test_shared_storage_connector.py
```diff
@@ -239,9 +256,34 @@ def test_shared_storage_connector_hashes(tmp_path):
         ),
     ]
 
-    # Run tests
+
+@pytest.mark.parametrize("eager", [False, True])
```
mgoin: Do we need to test both eager and non-eager? I think we can just test non-eager

NickLucche: agree @mgoin, I think I got caught up in reordering the test I lost sight of the objective lol

NickLucche:
Currently leading to OOM : `torch.OutOfMemoryError: CUDA out of memory`, will play around with test config

PR #28549, opened 2025-11-12
Add support for Eagle with separate lm-head and embed_tokens layers

eldarkurtic:
Some EAGLE3 drafters might have their own lm_head and/or embed_tokens layers. Existing codebase ignores this and always overrides them with target model's layers.

#### Test-case 1: for a model which needs copying of lm_head and embed_tokens from verifier model, the behavior should not change

Eval command to verify:

```bash
CUDA_VISIBLE_DEVICES=0,1 python examples/offline_inference/spec_decode.py \
  --method "eagle3" \
  --tp 2 \
  --model-dir "openai/gpt-oss-120b" \
  --eagle-dir "nvidia/gpt-oss-120b-Eagle3" \
  --dataset_name "hf" \
  --num-spec-tokens 3
```

Before this PR:
```bash
total_num_output_tokens: 241888
num_drafts: 105426
num_draft_tokens: 316278
num_accepted_tokens: 136330
mean acceptance length: 2.29
acceptance at token 0: 0.65
acceptance at token 1: 0.40
acceptance at token 2: 0.24
```

After this PR:
```
total_num_output_tokens: 241168
num_drafts: 105678
num_draft_tokens: 317034
num_accepted_tokens: 135362
mean acceptance length: 2.28
acceptance at token 0: 0.65
acceptance at token 1: 0.39
acceptance at token 2: 0.24
```

#### Test-case 2: for a model with has its own lm_head and embed_tokens, and therefore does not require copying from target's layers, acceptance rates look significantly better

Before this PR:
```bash
total_num_output_tokens: 247973
num_drafts: 187566
num_draft_tokens: 562698
num_accepted_tokens: 59726
mean acceptance length: 1.32
acceptance at token 0: 0.23
acceptance at token 1: 0.07
acceptance at token 2: 0.02
```

After this PR:
```bash
total_num_output_tokens: 247974
num_drafts: 99354
num_draft_tokens: 298062
num_accepted_tokens: 148677
mean acceptance length: 2.50
acceptance at token 0: 0.70
acceptance at token 1: 0.48
acceptance at token 2: 0.31
```

Note: idea for lm_head check inspired by https://github.com/vllm-project/vllm/pull/27688


vllm/model_executor/models/utils.py
```diff
@@ -824,3 +825,25 @@ def sequence_parallel_chunk_impl_fake(x: torch.Tensor) -> torch.Tensor:
...
+        name: The name of the weight to process
+    """
+    if not supports_any_eagle(model):
+        return
+
+    # To prevent overriding with target model's layers
+    if "lm_head" in name:
+        model.has_own_lm_head = True
+    if "embed_tokens" in name:
+        model.has_own_embed_tokens = True
```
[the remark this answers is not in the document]
eldarkurtic: this is the standard approach to check for these layers in the entire vllm codebase

vllm/model_executor/models/utils.py
```diff
@@ -824,3 +825,25 @@ def sequence_parallel_chunk_impl_fake(x: torch.Tensor) -> torch.Tensor:
     fake_impl=sequence_parallel_chunk_impl_fake,
     tags=(torch.Tag.needs_fixed_stride_order,),
 )
+
+
+def process_eagle_weight(
+    model: nn.Module,
```
NickLucche, line 832:
```suggestion
    model: SupportsEagleBase,
```

vllm/model_executor/models/utils.py
```diff
@@ -19,6 +19,7 @@
 )
 from vllm.logger import init_logger
 from vllm.model_executor.model_loader.weight_utils import default_weight_loader
+from vllm.model_executor.models.interfaces import supports_any_eagle
```
NickLucche, line 22:
```suggestion
from vllm.model_executor.models.interfaces import supports_any_eagle, SupportsEagleBase
```

NickLucche (commented):
This LGTM, left one minor comment. As an alternative to this approach, have you considered making `AutoWeightsLoader` eagle-aware? 
Something like `supports_any_eagle(module)` in `AutoWeightsLoader.__init__` and then chaining the map ops inside load_weights with the eagle-specific `process_eagle_weight`, wdyt?

vllm/model_executor/models/utils.py
eldarkurtic: it can be any nn.Module here, that's why there is a check below for `if supports_any_eagle`

NickLucche:
I think this is a valid concern, although I think we have mostly stabilized on using the utility since it was refactored. For this particular use-case though I don't think we're adding it to every possible model anyway right?
@DarkLight1337 may have more to say on it

vllm/v1/spec_decode/eagle.py
```diff
@@ -1007,24 +1007,22 @@ def load_model(self, target_model: nn.Module) -> None:
...
-            eagle_shape = self.model.model.embed_tokens.weight.shape
-            target_shape = target_embed_tokens.weight.shape
-            if eagle_shape == target_shape:
+            if not self.model.has_own_embed_tokens:
                 logger.info(
-                    "Assuming the EAGLE head shares the same vocab embedding"
-                    " with the target model."
+                    "Draft model embed_tokens are not initialized from the checkpoint "
+                    "weights. Assuming draft model should share the embedding weights "
+                    "with the target model to save memory."
```
benchislett: If the embed_tokens are uninitialized, then we aren't doing it to save memory anymore, but instead because we don't have any weights.
```suggestion
                    "Draft model embed_tokens are not initialized from the checkpoint "
                    "weights. Assuming draft model should share the embedding weights."
```

vllm/v1/spec_decode/eagle.py
```diff
@@ -1039,19 +1037,19 @@ def load_model(self, target_model: nn.Module) -> None:
...
-                and self.model.lm_head.weight.shape
-                == target_language_model.lm_head.weight.shape
+                and not self.model.has_own_lm_head
             ):
                 logger.info(
-                    "Assuming the EAGLE head shares the same lm_head"
-                    " with the target model."
+                    "Draft model lm_head is not initialized from the checkpoint weights"
+                    ". Assuming draft model should share the lm_head weights with the "
+                    "target model to save memory."
```
benchislett:
```suggestion
                    "Draft model lm_head is not initialized from the checkpoint weights"
                    ". Assuming draft model should share the lm_head weights with the "
                    "target model."
```

benchislett (changes requested):
I think this can still be greatly simplified if we just check the contents of the loaded weights. Discussed offline, @eldarkurtic will try it out and report back if it works.

vllm/v1/spec_decode/eagle.py
```diff
@@ -1003,52 +1004,96 @@ def load_model(self, target_model: nn.Module) -> None:
...
-                logger.info(
-                    "Assuming the EAGLE head shares the same vocab embedding"
-                    " with the target model."
-                )
-                del self.model.model.embed_tokens
-                self.model.model.embed_tokens = target_embed_tokens
+            share_embeddings = False
+            reason = ""
+            if hasattr(self.model, "has_own_embed_tokens"):
+                # EAGLE-{1,2,3} model
```
benchislett: I don't think there's a good reason to specify this here, since EAGLE1,2,3 is all of them so far
```suggestion
                # EAGLE model
```

vllm/v1/spec_decode/eagle.py
```diff
@@ -1003,52 +1004,96 @@ def load_model(self, target_model: nn.Module) -> None:
...
+                if not self.model.has_own_embed_tokens:
+                    share_embeddings = True
+                    reason = (
```
benchislett: Why have "reason" if it's just going to get passed into `logger.info` anyways? Can't you just call it here instead of having a variable?

eldarkurtic:
Summary of changes introduced by this PR (based on our offline discussions):

- We cannot simplify the implementation by checking only weight tensor values, because it is not possible to infer whether those values were loaded from a checkpoint or randomly initialized.
- To address this, we introduce two new flags: `has_own_lm_head` and `has_own_embed_tokens`, which all models supporting EAGLE-{1,2,3} must provide via two new interfaces: `SupportsEagle` (for EAGLE-{1,2}) and `SupportsEagle3`.
- If the drafter is an MTP model, it automatically shares `lm_head` and `embed_tokens` with the verifier model.
- If the drafter is an EAGLE model, we first check whether it loaded `lm_head` / `embed_tokens` from the checkpoint.
- If it *did* load them, we compare the loaded layers with those of the target model:
  - If they are exactly the same, we delete the duplicated layers and share the verifier’s layers to save memory.
  - If they differ, we keep them as distinct layers.
- If it *did not* load them, we assume they should match the target model’s layers, and we share them accordingly.
- Tests were updated to reflect the new flow.
- All new models added to vLLM are expected to inherit from `SupportsEagle` or `SupportsEagle3`, similar to the existing `SupportsLoRA` inheritance pattern.

Pinging for re-reviews: @benchislett @robertgshaw2-redhat @NickLucche

vllm/v1/spec_decode/eagle.py
eldarkurtic: done

vllm/v1/spec_decode/eagle.py
eldarkurtic: it looked a bit cleaner, but I don't have a strong preference. removed in the newest commit

benchislett (changes requested):
I'm satisfied with the strategy, but the naming feels odd.

"SupportsEagle" sounds like an attribute for a base model, but it seems like this is being used mostly for the EAGLE models as a mixin. This is fine, except "DeepSeekV2" implements `SupportsEagle` for some reason? Please either rename&clarify intent or remove from DSV2 and Llama and inherit on the EAGLE classes individually.

benchislett (approved):
Approved

PR #28551, opened 2025-11-12
Update xformers to 0.0.32.post2 instead of dev commit

EmilienM:
## Purpose

We were installing a non-released version of xformers for CUDA:
0.0.33+5d4b92a5.d20251029

5d4b92a5 is actually the commit for 0.0.32.post2 tag, so this PR just
replace the version so it's cleaner and we now consumed a stable tag
(which by the way is the latest).

### DIFF
--- a/requirements/cuda.txt
+++ b/requirements/cuda.txt
@@ -9,7 +9,6 @@ torch==2.9.0
 torchaudio==2.9.0
 # These must be updated alongside torch
 torchvision==0.24.0 # Required for phi3v processor. See https://github.com/pytorch/vision?tab=readme-ov-file#installation for corresponding version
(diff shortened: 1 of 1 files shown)


dtrifiro (approved):

NickLucche (commented):
I think the plan it's actually to get rid of xformers cc @ywang96.
I stil couldn't find a torch2.9 built wheel with this PR (though I suppose this wasn't the intent):
```
VLLM_USE_PRECOMPILED=1 uv pip install -e  .
  × No solution found when resolving dependencies:
  ╰─▶ Because xformers==0.0.32.post2 depends on torch==2.8.0 and vllm==0.11.1rc7.dev72+g074b03cb5.precompiled depends on torch==2.9.0, we..
```

PR #28112, opened 2025-11-05
[V0 deprecation] Deprecate use_v1 parameter

wangxiyuan:
## Purpose
`use_v1` for `get_attn_backend_cls` is useless now. Since this is an interface impact change, let's deprecate it  and remove it in the next release. 

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #28556, opened 2025-11-12
Fix pre-commit (and XPU) on `main`

hmellor:
- https://github.com/vllm-project/vllm/pull/28112 didn't cause conflicts but main still contained a `use_v1`
- Also fixes a pre-commit warning


DarkLight1337 (approved):

jikunshang:
thanks for fixing this!

PR #28562, opened 2025-11-12
[EPLB] Weight rearrangement optimization

ilmarkov:
In this PR we refactor eplb `rearrange_expert_weights` phase.

Instead of multiple small p2p operations which are done per layer, we take a group of layers and pack all p2p ops. It requires additional send and recv buffers but allows to reduce communication costs.

## Benchmarking.

Isolated `rearrange_expert_weights` microbenchmark.
QwenNext 80B with 128 redundant experts on 4 H100 GPUs.
**Total kernel duration for ncclSendRecv**
Before: 1.7s
After: 0.16s

As a result more than 10X communication kernel reduction.

## Purpose
EPLB weights distribution optimization

## Test Plan
PR also refactors the tests allowing nicer output and test failure without hangs.

`tests/distributed/test_eplb_execute.py`

## Test Result
Tests passed.


hmellor:
Closing as superseded: the numpy rebalance primitives and `preserve_intragpu_slots` landed in #29499, and the batched P2P collector landed as `EplbCommunicator` in #33176.

PR #28563, opened 2025-11-12
[Model] [Config] Correctly identify granite-4.0-micro as non-hybrid model

tdoublep:
## Purpose

The model `ibm-granite/granite-4.0-micro` uses the `GraniteMoeHybrid` config, but actually is not a hybrid model in that all of the layers are attention. This is encoded in the config using the `layer_types` field. 

Right now on main, we don't check layer_types when deciding if a model is hybrid or not. Thus, for this model we are (a) disabling prefix caching by default and (b) inflating the attention block size unncessarily. 

This PR adds a simple check of the `layer_types` field before deciding if a model is hybrid or not. 

## Test Plan

```
vllm serve ibm-granite/granite-4.0-micro
```

## Test Result

After this change, we no longer see prefix caching being disabled and the attention block size being inflated. 

---
<details>

</details>

### DIFF
--- a/vllm/config/model.py
+++ b/vllm/config/model.py
@@ -1619,6 +1619,13 @@ class ModelConfig:
 
     @property
     def is_hybrid(self) -> bool:
+        # Handle granite-4.0-micro case which uses hybrid config but does not
+        # actually contain any non-attention layers.
+        layer_types = getattr(self.hf_config, "layer_types", None)
+        if layer_types is not None and all(
+            layer == "attention" for layer in layer_types
+        ):
+            return False
         return self._model_info.is_hybrid
 
     @property


hmellor (approved):

vllm/config/model.py
```diff
@@ -1619,6 +1619,13 @@ def is_attention_free(self) -> bool:
...
     @property
     def is_hybrid(self) -> bool:
+        # Handle granite-4.0-micro case which uses hybrid config but does not
+        # actually contain any non-attention layers.
+        layer_types = getattr(self.hf_config, "layer_types", None)
+        if layer_types is not None and all(
+            layer == "attention" for layer in layer_types
+        ):
+            return False
         return self._model_info.is_hybrid
```
mgoin, line 1629: Could we put this information into `_model_info.is_hybrid` instead? I feel like we should define this as "low level" as possible

hmellor: The problem is that `_model_info` only has access to the class itself, not an instance of the class. So it cannot know anything about the config.

PR #28570, opened 2025-11-12
[Feature]: Implement naive prepare/finalize class to replace naive dispatching in fused_moe/layer.py

### LINKED ISSUE #28236: [Feature]: Implement naive prepare/finalize class to replace naive dispatching in fused_moe/layer.py

teddygood:
## Purpose
- Resolve #28236  by moving the legacy EP+DP “naive” dispatch/combine path into a dedicated `FusedMoENaivePrepareAndFinalize` subclass and wiring it through the modular kernel hooks.
- Extend `FusedMoEPrepareAndFinalize` with lightweight pre/post hooks so all prepare/finalize implementations can control routing/combine, eliminating the special-case logic from fused_moe/layer.py.
- Add unit coverage that proves the new subclass dispatches/combines correctly for plain, shared-expert, and zero-expert outputs.

## Test Plan

```bash
pytest tests/kernels/moe/test_naive_prepare_finalize.py
```

## Test Result

```bash
================================= test session starts =================================
platform linux -- Python 3.12.11, pytest-8.3.5
collected 2 items

tests/kernels/moe/test_naive_prepare_finalize.py ..                            [100%]

=========================== 2 passed, 3 warnings in 10.08s ===========================
```

---
<details>

</details>


heheda12345:
CC @tlrmchlsmth

vllm/model_executor/layers/fused_moe/naive_prepare_finalize.py
tlrmchlsmth, line 1: Could we rename this to `simple_prepare_finalize`, actually? Now that we have the allgather_reducescatter all2all backend, I think that calling this "naive" is a little disparaging 😄

vllm/model_executor/layers/fused_moe/naive_prepare_finalize.py
```diff
@@ -0,0 +1,48 @@
...
+import torch
+
+from vllm.distributed import get_ep_group
+from vllm.model_executor.layers.fused_moe.prepare_finalize import (
+    MoEPrepareAndFinalizeNoEP,
+)
+
+
+class FusedMoENaivePrepareAndFinalize(MoEPrepareAndFinalizeNoEP):
+    """Mirror the legacy dispatch/combine path via prepare/finalize hooks."""
```
tlrmchlsmth: I think we should ditch the reference to the legacy path since this PR removes that codepath

vllm/model_executor/layers/fused_moe/fused_moe_method_base.py
```diff
@@ -53,7 +53,15 @@ def uses_weight_scale_2_pattern(self) -> bool:
     def maybe_make_prepare_finalize(self) -> FusedMoEPrepareAndFinalize | None:
         from .all2all_utils import maybe_make_prepare_finalize
 
-        return maybe_make_prepare_finalize(self.moe, self.moe_quant_config)
+        prepare_finalize = maybe_make_prepare_finalize(self.moe, self.moe_quant_config)
+
+        if prepare_finalize is None and self.moe.dp_size > 1 and self.moe.use_ep:
+            # EP+DP without a specialized all2all backend: use naive dispatch/combine
+            from .naive_prepare_finalize import FusedMoENaivePrepareAndFinalize
+
```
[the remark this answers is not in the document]
tlrmchlsmth: This sounds like a very plausible issue and should be investigated + fixed if so

vllm/model_executor/layers/fused_moe/naive_prepare_finalize.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def preprocess_inputs(
+        self,
+        hidden_states: torch.Tensor,
+        router_logits: torch.Tensor,
+        layer: torch.nn.Module,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        is_sequence_parallel = getattr(layer, "is_sequence_parallel", False)
```
tlrmchlsmth: It would be better to not default to anything if `is_sequence_parallel` isn't set.  I'd rather vLLM loudly crash here rather than silently continue and maybe get the wrong answer.

vllm/model_executor/layers/fused_moe/naive_prepare_finalize.py
```diff
@@ -0,0 +1,48 @@
...
+            if zero_expert_num > 0:
+                expert_output, aux = result
+                return self._combine(expert_output, layer), aux
+        return self._combine(result, layer)
+
+    @staticmethod
+    def _combine(tensor: torch.Tensor, layer: torch.nn.Module) -> torch.Tensor:
+        if tensor.numel() == 0:
+            return tensor
+        is_sequence_parallel = getattr(layer, "is_sequence_parallel", False)
```
tlrmchlsmth: ditto, dangerous to default to false

tests/kernels/moe/test_naive_prepare_finalize.py
tlrmchlsmth, line 1: What failure conditions are covered by this test? I can't tell at a glance what it's trying to test against, as it isn't running performing any all2alls

tlrmchlsmth (commented):
Thanks for the contribution, looks like a good cleanup

Could you run an lm_eval test to sanity check? If you have access to a machine with 4 GPUs, we could test TP+DP with the following:
```
vllm serve {{MODEL}} -tp 2 -dp 2 --enable-expert-parallel --port 8192

lm_eval --model local-completions --tasks gsm8k --model_args model={{MODEL}},base_url={{BASE_URL}}/v1/completions,num_concurrent=50,max_retries=3,tokenized_requests=False --limit 100
```

vllm/model_executor/layers/fused_moe/naive_prepare_finalize.py
teddygood: I'm good with renaming to 'simple'. Should I do a full sweep (file/class names, imports, etc.) to 'simple' and drop 'naive' entirely, or keep 'naive' aliases for compatibility?

vllm/model_executor/layers/fused_moe/fused_moe_method_base.py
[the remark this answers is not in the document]
teddygood: I'll restore DP+EP cross-rank dispatch/combine when `prepare_finalize` is `None` and add a base helper to ensure all overrides receive the proper fallback.

vllm/model_executor/layers/fused_moe/naive_prepare_finalize.py
teddygood: Sure, I’ll update the docstring.

vllm/model_executor/layers/fused_moe/naive_prepare_finalize.py
teddygood: I’ll make `naive_prepare_finalize` require `layer.is_sequence_parallel`, so it fails instead of defaulting to False when the attribute is missing.

vllm/model_executor/layers/fused_moe/naive_prepare_finalize.py
teddygood: Updated `_combine` too to require `layer.is_sequence_parallel` (no default False), so it will fail if missing. Thanks.

tests/kernels/moe/test_naive_prepare_finalize.py
teddygood: On second thought, that makes sense. I’ll remove this test.

teddygood:
> Thanks for the contribution, looks like a good cleanup
> 
> Could you run an lm_eval test to sanity check? If you have access to a machine with 4 GPUs, we could test TP+DP with the following:
> 
> ```
> vllm serve {{MODEL}} -tp 2 -dp 2 --enable-expert-parallel --port 8192
> 
> lm_eval --model local-completions --tasks gsm8k --model_args model={{MODEL}},base_url={{BASE_URL}}/v1/completions,num_concurrent=50,max_retries=3,tokenized_requests=False --limit 100
> ```

Thanks for reviewing! Sorry for the slight delay since I needed some time to secure a gpu.

I ran the LM Eval tests you suggested using 4×A100 GPUs, with the Qwen/Qwen1.5-MoE-A2.7B model.

This is my first contribution, so thank you for your guidance. Please let me know if anything else needs to be updated.

```bash
|gsm8k|      3|flexible-extract|     5|exact_match|↑  | 0.01|±  |  0.01|
|     |       |strict-match    |     5|exact_match|↑  | 0.00|±  |  0.00|
```

PR #28576, opened 2025-11-12
[Kernel] Improve 2D Triton Attention Kernel

jvlunteren:
## Purpose
~~This pull request depends on PR #28306.~~

This pull request introduces the following improvements to the 2D attention kernel:

* **Exact Q-block calculation**: replace the previous estimation of the number of Q blocks per sequence and the overall Q-block count with an exact calculation.
* **Q-block to sequence mapping**: replace the 2D kernel function that determines the sequence for each Q block with a version based on the new exact calculation.
* ~~**Separate launch paths**: add distinct code paths for launching the 2D kernel for mixed prefill/decode batches versus decode-only batches.~~
* ~~**Kernel simplification for decode**: adapt the 2D kernel to remove the execution of instructions required only for prefill when running decode attention.~~

Impact and benefits:
* The first two changes reduce redundant Q blocks and eliminate unnecessary launches of Triton program instances that perform no actual work, improving GPU resource utilization.
* Both rely on a data structure built by the attention metadata builder class, which amortizes the (small) computation cost across multiple layers. This also removes CPU–GPU synchronization in the attention kernel launch code, enabling CUDA Graph capture.
* ~~The last two changes allow separate tuning of the 2D kernel for mixed prefill/decode and decode-only batches, and also simplify CUDA Graph support.~~

The changes in this PR are not intended to deliver significant performance gains immediately, but rather to lay the groundwork for future optimizations in upcoming PRs.

Note that the original dependencies on other PRs were removed. Additionally, the first two modifications listed above (exact Q-block calculation and Q-block to sequence mapping) have been applied only to the 2D kernel, not the 3D kernel. The latter will be optimized using a different approach in an upcoming PR.

## Test Plan

The unit test `./tests/kernels/attention/test_triton_unified_attention.py` has been updated to reflect the above changes. Other tests, such as `lm_eval`, remain compatible and can be used without modification.

## Test Result

================================================ 512 passed in 155.00s (0:02:34) ================================================
</pre>

<br>
<br>

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.798|±  |0.0180|
|     |       |strict-match    |     5|exact_match|↑  |0.786|±  |0.0184|</pre>

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.794|±  |0.0181|
|     |       |strict-match    |     5|exact_match|↑  |0.772|±  |0.0188|</pre>

@tdoublep @bringlein

---

> [!NOTE]
> No changes detected in the provided diff; no files modified, added, or removed.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9101dc756c14b6467c192e7bfc67036f40d76320. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit d637bc7fa39ccfd7fcaf6702cde878a788e3ee5c. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Introduces exact Q-block scheduling and optimized 2D kernel execution.
> 
> - Precompute `BLOCK_M`, `BLOCK_Q`, `num_q_blocks`, and `block_q_seq_boundaries_tensor` in attention metadata; pass them through backend to `unified_attention`
> - Update 2D kernel to use exact Q-block→sequence mapping and add a decode-only path (`only_decode`) that skips prefill-specific work
> - Always use 2D kernel for prefill; for decode, choose 2D vs 3D based on allocated softmax buffers and `seq_threshold_3D`
> - Refactor tile-size selection (`_get_tile_size`) to distinguish 2D vs 3D decode and keep Gemma3 optimizations
> - Add backward-compat fallback in `unified_attention` to compute missing metadata when not provided
> - Adjust tests to construct and pass new metadata fields and to validate both 2D/3D decode paths
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d637bc7fa39ccfd7fcaf6702cde878a788e3ee5c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Lays groundwork for more efficient Triton unified attention by eliminating redundant Q-block work and enabling clearer 2D/3D execution paths.
> 
> - Precompute and pass `BLOCK_M`, `BLOCK_Q`, `num_q_blocks`, and `block_q_seq_boundaries_tensor` via attention metadata/backend; add fallback computation in `unified_attention` for backward compatibility
> - 2D kernel now uses exact Q-block→sequence mapping and a decode-only path (`only_decode`) that skips prefill-only logic
> - Launch policy: always use 2D for prefill; for decode, choose 2D vs 3D based on allocated softmax buffers and `seq_threshold_3D`
> - Refactor tile-size selection (`_get_tile_size`) to distinguish 2D vs 3D decode while keeping Gemma3 optimizations
> - Tests updated to build and pass new fields and to exercise both decode paths (`seq_threshold_3D` param)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cdede79441f7bdd826695762273f6451b880d7f5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


heheda12345:
CC @tdoublep

vllm/attention/ops/triton_unified_attention.py
```diff
@@ -749,6 +708,15 @@ def unified_attention(
...
     v_descale,
+    BLOCK_M,
+    BLOCK_Q,
+    num_q_blocks,
+    block_q_seq_boundaries_tensor,
+    seq_threshold_3D,
+    num_par_softmax_segments,
+    softmax_segm_output,
+    softmax_segm_max,
+    softmax_segm_expsum,
```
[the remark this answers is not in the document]
jvlunteren: This has been resolved by a new update.

vllm/attention/ops/triton_unified_attention.py
```diff
@@ -819,7 +831,7 @@ def _get_tile_size(
         return 32
 
     # Default behavior
-    if is_prefill:
+    if is_prefill or use_2d_kernel:
```
tdoublep: Can you explain a bit more about why we are changing it to always use tile size 32 when using the 2D kernel?

vllm/v1/attention/backends/triton_attn.py
```diff
@@ -212,6 +221,33 @@ def build(
         block_table_tensor = common_attn_metadata.block_table_tensor
         slot_mapping = common_attn_metadata.slot_mapping
 
+        num_queries_per_kv = self.num_heads_q // self.num_heads_kv
+        BLOCK_M = (
+            16 if num_queries_per_kv <= 16 else next_power_of_2(num_queries_per_kv)
+        )
+        BLOCK_Q = BLOCK_M // num_queries_per_kv
+
+        if max_seq_len > 1:
```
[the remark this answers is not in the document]
jvlunteren: Fixed.

vllm/v1/attention/ops/triton_unified_attention.py
```diff
@@ -939,56 +955,68 @@ def unified_attention(
...
-    #   <= \sum_i[floor(query_len[i] / BLOCK_Q) + 1]
-    #    = \sum_i[floor(query_len[i] / BLOCK_Q)] + num_seqs
-    #   <= floor(\sum_i(query_len[i]) / BLOCK_Q) + num_seqs
-    #    = floor(q.shape[0] / BLOCK_Q) + num_seqs
-    total_num_q_blocks = q.shape[0] // BLOCK_Q + num_seqs
+    # Assign the following variables if they are not assigned in the attention metadata.
+    # This ensures backward compatibility with callers using an earlier version of this
+    # function. However, it is recommended to include these assignments in the
+    # attention metadata itself, as performing them here may negatively impact
+    # performance.
```
tdoublep: Wouldn't it be preferable to just change all the meta-data builders that call this function to include these tensors, rather than having this logic here? The meta-data builders should evolve in a tightly-coupled way imo, if we need to worry about backwards compatibility the code complexity will spiral.

vllm/v1/attention/ops/triton_unified_attention.py
```diff
@@ -104,28 +104,39 @@ def kernel_unified_attention_2d(
...
 
-    q_block_start_idx = tl.load(query_start_len_ptr + seq_idx) // BLOCK_Q + seq_idx
+    else:
+        q_block_global_idx = tl.program_id(0)
+        kv_head_idx = tl.program_id(1)
 
-    q_block_local_idx = q_block_global_idx - q_block_start_idx
+        seq_idx = find_seq_idx(
+            block_q_seq_boundaries_ptr, q_block_global_idx, num_seqs, BLOCK_Q, False
+        )
```
tdoublep: Is there a reason we can't compute the q_block to seq_idx mapping directly in the attention metadata and skip having to do the binary search?

vllm/v1/attention/ops/triton_unified_attention.py
```diff
@@ -104,28 +104,39 @@ def kernel_unified_attention_2d(
...
     BLOCK_M: tl.constexpr,  # int
+    block_q_seq_boundaries_ptr,  # [num_seqs+1]
+    only_decode,  # bool
     USE_FP8: tl.constexpr,  # bool
     FP8_MIN: tl.constexpr = float8_info.min,
     FP8_MAX: tl.constexpr = float8_info.max,
 ):
-    q_block_global_idx = tl.program_id(0)
-    kv_head_idx = tl.program_id(1)
+    if only_decode:
```
tdoublep: This `only_decode` change isn't really coupled to the changes related to the number of q_blocks right? I think I would prefer to consider the q_block changes as a standalone PR. The reason is that I'm concerned that 'only_decode' assumes query length is 1 which I think is kind of limited for spec decode case.

vllm/v1/attention/ops/triton_unified_attention.py
```diff
@@ -876,7 +888,7 @@ def _get_tile_size(
         return 32
 
     # Default behavior
-    if is_prefill:
+    if is_prefill or use_2d_kernel:
```
tdoublep: What motivates the change in the default tile size?

vllm/v1/attention/ops/triton_unified_attention.py
```diff
@@ -1039,77 +1067,151 @@ def unified_attention(
...
-            query_start_len_ptr=cu_seqlens_q,
-            BLOCK_Q=BLOCK_Q,
-            NUM_SEGMENTS_PER_SEQ=num_par_softmax_segments,
-            USE_FP8=output_scale is not None,
-        )
+    else:  # Decode
+        # Launch the 2D kernel if
+        # 1. No intermediate tiled softmax buffers have been allocated, or
+        # 2. The number of sequences exceeds the configured threshold
+        # 3. Batch invariance is enabled
```
tdoublep: I think adding a new branch here increases the complexity of the code. I understand that doing it may bring the possibility of tuning the kernel in the future, but I'd prefer to do it if and when we can demonstrate material improvement.

vllm/v1/attention/ops/triton_unified_attention.py
jvlunteren: OK. I will do that.

vllm/v1/attention/ops/triton_unified_attention.py
jvlunteren: Yes, we can do that. But prefills involving long prompts, that would require a much larger tensor with a size equal to the number of q blocks (containing the sequence id for each q block) instead of a tensor with a size equal to the number of sequences in the batch + 1 (containing the boundaries).

vllm/v1/attention/ops/triton_unified_attention.py
jvlunteren: It is actually related. The block_q_seq_boundaries tensor is only created as part of the attention metadata when there is at least one prefill in the batch. The latter is checked for by testing if max_query_len > 1. The latter may have to be updated for speculative decoding.

Therefore, to simplify the code, I can also modify that the block_q_seq_boundaries tensor is always created and searched also for decode-only batches, and the only_decode condition is removed.

vllm/v1/attention/ops/triton_unified_attention.py
jvlunteren: This additional condition ('or use_2d_kernel'), is only there to allow setting different tile sizes for the 2D kernel (separate for prefill and decode) and for the 3D kernel (decode):

```
    TILE_SIZE_2D_PREFILL = _get_tile_size(
        head_size,
        sliding_window_val,
        q.element_size(),
        is_prefill=True,
        use_2d_kernel=True,
    )
    TILE_SIZE_2D_DECODE = _get_tile_size(
        head_size,
        sliding_window_val,
        q.element_size(),
        is_prefill=False,
        use_2d_kernel=True,
    )
    TILE_SIZE_3D_DECODE = _get_tile_size(
        head_size,
        sliding_window_val,
        q.element_size(),
        is_prefill=False,
        use_2d_kernel=False,
    )
```
I assigned the default tile size for decodes only using 2D kernel based on a few experiments. I can also assign it a value equal to the tile size for decodes using the 3D kernel.

vllm/v1/attention/ops/triton_unified_attention.py
jvlunteren: This branch is essential for the 3rd objective (see above) of this PR: It implements distinct code paths for launching the 2D kernel for mixed prefill/decode batches versus decode-only batches.

vllm/v1/attention/ops/triton_unified_attention.py
jvlunteren: I have applied the latter approach. It also provides better support for correct speculative decoding.

vllm/v1/attention/ops/triton_unified_attention.py
jvlunteren: I restored the original default tile size assignment (removed this update).

vllm/v1/attention/ops/triton_unified_attention.py
jvlunteren: Done.

vllm/v1/attention/ops/triton_unified_attention.py
jvlunteren: It is probably better not to combine all these optimizations, so I will remove this one.

jvlunteren: Done.

jvlunteren: Also updated the PR description above.

vllm/v1/attention/ops/triton_unified_attention.py
```diff
@@ -112,10 +113,10 @@ def kernel_unified_attention_2d(
     kv_head_idx = tl.program_id(1)
 
     seq_idx = find_seq_idx(
-        query_start_len_ptr, q_block_global_idx, num_seqs, BLOCK_Q, True
+        block_q_seq_boundaries_ptr, q_block_global_idx, num_seqs, BLOCK_Q, False
```
tdoublep, line 116: Do we still need `query_start_len_ptr` anymore? And can we remove the final boolean argument from find_seq_idx?

vllm/v1/attention/backends/tree_attn.py
```diff
@@ -128,6 +166,20 @@ def decode_metadata(self) -> "TreeAttentionMetadata | None":
...
+        decode_block_q_seq_boundaries = torch.empty(
+            num_decode_seqs + 1, dtype=torch.int32, device=q_start_loc.device
+        )
+        decode_block_q_seq_boundaries[0] = 0
+        decode_block_q_seq_boundaries[1:].copy_(q_start_loc[1:])
+        decode_block_q_seq_boundaries[1:].sub_(q_start_loc[:-1])
+        decode_block_q_seq_boundaries[1:].add_(self.BLOCK_Q - 1)
+        decode_block_q_seq_boundaries[1:].floor_divide_(self.BLOCK_Q)
+        decode_block_q_seq_boundaries.cumsum_(dim=0)
+        decode_num_q_blocks = int(decode_block_q_seq_boundaries[-1].item())
```
tdoublep, line 181: We shouldn't have this logic 3x in the code - could we put in a utility function? I also think it can be simplified significantly.

tests/kernels/attention/test_triton_unified_attention.py
```diff
@@ -283,6 +301,21 @@ def test_triton_unified_attn_fp16_input_fp8_output(
...
+    num_queries_per_kv = num_query_heads // num_kv_heads
+    BLOCK_M = (
+        16 if num_queries_per_kv <= 16 else 1 << (num_queries_per_kv - 1).bit_length()
+    )  # next power of 2 value
+    BLOCK_Q = BLOCK_M // num_queries_per_kv
+
+    block_q_seq_boundaries_tensor = torch.empty(num_seqs + 1, dtype=torch.int32)
+    block_q_seq_boundaries_tensor[0] = 0
+    block_q_seq_boundaries_tensor[1 : cu_query_lens.numel()].copy_(cu_query_lens[1:])
+    block_q_seq_boundaries_tensor[1 : cu_query_lens.numel()].sub_(cu_query_lens[:-1])
```
tdoublep, line 313: I think we can call `naive_query_lens` on `CommonAttentionMetadata` to get the same thing: https://github.com/vllm-project/vllm/blob/main/vllm/v1/attention/backend.py#L345

PR #28579, opened 2025-11-12
[Core] Refactor padding logic and pad for CUDA graphs before attention metadata building 

### LINKED ISSUE #23789: [Attention]: Pad for cudagraphs before constructing attention metadata

LucasWilkinson:
FIX https://github.com/vllm-project/vllm/issues/23789

The goal of this PR is to:

1) Pad for cudagraphs before building attention metadata; this will allow us to 
                 - update to the latest FA3 (https://github.com/vllm-project/flash-attention/pull/82)
                 - remove hacks like: https://github.com/vllm-project/FlashMLA/pull/3
                 - remove `pad_for_cudagraphs` from attention backends; this is done for FlashInfer but will be done for `GDNAttentionBackend`, `Mamba1AttentionBackend`, `Mamba2AttentionMetadata`, `ShortConvAttentionBackend` in future PRs
2) Pad for cudagraphs in less places; prior to this PR `pad_for_cudagraphs` was called multiple times inside `execute_model` before the forward pass making it challenging to reason about the padding order. This PR starts to make the padding order in gpu_model_runner clearer but more work still needs to be done.
3) Generally make the padding logic more isolated and easier to reason about

Future related work that will be based off this PR:
- remove `pad_for_cudagraphs` from attention backends (see 1)
- remove `pad_for_cudagraphs` from config; transferring ownership to CUDAGraphDispatcher
        -  this will make it easier to have seperate cudagraph sizes for FULL and PIECEWISE; important for a more robust and long term solution to: https://github.com/vllm-project/vllm/issues/28207
- refactor dummy_run to seperate cudagraph capture

Shout-out to @ayushsatyam146 for the preliminary work in https://github.com/vllm-project/vllm/pull/24002
Co-authored-by: ayushsatyam146 <ayushsatyam146@gmail.com>

### DIFF
--- a/vllm/forward_context.py
+++ b/vllm/forward_context.py
@@ -35,23 +35,27 @@ class BatchDescriptor(NamedTuple):
     """
 
     num_tokens: int
-    uniform_decode: bool = False
+    num_reqs: int | None = None
     """
-    False can also be used for an uniform decode batch to dispatch to the 
-    cudagraph supporting non-uniform batches.
+    Number of requests in the batch. Can be None for PIECEWISE cudagraphs where
+    the cudagraphs can handle any number of requests.
+    """
+    uniform: bool = False
+    """
+    True if all the requests in the batch have the same number of tokens.
     """
     has_lora: bool = False
     """
     Whether this batch has active LoRA adapters.
     """
 
-    @property
-    def non_uniform(self) -> "BatchDescriptor":
+    def relax_for_mixed_batch_cudagraphs(self) -> "BatchDescriptor":
         """
-        Return a non-uniform version of current batch descriptor.
+        Return a relaxed version of current batch descriptor that is still compatible
+        with PIECEWISE cudagraphs (or mixed prefill-decode FA cudagraphs).
         """
         return BatchDescriptor(
-            self.num_tokens, uniform_decode=False, has_lora=self.has_lora
+            self.num_tokens, num_reqs=None, uniform=False, has_lora=self.has_lora
         )
 
 
--- a/vllm/v1/attention/backends/utils.py
+++ b/vllm/v1/attention/backends/utils.py
@@ -72,6 +72,7 @@ class CommonAttentionMetadata:
 
     num_reqs: int
     """Number of requests"""
+    # TODO(lucas): rename to num_tokens since it may be padded and this is misleading
     num_actual_tokens: int
     """Total number of tokens in batch"""
     max_query_len: int
@@ -857,7 +858,9 @@ def split_decodes_and_prefills(
     if require_uniform:
         is_prefill = query_lens != query_lens[0]
     else:
-        is_prefill = query_lens > decode_threshold
+        # 0-query len indicates a padded request; leave this at the back
+        # of the batch with the prefills
+        is_prefill = (query_lens > decode_threshold) | (query_lens == 0)
 
     if not torch.any(is_prefill):
         return num_reqs, 0, num_tokens, 0
--- a/vllm/v1/cudagraph_dispatcher.py
+++ b/vllm/v1/cudagraph_dispatcher.py
@@ -4,6 +4,9 @@ from itertools import product
 
 from vllm.config import CUDAGraphMode, VllmConfig
 from vllm.forward_context import BatchDescriptor
+from vllm.logger import init_logger
+
+logger = init_logger(__name__)
 
 
 class CudagraphDispatcher:
@@ -28,7 +31,11 @@ class CudagraphDispatcher:
     def __init__(self, vllm_config: VllmConfig):
         self.vllm_config = vllm_config
         self.compilation_config = vllm_config.compilation_config
-        self.cudagraph_mode = self.compilation_config.cudagraph_mode
+        self.uniform_decode_query_len = (
+            1
+            if not self.vllm_config.speculative_config
+            else 1 + self.vllm_config.speculative_config.num_speculative_tokens
+        )
 
         # Dict to store valid cudagraph dispatching keys.
         self.cudagraph_keys: dict[CUDAGraphMode, set[BatchDescriptor]] = {
@@ -36,24 +43,42 @@ class CudagraphDispatcher:
             CUDAGraphMode.FULL: set(),
         }
 
-        not_use_piecewise_compilation = (
-            not self.cudagraph_mode.requires_piecewise_compilation()
-        )
-
         assert (
-            not_use_piecewise_compilation
+            not self.compilation_config.cudagraph_mode.requires_piecewise_compilation()
             or self.compilation_config.is_attention_compiled_piecewise()
         ), (
             "Compilation mode should be CompilationMode.VLLM_COMPILE when "
             "cudagraph_mode piecewise cudagraphs is used, "
             "and attention should be in splitting_ops or "
             "inductor splitting should be used. "
-            f"cudagraph_mode={self.cudagraph_mode}, "
+            f"cudagraph_mode={self.compilation_config.cudagraph_mode}, "
             f"compilation_mode={self.compilation_config.mode}, "
             f"splitting_ops={self.compilation_config.splitting_ops}"
         )
 
         self.keys_initialized = False
+
+    def _create_padded_batch_descriptor(
+        self, num_tokens: int, uniform_decode: bool, has_lora: bool
+    ) -> BatchDescriptor:
+        max_num_seqs = self.vllm_config.scheduler_config.max_num_seqs
+        uniform_decode_query_len = self.uniform_decode_query_len
+        num_tokens_padded = self.vllm_config.pad_for_cudagraph(num_tokens)
+
+        if uniform_decode and self.cudagraph_mode.has_mode(CUDAGraphMode.FULL):
+            num_reqs = num_tokens_padded // uniform_decode_query_len
+            assert num_tokens_padded % uniform_decode_query_len == 0
+            assert num_reqs <= max_num_seqs
+        else:
+            uniform_decode = False
+            num_reqs = min(num_tokens_padded, max_num_seqs)
+
+        return BatchDescriptor(
+            num_tokens=num_tokens_padded,
+            num_reqs=num_reqs,
+            uniform=uniform_decode,
+            has_lora=has_lora,
+        )
 
     def add_cudagraph_key(
         self, runtime_mode: CUDAGraphMode, batch_descriptor: BatchDescriptor
@@ -66,7 +91,9 @@ class CudagraphDispatcher:
     def initialize_cudagraph_keys(
         self, cudagraph_mode: CUDAGraphMode, uniform_decode_query_len: int
     ):
-        # This should be called only after attention backend is initialized.
+        # This should be called only after attention backend is initialized. So we can
+        # get the correct cudagraph mode after backend support is resolved.
+        self.cudagraph_mode = cudagraph_mode
 
         # LoRA activation cases to specialize the cuda graphs on
         if self.vllm_config.lora_config:
@@ -86,9 +113,9 @@ class CudagraphDispatcher:
             ):
                 self.add_cudagraph_key(
                     cudagraph_mode.mixed_mode(),
-                    BatchDescriptor(
-                        num_tokens=bs, uniform_decode=False, has_lora=has_lora
-                    ),
+                    self._create_padded_batch_descriptor(
+                        bs, False, has_lora
+                    ).relax_for_mixed_batch_cudagraphs(),
                 )
 
         # if decode cudagraph mode is FULL, and we don't already have mixed
@@ -109,40 +136,49 @@ class CudagraphDispatcher:
             for bs, has_lora in product(cudagraph_capture_sizes_for_decode, lora_cases):
                 self.add_cudagraph_key(
                     CUDAGraphMode.FULL,
-                    BatchDescriptor(
-                        num_tokens=bs, uniform_decode=True, has_lora=has_lora
-                    ),
+                    self._create_padded_batch_descriptor(bs, True, has_lora),
                 )
+
         self.keys_initialized = True
 
     def dispatch(
-        self, batch_descriptor: BatchDescriptor, use_cascade_attn: bool = False
-    ) -> tuple[CUDAGraphMode, BatchDescriptor | None]:
+        self,
+        num_tokens: int,
+        uniform_decode: bool,
+        has_lora: bool,
+        use_cascade_attn: bool = False,
+    ) -> tuple[CUDAGraphMode, BatchDescriptor]:
         """
         Given conditions(e.g.,batch descriptor and if using cascade attention),
         dispatch to a cudagraph runtime mode and the valid batch descriptor.
         A new batch descriptor is returned as we might dispatch a uniform batch
         to a graph that supports a more general batch (uniform to non-uniform).
         """
-        # if not initialized, just skip dispatching.
-        if not self.keys_initialized:
-            return CUDAGraphMode.NONE, None
+        if (
+            not self.keys_initialized
+            or self.cudagraph_mode == CUDAGraphMode.NONE
+            or num_tokens > self.compilation_config.max_cudagraph_capture_size
+        ):
+            return CUDAGraphMode.NONE, BatchDescriptor(num_tokens)
 
-        non_uniform_key = batch_descriptor.non_uniform
-        # if a batch use cascade attention, bypass checking full cudagraphs
+        batch_desc = self._create_padded_batch_descriptor(
+            num_tokens, uniform_decode, has_lora
+        )
+        relaxed_batch_desc = batch_desc.relax_for_mixed_batch_cudagraphs()
+
         if not use_cascade_attn:
             # check if key exists for full cudagraph
-            if batch_descriptor in self.cudagraph_keys[CUDAGraphMode.FULL]:
-                return CUDAGraphMode.FULL, batch_descriptor
+            if batch_desc in self.cudagraph_keys[CUDAGraphMode.FULL]:
+                return CUDAGraphMode.FULL, batch_desc
 
-            # otherwise, check if non-uniform key exists
-            if non_uniform_key in self.cudagraph_keys[CUDAGraphMode.FULL]:
-                return CUDAGraphMode.FULL, non_uniform_key
+            # otherwise, check if the relaxed key exists
+            if relaxed_batch_desc in self.cudagraph_keys[CUDAGraphMode.FULL]:
+                return CUDAGraphMode.FULL, relaxed_batch_desc
 
-        # also check if non-uniform key exists for more "general"
+        # also check if the relaxed key exists for more "general"
         # piecewise cudagraph
-        if non_uniform_key in self.cudagraph_keys[CUDAGraphMode.PIECEWISE]:
-            return CUDAGraphMode.PIECEWISE, non_uniform_key
+        if relaxed_batch_desc in self.cudagraph_keys[CUDAGraphMode.PIECEWISE]:
+            return CUDAGraphMode.PIECEWISE, relaxed_batch_desc
 
-        # finally, just return no cudagraphs
-        return CUDAGraphMode.NONE, None
+        # finally, just return no cudagraphs and a trivial batch descriptor
+        return CUDAGraphMode.NONE, BatchDescriptor(num_tokens)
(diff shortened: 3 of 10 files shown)


vllm/forward_context.py
```diff
@@ -35,23 +35,27 @@ class BatchDescriptor(NamedTuple):
     """
 
     num_tokens: int
-    uniform_decode: bool = False
+    num_reqs: int | None = None
     """
-    False can also be used for an uniform decode batch to dispatch to the 
-    cudagraph supporting non-uniform batches.
+    Number of requests in the batch. Can be None for PIECEWISE cudagraphs where
```
SageMoore, line 40: Nit: Typo

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2508,6 +2445,78 @@ def _model_forward(
...
+                num_scheduled_tokens_per_request=num_scheduled_tokens_np,
+            )
+
+            # Extract DP padding if there is any
+            if num_tokens_across_dp is not None:
+                dp_rank = self.parallel_config.data_parallel_rank
+                num_tokens_padded = int(num_tokens_across_dp[dp_rank].item())
+
+                # Re-dispatch with DP padding
+                cudagraph_mode, batch_descriptor = dispatch_cudagraph(num_tokens_padded)
```
SageMoore, line 2770: Nit: I'd like for us to assert that the total number of tokens in `ubatch_slices` is equal to `num_tokens_padded`.

SageMoore (commented):
Looks like a good cleanup @LucasWilkinson. Thanks for the contribution.

vllm/v1/worker/gpu_model_runner.py
LucasWilkinson: done in `coordinate_batch_across_dp`

vllm/v1/cudagraph_dispatcher.py
```diff
@@ -55,6 +63,32 @@ def __init__(self, vllm_config: VllmConfig):
 
         self.keys_initialized = False
 
+    def _create_padded_batch_descriptor(
+        self, num_tokens: int, uniform_decode: bool, has_lora: bool
```
ProExpertProg, line 62: Call this `uniform` if we're calling the property `uniform`?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2622,6 +2576,78 @@ def _model_forward(
...
+        cudagraph_mode, batch_descriptor = dispatch_cudagraph(num_tokens_padded)
+        num_tokens_padded = batch_descriptor.num_tokens
+
+        # Extra coordination when running data-parallel since we need to coordinate
+        # across ranks
+        ubatch_slices, num_tokens_across_dp = None, None
+        if self.vllm_config.parallel_config.data_parallel_size > 1:
+            # Disable DP padding when running eager to avoid excessive padding when
+            # running prefills. This lets us set enforce_eager on the prefiller in
+            # a P/D setup and still use CUDA graphs (enabled by this padding) on the
```
ProExpertProg:
```suggestion
            # running prefills. This lets us set cudagraph_mode="NONE" on the prefiller in
            # a P/D setup and still use CUDA graphs (enabled by this padding) on the
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2622,6 +2576,78 @@ def _model_forward(
...
+        num_scheduled_tokens_np: np.ndarray,
+        max_num_scheduled_tokens: int,
+        use_cascade_attn: bool,
+        allow_microbatching: bool = True,
+        force_eager: bool = False,
+        force_uniform_decode: bool | None = None,
+    ) -> tuple[
+        CUDAGraphMode, BatchDescriptor, UBatchSlices | None, torch.Tensor | None
+    ]:
+        num_tokens_padded = self._pad_for_sequence_parallelism(num_tokens)
```
ProExpertProg, line 2712: So we potentially do this twice? Once in the worker and once in the `_determine_batch_execution_and_padding`? I get this has to happen inside the `_dummy_run` but for `execute_model` this now happens 2x, which is benign just perhaps unnecessary & more confusing

ProExpertProg (approved):
LGTM overall, just a few qs above

vllm/v1/cudagraph_dispatcher.py
LucasWilkinson: Im planning on refactoring this in a future PR but `uniform_decode` implies `uniform and query_len == num_speculated_tokens + 1` while the uniform in the property is a bit more flexible; implies just `uniform` but the `query_len` could be anything (just leaving it open to uniform graphs that aren't `num_speculated_tokens + 1` to be used potentially in the speculator where we speculate with a query_len = 1 for additional tokens; granted the graph would still probably be `num_speculated_tokens + 1` since with padded speculation we'd start with that for the first eagle head)

SageMoore (approved):
This all looks good to me, Lucas.

vllm/v1/cudagraph_dispatcher.py
```diff
@@ -109,40 +135,49 @@ def initialize_cudagraph_keys(
...
 
     def dispatch(
-        self, batch_descriptor: BatchDescriptor, use_cascade_attn: bool = False
-    ) -> tuple[CUDAGraphMode, BatchDescriptor | None]:
+        self,
+        num_tokens: int,
+        uniform_decode: bool,
+        has_lora: bool,
+        use_cascade_attn: bool = False,
+    ) -> tuple[CUDAGraphMode, BatchDescriptor]:
```
mgoin, line 149: It seems like a step backward to change from taking BatchDescriptor to individual params - why did you choose this?

vllm/v1/cudagraph_dispatcher.py
```diff
@@ -28,33 +31,54 @@ class CudagraphDispatcher:
...
+            num_reqs = num_tokens_padded // uniform_decode_query_len
+            assert num_tokens_padded % uniform_decode_query_len == 0
+        else:
+            uniform_decode = False
+            num_reqs = min(num_tokens_padded, max_num_seqs)
+
+        return BatchDescriptor(
+            num_tokens=num_tokens_padded,
+            num_reqs=num_reqs,
+            uniform=uniform_decode,
```
mgoin, line 78: nit: make BatchDescriptor's field also be uniform_decode to match

vllm/v1/attention/backends/utils.py
```diff
@@ -856,7 +857,9 @@ def split_decodes_and_prefills(
     if require_uniform:
         is_prefill = query_lens != query_lens[0]
     else:
-        is_prefill = query_lens > decode_threshold
+        # 0-query len indicates a padded request; leave this at the back
+        # of the batch with the prefills
+        is_prefill = (query_lens > decode_threshold) | (query_lens == 0)
```
mgoin, line 863: Is this function only used for reordering?

vllm/v1/worker/dp_utils.py
```diff
@@ -88,6 +89,18 @@ def _post_process_dp_padding(tensor: torch.Tensor, should_dp_pad: bool) -> torch
...
+# Should be called after attention metadata creation. This just pads
+# the second ubatch slice out to the total number of tokens
+# (num_tokens + padding)
+def _pad_out_ubatch_slice(ubatch_slices: UBatchSlices, num_total_tokens: int):
+    padded_second_ubatch_slice = slice(
+        ubatch_slices[1].token_slice.start, num_total_tokens
+    )
+    ubatch_slices[1] = UBatchSlice(
+        padded_second_ubatch_slice, padded_second_ubatch_slice
+    )
```
mgoin: Where is num_total_tokens relative to in this case? Are we having to pad to the total of tokens across all ranks?

This seems particularly tricky to avoid bad double padding. Could there be a test for this if this is the intended strategy for the future?

vllm/v1/attention/backends/utils.py
LucasWilkinson: yes; only used by attention backends to split the batch

vllm/v1/cudagraph_dispatcher.py
LucasWilkinson: The idea is to move to a regime where `BatchDescriptor` describes the exact workload captured in the cuda-graph but since some cudagraphs (PIECEWISE for example) can match different batches so the idea is to pass here all the parameters needed to find a matching cudagraph which may differ from the params that describe the graph (`BatchDescriptor`); id be open to renaming this CUDAGraphDescriptor, the only edge case is when no cudagraph mode is used (eager) we still return a batch descriptor that matches the current workload (basically implicitly indicating no padding)

Does that make sense?

vllm/v1/cudagraph_dispatcher.py
LucasWilkinson: Im planning on refactoring this in a future PR but `uniform_decode` implies uniform and `query_len == num_speculated_tokens + 1` while the uniform property in BatchDescriptor is a bit more flexible and implies just uniform but the query_len could be anything (just leaving it open to uniform graphs that aren't num_speculated_tokens + 1 to be used potentially in the speculator where we speculate with a query_len = 1 for additional tokens; granted the graph would still probably be num_speculated_tokens + 1 since with padded speculation we'd start with that for the first eagle head)

vllm/v1/worker/dp_utils.py
LucasWilkinson: Updated the comment :+1:; basically this is because `create_ubatch_slices` is called on unpadded `num_scheduled_tokens` then we need to use `_pad_out_ubatch_slice` to pad for cudagraphs/dp-padding

vllm/v1/worker/gpu_model_runner.py
LucasWilkinson: ya I ran into this rough edge; for now im just calling `_determine_batch_execution_and_padding` in the worker but I do think we need to have a broader rethink of PP+SP since having to replicate the padding logic in the worker is rough

for reference the padding logic was already being run twice just via `_get_num_input_tokens` on main (which also made it impossible to have separate cudagraph shapes for decode and mixed-decode-prefill cudagraphs since we would need the CG padding in the worker to match the model runner)

imo we should move the PP stuff into the model runner (since alot of logic in the model runner already https://github.com/vllm-project/vllm/blob/0abc79482a6d476f58530907443c46134ba6e2e1/vllm/v1/worker/gpu_model_runner.py#L2927-L2974; its kinda split between model runner and worker right now) or `_determine_batch_execution_and_padding` into the work and pass the result into the execute_model (this is less preferable since it results in repeated construction of `num_scheduled_tokens_np` and computation of `max_num_scheduled_tokens`. I started to look and moving the PP stuff into the model runner so we only have to call `_determine_batch_execution_and_padding` once but then ran into issues with Ray+PP so started to feel out of scope of this PR.

LucasWilkinson: Im going to open a GitHub issue around this topic

mgoin (approved):
Thanks for the clarification. Looks good to me!

vllm/v1/worker/gpu_model_runner.py
ProExpertProg: Sounds good, thanks for explaining! I just wanted to confirm my understanding, it made sense to keep it this way for this PR

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2737,6 +2693,87 @@ def _model_forward(
...
+            # Disable DP padding when running eager to avoid excessive padding when
+            # running prefills. This lets us set cudagraph_mode="NONE" on the prefiller
+            # in a P/D setup and still use CUDA graphs (enabled by this padding) on the
+            # decoder.
+            allow_dp_padding = (
+                self.compilation_config.cudagraph_mode != CUDAGraphMode.NONE
+            )
+
+            ubatch_slices, num_tokens_across_dp = coordinate_batch_across_dp(
+                num_tokens_unpadded=num_tokens_padded,
```
hidva, line 2755: `num_tokens_unpadded = num_tokens`?

minosfuture:
@LucasWilkinson I'm hitting this issue:
```
(EngineCore_DP4 pid=858652) Traceback (most recent call last):
(EngineCore_DP4 pid=858652)   File "/usr/lib/python3.12/multiprocessing/process.py", line 314, in _bootstrap
(EngineCore_DP4 pid=858652)     self.run()
(EngineCore_DP4 pid=858652)   File "/usr/lib/python3.12/multiprocessing/process.py", line 108, in run
(EngineCore_DP4 pid=858652)     self._target(*self._args, **self._kwargs)
(EngineCore_DP4 pid=858652)   File "/data/nfs01/ming/vllm/vllm/v1/engine/core.py", line 850, in run_engine_core
(EngineCore_DP4 pid=858652)     raise e
(EngineCore_DP4 pid=858652)   File "/data/nfs01/ming/vllm/vllm/v1/engine/core.py", line 839, in run_engine_core
(EngineCore_DP4 pid=858652)     engine_core.run_busy_loop()
(EngineCore_DP4 pid=858652)   File "/data/nfs01/ming/vllm/vllm/v1/engine/core.py", line 1228, in run_busy_loop
(EngineCore_DP4 pid=858652)     self.execute_dummy_batch()
(EngineCore_DP4 pid=858652)   File "/data/nfs01/ming/vllm/vllm/v1/engine/core.py", line 499, in execute_dummy_batch
(EngineCore_DP4 pid=858652)     self.model_executor.execute_dummy_batch()
(EngineCore_DP4 pid=858652)   File "/data/nfs01/ming/vllm/vllm/v1/executor/abstract.py", line 229, in execute_dummy_batch
(EngineCore_DP4 pid=858652)     self.collective_rpc("execute_dummy_batch")
(EngineCore_DP4 pid=858652)   File "/data/nfs01/ming/vllm/vllm/v1/executor/uniproc_executor.py", line 75, in collective_rpc
(EngineCore_DP4 pid=858652)     result = run_method(self.driver_worker, method, args, kwargs)
(EngineCore_DP4 pid=858652)              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP4 pid=858652)   File "/data/nfs01/ming/vllm/vllm/v1/serial_utils.py", line 479, in run_method
(EngineCore_DP4 pid=858652)     return func(*args, **kwargs)
(EngineCore_DP4 pid=858652)            ^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP4 pid=858652)   File "/data/nfs01/ming/vllm/vllm/v1/worker/gpu_worker.py", line 632, in execute_dummy_batch
(EngineCore_DP4 pid=858652)     self.model_runner._dummy_run(1, uniform_decode=True)
(EngineCore_DP4 pid=858652)   File "/data/nfs01/ming/vllm/.venv/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 120, in decorate_context
(EngineCore_DP4 pid=858652)     return func(*args, **kwargs)
(EngineCore_DP4 pid=858652)            ^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP4 pid=858652)   File "/data/nfs01/ming/vllm/vllm/v1/worker/gpu_model_runner.py", line 3947, in _dummy_run
(EngineCore_DP4 pid=858652)     self._determine_batch_execution_and_padding(
(EngineCore_DP4 pid=858652)   File "/data/nfs01/ming/vllm/vllm/v1/worker/gpu_model_runner.py", line 2825, in _determine_batch_execution_and_padding
(EngineCore_DP4 pid=858652)     assert batch_descriptor.num_tokens == num_tokens_padded
(EngineCore_DP4 pid=858652)            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP4 pid=858652) AssertionError
```

couldn't do a clean revert so I'm not sure if it is caused by this PR yet.  Checking.

hjjq:
```
export VLLM_ATTENTION_BACKEND=FLASHINFER_MLA
export VLLM_USE_FLASHINFER_MOE_FP8=1
export VLLM_USE_NCCL_SYMM_MEM=1
export NCCL_NVLS_ENABLE=1
export NCCL_CUMEM_ENABLE=1
export VLLM_USE_TRTLLM_RAGGED_DEEPSEEK_PREFILL=1

python3 -m vllm.entrypoints.openai.api_server --model nvidia/DeepSeek-R1-0528-FP4 --tokenizer nvidia/DeepSeek-R1-0528-FP4 --dtype auto --kv-cache-dtype fp8 --tensor-parallel-size 1 --pipeline-parallel-size 1 --data-parallel-size 4 --enable-expert-parallel --swap-space 16 --max-num-seqs 1024 --trust-remote-code --max-model-len 10240 --gpu-memory-utilization 0.9 --max-num-batched-tokens 8192 --no-enable-prefix-caching --async-scheduling --compilation_config.pass_config.fuse_allreduce_rms true --compilation_config.pass_config.fuse_attn_quant true --compilation_config.pass_config.eliminate_noops true --compilation_config.custom_ops+=+quant_fp8,+rms_norm --max-cudagraph-capture-size 2048 --compilation_config.cudagraph_mode FULL_DECODE_ONLY --stream-interval=20 --api-server-count=20

vllm bench serve --dataset-name random --ignore-eos --num-prompts 10240 --max-concurrency 2048 --random-range-ratio 0.8 --random-input-len 2048 --random-output-len 1024 --model nvidia/DeepSeek-R1-0528-FP4 --base-url http://0.0.0.0:8000/ --ready-check-timeout-sec 0
```

@minosfuture 's error can also be reproduced with above. Only 4xGB200/B200 required.

LucasWilkinson:
> ```
> export VLLM_ATTENTION_BACKEND=FLASHINFER_MLA
> export VLLM_FLASHINFER_MOE_BACKEND=latency
> export VLLM_USE_FLASHINFER_MOE_FP8=1
> export VLLM_USE_FLASHINFER_MOE_FP4=1
> export VLLM_USE_NCCL_SYMM_MEM=1
> export NCCL_NVLS_ENABLE=1
> export NCCL_CUMEM_ENABLE=1
> export VLLM_USE_TRTLLM_RAGGED_DEEPSEEK_PREFILL=1
> 
> python3 -m vllm.entrypoints.openai.api_server --model nvidia/DeepSeek-R1-0528-FP4 --tokenizer nvidia/DeepSeek-R1-0528-FP4 --dtype auto --kv-cache-dtype fp8 --tensor-parallel-size 1 --pipeline-parallel-size 1 --data-parallel-size 4 --enable-expert-parallel --swap-space 16 --max-num-seqs 1024 --trust-remote-code --max-model-len 10240 --gpu-memory-utilization 0.9 --max-num-batched-tokens 8192 --no-enable-prefix-caching --async-scheduling --compilation_config.pass_config.fuse_allreduce_rms true --compilation_config.pass_config.fuse_attn_quant true --compilation_config.pass_config.eliminate_noops true --compilation_config.custom_ops+=+quant_fp8,+rms_norm --max-cudagraph-capture-size 2048 --compilation_config.cudagraph_mode FULL_DECODE_ONLY --stream-interval=20 --api-server-count=20
> 
> vllm bench serve --dataset-name random --ignore-eos --num-prompts 10240 --max-concurrency 2048 --random-range-ratio 0.8 --random-input-len 2048 --random-output-len 1024 --model nvidia/DeepSeek-R1-0528-FP4 --base-url http://0.0.0.0:8000/ --ready-check-timeout-sec 0
> ```
> 
> @minosfuture 's error can also be reproduced with above. Only 4xGB200/B200 required.

Thank you! I do not have access to a GB200 so this helps; what torch and cuda version are you using? im getting NCCL crashes when I run this command

hjjq:
We caught this from our ci which uses the vllm nightly image. I was also able to reproduce locally on 4xGB200 with torch 2.9.0 cu129 from a base image `nvcr.io/nvidia/pytorch:25.06-py3` and vllm built from source. I was just trying to run on 4xB200 with latest main but I hit some allreduce fusion pass error. I'll get back once I can reproduce on B200.

Issue #25623, opened 2025-09-25
[Bug]: torch.compile padding causes IMA on Hopper + DBO


ProExpertProg:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
Collecting environment information...
uv is set
==============================
        System Info
==============================
OS                           : Ubuntu 22.04.4 LTS (x86_64)
GCC version                  : (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Clang version                : Could not collect
CMake version                : Could not collect
Libc version                 : glibc-2.35

==============================
       PyTorch Info
==============================
PyTorch version              : 2.8.0+cu128
Is debug build               : False
CUDA used to build PyTorch   : 12.8
ROCM used to build PyTorch   : N/A

==============================
      Python Environment
==============================
Python version               : 3.10.12 (main, Jan 17 2025, 14:35:34) [GCC 11.4.0] (64-bit runtime)
Python platform              : Linux-5.15.0-113-generic-x86_64-with-glibc2.35

==============================
       CUDA / GPU Info
==============================
Is CUDA available            : True
CUDA runtime version         : 12.8.61
CUDA_MODULE_LOADING set to   : LAZY
GPU models and configuration : 
GPU 0: NVIDIA H100 80GB HBM3
GPU 1: NVIDIA H100 80GB HBM3
GPU 2: NVIDIA H100 80GB HBM3
GPU 3: NVIDIA H100 80GB HBM3
GPU 4: NVIDIA H100 80GB HBM3
GPU 5: NVIDIA H100 80GB HBM3
GPU 6: NVIDIA H100 80GB HBM3
GPU 7: NVIDIA H100 80GB HBM3

Nvidia driver version        : 570.86.10
cuDNN version                : Could not collect
HIP runtime version          : N/A
MIOpen runtime version       : N/A
Is XNNPACK available         : True

==============================
          CPU Info
==============================
Architecture:                       x86_64
CPU op-mode(s):                     32-bit, 64-bit
Address sizes:                      52 bits physical, 57 bits virtual
Byte Order:                         Little Endian
CPU(s):                             384
On-line CPU(s) list:                0-383
Vendor ID:                          AuthenticAMD
Model name:                         AMD EPYC 9654 96-Core Processor
CPU family:                         25
Model:                              17
Thread(s) per core:                 2
Core(s) per socket:                 96
Socket(s):                          2
Stepping:                           1
[...]

PR #24666, opened 2025-09-11
[Performance] Move apply_w8a8_block_fp8_linear to an op class

ElizaWszola:
This PR unwraps `apply_w8a8_block_fp8_linear` custom op by implementing a class that we can instantiate and run from `Fp8LinearMethod`. Part of work outlined in https://github.com/vllm-project/vllm/issues/24629

#### Testing:
E2E tested with on Blackwell and Hopper
```
llm = LLM(model="Qwen/Qwen3-30B-A3B-FP8")
llm = LLM(model="RedHatAI/Qwen3-0.6B-FP8-BLOCK")
```

PR #28580, opened 2025-11-12
[torch.compile] Try using Dynamo-native padding

ProExpertProg:
## Purpose

Reapply padding from #24666, which caused bug #25623. Currently fails with different error due to torch==2.9 (works with torch==2.8):

```
VLLM_USE_DEEP_GEMM=0 vllm serve Qwen/Qwen3-30B-A3B-FP8 --trust-remote-code
```

**UPDATE**: this now works!!

## Test Result

### DIFF
--- a/vllm/model_executor/layers/quantization/utils/fp8_utils.py
+++ b/vllm/model_executor/layers/quantization/utils/fp8_utils.py
@@ -149,7 +149,12 @@ def _padded_cutlass2(
 ) -> torch.Tensor:
     pad_multiple = 4
 
-    padding = -qx.shape[0] % pad_multiple
+    # TODO(torch==2.10): torch semantics broken for PythonMod in 2.9:
+    #  so x<=0 -> x % y <= 0 (not actually python mod semantics).
+    #  https://github.com/pytorch/pytorch/issues/169602
+    #  We increment the padded dim to avoid a negative.
+    # padding = -qx.shape[0] % pad_multiple
+    padding = (qx.shape[0] * pad_multiple - qx.shape[0]) % pad_multiple
     # print(f"{padding=}")
     padded_qx = torch.nn.functional.pad(qx, (0, 0, 0, padding))
     assert len(x_scale.size()) == 2
--- a/vllm/model_executor/layers/quantization/utils/fp8_utils.py
+++ b/vllm/model_executor/layers/quantization/utils/fp8_utils.py
@@ -139,6 +139,28 @@ def _padded_cutlass(
         )
 
 
+def _padded_cutlass2(
+    qx: torch.Tensor,
+    weight: torch.Tensor,
+    x_scale: torch.Tensor,
+    weight_scale: torch.Tensor,
+    block_size: list[int],
+    output_dtype: torch.dtype,
+) -> torch.Tensor:
+    pad_multiple = 4
+
+    padding = -qx.shape[0] % pad_multiple
+    # print(f"{padding=}")
+    padded_qx = torch.nn.functional.pad(qx, (0, 0, 0, padding))
+    assert len(x_scale.size()) == 2
+    padded_x_scale = torch.nn.functional.pad(x_scale, (0, 0, 0, padding), value=1.0)
+
+    output = cutlass_scaled_mm(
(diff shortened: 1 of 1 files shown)


ProExpertProg:
I tried running the full repro through lm_eval and it seems like there's an issue:

```
# serve
VLLM_USE_DEEP_GEMM=0 VLLM_ALL2ALL_BACKEND=deepep_low_latency vllm serve \
                Qwen/Qwen3-30B-A3B-FP8 \
                --port 7557 \
                --disable-uvicorn-access-log \
                --trust-remote-code \
                --enable-expert-parallel \
                --data-parallel-hybrid-lb \
                --tensor-parallel-size 2 \
                --data-parallel-size 2 \
                --data-parallel-size-local 2 \
                --data-parallel-address localhost \
                --data-parallel-rpc-port 5555 \
                --data-parallel-start-rank 0 \
                --enable-eplb \
                --eplb-config '{"window_size":"1000",
                                "step_interval":"3000",
                                "num_redundant_experts":"32",
                                "log_balancedness":"False"}' \
                --enable-dbo \
                --dbo-decode-token-threshold 32 \
                --max_num_seqs 256

# eval
lm_eval --model local-completions --tasks gsm8k --model_args model=Qwen/Qwen3-30B-A3B-FP8,base_url=http://0.0.0.0:7557/v1/completions,max_retries=3,tokenized_requests=False,timeout=1200,max_gen_toks=64,max_length=8192 --batch_size 2049 --trust_remote_code --limit 0.5
```

`main` (`949a6a19d`):
```
local-completions (model=Qwen/Qwen3-30B-A3B-FP8,base_url=http://0.0.0.0:7557/v1/completions,max_retries=3,tokenized_requests=False,timeout=1200,max_gen_toks=64,max_length=8192,trust_remote_code=True), gen_kwargs: (None), limit: 0.5, num_fewshot: None, batch_size: 2049
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.1955|±  |0.0154|
|     |       |strict-match    |     5|exact_match|↑  |0.1091|±  |0.0121|
```

This PR:
```
local-completions (model=Qwen/Qwen3-30B-A3B-FP8,base_url=http://0.0.0.0:7557/v1/completions,max_retries=3,tokenized_requests=False,timeout=1200,max_gen_toks=64,max_length=8192,trust_remote_code=True), gen_kwargs: (None), limit: 0.5, num_fewshot: None, batch_size: 2049
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.0061|±  | 0.003|
|     |       |strict-match    |     5|exact_match|↑  |0.0000|±  | 0.000|
```

vllm/model_executor/layers/quantization/utils/fp8_utils.py
```diff
@@ -139,6 +139,33 @@ def _padded_cutlass(
...
+    # TODO(torch==2.10): torch semantics broken for PythonMod in 2.9:
+    #  so x<=0 -> x % y <= 0 (not actually python mod semantics).
+    #  https://github.com/pytorch/pytorch/issues/169602
+    #  We increment the padded dim to avoid a negative.
+    # padding = -qx.shape[0] % pad_multiple
+    padding = (qx.shape[0] * pad_multiple - qx.shape[0]) % pad_multiple
+    # print(f"{padding=}")
+    padded_qx = torch.nn.functional.pad(qx, (0, 0, 0, padding))
+    assert len(x_scale.size()) == 2
+    padded_x_scale = torch.nn.functional.pad(x_scale, (0, 0, 0, padding), value=1.0)
```
ElizaWszola, line 161: Looks like `torch.nn.functional.pad` doesn't preserve the col-major layout of scales and the returned tensor is row-major - this leads to bad output of `cutlass_scaled_mm`

ProExpertProg: Nice find!

ElizaWszola: I checked the transpose trick and it works - can you verify?

```suggestion
    padded_x_scale = torch.nn.functional.pad(
        x_scale.transpose(0, 1), (0, padding, 0, 0), value=1.0
    ).transpose(0, 1)
```

Issue #26157, opened 2025-10-03
[Feature]: Make Custom I/O Processing More General


alex-jw-brooks:
### 🚀 The feature, motivation and pitch

Granite Speech models, e.g., [granite-speech-3.3-8b](https://huggingface.co/ibm-granite/granite-speech-3.3-8b) have a two pass design for responding to the content of provided audio, since it first needs to transcribe text. As an example, in order to get the answer `4` from an audio clip with the content `what is 2+2?`, a user would first need to call the model to transcribe to text, and then call it again with the resulting input to get the response `4` out.

This has turned out to be a big point of confusion for people adopting this model, and we would like the ability to abstract the two passes for non-transcription use-cases. This essentially boils down to:
- Calling it normally if there's no audio
- If there's audio, call generate w/ the audio, then pass the result into generate again

With that in mind, I think that the cleanest way to support this would be to make [IO processor plugins](https://github.com/vllm-project/vllm/pull/22820), which are currently only used for pooling models, more general. If we were to take this approach and allow support for IO processor plugins on different model types / entrypoints, this would enable us to potentially write a plugin for granite speech that makes the first call to process the audio from within the plugin, and then pass the result to the main generate call to process the text, which I think should be fine with the current signatures since `pre_process` is already returning a `PromptType`.

@DarkLight1337 @christian-pinto any thoughts on this?

CC @gsaon

### Alternatives

The alternative approach would be to potentially add support for piping the output of one generation call into another, but IMO this is not ideal as it could introduce features that could easily be misunderstood or misused (e.g., for stuff like agent orchestration). Extending the plugin support here would be more sustainable as it would let us make this a lot cleaner on our side while keeping the extra generation logic as an implementation detail of our plugin instead of potentially putting it in core vLLM logic

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

christian-pinto:
In general I would be more than happy to see the IOProcessor plugins generalized to other task such as generation. The one thing to take into account is that, at the moment, the `pre_process` function of the plugin is applied once and it can generate one or more prompts. These prompts are then submitted to vLLM assuming they are independent from each other, meaning they can be batched etc. 

For your use-case you would need to modify how data is fed to vLLM for generation:

[...]

PR #28581, opened 2025-11-12
[Core] Extending I/O Plugins

alex-jw-brooks:
## Purpose
This PR reworks and expands the input/output plugins, which currently are largely used for pooling / embedding models, e.g., [Prithvi-EO-2.0-300M-TL-Sen1Floods11](https://huggingface.co/ibm-nasa-geospatial/Prithvi-EO-2.0-300M-TL-Sen1Floods11) to allow custom processing options to do tasks like semantic segmentation using a more integrated interface with vLLM.

This PR addresses [this](https://github.com/vllm-project/vllm/issues/26157) issue, which aims to extend I/O plugin capabilities to `.generate` and extend the interface so that we can write a plugin to simplify the experience of using granite speech models, which need to call vLLM twice to act as an integrated voice model with the underlying granite LLM, since this is accomplished by intermediate transcription (e.g., audio with `what is 2+2?` -(first call to vllm, which also uses a lora)> `"what is 2+2"` -(second call to vllm, which doesn't use a lora)> `4`). To accomplish this with a minimally invasive change to existing I/O plugins, this PR approaches this with the following objectives:

- I/O plugins should work with `.generate`
- It needs the ability to access more things from the plugin itself, specifically 
     - The plugin needs to be able to submit things to the vLLM engine/client the request
     - The plugin should be able to update the LoRA to be used after the input part of the plugin is used, when the result is used for generation by vLLM
- We should have a clear pattern for wrapping preprocessing capabilities in online serving endpoints that can be invoked from the plugin if needed, since this is what people will often need to do to make their plugins work well with different async endpoints if they want to invoke the client multiple times in one vLLM call
- Plugins should be extended to allow for `input only` & `output only` plugins for convenience (but be input + output by default)
- It should pass what it needs to in order to make plugins flexible and as easy to write and well-patterned as possible against public APIs (as opposed to passing something like the whole LLM entrypoint class, which will likely tempt people with this type of use case to write plugins using internal methods like `LLM._validate_and_add_requests` etc).
- It should be as nonabrasive to existing plugins as possible

### Overview 
Quick summary of state:
- Works for .generate & chat completions
- preprocess is given access to the `lora request`, `engine` / `engine client`, and `processor` so that users can directly submit and await requests to the engine as needed.
    - For preprocessing logic that may be complex for different serving endpoints, e.g., chat templating, the preprocessing logic is wrapped into a method that is passed as a function `partial`, which should have a unified signature across different endpoints. This allows users to be able to call vLLM's preprocessing directly in a way that makes sense, while also ensuring that we have a clear path to enabling it for other endpoints.
    - ^ this is also useful for plugin flexibility, because in some cases like geospatial, we want the plugin to its own preprocessing and skip everything it normally does, but in others like granite speech, we essentially want to the normal preprocessing + extra things.

A full plugin for granite speech can be found [here](https://github.com/alex-jw-brooks/granite_speech_io_processor_plugin) that works for both synchronous generate + async chat completions (still under development and cleaning it up, but most of the hackery is in this plugin, since I tried to keep the vLLM changes as clean as possible for review).

### Changes in Behavior
TODO - there are a few important things to be aware of and caveats for existing plugins, so I will be sure to add an explicit list of default behaviors & how to rectify them.

For the most part, this may require small changes, but should be minimally invasive (geospatial test is currently passing locally with a change to one func signature, which we can also make backwards compatible if we like)

## Test Plan
I will add a test with granite speech's plugin as an example for generate. Geospatial plugin tests should also pass!

## Test Result
Geospatial tests currently pass, granite speech example in the plugin works, but is not added to vLLM tests yet since the plugin is still in a brittle state.

Will revise this and move it out of draft once I have updated docs and solidified potential changes to stuff like output interfaces for the plugin, but opening it in case anyone has thoughts!

CC @DarkLight1337 @njhill @NickLucche @christian-pinto @lastras @gsaon

### DIFF
--- a/vllm/entrypoints/llm.py
+++ b/vllm/entrypoints/llm.py
@@ -427,7 +427,32 @@ class LLM:
             sampling_params = self.get_default_sampling_params()
 
         # Add any modality specific loras to the corresponding prompts
-        lora_request = self._get_modality_specific_lora_reqs(prompts, lora_request)
+        user_lora_request = lora_request
+        lora_request = self._get_modality_specific_lora_reqs(prompts, user_lora_request)
+
+        # TODO - currently generate always applies the io processor, but we should
+        # ensure this behavior is consistent with pooling, which looks for the `data`
+        # key in a dict.
+        if self.io_processor is not None:
+            # Validate the request data is valid for the loaded plugin
+            validated_prompt = self.io_processor.parse_request(prompts)
+
+            # obtain the actual model prompts from the pre-processor
+            prompts = self.io_processor.pre_process(
+                prompt=validated_prompt,
+                params=sampling_params,
+                llm_instance=self,
+                lora_request=lora_request,
+                priority=priority,
+            )
+
+            # reset the lora request unless it's one the user explicitly gave,
+            # because plugins may submit their own requests to the engine,
+            # which may change the modality.
+            if user_lora_request is None:
+                lora_request = self._get_modality_specific_lora_reqs(
+                    prompts, user_lora_request
+                )
 
         self._validate_and_add_requests(
             prompts=prompts,
@@ -438,6 +463,7 @@ class LLM:
         )
 
         outputs = self._run_engine(use_tqdm=use_tqdm)
+        # TODO - handle output side of output plugins
         return self.engine_class.validate_outputs(outputs, RequestOutput)
 
     def _get_modality_specific_lora_reqs(
@@ -458,7 +484,7 @@ class LLM:
         ):
             return lora_request
 
-        if not isinstance(prompts, Sequence):
+        if not isinstance(prompts, Sequence) or isinstance(prompts, str):
             prompts = [prompts]
 
         optional_loras = (
--- a/vllm/entrypoints/llm.py
+++ b/vllm/entrypoints/llm.py
@@ -352,10 +352,22 @@ class LLM:
         return self.llm_engine.get_tokenizer()
 
     def _apply_input_processor_plugin(
-        self, prompts, lora_request, sampling_params, priority
+        self,
+        prompts,
+        lora_request,
+        params,
     ):
         """Apply the input side of the custom IO processor if there is one."""
         if self.io_processor is None:
+            if isinstance(prompts, dict) and "data" in prompts:
+                # Previously this was what we used to distinguish IOProcessor
+                # prompts with pooling, but now we just pass everything to the
+                # input processor if there is one. We warn in this situation
+                # as it's likely a mistake.
+                logger.warning(
+                    "No IOProcessor is installed, but the prompt contains"
+                    " a 'data' key; did you mean to install an IOProcessor?"
+                )
             return (prompts, lora_request)
 
         user_lora_request = lora_request
@@ -370,10 +382,9 @@ class LLM:
         # obtain the actual model prompts from the pre-processor
         prompts = self.io_processor.pre_process(
             prompt=validated_prompt,
-            params=sampling_params,
+            params=params,
             llm_instance=self,
             lora_request=lora_request,
-            priority=priority,
         )
 
         # reset the lora request unless it's one the user explicitly gave,
@@ -390,6 +401,26 @@ class LLM:
         if self.io_processor is None:
             return model_outputs
         return self.io_processor.post_process(model_output=model_outputs)
+
+    def _get_params(self, params):
+        """Applies the custom io processor to get the sampling params.
+
+
+        Args:
+            params: The Pooling or Sampling Params provided by the user,
+                or the default params if None were provided.
+        """
+        if is_list_of(params, PoolingParams) or is_list_of(params, SamplingParams):
+            validated_params = []
+            for param in as_iter(params):
+                validated_params.append(
+                    self.io_processor.validate_or_generate_params(param)
+                )
+            params = validated_params
+        else:
+            assert not isinstance(params, Sequence)
+            params = self.io_processor.validate_or_generate_params(params)
+        return params
 
     @deprecated("`set_tokenizer` is deprecated and will be removed in v0.13.")
     def set_tokenizer(self, tokenizer: AnyTokenizer) -> None:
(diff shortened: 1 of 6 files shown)


vllm/entrypoints/llm.py
```diff
@@ -422,12 +476,22 @@ def generate(
                 "generative model."
             )
 
-        if sampling_params is None:
-            # Use default sampling params.
-            sampling_params = self.get_default_sampling_params()
+        sampling_params = self._get_params(
+            sampling_params
+            if sampling_params is not None
+            else self.get_default_sampling_params
```
christian-pinto, line 482:
```suggestion
            else self.get_default_sampling_params()
```

vllm/entrypoints/llm.py
```diff
@@ -351,6 +352,59 @@ def __init__(
     def get_tokenizer(self) -> AnyTokenizer:
         return self.llm_engine.get_tokenizer()
 
+    def _apply_input_processor_plugin(
+        self,
+        prompts,
+        lora_request,
+        params,
+    ):
```
christian-pinto, line 360: Please use typing for params and return values

vllm/entrypoints/llm.py
```diff
@@ -351,6 +352,59 @@ def __init__(
...
+        """Apply the input side of the custom IO processor if there is one."""
+        if (
+            self.io_processor is None
+            or self.io_processor.plugin_type == IOProcessorPluginType.OUTPUT_ONLY
+        ):
+            return (prompts, lora_request)
```
christian-pinto, line 366: I would personally have this check in `generate()`. Otherwise this gives the impression that all requests pass via the plugin, which they don't.

Another thing is about applying the processor to any request if a plugin is installed. Previously, in `encode()` we would check for the structure of the input and infer if applying a plugin is needed. This makes it possible for the same vLLM instance to serve the same model with and without plugin. 

Is the behavior here intentional? I am fine with it, as long as we document it well for people to realize that if they load a plugin then everything goes through the plugin.

vllm/entrypoints/llm.py
```diff
@@ -438,7 +502,31 @@ def generate(
...
+        model_outputs = self.engine_class.validate_outputs(outputs, RequestOutput)
+
+        if (
+            self.io_processor is not None
+            and self.io_processor.plugin_type != IOProcessorPluginType.INPUT_ONLY
+        ):
+            # get the post-processed model outputs
+            processed_outputs = self.io_processor.post_process(
+                model_output=model_outputs
+            )
```
christian-pinto, line 514: I would do exactly the same for the pre-processing part, figuring out if the plugin should be applied in this function.

vllm/plugins/io_processors/interface.py
```diff
@@ -37,14 +51,19 @@ async def pre_process_async(
     ) -> PromptType | Sequence[PromptType]:
         return self.pre_process(prompt, request_id, **kwargs)
 
-    @abstractmethod
     def post_process(
         self,
         model_output: Sequence[PoolingRequestOutput],
```
christian-pinto, line 56:
```suggestion
        model_output: Union(Sequence[RequestOutput], Sequence[PoolingRequestOutput]),
```

same for the async one

vllm/entrypoints/openai/serving_chat.py
```diff
@@ -218,48 +223,45 @@ async def create_chat_completion(
...
+            if (
+                self.io_processor is not None
+                and self.io_processor.plugin_type != IOProcessorPluginType.OUTPUT_ONLY
+            ):
                 (
                     conversation,
                     request_prompts,
                     engine_prompts,
-                ) = await self._preprocess_chat(
+                ) = await self._apply_input_processor_plugin(
```
christian-pinto, line 241: `_apply_input_processor_plugin` is not an async function

vllm/entrypoints/openai/serving_chat.py
```diff
@@ -282,14 +284,23 @@ async def create_chat_completion(
 
                 sampling_params: SamplingParams | BeamSearchParams
                 if request.use_beam_search:
+                    if self.io_processor is not None:
+                        logger.warning(
+                            "IO Processor Plugins do not currently support beam search"
+                        )
```
christian-pinto, line 290: Doesn't seem to be relevant here?

vllm/entrypoints/openai/serving_engine.py
```diff
@@ -1176,6 +1176,48 @@ async def _process_inputs(
         )
         return engine_request, tokenization_kwargs
 
+    async def _apply_input_processor_plugin(
+        self,
+        request,
+        request_id,
+        lora_request,
```
christian-pinto, line 1183: The sync version of this function (in llm.py) checks if the IO plugin is to be applied and is called on each generate/encode. While for the async versions we check in `serving_x` if there is an IO processor and apply only if true.

This should be standardized across all the use-cases (llm vs serving)

vllm/plugins/io_processors/interface.py
```diff
@@ -37,14 +51,19 @@ async def pre_process_async(
...
         **kwargs,
     ) -> IOProcessorOutput:
-        raise NotImplementedError
+        if self.plugin_type != IOProcessorPluginType.INPUT_ONLY:
+            raise NotImplementedError(
+                ".post_process() is not implemented for this IO plugin"
+            )
+        raise AssertionError(
+            ".post_process() should not be invoked for input only plugins"
+        )
```
christian-pinto, line 67: I assume this is not abstract anymore to give the hint that this should not be applied if the plugin is INPUT_ONLY. However, since in the code we always check the IO processor is available and its type  before invoking, this additional check here seems unnecessary.

I would just revert to the abstract implementation.

Same comment applies to the other methods in the interface.

christian-pinto (commented):
I have done a first pass. The changes are mostly ok but I would standardize the way we check whether to apply the IO processor throughout the whole code -- i.e., in llm vs serving we have a different approach for when we apply the pre-processor. 

Still in LLM for the preprocessor we have a function applied to all requests that then checks if an IO processor is to be used, while for the  post_process we always check before applying the post processing function. 


It would be nice to see an example with a plugin applied to the speech use case.

vllm/entrypoints/openai/serving_chat.py
alex-jw-brooks: This one is, since this subclass inherits from `OpenAIServing`. The definition is here: https://github.com/vllm-project/vllm/blob/d6d93da342497d310d0b6a66346f897759180db7/vllm/entrypoints/openai/serving_engine.py#L1179

hmellor:
Closing as obsolete. The issue this implements, #26157, was closed as not planned on 2026-02-02, the interface went the other way in #34236, and all three serving files this patches have since been deleted.

Issue #25700, opened 2025-09-25
[RFC]: Limit the use of envvars in vLLM


zhuohan123:
### Motivation.

vLLM's envvars are on the edge of getting out of control. There are many envvars should instead be configs. Like the attention backend, all2all kernel backend, and even there is a flag to control KV cache layout. envvars are evil because:

1. It is equivalent of using global variables everywhere in the code, which is a bad programming practice.
2. envvars have no advanced structure like, hierarchy, typechecks, etc. 

In summary, I think envvars are the kind of thing that is very easy to add so people tend to add envvars as the shortest path to implement their feature, but this can quickly becomes unmanageable and makes the project very hard to use. 

### Proposed Change.

I think we should:

1. Spend some effort on reviewing the current envvars and move many of them to configs.
2. Have very strict bar on what can be an envvar and question every new envvars in vllm.

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

DarkLight1337:
Related to https://github.com/vllm-project/vllm/issues/22914

ProExpertProg:
I agree we should reduce our reliance on environment variables. I see a couple of reasonable uses for them:
- debugging probes and other properties we want to be able to control directly from the outside without changing code or access to the actual vllm command that's running, for example:
  - `VLLM_LOGGING_LEVEL`
  - `VLLM_CACHE_DIR`
  - `VLLM_DEBUG_DUMP_PATH` (in #25651)
- temporary switches for unstable experimental features

For the ones that do stay around, we should improve their handling. I sent this in the #contributors channel a while ago, I'm happy to create a separate RFC.

I'm hoping to get some preliminary feedback on env variable restructuring. Instead of duplicating the type annotations and getter dictionary, I propose the following:
- Use the type-hinted declaration as the single source of truth, including defaults and custom parsing, and move them into envs/_variables.py
- Import those directly if `TYPE_CHECKING`, otherwise use them as a default store
- Allow defaults
- Standardize argument parsing for `str`/`int`/`float`/`bool`/`Path`/`list[str]` "trivial" types
- Add a pre-commit check to validate _variables.py and that they are never imported directly

<details>
<summary> Skeleton for `vllm/envs/__init__.py`: </summary>

```
from typing import TYPE_CHECKING, get_type_hints

import envs._variables
[...]

PR #28582, opened 2025-11-12
[Feature] Conditional prefix logging for stdout/stderr

NargiT:
~~Add conditional logging prefixing based on `VLLM_ENABLE_PREFIX_PROC_LOGS` environment variable.~~

**Update:** per [@markmc's suggestion](https://github.com/vllm-project/vllm/pull/28582#issuecomment-3526137301) (see also #25700, #28666), this now uses an explicit `--log-prefix` / `--no-log-prefix` CLI argument and `ObservabilityConfig.log_prefix` config setting instead of an environment variable.

## Suggestion
Python's logging module already exposes `processName` as a standard LogRecord attribute `(%(processName)s)`, giving users full control over format without custom code. 

Separately: in enterprise environments, logs need to be parseable — no ANSI colors (breaks parsers/log aggregators), and a consistent format across all services.

__add_prefix works against both: it hardcodes format/color outside the standard Formatter pipeline, so users can't override or reposition it, and it likely forces colors even in non-TTY/production contexts.

**Basic example**: Companies using ELK stack, must update their parsers just for vLLM. 

I would recommend to deprecated this feature and instead encourage people to use logging format. 
- https://docs.python.org/3/library/logging.html#logging.LogRecord
- https://docs.vllm.ai/en/stable/examples/others/logging_configuration.html?h=logging


## Purpose
Allow full control over log format

## Test Plan

~~VLLM_ENABLE_PREFIX_PROC_LOGS=0 vllm serve [...]~~
`vllm serve [...] --no-log-prefix`

## Test Result
```bash
INFO:     127.0.0.6:56097 - "GET /health HTTP/1.1" 200 OK                                                                                                                                                                                                                                                                                                              
# instead of
(APIServer pid=1) INFO:     127.0.0.6:56097 - "GET /health HTTP/1.1" 200 OK                                                                                                                                                                                                                                                                                              
# with processName
INFO:     127.0.0.6:56097 - MainProcess "GET /health HTTP/1.1" 200 OK                                                                                                                                                                                                                                                                                              
# Worker -> VllmWorker-0
# EngineCore_DP0 -> EngineCore_DP0
```
---
<details>

</details>


markmc:
See also #28666

njhill:
The main reason for doing the stdout/err prefix this way is to cover output that vllm itself isn't logging (e.g. in each worker process), from cuda / torch / flash infer / other libraries.

Having an option to disable this though sounds like a good idea.

NargiT:
Sorry I don't have any more time for the extra mile here.

NargiT:
Hello, I will fix this.

NargiT:
After @markmc  suggestion, I switch environment based configuration to a cli arguments. 
Looks like one of you needs to add a label in order for pre-commit action runs.

PR #28585, opened 2025-11-12
[KV Connector] Fix async connector prefix cache metrics

markmc:
Currently we are recording async connector prefix cache queries and hits twice, for the same reason that `update_state_after_alloc()` is called twice:

> If get_num_new_matched_tokens previously returned True for a
> request, this function may be called twice for that same request -
> first when blocks are allocated for the connector tokens to be
> asynchronously loaded into, and second when any additional blocks
> are allocated, after the load/transfer is complete.

Worse, the second time we are recording with `num_external_computed_tokens=0` so effectively we are halving the hit rate.

Before

```
External prefix cache hit rate: 100.0%
```

After

```
External prefix cache hit rate: 50.0%
```

Borrows part of #27569 to track `num_external_computed_tokens` for use when the KV transfer completes.

Will use #28550 for testing this scenario.

### DIFF
--- a/tests/v1/core/test_scheduler.py
+++ b/tests/v1/core/test_scheduler.py
@@ -1057,7 +1057,8 @@ def test_kv_connector_basic(is_async: bool):
     )
 
 
-def test_external_prefix_cache_metrics():
+@pytest.mark.parametrize("is_async", [False, True])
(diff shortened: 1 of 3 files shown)


robertgshaw2-redhat (approved):
nice fix!

PR #28587, opened 2025-11-12
[Hybrid][torch.compile] Refactor mamba2 forward to avoid obscuring linear projections under custom op

tomeras91:
## Purpose

Currently, the mamba_mixer2 custom op encapsulates the entire forward pass of the Mamba2 block. This is including: the input projection, the token mixing conv + SSM ops, the gated RMSnorm, and the output proj. 

The reason the custom op exists is to enable piecewise CUDA graphs - the token mixing ops aren't easily CUDA-graphable for a prefill/mixed batch, due input dependent control flow (cu_seqlens for example). This is similar to why we want to break CUDA graphs on Attention ops. Yet, everything except the token mixing ops can be moved out of the custom op since it doesn't have input dependent control flow. This is similar to how the QKV and output projections in Attention are **not** part of the Attention custom op. Rather, the custom op performs just the token mixing Attention op itself. Thus, **this PR makes the Mamba2 forward much more similar to how the Attention forward is implemented.**

The motivation for this change is to allow torch.compile to run on these projections and RMSnorm. These, especially the RMSnorm, are implemented using many torch ops In the default flow (i.e. without activating the `mixer2_gated_rms_norm` custom op). When they are part of the custom op, they are obscured from torch.compile and many simple kernel fusion opportunities are missed. By moving them out of the custom op, we allow Inductor to fuse them.

This is even more important when running quantized models where the RMSnorm can be fused with the quantization op before the projections. I this case, even when enabling the `mixer2_gated_rms_norm` and the `quant` custom ops we don't get the norm + quant fusion since there is no custom pass replacing the `mixer2_gated_rms_norm` -> `quant` pattern with a fused custom kernel.

## Performance effect

This PR boosts overall performance of hybrid models significantly. For [nvidia/NVIDIA-Nemotron-Nano-9B-v2-FP8](https://huggingface.co/nvidia/NVIDIA-Nemotron-Nano-9B-v2-FP8) with ISL/OSL/BS = 1K/1K/256 on a single H100, we observed 10%-15% increase in throughput, and ~22% decrease in TTFT:

```
========================== Serving Benchmark Result Comparison ===========================
Metric                                    Main            PR              Diff (%)
--------------------------------------------------------------------------------------------
Successful requests:                     1280            1280            +0.00%
Failed requests:                         0               0               +0.00%
Maximum request concurrency:             256             256             +0.00%
Benchmark duration (s):                  226.32          201.32          -11.04%
Total input tokens:                      1309444         1309444         +0.00%
Total generated tokens:                  1310720         1310720         +0.00%
Request throughput (req/s):              5.66            6.36            +12.37%
Output token throughput (tok/s):         5791.47         6510.57         +12.42%
Peak output token throughput (tok/s):    7936.00         8704.00         +9.68%
Peak concurrent requests:                288.00          295.00          +2.43%
Total Token throughput (tok/s):          11577.31        13014.81        +12.42%
-------------------------------Time to First Token---------------------------------------
Mean TTFT (ms):                          2021.85         1576.73         -22.02%
Median TTFT (ms):                        1077.14         843.59          -21.69%
P99 TTFT (ms):                           9803.03         7490.96         -23.59%
-------------------------------Time per Output Token (excl. 1st token)-------------------
Mean TPOT (ms):                          42.21           37.77           -10.52%
Median TPOT (ms):                        42.20           37.81           -10.40%
P99 TPOT (ms):                           45.55           40.55           -10.98%
-------------------------------Inter-token Latency---------------------------------------
Mean ITL (ms):                           42.21           37.77           -10.52%
Median ITL (ms):                         33.00           30.74           -6.85%
P99 ITL (ms):                            349.69          273.94          -21.67%
-------------------------------End-to-end Latency----------------------------------------
Mean E2EL (ms):                          45204.97        40210.64        -11.05%
Median E2EL (ms):                        44381.41        39479.33        -11.05%
P99 E2EL (ms):                           55504.54        48002.33        -13.52%
=========================================================================================
``` 

<details>
<summary> Benchmark details </summary>

Server run command:
```
vllm serve <model> --trust-remote-code --no-enable-prefix-caching --async-scheduling --max-num-seqs 2048
```

Bench command:
```
vllm bench serve --backend vllm --model <model> --tokenizer <model> --trust-remote-code \
  --num-prompts 1280 \
  --ignore-eos \
  --num-warmups 20 \
  --max-concurrency 256 \
  --random-input-len 1024 \
  --random-output-len 1024 \
  --no-stream \
  --dataset-name random \
  --save-result
```
</details>

The following screenshots, taken from nsys profiles from the benchmark above (decode forward pass), show that the RMS + quant ops after the SSM update and before the output proj are indeed fused now, reducing the kernel count from 15 to 2, and taking ~14.5 microsec instead of ~91.8 microsec:

Profile on `main`:
main

Profile on PR:
pr

## Test Plan

Validate quality of hybrid models hasn't changed, by running GSM8K eval.

Run command:

```
lm_eval --model vllm --model_args pretrained=<model> --batch_size auto --trust_remote_code --cache_requests true --tasks gsm8k
```

### model: [ibm-granite/granite-4.0-h-tiny](https://huggingface.co/ibm-granite/granite-4.0-h-tiny)

results on `main`:
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.7961|±  |0.0111|
|     |       |strict-match    |     5|exact_match|↑  |0.7953|±  |0.0111|

results on PR:
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8044|±  |0.0109|
|     |       |strict-match    |     5|exact_match|↑  |0.8036|±  |0.0109|

### model: [nvidia/NVIDIA-Nemotron-Nano-9B-v2](https://huggingface.co/nvidia/NVIDIA-Nemotron-Nano-9B-v2)

results on `main`:
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.7983|±  |0.0111|
|     |       |strict-match    |     5|exact_match|↑  |0.8650|±  |0.0094|

results on PR:
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.7961|±  |0.0111|
|     |       |strict-match    |     5|exact_match|↑  |0.8635|±  |0.0095|

**Results are identical up to errorbar**


vllm/model_executor/layers/mamba/mamba_mixer2.py
```diff
@@ -524,46 +576,13 @@ def forward_cuda(
...
             # profile run
             hidden_states_B_C = (
                 hidden_states_B_C.transpose(0, 1).clone().transpose(0, 1)
             ).contiguous()
-            hidden_states, _B, _C = split_hidden_states_B_C_fn(hidden_states_B_C)
-            hidden_states = self.norm(hidden_states, gate)
-            out, _ = self.out_proj(hidden_states)
-            return out
+            hidden_states, _B, _C = self.split_hidden_states_B_C_fn(hidden_states_B_C)
+            return hidden_states
```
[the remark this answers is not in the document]
tomeras91: This is similar to how it's done in other custom ops

tdoublep:
Wow!! This look amazing. Will review carefully asap

vllm/model_executor/layers/mamba/mamba_mixer2.py
```diff
@@ -473,29 +488,67 @@ def __init__(
...
-        output: torch.Tensor,
         mup_vector: torch.Tensor | None = None,
     ):
+        # 1. Gated MLP's linear projection
+        projected_states, _ = self.in_proj(hidden_states)
+        if mup_vector is not None:
+            projected_states = projected_states * mup_vector
+
+        # 2. Prepare inputs for conv + SSM
+        gate_size = self.intermediate_size // self.tp_size
```
amirkl94: A bit of nitpicking: I think you can move these definitions to `__init__`, then you can use `gate_size` in the definition of `split_hidden_states_B_C_fn` which will make the code a bit more readable.

tomeras91: Done. Thanks :)

ZJY0516:
LGTM

vllm/model_executor/layers/mamba/mamba_mixer2.py
```diff
@@ -901,21 +917,21 @@ def get_attn_backend(self) -> type["AttentionBackend"]:
 
 
 def mamba_mixer2(
```
tdoublep, line 917: I wonder if we should rename the op to reflect that it is no longer doing the whole mamba_mixer2 layer? I'm tempted to say let's leave it for now. I think the right approach is to try and refactor things a bit to have a `unified_mamba` op kind of like how we have it for attention (as a follow-up).

tdoublep (approved):
Thank you for this great contribution.

I ran a quick benchmark myself for granite 4.0. 

Server (1x H100):
```
vllm serve ibm-granite/granite-4.0-h-small-fp8
```

Client:
```
vllm bench serve \
        --model ibm-granite/granite-4.0-h-small-fp8 \
        --dataset-name sharegpt \
        --ignore_eos
```

Results (second run) on main:
```
Successful requests:                     998       
Failed requests:                         2         
Benchmark duration (s):                  86.00     
Total input tokens:                      214249    
Total generated tokens:                  199536    
Request throughput (req/s):              11.60     
Output token throughput (tok/s):         2320.12   
Peak output token throughput (tok/s):    3200.00   
Peak concurrent requests:                998.00    
Total Token throughput (tok/s):          4811.32   
Mean TTFT (ms):                          33351.37  
Median TTFT (ms):                        33633.66  
P99 TTFT (ms):                           70589.64  
Mean TPOT (ms):                          50.69     
Median TPOT (ms):                        50.04     
P99 TPOT (ms):                           99.75     
Mean ITL (ms):                           47.87     
Median ITL (ms):                         37.87     
P99 ITL (ms):                            94.47     
```

Results (second run) from this PR:
```
Successful requests:                     998       
Failed requests:                         2         
Benchmark duration (s):                  52.10     
Total input tokens:                      214223    
Total generated tokens:                  199066    
Request throughput (req/s):              19.16     
Output token throughput (tok/s):         3821.11   
Peak output token throughput (tok/s):    5816.00   
Peak concurrent requests:                998.00    
Total Token throughput (tok/s):          7933.17   
Mean TTFT (ms):                          10330.10  
Median TTFT (ms):                        5983.26   
P99 TTFT (ms):                           29918.19  
Mean TPOT (ms):                          102.28    
Median TPOT (ms):                        90.12     
P99 TPOT (ms):                           373.78    
Mean ITL (ms):                           77.62     
Median ITL (ms):                         76.35     
P99 ITL (ms):                            286.91    
```

The TTFT is reduced by 3x, and the throughput improved by 66%. There is an increase in the ITL but I can see it is because we are able to drive much larger decode batch sizes in the benchmark whereas on main it is all bottlenecked by prefill.

vllm/model_executor/layers/mamba/mamba_mixer2.py
tomeras91: I was wondering about that too. Decided to leave as is since I wanted to limit the scope of the PR, didn't want to change default graph splitting ops, etc. I did change the name of the function under the custom op though.. I think now `forward_cuda` is really misleading so I changed it to `conv_ssm_forward`

mgoin (approved):

mgoin:
The CI failure seems related, PTAL
```
[2025-11-16T21:42:14Z] (EngineCore_DP0 pid=2149) AssertionError: expected size 4096==4096, stride 1088==1048 at dim=0
[2025-11-16T21:42:14Z] (EngineCore_DP0 pid=2149) This error most often comes from a incorrect fake (aka meta) kernel for a custom op.
```

ProExpertProg (approved):
Nice improvement! In the future, we can also extract the prefill-decode split using data-dependent shapes, like proposed for MLA in #26516

mgoin (approved):

PR #17320, opened 2025-04-28
[Build][Bugfix] Restrict setuptools version to <80

gshtras:
Setuptools appears to have introduced a bug in v80.0.0 resulting in a crash when trying to run `python setup.py develop` with a `--prefix`
Original issue filed at https://github.com/pypa/setuptools/issues/4966
Until it is addressed we should restrict the version
This is a generalization of https://github.com/vllm-project/vllm/pull/17298

Another issue is that v80 will replace `python setup.py develop` with `pip install -e .` which does not work with ROCm, in part due to torch requirement hard coded in pyproject.toml

PR #28592, opened 2025-11-12
[ROCm][Bugfix] Revert removing setuptools version restriction

gshtras:
Reverting #28337 on ROCm specific files until a proper solution is found. Other platforms might suffer from a similar issue

The underlying issue mentioned in #17320 is still not fixed properly.
To elaborate, setuptools>=80 will replace `python setup.py develop` with `pip install -e .`, which creates an isolateb build environment and pulls build dependencies from pyprojec.toml. In particulat it would try to install torch==2.9.0, which is a CUDA version of torch package by default, so the ROCm build would fail.
The only possible way to build currently is with bdist_wheel, which doesn't allow creating a development environment

### DIFF
--- a/requirements/rocm-build.txt
+++ b/requirements/rocm-build.txt
@@ -9,7 +9,7 @@ torchaudio==2.9.0
 triton==3.5.0
 cmake>=3.26.1,<4
 packaging>=24.2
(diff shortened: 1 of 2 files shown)


tjtanaa (approved):
LGTM

Issue #26191, opened 2025-10-03
[Bug]: Performance Regression in Acceptance length for EAGLE3


tomasruizt:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
Collecting environment information...
==============================
        System Info
==============================
OS                           : Ubuntu 22.04.4 LTS (x86_64)
GCC version                  : (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Clang version                : Could not collect
CMake version                : version 3.26.4
Libc version                 : glibc-2.35

==============================
       PyTorch Info
==============================
PyTorch version              : 2.8.0+cu128
Is debug build               : False
CUDA used to build PyTorch   : 12.8
ROCM used to build PyTorch   : N/A

==============================
      Python Environment
==============================
Python version               : 3.11.9 (main, Apr 19 2024, 16:48:06) [GCC 11.2.0] (64-bit runtime)
Python platform              : Linux-5.15.0-153-generic-x86_64-with-glibc2.35

==============================
       CUDA / GPU Info
==============================
Is CUDA available            : True
CUDA runtime version         : 12.4.131
CUDA_MODULE_LOADING set to   : LAZY
GPU models and configuration : GPU 0: NVIDIA H100
Nvidia driver version        : 535.261.03
cuDNN version                : Could not collect
HIP runtime version          : N/A
MIOpen runtime version       : N/A
Is XNNPACK available         : True

==============================
          CPU Info
==============================
Architecture:                            x86_64
CPU op-mode(s):                          32-bit, 64-bit
Address sizes:                           52 bits physical, 57 bits virtual
Byte Order:                              Little Endian
CPU(s):                                  96
On-line CPU(s) list:                     0-95
Vendor ID:                               AuthenticAMD
Model name:                              AMD EPYC 9254 24-Core Processor
CPU family:                              25
Model:                                   17
Thread(s) per core:                      2
Core(s) per socket:                      24
Socket(s):                               2
Stepping:                                1
Frequency boost:                         enabled
CPU max MHz:                             2900.0000
CPU min MHz:                             1500.0000
BogoMIPS:                                5800.48
[...]

PR #28597, opened 2025-11-12
[Perf] Optimize EAGLE prepare_inputs_padded with triton kernels

benchislett:
## Purpose

Part 1 of a series of planned efforts to reduce the kernel call overhead in the EAGLE padded speculation pathway.

This PR tackles `next_token_ids_padded` and `prepare_inputs_padded` with a triton kernel for each that handles the GPU logic. The fused kernels are simple and cheap to run, reducing the overhead in these sections by over 10x.

This PR also inlines the positions slicing and removes `token_indices` from `prepare_inputs_padded`, since it is always an `arange` which can be more efficiently represented using an integer slice on the indexed tensor.

## Test Plan

Previous EAGLE tests should fully cover both functions already.

## Performance Validation

I constructed an artificial scenario by modifying the tests with an NVTX range and a spin-loop that allows the CPU to run ahead, as a realistic proxy for the model execution. Here is the measured overhead in each case. Note that the time scale in the screenshots (x axis) is not normalized.

See early commits on this PR for the setup for this testing.

### Next token ids (reference)
About **50us**
image

### Next token ids (kernel)
Less than **5us (10x)**
image

### Prepare inputs (reference)
About **40us**
image

### Prepare inputs (kernel)
Less than **2us (20x)**
image


vllm/v1/spec_decode/eagle.py
```diff
@@ -530,59 +535,57 @@ def prepare_next_token_ids_padded(
...
-        It also accounts for the rejected tokens in `sampled_token_ids`.
-        This function must use device functions to operate on the inputs, and
-        should not introduce any blocking CPU-GPU synchronization.
+        is not sampled and comes from `request.get_token_id()` instead. This is denoted
+        the "backup" token id. It also counts rejected tokens via `sampled_token_ids`.
         """
-        # TODO(Ben): Combine this into a custom fused kernel
-
-        # Precompute get_token_id for when there is no valid next token
         num_reqs = gpu_input_batch.num_reqs
```
benchislett, line 573:
```suggestion
        # Precompute get_token_id for when there is no valid next token
        num_reqs = gpu_input_batch.num_reqs
```

vllm/v1/spec_decode/utils.py
```diff
@@ -14,3 +15,122 @@ def is_spec_decode_unsupported(sampling_params: SamplingParams) -> bool:
...
+        num_draft_tokens = cu_draft_curr - cu_draft_prev
+
+    valid_count = tl.load(valid_sampled_tokens_count_ptr + req_idx)
+    num_rejected_tokens = num_draft_tokens + 1 - valid_count
+
+    # query_start_loc[req_idx + 1] is the start position of the next request,
+    # which is one past the last token of this request.
+    q_last_tok_idx = tl.load(query_start_loc_gpu_ptr + req_idx + 1) - 1
+
+    index_to_sample = q_last_tok_idx - num_rejected_tokens
```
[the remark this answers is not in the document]
benchislett: Ah, I really thought about this and thought it was fine, but this is probably a realistic edge-case.

vllm/v1/spec_decode/utils.py
```diff
@@ -14,3 +15,122 @@ def is_spec_decode_unsupported(sampling_params: SamplingParams) -> bool:
...
+            mask=mask,
+            other=-1,
+        )
+        # Check if req_idx is in this block
+        if tl.sum(tl.where(discarded_idx_block == req_idx_64, 1, 0)) > 0:
+            is_discarded = True
+
+    if is_discarded:
+        backup_token = tl.load(backup_next_token_ids_ptr + req_idx)
+        valid_count = tl.full((), 0, dtype=tl.uint32)
```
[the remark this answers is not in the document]
benchislett: this is wrong. it's unsigned in both branches.

vllm/v1/spec_decode/eagle.py
```diff
@@ -545,44 +546,47 @@ def prepare_next_token_ids_padded(
...
-            :num_discarded_requests
-        ]
+        batch_size, num_tokens = sampled_token_ids.shape
+        device = sampled_token_ids.device
 
-        valid_sampled_token_ids_gpu = sampled_token_ids.clone()
-        valid_sampled_token_ids_gpu.index_fill_(
-            0, discard_sampled_tokens_req_indices, -1
+        assert discard_request_indices.dtype == torch.int64, (
+            "discard_request_indices must be torch.int64"
```
LucasWilkinson: nit: I think the assert is self explanatory enough that the additional message is not needed

vllm/v1/spec_decode/eagle.py
```diff
@@ -545,44 +546,47 @@ def prepare_next_token_ids_padded(
...
         )
-
-        # Generate a mask for all valid tokens within those requests
-        valid_mask = (valid_sampled_token_ids_gpu != -1) & (
-            valid_sampled_token_ids_gpu < gpu_input_batch.vocab_size
+        assert backup_tokens_gpu.dtype == torch.int32, (
```
LucasWilkinson: nit: I think the assert is self explanatory enough that the additional message is not needed

vllm/v1/spec_decode/utils.py
```diff
@@ -14,3 +15,123 @@ def is_spec_decode_unsupported(sampling_params: SamplingParams) -> bool:
...
+    for i in range(0, tl.cdiv(num_discarded_requests, BLOCK_SIZE_DISCARD)):
+        mask = (discard_offs + i * BLOCK_SIZE_DISCARD) < num_discarded_requests
+        discarded_idx_block = tl.load(
+            discard_request_indices_ptr + discard_offs + i * BLOCK_SIZE_DISCARD,
+            mask=mask,
+            other=-1,
+        )
+        # Check if req_idx is in this block
+        if tl.sum(tl.where(discarded_idx_block == req_idx_64, 1, 0)) > 0:
+            is_discarded = True
```
LucasWilkinson: Do you think it makes sense to just stop `discard_requests_mask` instead of `discard_request_indices` and use that instead? 
and then move 
```
discard_request_indices = np.nonzero(discard_requests_mask)[0]
```
into the bookkeeping sync?

I think that would simplify this code quite a bit

LucasWilkinson (commented):
Overall looks good to me; left a couple comments 

(side note: I think we should consider removing `disable_padded_drafter_batch` soonish now that its diverging more and its pretty robustly tested at this point)

LucasWilkinson (approved):

vllm/v1/spec_decode/utils.py
benchislett: You're right, that does seem cleaner here

LucasWilkinson:
LGTM! Thanks for the updates

> @LucasWilkinson we need to address https://github.com/vllm-project/vllm/pull/26498 (https://github.com/vllm-project/vllm/issues/26191) before removing the non-padded pathway.

Makes sense, lets revisit after that is done  :+1:

PR #28599, opened 2025-11-12
[BugFix] Fix `mm_encoder_attn_backend` arg type checking

njhill:
Currently, use of `--mm-encoder-attn-backend` fails with the following error:

```
self = typing.Any, obj = 'FLASH_ATTN'

    def __instancecheck__(self, obj):
        if self is Any:
>           raise TypeError("typing.Any cannot be used with isinstance()")
E           TypeError: typing.Any cannot be used with isinstance()

../../../.local/share/uv/python/cpython-3.12.12-linux-x86_64-gnu/lib/python3.12/typing.py:530: TypeError
```

There is a [test](https://github.com/vllm-project/vllm/blob/main/tests/config/test_multimodal_config.py) for this but the `config` tests aren't referenced in `test-pipeline.yaml` and so weren't actually running in the CI!

cc @ywang96

### DIFF
--- a/.buildkite/test-pipeline.yaml
+++ b/.buildkite/test-pipeline.yaml
@@ -57,7 +57,7 @@ steps:
   - pytest -v -s -m 'not cpu_test' multimodal
   - pytest -v -s utils_
 
-- label: Async Engine, Inputs, Utils, Worker Test (CPU) # 4 mins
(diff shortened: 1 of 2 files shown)


ywang96 (approved):

njhill:
Looks like this was probably broken by https://github.com/vllm-project/vllm/pull/24794.

hmellor (approved):

Issue #27851, opened 2025-10-31
[Installation]: xformers issue


seindum:
### Your current environment

```text
The output of `python collect_env.py`
```

### How you are installing vllm

× No solution found when resolving dependencies:
  ╰─▶ Because there is no version of xformers{platform_machine ==
      'x86_64' and sys_platform == 'linux'}==0.0.33+5d4b92a5.d20251029
      and vllm==0.11.1rc6.dev12+gb2e65cb4a.precompiled depends on
      xformers{platform_machine == 'x86_64' and sys_platform ==
      'linux'}==0.0.33+5d4b92a5.d20251029, we can conclude that
      vllm==0.11.1rc6.dev12+gb2e65cb4a.precompiled cannot be used.
      And because only vllm==0.11.1rc6.dev12+gb2e65cb4a.precompiled is
      available and you require vllm, we can conclude that your requirements
      are unsatisfiable.

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

0xC45:
I am facing this same issue when trying to build from source.
```
$ uv venv --python 3.12 --seed

$ source .venv/bin/activate
Using CPython 3.12.3 interpreter at: /usr/bin/python3.12
Creating virtual environment with seed packages at: .venv
 + pip==25.3
Activate with: source .venv/bin/activate

$ uv pip install -e .
  × No solution found when resolving dependencies:
  ╰─▶ Because there is no version of xformers{platform_machine == 'x86_64' and sys_platform == 'linux'}==0.0.33+5d4b92a5.d20251029 and vllm==0.11.1rc6.dev12+gb2e65cb4a.d20251031.cu130 depends on xformers{platform_machine == 'x86_64' and
      sys_platform == 'linux'}==0.0.33+5d4b92a5.d20251029, we can conclude that vllm==0.11.1rc6.dev12+gb2e65cb4a.d20251031.cu130 cannot be used.
      And because only vllm==0.11.1rc6.dev12+gb2e65cb4a.d20251031.cu130 is available and you require vllm, we can conclude that your requirements are unsatisfiable.
```

Seems like #27768 could potentially be related? CC @huydhn

seindum:
[...]

PR #28600, opened 2025-11-12
Use official xformers-0.0.33 built for PT 2.9

huydhn:
## Purpose

`xformers-0.0.33` has been released today for PyTorch 2.9 https://pypi.org/project/xformers/#history, and we can finally switch to that version instead of the custom one we build on PyTorch

FIX https://github.com/vllm-project/vllm/issues/27880 https://github.com/vllm-project/vllm/issues/27851
## Test Plan

CI

cc @simon-mo @ywang96 @ProExpertProg


ProExpertProg (approved):

EmilienM (approved):

houseroad (approved):
finally :-)

mgoin (approved):

ywang96 (approved):
I think this PR is now good to be force-merged.  All failed tests are not related to this dependency change but we should address them separately.

PR #28611, opened 2025-11-13
[XPU] add sym params to IPEXConfig

zufangzhu:
This PR adds a sym parameter to the configuration function for IPEX WeightOnlyQuantizedLinear.
The parameter allows explicit control over whether symmetric or asymmetric quantization is used.

To support run GPTQ model with asym quant.
Command: python examples/offline_inference/basic/generate.py --enforce-eager --model  dj271/meta-llama-Llama-3.1-8B-Instruct-gptq-4bit-asym  --max-model-len 1024 --max-num-batched-tokens 2048 --dtype float16 --temperature 0 -tp 2

Before:
image

After:
image


jikunshang (approved):
LGTM. Can you add some examples/reproducer in description?

PR #28619, opened 2025-11-13
[Bugfix] fix IMA issue in certain cases of the moe marlin kernel

### LINKED ISSUE #28220: [Bug]: Find the root cause of `apply_repetition_penalties` fails from #28180

jinzhen-lin:
fix https://github.com/vllm-project/vllm/issues/28220


youkaichao:
cc @mgoin @vadiklyutiy

vadiklyutiy:
So, the illegal memory access appeared when we read some garbage from `x=sh_block_topk_weights[0]` and then use `x` as index in `topk_weights_ptr[x]`?

vadiklyutiy (approved):

youkaichao (approved):

vllm/model_executor/layers/fused_moe/shared_fused_moe.py
```diff
@@ -38,7 +38,6 @@ def __init__(
                 # TODO(wentao): find the root cause and remove this condition
                 self.enable_eplb
                 or (self.moe_config.use_flashinfer_cutlass_kernels and self.dp_size > 1)
-                or self.use_marlin_kernels
```
yewentao256, line 41: Are we sure this is well tested?
Could you:
1. test with a smaller batch size and get the profile. `vllm bench serve --model deepseek-ai/DeepSeek-V3.1  --dataset-name random --host 127.0.0.1 --port 9256 --random-input-len 2 --random-output-len 10 --request-rate inf --num-prompts 256 --max-concurrency 32`, making sure this works with shared experts.
2. test with accuracy, `lm_eval --model local-completions --model_args "base_url=http://127.0.0.1:9256/v1/completions,model=deepseek-ai/DeepSeek-V3.1,num_concurrent=1024" --tasks gsm8k`

jinzhen-lin: Sorry, I don’t currently have enough resources to run deepseek-v3.1. But all model tests in CI seem to have passed, including the one that previously reported an error: `nm-testing/Qwen1.5-MoE-A2.7B-Chat-quantized.w4a16`.

https://buildkite.com/vllm/ci/builds/40015/steps/canvas?jid=019aa409-829f-490c-8209-73b0f110f99a

(Search "GSM8K Results for nm-testing/Qwen1.5-MoE-A2.7B-Chat-quantized.w4a16" in the logs)

mgoin: We only need to test that failing w4a16 model in CI. Testing DeepSeek block fp8 is not related to this PR

yewentao256: Make sense, thanks!

yewentao256 (approved):

yewentao256 (approved):
Let's try again before force merge

mgoin (approved):

PR #28623, opened 2025-11-13
Allow oot custom compiler extension via CompilerInterface

### LINKED ISSUE #23612: [Feature]: Allow oot custom compiler extension via CompilerInterface and reuse backend-agnostic FX passes

From vllm-ascend, we are trying to enable some graph fusion passes by extending the vllm core. But it seems that the compilation backend is currently coupled with the "inductor" backend which not every HW backend supports. In vllm-ascend, the inductor backend is not yet enabled and we are adding direct FX-based graph fusion. Currently, we have to do monkey-patches to plugin our own passes (see the rmsnorm+quant fusion PR from
@ganyi1996ppo: https://github.com/vllm-project/vllm-ascend/pull/2389). We also want to reuse the existing backend-agnostic FX passes from vllm core, e.g., sequence parallel fusion. But these are all "inductor" passes.

In summary, it would be great if

A HW backend can extend vllm to support custom compiler backend that implements CompilerInterface.
A HW backend with custom compiler backend can reuse existing backend-agnostic FX passes..
Would love to get your inputs.

PS: For "2", we can manually apply the AotDispatch passes and invoke those vllm "inductor" passes from inside the custom compiler backend. Not sure if that would work well though.

Discussion context from:
https://github.com/vllm-project/vllm/pull/23332#issuecomment-3217450860
https://github.com/vllm-project/vllm/pull/23332#issuecomment-3221935847

cc @ProExpertProg @Yikun @wangxiyuan @zou3519 @BoyuanFeng

### Implementation proposal
We propose to add an extra field `oot_compiler` to the `CompilationConfig` for out-of-tree extension to plug in their own piece-wise compiler. Then, `make_compiler` is updated to check and load the `oot_compiler` before instantiating other compilers like inductor.

The changes are listed below:
```diff
diff --git a/vllm/compilation/backends.py b/vllm/compilation/backends.py
index 3361b65a9..a62ce75f8 100644
--- a/vllm/compilation/backends.py
+++ b/vllm/compilation/backends.py
@@ -30,6 +30,8 @@ logger = init_logger(__name__)


 def make_compiler(compilation_config: CompilationConfig) -> CompilerInterface:
+    if compilation.oot_compiler:
+        return resolve_obj_by_qualname(compilation.oot_compiler)()
     if compilation_config.use_inductor:
         if envs.VLLM_USE_STANDALONE_COMPILE and is_torch_equal_or_newer(
                 "2.8.0.dev"):
diff --git a/vllm/config/compilation.py b/vllm/config/compilation.py
index 5c3b22001..90efd89c0 100644
--- a/vllm/config/compilation.py
+++ b/vllm/config/compilation.py
@@ -342,6 +342,8 @@ class CompilationConfig:
         "vllm.linear_attention",
     ]

+    oot_compiler: Optional[str] = None
+
     def compute_hash(self) -> str:
         """
         WARNING: Whenever a new field is added to this config,
```

Below is sample code for how an OOT extension can define its own OOT compiler and plug into vllm. In the OOT compiler, we use aot_autograd to preprocess the graph (do normalization and functionalization etc.) so that we can reuse existing inductor passes or define our own.

```python
...
from torch._dynamo.backends.common import aot_autograd

class TestOotCompiler(CompilerInterface):
    name: str = "TestOotCompiler"

    def compile(
        self,
        graph: fx.GraphModule,
        example_inputs: list[Any],
        compiler_config: dict[str, Any],
        runtime_shape: Optional[int] = None,
        key: Optional[str] = None,
    ) -> tuple[Optional[Callable], Optional[Any]]:
        def compile_inner(fx_graph, inputs):
            # OOT custom fusion pass goes here
            # can reuse existing Inductor passes from vllm core
            return graph

        # AotAutograd
        return aot_autograd(fw_compiler=compile_inner)(graph, example_inputs), None


def test_fusion_with_oot_compiler():
    compilation_config = CompilationConfig(
        level=CompilationLevel.PIECEWISE,
        oot_compiler=TestOotCompiler.__module__ + "." + TestOotCompiler.__name__
    )
    vllm_config = VllmConfig(compilation_config=compilation_config)
    with vllm.config.set_current_vllm_config(vllm_config):
        backend = compilation_config.init_backend(vllm_config)
        model = torch.nn.RMSNorm([2, 3])
        model = torch.compile(
            model,
            fullgraph=True,
            backend=backend)
        input = torch.randn(2, 2, 3)
        model(input)
```

### Alternatives

One may also directly invoke `torch.compile` on the full model with its own backend backend in the OOT extension, such as the practice in TPU (https://github.com/vllm-project/vllm/blob/0235103cbbdb511e6708aae600f759060a797c16/vllm/v1/worker/tpu_model_runner.py#L299). The downside is that we are not able to reuse the piece-wise compilation infrastructure from vllm core.

### Additional context

_No response_

wxsIcey:
## Purpose

From vLLM-Ascend, I am trying to enable some graph fusion operations by extending the vLLM core. However, it seems that the compiled backend is currently coupled with the "inductor" backend, and not all hardware backends support this backend. In vLLM-Ascend, `pattern matcher` for the inductor is supported, but code generation is not yet supported. 

Therefore, I hope that the hardware backend can be extended to support `custom compiler backend`, as detailed in the PR.

## Test Plan
ci passed

---
<details>

</details>


vllm/platforms/interface.py
```diff
@@ -134,6 +134,10 @@ class Platform:
 
     _global_graph_pool: Any | None = None
 
+    @property
+    def pass_key(self) -> str:
+        return "post_grad_custom_post_pass"
```
wangxiyuan, line 140: Not related to this PR, but I just have a question that can `pass_key` be a list?

IIUC, this property is used to add custom InductorPass  by config way dynamicly. So a list may be better for multi InductorPass case?

wangxiyuan (commented):
LGTM.

vllm/platforms/interface.py
wxsIcey: Thank you for your review. Regarding `pass_key`, I ​​understand it to be an Inductor pass configuration item, such as `post_grad_custom_post_pass` or `post_grad_custom_pre_pass`, corresponding to a `PassManager` or `InductorPass`. However, I'm currently unsure if other `pass_key` will be used, so I'll keep the current changes for now.

vllm/compilation/backends.py
```diff
@@ -58,13 +58,12 @@ def make_compiler(compilation_config: CompilationConfig) -> CompilerInterface:
...
-        assert compilation_config.backend == "eager", (
-            "Custom backends not supported with CompilationMode.VLLM_COMPILE"
-        )
-
+    elif compilation_config.backend == "eager":
         logger.debug("Using EagerAdaptor")
         return EagerAdaptor()
+    else:
+        logger.debug("Using custom backend: %s", compilation_config.backend)
+        return resolve_obj_by_qualname(compilation_config.backend)()
```
ProExpertProg: Can we check the interface here?
```suggestion
        compiler = resolve_obj_by_qualname(compilation_config.backend)()
        assert isinstance(compiler, CompilerInterface)
        return compiler
```

vllm/platforms/interface.py
```diff
@@ -134,6 +134,10 @@ class Platform:
 
     _global_graph_pool: Any | None = None
 
+    @property
+    def pass_key(self) -> str:
```
ProExpertProg, line 138:
```suggestion
    def pass_key(self) -> str:
    """Inductor config key for the PassManager custom pass"""
```

vllm/platforms/interface.py
```diff
@@ -177,6 +181,11 @@ def is_sleep_mode_available(self) -> bool:
         # all ROCm platforms for now.
         return self._enum in (PlatformEnum.CUDA, PlatformEnum.ROCM)
 
+    @classmethod
+    def get_pass_manager_cls(cls) -> str:
+        """Get the pass manager class of a device."""
```
ProExpertProg:
```suggestion
        """
        Get the pass manager class for this platform.
        It will be registered as a custom pass under the current_platform.pass_key.
        """
```

ProExpertProg (commented):
Thanks for adding this extension point, a few notes:

vllm/compilation/backends.py
wxsIcey: Thank you for your reply. I have revised it according to your feedback.

ProExpertProg (approved):

ProExpertProg:
Can you actually just add some documentation to `CompilationConfig.backend` that explains this can now be used with `CompilationMode.VLLM_COMPILE` as well?

vllm/compilation/backends.py
```diff
@@ -59,13 +59,14 @@ def make_compiler(compilation_config: CompilationConfig) -> CompilerInterface:
...
-        assert compilation_config.backend == "eager", (
-            "Custom backends not supported with CompilationMode.VLLM_COMPILE"
-        )
-
+    elif compilation_config.backend == "eager":
         logger.debug("Using EagerAdaptor")
         return EagerAdaptor()
+    else:
+        logger.debug("Using custom backend: %s", compilation_config.backend)
+        compiler = resolve_obj_by_qualname(compilation_config.backend)()
```
wangxiyuan: After @ProExpertProg reminding to update backend docs, I just revisit the logic here. Then I think there is a problem:

compilation_config.backend have three choice https://github.com/vllm-project/vllm/blob/3fb0d90999887949629d1e9bac4d98336a35c475/vllm/config/compilation.py#L693-L696
If it's emtpy, it will be set to `simple_compile_backend ` https://github.com/vllm-project/vllm/blob/3fb0d90999887949629d1e9bac4d98336a35c475/vllm/config/compilation.py#L710-L711

So there is no way to update the backend to a `resolve_obj_by_qualname` acutally unless you overrite it in `check_and_update_config` platform function(This is really a hack way)

So I suggest to add a new platform interface like `get_compiler_backend` to return custom compile cls. If it's not implemented, then fallbcak to `simple_compile_backend `

wxsIcey: Thanks for your suggestion, it's very important. I've modified the implementation logic. Could you please take another look?

vllm/platforms/interface.py
```diff
@@ -177,6 +182,21 @@ def is_sleep_mode_available(self) -> bool:
...
+        It will be registered as a custom pass under the current_platform.pass_key.
+        """
+        return "vllm.compilation.pass_manager.PostGradPassManager"
+
+    @classmethod
+    def get_compile_backend(cls) -> str:
+        """
+        Get the custom compile backend for current platform.
+        """
+        return ""
```
wangxiyuan: raise 
```suggestion
        raise NotImplementedError
```

vllm/config/compilation.py
```diff
@@ -704,7 +704,10 @@ def __post_init__(self) -> None:
             self.backend = "inductor" if self.use_inductor else "eager"
 
         if self.backend == "":
-            self.backend = current_platform.simple_compile_backend
+            try:
+                self.backend = current_platform.get_compile_backend()
+            except Exception:
```
wangxiyuan: please catch NotImplementedError instead of `Exception`

vllm/platforms/interface.py
wxsIcey: done

vllm/config/compilation.py
wxsIcey: done

vllm/config/compilation.py
```diff
@@ -704,7 +704,10 @@ def __post_init__(self) -> None:
             self.backend = "inductor" if self.use_inductor else "eager"
 
         if self.backend == "":
-            self.backend = current_platform.simple_compile_backend
+            try:
+                self.backend = current_platform.get_compile_backend()
+            except NotImplementedError:
+                self.backend = current_platform.simple_compile_backend
```
ProExpertProg: Why not make `get_compile_backend` return `simple_compile_backend` by default?

vllm/platforms/interface.py
```diff
@@ -177,6 +182,21 @@ def is_sleep_mode_available(self) -> bool:
...
+        It will be registered as a custom pass under the current_platform.pass_key.
+        """
+        return "vllm.compilation.pass_manager.PostGradPassManager"
+
+    @classmethod
+    def get_compile_backend(cls) -> str:
+        """
+        Get the custom compile backend for current platform.
+        """
+        raise NotImplementedError
```
ProExpertProg:
```suggestion
        return self.simple_compile_backend
```

ProExpertProg (approved):

vllm/config/compilation.py
wxsIcey: Thank you again for your suggestion. I have revised it.

wangxiyuan (approved):

ProExpertProg (approved):

vllm/config/compilation.py
```diff
@@ -736,9 +736,7 @@ def init_backend(self, vllm_config: "VllmConfig") -> str | Callable:
 
         assert self.mode == CompilationMode.VLLM_COMPILE
         if self.backend not in ["eager", "inductor"]:
-            raise ValueError(
-                f"Invalid backend for piecewise compilation: {self.backend}"
-            )
+            logger.info("Using oot custom backend for compilation.")
```
hmellor: nit

```suggestion
            logger.info("Using OOT custom backend for compilation.")
```

wxsIcey: Thanks. I have modified it.

wxsIcey:
The test has passed all tests. Hopefully, it can be merged as soon as possible. Thank you. @ProExpertProg

PR #28662, opened 2025-11-13
[bugfix] check whether MoE Layers are ignored by quantization

HDCharles:
Applying quantization while ignoring some MoE layers but not others would cause model load errors due to vllm assuming all the layers were quantized since there was no check of the ignore list unlike the one for [linear](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py#L138)

Changes
1) previously the code had to [duplicate the get_scheme functionality for FusedMoE](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors_moe.py#L113-L136) due to layer name changes and the layer no longer being linear. This duplicated code had issues like not taking into account the ignore list. I got rid of the duplication by making it so FusedMoE can use get_scheme if we modify the CTconfig slightly.
2) refactor the get_moe_method to use quant_scheme rather than quant_config, 
2a)^ requires changing a lot of non-static CompressedTensorConfig methods like _is_wNa16_group_channel to be converted to static/class methods since we won't have access to teh quant_config to call them. (Is this fine? TODO)
3) we check for qschemes of all 3 projections but for only expert 0, this aligns with the old logic which used a regex and only returned the first match but could be corrected. It may be slow though.

depends on [28770](https://github.com/vllm-project/vllm/pull/28870) and [28771](https://github.com/vllm-project/vllm/pull/28871)

@kylesayrs @dsikka @brian-dellabetta

### DIFF
--- a/vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
+++ b/vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
@@ -266,8 +266,9 @@ class CompressedTensorsConfig(QuantizationConfig):
     def get_config_filenames(cls) -> list[str]:
         return []
 
+    @staticmethod
     def _check_scheme_supported(
-        self, min_capability: int, error: bool = True, match_exact: bool = False
+        min_capability: int, error: bool = True, match_exact: bool = False
     ) -> bool:
         capability_tuple = current_platform.get_device_capability()
 
@@ -293,9 +294,8 @@ class CompressedTensorsConfig(QuantizationConfig):
         else:
             return False
 
-    def _is_fp4a4_nvfp4(
-        self, weight_quant: QuantizationArgs, input_quant: QuantizationArgs
-    ):
+    @staticmethod
+    def _is_fp4a4_nvfp4(weight_quant: QuantizationArgs, input_quant: QuantizationArgs):
         if weight_quant is None or input_quant is None:
             return False
 
@@ -322,9 +322,8 @@ class CompressedTensorsConfig(QuantizationConfig):
             and is_symmetric
         )
 
-    def _is_fp4a16_nvfp4(
-        self, weight_quant: QuantizationArgs, input_quant: QuantizationArgs
-    ):
+    @staticmethod
+    def _is_fp4a16_nvfp4(weight_quant: QuantizationArgs, input_quant: QuantizationArgs):
         is_weight_only = weight_quant is not None and input_quant is None
         is_tensor_group_quant = (
             weight_quant.strategy == QuantizationStrategy.TENSOR_GROUP.value
@@ -344,8 +343,9 @@ class CompressedTensorsConfig(QuantizationConfig):
             and is_symmetric
         )
 
+    @staticmethod
     def _is_static_tensor_w8a8(
-        self, weight_quant: QuantizationArgs, input_quant: QuantizationArgs
+        weight_quant: QuantizationArgs, input_quant: QuantizationArgs
     ) -> bool:
         is_8_bits = weight_quant.num_bits == input_quant.num_bits == 8
         weight_strategy = (
@@ -362,8 +362,9 @@ class CompressedTensorsConfig(QuantizationConfig):
         # Only symmetric weight quantization supported.
         return is_8_bits and is_tensor and weight_quant.symmetric and is_static
 
+    @staticmethod
     def _is_dynamic_token_w8a8(
-        self, weight_quant: QuantizationArgs, input_quant: QuantizationArgs
+        weight_quant: QuantizationArgs, input_quant: QuantizationArgs
     ) -> bool:
         is_8_bits = weight_quant.num_bits == input_quant.num_bits == 8
         weight_strategy = (
@@ -379,8 +380,9 @@ class CompressedTensorsConfig(QuantizationConfig):
         # Only symmetric weight quantization supported.
         return is_8_bits and is_token and weight_quant.symmetric and is_dynamic
 
+    @staticmethod
     def _is_dynamic_token_w4a8_int(
-        self, weight_quant: QuantizationArgs, input_quant: QuantizationArgs
+        weight_quant: QuantizationArgs, input_quant: QuantizationArgs
     ) -> bool:
         is_weight_4_bits = weight_quant.num_bits == 4
         is_activation_8_bits = input_quant.num_bits == 8
@@ -403,8 +405,9 @@ class CompressedTensorsConfig(QuantizationConfig):
             and is_dynamic
         )
 
+    @staticmethod
     def _is_fp8_w8a8(
-        self, weight_quant: QuantizationArgs, input_quant: QuantizationArgs
+        weight_quant: QuantizationArgs, input_quant: QuantizationArgs
     ) -> bool:
         # Confirm weights and activations quantized.
         if weight_quant is None or input_quant is None:
@@ -439,8 +442,9 @@ class CompressedTensorsConfig(QuantizationConfig):
         is_per_tensor_activation = input_quant.strategy == QuantizationStrategy.TENSOR
         return is_symmetric_activation and is_per_tensor_activation
 
+    @staticmethod
     def _is_fp8_w4a8(
-        self, weight_quant: QuantizationArgs, input_quant: QuantizationArgs
+        weight_quant: QuantizationArgs, input_quant: QuantizationArgs
     ) -> bool:
         if not weight_quant or not input_quant:
             return False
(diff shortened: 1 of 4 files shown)


vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
```diff
@@ -723,6 +729,47 @@ def get_scheme(
...
+                               fused_mapping=self.packed_modules_mapping):
+            return True
+
+        # Check individual expert projections
+        # We need to check if expert 0's projections are ignored as a heuristic
+        # for whether all experts should be ignored
+        expert_proj_names = ["gate_proj", "up_proj", "down_proj"]
+
+        for proj_name in expert_proj_names:
+            # Check expert 0's projection
```
brian-dellabetta: why only expert 0? we are assuming all experts have the same quant config?

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
```diff
@@ -723,6 +729,47 @@ def get_scheme(
...
+            expert_layer_name = f"{prefix}.0.{proj_name}"
+            if not should_ignore_layer(
+                    expert_layer_name,
+                    ignore=self.ignore,
+                    fused_mapping=self.packed_modules_mapping):
+                # If any projection is NOT ignored, the layer should be quantized
+                return False
+
+        # All projections for expert 0 are ignored, so ignore the whole layer
+        return True
```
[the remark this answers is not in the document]
brian-dellabetta: Your diff looks correct to me, if we want to ignore the FusedMoe layer only if "If all experts' projections (gate_proj, up_proj, down_proj) are ignored" I think gemini's suggestion is wrong here. Gemini's suggestion will return True if gate_proj should be ignored, without ever checking up_proj or down_proj

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
```diff
@@ -508,8 +528,9 @@ def _is_fp8_w8a16(
             and is_tensor_or_channel_or_block_weight
         )
 
+    @staticmethod
```
HDCharles, line 555: why are all these _is_X methods not class/static methods? is there an expectation of needing the state of the config rather than the qschemes being passed in?

There are like 15 of them, this change was made as an example.

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors_moe.py
```diff
@@ -106,41 +106,13 @@ class CompressedTensorsMoEMethod(FusedMoEMethodBase):
     def get_moe_method(
         quant_config: "CompressedTensorsConfig",  # type: ignore # noqa E501
```
HDCharles: we can get rid of quant_config entirely here since we have quant_scheme but it would require us converting 5 or so CompressedTensorsConfig.is_X methods to class/static methods at which point i'd probably just change them all. Want to make sure i'm on the right track before i start doing all that.

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors_moe.py
```diff
@@ -106,41 +106,13 @@ class CompressedTensorsMoEMethod(FusedMoEMethodBase):
...
-
-                # Only valid if down_proj, gate_proj, and up_proj
-                # are mapped to the same quant scheme in the quant_config
-                if current_scheme is None:
-                    current_scheme = quant_config.target_scheme_map.get(matched_target)
-                else:
-                    assert current_scheme == quant_config.target_scheme_map.get(
-                        matched_target
-                    )
-
```
HDCharles, line 137: we now check that qschemes are the same in the config

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
```diff
@@ -158,9 +160,27 @@ def get_quant_method(
         if isinstance(layer, Attention):
             return CompressedTensorsKVCacheMethod(self)
         if isinstance(layer, FusedMoE):
-            return CompressedTensorsMoEMethod.get_moe_method(self, layer)
+            # FusedMoE was made by combining multiple Linears so need to make sure quantization config for Linear can target it
+            self._add_fused_moe_to_target_scheme_map()
```
HDCharles: not sure if this is kosher, it feels bad that we basically have to replicate the whole get_scheme logic elsewhere because we're using a fused module. 

if we make this change to copy the scheme from target='Linear' to target='FusedMoE' then we can just use get scheme.

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
```diff
@@ -158,9 +160,27 @@ def get_quant_method(
         if isinstance(layer, Attention):
             return CompressedTensorsKVCacheMethod(self)
         if isinstance(layer, FusedMoE):
-            return CompressedTensorsMoEMethod.get_moe_method(self, layer)
+            # FusedMoE was made by combining multiple Linears so need to make sure quantization config for Linear can target it
+            self._add_fused_moe_to_target_scheme_map() 
+            unfused_names = [[prefix + proj_name] for proj_name in [".0.gate_proj", ".0.up_proj", ".0.down_proj"]]
```
HDCharles: potential issue if expert 0 has one scheme and expert 1 has another. However i don't think this would be a NEW issue since the old logic was even weirder where it pretended the layer names were regexes. We can check all the experts and make sure it all aligns but it seems like overkill

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
```diff
@@ -158,6 +158,12 @@ def get_quant_method(
         if isinstance(layer, Attention):
             return CompressedTensorsKVCacheMethod(self)
         if isinstance(layer, FusedMoE):
+            # Check if the FusedMoE layer should be ignored by checking if
+            # the expert projections are in the ignore list
+            if self._should_ignore_fused_moe(prefix, layer):
+                from vllm.model_executor.layers.fused_moe import (
+                    UnquantizedFusedMoEMethod)
+                return UnquantizedFusedMoEMethod(layer.moe_config)
```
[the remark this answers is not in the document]
HDCharles: outdated

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
HDCharles: we can match all of them but this aligns with the old logic. 

before we would [pretend the layer names were regexes](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors_moe.py#L118)

then just return the [first match](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/quantization/compressed_tensors/utils.py#L120)

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
kylesayrs: There's a lot of code clean up that could be done in CT matching logic. I have a backlog ticket to clean a lot of this up and use CT matching utilities.

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
```diff
@@ -158,9 +158,36 @@ def get_quant_method(
...
+            all_quant_schemes = set(
+                [self.get_scheme(layer, name) for name in unfused_names]
+            )
+
+            if len(all_quant_schemes) > 1:  # inconsistent qscheme
+                raise ValueError(
+                    "All MoE projections need to have same quantization scheme"
+                )
+
+            quant_scheme = all_quant_schemes.pop()
```
kylesayrs: I think this logic should go in `CompressedTensorsMoEMethod.get_moe_method`

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
```diff
@@ -158,9 +158,36 @@ def get_quant_method(
...
+            # FusedMoE was made by combining multiple Linears so need to
+            # make sure quantization config for Linear can target it
+            self._add_fused_moe_to_target_scheme_map()
+            unfused_names = [
+                prefix + proj_name
+                for proj_name in [".0.gate_proj", ".0.up_proj", ".0.down_proj"]
+            ]
+            all_quant_schemes = set(
+                [self.get_scheme(layer, name) for name in unfused_names]
+            )
```
kylesayrs: I don't really understand this logic? Is this supposed to only be checking the first expert?

Also, not all models have experts in this format. Some checkpoints have fused experts, don't use names like gate_proj, ect.

I think you'll have to use the `moe.expert_mapping` in order to determine the names of all the checkpoint modules, then check the schemes of those checkpoint modules.

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
brian-dellabetta: ok, good to know, thanks

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
HDCharles: shouldn't it match the way we're doing it for linear? Feels like it makes way more sense to pass in a quantization scheme for a layer to get a moe method for a layer, rather than passing in a quantization config for a model to get a moe method for a layer

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
HDCharles: its not new logic, its just the logic from below that I moved here per @mgoin 's [todo comment ](https://github.com/vllm-project/vllm/pull/28662/files#diff-ff1294eac1d5278b98c34982811d183967f327696bfdcc7f546d3b8c0d44f6ebL110-L112) and getting the q_scheme at the same part of the code as we do for Linear, so that we don't have multiple interfaces for getting q_scheme that we need to maintain separately and forget that one doesn't check the ignore list.

A) [matching only the first expert](https://github.com/vllm-project/vllm/blob/da14ae0fad3165b88fcdc03a8f59f1813f8e832a/vllm/model_executor/layers/quantization/compressed_tensors/utils.py#L120)
B) [hard coding gate_proj, up_proj, down_proj](https://github.com/vllm-project/vllm/blob/da14ae0fad3165b88fcdc03a8f59f1813f8e832a/vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors_moe.py#L118)

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
HDCharles: for the _is_X methods? there's a comment about CT matching but it seems more applicable to the matching logic in get_scheme

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
HDCharles: also it looks like layer.expert_mapping is unpopulated at this point, is up/gate/down default and i should use that if expert_mapping is None?

PR #28665, opened 2025-11-13
[Config] Clean up SchedulerConfig initialization

DarkLight1337:
## Purpose

- Move default values of `SchedulerConfig` from utils to `SchedulerConfig` itself.
- In normal usage of vLLM, `max_num_batched_tokens` and `max_num_seqs` are actually set by `UsageContext` when constructing `EngineArgs`, so there is no point in making the default values of `SchedulerConfig` dynamic (the defaults for pooling and multimodal models are actually being overwritten). This simplifies the initialization code a lot.
- Split up the logic of `EngineArgs._set_default_args` to be more modular.
- Make `SchedulerConfig.chunked_prefill_enabled` a property-based alias of `SchedulerConfig.enable_chunked_prefill` to avoid having to set both when overriding the config in each platform.

## Test Result

---
<details>

</details>

### DIFF
--- a/tests/models/language/generation/test_hybrid.py
+++ b/tests/models/language/generation/test_hybrid.py
@@ -348,9 +348,14 @@ def test_fp32_cache_state(
 
 
 # Helper functions for the APC tests
-def _get_vllm_runner_params(model, max_model_len, tensor_parallel_size=1):
+def _get_vllm_runner_params(
+    model: str,
+    max_model_len: int,
+    tensor_parallel_size: int = 1,
+):
     return {
         "model_name": model,
+        "enable_chunked_prefill": True,
         "enable_prefix_caching": False,
         "max_model_len": max_model_len,
         "tensor_parallel_size": tensor_parallel_size,
--- a/tests/v1/sample/test_logprobs.py
+++ b/tests/v1/sample/test_logprobs.py
@@ -47,6 +47,7 @@ def vllm_model(vllm_runner, request) -> Generator[VllmRunner, None, None]:
         max_num_batched_tokens=16,
         max_num_seqs=16,
         max_model_len=128,
+        enable_chunked_prefill=True,
         enforce_eager=True,
         # TODO: enable this once we support it for
         # prompt logprobs.
--- a/vllm/config/scheduler.py
+++ b/vllm/config/scheduler.py
@@ -4,7 +4,7 @@
 import hashlib
 from collections.abc import Callable
 from dataclasses import InitVar
-from typing import TYPE_CHECKING, Any, Literal, cast
+from typing import TYPE_CHECKING, Any, ClassVar, Literal, cast
 
 from pydantic import Field, field_validator, model_validator
 from pydantic.dataclasses import dataclass
@@ -12,11 +12,6 @@ from typing_extensions import Self
 
 from vllm.config.utils import config
 from vllm.logger import init_logger
-from vllm.utils import (
-    DEFAULT_MAX_NUM_BATCHED_TOKENS,
-    MULTIMODAL_MODEL_MAX_NUM_BATCHED_TOKENS,
-    POOLING_MODEL_MAX_NUM_BATCHED_TOKENS,
-)
 from vllm.utils.import_utils import resolve_obj_by_qualname
 
 if TYPE_CHECKING:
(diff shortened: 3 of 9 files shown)


vllm/engine/arg_utils.py
```diff
@@ -1838,39 +1840,59 @@ def _set_default_args(
...
+                    "Enabling" if default_chunked_prefill else "Disabling",
+                )
+        if self.enable_prefix_caching is None:
+            self.enable_prefix_caching = default_prefix_caching
+
+            if model_config.runner_type == "pooling":
+                logger.info(
+                    "%s chunked prefill by default",
+                    "Enabling" if default_prefix_caching else "Disabling",
+                )
```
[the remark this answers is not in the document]
DarkLight1337: Good catch, fixed

vllm/config/scheduler.py
```diff
@@ -76,9 +78,13 @@ class SchedulerConfig:
     NOTE: This will be replaced by speculative config in the future; it is
     present to enable correctness tests until then."""
 
-    enable_chunked_prefill: bool = Field(default=None)
+    enable_chunked_prefill: bool = False
     """If True, prefill requests can be chunked based
-    on the remaining max_num_batched_tokens."""
+    on the remaining `max_num_batched_tokens`.
+
+    The default value here is mainly for convenience when testing.
```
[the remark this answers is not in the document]
DarkLight1337: It is set by `get_chunked_prefill_prefix_caching_defaults`

vllm/config/scheduler.py
```diff
@@ -199,54 +194,30 @@ def _skip_none_validation(cls, value: Any, handler: Callable) -> Any:
...
 
+        # Ensure max_num_batched_tokens does not exceed model limit.
+        # Some models (e.g., Whisper) have embeddings tied to max length.
+        self.max_num_batched_tokens = min(
+            self.max_num_seqs * self.max_model_len,
+            self.max_num_batched_tokens,
+        )
+
         self.max_num_encoder_input_tokens = self.max_num_batched_tokens
         self.encoder_cache_size = self.max_num_batched_tokens
```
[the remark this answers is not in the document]
DarkLight1337: The case of no `UsageContext` is not normal usage of vLLM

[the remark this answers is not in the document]
DarkLight1337: @njhill @WoosukKwon @robertgshaw2-redhat correct me if I'm wrong about this

Isotr0py (approved):

vllm/config/scheduler.py
```diff
@@ -33,25 +28,32 @@
...
-    max_model_len: int = Field(default=None, ge=1)
-    """Maximum length of a sequence (including prompt and generated text). This
-    is primarily set in `ModelConfig` and that value should be manually
-    duplicated here."""
+    max_model_len: int = Field(default=8192, ge=1)
+    """Maximum length of a sequence (including prompt and generated text).
+
+    The default value here is mainly for convenience when testing.
+    In real usage, this should duplicate `ModelConfig.max_model_len` via
+    `EngineArgs`."""
```
hmellor, line 56: Could we remove this entirely?

vllm/config/scheduler.py
```diff
@@ -276,6 +232,14 @@ def __post_init__(self, is_encoder_decoder: bool) -> None:
...
             )
 
+    @property
+    def chunked_prefill_enabled(self) -> bool:
+        return self.enable_chunked_prefill
+
+    @chunked_prefill_enabled.setter
+    def chunked_prefill_enabled(self, value: bool):
+        self.enable_chunked_prefill = value
+
```
hmellor, line 242: Can we just remove this? It used to be `init=False` so it's not part of the normal API of `SchedulerConfig`

vllm/engine/arg_utils.py
```diff
@@ -428,11 +428,11 @@ class EngineArgs:
     cpu_offload_gb: float = CacheConfig.cpu_offload_gb
     gpu_memory_utilization: float = CacheConfig.gpu_memory_utilization
     kv_cache_memory_bytes: int | None = CacheConfig.kv_cache_memory_bytes
-    max_num_batched_tokens: int | None = SchedulerConfig.max_num_batched_tokens
+    max_num_batched_tokens: int | None = None
```
hmellor, line 431: Could you group these fields which don't copy the defaults from their respective config and add a comment saying why?

tests/v1/sample/test_logprobs.py
```diff
@@ -47,6 +47,7 @@ def vllm_model(vllm_runner, request) -> Generator[VllmRunner, None, None]:
         max_num_batched_tokens=16,
         max_num_seqs=16,
         max_model_len=128,
+        enable_chunked_prefill=True,
```
hmellor, line 50: Why is this necessary if this is the default value?

tests/models/language/generation/test_hybrid.py
```diff
@@ -348,9 +348,14 @@ def test_fp32_cache_state(
...
 # Helper functions for the APC tests
-def _get_vllm_runner_params(model, max_model_len, tensor_parallel_size=1):
+def _get_vllm_runner_params(
+    model: str,
+    max_model_len: int,
+    tensor_parallel_size: int = 1,
+):
     return {
         "model_name": model,
+        "enable_chunked_prefill": True,
```
hmellor, line 358: Same here given that it's the default value?

hmellor (commented):
This is a really nice change, I've left a few comments for clarification

vllm/config/scheduler.py
DarkLight1337: It is used in some other places like `vllm.v1.core.sched.Scheduler`. We can try to refactor this in another PR.

vllm/config/scheduler.py
DarkLight1337: Same as above

tests/v1/sample/test_logprobs.py
DarkLight1337: `VllmRunner` disables chunked prefill by default, and when that's the case, an error is raised if `max_num_batched_tokens < max_model_len`

tests/models/language/generation/test_hybrid.py
DarkLight1337: As above

vllm/engine/arg_utils.py
DarkLight1337: We can do that in the next cleanup

vllm/config/scheduler.py
hmellor: Yeah this would need a small refactor. A follow up PR sounds good.

hmellor (approved):

WoosukKwon:
@DarkLight1337 I think this PR changes the default values for `max_num_batched_tokens` and `max_num_seqs` unexpectedly.

hmellor:
If the behaviour has changed I don't think it's because of the changed defaults in `SchedulerConfig`. The if blocks linked will be triggered if the values of `EngineArgs` are `None`, which still are in the default case:

https://github.com/vllm-project/vllm/blob/d4acf518d09515560e1082a80b8a4d6550e20d9b/vllm/engine/arg_utils.py#L436
https://github.com/vllm-project/vllm/blob/d4acf518d09515560e1082a80b8a4d6550e20d9b/vllm/engine/arg_utils.py#L440

DarkLight1337:
Ok I figured out the issue, the CLI defaults are still using the ones from `SchedulerConfig`.

PR #28671, opened 2025-11-13
[Bugfix] Respect VLLM_CONFIGURE_LOGGING value

elizabetht:
## Purpose
Fix https://github.com/vllm-project/vllm/issues/28666 

## Test Result

---
<details>

</details>


vllm/utils/system_utils.py
```diff
@@ -168,6 +168,21 @@ def write_with_prefix(s: str):
...
+    except (ValueError, AttributeError) as e:
+        # If VLLM_CONFIGURE_LOGGING is set to an invalid value, log a warning
+        # and disable log decoration to avoid crashing the process
+        logger.warning(
+            "Invalid value for VLLM_CONFIGURE_LOGGING environment variable: %s. "
+            "Disabling log decoration. Error: %s",
+            os.getenv("VLLM_CONFIGURE_LOGGING"),
+            e,
+        )
+        return
```
mgoin, line 209: You should be able to avoid all this by updating the entry in envs.py to bool rather than an int e.g.
```
    "VLLM_CONFIGURE_LOGGING": lambda: int(os.getenv("VLLM_CONFIGURE_LOGGING", "1")),
```
into
```
    "VLLM_CONFIGURE_LOGGING": lambda: bool(int(os.getenv("VLLM_CONFIGURE_LOGGING", "1"))),
```

elizabetht: Ah, thanks for the pointer, very new to the code @mgoin  - I will incorporate the change.

markmc:
See also #28582 adding an env var to disable `decorate_logs()`

As per #25700 I think we should add a command line parameter instead of an env var

elizabetht:
> See also #28582 adding an env var to disable `decorate_logs()`
> 
> As per #25700 I think we should add a command line parameter instead of an env var

@markmc Since the conversion of environment variables to command-line arguments is being handled in a separate RFC (which I’m also involved in), could we keep this PR focused on the originally reported issue and avoid expanding its scope?

vllm/utils/system_utils.py
elizabetht: @mgoin I have incorporated your suggestion. Could you take another look at the PR?

FelixSchneiderZoom (approved):

vllm/utils/system_utils.py
```diff
@@ -204,6 +204,18 @@ def write_with_prefix(s: str):
...
+        if not envs.VLLM_CONFIGURE_LOGGING:
+            return
+    except ValueError:
+        logger.warning(
+            "Invalid value for VLLM_CONFIGURE_LOGGING: %s. Expected a boolean. "
+            "Proceeding to decorate logs.",
+            envs.VLLM_CONFIGURE_LOGGING,
+        )
+        # Fall through to decorate logs as stated in the warning
+
```
hmellor: `decorate_logs` shouldn't have to care if an environment variable was set correctly. If it's been set incorrectly, it should error

```suggestion
    if not envs.VLLM_CONFIGURE_LOGGING:
        return

```

elizabetht: @hmellor I initially had the proposed changes and added the exception handling later..

vllm/envs.py
hmellor, line 1: Could this change also be reverted? The rest of vLLM assumes it's an `int`

elizabetht: @hmellor To my understanding, this change as suggested by @mgoin is safe and actually aligns the implementation with the test expectations. All usages treat it as a boolean (using if / if not), so changing from [int](0/1) to [bool](False/True) is fully compatible. Let me know if I am missing anything.

hmellor: Ok, we can change it. Please update the type hint and default on line 40

```py
    VLLM_CONFIGURE_LOGGING: int = 1
```

elizabetht: Updated, thanks for catching it @hmellor

tests/test_envs.py
```diff
@@ -365,3 +365,81 @@ def test_duplicate_values_deduped(self):
...
+                assert isinstance(result, bool), (
+                    f"Expected bool for VLLM_CONFIGURE_LOGGING={env_value!r}, "
+                    f"got {type(result).__name__}"
+                )
+
+                # Should have the correct value
+                assert result == expected, (
+                    f"Expected {expected} for VLLM_CONFIGURE_LOGGING={env_value!r}, "
+                    f"got {result}"
+                )
```
hmellor: This test is the same as the previous 2

```suggestion
```

hmellor (approved):
LGTM after the last nit is resolved

tests/test_envs.py
elizabetht: Done, removed the duplicative test

yewentao256 (approved):
LGTM, thanks for the work!

Issue #28326, opened 2025-11-07
[RFC]: vLLM Support for Generic Model Definitions


bwasti:
### Motivation.

As users experiment with model architectures they often need a fast inference path to explore post-training (with reinforcement learning).  Since training in this domain is inference dominated, vLLM should increase support for research by providing utilities to construct fast-inference models directly from training code with low cognitive overhead.

Recent exploration in [bitwise-exact reinforcement learning](https://github.com/pytorch/torchtitan/tree/main/torchtitan/experiments/deterministic_vllm_rl) with vLLM has shown that leveraging the features of vLLM are crucial to ensure stability in training runs.  Further, vLLM is already used in many RL frameworks (such as VERL and TorchForge), and by simplifying integration, we can continue to help grow the vLLM impact and community.

### Proposed Change.

The goal is to support generic training/etc. ⇔ inference compatibility.  We propose a formal **specification** and **helper utilities** for use of `ModelRegistry.register_model` with training and other workloads.

- A specification of the minimum set of features needed to use vLLM with a user-defined model
- A wrapper that will enable simplistic “training-only” model specifications (by shimming interfaces or features for use with `register_model`)
- A set of vLLM-provided `nn.Module`s to support both training and efficient inference, such as `Attention`
- Flexibility to support user-defined parallelism (specifically exposing process group instantiation to model registration)

### Feedback Period.

1-2 weeks

### CC List.

@zhuohan123 @tianyu-l @teja-rao

### Any Other Things.

Current very-close interface (that requires invasive changes without utilities):
https://docs.vllm.ai/en/latest/contributing/model/basic/

Additional features (module boilerplate) here: 
https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/registry.py

Additional backwards passes for attention could be added here:
https://github.com/vllm-project/vllm/tree/main/vllm/model_executor/layers

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

hmellor:
vLLM does actually already support this using the Transformers backend. 

If you have any questions about it I'd be happy to elaborate.

hmellor:
I'm interested to know what would be in the superset. As the developer of the Transformers backend, I'm working on making it support everything that vLLM does. Right now some key omissions are:

- Mamba (should be able to get this working pretty easily)
[...]

PR #28685, opened 2025-11-13
[Feature] Generic Model Support via TrainableAttention and ModelRegistry parallelism constructor callback

bwasti:
## Purpose

Implement RFC #28326 to enable users to easily integrate custom models with vLLM for both training and inference, with support for external parallelism libraries (e.g., Megatron-LM, FSDP, DeepSpeed).

Enables use cases like:
  - RL training with vLLM acceleration
  - Integrating third-party models without rewriting parallelism logic
  - Mixing external tensor parallel layers (Megatron) with vLLM attention

## Test Plan

```
# example, should be a test
python examples/offline_inference/custom_model_with_megatron.py
```
  - Full transformer using Megatron-LM's ColumnParallelLinear + RowParallelLinear for MLPs
  - TrainableFlashAttention for attention (with KV cache + backward pass)
  - Ground truth validation against independent PyTorch run
  - Tests TP=4, dynamic batching, training backward pass
 
```
# actual test (simplistic)
pytest tests/models/test_generic_models.py
```

## Test Result

Screenshot 2025-11-13 at 4 32 51 PM

Both pass

---
<details>

</details>


examples/custom_models/deepseek_v3_torchtitan.py
tianyu-l, line 1: For an RL framework to use such vLLM-version of torchtitan models, where should such models eventually be put?

A related question would be, for batch invariance implementation we can use the "canonical" version of modules (Attention, RoPE, etc.) showcased here, then for other versions (e.g. FlexAttention with arbitrary masks), how should we configure them? Is the idea that we configure the RL framework to just not call into these compat functions including `replace_with_trainable_attention`?

bwasti: > where should such models eventually be put

that's up to the RL framework

> batch invariance implementation

out of scope for this PR.  every kernel needs to be audited in a batch invariant impl - user code has no guarantees about that.  we'll need some type of test that can highlight issues and allow users to pick which tools are needed (overriding pytorch, torch.compile, replace_with_trainable_attn)

wwwjn: From what I read (mainly `TrainableMLA`, `TrainableAttention`, `model_excutor/custom_models`), this PR doesn't change vLLM's interface of supporting generic model that much (basically following instructions here: https://docs.vllm.ai/en/latest/contributing/model/registration/) , but provide an practical example that how user could replace a customized model's attention with vllm's attention (so it has kv cache capability). This PR also provide a model wrapper (`DeepSeekV3TorchTitanForCausalLM`) of Torchtitan model which implemented APIs needed by vLLM. Please correct me if it's wrong.

There are 2 questions to be discussed:
1. Where does the customize model wrapper should be placed? 
    I'd prefer vLLM could have an example of how to plug in customized model - I was struggling with the doc to run customized model. However, the model wrapper is closed related to torchtitan model definition, even the TrainableMLA. I would prefer to put it in torchtitan, and showcase "vLLM+custom model" with a simpler nn.Module model. 

2. How does it work with batch-invariant model in the future? (This might be a very long-term goal) 
    We want the "canonical" model definition between training and inference, because we want to run batch-invariant mode and achieve (almost) bit-wise identity between training and inference. Currently TrainableAttention has 2 separate path for training and inference. Do you have some idea how we should achieve bitwise-identity here?

examples/custom_models/benchmark_deepseek_v3.py
```diff
@@ -0,0 +1,411 @@
...
+            traceback.print_exc()
+
+    # Benchmark built-in vLLM model
+    if run_builtin:
+        try:
+            print("\n" + "=" * 70)
+            print("Built-in DeepSeek (vLLM Native Implementation)")
+            print("=" * 70)
+            results["builtin"] = benchmark_model(
+                model_name=args.model,
```
wwwjn, line 340: Hi @bwasti  Thanks for the great PR, it helps me a lot of understanding vllm model registering... I have a question about running torchtitan model here:
- The model_name= "deepseek-ai/DeepSeek-V3-Base", so its running the following code:`llm = LLM(
    model="deepseek-ai/DeepSeek-V3-Base",
    trust_remote_code=True,
    tensor_parallel_size=8,
)`, which will download `config.json` from [HF](https://huggingface.co/deepseek-ai/DeepSeek-V3.1-Base/blob/main/config.json), and the architecture is `DeepseekV3ForCausalLM ` 

- When registering model in `examples/custom_models/deepseek_torchtitan.py`, the registered model architecture is "DeepSeekV3TorchTitan"

How does vllm find registered customized model, as these 2 model architectures are different? I run the following command `python examples/custom_models/benchmark_deepseek_v3.py --run-both --num-requests 10` . Did I missed something here? Thank you!

examples/custom_models/qwen3_torchtitan.py
```diff
@@ -0,0 +1,295 @@
...
+        # Get embeddings
+        h = (
+            inputs_embeds
+            if inputs_embeds is not None
+            else self.model.tok_embeddings(input_ids)
+        )
+
+        # Get RoPE cache
+        seqlen = h.shape[1] if h.dim() == 3 else h.shape[0]
+        rope_cache = self.model.rope_cache[:seqlen]
```
acisseJZhong, line 128: qq: should this be `self.model.rope_cache[positions]`?

PR #28687, opened 2025-11-13
[Performance] Reduce DeepGEMM N dim restriction from 128 to 64 multiplier 

alexm-redhat:
W8A8BlockFp8LinearOp class uses should_use_deepgemm_for_fp8_linear() function to determine which gemm to run with deepgemm fp8_gemm_nt kernel. Originally it was checking that N dim is divisible by 128, however, after some inspection of the deepgemm tests, I found that 64 works as well. This allows fused_qkv_a_proj to run with deepgemm kernel (and not cutlass) and is ~1.83X faster (after inspecting profiles).

With cutlass:
image

With deepgemm:
image

* lm_eval correctness verification shows that no accuracy loss.
* e2e TPOT improvement is from 28.32ms to 27.72 so about 2.2%


ElizaWszola:
If we successfully remove the size restriction, it would be good if we could also remove the `should_use_deepgemm_for_fp8_linear` condition from `W8A8BlockFp8LinearOp.apply()` and decide whether we want to run deepgemm with `_dispatch_w8a8_blockscale_op()`

examples/offline_inference/basic/basic.py
```diff
@@ -16,7 +16,7 @@
 
 def main():
     # Create an LLM.
-    llm = LLM(model="facebook/opt-125m")
+    llm = LLM(model="deepseek-ai/DeepSeek-R1-0528", tensor_parallel_size=8)
```
heheda12345: can you revert?

yewentao256 (commented):
We should be more careful about removing this.
I tried to remove before, and got a lot performance loss. Seems to be like dp=8 case of R1

alexm-redhat:
@yewentao256 I have verified dp==8 DSR1 (on H200) and the performance is actually better with this PR, about 1.5% for TPOT.

alexm-redhat:
@yewentao256 @robertgshaw2-redhat Verified performance for DSR1 on 8xB200 system that has DeepGEMM installed. For TP==8 and DP==8 (+EP), performance with the PR or without is the same.

examples/offline_inference/basic/basic.py
alexm-redhat: reverted

alexm-redhat:
@yewentao256 commands used:

**For client:**
`vllm bench serve --port 8123 --model deepseek-ai/DeepSeek-R1-0528 --dataset-name random --max-concurrency 32 --random-input-len 4 --random-output-len 1024 --num-prompts 80 --seed 1763417423 --percentile-metrics ttft,tpot,itl,e2el --metric-percentiles 90,95,99 --ignore-eos --trust-remote-code`

**For server:**

_For TP_
`vllm serve deepseek-ai/DeepSeek-R1-0528 --port 8123 --no-enable-prefix-caching --max-num-seqs 128 --max-model-len 16384 --tensor-parallel-size 8
`
_For DP_
`vllm serve deepseek-ai/DeepSeek-R1-0528 --port 8123 --no-enable-prefix-caching --max-num-seqs 128 --max-model-len 16384 --enable-expert-parallel --data-parallel-size 8`

Results, for TP, TPOT is 38.12ms with PR, 38.89 without PR, for DP, TPOT is 37.9 with PR and 37.6 without PR.

yewentao256 (approved):
`vllm bench serve --model deepseek-ai/DeepSeek-R1  --dataset-name random --host 127.0.0.1  --random-input-len 4 --random-output-len 1024 --request-rate inf --num-prompts 1024 --port 9256`

For larger batch size, seems not hurt in performance now

This PR

```bash
Successful requests:                     1024      
Benchmark duration (s):                  80.35     
Total input tokens:                      3072      
Total generated tokens:                  1048576   
Request throughput (req/s):              12.74     
Output token throughput (tok/s):         13050.34  
Peak output token throughput (tok/s):    14209.00  
Peak concurrent requests:                1024.00   
Total Token throughput (tok/s):          13088.57  
Mean TTFT (ms):                          1401.40   
Median TTFT (ms):                        1430.52   
P99 TTFT (ms):                           1462.59   
Mean TPOT (ms):                          77.08     
Median TPOT (ms):                        77.07     
P99 TPOT (ms):                           77.21     
Mean ITL (ms):                           77.08     
Median ITL (ms):                         76.91     
P99 ITL (ms):                            116.87    
```

Without

```bash
Successful requests:                     1024      
Benchmark duration (s):                  80.23     
Total input tokens:                      3072      
Total generated tokens:                  1048576   
Request throughput (req/s):              12.76     
Output token throughput (tok/s):         13069.67  
Peak output token throughput (tok/s):    14143.00  
Peak concurrent requests:                1024.00   
Total Token throughput (tok/s):          13107.96  
Mean TTFT (ms):                          1141.24   
Median TTFT (ms):                        1143.51   
P99 TTFT (ms):                           1225.30   
Mean TPOT (ms):                          77.23     
Median TPOT (ms):                        77.24     
P99 TPOT (ms):                           77.35     
Mean ITL (ms):                           77.23     
Median ITL (ms):                         77.24     
P99 ITL (ms):                            118.71    
```

And `lm_eval` looks good.

```bash
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9507|±  | 0.006|
|     |       |strict-match    |     5|exact_match|↑  |0.9492|±  | 0.006|
```

alexm-redhat:
@yewentao256 thanks for checking

mgoin (changes requested):
@alexm-redhat Can we please add unit test cases before landing? See where we currently skip many shapes https://github.com/vllm-project/vllm/blob/c3e29786209d91d3842e839b62f4d1d815902262/tests/kernels/quantization/test_block_fp8.py#L159-L162
And I think the current N and K in this file don't test all shapes of interest. If this test isn't currently running due to lack of Hopper in CI, we should add it to the Blackwell runner

mgoin (approved):

PR #28693, opened 2025-11-14
[feat] Implement Elastic Speculation: Adaptive Draft Length + Confidence-Based Early Exit

yuz207:
# Purpose

We implemented Elastic Speculation, an adaptive control layer for EAGLE speculative decoding that delivers __double-digit latency improvements__ over fixed-length speculation while __reducing KV-cache DRAM traffic__.

Two independent features:
1. Adaptive Draft Length: Dynamically adjusts speculation depth based on acceptance rates
2. Confidence-Based Early Exit: Gates KV writes for low-confidence draft tokens

# Summary

## Adaptive Draft Length

Captures multiple draft lengths as CUDA graphs (default: [5, 10, 15]), tracks
acceptance rate via EWMA, selects based on thresholds:

>70% acceptance  → d = high (maximize benefit)
50-70%          → d = med (balanced)
30-50%          → d = low (conservative)
<30%            → d = very low (minimize waste)

## Confidence-Based Early Exit

When draft token confidence drops below threshold, set slot_mapping = -1. CUDA
kernel early-returns, skipping KV write:

In reshape_and_cache kernel:
```
if (slot_idx < 0) return;  // No DRAM write
```

# Implementation

Core implemntation:
- vllm/config/speculative.py - Config options
- vllm/v1/cudagraph_dispatcher.py - Multi-draft-length graph support
- vllm/v1/spec_decode/eagle.py - Adaptive + early exit logic
- vllm/v1/spec_decode/metrics.py - EWMA tracking, draft selection

Environment variables:
```
VLLM_SPEC_ADAPTIVE_DRAFT_LENGTH = 0,1
VLLM_SPEC_CONFIDENCE_THRESHOLD = [0,1]
```

Config options:
```
SpeculativeConfig(
  draft_length_options = [d_1 ... d_n],  # None = auto-compute
  draft_confidence_threshold = c,    # 0.0 = disabled
)
```

# Testing and Results

Implementation was tested across draft lengths (5, 10, 15), early exit thresholds (0.3, 0.5, 0.7), and two Llama target + draft model pairs and evaluated on four benchmark datasets to simulate diverse inference workloads (Alpaca, SQuAD, BigCodeBench, and CNN DailyMail). 

We observe a ~20-50% reduction in latency depending on the task and configuration and confirm a threshold-proportional reduction in KV writes (~50% at 0.5 at 1-3% latency cost). In production, where as much as 70% of memory can be KV cache and as up as 50% of attention cycles stall on bandwidth constraints, this could potentially translate to latency savings as well. 

Our blog post with background, methodology, and full results can be found here: https://iluvatarlabs.com/blog/2025/11/elastic-speculation/

We're excited to PR this to vLLM and happy to address any feedback!


vllm/v1/spec_decode/eagle.py
```diff
@@ -360,11 +391,20 @@ def propose(
...
+            # Mask stopped sequences to PAD token for early stopping
+            if token_index > 0:
+                input_ids = torch.where(
+                    self.continue_mask[:batch_size],
+                    input_ids,
+                    self.pad_token_id
+                )
             if self.uses_mrope:
                 positions += 1
                 # NOTE(woosuk): We should handle the case where the draft model
```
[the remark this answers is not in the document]
yuz207: Good feedback. Regarding the first issue, we have re-implemented the full graph capture as an option, which provides an additional latency speed up over not using it, albeit at 3x the memory. The default remains to not use it though as adaptive draft length is actually still faster _without_ it than no adaptive baseline speculation _with_ it. The decision of whether to use it or will depend on specific workloads as larger batches will benefit less and the memory penalty increases with model size. It's nothing that the granularity of draft lengths can be changed as well in the code. Finally, we have also corrected the seq_lens counter to KV slot correctness; this should not have any effect on token stoppage, just potentially KV accuracy.

tools/profiling/run_nwor_microbench.py
```diff
@@ -0,0 +1,714 @@
...
+
+
+def run_microbenchmark(config: RunConfig) -> tuple[list[dict[str, Any]], dict[tuple[str, str], dict[str, float]]]:
+    prompts = pick_prompts(config)
+    results: list[dict[str, Any]] = []
+    metrics_delta: dict[tuple[str, str], dict[str, float]] = {}
+
+    # Set NWOR environment variables from config
+    os.environ["VLLM_NWOR_ADAPTIVE_DRAFT_LENGTH"] = str(config.adaptive_draft_length)
+    os.environ["VLLM_NWOR_CONFIDENCE_THRESHOLD"] = str(config.confidence_threshold)
```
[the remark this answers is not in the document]
yuz207: Developmental microbench script has been removed from PR.

heheda12345:
CC @benchislett @luccafong

robertgshaw2-redhat:
This looks very interesting!

robertgshaw2-redhat:
@MatthewBonanni @LucasWilkinson

vllm/config/speculative.py
```diff
@@ -454,6 +476,36 @@ def __post_init__(self):
...
+                n,  # Target
+                min(8, n * 2),  # Double (capped at 8)
+            ]
+            # Remove duplicates and sort
+            self.draft_length_options = sorted(set(self.draft_length_options))
+        # else: keep draft_length_options as None (fixed draft length)
+
+        # Override confidence threshold from environment if set
+        # This allows independent control of early exit
+        env_confidence = envs.VLLM_SPEC_CONFIDENCE_THRESHOLD
```
MatthewBonanni: Same as above, would prefer CLI argument. That would also simplify managing the default

vllm/model_executor/models/llama4_eagle.py
```diff
@@ -126,6 +127,14 @@ def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
                 weight_loader(param, loaded_weight, shard_id)
                 break
             else:
+                # if PP disabled then draft will share embed with target
+                if get_pp_group().world_size == 1 and "embed_tokens." in name:
+                    continue
+
+                # Skip loading extra bias for GPTQ models.
+                if name.endswith(".bias") and name not in params_dict:
```
MatthewBonanni, line 135: Why is this necessary?

vllm/model_executor/models/llama_eagle.py
```diff
@@ -154,6 +155,14 @@ def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
                 weight_loader(param, loaded_weight, shard_id)
                 break
             else:
+                # if PP disabled then draft will share embed with target
+                if get_pp_group().world_size == 1 and "embed_tokens." in name:
+                    continue
+
+                # Skip loading extra bias for GPTQ models.
+                if name.endswith(".bias") and name not in params_dict:
```
MatthewBonanni, line 163: ditto

vllm/model_executor/models/llama_eagle3.py
```diff
@@ -246,6 +246,10 @@ def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
                 weight_loader(param, loaded_weight, shard_id)
                 break
             else:
+                # Skip loading extra bias for GPTQ models.
+                if name.endswith(".bias") and name not in params_dict:
```
MatthewBonanni, line 250: ditto

vllm/v1/spec_decode/metrics.py
```diff
@@ -36,13 +39,94 @@ def new(cls, num_spec_tokens: int) -> "SpecDecodingStats":
...
+            finer-grained control and differentiate from very low acceptance.
+        """
+        if not draft_length_options:
+            return self.num_spec_tokens
+
+        # Sort to ensure consistent ordering
+        sorted_options = sorted(draft_length_options)
+
+        # Threshold-based selection
+        if self.acceptance_rate_ewma > 0.7:
```
MatthewBonanni: Is this something we would want to be tunable by the user? How universal is this default choice?

vllm/config/speculative.py
```diff
@@ -454,6 +476,36 @@ def __post_init__(self):
                         self.target_parallel_config, self.draft_tensor_parallel_size
                     )
                 )
+
+        # Auto-compute draft_length_options if not specified
+        if (
+            self.draft_length_options is None
+            and self.num_speculative_tokens is not None
+            and envs.VLLM_SPEC_ADAPTIVE_DRAFT_LENGTH
```
MatthewBonanni: Could we add this to SpeculativeConfig instead of using an env variable?

vllm/v1/spec_decode/eagle.py
```diff
@@ -406,13 +450,18 @@ def propose(
...
+            # consistent with actual KV writes (inactive ones are masked out).
+            # For data integrity when async scheduling, we use out-of-place operations
+            # in case they are modified in next step's `prepare_input` of main model.
+            length_increments = active_mask.to(common_attn_metadata.seq_lens.dtype)
+            common_attn_metadata.seq_lens += length_increments
             # This is an out-of-place operation to avoid modifying the original tensor.
-            common_attn_metadata.seq_lens_cpu = common_attn_metadata.seq_lens_cpu + 1
+            common_attn_metadata.seq_lens_cpu = (
+                common_attn_metadata.seq_lens_cpu
+                + active_mask.cpu().to(common_attn_metadata.seq_lens_cpu.dtype)
```
MatthewBonanni: This introduces a GPU-CPU sync point (which we've been trying to eliminate for `--async-scheduling`). Would it be possible to avoid this?

vllm/v1/spec_decode/eagle.py
```diff
@@ -379,11 +413,21 @@ def propose(
...
+            # Snapshot which requests are still emitting draft tokens before this
+            # iteration mutates the continue mask.
+            active_mask = self.continue_mask[:batch_size]
             # Update the inputs.
             # cast to int32 is crucial when eagle model is compiled.
             # tensor.argmax() returns int64 by default.
             input_ids = draft_token_ids_list[-1].int()
+            # Mask stopped sequences to PAD token for early stopping
+            if token_index > 0:
+                input_ids = torch.where(
```
MatthewBonanni: Could we use `masked_fill_` instead to avoid an extra allocation?

MatthewBonanni (commented):
Thanks for this contribution! This is a cool technique; sorry to make you wait so long for a review

MatthewBonanni:
Hi @yuz207, just wanted to check in! Do you think you'll have a chance to update the PR?

yuz207:
First off, happy new year and thanks everyone for the thoughtful review and comments! Apologies for the very belated response, I was offline for most of the holidays and finally catching up on everything. 

----

@MatthewBonanni Really appreciate the detailed comments in your review, Matthew. We're preparing an update we'll push shortly. 

Regarding the envs, we've switched over using the standard args in --speculative-config. 

For the model_executor scripts, it was something was mirrored from llama.py (it's already in deepseek_eagle.py and minicpm_eagle.py) because vLLM doesn't like the mismatch that occurs when quantized models are loaded from HF if they have bias params. 

Great call on using masked_fill_, now modifies tensor in place. 

On the extra CPU-GPU sync, It's a great call out since the whole point of this is for efficiency while minimizing unnecessary overhead. I'll have to think a little more about how we can avoid this and switch everything downstream to GPU tensors. In terms of low hanging fruit, we can reduce the # of syncs to just 1 for early exit if we move it outside the loop (and skip it entirely if not using early exit and just do seq_lens_cpu += draft_length). 

Finally, for the tunable threshold, we should absolutely make it configurable. We zeroed in on 0.7 as a default based on our tests but users should be able to select a threshold that meets their speed vs quality calculus. 

----

@fluctlux Thanks for sharing those links! I wasn't aware of them previously so I'll take some time this week to compare those proposed methods with how we implemented this approach. At a quick glance, these appear more similar to the adaptive draft length component rather than the confidence-based early exit. 

Terrific questions. I do believe these are valuable features to integrate as the savings to both latency and memory bandwidth can be quite substantial depending on the workload. Though I don't have hard evidence of this, it is plausible that frontier labs are already using similar techniques in their proprietary stacks given their current emphasis on latency tricks for serving. 

IMO the biggest obstacle is determining when to use elastic speculation in practice. While confidence-based early exit is pretty much universally beneficial in bandwidth-constrained settings, adaptive draft length has a small amount overhead so the cost-benefit depends on the target-draft model pair and accept rate. It's worth noting that in our benchmarking, we found that adaptive draft length even WITHOUT graph capture was faster than fixed draft length WITH it. Speaking of graph capture...

----

@Fsanic This is another terrific question. We actually only need to capture a single graph per batch size as multiple draft lengths produce duplicate graphs. The num_tokens dimension already differentiates CUDA graphs based on actual tensor shapes. So in practice, no additional memory overhead. 

Your question actually reminded me of some dead code that snuck into this PR that we're going to clean up. We did try a couple approaches to optimize multiple draft length graph capture (behind enable_draft_length_cudagraph_specialization), which is how we found doing so to be redundant and just wastes memory.

WoosukKwon (changes requested):
Thanks for the PR. I think we need more discussion before merging this PR, as dynamism is typically the biggest obstacle in achieving high gpu utilization.

PR #28701, opened 2025-11-14
Upstreaming aiter triton attention backend as a new backend

maleksan85:
Adding new Triton MLA to handle prefills in DS

## Server command:
```
VLLM_DISABLE_COMPILE_CACHE=1 \
AMDGCN_USE_BUFFER_OPS=1 \
VLLM_ROCM_USE_AITER=1 \
VLLM_ATTENTION_BACKEND=ROCM_AITER_TRITON_MLA \
VLLM_ROCM_USE_AITER_MLA=1 \
VLLM_ROCM_USE_TRITON_ROPE=1 \
vllm serve /data/models/deepseek-ai/DeepSeek-R1-0528 \
    --host localhost \
    --port 8000 \
    --swap-space 64 \
    --disable-log-requests \
    --dtype auto \
    --max-model-len 8192 \
    --tensor-parallel-size 8 \
    --max-num-seqs 1024  \
   --trust-remote-code \
    --block-size 1 \
    --gpu-memory-utilization 0.90  \
   --max-num-batched-tokens 131072 \
    --no-enable-prefix-caching  \
   --async-scheduling  \
   --enforce-eager
```

```
curl http://localhost:8000/v1/chat/completions -H "Content-Type: application/json" -d '{
    "model": "/data/models/deepseek-ai/DeepSeek-R1-0528",
    "messages": [
        {"role": "user", "content": "Who won the world series in 2023?"}
    ]}'
```

```
<think>\nOkay, the user is asking about the 2023 World Series winner. Hmm, this seems like a straightforward sports fact 
question, but I should double-check because the 2023 MLB season actually concluded in late 2023 (fall classic) for the 2023 
championship title. \n\nWait, I recall the Texas Rangers won their first-ever championship last year against the Diamondbacks. 
Let me mentally verify the details: It was a 4-1 series, Corey Seager got MVP, and Bruce Bochy managed them. The clinching 
Game 5 was especially dramatic with that late-inning comeback. \n\nThe user might just want a quick answer, but including the 
\"first title in franchise history\" context could add value since it's historically significant. No need to overcomplicate this unless 
they follow up with deeper questions about the series. \n\n...Though I wonder if they're testing if I know recent sports results? 
Either way, accurate and concise is best here. No signs this is homework – seems like casual curiosity. Just serve the facts 
cleanly.\n</think>....
```
## Correctness
```
lm_eval \
  --model local-completions \
  --model_args "base_url=http://localhost:8000/v1/completions,pretrained=/data/models/deepseek-ai/DeepSeek-R1-0528,tensor_parallel_size=8,add_bos_token=true,trust_remote_code=true" \
  --tasks gsm8k --num_fewshot 5 --limit 250   --batch_size 64
```
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.976|±  |0.0097|
|     |       |strict-match    |     5|exact_match|↑  |0.972|±  |0.0105|
```

## Benchmark
```
vllm bench serve \
  --host localhost \
  --port 8000 \
  --model /data/models/deepseek-ai/DeepSeek-R1-0528 \
  --dataset-name random \
  --random-input-len 1024 \
  --random-output-len 1024 \
  --max-concurrency 64 \
  --num-prompts 256 \
  --percentile-metrics ttft,tpot,itl,e2el \
  --ignore-eos
```

### ROCM_AITER_TRITON_MLA
```
Successful requests:                     256
Maximum request concurrency:             64
Benchmark duration (s):                  126.35
Total input tokens:                      261888
Total generated tokens:                  262144
Request throughput (req/s):              2.03
Output token throughput (tok/s):         2074.70
Peak output token throughput (tok/s):    2304.00
Peak concurrent requests:                128.00
Total Token throughput (tok/s):          4147.37
Mean TTFT (ms):                          1971.39
Median TTFT (ms):                        1965.70
P99 TTFT (ms):                           3307.36
Mean TPOT (ms):                          28.94
Median TPOT (ms):                        29.04
P99 TPOT (ms):                           30.26
Mean ITL (ms):                           28.94
Median ITL (ms):                         28.58
P99 ITL (ms):                            32.11
Mean E2EL (ms):                          31580.53
Median E2EL (ms):                        31330.76
P99 E2EL (ms):                           32561.38
```

### ROCM_AITER_MLA
```
Successful requests:                     256
Maximum request concurrency:             64
Benchmark duration (s):                  125.53
Total input tokens:                      261888
Total generated tokens:                  262144
Request throughput (req/s):              2.04
Output token throughput (tok/s):         2088.33
Peak output token throughput (tok/s):    2304.00
Peak concurrent requests:                128.00
Total Token throughput (tok/s):          4174.63
Mean TTFT (ms):                          1655.11
Median TTFT (ms):                        1503.20
P99 TTFT (ms):                           2111.98
Mean TPOT (ms):                          29.05
Median TPOT (ms):                        29.15
P99 TPOT (ms):                           30.46
Mean ITL (ms):                           29.05
Median ITL (ms):                         28.71
P99 ITL (ms):                            29.96
Mean E2EL (ms):                          31375.56
Median E2EL (ms):                        31374.27
P99 E2EL (ms):                           31465.52
```


vllm/v1/attention/backends/mla/triton_mla.py
```diff
@@ -35,14 +41,77 @@ def get_name() -> str:
...
+        if rocm_aiter_ops.is_mla_enabled():
+            return AiterMLAMetadataBuilder
+        return MLACommonMetadataBuilder
+
     @classmethod
     def supports_compute_capability(cls, capability: DeviceCapability) -> bool:
         return True
 
 
+class AiterTritonMLAImpl(AiterMLAImpl):
```
tjtanaa: Let's separate out into another class called `aiter_triton_mla.py`

maleksan85: sure, extracted

vllm/v1/attention/backends/mla/aiter_triton_mla.py
```diff
@@ -0,0 +1,56 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+from vllm.v1.attention.backends.mla.rocm_aiter_mla import AiterMLAImpl
+
+
+class AiterTritonMLAImpl(AiterMLAImpl):
```
tjtanaa, line 24: Could I know if you are intending to use AITER's mla decode or vLLM's triton mla decode backend?

From my understanding, what this PR is to fix the accuracy of Triton MLA backend on mi355x.
The cause is that ck flash attention seems to have some precision issues on mi355x.

Current changes are using AITER MLA Backend instead of Triton MLA Backend.
If you inherit `AiterMLAImpl`, you are essentially using `AITER MLA Backend` for the Decode pass.

There was once a vLLM's triton flash attention implementation before this PR https://github.com/vllm-project/vllm/pull/27611 . Generally, I would prefer to directly use a vLLM's own triton flash attention implementation is more suitable approach for the Triton MLA backend.

On second pass of reviewing the code, if you are reusing the vLLM's triton mla decode backend, we don't need to separate out AITER Triton MLA backend. 
We don't have to introduce new metadata, metadatabuilder nor impl class given that on MI355, the only way to get correct accuracy is to call the aiter's triton flash attention (as vLLM's triton flash attention was removed in PR https://github.com/vllm-project/vllm/pull/27611 .


In the `__init__()` of `TritonMLAImpl`,

```
    # check if it is gfx950:
    assert AITER_IS_FOUND, "TRITON MLA Backend on gfx950 requires AITER"

    # override the self.flash_attn_varlen_func
    from aiter.ops.triton.mha import flash_attn_varlen_func

    self.flash_attn_varlen_func = flash_attn_varlen_func

    self._flash_attn_varlen_diff_headdims = self._flash_attn_varlen_diff_headdims_aiter
```


Remove 

```
    def _flash_attn_varlen_diff_headdims(
        self, q, k, v, return_softmax_lse=False, softmax_scale=None, **kwargs
    ):
        return super()._flash_attn_varlen_diff_headdims(
            q,
            k,
            v,
            return_softmax_lse=return_softmax_lse,
            softmax_scale=softmax_scale,
            **kwargs,
        )
```

Add the following function
```
    def _flash_attn_varlen_diff_headdims_aiter(
        self, q, k, v, return_softmax_lse=False, softmax_scale=None, **kwargs
    ):
        result = self.flash_attn_varlen_func(
            q,
            k,
            v,
            softmax_scale=softmax_scale,
            return_lse=return_softmax_lse,
            **kwargs,
        )
        # Transpose the LSE if Triton MHA is used:
        # (q.shape[0], num_q_heads) to (num_q_heads, q.shape[0])
        if type(result) is tuple and return_softmax_lse:
            output, lse = result
            lse = lse.T.contiguous()
            return (output, lse)
        return result
```

maleksan85: no, added Trtiton MLA is particular for prefills in AITER MLA. In order to get alternative and upstream from internal fork.

vllm/v1/attention/backends/mla/triton_mla.py
```diff
@@ -35,9 +41,19 @@ def get_name() -> str:
         return "TRITON_MLA"
 
     @staticmethod
-    def get_impl_cls() -> type["TritonMLAImpl"]:
+    def get_impl_cls() -> type["TritonMLAImpl"] | type["AiterTritonMLAImpl"]:
```
gshtras: Why not make it a separate backend class, since it's a separate implementation?
Since there seem to now be:
triton MLA
AITER mla
AITER triton MLA
And only 2 backends to select - TRITON_MLA and ROCM_AITER_MLA
It's unclear which one you need to select with a combination of an AITER env var to get which outcome.
I suggest adding a whole new backend, and enabling it only through explicitly setting it, and not ROCM_VILLM_USE_AITER_...

maleksan85: just tried to avoid adding new flag as upstream do not appreciate new flags.

maleksan85: refactored

tjtanaa: @maleksan85 Thank you for the consideration. To add on, vLLM offers an explicit backend selection feature through `VLLM_ATTENTION_BACKEND` (this is for all platforms), we can just use `VLLM_ATTENTION_BACKEND=ROCM_AITER_TRITON_MLA` and `VLLM_ATTENTION_BACKEND=ROCM_AITER_MLA`. We don't have to introduce new flags.

SageMoore (approved):

vllm/v1/attention/backends/mla/triton_mla.py
maleksan85: yep, did that in the last revision, thanks!

gshtras (approved):

PR #28718, opened 2025-11-14
[Feature] Prefill Context Parallel (PCP) basic support

pisceskkk:
# Purpose
This PR, splited from full PR #26864, adds the basic supports for the Prefill Context Parallelism (PCP) feature, which corresponds to DCP. For specific implementation details, please refer to the RFC #25749.

- Modified files such as `block_tables.py` to extend the KV cache storage based on DCP & PCP;
- Added a communication group `pcp_group` for PCP;
- Added necessary command-line arguments to control parallelism for PCP. Temporarily disabled PCP parameters until backend support is complete, then re-enable;
- Added PCP-related parameters to the attention backend prototype class;

CC @LookAround0301 @FENP @LucasWilkinson


vllm/engine/arg_utils.py
```diff
@@ -758,7 +757,12 @@ def add_cli_args(parser: FlexibleArgumentParser) -> FlexibleArgumentParser:
         )
         parallel_group.add_argument(
             "--dcp-kv-cache-interleave-size",
```
luccafong, line 772: we may need to add both `--dcp-kv-cache-interleave-size` and `--cp-kv-cache-interleave-size` to align the interface  and deprecate --dcp-xx later

pisceskkk: yes, this makes sense, I changed this.

vllm/v1/core/kv_cache_utils.py
```diff
@@ -1219,10 +1219,21 @@ def _report_kv_cache_config(
...
+    if (
+        vllm_config.parallel_config.decode_context_parallel_size
+        * vllm_config.parallel_config.prefill_context_parallel_size
+        > 1
+    ):
+        cp_size = (
+            vllm_config.parallel_config.prefill_context_parallel_size
+            * vllm_config.parallel_config.decode_context_parallel_size
+        )
+        num_tokens *= cp_size
```
LucasWilkinson: nit: can you reduce line spill by doing
```
pcp_size = vllm_config.parallel_config.decode_context_parallel_size
dcp_size = vllm_config.parallel_config.prefill_context_parallel_size
if pcp_size * dcp_size > 1:
       num_tokens *= (pcp_size * dcp_size)
```

vllm/v1/attention/backends/utils.py
```diff
@@ -1079,37 +1079,37 @@ def compute_causal_conv1d_metadata(query_start_loc_p: torch.Tensor):
 
 def get_dcp_local_seq_lens(
     seq_lens: torch.Tensor,
-    dcp_world_size: int = 1,
-    dcp_rank: int | None = None,
-    dcp_kv_cache_interleave_size: int = 1,
+    cp_size: int = 1,
+    cp_rank: int | None = None,
```
LucasWilkinson: nit: these will still be `dcp` specific right?

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1809,6 +1834,20 @@ def forward_impl(
...
+                        states,
+                        dim=0,
+                    )
+
+                if (
+                    not self.is_sequence_parallel
+                    and self.reduce_results
+                    and (self.tp_size > 1 or self.ep_size > 1)
+                ):
+                    states = self.maybe_all_reduce_tensor_model_parallel(states)
```
LucasWilkinson: what is this change for?

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1753,6 +1765,19 @@ def forward_impl(
...
+            # we should modify All2AllManager abstract to better support PCP.
+            if self.pcp_size > 1:
+                hidden_states = get_pcp_group().all_gather(
+                    hidden_states,
+                    dim=0,
+                )
+                router_logits = get_pcp_group().all_gather(
+                    router_logits,
+                    dim=0,
+                )
```
LucasWilkinson, line 1779: can you explain why this all_gather is needed? Why can't we just dispatch tokens directly from there existing PCP ranks?

vllm/v1/core/kv_cache_utils.py
pisceskkk: sure, we will change this

vllm/v1/attention/backends/utils.py
pisceskkk: no, we will split KVCache among PCP devices too, so this local_seq_len func will be used as following:
```python
total_cp_size = pcp_size * dcp_size
total_cp_rank = pcp_rank * dcp_size + dcp_rank
local_lens = get_cp_local_seq_lens(lens, total_cp_size, total_cp_rank, cp_kv_cache_interleave_size)
```
but, maybe we could change this func in another PR insteal of this basic support?

actually, I'm not entirely sure whether some of the shared functions and variables for PCP and DCP should be included in this PR. I would appreciate your guidance on this matter!

vllm/model_executor/layers/fused_moe/layer.py
pisceskkk: Apologies, this code was mistakenly included due to my error. It existed in a previous PR but was optimized out in a later commit. However, I incorrectly copied it over during the migration process. The code has now been removed to ensure proper functionality of TP+MoE.

vllm/model_executor/layers/fused_moe/layer.py
FENP: > can you explain why this all_gather is needed? Why can't we just dispatch tokens directly from there existing PCP ranks?

Yes, indeed, the all-gather here acts as the dispatch operation. Likewise, the following reduce-scatter handles the combine phase. This is equivalent to the code below.
https://github.com/vllm-project/vllm/blob/f67299f66d71b593401b1f77540726005dfe67da/vllm/distributed/device_communicators/all2all.py#L111-L147

This is necessary because when PCP is enabled, `do_naive_dispatch_combine` is not set to True, so the code snippet above will not be executed.
```python
  if do_naive_dispatch_combine:
      hidden_states_combined, router_logits = get_ep_group().dispatch(
          hidden_states, router_logits, self.is_sequence_parallel
      )
```

Frankly, I attempted to set `do_naive_dispatch_combine` to True with PCP enabled, so we could reuse the current implementation. However, the current `All2AllManager` implementation is tightly coupled with DP, and supporting PCP would require extensive code changes. For example.
https://github.com/vllm-project/vllm/blob/f67299f66d71b593401b1f77540726005dfe67da/vllm/distributed/device_communicators/all2all.py#L120-L131

For simplicity, I went with the current approach. Any suggestion?

LucasWilkinson: got it, thanks!; yes this is fine for now; what test setup are you using (like `vllm serve ...` I'd love to poke around a bit)

how are we guaranteeing that all ranks have same number of tokens in the PCP case?

vllm/v1/attention/backends/utils.py
LucasWilkinson: lets not rename the variables in this function yet since it will cause confusion with the function name `get_dcp_local_seq_lens`; we can update it in the following PR

vllm/model_executor/layers/fused_moe/layer.py
FENP: > got it, thanks!; yes this is fine for now; what test setup are you using (like `vllm serve ...` I'd love to poke around a bit)

This PR includes only partial code. To fully test PCP, you can give a try in this [branch](https://github.com/LookAround0301/vllm/tree/pcp_pr).
Here’s an setup example.
```shell
vllm serve Qwen/Qwen3-32B --gpu-memory-utilization 0.9 --tensor-parallel-size 2 --prefill-context-parallel-size 2
```
> how are we guaranteeing that all ranks have same number of tokens in the PCP case?

We currently apply padding so that each rank handles an equal number of tokens after splitting. This logic is not included in this PR — the detailed code can be found here.

https://github.com/LookAround0301/vllm/blob/a58383337bbd7ef0a6b072737af257b288e36c2c/vllm/v1/worker/gpu_model_runner.py#L1081-L1083

vllm/v1/attention/backends/utils.py
pisceskkk: > lets not rename the variables in this function yet since it will cause confusion with the function name `get_dcp_local_seq_lens`; we can update it in the following PR

sure, we will change this

vllm/v1/core/kv_cache_manager.py
```diff
@@ -124,12 +125,12 @@ def __init__(
...
             ].kv_cache_spec.block_size
 
-            if dcp_world_size > 1:
+            if dcp_world_size * pcp_world_size > 1:
                 assert len(kv_cache_config.kv_cache_groups) == 1
                 # Note(hc): need revisit. When both DCP and any future
                 # PCP are enabled, the block_size may need to be scaled
                 # by a factor of dcp_size × pcp_size?
-                self.block_size *= dcp_world_size
+                self.block_size *= dcp_world_size * pcp_world_size
```
LucasWilkinson, line 130: nit: should we remove the comment?

vllm/v1/core/kv_cache_utils.py
```diff
@@ -1219,10 +1219,15 @@ def _report_kv_cache_config(
...
+    pcp_size = vllm_config.parallel_config.prefill_context_parallel_size
+    if pcp_size * dcp_size > 1:
+        num_tokens *= pcp_size * dcp_size
         logger.info(
-            "Multiplying the GPU KV cache size by the dcp_world_size %d.",
+            "Multiplying the GPU KV cache size by the cp_world_size %d "
+            "(pcp_world_size %d * dcp_world_size %d).",
+            pcp_size * dcp_size,
+            vllm_config.parallel_config.prefill_context_parallel_size,
             vllm_config.parallel_config.decode_context_parallel_size,
```
LucasWilkinson: nit: replace with `pcp_size`, `dcp_size`

vllm/model_executor/layers/fused_moe/config.py
LucasWilkinson: can we remove this no and replace usages with `flatten_tp_across_dp_and_pcp`

LucasWilkinson (commented):
Overall looks pretty good to me; running the CI, left a couple of final nits/comments

vllm/v1/core/kv_cache_manager.py
pisceskkk: thanks for review, removed this comment

vllm/v1/core/kv_cache_utils.py
pisceskkk: sure, changed this

vllm/model_executor/layers/fused_moe/config.py
pisceskkk: sure, removed  and replaced this function

LucasWilkinson (approved):
LGTM

heroes999:
> > If we set PCP=2, TP=8 but without DCP, what's the KVCache split strategy? Input prompt segmented between ranks, but decoding output tokens containing copies?
> 
> Different PCP ranks contain split sequence dimensions and distinct KV caches, while copies may exist within the same PCP rank (depending on MLA or GQA configurations, specifically the relationship between num_kv_heads and tp_size).
> 
> In your example, assuming we are using an MLA model, all KV caches belonging to PCP0 are identical. The KV caches of PCP0 and PCP1 are different, following the same distribution pattern as DCP, where they are stored in interleave-style on cp_kvcache_interleave_size.

Does it mean in decoding phase，pcp acts similarly to dcp, which requires gather q, calc local attn, correct lse..., sharding new tokens' kv cache interleavely?

pisceskkk:
> > > If we set PCP=2, TP=8 but without DCP, what's the KVCache split strategy? Input prompt segmented between ranks, but decoding output tokens containing copies?
> > 
> > 
> > Different PCP ranks contain split sequence dimensions and distinct KV caches, while copies may exist within the same PCP rank (depending on MLA or GQA configurations, specifically the relationship between num_kv_heads and tp_size).
> > In your example, assuming we are using an MLA model, all KV caches belonging to PCP0 are identical. The KV caches of PCP0 and PCP1 are different, following the same distribution pattern as DCP, where they are stored in interleave-style on cp_kvcache_interleave_size.
> 
> Does it mean in decoding phase，pcp acts similarly to dcp, which requires gather q, calc local attn, correct lse..., sharding new generated tokens' kv cache interleavely?

> https://github.com/vllm-project/vllm/pull/28988 , it seems dcp&pcp are merged to cp concept, so they do the same thing in decoding phase?

yes, in decoding phase, they act similarly for kvcache. If I remember correctly, there are only two differences: one is that dcp introduce an all-gather op for Q in head dim, and another is that dcp need a reduce-scatter after update output using lse.

PR #28723, opened 2025-11-14
[Feature] Support Prefill Context Parallel (PCP) for GQA flashinfer

pisceskkk:
# Purpose
This PR, splited from full PR #26864, adds the supports for the Prefill Context Parallelism (PCP) with GQA flashinfer, following PR #28718. For specific implementation details, please refer to the RFC [#25749](https://github.com/vllm-project/vllm/issues/25749).  

- Modified `ModelRunner.py` for PCP partitioning logic for tokens;  
- Modified `flashinfer.py` to adapt the FlashInfer backend for GQA to PCP.
- Added `PrefillContextParallelMetadata` shared across attention backends;
- Renamed variables and functions shared by both PCP and DCP.

# Test Plan
Qwen/Qwen2.5-3B

```
vllm serve Qwen/Qwen2.5-3B --tensor-parallel-size 4 --decode-context-parallel-size 2 --prefill-context-parallel-size 2 --dcp-kv-cache-interleave-size 8
```

# Test Result
gsm8k eval

tp4 17c540a993af88204ad1b78345c8a865cf58ce44
| dataset | version | metric | mode | vllm-api-general-stream |
| gsm8kdataset | - | avg@5 | gen | 72.78 |

tp4 dcp2 interleave 8
| dataset | version | metric | mode | vllm-api-general-stream |
| gsm8kdataset | - | avg@5 | gen | 72.43 |

tp4 pcp2 interleave 8
| dataset | version | metric | mode | vllm-api-general-stream |
| gsm8kdataset | - | avg@5 | gen | 72.51 |

tp4 dcp2 pcp2 interleave 8
| dataset | version | metric | mode | vllm-api-general-stream |
| gsm8kdataset | - | avg@5 | gen | 72.98 |

CC @LookAround0301 @FENP @gjc0824 @LucasWilkinson

### DIFF
--- a/vllm/v1/attention/backends/flashinfer.py
+++ b/vllm/v1/attention/backends/flashinfer.py
@@ -4,6 +4,7 @@
 
 from dataclasses import dataclass
 from typing import ClassVar
+from xxlimited import Str
 
 import numpy as np
 import torch
@@ -57,7 +58,9 @@ from vllm.v1.attention.backends.utils import (
     PrefillContextParallelMetadata,
     get_cp_local_seq_lens,
     get_kv_cache_layout,
+    get_kv_indices,
     get_per_layer_parameters,
+    get_q_indices,
     infer_global_hyperparameters,
     split_decodes_and_prefills,
 )
@@ -171,6 +174,7 @@ class BatchCPPrefillWrapper:
         self,
         dcp_world_size,
         pcp_world_size,
+        max_num_reqs,
         workspace_buffer: torch.Tensor | None = None,
     ):
         self.dcp_world_size = dcp_world_size
@@ -185,10 +189,61 @@ class BatchCPPrefillWrapper:
             self._new_tokens_tail = BatchPrefillWithRaggedKVCacheWrapper(
                 workspace_buffer, get_kv_cache_layout()
             )
+            pin_memory = is_pin_memory_available()
+            self.pcp_q_indptr_cpu = torch.zeros(
+                max_num_reqs + 1, dtype=torch.int32, device="cpu", pin_memory=pin_memory
+            )
+            self.pcp_q_indptr_np = self.pcp_q_indptr_cpu.numpy()
+            self.kv_for_head_indptr_cpu = torch.zeros(
+                max_num_reqs + 1, dtype=torch.int32, device="cpu", pin_memory=pin_memory
+            )
+            self.kv_for_head_indptr_np = self.kv_for_head_indptr_cpu.numpy()
+            self.kv_for_tail_indptr_cpu = torch.zeros(
+                max_num_reqs + 1, dtype=torch.int32, device="cpu", pin_memory=pin_memory
+            )
+            self.kv_for_tail_indptr_np = self.kv_for_tail_indptr_cpu.numpy()
         else:
             self._new_tokens = BatchPrefillWithRaggedKVCacheWrapper(
                 workspace_buffer, get_kv_cache_layout()
             )
+
+    def _iter_wrappers(self) -> list:
+        wrappers = [self._context]
+        if self.pcp_world_size > 1:
+            wrappers.extend([self._new_tokens_head, self._new_tokens_tail])
+        else:
+            wrappers.append(self._new_tokens)
+        return wrappers
+
+    def _get_consistent_attr(self, attr_name: str):
+        wrappers = self._iter_wrappers()
+        base_value = getattr(wrappers[0], attr_name)
+        for wrapper in wrappers[1:]:
+            value = getattr(wrapper, attr_name)
+            assert (
+                value == base_value
+            ), f"Inconsistent {attr_name} detected across CP prefill wrappers."
+        return base_value
+
+    @property
+    def _window_left(self):
+        return self._get_consistent_attr("_window_left")
+
+    @property
+    def _logits_soft_cap(self):
+        return self._get_consistent_attr("_logits_soft_cap")
+
+    @property
+    def _sm_scale(self):
+        return self._get_consistent_attr("_sm_scale")
+
+    def _assert_causal(self) -> None:
+        assert not self._context._causal
+        if self.pcp_world_size > 1:
+            assert self._new_tokens_head._causal
+            assert self._new_tokens_tail._causal
+        else:
+            assert self._new_tokens._causal
 
     def plan(
         self,
@@ -208,8 +263,8 @@ class BatchCPPrefillWrapper:
         kv_cache_dtype: torch.dtype,
         prefill_fixed_split_size: int,
         disable_split_kv: bool,
-        pcp_metadata: PrefillContextParallelMetadata | None,
-    ):
+        device: str,
+    ) -> PrefillContextParallelMetadata:
         """Plan the prefill operation with given parameters."""
         common_args = [
             num_kv_heads,
@@ -236,13 +291,30 @@ class BatchCPPrefillWrapper:
             **common_kwargs,
         )
         if self.pcp_world_size > 1:
-            assert pcp_metadata is not None
-            qo_indptr_cpu = pcp_metadata.q_head_start_loc
-            kv_for_head_indptr = pcp_metadata.kv_for_head_indptr
-            kv_for_tail_indptr = pcp_metadata.kv_for_tail_indptr
-            self._new_tokens_head.plan(
+            self.pcp_q_indptr_cpu[: qo_indptr_cpu.shape[0]] = qo_indptr_cpu // 2
+            pcp_q_indptr_cpu = self.pcp_q_indptr_cpu[: qo_indptr_cpu.shape[0]]
+            self.kv_for_head_indptr_cpu[: qo_indptr_cpu.shape[0]] = (
+                (self.pcp_rank + 1) * pcp_q_indptr_cpu
+            )
+            self.kv_for_tail_indptr_cpu[: qo_indptr_cpu.shape[0]] = (
+                (2 * self.pcp_world_size - self.pcp_rank) * pcp_q_indptr_cpu
+            )
+            q_head_indices, q_tail_indices = get_q_indices(
                 qo_indptr_cpu,
-                kv_for_head_indptr,
+                self.pcp_q_indptr_np[: qo_indptr_cpu.shape[0]],
+            )
+            q_head_indices_gpu = q_head_indices.to(device)
+            q_tail_indices_gpu = q_tail_indices.to(device)
+            q_full_indices = torch.cat([q_head_indices_gpu, q_tail_indices_gpu]).argsort()
+            kv_for_head_indices, kv_for_tail_indices = get_kv_indices(
+                qo_indptr_cpu * self.pcp_world_size,
+                self.kv_for_head_indptr_np[: qo_indptr_cpu.shape[0]],
+                self.kv_for_tail_indptr_np[: qo_indptr_cpu.shape[0]],
+            )
+            
+            self._new_tokens_head.plan(
+                pcp_q_indptr_cpu,
+                self.kv_for_head_indptr_cpu[: qo_indptr_cpu.shape[0]],
                 num_qo_heads,
                 *common_args,
                 causal=True,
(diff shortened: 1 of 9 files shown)


vllm/v1/worker/gpu_model_runner.py
```diff
@@ -492,7 +498,55 @@ def __init__(
 
         # Only relevant for multimodal models
         if self.supports_mm_inputs:
-            self.is_mm_embed = self._make_buffer(self.max_num_tokens, dtype=torch.bool)
+            self.is_mm_embed = self._make_buffer(max_num_tokens, dtype=torch.bool)
+
+        # Persistent buffers for Prefill Context Parallism
+        if self.pcp_world_size > 1:
```
LucasWilkinson, line 504: can we please separate all of this into a `PCPManager` or a utils file to make it more modular and easier to migrate to model runner v2?

LucasWilkinson (commented):
Left some comments on https://github.com/vllm-project/vllm/pull/28988 which I think similarly apply here

vllm/v1/spec_decode/eagle.py
```diff
@@ -644,7 +644,9 @@ def prepare_inputs_padded(
             block_table_tensor=common_attn_metadata.block_table_tensor,
             slot_mapping=common_attn_metadata.slot_mapping[token_indices],
             causal=True,
-            dcp_local_seq_lens=common_attn_metadata.dcp_local_seq_lens,
+            cp_local_seq_lens=common_attn_metadata.cp_local_seq_lens,
+            cp_local_seq_lens_cpu=common_attn_metadata.cp_local_seq_lens_cpu,
+            pcp_metadata=common_attn_metadata.pcp_metadata,
```
LucasWilkinson: if PCP + spec-decode is not yet supported should we be passing this blindly?

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -1218,28 +1353,37 @@ def forward(
...
+                    wrappers_to_check = [(prefill_wrapper._context, False)]
+                    if self.pcp_world_size > 1:
+                        wrappers_to_check.extend(
+                            [
+                                (prefill_wrapper._new_tokens_head, True),
+                                (prefill_wrapper._new_tokens_tail, True),
+                            ]
+                        )
+                    else:
+                        wrappers_to_check.append((prefill_wrapper._new_tokens, True))
```
LucasWilkinson: this is kinda messy imo; can you just do something like:

```
class BatchCPPrefillWrapper:
        @property
         def _window_left(self):
                assert self._context._window_left == self._new_tokens._window_left
                return self._context._window_left
         ...
```

Lests also do:
```
class FlashInferImpl:
       def __init__(self, ...):
             self.is_debugging_mode = envs.VLLM_LOGGING_LEVEL == "DEBUG"
```
and then only run the asserts if `self.is_debugging_mode` to avoid excessive CPU overhead on each forward pass

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -1161,6 +1279,23 @@ def forward(
 
         num_actual_tokens = attn_metadata.num_actual_tokens
 
+        if self.pcp_world_size > 1:
```
LucasWilkinson, line 1389: can this be moved into `BatchCPPrefillWrapper.run(...)`?

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -1334,28 +1478,27 @@ def forward(
...
                     )
-                    output[:num_decode_tokens] = cp_lse_ag_out_rs(
-                        output_tmp, lse, get_dcp_group()
-                    )
+                    if self.dcp_world_size > 1:
+                        out, lse = cp_lse_ag_out_rs(
+                            out,
+                            lse,
+                            get_dcp_group(),
+                            return_lse=True,
```
LucasWilkinson: should this be `return_lse=self.pcp_world_size > 1`

vllm/v1/attention/backends/utils.py
```diff
@@ -49,6 +49,27 @@ def is_valid_kv_cache_layout(value: str) -> bool:
...
+    We split and concatenate the sequence in a head-tail style,
+    and use this variable to restore the original order.
+    """
+    q_head_indices: torch.Tensor | None = None
+    q_tail_indices: torch.Tensor | None = None
+    q_head_start_loc: torch.Tensor | None = None
+    kv_for_head_indices: torch.Tensor | None = None
+    kv_for_tail_indices: torch.Tensor | None = None
+    kv_for_head_indptr: torch.Tensor | None = None
+    kv_for_tail_indptr: torch.Tensor | None = None
```
LucasWilkinson: this seems FlashInfer specific? we shouldn't have backend specific things (or styles) in `CommonAttentionMetadata`

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -195,39 +208,126 @@ def plan(
...
+        kv_for_head_indices = metadata.kv_for_head_indices
+        kv_for_tail_indices = metadata.kv_for_tail_indices
+        q_full_indices = metadata.q_full_indices
+
+        q_head = torch.index_select(query, 0, q_head_indices)
+        q_tail = torch.index_select(query, 0, q_tail_indices)
+        k_head = torch.index_select(key, 0, kv_for_head_indices)
+        v_head = torch.index_select(value, 0, kv_for_head_indices)
+        k_tail = torch.index_select(key, 0, kv_for_tail_indices)
+        v_tail = torch.index_select(value, 0, kv_for_tail_indices)
```
LucasWilkinson: so many `index_select` seems very expensive; have you profiled this?

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -195,39 +208,126 @@ def plan(
...
+                qo_indptr_cpu,
+                num_qo_heads,
+                num_kv_heads=num_kv_heads,
+                head_dim_qk=head_dim,
+                head_dim_vo=head_dim,
+                causal=True,  # This is newtokens run
+                **common_kwargs,
+            )
+
+    def _attention_with_head_and_tail(
```
LucasWilkinson: can this be moved into the `BatchCPPrefillWrapper`?

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -195,39 +208,126 @@ def plan(
...
+                q_full_indices,
+            )
+            return output, lse
+        else:
+            output = torch.index_select(
+                torch.cat([output_head, output_tail], dim=0),
+                0,
+                q_full_indices,
+            )
+            return output
```
LucasWilkinson:
```suggestion
        if return_lse:
            output_head, lse_head = output_head
            output_tail, lse_tail = output_tail
              lse = torch.index_select(
                  torch.cat([lse_head, lse_tail], dim=0),
                  0,
                  q_full_indices,
              )
             
        output = torch.index_select(
            torch.cat([output_head, output_tail], dim=0),
            0,
            q_full_indices,
        )
         
         return output if not return_lse else (output, lse)
```

LucasWilkinson (commented):
Thanks for the contribution! A few more comments

vllm/v1/attention/backends/flashinfer.py
gjc0824: Thanks. We have changed it.

vllm/v1/attention/backends/flashinfer.py
gjc0824: Sure. We have moved it into `BatchCPPrefillWrapper.run()`.

vllm/v1/attention/backends/flashinfer.py
gjc0824: Yes, we profiled it. Because PCP is only meaningful when the input sequence is long, we conducted performance tests in scenarios where the input sequence length is 32k.
`index_select` is negligible compared to the benefits brought by PCP.

vllm/v1/attention/backends/flashinfer.py
gjc0824: Sure. We have changed it for less unnecessary calculation.

vllm/v1/attention/backends/flashinfer.py
gjc0824: No. All gather & restore KV is required in both prefilling and decoding phrase so it cannot be moved into PrefillWrapper. But we have extracted it to `vllm/v1/attention/backends/utils.py` as a general function, which can be used by other backends.

vllm/v1/attention/backends/flashinfer.py
gjc0824: Thanks for your comments. We have changed it.

vllm/v1/spec_decode/eagle.py
gjc0824: Sure. We have removed these params to avoid misunderstanding.

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -238,37 +337,48 @@ def run(
...
-            kv_cache_permute,
-            k_scale=layer._k_scale_float,
-            v_scale=layer._v_scale_float,
-            return_lse=True,
-        )
-        output_context, lse_context = cp_lse_ag_out_rs(
-            output_context_tmp, lse_context_tmp, get_dcp_group(), return_lse=True
-        )
-        lse_context = lse_context.transpose(0, 1).contiguous()
+        if self.pcp_world_size > 1:
```
Livinfly, line 407: I wonder whether we forgot to add the calculation between `kv_cache_permute` and `prefill_query` here?
Then something like `cp_lse_ag_out_rs`, `cp_lse_ag_out_ar` and `merge_attn_states` correct the output?

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -238,37 +337,48 @@ def run(
...
+                key,
+                value,
+                pcp_metadata,
+                return_lse=True,
+            )
+        else:
+            prefill_query_across_dcp = get_dcp_group().all_gather(
+                prefill_query.contiguous(), dim=1
+            )
+            output_context_tmp, lse_context_tmp = self._context.run(
```
Livinfly, line 458: As we store the input keys and values in the cache via `reshape_and_cache_flash` in `FlashInferImpl.forward`, this `_context` calculate between `kv_cache_permute` and `prefill_query_across_dcp` will lead to leak / recompute **the input keys and values** ?

vllm/v1/worker/gpu_model_runner.py
gjc0824: OK. We will make similar changes as [#28988 ](https://github.com/vllm-project/vllm/pull/28988).

vllm/v1/attention/backends/utils.py
gjc0824: Sure. `*_indptr` params are Flashinfer specific. So we remove `*_indptr` params from general `PrefillContextParallelMetadata` and moved the computation of `PrefillContextParallelMetadata` to the stage of building attention wrapper. 
When building attention wrapper, we compute the `*_indptr` params in specific `flashinfer.py` and then use general functions such as `get_q_indices` to get `PrefillContextParallelMetadata` [code](https://github.com/vllm-project/vllm/pull/28723/files#diff-08fafcaeafe195852ef3ea32b75cdaeb2b1fb3d7617857f4231f7e0dbade7623R304). We think this refactoring not only distinguishes the handling of backend-specific parameters but also reduces the dependency on modelrunner.

vllm/v1/attention/backends/flashinfer.py
gjc0824: Thanks for your comments. Current PCP on flashinfer temporarily does not support chunk prefilling and prefix caching so  the kv context can be ignored. Subsequent PRs will add further support.

vllm/v1/attention/backends/flashinfer.py
gjc0824: Thanks. Similar reasons as last comment.

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -164,17 +170,83 @@ def trtllm_prefill_attn_kvfp8_dequant(
...
-        )
+        if self.pcp_world_size > 1:
+            self._new_tokens_head = BatchPrefillWithRaggedKVCacheWrapper(
+                workspace_buffer, get_kv_cache_layout()
+            )
+            self._new_tokens_tail = BatchPrefillWithRaggedKVCacheWrapper(
+                workspace_buffer, get_kv_cache_layout()
+            )
+            pin_memory = is_pin_memory_available()
+            self.pcp_q_indptr_cpu = torch.zeros(
```
FENP, line 196: Just wondering — is this wrapper expected to be instantiated only once? I'm concerned about potential repeated pinned memory allocation across instances.

pisceskkk: Yes, to my knowledge, these wrappers are initialized only once. Subsequent inference processes will only invoke the `plan` and `run` functions.

PR #24812, opened 2025-09-13
[Kernels] Modular kernel refactor

bnellnm:
## Purpose
Simplify `modular_kernel.py`, mainly the `forward` method of FusedMoEModularKernel.
- Fewer levels of indirection in `forward`
- Simplify `workspace_shape` method by passing in `M` and current/chunked `M` sizes instead of tensors.
- Separate getting the workspace type into a method that defaults to the activation dtype.  Only a few experts can override this.
- Factor out `FusedMoEModularKernel` prepare/finalize code into methods since the async/DBO features have made them more complicated.
- Reduce number of special cases, esp. wrt chunking.
- More robust chunking/reduce/do naive flags in `layer.py`

## Test Plan
Run tests/kernels/moe.

## Test Result
All tests pass.

cc @varun-sundar-rabindranath , @SageMoore , @LucasWilkinson

PR #28728, opened 2025-11-14
[BugFix] Fix misprint introduced by modular_kernel refactoring.

halyavin:
This bug was introduced during modular_kernel refactoring in #24812. 

Correct code before the refactoring:
https://github.com/vllm-project/vllm/blame/f08919b7d17f6adf4eea26ca410c4a3de41a71da/vllm/model_executor/layers/fused_moe/modular_kernel.py#L833


mgoin (approved):
Nice find, thank you for the fix cc @bnellnm

bnellnm (approved):
Good catch!

PR #28729, opened 2025-11-14
[Bugfix] Multiple fixes for gpt-oss Chat Completion prompting

bbrowning:
## Purpose

This fixes multiple issues with how we were prompting gpt-oss models for Chat Completion requests. A summary of the fixes:

- We were not setting the recipient to "assistant" on tool responses
- We were not mapping the tool_call_id to the proper function name on tool responses, resulting in the model getting confused about which tool response matched which tool call.
- When the model generates tool calls during a turn, we return any model output to the commentary channel as the `content` of that tool call response. We were not properly restoring this content as a commentary message when those tool calls were passed back in via Chat Completion messages.
- When the model generates reasoning messages during a tool call turn, we return that as a `reasoning` properly on the Chat Completion response. We were not properly restoring that reasoning content as a message to the analysis channel when subsequent turns passed that back in.
- When non-tool reasoning responses from the model were passed back in via the `reasoning` extra_body in subsequent turns, we were not properly turning those back into analysis messages. This is the same problem we had with reasoning in tool calls, but a slightly different fix for the non-tool call path.
- We were not properly marking the channel as `final` for non-tool, non-reasoning model output when generating the Harmony messages for subsequent turns.
- We were not dropping analysis messages that precede a model output to the `final` channel in all cases, resulting in us often showing the model analysis messages from previous turns where it has already generated a final output leading to increased token usage, context pollution, and not following the best practices for handling raw chain of thought outlined by OpenAI.

I added a series of tests that pass in Chat Completion requests and thoroughly validate the exact Harmony messages generated from those requests to test for all of the cases above.


## Test Plan

### Unit Tests

```
pytest tests/entrypoints/openai/test_serving_chat.py::TestServingChatWithHarmony

pytest tests/entrypoints/openai/parser/test_harmony_utils.py
```

### Berkeley Function Calling Leaderboard v4 Results

Run vLLM built with this fix:
```
vllm serve openai/gpt-oss-120b \
  --tool-call-parser openai \
  --reasoning-parser openai_gptoss \
  --enable-auto-tool-choice \
  --tensor-parallel-size 4
```

Run `bfcl generate` against that vLLM:
```
# Assumes you're inside a local clone of bfcl and adjusted it to use openai/gpt-oss-120b model
cat <<EOF >> bfcl_eval/constants/model_config.py
MODEL_CONFIG_MAPPING = {
    "openai/gpt-oss-120b": ModelConfig(
        model_name="openai/gpt-oss-120b",
        display_name="openai/gpt-oss-120b (FC) (vLLM)",
        url="https://huggingface.co/openai/gpt-oss-120b",
        org="OpenAI",
        license="apache-2.0",
        model_handler=OpenAICompletionsHandler,
        input_price=None,
        output_price=None,
        is_fc_model=True,
        underscore_to_dot=True,
    ),
}
EOF

OPENAI_BASE_URL="http://localhost:8000/v1" \
OPENAI_API_KEY="fake" \
bfcl generate \
  --model openai/gpt-oss-120b \
  --test-category multi_turn \
  --num-threads 8
```

## Test Result

### Unit Tests

```
pytest -q --disable-warnings \
  tests/entrypoints/openai/test_serving_chat.py::TestServingChatWithHarmony
...........                                                       [100%]
11 passed, 5 warnings in 8.65s

pytest -q --disable-warnings \
  tests/entrypoints/openai/parser/test_harmony_utils.py
........................................................................        [100%]
72 passed, 2 warnings in 14.47s
```

### Berkeley Function Calling Leaderboard v4 Results

**Baseline, from vLLM main as of Nov 12, 2025**

commit 304419576ae9dc2ecaa28c4506d3870f7c68bd85

45.0% overall accuracy

```
🔍 Running test: multi_turn_base
✅ Test completed: multi_turn_base. 🎯 Accuracy: 50.00%
🔍 Running test: multi_turn_long_context
✅ Test completed: multi_turn_long_context. 🎯 Accuracy: 43.00%
🔍 Running test: multi_turn_miss_func
✅ Test completed: multi_turn_miss_func. 🎯 Accuracy: 45.50%
🔍 Running test: multi_turn_miss_param
✅ Test completed: multi_turn_miss_param. 🎯 Accuracy: 41.50%

$ cat score-baseline/data_multi_turn.csv 
Rank,Model,Multi Turn Overall Acc,Base,Miss Func,Miss Param,Long Context
1,openai/gpt-oss-120b (FC) (vLLM),45.00%,50.00%,45.50%,41.50%,43.00%
```

**With this fix on top of same commit from main**

* 53.5% overall accuracy, an improvement of 8.5%
* 12% improvement in multi_turn_base
* 8.5% improvement in multi_turn_long_context
* 4.5% improvement in multi_turn_miss_func
* 8.0% improvement in multi_turn_miss_param

```
✅ Test completed: multi_turn_base. 🎯 Accuracy: 62.00%                                                                                                                                                                                                                                         
🔍 Running test: multi_turn_long_context                                                                                                                                                                                                                                                        
✅ Test completed: multi_turn_long_context. 🎯 Accuracy: 51.50%                                                                                                                                                                                                                                 
🔍 Running test: multi_turn_miss_func                                                                                                                                                                                                                                                           
✅ Test completed: multi_turn_miss_func. 🎯 Accuracy: 50.00%                                                                                                                                                                                                                                    
🔍 Running test: multi_turn_miss_param                                                                                                                                                                                                                                                          
✅ Test completed: multi_turn_miss_param. 🎯 Accuracy: 49.50%

$ cat score-harmony_fixes/data_multi_turn.csv 
Rank,Model,Multi Turn Overall Acc,Base,Miss Func,Miss Param,Long Context
1,openai/gpt-oss-120b (FC) (vLLM),53.25%,62.00%,50.00%,49.50%,51.50%
```

**Updated results after rebase on latest main as of Dec 10th**

```
✅ Test completed: multi_turn_base. 🎯 Accuracy: 62.00%
🔍 Running test: multi_turn_long_context
✅ Test completed: multi_turn_long_context. 🎯 Accuracy: 59.00%
🔍 Running test: multi_turn_miss_func
✅ Test completed: multi_turn_miss_func. 🎯 Accuracy: 52.50%
🔍 Running test: multi_turn_miss_param
✅ Test completed: multi_turn_miss_param. 🎯 Accuracy: 46.50%
```

```
$ cat score/data_multi_turn.csv 
Rank,Model,Multi Turn Overall Acc,Base,Miss Func,Miss Param,Long Context
1,openai/gpt-oss-120b (FC) (vLLM),55.00%,62.00%,52.50%,46.50%,59.00%
```


vllm/entrypoints/harmony_utils.py
```diff
@@ -259,7 +346,8 @@ def parse_input_to_harmony_message(chat_msg) -> list[Message]:
 
     # Tool role message (tool output)
     if role == "tool":
-        name = chat_msg.get("name", "")
+        tool_call_id = chat_msg.get("tool_call_id", "")
+        name = tool_id_names.get(tool_call_id, "")
         content = chat_msg.get("content", "") or ""
```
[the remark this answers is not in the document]
bbrowning: I have not tested this in the Responses API code path. In Chat Completions, there is no `name` field that is passed in by typical clients in the Chat Completions API here and everything should be based off `tool_call_id`. I'll look into how Responses does that, just to ensure it wasn't using the `name` key. In practice, `chat_msg.get("name", "")` was returning an empty string so unless Responses message use this `name` field we don't need to check that field.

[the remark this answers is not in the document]
bbrowning: @alecsolder Do you know, when `request.previous_input_messages` is a dict (for Responses API requests), how are tool call responses identified? Are they role of `tool`? And if so, do they use `tool_call_id` or `name` to map back to their corresponding tool call generated? Responses and Chat Completions sharing this code path makes this logic trickier to get right for both and I want to make sure I didn't mess up something for Responses.

vllm/entrypoints/harmony_utils.py
```diff
@@ -235,17 +235,104 @@ def parse_response_input(
...
+def parse_inputs_to_harmony_messages(chat_msgs: list) -> list[Message]:
+    msgs: list[Message] = []
+    tool_id_names: dict[str, str] = {}
+
+    # Collect tool id to name mappings for tool response recipient values
+    for chat_msg in chat_msgs:
+        # Convert from Pydantic ValidatorIterators to a real list
+        tool_calls = list(chat_msg.get("tool_calls") or [])
+        chat_msg["tool_calls"] = tool_calls
+        for tool_call in chat_msg.get("tool_calls"):
```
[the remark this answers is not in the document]
bbrowning: Hmm this suggestion makes some tests fail, because then the tool calls end up not being a list later on. I understand the intent, so need to dig more into why this causes some Pydantic ValidatorIterator things for lazy validation to leak into other places if I change as requested here. I really dislike Pydantic lazy validation...

ehfd:
Are we able to get this merged before v0.11.1?

vllm/entrypoints/harmony_utils.py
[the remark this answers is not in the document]
bbrowning: Ok this is running into https://github.com/pydantic/pydantic/issues/9541 which is poor behavior in Pydantic where the `tool_calls` here are an iterator and the first reader consumes them, which is why I had to re-assign the value here as otherwise any code that looks at `tool_calls` after this sees an empty list. There's already special code in vLLM to handle this for the [MistralTokenizer case](https://github.com/vllm-project/vllm/blob/89d3679221023fc18fd47df8fc426347fa9694e1/vllm/transformers_utils/tokenizers/mistral.py#L24), but we'll also need to handle this for Harmony cases. That's why my code above does, but I agree with Gemini it's not the ideal place and we should probably just do this in a Pydantic `field_validator` on our `ChatCompletionRequest` object to always materialize these as opposed to special-casing MistralTokenizer or Harmony models, since this could bite any future code working with tool_calls directly before applying a chat template.

vllm/entrypoints/harmony_utils.py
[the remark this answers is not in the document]
bbrowning: So, there is a Responses test that passes in a `name` at https://github.com/vllm-project/vllm/blob/085a5253321a66d7aac0f990d82417ad85ec0eb0/tests/entrypoints/openai/test_response_api_with_harmony.py#L910-L916

To me, that code looks quite wrong, as that is not the format that tool call results should be passed back in when using Responses. But, I also don't necessarily want to go digging into that part of Responses as part of fixing Chat Completions. This means I may need to fork the logic here and handle turning Chat Completions requests into Harmony messages differently from how Responses does it.

bbrowning:
I'm going to make some minor changes to address the review feedback from the two code review bots and to isolate this change entirely from the Responses API path so that we can merge this quicker without worry of it impacting Responses handling in any way. I'll get that done, re-tested, and push the updates today.

vllm/entrypoints/harmony_utils.py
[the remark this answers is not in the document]
bbrowning: Resolved this by reusing the existing `maybe_serialize_tool_calls` that some other code paths already use to handle this pydantic validatoriterator issue.

bbrowning:
> Are we able to get this merged before v0.11.1? Looks very critical.

Note that even without this change, the gpt-oss support on latest merged main branch is substantially better than what v0.11.0 or v0.10.2 shipped with. I don't know the timing or cutoff for fixes in v0.11.1, but even if this does not make that release there will still be measurably better gpt-oss support in v0.11.1. With that said, I support getting this in if the window for additional fixes is still open for v0.11.1.

vllm/entrypoints/harmony_utils.py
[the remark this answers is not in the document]
bbrowning: I separated this change out from any Responses API path. And, the previous usage of `name` is not part of the spec for Chat Completions API nor something our other Chat Completions specs or examples expect to work, so there's no need to retain it vs mapping by the tool_call_id. Marking this as resolved.

bbrowning:
Ok, from my perspective I've addressed the review bot comments, entirely isolated these changes to only impact the Chat Completions path, expanded the existing test_harmony_utils.py test suite so that it tests both Chat Completions and Responses paths with a common set of tests while moving logic specific to each into its own test class, and re-run my BFCL evaluation on gpt-oss-120b with the latest changes to ensure the numbers are approximately equal as before (approximately because there is a bit of run-to-run variance in this test suite as even though the inputs are identical in every test run the outputs are not entirely deterministic).

I'll keep an eye out for any other review feedback or comments required to get this in.

vllm/entrypoints/harmony_utils.py
```diff
@@ -235,7 +235,177 @@ def parse_response_input(
...
+        msg = msgs[i]
+        if msg.author.role == "assistant" and msg.channel == "final":
+            last_assistant_final_index = i
+            break
+
+    cleaned_msgs: list[Message] = []
+    for i, msg in enumerate(msgs):
+        if (
+            i < last_assistant_final_index
+            and msg.author.role == "assistant"
```
alecsolder: I don't think you should restrict for the role of assistant, [they don't in harmony](https://github.com/openai/harmony/blob/ec7606df9e87e3d0a1fec9f50928c1e407f0c438/src/encoding.rs#L215). Primarily this is to save context and remove tool calls and tool call outputs after there is a message to the final channel after them, and requiring the role to be assistant would prevent that

vllm/entrypoints/openai/parser/harmony_utils.py
```diff
@@ -235,7 +235,177 @@ def parse_response_input(
...
+
+    if role == "assistant" and tool_calls:
+        content = chat_msg.get("content")
+        if isinstance(content, list):
+            content = "".join(
+                item.get("text", "")
+                for item in content
+                if isinstance(item, dict) and item.get("type") == "text"
+            )
+        if content:
```
alecsolder, line 324: what is this case for? If tool calls is populated, I can understand reasoning being populated, but not this (which would imply it would be more like a "final" message)

vllm/entrypoints/openai/parser/harmony_utils.py
```diff
@@ -235,7 +235,177 @@ def parse_response_input(
...
+            msgs.append(commentary_msg)
+
+        reasoning_content = chat_msg.get("reasoning") or chat_msg.get(
+            "reasoning_content"
+        )
+        if reasoning_content:
+            analysis_msg = Message.from_role_and_content(
+                Role.ASSISTANT, reasoning_content
+            )
+            analysis_msg = analysis_msg.with_channel("analysis")
```
alecsolder, line 336: This runs counter to how I _feel_ like it should work, but [I think is correct](https://github.com/openai/gpt-oss/blob/48db88d8e29f48493fe75f084a8c9bd900a2b92f/gpt_oss/responses_api/api_server.py#L1297). 

Trying to understand OpenAI's implementation, when reasoning is output to the `commentary` channel, that information is meant to signal that this reasoning should be shown to the user as-is (vs needing a summarization) i.e they want to provide the actual reasoning for the tool call so the user can approve it or not vs only a summarization. But then when it is provided as input again, it doesn't matter so they just always change it back to analysis so that it can be dropped.

vllm/entrypoints/harmony_utils.py
```diff
@@ -235,7 +235,177 @@ def parse_response_input(
...
+            msg = msg.with_content_type("json")
+            msgs.append(msg)
+        return msgs
+
+    # Tool role message (tool output)
+    if role == "tool":
+        tool_call_id = chat_msg.get("tool_call_id", "")
+        name = tool_id_names.get(tool_call_id, "")
+        content = chat_msg.get("content", "") or ""
+        if isinstance(content, list):
```
alecsolder: nit: this could be pulled out because it is repeated multiple times

vllm/entrypoints/harmony_utils.py
```diff
@@ -235,7 +235,177 @@ def parse_response_input(
...
+        # TODO: Support refusal.
+        contents = [TextContent(text=c.get("text", "")) for c in content]
+
+    # Only add assistant messages if they have content, as reasoning or tool calling
+    # assistant messages were already added above.
+    if role == "assistant" and contents and contents[0].text:
+        msg = Message.from_role_and_contents(role, contents)
+        # Send non-tool assistant messages to the final channel if they don't have a
+        # channel already.
+        if not msg.channel:
```
alecsolder: when would this if case not trigger? I think from_role_and_contents [always returns None channel](https://github.com/openai/harmony/blob/ec7606df9e87e3d0a1fec9f50928c1e407f0c438/src/chat.rs#L168)

vllm/entrypoints/harmony_utils.py
alecsolder: It's a little confusing but if we want to provide a reference for this, [we can use this](https://github.com/openai/gpt-oss/blob/48db88d8e29f48493fe75f084a8c9bd900a2b92f/gpt_oss/responses_api/api_server.py#L1267). It says the index of the "last assistant message", where message is the important word because the message type in responses API means final channel, so this is where our logic comes from.

vllm/entrypoints/openai/parser/harmony_utils.py
```diff
@@ -235,7 +235,177 @@ def parse_response_input(
...
+            msgs.append(analysis_msg)
+
+        for call in tool_calls:
+            func = call.get("function", {})
+            name = func.get("name", "")
+            arguments = func.get("arguments", "") or ""
+            msg = Message.from_role_and_content(Role.ASSISTANT, arguments)
+            msg = msg.with_channel("commentary")
+            msg = msg.with_recipient(f"functions.{name}")
+            msg = msg.with_content_type("json")
```
alecsolder, line 349: I think this could use a comment as it [differs from the OpenAI implementation](https://github.com/openai/gpt-oss/blob/48db88d8e29f48493fe75f084a8c9bd900a2b92f/gpt_oss/responses_api/api_server.py#L1301), however I think there has been enough proof at this point that it increases benchmark scores so should be kept

vllm/entrypoints/harmony_utils.py
```diff
@@ -235,7 +235,177 @@ def parse_response_input(
...
+    elif role != "assistant":
+        msg = Message.from_role_and_contents(role, contents)
+        msgs.append(msg)
+
+    return msgs
+
+
 def parse_input_to_harmony_message(chat_msg) -> list[Message]:
+    """
+    Parse a message from request.preview_input_messages in the Responsees API to
```
alecsolder: nit: previous?

vllm/entrypoints/openai/parser/harmony_utils.py
```diff
@@ -235,7 +235,177 @@ def parse_response_input(
...
+    # Only add assistant messages if they have content, as reasoning or tool calling
+    # assistant messages were already added above.
+    if role == "assistant" and contents and contents[0].text:
+        msg = Message.from_role_and_contents(role, contents)
+        # Send non-tool assistant messages to the final channel if they don't have a
+        # channel already.
+        if not msg.channel:
+            msg = msg.with_channel("final")
+        msgs.append(msg)
+    # For user/system/developer messages, add them directly even if no content.
```
alecsolder, line 393: I think system and developer cases here need more complex logic. 

I think from a high level:
- We should never have more than 1 system message
- We should never have more than 1 developer message

(Can this validation be added to a test?)

Since we create system message and developer message every time in _make_request_with_harmony I'm worried about what happens with system messages here.

Since there is no "instructions" field like in responses, we need to merge the developer message created with the included tools with the developer message here, but not duplicating the tools for example. Complicated and definitely worth the tests I think.

alecsolder (commented):
This is great! I think we're finally landing on an implementation that takes the learnings from all of us over the past few months :)

vllm/entrypoints/harmony_utils.py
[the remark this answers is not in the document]
alecsolder: > @alecsolder Do you know, when `request.previous_input_messages` is a dict (for Responses API requests), how are tool call responses identified? Are they role of `tool`? And if so, do they use `tool_call_id` or `name` to map back to their corresponding tool call generated? Responses and Chat Completions sharing this code path makes this logic trickier to get right for both and I want to make sure I didn't mess up something for Responses.

I think they would be identified by [recipient = assistant](https://cookbook.openai.com/articles/openai-harmony#handling-tool-calls)? And they don't need the ID mapping because the role name is the function name, the id mapping is only for completions as calls and outputs are separate without the output having a reference to the tool name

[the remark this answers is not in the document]
bbrowning: That's useful context. I split the Chat Completion path out from this because trying to handle two different shapes of input from the single source of logic makes it harder to reason about, at least for me.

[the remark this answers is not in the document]
alecsolder: agreed!

vllm/entrypoints/openai/parser/harmony_utils.py
bbrowning: I think you're right that this content closer matches the `final` content, but now you made me double-check how we generate the outputs and we have some issues in mapping of reasoning and tool call outputs when generating Chat Completion responses. In my head, the `content` here matched `commentary` channel responses from tool calls while the `reasoning` key matches `analysis` channel responses. However, that's not necessarily the case and it looks like there are some edge cases in our reasoning and tool call parsers for gpt-oss models that can sometimes end up with this, sometimes end up with analysis or commentary messages missed entirely, or most often end up with analyis messages as reasoning and final messages as the tool call content.

In practice, this still seems to work quite well as-is. But, we can likely pick up a bit more accuracy by doing a once-over of Chat Completions reasoning/tool parsers and chat_completion_*_generator paths for harmony models to ensure streaming/non-streaming cases are all aligned with exactly how we parse input. I probably need a new test suite here that starts with Harmony messages, goes through our response parsing to Chat Completion responses, sends those back in as Chat Completion requests, and ensures the Harmony messages generated from that match what we originally sent.

bbrowning: Part of what worries me here is on the response generation side, where it makes assumptions about which messages to map as reasoning vs content solely based on order of the output messages without checking of any channels: https://github.com/vllm-project/vllm/blob/63fed5550609b96b578d2512aefced09efe76e1e/vllm/entrypoints/harmony_utils.py#L522-L534

Then, we overwrite he `content` from that if using a tool call parser at https://github.com/vllm-project/vllm/blob/63fed5550609b96b578d2512aefced09efe76e1e/vllm/entrypoints/openai/serving_chat.py#L1340-L1351

The tool call parser at https://github.com/vllm-project/vllm/blob/63fed5550609b96b578d2512aefced09efe76e1e/vllm/entrypoints/openai/tool_parsers/openai_tool_parser.py only looks at harmony messages with function recipients or to the final channel.

So, we may end up mixing analysis and commentary channel messages in reasoning or even dropping one of those entirely when making the Chat Completion response. To get this request mapping to Harmony messages exactly right, I'll need to clean up and pair this with some adjustments to the Harmony message to response mapping path so I can do this deterministically.

That will take a bit of time to match up, add some round-trip tests to ensure it doesn't deviate in the future, and re-run the eval. I can tackle this next week, either as part of this or a folllow-up based on timing preferences.

vllm/entrypoints/harmony_utils.py
bbrowning: You're right, thanks! A previous iteration of this had `msg` differently so I just missed that I could remove this conditional in the latest version. Removed this conditional in a new commit.

vllm/entrypoints/harmony_utils.py
bbrowning: That was a poor assumption on my part, so I loosened this up to drop the analysis messages from any role (ie removed the role check) instead of just assistant. And I added tests for various permutations of this method, including this one.

vllm/entrypoints/openai/parser/harmony_utils.py
bbrowning: I added a comment to this effect, but also indicating it may need further evaluation.

vllm/entrypoints/openai/parser/harmony_utils.py
bbrowning: So, there were some bugs on the Chat Completion response generation side around what I outlined above. I vastly expanded the new tests in test_serving_chat.py to verify various round-trips of Chat Completion requests to Harmony messages and Harmony outputs to Chat Completion responses including multi-turn variants that then send those Chat Completion response messages back in to future Chat Completion requests.

vllm/entrypoints/openai/parser/harmony_utils.py
bbrowning: I have not addressed this yet. This behavior should be unchanged from how it was before my PR, and I'm tempted to defer addressing this specific system/developer concern as this one is already large.

austin-mccall (approved):

chaunceyjiang (approved):
Thanks~ @bbrowning

rhajou:
This is part of which vllm version? 0.13.0 ? is there anything not released yet in 0.13.0?

bbrowning:
These fixes are in v0.13.0, yes. If you're still seeing any issues with gpt-oss models in vLLM 0.13.0, please raise an issue so we can dig into things.

PR #28730, opened 2025-11-14
[Model][Qwen3VL] Use `mm_position` to compute mrope positions

lgeiger:
## Purpose

This is a followup to #28399 and updates `get_mrope_input_positions` for Qwen3VL to make use of `mm_feature.mm_position`.

/cc @DarkLight1337 

## Test Plan

I manually verified that the implementation generates the same output for both images and videos. And ran an end2end test:

```
VLLM_WORKER_MULTIPROC_METHOD=spawn lm_eval --model vllm-vlm --model_args "pretrained=Qwen/Qwen3-VL-30B-A3B-Instruct-FP8,max_model_len=10000" --tasks chartqa --batch_size auto --apply_chat_template
```

## Test Result

**Before:**

| Tasks   | Version | Filter | n-shot | Metric            |     |  Value |     | Stderr |
| ------- | ------: | ------ | -----: | ----------------- | --- | -----: | --- | -----: |
| chartqa |       0 | none   |      0 | anywhere_accuracy | ↑   | 0.8732 | ±   | 0.0067 |
|         |         | none   |      0 | exact_match       | ↑   | 0.6388 | ±   | 0.0096 |
|         |         | none   |      0 | relaxed_accuracy  | ↑   | 0.8644 | ±   | 0.0068 |

**After:**

| Tasks   | Version | Filter | n-shot | Metric            |     |  Value |     | Stderr |
| ------- | ------: | ------ | -----: | ----------------- | --- | -----: | --- | -----: |
| chartqa |       0 | none   |      0 | anywhere_accuracy | ↑   | 0.8732 | ±   | 0.0067 |
|         |         | none   |      0 | exact_match       | ↑   | 0.6388 | ±   | 0.0096 |
|         |         | none   |      0 | relaxed_accuracy  | ↑   | 0.8644 | ±   | 0.0068 |


vllm/model_executor/models/qwen3_vl.py
```diff
@@ -1410,72 +1410,47 @@ def _parse_and_validate_multimodal_inputs(self, **kwargs: object) -> dict:
                 )
         return mm_input_by_modality
 
+    def get_mm_grid_hw(
```
DarkLight1337: I would rename this to `iter_mm_grid_hw`

DarkLight1337 (commented):
Also, I wonder how efficient this is compared to the original, can you run some benchmarks?

DarkLight1337 (approved):

PR #28734, opened 2025-11-14
fix(compilation): include input hash for model compilation to avoid c…

roadr:
Models with the same shapes have a colliding cache key even if they have different input signature. This was encountered using vllm's llama.py MLP decoder layer. This model being present with bias and without bias collides when it comes to compilation cache key.

The proposed fix includes the generated graph's input signature in the cache key.

### DIFF
--- a/tests/compile/test_input_signature_cache.py
+++ b/tests/compile/test_input_signature_cache.py
@@ -1,0 +1,379 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""
+Test that the input signature hash prevents cache collisions between
+graphs with the same structure but different input signatures.
+
+This test specifically verifies the fix for cache collisions when models
+have structurally similar graphs but different input argument counts
+(e.g., with/without bias weights).
+"""
+
+import pytest
+import torch
+from torch import nn
+
+from vllm.compilation.backends import CompilerManager
+from vllm.compilation.compiler_interface import compute_input_signature_hash
+from vllm.compilation.decorators import support_torch_compile
+from vllm.config import (
+    CompilationConfig,
+    CompilationMode,
+    VllmConfig,
+    set_current_vllm_config,
+)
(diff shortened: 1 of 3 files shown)


tests/compile/test_input_signature_cache.py
```diff
@@ -0,0 +1,379 @@
...
+    """
+    hidden_size = 128
+    intermediate_size = 256
+    batch_size = 4
+
+    x = torch.randn(batch_size, hidden_size)
+
+    # Create two MLPs with different bias settings
+    mlp_no_bias = LlamaMLP(hidden_size, intermediate_size, bias=False)
+    mlp_with_bias = LlamaMLP(hidden_size, intermediate_size, bias=True)
```
zou3519, line 224: I mentioned this in the other thread, but if the problem is that one model has a bias and the other doesn't, the general approach is that "the existence of a bias" should be reflected in the cache key.

There's a second approach which is to use `set_model_tag` to specify the two models should be cached separately

roadr: I'll try to test the set_model_tag when I have time.

But in general I'd love for this to just work transparently. I might be missing some detail knowledge wise about what you are suggesting, doesn't including input shape in the hash cover the "bias existence" detail?

zou3519: If you want this to work transparently, use compilation_mode=STOCK_TORCH_COMPILE . The default vLLM compilation mode (VLLM_COMPILE) trades off the transparency for performance.

It's not clear to me which input shape we're including in the hash, especially when the number of tokens is dynamic.

roadr: Ah good point about the fact it's dynamic

hmellor:
Closing: the approach was ruled out in review (input shapes are dynamic under VLLM_COMPILE), and the supported alternatives, `set_model_tag` or `compilation_mode=STOCK_TORCH_COMPILE`, are both in main. Happy to reopen if a cache-key-side fix along those lines materialises.

PR #28739, opened 2025-11-14
[Bugfix] Fix ChunkedLocalAttention CUDA Graph setting

benchislett:
## Purpose

Bugfix for incorrect outputs on llama4 caused by the refactor in #28479: ChunkedLocalAttentionBuilder incorrectly inherits FlashInfer's `get_cudagraph_support` method, causing the `_cudagraph_support` to be ignored and CUDA graphs to always get used.

## Test Plan

Breaking command:

```
python examples/offline_inference/basic/generate.py --model=nvidia/Llama-4-Scout-17B-16E-Instruct-FP8 --kv-cache-dtype=fp8 --max-model-len=1024
```

## Test Result

Now works. Ongoing discussion to make ChunkedLocalAttention compatible with CUDA graphs, but for now it should stay as NEVER to avoid errors.

FIX #28604

### DIFF
--- a/vllm/attention/layers/chunked_local_attention.py
+++ b/vllm/attention/layers/chunked_local_attention.py
@@ -1,7 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 import functools
-from typing import ClassVar
 
 import torch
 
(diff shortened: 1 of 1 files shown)


mgoin (approved):
LGTM as an immediate fix, thank you

LucasWilkinson (approved):
LGTM; thanks for the quick fix!

PR #28142, opened 2025-11-05
[CI/Build] Install uv for AMD MI300: Language Models Tests (Hybrid) %N

amdfaa:
## Purpose
This PR is for fixing the AMD MI300: Language Models Tests (Hybrid) %N failing with uv: command not found
See failing build:
https://buildkite.com/vllm/ci/builds/37631/steps/canvas?sid=019a5264-347a-407b-8a34-6d54a2f8e82f

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #28741, opened 2025-11-14
[ROCm][CI/Build] Change install location of uv

gshtras:
A follow up for #28142 to install uv under /usr/local/bin

The rationale is that /root/.local/bin is inaccessible to non-root users, so when running a container with a custom user, there is a folder in the PATH that the user doesn't have access to.
This becomes critical in any subprocess call that is bound to fail, such as the one in /usr/local/lib/python3.12/dist-packages/torch/_dynamo/debug_utils.py that calls nvcc --version
On ROCm (and other non-NVidia platforms) it is expected to throw FileNotFound, but is currently throwing EACCESS, which has precidence over FileNotFound. So any vllm run from a non-root user would fail

# Testing plan
Run a docker with `--user $(id -u):$(id -g) -v /etc/passwd:/etc/passwd:ro -v /etc/group:/etc/group:ro --group-add video --group-add render`
Before this PR any vllm run fails with `torch._inductor.exc.InductorError: PermissionError: [Errno 13] Permission denied: 'nvcc'`
After this PR runs succeed


tjtanaa (approved):
LGTM

PR #28754, opened 2025-11-14
[Bugfix]Fix a conditional to not check zero value

gmagogsfm:
`if inp` unintentionally checks zero value in addition to none-ness. This would introduce an unnecessary and failing dynamic shape guard in some cases. Checking none-ness only is sufficient and correct.


ProExpertProg (commented):
LGTM< cc @zhxchen17 @zou3519

zhxchen17 (approved):
looks good.

zou3519 (approved):

mgoin (approved):

PR #28761, opened 2025-11-14
add support for --fully-sharded-loras in fused_moe

gnovack:
## Purpose
This PR adds support for `--fully-sharded-loras` to MoE LoRA adapters to allow S-LoRA style sharding of adapter weights. This reduces the amount of GPU memory required per rank when using LoRA w/ tensor parallelism.

### Example using Qwen3-30B

Before PR:
```
vllm serve Qwen/Qwen3-Coder-30B-A3B-Instruct -tp 4 --enable-lora --max-loras 16 --max-lora-rank 32 --fully-sharded-loras
```

```
...
[gpu_model_runner.py:3126] Model loading took 53.9761 GiB memory and 11.972274 seconds
...
```

After PR:
```
vllm serve Qwen/Qwen3-Coder-30B-A3B-Instruct -tp 4 --enable-lora --max-loras 16 --max-lora-rank 32 --fully-sharded-loras
```

```
...
[gpu_model_runner.py:3126] Model loading took 26.9761 GiB memory and 12.145540 seconds
...
```

## Test Plan

- Added `fully_sharded_loras` case to existing MoE LoRA TP test cases to validate that the model output remains the 
same when enabling this flag.
- Added new kernel-level test cases for `test_fused_moe_lora_kernel` to validate that kernel outputs matches before and after this change


vllm/lora/ops/triton_ops/fused_moe_lora_op.py
```diff
@@ -480,6 +486,16 @@ def _fused_moe_lora(
...
+                a_intermediate_cache1
+            )
+        else:
+            a_intermediate_cache1 = tensor_model_parallel_all_gather(
+                a_intermediate_cache1
+            )
+
     _fused_moe_lora_expand(
         output,
         a_intermediate_cache1,
```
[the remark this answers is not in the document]
gnovack: The bot has a good point. I'm surprised that this would not cause the test cases to fail. Let me take a close look here and see if we do need to modify the `max_lora_rank` after the shrink call

vllm/lora/ops/triton_ops/fused_moe_lora_op.py
```diff
@@ -480,6 +486,16 @@ def _fused_moe_lora(
         mul_routed_weight,
     )
 
+    if fully_sharded:
+        if max_lora_rank == w1_lora_b_stacked.shape[-1]:
+            a_intermediate_cache1 = tensor_model_parallel_all_reduce(
+                a_intermediate_cache1
+            )
+        else:
+            a_intermediate_cache1 = tensor_model_parallel_all_gather(
```
jeejeelee, line 495: QQ: Is PDL compatible with `fully_sharded`?

gnovack: I'll test this out today and let you know. my assumption is that PDL probably won't work here due to the collective comms between shrink and expand, but i will confirm this

vllm/lora/ops/triton_ops/fused_moe_lora_op.py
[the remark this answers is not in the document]
gnovack: It turns out that resetting `max_lora_rank` to the full rank is required here, even though the e2e model test cases were passing without it.

I added some additional kernel-level test cases to verify that the output is the same with and without `fully_sharded`, since the e2e model tests were not sensitive enough to detect this.

vllm/lora/ops/triton_ops/fused_moe_lora_op.py
gnovack: From what i can tell based on profiling results, it seems that PDL does not provide benefit when `fully_sharded` is enabled, as the communication op between shrink and expand is not overlapped with either one (see below):
Image 11-17-25 at 4 17 PM

However, even when `fully_sharded` is disabled, it does not look like the moe shrink/expand calls are getting overlapped (seems this is because of the `torch.zeros` call before the expand kernel):
Image 11-17-25 at 4 19 PM

Let me know if there is some other/better test i can run to see if PDL is working as expected before/after my changes.

jeejeelee: I've observed similar behavior as well. At least DPL didn't cause any undefined behavior. I'll take some more time to look into it later.

vllm/lora/ops/triton_ops/fused_moe_lora_op.py
```diff
@@ -380,7 +385,7 @@ def _fused_moe_lora_expand(
         **expand_config,
     )
     for i in range(num_slices):
-        output[:, :, i * N : (i + 1) * N] += b_intermediate_cache1[i]
+        output[:, :, i * N + offset : (i + 1) * N + offset] += b_intermediate_cache1[i]
```
jeejeelee, line 388: BTW, we can  fuse this `add` into the expand kernel, so we don't need to create `b_intermediate_cache1` explicitly.
This should save the overhead of 2 kernels (empty and add). We can complete this in a follow-up PR.

jeejeelee (approved):

vllm/lora/ops/triton_ops/fused_moe_lora_op.py
dcmaddix: Thanks @jeejeelee !

PR #28778, opened 2025-11-15
[NIXL][XPU] update install script of NIXL

zhenwei-intel:
update install script of NIXL
pin to 0.7.0 in dockerfile

### DIFF
--- a/docker/Dockerfile.xpu
+++ b/docker/Dockerfile.xpu
@@ -14,6 +14,7 @@ RUN apt clean && apt-get update -y && \
     libxext6 \
     libgl1 \
     lsb-release \
+    libaio-dev \
     numactl \
     wget \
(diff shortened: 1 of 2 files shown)


docker/Dockerfile.xpu
```diff
@@ -14,6 +14,7 @@ RUN apt clean && apt-get update -y && \
     libxext6 \
     libgl1 \
     lsb-release \
+    libaio-dev \
```
jikunshang, line 17: what's this for? if nixl have libaio dependency, I am not sure whether this is a good manner since version may matter.

zhenwei-intel: which is needed by nixl v0.7.1

jikunshang (approved):

PR #28779, opened 2025-11-15
[Bugfix]: nccl connnector memory leak

weichengz0616:
## Purpose
This PR addresses two issues:

1. In the current implementation of `recv_tensor` in the nccl connector/engine, `self.buffer_size` is restored, but the corresponding tensor is released in `get_finished`, with the decoding step in between. This can potentially lead to GPU memory leaks. For example, during this period, the decode instance continues to receive KV cache. When `self.buffer_size` is used to check for available GPU memory, it may incorrectly assume that there is free memory, potentially leading to out-of-memory errors.

2. The scheduler currently allows None when generating `new_block_ids` in `CachedRequestData`, but the nccl connector does not support this. This can cause errors in chunked prefill scenarios.
image

---
<details>

</details>

### DIFF
--- a/vllm/distributed/kv_transfer/kv_connector/v1/p2p/p2p_nccl_connector.py
+++ b/vllm/distributed/kv_transfer/kv_connector/v1/p2p/p2p_nccl_connector.py
@@ -228,6 +228,10 @@ class P2pNcclConnector(KVConnectorBase_V1):
                     layer, kv_cache, request.block_ids, request.request_id
                 )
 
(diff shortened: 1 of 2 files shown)


heheda12345:
CC @NickLucche

heheda12345:
CC @NickLucche

NickLucche (commented):
Thanks a lot for contributing @weichengz0616 !

I feel like a unit test here would go a long way in ensuring this issue is pinned down.

PR #28786, opened 2025-11-15
[Perf] Optimize multi-token incremental detokenization

OthmanMohammad:
## Purpose

This PR optimizes `BaseIncrementalDetokenizer.update()` to batch-process multiple tokens instead of processing them one-by-one in a loop. This addresses a TODO from @woosuk regarding inefficiency when handling multiple tokens, which is increasingly common with speculative decoding (EAGLE, Medusa, n-gram proposer).

**Addresses:** TODO at `vllm/v1/engine/detokenizer.py:115-116`

## Problem

The current implementation processes tokens sequentially:
```python
for new_token_id in new_token_ids:
    self.token_ids.append(new_token_id)
    self.output_text += self.decode_next(new_token_id)
```

This becomes inefficient when speculative decoding generates multiple tokens per step (up to 128 with `MAX_SPEC_LEN`).

## Solution

- Introduced `_decode_tokens_batch()` method for batch processing
- Smart path selection:
  - **Fast path:** Batch-process when not crossing `min_tokens` threshold
  - **Slow path:** Process one-by-one only when crossing threshold (to maintain accurate `stop_check_offset` for stop string detection)
- Added optimized override in `SlowIncrementalDetokenizer`

## Changes

- `vllm/v1/engine/detokenizer.py`: Refactored `update()` method and added batch processing support

## Test Plan
```bash
pytest tests/tokenization/test_detokenize.py -v
```

## Test Result

Verified existing detokenization tests pass with no regressions. The optimization is backward compatible with no changes to public API.

---

**Summary:**
- Purpose: Optimize multi-token detokenization (addresses existing TODO)
- Test plan: Run existing tokenization tests
- Test results: No regressions, backward compatible
- Documentation: Not needed (internal optimization, no API changes)
- Release notes: Not needed (internal performance improvement)


njhill:
Thanks @OthmanMohammad. Could you share an e2e benchmark that shows the improvement?

We also have `--api-server-count N` which can be used to scale out the front end cpu processing.

njhill:
Thanks @OthmanMohammad for doing this analysis :pray:!

PR #28789, opened 2025-11-15
[Bugfix] Fix Llama3JsonToolParser to support deeply nested JSON parameters

ym820:
## Summary
Fixes a bug in `Llama3JsonToolParser` where the regex pattern could only match JSON objects with one level of nesting, causing tool calls with deeply nested parameters to fail.

## Problem
The old regex pattern could only match patterns like `{"a": {"b": "c"}}` but failed on `{"a": {"b": {"c": "d"}}}`.

Example failing input:
```json
{"name": "get_current_conditions", "parameters": {"location": {"city": "San Francisco", "state": "CA"}, "unit": "Fahrenheit"}}
```
The parser would extract only the inner parameters value and miss the tool name entirely.

## Solution

Replaced with a recursive regex pattern to allow unlimited nesting depth.

## Testing

- Added `test_extract_tool_calls_deeply_nested_json` for 2-level nesting
- Added `test_extract_tool_calls_very_deeply_nested_json` for 3+ levels
- All existing tests continue to pass

## Changes

- `vllm/entrypoints/openai/tool_parsers/llama_tool_parser.py`: Updated regex pattern
- `tests/entrypoints/openai/tool_parsers/test_llama3_json_tool_parser.py`: Added regression tests

### DIFF
--- a/vllm/entrypoints/openai/tool_parsers/llama_tool_parser.py
+++ b/vllm/entrypoints/openai/tool_parsers/llama_tool_parser.py
@@ -57,9 +57,10 @@ class Llama3JsonToolParser(ToolParser):
             0
         ]
         # Updated regex to match multiple JSONs separated by semicolons
-        # This pattern is more robust and can handle nested JSON objects
+        # This pattern uses recursion to handle arbitrarily nested JSON objects
+        # (?R) is a recursive pattern that matches the entire pattern again
         self.tool_call_regex = re.compile(
-            r"{[^{}]*(?:{[^{}]*}[^{}]*)*}(?:\s*;\s*{[^{}]*(?:{[^{}]*}[^{}]*)*})*",
+            r"\{(?:[^{}]|(?R))*\}(?:\s*;\s*\{(?:[^{}]|(?R))*\})*",
             re.DOTALL,
         )
 
--- a/vllm/entrypoints/openai/tool_parsers/llama_tool_parser.py
+++ b/vllm/entrypoints/openai/tool_parsers/llama_tool_parser.py
@@ -5,7 +5,6 @@ import json
 from collections.abc import Sequence
 
 import partial_json_parser
-import regex as re
 from partial_json_parser.core.options import Allow
 from transformers import PreTrainedTokenizerBase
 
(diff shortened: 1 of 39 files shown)


vllm/entrypoints/openai/tool_parsers/llama_tool_parser.py
```diff
@@ -57,9 +57,10 @@ def __init__(self, tokenizer: PreTrainedTokenizerBase):
...
         ]
         # Updated regex to match multiple JSONs separated by semicolons
-        # This pattern is more robust and can handle nested JSON objects
+        # This pattern uses recursion to handle arbitrarily nested JSON objects
+        # (?R) is a recursive pattern that matches the entire pattern again
         self.tool_call_regex = re.compile(
-            r"{[^{}]*(?:{[^{}]*}[^{}]*)*}(?:\s*;\s*{[^{}]*(?:{[^{}]*}[^{}]*)*})*",
+            r"\{(?:[^{}]|(?R))*\}(?:\s*;\s*\{(?:[^{}]|(?R))*\})*",
             re.DOTALL,
         )
```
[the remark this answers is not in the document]
ym820: Thanks. You're absolutely right. I've replaced the regex approach with an iterative parser that tracks string boundaries and escape sequences. I also added a few tests, `test_extract_tool_calls_with_braces_in_strings`, `test_extract_tool_calls_with_code_snippets`, `test_extract_tool_calls_with_escaped_quotes`, to cover the concerns.

vllm/entrypoints/openai/tool_parsers/llama_tool_parser.py
```diff
@@ -121,6 +117,64 @@ def extract_tool_calls(
                 tools_called=False, tool_calls=[], content=model_output
             )
 
+    def _extract_json_objects(self, text: str) -> list[str]:
```
chaunceyjiang, line 120: Just a question: why was regex deprecated?

ym820: Previous regex approach couldn't distinguish between braces in strings vs structural braces. For example, `{"name": "search", "parameters": {"query": "find users with status {active}"}}` would fail. The new parser tracks string boundaries and only counts braces outside of quoted strings.

chaunceyjiang: Honestly, I'm worried that doing these checks character by character might introduce more issues.

PR #26468, opened 2025-10-09
[torch.compile] caching of config fields should be opt-out by default

ghost:
Referring: https://github.com/vllm-project/vllm/issues/23107

Implements opt-out system of hashing, suggested by https://github.com/vllm-project/vllm/issues/16501 

```
def compute_hash(self):
   factors = list(self.__dict__.values())
   factors.remove("enforce_eager")
   factors.remove("tokenizer_config")
   ...
```

PR #28795, opened 2025-11-16
[Misc] Add backup hash algorithm for FIPS constrained environments

geodavic:
There are many places where `hashlib.md5` is used to compute hashes in configs. This algorithm isn't Federal Information Processing Standards (FIPS) compliant, and so the vLLM engine will fail to start inside environments where `md5` is forcefully disabled. 

## Purpose

Add a backup hashing algorithm for environments where `hashlib.md5` is disabled. These changes will still use `md5` when it is available, ensuring no existing behavior is changed. But when it is not available, it will fall back to a FIPS compliant algorithm (sha256).

## Test Plan

I spun up a server locally and performed inference

## Test Result

---
<details>

</details>


russellb (approved):
thanks!

hmellor (dismissed):
This is missing the `hash_factors` method in `vllm/config/utils.py` making this PR incomplete (https://github.com/vllm-project/vllm/pull/26468 merged while this was in review)

geodavic:
@russellb can I get one more stamp? Should go through now (finally)

vllm/utils/hashing.py
```diff
@@ -61,3 +62,20 @@ def get_hash_fn_by_name(hash_fn_name: str) -> Callable[[Any], bytes]:
         return sha256_cbor
 
     raise ValueError(f"Unsupported hash function: {hash_fn_name}")
+
+
+def safe_hash(data: bytes, usedforsecurity: bool = False) -> HASH:
```
russellb:
```suggestion
def safe_hash(data: bytes, usedforsecurity: bool = True) -> HASH:
```

I think `True` is a safer default, and it matches Python's default.

PR #28803, opened 2025-11-16
Starscream Changes to support CPX NPS4

sgudapar:
## Purpose
To add support for Starscream when operating AMD Instinct lineup GPUs in CPX+NPS4 operating mode. The is expected to improve the performance of attention kernels on Instinct GPUs.

## Test Plan
include `--enable-starscream` argument when running vllm to launch the updated code. If `--enable-starscream` argument is not set, the default vllm code is launched. 

A sample command would be as follows:
`python3 /app/vllm/benchmarks/benchmark_throughput.py --model $model --num-prompts 2 --input-len $in_len --output-len $out_len --dtype float16 --tensor-parallel-size $tp --distributed-executor-backend mp --enforce-eager --enable-starscream`

## Test Result
The output should look exactly the same but with improved performance for attention.

## Additional Documentation
When the instinct node is set to CPX NPS4 operating mode, each node will now have 64 VGPUs. This would result in replication of KV caches in cases where the num_kv_heads are less than 64. What the code is attempting to do is split the KV caches across XCDs within a GPU instead of replicating it when operating it in CPX NPS4 mode. This requires changes in slot_mapping, KV cache population, and how attention operation is performed with this split. 

1. The changes to slot_mapping is done in `gpu_model_runner.py`
2. The changes to KV cache population by splitting the KV entries is done in `triton_attn.py`
3. The changes to attention implementation to support this change is done in `triton_unified_attention.py`
4. There are additional inter-XCD collectives needed (specifically 2 Allgather operations) for which support for custom calls are made in `communication_op.py` and `parallel_state.py`

---
<details>

</details>

### DIFF
--- a/vllm/config/parallel.py
+++ b/vllm/config/parallel.py
@@ -104,8 +104,6 @@ class ParallelConfig:
     --data-parallel-start-rank."""
     enable_starscream: bool = False
     """ Enable starscream for cpx partitioning. """
-    num_prompts: int = 0
-    """ batch size """
     enable_expert_parallel: bool = False
     """Use expert parallelism instead of tensor parallelism for MoE layers."""
     enable_eplb: bool = False
--- a/vllm/config/parallel.py
+++ b/vllm/config/parallel.py
@@ -102,6 +102,10 @@ class ParallelConfig:
     between local data parallel ranks, but an external LB balances
     between vLLM nodes/replicas. Set explicitly in conjunction with
     --data-parallel-start-rank."""
+    enable_starscream: bool = False
+    """ Enable starscream for cpx partitioning. """
+    num_prompts: int = 0
+    """ batch size """
     enable_expert_parallel: bool = False
     """Use expert parallelism instead of tensor parallelism for MoE layers."""
     enable_eplb: bool = False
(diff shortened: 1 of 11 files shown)


LucasWilkinson (changes requested):
This looks alot like decode context parallel which is already supported by vLLM (by certain backends). Can you please provide a more detailed description of what you are trying to achieve here? a design doc or better comments would help. That will help us assess if we can add support for this using existing systems like decode context parallel (DCP). FYI: We are unlikely to accept system specific modification to gpu_model_runner given this code is shared by a vast amount of HW and is already very complex.

vllm/config/parallel.py
```diff
@@ -102,6 +102,10 @@ class ParallelConfig:
     between local data parallel ranks, but an external LB balances
     between vLLM nodes/replicas. Set explicitly in conjunction with
     --data-parallel-start-rank."""
+    enable_starscream: bool = False
+    """ Enable starscream for cpx partitioning. """
+    num_prompts: int = 0
```
hmellor: This seems like a really strange field to add to the global `ParallelConfig`

vllm/config/parallel.py
```diff
@@ -102,6 +102,10 @@ class ParallelConfig:
     between local data parallel ranks, but an external LB balances
     between vLLM nodes/replicas. Set explicitly in conjunction with
     --data-parallel-start-rank."""
+    enable_starscream: bool = False
```
hmellor, line 105: This should probably be an environment variable, like the other AMD specific hardware features

sgudapar:
Adding additional optimizations for the additional collective parts.

sgudapar:
Adding a comment to leave it open.

PR #28812, opened 2025-11-16
v1: clamp max_model_len when KV cache exceeds GPU budget

m0nk111:
## Summary
- snapshot and restore `max_model_len` while probing memory usage so we stop mutating the config globally
- when CPU offload is enabled but the KV cache still exceeds GPU capacity, clamp `max_model_len` down to what fits to avoid crashes and warn the operator instead
- keep CPU offload reporting in the error message for cases where the clamp still cannot fit

Addresses the second item in #27934.

## Testing
- `pytest tests/v1/kv_offload/test_cpu_offloading.py`

### DIFF
--- a/vllm/v1/core/kv_cache_utils.py
+++ b/vllm/v1/core/kv_cache_utils.py
@@ -640,9 +640,7 @@ def estimate_max_model_len(
         vllm_config.model_config.max_model_len = model_len
         try:
             # Calculate memory needed for the given model length
-            memory_needed = max_memory_usage_bytes(
-                vllm_config, kv_cache_spec.values()
-            )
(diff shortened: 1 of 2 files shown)


vllm/v1/core/kv_cache_utils.py
```diff
@@ -657,6 +662,7 @@ def fits_in_memory(model_len: int) -> bool:
             left = mid + 1
         else:
             right = mid - 1
+    vllm_config.model_config.max_model_len = original_max_model_len
```
heheda12345, line 665: I don't think it's a good idea to reset max_model_len here. This function is called quite late and it's hard to ensure the old max_model_len wasn't used by other steps.

PR #28827, opened 2025-11-17
[Core] Support resetting all running requests' KV while calling `reset_prefix_cache`

zhuohan123:
## Purpose

Currently, the `reset_prefix_cache` call can actually fail: When there are still running requests and kv cache with ref count > 0 in the system, the call will refuse to reset the prefix cache. This PR adds an option to move running requests back to waiting queue and ensure the reset call can be successful.

The original behavior is actually pretty dangerous because a user might not immediately realize the reset has failed, since it just prints out a warning message but not crash the engine. I suggest we actually crash the engine when the reset fail.

## Test Plan

New test in `tests/v1/core/test_scheduler.py`.

---
<details>

</details>


vllm/entrypoints/openai/api_server.py
```diff
@@ -1064,17 +1064,15 @@ async def show_server_info(
...
         """
-        device = None
-        device_str = raw_request.query_params.get("device")
-        if device_str is not None:
-            device = Device[device_str.upper()]
-        logger.info("Resetting prefix cache with specific %s...", str(device))
-        await engine_client(raw_request).reset_prefix_cache(device)
+        logger.info("Resetting prefix cache...")
+        await engine_client(raw_request).reset_prefix_cache(reset_running_requests)
         return Response(status_code=200)
```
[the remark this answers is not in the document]
zhuohan123: We explicitly mentioned the following:
```
Note that we currently do not check if the prefix cache is successfully reset in the API server.
```

zhuohan123:
@codex review

vllm/v1/core/sched/scheduler.py
```diff
@@ -1306,8 +1320,33 @@ def get_num_unfinished_requests(self) -> int:
...
+            while self.running:
+                self._preempt_request(self.running.pop(), timestamp)
+            self.running = []
+
+            # Clear scheduled request ids cache
+            self.prev_step_scheduled_req_ids.clear()
+
+        reset_successful = self.kv_cache_manager.reset_prefix_cache()
+        if reset_running_requests:
+            assert reset_successful
```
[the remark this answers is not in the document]
heheda12345: @zhuohan123 I think this problem could happen.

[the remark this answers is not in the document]
zhuohan123: +1 Fixing this now

vllm/v1/core/sched/scheduler.py
```diff
@@ -1306,8 +1320,33 @@ def get_num_unfinished_requests(self) -> int:
...
+    def reset_prefix_cache(self, reset_running_requests: bool = False) -> bool:
+        """Reset the KV prefix cache.
+
+        If reset_running_requests is True, all the running requests will be
+        preempted and moved to the waiting queue.
+        Otherwise, this method will only reset the KV prefix cache when there
+        is no running requests taking KV cache.
+        """
+        # For logging.
+        timestamp = time.monotonic()
```
njhill: nit: can move this within the `if`

vllm/v1/core/sched/scheduler.py
```diff
@@ -1335,8 +1349,42 @@ def get_num_unfinished_requests(self) -> int:
...
+
+            # Clear scheduled request ids cache
+            self.prev_step_scheduled_req_ids.clear()
+
+        reset_successful = self.kv_cache_manager.reset_prefix_cache()
+        if reset_running_requests and not reset_successful:
+            raise RuntimeError(
+                "Failed to reset KV cache even when all the running requests are "
+                "preempted and moved to the waiting queue. This is likely due to "
+                "there are running requests waiting for remote KV transfer, which "
```
njhill: wording nit:
```suggestion
                "the presence of running requests waiting for remote KV transfer, which "
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -774,7 +774,8 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> None:
         # they will be scheduled again sometime in the future.
         scheduled_req_ids = scheduler_output.num_scheduled_tokens.keys()
         cached_req_ids = self.input_batch.req_id_to_index.keys()
-        unscheduled_req_ids = cached_req_ids - scheduled_req_ids
+        resumed_req_ids = scheduler_output.scheduled_cached_reqs.resumed_req_ids
+        unscheduled_req_ids = cached_req_ids - (scheduled_req_ids - resumed_req_ids)
```
njhill, line 784: This is probably worth an explanatory comment - that `cached_req_ids` and `resumed_req_ids` are usually disjoint, apart from the forced-preemption case. And in that case we include the resumed_req_ids in the unscheduled set so that they get cleared from the persistent batch before being re-scheduled in the normal resumed request path.

vllm/v1/core/sched/scheduler.py
```diff
@@ -1335,8 +1349,42 @@ def get_num_unfinished_requests(self) -> int:
...
+            # running queue in FIFO order.
+            while self.running:
+                request = self.running.pop()
+                self._preempt_request(request, timestamp)
+                # NOTE(zhuohan): For async scheduling, we need to discard the latest
+                # output token on the fly to avoid a redundant repetitive output token.
+                request.num_output_placeholders = 0
+                request.discard_latest_async_token = True
+
+            # Clear scheduled request ids cache
```
```suggestion
            # Clear scheduled request ids cache. Since we are forcing preemption
            # + resumption in the same step, we must act as if these requests were
            # not scheduled in the prior step. They will be flushed from the persistent
            # batch in the model runner.
```

vllm/v1/request.py
```diff
@@ -93,7 +93,12 @@ def __init__(
             if self.prompt_token_ids is not None
             else [0] * self.num_prompt_tokens
         )
-        self.num_output_placeholders = 0  # Used in async scheduling.
+
+        # Used in async scheduling.
+        self.num_output_placeholders = 0
+        # Used in forced preemption (reset_prefix_cache) with async scheduling.
+        self.discard_latest_async_token = False
```
njhill: in spec decoding case it may be more than one.
```suggestion
        self.discard_latest_async_tokens = False
```

tests/v1/core/test_reset_prefix_cache_e2e.py
```diff
@@ -0,0 +1,65 @@
...
+
+PROMPTS = [
+    "A robot may not injure a human being ",
+    "To be or not to be,",
+    "What is the meaning of life?",
+    "What does the fox say? " * 20,  # Test long prompt
+]
+
+
+def test_reset_prefix_cache_e2e():
```
njhill, line 14: I wonder if we should also include a test with spec decoding?

njhill (commented):
Thanks @zhuohan123, it looks good to me, mainly some suggestions for additional clarifying comments.

It does still feel slightly hacky but isn't too invasive and will likely need a bit of rework for model runner v2 anyhow.

vllm/v1/worker/gpu_model_runner.py
zhuohan123: Oh yeah! Great catch.

tests/v1/core/test_reset_prefix_cache_e2e.py
zhuohan123: Will add it in a follow up PR!

njhill (approved):
Thanks @zhuohan123!

PR #28829, opened 2025-11-17
[XPU][QWEN] support qwen3-30b-a3b gptq int4 and tp=1,2

mayuyuace:
1. Add XPUGPTQMarlinMoEMethod
2. Add moe weights padding to qwen_moe.py. Because moe_intermediate_size of qwen3-30b-a3b is 768 and group_size is 128 ===> moe_intermediate_size // group_size = 6, so qwen3-30b-a3b gptq int4 cannot run with tp=4 or 8. Padding is needed.

Lm_eval CMD:
```
python3 -m vllm.entrypoints.openai.api_server --model Qwen/Qwen3-30B-A3B-GPTQ-Int4 --dtype=float16  --enforce-eager --port 8123 --host 0.0.0.0 --trust-remote-code --gpu-memory-util=0.9 --no-enable-prefix-caching --max-num-batched-tokens=4096  --disable-log-requests  --max-model-len=8192 --block-size 64 -tp=4 
lm_eval --model local-chat-completions --tasks gsm8k --num_fewshot 1 --batch_size 1 --model_args "model=Qwen/Qwen3-30B-A3B-GPTQ-Int4,base_url=http://localhost:8123/v1/chat/completions,max_gen_toks=4096,num_concurrent=32" --apply_chat_template  --output_path ./lm_eval_output --log_samples --limit 100
```
Lm_eval results on XPU as below:
image

### DIFF
--- a/vllm/model_executor/layers/quantization/ipex_quant.py
+++ b/vllm/model_executor/layers/quantization/ipex_quant.py
@@ -207,7 +207,6 @@ class IPEXConfig(QuantizationConfig):
                 return get_linear_quant_method(
                     self, layer, prefix, IPEXGPTQLinearMethod
                 )
-                # return IPEXGPTQLinearMethod(self)
         if isinstance(layer, FusedMoE) and self.method == "gptq":
             return XPUGPTQMarlinMoEMethod(self, layer.moe_config)
         return None
--- a/vllm/model_executor/layers/quantization/ipex_quant.py
+++ b/vllm/model_executor/layers/quantization/ipex_quant.py
@@ -2,7 +2,7 @@
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 
 from collections.abc import Callable
-from typing import Any, Optional
+from typing import Any
 
 import torch
 from packaging import version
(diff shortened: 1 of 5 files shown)


vllm/model_executor/layers/quantization/ipex_quant.py
```diff
@@ -155,9 +204,42 @@ def get_quant_method(
                     return UnquantizedLinearMethod()
                 return IPEXAWQLinearMethod(self)
             if self.method == "gptq":
-                return IPEXGPTQLinearMethod(self)
+                return get_linear_quant_method(
+                    self, layer, prefix, IPEXGPTQLinearMethod
+                )
+                # return IPEXGPTQLinearMethod(self)
```
jikunshang: please remove this

mayuyuace: Done.

jikunshang:
tp=4,8 will also throw error on CUDA device right?

mayuyuace:
> tp=4,8 will also throw error on CUDA device right?

I think yes, because xpu & cuda use the same path.
The problem that 6 is not divisible by 4 is unrelated to the device.

vllm/model_executor/models/qwen3_moe.py
```diff
@@ -510,7 +511,76 @@ def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
...
+                per_rank_moe_intermediate_size = (
+                    per_rank_moe_intermediate_size_group * group_size
+                )
+                moe_intermediate_size_padded = per_rank_moe_intermediate_size * tp_size
+                grouped_moe_intermediate_size_padded = (
+                    moe_intermediate_size_padded // group_size
+                )
+
         for name, loaded_weight in weights:
+            if quantization_config is not None and is_gptq_or_awq:
```
jikunshang: @dsikka can you take a review for padding part?

PR #28832, opened 2025-11-17
[tiny] Remove unsupported TRITON_MLA backend from batch invariance

bwasti:
TRITON_MLA is not well tested enough and not actually supported (but was incorrectly listed in the batch invariance supported backends and test parametrizations).

TRITON_MLA has two codepaths for prefill and decode that have not been unified. only the decode path shows batch-invariance, but the property that generated tokens have bitwise identical logprobs to prefilled does not hold.

This removes it from:
- override_envs_for_invariance() supported_backends list
- test_v1_generation_is_deterministic_across_batch_sizes_with_needle
- test_logprobs_bitwise_batch_invariance_bs1_vs_bsN
- test_simple_generation
- test_logprobs_without_batch_invariance_should_fail

## Test Plan

all listed tests

## Test Result

pass

### DIFF
--- a/vllm/model_executor/layers/batch_invariant.py
+++ b/vllm/model_executor/layers/batch_invariant.py
@@ -805,10 +805,11 @@ def override_envs_for_invariance():
         "FLASH_ATTN",  # best supported backend
         "FLASHINFER",
         "FLASH_ATTN_MLA",
-        "FLASHINFER_MLA",
         # Not yet supported MLA backends
         # "FLASHMLA",
         # "FLEX_ATTENTION", # IMA issue even if we disable batch invariance
+        # "FLASHINFER_MLA", https://github.com/vllm-project/vllm/pull/28967
+        # "TRITON_MLA",
     ]
     if curr_attn_backend not in supported_backends:
         warning = (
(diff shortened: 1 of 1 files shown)


ZJY0516:
cc @yewentao256

yewentao256 (approved):
Thanks for the work! Please also add more context in the PR description, eg. why it is not well supported, in what context it will fail for future reference.

bwasti:
added details!

vllm/model_executor/layers/batch_invariant.py
```diff
@@ -877,7 +877,6 @@ def override_envs_for_invariance():
         "FLASHINFER",
         "FLASH_ATTN_MLA",
         "FLASHINFER_MLA",
-        "TRITON_MLA",
         # Not yet supported MLA backends
         # "FLASHMLA",
         # "FLEX_ATTENTION", # IMA issue even if we disable batch invariance
```
yewentao256, line 810:
```suggestion
        # "FLEX_ATTENTION", # IMA issue even if we disable batch invariance
        # "TRITON_MLA",
```

yewentao256 (commented):
> added details!

@bwasti Did you save your update? Seems nothing change in the description.
Or we can add it in the code as well.

vllm/model_executor/layers/batch_invariant.py
```diff
@@ -805,11 +805,11 @@ def override_envs_for_invariance():
         "FLASH_ATTN",  # best supported backend
         "FLASHINFER",
         "FLASH_ATTN_MLA",
-        "TRITON_MLA",
+        "FLASHINFER_MLA",
```
yewentao256: Flashinfer is removed at https://github.com/vllm-project/vllm/pull/28967

Issue #24190, opened 2025-09-03
[Feature]: Model FLOPs Utilization Reporting


bwasti:
### 🚀 The feature, motivation and pitch

I'm working on inference optimization and would like Model FLOPs Utilization (MFU) to be more readily reported by vLLM.  It knows the model and it knows the hardware -- that's all you need!

I'd like to open this issue for discussion on the best way to report MFU.  Some open questions include:
- where does this get dumped? periodically printed like the speculative decoding metrics?
- should it be always on or require something like "/start_profile"?
- should care be taken to make FLOP counts exact (and unreported if the model isn't supported) or can simple approximations be used? (e.g. 2x # params)
- for hardware peak should we compile + check in vendor metrics? should we use simple approximations (e.g. number of tensorcores * clock)?

In general I would like to volunteer to help build out a calculator.

### Alternatives

manual calculation

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

zhuohan123:
I think from vLLM's perspective, let's get the numerator of the MFU correct. It should be easy for us to compute the actual flop used by each inference step, then we can report the flops by dividing the actual flop by the step time. We only need to report the flops and any user can compute the MFU by finding the proper denominator through their specific hardware spec.

bwasti:
> let's get the numerator of the MFU correct

this can be quite hard to do precisely.  some questions about that:
- do you think if we don't have new modules supported should we just bail and not report MFU?
- how should we handle speculative decoding, which "wastes" compute to some degree (but certainly improves utilization)?

I like the splitting up the denominator -- which allows us to ignore things like quantization (although this might be something the user wants in the future, since only parts of the model will run fp8 for example).

zhuohan123:
> where does this get dumped? periodically printed like the speculative decoding metrics?

It should be dumped together with where we dump our throughput numbers?

> should it be always on or require something like "/start_profile"?

I think the cost of this is very low and it should be always on.

> should care be taken to make FLOP counts exact (and unreported if the model isn't supported) or can simple approximations be used? (e.g. 2x # params)

If getting the exact count is too much work then we can start from params approach and we can report both eventually. But even just counting params we still need to have special handling for MoE layers to make sure we only count active params.

[...]

PR #28859, opened 2025-11-17
[Metrics] Model FLOPs Utilization estimation

SungMinCho:
Implements #24190

Summary:
Signed-off-by: Sung Min Cho sungmincho@meta.com

Use D82721353 to record average TF/s/GPU and GB/s/GPU periodically along with other existing stats like token throughput.

---

### Example outputs

common setup:
- model: gpt-oss
- batch size: 64
- input length: 6100
- output length: 1000

- TP8
  - full server log P2043309463
  - logger grep P2043310377
- TP4
  - full server log P2043310682
  - logger grep P2043310909
- TP4PP2
  - full server log P2043311278
  - logger grep P2043311447
- DP2TP4 - TP8
  - full server log P2043311757
  - logger grep P2043312147
- DP2TP4 - EP8
  - full server log P2043312361
  - logger grep P2043312522

Test Plan: CI

Differential Revision: D82742756

### DIFF
--- a/vllm/v1/core/sched/scheduler.py
+++ b/vllm/v1/core/sched/scheduler.py
@@ -42,7 +42,9 @@ from vllm.v1.core.sched.request_queue import SchedulingPolicy, create_request_qu
 from vllm.v1.core.sched.utils import check_stop, remove_all
 from vllm.v1.engine import EngineCoreEventType, EngineCoreOutput, EngineCoreOutputs
 from vllm.v1.kv_cache_interface import KVCacheConfig
+from vllm.v1.metrics.perf import ModelMetrics
 from vllm.v1.metrics.stats import (
+    PerfStats,
     PrefixCacheStats,
     SchedulerStats,
 )
@@ -186,6 +188,11 @@ class Scheduler(SchedulerInterface):
             if speculative_config.use_eagle():
                 self.use_eagle = True
                 self.num_lookahead_tokens = self.num_spec_tokens
+
+        # Used to derive model's MFU stats for each SchedulerOutput
+        self.perf_metrics: ModelMetrics | None = None
+        if self.log_stats and envs.VLLM_ENABLE_MFU_STATS:
+            self.perf_metrics = ModelMetrics(vllm_config)
 
         # Create the KV cache manager.
         self.kv_cache_manager = KVCacheManager(
@@ -1039,6 +1046,10 @@ class Scheduler(SchedulerInterface):
         kv_connector_output = model_runner_output.kv_connector_output
         cudagraph_stats = model_runner_output.cudagraph_stats
 
+        perf_stats: PerfStats | None = None
+        if self.perf_metrics and self.perf_metrics.is_enabled():
+            perf_stats = self.perf_metrics.get_step_perf_stats_per_gpu(scheduler_output)
+
         outputs: dict[int, list[EngineCoreOutput]] = defaultdict(list)
         spec_decoding_stats: SpecDecodingStats | None = None
         kv_connector_stats: KVConnectorStats | None = (
@@ -1221,7 +1232,7 @@ class Scheduler(SchedulerInterface):
 
         if (
             stats := self.make_stats(
-                spec_decoding_stats, kv_connector_stats, cudagraph_stats
+                spec_decoding_stats, kv_connector_stats, cudagraph_stats, perf_stats
             )
         ) is not None:
             # Return stats to only one of the front-ends.
@@ -1444,6 +1455,7 @@ class Scheduler(SchedulerInterface):
         spec_decoding_stats: SpecDecodingStats | None = None,
         kv_connector_stats: KVConnectorStats | None = None,
         cudagraph_stats: CUDAGraphStat | None = None,
+        perf_stats: PerfStats | None = None,
     ) -> SchedulerStats | None:
         if not self.log_stats:
             return None
@@ -1469,6 +1481,7 @@ class Scheduler(SchedulerInterface):
             spec_decoding_stats=spec_stats,
             kv_connector_stats=connector_stats_payload,
             cudagraph_stats=cudagraph_stats,
+            perf_stats=perf_stats,
         )
 
     def make_spec_decoding_stats(
--- a/vllm/v1/metrics/loggers.py
+++ b/vllm/v1/metrics/loggers.py
@@ -19,6 +19,7 @@ from vllm.distributed.kv_transfer.kv_connector.v1.metrics import (
 from vllm.logger import init_logger
 from vllm.plugins import STAT_LOGGER_PLUGINS_GROUP, load_plugins_by_group
 from vllm.v1.engine import FinishReason
+from vllm.v1.metrics.perf import PerfMetricsLogging
 from vllm.v1.metrics.prometheus import unregister_vllm_metrics
 from vllm.v1.metrics.stats import (
     CachingMetrics,
@@ -113,6 +114,10 @@ class LoggingStatLogger(StatLoggerBase):
                 self.vllm_config.compilation_config.cudagraph_mode,
                 self.vllm_config.compilation_config.cudagraph_capture_sizes,
             )
+
+        if self._enable_perf_stats():
+            self.perf_metrics_logging = PerfMetricsLogging(vllm_config)
+
         self.last_prompt_throughput: float = 0.0
         self.last_generation_throughput: float = 0.0
         self.engine_is_idle = False
@@ -133,6 +138,9 @@ class LoggingStatLogger(StatLoggerBase):
         self.num_generation_tokens += iteration_stats.num_generation_tokens
         self.num_corrupted_reqs += iteration_stats.num_corrupted_reqs
         self.num_preemptions += iteration_stats.num_preempted_reqs
+
+    def _enable_perf_stats(self) -> bool:
+        return True
 
     def _get_throughput(self, tracked_stats: int, now: float) -> float:
         # Compute summary metrics for tracked stats
@@ -175,6 +183,8 @@ class LoggingStatLogger(StatLoggerBase):
                 self.cudagraph_logging.observe(scheduler_stats.cudagraph_stats)
             if not self.aggregated:
                 self.last_scheduler_stats = scheduler_stats
+            if (perf_stats := scheduler_stats.perf_stats) and self._enable_perf_stats():
+                self.perf_metrics_logging.observe(perf_stats)
         if mm_cache_stats:
             self.mm_caching_metrics.observe(mm_cache_stats)
 
@@ -241,6 +251,10 @@ class LoggingStatLogger(StatLoggerBase):
         if not self.connector_prefix_caching_metrics.empty:
             log_parts.append("External prefix cache hit rate: %.1f%%")
             log_args.append(self.connector_prefix_caching_metrics.hit_rate * 100)
+
+        if self._enable_perf_stats():
+            self.perf_metrics_logging.log(log_parts, log_args)
+
         if not self.mm_caching_metrics.empty:
             log_parts.append("MM cache hit rate: %.1f%%")
             log_args.append(self.mm_caching_metrics.hit_rate * 100)
@@ -281,6 +295,10 @@ class AggregatedLoggingStatLogger(LoggingStatLogger, AggregateStatLoggerBase):
     @property
     def log_prefix(self):
         return "{} Engines Aggregated: ".format(len(self.engine_indexes))
+
+    def _enable_perf_stats(self) -> bool:
+        # Adding per_gpu perf stats across engines can lead to misleading numbers.
+        return False
 
     def record(
         self,
--- a/vllm/v1/metrics/perf.py
+++ b/vllm/v1/metrics/perf.py
@@ -6,6 +6,7 @@ Analytic flops/memory estimation module for transformer components,
 to help derive MFU (Model Flops Utilization) stats for a running model.
 """
 
+import time
 from abc import ABC, abstractmethod
 from collections.abc import Iterable
 from dataclasses import dataclass
@@ -22,6 +23,7 @@ from vllm.utils.torch_utils import (
     get_dtype_size,
     get_kv_cache_torch_dtype,
 )
+from vllm.v1.core.sched.output import SchedulerOutput
 
 logger = init_logger(__name__)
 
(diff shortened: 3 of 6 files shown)


vllm/v1/metrics/perf.py
```diff
@@ -0,0 +1,898 @@
...
+    def parse(self, args: ParsedArgs, vllm_config: VllmConfig) -> ParsedArgs:
+        cfg = vllm_config.model_config.hf_config
+        if hasattr(cfg, "text_config") and cfg.text_config is not None:
+            cfg = cfg.text_config
+
+        args.vocab_size = get_required(cfg, "vocab_size")
+        args.hidden_size = get_required(cfg, "hidden_size")
+        args.num_attention_heads = get_required(cfg, "num_attention_heads")
+        args.num_hidden_layers = get_required(cfg, "num_hidden_layers")
+
```
zhuohan123: In vLLM, we have some more general way to get these model parameters: https://github.com/vllm-project/vllm/blob/5f3cd7f7f20a8e4445d70cbd1f5475175ef391e3/vllm/config/model.py#L1230-L1353

In the future, we are moving towards a more unified config type to make this parsing easier: https://github.com/vllm-project/vllm/issues/24384

zhuohan123 (commented):
Thanks for the work! The integration is pretty standalone so it looks pretty nice to me. The biggest comment is that vLLM has some more general functions to get some of the parameters you want. We can make use of these functions to make the estimation more general.

vllm/v1/metrics/perf.py
SungMinCho: Thanks @zhuohan123 ! I updated the code so that it partially uses these pre-defined methods to parse the config! (But I couldn't replace "all" of my parsing because some still required manual parsing. e.g.) I still had to parse num_attention_heads manually because `model_config.get_attention_heads()` automatically divides by TP size which we don't want.)

vllm/v1/metrics/loggers.py
```diff
@@ -268,6 +311,18 @@ def __init__(
     def log_prefix(self):
         return "{} Engines Aggregated: ".format(len(self.engine_indexes))
 
+    def _enable_perf_stats(self) -> bool:
+        # Enable only when DP==1 for multiple reasons
+        # - Currently, LoggingStatLogger will not get instantiated and only
+        #   AggregatedLoggingStatLogger will be used in case of DP==1 and
+        #   if entered from FB-entrypoint.
```
zhuohan123: What's `FB-entrypoint`? Let's try to avoid internal-only logic here.

zhuohan123: Also, why do we have `_enable_perf_stats` for one logger but not for another logger? It is a bit confusing to me that logging method itself should matter to whether we can get perf stats or not.

vllm/v1/metrics/perf.py
```diff
@@ -0,0 +1,939 @@
...
+      -> ExecutionContext(2048, 2048, True)
+    - partial prefill with 1024 tokens 2048 context
+      -> ExecutionContext(1024, 2048, True)
+    - decode with 8192 context
+      -> ExecutionContext(1, 8192, False)
+    """
+
+    num_tokens: int
+    context_len: int
+    is_prefill: bool
```
zhuohan123: Why do we need to distinguish prefill and decode? From flops perspective they should be the same right?

vllm/v1/metrics/perf.py
```diff
@@ -0,0 +1,939 @@
...
+            logger.warning(
+                "Unknown model_dtype %s, defaulting to bfloat16",
+                model_dtype,
+            )
+            torch_dtype = torch.bfloat16
+
+        args.weight_byte_size = get_dtype_size(torch_dtype)
+
+        # FIXME: handle this better
+        args.activation_byte_size = 2
```
zhuohan123: Can you leave a more detailed TODO for the hardcode here and how exactly should we fix it?

vllm/v1/metrics/perf.py
```diff
@@ -0,0 +1,939 @@
...
+
+        quant_method = cfg.get_name()
+        if quant_method in ["fp8", "fbgemm_fp8"]:
+            # FIXME: This is a hacky coarse-grained fp8 quantization detection.
+            # FIXME: These configs also have concept of "ignored layers" and we
+            # need to solve the same problem as above.
+            args.weight_byte_size = 1
+        else:
+            # FIXME: Add more parsing logic for different quant methods.
+            pass
```
zhuohan123: Should we error out here?

vllm/v1/metrics/perf.py
```diff
@@ -0,0 +1,939 @@
...
+            # need to solve the same problem as above.
+            args.weight_byte_size = 1
+            pass
+        elif quant_method == "mxfp4":
+            # FIXME: Also has "ignored layers" issue above
+            args.weight_byte_size = 0.5
+        else:
+            # FIXME: Add more parsing logic for different quant methods.
+            pass
+
```
zhuohan123, line 607: ditto, should we error out here?

vllm/v1/metrics/perf.py
```diff
@@ -0,0 +1,939 @@
...
+class AttentionMetrics(ComponentMetrics):
+    num_hidden_layers: int = Field(..., gt=0)
+    hidden_size: int = Field(..., gt=0)
+    num_attention_heads: int = Field(..., gt=0)
+    num_key_value_heads: int = Field(..., gt=0)
+    head_dim: int = Field(..., gt=0)
+    weight_byte_size: int = Field(..., gt=0)
+    activation_byte_size: int = Field(..., gt=0)
+    cache_byte_size: int = Field(..., gt=0)
```
zhuohan123, line 354: Does this metric has assumption that all the layers are the same? what if say some layers have sliding window attention and some layers have full attention? Or some layers have MLA?

vllm/v1/metrics/perf.py
```diff
@@ -0,0 +1,939 @@
...
+        if quant_method in ["fp8", "fbgemm_fp8"]:
+            # FIXME: This is a hacky coarse-grained fp8 quantization detection.
+            # (there might be more quantization methods for fp8).
+            # FIXME: These configs also have concept of "ignored layers" and we
+            # need to solve the same problem as above.
+            args.weight_byte_size = 1
+            pass
+        elif quant_method == "mxfp4":
+            # FIXME: Also has "ignored layers" issue above
+            args.weight_byte_size = 0.5
```
zhuohan123, line 603: Do we also need to count the block scale's size?

vllm/v1/metrics/perf.py
```diff
@@ -0,0 +1,939 @@
...
+    def __init__(self, vllm_config: VllmConfig | None = None) -> None:
+        """
+        Parse vllm_config to instantiate metrics for each component.
+        Will raise ValueError if we cannot instantiate any components.
+        """
+
+        self.vllm_config = vllm_config
+
+        self.metrics: list[ComponentMetrics] = []
+        for metric_cls in ComponentMetrics.registered_metrics():
```
zhuohan123, line 944: Just to clarify, this for loop is for every type of layer, not for every layer instance right?

vllm/v1/metrics/perf.py
```diff
@@ -0,0 +1,939 @@
...
+    ) -> dict[str, int]:
+        total = {}
+        for metric in self.metrics:
+            breakdown = metric.get_write_bytes_breakdown(ctx, per_gpu)
+            component = metric.component_type()
+            prefixed = {
+                f"{component}.{key}": val for key, val in breakdown.model_dump().items()
+            }
+            total.update(prefixed)
+        return total
```
zhuohan123, line 998: nit: logic here seems a bit repetitive.

vllm/v1/core/sched/scheduler.py
```diff
@@ -188,6 +189,15 @@ def __init__(
         )
         self.use_pp = self.parallel_config.pipeline_parallel_size > 1
 
+        # Used to derive model's MFU stats for each SchedulerOutput
+        self.perf_metrics: ModelMetrics | None
+        try:
+            self.perf_metrics = ModelMetrics(vllm_config)
+            logger.info("Successfully initialized ModelMetrics")
+        except ValueError as e:
+            logger.info("Failed to initialize ModelMetrics due to %s", e)
```
markmc: We should distinguish here between `ValueError` that could occur for any number of reasons, an actual failure of some sort, and the case where none of our registered component perf metrics handlers can provide any insight into this particular model. Only this last case should be caught and handled gracefully here with an `INFO` log

vllm/v1/metrics/perf.py
```diff
@@ -0,0 +1,939 @@
...
+    @classmethod
+    def from_vllm_config(cls, vllm_config: VllmConfig) -> Self:
+        """
+        Instantiate this class from VllmConfig.
+        Raises ValueError if parsing fails.
+        """
+
+        parser = cls.get_parser()
+        parsed_args = parser.parse(vllm_config)
+        return cls.model_validate(parsed_args.model_dump())
```
markmc: I'm a bit confused where the `ValueError` is coming from - would it not be `ValidationError` here from pydantic?

In any case, I think here is where we'd want a "this component isn't relevant to this model" type thing that is distinguished from "parsing failed"

vllm/v1/metrics/perf.py
```diff
@@ -0,0 +1,939 @@
...
+        """
+
+        self.vllm_config = vllm_config
+
+        self.metrics: list[ComponentMetrics] = []
+        for metric_cls in ComponentMetrics.registered_metrics():
+            try:
+                metric = metric_cls.from_vllm_config(vllm_config)
+                self.metrics.append(metric)
+            except ValueError:
```
markmc: I'm wondering whether debug logging would be valuable here - e.g. if you were extending or adding a component, this would currently just swallow a useful error

vllm/v1/core/sched/scheduler.py
```diff
@@ -188,6 +189,15 @@ def __init__(
         )
         self.use_pp = self.parallel_config.pipeline_parallel_size > 1
 
+        # Used to derive model's MFU stats for each SchedulerOutput
+        self.perf_metrics: ModelMetrics | None
+        try:
+            self.perf_metrics = ModelMetrics(vllm_config)
```
markmc: We should only enable this if `log_stats` is `True`

vllm/v1/metrics/perf.py
```diff
@@ -0,0 +1,939 @@
+"""
+Analytic flops/memory estimation module for transformer components.
```
markmc: This would be a good place to expand the "MFU" acronym

vllm/v1/metrics/loggers.py
```diff
@@ -229,6 +262,16 @@ def log(self):
         if not self.connector_prefix_caching_metrics.empty:
             log_parts.append("External prefix cache hit rate: %.1f%%")
             log_args.append(self.connector_prefix_caching_metrics.hit_rate * 100)
+
+        if self.has_perf_stats and self._enable_perf_stats():
```
markmc: Can't we just make this conditional on `last_avg_tflops_per_gpu` and `last_avg_gbps_per_gpu` being non-zero?

vllm/v1/metrics/loggers.py
```diff
@@ -111,6 +112,12 @@ def __init__(self, vllm_config: VllmConfig, engine_index: int = 0):
         self.engine_is_idle = False
         self.aggregated = False
 
+        self.has_perf_stats = False
+        self.last_avg_tflops_per_gpu = 0.0
+        self.last_avg_gbps_per_gpu = 0.0
+        self.tp_size = vllm_config.parallel_config.tensor_parallel_size
+        self.pp_size = vllm_config.parallel_config.pipeline_parallel_size
```
markmc: `tp_size` and `pp_size` appear to be unused

vllm/v1/metrics/perf.py
```diff
@@ -0,0 +1,939 @@
...
+    @classmethod
+    def component_type(cls) -> str:
+        return "attn"
+
+    @classmethod
+    def get_parser(cls) -> ParserChain:
+        return ParserChain(
+            BaseConfigParser(),
+            BaseAttentionConfigParser(),
+            AttentionQuantizationConfigParser(),
```
markmc, line 371: The list of fields above is the union of the args from these three parsers, right? I'm not loving that we need to keep args/fields in sync across classes like this. Could `ComponentMetrics` not be a parser too?

markmc (changes requested):
Some comments on a first pass through the implementation

vllm/v1/metrics/loggers.py
SungMinCho: Removed tp_size which was not being used. Kept pp_size because it is used to divide duration.

vllm/v1/metrics/perf.py
SungMinCho: For example, the read_bytes() will differ depending on whether it is prefill or decode (i.e. decode will "read" context_len amount of KV cache, whose precision might differ from activation_byte_size, and so on). So I think it's good to keep them separate.

vllm/v1/metrics/perf.py
SungMinCho: Added a TODO comment below.

vllm/v1/metrics/perf.py
SungMinCho: I see what you're saying and my initial approach was actually what you suggested but I found many duplicate code involving basic config parsing (repeated for each component) which was uglier than this... so I switched to composable and stack-able chain of parsers. I think it helps with extensibility. Same reason why I keep a list of ComponentMetrics and not a monolithic ModelMetrics that's responsible for everything. I guess I'm implicitly nudging people to add parsers/metrics/etc that they are interested in to iteratively make the whole thing more accurate in a gradual manner.

vllm/v1/metrics/perf.py
SungMinCho: Ehh I guess scale tensor is very negligible compared to weights. Correct me if I'm wrong.

vllm/v1/metrics/perf.py
SungMinCho: Yes for every type.

vllm/v1/metrics/loggers.py
SungMinCho: Removed the internal related comment. JFYI the gist of the problem is that aggregated logger just naively "adds" every stat, which leads to misleading results for per-GPU MFU so we had to disable it. But then also, we have to still enable it for DP==1 because I discovered that when we run vLLM with DP==1 from fb predictor entrypoint, it does not instantiate LoggingStatLogger and instead instantiates AggregatedLoggingStatLogger.

vllm/v1/metrics/perf.py
SungMinCho: Done.

vllm/v1/metrics/perf.py
SungMinCho: Updated to throw error.

vllm/v1/metrics/perf.py
SungMinCho: Updated to throw error.

vllm/v1/core/sched/scheduler.py
SungMinCho: Updated to rely on `is_enabled()` method to figure out the latter case. All other types of errors will not be caught.

vllm/v1/metrics/perf.py
SungMinCho: Sorry you're right. Updated to ValidationError. I also defined custom `InvalidComponent` error and used it for all relevant cases.

vllm/v1/metrics/perf.py
SungMinCho: Done.

vllm/v1/core/sched/scheduler.py
SungMinCho: Done.

vllm/v1/metrics/perf.py
SungMinCho: Done.

vllm/v1/metrics/loggers.py
SungMinCho: Done, but keeping _enable_perf_stats() for reasons mentioned in https://github.com/vllm-project/vllm/pull/28859#discussion_r2557955671

vllm/v1/metrics/perf.py
```diff
@@ -0,0 +1,962 @@
...
+    def get_parser(cls) -> ParserChain: ...
+
+    def __init_subclass__(cls):
+        _COMPONENT_METRICS_REGISTRY[cls.component_type()] = cls
+
+    @classmethod
+    def from_vllm_config(cls, vllm_config: VllmConfig) -> Self:
+        """
+        Instantiate this class from VllmConfig.
+        Raises ValidationError if parsing fails.
```
markmc: Raises `InvalidComponentError`

markmc:
I've hacked on this a bit and sent you some PRs:

- SungMinCho/vllm#1
- SungMinCho/vllm#2
- SungMinCho/vllm#3

markmc:
>  (2) merge @markmc 's PRs onto this PR

Treat my PRs as changes I request through code review - I'm just making it easier to communicate the changes I'd like to see and easier for you to incorporate those changes

SungMinCho:
Hi @markmc I included the first two of your PRs as a part of this PR and pushed just now. (I ask that we make the third prometheus PR as a separate follow up PR from you because I'm not really familiar with prometheus and don't feel too comfortable including it in this PR without being able to thoroughly test it myself). Please take a look.

I also dug deeper into the performance degradation issue. I found some inefficiencies and removed them. For example, using pydantic just as syntactic sugar for Breakdown() turned out to be very slow so I replaced them with plain dictionaries. **But the perf degradation still persists**. Please compare the following logs:
- with MFU: https://gist.github.com/SungMinCho/fe6bb9e1a67a94e4e46b57a92fdb8bc3
- without MFU: https://gist.github.com/SungMinCho/1265fae60f4e78af21cb9d572efeb2a7

You can see that without MFU, avg generation throughput hovers around 6000 tokens/s whereas with MFU it is brought down to around 5700 tokens/s. If you check the end of the lines you will see that "mfu calc overhead" is around 5%, which means total 5% of the interval was spent on calculating the MFU stats... I guess pure arithmetic **can** accumulate to non-trivial overhead when there are many requests to calculate the stats for and when the actual GPU operation is very fast in contrast (e.g. decode)... (Still, it only makes half-sense to me so I'm continuing to debug).

I tried some optimizations like lazily calculating the MFU stats by only accumulating `num_tokens`, `context_len` (and their product) across all the observed requests, and then running these sums through our flops/bytes equations **only once** at the last moment to calculate the total MFU stats for that logging interval, and I did see that the performance degradation was mitigated by a lot (overhead decreases from 5% -> 0.x%?), but then later I realized that this leads to incorrect MFU stats (e.g. we can't naively accumulate num_tokens and derive the sum of number of activated experts because the relationship is nonlinear)...

I'll keep trying to optimize as 5% perf degradation will definitely not be approved. Just wanted to share the progress.

SungMinCho:
Moved to https://github.com/vllm-project/vllm/pull/30737

Closing this.

PR #28878, opened 2025-11-17
[Bugfix] Make compressed-tensors MoEs respect ignored layers

HDCharles:
## Purpose
Applying quantization to some MoE layers but not others would cause model load errors due to vllm assuming all the layers were quantized since there was no check of the ignore list.

## Changes
- added helper function get_scheme_dict used by get_scheme so only a single interface for MoE and Linear to match layers
- MoE matching previously would assume the 'Linear' target is used for MoE. instead, added helper which adds 'FusedMoE' to target_scheme_map and then match normally by either layer or Module type
- New test to check this
- Added conch-triton-kernels to be installed for the new test otherwise this tiny model has no kernels

## Test Plan
pytest tests/quantization/test_compressed_tensors.py::test_compressed_tensors_moe_ignore_with_model -vs -rs

## Test Result

```
PASSED

=============================== warnings summary ===============================
<frozen importlib._bootstrap>:241
  <frozen importlib._bootstrap>:241: DeprecationWarning: builtin type SwigPyPacked has no __module__ attribute

<frozen importlib._bootstrap>:241
  <frozen importlib._bootstrap>:241: DeprecationWarning: builtin type SwigPyObject has no __module__ attribute

======================== 1 passed, 2 warnings in 25.78s ========================
sys:1: DeprecationWarning: builtin type swigvarlink has no __module__ attribute
```

@kylesayrs @dsikka

### DIFF
--- a/.buildkite/test-pipeline.yaml
+++ b/.buildkite/test-pipeline.yaml
@@ -632,7 +632,7 @@ steps:
   # we can only upgrade after this is resolved
   # TODO(jerryzh168): resolve the above comment
   - uv pip install --system torchao==0.13.0 --index-url https://download.pytorch.org/whl/cu129
-  - uv pip install --system bitblas>=0.1.0
+  - uv pip install --system conch-triton-kernels
   - VLLM_TEST_FORCE_LOAD_FORMAT=auto pytest -v -s quantization/ --ignore quantization/test_blackwell_moe.py
 
 - label: LM Eval Small Models # 53min
--- a/.buildkite/test-pipeline.yaml
+++ b/.buildkite/test-pipeline.yaml
@@ -632,6 +632,7 @@ steps:
   # we can only upgrade after this is resolved
   # TODO(jerryzh168): resolve the above comment
   - uv pip install --system torchao==0.13.0 --index-url https://download.pytorch.org/whl/cu129
+  - uv pip install --system bitblas>=0.1.0
   - VLLM_TEST_FORCE_LOAD_FORMAT=auto pytest -v -s quantization/ --ignore quantization/test_blackwell_moe.py
 
 - label: LM Eval Small Models # 53min
--- a/tests/quantization/test_compressed_tensors.py
+++ b/tests/quantization/test_compressed_tensors.py
@@ -784,7 +784,7 @@ def test_compressed_tensors_moe_ignore_with_model(vllm_runner):
     - Config with ignore list containing specific MoE layers
     - Multiple MoE layers where some are quantized and some are not
     """
-    model_path = "nm-testing/tinysmokeqwen3moe-W4A16-first-only"
+    model_path = "nm-testing/tinysmokeqwen3moe-W4A16-first-only-CTstable"
 
     with vllm_runner(model_path, enforce_eager=True) as llm:
 
--- a/tests/quantization/test_compressed_tensors.py
+++ b/tests/quantization/test_compressed_tensors.py
@@ -770,6 +770,10 @@ def test_compressed_tensors_fp8_block_enabled(vllm_runner):
         assert output
 
 
+@pytest.mark.skipif(
+    not current_platform.is_cuda(),
+    reason="This test is not for non-CUDA platforms",
+)
 def test_compressed_tensors_moe_ignore_with_model(vllm_runner):
     """
     Integration test for MoE layer ignore functionality with a real model.
@@ -784,7 +788,9 @@ def test_compressed_tensors_moe_ignore_with_model(vllm_runner):
     - Config with ignore list containing specific MoE layers
     - Multiple MoE layers where some are quantized and some are not
     """
-    model_path = "nm-testing/tinysmokeqwen3moe-W4A16-first-only-CTstable"
+
+    # model_path = "nm-testing/tinysmokeqwen3moe-W4A16-first-only" # CT 12.3
+    model_path = "nm-testing/tinysmokeqwen3moe-W4A16-first-only-CTstable"  # CT 12.2
 
     with vllm_runner(model_path, enforce_eager=True) as llm:
 
--- a/tests/quantization/test_compressed_tensors.py
+++ b/tests/quantization/test_compressed_tensors.py
@@ -10,6 +10,7 @@ import torch
 from compressed_tensors.quantization import QuantizationType
 
 from tests.models.utils import check_logprobs_close
+from vllm.model_executor.layers.fused_moe import UnquantizedFusedMoEMethod
 from vllm.model_executor.layers.quantization.compressed_tensors.compressed_tensors import (  # noqa: E501
     CompressedTensors24,
     CompressedTensorsLinearMethod,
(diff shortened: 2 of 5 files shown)


vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors_moe.py
```diff
@@ -106,39 +104,35 @@ class CompressedTensorsMoEMethod(FusedMoEMethodBase):
...
+        # make sure quantization config for Linear can target it
+        quant_config._add_fused_moe_to_target_scheme_map()
+        unfused_names = [
+            prefix + proj_name
+            for proj_name in [".0.gate_proj", ".0.up_proj", ".0.down_proj"]
+        ]
+        # TODO: refactor this to use expert_mapping and check all layer numbers
+        all_scheme_dicts = [
+            quant_config.get_scheme_dict(layer, name) for name in unfused_names
+        ]
```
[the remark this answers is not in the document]
HDCharles: this logic is the same as before just explicated.

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
```diff
@@ -664,10 +672,35 @@ def get_scheme(
...
+        Detect whether a layer_name is found in any target and
+        use the quantization scheme corresponding to the matched target
+        to select the CompressedTensorsScheme used for inference.
+        """
+
+        # Use the new get_quant_args method to extract QuantizationArgs
+        scheme_dict = self.get_scheme_dict(layer, layer_name)
+
+        if scheme_dict is None:
+            return None
```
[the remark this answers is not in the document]
HDCharles: fixed

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors_moe.py
```diff
@@ -106,39 +104,35 @@ class CompressedTensorsMoEMethod(FusedMoEMethodBase):
...
-                    module=layer,
-                    targets=quant_config.target_scheme_map.keys(),
-                    fused_mapping=quant_config.packed_modules_mapping,
-                )
+        # FusedMoE was made by combining multiple Linears so need to
+        # make sure quantization config for Linear can target it
+        quant_config._add_fused_moe_to_target_scheme_map()
+        unfused_names = [
+            prefix + proj_name
+            for proj_name in [".0.gate_proj", ".0.up_proj", ".0.down_proj"]
```
HDCharles, line 121: this functionally is the same as previously where we matched based on hardcoded regex to the first layer. In practice the layers are normally matched based on Linear/FusedMoELayer so this won't matter except for models with ignore lists (previously entirely unsupported) or different expert_mapping or targetting using layer names with a different structure (same as before)

tests/quantization/test_compressed_tensors.py
```diff
@@ -767,3 +768,44 @@ def check_model(model):
...
+    MoE model where some MoE layers are in the ignore list, those layers
+    use UnquantizedFusedMoEMethod while non-ignored layers use the
+    quantized method.
+
+    Expected model structure:
+    - Compressed-tensors quantized MoE model (e.g., Mixtral-based)
+    - Config with ignore list containing specific MoE layers
+    - Multiple MoE layers where some are quantized and some are not
+    """
+    model_path = "nm-testing/Qwen3-30B-A3B-W4A16-first-10"
```
mgoin: This is a big model for CI and we don't actually need to load real weights here. Can you make this for a <1B model and use `load_format="dummy"` if the outputs aren't checked?

kylesayrs (approved):

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
```diff
@@ -158,9 +158,23 @@ def get_quant_method(
...
+        since linear layers get fused into FusedMoE
+        targetting 'Linear' needs to also match
+        FusedMoE modules.
+        """
+        if (
+            "Linear" not in self.target_scheme_map
+            or "FusedMoE" in self.target_scheme_map
+        ):
+            return
+        self.target_scheme_map["FusedMoE"] = self.target_scheme_map["Linear"]
```
mgoin, line 176: Shouldn't this be doing the filtering by looking for layer ignores that might disqualify FusedMoE?

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors_moe.py
mgoin: I don't understand the purpose of `.0` here? Are we making an assumption of just looking at the first moe layer to define the rules for the rest? I don't think this will work well since many MoE models skip the first couple of layers as dense. For instance deepseek v3 doesn't use MoE until the fourth layer, see `"first_k_dense_replace": 3,`

HDCharles: yeah that wouldn't be an issue since this is indexing experts, not layers.

e.g. for 
model.layers.0.mlp.experts.(0.down_proj).weight << we only try to match for this one
model.layers.0.mlp.experts.(1.down_proj).weight
...
model.layers.0.mlp.experts.(127.down_proj).weight
model.layers.0.mlp.experts.(0.gate_proj).weight
...
model.layers.0.mlp.experts.(0.up_proj).weight
...

we're assuming that all of the experts for that layer have the same quantization

whereas 

model.layers.[0].mlp.experts.0.down_proj.weight 
model.layers.[1].mlp.experts.0.down_proj.weight 
...
model.layers.[47].mlp.experts.0.down_proj.weight 

are all handled independently since the prefix will be model.layers.X.mlp.experts

its the same logic as before, just explicated. Before it'd try to see if there was a regex in the target list (which wouldn't handle manual layer listing) and only match on the first one. Here we're explicitly matching the first (0th) expert.

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
HDCharles: This isn't doing the filtering/matching, its just adding FusedMoE to the target_scheme_map as an option.

i.e.

before {"Linear": <scheme_dict_A>, "<other targets>": <scheme_dict_B>} -> 
after {"Linear": <scheme_dict_A>, "FusedMoE": <scheme_dict_A>,  "<other targets>": <scheme_dict_B>}

In the past what would happen is the MoE specific matching code would check to see if "Linear" is in targets and use that, and if not, it would try to match the regex. https://github.com/vllm-project/vllm/pull/28878/files#diff-ff1294eac1d5278b98c34982811d183967f327696bfdcc7f546d3b8c0d44f6ebL113-L127 . Note: This fails a bunch of corner cases like specifying layers manually...etc if Linear wasn't set as a target.

Instead what I want to do is to take that target_scheme_map and add FusedMoE as a target if Linear was in the targets (Since we know the original layers that became FusedMoE were linears). If we have FusedMoE as a target in the target_scheme_map,  we can use the default matching logic instead of the hacky MoE specific matching logic from before. This handles those corner cases more cleanly and is more maintainable since we only have some piece of matching logic.

tests/quantization/test_compressed_tensors.py
HDCharles: yeah i'll make a fake model for this, i don't think there's an actual MoE that's smaler than this one.

HDCharles: ok done, made a tiny model for this

kylesayrs (approved):

mgoin (approved):

.buildkite/test-pipeline.yaml
```diff
@@ -632,6 +632,7 @@ steps:
   # we can only upgrade after this is resolved
   # TODO(jerryzh168): resolve the above comment
   - uv pip install --system torchao==0.13.0 --index-url https://download.pytorch.org/whl/cu129
+  - uv pip install --system conch-triton-kernels
```
mgoin, line 635: Why is this needed now if we didn't need this before? Is it needed for the new model somehow?

mgoin (approved):

.buildkite/test-pipeline.yaml
HDCharles: the new test needs this or else there's no kernel for the test of the tiny model I made

Issue #28745, opened 2025-11-14
[Bug]: LongCat-Flash fails to start with TorchDynamo due to “Data-dependent assertion failed” in FusedMoE


baonudesifeizhai:
### Your current environment

<details>
 8*h200 
 
2.9.0+cu128

</details>

### 🐛 Describe the bug

vllm serve meituan-longcat/LongCat-Flash-Chat-FP8 \
  --trust-remote-code \
  --enable-expert-parallel \
  --tensor-parallel-size 8 \
  --max-model-len 32768 \
  --max-num-seqs 4

 goes to this error  https://paste.ubuntu.com/p/ZBjQdRx9v6/

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

PR #28891, opened 2025-11-18
[MoE Refactor][5/N] Isolate zero expert to LongCatFlash

baonudesifeizhai:
## Purpose
#28152
#28745
Fix dimension mismatch error in LongCat Flash MoE model when using zero experts. The issue occurred when `router_logits` (including zero experts) and `e_score_correction_bias` (only real experts) had different dimensions during expert selection.

## Test Plan
pytest tests/models/test_registry.py::test_registry_imports -k LongcatFlash 
vllm serve meituan-longcat/LongCat-Flash-Chat-FP8 \
  --trust-remote-code \
  --enable-expert-parallel \
  --tensor-parallel-size 8 \
  --max-model-len 32768 \
  --max-num-seqs 4

main branch : broken
this branch : 
 
2b01b00bad4f0e9abb4549dcc88cc51b

## Test Result

---
<details>

</details>


robertgshaw2-redhat:
@SageMoore @bnellnm @varun-sundar-rabindranath can you take a look?

vllm/model_executor/layers/fused_moe/fused_moe_modular_method.py
```diff
@@ -103,11 +103,7 @@ def apply(
         expert_load_view: torch.Tensor | None = None,
         logical_to_physical_map: torch.Tensor | None = None,
         logical_replica_count: torch.Tensor | None = None,
-    ) -> torch.Tensor | tuple[torch.Tensor, torch.Tensor]:
-        # Is getattr needed?
-        zero_expert_num = getattr(layer, "zero_expert_num", 0)
-        zero_expert_type = getattr(layer, "zero_expert_type", None)
-
+    ) -> torch.Tensor:
```
bnellnm: The return type can still be a tuple even without the zero experts.  This is due to the (optional) shared experts output.

vllm/model_executor/models/longcat_flash.py
```diff
@@ -292,31 +293,125 @@
...
+            routed_scaling_factor=self.experts.routed_scaling_factor,
+            e_score_correction_bias=self.router.e_score_correction_bias,
+            indices_type=None,
+            enable_eplb=False,
+            expert_map=None,
+            expert_load_view=None,
+            logical_to_physical_map=None,
+            logical_replica_count=None,
+            num_fused_shared_experts=self.experts.num_fused_shared_experts,
+        )
```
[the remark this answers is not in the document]
bnellnm: The redundant computation does seem like it could be a problem.

vllm/model_executor/layers/fused_moe/fused_moe_modular_method.py
baonudesifeizhai: changed back to tuple

vllm/model_executor/models/longcat_flash.py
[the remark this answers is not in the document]
baonudesifeizhai: for here i just move it to forward ...right now ...

any better solution?

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1384,6 +1376,18 @@ def select_experts(
...
+            # but bias doesn't
+            bias_dim = (
+                e_score_correction_bias.shape[-1]
+                if e_score_correction_bias.dim() > 0
+                else len(e_score_correction_bias)
+            )
+            router_logits_dim = router_logits.shape[-1]
+            if router_logits_dim > bias_dim:
+                router_logits = router_logits[..., :bias_dim]
+
```
bnellnm: Is it ok to just discard the extra router logits?  Should the bias be padded with 1s or something instead?  Or are the extra logits junk because they are for the zero expert?

vllm/model_executor/models/longcat_flash.py
```diff
@@ -292,31 +293,139 @@ def __init__(
...
+            routed_scaling_factor=self.experts.routed_scaling_factor,
+            e_score_correction_bias=self.experts.e_score_correction_bias,
+            indices_type=None,
+            enable_eplb=False,
+            expert_map=None,
+            expert_load_view=None,
+            logical_to_physical_map=None,
+            logical_replica_count=None,
+            num_fused_shared_experts=self.experts.num_fused_shared_experts,
+        )
```
bnellnm: I've got a suggestion for removing the extra routing computation.  I recall another PR that used a custom routing function whose only purpose was to "memoize" the results of `select_experts` and when used by the main `self.experts(...)` call, the memoized routing results would be used.

`vllm/model_executor/models/transformers/moe.py` is the example I was thinking of.

bnellnm (commented):
Do you know if there are plans to use the zero expert for other MoE models?  If so, it might be worth making a `ZeroExpertFusedMoE` class that encapsulates the zero expert logic (similar to the `SharedFusedMoE` class).

vllm/model_executor/layers/fused_moe/layer.py
baonudesifeizhai: i just remove this part ，use full bias  and logits works

vllm/model_executor/models/longcat_flash.py
baonudesifeizhai: changed to new file vllm/model_executor/layers/fused_moe/zero_expert_fused_moe.py

vllm/model_executor/layers/fused_moe/zero_expert_fused_moe.py
```diff
@@ -0,0 +1,246 @@
...
+        def custom_routing_function(hidden_states, gating_output, topk, renormalize):
+            """Return memoized `topk_weights` and `topk_ids`."""
+            if self._memoized_topk_weights is None or self._memoized_topk_ids is None:
+                raise RuntimeError(
+                    "ZeroExpertFusedMoE: routing results not memoized. "
+                    "Call select_experts first to compute routing."
+                )
+            return self._memoized_topk_weights, self._memoized_topk_ids
+
+        self.custom_routing_function = custom_routing_function
```
bnellnm, line 81: Can you add an assert that if `custom_routing_function` is not in `kwargs` or is `None` if it is in `kwargs`?

baonudesifeizhai: sure ！

vllm/model_executor/layers/fused_moe/zero_expert_fused_moe.py
```diff
@@ -0,0 +1,246 @@
...
+
+        # Compute real expert results (will reuse memoized routing via
+        # custom_routing_function)
+        fused_out = super().forward(
+            hidden_states=hidden_states,
+            router_logits=router_logits_sliced,
+        )
+
+        # Combine results
+        if zero_expert_result is not None:
```
baonudesifeizhai: this part maybe unuse.. .i will start a pod and have a check

vllm/model_executor/models/longcat_flash.py
```diff
@@ -280,48 +280,74 @@ def __init__(
...
             rounter_params_dtype=self.rounter_params_dtype,
             prefix=f"{prefix}.gate",
         )
 
-        self.experts = FusedMoE(
+        # Slice e_score_correction_bias to only include real experts (not zero_experts)
+        # ZeroExpertFusedMoE will handle zero experts internally
+        e_score_correction_bias_for_experts = self.router.e_score_correction_bias[
+            :num_experts
+        ]
```
bnellnm: nit: should/could the `ZeroExpertFusedMoE` handle slicing the correction bias?

bnellnm (approved):
Nice cleanup.  Thanks!

baonudesifeizhai:
image
 works after refactoring

vllm/model_executor/layers/fused_moe/zero_expert_fused_moe.py
```diff
@@ -0,0 +1,184 @@
...
+
+        self.custom_routing_function = custom_routing_function
+
+    @contextmanager
+    def _temporarily_set_attrs(self, **attrs):
+        """
+        Temporarily set attributes using object.__setattr__ and restore them.
+
+        This bypasses nn.Module.__setattr__ to avoid Dynamo tracing issues.
+        """
```
bnellnm, line 94: Can you elaborate on what the dynamo issues were?

baonudesifeizhai: added

bnellnm (approved):

robertgshaw2-redhat (approved):

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -73,7 +72,10 @@ def _eplb_map_to_physical_and_record(
         return topk_ids
 
     eplb_map_to_physical_and_record = _eplb_map_to_physical_and_record
-from vllm.model_executor.layers.fused_moe.fused_moe import grouped_topk
+from vllm.model_executor.layers.fused_moe.fused_moe import (
+    grouped_topk,
```
robertgshaw2-redhat: shouldnt these be removed?

robertgshaw2-redhat (changes requested):

vllm/model_executor/layers/fused_moe/layer.py
baonudesifeizhai: we use both of that in layer

baonudesifeizhai:
image
 works   for  vllm serve meituan-longcat/LongCat-Flash-Chat-FP8 \
  --trust-remote-code \
  --enable-expert-parallel \
  --tensor-parallel-size 8 \
  --max-model-len 32768 \
  --max-num-seqs 4

baonudesifeizhai:
image
 finally works normal and refactoring all zero_expert_num/zero_expert_type

baonudesifeizhai:
https://paste.ubuntu.com/p/MbssQ9KHBx/ meet this error..
god

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1633,23 +1630,7 @@ def select_experts(
...
-                expert_indices=topk_ids,
-                expert_scales=topk_weights,
-                num_experts=self.global_num_experts,
-                zero_expert_type=self.zero_expert_type,
-                hidden_states=hidden_states,
-            )
-        else:
-            zero_expert_result = None
-        return topk_weights, topk_ids, zero_expert_result
+        return topk_weights, topk_ids, None
```
zyongye: Can we get rid of `None` at this point and just keep two return value? Seems no other function need the third field?

baonudesifeizhai: okay  i will figure it out today

vllm/model_executor/layers/fused_moe/zero_expert_fused_moe.py
zyongye, line 1: I am also wondering if we need this file at all. Is it easy to do [something](https://github.com/sgl-project/sglang/blob/3f0482174aa45e649c7d39e27878c3ce80b6d5cb/python/sglang/srt/models/longcat_flash.py#L271-L290) like in sglang? So we don't need to create a separate moe class.

baonudesifeizhai: image
 really depends on how this zero expert moe  future looks like....

zyongye: Ok, we can keep it for extensibility

robertgshaw2-redhat (approved):
Thanks for the work on this.

Verified correctness of longcat. The implementation of ZeroExpertsFusedMoE is quite complicated, but we can revisit this once we get through the rest of the refactoring (once we break apart fused moe) to make it simpler

This effectively removes the zero expert from FusedMoE

Thanks!

PR #28893, opened 2025-11-18
Add gpu memory wait before test_async_tp

angelayi:
Attempting to fix [test failure](https://buildkite.com/vllm/ci/builds/39378/steps/canvas?sid=019a935d-582b-4209-afea-dc40df8de33e)


ProExpertProg:
Seems like this works, can you add it to E2E tests as well? Or maybe better add a test util file that just calls this function, so that we can insert it into the test pipeline and not pollute the tests?

ProExpertProg (approved):

ProExpertProg (approved):
I like the envvar approach! I think you should take into account the CUDA_VISIBLE_DEVICES variable, as I think that these tests should only be running on 2xH200 at a time, and the CI failure implies that the fixture is waiting on all 8 GPUs on the node.

ProExpertProg (approved):

ProExpertProg:
Okay this looks good now! We finally fixed the failure with e2e model tests. There's now a DBO failure which I assume snuck in from Lucas's refactor PR because this test failed before the DBO tests could run

khluu (approved):

PR #28895, opened 2025-11-18
[ROCm][CI] Fix tests/compile unit tests

charlifu:


tests/compile/test_noop_elimination.py
```diff
@@ -25,10 +25,13 @@ def test_noop_elimination(dtype, num_tokens, hidden_size, buffer_size):
...
             super().__init__()
-            self.pos_embed = torch.empty(buffer_size, hidden_size, dtype=dtype)
+            # Avoid using empty, since on rocm torch.empty
+            # does not initialize the memory.
+            self.pos_embed = torch.randn(buffer_size, hidden_size, dtype=dtype)
 
         def forward(self, x):
-            x += self.pos_embed[: x.shape[0]]
+            # Avoid += to prevent inplace addition.
+            x = x + self.pos_embed[: x.shape[0]]
```
SageMoore, line 34: Why is this change necessary?

SageMoore (commented):
I'm somewhat surprised that these graph partitioning tests don't work on ROCm. CC @ProExpertProg. PR looks reasonable, just one question.

ProExpertProg (changes requested):
Inductor partition will soon be the default, what about it doesn't work? If we can't get it to work, can you file an issue in upstream pytorch so it can be fixed? And add the issue number to all places where the tests are disabled

tests/compile/test_noop_elimination.py
charlifu: '''
x[0][0]: tensor(0.6719, device='cuda:0')
embed[0][0]: tensor(-0.2964, device='cuda:0')
result[0][0][0]: tensor(0.3755, device='cuda:0')
result2[0][0][0]: tensor(0.0791, device='cuda:0')
'''
result = x + embed
result2 = x + embed + embed

`+=` is doing inplace addition will cause all test cases fail. On NV side, because `torch.empty` always initializes memory to zero, so we didn't see the failure before.

charlifu: But on Rocm, we can not use `torch.empty()` since it does not initialize memory, so the value could be nan, causing test failure.

tests/compile/test_decorator.py
```diff
@@ -67,7 +68,10 @@ def run_model(
         return output.cpu()
 
 
-@pytest.mark.parametrize("use_inductor_graph_partition", [True, False])
+@pytest.mark.parametrize(
+    "use_inductor_graph_partition",
+    [True, False] if not current_platform.is_rocm() else [False],
+)
```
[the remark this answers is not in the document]
tjtanaa: @charlifu can you follow this suggestion?

SageMoore (approved):

tjtanaa (approved):
LGTM

tests/compile/test_decorator.py
ProExpertProg: Please skip tests instead of generating test cases dynamically
```suggestion
@pytest.mark.parametrize("use_inductor_graph_partition", [True, False])
def test_ignore_torch_compile_decorator(...)
    if not inductor_partition_supported()
        pytest.skip("Inductor partition not supported with torch==2.9.0a")
```

tests/compile/fullgraph/test_full_cudagraph.py
```diff
@@ -114,7 +122,7 @@ def llm_pair(request):
     [
         pytest.param((model, backend_config, use_inductor_graph_partition))
         for model, backend_config in model_backends_full_cudagraph
-        for use_inductor_graph_partition in [True, False]
+        for use_inductor_graph_partition in INDUCTOR_GRAPH_PARTITION_PARAMS
```
ProExpertProg: Skip instead of generating dynamically

tests/compile/fullgraph/test_full_cudagraph.py
```diff
@@ -70,6 +74,10 @@ def llm_pair(request):
         elif backend_config.specific_gpu_arch == (10, 0):
             pytest.skip("Only Blackwell GPUs support Cutlass MLA")
 
+    # FlashInfer is not supported on ROCm
+    if backend_config.name == "FlashInfer" and current_platform.is_rocm():
```
ProExpertProg:
```suggestion
    if backend_config == AttentionBackendEnum.FLASHINFER and current_platform.is_rocm():
```

ProExpertProg (changes requested):
Please address nits before merging

tests/compile/test_decorator.py
charlifu: torch updated to 2.9, no need for this.

ProExpertProg (approved):

PR #28904, opened 2025-11-18
[NIXL] fix cpu PD after physical <> logical block_size PR

xuechendi:
## Purpose

CPU KV layout is using HND as view shape => which leads to incorrect physical kernel block_size before => Fix in this PR

verified locally with GNR, accuracy looks good now.

```
PREFILL_BLOCK_SIZE=64 DECODE_BLOCK_SIZE=64 bash tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh --kv_buffer_device cpu 2>&1 | tee cpu_pd.log
```

## Test Result

---
<details>

</details>


bigPYJ1151 (approved):

tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh
```diff
@@ -91,8 +91,10 @@ get_model_args() {
 get_num_gpus() {
   if [[ "$SMI_BIN" == *"nvidia"* ]]; then
     echo "$($SMI_BIN --query-gpu=name --format=csv,noheader | wc -l)"
-  else
+  elif [[ "$SMI_BIN" == *"rocm"* ]]; then
     echo "$($SMI_BIN -l | grep GPU | wc -l)"
+  else
+    echo "1"
```
NickLucche, line 100: can you comment this line is for cpu?

tools/install_nixl_from_source_ubuntu.py
```diff
@@ -95,6 +95,7 @@ def install_system_dependencies():
         "meson",
         "libtool",
         "libtool-bin",
+        "pkg-config",
```
NickLucche, line 98: is this a new dep listed by nixl?

NickLucche (commented):
Thanks for the fix @xuechendi , this looks correct.

I wonder if we could achieve the same result by swapping H<>N dims in `get_kv_cache_shape` and then implementing `get_kv_cache_stride_order` with `(0, 1, 3, 2, 4)` to keep things aligned with gpu backends. 
Just an idea, wdyt? Otherwise lgtm

tools/install_nixl_from_source_ubuntu.py
xuechendi: Pkg-config is needed for compiling UCX, It is pre-installed in Intel GPU and Gaudi docker, but missed in CPU docker, without it, UCX compilation will soft fail without libplugin_UCX.so

tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh
xuechendi: This is not only for CPU, it also works for Intel Gaudi and intel GPU as well. Idea here is for non-cuda environment, we can leverage system to decide which card to use.

NickLucche: yep, fine as is just making sure test is readable

tools/install_nixl_from_source_ubuntu.py
NickLucche: ok I see, should we change said dockerfiles to omit this dependency?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1172,7 +1172,12 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
 
                 # TODO (NickLucche): Get kernel_block_size in a cleaner way
                 # NHD default "view" for non-MLA cache
-                kernel_block_size = cache.shape[-2] if self.use_mla else cache.shape[-3]
+                if self.device_type == "cpu":
+                    block_size_position = -2
+                else:
+                    block_size_position = -2 if self.use_mla else -3
```
NickLucche: can we move this if/else block out of the for lops to keep things cleaner?

NickLucche (approved):
@xuechendi alright thanks for considering layout swap, approved but left one minor comment to address before we merge.

tools/install_nixl_from_source_ubuntu.py
xuechendi: I don't think we need to omit, if system has pkg-config installed already, apt install will skip the installation.
This line here is to try install in case the system does not come with pkg-config pre-installed

tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh
xuechendi: Done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py

mgoin (approved):

PR #28917, opened 2025-11-18
[Core] Reuse created spec tokens lists to mitigate GC cost

Jialin:
## Purpose
Reuse the create spec tokens lists in InputBatch to mitigate GC costs.

In the current implementation, InputBatch.spec_token_ids would be frequently replaced by tokens newly created in scheduler output. So we're keep introducing medium / long living lists in the system, which means high GC costs.

## Test Plan & Test Result
CI signals

---
<details>

</details>


vllm/v1/worker/gpu_input_batch.py
```diff
@@ -656,7 +656,7 @@ def condense(self) -> None:
 
             spec_token_ids = self.spec_token_ids[last_req_index]
             self.spec_token_ids[empty_index] = spec_token_ids
-            self.spec_token_ids[last_req_index] = None
+            self.spec_token_ids[last_req_index].clear()
```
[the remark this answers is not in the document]
Jialin: Great catch, fixed.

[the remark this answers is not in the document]
Jialin: ditto, great catch, fixed in the latest change.

22quinn (approved):

vllm/v1/worker/gpu_input_batch.py
```diff
@@ -654,9 +654,15 @@ def condense(self) -> None:
             self.req_output_token_ids[last_req_index] = None
             self.req_id_to_index[req_id] = empty_index
 
-            spec_token_ids = self.spec_token_ids[last_req_index]
-            self.spec_token_ids[empty_index] = spec_token_ids
-            self.spec_token_ids[last_req_index] = None
+            if last_req_index != empty_index:
```
njhill, line 657: I don't think this if is needed, it will always be true.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -892,7 +892,8 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> None:
             # conform to the schema. This can result in
             # scheduler_output.scheduled_spec_decode_tokens being empty,
             # even when speculative decoding is enabled.
-            self.input_batch.spec_token_ids[req_index] = spec_token_ids
+            self.input_batch.spec_token_ids[req_index].clear()
+            self.input_batch.spec_token_ids[req_index].extend(spec_token_ids)
```
njhill, line 896: I don't understand this one, how is it better to repopulate an existing list and then discard the other one. Surely better to avoid that extra work and just use the other one? Either way a list is going to get garbage collected..

njhill (commented):
@Jialin sorry missed this before it was merged

vllm/v1/worker/gpu_model_runner.py
Jialin: Great question! The magic is hidden behind gc.freeze.

If we preallocate spec_token_ids lists, then run gc.freeze. The preallocate lists will NOT be GC tracked. And the spec_token_ids in this function is short living, majority of the cases, it would be gone due to reference counting, it will not be handled by gc.collect in most of the cases.

Let me know if you want more clarification :)

vllm/v1/worker/gpu_input_batch.py
Jialin: That's true! If not, we will have an even worse scenarios (e.g. req_id got reset to None above)

Get rid of it in #29031

vllm/v1/worker/gpu_model_runner.py
njhill: Thanks @Jialin, my general concern about these changes is that they sometimes come at the cost of added complexity and can negatively affect the readability/maintainability.

There is a balance where it's reasonable I think to give up minor perf benefits to keep the code simpler (of course case by case weighing magnitude of complexity vs magnitude of perf benefit).

This particular change for example I think is quite fragile, especially without comments explain why we are expending extra cycles here to optimize the object lifecycle in relation to GC. It's very likely someone could change this in future without being aware. I'm not sure whether it's realistic to enforce this (actually these lists will go away anyhow soon with MRV2).

Jialin: @njhill Totally agree with you for the tradeoff. And @zhuohan123 actually raised similar concerns with my other PRs as well. And I totally aligned with it.

We had internal TPGS runs to justify the impact. But moving forward, I think I should provide more data which is accessible for open access to avoid such confusion. And I'll hold off other GC changes landing to OSS side, until I came up with e2e benchmark to show e2e wins (both latency and GC costs) to justify such changes.

PR #28920, opened 2025-11-18
[ROCm] Validate quantization consistency for AITER shared experts fusion

Jasen2201:
## Purpose

Fix quantization mismatch issue for AITER shared experts fusion. When `VLLM_ROCM_USE_AITER_FUSION_SHARED_EXPERTS` is enabled, shared_experts and routed experts must use identical quantization configs. For example, Kimi v2 Think model uses bf16 for shared_experts but w4a16 for routed experts, causing fusion failures.

Changes:
- Set `VLLM_ROCM_USE_AITER_FUSION_SHARED_EXPERTS` default to False (was True)
- Add quantization consistency validation when fusion is enabled

## Test Plan

Test with Kimi v2 Think model with mixed quantization (bf16 shared_experts + w4a16 routed experts).

## Test Result

With `VLLM_ROCM_USE_AITER_FUSION_SHARED_EXPERTS=1`, validation correctly detects mismatch and provides clear error message.

### DIFF
--- a/vllm/envs.py
+++ b/vllm/envs.py
@@ -112,7 +112,7 @@ if TYPE_CHECKING:
     VLLM_ROCM_USE_AITER_TRITON_ROPE: bool = False
     VLLM_ROCM_USE_AITER_FP8BMM: bool = True
     VLLM_ROCM_USE_AITER_UNIFIED_ATTENTION: bool = False
-    VLLM_ROCM_USE_AITER_FUSION_SHARED_EXPERTS: bool = True
+    VLLM_ROCM_USE_AITER_FUSION_SHARED_EXPERTS: bool = False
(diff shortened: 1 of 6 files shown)


tjtanaa:
@zhyajie yea, let's disable by default in the other PR. However, I think we do need your logic from this PR. There is a PR to enable fused shared experts for CUDA as well. 
Can you help to make sure at least the the `fp8.py` also works by testing the official DeepSeek-R1 model? 
Basically, ensure the DeepSeek-R1 blockscale (official weights) and the DeepSeek-R1 (ptpc, quantized with llm-compressors) works).

We should be able to catch the abstraction issue for all other quantization config when we run the full CUDA CI suites.

PR #28908, opened 2025-11-18
[CI/Build] Replace wikipedia url with local server ones

Isotr0py:
## Purpose
- Fix flaky `entrypoints/openai/test_metrics.py::test_metrics_exist[-multimodal]` test (https://buildkite.com/vllm/ci/builds/39344/steps/canvas?jid=019a951c-ed1d-4bd1-a6a9-4a92c2fc3ecd):
```
FAILED entrypoints/openai/test_metrics.py::test_metrics_exist[-multimodal] - openai.InternalServerError: Error code: 500 - {'error': {'message': "403, message='Forbidden', url='https://upload.wikimedia.org/wikipedia/commons/thumb/d/dd/Gfp-wisconsin-madison-the-nature-boardwalk.jpg/2560px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg'", 'type': 'Internal Server Error', 'param': None, 'code': 500}}
```

## Test Plan
```
pytest -s -v tests/entrypoints/openai/test_metrics.py::test_metrics_exist[-multimodal]
```

## Test Result
Test should use image on s3 now.

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x] The test plan, such as providing test command.
- [x] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #28921, opened 2025-11-18
[chore] Move the rest of wikimedia url to S3

khluu:
No more wikimedia...

### DIFF
--- a/tests/v1/entrypoints/openai/serving_responses/test_image.py
+++ b/tests/v1/entrypoints/openai/serving_responses/test_image.py
@@ -15,10 +15,10 @@ MODEL_NAME = "Qwen/Qwen2.5-VL-3B-Instruct"
 MAXIMUM_IMAGES = 2
 # Test different image extensions (JPG/PNG) and formats (gray/RGB/RGBA)
 TEST_IMAGE_ASSETS = [
-    "2560px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg",  # "https://upload.wikimedia.org/wikipedia/commons/thumb/d/dd/Gfp-wisconsin-madison-the-nature-boardwalk.jpg/2560px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg"
-    "Grayscale_8bits_palette_sample_image.png",  # "https://upload.wikimedia.org/wikipedia/commons/f/fa/Grayscale_8bits_palette_sample_image.png",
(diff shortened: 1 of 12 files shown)


tests/v1/entrypoints/openai/serving_responses/test_image.py
```diff
@@ -15,10 +15,10 @@
 MAXIMUM_IMAGES = 2
 # Test different image extensions (JPG/PNG) and formats (gray/RGB/RGBA)
 TEST_IMAGE_ASSETS = [
-    "2560px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg",  # "https://upload.wikimedia.org/wikipedia/commons/thumb/d/dd/Gfp-wisconsin-madison-the-nature-boardwalk.jpg/2560px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg"
-    "Grayscale_8bits_palette_sample_image.png",  # "https://upload.wikimedia.org/wikipedia/commons/f/fa/Grayscale_8bits_palette_sample_image.png",
-    "1280px-Venn_diagram_rgb.svg.png",  # "https://upload.wikimedia.org/wikipedia/commons/thumb/9/91/Venn_diagram_rgb.svg/1280px-Venn_diagram_rgb.svg.png",
-    "RGBA_comp.png",  # "https://upload.wikimedia.org/wikipedia/commons/0/0b/RGBA_comp.png",
+    "2560px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg",  # "https://vllm-public-assets.s3.us-west-2.amazonaws.com/vision_model_images/2560px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg"
```
DarkLight1337, line 18: I think we can also use local server for these ones just like in #28908?

noooop (approved):

tests/v1/entrypoints/openai/serving_responses/test_image.py
khluu: I think this is using local server? It's just a comment to reference the source of the image

ywang96 (approved):
LGTM - we probably need to doc somewhere on how to add a new asset to our S3 for future tests? (If we don't already have one)

tests/v1/entrypoints/openai/serving_responses/test_image.py
noooop: I think we should gradually store all resources used by CI locally, so that we can complete all CI tests without using the network. Otherwise, there will be failures that suddenly occur in large numbers during the busiest release time LOL

DarkLight1337: Oh, right, I got confused, this just changes the comments

DarkLight1337 (approved):
LGTM, thanks

Isotr0py (approved):

PR #28927, opened 2025-11-18
[CI][NIXL] Change default `block_size` for tests

NickLucche:
Testing on `block_size=16` is less aligned with real use-cases in which nixl is used.
Bumping up default value so that we are tracking this important value.


njhill (commented):
Thanks @NickLucche

njhill (approved):

PR #27126, opened 2025-10-17
[compile] Enable sequence parallelism matching w/o custom ops enabled 

angelayi:
## Purpose

Based on https://github.com/vllm-project/vllm/pull/24604, modified sequence-parallelism pass to do custom op matching w/o needing to enable the custom op

## Test Plan

`pytest -sv tests/compile/test_sequence_parallelism.py`

## Performance numbers

I did some benchmarking with the command on H100 w/o flashinfer

```
VLLM_DISABLE_COMPILE_CACHE=1 VLLM_USE_STANDALONE_COMPILE=1 VLLM_LOGGING_LEVEL=DEBUG vllm bench latency --model=nvidia/Llama-3.3-70B-Instruct-FP8 --output-len 1 --input-len 8192 --batch-size 1 --tensor-parallel-size 8 --load-format dummy --num_iters_warmup 5 --num_iters 15 -O '{"level": 3, "use_inductor_graph_partition": false, "splitting_ops":[], "cudagraph_mode": "FULL", }' --no-enable-prefix-caching
```

while varying
* `"pass_config": {"enable_async_tp": true, "enable_sequence_parallelism": true}` vs. `"pass_config": {"enable_async_tp": false, "enable_sequence_parallelism": false}`
* `"custom_ops":["+quant_fp8", "+rms_norm"]` vs. `"custom_ops":[]`

image

PR #28935, opened 2025-11-18
[Ascend]: Fixed the issue where OOT Platform vllm-ascend could not enable SP in Eager mode

leo-pony:
Before get lens of splitting_ops should first check it is not None; otherwise, in piecewise mode without set splitting_ops error will occur:
```
        if self.compilation_config.pass_config.enable_sequence_parallelism:
            # With pipeline parallelism or dynamo partitioning,
            # native rms norm tracing errors due to incorrect residual shape.
            # Use custom rms norm to unblock. In the future,
            # the pass will operate on higher-level IR to avoid the issue.
            # TODO: https://github.com/vllm-project/vllm/issues/27894
            is_fullgraph = (
                self.compilation_config.use_inductor_graph_partition
>               or len(self.compilation_config.splitting_ops) == 0
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
            )
E           TypeError: object of type 'NoneType' has no len()
```

The pull request for this issue is here: https://github.com/vllm-project/vllm/pull/27126

For the value of splitting_ops: If None, defaults to attention ops for piecewise cudagraphs , and If empty list [], no ops are excluded (suitable for full cudagraphs).  Detail see comments from vllm/config/compilation.py:

    splitting_ops: list[str] | None = None
    """A list of ops to exclude from cudagraphs, used in piecewise compilation.

    The behavior depends on use_inductor_graph_partition:

    - When use_inductor_graph_partition=False (default):
        These ops are used for Dynamo FX-level graph splitting. The graph is
        split at these ops before Inductor compilation, creating separate
        subgraphs for cudagraph capture.

    - When use_inductor_graph_partition=True:
        These ops are used to register Inductor partition rules. The graph
        partitioning happens at Inductor codegen time after all passes and
        fusions are finished, allowing compilation and custom passes to operate
        on the full graph while still excluding these ops from cudagraphs.

    If None, defaults to attention ops for piecewise cudagraphs.
    If empty list [], no ops are excluded (suitable for full cudagraphs).
    """

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/config/vllm.py
+++ b/vllm/config/vllm.py
@@ -835,8 +835,7 @@ class VllmConfig:
         ), "MTP with cp_kv_cache_interleave_size > 1 is not supported now."
 
         # Do this after all the updates to compilation_config.mode
-        if self.compilation_config.mode == CompilationMode.VLLM_COMPILE:
-            self.compilation_config.set_splitting_ops_for_v1()
+        self.compilation_config.set_splitting_ops_for_v1()
 
         if self.compilation_config.pass_config.enable_sequence_parallelism:
             # With pipeline parallelism or dynamo partitioning,
@@ -844,6 +843,13 @@ class VllmConfig:
             # Use custom rms norm to unblock. In the future,
             # the pass will operate on higher-level IR to avoid the issue.
             # TODO: https://github.com/vllm-project/vllm/issues/27894
+            if self.compilation_config.mode != CompilationMode.VLLM_COMPILE:
+                logger.warning(
+                    "Sequence parallelism is enabled, but running in wrong "
+                    "vllm compile mode: %s.",
+                    self.compilation_config.mode,
+                )
+
             is_fullgraph = (
                 self.compilation_config.use_inductor_graph_partition
                 or len(self.compilation_config.splitting_ops) == 0
--- a/vllm/config/compilation.py
+++ b/vllm/config/compilation.py
@@ -855,6 +855,13 @@ class CompilationConfig:
         self.compute_bs_to_padded_graph_size()
 
     def set_splitting_ops_for_v1(self):
+        # To compatible with OOT hardware plugin platform (for example vllm-ascend)
+        # which currently only supports sequence parallelism in eager mode.
+        if self.mode != CompilationMode.VLLM_COMPILE:
+            if self.splitting_ops is None:
+                self.splitting_ops = []
+            return
+
         # NOTE: this function needs to be called only when mode is
         # CompilationMode.VLLM_COMPILE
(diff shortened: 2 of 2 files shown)


angelayi (approved):
thanks!

yewentao256 (commented):
Could you take a look the original cause, namely, why there is a None in the context and fix that?

vllm/config/vllm.py
```diff
@@ -639,7 +639,7 @@ def __post_init__(self):
             # TODO: https://github.com/vllm-project/vllm/issues/27894
             is_fullgraph = (
                 self.compilation_config.use_inductor_graph_partition
-                or len(self.compilation_config.splitting_ops) == 0
+                or self.compilation_config.splitting_ops == []
```
ProExpertProg: Right above this line, we call `self.compilation_config.set_splitting_ops_for_v1()` which sets the splitting ops to a non-None value. What case are you running that has the SP pass enabled but compilation mode != VLLM_COMPILE? That's not a supported case.

I do agree we should handle invalid configuration gracefully. If you want to fix this, can you move the call to `set_splitting_ops_for_v1` outside of the if, and just set to empty list if mode != VLLM_COMPILE? and then if `enable_sequence_parallelism` is True but mode is wrong, add a warning?

ProExpertProg (changes requested):

vllm/config/vllm.py
wangxiyuan: if enalbe sp with eager mode, the compile mode is None
https://github.com/vllm-project/vllm/blob/da94c7c0eb8dabea9c500dbd70fa042497497689/vllm/config/vllm.py#L440-L444

I know that sp should work with graph mode in vLLM. But for some platform, such as vllm-ascend, it doesn't work with graph pass currently, so it implements sp with another way to let sp work with eager mode. The e2e test is:
https://github.com/vllm-project/vllm-ascend/blob/main/tests/e2e/multicard/test_offline_inference_distributed.py#L169

Future plan:
We're working on custom graph pass feature https://github.com/vllm-project/vllm/pull/28623 Once it's done, we'll implement sp with the same way with vLLM.

ProExpertProg: Okay thanks for the clarification. Can you update `set_splitting_ops_for_v1` so that splitting ops aren't empty during enforce eager?

22quinn: Facing the same issue. @wangxiyuan Gentle ping if there's any update

wangxiyuan: @leo-pony please update the PR ASAP, thanks.

ProExpertProg (approved):

PR #28938, opened 2025-11-18
[NVIDIA] Guard SM100 CUTLASS MoE macro to SM100 builds v2

johnnynunez:
fix DGX Spark vllm issue

## Purpose

Continue with this PR. No answer from owner. https://github.com/vllm-project/vllm/pull/26844

cc @mgoin


wrmedford (approved):
Looks good to me, and preserves functionality on sm110a across its rename.

mgoin (approved):
Thank you!

ericcurtin:
I see we released a new wheel with this fix for DGX Spark, should we expect the aarch64 wheel to be compatible with DGX Spark and run accelerated workloads soon?

https://github.com/vllm-project/vllm/releases/tag/v0.11.2

bbrowning:
I was able to install release v0.11.2 via these commands on a DGX spark:

```
mkdir -p ~/tmp/vllm-v0.11.2
cd ~/tmp/vllm-v0.11.2
uv venv --python 3.12 --seed
source .venv/bin/activate
uv pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cu130
uv pip install "vllm==v0.11.2" --no-binary vllm --torch-backend=auto
```

That compiled from source without issue. So, while we don't have any wheel releases that work yet for DGX Spark, release v0.11.2 does install on the system without extra hacks.

johnnynunez:
> Does `vllm serve "HuggingFaceTB/SmolLM-135M-Instruct"` work with this installation technique?

just try, now the best backend for spark is flash infer

bbrowning:
@ericcurtin I was able to build and run a functioning v0.11.2 container directly on my DGX Spark with the Dockerfile at https://gist.github.com/bbrowning/e2efe77b617b741a23ed31333a7ecba9 - it just takes the first bits of the official vLLM container and installs from releases instead of source, along with removing as much of the unnecessary bits I could find to simplify things for just this one use case.

Make sure to pass `--gpus=all` when running the built container. The entrypoint is set to `vllm serve` just like the official container, so pass whatever args you need after that.

ericcurtin:
@bbrowning I wonder is it worth trying to open a PR like that against here:

https://github.com/vllm-project/vllm/blob/main/docker/Dockerfile

```
if aarch64; then
  do things a little differently...
fi
```

johnnynunez:
hello,
@bbrowning @ericcurtin 0.11.2 is compatible with all aarch64, the problem of that type of models is the missing support for nvfp4 around sm12x

ericcurtin:
> @ericcurtin I think that if we start releasing wheels built with CUDA acceleration for aarch64 then we may not need a custom Dockerfile, and can just use the standard one? It may take a few iterations to get that right, but I don't think it has to be a fork of the actual logic in the Dockerfile itself if we get usable accelerated wheels for this platform.

Sure that's even better, vLLM CUDA on an x86_64 machine is pretty much:

uv pip install vllm

if DGX Spark became pretty much that too, that would be even better than a Dockerfile.

But testing that same Dockerfile on both would be worth doing of course :)

PR #28948, opened 2025-11-18
[Log] Optimize startup log

yewentao256:
## Purpose

Reduce duplicate logs when using dp/tp


vllm/v1/worker/gpu_worker.py
```diff
@@ -91,18 +91,19 @@ def __init__(
...
-            )
-            logger.debug(
-                "Profiler config: record_shapes=%s,"
-                "profile_memory=%s,with_stack=%s,with_flops=%s",
-                envs.VLLM_TORCH_PROFILER_RECORD_SHAPES,
-                envs.VLLM_TORCH_PROFILER_WITH_PROFILE_MEMORY,
-                envs.VLLM_TORCH_PROFILER_WITH_STACK,
-                envs.VLLM_TORCH_PROFILER_WITH_FLOPS,
-            )
+            if getattr(self.parallel_config, "data_parallel_rank", 0) == 0:
```
njhill: I don't think we need to use `get_attr` here since it should always be an attribute.

Also we should maybe use `data_parallel_rank_local` instead so that this is logged in each node?
e.g.
```suggestion
            if self.parallel_config.data_parallel_rank_local in (None, 0):
```

njhill (commented):
Thanks @yewentao256!

njhill (approved):
Thanks @yewentao256

PR #28952, opened 2025-11-18
Relax Transformers modeling backend MoE experts check

hmellor:
The experts module could now also be a 3D tensor.

Note that there are other issues on Transformers main which currently break Transformers modeling backend MoE support. These are being worked on separately.

### DIFF
--- a/vllm/model_executor/models/transformers/moe.py
+++ b/vllm/model_executor/models/transformers/moe.py
@@ -256,7 +256,13 @@ class MoEMixin(MixtureOfExperts):
         def _recursive_replace(module: nn.Module, prefix: str):
             for child_name, child_module in module.named_children():
                 qual_name = maybe_prefix(prefix, child_name)
-                if child_name == "experts" and isinstance(child_module, nn.ModuleList):
+                # Naive implementations will have experts as ModuleList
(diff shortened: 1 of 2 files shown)


vllm/model_executor/models/transformers/moe.py
```diff
@@ -256,7 +256,14 @@ def forward(self, *args, **kwargs):
...
             for child_name, child_module in module.named_children():
                 qual_name = maybe_prefix(prefix, child_name)
-                if child_name == "experts" and isinstance(child_module, nn.ModuleList):
+                # Naive implementations will have experts as ModuleList
+                is_modulelist = isinstance(child_module, nn.ModuleList)
+                # Packed implementations will have experts as 3D tensors of shapes like:
+                # gate_up_proj = (num_experts, 2 * intermediate_size, hidden_size)
+                # down_proj = (num_experts, intermediate_size, hidden_size)
+                params = list(child_module.parameters())
+                is_3d = bool(params) and all(p.ndim == 3 for p in params)
```
DarkLight1337:
```suggestion
                is_3d = len(params) > 0 and all(p.ndim == 3 for p in params)
```

I prefer to be more explicit

Isotr0py (approved):

PR #28957, opened 2025-11-18
Update Dockerfile to use gcc-toolset-14 and fix test case failures on power (ppc64le)

bhagyashrigai:
## Purpose
This PR updates the ppc64le Dockerfile to use gcc-toolset-14 instead of gcc-toolset-13 and includes additional environment fixes required for successful builds. The update resolves multiple test case failures observed during vLLM and dependency builds on the ppc64le platform.

**Key Changes**

- Updated all Dockerfile stages from gcc-toolset-13 → gcc-toolset-14.
- Updated compiler activation paths (source /opt/rh/gcc-toolset-14/enable).
- Exported LLVM_CONFIG and updated PATH to ensure correct LLVM detection.
- Added missing dependency (libicu) for ppc64le build stability.
- Updated platform selectors in requirements/common.txt to include:platform_machine == "ppc64le" for llguidance and xgrammar

## Test Result
[test_logs.txt](https://github.com/user-attachments/files/23656661/test_logs.txt)

---
<details>

</details>


bigPYJ1151 (approved):

requirements/common.txt
hmellor, line 1: Are there any other platforms we support? If we can now install these for all platforms we could remove the specifiers?

bigPYJ1151: There still has ```Darwin```, the macOS.

PR #25444, opened 2025-09-23
[Perf] Change default CUDAGraphMode from PIECEWISE to FULL_AND_PIECEWISE

mgoin:
## Purpose

This PR proposes to enable using full cudagraphs by default in vLLM V1. Support for cudagraphs beyond piecewise only was added over the past months (notably https://github.com/vllm-project/vllm/pull/20059) and while the startup time increase is measurable, we believe the performance gain for full cudagraphs is worth it, especially for low latency with small models or MoEs.

For instance, running Qwen/Qwen3-Coder-30B-A3B-Instruct-FP8 on 1xH100 with vLLM for 10 requests of 1024 in and 128 out results in the following throughputs:
* `vllm serve` (default): 4.07 req/s
* `vllm serve --async-scheduling`: 4.29 req/s
* `vllm serve -O.cudagraph_mode=FULL`: 5.83 req/s
* `vllm serve --async-scheduling -O.cudagraph_mode=FULL`: 6.03 req/s
* `vllm serve -O.cudagraph_mode=FULL_AND_PIECEWISE`: 5.98 req/s
* `vllm serve --async-scheduling -O.cudagraph_mode=FULL_AND_PIECEWISE`: 6.20 req/s

Benchmark command: `vllm bench serve --model Qwen/Qwen3-Coder-30B-A3B-Instruct-FP8 --port 8000 --num-prompts 10`

====

Startup time impact:

```
main: python -c "from vllm import LLM;LLM('Qwen/Qwen3-0.6B')"  34.05s user 7.32s system 190% cpu 21.684 total
pr:   python -c "from vllm import LLM;LLM('Qwen/Qwen3-0.6B')"  34.99s user 9.01s system 175% cpu 25.005 total
```

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #28971, opened 2025-11-18
[LoRA] Support FusedMoE LoRA Triton kernel for mxfp4

xyang16:
## Purpose

This PR is to support FusedMoE LoRA Triton kernel for mxfp4 model.
* Add UnfusedOAITritonExperts
* Inject lora module in activation: Since matmul_ogs can fuse activation, set fused_activation to None to unfuse activation in the first matmul_ogs.
* Inject lora module in moe_sum: This need to unfuse sum in second matmul_ogs. Grouped reduction does scatter + accumulate, it is essentially equal to: `y[dst_indx // n_expts_act, :] += x[src_indx, :]`, so that scatter sum across multiple experts, and collapse `M * topk` to `M` rows. Therefore, we need to set `routing_data.n_expts_act` to 1, so it doesn't sum across multiple experts, in order unfuse moe_sum in the second matmul_ogs.
* Added test_modular_oai_triton_moe.py

## Test Plan

```
pytest -s -v tests/kernels/moe/test_modular_oai_triton_moe.py
```

## Test Result

Tests passed.

## Benchmark

Baseline (marlin):

```
VLLM_MXFP4_USE_MARLIN=1 vllm serve openai/gpt-oss-20b \
  --tensor-parallel-size 1 \
  --max-num-seqs 16 \
  --compilation-config '{"max_cudagraph_capture_size": 128, "compile_sizes": [1, 2, 4, 8, 16]}' \
  --enable-lora \
  --max-loras 6 \
  --lora-modules lora1=/opt/dlami/nvme/models/gpt-oss-20b-lora-gsm8k \
  --max-lora-rank 64
```

```
vllm bench serve \
  --model openai/gpt-oss-20b \
  --lora-modules lora1 \
  --dataset-name sharegpt \
  --max-concurrency 16 \
  --num-prompts 1000 \
  --num-warmups 60 \
  --ignore-eos
```

```
Maximum request concurrency:             16        
Benchmark duration (s):                  142.34    
Total input tokens:                      215312    
Total generated tokens:                  199033    
Request throughput (req/s):              7.03      
Output token throughput (tok/s):         1398.29   
Peak output token throughput (tok/s):    1677.00   
Peak concurrent requests:                29.00     
Total Token throughput (tok/s):          2910.95   
Mean TTFT (ms):                          43.24     
Median TTFT (ms):                        26.77     
P99 TTFT (ms):                           140.09    
Mean TPOT (ms):                          11.19     
Median TPOT (ms):                        10.93     
P99 TPOT (ms):                           17.44     
Mean ITL (ms):                           11.12     
Median ITL (ms):                         9.73      
P99 ITL (ms):                            56.55     
```

PR (triton):

Install triton_kernels

```
pip install "git+https://github.com/triton-lang/triton.git@0a2e3a391cbb9e13d29bf12a2a0005e358102d74#subdirectory=python/triton_kernels"
```

```
  --tensor-parallel-size 1 \
  --max-num-seqs 16 \
  --compilation-config '{"max_cudagraph_capture_size": 128, "compile_sizes": [1, 2, 4, 8, 16]}' \
  --enable-lora \
  --max-loras 6 \
  --lora-modules lora1=/opt/dlami/nvme/models/gpt-oss-20b-lora-gsm8k \
  --max-lora-rank 64
```

```
vllm bench serve \
  --model openai/gpt-oss-20b \
  --lora-modules lora1 \
  --dataset-name sharegpt \
  --max-concurrency 16 \
  --num-prompts 1000 \
  --num-warmups 60 \
  --ignore-eos
```

```
Maximum request concurrency:             16        
Benchmark duration (s):                  125.59    
Total input tokens:                      215312    
Total generated tokens:                  199033    
Request throughput (req/s):              7.96      
Output token throughput (tok/s):         1584.77   
Peak output token throughput (tok/s):    2171.00   
Peak concurrent requests:                29.00     
Total Token throughput (tok/s):          3299.15   
Mean TTFT (ms):                          52.14     
Median TTFT (ms):                        19.03     
P99 TTFT (ms):                           129.60    
Mean TPOT (ms):                          9.82      
Median TPOT (ms):                        9.28      
P99 TPOT (ms):                           21.31     
Mean ITL (ms):                           9.72      
Median ITL (ms):                         7.32      
P99 ITL (ms):                            90.10     
```

## Accuracy Testing

* 20b deepep triton

```
VLLM_ALL2ALL_BACKEND="deepep_high_throughput" vllm serve openai/gpt-oss-20b \
  --tensor-parallel-size 1 \
  --data-parallel-size 2 \
  --enable-expert-parallel \
  --no-enable-prefix-caching
```

```
OPENAI_API_KEY=EMPTY python3 -m gpt_oss.evals --model openai/gpt-oss-20b --eval gpqa --n-threads 200 --reasoning-effort low
```

```
Writing report to /tmp/gpqa_openai__gpt-oss-20b-low_temp1.0_20251125_122239.html
{'chars': np.float64(55.219065656565654), 'chars:std': np.float64(220.291583893894), 'score': np.float64(0.5707070707070707), 'score:std': np.float64(0.4949752621616814)}
Writing results to /tmp/gpqa_openai__gpt-oss-20b-low_temp1.0_20251125_122239.json
Writing all results to /tmp/gpqa_openai__gpt-oss-20b-low_temp1.0_20251125_122239_allresults.json
[{'eval_name': 'gpqa', 'model_name': 'openai__gpt-oss-20b-low_temp1.0_20251125_122239', 'metric': 0.5707070707070707}]
```

* 20b lora triton

```
  --tensor-parallel-size 1 \
  --compilation-config '{"cudagraph_mode": "PIECEWISE"}' \
  --max-num-seqs 16 \
  --enable-lora \
  --max-loras 6 \
  --lora-modules lora1=/opt/dlami/nvme/models/gpt-oss-20b-lora/checkpoint-13 \
  --max-lora-rank 64
```

```
OPENAI_API_KEY=EMPTY python3 -m gpt_oss.evals --model lora1 --eval gpqa --n-threads 200 --reasoning-effort low
```

```
Writing report to /tmp/gpqa_openai__gpt-oss-20b-low_temp1.0_20251126_124509.html
{'chars': np.float64(62.859848484848484), 'chars:std': np.float64(230.21663448129942), 'score': np.float64(0.577020202020202), 'score:std': np.float64(0.4940322747359399)}
Writing results to /tmp/gpqa_openai__gpt-oss-20b-low_temp1.0_20251126_124509.json
Writing all results to /tmp/gpqa_openai__gpt-oss-20b-low_temp1.0_20251126_124509_allresults.json
[{'eval_name': 'gpqa', 'model_name': 'openai__gpt-oss-20b-low_temp1.0_20251126_124509', 'metric': 0.577020202020202}]
```

* 20b lora marlin

```
VLLM_MXFP4_USE_MARLIN=1 vllm serve openai/gpt-oss-20b \
  --tensor-parallel-size 1 \
  --compilation-config '{"cudagraph_mode": "PIECEWISE"}' \
  --max-num-seqs 16 \
  --enable-lora \
  --max-loras 6 \
  --lora-modules lora1=/opt/dlami/nvme/models/gpt-oss-20b-lora/checkpoint-13 \
  --max-lora-rank 64
```

```
OPENAI_API_KEY=EMPTY python3 -m gpt_oss.evals --model lora1 --eval gpqa --n-threads 200 --reasoning-effort low
```

```
Writing report to /tmp/gpqa_lora1-low_temp1.0_20251126_125725.html
{'chars': np.float64(25.56691919191919), 'chars:std': np.float64(165.94325370775), 'score': np.float64(0.5669191919191919), 'score:std': np.float64(0.4955015860245882)}
Writing results to /tmp/gpqa_lora1-low_temp1.0_20251126_125725.json
Writing all results to /tmp/gpqa_lora1-low_temp1.0_20251126_125725_allresults.json
[{'eval_name': 'gpqa', 'model_name': 'lora1-low_temp1.0_20251126_125725', 'metric': 0.5669191919191919}]
```

---
<details>

</details>


robertgshaw2-redhat:
nice speedup!

vllm/lora/layers/fused_moe.py
```diff
@@ -28,6 +28,9 @@
 from vllm.model_executor.layers.fused_moe.fused_moe_modular_method import (
```
varun-sundar-rabindranath, line 29: @xyang16  can you add a unit test for gpt-oss lora + triton_kernels. The test can be predicated on `has_triton_kernels` like in https://github.com/vllm-project/vllm/blob/main/tests/kernels/moe/test_gpt_oss_triton_kernels.py

vllm/lora/ops/triton_ops/fused_moe_lora_op.py
```diff
@@ -430,7 +430,7 @@ def _fused_moe_lora(
         == expert_ids.shape[0]
         == num_tokens_post_padded.shape[0]
     )
-    assert len(lora_b_stacked) * lora_b_stacked[0].shape[-2] == output.shape[-1]
+    assert output.shape[-1] // lora_b_stacked[0].shape[-2] == len(lora_b_stacked)
```
varun-sundar-rabindranath: is this change required ? It looks like `a * b == c` is transformed into `c // b == a` ? I guess it is required because of the `floor` operation `//` . To better understand, can you provide and example where the first one fails and replacement passes. Thanks.

vllm/model_executor/layers/quantization/mxfp4.py
```diff
@@ -74,24 +74,9 @@ class Mxfp4Backend(Enum):
...
-    return Mxfp4Backend.MARLIN
-
-
-def get_mxfp4_backend(with_lora_support: bool) -> Mxfp4Backend:
+def get_mxfp4_backend() -> Mxfp4Backend:
     # Backend Selection
 
-    if with_lora_support:
-        return get_mxfp4_backend_with_lora()
-
```
varun-sundar-rabindranath, line 94: I think we should maintain `get_mxfp4_backend_with_lora()` and return the appropriate backend from within that function. This is because, there is no guarantee that the logic below will choose a LoRA compatible backend.

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
```diff
@@ -297,20 +341,39 @@ def apply(
...
             quant_config=self.quant_config,
             apply_router_weight_on_input=False,
             global_num_experts=local_num_experts,
             expert_map=None,  # applied already
+            intermediate_cache13=workspace2,
+            intermediate_cache2=workspace13,
             a1q_scale=a1q_scale,
         )
 
-        output.copy_(experts_output, non_blocking=True)
```
varun-sundar-rabindranath: I am not sure I understand why this removal is required. It looks like the `output` tensor isn't being filled anywhere for the non LoRA case ? Am I missing something ? 

also since this class is declared to return `TopKWeightAndReduceNoOP()` in `finalize_weight_and_reduce_impl` above,  the `apply` method is expected to fill in the output in the output tensor and other parts of ModularKernel depend on that contract. 

I see the redundant copy below in `topk_weight_and_reduce.py` below, for that I think we should avoid it by doing a `.data_ptr()` equivalence check between tensors.

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
```diff
@@ -157,16 +178,37 @@ def triton_kernel_fused_experts(
         fused_activation=act,
     )
```
varun-sundar-rabindranath, line 183: @xyang16  To guarantee that `matmul_ogs` will return an output shaped [M * topk, N], I think it is better to pass in the output tensor ourselves using the argument `y`. Note that `matmul_ogs` actually checks if the output is of expected size here https://github.com/triton-lang/triton/blob/c3c476f357f1e9768ea4e45aa5c17528449ab9ef/python/triton_kernels/triton_kernels/matmul_ogs.py#L180 . That way it is guaranteed that `matmul_ogs` will respect the contract. 

Same for the second matmul_ogs also.

vllm/model_executor/layers/fused_moe/prepare_finalize.py
```diff
@@ -65,10 +65,10 @@ def finalize(
         topk_ids: torch.Tensor,
         apply_router_weight_on_input: bool,
         weight_and_reduce_impl: mk.TopKWeightAndReduce,
-    ) -> None:
+    ) -> torch.Tensor:
```
varun-sundar-rabindranath: This violates the base class contract at https://github.com/vllm-project/vllm/blob/814843e021a3618f7f8e494d4f0d4fd561cf3225/vllm/model_executor/layers/fused_moe/modular_kernel.py#L251

vllm/model_executor/layers/fused_moe/topk_weight_and_reduce.py
```diff
@@ -68,8 +68,7 @@ def apply(
             f"But got output={output.size()}, "
             f"used_expert_output={fused_expert_output.size()}"
         )
-        output.copy_(fused_expert_output, non_blocking=True)
```
varun-sundar-rabindranath, line 71: When `output` and `fused_expert_output` tensors are the same, I believe this copy should be avoided by doing, 
```
if (output.data_ptr() != fused_expert_output.data_ptr()):
       output.copy_(fused_expert_output, non_blocking=True)
```
I think this will prevent changing the signature of `MoEPrepareAndFinalizeNoEP::finalize()` method.

varun-sundar-rabindranath (changes requested):
The changes to `triton_kernel_fused_experts` are invasive and it is a bit confusing reason about the `fused_act=True, fuse_sum=True` and `fuse_act=False,fused_sum=False` cases as the assumptions and expectations from `matmul_ogs` is different in both cases. 

The main difference between the non-LoRA and the LoRA case, 
 - For the non-LoRA case, no assumptions about the sizes of `matmul_ogs` output tensors are made. The only requirement here is that the second `matmul_ogs` must return a tensor of size [M, K]. For the LoRA case, we expect the outputs to be of a specific shape - This pattern is similar to [TritonExperts](https://github.com/vllm-project/vllm/blob/4c23690f43e51eccf6ce5866ac47adcf39215e4d/vllm/model_executor/layers/fused_moe/fused_moe.py#L1984) 
 - For the non-LoRA case, there are no requirements on the `gather_indx` and `scatter_indx` sizes. The LoRA case requires the tensors in these objects to be a specific shape. 
 
 For these reasons, I think it will be better to create a separate implementation of `BaseOAITritonExperts` class for the LoRA case, naming it something like `UnfusedOAITritonExperts`. Apart of being easier to assert for expectations, with this we can create correct and adequate workspace shapes  for both workspace13 and workspace2 and reuse them properly in the implementation. Please refer to [TritonExperts](https://github.com/vllm-project/vllm/blob/4c23690f43e51eccf6ce5866ac47adcf39215e4d/vllm/model_executor/layers/fused_moe/fused_moe.py#L1984)  I think the implementation here would be very similar and all the logic could be contained within the `apply` function, thus not disturbing the existing `triton_kernel_fused_experts` function. something like, 
 
```
def apply():
        routing_data, gather_indx, scatter_indx = self._make_routing_data(
            topk_ids, topk_weights, local_num_experts
        )
        matmul_ogs(..., 
                             y = intermediate_cache1)
        activation(intermediate_cache2, intermediate_cache1)
       matmul_ogs(...,
                            y = intermediate_cache3)
```

vllm/lora/ops/triton_ops/fused_moe_lora_op.py
xyang16: Thanks for reviewing! I changed this since output.shape[-1] is padded because of mxfp4 swizzle. I put some explanations for my other changes as well in the Notes in description.

vllm/model_executor/layers/fused_moe/prepare_finalize.py
xyang16: Changed back. Thanks.

vllm/model_executor/layers/fused_moe/topk_weight_and_reduce.py
xyang16: Changed back.

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
xyang16: Raised a separate PR https://github.com/vllm-project/vllm/pull/29219 for this

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
xyang16: Changed back. Thanks.

vllm/model_executor/layers/quantization/mxfp4.py
xyang16: Added back get_mxfp4_backend_with_lora. Thanks for review!

vllm/lora/layers/fused_moe.py
xyang16: Added test_modular_oai_triton_moe.py. Thanks!

tests/lora/test_gptoss_tp.py
```diff
@@ -93,6 +94,7 @@ def test_gpt_oss_lora_tp2(gptoss20b_lora_files):
         max_lora_rank=8,
         tensor_parallel_size=2,
         compilation_config=vllm.config.CompilationConfig(  # Avoid OOM
+            max_cudagraph_capture_size=48,
```
varun-sundar-rabindranath: is this required to avoid OOMs ?

vllm/lora/layers/fused_moe.py
```diff
@@ -110,16 +112,15 @@ def _inject_lora_into_fused_moe(self):
...
-                quant_config, shared_experts=self.base_layer.shared_experts
-            )
+        prepare_finalize = MoEPrepareAndFinalizeNoEP()
+        m_fused_moe_fn = FusedMoEModularKernel(
+            prepare_finalize,
+            self.base_layer.quant_method.select_gemm_impl(
+                prepare_finalize, self.base_layer
+            ),
+            self.base_layer.shared_experts,
+            getattr(self.base_layer, "shared_experts_stream", None),
```
varun-sundar-rabindranath, line 133: I think we should assert below that we are getting the Experts that we are expecting for the different cases.

tests/kernels/moe/test_modular_oai_triton_moe.py
```diff
@@ -0,0 +1,250 @@
...
+            w2_precision_config,
+            w1_bias_tri,
+            w2_bias_tri,
+            num_experts,
+            topk_weights,
+            topk_ids,
+            unfused,
+        )
+
+    assert_close(ref=out_ref, tri=out, maxtol=0.025, rmstol=0.005)
```
varun-sundar-rabindranath, line 250: Nice tests ! Thanks @xyang16

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
```diff
@@ -221,6 +224,47 @@ def __init__(self, quant_config: FusedMoEQuantConfig):
...
+        K = a1.size(-1)
+
+        if a1.dim() == 2:
+            # Make sure we are using the correct a1 (pre-permute).
+            assert topk_ids.size(0) == a1.size(0), f"{topk_ids.size(0)} != {a1.size(0)}"
+            M = a1.size(0)
+        else:
+            assert a1.dim() == 3
+            assert a1.size(0) == E, f"{a1.size(0)} == {E}"
+            M = a1.size(1)  # This is max_num_tokens
```
varun-sundar-rabindranath: Given both `UnfusedOAITritonExperts` and `OAITritonExperts` support only the `mk.FusedMoEActivationFormat.Standard` format, we don't need to handle the `batched` case and can can remove batch related logic.

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
```diff
@@ -314,3 +358,139 @@ def apply(
...
+            scatter_indx=scatter_indx,
+            precision_config=self.quant_config.w2_precision,
+            gammas=None if apply_router_weight_on_input else gammas,
+            y=intermediate_cache3,
+        )
+
+        self.moe_sum(intermediate_cache3.view(-1, topk, K), output)
+
+        # Set the original n_expts_act back
+        routing_data.n_expts_act = n_expts_act
```
varun-sundar-rabindranath: is resetting `routing_data.n_expts_act` required ? it looks like we are done using `routing_data` at this point?

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
```diff
@@ -314,3 +358,139 @@ def apply(
...
+        assert (
+            self.quant_config.w2_bias is None
+            or self.quant_config.w2_bias.dtype == torch.float32
+        )
+
+        # Shape check, only check non-mxfp4
+        assert hidden_states.shape[-1] == w1.shape[-2]
+        assert w2.shape[-1] == w1.shape[1]
+
+        batch_dim = hidden_states.shape[0] if hidden_states.ndim == 3 else 1
```
varun-sundar-rabindranath: I think it is safer to assert `batch_dim` == 1 as the workspace allocation is only setup for batch_dim == 1.

vllm/model_executor/layers/quantization/mxfp4.py
```diff
@@ -850,6 +860,8 @@ def select_gemm_impl(
             elif self.mxfp4_backend == Mxfp4Backend.MARLIN:
                 return MarlinExperts(self.moe_quant_config)
             elif self.mxfp4_backend == Mxfp4Backend.TRITON:
+                if self.moe.is_lora_enabled:
+                    return UnfusedOAITritonExperts(self.moe_quant_config)
```
varun-sundar-rabindranath, line 873: It is better to call `get_mxfp4_backend_with_lora()`  in line 847 ? something like, 
```
if self.moe.is_lora_enabled:
      return get_mxfp4_backend_with_lora()
```
so all the selection logic for LoRA exists in one place. Thanks

vllm/model_executor/layers/quantization/mxfp4.py
```diff
@@ -82,8 +83,17 @@ def get_mxfp4_backend_with_lora() -> Mxfp4Backend:
...
-    logger.info_once("[get_mxfp4_backend_with_lora] Using Marlin backend")
-    return Mxfp4Backend.MARLIN
+    if (
+        envs.VLLM_MXFP4_USE_MARLIN
+        or current_platform.get_device_capability()[0] < 9
+        or not has_triton_kernels()
+        or not is_torch_equal_or_newer("2.8.0")
+    ):
+        logger.info_once("[get_mxfp4_backend_with_lora] Using Marlin backend")
+        return Mxfp4Backend.MARLIN
```
varun-sundar-rabindranath, line 97: This logic changed recently  - Please take a look at https://github.com/vllm-project/vllm/commit/c17610e2baf5e40b3b0638b272bfe7e04e471bfe .

vllm/model_executor/layers/quantization/mxfp4.py
varun-sundar-rabindranath: nvm - I see why it is required. Thanks 👍

tests/lora/test_gptoss_tp.py
xyang16: Yes. I see generated text become "SELECT!!!!!!!!". After debugging, I found it's related to CUDA graph.

CUDA graph memory is allocated from a pool and is reused, and output tensor is [weak_ref_tensors](https://github.com/vllm-project/vllm/blob/v0.11.2/vllm/compilation/cuda_graph.py#L186). When bucket size is large, a tensor is reused or overwritten, but the weak_ref was still pointing to it, resulting in nans in output.

After I change max_cudagraph_capture_size to a smaller number, the nan issue got resolved.

vllm/lora/layers/fused_moe.py
xyang16: Added assert

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
xyang16: I removed batch related logic

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
xyang16: Yes, I also thought about it. Removed. Thanks!

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
xyang16: Yes, added assert. Thanks.

vllm/model_executor/layers/quantization/mxfp4.py
xyang16: Updated. Thanks!

tests/lora/test_gptoss_tp.py
varun-sundar-rabindranath: Does this happen in `main` ? If the changes in the PR are triggering it, I think we should either, 
 - fix the root-cause of this issue before landing as I think, or
 - have a safer workaround fix for now to land this PR. 
 I believe people actively use the gpt-oss TP code-path and we wouldn't want to disrupt on `main`. Thanks.

jeejeelee: @varun-sundar-rabindranath  Makes sense

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
```diff
@@ -314,3 +353,139 @@ def apply(
         )
 
         output.copy_(experts_output, non_blocking=True)
+
+
+class UnfusedOAITritonExperts(BaseOAITritonExperts):
```
jeejeelee, line 382: It would be perfect if we could add  doc describing the use cases for `UnfusedOAITritonExperts`

vllm/lora/layers/fused_moe.py
```diff
@@ -112,15 +121,21 @@ def _inject_lora_into_fused_moe(self):
...
         )
+        if quant_config.use_mxfp4_w4a16:
+            assert isinstance(
+                m_fused_moe_fn.fused_experts, (MarlinExperts, UnfusedOAITritonExperts)
+            )
+        else:
+            assert isinstance(m_fused_moe_fn.fused_experts, TritonExperts)
```
jeejeelee: Let's remove this assert, some quantization methods also use `MarlinExperts`, for example
```suggestion

```

xyang16: Removed. Thanks for review!

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
xyang16: Added docstring. Please take a look. Thanks!

tests/lora/test_gptoss_tp.py
xyang16: Yes it happens in main.

varun-sundar-rabindranath: Thanks @xyang16  - I see the nightly failing on this test. In that case, I think we should revert the change and let it fail on the PR as well. 
I am trying to avoid a situation where this change "fixes" `main` and we assume the issue is resolved. I believe this needs to be investigated further. 
cc @dcmaddix  @robertgshaw2-redhat

vllm/lora/layers/fused_moe.py
varun-sundar-rabindranath: Thanks. Can you also assert for the  `TritonExperts` case.

varun-sundar-rabindranath: Sorry, I see @jeejeelee 's comment about this here https://github.com/vllm-project/vllm/pull/28971#discussion_r2560768746 - @jeejeelee do you think an else assert like,
```
           assert isinstance(
                m_fused_moe_fn.fused_experts, (MarlinExperts,TritonExperts)
            )
```
makes sense ?

jeejeelee: @varun-sundar-rabindranath  makes sense

tests/lora/test_gptoss_tp.py
jeejeelee: > Yes. I see generated text become "SELECT!!!!!!!!". After debugging, I found it's related to CUDA graph.
> 
> CUDA graph memory is allocated from a pool and is reused, and output tensor is [weak_ref_tensors](https://github.com/vllm-project/vllm/blob/v0.11.2/vllm/compilation/cuda_graph.py#L186). When bucket size is large, a tensor is reused or overwritten, but the weak_ref was still pointing to it, resulting in nans in output.
> 
> After I change max_cudagraph_capture_size to a smaller number, the nan issue got resolved.


Is the current gpt-oss LoRA test failure caused by this reason?

xyang16: > > Yes. I see generated text become "SELECT!!!!!!!!". After debugging, I found it's related to CUDA graph.
> > CUDA graph memory is allocated from a pool and is reused, and output tensor is [weak_ref_tensors](https://github.com/vllm-project/vllm/blob/v0.11.2/vllm/compilation/cuda_graph.py#L186). When bucket size is large, a tensor is reused or overwritten, but the weak_ref was still pointing to it, resulting in nans in output.
> > After I change max_cudagraph_capture_size to a smaller number, the nan issue got resolved.
> 
> Is the current gpt-oss LoRA test failure caused by this reason?

I just checked the current gpt-oss LoRA test failure. Looks like it is using Marlin kernel. It is generating some garbage output instead of sql, but not !!!. So it looks a little different.

xyang16:
> @xyang16 Could you please resolve conflicts?

@jeejeelee I have resolved the conflicts. Thanks a lot!

vllm/lora/layers/fused_moe.py
xyang16: I have added an else assert. Thanks!

tests/lora/test_gptoss_tp.py
xyang16: @varun-sundar-rabindranath @jeejeelee @robertgshaw2-redhat I built wheel before and after https://github.com/vllm-project/vllm/pull/25444. And found before the PR, there's no !!! in output. After the PR, I can see !!! in output.

I have created an issue for this https://github.com/vllm-project/vllm/issues/29539

jeejeelee: @xyang16 your mean  the `Marlin kernel` can generate reasonable result?

xyang16:
> @xyang16 All tests pass correctly, but I realized that the GPT-OSS LoRA test does not select the Triton backend. Could you verify whether the Triton backend can pass correctly?

@jeejeelee Thanks a lot! Triton backend can pass with slight modification of `test_gpt_oss_lora_tp2` like below. Because FULL_AND_PIECEWISE cudagraph mode is causing some !!! in output, you can see the details in https://github.com/vllm-project/vllm/issues/29539. But fixing that problem is beyond the scope of this PR.

```
        compilation_config=vllm.config.CompilationConfig(  # Avoid OOM
            cudagraph_mode=CUDAGraphMode.PIECEWISE,
            cudagraph_specialize_lora=False,
        ),
```

xyang16:
> @xyang16 Any plan to fix this issue, if we have ,I think we can land this PR now

@jeejeelee Yes, I think the issue should be fixed soon. I talked about this issue with @varun-sundar-rabindranath and he suggested me to create an issue for this, and see what other people think how to fix it. Maybe quickest is to not make [FULL_AND_PIECEWISE](https://github.com/vllm-project/vllm/blob/v0.11.2/vllm/config/compilation.py#L371) default (FULL_AND_PIECEWISE was made default recently starting from 0.11.0).

jeejeelee (approved):
Thank you for contribution

DarkLight1337:
Reverted by https://github.com/vllm-project/vllm/pull/29697, please redo

PR #25463, opened 2025-09-23
[Model] Add qwen3-guard model support with streaming input and output.

sighingnow:
FIX #31975

Issue #25066, opened 2025-09-17
[Feature]: Streaming multi-modal input/output


DarkLight1337:
This is a tracking issue for enabling streaming MM I/O.

## Outline
 
Streaming input:

- Support streaming multi-modal inputs at API level.
- Handle streaming inputs in the multi-modal processor.
- Define an interface for models to indicate support for streaming inputs.
- Update V1 model runner and scheduler to handle partial MM encoding requests (this is the hardest part IMO)

Streaming output:

- Implement `RequestOutputKind.DELTA` for multi-modal outputs in V1 output processor.
- Support streaming multi-modal outputs at API level.

## Notes

- Currently we are waiting for AWS's proposal
- Take inspiration from https://github.com/vllm-project/vllm/pull/16347?
- See also [#22695](https://github.com/vllm-project/vllm/issues/22695)

qandrew:
Hi! some quick questions
- is the preferred way to implement MM streaming via responsesAPI or the legacy chat competions API?

DarkLight1337:
I think we can enable it for both, the API additions should be easy compared to what needs to be changed in the core engine.

KevinZeng08:
> Also wondering — is there any plan or existing support for streaming text input, not just output?

https://github.com/vllm-project/vllm/pull/25463 implements streaming inputs for text, you can check it.

ywang96:
Hey everyone! We have some updates and please join our the channel #sig-streaming-input in our slack slack.vllm.ai for collaborations!

PR #28973, opened 2025-11-19
[Feature] add session based streaming input support to v1

joshuadeng:
## Purpose
Proposal for session-based streaming with sequenced updates with minimal changes to the core scheduler and engine interfaces.

## Latest design

`AsyncLLM.generate` can now optionally take an async generator that produces a stream of `StreamingInput` objects, instead of a single prompt:
```python
@dataclass
class StreamingInput:
    prompt: PromptType
    sampling_params: SamplingParams | None = None
```
The `SamplingParams` for each input can differ, if omitted will use those provided in the original `generate` call.

The `StreamingInput` chunks are handled internally as separate requests, where the prompt of each request is the cumulative concatenation of all input prompts so far + their corresponding output tokens - excluding the final sampled token from each request. All generated tokens are returned to the caller in the session's output stream.

So for streaming inputs `[A1, B1, C1]`, `[A2, B2]`, `[A3, B3]`:
1. First prompt `[A1, B1, C1]`, generates `[D1]`
2. Second prompt is `[A1, B1, C1, A2, B2]`, generates `[C2, D2, E2]`  (`D1` discarded)
3. Third prompt is `[A1, B1, C1, A2, B2, C2, D2]`, generates `[C3, D3]` (`E2` discarded)

Streamed output tokens would be `D1, C2, D2, E2, C3, D3`. Note that we expect to generalize/parameterize the behaviour w.r.t. which output tokens to retain in successive prompts as a follow-on based on use case requirements.

Inputs are considered completed for a particular session when the provided async generator exits or is closed/garbage collected.

It is not necessary (but also not required) to wait for all of the outputs corresponding to a particular input chunk prior to sending the next one, they are queued internally.

Certain input types/parameters are not yet supported with streaming input:
- pooling, `prompt_embeds`, `n > 1`, `output_kind == FINAL_ONLY`, and stop strings.

---

### Original PR summary (superseded)

design (please request access):
https://docs.google.com/document/d/16iE0pUsjdlfEdcghiCSLlicKFdvVtxn2aiic3kL9RVY/edit?usp=sharing

- Enable low‑latency, interactive workloads (e.g., ASR) where inputs arrive incrementally.
- Reuse vLLM’s request‑centric architecture and KV blocks, minimizing scheduling changes and avoiding re‑prefill.

### High‑level design
- A session is opened by the first request with `streaming_sequence_id = 0` for a given `request_id`.
- While the session is RUNNING, new updates are queued to the session's queue (`streaming_queue`).
- When a chunk finishes, the session transitions to `WAITING_FOR_STREAMING_REQ`:
- Queued updates are applied from `streaming_queue` in place and the session goes back to `WAITING` to decode the next chunk.
  - If the update sets `close_session = True`, we mark the session finished with `stop_reason = "close_session"` and free resources.
- To keep KV and multimodal alignment correct, each updated session request sends the full prompt history. The unscheduled output token from the previous step, will be pruned before appending new prompt tokens (preserves `num_new_tokens` correctness and KV alignment).

## Test Plan

`pytest tests/v1/streaming/test_streaming_scheduler.py`
`pytest tests/v1/streaming/test_streaming_async_llm.py`
`pytest tests/v1/streaming/test_streaming_gpu_model_runner.py`

## Test Result
they pass
<details>

</details>

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit e38b36d854490248995721493c6843ace1017bb9. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Enables multi-turn, session-based streaming with resumable updates and correct KV/MM alignment across the stack.
> 
> - Add `StreamingInput` and new APIs `AsyncLLM.generate_streaming` and `AsyncLLM.generate_from_stream`; update `generate`/`add_request` to support `resumable` and reuse request queues
> - Extend `Request` with `resumable` and `streaming_queue` plus `StreamingUpdate`; add `RequestStatus.WAITING_FOR_STREAMING_REQ`
> - Scheduler: queue streaming updates, `_update_request_as_session` (prunes last unscheduled output, merges mm offsets, updates tokens/params), propagate `resumable`, exclude waiting-for-streaming from unfinished counts, and build `NewRequestData` via `_make_new_request_data` with copied `prompt_token_ids` to avoid aliasing
> - GPUModelRunner: handle updated streaming sessions in `scheduled_new_reqs` via `_update_streaming_request` (remove from `InputBatch`, refresh fields, clear `output_token_ids`)
> - OutputProcessor: allow in-place streaming state updates, only free request when finished and not `resumable`
> - Minor: compute prompt/encoder properties via accessors; plumb `resumable` into `EngineCoreOutput`
> - Add targeted unit tests for AsyncLLM, Scheduler (including aliasing bug prevention), GPUModelRunner, and RequestStatus
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e38b36d854490248995721493c6843ace1017bb9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Introduces resumable, session-based streaming that queues incremental inputs and preserves KV/MM alignment end-to-end.
> 
> - Add `StreamingInput` and extend `AsyncLLM.generate` to accept async generators; plumb `resumable` through `add_request` and reuse per-request `RequestOutputCollector`
> - Extend `Request` with `resumable`, `streaming_queue`, and `StreamingUpdate`; add `RequestStatus.WAITING_FOR_STREAMING_REQ` and expose it in `__str__`
> - Scheduler: queue and apply streaming updates via `_update_request_as_session` (prunes last unscheduled output, merges MM offsets, updates tokens/params), propagate `resumable`, exclude waiting-for-streaming from unfinished counts, and build `NewRequestData` via `_make_new_request_data` using a copied `prompt_token_ids` to avoid aliasing
> - GPUModelRunner: handle updated streaming sessions in `scheduled_new_reqs` via `_update_streaming_request` (remove from `InputBatch`, refresh fields, clear `output_token_ids`)
> - OutputProcessor: support in-place streaming state updates and only free requests when finished and not `resumable`; propagate `resumable` in `EngineCoreOutput`
> - Minor: compute prompt/encoder properties via accessors; add unit tests covering AsyncLLM streaming, scheduler lifecycle (incl. aliasing bug prevention), GPUModelRunner updates, and new status
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit eba8018bf5ab15b2826c61324663a2e4cb2e66bd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


vllm/v1/engine/__init__.py
```diff
@@ -71,6 +71,7 @@ class EngineCoreRequest(
     priority: int = 0
 
     trace_headers: Mapping[str, str] | None = None
+    close_session: bool = True
```
patrickvonplaten:
```suggestion
    close_streaming_session: bool | None = None
```

Then maybe we could check that this can only be True/False for streaming requests but never for normal requests where it always has to stay None?

vllm/v1/engine/__init__.py
```diff
@@ -126,6 +127,7 @@ class EngineCoreOutput(
     # The number of NaNs in logits.
     # A value greater than 0 indicates that the output is corrupted.
     num_nans_in_logits: int = 0
+    close_session: bool = True
```
patrickvonplaten: Why does the EngineCoreOutput need to have a close_session bool - is it not enough that the input has it? It's always the user that decides whether the session is closed or not?

vllm/v1/engine/async_llm.py
```diff
@@ -285,6 +285,7 @@ async def add_request(
         priority: int = 0,
         data_parallel_rank: int | None = None,
         prompt_text: str | None = None,
+        close_session: bool = True,
```
patrickvonplaten: why is close_session an arg here? It's part of `EngineCoreRequest` already no?

vllm/v1/engine/__init__.py
joshuadeng: it's needed in output_processor.py, to know if we need to close the streaming session for finished request

vllm/v1/engine/async_llm.py
joshuadeng: it's used in the else statement in line 307, if the input prompt isn't an EngineCoreRequest

vllm/v1/request.py
```diff
@@ -259,4 +275,5 @@ def get_finished_reason(status: "RequestStatus") -> FinishReason | None:
     RequestStatus.FINISHED_LENGTH_CAPPED: FinishReason.LENGTH,
     RequestStatus.FINISHED_ABORTED: FinishReason.ABORT,
     RequestStatus.FINISHED_IGNORED: FinishReason.LENGTH,
+    RequestStatus.WAITING_FOR_STREAMING_REQ: FinishReason.STOP,
```
zhuohan123, line 324: Why do we count `RequestStatus.WAITING_FOR_STREAMING_REQ` as a finished state?

vllm/v1/engine/output_processor.py
```diff
@@ -435,6 +439,19 @@ def add_request(
...
+    ) -> None:
+        req_state = self.request_states[request.request_id]
+        if req_state.prompt and prompt:
+            req_state.prompt += prompt
+        if req_state.prompt_token_ids is not None and request.prompt_token_ids:
+            req_state.prompt_token_ids.extend(request.prompt_token_ids)
+        req_state.prompt_embeds = request.prompt_embeds
+        if req_state.stats is not None:
+            req_state.stats.arrival_time = request.arrival_time
+        req_state.is_prefilling = True
```
zhuohan123: We need to raise error if we are updating a non-streaming/`close_session=True` request.

vllm/v1/engine/processor.py
```diff
@@ -394,7 +394,26 @@ def process_inputs(
...
+                pooling_params=params if isinstance(params, PoolingParams) else None,
+                eos_token_id=None,
+                arrival_time=arrival_time if arrival_time else time.time(),
+                lora_request=lora_request,
+                cache_salt=None,
+                priority=priority,
+                data_parallel_rank=data_parallel_rank,
+                close_streaming_session=True,
+            )
+
```
zhuohan123: For non-streaming requests, do we default `close_streaming_session` to `None` or `True`?  I personally prefer `True` and not distinguishing between streaming and non-streaming requests. We can view a non-streaming request as a request that a user has finished their input in the first round.

vllm/v1/engine/async_llm.py
```diff
@@ -439,17 +447,37 @@ async def generate(
...
 
+        except GeneratorExit:
+            if close_streaming_session is not None:
+                if self.log_requests:
+                    logger.info(
+                        "Request %s generator completed, session remains alive.",
+                        request_id,
+                    )
+                # For streaming sessions, generator completion is normal
+                return
```
zhuohan123: Can you elaborate a bit for the behavior here?

vllm/v1/core/sched/scheduler.py
```diff
@@ -1207,6 +1248,36 @@ def _update_request_with_output(
...
+    ) -> None:
+        pass
+
+    def _handle_finished(
+        self,
+        finished_req_ids: set[str],
+        outputs: dict[int, list[EngineCoreOutput]],
+    ) -> None:
+        pass
+
```
zhuohan123: what are these functions for?

vllm/v1/engine/__init__.py
```diff
@@ -71,6 +71,7 @@ class EngineCoreRequest(
     priority: int = 0
 
     trace_headers: Mapping[str, str] | None = None
+    close_streaming_session: bool | None = None
```
zhuohan123: Nit: We can maybe call this field something like `more_tokens_coming` (probably also not the best name but you got the point)? Then we can default this field to False for all requests but only set this to True for requests that have more streaming inputs coming.

vllm/v1/streaming/streaming_scheduler.py
```diff
@@ -0,0 +1,171 @@
...
+from typing import Any
+
+from vllm.multimodal.inputs import PlaceholderRange
+from vllm.v1.core.sched.output import NewRequestData
+from vllm.v1.core.sched.scheduler import Scheduler
+from vllm.v1.engine import EngineCoreEventType, EngineCoreOutput
+from vllm.v1.request import Request, RequestStatus
+
+
+class StreamingScheduler(Scheduler):
```
zhuohan123: What's the difficulty of merging StreamingScheduler and normal scheduler?

vllm/v1/engine/processor.py
joshuadeng: right now it's defaulted to None, per Patrick's comment, but i'll change it to True

vllm/v1/core/sched/scheduler.py
joshuadeng: for our internal use case of streaming, we want to be able to mutate request for sessions that are runnning still, and it's handled by overriding this function _handle_non_stopped

vllm/v1/request.py
joshuadeng: we need to provide a finish reason when creating EngineCoreOutput, otherwise it would be none. the client will expect a finish reason to know generation stopped (temporarily for streaming), so we can consume the output chunk, and it won't hang.

vllm/v1/streaming/streaming_scheduler.py
joshuadeng: - for stopped requests (normally we would finish and free request) for streaming we make status WAITING_FOR_STREAMING_REQ
- for streaming requests, we create NewRequestData with _all_token_ids (not just prompt), to provide all tokens for the prompt field to include all past inputs (including decoded outputs). Updated streaming requests will create new entries in InputBatch, so we need the full input history to ensure alignment

these are the main 2 issues, however I think if we leverage the `more_tokens_coming` field we can gate this logic from non streaming behavior and merge StreamingScheduler into normal scheduler

vllm/v1/streaming/streaming_scheduler.py
```diff
@@ -0,0 +1,171 @@
...
+        num_new_tokens = (
+            session_request.num_tokens - session_request.num_computed_tokens
+        )
+        assert num_new_tokens in {0, 1}
+        if num_new_tokens == 1:
+            assert (
+                session_request._all_token_ids[-1]
+                == session_request._output_token_ids[-1]
+            )
+            del session_request._all_token_ids[-1]
```
joshuadeng: I also want to highlight this behavior. This removes the last output token (which hasn't been scheduled) from `_all_token_ids`, as the new request's prompt tokens will replace it. The last output token (unscheduled) thus will not go in the kv cache as it will be replaced by the streaming update request's prompt, but will be returned to the client.

We needed this behavior for our internal use case, but we can handle it another way for oss, i.e. make the new scheduled tokens the last output token + new prompt tokens

vllm/v1/streaming/streaming_scheduler.py
```diff
@@ -0,0 +1,171 @@
...
+
+        if request.mm_features:
+            base = session_request.num_tokens
+            for mm_feature in request.mm_features:
+                mm_feature.mm_position = PlaceholderRange(
+                    offset=mm_feature.mm_position.offset + base,
+                    length=mm_feature.mm_position.length,
+                )
+            session_request.mm_features.extend(request.mm_features)
```
joshuadeng: also highlighting this behavior here, for multimodal this extends new mm features to the current session's mm features and offsets it by the previous num tokens to be in alignment.

we can also change this behavior to be replacing session's mm features (if old mm features are expected to be consumed) and if we need to change the alignment based on session's token history or keep unchanged for the new streaming request

vllm/v1/engine/async_llm.py
ErickLuo90: so this is needed because let's say for caller side we have this code
```
output_generator = await generate()
async for request_output in output_generator:
  client_handle_output

--> here triggers GeneratorExit
do_more_things
```

for non-streaming before, in GeneratorExit, it's fine to just do abort(request_id) which will free everything.
Now, for streaming session, we shouldn't abort(request_id) because client hasn't close session yet.

vllm/v1/core/sched/scheduler.py
ErickLuo90: I guess we can remove this for now to avoid confusion?

zhuohan123: yeah also I think this seems to be a bit model specific and it's weird to have that kind of logic in the shared scheduler.

vllm/v1/engine/processor.py
patrickvonplaten: The reason I think `None` is better is because it cleanly declares `close_streaming_session` as irrelevant for non-streaming use case (which it is I think). We can see non-streaming requests as "single-round" streaming requests, but I think we don't do this really at the moment (for example because we're abstracting the scheduler into a specialized streaming_scheduler design that should never be used for non-streaming requests)

So streaming requests have to go trough the new streaming_scheduler and are not allowed to call certain functionality from the non-streaming scheduler. By declaring the variable `close_streaming_session` "not used" with None I think we have an easy mechanism to verify that something doesn't accidentally ends up in the wrong path.

For example we can nicely assert everywhere in `streaming_scheduler` that `close_streaming_session` is not None (similar for functions that are used for both streaming and non-streaming `assert close_streaming_session is None` can be useful for statements / functions that are not relevant for streaming.

But obviously no strong opinion and I'm not super familiar with general design logic of vllm.

patrickvonplaten: Ah I see now that you also proposed to merge the streaming_scheduler here: https://github.com/vllm-project/vllm/pull/28973/files#r2579527152 => if we're able to merge everything into one then yeah agree

vllm/v1/streaming/streaming_scheduler.py
```diff
@@ -0,0 +1,171 @@
...
+        updating session requests, we need to provide all tokens for the prompt field
+        to include all past inputs, including decoded outputs. New and updated session
+        requests create new entries in InputBatch, so we need the full input history to
+        ensure alignment of mm offsets, kv cache, and token ids.
+
+        Make sure that prompt_token_ids is a copy of the original request's
+        _all_token_ids. Since the scheduler updates _all_token_ids each iteration, when
+        continuing decoding for a text or special token an updated NewRequestData won't
+        be sent from the scheduler, the corresponding prompt_token_ids reference in
+        SpeechGPUModelRunner will be mistakenly updated.
```
patrickvonplaten: What's `SpeechGPUModelRunner`?

vllm/v1/engine/__init__.py
patrickvonplaten: Should we maybe re-use `resumable` that was used in another PR: https://github.com/vllm-project/vllm/pull/25463 and which I think is quite intuitive

vllm/v1/request.py
```diff
@@ -132,6 +129,10 @@ def __init__(
 
         self.skip_reading_prefix_cache = self.get_skip_reading_prefix_cache()
 
+        # Used for streaming
+        self.close_streaming_session = close_streaming_session
+        self.streaming_queue: deque[Request] = deque()
```
patrickvonplaten: let's maybe not init a streaming queue if it's not a streaming request no?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -994,6 +1001,39 @@ def _update_states_after_model_execute(
         for i, num_tokens in enumerate(num_accepted_tokens):
             self.input_batch.num_accepted_tokens_cpu[i] = num_tokens
 
+    def _update_streaming_request(
+        self, req_id: str, new_req_data: NewRequestData
+    ) -> CachedRequestState:
+        """Updates streaming session request from `scheduled_new_reqs`.
```
patrickvonplaten, line 1139: we should probably also throw here somehow if we're not in a streaming regime no?

vllm/v1/streaming/streaming_scheduler.py
joshuadeng: leftover from internal, fixed

vllm/v1/worker/gpu_model_runner.py
joshuadeng: this function is only called for streaming scenarios. the only way to check in the model runner is if for new scheduled requests, if the request already exists in model_runner.requests

vllm/v1/core/sched/scheduler.py
```diff
@@ -802,6 +812,75 @@ def _update_after_schedule(
...
+            del session._all_token_ids[-1]
+
+        session.continue_session = request.continue_session
+        if request.mm_features:
+            base = session.num_tokens
+            for mm_feature in request.mm_features:
+                mm_feature.mm_position = PlaceholderRange(
+                    offset=mm_feature.mm_position.offset + base,
+                    length=mm_feature.mm_position.length,
+                )
```
ywang96: IMO we should consider adding a class method for `PlaceholderRage` to update itself in place rather than creating a new one to overwrite it. WDYT?

This can also be a TODO for later

vllm/v1/core/sched/scheduler.py
```diff
@@ -802,6 +812,75 @@ def _update_after_schedule(
         # it will also affect the scheduler output.
         self.finished_req_ids = set()
 
+    def _update_session(self, session: Request) -> None:
```
ywang96: Nit
```suggestion
    def _update_request_as_session(self, session: Request) -> None:
```

vllm/v1/core/sched/scheduler.py
ywang96: IMO we should consider adding a class method for `PlaceholderRage` to update itself in place rather than creating a new one to overwrite it. WDYT?

This can also be a TODO for later

vllm/v1/core/sched/scheduler.py
```diff
@@ -445,6 +446,15 @@ def schedule(self) -> SchedulerOutput:
                         skipped_waiting_requests.prepend_request(request)
                         continue
 
+                # Streaming: skip request if still waiting for next streaming req.
+                if request.status == RequestStatus.WAITING_FOR_STREAMING_REQ:
+                    if request.streaming_queue:
+                        self._update_session(request)
```
ywang96: Nit: add a note here that this updates request status to be `WAITING`

vllm/v1/core/sched/scheduler.py
```diff
@@ -802,6 +812,75 @@ def _update_after_schedule(
...
+
+        session._all_token_ids.extend(request.prompt_token_ids or [])
+        if session.prompt_token_ids is None:
+            session.prompt_token_ids = []
+        session.prompt_token_ids.extend(request.prompt_token_ids or [])
+        session.prompt_embeds = request.prompt_embeds
+        session.max_tokens = session.num_output_tokens + request.max_tokens
+        session.arrival_time = request.arrival_time
+        session.priority = request.priority
+        session.sampling_params = request.sampling_params
```
ywang96: What would be the use case when these are different?

I'm wondering if we should simply assert them to be equal - allowing updates on priority or sampling parameters on the fly when streaming doesn't make too much sense to me but I could wrong here.

vllm/v1/core/sched/scheduler.py
```diff
@@ -1375,7 +1467,11 @@ def _free_blocks(self, request: Request):
         del self.requests[request.request_id]
 
     def get_num_unfinished_requests(self) -> int:
-        return len(self.waiting) + len(self.running)
+        num_waiting = sum(
+            req.status != RequestStatus.WAITING_FOR_STREAMING_REQ
+            for req in self.waiting
+        )
+        return num_waiting + len(self.running)
```
ywang96: Why do we exclude them from unfinished?

vllm/v1/engine/output_processor.py
```diff
@@ -438,6 +442,24 @@ def add_request(
         if parent_req:
             self.parent_requests[parent_req.request_id] = parent_req
 
+    def _update_streaming_request_state(
+        self, request: EngineCoreRequest, prompt: str | None
+    ) -> None:
+        if not EngineCoreRequest.continue_session:
```
ywang96: Typo?
```suggestion
        if not request.continue_session:
```

vllm/v1/core/sched/scheduler.py
```diff
@@ -802,6 +812,75 @@ def _update_after_schedule(
...
+                    offset=mm_feature.mm_position.offset + base,
+                    length=mm_feature.mm_position.length,
+                )
+            session.mm_features.extend(request.mm_features)
+
+        session._all_token_ids.extend(request.prompt_token_ids or [])
+        if session.prompt_token_ids is None:
+            session.prompt_token_ids = []
+        session.prompt_token_ids.extend(request.prompt_token_ids or [])
+        session.prompt_embeds = request.prompt_embeds
```
ywang96: Should the prompt_embeds be concatenated?

vllm/v1/engine/output_processor.py
```diff
@@ -438,6 +442,24 @@ def add_request(
...
+            raise ValueError(
+                f"Existing session {request.request_id} has to be updated with a \
+                streaming request with `continue_session=True`"
+            )
+        req_state = self.request_states[request.request_id]
+        if req_state.prompt and prompt:
+            req_state.prompt += prompt
+        if req_state.prompt_token_ids is not None and request.prompt_token_ids:
+            req_state.prompt_token_ids.extend(request.prompt_token_ids)
+        req_state.prompt_embeds = request.prompt_embeds
```
ywang96: Ditto

vllm/v1/core/sched/scheduler.py
```diff
@@ -802,6 +812,75 @@ def _update_after_schedule(
...
+        create new entries in InputBatch, so we need the full input history to
+        ensure alignment of mm offsets, kv cache, and token ids.
+
+        NOTE: Make sure that prompt_token_ids is a copy of the original request's
+        _all_token_ids. Since the scheduler updates _all_token_ids each iteration, the
+        corresponding prompt_token_ids reference in NewRequestData will be mistakenly
+        updated while decoding if we don't make a copy.
+        """
+        req_data = NewRequestData.from_request(request, block_ids, prefill_token_ids)
+        if request.continue_session:
```
ywang96: Should this be `request.streaming_queue is not None`? What happens if a session receives a closing streaming request?

ywang96 (commented):
Thanks for the contribution - I left a few comments and please take a look.

A few thoughts on my mind that are not reflected in comments but I think we should address (open to discussion):

1. Request validation: there should be a request validation that happens at `AsyncLLM.add_request` before we send them to EngineCore
2. The relationship between `continue_session` and `streaming_queue` is not very clearly reflected in the code logic/documentation (I also personally feel like `resumable` is probably a better name than `continue_session`)
3. (related to 1) In enginecore we have a lot of places where we implicitly assume that "if we have seen this request id before then it must be a streaming request" - I think this is slightly risky if we don't have some validation in place.
4. Providing a good usage example would be great!

Regarding ordering of streaming requests that belong to the same session: I think it's okay and natural to assume that they come in order. We can iterate on it if there's a strong need on user-specified streaming flow. We should also add some validation on arrival time.

vllm/v1/core/sched/scheduler.py
joshuadeng: I think sampling params should be able to be changed, i.e. adjusting max_tokens from request to request. priority we can assert the same

vllm/v1/core/sched/scheduler.py
joshuadeng: this allows the busy loop of engine core (https://github.com/vllm-project/vllm/blob/67312cad11835bd75ca55fda83708d4806b82436/vllm/v1/engine/core.py#L1252) to properly block instead of continue polling iirc. WAITING_FOR_STREAMING_REQ is closer to a "finished" status then a true WAITING one

cc @ErickLuo90

ErickLuo90: Yeah, this is related to the tight polling loop. Basically, request WAITING_FOR_STREAMING_REQ should be also considered as nothing to do. when we have nothing to process, we should do the block get in stead of get_nowait (which leads to the tight polling loop and weird stuck issues)
otherwise, we will see get_nowait keep getting called with wasted cycles (tight pooling loop) when _process_input_queue -> self.scheduler.has_requests() is called and latency would be pretty bad

ywang96:
Hey @joshuadeng ! I've discussed with @zhuohan123 @WoosukKwon @ErickLuo90 and there are a few things to mention here

1. We shouldn't need to use `deque[Request]` if we're always going to overwrite the request data with the latest received streaming request - this means we should only need to keep the latest update needed. If we indeed want to keep all individual updates with `deque`, we should have a clear separation on what are truly stateful for a session and only keep a lightweight data structure for the changing ones instead of `Request` itself.

2. Currently this design means the server can be held hostage (i.e having KV cache not released) if the client never sends a terminate signal for a session. We should add a guard ourselves to terminate a session if it hasn't received an update in X time - we can leave it as a TODO in a follow-up.

3. There's [#27987](https://github.com/vllm-project/vllm/pull/27987) that changes some underlying assumption of `req_id` and we're discussing how these two PRs should coordinate.

vklimkov-nvidia:
Thanks @joshuadeng for this contribution, we were looking for smth similar for a bit. we have a draft of an implementation that aims at a similar thing [here](https://github.com/vklimkov-nvidia/vllm/tree/alm_api_v3). But I extremely happy to find such a feature on the way to main.

* Is the design doc somewhat outdated? there is no `streaming_sequence_id` or `StreamingRequest`

* I have some troubles to use it however. My usecase is a conversational agent, where need to feed some input and produce some output synchronously. In case of a simple LLM, that would be "generating a token, injecting a different token for a subsequent decoding step". the only way i managed to make it work is to set max_tokens=1 during sampling and run `add_request` in a loop:

```
from vllm import SamplingParams
from vllm.engine.arg_utils import AsyncEngineArgs
from vllm.v1.engine.async_llm import AsyncLLM

engine_args = AsyncEngineArgs(
    model="TinyLlama/TinyLlama-1.1B-Chat-v1.0",
    max_model_len=256,
    gpu_memory_utilization=0.8,
)
engine = AsyncLLM.from_engine_args(engine_args)

# Get tokenizer
tokenizer = engine.tokenizer

# Initial prompt and injection phrase
initial_prompt = "tell me about yourself."
inject_phrase = "i am a robot, beep boop."

prompt_tokens = tokenizer.encode(initial_prompt, add_special_tokens=False)
inject_tokens = tokenizer.encode(inject_phrase, add_special_tokens=False)

print(f"Initial prompt: '{initial_prompt}'")
print(f"Prompt tokens: {prompt_tokens}")
print(f"\nInjection phrase: '{inject_phrase}'")
print(f"Injection tokens ({len(inject_tokens)}): {inject_tokens}")

# max_tokens=1 means decode exactly 1 token, then pause for next streaming request
sampling_params = SamplingParams(max_tokens=1, temperature=0.7, top_p=0.9)

# Reset for new test
session_id = "streaming_session"

# First request - get the queue
queue = await engine.add_request(
    session_id,
    {"prompt_token_ids": prompt_tokens},
    params=sampling_params,
    resumable=True,
)

kv_context = prompt_tokens.copy()
decoded = []

# Iterate step by step
for step in range(len(inject_tokens) + 1):
    is_last = (step >= len(inject_tokens))
    
    # Wait for output from queue
    output = await queue.get()
    
    if output.finished:
        decoded_token = output.outputs[0].token_ids[-1]
        decoded_text = tokenizer.decode([decoded_token])
        decoded.append(decoded_token)
        
        print(f"Step {step}: decoded '{decoded_text}'")
        
        if not is_last:
            # Inject next token
            inject_token = inject_tokens[step]
            inject_text = tokenizer.decode([inject_token])
            print(f"  Injecting: '{inject_text}'")
            kv_context.append(inject_token)
            
            # Add continuation - update queue reference since add_request
            # may return a new queue if the previous state was cleaned up
            queue = await engine.add_request(
                session_id,
                {"prompt_token_ids": [inject_token]},
                params=sampling_params,
                resumable=(step < len(inject_tokens) - 1),
            )
        else:
            kv_context.append(decoded_token)
            print("  Session complete")

print(f"\nDecoded: '{tokenizer.decode(decoded)}'")
print(f"Context: '{tokenizer.decode(kv_context)}'")
```

But i suspect that is not a correct way to do it. I was trying to make it work with:

```
for output in engine.generate(...):
    engine.add_request(...)
```

but it seems tokens are not getting injected into decoding process.

* Looks like in `request.py` in method `from_engine_core_request`, the `resumable` is not getting forwarded to request. 

* in my implementation we prompted to introduce a new method "append_request". We thought it's less confusing when user is passing same `request_id`

* i benchmarked my implementation sync vs async exection. I noticed that for concurrency=1, the main scheduler loop may go to sleep waiting for messages in queue, which can be somewhat slow (acquiring lock takes time). For concurrency>1 this doesnt happen. I'd like to benchmark this implementation, but maybe let me now what is a proper way to run it.

thanks @ywang96 for pointing out this change. Looking forward to start using it!

vllm/v1/engine/output_processor.py
```diff
@@ -136,6 +133,12 @@ def __init__(
         self.stream_interval = stream_interval
         self.sent_tokens_offset = 0  # Offset of sent tokens
 
+    @property
+    def prompt_len(self) -> int:
+        return length_from_prompt_token_ids_or_embeds(
+            self.prompt_token_ids, self.prompt_embeds
+        )
```
njhill: We may want to check where/how this is accessed since it introduces some overhead to the normal case.

vllm/v1/core/sched/scheduler.py
```diff
@@ -806,6 +819,79 @@ def _update_after_schedule(
...
+        decoded outputs, through the prompt field. Updated streaming requests
+        create new entries in InputBatch, so we need the full input history to
+        ensure alignment of mm offsets, kv cache, and token ids.
+
+        NOTE: Make sure that prompt_token_ids is a copy of the original request's
+        _all_token_ids. Since the scheduler updates _all_token_ids each iteration, the
+        corresponding prompt_token_ids reference in NewRequestData will be mistakenly
+        updated while decoding if we don't make a copy.
+        """
+        req_data = NewRequestData.from_request(request, block_ids, prefill_token_ids)
```
njhill: IIUC here (or somewhere similar), in the case that `request.resumable` is `True`, we should replace `req_data.sampling_params` with a default `SamplingParams` object that has `max_tokens=1`, since we don't want any sampling/decoding to be done.

vllm/v1/core/sched/scheduler.py
```diff
@@ -1324,8 +1415,16 @@ def get_request_counts(self) -> tuple[int, int]:
         return len(self.running), len(self.waiting)
 
     def add_request(self, request: Request) -> None:
-        self.waiting.add_request(request)
-        self.requests[request.request_id] = request
+        if request.request_id not in self.requests:
```
njhill: I think we should also check `request.resumable` here.

vllm/v1/core/sched/scheduler.py
```diff
@@ -1324,8 +1415,16 @@ def get_request_counts(self) -> tuple[int, int]:
...
+            self.waiting.add_request(request)
+            self.requests[request.request_id] = request
+        else:
+            session = self.requests[request.request_id]
+            assert session.streaming_queue is not None, (
+                "streaming queue must be initialized for session"
+            )
+            session.streaming_queue.append(request)
```
njhill: As I think @ywang96 is suggesting, rather than a deque, it would be better to have a single "next request" field, and here rather than appending it should either set the next request, or if already set then merge this one into it (e.g. should be the full/combined prompt)

vllm/v1/request.py
```diff
@@ -190,6 +191,20 @@ def num_tokens_with_spec(self) -> int:
     def num_output_tokens(self) -> int:
         return len(self._output_token_ids)
 
+    @property
+    def num_prompt_tokens(self) -> int:
+        return length_from_prompt_token_ids_or_embeds(
+            self.prompt_token_ids, self.prompt_embeds
+        )
```
njhill: similar comment to the other one

vllm/v1/core/sched/scheduler.py
```diff
@@ -806,6 +819,79 @@ def _update_after_schedule(
...
+        outputs are scheduled as the next input in autoregressive decoding. When we
+        receive a new streaming request, the new prompt becomes our next input, so the
+        last output token is no longer needed and will not join the kv cache. This
+        ensures correct calculation of `num_new_tokens` in `schedule`.
+        """
+        assert session.streaming_queue is not None
+        request = session.streaming_queue.popleft()
+
+        num_new_tokens = session.num_tokens - session.num_computed_tokens
+        assert num_new_tokens in {0, 1}, f"got {num_new_tokens=}"
```
njhill: nit: tuple probably better than set here
```suggestion
        assert num_new_tokens in (0, 1), f"got {num_new_tokens=}"
```

vllm/v1/core/sched/scheduler.py
```diff
@@ -806,6 +819,79 @@ def _update_after_schedule(
...
+        session.resumable = request.resumable
+        if request.mm_features:
+            base = session.num_tokens
+            for mm_feature in request.mm_features:
+                mm_feature.mm_position = replace(
+                    mm_feature.mm_position, offset=mm_feature.mm_position.offset + base
+                )
+            session.mm_features.extend(request.mm_features)
+
+        session._all_token_ids.extend(request.prompt_token_ids or [])
```
njhill: nit:
```suggestion
        session._all_token_ids.extend(request.prompt_token_ids or ())
```

njhill (commented):
Thanks a lot for this @joshuadeng, sorry again for getting to it late.

A lot of the mechanics look good to me but feel we should change the AsyncLLM API semantics to be more like a single request, i.e. a single call to `generate()` but additionally passing an async generator, or possibly streaming additional inputs into the returned async generator before using it to stream the outputs.

In particular this would avoid the need for the "fake" stopping and related logic, and also avoid complications related to duplicate request ids.

But I also think that would be more idiomatic and would integrate more naturally with streaming front-ends such as SSE / web sockets and gRPC. It would also avoid the issue of partial requests being abandoned and leaving a bunch of kv cache reserved - when the output generator is closed (incl. going out of scope) the whole request would be automatically cleaned up.

One question is what the contents of the streamed messages should be. I'm a bit uneasy about having each be the full set of args that generate() currently accepts, since a subset are redundant (sampling params, request id, etc.) ... which is kind of messy / potentially confusing.

Other implementation related things:
- As mentioned inline, if I've understood correctly, we should be enforcing max_tokens=1 for the non-final sub requests or else we'll be decoding extra tokens prematurely which will get sandwiched between the input chunks.
- We need to verify this works properly with preemption and existing prefill chunking since these can happen any time.
- We'll also want to ensure that it works with async scheduling since we want to enable that by default imminently. Worst case we could require async scheduling to be disabled to use input streaming initially.

I don't see any major compatibility issues for the last two things but they might require some adjustments.

One more comment is that I think in some of the field/method names it would be better to qualify as "input_streaming" because currently when streaming is mentioned in the code it applies to streaming output tokens.

vllm/v1/core/sched/scheduler.py
```diff
@@ -1394,7 +1493,11 @@ def _free_blocks(self, request: Request):
         del self.requests[request.request_id]
 
     def get_num_unfinished_requests(self) -> int:
-        return len(self.waiting) + len(self.running)
+        num_waiting = sum(
+            req.status != RequestStatus.WAITING_FOR_STREAMING_REQ
+            for req in self.waiting
+        )
```
njhill: For performance reasons it might be better to maintain a separate count for these. This could become expensive if the waiting queue is very long (even when folks aren't using input streaming)

vllm/v1/engine/async_llm.py
njhill: This change wouldn't be needed with the alternative proposed API

vllm/v1/core/sched/scheduler.py
```diff
@@ -1128,7 +1214,11 @@ def update_from_output(
                 stopped = check_stop(request, self.max_model_len, pooler_output)
 
             if stopped:
-                kv_transfer_params = self._free_request(request)
+                if request.resumable:
+                    request.status = RequestStatus.WAITING_FOR_STREAMING_REQ
+                    self.waiting.add_request(request)
```
njhill: Here I think it might make sense to check whether there is another request in the streaming queue and "update the session" here if so (leaving the request state as `WAITING` rather than `WAITING_FOR_STREAMING_REQ`).

This might avoid the performance issue mentioned by @vklimkov-nvidia above:
> I noticed that for concurrency=1, the main scheduler loop may go to sleep waiting for messages in queue, which can be somewhat slow (acquiring lock takes time).

vllm/v1/core/sched/scheduler.py
joshuadeng: for the last request, however,  resumable will be false. though we should check if the session request has a next request queue

joshuadeng:
> * As mentioned inline, if I've understood correctly, we should be enforcing max_tokens=1 for the non-final sub requests or else we'll be decoding extra tokens prematurely which will get sandwiched between the input chunks.

this depends on your requirements. we have some models that use streaming while also allow autoregressive decoding in between streaming request updates. also see: https://github.com/vllm-project/vllm/pull/28973#discussion_r2582483630

vllm/v1/core/sched/scheduler.py
joshuadeng: it depends on the use case, it should be allowed that autoreg decoding can happen between streaming updates

joshuadeng:
> If we are including decoded tokens from the intermediate chunks in subsequent prompts:
> 
> * Presumably those decoded tokens would also be returned as output?

yes

> * Would the caller consume these prior to sending the next input chunk?
> * Would the prompt for the second chunk be the concatenation of the two inputs or also include the decoded tokens in between?

2 & 3. see https://github.com/vllm-project/vllm/pull/28973#discussion_r2582483630, when we update with the next chunk, the behavior is to remove the last decoded token (would've been scheduled to decode) and use the new prompt tokens from the next streaming request. the previous last decoded token would not enter the kv cache. another alternative that we use internally (for a specific use case) is to keep the last decoded token and combine it with the prompt of the next streaming request as the new input.

njhill:
> 2 & 3. see [#28973 (comment)](https://github.com/vllm-project/vllm/pull/28973#discussion_r2582483630), when we update with the next chunk, the behavior is to remove the last decoded token (would've been scheduled to decode) and use the new prompt tokens from the next streaming request. the previous last decoded token would not enter the kv cache. another alternative that we use internally (for a specific use case) is to keep the last decoded token and combine it with the prompt of the next streaming request as the new input.

Thanks @joshuadeng. To understand better, does this example make sense:
```
input1: [P0, P1, P2] -> prompt [P0, P1, P2], decode [D0, D1, D2]
<- [D0]
<- [D1]
<- [D2]
input 2: [P3, P4] -> prompt [P0, P1, P2, D0, D1, P3, P4]. decode [D3, D4]     (discard last decoded token)
<- [D3]
<- [D4]
input 3: [P5, P6, P7] -> prompt [P0, P1, P2, D0, D1, P3, P4, D3, P5, P6, P7], ...    (discard last decoded token)
```

And in particular, are the discarded decode tokens still returned?
Would the caller always consume the D tokens before sending the next input, or could the sequence also be:
```
input1: [P0,P1,P2] ->
input 2: [P3, P4] ->
input 3: [P5, P6, P7] ->

 prompt [P0, P1, P2], decode [D0, D1, D2]
<- [D0]
<- [D1]
<- [D2]
  prompt [P0, P1, P2, D0, D1, P3, P4]. decode [D3, D4]     (discard last decoded token)
<- [D3]
<- [D4]
  prompt [P0, P1, P2, D0, D1, P3, P4, D3, P5, P6, P7], ...    (discard last decoded token)
```
In which case, does the caller need to know which decode tokens came after which inputs? (since the processing following each input could generate a variable/arbitrary number of tokens)

njhill:
Here's examples of the API designs I'm proposing:

1. This is probably best suited to the case that there's a distinct input/prefill and output phase, i.e. you don't consume output until the inputs have finished:
```python
generator = await async_llm.generate(resumable_input1)  # with resumable=True set

# Stream inputs
await generator.asend(resumable_input2)
await generator.asend(resumable_input3)
await generator.asend(final_input)  # with resumable=False set

# Stream outputs
async for output in generator:
    print("got output:", output)
```

2. This is a bit more general, allowing independent input and output streams (similar to async python gRPC API for bidi streaming rpcs).
```python

async def input_generator() -> AsyncGenerator:
     # this could be reading from a queue
     yield input1
     yield input2
     yield input3

generator = await async_llm.generate(input_stream=input_generator())

# inputs/outputs are independent streams that can be handled in separate async tasks

# Stream outputs
async for output in generator:
    print("got output:", output)
```

I've been deliberately vague about the actual input object type. It would "work" to use the existing `EngineCoreRequest` for this but I don't think that's a good idea because there are some args that are request/session scoped and others that are input chunk scoped and it's confusing/messy to have the the request-scoped args redundantly duplicated in every chunk - what if different values are sent, etc.

Note that there's no request_id involved here.

But I think for both of these approaches the existing `generate()` method can easily be generalized to support streaming input while still working as it does now for non-streaming input.. There would be a single call to `generate()` whether or not it's a streaming input request.

njhill:
Thanks @joshuadeng @ErickLuo90.

My feeling is that we should get the API design right here and not hack the current one to support it, mainly because we are trying to move away from accumulating this kind of technical debt, and especially given it's an external API which will then become harder to change later.

> Yes, internally, we have something similar to what you described. What we do is that instead of having current generate function that returns the output generator, we expose RequestOutputCollector directly

`RequestOutputCollector` relates to the outputs so I'm not sure how relevant that is? And it had really been intended as an internal class (though it's returned from `add_request`, the expectation is that folks use `generate()` directly in the vast majority of cases ... the public interface boundary is definitely something we need to formalize better though).

Really though if you prefer to consume outputs via the `RequestOuputCollector` queue-equivalent interface you can just call `add_request()` rather than `generate()`, the args are the same. I am seeing now there's a small amount of additional logic in `generate()` which I think would make sense to push into `add_request()`.

I feel bad about holding this up but I do think we could implement the proposed design quickly. IMO we should be very clear on the details/scope of the streaming input use cases that we intend to cover though since it may have implications for this API design. i.e. get precise about the different ways we'll support mapping the external input/output streams to the actual token sequences that end up getting processed and which decoded tokens are returned. Even if we don't support some of these now, we should consider their implications to the API.

> so, the caller doesn't need to know the exact decode tokens that come after each input, you could send all 3 requests at the same time and each new request would be processed after the previous request's decodes finished. this is just the default design, but you can extend this to have decoding stop (max_tokens=1) after each request and use the next streaming request prompt. Also, if you wanted to keep last decoded token, we can remove the discard last token logic and on streaming request update the new scheduled tokens would be last output token + new prompt tokens, and the previous output token would be put in kv cache.

@joshuadeng these variations make sense to me I think but are they exhaustive? Does this cover everything we had in mind in https://github.com/vllm-project/vllm/issues/25066 @ywang96 @DarkLight1337?

I also realized that we might be able to exploit this in our beam search impl.

vllm/v1/engine/output_processor.py
njhill: I think this could addressed in a simple way by keeping the prompt len field and having it be updated when the prompt_token_ids / prompt_embeds are updated

vllm/v1/core/sched/scheduler.py
njhill: I guess this suggestion isn't applicable now - it was based on my original incorrect assumption that we were just streaming prefill tokens rather than a sequence of independent requests.

vllm/v1/engine/output_processor.py
```diff
@@ -561,7 +583,10 @@ def process_outputs(
                     request_outputs.append(request_output)
 
             # Free completed requests.
-            if finish_reason is not None:
+            if (
+                engine_core_output.finish_reason is not None
```
njhill: This change will break the existing behavior. Here, `engine_core_output.finish_reason` is not the same as `finish_reason`. The latter may be set in the case that we detect a stop string (checking is done in this process rather than the engine core process), in which case we send an abort to the engine below. (the `if not_engine_core_output.finished`).

I think this will need some kind of different handling in the streaming input case. If one input "sub-request" completes with a stop string, we want to stop that sub-request from generating further tokens without necessarily aborting the "session".

Because of the above behavior, it's possible to get additional outputs here for requests that have already finished. We currently rely on the fact that we remove the request state and so discard any such requests. However, with the current changes in this PR that will no longer be the case. Such requests will continue generating and returning tokens until they stop for some other reason (like max length).

vllm/v1/engine/output_processor.py
```diff
@@ -459,7 +462,8 @@ def add_request(
     ) -> None:
         request_id = request.request_id
         if request_id in self.request_states:
-            raise ValueError(f"Request id {request_id} already running.")
+            self._update_streaming_request_state(request, prompt)
```
njhill: I realized this will be a problem if multiple inputs are sent while prior inputs are still in progress. We update the request state immediately based on the latest received input, but subsequent outputs might still correspond to prior input states (shorter prompt etc.). As implemented currently, these outputs will incorrectly contain state associated with the latest input (like the new concatenated prompt etc.)

njhill (changes requested):
> hi @njhill, just added new streaming apis you proposed to async_llm.py, feel free to take a look and let me know what you think!

Thanks @joshuadeng!  Actually I what I had in mind was generalizing the existing `generate()` method rather than adding new input-streaming-specific methods. And my two suggestions were alternatives, I think we should have one or the other rather than both. I am working on a change to illustrate what I mean. I do not think we should have a `resumable` arg in the public `AsyncLLM` methods.

> I am seeing now there's a small amount of additional logic in `generate()` which I think would make sense to push into `add_request()`.

Just an FYI that this has now been addressed in https://github.com/vllm-project/vllm/pull/31510.

vllm/v1/engine/async_llm.py
```diff
@@ -51,6 +53,24 @@
...
+    """
+
+    prompt: EngineCoreRequest | PromptType
+    sampling_params: SamplingParams
+    resumable: bool = True
+    lora_request: LoRARequest | None = None
+    tokenization_kwargs: dict[str, Any] | None = None
+    trace_headers: Mapping[str, str] | None = None
+    priority: int = 0
+    data_parallel_rank: int | None = None
```
njhill: I think the per-input fields could be cut down to just `prompt`, `sampling_params`. It should not be hard to add more later but it is difficult to remove them.

We should avoid the resumable field here if possible since completion can be cleanly signaled by the generator finishing, and for robustness we need to handle that case anyhow.

vllm/v1/engine/async_llm.py
```diff
@@ -472,6 +516,301 @@ async def generate(
...
+                await self.abort(q.request_id, internal=True)
+            if self.log_requests:
+                logger.info("Streaming request %s failed.", request_id)
+            raise EngineGenerateError() from e
+
+    async def generate_from_stream(
+        self,
+        input_stream: AsyncGenerator[StreamingInput, None],
+        request_id: str,
+    ) -> AsyncGenerator[RequestOutput, None]:
```
njhill: Given that input and output messages can be arbitrarily interleaved, I think this is the better option of the two, but as mentioned I think we should extend the existing generate method so that the prompt is of type
```python
EngineCoreRequest | PromptType | AsyncGenerator[StreamingInput, None]
```
and actually implement this within `add_request` (so that `AsyncLLM.add_request` is still only called once, and doesn't require a `resumable` arg itself.

vllm/v1/engine/async_llm.py
ErickLuo90: just to confirm, you are suggesting to remove resumable from the generate side (StreamingInput), right? resumable is still needed on request level for the engine

vllm/v1/engine/output_processor.py
joshuadeng: for model correctness this case won't affect correctness bc the model will be decoding and the output processor state won't affect this. the stats using req_state.prompt_len could be wrong, and RequestOutput would have an accumulated prompt. if this isn't ok, we'd need to add a queue for pending updates that we can apply once prior inputs finish, which we could add in a follow up pr (or here if we prefer)

vllm/v1/engine/output_processor.py
joshuadeng: I think changing from engine_core_output.finish_reason to just finish_reason should solve this. finish_reason vs engine_core_output.finish_reason captures the stop string finish as well and we have the resumable check to know whether to fully free the request state and possibly abort, versus just to no-op (waiting for next streaming update)

njhill:
Thanks @joshuadeng for all of the updates. I've some more in another [branch](https://github.com/njhill/vllm/commit/eded02a98054d72bdcb54572507769a3bf6718cd#diff-dfd61c34d016e1134939e443b94b5a49e0a5ace1679a027a3ec1f3e4e2073dd1) based on top of this one, which should address the remaining concerns I had. Here is a summary:
- Pushed down the streaming input logic into the `AsyncLLM.add_request` method, so that generate itself remains largely unchanged.
- Removed the `resumable` flag completed from the _external_ API, input stream completion is now communicated by just finishing/closing the input generator, including if it's garbage collected.
- To handle outputs having inconsistent metadata I've introduced a queue in the output processor too, so that the updates happen when each sub request completed.
- To address the issue related to stop checking happening in the output processor, I've blocked the use of stop strings with streaming input requests for now.
- Also blocked for n > 1 case, pooling models and `output_kind == FINAL_ONLY` since they will require some additional consideration I think.
- Removed the `num_prompt_tokens` `@property` methods (update the field when num tokens change rather than having the logic in the getter).

Other notes:
- I haven't tested this yet (!) nor updated your tests.
- I haven't looked closely at all the metrics/tracing aspects. This isn't a blocker from my pov but we may want to make sure various request-level metrics make sense and are recorded in a coherent way for streaming input reqs.
- I think the current logic is sound w.r.t. preemption, but not optimal in that we won't preempt streaming-input requests in `WAITING_FOR_STREAMING_REQ` state. These should probably be high on the preemption list given they are (temporarily) idle. We may also want different resumption logic so that they can be re-prefilled if kvcache space frees up, even if the next input chunk still hasn't arrived. In any case I don't consider that a blocker either.

Please take a look and let me know what you think, in the meantime I'll try to test too.

njhill (approved):
@joshuadeng @patrickvonplaten @zhuohan123 @ErickLuo90 @ywang96 thanks for all of the work / input / reviews of this! And thanks for the patience while iterating.

I've updated the PR summary [above](https://github.com/vllm-project/vllm/pull/28973#issue-3640559452) with the final design (though we expect there will be follow-on changes, especially w.r.t. handling of generated tokens).

PR #28979, opened 2025-11-19
[ROCm][CI] Fix entrypoints tests and Python-only installation test on ROCm

AndreasKaratzas:
This PR fixes ROCm CI failures by:

1. **Setting correct attention backends for entrypoints tests on ROCm**
   - Encoder-only models (embeddings/pooling) require `FLEX_ATTENTION` backend as it's the only attention backend that supports encoder-only self-attention on ROCm
   - Audio transcription/translation tests require `ROCM_AITER_FA` backend

2. **Fixing Python-only installation test on ROCm**

## Test Plan

ROCm CI tests should pass after this PR.


docker/Dockerfile.rocm_base
```diff
@@ -66,20 +69,30 @@ RUN mkdir -p /app/install && cp /opt/rocm/share/amd_smi/dist/*.whl /app/install
...
 ARG PYTORCH_VISION_BRANCH
+ARG PYTORCH_AUDIO_BRANCH
 ARG PYTORCH_REPO
 ARG PYTORCH_VISION_REPO
-RUN git clone ${PYTORCH_REPO} pytorch
-RUN cd pytorch && git checkout ${PYTORCH_BRANCH} && \
-    pip install -r requirements.txt && git submodule update --init --recursive \
+ARG PYTORCH_AUDIO_REPO
+
+RUN git clone ${PYTORCH_REPO} pytorch && cd pytorch \
```
gshtras: Please don't combine these steps
It won't affect the final docker size because of the multi stage build process.
But this allows to cache the git clone stage when you're debugging  torch build issues

AndreasKaratzas: Thanks for pointing this out. Let me know if the modifications are now inline with the above feedback.

docker/Dockerfile.rocm
```diff
@@ -67,6 +67,7 @@ COPY --from=build_vllm ${COMMON_WORKDIR}/vllm/.buildkite /.buildkite
 # -----------------------
 # Test vLLM image
 FROM base AS test
+ARG PYTHON_VERSION=3.12
```
gshtras: This is being set in the base docker. A better way would be to set it as an env there and inherit in this image

docker/Dockerfile.rocm
```diff
@@ -86,10 +87,22 @@ COPY --from=build_vllm ${COMMON_WORKDIR}/vllm /vllm-workspace
...
     && python3 -m pip install -e tests/vllm_test_utils \
     && python3 -m pip install pytest-shard
 
+# enable fast downloads from hf (for testing)
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv pip install --system hf_transfer
+ENV HF_HUB_ENABLE_HF_TRANSFER=1
+
+# Copy in the v1 package for testing (it isn't distributed yet)
+COPY vllm/v1 /usr/local/lib/python${PYTHON_VERSION}/dist-packages/vllm/v1
```
gshtras: Looks kinda hacky, why do we need it for tests, but not for the normal distribution?

tests/entrypoints/openai/test_translation_validation.py
```diff
@@ -18,6 +19,26 @@
 SERVER_ARGS = ["--enforce-eager"]
 
 
+@pytest.fixture(scope="module", autouse=True)
```
rasmith: Could this be put in conftest.py?

AndreasKaratzas: Thanks for the review. I moved this into `conftest`. Let me know if it looks better now.

AndreasKaratzas:
Added support for `Entrypoints Integration Test (Pooling)` test group on ROCm.

AndreasKaratzas:
Added FLEX_ATTENTION back since it was an integral part of some encoder only self attention tests under entrypoints test group. Also, updated version of terratorch to 1.1.1 because current version gave error:
```log
FAILED entrypoints/openai/test_vision_embeds.py::test_single_request[ibm-nasa-geospatial/Prithvi-EO-2.0-300M-TL-Sen1Floods11]
```

<details>
<summary>Click to expand full log</summary>
```log
FAILED entrypoints/openai/test_vision_embeds.py::test_single_request[ibm-nasa-geospatial/Prithvi-EO-2.0-300M-TL-Sen1Floods11]

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
self = _ModelRegistry(models={'AfmoeForCausalLM': _LazyRegisteredModel(module_name='vllm.model_executor.models.afmoe', class_...(module_name='vllm.model_executor.models.transformers', class_name='TransformersMultiModalForSequenceClassification')})
architectures = ['Terratorch']
model_config = ModelConfig(model='ibm-nasa-geospatial/Prithvi-EO-2.0-300M-TL-Sen1Floods11', runner='pooling', convert='auto', task=No...its_processors=None, io_processor_plugin=None, pooler_config=None, override_pooler_config=None, multimodal_config=None)
    def is_text_generation_model(
        self,
        architectures: str | list[str],
        model_config: ModelConfig,
    ) -> bool:
        model_cls, _ = self.inspect_model_cls(architectures, model_config)
>       return model_cls.is_text_generation_model
E       AttributeError: 'NoneType' object has no attribute 'is_text_generation_model'
```

</details>

vllm/platforms/rocm.py
```diff
@@ -384,7 +390,17 @@ def check_and_update_config(cls, vllm_config: "VllmConfig") -> None:
         use_aiter_rms_norm = rocm_aiter_ops.is_rmsnorm_enabled()
```
AndreasKaratzas, line 430: Added also logic to auto set block size to 64 on ROCm if attention backend is AITER Unified Attention since it yields significant performance gains.

vllm/v1/attention/backends/flex_attention.py
```diff
@@ -927,7 +927,23 @@ def ensure_divisible(candidate: int, block_size: int) -> int:
...
             device_props = torch.cuda.get_device_properties()
-            max_shared_memory = device_props.shared_memory_per_block_optin
+            # ROCm doesn't expose shared_memory_per_block_optin attribute
+            # AMD GPUs typically have 64KB LDS (Local Data Share) per workgroup
+            if hasattr(device_props, "shared_memory_per_block_optin"):
+                max_shared_memory = device_props.shared_memory_per_block_optin
+            else:
+                # ROCm fallback: use 64KB (typical for MI300 series)
+                # Map GCN architecture to known LDS sizes
+                ROCM_LDS_SIZE_MAP = {
```
gshtras: Consider putting the fallback under if is_rocm

tests/entrypoints/openai/test_chat.py
```diff
@@ -267,10 +269,21 @@ async def test_single_chat_session(client: openai.AsyncOpenAI, model_name: str):
     assert len(chat_completion.choices) == 1
 
     choice = chat_completion.choices[0]
-    assert choice.finish_reason == "length"
-    assert chat_completion.usage == openai.types.CompletionUsage(
-        completion_tokens=10, prompt_tokens=37, total_tokens=47
-    )
+
+    if current_platform.is_rocm():
+        assert choice.finish_reason in ["length", "stop"]
```
rasmith: Might be good to add a reason into the assert

AndreasKaratzas: Not sure if I understand this. You mean a text message explaining the reason? Code is a bit self-explanatory here, that's why I didn't feel the need.

rasmith: When user runs it, the reason will be printed, instead of them having them to go look at the code.  Can be helpful.

AndreasKaratzas: Done :)

tests/entrypoints/openai/test_response_api_with_harmony.py
```diff
@@ -503,11 +503,15 @@ async def test_web_search(client: OpenAI, model_name: str):
...
+    # Code interpreter needs more time for container init + code execution
+    # Extend timeout especially for ROCm
+    from vllm.platforms import current_platform
+
+    timeout_value = client.timeout * 3 if current_platform.is_rocm() else client.timeout
+    client_with_timeout = client.with_options(timeout=timeout_value)
+
+    response = await client_with_timeout.responses.create(
         model=model_name,
-        # TODO: Ideally should be able to set max tool calls
```
tjtanaa, line 509: Is this resolved in this PR? I only saw fixes done to tests. If it is not, can you add back the comments?

tests/entrypoints/openai/test_serving_chat.py
```diff
@@ -407,12 +407,16 @@ async def _fake_process_inputs(
 @dataclass
 class MockEngine:
     model_config: MockModelConfig = field(default_factory=MockModelConfig)
-    input_processor: MagicMock = field(default_factory=MagicMock)
     io_processor: MagicMock = field(default_factory=MagicMock)
+    input_processor: MagicMock = field(default_factory=MagicMock)
```
tjtanaa, line 487: Let's revert this line order changes

tests/entrypoints/openai/test_serving_tokens.py
```diff
@@ -16,10 +16,18 @@
 
 
 def get_vocab_size(model_name):
+    # ROCm/AMD specific configuration:
+    # 1. Use float16 to ensure compatibility with MI3XX where
```
tjtanaa: @AndreasKaratzas 
Thanks for adding the description.
I noticed that `bfloat16` makes many tests flaky and not passing. Will AMD investigate why bfloat16 is causing large numerical variance? Because it could affect accuracy as almost all models out there are `bfloat16` in nature. Users in most time does not explicitly launch vllm server in `float16`, they will be using `auto` which uses the original dtype `bfloat16`.

AndreasKaratzas: @tjtanaa We will investigate this. There is an effort to use bfloat16 more.

tests/entrypoints/openai/test_serving_chat.py
AndreasKaratzas: Done :)

tests/entrypoints/openai/test_response_api_with_harmony.py
AndreasKaratzas: Nice catch, I brought the TODO comment back. Thanks :)

tjtanaa (approved):
LGTM

tests/entrypoints/openai/test_chat.py
```diff
@@ -250,11 +250,13 @@ async def test_more_than_one_prompt_logprobs_chat(
     [MODEL_NAME, "zephyr-lora"],
 )
 async def test_single_chat_session(client: openai.AsyncOpenAI, model_name: str):
+    # Finish reason may be "length" or "stop" on ROCm due to different tokenization
```
robertgshaw2-redhat: Why would tokenization be different in the platforms?

I would suggest that instead, we either change the prompt or change the max_completion_tokens to be something smaller. This is going to be a PITA to maintain

tests/entrypoints/openai/test_optional_middleware.py
```diff
@@ -20,6 +20,12 @@
 
 @pytest.fixture(scope="module")
 def server(request: pytest.FixtureRequest):
+    # ROCm SPECIFIC CONFIGURATION:
```
robertgshaw2-redhat: Isnt a better and more user friendly solution to this problem to make the default attn backend for this type of model be FLEX_ATTENTION?

tests/entrypoints/openai/test_response_api_with_harmony.py
```diff
@@ -504,7 +504,14 @@ async def test_web_search(client: OpenAI, model_name: str):
 @pytest.mark.asyncio
 @pytest.mark.parametrize("model_name", [MODEL_NAME])
 async def test_code_interpreter(client: OpenAI, model_name: str):
-    response = await client.responses.create(
+    # Code interpreter needs more time for container init + code execution
```
robertgshaw2-redhat: just make the timeout longer for all platforms

tests/entrypoints/openai/test_run_batch.py
```diff
@@ -43,6 +43,20 @@
 {"custom_id": "request-2", "method": "POST", "url": "/v1/chat/completions", "body": {"model": "Qwen/Qwen3-0.6B", "messages": [{"role": "system", "content": "You are a helpful assistant."},{"role": "user", "content": "What is the capital of France?"}]}}"""
 
 
+def _get_rocm_attention_args():
```
robertgshaw2-redhat: We should just make this the default on ROCm for encoder-decoder models

tests/entrypoints/openai/test_serving_chat.py
```diff
@@ -489,7 +489,11 @@ class MockEngine:
 
 
 async def _async_serving_chat_init():
-    engine = MockEngine()
+    engine = MagicMock()
```
robertgshaw2-redhat: why is this needed?

tests/entrypoints/openai/test_serving_tokens.py
```diff
@@ -16,10 +16,18 @@
...
+    # 2. Use an explicit non-zero seed (42) because seed=0 can
+    #       be interpreted inconsistently
+    #    (as 'default' vs 'explicit 0') on ROCm backends, affecting initialization.
+    from vllm.platforms import current_platform
+
     config = ModelConfig(
         model=model_name,
-        seed=0,
-        dtype="bfloat16",
+        seed=42 if current_platform.is_rocm else 0,
```
robertgshaw2-redhat: just use the same seed for both

tests/entrypoints/openai/test_serving_tokens.py
```diff
@@ -93,6 +104,17 @@ async def test_same_response_as_chat_completions(client, tokenizer, messages):
         add_generation_prompt=True,
         enable_thinking=False,  # default with Qwen3
     )
+
+    # ROCm SPECIFIC CONFIGURATION:
```
robertgshaw2-redhat: just pass a nonzero seed to both

tests/entrypoints/openai/test_serving_tokens.py
robertgshaw2-redhat: I don't think its a good idea to change the dtype in these tests

tests/entrypoints/openai/test_shutdown.py
```diff
@@ -9,10 +9,16 @@
...
 import pytest
 
+from vllm.platforms import current_platform
 from vllm.utils.network_utils import get_open_port
 
 MODEL_NAME = "hmellor/tiny-random-LlamaForCausalLM"
 
+# ROCm: GPU initialization takes longer, requiring extended timeouts.
+_IS_ROCM = current_platform.is_rocm()
+_SERVER_STARTUP_TIMEOUT = 120 if _IS_ROCM else 30
```
robertgshaw2-redhat: just use the same timeouts

tests/entrypoints/openai/test_translation_validation.py
```diff
@@ -49,8 +61,13 @@ async def test_non_asr_model(foscolo):
 
 
 @pytest.mark.asyncio
-async def test_basic_audio_with_lora(mary_had_lamb):
+async def test_basic_audio_with_lora(mary_had_lamb, rocm_aiter_fa_attention):
     """Ensure STT (translate) requests can pass LoRA through to generate."""
+    # ROCm SPECIFIC CONFIGURATION:
+    # To ensure the test passes on ROCm, we modify the max model length to 512.
```
robertgshaw2-redhat, line 67: why does changing max-model-len have an impact?

tests/entrypoints/openai/test_transcription_validation.py
```diff
@@ -44,8 +47,13 @@ async def test_basic_audio(mary_had_lamb, model_name):
 
 
 @pytest.mark.asyncio
-async def test_basic_audio_with_lora(mary_had_lamb):
+async def test_basic_audio_with_lora(mary_had_lamb, rocm_aiter_fa_attention):
     """Ensure STT (transcribe) requests can pass LoRA through to generate."""
+    # ROCm SPECIFIC CONFIGURATION:
+    # To ensure the test passes on ROCm, we modify the max model length to 512.
```
robertgshaw2-redhat, line 53: why does this make the tests pass?

robertgshaw2-redhat:
I have a couple pieces of feedback:
* Instead of having special cases for attention backend, why can't we make FLEX_ATTENTION the default on ROCm for encoder and cross attention?
* In all cases where we tweak timeouts, we can just tweak the timeouts for both backends. We dont need a special case for it

tests/entrypoints/openai/test_transcription_validation.py
AndreasKaratzas: Unfortunately, the other backends do not support that test. They trigger different errors each.

tests/entrypoints/openai/test_chat.py
AndreasKaratzas: I reduced the number of tokens to 5 as per your suggestion and simplified the test structure. Thank you for your feedback :)

tests/entrypoints/openai/test_optional_middleware.py
AndreasKaratzas: I removed the hard coded attn backend setting since attn dispatching has already been taken care of in: https://github.com/vllm-project/vllm/pull/30270.

tests/entrypoints/openai/test_response_api_with_harmony.py
AndreasKaratzas: Done :)

tests/entrypoints/openai/test_run_batch.py
AndreasKaratzas: I removed the hard coded attn backend setting since attn dispatching has already been taken care of in: https://github.com/vllm-project/vllm/pull/30270.

tests/entrypoints/openai/test_serving_chat.py
AndreasKaratzas: There might have been a bug in the past which rendered this code block legit, but I tested it now without this and the tests are passing, so I removed it. Good catch, thanks :)

tests/entrypoints/openai/test_serving_tokens.py
AndreasKaratzas: I actually found out that the failure in the test happens post-eos, so I created a logic to stick to the legit response part of the model for the test comparison. I reverted the seed and float logic.

tests/entrypoints/openai/test_serving_tokens.py
AndreasKaratzas: I reverted the seed and bfloat16 logic in favor of better comparison mechanism.

tests/entrypoints/openai/test_serving_tokens.py
AndreasKaratzas: I reverted the seed and bfloat16 logic in favor of better comparison mechanism.

tests/entrypoints/openai/test_shutdown.py
AndreasKaratzas: Done :)

tests/entrypoints/openai/test_translation_validation.py
AndreasKaratzas: Because we use SPDA which computes full matrix, and triggers memory error no ROCm with this large model len since it also compiles the model.

PR #28987, opened 2025-11-19
[Feat] Iteration-level profiling for Torch and CUDA profiler

benchislett:
## Purpose

This PR introduces `VLLM_PROFILER_DELAY_ITERS` and `VLLM_PROFILER_MAX_ITERS` that control the behaviour of the Torch and CUDA profilers. Respectively, they allow to offset the start of profiling and limit the total number of engine iterations that are profiled. 

This mostly serves to facilitate lightweight profiling of heavy workloads on large models: when trying to measure high-throughput traffic, we need to send a lot of requests to get the batch size sufficiently high to be realistic, and we need to run for some time in order to warm up the hardware and get into a steady state. This leads to extremely large profiling footprints which can take a long time to record, serialize, store on disk, and process. This PR enables us to capture a small slice of a long benchmarking run, making it easy to capture a precise engine time slice.

This change is compatible with both the Torch profiler and the CUDA profiler, with the exception that it is not currently integrated with the CPU-side AsyncLLM Torch profiling which remains on for the full duration by default. I added an opt-in variable to disable this for cases when we need a torch profile capturing small slice of a long run. 

In doing this, I refactored most of the torch profiling code out of `gpu_worker.py` and into `gpu_profiler.py` alongside the CUDA profiler. A common base class coordinates the delay and max-limit calculations, and is thoroughly unit-tested.

Also, I extended `annotate_profile` to work with the CUDA profiler, adding a handy NVTX range over each engine iteration.

**TODO**: update documentation

## Test Plan

The common logic introduced in this PR is unit tested. I manually checked that the integration works properly for both CUDA and Torch profiler. Here are screenshots of a 256-batch-size slice with a delay of 128 and a max of 4 iterations:

image

image

Unit test passes locally. **I need to make sure it gets run in the CI**, I think it should get covered by `tests/v1/worker` runs but should double-check.

### DIFF
--- a/vllm/profiler/gpu_profiler.py
+++ b/vllm/profiler/gpu_profiler.py
@@ -130,7 +130,7 @@ class WorkerProfiler(ABC):
 
     def annotate_context_manager(self, name: str):
         """Return a context manager to annotate profiler traces."""
-        raise nullcontext()
+        return nullcontext()
 
 
 class TorchProfilerWrapper(WorkerProfiler):
--- a/vllm/profiler/gpu_profiler.py
+++ b/vllm/profiler/gpu_profiler.py
@@ -1,37 +1,212 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 
+from abc import ABC, abstractmethod
+from contextlib import nullcontext
+
+import torch
+from typing_extensions import override
+
+import vllm.envs as envs
 from vllm.logger import init_logger
 
 logger = init_logger(__name__)
 
 
-class CudaProfilerWrapper:
+class WorkerProfiler(ABC):
     def __init__(self) -> None:
-        self._profiler_running = False
+        self._delay_iters = envs.VLLM_PROFILER_DELAY_ITERS
+        if self._delay_iters > 0:
+            logger.info_once(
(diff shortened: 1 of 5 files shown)


vllm/profiler/gpu_profiler.py
```diff
@@ -1,37 +1,212 @@
...
+            logger.debug(
+                "start_profile received when profiler is already active. "
+                "Ignoring request."
+            )
+            return
+        self._active = True
+        if self._delay_iters == 0:
+            self._call_start()
+
+    def step(self) -> None:
```
wangshangsam, line 80: We also need the ability to log the number of requests that got scheduled/executed per iteration and how many tokens per request.

benchislett: The profiling annotation will add `num_new` and `num_cached` request count information to the NVTX range for each iteration already, thanks to the changes to `annotate_profile`. As for the number of tokens, this should probably just be an NVTX mark or range inside the model runner. This can be added separately and is independent of this PR.

mgoin (approved):
I think this looks great! It makes the torch profiler feel better too, so I'm fairly confident this is a right step. If you can fast-follow a ProfilerConfig before the next release, that would be best so we can remove the new env vars without need for deprecation

PR #28988, opened 2025-11-19
[Feature][Attention][PCP] Support PCP (Prefill Context Parallel) with MLA

FENP:
## Purpose
Ref to issue https://github.com/vllm-project/vllm/issues/25749. Enable PCP for MLA models.
This PR mainly includes the following changes:

- Add PCP Manager to `vllm/v1/worker/cp_utils.py` for model runner.
- Modified `vllm/v1/worker/gpu_model_runner.py` for PCP splitting logic for tokens
- Modified `vllm/v1/attention/backends/mla/common.py` to adapt the MLA backend to PCP
- Add utility functions required by PCP to `vllm/v1/attention/backends/utils.py` and `vllm/attention/ops/common.py`
- Renamed variables and functions shared by both PCP and DCP

## Test Plan
```shell
vllm serve deepseek-ai/DeepSeek-V2-Lite-Chat --gpu-memory-utilization 0.9 --tensor-parallel-size 1 --prefill-context-parallel-size 2
```

## Test Result
- PCP1TP8 (Baseline)
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.6277|±  |0.0094|
|     |       |strict-match    |     5|exact_match|↑  |0.6179|±  |0.0095|
```
- PCP2TP4
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.6259|±  |0.0094|
|     |       |strict-match    |     5|exact_match|↑  |0.6164|±  |0.0095|
```
- PCP4TP2
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.6308|±  |0.0094|
|     |       |strict-match    |     5|exact_match|↑  |0.6236|±  |0.0094|
```
- PCP8TP1
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.6240|±  |0.0094|
|     |       |strict-match    |     5|exact_match|↑  |0.6168|±  |0.0095|
```

## Benchmark
In addition to reducing GPU memory redundancy and **increasing KV cache capacity**, PCP can also reduce the all-reduce communication overhead of `o_proj` and **lower TTFT**. We evaluated the performance of PCP and TP on DeepSeek-R1 and Kimi-K2 using 4K-length inputs on the H20-3e.

### DeepSeek-R1

```shell
vllm bench serve --backend vllm  --model deepseek-ai/DeepSeek-R1/  --endpoint /v1/completions   --dataset-name random   --random-input 4096   --random-output 1   --max-concurrency 1  --num-prompt 10 --ignore-eos --metric-percentiles "50,90,99"
```

| Parallel Config |P50 TTFT (ms)  |     P90 TTFT (ms)    |  P99 TTFT (ms)   |   Performance gain  |
|:-----:|:------:|:------:|:-----:|:----:|
| PCP1TP8 |    414  | 417 |  418   | --- |
|   PCP2TP4  |   400  |  403  |   404   |  3.2%  |
|   PCP4TP2  |   392    |  394   |   395  |  5.3%  |

### Kimi-K2

```shell
vllm bench serve --backend vllm  --model moonshotai/Kimi-K2-Instruct/  --endpoint /v1/completions   --dataset-name random   --random-input 4096   --random-output 1   --max-concurrency 1  --num-prompt 10 --ignore-eos --metric-percentiles "50,90,99"  --trust-remote-code
```

| Parallel Config |P50 TTFT (ms)  |     P90 TTFT (ms)    |  P99 TTFT (ms)   |   Performance gain  |
|:-----:|:------:|:------:|:-----:|:----:|
| PCP1TP8 |    394  |  396  |  396  | --- |
|   PCP2TP4  |  371   |  373  |   373   |  5.9%  |
|   PCP4TP2  |   351    |  353   |  354   |  10.9%  |

Of course, PCP additionally introduces communication overhead from KV all-gather and index select kernel launch overhead for restoring KV. Further tuning is still needed to improve performance.

- [x] Support piecewise graph
- [x] Support chunk prefill and prefix caching
- [x] Add ci test
- [x] Accuracy test
- [x] Benchmark

## Limitations
Although the current PCP logic is fully compatible with decoding, the lack of splitting of decode tokens means that every PCP rank holds the full set of decode tokens, leading to significant redundant communication and computation (include attention and MoE). We therefore recommend enabling PCP on P instances in a P/D-disaggregation case.

## Future work
These items will be tackled in follow-up PRs; community contributions are warmly welcomed.
- [ ] PCP support decode length > 1
- [ ] PCP support fullgraph
- [ ] PCP support others MLA prefill attention backends (currently only FlashAttention prefill is supported)
- [ ] Ring-CP style attention backend algorithm, ref RFC https://github.com/vllm-project/vllm/issues/26133.
- [ ] PCP support GQA
- [ ] PCP support P/D Disaggregation
- [ ] PCP suppots DSA
- [ ] PCP support distributed PCP (multi-machine)
- [ ] dynamic PCP？(https://arxiv.org/pdf/2510.10620)

---
<details>

</details>

### DIFF
--- a/vllm/attention/ops/common.py
+++ b/vllm/attention/ops/common.py
@@ -467,3 +467,220 @@ def unpack_seq_triton(
         out = out.reshape(output_shape)
 
     return out
+
+
+@triton.jit
+def _fused_pcp_qkv_select_kernel(
+    q_ptr,
+    q_stride_B,
+    q_stride_H,
+    k_ptr,
+    k_stride_B,
+    k_stride_H,
+    v_ptr,
+    v_stride_B,
+    v_stride_H,
+    query_start_ptr,
+    out_q_head_ptr,
+    out_q_tail_ptr,
+    out_k_head_ptr,
+    out_k_tail_ptr,
+    out_v_head_ptr,
+    out_v_tail_ptr,
+    pcp_world_size: tl.constexpr,
+    pcp_rank: tl.constexpr,
+    n_head: tl.constexpr,
+    q_head_dim: tl.constexpr,
+    k_head_dim: tl.constexpr,
+    v_head_dim: tl.constexpr,
+    SEQ_BLOCK_SIZE: tl.constexpr,
+    DIM_BLOCK_SIZE: tl.constexpr,
+):
+    req_id = tl.program_id(0) // (2 * pcp_world_size)
+    seq_block_id = tl.program_id(0) % (2 * pcp_world_size)
+    head_id = tl.program_id(1)
+    dim_block_id = tl.program_id(2)
+    dim_off = tl.arange(0, DIM_BLOCK_SIZE) + dim_block_id * DIM_BLOCK_SIZE
+
+    q_start_loc = tl.load(query_start_ptr + req_id)
+    q_end_loc = tl.load(query_start_ptr + req_id + 1)
+    q_select_len = (q_end_loc - q_start_loc) // 2
+
+    # Select Q
+    if seq_block_id < 2:
+        block_q_start_loc = q_start_loc + seq_block_id * q_select_len
+        out_ptr = out_q_head_ptr if seq_block_id == 0 else out_q_tail_ptr
+        for qi in range(tl.cdiv(q_select_len, SEQ_BLOCK_SIZE)):
+            q_offset = tl.arange(0, SEQ_BLOCK_SIZE) + qi * SEQ_BLOCK_SIZE
+            mask = (dim_off[None, :] < q_head_dim) & (q_offset[:, None] < q_select_len)
+            q_src_idx = block_q_start_loc + q_offset[:, None]
+            q_dst_idx = q_start_loc // 2 + q_offset[:, None]
+            q_val = tl.load(
+                q_ptr
+                + q_src_idx * q_stride_B
+                + head_id * q_stride_H
+                + dim_off[None, :],
+                mask=mask,
+            )
+            tl.store(
+                out_ptr
+                + q_dst_idx * n_head * q_head_dim
+                + head_id * q_head_dim
+                + dim_off[None, :],
+                q_val,
+                mask=mask,
+            )
+
+    # Select KV
+    kv_start_loc = q_start_loc * pcp_world_size
+    kv_select_len = q_select_len
+    k_d_mask = dim_off[None, :] < k_head_dim
+    v_d_mask = dim_off[None, :] < v_head_dim
+    block_src_kv_start_loc = kv_start_loc + seq_block_id * kv_select_len
+    block_dst_kv_head_start_loc = (
+        kv_start_loc // 2 // pcp_world_size * (pcp_rank + 1)
+        + seq_block_id * kv_select_len
+    )
+    block_dst_kv_tail_start_loc = (
+        kv_start_loc // 2 // pcp_world_size * (2 * pcp_world_size - pcp_rank)
+        + seq_block_id * kv_select_len
+    )
+    for ki in range(tl.cdiv(kv_select_len, SEQ_BLOCK_SIZE)):
+        kv_offset = tl.arange(0, SEQ_BLOCK_SIZE) + ki * SEQ_BLOCK_SIZE
+        kv_block_mask = kv_offset[:, None] < kv_select_len
+        kv_src_idx = block_src_kv_start_loc + kv_offset[:, None]
+        kv_dst_idx_head = block_dst_kv_head_start_loc + kv_offset[:, None]
+        kv_dst_idx_tail = block_dst_kv_tail_start_loc + kv_offset[:, None]
+        k_val = tl.load(
+            k_ptr + kv_src_idx * k_stride_B + head_id * k_stride_H + dim_off[None, :],
+            mask=k_d_mask & kv_block_mask,
+        )
+        v_val = tl.load(
+            v_ptr + kv_src_idx * v_stride_B + head_id * v_stride_H + dim_off[None, :],
+            mask=v_d_mask & kv_block_mask,
+        )
+        if seq_block_id < pcp_rank + 1:
+            tl.store(
+                out_k_head_ptr
+                + kv_dst_idx_head * n_head * k_head_dim
+                + head_id * k_head_dim
+                + dim_off[None, :],
+                k_val,
+                mask=k_d_mask & kv_block_mask,
+            )
+            tl.store(
+                out_v_head_ptr
+                + kv_dst_idx_head * n_head * v_head_dim
+                + head_id * v_head_dim
+                + dim_off[None, :],
+                v_val,
+                mask=v_d_mask & kv_block_mask,
+            )
+        if seq_block_id < 2 * pcp_world_size - pcp_rank:
+            tl.store(
+                out_k_tail_ptr
+                + kv_dst_idx_tail * n_head * k_head_dim
+                + head_id * k_head_dim
+                + dim_off[None, :],
+                k_val,
+                mask=k_d_mask & kv_block_mask,
+            )
+            tl.store(
+                out_v_tail_ptr
+                + kv_dst_idx_tail * n_head * v_head_dim
+                + head_id * v_head_dim
+                + dim_off[None, :],
+                v_val,
+                mask=v_d_mask & kv_block_mask,
+            )
+
+
+def fused_pcp_qkv_select(
+    q: torch.Tensor,
+    k: torch.Tensor,
+    v: torch.Tensor,
+    query_start_loc: torch.Tensor,
+    pcp_world_size: int,
+    pcp_rank: int,
+):
+    """
+    Select the query and kv tensors for PCP. Instead of calling
+    `torch.index_select` multiple times, this function fuses the
+    selection for Q, K, and V into a single kernel to reduce
+    kernel launch overhead.
+    Args:
+        q: query tensor on the current PCP rank.
+        k: key tensor across PCP ranks.
+        v: value tensor across PCP ranks.
+        query_start_loc: start location of each query.
+        pcp_world_size: number of PCP ranks.
+        pcp_rank: rank of the current PCP rank.
+    Returns:
+        q_head: selected query tensor for pcp head.
+        k_head: selected key tensor for pcp head.
+        v_head: selected value tensor for pcp head.
+        q_tail: selected query tensor for pcp tail.
+        k_tail: selected key tensor for pcp tail.
+        v_tail: selected value tensor for pcp tail.
+
+    """
+    q_head = torch.empty(
+        (q.size(0) // 2,) + q.shape[1:], device=q.device, dtype=q.dtype
+    )
+    q_tail = torch.empty_like(q_head)
+    k_head = torch.empty(
+        (q.size(0) // 2 * (pcp_rank + 1),) + k.shape[1:], device=k.device, dtype=k.dtype
+    )
+    v_head = torch.empty(
+        (q.size(0) // 2 * (pcp_rank + 1),) + v.shape[1:], device=v.device, dtype=v.dtype
+    )
+    k_tail = torch.empty(
+        (q.size(0) // 2 * (2 * pcp_world_size - pcp_rank),) + k.shape[1:],
+        device=k.device,
+        dtype=k.dtype,
+    )
+    v_tail = torch.empty(
+        (q.size(0) // 2 * (2 * pcp_world_size - pcp_rank),) + v.shape[1:],
+        device=v.device,
+        dtype=v.dtype,
+    )
+    BS = len(query_start_loc) - 1
+    DIM_BLOCK_SIZE: int = 64
+    SEQ_BLOCK_SIZE: int = 256
+    assert q.shape[1] == k.shape[1] == v.shape[1]
+    n_head = q.shape[1]
+    n_dim_block = (
+        max(q.shape[2], k.shape[2], v.shape[2]) + DIM_BLOCK_SIZE
+    ) // DIM_BLOCK_SIZE
+    grid = (
+        2 * pcp_world_size * BS,
+        n_head,
+        n_dim_block,
+    )
+    _fused_pcp_qkv_select_kernel[grid](
+        q,
+        q.stride(0),
+        q.stride(1),
+        k,
+        k.stride(0),
+        k.stride(1),
+        v,
+        v.stride(0),
+        v.stride(1),
+        query_start_loc,
+        q_head,
+        q_tail,
+        k_head,
+        k_tail,
+        v_head,
+        v_tail,
+        pcp_world_size,
+        pcp_rank,
+        n_head,
+        q.shape[2],
+        k.shape[2],
+        v.shape[2],
+        SEQ_BLOCK_SIZE,
+        DIM_BLOCK_SIZE,
+    )
+    return q_head, k_head, v_head, q_tail, k_tail, v_tail
--- a/vllm/distributed/parallel_state.py
+++ b/vllm/distributed/parallel_state.py
@@ -1092,9 +1092,6 @@ def get_dcp_group() -> GroupCoordinator:
     return _DCP
 
 
-# kept for backward compatibility
-get_context_model_parallel_group = get_dcp_group
-
 _PP: GroupCoordinator | None = None
 
 
--- a/vllm/v1/attention/backends/mla/common.py
+++ b/vllm/v1/attention/backends/mla/common.py
@@ -205,11 +205,15 @@ from vllm.attention.backends.abstract import (
     MLAAttentionImpl,
 )
 from vllm.attention.backends.utils import get_mla_dims
-from vllm.attention.ops.common import cp_lse_ag_out_rs
+from vllm.attention.ops.common import cp_lse_ag_out_ar, cp_lse_ag_out_rs
 from vllm.attention.ops.merge_attn_states import merge_attn_states
 from vllm.attention.utils.fa_utils import get_flash_attn_version
 from vllm.config import VllmConfig, get_current_vllm_config
-from vllm.distributed.parallel_state import get_dcp_group, is_global_first_rank
+from vllm.distributed.parallel_state import (
+    get_dcp_group,
+    get_pcp_group,
+    is_global_first_rank,
+)
 from vllm.logger import init_logger
 from vllm.model_executor.layers.batch_invariant import (
     vllm_is_batch_invariant,
@@ -225,9 +229,12 @@ from vllm.utils.math_utils import cdiv, round_down
 from vllm.v1.attention.backends.utils import (
     AttentionMetadataBuilder,
     CommonAttentionMetadata,
-    get_dcp_local_seq_lens,
+    get_cp_local_seq_lens,
+    get_pcp_kv_indices,
+    get_pcp_query_indices,
     get_per_layer_parameters,
     infer_global_hyperparameters,
+    pcp_kv_allgather_and_restore,
     split_decodes_and_prefills,
 )
 from vllm.v1.kv_cache_interface import AttentionSpec
@@ -350,6 +357,16 @@ class MLACommonPrefillMetadata:
         cu_seq_lens_lst: list[list[int]] | None = None
         chunk_size: int | None = None
 
+    @dataclass
+    class PCPMetadata:
+        # For PCP
+        pcp_allgather_restore_idx: torch.Tensor | None = None
+        kv_head_indices: torch.Tensor | None = None
+        kv_tail_indices: torch.Tensor | None = None
+        query_head_indices: torch.Tensor | None = None
+        query_tail_indices: torch.Tensor | None = None
+        output_restore_idx: torch.Tensor | None = None
+
     block_table: torch.Tensor
     query_start_loc: torch.Tensor
     max_query_len: int
@@ -357,6 +374,7 @@ class MLACommonPrefillMetadata:
     query_seq_lens: torch.Tensor | None = None
     workspace_buffer: torch.Tensor | None = None
     q_data_type: torch.dtype | None = None
+    pcp_metadata: PCPMetadata | None = None
 
 
 @dataclass
@@ -379,7 +397,7 @@ class CudnnPrefillMetadata(MLACommonPrefillMetadata):
 class MLACommonDecodeMetadata:
     block_table: torch.Tensor
     seq_lens: torch.Tensor
-    dcp_tot_seq_lens: torch.Tensor | None
+    cp_tot_seq_lens: torch.Tensor | None
 
 
 D = TypeVar("D", bound=MLACommonDecodeMetadata)
@@ -535,7 +553,7 @@ class MLACommonMetadataBuilder(AttentionMetadataBuilder[M]):
         vllm_config: VllmConfig,
         device: torch.device,
         metadata_cls: type[M] | None = None,
-        supports_dcp_with_varlen: bool = False,
+        supports_cp_with_varlen: bool = False,
     ):
         self.metadata_cls = (
             metadata_cls if metadata_cls is not None else MLACommonMetadata
@@ -558,9 +576,18 @@ class MLACommonMetadataBuilder(AttentionMetadataBuilder[M]):
             # DCP might not be initialized in testing
             self.dcp_world_size = 1
             self.dcp_rank = 0
-        self.dcp_local_block_size = parallel_config.cp_kv_cache_interleave_size
-        self.dcp_virtual_block_size = self.dcp_local_block_size * self.dcp_world_size
-        self.cp_kv_cache_interleave_size = parallel_config.cp_kv_cache_interleave_size
+        try:
+            self.pcp_world_size = get_pcp_group().world_size
+            self.pcp_rank = get_pcp_group().rank_in_group
+        except AssertionError:
+            # PCP might not be initialized in testing
+            self.pcp_world_size = 1
+            self.pcp_rank = 0
+        self.cp_world_size = self.dcp_world_size * self.pcp_world_size
+        self.cp_local_block_size = parallel_config.cp_kv_cache_interleave_size
+        self.cp_virtual_block_size = self.cp_local_block_size * self.cp_world_size
+        # TODO(yyj) Remove this once the PCP bug for decode_length > 1 is fixed.
+        supports_cp_with_varlen = supports_cp_with_varlen and self.pcp_world_size == 1
 
         # Don't try to access the runner on AMD
         if self.aot_schedule:
@@ -570,16 +597,16 @@ class MLACommonMetadataBuilder(AttentionMetadataBuilder[M]):
             self.determine_chunked_prefill_workspace_size(vllm_config)
         )
 
-        if self.dcp_world_size > 1:
-            # Note(hc): The local kvcache is incomplete when DCP is triggered,
-            # an additional kvcache allgather across the DCP group is therefore
-            # required, so the workspace has to be enlarged by 1/DCP relative
+        if self.cp_world_size > 1:
+            # Note(hc): The local kvcache is incomplete when DCP or PCP is triggered,
+            # an additional kvcache allgather across the DCP&PCP group is therefore
+            # required, so the workspace has to be enlarged by 1/CP relative
             # to the original TP allocation.
-            assert self.chunked_prefill_workspace_size % self.dcp_world_size == 0
+            assert self.chunked_prefill_workspace_size % self.cp_world_size == 0
             self.chunked_prefill_workspace = torch.empty(
                 (
                     self.chunked_prefill_workspace_size
-                    + self.chunked_prefill_workspace_size // self.dcp_world_size,
+                    + self.chunked_prefill_workspace_size // self.cp_world_size,
                     self.model_config.get_head_size(),
                 ),
                 dtype=self.model_config.dtype,
@@ -636,7 +663,7 @@ class MLACommonMetadataBuilder(AttentionMetadataBuilder[M]):
 
         supports_spec_decode = self.query_len_support != QueryLenSupport.SINGLE_ONLY
         self._init_reorder_batch_threshold(
-            self.reorder_batch_threshold, supports_spec_decode, supports_dcp_with_varlen
+            self.reorder_batch_threshold, supports_spec_decode, supports_cp_with_varlen
         )
 
         # Validate consistency between query_len_support and reorder_batch_threshold
@@ -730,12 +757,12 @@ class MLACommonMetadataBuilder(AttentionMetadataBuilder[M]):
         query_start_loc_cpu: torch.Tensor,
         query_start_loc_device: torch.Tensor,
         num_decode_tokens: int,
-        dcp_tot_seq_lens_device: torch.Tensor | None,
+        cp_tot_seq_lens_device: torch.Tensor | None,
     ) -> MLACommonDecodeMetadata:
         return MLACommonDecodeMetadata(
             block_table=block_table_tensor,
             seq_lens=seq_lens_device,
-            dcp_tot_seq_lens=dcp_tot_seq_lens_device,
+            cp_tot_seq_lens=cp_tot_seq_lens_device,
         )
 
     def build_for_cudagraph_capture(
@@ -765,6 +792,7 @@ class MLACommonMetadataBuilder(AttentionMetadataBuilder[M]):
         num_tokens = common_attn_metadata.num_actual_tokens
         max_query_len = common_attn_metadata.max_query_len
         max_seq_len = common_attn_metadata.max_seq_len
+        pcp_allgather_restore_idx = common_attn_metadata.pcp_allgather_restore_idx
 
         # Note(simon): be careful about the CPU <> GPU memory movement in this
         # function. We should avoid GPU -> CPU sync as much as possible because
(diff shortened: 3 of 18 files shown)


vllm/v1/attention/backends/mla/common.py
```diff
@@ -1342,24 +1404,74 @@ def _flash_attn_varlen_diff_headdims(
...
+                max_seqlen_q=prefill.max_query_len // 2,
+                max_seqlen_k=prefill.max_query_len
+                // 2
+                * (self.pcp_world_size * 2 - self.pcp_rank),
+                softmax_scale=self.scale,
+                causal=True,
+                return_softmax_lse=True,
+            )
+
+            output = torch.cat([output_head, output_tail], dim=0)
```
LucasWilkinson, line 1515: nit: we should try to avoid `torch.cat` as much as possible; these can be very slow

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -492,7 +507,31 @@ def __init__(
 
         # Only relevant for multimodal models
         if self.supports_mm_inputs:
-            self.is_mm_embed = self._make_buffer(self.max_num_tokens, dtype=torch.bool)
+            self.is_mm_embed = self._make_buffer(
+                max_buffer_num_tokens, dtype=torch.bool
+            )
+
+        # Persistent buffers for Prefill Context Parallism
+        if self.pcp_world_size > 1:
```
LucasWilkinson, line 538: can we separate all of this into a `PCPManager`? We should modularize this as much as possible to make it easier to support model runner v2 and not pollute the gpu model runner so much.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1022,6 +1061,113 @@ def _dummy_mm_kwargs(self, num_seqs: int) -> BatchedTensorInputs:
...
+        `self.num_pcp_pads_cpu` the number of padding tokens
+            per request for alignment,
+        `self.pcp_unpad_mask_cpu` the mask for non-padded tokens,
+        `self.pcp_allgather_restore_idx` indices to restore the
+            original vector order after PCP allgather.
+        Example:
+        >>> tokens = [1, 5, 8]
+        >>> pcp_world_size = 2
+        >>> pcp_rank = 0
+        >>> _update_tokens_for_pcp(tokens)
```
LucasWilkinson: can you please improve the doc string? it doesn't cover what `_update_tokens_for_pcp` so its hard to read through the example without first reading the code

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1022,6 +1061,113 @@ def _dummy_mm_kwargs(self, num_seqs: int) -> BatchedTensorInputs:
...
+        )
+        num_decode_reqs = sum(tokens <= self.reorder_batch_threshold)
+
+        self.num_pcp_pads_cpu[:num_reqs] = 0
+
+        num_decode_tokens = sum(tokens[:num_decode_reqs])
+
+        num_padded_scheduled_tokens = np.ceil(
+            tokens / (2 * self.pcp_world_size)
+        ).astype(np.int32) * (2 * self.pcp_world_size)
```
LucasWilkinson: please document where the `2 *` come from

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1022,6 +1061,113 @@ def _dummy_mm_kwargs(self, num_seqs: int) -> BatchedTensorInputs:
...
+        # we duplicate scheduled tokens of decode reqs to pcp_world_size
+        num_padded_scheduled_tokens[:num_decode_reqs] = (
+            tokens[:num_decode_reqs] * self.pcp_world_size
```
LucasWilkinson: it appears this assumes a decode length of 1? we should document assumptions like this everywhere so that we can more easily add spec decode support in the future

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1022,6 +1061,113 @@ def _dummy_mm_kwargs(self, num_seqs: int) -> BatchedTensorInputs:
...
+        self.pcp_unpad_mask_cpu[: pcp_padded_arange.shape[0]] = (
+            pcp_padded_arange < np.repeat(tokens, num_padded_scheduled_tokens)
+        )
+
+        pcp_tokens = num_padded_scheduled_tokens // self.pcp_world_size
+        pcp_chunk_sizes = (pcp_tokens // 2).clip(min=1)
+        pcp_chunk_sizes[:num_decode_reqs] = pcp_tokens[:num_decode_reqs]
+        _, pcp_arange = self._get_cumsum_and_arange(pcp_tokens)
+        _, pcp_chunk_arange = self._get_cumsum_and_arange(pcp_chunk_sizes)
+        pcp_head_chunk_mask = pcp_arange < np.repeat(pcp_chunk_sizes, pcp_tokens)
```
LucasWilkinson: can you please document what this is?

vllm/v1/worker/gpu_model_runner.py
FENP: Done.

vllm/v1/worker/gpu_model_runner.py
FENP: The description of function `_update_tokens_for_pcp` has been improved.

vllm/v1/worker/gpu_model_runner.py
FENP: Sure.

vllm/v1/worker/gpu_model_runner.py
FENP: How about something like this?
```python
Class PCPManager:
    """
    Manager for Prefill Context Parallelism (PCP) metadata and buffers.

    This manager encapsulates all PCP-related buffers and logic so that the
    GPUModelRunner can access them via `self.pcp_manager`..
    """

    def __init__(self, runner: ModelRunner, ...):
        self.pcp_allgather_restore_idx = runner._make_buffer(
            max_buffer_num_tokens, dtype=torch.int64
        )

        self.pcp_padded_slot_mapping = torch.empty(
            (max_buffer_num_tokens,), dtype=torch.int64, device=self.device
        )

        self.num_pcp_pads_cpu_tensor = torch.zeros(
            (max_num_reqs,), device="cpu", dtype=torch.int64, pin_memory=True
        )
        self.num_pcp_pads_cpu = self.num_pcp_pads_cpu_tensor.numpy()

        self.pcp_unpad_mask_cpu_tensor = torch.zeros(
            (max_buffer_num_tokens,), device="cpu", dtype=torch.bool, pin_memory=True
        )
        self.pcp_unpad_mask_cpu = self.pcp_unpad_mask_cpu_tensor.numpy()
        ...

    def update_tokens_for_pcp(self, tokens: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
        ...

```

vllm/v1/worker/gpu_model_runner.py
FENP: No. PCP does not split the decode tokens, so duplicate decode tokens here ensures that after splitting, all pcp ranks handle the all decode tokens.

Actually, decode length is just based on `tokens[:num_decode_reqs]`, i.e., how we define decode requests. 
Currently, the implementation works as follows:
```python
def _update_tokens_for_pcp(
    self,
    tokens: np.ndarray,
) -> tuple[np.ndarray, np.ndarray]:

    assert self.reorder_batch_threshold is not None, (
        "PCP depends on reorder batch to split decode and prefill requests."
    )
    num_decode_reqs = sum(tokens <= self.reorder_batch_threshold)
    num_decode_tokens = sum(tokens[:num_decode_reqs])
```

vllm/v1/worker/gpu_model_runner.py
LucasWilkinson: I think generally looks good other than if possible should not pass in `runner: ModelRunner`; we should just pass-in needed info/references to `update_tokens_for_pcp` to avoid a tight coupling to the model runner (making the transition to V2 smoother)

vllm/v1/worker/gpu_model_runner.py
LucasWilkinson: Ah apologies; my bad! I miss understood what `tokens` was; we should rename this to `num_scheduled_tokens` or `num_scheduled_tokens_per_request` to match more closely with the existing naming scheme within the gpu_model_runner/codebase

vllm/attention/ops/common.py
```diff
@@ -194,7 +194,8 @@ def _cp_lse_common(
     cp_attn_lse = cp_attn_lse.contiguous()
     lses = cp_group.all_gather(cp_attn_lse, dim=0).view_as(lses)
     out, lse = correct_attn_out(cp_attn_out, lses, cp_group.rank_in_group, ctx)
-    assert out.is_contiguous()
+    # remove out.is_contiguous() check
+    # Ref: https://github.com/vllm-project/vllm/pull/25414
```
LucasWilkinson: nit: rebase, this should be removed now

vllm/v1/attention/backends/mla/common.py
```diff
@@ -354,6 +361,14 @@ class ChunkedContextMetadata:
     chunked_context: ChunkedContextMetadata | None = None
     query_seq_lens: torch.Tensor | None = None
 
+    # For PCP
+    pcp_allgather_restore_idx: torch.Tensor | None = None
+    kv_head_indices: torch.Tensor | None = None
+    kv_tail_indices: torch.Tensor | None = None
+    query_head_indices: torch.Tensor | None = None
+    query_tail_indices: torch.Tensor | None = None
+    output_restore_idx: torch.Tensor | None = None
```
LucasWilkinson: nit: can we separate this into a `PCPMetadata` class?

vllm/v1/attention/backends/mla/common.py
```diff
@@ -1342,24 +1404,74 @@ def _flash_attn_varlen_diff_headdims(
...
+                max_seqlen_k=prefill.max_query_len // 2 * (self.pcp_rank + 1),
+                softmax_scale=self.scale,
+                causal=True,
+                return_softmax_lse=True,
+            )
+
+            output_tail, lse_tail = self._flash_attn_varlen_diff_headdims(
+                q=torch.index_select(q, 0, prefill.query_tail_indices),
+                k=torch.index_select(k, 0, prefill.kv_tail_indices),
+                v=torch.index_select(v, 0, prefill.kv_tail_indices),
```
LucasWilkinson: These index selects seem expensive; have you done any profiling / benchmarking yet? Id curious to know how bad (or not) these are

vllm/v1/attention/backends/mla/common.py
```diff
@@ -1342,24 +1404,74 @@ def _flash_attn_varlen_diff_headdims(
...
-            causal=True,
-            return_softmax_lse=return_softmax_lse,
-        )
+        assert self.pcp_world_size is not None
+        assert self.pcp_rank is not None
+        if self.pcp_world_size > 1:
+            # NOTE(yyj) PCP split the sequence using the DualChunkSwap strategy
+            # to ensure load balancing, so we need to split the query, key and
+            # value into two parts and run the attention twice.
+            # TODO(yyj) Add comments to clarify ...
```
LucasWilkinson: nit: please dont forget to clarify

vllm/distributed/parallel_state.py
```diff
@@ -1093,8 +1093,7 @@ def get_dcp_group() -> GroupCoordinator:
     return _DCP
 
 
-# kept for backward compatibility
-get_context_model_parallel_group = get_dcp_group
+_CP: GroupCoordinator | None = None
```
zhenwenqi2024: In the code of PCP and DCP, there are many places where we need to obtain world_size and rank. Could we abstract this into a separate function? like this:
def get_prefill_context_model_parallel_world_size():
    """Return world size for the tensor model parallel group."""
    return get_pcp_group().world_size


def get_prefill_context_model_parallel_rank():
    """Return my rank for the tensor model parallel group."""
    return get_pcp_group().rank_in_group

gjc0824: It would be best to add the following tool functions to this file, just like tp and dcp.
```
def get_prefill_context_model_parallel_world_size():
    """Return world size for the prefill context model parallel group."""
    return get_pcp_group().world_size


def get_prefill_context_model_parallel_rank():
    """Return my rank for the prefill context model parallel group."""
    return get_pcp_group().rank_in_group
```

vllm/v1/attention/backends/mla/common.py
```diff
@@ -933,11 +960,43 @@ def build(
...
+                output_restore_idx = torch.cat([q_head_idx, q_tail_idx]).argsort()
+                prefill_kv_start_loc_cpu = (
+                    prefill_query_start_loc_cpu * self.pcp_world_size
+                )
+                kv_head_idx, kv_tail_idx = get_pcp_kv_indices(
+                    prefill_kv_start_loc_cpu,
+                    self.pcp_rank,
+                    self.pcp_world_size,
+                )
+                kv_head_indices = kv_head_idx.to(device, dtype=torch.int32)
```
zhenwenqi2024: Will a synchronization operation be triggered here?

vllm/v1/worker/gpu_model_runner.py
FENP: Got it.

vllm/distributed/parallel_state.py
FENP: I couldn't find any usage of these functions. Also, it seems that PP and DP don't have similar ones. Why not just use get_pcp_group().world_size and get_pcp_group().rank_in_group directly?

gjc0824: Because the vLLM plugin for other hardware such as vllm-ascend has already used these functions in the existing implementation. This depends on vLLM.

vllm/v1/attention/backends/mla/common.py
FENP: Sure, I profiled with a 32K input. The results show that single index_select: ~9 μs. Given that attention computation takes several milliseconds, this overhead is insignificant.

image

vllm/v1/attention/backends/mla/common.py
FENP: Good suggestion

vllm/v1/attention/backends/mla/common.py
```diff
@@ -956,11 +983,43 @@ def build(
...
+                kv_tail_indices = kv_tail_idx.to(device, dtype=torch.int32)
```
zhenwenqi2024: Will a synchronization operation be triggered here?

FENP: I'm not sure if setting `non_blocking=True` makes sense with non-pinned memory. But it's worth a try.
BTW, I plan to switch these tensors to pinned memory when full graph support is available. This will be implemented in a follow-up PR.

vllm/distributed/parallel_state.py
FENP: I think get_pcp_group().* is already concise and clear enough. If there are compatibility concerns, we can consider keeping or adding the functions you need.

FENP: Ok, I see.

zhenwenqi2024: https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/fused_moe/config.py

we should add some property in this file for pcp,
image
image

vllm/v1/attention/backends/mla/common.py
FENP: Yes, done — I’ve improved the comments and included an example.

vllm/v1/attention/backends/mla/common.py
FENP: Done

vllm/distributed/parallel_state.py
FENP: Make sense.

FENP: https://github.com/vllm-project/vllm/blob/275de34170654274616082721348b7edd9741d32/vllm/distributed/parallel_state.py#L1079-L1083

@gjc0824 @zhenwenqi2024 Are you considering updating the usage in vllm-ascend? This could help improve consistency with the main codebase.

FENP:
After merging with `main`, I found a bug in PCP when decode length is greater than 1. I tried to investigate but couldn't locate the root cause. For now, I've set `supports_cp_with_varlen` to `False` when PCP is enabled.
This is a temporary workaround to prevent incorrect behavior. We should re-enable this feature once the bug is identified and fixed. cc @LucasWilkinson @pisceskkk @zhenwenqi2024 
commit：https://github.com/vllm-project/vllm/pull/28988/commits/53137ed3a447e7a198a61264dab40e01e489d7a2

vllm/v1/worker/gpu_model_runner.py
FENP: Done

vllm/v1/worker/gpu_model_runner.py
FENP: Fix

vllm/v1/attention/backends/mla/common.py
FENP: While I considered pre-allocating an appropriately sized empty tensor, it may still be insufficient to eliminate memory copies, as flash_attn_varlen_func internally allocates a new output tensor for the results.

What do you think about it?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1456,10 +1510,19 @@ def _prepare_inputs(
...
             # TODO: Support prompt logprobs.
             logits_indices = query_start_loc[1:] - 1
+            if self.pcp_world_size > 1:
+                logits_indices = (
+                    torch.from_numpy(cu_num_tokens) * self.pcp_world_size
+                    - self.pcp_manager.num_pcp_pads_cpu_tensor[:num_reqs]
+                    - 1
+                )
+            else:
+                logits_indices = query_start_loc[1:] - 1
```
LucasWilkinson: why repeated `logits_indices = query_start_loc[1:] - 1`?

can we do something like? the ultimate goal is to lower the visual presence and cognitive load so people can easily ignore the PCP stuff when reading gpu_model_runner if they don't care about PCP

```
logits_indices = query_start_loc[1:] - 1
if self.pcp_world_size > 1:
     logits_indices = self.pcp_manager.get_logits_indices(cu_num_tokens, num_reqs)
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1592,11 +1659,18 @@ def _build_attention_metadata(
...
+                else num_tokens * self.pcp_world_size
+                - sum(self.pcp_manager.num_pcp_pads_cpu[:num_reqs])
+            )
+
             if isinstance(kv_cache_group.kv_cache_spec, EncoderOnlyAttentionSpec):
                 # Encoder-only layers do not have KV cache, so we need to
                 # create a dummy block table and slot mapping for them.
                 blk_table_tensor = torch.zeros(
-                    (num_reqs_padded, 1),
+                    (num_tokens_padded, 1),
```
LucasWilkinson: nit: rebase error?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1420,9 +1467,16 @@ def _prepare_inputs(
...
+        if self.pcp_world_size > 1:
+            self.discard_request_mask.np[:num_reqs] = (
+                self.input_batch.num_computed_tokens_cpu[:num_reqs]
+                + num_scheduled_tokens * self.pcp_world_size
+                - self.pcp_manager.num_pcp_pads_cpu[:num_reqs]
+            ) < num_tokens_np
+        else:
+            self.discard_request_mask.np[:num_reqs] = (
+                self.seq_lens.np[:num_reqs] < num_tokens_np
+            )
```
LucasWilkinson, line 1497: nit similar to: https://github.com/vllm-project/vllm/pull/28988/files#r2587693784

can we do:

```
if self.pcp_world_size > 1:
      self.discard_request_mask.np[:num_reqs] = self.pcp_manager.get_discard_request_mask(...)
else:
      self.discard_request_mask.np[:num_reqs] = (
         self.seq_lens.np[:num_reqs] < num_tokens_np
      )

```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1608,12 +1682,26 @@ def _build_attention_metadata(
...
-                slot_mapping = blk_table.slot_mapping.gpu[:num_tokens_padded]
+                slot_mapping = blk_table.slot_mapping.gpu[:maybe_pcp_full_tokens]
 
                 # Fill unused with -1. Needed for reshape_and_cache in full cuda
                 # graph mode. `blk_table_tensor` -1 to match mamba PAD_SLOT_ID
-                slot_mapping[num_tokens:num_tokens_padded].fill_(-1)
-                blk_table_tensor[num_reqs:num_reqs_padded].fill_(-1)
+                if self.pcp_world_size == 1:
+                    slot_mapping[num_tokens:num_tokens_padded].fill_(-1)
+                    blk_table_tensor[num_reqs:num_reqs_padded].fill_(-1)
```
LucasWilkinson, line 1666: won't this cause issues for decode batches with full-cudagraphs (we should try to get FULL_AND_PIECEWISE turned on for DCP)

LucasWilkinson (commented):
some very initial comments; will review more thoroughly tmrw

vllm/v1/worker/gpu_model_runner.py
FENP: Yes, sorry for the mistake, fixed it.

vllm/v1/worker/gpu_model_runner.py
FENP: 👍 Modified as suggested. Refactor the logic of other PCPs into functions in this way, including: get_restore_hidden_states, get_discard_request_mask, and get_padded_slot_mapping.

vllm/v1/worker/gpu_model_runner.py
FENP: Modified as suggested.

vllm/v1/worker/gpu_model_runner.py
FENP: Since the PCP logic is incompatible with pad attn by now, I've added an `if` condition here to ensure that these two lines of code are executed only when PCP is disabled. I don't think this will have any impact on DCP.

vllm/v1/worker/gpu_model_runner.py
zhenwenqi2024: hello， in  _dummy_run ， the tokens should be splited by pcp_size.
also ,I think in _dummy_run， we should call self._model_forward other thab self.model()

vllm/v1/attention/backends/utils.py
```diff
@@ -1109,41 +1115,143 @@ def compute_causal_conv1d_metadata(query_start_loc_p: torch.Tensor):
...
+    pcp_rank: int,
+    pcp_size: int,
+):
+    kv_head_indices, _ = get_pcp_part_indices(
+        cu_num_tokens,
+        pcp_rank + 1,
+        2 * pcp_size,
+        return_head=True,
+    )
+    kv_tail_indices, _ = get_pcp_part_indices(
```
LucasWilkinson: shouldn't this be `_, kv_tail_indices`?

FENP: No, `return_head` is setted True here. For query, the required KV is starting from the first token, so we always need to return the indices starting from the head.

vllm/v1/worker/gpu_model_runner.py
FENP: > hello， in _dummy_run ， the tokens should be splited by pcp_size. also ,I think in _dummy_run， we should call self._model_forward other thab self.model()

Done. Split num_tokens by pcp_world_size.

vllm/v1/attention/backends/mla/common.py
```diff
@@ -1372,24 +1444,92 @@ def _flash_attn_varlen_diff_headdims(
...
+            pcp_metadata = prefill.pcp_metadata
+            assert pcp_metadata is not None
+            output_head, lse_head = self._flash_attn_varlen_diff_headdims(
+                q=torch.index_select(q, 0, pcp_metadata.query_head_indices),
+                k=torch.index_select(k, 0, pcp_metadata.kv_head_indices),
+                v=torch.index_select(v, 0, pcp_metadata.kv_head_indices),
+                cu_seqlens_q=prefill.query_start_loc // 2,
+                cu_seqlens_k=prefill.query_start_loc // 2 * (self.pcp_rank + 1),
+                max_seqlen_q=prefill.max_query_len // 2,
+                max_seqlen_k=prefill.max_query_len // 2 * (self.pcp_rank + 1),
```
LucasWilkinson, line 1494: can we move the `// 2` floor divs of the hot path? im seeing them show up in the profiles 

Image

vllm/v1/attention/backends/mla/common.py
```diff
@@ -1372,24 +1444,92 @@ def _flash_attn_varlen_diff_headdims(
...
+            #    Q\KV  0 1 2 3
+            # head 1   1 1 0 0
+            #      -----------
+            # tail 2   1 1 1 0
+
+            pcp_metadata = prefill.pcp_metadata
+            assert pcp_metadata is not None
+            output_head, lse_head = self._flash_attn_varlen_diff_headdims(
+                q=torch.index_select(q, 0, pcp_metadata.query_head_indices),
+                k=torch.index_select(k, 0, pcp_metadata.kv_head_indices),
```
LucasWilkinson: could we replace the k/v index selects by instead treating `pcp_metadata.kv_head_indices` as the block table for a page_size==1 kv-cache (and pass k/v directly)? FA3 supports page size 1

LucasWilkinson (commented):
i ran some profiles and I think theres way to many small torch ops in the current implementation (e.g. index_select) leading to excessive CPU overhead; i know this is targeting very long prefills (this was tested with 16k input) but I think we should try to optimize this more before landing

Image

vllm/v1/attention/backends/mla/common.py
FENP: 👍  It's worth a try

FENP: Hi @LucasWilkinson, I've thought about this issue a bit more.
Could we consider using a single Triton kernel to perform the `index select` for qkv all at once? Compared to treating `pcp_metadata.kv_head_indices` as a block table, this approach could avoid the additional `index select` call for q. In addition, I think this approach would also make it easier to migrate to other attention backends.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1463,10 +1527,15 @@ def _prepare_inputs(
             # We will ignore the sampled tokens from the partial requests.
             # TODO: Support prompt logprobs.
             logits_indices = query_start_loc[1:] - 1
+            if self.pcp_world_size > 1:
+                logits_indices = self.pcp_manager.get_logits_indices(
```
zhenwenqi2024, line 1532: logits_indices here seem a cpu_tensor , maybe it should be a devie_tensor?

vllm/v1/attention/backends/mla/common.py
LucasWilkinson: makes sense to me; assuming the triton launch overheads are reasonable since we unfortunately don't have prefill CGs to hide this overhead

FENP: Thanks for the feedback! I'm currently working on this change — will push the update soon.

vllm/v1/worker/gpu_model_runner.py
FENP: Make sense.

vllm/v1/worker/utils.py
```diff
@@ -128,6 +131,290 @@ def reset_cache(self) -> None:
...
+        # Build the restore index used after allgather.
+        padded_pos_start_loc = np.roll(cu_padded_tokens, 1)
+        padded_pos_start_loc[0] = 0
+        all_positions_lst = [
+            get_current_rank_positions(padded_pos_start_loc, rank_i)
+            for rank_i in range(self.pcp_world_size)
+        ]
+        all_positions = np.concatenate(all_positions_lst)
+        self.pcp_allgather_restore_idx.np[: all_positions.shape[0]] = (
+            all_positions.argsort()
```
zhenwenqi2024: now, all_positions is a cpu_tensor, how do you think put the tensor to device and then sort?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1615,12 +1695,19 @@ def _build_attention_metadata(
...
                 # Fill unused with -1. Needed for reshape_and_cache in full cuda
                 # graph mode. `blk_table_tensor` -1 to match mamba PAD_SLOT_ID
-                slot_mapping[num_tokens:num_tokens_padded].fill_(-1)
-                blk_table_tensor[num_reqs:num_reqs_padded].fill_(-1)
+                if self.pcp_world_size == 1:
+                    slot_mapping[num_tokens:num_tokens_padded].fill_(-1)
+                    blk_table_tensor[num_reqs:num_reqs_padded].fill_(-1)
+
+            if self.pcp_world_size > 1:
+                slot_mapping = self.pcp_manager.get_padded_slot_mapping(
```
zhenwenqi2024, line 1668: the address of slot_mapping here seems changed, I think it may infulence cudagraph in future。maybe we could make the address fixed？

vllm/v1/worker/utils.py
```diff
@@ -128,6 +131,290 @@ def reset_cache(self) -> None:
...
+    def get_padded_slot_mapping(self, num_tokens: int, slot_mapping: torch.Tensor):
+        # After pcp allgather and restore, there are padded tokens in kv,
+        # so we need pad slotmapping for alignment.
+        pcp_padded_slot_mapping = self.pcp_padded_slot_mapping[
+            : num_tokens * self.pcp_world_size
+        ]
+        cp_unpad_mask = self.pcp_unpad_mask_cpu_tensor[
+            : num_tokens * self.pcp_world_size
+        ]
+        pcp_padded_slot_mapping.fill_(-1)
```
zhenwenqi2024: could we delete this operation and initilized with full of -1

chaunceyjiang:
hi @FENP Can you resolve the conflict? I want to test it locally.

FENP:
> hi @FENP Can you resolve the conflict? I want to test it locally.

Thank you for your attention. Would you like to test it first on current development branch? I'm still working on fusing the QKV index-select operation using Triton to reduce kernel launch overhead. I've implemented a initial version, but its performance is not great yet—I'm still optimizing it. After that, I'll rebase onto main.

vllm/v1/worker/gpu_model_runner.py
FENP: I'll take a look after the index select optimization is completed.

vllm/v1/attention/backends/mla/common.py
FENP: Hi @LucasWilkinson, I have fused index select for qkv https://github.com/vllm-project/vllm/pull/28988/changes/ae8e5f35f2cfcea61daec8c3a182e8502e00005f.
The launch overhead of the fused QKV select op has been successfully reduced to 4 μs.
image

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -481,25 +486,38 @@ def __init__(
         # Cache the device properties.
         self._init_device_properties()
 
+        if self.pcp_world_size > 1:
+            # NOTE For PCP, we will pad the tokens of each request
+            # to a multiple of 2 * pcp_size that is possible greater
+            # than the max_num_batched_tokens.
+            max_buffer_num_tokens = (
```
LucasWilkinson: nit: can we rename this to `max_padded_num_tokens`?

vllm/v1/worker/utils.py
FENP: How can we ensure that the data in the slot mapping won't be dirty after multiple forward?

vllm/v1/worker/gpu_model_runner.py
FENP: I took a look, and I think its address is fixed. As long as PCP is enabled, we always use pcp_padded_slot_mapping.
```python
self.pcp_padded_slot_mapping = torch.empty(
    (max_padded_num_tokens,),
    dtype=torch.int64,
    device=device,
)
```

vllm/v1/worker/utils.py
FENP: Considering chunked prefill, the data size of `all_positions` is not large—could the GPU kernel launch overhead be too high? Have you tried measuring the overhead of doing it this way?

vllm/v1/worker/gpu_model_runner.py
FENP: Of course. Renamed.

LucasWilkinson:
@FENP is this ready for another round of review? happy to start reviewing whenever it is ready

FENP:
> @FENP is this ready for another round of review? happy to start reviewing whenever it is ready

@LucasWilkinson It’s ready for your review. Thanks in advance!
BTW, the current pre-commit error is a bit strange — I don't understand why it's reporting "Undefined name `num_scheduled_tokens_np`" at `vllm/v1/worker/gpu_model_runner.py:3252:52`.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -3118,129 +3149,134 @@ def execute_model(
...
-            if self.cache_config.kv_sharing_fast_prefill:
-                assert not self.num_prompt_logprobs, (
-                    "--kv-sharing-fast-prefill produces incorrect "
-                    "logprobs for prompt tokens, tokens, please disable "
-                    "it when the requests need prompt logprobs"
+                    num_scheduled_tokens_np,
                 )
+                if self.pcp_world_size > 1:
+                    max_num_scheduled_tokens = int(num_scheduled_tokens_np.max())
+                    num_tokens_unpadded = scheduler_output.total_num_scheduled_tokens
```
LucasWilkinson, line 3207: I find it quite messy that `_prepare_inputs` modifies `num_scheduled_tokens_np` and `scheduler_output` internally. Its not very clear to the reader here thats whats happening and thats why this recomputation/re-assignment is required. 

I think we should try harder to keep PCP more isolated for now, im working on an idea here (vibe coded and not tested yet): https://github.com/FENP/vllm/pull/4 to leave `_prepare_inputs` untouched. It just shuffles the inputs after preparation to select for the tokens this PCP rank cares about. It means we do some duplicated/wasted work, but I think better for the initial implementations and we can do broader refactors on the model runner later to make it support this more naturally with less duplicated/wasted work, potentially by breaking up prepare inputs. Thoughts? (sorry not a complete review yet, will continue tmrw)

Hg-Spark:
Accidental connector test comment; please disregard.

PR #25881, opened 2025-09-29
[Bugfix][Multi Modal] Fix broken frames in video input

Jixin10:
## Purpose
Fix #20313  

Fix the internal server error when the video input has broken frames. 

The engine get frames num via `total_frames_num = int(cap.get(cv2.CAP_PROP_FRAME_COUNT))` , however some of the frames are broken and cannot be read via `cap.grab()` and `cap.retrieve()`. Therefore `i` and `num_frames` are not equal, causing internal server error.

## Test Plan
```
curl -X POST localhost:50799/v1/chat/completions -H "Content-Type: application/json" -d '{
  "model":"qwen2.5_vl",
  "messages": [
      {
        "role": "user",
        "content": [
          {
            "video_url": {
              "url": "{vido_url}"
            },
            "type": "video_url"
          },
          {
            "type": "text",
            "text": "Please tell me what the video describes"
          }
        ]
      }
  ],
  "stream": false,
  "max_tokens": 1000
}'
```

## Test Result
Before fix:
```
AssertionError: Expected reading 32 frames, but only loaded 31 frames from video.
INFO:     127.0.0.1:26426 - "POST /v1/chat/completions HTTP/1.1" 500 Internal Server Error
```

After fix:
```
{"id":"chatcmpl-9246e6c97b934cbb829897080ccc3822","object":"chat.completion","created":1759148580,"model":"qwen2.5_vl","choices":[{"index":0,"message":{"role":"assistant","reasoning_content":null,"content":"The video shows xxx","tool_calls":[]},"logprobs":null,"finish_reason":"stop","stop_reason":null}],"usage":{"prompt_tokens":6281,"total_tokens":6469,"completion_tokens":188,"prompt_tokens_details":null},"prompt_logprobs":null}
```
---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #29001, opened 2025-11-19
[Bugfix] Handle broken frames in video loading

gcanlin:
## Purpose

Fixes [#20313](https://github.com/vllm-project/vllm/issues/20313), [vllm-ascend#4186](https://github.com/vllm-project/vllm-ascend/issues/4186) Following [#25881](https://github.com/vllm-project/vllm/pull/25881).

Fix assertion error when videos contain broken or corrupted frames. Previously, the code would fail with an assertion error if `grab()` or `retrieve()` failed for any frame in the sampling set. Now we track valid frames during iteration and return only successfully loaded frames.

Co-Authored-By: @Jixin10 <lvjiang.lj@alibaba-inc.com>

## Test Plan

`pytest tests/multimodal/test_video.py::test_video_backend_handles_broken_frames -v --log-cli-level=WARNING`

## Test Result

```
(APIServer pid=1135632) WARNING 11-19 15:28:19 [video.py:87] Failed to grab frame 17 during video loading. This frame will be skipped.
(APIServer pid=1135632) WARNING 11-19 15:28:19 [video.py:109] Video loading completed with 1 broken/unreadable frames. Expected 26 frames but only loaded 25 frames.
(APIServer pid=1135632) INFO 11-19 15:28:19 [video.py:206] OpenCVVideoBackend.load_bytes took 27.460 ms
```

---
<details>

</details>


Isotr0py (approved):
LGTM. Would be better if we can add a regression test for this.

vllm/multimodal/video.py
```diff
@@ -154,7 +180,7 @@ def load_bytes(
             "do_sample_frames": num_frames_to_sample == total_frames_num,
         }
 
-        return frames, metadata
+        return frames[:valid_num_frames], metadata
```
Isotr0py: BTW, can you also update the `metadata` part? Otherwise video processor needed metadata can still fail due to inconsistency between exact video and metadata:
```
        metadata = {
            "total_num_frames": total_frames_num,
            "fps": original_fps,
            "duration": duration,
            "video_backend": "opencv",
            "frames_indices": list(valid_frame_idx),
            # extra field used to control hf processor's video
            # sampling behavior
            "do_sample_frames": valid_num_frames == total_frames_num,
        }
```

gcanlin: Thanks. I have updated it. And I extract the duplicate code to a function `_read_frames`. Not sure if break something I miss. But for one example, it doesn't take much regression.

vllm/multimodal/video.py
```diff
@@ -185,10 +228,10 @@ def load_bytes(
 
         # Refer to:
         # https://github.com/huggingface/transformers/blob/v4.55.4/src/transformers/models/glm4v/video_processing_glm4v.py#L103-L140
-        frame_indices: range | list[int]
+        frame_indices_list: list[int]
```
gcanlin: Is the type `range` only for performance? I want to make two backends keep consistent so that we can extract them into the function `_read_frames`. And from the testing result, it didn't cause any performance regression.

wangxiyuan:
please add the origin author as co-authored-by, thanks.

Isotr0py (approved):
LGTM. Thanks!

gcanlin:
> please add the origin author as co-authored-by, thanks.

Could @Isotr0py help add the co-author @Jixin10 when squash-merging? Thanks for the original contribution. (Want to keep the almost ready CI)

```
Co-authored-by: 凌葭 <lvjiang.lj@alibaba-inc.com>
```

Jixin10:
@gcanlin @Isotr0py Thank you! Glad to see that the issue has been traced and resolved. I previously thought that the assert for `i == num_frames_to_sample` couldn't be modified. I'm still wondering why this assert was added. It seems it could be completely removed and just need to add some warning logs?

gcanlin:
> It seems it could be completely removed and just need to add some warning logs?

~~I guess when `cap.grab()` is not ok and `idx not in frame_indices`,  path will hit the assert, which follows the original logic. But not sure whether `idx not in frame_indices` would happen.~~

Sorry, seems that we actually doesn't need `assert i == valid_num_frames` anymore.

PR #24945, opened 2025-09-16
[EP/DP][API Server] Enable DP-aware routing in OpenAI API requests

Prowindy:
Signed-off-by: Cong Chen <prowindy@gmail.com>
  ## Purpose

  This PR adds support for the `data_parallel_rank` parameter in vLLM's OpenAI API, enabling external routers to specify which data parallel rank
   should handle a request. This enhancement improves load distribution and cache locality in multi-GPU deployments by allowing DP-aware routing
  from router to engine core.

  **Key functionality:**
  - External routers can inject `data_parallel_rank` into API requests
  - vLLM server routes requests to specific GPU ranks based on the parameter
  - Maintains backward compatibility when parameter is not provided
  - Enables intelligent load balancing and cache-aware routing strategies

  ## Test Plan

  1. **End-to-End DP-Aware Routing Test**:
     ```bash
     # Start vLLM server with 8 DP ranks
     vllm serve meta-llama/Meta-Llama-3.1-8B-Instruct \
       --data-parallel-size 8 --tensor-parallel-size 1 --port 8000

     # Start vLLM-router with DP-aware routing (selective ranks 0,2,4,6)
     vllm-router --worker-urls http://0.0.0.0:8000 \
       --dp-aware --policy round_robin --port 30000

     # Generate test traffic
     vllm bench serve --num-prompts 100 --port 30000 \
       --endpoint /v1/completions --max-concurrency 32

  2. Validation Methods:
    - Added comprehensive logging throughout the request pipeline
    - Implemented rank counters to track request distribution
    - Verified GPU device assignment for each request
    - Tested with multiple routing policies (round_robin, cache_aware)

  Test Result

  ✅ Successful DP-Aware Routing Validation:

  1. Router Discovery: Successfully identified 4 DP-aware workers
  workers: ["http://0.0.0.0:8000@0", "http://0.0.0.0:8000@2", "http://0.0.0.0:8000@4", "http://0.0.0.0:8000@6"]
  2. Request Distribution: Balanced traffic across selected ranks
  Current counters: {0: 2, 2: 3, 4: 3, 6: 2}
  3. End-to-End Pipeline Verification:
  API_SERVER_DEBUG: Received completion request with data_parallel_rank=4
  ROUTING_DEBUG: Using data_parallel_rank=4 directly
  ENGINE_DISPATCH DEBUG: Request dispatched to engine identity 0400 (rank 4)
  GPU_DEVICE_DEBUG: Request received on GPU device, parallel_config.rank=4
[...]

PR #29002, opened 2025-11-19
[Bugfix] fix DP-aware routing in OpenAI API requests

inkcherry:
## Purpose
fix  https://github.com/vllm-project/vllm/pull/24945
In ```add_request```, duplicate initialization is skipped, but during the previous```self.processor.process_inputs```, ```data_parallel_rank``` is not initialized. Using ```-H 'X-data-parallel-rank'``` to specify the data parallel rank would be invalid in this case., cc @njhill 

https://github.com/vllm-project/vllm/blob/d69062c67af46a2e624be92162e9db585eef329b/vllm/v1/engine/async_llm.py#L283-L302

## Test Result

---
<details>

</details>


vllm/entrypoints/openai/serving_completion.py
```diff
@@ -225,6 +225,7 @@ async def create_completion(
                         lora_request=lora_request,
                         trace_headers=trace_headers,
                         priority=request.priority,
+                        data_parallel_rank=data_parallel_rank,
```
markmc, line 233: Looks like a similar fix is required in `serving_chat.py`?

Also would be good to catch this issue in the `test_serving_chat_data_parallel_rank_extraction` test

Even better, it would be great to add a similar test for `serving_completion` !

markmc (dismissed):
Nice catch!

/cc @Prowindy

vllm/entrypoints/openai/serving_completion.py
ChuanLi1101, line 233: Could this be a breaking change for the current API?

inkcherry: @markmc  Thanks for the comment. I've added it to ```serving_chat.py```.

I noticed that mock objects won't trigger this error. So I added a test with a real engine for coverage, placed after the ```test_dp_rank_argument``` test.

inkcherry: Actually not. When unspecified, it defaults to None and uses the default DP load algorithm

njhill (approved):
Thanks @inkcherry!

Doesn't need to be addressed in this PR but I don't see why we would only want to support this header on the chat and completion endpoints, it could apply similarly to all of the endpoints.

PR #29003, opened 2025-11-19
[CI/Build] Add max_jobs docker build arg to rocm_base to prevent oom

epipho:
## Purpose
Building the rocm_base docker image uses a a lot of memory on machines with large numbers of CPU cores, consistently going OOM on a 32-core / 64GB machine. 

This adds a build arg to set the `MAX_JOBS` env variable to limit the number of processes Ninja will spin up while building. The default is set to `2` to mirror the values in the main `Dockerfile`.

## Test Plan
N/A

## Test Result
N/A


gshtras:
Is it worth using $(nproc) for this? Or a higher default, matching the environment where we build the base images - 2T RAM, 200+CPUs (rationale being, the end user should not need to build the base image themselves, we publish them as rocm/vllm-dev:base)

PR #29013, opened 2025-11-19
CPU KV Offloading: Use more CUDA streams

orozery:
Prior to this PR, the CPU offloading connector used 2 global CUDA streams, one for loads and one for stores.
When there are a lot of concurrent loads, the stream queue (which has a capacity of about 1000 on A100) can get full.
When this happens, submitting more operations to the stream becomes blocking, which is harmful to model execution.
To overcome this, this PR changes the CPU KV offloading to use a unique CUDA stream per transfer.

I have witnessed this is a true issue by timing the time it takes to call `cudaMemcpyAsync`, which increases from 2us to 180us when the stream queue gets full.


vllm/v1/kv_offload/worker/cpu_gpu.py
```diff
@@ -51,7 +52,125 @@ def expand_block_ids(
...
+            skip_count=src_sub_blocks_to_skip,
+        )
+        expand_block_ids(dst_blocks, self.dst_block_size_factor, src_to_dst[:, 1])
+        src_to_dst_tensor = torch.from_numpy(src_to_dst)
+
+        stream = (
+            self._stream_pool.pop()
+            if self._stream_pool
+            else torch.cuda.Stream(priority=self.priority)
+        )
```
njhill, line 137: Should we keep a bound on the max number of streams (start to share after this)? Or is the number of concurrent transfers already bounded somehow?

vllm/v1/kv_offload/worker/cpu_gpu.py
```diff
@@ -51,7 +52,125 @@ def expand_block_ids(
...
+            for src_tensor, dst_tensor, kv_dim in zip(
+                self.src_tensors, self.dst_tensors, self.kv_dim_before_num_blocks
+            ):
+                if kv_dim:
+                    src_key_cache = src_tensor[0]
+                    dst_key_cache = dst_tensor[0]
+                    ops.swap_blocks(src_key_cache, dst_key_cache, src_to_dst_tensor)
+                    src_value_cache = src_tensor[1]
+                    dst_value_cache = dst_tensor[1]
+                    ops.swap_blocks(src_value_cache, dst_value_cache, src_to_dst_tensor)
```
njhill: nit: bit cleaner
```suggestion
                    src_key_cache, src_value_cache = src_tensor
                    dst_key_cache, dst_value_cache = dst_tensor
                    ops.swap_blocks(src_key_cache, dst_key_cache, src_to_dst_tensor)
                    ops.swap_blocks(src_value_cache, dst_value_cache, src_to_dst_tensor)
```

njhill (commented):
Thanks @orozery, looks great just a couple of minor comments.

vllm/v1/kv_offload/worker/cpu_gpu.py
orozery: I don't think there's a reason to limit.
On the GPU side, it seems there's no limit on the number of streams.
Also, note that the streams do not run in parallel, as each transfer stream waits for the previous transfer to finish.
The only purpose for creating multiple streams is to overcome the 1024-ops-per-stream limit.
I tested this PR with 100000 concurrent small requests loading from CPU and it handles it well.
It does not really create 100000 as the GPU KV cache gets full so only around 1500 requests get allocated in any point in time, so basically my test used 1500 streams with no issues.

I am planning on limiting the number of requests being async loaded, to count towards `max_num_seqs` (which equals 1024 for H100).
See this draft PR #29877.
So once we get that in, the number of streams will be bounded by, roughly, `max_num_seqs`.

njhill (approved):

vllm/v1/kv_offload/worker/cpu_gpu.py
```diff
@@ -51,7 +52,123 @@ def expand_block_ids(
...
+        )
+        expand_block_ids(dst_blocks, self.dst_block_size_factor, src_to_dst[:, 1])
+        src_to_dst_tensor = torch.from_numpy(src_to_dst)
+
+        stream = (
+            self._stream_pool.pop()
+            if self._stream_pool
+            else torch.cuda.Stream(priority=self.priority)
+        )
+        event = self._event_pool.pop() if self._event_pool else torch.Event()
```
lixiaobai09, line 138: Hi, there seems to be a problem.

In the swapping-out situation, this workflow does not guarantee that the swap operation occurs after the forward computation, potentially resulting in the incorrect KV cache being swapped out.

So, maybe we can insert such lines of code?
```
        current_forward_stream = torch.cuda.current_stream()
        if (src_spec.medium() == "gpu") and (dst_spec.medium() == "cpu"):
            stream.wait_stream(current_forward_stream)
```

orozery: Even before this PR, the offloading connector used a separate cuda stream for executing the transfers.
This stream was not synced with the model execution stream.
This is also the case for other connectors.
Looking at the connector API itself, specifically `save_kv_layer` and `wait_for_save`. 
My assumption was that when these functions are called we can be sure the KV cache is safe to read (without syncing on the default stream).
@NickLucche @njhill can you confirm?

orozery: Digging more into the code, looks like you're right @lixiaobai09.
This is indeed a bug, though not introduced by this PR, but existed from the beginning of the OffloadingConnector.
I just opened #31341 to fix it.
@lixiaobai09 thank you so much for point this out!

PR #29016, opened 2025-11-19
Updating the mirror of test-amd.yaml as of 2025-11-18

Alexei-V-Ivanov-AMD:
Updating the mirror of test-amd.yaml as of 2025-11-18

Signed-off-by: Alexei V. Ivanov <alexei.ivanov@amd.com>


.buildkite/test-amd.yaml
```diff
@@ -694,7 +711,7 @@ steps:
   - vllm/model_executor/models/whisper.py
   commands: # LMEval
   # Transcription WER check is skipped because encoder-decoder models are not supported on ROCm, see https://github.com/vllm-project/vllm/issues/27442
-  - pytest -s entrypoints/openai/correctness/  --ignore entrypoints/openai/correctness/test_transcription_api_correctness.py
+  - pytest -s entrypoints/openai/correctness/
 
 - label: OpenAI-Compatible Tool Use # 23 min
   timeout_in_minutes: 35
```
zhewenl, line 700: we need to ignore this since encoder-decoder models are not supported on ROCm

Alexei-V-Ivanov-AMD: There is an effort to support specifically OpenAI Whisper (e.g. https://github.com/vllm-project/vllm/pull/28376)
for which Word Error Rate (WER) is the predominant quality metric.

tjtanaa (approved):

.buildkite/test-amd.yaml
tjtanaa: @Alexei-V-Ivanov-AMD Could we disable the test first? The PR are still under reviewing. It will take some time.

PR #29031, opened 2025-11-19
[Core] DO NOT trim spec token lists for future reuse

Jialin:
## Purpose

Continuous effort on top of #28917, DO NOT trim spec token lists ever to ensure lists are being reuse consistently.

## Test Plan & Test Result
CI signals

---
<details>

</details>

### DIFF
--- a/vllm/v1/worker/gpu_input_batch.py
+++ b/vllm/v1/worker/gpu_input_batch.py
@@ -309,14 +309,12 @@ class InputBatch:
         if req_index == len(self._req_ids):
             self._req_ids.append(req_id)
             self.req_output_token_ids.append(request.output_token_ids)
-            self.spec_token_ids.append([])
         else:
             self._req_ids[req_index] = req_id
             self.req_output_token_ids[req_index] = request.output_token_ids
-            self.spec_token_ids[req_index].clear()
 
         self.req_id_to_index[req_id] = req_index
-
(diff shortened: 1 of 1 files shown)


vllm/v1/worker/gpu_input_batch.py
```diff
@@ -732,10 +731,12 @@ def condense(self) -> None:
             # Decrement last_req_index since it is now empty.
             last_req_index -= 1
 
+        for i in range(num_reqs, len(self.spec_token_ids)):
+            self.spec_token_ids[i].clear()
```
njhill: we are introducing a new loop of ~max_num_reqs iterations here, I'm unsure about this

Jialin: I think we could get rid of this, and ensure to clear() before usage instead (which I believe we're doing so)

vllm/v1/worker/gpu_input_batch.py
```diff
@@ -732,10 +731,12 @@ def condense(self) -> None:
...
+
         # Trim lists to the batch size.
         del self._req_ids[num_reqs:]
         del self.req_output_token_ids[num_reqs:]
-        del self.spec_token_ids[num_reqs:]
```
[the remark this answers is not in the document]
Jialin: Updated, as we never trim the list with the change, we could safely remove this out from the checks.

PR #29035, opened 2025-11-19
[Core] Replace LogprobsLists with multiple numpy arrays in EngineCoreOutput

Jialin:
## Purpose
Similar to quite a few other GC optimizations, each batch would create <batch_size> of EngineCoreOutputs. For large batch size scenarios which means it would guarantee to invoke GC0 which is less ideal.

In this PR, we only replace LogprobsLists with a few numpy arrays in EngineCoreOutput, which would get rid of the GC0 invocations. For large batch size cases, it turned out to improve throughput a lot.

## Test Plan & Test Result
CI Signals

---
<details>

</details>

### DIFF
--- a/vllm/v1/core/sched/scheduler.py
+++ b/vllm/v1/core/sched/scheduler.py
@@ -6,6 +6,8 @@ from collections import defaultdict
 from collections.abc import Iterable
 from typing import Any
 
+import numpy as np
+
 from vllm.config import VllmConfig
 from vllm.distributed.ec_transfer.ec_connector.base import (
     ECConnectorMetadata,
@@ -1040,7 +1042,6 @@ class Scheduler(SchedulerInterface):
                 )
 
             stopped = False
-            new_logprobs = None
             new_token_ids = generated_token_ids
             kv_transfer_params = None
             status_before_stop = request.status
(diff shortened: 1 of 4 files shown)


vllm/v1/core/sched/scheduler.py
```diff
@@ -1065,14 +1066,23 @@ def update_from_output(
...
             ):
                 # NOTE: once we support N tokens per step (spec decode),
                 # the outer lists can be of length > 1.
-                new_logprobs = logprobs.slice(req_index, req_index + 1)
+                (
+                    new_logprob_token_ids,
+                    new_logprobs,
+                    new_sampled_token_ranks,
+                    new_cu_num_generated_tokens,
+                ) = logprobs.slice(req_index, req_index + 1)
```
njhill, line 1085: @Jialin I don't understand this one - why not just pass the tuple rather than unpacking it. We are paying nontrivial code complexity cost for this and I don't see how it reduces GC

vllm/v1/outputs.py
```diff
@@ -27,20 +27,17 @@ class LogprobsLists(NamedTuple):
     # Used for slicing the logprobs in cases like speculative
     # decoding where the number of generated tokens may be
     # different for each request.
-    cu_num_generated_tokens: list[int] | None = None
+    cu_num_generated_tokens: np.ndarray | None = None
```
njhill, line 30: This change seems reasonable if it helps

vllm/v1/core/sched/scheduler.py
Jialin: @njhill Each decode cycle would return <BS> EngineCoreOutput to upstream, and if we use LogprobLists / tuples as member variables in EngineCoreOutput, this would double the object allocation count counts.

As GC0 is triggered per (# allocate - # deallocate) count, LogprobLists / tuples would double this counter, so much easier to trigger GC0.

Please let me know if you need more clarification.

But I think the statement might not be plausible enough, I will introduce an e2e GC unit tests to show GC invocation difference. In that way, I should have a more data driven way to show the win. (And that could be used as arenas of GC optimization as well.

Issue #28635, opened 2025-11-13
[Bug]: Streaming=True causes missing or scrambled tokens with GPT-OSS 120B on vLLM v0.11.0


cxz1418:
### Your current environment

| NVIDIA-SMI 550.127.05             Driver Version: 550.127.05     CUDA Version: 12.4     |

* vLLM versions: v0.10.2 and v0.11.0
* Model: GPT-OSS 120B
* Streaming: True
* --enforce-eager: Tested

### 🐛 Describe the bug

Issue:
When using Streaming=True, some tokens are missing or arrive in scrambled order.

* With Streaming=False, all tokens are generated correctly.
* Using --enforce-eager produces the correct token sequence but significantly slows down generation.

This issue occurs in both v0.10.2 and v0.11.0.

Expected behavior:
Streaming should produce all tokens in the correct order, similar to --enforce-eager, without the performance penalty.

Steps to reproduce:

1. Run GPT-OSS 120B with vLLM v0.11.0 (or v0.10.2)
2. Enable streaming (Streaming=True)
3. Generate text and observe missing or scrambled tokens

Additional notes:
The problem appears to be specific to asynchronous streaming. Using eager execution ensures correct token order but reduces speed.

Command Example

```
CUDA_VISIBLE_DEVICES=0,1,2,3 vllm serve openai/gpt-oss-120b \
   --tensor-parallel-size 4 \
   --enable-expert-parallel \
   --tool-call-parser openai \
   --reasoning-parser openai_gptoss \
   --enable-auto-tool-choice \
   --async-scheduling \
   --max-model-len 131072 \
   --gpu-memory-utilization 0.90 \
   --max-num-seqs 32 \
   --host 0.0.0.0 \
   --max-num-batched-tokens 8192 \
   --port 20003
```

---

robertgshaw2-redhat:
Thanks for this report. Do you have any client side that consistent reproduces this?

cxz1418:
@robertgshaw2-redhat 
Yes, We tested it from multiple clients, but the same issue occurs where some tokens are missing in the output.
@njhill 
Disabling --async-scheduling also results in the same behavior.

danladis:
I have the same problem. Both with and without the `--async-scheduling` flag. Also with older and the newest version (v0.11.0)

robertgshaw2-redhat:
[...]

PR #29044, opened 2025-11-19
[BugFix] Fix malformed output with streaming responses API

njhill:
Attempting to fix issues described in https://github.com/vllm-project/vllm/issues/28635.

This was written by claude code, I haven't yet scrutinized the logic.

```
● Perfect! I've identified and fixed the race condition. Here's a summary:

  Root Cause

  The race condition was caused by incomplete implementation in StreamingHarmonyContext.append_tool_output() at vllm/entrypoints/context.py:550. The method had a TODO comment indicating that tool
  output messages should be added to self._messages, but this was never implemented.

  The Bug Flow

  1. Tool execution completes: append_tool_output() processes tool output tokens through the parser and updates self.last_tok
  2. Critical bug: Tool message is NOT added to self._messages (the TODO was never completed)
  3. Next turn begins: render_for_completion() is called to prepare the next prompt
  4. Token mismatch: It renders self._messages (missing the tool output) and tries to find self.last_tok in the rendered tokens
  5. Wrong position: If self.last_tok happens to match a token at the wrong position, it processes tokens from an incorrect location
  6. Parser corruption: The harmony parser receives garbled token sequences and fails with "Unknown role: assistantary"

  The error "assistantary" is likely a result of the parser trying to extract a role from tokens that span across message boundaries or are otherwise malformed due to processing from the wrong
  offset.

  Fixes Applied

  1. Primary fix (vllm/entrypoints/context.py:551): Completed the TODO by adding self._messages.extend(output) to properly track tool output messages
  2. Secondary fix (vllm/entrypoints/context.py:534-535): Added guard to only update self.last_tok if tokens were actually processed, preventing potential NameError if token_ids is empty
  3. Additional fix (vllm/entrypoints/context.py:217): Removed erroneous self.parser = get_streamable_parser_for_assistant() that was resetting the parser state on every append_output() call in
  non-streaming HarmonyContext

  These fixes ensure that the message list stays synchronized with the parser state, preventing token processing from incorrect positions that lead to the race condition errors.
```

### DIFF
--- a/vllm/entrypoints/context.py
+++ b/vllm/entrypoints/context.py
@@ -552,8 +552,11 @@ class StreamingHarmonyContext(HarmonyContext):
         for tok in toks:
             self.parser.process(tok)
         self.last_tok = toks[-1]
-        # Add tool output messages to self._messages
-        self._messages.extend(output)
+        # Add tool output messages from parser to self._messages
+        # (same pattern as append_output)
+        msg_count = len(self._messages) - self.num_init_messages
+        if msg_count < len(self.parser.messages):
+            self._messages.extend(self.parser.messages[msg_count:])
 
     def is_expecting_start(self) -> bool:
         return self.parser.state == StreamState.EXPECT_START
--- a/vllm/entrypoints/context.py
+++ b/vllm/entrypoints/context.py
@@ -214,6 +214,9 @@ class HarmonyContext(ConversationContext):
 
     def append_output(self, output: RequestOutput) -> None:
         output_token_ids = output.outputs[0].token_ids
+        # Reset parser for each append_output call to handle multi-turn scenarios
+        # where the parser needs to start fresh for each assistant response
+        self.parser = get_streamable_parser_for_assistant()
         for token_id in output_token_ids:
             self.parser.process(token_id)
             # Check if the current token is part of reasoning content
--- a/vllm/entrypoints/context.py
+++ b/vllm/entrypoints/context.py
@@ -506,6 +506,8 @@ class StreamingHarmonyContext(HarmonyContext):
         self.encoding = get_encoding()
         self.last_tok = None
         self.first_tok_of_message = True
+        # Track how many tokens have been processed to avoid buggy token search
+        self.processed_token_count = 0
 
     @property
     def messages(self) -> list:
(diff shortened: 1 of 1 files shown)


vllm/entrypoints/context.py
```diff
@@ -214,7 +214,6 @@ def _update_num_reasoning_tokens(self):
 
     def append_output(self, output: RequestOutput) -> None:
         output_token_ids = output.outputs[0].token_ids
-        self.parser = get_streamable_parser_for_assistant()
```
njhill, line 217: @heheda12345 it looks like you added this line in #22512... wonder if you could check this and claude's logic when you have a chance?

vllm/entrypoints/context.py
```diff
@@ -547,7 +549,8 @@ def append_tool_output(self, output: list[Message]) -> None:
         for tok in toks:
             self.parser.process(tok)
         self.last_tok = toks[-1]
-        # TODO: add tool_output messages to self._messages
+        # Add tool output messages to self._messages
+        self._messages.extend(output)
```
bbrowning: This looks like a key fix to any scenario involving server-side builtin tools executed via the Responses API. Without this, the logic in `render_for_completion(self)` of the `StreamingHarmonyContext` would backtrack much further than just the `<|start|>assistant` tokens it is trying to do and would go all the way back to before the tool output, leading to who knows what as far as next tokens and state.

njhill:
I ran the CI on this change in https://github.com/vllm-project/vllm/pull/27614, unfortunately a different test failed before it reached `test_streaming`:
```
[2025-11-20T00:15:34Z] FAILED entrypoints/openai/test_response_api_mcp_tools.py::test_mcp_tool_env_flag_enabled[openai/gpt-oss-20b] - openai.BadRequestError: Error code: 400 - {'error': {'message': 'Unexpected token 815 while expecting start token 200006', 'type': 'BadRequestError', 'param': None, 'code': 400}}
```

vllm/entrypoints/context.py
njhill: Actually in response to the new test failure [below](https://github.com/vllm-project/vllm/pull/29044#issuecomment-3555262490), claude suggested to reinstate this line.

bbrowning: What's not clear to me is why the non-streaming case resets the parser between turns but the streaming case uses the same parser and tries to feed the missing tokens into it to catch it back up with the expected state. On the surface, it feels like both of these should take the same approach - either resetting the parser between turns or doing the extra logic the streaming side does down in streaming's render_for_completion.

But, for the sake of trying to get this going without digging deeper into things here, I think putting this back like it was with the added comment at least makes it clear why it's doing this.

njhill:
Going to close this, the issue turned out to be related to duplicate request ids, see https://github.com/vllm-project/vllm/pull/29355.

PR #29051, opened 2025-11-20
[DO NOT LAND] Prototype Helion kernel in vLLM

gmagogsfm:
Prototype Helion kernels in vLLM

- [x] Reduce boiler plate in authoring/registering Helion kernel
  - [x] Use direct_register_custom_op to reduce overhead 
- [x] Performance tuning
  - [x] allreduce_add_rmsnorm hangs on H200
  - [x] Fine grained dynamic shape control in Helion kernel settings 
  - [x] flashinfer.jit autotuning enabled?
- [x] Add runtime dispatch logic based on model param
- [x] Is it OK for HelionKernelWrapper to be singleton?
- [x] Specialize on dtype?
- [ ] Add a simple wrapper for one-Helion-kernel CustomOp
- [ ] Can Helion guarantee numerical parity with baseline (like eager)?
- [ ] Add batch-size based dispatching
- [ ] Add lowering pass that allow Helion kernel to fuse with other ops

Future improvments:
- [ ] Logging support to help debug Helion kernels, specially while config compilation.
- [ ] Consolidate input generation between autotuning and benchmarking
- [ ] Support out-variant kernel, auto-functionalize
  - [ ] Support more dtypes in fake kernel
- [ ] Automatically generate fake kernel
  - [ ] Helion kernel can't handle non-nested SymInt input (bound kernel cache error when computing cache key with Symint)
  - [ ] Can helion automatically detect mutate_args?
  - [ ] Disable 0/1 specialization in Helion kernel.bind
- [ ] tests should be more flexible in checking helion op names (dynamic)
- [ ] Test infra for mocking Helion configs
- [ ] Benchmark script to support benchmarking multiple hidden sizes at the same time
- [ ] Support quick autotuning mode in autotune framework
- [ ] Support memory footprint comparison in benchmarking
- [ ] use quick autotuning mode by default
- [ ] Worthwhile to fuse across Helion kernel boundary? with other helion kernels or non-helion ops


gmagogsfm:
RMS Norm Quant 8
```
Summary Statistics
============================================================
Total configurations tested: 199

Speedup:
  Average: 1.62x
  Median:  1.59x
  Min:     0.96x
  Max:     2.58x

Latency (ms):
  Baseline - Avg: 0.0040, Min: 0.0022, Max: 0.0190
  Helion   - Avg: 0.0024, Min: 0.0014, Max: 0.0125
============================================================
```

docs/design/helion_kernels_vllm_rfc.md
```diff
@@ -0,0 +1,413 @@
...
+| | 5120 | 0.0122 | 0.0163 | **1.28x** | 1.23x | 0.85x - 2.42x |
+| | 8192 | 0.0181 | 0.0165 | **1.04x** | 0.98x | 0.66x - 2.08x |
+| **rms_norm_fp8** | 2048 | 0.0050 | 0.0057 | **1.08x** | 1.00x | 0.66x - 2.12x |
+| | 4096 | 0.0050 | 0.0059 | **1.11x** | 1.00x | 0.66x - 2.69x |
+| | 5120 | 0.0050 | 0.0057 | **1.08x** | 1.00x | 0.66x - 2.14x |
+| | 8192 | 0.0054 | 0.0056 | **1.03x** | 1.00x | 0.41x - 1.94x |
+| **allreduce_add_rmsnorm** | 2048 | 0.1365 | 0.1283 | **0.92x** | 0.86x | 0.81x - 2.29x |
+| | 4096 | 0.1297 | 0.1256 | **0.94x** | 0.88x | 0.83x - 2.46x |
+| | 5120 | 0.1338 | 0.1276 | **0.93x** | 0.87x | 0.82x - 2.40x |
+| | 8192 | 0.1350 | 0.1274 | **0.92x** | 0.86x | 0.81x - 2.42x |
```
mgoin, line 475: It seems like the "Baseline Avg (ms)" is the same time for all hidden sizes, so i think something is wrong there. We must be benchmarking overhead instead of the kernel

gmagogsfm: Good point, will look into it.

gmagogsfm: @mgoin I updated benchmark script to use [triton's benchmark API](https://triton-lang.org/main/python-api/generated/triton.testing.do_bench_cudagraph.html) for non-distributed ops, indeed some overhead was removed and Helion exhibited slightly higher speedups. I updated the RFC with latest numbers.

gmagogsfm: As for why "baseline avg" is the same for all hidden sizes, that is because they share the same baseline. "hidden size" column indicates what hidden size was used to autotune the kernel, they all share the same set of benchmark input sizes to simulate the worst case scenario (like when we don't have the right config for a input shape)

vllm/compilation/helion/benchmark.py
```diff
@@ -0,0 +1,1391 @@
...
+        # Fallback to manual implementation for:
+        # - Multi-GPU scenarios (distributed=True)
+        # - Non-CUDA graph timing
+        # - When Triton fails or is disabled
+        return KernelBenchmark._manual_time_kernel(
+            kernel_fn, num_iterations, warmup, use_cudagraph, distributed
+        )
+
+    @staticmethod
+    def _manual_time_kernel(
```
gmagogsfm, line 326: Here

hmellor:
Closing: superseded by the landed Helion integration in vllm/kernels/helion (#32740, #32964, #33203, #33373), and this prototype's vllm/compilation/helion/ tree no longer exists on main.

PR #29057, opened 2025-11-20
[torch.compile] split shared compilation hash helpers

ghost:
## Motivation
PR: https://github.com/vllm-project/vllm/pull/26468 removed `envs.compute_hash` which was being used by compilation_config_hash_factors, which [decorators.py](https://github.com/vllm-project/vllm/blob/main/vllm/compilation/decorators.py#L357) relies on for aot cache keys.

Both `_compute_code_hash` and `compilation_config_hash_factors` exist to keep aot and jit hashing consistent. `backends.py` had a duplicate of their logic

### The Fix
This pr creates a shared helper instead of inlining them and updates the helper itself to call the new `envs.compile_factors()`. 

Thanks for [pointing this out](https://github.com/vllm-project/vllm/pull/26468#discussion_r2543823745) @zhxchen17 

### Test
@zhxchen17 does this fix the issue with pytorch nightly?

cc @ProExpertProg @WorldExplored


vllm/compilation/caching.py
```diff
@@ -134,11 +135,15 @@ def co_name(self):
         return "VllmSerializableFunction"
 
 
-def compilation_config_hash_factors(vllm_config: VllmConfig) -> list[str]:
+def compilation_config_hash_factors(
```
ProExpertProg: This utility method is not the right abstraction, let's just inline it for now, or you can organize the code better if you have a better idea.

ProExpertProg (commented):
Can we instead refactor the decorator to use the new & cleaner approach instead of restoring the older code?

zhxchen17:
> @zhxchen17 i tried to hit ur concern here:
> 
> > In other words, we don't have the same hashing algorithm between VLLM_USE_AOT_COMPILE=1 and VLLM_USE_AOT_COMPILE=0. I wonder does anyone think that will be a problem?

@vnadathur Deleted my original comment there since I think that's not accurate. 

@ProExpertProg I like this fix slightly better because it shares the code for code hashing. I think in the short term it's okay to remove compilation_config_hash_factors tho. I wonder whether it will be a problem if in the future we decide to introduce a new common source of hashing (in addition to env+compiler config) that only got added to one path but not the other one. If we think it's unlikely, I think it's fine to have small duplication.

vllm/compilation/caching.py
```diff
@@ -133,6 +135,24 @@ def co_name(self):
...
+    """
+    Return the hashed environment factors and vLLM config hash.
+
+    Both AOT and JIT cache paths rely on this helper to ensure their cache keys
+    stay in sync. When ``env_factors`` is omitted the helper will gather them
+    via ``envs.compile_factors()``.
+    """
+
+    if env_factors is None:
+        env_factors = envs.compile_factors()
```
ProExpertProg: We should either always or never pass this in

ProExpertProg (changes requested):
I think if we're cleaning this up, let's really clean it up. I would prefer to land #29058 for now. I agree it would be good to keep the two cache hashes in sync but they are already different and the only common part is hash of config and envs, both of which come from the same member methods on the config and envs already.

ProExpertProg (approved):
This looks fine as well, @zou3519 @drisspg @zhxchen17 can you check this works for you, happy to merge as is and improve structure later

ProExpertProg (changes requested):
Let's merge #29058 for now. After that we can properly restructure the config, including pulling out the compile cache factors so they can be stored with other cache artifacts

PR #29058, opened 2025-11-20
Fixes bench

drisspg:
## Purpose
Fix da thingsss
## Test Plan
`vllm bench throughput --input-len 1024`

## Test Result

---
<details>

</details>


zhxchen17:
@ProExpertProg @zou3519 I think this is fine for fixing the issue.

There'll be one subtle implication to do the fix this way: If we do VLLM_USE_AOT_COMPILE=0 to compile the model first, the inductor compile cache will be saved to `~/.cache/vllm/torch_compile_cache/<hash1>/rank_x_y`, and then when we run vllm again with VLLM_USE_AOT_COMPILE=1, inductor compile cache will be saved to `~/.cache/vllm/torch_compile_cache/<hash2>/rank_x_y`, and there's no guarantee now that `hash1` and `hash2` are equal.

In other words, we don't have the same hashing algorithm between VLLM_USE_AOT_COMPILE=1 and VLLM_USE_AOT_COMPILE=0. I wonder does anyone think that will be a problem? In my understanding, this shouldn't be a big issue because we could expect user to always stick one config in production? If this is the case I think we should land this fix and keep it this way.

zhxchen17:
I think the current code works but to make type checker happier we could follow gemini's advice.

drisspg:
@ProExpertProg yup yup closing in favor of: https://github.com/vllm-project/vllm/pull/29057

ProExpertProg (approved):

zou3519 (approved):

PR #29064, opened 2025-11-20
[CI/Build] Make test_attention_selector.py run tests on correct platform

rasmith:
This PR fixes test_attention_selector.py to make sure that tests run only correct platform, so if `hip `or `cuda` is selected, the test will only run on the corresponding platform, since the `get_attn_backend`  needs to be on the correct platform to work correctly.

### DIFF
--- a/tests/kernels/attention/test_attention_selector.py
+++ b/tests/kernels/attention/test_attention_selector.py
@@ -50,13 +50,9 @@ DEVICE_MLA_BLOCK_SIZES = {
 def generate_params():
     is_rocm = current_platform.is_rocm()
     params = []
+    device_list = ["cuda", "cpu"] if not is_rocm else ["hip", "cpu"]
     for use_mla in [True, False]:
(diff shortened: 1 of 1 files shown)


SageMoore (approved):

yewentao256 (approved):
LGTM, thanks for the work!

tests/kernels/attention/test_attention_selector.py
```diff
@@ -78,6 +79,10 @@ def test_env(
     block_size: int,
     monkeypatch: pytest.MonkeyPatch,
 ):
+    if device == "hip" and not current_platform.is_rocm():
```
gshtras: Maybe make the for device in ["cuda", "hip", "cpu"]: conditional on the current_platform?

rasmith: Done

gshtras (approved):

PR #29066, opened 2025-11-20
[MoE][Refactor] Remove most arguments to FusedMoEMethodBase.apply

bnellnm:
## Purpose
After making `select_experts` a non-static method (https://github.com/vllm-project/vllm/pull/29067), we can avoid passing most of the arguments to `FusedMoEMethodBase.apply` and get them from the layer directly.

This doesn't really decrease flexibility, the MoE quantization methods were already fairly tightly coupled with `FusedMoE`.  A further step could be making `FusedMoE` an abstract base class that simply provides the `quant_method` parameters.  Then the current `FusedMoE` class would become a subclass of that.

## Test Plan
CI

## Test Result
Blackwell failure is known bug fixed by https://github.com/vllm-project/vllm/pull/30336

cc @varun-sundar-rabindranath 

---
<details>

</details>


LucasWilkinson:
Looks like we need to land https://github.com/vllm-project/vllm/pull/29067 first? marking draft until that lands :+1:

LucasWilkinson (approved):
LGTM; thanks for doing this! this is amazing!

bnellnm:
> LGTM; thanks for doing this! this is amazing!

Thanks for the review!

vllm/model_executor/layers/quantization/rtn.py
```diff
@@ -394,10 +376,10 @@ def apply(
...
             topk_ids,
             quant_type_id=self.quant_config.quant_type.id,
-            apply_router_weight_on_input=apply_router_weight_on_input,
-            global_num_experts=global_num_experts,
-            expert_map=expert_map,
-            workspace=workspace,
+            apply_router_weight_on_input=layer.apply_router_weight_on_input,
+            global_num_experts=layer.global_num_experts,
+            expert_map=layer.expert_map,
+            workspace=workspace,  # ???????
```
tlrmchlsmth: Could we remove this comment before landing?

tlrmchlsmth: I'll push a commit to remove

mgoin (approved):
Great work Bill! Thinking now on the state, the main downside I see of this move is that since we aren't passing arguments into the function, it is less explicit for each apply method to assert against unsupported modes as we add more. Hopefully we can move to a system like the attention backends for each kernel backend

PR #29067, opened 2025-11-20
[MoE][Refactor] Make select_experts a non-static method

bnellnm:
## Purpose
There didn't appear to be a good reason why `select_experts` was a static method.  Making it non-static allows access to the attributes of the `FusedMoE` layer so that we don't need to pass so many arguments around.  This can lead to removing most of the arguments to `FusedMoEMethodBase.apply` (see https://github.com/vllm-project/vllm/pull/29066).

## Test Plan
CI

## Test Result

cc @varun-sundar-rabindranath 

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/layers/fused_moe/layer.py
+++ b/vllm/model_executor/layers/fused_moe/layer.py
@@ -1469,30 +1469,13 @@ class FusedMoE(CustomOp):
             logits_shape, dtype=moe.in_dtype, device=torch.cuda.current_device()
         )
 
-    @staticmethod
     def select_experts(
+        self,
         hidden_states: torch.Tensor,
         router_logits: torch.Tensor,
-        top_k: int,
-        use_grouped_topk: bool,
-        renormalize: bool,
-        topk_group: int | None = None,
-        num_expert_group: int | None = None,
-        custom_routing_function: Callable | None = None,
-        scoring_func: str = "softmax",
-        routed_scaling_factor: float = 1.0,
-        e_score_correction_bias: torch.Tensor | None = None,
-        indices_type: torch.dtype | None = None,
-        enable_eplb: bool = False,
-        expert_map: torch.Tensor | None = None,
-        expert_load_view: torch.Tensor | None = None,
-        logical_to_physical_map: torch.Tensor | None = None,
-        logical_replica_count: torch.Tensor | None = None,
-        global_num_experts: int | None = None,
         zero_expert_num: int | None = None,
         zero_expert_type: str | None = None,
-        num_fused_shared_experts: int = 0,
-    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor | None]:
         """
         Route the input hidden states to the top-k experts based on the
         router logits.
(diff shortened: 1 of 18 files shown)


vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1416,30 +1416,11 @@ def ensure_dp_chunking_init(self):
...
-        expert_map: torch.Tensor | None = None,
-        expert_load_view: torch.Tensor | None = None,
-        logical_to_physical_map: torch.Tensor | None = None,
-        logical_replica_count: torch.Tensor | None = None,
-        global_num_experts: int | None = None,
-        zero_expert_num: int | None = None,
-        zero_expert_type: str | None = None,
-        num_fused_shared_experts: int = 0,
-    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor | None]:
```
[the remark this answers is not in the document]
LucasWilkinson: @bnellnm is this real?

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1472,97 +1439,112 @@ def select_experts(
...
+            if self.quant_method.supports_eplb:
+                assert self.expert_load_view is not None
+                assert self.logical_to_physical_map is not None
+                assert self.logical_replica_count is not None
+            else:
+                raise NotImplementedError(
+                    f"EPLB is not supported for {self.quant_method.__class__.__name__}."
+                )
+
+        # ?
```
LucasWilkinson: nit cruft

LucasWilkinson (approved):
Overall looks like a very nice cleanup! thanks! LGTM, left a couple nits

vllm/model_executor/layers/fused_moe/layer.py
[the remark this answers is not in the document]
bnellnm: Yeah, I've fixed these but haven't updated the PR yet.

vllm/model_executor/layers/fused_moe/layer.py
bnellnm: The last output of `select_experts` is for the zero expert output.  In most cases it is `None`.  Hopefully, this will get cleaned up by https://github.com/vllm-project/vllm/pull/28891

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1511,78 +1492,95 @@ def select_experts(
             fused_topk_bias,
         )
 
+        if self.enable_eplb:
+            if self.quant_method.supports_eplb:
```
mgoin, line 1496: How does this handle methods where not all kernels support eplb?

bnellnm: Right now, for all the quant_methods that support eplb, all the implementation kernels support it also.  I've left individual eplb checks in the various apply methods for paths that don't.  I'll double check the logic though.

bnellnm: Currently only three quant methods support eplb and they all seems to unconditionally support it (at least according to the code in main).  The rest do not support eplb at all.  If there happens to be a mixed quant method where some kernels support eplb and some don't, then `supports_eplb` would need to be true and manual checks for the non-supporting kernels would need to be added.

PR #29068, opened 2025-11-20
[Model] Add OpenCUA-7B support

lim4349:
## Purpose
Add support for OpenCUA-7B, a multimodal model based on Qwen2.5-VL that uses 1D-RoPE instead of M-RoPE for the vision encoder. This PR enables vLLM to load and run inference with the OpenCUA-7B model from HuggingFace (`xlangai/OpenCUA-7B`).

### 1. Registry Import Test
✅ **PASSED** - Model class can be imported and registered successfully

The model is properly registered in the vLLM model registry:
- `vllm/model_executor/models/registry.py`: Added to `_MULTIMODAL_MODELS` dictionary
- `tests/models/registry.py`: Added test example model entry

### 2. API Inference Test with System Prompt
✅ **PASSED** - Model successfully processes multimodal inputs via OpenAI-compatible API

**Test Command:**
```bash
curl -X POST "http://localhost:8000/v1/chat/completions" \
  -H "Authorization: Bearer EMPTY" \
  -d '{
    "model": "xlangai/OpenCUA-7B",
    "messages": [
      {
        "role": "system",
        "content": "You are a GUI agent. You are given a task and a screenshot of the screen. You need to perform a series of pyautogui actions to complete the task."
      },
      {
        "role": "user",
        "content": [
          {"type": "image_url", "image_url": {"url": "https://..."}},
          {"type": "text", "text": "Close file explorer"}
        ]
      }
    ],
    "max_tokens": 512,
    "temperature": 0
  }'
```
## Test Result
- ✅ Model successfully loads and processes requests
- ✅ System prompt is correctly handled
- ✅ Image input is properly processed via multimodal pipeline
- ✅ Response generation works correctly
- ✅ Output format matches expected structure (e.g., `pyautogui.click(x=1428, y=264)`)
- ✅ Input/output format is identical to the official HuggingFace Spaces demo ([xlangai/OpenCUA-demo](https://huggingface.co/spaces/xlangai/OpenCUA-demo))
---
<details>

</details>


vllm/model_executor/models/opencua.py
```diff
@@ -0,0 +1,471 @@
...
+    dummy_inputs=OpenCUADummyInputsBuilder,
+)
+class OpenCUAForConditionalGeneration(
+    nn.Module,
+    SupportsMultiModal,
+    SupportsLoRA,
+    SupportsPP,
+    SupportsQuant,
+    SupportsEagle3,
+):
```
Isotr0py:
```suggestion
class OpenCUAForConditionalGeneration(Qwen2_5_VLForConditionalGeneration):
```
I think we can simplify the model implementation here.

vllm/model_executor/models/opencua.py
```diff
@@ -0,0 +1,471 @@
...
+                    dtype=torch.get_default_dtype(),
+                    attn_backend_override=attn_backend_override,
+                )
+                self.attn_backend, self.flash_attn_varlen_func = (
+                    maybe_get_vit_flash_attn_backend(
+                        self.attn_backend,
+                        False,
+                        attn_backend_override=attn_backend_override,
+                    )
+                )
```
Isotr0py: Our FA fork has supported head_size=72 and 80, so I think this is no longer necessary.

vllm/transformers_utils/configs/opencua.py
Isotr0py, line 1: Seems it's not very necessary to add this config. We can just pass `truust_remote_code=True` to use config in model repo.

vllm/model_executor/models/opencua.py
lim4349: Thanks for your comment. modified

vllm/model_executor/models/opencua.py
lim4349: Thanks. I removed it.

vllm/model_executor/models/opencua.py
```diff
@@ -0,0 +1,415 @@
...
+            quant_config=quant_config,
+            prefix=prefix,
+            use_data_parallel=use_data_parallel,
+            attn_backend_override=attn_backend_override,
+        )
+
+
+class OpenCUAProcessingInfo(Qwen2VLProcessingInfo):
+    def get_hf_config(self):
+        return self.ctx.get_hf_config(None)
```
DarkLight1337: I think there's no need to override this if we just load the config from HF Hub given `trust_remote_code=True`

vllm/model_executor/models/opencua.py
jeejeelee: Because this model  directly inherits from `Qwen2_5_VLForConditionalGeneration`, functions like `get_mm_mapping`, `load_weights`, `compute_logits` etc., and attributes like `packed_modules_mapping`, `hf_to_vllm_mapper`, etc. don't need to be implemented.

vllm/transformers_utils/configs/opencua.py
lim4349: Thanks.

vllm/model_executor/models/opencua.py
lim4349: I tried removing this override, but the model fails during initialization.
The base Qwen2VLProcessingInfo.get_hf_config() enforces that the config must be a Qwen2VLConfig, while OpenCUA uses its own OpenCUAConfig. As a result, removing the override produces:

TypeError: Invalid type of HuggingFace config.
Expected Qwen2VLConfig, found OpenCUAConfig

So it looks like the Qwen2-VL path is too strict for this model, and overriding get_hf_config() is necessary to return the correct OpenCUA config.

DarkLight1337:
Ok let's keep the override then, thanks for testing

vllm/model_executor/models/opencua.py
```diff
@@ -0,0 +1,418 @@
...
+    WeightsMapper,
+    init_vllm_registered_model,
+    maybe_prefix,
+)
+from .vision import (
+    run_dp_sharded_mrope_vision_model,
+)
+
+
+class OpenCUAVisionTransformer(Qwen2_5_VisionTransformer):
```
DarkLight1337: I think we don't need to create a subclass if it's the exact same. Just use `import ... as ...` syntax to rename the imported class

vllm/model_executor/models/opencua.py
```diff
@@ -0,0 +1,396 @@
...
+            image_processor=image_processor,
+            tokenizer=tokenizer,
+            video_processor=video_processor,
+            chat_template=chat_template,
+            **kwargs,
+        )
+
+        self.image_token = "<|media_placeholder|>"
+
+    def __call__(
```
DarkLight1337, line 99: QQ: Why is it necessary to override `__call__`?

vllm/model_executor/models/opencua.py
```diff
@@ -0,0 +1,396 @@
...
+        )
+
+        self.make_empty_intermediate_tensors = (
+            self.language_model.make_empty_intermediate_tensors
+        )
+
+    def get_language_model(self) -> torch.nn.Module:
+        return self.language_model
+
+    def set_aux_hidden_state_layers(self, layers: tuple[int, ...]) -> None:
```
DarkLight1337: If the override is the same as the base class then I think no need to include them here

lim4349:
@DarkLight1337 

Thanks for the tip I removed the unnecessary subclass and used an alias instead — the code is much cleaner now.

Just for context: we still need this minimal `__call__` override because OpenCUA uses `<|media_placeholder|>` and relies on `_get_prompt_updates` later, while the upstream processor enforces `image`/`video` token checks and raises:

`ValueError: Failed to apply OpenCUAProcessor on data={'text': '<|media_placeholder|>'}`

This wrapper bypasses that check and works with OpenCUA's flow.

Really appreciate the feedback!

vllm/transformers_utils/configs/opencua.py
```diff
@@ -0,0 +1,46 @@
+# SPDX-License-Identifier: Apache-2.0
```
DarkLight1337: Did you forget to remove this?

DarkLight1337: I mean that this whole config isn't necessary to be defined in vLLM if you use `trust_remote_code=True`

DarkLight1337 (approved):
Thanks for your patience!

DarkLight1337 (approved):
Sorry for the delay, let's see if the tests pass now

PR #29071, opened 2025-11-20
[BugFix] Fix chunked prompt logprobs + preemption

njhill:
Thanks @zhuohan123 for helping discover this.

TPU changes are just for consistency (TPU doesn't support prompt logprobs).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -845,6 +845,7 @@ class VllmRunner:
     @staticmethod
     def _final_steps_generate_w_logprobs(
         req_outputs: list[RequestOutput],
+        include_prompt_token_ids: bool = False,
     ) -> list[TokensTextLogprobsPromptLogprobs]:
         outputs: list[TokensTextLogprobsPromptLogprobs] = []
         for req_output in req_outputs:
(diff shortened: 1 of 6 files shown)


njhill:
> 
> With `num_prompt_logprobs` now retained on the runner, resumed requests still enter `_get_prompt_logprobs` with `start_idx = request.num_computed_tokens`, meaning previously processed prompt tokens are skipped when filling the logprob tensor (see `gpu_model_runner.py` around lines 3489–3492). But `InputBatch.remove_request` still clears `in_progress_prompt_logprobs_cpu` on every preemption, so when a chunked prefill is preempted the partial prompt logprob buffer is dropped and the resumed request continues from a fresh tensor, leaving all earlier prompt positions uninitialized/missing. Any prompt that requires logprobs and is chunked then preempted will still return incorrect prompt logprobs despite this fix.
> 

This is incorrect - num_computed_tokens is reset to 0 when a request is preempted, so any chunks of prompt logprobs already computed will be re-computed.

mgoin (approved):
LGTM 🙏

PR #29074, opened 2025-11-20
[Bugfix] Missing tokens in `return_token_ids` when tool parsers is enabled in streaming mode

### LINKED ISSUE #27482: [Bug]: `return_token_ids` missing tokens when using tool calls
### Your current environment

Testing with latest vLLM builds from main, as of Fri Oct 24th 2025 (when this bug was opened).

### 🐛 Describe the bug

The `return_token_ids` parameter that is supposed to return all generated token ids back to the client is missing quite a few tokens for Chat Completion streaming requests that result in tool calls being generated. Exactly how many and where they are missing in the request will depend on the tool call parser in use as well as the exact request format.

Here's a minimal reproducer.

First, run vLLM with a tool call parser and model. I use a Granite model for testing here, but it should be roughly the same for any model with a tool call parser.

```
vllm serve ibm-granite/granite-3.3-8b-instruct \
  --enable-auto-tool-choice \
  --tool-call-parser granite
```

Then, send a streaming tool call request to the server and check the response for missing tokens:

```python
from openai import OpenAI

client = OpenAI(base_url="http://localhost:8000/v1", api_key="fake")

tools = [
    {
        "type": "function",
        "function": {
            "name": "get_current_weather",
            "description": "Get the current weather in a given location",
            "parameters": {
                "type": "object",
                "properties": {
                    "location": { "type": "string", "description": "The city, e.g. San Francisco, CA" },
                    "unit": { "type": "string", "enum": ["celsius", "fahrenheit"] }
                },
                "required": ["location"]
            }
        }
    },
]
response = client.chat.completions.create(
    model="ibm-granite/granite-3.3-8b-instruct",
    messages=[{"role": "user", "content": "What is the weather in Sydney in celsius?"}],
    tools=tools,
    tool_choice="auto",
    stream=True,
    stream_options={
        "include_usage": True,
        "continuous_usage_stats": True,
    },
    extra_body={"return_token_ids": True},
)

returned_token_ids = []
last_completion_tokens = 0
for event in response:
    if not getattr(event, "choices", None):
        continue
    choice = event.choices[0]
    usage = event.usage
    if hasattr(choice, "token_ids"):
        returned_token_ids.extend(choice.token_ids)
        num_token_ids = len(choice.token_ids)
    else:
        num_token_ids = 0
    elapsed_completion_tokens = usage.completion_tokens - last_completion_tokens
    if elapsed_completion_tokens != num_token_ids:
        raise ValueError(
            "Model generated more tokens than returned by return_token_ids!\n"
            f"All tokens returned so far: {returned_token_ids}"
        )
    last_completion_tokens = usage.completion_tokens
```

Running that, I get the following output:

```
python return_token_ids_test.py
  File "/Volumes/SourceCode/vllm/return_token_ids_test.py", line 49, in <module>
    raise ValueError(
ValueError: Model generated more tokens than returned by return_token_ids!
All tokens returned so far: [49154, 48685]
```

If I add a bit of debug logging into vLLM server side and run it again, I can see the list of tokens that should have been returned:

`current_token_ids: [49154, 7739, 8299, 563, 3447, 2645, 563, 313, 16716, 6161, 910, 392, 313, 2243, 563, 313, 3308, 101, 3263, 3918, 313, 426, 563, 313, 371, 81, 1700, 81, 15859, 48685]`

All of the tokens between the first and last in that list were missed by `return_token_ids`.

This code is not executed for every generated token when tool call parser (or reasoning parsers, most likely) are in use: https://github.com/vllm-project/vllm/blob/61089465a6101790635ed96c26df3e9a57d8d2c9/vllm/entrypoints/openai/serving_chat.py#L1090

The reason is because we return early at: https://github.com/vllm-project/vllm/blob/61089465a6101790635ed96c26df3e9a57d8d2c9/vllm/entrypoints/openai/serving_chat.py#L1063

Peng-YM:
# Fix: Missing tokens in `return_token_ids` when using tool parsers in streaming mode

## Purpose

Fix missing tokens in `return_token_ids` response when using tool parsers in streaming mode.

This PR resolves the issue where tool call control tokens (e.g., `<tool_call>`, `</tool_call>`) and JSON structure tokens (e.g., `"name":`, `"arguments":`) are mysteriously missing from the `token_ids` field in chat completion responses when using tool parsers with `return_token_ids=True` in streaming mode.

**Related Issue**: [If there's an existing issue, link it here]

## Problem Background

When users enable `return_token_ids=True` with tool parsers in streaming chat completion, they expect to receive all generated tokens in the response. However, a significant portion of tokens related to tool calls are missing from the `token_ids` field.

**This issue affects ALL tool parsers**, not just Hermes. Any tool parser that returns `None` during token processing will cause tokens to be dropped from the response.

### Example of Missing Tokens

For a tool call that should generate:
```
<tool_call>
{"name": "add_numbers", "arguments": {"a": 15, "b": 27}}
</tool_call>
```

The actual `token_ids` response only contains tokens for `",": 27}}`, missing:
- `<tool_call>` tag tokens
- `"name": "add_numbers"` tokens  
- `"arguments":` key tokens
- Most of `{"a": 15, "b":` tokens
- `</tool_call>` tag tokens

## Root Cause Analysis

The issue occurs in the streaming response handling logic in `serving_chat.py` (lines 1066-1070):

```python
if delta_message is None:
    if output.finish_reason is None:
        continue  # Skips current token without streaming
    else:
        delta_message = DeltaMessage()
```

When tool parsers process tool call tokens, they frequently return `None` in various scenarios. This is a common pattern across **all tool parser implementations**:

#### 1. Hermes Tool Parser (`hermes_tool_parser.py`)
- **Malformed JSON during parsing** (`hermes_tool_parser.py:323-324`):
  ```python
  except partial_json_parser.core.exceptions.MalformedJSON:
      return None  # Causes token to be skipped
  ```

- **Tool name not yet available** (`hermes_tool_parser.py:332-333`, `349-350`):
  ```python
  if not self.current_tool_name_sent:
      if current_tool_call is None:
          return None
  ```

- **Delta text not found in arguments** (`hermes_tool_parser.py:427-428`):
  ```python
  if delta_text not in cur_arguments_json:
      return None
  ```

#### 2. Mistral Tool Parser (`mistral_tool_parser.py`)
- Returns `None` when processing control tokens
- Returns `None` during partial JSON parsing

#### 3. Llama Tool Parser (`llama_tool_parser.py`)
- Returns `None` when tool call start/end tokens are being buffered
- Returns `None` during JSON validation

#### 4. InternLM2 Tool Parser (`internlm2_tool_parser.py`)
- Returns `None` when parsing incomplete tool call structures
- Returns `None` during argument extraction

#### 5. Other Tool Parsers
Similar patterns exist in:
- `qwen3xml_tool_parser.py`
- `deepseekv3_tool_parser.py`
- `granite_tool_parser.py`
- `jamba_tool_parser.py`
- And many others...

When `delta_message` is `None`, the code executes `continue`, which skips sending a chunk for that token. This means the token is never included in any response chunk, causing it to be permanently lost from the `token_ids` field.

**Important**: This issue only affects streaming mode. In non-streaming mode, `token_ids` are returned directly from `output.token_ids` without tool parser processing, so all tokens are preserved.

## Fix Implementation

The fix modifies the streaming logic to send a chunk with `token_ids` even when `delta_message` is `None`:

```python
if delta_message is None:
    if output.finish_reason is None:
        # If return_token_ids is enabled, we still need to
        # send a chunk with token_ids even if delta_message is None
        # to ensure all tokens are included in the response
        if request.return_token_ids:
            delta_message = DeltaMessage(content="")
        else:
            continue
    else:
        delta_message = DeltaMessage()
```

### Key Changes:
1. When `delta_message` is `None` and `output.finish_reason` is `None` (token still being generated)
2. Check if `request.return_token_ids` is `True`
3. If yes, create a `DeltaMessage(content="")` instead of skipping
4. This ensures the chunk is sent with `token_ids` populated (lines 1101-1105)

### Why This Works:
- The empty `DeltaMessage(content="")` satisfies the non-None check
- The subsequent code (lines 1101-1105) still populates `token_ids` based on `request.return_token_ids`
- Clients receive all token IDs, even for control tokens that don't have visible content
- The content field being empty string is acceptable since these are control tokens anyway
- Backward compatibility is maintained: users not using `return_token_ids` see no change
---

## Checklist

- [x] The purpose of the PR is clearly stated
- [x] Root cause analysis provided
- [x] Test plan included with expected results
- [x] Backward compatibility maintained
- [x] No breaking changes to existing API
- [ ] Manual testing completed (to be done by reviewer/maintainer)
- [ ] Existing tests pass (to be verified by CI)


Peng-YM:
> @hmellor
> 
> Can you help review this PR? The `return_token_ids` feature is currently relied upon by the Agent-Lightning project. This bug causes severe Training-Rollout Mismatch issues in RL training, leading to rapid model training crashes. We hope to merge this PR as soon as possible.
> 
> Corresponding Agent-Lightning issue: [microsoft/agent-lightning#326](https://github.com/microsoft/agent-lightning/issues/326)

Also cc: @aarnphm @chaunceyjiang

LucasWilkinson:
cc @aarnphm

vllm/entrypoints/openai/serving_chat.py
```diff
@@ -1065,7 +1065,13 @@ async def chat_completion_stream_generator(
                     #   get the next token without streaming a chunk
                     if delta_message is None:
                         if output.finish_reason is None:
-                            continue
+                            # NOTE: If return_token_ids is enabled, we still need to
+                            # send a chunk with token_ids even if delta_message is None
+                            # to ensure all tokens are included in the response
+                            if request.return_token_ids:
+                                delta_message = DeltaMessage(content="")
```
aarnphm:
```suggestion
                                delta_message = DeltaMessage()
```

aarnphm (approved):
nit

vllm/entrypoints/openai/serving_chat.py
hmellor: If this change is committed, then this whole if block could be greatly simplified

Peng-YM: Thank you for your suggestion! This modification can indeed effectively reduce the complexity of this `if` block.

Peng-YM:
@aarnphm 

Sorry, I'm submitting a PR to vLLM for the first time. Before merging this PR, what do I need to do?

PR #29075, opened 2025-11-20
[CI Failure] Add working_dir to Kernels DeepGEMM Test (H100)

hl475:
## Purpose

From nightly https://buildkite.com/vllm/ci/builds/39860/steps/canvas?sid=019aa011-0c6b-41b7-a847-2a74913e51b1 , this test is failing due to `ERROR: file or directory not found: tests/kernels/quantization/test_block_fp8.py`. According to https://github.com/vllm-project/vllm/blob/main/.buildkite/test-pipeline.yaml#L26 , we can override `working_dir`

## Test Result

---
<details>

</details>

### DIFF
--- a/.buildkite/test-pipeline.yaml
+++ b/.buildkite/test-pipeline.yaml
@@ -551,6 +551,7 @@ steps:
     - pytest -v -s kernels/mamba
 
 - label: Kernels DeepGEMM Test (H100)
+  working_dir: "/vllm-workspace/"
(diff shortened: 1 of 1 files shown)


khluu:
> Thanks for the fix! I wonder if we should notify code author if the test is been added but never success? cc @khluu
I think in this case, the test was added as optional so it didn't run until nightly.

PR #29083, opened 2025-11-20
[Bugfix] Fix logprobs to support rank sequences and empty slices

xrwang8:
## Purpose

This PR fixes two critical issues in the logprobs module:

1. **Support rank sequences**: The `append_logprobs_for_next_position()` function now accepts either a single rank (int) or a sequence of ranks (list/tuple). This is required for V1 engine's spec decoding and prompt logprobs features, which pass rank arrays from numpy tensors.

2. **Fix empty slice bug**: `FlatLogprobs.__getitem__()` now correctly handles empty slices (e.g., `logprobs[1:1]`) without raising `IndexError`.

**Additional improvements:**
- Improved type checking to use `isinstance(rank, (list, tuple))` instead of the less precise `Iterable` check, avoiding issues with strings
- Updated `append_fast()` type signature to accept `Iterable[int | None]`
- Added comprehensive test coverage including error case validation

## Test Plan

```bash
pytest tests/test_logprobs.py -v
```

### DIFF
--- a/vllm/logprobs.py
+++ b/vllm/logprobs.py
@@ -193,7 +193,9 @@ def append_logprobs_for_next_position(
     # Accept either a single sampled-token rank (legacy path) or
     # a full sequence of ranks (e.g., prompt logprobs already provide
     # per-token ranks). When provided, enforce 1:1 with token_ids.
-    if isinstance(rank, (list, tuple)):
+    if not isinstance(rank, int):
+        if isinstance(rank, str):
+            raise TypeError("rank cannot be a string")
         ranks = list(rank)
         if len(ranks) != len(token_ids):
             raise ValueError(
--- a/vllm/logprobs.py
+++ b/vllm/logprobs.py
@@ -75,7 +75,7 @@ class FlatLogprobs(MutableSequence[LogprobsOnePosition]):
         self,
         token_ids: list[int],
         logprobs: list[float],
-        ranks: itertools.chain[int],
+        ranks: Iterable[int | None],
         decoded_tokens: Iterable[str | None],
     ) -> None:
         """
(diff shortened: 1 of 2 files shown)


vllm/logprobs.py
```diff
@@ -119,13 +119,19 @@ def __getitem__(self, index: int | slice):
                 for i in range(self.start_indices[index], self.end_indices[index])
             }
         elif isinstance(index, slice):
-            min_index = self.start_indices[index][0]
-            max_index = self.end_indices[index][-1]
+            start_indices = self.start_indices[index]
+            end_indices = self.end_indices[index]
```
njhill, line 123: can move this below the `if`

vllm/logprobs.py
```diff
@@ -177,17 +183,31 @@ def append_logprobs_for_next_position(
...
-    # into a dictionary twice is the same as doing it once.
-    topk_ranks = range(1, num_logprobs + 1)
-    ranks = itertools.chain((rank,), topk_ranks)
+
+    # Accept either a single sampled-token rank (legacy path) or
+    # a full sequence of ranks (e.g., prompt logprobs already provide
+    # per-token ranks). When provided, enforce 1:1 with token_ids.
+    if not isinstance(rank, int):
+        if isinstance(rank, str):
+            raise TypeError("rank cannot be a string")
```
njhill, line 198: I don't think this check is needed, we only call this method internally

vllm/logprobs.py
```diff
@@ -177,17 +183,31 @@ def append_logprobs_for_next_position(
...
+        ranks = list(rank)
+        if len(ranks) != len(token_ids):
+            raise ValueError(
+                f"Expected {len(token_ids)} ranks for {len(token_ids)} token_ids, "
+                f"got {len(ranks)}."
+            )
```
njhill, line 204: I'm not sure that we need this check either, we can instead add `strict=True` to the `zip` below and in `append_fast`.

So we could just simplify this to:
```python
    if isinstance(rank, int):
        # We do not need a special case for the sampled token
        # being in the topk, since inserting duplicated data
        # into a dictionary twice is the same as doing it once.
        topk_ranks = range(1, num_logprobs + 1)
        ranks = itertools.chain((rank,), topk_ranks)
```

njhill (commented):
Thanks @xrwang8

I have not looked in full detail yet but something here looks incorrect to me.

This method I think is for adding logprobs corresponding to a single position. Each position may have multiple candidate tokens along with their ranks.

In the prompt logprobs / spec decoding case, there will be more than one position so the method I think should be called multiple times.

The change in this PR looks like it's conflating the multiple per-position tokens with multiple positions (but sorry if I misinterpreted it).

---

Re tests, thanks a lot for adding the unit tests. Do we also have tests though which exercise this in an end-to-end context. For example testing the prompt logprobs functionality from the outside and checking that the returned output with ranks etc. are all correct? These kind of tests are more valuable imo since they aren't tightly coupled to implementation details that may change.

cc @Jialin

PR #27753, opened 2025-10-29
[Hybrid] Pass kernel block size to builders

tdoublep:
## Purpose

Solves https://github.com/vllm-project/vllm/issues/27264 and https://github.com/vllm-project/vllm/issues/26936

This PR makes two changes:
- The GPU model runner will now pass the kernel block size to the metadata builders, fixing a pretty bad bug that exists on main.
- I also restrict the kernel block sizes for the FlashAttention backend based on the discussion in https://github.com/vllm-project/vllm/issues/27264. This is required since for block_size >= 128, FA will read partial block data that may contain NaN's if mamba state is in fp32, which leads to NaN's appearing in output. 

## Test Plan

I'm using the following test script:
```python
from vllm import LLM, SamplingParams
import os

os.environ['VLLM_ATTENTION_BACKEND'] = 'FLASH_ATTN'
#os.environ['VLLM_ATTENTION_BACKEND'] = 'FLASHINFER'

prompts = ["Solve the following math problem step by step. The last line of your response should be of the form Answer: $Answer (without quotes) where $Answer is the answer to the problem.\n\nConsider the paths of length $16$ that follow the lines from the lower left corner to the upper right corner on an $8\\times 8$ grid. Find the number of such paths that change direction exactly four times, as in the examples shown below.\n\nRemember to put your answer on its own line after \"Answer:\"."]
sampling_params = SamplingParams(temperature=0.8, top_p=0.95)

llm = LLM(
    model="nvidia/NVIDIA-Nemotron-Nano-9B-v2",
    trust_remote_code=True,
    num_gpu_blocks_override=10,
    compilation_config={'cudagraph_capture_sizes': [1]},
    mamba_ssm_cache_dtype="float32",
    max_num_seqs=1,
    enable_prefix_caching=True,
)

outputs = []
for _ in range(10):
    outputs.append(llm.generate(prompts, sampling_params)[0])

# Print the outputs.
print(f"Prompt: {prompts[0]!r}")
for output in outputs:
    prompt = output.prompt
    generated_text = output.outputs[0].text
    print(f"Generated text: {generated_text!r}")
    print(f"Generated token IDs: {output.outputs[0].token_ids!r}")
```
Which on main using `FLASH_ATTN` produces:
```
Generated text: '</think>\nThe problem requires finding the number of paths of length 16'
Generated token IDs: [1885, 74045, 1561, 1784, 4127, 10867, 13170, 1278, 2782, 1307, 22344, 1307, 5592, 1032, 1049, 1054]
[...]

PR #29084, opened 2025-11-20
[Attention] Refactor FA `block_size` limitations to hybrid models only 

NickLucche:
This PR limits the blocks-size changes introduced in https://github.com/vllm-project/vllm/pull/27753/ to hybrid-models only.
As non hybrid-models are not affected by the issues reported in the PR above, not allowing a supported physical block_size can hinder performance of kv cache transfers.

Eg on the likes of PD disaggregation with NIXL, we're not limited by bw so reducing the size of blocks reduces optimal saturation of medium.

To do so I have *soft-reverted* some of the changes introduced in the https://github.com/vllm-project/vllm/pull/24794 refactor, regarding `supported_kernel_block_size` being a class attribute rather than a staticmethod. Happy to discuss other options, but I found I couldn't have a cleaner solution with classvars. 

cc @tdoublep for hybrid models
cc @MatthewBonanni for attn metadata refactor

## Test

Reproducing same setup as for hybrid models.
```
# after PR (same as without) FA backend
Prompt: 'Solve the following math problem step by step. The last line of your response should be of the form Answer: $Answer (without quotes) where $Answer is the answer to the problem.\n\nConsider the paths of length $16$ that follow the lines from the lower left corner to the upper right corner on an $8\\times 8$ grid. Find the number of such paths that change direction exactly four times, as in the examples shown below.\n\nRemember to put your answer on its own line after "Answer:".'
Generated text: '</think>\nThe problem requires finding the number of paths of length 16'
Generated token IDs: [1885, 74045, 1561, 1784, 4127, 10867, 13170, 1278, 2782, 1307, 22344, 1307, 5592, 1032, 1049, 1054]
Generated text: ' \n\nExample paths:\n- Path 1: Right, Right, Down,'
Generated token IDs: [1032, 1267, 20396, 22344, 1877, 1045, 17669, 1032, 1049, 1058, 21285, 1044, 21285, 1044, 16999, 1044]
Generated text: '</think>\nThe number of paths of length 16 from the lower left'
Generated token IDs: [1885, 74045, 1561, 1784, 2782, 1307, 22344, 1307, 5592, 1032, 1049, 1054, 1562, 1278, 4953, 3979]
Generated text: '</think>\nThe problem involves finding the number of paths of length 16'
Generated token IDs: [1885, 74045, 1561, 1784, 4127, 19263, 13170, 1278, 2782, 1307, 22344, 1307, 5592, 1032, 1049, 1054]
Generated text: ' output: To solve the problem of finding the number of paths of length 1'
Generated token IDs: [4848, 1058, 3870, 15047, 1278, 4127, 1307, 13170, 1278, 2782, 1307, 22344, 1307, 5592, 1032, 1049]
Generated text: ' \n</think>\nThe problem requires finding the number of paths of length '
Generated token IDs: [1032, 1010, 1885, 74045, 1561, 1784, 4127, 10867, 13170, 1278, 2782, 1307, 22344, 1307, 5592, 1032]
Generated text: '</think>\nThe problem requires finding the number of paths of length 16'
Generated token IDs: [1885, 74045, 1561, 1784, 4127, 10867, 13170, 1278, 2782, 1307, 22344, 1307, 5592, 1032, 1049, 1054]
Generated text: ' \n</think>\nThe paths of length 16 from the lower left'
Generated token IDs: [1032, 1010, 1885, 74045, 1561, 1784, 22344, 1307, 5592, 1032, 1049, 1054, 1562, 1278, 4953, 3979]
Generated text: " \n\n<think>\nOkay, let's see. I need to find the"
Generated token IDs: [1032, 1267, 49250, 2077, 1561, 44053, 1044, 2878, 1681, 3219, 1046, 1362, 2534, 1317, 3081, 1278]
Generated text: '<think>\nOkay, so I need to find the number of paths on an'
Generated token IDs: [1060, 74045, 1561, 44053, 1044, 1878, 1362, 2534, 1317, 3081, 1278, 2782, 1307, 22344, 1408, 1420]
```

### DIFF
--- a/vllm/attention/backends/abstract.py
+++ b/vllm/attention/backends/abstract.py
@@ -46,8 +46,11 @@ class AttentionBackend(ABC):
     # makes sure the output tensor is allocated inside the cudagraph.
     accept_output_buffer: bool = False
     supported_dtypes: ClassVar[list[torch.dtype]] = [torch.float16, torch.bfloat16]
-    supported_kernel_block_sizes: ClassVar[list[int | MultipleOf]] = [MultipleOf(1)]
     supported_kv_cache_dtypes: ClassVar[list["CacheDType"]] = ["auto"]
+
+    @staticmethod
+    def get_supported_kernel_block_size() -> list[int | MultipleOf]:
+        return [MultipleOf(1)]
 
     @staticmethod
     @abstractmethod
@@ -142,10 +145,11 @@ class AttentionBackend(ABC):
         if block_size not in valid_sizes:
             return False
 
-        if not cls.supported_kernel_block_sizes:
+        supported_kernel_block_sizes = cls.get_supported_kernel_block_size()
+        if not supported_kernel_block_sizes:
             return True
 
-        for supported_size in cls.supported_kernel_block_sizes:
+        for supported_size in supported_kernel_block_sizes:
             if isinstance(supported_size, MultipleOf):
                 supported_size = supported_size.base
             # With hybrid_blocks feature, the framework-level block size
--- a/vllm/attention/backends/abstract.py
+++ b/vllm/attention/backends/abstract.py
@@ -49,7 +49,7 @@ class AttentionBackend(ABC):
     supported_kv_cache_dtypes: ClassVar[list["CacheDType"]] = ["auto"]
 
     @staticmethod
-    def get_supported_kernel_block_size() -> list[int | MultipleOf]:
+    def get_supported_kernel_block_sizes() -> list[int | MultipleOf]:
         return [MultipleOf(1)]
 
     @staticmethod
@@ -145,7 +145,7 @@ class AttentionBackend(ABC):
         if block_size not in valid_sizes:
             return False
 
-        supported_kernel_block_sizes = cls.get_supported_kernel_block_size()
+        supported_kernel_block_sizes = cls.get_supported_kernel_block_sizes()
         if not supported_kernel_block_sizes:
             return True
 
--- a/vllm/v1/attention/backends/flash_attn.py
+++ b/vllm/v1/attention/backends/flash_attn.py
@@ -61,7 +61,11 @@ class FlashAttentionBackend(AttentionBackend):
     def get_supported_kernel_block_size() -> list[int | MultipleOf]:
         vllm_config = get_current_vllm_config()
         model_config = vllm_config.model_config
-        if model_config and model_config.is_hybrid:
+        cache_config = vllm_config.cache_config
+        if model_config.is_hybrid and (
+            cache_config.mamba_ssm_cache_dtype == "float32"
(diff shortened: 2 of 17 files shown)


vllm/v1/attention/backends/flashinfer.py
```diff
@@ -558,7 +559,6 @@ def __init__(
             )
 
     @classmethod
-    @override
```
NickLucche, line 569: pre-commit rightfully complaining as this is not an ovveride

vllm/v1/attention/backends/flash_attn.py
```diff
@@ -56,11 +56,18 @@
...
-    # MultipleOf(16), these are the block sizes that do not
-    # suffer from the NaN propagation problem described here:
-    # https://github.com/Dao-AILab/flash-attention/issues/1974
-    supported_kernel_block_sizes: ClassVar[list[int | MultipleOf]] = [16, 32, 64]
+
+    @staticmethod
+    def get_supported_kernel_block_size() -> list[int | MultipleOf]:
+        vllm_config = get_current_vllm_config()
+        model_config = vllm_config.model_config
+        if model_config and model_config.is_hybrid:
```
tdoublep: If we want, we could refine this further and only restrict the block size only when `cache_config. mamba_cache_dtype == "float32"` or `cache_config.mamba_ssm_cache_dtype == "float32"` in addition to these conditions. I'm also OK with how you have it though. 

Why is it necessary to check `model_config` exists btw? Can this function get called before the configs are created?

tdoublep (commented):
Thanks for doing this!

vllm/attention/backends/abstract.py
```diff
@@ -46,9 +46,12 @@ class AttentionBackend(ABC):
     # makes sure the output tensor is allocated inside the cudagraph.
     accept_output_buffer: bool = False
     supported_dtypes: ClassVar[list[torch.dtype]] = [torch.float16, torch.bfloat16]
-    supported_kernel_block_sizes: ClassVar[list[int | MultipleOf]] = [MultipleOf(1)]
     supported_kv_cache_dtypes: ClassVar[list["CacheDType"]] = ["auto"]
 
+    @staticmethod
+    def get_supported_kernel_block_size() -> list[int | MultipleOf]:
```
MatthewBonanni: nit: can we make this `get_supported_kernel_block_sizes()` (plural)?

MatthewBonanni (approved):
Thanks! I have no objections to making this a static method.

vllm/v1/attention/backends/flash_attn.py
NickLucche: added! Actually I was only checking due to mypy, but it looks like it's not complaining anymore

tdoublep (approved):
LGTM

PR #29086, opened 2025-11-20
Revert back to torch.equal over torch.allclose from #28819 

eldarkurtic:
As pointed out in https://github.com/vllm-project/vllm/pull/28819 , we already had some offline discussions (cc @benchislett) on whether `torch.equal` or `torch.allclose` should be used here to detect whether draft model has exactly the same layers as the target model to enable weight-sharing. 
I have two major concerns with `torch.allclose` approach:
1. `embed_tokens` tend to have very small magnitudes of updates during speculators training, making them look very close **but not equal** to the verifier model. `torch.allclose` opens up a pathway of overriding those weights due to their similarity, which is not something that a user who trained a speculator model with distinct embed_tokens expects.
2. `torch.allclose` also opens up an additional set of questions for how to tune `atol/rtol`, especially for different dtypes (which will become even more problematic ones people start using quantized models)

I would advocate here to go back to `torch.equal` until someone demonstrates a use case where `torch.equal` fails to identify equal layers due to issues with FP precision. If that happens, we can update the logic to check for bitwise equality like this:
```python
a_bits = a.view(torch.int32)
b_bits = b.view(torch.int32)

bitwise_equal = torch.equal(a_bits, b_bits)
```

The PR also adds CPU offloading to prevent GPU OOM issues reported in #28819

### DIFF
--- a/vllm/v1/spec_decode/eagle.py
+++ b/vllm/v1/spec_decode/eagle.py
@@ -1028,8 +1028,8 @@ class EagleProposer:
                 elif (
                     isinstance(target_embed_tokens.weight, torch.Tensor)
                     and isinstance(self.model.model.embed_tokens.weight, torch.Tensor)
-                    # Offload to CPU for comparison to avoid extra GPU memory usage
-                    # in CI testing environments with limited GPU memory
+                    # TODO: Offload to CPU for comparison to avoid extra GPU memory
+                    # usage in CI testing environments with limited GPU memory
                     and torch.equal(
                         target_embed_tokens.weight.cpu(),
                         self.model.model.embed_tokens.weight.cpu(),
@@ -1078,8 +1078,8 @@ class EagleProposer:
                 hasattr(target_language_model, "lm_head")
                 and isinstance(target_language_model.lm_head.weight, torch.Tensor)
                 and isinstance(self.model.lm_head.weight, torch.Tensor)
-                # Offload to CPU for comparison to avoid extra GPU memory usage
-                # in CI testing environments with limited GPU memory
+                # TODO: Offload to CPU for comparison to avoid extra GPU memory
+                # usage in CI testing environments with limited GPU memory
                 and torch.equal(
                     target_language_model.lm_head.weight.cpu(),
                     self.model.lm_head.weight.cpu(),
--- a/vllm/v1/spec_decode/eagle.py
+++ b/vllm/v1/spec_decode/eagle.py
@@ -1077,7 +1077,8 @@ class EagleProposer:
                 and isinstance(target_language_model.lm_head.weight, torch.Tensor)
                 and isinstance(self.model.lm_head.weight, torch.Tensor)
                 and torch.equal(
-                    target_language_model.lm_head.weight, self.model.lm_head.weight
+                    target_language_model.lm_head.weight.cpu(),
+                    self.model.lm_head.weight.cpu(),
                 )
             ):
                 share_lm_head = True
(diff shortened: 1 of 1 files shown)


vllm/v1/spec_decode/eagle.py
```diff
@@ -1028,11 +1028,9 @@ def load_model(self, target_model: nn.Module) -> None:
                 elif (
                     isinstance(target_embed_tokens.weight, torch.Tensor)
                     and isinstance(self.model.model.embed_tokens.weight, torch.Tensor)
-                    and torch.allclose(
+                    and torch.equal(
                         target_embed_tokens.weight.cpu(),
                         self.model.model.embed_tokens.weight.cpu(),
```
NickLucche, line 1035: as I commented in the previous PR, I don't think this check should happen on cpu when both weights are already loaded.
If the memory is so tight it will crash during dummy run instead of during load.

NickLucche: also qq: can we just share these two layers?

NickLucche (commented):
Is there any other model with smaller footprint that we can use for the test that triggered this discussion in https://github.com/vllm-project/vllm/pull/28819?

vllm/v1/spec_decode/eagle.py
eldarkurtic: No, we can't share the layers just by default. A good amount of eagle-3 speculators are training embed_tokens and/or lm_head from scratch.

eldarkurtic: As for the memory issue, re-reading the original #28819 makes me think that the memory issue came from a completely different thing: https://github.com/vllm-project/vllm/pull/28819/files#diff-66eeb1275496be8941bef653c4e581ac25dfd2ba463f704801c0729213ab5f34L175
and not from `torch.equal` running on a GPU (basically one instantiation of a layer hasn't been treated properly in the weight-sharing logic). 

As for the general CPU-vs-GPU `torch.equal` comparison: if we can't fit `torch.equal(t1, t2)` on a GPU, then what is the purpose of that deployment, it won't have memory to fit a single forward pass? Shall we revert this check back to gpu?

eldarkurtic: @zhewenl could you perhaps confirm if the GPU OOM happened because of `torch.equal(t1, t2)` running on a GPU or because of the fact that one instantiation of a layer hasn't been treated properly in the weight-sharing logic (which you resolved here: https://github.com/vllm-project/vllm/pull/28819/files#diff-66eeb1275496be8941bef653c4e581ac25dfd2ba463f704801c0729213ab5f34L175)?

zhewenl: @eldarkurtic it's happening because `torch.equal(t1, t2)` running on GPU and OOMed on CI, which only has 24G memory.
> As for the general CPU-vs-GPU torch.equal comparison: if we can't fit torch.equal(t1, t2) on a GPU, then what is the purpose of that deployment, it won't have memory to fit a single forward pass? Shall we revert this check back to gpu?

Yes, I fully agree, and from CI's perspective, we can check if we can replace the test suite with smaller models

LucasWilkinson:
cc @benchislett

benchislett (commented):
I'm overall indifferent to running the computation on cpu vs gpu. I don't think it will make a big difference either way.

As for whether to use equal or allclose, it's hard to say. If we can't find a rigorous bound like dtype-epsilon or something, we should just use direct .equal comparison. We should be conservative here and avoid risking inaccuracy where possible.

vllm/v1/spec_decode/eagle.py
NickLucche: @eldarkurtic 
> No, we can't share the layers just by default

Yeah no I meant loading the same weight once and then sharing the tensor between draft-target model (if not done already)

NickLucche (approved):
Discussed offline, leaving check on cpu to unblock CI until we have a valid smaller model that we can use for the test

PR #28834, opened 2025-11-17
[MODEL] Implement plamo3

Alnusjaponica:
## Purpose

This PR adds support for [PLaMo3 falimies](https://huggingface.co/collections/pfnet/plamo-3). 

## Test Plan
Run unit test added to `tests/distributed/test_pipeline_parallel.py`

## Test Result
```
> uv run pytest tests/distributed/test_pipeline_parallel.py
/usr/local/lib/python3.12/dist-packages/pytest_asyncio/plugin.py:211: PytestDeprecationWarning: The configuration option "asyncio_default_fixture_loop_scope" is unset.
The event loop scope for asynchronous fixtures will default to the fixture caching scope. Future versions of pytest-asyncio will default the loop scope for asynchronous fixtures to function scope. Set the default fixture loop scope explicitly in order to avoid unexpected behavior in the future. Valid fixture loop scopes are: "function", "class", "module", "package", "session"

  warnings.warn(PytestDeprecationWarning(_DEFAULT_FIXTURE_LOOP_SCOPE_UNSET))
=================================== test session starts ====================================
platform linux -- Python 3.12.3, pytest-8.3.5, pluggy-1.5.0
rootdir: /home/team-prj-mfm-platform/pfnet-vllm
configfile: pyproject.toml
plugins: hydra-core-1.3.2, Faker-37.6.0, asyncio-1.1.0, typeguard-4.4.2, anyio-4.9.0
asyncio: mode=Mode.STRICT, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 3 items                                                                          

tests/distributed/test_pipeline_parallel.py .ss                                      [100%]

===================================== warnings summary =====================================
tests/distributed/test_pipeline_parallel.py::test_tp_language_generation[pfnet/plamo-3-nict-2b-base-parallel_setup0-mp-auto-test_options0]
  /home/team-prj-mfm-platform/pfnet-vllm/tests/utils.py:890: DeprecationWarning: This process (pid=2490685) is multi-threaded, use of fork() may lead to deadlocks in the child.
    pid = os.fork()

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=================== 1 passed, 2 skipped, 1 warning in 121.31s (0:02:01) ====================
```
---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
[...]

PR #29092, opened 2025-11-20
[Bugfix] Fix Plamo3 rope handling

DarkLight1337:
## Purpose

Fix a failure in model initialization tests caused by an incompatibility between #28834 and #28542 .

https://buildkite.com/vllm/ci/builds/39885#019aa0ed-01f7-4df6-a93b-2d95f144776c

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/models/plamo3.py
+++ b/vllm/model_executor/models/plamo3.py
@@ -157,7 +157,6 @@ class Plamo3AttentionMixer(nn.Module):
         layer_idx = extract_layer_index(prefix)
         layer_type = config.layer_types[layer_idx]
         is_sliding = layer_type == "sliding_attention"
-        self.sliding_window = config.sliding_window if is_sliding else None
 
         # Initialize the rotary embedding.
         if layer_type in config.rope_parameters:
--- a/vllm/model_executor/models/plamo3.py
+++ b/vllm/model_executor/models/plamo3.py
@@ -62,7 +62,7 @@ class Plamo3Config(PretrainedConfig):  # type: ignore
     # if `sliding_window` is list
     interleaved_sliding_window: list[int | None]
     sliding_window_pattern: int
-    rope_theta: int
+    rope_parameters: dict[str, Any]
     rope_local_theta: int
     # MLP
     intermediate_size: int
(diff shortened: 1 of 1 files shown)


vllm/model_executor/models/plamo3.py
```diff
@@ -170,8 +170,7 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = "", **kwargs) -> No
             self.head_dim,
             rotary_dim=self.head_dim,
             max_position=max_position,
-            base=self.rope_theta,
-            rope_scaling=self.rope_scaling,
+            rope_parameters=config.rope_parameters,
```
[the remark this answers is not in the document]
DarkLight1337: Fixed

hmellor:
Is Plamo2's config implemented in vLLM? 

If yes it can be updated to create a nested `paramaters_config` (see Gemma 3 on Transformers main for reference) and we don't need to do the init time check in the model code, it would always be nested.

vllm/model_executor/models/plamo3.py
```diff
@@ -153,13 +153,23 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = "", **kwargs) -> No
...
+        if layer_type in config.rope_parameters:
+            # Transformers v5 rope config.
+            rope_parameters = config.rope_parameters[layer_type]
+        else:
+            # Transformers v4 rope config.
+            # Global attention. Use the values in config.json.
+            rope_parameters = config.rope_parameters.copy()
+            # Local attention. Override the values in config.json.
+            if is_sliding:
+                rope_parameters["rope_theta"] = config.rope_local_theta
```
hmellor: This issue was found with Gemma too, where the rope_type could still be non-default, which could cause scaling to happen anyway 

```suggestion
                rope_parameters = dict(rope_type="default", rope_theta=config.rope_local_theta)
```

hmellor (changes requested):
Overall LGTM, just one comment on something I missed in Gemma 3 which someone has a fix for.

---

Side notes about v5 withthis model:

- It look like it's a custom model, which means that unless the model authors update their repo, it will never support v5 style RoPE.
- However, including the v5 block could be a good way to convince them to update it, because there would be no downtime where their model was unsupported.

vllm/model_executor/models/plamo3.py
```diff
@@ -153,13 +153,23 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = "", **kwargs) -> No
...
+        is_sliding = layer_type == "sliding_attention"
+
+        # Initialize the rotary embedding.
+        if layer_type in config.rope_parameters:
+            # Transformers v5 rope config.
+            rope_parameters = config.rope_parameters[layer_type]
+        else:
+            # Transformers v4 rope config.
+            # Global attention. Use the values in config.json.
+            rope_parameters = config.rope_parameters.copy()
```
hmellor: Making the previous change should make this copy unnecessary

```suggestion
            rope_parameters = config.rope_parameters
```

hmellor (approved):

PR #29096, opened 2025-11-20
[torch.compile] add condition to normalize_value

ghost:
Follow-up to this: https://github.com/vllm-project/vllm/pull/26468#discussion_r2545031784 

cc @hmellor


ProExpertProg (changes requested):
This is not good because it won't let us return the full pass config in `get_hash_factors` which we want to be able to add it to the vllm cache file

PR #29105, opened 2025-11-20
Add Multimodal Processor Benchmark 

reaganjlee:
Resolves #24171. 

## Purpose

This PR adds in a benchmark for the multimodal processor. This tests for the MM processor on a single vllm instance on a single GPU. 

Originally, I planned to additionally test for 1. DP on multiple GPUs and 2. multiple instances on a single GPU. On second thought, testing 1. DP seems equivalent to just running the benchmark multiple times. For 2. multiple instances on single GPU, I would like direction on which type of load balancing should be done on the #2 point. I could use a nginx round robin, or is the request router from the [production_stack repo](https://github.com/vllm-project/production-stack/tree/main/src/vllm_router) recommended? This also seems to be a good stopping place since the PR is just around 800 lines.

## Test Plan
```
# Using just HuggingFaceTB/SmolVLM-Instruct to save on compute during development, and random-mm dataset for simplicity
~vllm bench mm-processor     --model HuggingFaceTB/SmolVLM-Instruct     --dataset-name random-mm     --num-prompts 100     --max-concurrency 10~

# concurrency
~vllm bench mm-processor     --model HuggingFaceTB/SmolVLM-Instruct     --dataset-name random-mm     --num-prompts 100     --max-concurrency 50~

vllm bench mm-processor \
        --model HuggingFaceTB/SmolVLM-Instruct \
        --num-prompts 10 \
        --input-len 1024 \
        --output-len 128 \
        --num-images 1
```

## Test Result

```
# Example result
# vllm bench mm-processor \
        --model HuggingFaceTB/SmolVLM-Instruct \
        --num-prompts 10 \
        --input-len 1024 \
        --output-len 128 \
        --num-images 1
Multimodal Processor Benchmark Results                                                                                                                          
                                                                                                                                                                
MM Processor Timing (ms):                                                                                                                                       
             Stage   Mean Median    Std  P99.0                                                                                                                  
 hf_processor_time 228.01 156.81 204.41 780.97                                                                                                                  
      hashing_time   0.06   0.06   0.01   0.08
 cache_lookup_time   0.09   0.08   0.01   0.12
prompt_update_time   5.00   4.93   0.52   6.20
        total_time   0.00   0.00   0.00   0.00

End-to-End Latency (ms):
Metric Value (ms)
  Mean     430.99
Median     430.99
   Std       0.00
 P99.0     430.99
```

---
<details>

</details>

### DIFF
--- a/docs/cli/bench/mm_processor.md
+++ b/docs/cli/bench/mm_processor.md
@@ -1,0 +1,9 @@
+# vllm bench mm-processor
+
+## JSON CLI Arguments
+
+--8<-- "docs/cli/json_tip.inc.md"
+
+## Arguments
+
+--8<-- "docs/argparse/bench_mm_processor.inc.md"
--- a/docs/mkdocs/hooks/generate_argparse.py
+++ b/docs/mkdocs/hooks/generate_argparse.py
@@ -92,6 +92,7 @@ def auto_mock(module_name: str, attr: str, max_mocks: int = 100):
 
 
 bench_latency = auto_mock("vllm.benchmarks", "latency")
+bench_mm_processor = auto_mock("vllm.benchmarks", "mm_processor")
 bench_serve = auto_mock("vllm.benchmarks", "serve")
 bench_sweep_plot = auto_mock("vllm.benchmarks.sweep.plot", "SweepPlotArgs")
 bench_sweep_plot_pareto = auto_mock(
@@ -222,6 +223,7 @@ def on_startup(command: Literal["build", "gh-deploy", "serve"], dirty: bool):
         "run-batch": create_parser(openai_run_batch.make_arg_parser),
         # Benchmark CLI
         "bench_latency": create_parser(bench_latency.add_cli_args),
+        "bench_mm_processor": create_parser(bench_mm_processor.add_cli_args),
         "bench_serve": create_parser(bench_serve.add_cli_args),
         "bench_sweep_plot": create_parser(bench_sweep_plot.add_cli_args),
         "bench_sweep_plot_pareto": create_parser(bench_sweep_plot_pareto.add_cli_args),
--- a/vllm/benchmarks/datasets.py
+++ b/vllm/benchmarks/datasets.py
@@ -1437,148 +1437,12 @@ def add_dataset_parser(parser: FlexibleArgumentParser):
     )
 
     random_group = parser.add_argument_group("random dataset options")
-    random_group.add_argument(
-        "--random-input-len",
-        type=int,
-        default=1024,
-        help="Number of input tokens per request, used only for random sampling.",
-    )
-    random_group.add_argument(
-        "--random-output-len",
-        type=int,
-        default=128,
-        help="Number of output tokens per request, used only for random sampling.",
-    )
-    random_group.add_argument(
-        "--random-range-ratio",
-        type=float,
-        default=0.0,
-        help="Range ratio for sampling input/output length, "
-        "used only for random sampling. Must be in the range [0, 1) to define "
-        "a symmetric sampling range"
-        "[length * (1 - range_ratio), length * (1 + range_ratio)].",
-    )
-    random_group.add_argument(
-        "--random-prefix-len",
-        type=int,
-        default=0,
-        help=(
-            "Number of fixed prefix tokens before the random context "
-            "in a request. "
-            "The total input length is the sum of `random-prefix-len` and "
-            "a random "
-            "context length sampled from [input_len * (1 - range_ratio), "
-            "input_len * (1 + range_ratio)]."
-        ),
-    )
-    random_group.add_argument(
-        "--random-batch-size",
-        type=int,
-        default=1,
-        help=("Batch size for random sampling. Only used for embeddings benchmark."),
-    )
-    random_group.add_argument(
-        "--no-reranker",
-        action="store_true",
-        help=(
-            "Whether the model supports reranking natively."
-            " Only used for reranker benchmark."
-        ),
-    )
+    add_random_dataset_base_args(random_group)
 
-    # random multimodal dataset options
     random_mm_group = parser.add_argument_group(
         "random multimodal dataset options extended from random dataset"
     )
-    random_mm_group.add_argument(
-        "--random-mm-base-items-per-request",
-        type=int,
-        default=RandomMultiModalDataset.DEFAULT_BASE_ITEMS_PER_REQUEST,
-        help=(
-            "Base number of multimodal items per request for random-mm. "
-            "Actual per-request count is sampled around this base using "
-            "--random-mm-num-mm-items-range-ratio."
-        ),
-    )
-    random_mm_group.add_argument(
-        "--random-mm-num-mm-items-range-ratio",
-        type=float,
-        default=RandomMultiModalDataset.DEFAULT_NUM_MM_ITEMS_RANGE_RATIO,
-        help=(
-            "Range ratio r in [0, 1] for sampling items per request. "
-            "We sample uniformly from the closed integer range "
-            "[floor(n*(1-r)), ceil(n*(1+r))] "
-            "where n is the base items per request. "
-            "r=0 keeps it fixed; r=1 allows 0 items. The maximum is clamped "
-            "to the sum of per-modality limits from "
-            "--random-mm-limit-mm-per-prompt. "
-            "An error is raised if the computed min exceeds the max."
-        ),
-    )
-    random_mm_group.add_argument(
-        "--random-mm-limit-mm-per-prompt",
-        type=json.loads,
-        default=RandomMultiModalDataset.DEFAULT_LIMIT_MM_PER_PROMPT,
-        help=(
-            "Per-modality hard caps for items attached per request, e.g. "
-            '\'{"image": 3, "video": 0}\'. The sampled per-request item '
-            "count is clamped to the sum of these limits. When a modality "
-            "reaches its cap, its buckets are excluded and probabilities are "
-            "renormalized."
-            "OBS.: Only image sampling is supported for now."
-        ),
-    )
-
-    def _parse_mm_bucket_config(v: object) -> dict[tuple[int, int, int], float]:
-        # If already a dict (e.g., programmatic call), normalize keys
-        def normalize(d: dict) -> dict[tuple[int, int, int], float]:
-            out: dict[tuple[int, int, int], float] = {}
-            for k, val in d.items():
-                key = k
-                if isinstance(key, str):
-                    with suppress(Exception):
-                        key = ast.literal_eval(key)
-                if not (
-                    isinstance(key, tuple)
-                    and len(key) == 3
-                    and all(isinstance(x, int) for x in key)
-                ):
-                    raise ValueError(
-                        f"Invalid bucket key {k!r}. Expected tuple (H, W, T)."
-                    )
-                out[(int(key[0]), int(key[1]), int(key[2]))] = float(val)
-            return out
-
-        if isinstance(v, dict):
-            return normalize(v)
-        if isinstance(v, str):
-            # Python literal (supports tuple keys)
-            parsed = ast.literal_eval(v)
-            if not isinstance(parsed, dict):
-                raise ValueError("Bucket config must parse to a dict.")
-            return normalize(parsed)
-        raise ValueError("Unsupported value for --random-mm-bucket-config.")
-
-    random_mm_group.add_argument(
-        "--random-mm-bucket-config",
-        type=_parse_mm_bucket_config,
-        default=RandomMultiModalDataset.DEFAULT_MM_ITEM_BUCKET_CONFIG,
-        help=(
-            "The bucket config is a dictionary mapping a multimodal item"
-            "sampling configuration to a probability."
-            "Currently allows for 2 modalities: images and videos. "
-            "An bucket key is a tuple of (height, width, num_frames)"
-            "The value is the probability of sampling that specific item. "
-            "Example: "
-            "--random-mm-bucket-config "
-            "{(256, 256, 1): 0.5, (720, 1280, 1): 0.4, (720, 1280, 16): 0.10} "
-            "First item: images with resolution 256x256 w.p. 0.5"
-            "Second item: images with resolution 720x1280 w.p. 0.4 "
-            "Third item: videos with resolution 720x1280 and 16 frames w.p. 0.1"
-            "OBS.: If the probabilities do not sum to 1, they are normalized."
-            "OBS bis.: Only image sampling is supported for now."
-        ),
-    )
+    add_random_multimodal_dataset_args(random_mm_group)
 
     hf_group = parser.add_argument_group("hf dataset options")
     hf_group.add_argument(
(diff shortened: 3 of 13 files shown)


vllm/multimodal/processing.py
```diff
@@ -70,6 +73,134 @@
 
 _S = TypeVar("_S", str, list[int])
 
+
+@dataclass
+class MultiModalProcessorTimingStats:
+    """Per-request timing statistics for multimodal processor stages."""
+
```
Isotr0py, line 100: Let's move the stats objects to a new file `multimodal/stats.py`.

vllm/envs.py
```diff
@@ -769,6 +769,9 @@ def get_vllm_port() -> int | None:
     # [DEPRECATED] Cache size (in GiB per process) for multimodal input cache
     # Default is 4 GiB per API process + 4 GiB per engine core process
     "VLLM_MM_INPUT_CACHE_GIB": lambda: int(os.getenv("VLLM_MM_INPUT_CACHE_GIB", "4")),
+    "VLLM_ENABLE_MM_PROCESSOR_STATS": lambda: bool(
+        int(os.getenv("VLLM_ENABLE_MM_PROCESSOR_STATS", "0"))
+    ),
```
Isotr0py: Hmmm, our environment variables are quite bloated. Perhaps we can move this to `MultimodalConfig`?

vllm/benchmarks/multimodal_processor.py
```diff
@@ -0,0 +1,460 @@
...
+    
+    Returns:
+        List containing a single instance URL
+    """
+    if base_url:
+        return [base_url]
+    
+    host = getattr(args, "host", "127.0.0.1") if args else "127.0.0.1"
+    port = getattr(args, "port", 8000) if args else 8000
+    return [f"http://{host}:{port}"]
```
Isotr0py: IIRC, all bench commands except `vllm bench serve` are using offline inference?

I think we can extend mm processor benchmarks functionality for `vllm bench serve` with online serving, and leaving `vllm bench multimodal-processor` to use offline inference. 🤔

vllm/benchmarks/multimodal_processor.py
```diff
@@ -0,0 +1,460 @@
...
+        selected_percentiles = [
+            float(p) for p in getattr(args, "metric_percentiles", "99").split(",")
+        ]
+        for stage, metrics in result["mm_processor_stats"].items():
+            print(f"  {stage}:")
+            print(f"    Mean: {metrics['mean']:.2f}")
+            print(f"    Median: {metrics['median']:.2f}")
+            print(f"    Std: {metrics['std']:.2f}")
+            for p in selected_percentiles:
+                print(f"    P{p}: {metrics.get(f'p{p}', 0.0):.2f}")
```
Isotr0py: I think we can format the output's display as a table using `pandas`'s dataframe. Current output is a bit long with duplicate fields:
```
  hf_processor_time:
    Mean: 204.14
    Median: 201.04
    Std: 48.78
    P99.0: 319.08
  hashing_time:
    Mean: 1.58
    Median: 1.94
    Std: 1.25
    P99.0: 5.57
  cache_lookup_time:
    Mean: 0.16
    Median: 0.14
    Std: 0.10
    P99.0: 0.44
  prompt_update_time:
    Mean: 7.30
    Median: 7.02
    Std: 0.96
    P99.0: 8.80
  total_time:
    Mean: 232.79
    Median: 211.02
    Std: 208.11
    P99.0: 349.57
```

Isotr0py (commented):
Thanks for implementing this feature! I leave some initial comments. PTAL!

vllm/multimodal/processing.py
DarkLight1337: I feel that making the stats container a global is a bit hacky. Perhaps it's better to have it as a member of `InputProcessingContext`.

vllm/benchmarks/multimodal_processor.py
reaganjlee: Yeah I originally designed this for the multi-instance case, but having the separate API for this wont add too much code I think, I can separate this out

vllm/benchmarks/multimodal_processor.py
reaganjlee: I think it'd be helpful to keep the E2E latency stats as well, but lmk if you think otherwise

vllm/multimodal/processing.py
```diff
@@ -1492,16 +1692,18 @@ def _call_hf_processor(
         mm_data: Mapping[str, object],
         mm_kwargs: Mapping[str, object],
         tok_kwargs: Mapping[str, object],
+        request_id: str | None = None,
```
Isotr0py: Hmmm, I think it's not good to expose `request_id` for internal multimodal processor... We shouldn't let developer notice which request is being processed in internal processor since it's part of model implementation.

Can we only expose it to `InputProcessingContext`?

vllm/benchmarks/multimodal_processor.py
```diff
@@ -0,0 +1,470 @@
...
+import argparse
+import dataclasses
+import json
+import time
+from dataclasses import dataclass
+from datetime import datetime
+from typing import Any
+
+import numpy as np
+import pandas as pd
```
DarkLight1337: Need to use `PlaceholderModule` for pandas since it's an optional dependency

vllm/benchmarks/multimodal_processor.py
```diff
@@ -0,0 +1,470 @@
...
+        )
+
+    selected_percentiles = [
+        float(p) for p in getattr(args, "metric_percentiles", "99").split(",")
+    ]
+
+    freeze_gc_heap()
+
+    if not envs.VLLM_ENABLE_MM_PROCESSOR_STATS:
+        print(
```
DarkLight1337: I think we should simply raise an error in this case

vllm/config/multimodal.py
```diff
@@ -140,6 +141,17 @@ class MultiModalConfig:
     Value sits in range [0;1) and determines fraction of media tokens
     from each video to be pruned.
     """
+    enable_mm_processor_stats: bool = Field(
```
DarkLight1337: For easier code maintenance, we should set this via either config or env, not both. Personally I prefer using config

vllm/entrypoints/openai/api_server.py
```diff
@@ -352,6 +353,35 @@ async def get_server_load_metrics(request: Request):
     return JSONResponse(content={"server_load": request.app.state.server_load_metrics})
 
 
+@router.get("/mm-processor-stats")
```
DarkLight1337: I think we can combine this with the request metrics instead of having a separate endpoint for it

vllm/config/multimodal.py
hmellor: Should this also belong to the observability config instead?

vllm/multimodal/processing.py
reaganjlee: Got it, and there isn't a exactly clean way to have the data only in `InputProcessingContext` as the class itself is on the model level rather than request level, so we'd need something to keep track for individual requests. 

I think one simple solution is to use `contextvars` to not change any of the signatures, but this isn't a very widely used pattern in the codebase.

vllm/entrypoints/cli/benchmark/multimodal_processor.py
```diff
@@ -0,0 +1,21 @@
...
+import argparse
+
+from vllm.benchmarks.multimodal_processor import add_cli_args, main
+from vllm.entrypoints.cli.benchmark.base import BenchmarkSubcommandBase
+
+
+class BenchmarkMultimodalProcessorSubcommand(BenchmarkSubcommandBase):
+    """The `multimodal-processor` subcommand for `vllm bench`."""
+
+    name = "multimodal-processor"
```
DarkLight1337: Can we shorten this to `mm-processor`? (Same for the module name)

vllm/multimodal/processing.py
DarkLight1337: Looks much cleaner now from the perspective of someone maintaining the `BaseMultiModalProcessor`, thanks.

docs/cli/bench/multimodal_processor.md
```diff
@@ -0,0 +1,9 @@
+# vllm bench multimodal-processor
```
DarkLight1337: Here as well

docs/mkdocs/hooks/generate_argparse.py
```diff
@@ -92,6 +92,7 @@ def auto_mock(module_name: str, attr: str, max_mocks: int = 100):
 
 
 bench_latency = auto_mock("vllm.benchmarks", "latency")
+bench_multimodal_processor = auto_mock("vllm.benchmarks", "multimodal_processor")
```
DarkLight1337: And here

docs/mkdocs/hooks/generate_argparse.py
```diff
@@ -222,6 +223,9 @@ def on_startup(command: Literal["build", "gh-deploy", "serve"], dirty: bool):
         "run-batch": create_parser(openai_run_batch.make_arg_parser),
         # Benchmark CLI
         "bench_latency": create_parser(bench_latency.add_cli_args),
+        "bench_multimodal_processor": create_parser(
```
DarkLight1337: And here

vllm/benchmarks/datasets.py
```diff
@@ -1908,7 +1908,8 @@ def get_samples(args, tokenizer: TokenizerLike) -> list[SampleRequest]:
 
         try:
             # Enforce endpoint compatibility for multimodal datasets.
-            if args.dataset_name == "random-mm" and args.backend not in ["openai-chat"]:
+            backend = getattr(args, "backend", "openai-chat")
```
DarkLight1337: Why is this needed, shouldn't the CLI add this argument?

reaganjlee: Oops you're right, fixed

docs/mkdocs/hooks/generate_argparse.py
DarkLight1337: Rename this as well

docs/mkdocs/hooks/generate_argparse.py
DarkLight1337: Rename this as well

docs/mkdocs/hooks/generate_argparse.py
reaganjlee: I didnt push the change earlier mb, good now

vllm/benchmarks/mm_processor.py
```diff
@@ -0,0 +1,436 @@
...
+    seed: int = 0,
+) -> tuple[list[list[dict]], list[int]]:
+    """
+    Generate random multimodal prompts with synthetic images and text tokens.
+
+    Returns:
+        tuple: (prompts, expected_output_lens)
+            - prompts: List of OpenAI chat format messages with text and images
+            - expected_output_lens: List of expected output lengths
+    """
```
Isotr0py: I think we can use `get_requests` to reuse existing random mm dataset and other image/video/audio datasets.
https://github.com/vllm-project/vllm/blob/de08b8f61bd412f2bcc3a5f21da26871e39a9078/vllm/benchmarks/throughput.py#L334-L404

It should also return list of chat messages as well.

docs/cli/bench/mm_processor.md
```diff
@@ -0,0 +1,9 @@
+# vllm bench mm-processor
+
+## JSON CLI Arguments
+
+--8<-- "docs/cli/json_tip.inc.md"
+
+## Arguments
+
+--8<-- "docs/argparse/bench_multimodal_processor.inc.md"
```
DarkLight1337: Update this link as well?

docs/mkdocs/hooks/generate_argparse.py
```diff
@@ -222,6 +223,9 @@ def on_startup(command: Literal["build", "gh-deploy", "serve"], dirty: bool):
         "run-batch": create_parser(openai_run_batch.make_arg_parser),
         # Benchmark CLI
         "bench_latency": create_parser(bench_latency.add_cli_args),
+        "bench_mm_processor": create_parser(
+            bench_mm_processor.add_cli_args
```
DarkLight1337: Can we put this on one line now?

tests/models/multimodal/processing/test_tensor_schema.py
```diff
@@ -12,7 +12,11 @@
 import torch.nn as nn
 from PIL import Image
 
-from vllm.config import ModelConfig, VllmConfig, set_current_vllm_config
+from vllm.config import (
```
DarkLight1337: Same

vllm/benchmarks/mm_processor.py
reaganjlee: `get_requests` at the moment doesn't support `random-mm`, is there a reason why? 

I previously used the `add_dataset_parser` which does have it, but it led to a minor argument collision.

Btw, is there any reason for the separate functions `get_requests` in the `throughput` benchmark and `get_samples` in `datasets.py`, the only obvious difference I see is the filtering at the end

DarkLight1337: > `get_requests` at the moment doesn't support `random-mm`, is there a reason why?

Not in particular, feel free to add in logic to handle them so that `RandomMultiModalDataset` and `RandomDatasetForReranking` are returned for `args.dataset_name == "random-mm"` and `args.dataset_name == "random-rerank"` respectively and see if they can be run

DarkLight1337: > Btw, is there any reason for the separate functions `get_requests` in the throughput benchmark and `get_samples` in datasets.py, the only obvious difference I see is the filtering at the end

Maybe @mgoin has more context regarding this. IIRC it is because we previously supported more request formats.

reaganjlee:
Good for another look now cc @Isotr0py @DarkLight1337

vllm/benchmarks/mm_processor.py
```diff
@@ -0,0 +1,371 @@
...
+    tokenizer = get_tokenizer(
+        args.tokenizer,
+        tokenizer_mode=getattr(args, "tokenizer_mode", "auto"),
+        trust_remote_code=getattr(args, "trust_remote_code", False),
+    )
+
+    requests = get_requests(args, tokenizer)
+
+    engine_args = EngineArgs.from_cli_args(args)
+    llm = LLM(**dataclasses.asdict(engine_args))
```
DarkLight1337, line 132: I think we can simply access the tokenizer from `LLM` instead of having to construct one separately ourselves

vllm/benchmarks/throughput.py
```diff
@@ -342,16 +346,22 @@ def get_requests(args, tokenizer):
         "lora_path": args.lora_path,
         "max_loras": args.max_loras,
         "num_requests": args.num_prompts,
-        "input_len": args.input_len,
-        "output_len": args.output_len,
+        "input_len": getattr(args, "input_len", None),
+        "output_len": getattr(args, "output_len", None),
```
DarkLight1337: If `input_len` and `output_len` aren't applicable to all branches, better to explicitly set them in each branch so it's clear which datasets support it

vllm/benchmarks/throughput.py
```diff
@@ -574,14 +664,14 @@ def add_cli_args(parser: argparse.ArgumentParser):
         "--input-len",
         type=int,
         default=None,
-        help="Input prompt length for each request",
+        help="Input prompt length for each request. "
```
DarkLight1337: We should keep the trailing comma

DarkLight1337: Same for the other arguments

DarkLight1337: Same for the other arguments

reaganjlee: Ah got it, fixed two other instances I found, lmk if I missed any others

DarkLight1337 (approved):

DarkLight1337:
Please fix the merge conflicts

reaganjlee:
@DarkLight1337 CI failures seem unrelated

TimeYWL:
We built vllm from today's commit, but it seems 'vllm bench multimodal-processor' is missing. Is it possible that only the core code was merged, while the benchmark function have not been included yet?

PR #29108, opened 2025-11-20
[compile] Add utility for hashing the entire vllm codebase.

zhxchen17:
Summary:
Following up on the recent refactoring on compile hashing, instead of hashing only the dynamo traced files, we should be more conservative and able to hash the entire vllm codebase to address corner cases that are not guarded by traced files.

This PR adds the basic functionality to hash over the entire codebase (with some heuristics on what are the meaningful source files to be included). It doesn't change the current caching behavior (yet) and we can figure out how to refactor the caching code in discussion.

Test Plan:
time python vllm/compilation/source_checksum.py (measured <1sec on my local machine) pytest tests/compile/test_source_checksum.py

Reviewers:

Subscribers:

Tasks:

Tags:

## Test Result

---
<details>

</details>

### DIFF
--- a/tests/compile/test_source_checksum.py
+++ b/tests/compile/test_source_checksum.py
@@ -1,0 +1,140 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+from pathlib import Path
+
+import pytest
+
+from vllm.compilation.source_checksum import (
+    calculate_vllm_source_checksum,
+    calculate_vllm_source_factors,
+    should_exclude_dir,
+)
+
+
+def test_should_exclude_dir():
+    assert should_exclude_dir(".git")
+    assert should_exclude_dir(".github")
+    assert should_exclude_dir(".pytest_cache")
+
+    assert should_exclude_dir("__pycache__")
+    assert should_exclude_dir(".mypy_cache")
+    assert should_exclude_dir(".ruff_cache")
+
+    assert not should_exclude_dir("vllm")
+    assert not should_exclude_dir("csrc")
+    assert not should_exclude_dir("tests")
+    assert not should_exclude_dir("benchmarks")
+
+
+def test_calculate_vllm_source_checksum_consistency():
+    """Test that checksum calculation is consistent across multiple runs."""
+    factors1 = calculate_vllm_source_factors()
+    checksum1 = calculate_vllm_source_checksum(factors1)
+
+    factors2 = calculate_vllm_source_factors()
+    checksum2 = calculate_vllm_source_checksum(factors2)
+
+    assert checksum1 == checksum2
+    # Checksum should be a SHA256 hex string (64 characters)
+    assert len(checksum1) == 64
+
+    assert factors1 == factors2
+
+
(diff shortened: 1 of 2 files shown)


vllm/compilation/source_checksum.py
```diff
@@ -0,0 +1,128 @@
...
+    root_str_len = len(root_str)
+
+    # Scan all files in root directory
+    for file_path in root_path.rglob("*"):
+        if not file_path.is_file():
+            continue
+
+        file_str = str(file_path)
+        assert file_str.startswith(root_str)
+        relative_str = file_str[root_str_len:]
```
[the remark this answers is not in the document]
zhxchen17: Calling relative_to is super slow in my local test (takes > 1sec in total). I can figure out a better way to handle cross platform tho.

vllm/compilation/source_checksum.py
```diff
@@ -0,0 +1,127 @@
...
+    ".h",
+    ".hpp",
+    ".hxx",
+    ".cmake",
+    ".txt",
+    ".toml",
+}
+
+
+# Directories to exclude at top level only
```
ZJY0516:
```suggestion
# Directories to include at top level only
```

LucasWilkinson:
cc @ProExpertProg

tests/compile/test_source_checksum.py
```diff
@@ -0,0 +1,177 @@
...
+
+    # Should not exclude normal directories
+    assert not should_exclude_dir("vllm")
+    assert not should_exclude_dir("csrc")
+    assert not should_exclude_dir("tests")
+    assert not should_exclude_dir("benchmarks")
+
+
+def test_calculate_vllm_source_checksum_consistency():
+    """Test that checksum calculation is consistent across multiple runs."""
```
zou3519, line 30: these comments directly under the `def` are redundant, the test name is self-explanatory

vllm/compilation/source_checksum.py
```diff
@@ -0,0 +1,127 @@
...
+    ".c",
+    ".cu",
+    ".cuh",
+    ".h",
+    ".hpp",
+    ".hxx",
+    ".cmake",
+    ".txt",
+    ".toml",
+}
```
zou3519, line 27: @zhxchen17 do you have a pointer to how torch does this? I'm curious

ProExpertProg (approved):
LGTM once @zou3519's comments are addressed

ProExpertProg:
> It doesn't change the current caching behavior (yet) and we can figure out how to refactor the caching code in discussion.

I think the wheel build process should invoke this function and produce a `repo_compile_hash.txt` file (or a `.py` file with a variable that defines it) in the `vllm` folder, and then during hashing we can check for existence of this file - if it exists, we use it, if not, we manually hash.

vllm/compilation/source_checksum.py
zhxchen17: discussed offline with @zou3519, seems torch only filters .py files. Since it's possible for a developer to change cuda kernels and rebuild vllm, I think we might want to also include cpp/cuda files to prevent subtle caching issues.

PR #28742, opened 2025-11-14
[ci][amd] fix EPLB execution test

bradleyhd:
## Purpose
To fix the failing EPLB execution test for AMD CI. This test uses multiprocessing (defaulting to fork method), but tries to set cuda devices in each process. NV seems to be forgiving of this, but ROCM runtime is more strict. "spawn" MP start method should be used. In order to accomplish this, I've removed most local functions (so MP can pickle them), and set the MP start method to 'spawn'.

## Test Plan
AMD CI. `test_eplb_execute` passes (https://buildkite.com/vllm/amd-ci/builds/1064/steps/canvas?jid=019a84a1-e8cb-444f-bd2b-f7f239a02978).
Please note the overall test still fails, in a second command against `distributed/test_eplb_spec_decode.py`. This is a more complex fix and can be addressed in a follow-up PR

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #29119, opened 2025-11-20
[ROCm][CI] Fix "Cannot re-initialize CUDA in forked subprocess" in test_pynccl.py 

micah-wil:
Similar to https://github.com/vllm-project/vllm/pull/28742, `test_pynccl.py` was also failing on AMD CI due to the following error when running `pytest -v -s distributed/test_pynccl.py`:

```
  File "/usr/lib/python3.12/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/usr/lib/python3.12/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/projects/vllm/tests/distributed/test_pynccl.py", line 54, in wrapped_fn
    torch.cuda.set_device(device)
  File "/usr/local/lib/python3.12/dist-packages/torch/cuda/__init__.py", line 567, in set_device
    torch._C._cuda_setDevice(device)
  File "/usr/local/lib/python3.12/dist-packages/torch/cuda/__init__.py", line 398, in _lazy_init
    raise RuntimeError(
RuntimeError: Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method
```

If we just set `multiprocessing.set_start_method("spawn", force=True)` as the error suggests, we then run into pickling errors due to nested functions in the test. We simply switch to `import multiprocess` which can handle this without further changes to the test due to the `dill` pickling backend. Multiprocess is already a dependency in `requirements/test.txt.`


tjtanaa (approved):

DarkLight1337:
The docker build keeps failing, PTAL

micah-wil:
@DarkLight1337 Fixed the multiprocess version error that was causing build error, thanks for taking a look.

tests/distributed/test_pynccl.py
```diff
@@ -1,9 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 
-import multiprocessing
 import os
```
DarkLight1337, line 5: I wonder whether `torch.multiprocessing` would work...

micah-wil: Well, it _can_ work, but the test would have to be refactored otherwise you get pickling errors like this:
```
distributed/test_pynccl.py:37: in distributed_run
    p.start()
/usr/lib/python3.12/multiprocessing/process.py:121: in start
    self._popen = self._Popen(self)
                  ^^^^^^^^^^^^^^^^^
/usr/lib/python3.12/multiprocessing/context.py:224: in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/usr/lib/python3.12/multiprocessing/context.py:289: in _Popen
    return Popen(process_obj)
           ^^^^^^^^^^^^^^^^^^
/usr/lib/python3.12/multiprocessing/popen_spawn_posix.py:32: in __init__
    super().__init__(process_obj)
/usr/lib/python3.12/multiprocessing/popen_fork.py:19: in __init__
    self._launch(process_obj)
/usr/lib/python3.12/multiprocessing/popen_spawn_posix.py:47: in _launch
    reduction.dump(process_obj, fp)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

obj = <Process name='Process-4' parent=93586 initial>
file = <_io.BytesIO object at 0x7eec45ebb240>, protocol = None

    def dump(obj, file, protocol=None):
        '''Replacement for pickle.dump() using ForkingPickler.'''
>       ForkingPickler(file, protocol).dump(obj)
E       AttributeError: Can't get local object 'worker_fn_wrapper.<locals>.wrapped_fn'
```

If you look at this PR you will see what I mean (it overcame the same issues as those addressed in this PR but for a different test, and they chose to go the route that you are suggesting): https://github.com/vllm-project/vllm/pull/28742. It has to be refactored because both `multiprocessing` and `torch.multiprocessing` use a pickling backend that can not pickle nested functions. `multiprocess` uses `dill` which can handle nested functions. As far as I can tell, there is actually no benefit to using `torch.multiprocessing` over `multiprocessing` for a simple case such as this. I do see the downside of baking in `multiprocess` as a dependency (although it already is listed as one in `requirements/test.txt`, but I don't think it is explicitly used anywhere else), so if you would prefer that I refactor this test similarly to how https://github.com/vllm-project/vllm/pull/28742 does, I don't mind doing that. I just chose to go this route because imo `multiprocess` will allow us to write cleaner tests in these types of cases moving forward.

Please let me know your thoughts, and lmk if I am missing anything here. Thanks again for taking a look!

DarkLight1337:
Ok let's go with the current approach for now then

PR #29125, opened 2025-11-20
[Feature] Batch invariant: Enable `TRITON_MLA` without prefix-caching

yewentao256:
## Purpose

Enable `TRITON_MLA` without prefix-caching

It involves very large change if we want to support the kernel with prefix-caching, I will have a follow up issue for this.

### Other unit tests

```bash
(wentao) wentao@dgxB200-09:~/vllm-source/tests/v1/determinism$ pytest
==================================== test session starts ====================================
platform linux -- Python 3.12.3, pytest-9.0.1, pluggy-1.6.0
rootdir: /home/wentao/vllm-source
configfile: pyproject.toml
plugins: anyio-4.11.0
collected 105 items                                                                         

test_batch_invariance.py .............                                                [ 12%]
test_online_batch_invariance.py ...                                                   [ 15%]
test_rms_norm_batch_invariant.py .................................................... [ 64%]
.....................................                                                 [100%]

======================= 105 passed, 22 warnings in 647.81s (0:10:47) ========================
```

### R1

`VLLM_TEST_MODEL=deepseek-ai/DeepSeek-R1 VLLM_TP_SIZE=8 pytest test_online_batch_invariance.py::test_logprobs_bitwise_batch_invariance_bs1_vs_bsN[TRITON_MLA]`

```bash
test_online_batch_invariance.py .                                                     [100%]

========================= 1 passed, 2 warnings in 209.41s (0:03:29) =========================
```


vllm/v1/core/sched/scheduler.py
```diff
@@ -176,10 +178,16 @@ def __init__(
                 self.num_lookahead_tokens = self.num_spec_tokens
 
         # Create the KV cache manager.
+        enable_caching = bool(self.cache_config.enable_prefix_caching)
+
+        # TODO(wentao): fix prefix caching for batch invariance of TRITON_MLA.
+        if vllm_is_batch_invariant() and envs.VLLM_ATTENTION_BACKEND == "TRITON_MLA":
+            enable_caching = False
```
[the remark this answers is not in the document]
yewentao256: Fixed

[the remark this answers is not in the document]
yewentao256: Rocm is not supported with batch invariant

vllm/v1/core/sched/scheduler.py
```diff
@@ -176,10 +178,22 @@ def __init__(
                 self.num_lookahead_tokens = self.num_spec_tokens
 
         # Create the KV cache manager.
+        enable_caching = bool(self.cache_config.enable_prefix_caching)
+
+        # TODO(wentao): fix prefix caching for batch invariance of TRITON_MLA.
+        if vllm_is_batch_invariant() and envs.VLLM_ATTENTION_BACKEND == "TRITON_MLA":
```
mgoin: We should not be checking the environment variable as this isn't set when we choose the backend automatically. I think this only works for your test since you set the environment variable i.e. `monkeypatch.setenv("VLLM_ATTENTION_BACKEND", backend)`
This might need to wait for Matt's work to make a proper AttentionConfig https://github.com/vllm-project/vllm/pull/26315

Also, this decision should not be happening in the scheduler as it is too late. This should happen after attention selection to update `self.cache_config.enable_prefix_caching` itself

mgoin (changes requested):

Jakub227 (approved):

vllm/v1/core/sched/scheduler.py
yewentao256: I have fixed this and could you take another look? @mgoin 
I don't think we need to wait until #26315 as this is needed for batch invariant MLA

tests/v1/determinism/test_batch_invariance.py
```diff
@@ -185,7 +185,7 @@ def test_logprobs_bitwise_batch_invariance_bs1_vs_bsN(
     llm = LLM(
         model=model_name,
         tensor_parallel_size=tp_size,
-        enable_prefix_caching=False,
+        # enable_prefix_caching=False,
```
mgoin, line 188: Remove the comment

mgoin (approved):
LGTM

PR #29128, opened 2025-11-21
[Cleanup] Refactor FlashInferMetadataBuilder

benchislett:
## Purpose

Reorganize the FlashInferMetadata into clear `prefill` and `decode` sections that either belong to FlashInfer or TRTLLM execution pathways. This separation is desirable because it allows us to make explicit which metadata needs to be prepared for each backend, and therefore which computations can be omitted when a certain backend is not used.

As such, I refactor (and make skip-able) some computations related to the paged kv indices, see `_compute_paged_kv_indices`.

## Test Plan

I run MTBench for spec acceptance rate and GSM8k for accuracy benchmarking on the following configurations with both the new and old code, and with both TRTLLM and FI-native pathways:
- Llama 3.1-8B no-spec
- Llama 3.1-8B spec decoding
- Llama 3.1-8B decode-context-parallel (help needed, I have never run this before)

In addition, I will annotate the metadata build and forward calls via NVTX to microbenchmark the CPU and GPU runtime change for Llama 3.1-8B no-spec with both pathways. 

## Test Result

Launch with:
```
vllm serve meta-llama/Llama-3.1-8B-Instruct --speculative-config '{"method": "eagle3", "model": "yuhuili/EAGLE3-LLaMA3.1-Instruct-8B", "num_speculative_tokens": 3}' --async-scheduling --max-num-seqs 128 --no-enable-prefix-caching
```
- with/without spec,
- with/without async scheduling,
- with/without VLLM_USE_TRTLLM_ATTENTION=0

```
lm_eval \
  --model local-completions \
  --tasks gsm8k \
  --model_args base_url=http://0.0.0.0:8049/v1/completions,model=meta-llama/Llama-3.1-8B-Instruct,tokenized_requests=False,tokenizer_backend=None,num_concurrent=32,timeout=120,max_retries=5
```

Passes in all cases:

```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.7741|±  |0.0115|
|     |       |strict-match    |     5|exact_match|↑  |0.7028|±  |0.0126|
```

MTBench eval with and without VLLM_USE_TRTLLM_ATTENTION:

```
python3 examples/offline_inference/spec_decode.py --method eagle3 --model-dir meta-llama/Llama-3.1-8B-Instruct --eagle-dir yuhuili/EAGLE3-LLaMA3.1-Instruct-8B --tp 1 --num-spec-tokens 3 --dataset-name hf --dataset-path philschmid/mt-bench --num-prompts 80
```

```
total_num_output_tokens: 17154
num_drafts: 6157
num_draft_tokens: 18418
num_accepted_tokens: 10919
mean acceptance length: 2.77
acceptance at token 0: 0.77
acceptance at token 1: 0.57
acceptance at token 2: 0.43
```


vllm/v1/attention/backends/utils.py
```diff
@@ -100,6 +100,32 @@ class CommonAttentionMetadata:
     dcp_local_seq_lens_cpu: torch.Tensor | None = None
     """Sequence lengths of the local rank in decode context parallelism world"""
 
+    # TODO(lucas): remove once we have FULL-CG spec-decode support
```
benchislett, line 133: cherry-picked bugfix

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -848,91 +946,107 @@ def build(
...
+                    qo_indptr_prefill_cpu[1:] - qo_indptr_prefill_cpu[:-1]
                 )
-                # Since prefill_wrapper.run() will be called with
-                # query[num_decode_tokens:] we need to adjust the qo_indptr
-                # to be relative to the start of the prefill queries.
-                qo_indptr_cpu = (
-                    qo_indptr_cpu[prefill_start:] - qo_indptr_cpu[prefill_start]
+                max_q_len_prefill = int(query_lens_prefill_cpu.max().item())
+                attn_metadata.prefill = TRTLLMPrefill(
+                    block_tables=block_table_tensor[prefill_start:],  # todo
```
benchislett:
```suggestion
                    block_tables=block_table_tensor[prefill_start:],
```

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -380,40 +381,103 @@ def get_required_kv_cache_layout(cls) -> KVCacheLayoutType | None:
...
-    block_table_tensor: torch.Tensor
-    prefill_use_trtllm: bool
-    decode_use_trtllm: bool
+    """
+    The slice of the sequence lengths tensor corresponding *only* to decode requests.
+    Shape: [num_decodes]
+    """
+
+    max_seq_len: int
+    """The maximum sequence length (context length) for KV."""
```
pavanimajety: nit: calling it context length may be slightly confusing. "The maximum sequence length for KV Cache(`[max_seq_len, num_kv_heads, head_dim]`)". 

Same comment for usage in TRTLLMPrefill

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -532,34 +597,14 @@ def __init__(
...
-            max_num_pages, dtype=torch.int32, device="cpu", pin_memory=pin_memory
-        )
-        self.paged_kv_last_page_len_cpu = torch.zeros(
-            max_num_reqs, dtype=torch.int32, device="cpu", pin_memory=pin_memory
-        )
-        self.paged_kv_last_page_len_np = self.paged_kv_last_page_len_cpu.numpy()
+        # Preparing persistent buffers
+        self.pin_memory = is_pin_memory_available()
+        self.paged_kv_indptr = self._make_buffer(max_num_reqs + 1)
+        self.paged_kv_indptr_cpu_buffer = torch.zeros_like(
```
pavanimajety, line 606: Why do we need a copied buffer here as opposed to just using paged_kv_indptr.cpu?

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -761,12 +762,42 @@ def build(
...
+                decode_threshold=self.reorder_batch_threshold,
+                require_uniform=True,
+            )
+        )
+
+        page_size = self.page_size
+        max_seq_len = common_attn_metadata.max_seq_len
+        seq_lens = common_attn_metadata.seq_lens
+        seq_lens_cpu = common_attn_metadata.seq_lens_cpu
+        block_table_tensor = common_attn_metadata.block_table_tensor
```
pavanimajety, line 794: is it worth adding an assertion / check here to ensure seq_lens == seq_lens_cpu?

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -809,28 +847,88 @@ def build(
...
+            q_data_type=self.q_data_type,
             num_decodes=num_decodes,
             num_decode_tokens=num_decode_tokens,
             num_prefills=num_prefills,
             num_prefill_tokens=num_prefill_tokens,
+            prefill=None,
+            decode=None,
             use_cascade=use_cascade,
+            cascade_wrapper=None,
         )
```
pavanimajety, line 868: Suggestion: For better readability, does it make sense to initialize the MetaData once prefill, decode, cascade_wrapper's individual metadata is initialized? Or may be add a comment here saying that it will be done in the next steps

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -658,99 +715,43 @@ def _get_cascade_wrapper(self):
             )
         return self._cascade_wrapper
 
-    def build(
+    def _compute_paged_kv_indices(
```
pavanimajety: Nit: rename to `compute_flashinfer_metadata` to be more explicit for what kv_indices here mean

pavanimajety (approved):
Minor comments, LGTM. @LucasWilkinson , @MatthewBonanni for another set of eyes.

vllm/v1/attention/backends/flashinfer.py
MatthewBonanni: I think we wouldn't want to introduce a sync here since the TRTLLM backend also runs through here

vllm/v1/attention/backends/flashinfer.py
benchislett: This is a pre-existing problem. We modify paged_kv_indptr_cpu when padding for cuda graphs, so we need to make a separate buffer so we can safely make the gpu->cpu copy without it getting overwritten while in flight.

vllm/v1/attention/backends/flashinfer.py
benchislett: I think it's easier to keep it this way, since it's nice to have cascade attention as an early-exit pathway to reduce indentation / branching. In which case it's best to keep it pre-initialized so that we don't duplicate the initialization code block. I can definitely tidy this up so it's clearer what's happening here!

vllm/v1/attention/backends/flashinfer.py
benchislett: +1, we would have to copy the gpu tensor to the cpu to perform this check. We should avoid doing that wherever possible.

benchislett: Might make sense to have a debug-mode-only assertion but that seems like overkill

vllm/v1/attention/backends/flashinfer.py
```diff
@@ -761,12 +762,42 @@ def build(
...
+                common_attn_metadata,
+                decode_threshold=self.reorder_batch_threshold,
+                require_uniform=True,
+            )
+        )
+
+        page_size = self.page_size
+        max_seq_len = common_attn_metadata.max_seq_len
+        seq_lens = common_attn_metadata.seq_lens
+        seq_lens_cpu = common_attn_metadata.seq_lens_cpu
```
LucasWilkinson: nit: should we move this access to where its actually used to avoid triggering a sync post: https://github.com/vllm-project/vllm/pull/29624 and https://github.com/vllm-project/vllm/pull/29845 ?

LucasWilkinson (approved):
Overall looks good to me! thanks for doing this!

mgoin (approved):

PR #29129, opened 2025-11-21
[BugFix] skip combo kernel on cpu

### LINKED ISSUE #28982: [Bug]: CPU Backend: vllm / torch.compile issue with torch > 2.8
### Your current environment

<details>

```text
```

</details>

### 🐛 Describe the bug

If we run vllm on CPU Backend with `torch==2.9` we hit this inductor error:
```
(EngineCore_DP0 pid=73386)   File "/Users/mgoin/code/venvs/vllm/lib/python3.12/site-packages/torch/_inductor/scheduler.py", line 5345, in _codegen_partitions
(EngineCore_DP0 pid=73386)     self._codegen(partition)
(EngineCore_DP0 pid=73386)   File "/Users/mgoin/code/venvs/vllm/lib/python3.12/site-packages/torch/_inductor/scheduler.py", line 5440, in _codegen
(EngineCore_DP0 pid=73386)     raise AssertionError(f"{type(self)=}")
(EngineCore_DP0 pid=73386) torch._inductor.exc.InductorError: AssertionError: type(self)=<class 'torch._inductor.scheduler.Scheduler'>
```

This is not the case with `torch==2.8` which is the current version used in `vllm/requirements/cpu.txt`
However it's problem for Darwin which uses `torch==2.9` and a blocker to updating to `torch==2.9`

BoyuanFeng:
combo kernel does not support cpu yet. we have a fix but need pytorch 2.10. This pr skips combo kernel on cpu to unblock users.

Close #28982

cc @Radu2k @zou3519 @angelayi @ProExpertProg

### DIFF
--- a/vllm/config/compilation.py
+++ b/vllm/config/compilation.py
@@ -8,7 +8,6 @@ from dataclasses import asdict, field
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, ClassVar, Literal
 
-import torch
 from pydantic import TypeAdapter, field_validator
 from pydantic.dataclasses import dataclass
 
@@ -666,7 +665,7 @@ class CompilationConfig:
             and "combo_kernels" not in self.inductor_compile_config
             and "benchmark_combo_kernel" not in self.inductor_compile_config
             # (fixme @boyuan) combo kernel does not support cpu yet.
-            and torch.cuda.is_available()
+            and not current_platform.is_cpu()
         ):
             # use horizontal fusion, which is useful for fusing qk-norm and
             # qk-rope when query and key have different shapes.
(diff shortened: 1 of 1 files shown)


vllm/config/compilation.py
```diff
@@ -664,6 +664,8 @@ def __post_init__(self) -> None:
             is_torch_equal_or_newer("2.9.0.dev")
             and "combo_kernels" not in self.inductor_compile_config
             and "benchmark_combo_kernel" not in self.inductor_compile_config
+            # (fixme @boyuan) combo kernel does not support cpu yet.
+            and not current_platform.is_cpu()
```
zou3519, line 668: should use is_torch_equal_or_newer("2.10.0.dev") after the fix lands in pytorch

zou3519 (approved):

vllm/config/compilation.py
ProExpertProg: @BoyuanFeng can you submit followup after this lands?

ProExpertProg (approved):

vllm/config/compilation.py
BoyuanFeng: yes will do

PR #23795, opened 2025-08-28
[CI] Fail subprocess tests with root-cause error

njhill:
QOL improvement

We use a `create_new_process_for_each_test` decorator on many tests to avoid cleanup issues. However when such tests fail the pytest error is from the process failure not the actual test failure. This change propagates the actual test exception of interest so that it's reported by pytest as if the test was running inline.

Before:
image

After:
image

PR #7054, opened 2024-08-02
[ci][distributed] try to fix pp test

youkaichao:

Issue #7053, opened 2024-08-01
[Misc]: how to fix the proper cleanup problem in tests


youkaichao:
### Anything you want to discuss about vllm.

We are having a very difficult time in cleaning up resources properly, especially in distributed inference. This makes our tests suffer recently.

To understand this, we have to understand the process model of `pytest`:

For this simple test:

```python
import pytest
@pytest.mark.parametrize('arg', [1, 2, 3])
def test_pass(arg):
    import os
    print((arg, os.getpid()))
```

`pytest` will create one process, to run all the three tests one by one.

So the output is (Note the process id is the same for three tests):

```text
testf.py::test_pass[1] (1, 15068)
PASSED
testf.py::test_pass[2] (2, 15068)
PASSED
testf.py::test_pass[3] (3, 15068)
PASSED
```

The fact that these three tests share the same process, makes some low-level handling difficult.

1. when some test segfaults, the following test will not run because the process died. For example:

```python
import pytest
@pytest.mark.parametrize('arg', [1, 2, 3])
def test_pass(arg):
    import os
    print((arg, os.getpid()))

    if arg == 2:
        import ctypes
        func_ptr = ctypes.CFUNCTYPE(ctypes.c_int)(0)
        # Calling the function pointer with an invalid address will cause a segmentation fault
        func_ptr()
```

this test produces:

```text
Running 3 items in this shard: testf.py::test_pass[1], testf.py::test_pass[2], testf.py::test_pass[3]

testf.py::test_pass[1] (1, 24492)
PASSED
testf.py::test_pass[2] (2, 24492)
Fatal Python error: Segmentation fault
```

So test 3 is not executed in this case. Arguably, this is okay. If test 2 fails with segment fault, we should definitely investigate the reason and fix it.

2. when some tests make the process dirty, the process cannot be used for later tests anymore. For example:

```python
import pytest
@pytest.mark.parametrize('arg', [1, 2, 3])
def test_pass(arg):
    import torch
    assert not torch.cuda.is_initialized()
    data = torch.ones(10, 10).cuda()
    print(data)
    assert data.sum().item() == 100
```

In this example, every test need a clean process ( without cuda initialized ), but it will make the process dirty after the test finishes.

The process can also be dirty, if some objects are not garbage-collected, and GPU memory is not released.

[...]

PR #29130, opened 2025-11-21
[Test] Fix @create_new_process_for_each_test("fork") in interactive shell pipeline

markmc:
Various engine core tests use @create_new_process_for_each_test() which defaults to "fork" mode:

```
$ pytest -vs tests/v1/engine/test_engine_core.py::test_engine_core
...
Running 1 items in this shard: tests/v1/engine/test_engine_core.py::test_engine_core

tests/v1/engine/test_engine_core.py::test_engine_core Fork a new process to run a test 3071367
...
PASSED

===================== 1 passed, 18 warnings in 31.07s ========================
```

If you attempt to capture the output in a pipeline with `tee`:

```
$ pytest -vs tests/v1/engine/test_engine_core.py::test_engine_core 2>&1 | tee output.log
...
INFO 04-15 07:27:52 [gpu_model_runner.py:6072] Graph capturing finished in 1 secs, took 0.04 GiB
INFO 04-15 07:27:52 [gpu_worker.py:597] CUDA graph pool memory: 0.04 GiB (actual), 0.07 GiB (estimated), difference: 0.03 GiB (81.0%).
INFO 04-15 07:27:52 [core.py:299] init engine (profile, create kv cache, warmup model) took 3.83 s (compilation: 0.96 s)
Terminated
```

This appears to be because in an interactive shell with job control enabled, tee is also part of the same process group as the python process so while the python process uses SIG_IGN to try to ensure it doesn't get killed, the tee process gets killed so the whole pipeline fails.

We can confirm this by showing it working fine with job control disabled:

```
$ set +m
$ pytest -vs tests/v1/engine/test_engine_core.py::test_engine_core 2>&1 | tee output.log
...
======================= 1 passed, 18 warnings in 14.95s ========================
```

The fix is to put the child process in its own process group, and terminate that ... rather than terminating the parent and relying on SIG_IGN.

Related past PRs: #23795 and #7054 (#7053)

/cc @njhill @youkaichao

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1261,9 +1261,6 @@ def fork_new_process_for_each_test(func: Callable[_P, None]) -> Callable[_P, Non
 
     @functools.wraps(func)
     def wrapper(*args: _P.args, **kwargs: _P.kwargs) -> None:
-        # Make the process the leader of its own process group
-        # to avoid sending SIGTERM to the parent process
-        os.setpgrp()
         from _pytest.outcomes import Skipped
 
         # Create a unique temporary file to store exception info from child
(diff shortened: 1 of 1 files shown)


robertgshaw2-redhat (approved):
nice investigation!

PR #29133, opened 2025-11-21
Optimizations for beam_search

mgoin:
## Purpose

* Set `detokenize=False`  in SamplingParams during the beam search since we manually detokenize at the end
* Add `skip_clone` parameter to SamplingParams to skip the defensive clone in processor when trusted
* Slightly tweak the code in `append_logprobs_for_next_position` to boost perf for default case (using FlatLogprobs was slower due to the amount of indexing needed)
* Cache tokenizer related variables that trigger the tokenizer such as `get_eos_token_id`. These were tokenizing the eos text in every preprocess call for each request
* Cache `is_encoder_decoder` as it is expensive to read through the model's config
* Add a special case for `kv_cache_manager::get_computed_blocks` where we reuse the previous result when consecutive requests are the same, allowing us to skip find_longest_cache_hit for the rest of the requests

Benchmark command for beam_search with n=30 on H100:
```
vllm serve Qwen/Qwen3-8B --max-logprobs 100
python benchmarks/benchmark_beam_search_server.py \
    --base-url http://localhost:8000 \
    --beam-width 30 \
    --max-tokens 8 \
    --input-len 128 \
    --num-requests 10
```

Before:
```
Average latency: 282.96ms
Median latency: 282.68ms
Std deviation: 10.59ms
Min latency: 263.52ms
Max latency: 299.33ms
```

After
```
Average latency: 223.68ms
Median latency: 226.31ms
Std deviation: 15.06ms
Min latency: 200.60ms
Max latency: 243.34ms
```

## Test Result

---
<details>

</details>

### DIFF
--- a/benchmarks/benchmark_beam_search_server.py
+++ b/benchmarks/benchmark_beam_search_server.py
@@ -25,6 +25,7 @@ import time
 
 import requests
 from openai import OpenAI
+from tqdm import tqdm
 
 
 def start_profile(base_url: str) -> bool:
@@ -60,31 +61,27 @@ def send_beam_search_request(
     temperature: float = 0.0,
 ) -> dict | None:
     """Send a beam search completion request."""
-    try:
-        start_time = time.perf_counter()
-        completion = client.completions.create(
-            model=model,
-            prompt=prompt,
-            n=beam_width,  # Beam width
-            max_tokens=max_tokens,
-            temperature=temperature,
-            extra_body={"use_beam_search": True},  # This triggers beam search
-        )
-        end_time = time.perf_counter()
+    start_time = time.perf_counter()
+    completion = client.completions.create(
+        model=model,
+        prompt=prompt,
+        n=beam_width,  # Beam width
+        max_tokens=max_tokens,
+        temperature=temperature,
+        extra_body={"use_beam_search": True},  # This triggers beam search
+    )
+    end_time = time.perf_counter()
 
-        latency = end_time - start_time
+    latency = end_time - start_time
 
-        # Convert completion object to dict-like structure for compatibility
-        num_choices = len(completion.choices) if completion.choices else 0
+    # Convert completion object to dict-like structure for compatibility
+    num_choices = len(completion.choices) if completion.choices else 0
 
-        return {
-            "latency": latency,
-            "result": completion,
-            "num_choices": num_choices,
-        }
-    except Exception as e:
-        print(f"✗ Request failed: {e}")
-        return None
+    return {
+        "latency": latency,
+        "result": completion,
+        "num_choices": num_choices,
+    }
 
 
 def generate_prompt(input_len: int, seed: int | None = None) -> str:
@@ -171,16 +168,13 @@ def main():
 
     args = parser.parse_args()
 
-    print("=" * 80)
     print("Beam Search Server Benchmark")
-    print("=" * 80)
     print(f"Server URL: {args.base_url}")
     print(f"Beam width (n): {args.beam_width}")
     print(f"Max tokens: {args.max_tokens}")
     print(f"Input length: ~{args.input_len} tokens")
     print(f"Number of requests: {args.num_requests}")
     print(f"Temperature: {args.temperature}")
-    print("=" * 80)
 
     # Initialize OpenAI client
     openai_api_base = f"{args.base_url}/v1"
@@ -235,32 +229,27 @@ def main():
             print("Warning: Could not start profiling. Continuing anyway...")
         time.sleep(1)  # Brief pause to ensure profiler is ready
 
-    # Send requests
     print(f"\nSending {args.num_requests} requests...")
     latencies = []
     successful_requests = 0
-
-    for i in range(args.num_requests):
+    for _ in tqdm(range(args.num_requests)):
         # Generate a unique random prompt for each request
         prompt = args.prompt or generate_prompt(args.input_len, seed=None)
-
-        print(f"\nRequest {i + 1}/{args.num_requests}...", end=" ", flush=True)
         result = send_beam_search_request(
-            client=client,
-            model=model_name,
-            prompt=prompt,
-            beam_width=args.beam_width,
-            max_tokens=args.max_tokens,
-            temperature=args.temperature,
+            client,
+            model_name,
+            prompt,
+            args.beam_width,
+            args.max_tokens,
+            args.temperature,
         )
 
         if result:
             latency = result["latency"]
             latencies.append(latency)
             successful_requests += 1
-            print(f"✓ Latency: {latency:.4f}s ({latency * 1000:.2f}ms)")
         else:
-            print("✗ Failed")
+            print("✗ Failed request")
 
     # Stop profiling
     if args.profile:
@@ -271,9 +260,7 @@ def main():
     if latencies:
         import statistics
 
-        print("\n" + "=" * 80)
-        print("Benchmark Results")
-        print("=" * 80)
+        print("\nResults")
         print(f"Successful requests: {successful_requests}/{args.num_requests}")
         print(f"Average latency: {statistics.mean(latencies) * 1000:.2f}ms")
         print(f"Median latency: {statistics.median(latencies) * 1000:.2f}ms")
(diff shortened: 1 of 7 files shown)


vllm/logprobs.py
```diff
@@ -186,21 +185,16 @@ def append_logprobs_for_next_position(
     # We do not need a special case for the sampled token
     # being in the topk, since inserting duplicated data
     # into a dictionary twice is the same as doing it once.
-    topk_ranks = range(1, num_logprobs + 1)
-    ranks = itertools.chain((rank,), topk_ranks)
+    ranks = [rank] + list(range(1, num_logprobs + 1))
```
hmellor, line 188: This is about 1.4x faster if we want to micro-optimise

```suggestion
    ranks = [rank, *range(1, num_logprobs + 1)]
```

hmellor (approved):
LGTM, just one micro-optimisation nit

benchmarks/benchmark_beam_search_server.py
```diff
@@ -0,0 +1,277 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""Simple benchmark script to test beam search performance via OpenAI API.
```
heheda12345, line 4: I think we have can benchmark beam search with vllm bench latency and vllm bench throughput

vllm/v1/core/kv_cache_manager.py
```diff
@@ -198,11 +216,34 @@ def get_computed_blocks(self, request: Request) -> tuple[KVCacheBlocks, int]:
...
         max_cache_hit_length = request.num_tokens - 1
-        computed_blocks, num_new_computed_tokens = (
-            self.coordinator.find_longest_cache_hit(
-                request.block_hashes, max_cache_hit_length
+
+        # Optimization: Reuse previous result when consecutive requests are the same.
+        if (
+            request.block_hashes
+            and self._cache_hit_cache is not None
+            and max_cache_hit_length == self._cache_hit_cache.max_cache_hit_length
```
heheda12345, line 224: Does this constraint mean we need to recompute `find_longest_cache_hit` for each beam search iteration?

vllm/v1/core/kv_cache_manager.py
```diff
@@ -198,11 +216,34 @@ def get_computed_blocks(self, request: Request) -> tuple[KVCacheBlocks, int]:
...
+            and len(request.block_hashes) == self._cache_hit_cache.num_block_hashes
+            and request.block_hashes[-1] == self._cache_hit_cache.last_block_hash
```
heheda12345, line 226: does this mean the optimization works only when only the last partial block of the current request and the previous request is different and has no effect if different candidates have different prefix at the beginning? Do you know the cache hit rate of this optimization?

vllm/v1/core/kv_cache_manager.py
```diff
@@ -198,11 +216,34 @@ def get_computed_blocks(self, request: Request) -> tuple[KVCacheBlocks, int]:
         # num_computed_tokens to be block-size aligned. Removing this limitation
         # could slightly improve performance in the future.
         max_cache_hit_length = request.num_tokens - 1
-        computed_blocks, num_new_computed_tokens = (
-            self.coordinator.find_longest_cache_hit(
-                request.block_hashes, max_cache_hit_length
+
+        # Optimization: Reuse previous result when consecutive requests are the same.
```
heheda12345, line 220: can you make this optimization optional to beam search? e.g., add a flag?

vllm/v1/core/kv_cache_manager.py
```diff
@@ -198,11 +216,34 @@ def get_computed_blocks(self, request: Request) -> tuple[KVCacheBlocks, int]:
...
+        if (
+            request.block_hashes
+            and self._cache_hit_cache is not None
+            and max_cache_hit_length == self._cache_hit_cache.max_cache_hit_length
+            and len(request.block_hashes) == self._cache_hit_cache.num_block_hashes
+            and request.block_hashes[-1] == self._cache_hit_cache.last_block_hash
+        ):
```
heheda12345, line 227: We can reuse the computed_blocks of the previous request only when we can ensure these blocks are not evicted.
1. If the model only has full attention, I guess it's fine but want some proof on it
2. If the model has sliding window attention, I feel there may be some problem as the previous request may free the early blocks outside the sliding window and make these blocks evocable.

heheda12345:
We are using prefix caching to do beam search. Does it mean with block_size 16 and prompt length 30, we always need to recompute the prefill of 14 tokens?

vllm/sampling_params.py
```diff
@@ -618,8 +624,13 @@ def clone(self) -> "SamplingParams":
         data that is expensive to copy. However, if not copied, the processor
         needs to support parallel decoding for multiple sequences
         See https://github.com/vllm-project/vllm/issues/3087
+
```
njhill, line 627: We can also cleanup the logitsprocessor stuff in this method since they aren't supported in SamplingParams anymore

vllm/entrypoints/openai/serving_engine.py
```diff
@@ -395,6 +395,8 @@ async def beam_search(
             logprobs=logprobs_num,
             max_tokens=1,
             temperature=temperature,
+            detokenize=False,  # We detokenize the output after the search
+            skip_clone=True,  # Safe to reuse params without cloning
```
njhill, line 399: I think we could set this for all of the API server endpoints

benchmarks/benchmark_beam_search_server.py
mgoin: I didn't realize, will try to use it

vllm/v1/core/kv_cache_manager.py
mgoin: Yes, each beam search iteration submits `n` identical requests to the engine core and the core is unaware of this

vllm/v1/core/kv_cache_manager.py
mgoin: The intent of this optimization is just to deal with the case when the current request is identical to the previous request. My assumption was that because a given block_hash "includes" all the previous blocks, I could just look at the last one to check for equality. This is a huge boost to beam search n=30 because all requests perform the same search

Here are profiles showing that "Self duration" of `schedule` decreases from 6.060ms to 1.471ms 

Before:
Screenshot 2025-11-24 at 2 52 51 PM

After:
Screenshot 2025-11-24 at 2 53 11 PM

vllm/logprobs.py
Jialin: As the construction of ranks are fixed, I was thinking about simply get rid of this list, and just pass in the sampled rank here.

Then we could just do
```
# insert sampled token
...
# insert top ranked tokens in order
for rank, (token, logprob) in enumerate(zip(tokens[1:], logprobs[1:]))
```

vllm/sampling_params.py
```diff
@@ -231,6 +231,12 @@ class SamplingParams(
     set to an integer k, will use only the last k tokens from the prompt
     (i.e., left truncation). If set to `None`, truncation is disabled."""
     output_kind: RequestOutputKind = RequestOutputKind.CUMULATIVE
+    skip_clone: bool = False
+    """Internal flag indicating that this SamplingParams instance is safe to
+    reuse without cloning. When True, clone() will return self without
+    performing a deep copy. This should only be set when the params object
+    is guaranteed to be dedicated to a single request and won't be modified
+    in ways that would affect other uses."""
```
Jialin, line 239: IIUC, this should be helpful for n-gen and beam search, correct?

And just curious, what's the risk to make it default to true? Or even make it as a consistent behaviour?

vllm/v1/core/kv_cache_manager.py
Jialin: @mgoin @heheda12345 I thought about it before, and would love to see if it's possible. For find_longest_cache_hit, should we do binary search instead of linear search?

In this way, maybe we could consolidate the code without special handling for beam search?

vllm/entrypoints/openai/serving_engine.py
```diff
@@ -395,6 +395,8 @@ async def beam_search(
             logprobs=logprobs_num,
             max_tokens=1,
             temperature=temperature,
+            detokenize=False,  # We detokenize the output after the search
```
Jialin, line 398: +1 on this. I'm wondering if we should directly remove tokenized string in Logprob. IMO,
- detokenization should never live inside Engine.
- clients are rarely used the data, if they do, it should be cleaner to do it on their side instead.

PR #24949, opened 2025-09-16
[Hybrid Allocator] Support KV cache groups with different block_size

heheda12345:
## Purpose
Hybrid allocator requires all layer has the same physical memory per block now. But models like https://github.com/vllm-project/vllm/issues/24916 (bf16 for sliding window attention and fp8 for full attention) will have different memory per layer.

This pr supports these cases by giving different layers different block_sizes to make the physical memory per block the same. Require one layer's memory per block is a multiple of the other now.

To support prefix caching, we need:
1. make sure the prefix caching hit length is a multiple of all block_sizes. This is achieved by introducing an alignment factor to `get_longest_cache_hit` and set the alignment requirement to the LCM of all block_sizes.
2. generate the BlockHash of each block size from the BlockHash with `block_size=cache_config.block_size`.

For 2, we can generate the block hash with a larger block_size from that with a smaller block size. For example, with block hash of block_size 16, we can get the block hash with block_hash 32 by concatenating two hash value with block_size 16 to one hash value with block_size 32:

block_hash with block_size 16:
|  token   | 0-15 | 16-31   | 32-47 | 48-63 |
|----------|-----|------------|----|----|
| Hash    | A  | B  | C| D|

block_hash with block_size 32:
|  token   | 0-31 | 32-63   |
|----------|-----|------------|
| Hash    | AB | CD  |

Note: for non-hybrid model with different hidden size per layer like https://github.com/vllm-project/vllm/issues/22432 , we may still keep the block size the same for all layers. I plan to do it in a future PR.

## Test Plan
Set the kv_dtype of either sliding window attention or full attention to fp8 and run
```
test_correctness_sliding_window.py::test_sliding_window_retrieval[False-1-5-google/gemma-3-1b-it] 
```

And also necessary unit tests.

## Test Result
Success
---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

Issue #24916, opened 2025-09-15
[Feature]: Make FP8 Attention fast for GPT-OSS w/ FA3 on Hopper


jmkuebler:
### 🚀 The feature, motivation and pitch

When using `kv_cache_dtype="fp8` on Hopper, we are using FA3 FP8 backend. This means we are quantizing Keys, Values, and Queries and are using the complete FP8 forward pass of FA3.

Whilst this saves KV cache space and leads to TTFT speedups it currently leads to slowdowns in ITL which are quite considerable (see below)

The biggest issue are that quantizing the queries and the K, V causes overheads. Furthermore FP8 path of FA3 is very bad for decoding for GPT-OSS (head_dim=64).

To adress this we propose 3 optimizations:
1/ Only quantize the full attention layers and leave the sliding window layers in BF16 --> #24912 (specific to GPT-OSS architecture)
2/ [merged] Add an environment flag such that for static scales torch.compile can fuse the quantization of the queries into previous operations --> #24914  (applies generally)
3/ [merged] Optimize the tiling configuration for head_dim 64 of FA3 --> https://github.com/vllm-project/flash-attention/pull/91 

We run `bench serve` with 25k and 1k inputs to illustrate the overheads during decoding (always running with `"cudagraph_mode": "FULL_AND_PIECEWISE"`). On mainline turning on KV quantization increases ITL at 1k context from 4.83ms to 5.06ms.
With our proposed optimizations, this get's optimized to 4.86ms. 

Already at 25k inputs we then start seeing actual speedups from FP8 quantization.

With the skipping of SW layers quantization, we quantize less than normally. We also checked the accuracy via gpt-oss evals and are within the usual variations.

| model                                            | applied optimization | input  | output | median ITL [ms] | median ttft [ms] |
|--------------------------------------------------|----------------------|--------|--------|-----------------|------------------|
| gpt-oss (full bf 16)                             |        mainline              |  25000 |    200 |            5.08 |          1252.43 |
| gpt-oss (full bf 16)                             |       mainline               |   1000 |    200 |            4.83 |            76.16 |
| kv fp8                                           |        mainline              |  25000 |    200 |            5.38 |          1211.33 |
| kv fp8                                           |       mainline               |   1000 |    200 |            5.06 |            76.36 |
| gpt-oss-KV8-skip-sw                              |                    1 |  25000 |    200 |            5.23 |           1210.6 |
[...]

Issue #22432, opened 2025-08-07
[Feature]: Support Eagle Draft Model with different number of KV heads


MMuzzammil1:
### 🚀 The feature, motivation and pitch

The current EAGLE implementation raises `NotImplementedError` (https://github.com/vllm-project/vllm/blob/8e8e0b6af189d262bcfdaef6c0cfb94772e86b0b/vllm/v1/core/kv_cache_utils.py#L1098), because of different  `page_size_bytes` calculation ( 
https://github.com/vllm-project/vllm/blob/main/vllm/v1/kv_cache_interface.py#L68 ) when using a draft model with different `num_kv_heads` compared to the target model. 
However in practice, we can have draft models trained with different `num_kv_heads` than the target model. Would that be possible to implement this change? Or it would require a lot of architectural changes?

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

PR #29143, opened 2025-11-21
[Hybrid Allocator] Support KV cache groups with different block_size

ivanium:
## Purpose

This PR rebases #24949, authored by @heheda12345, onto the latest main branch and includes minor fixes along with updated docstrings and comments. For additional background and discussion, please refer to #24949.

> [!NOTE] 
> Below is a copy of the description of #24949

Hybrid allocator requires all layer has the same physical memory per block now. But models like https://github.com/vllm-project/vllm/issues/24916 (bf16 for sliding window attention and fp8 for full attention) will have different memory per layer.

This pr supports these cases by giving different layers different block_sizes to make the physical memory per block the same. Require one layer's memory per block is a multiple of the other now.

To support prefix caching, we need:
1. make sure the prefix caching hit length is a multiple of all block_sizes. This is achieved by introducing an alignment factor to `get_longest_cache_hit` and set the alignment requirement to the LCM of all block_sizes.
2. generate the BlockHash of each block size from the BlockHash with `block_size=cache_config.block_size`.

For 2, we can generate the block hash with a larger block_size from that with a smaller block size. For example, with block hash of block_size 16, we can get the block hash with block_hash 32 by concatenating two hash value with block_size 16 to one hash value with block_size 32:

block_hash with block_size 16:
|  token   | 0-15 | 16-31   | 32-47 | 48-63 |
|----------|-----|------------|----|----|
| Hash    | A  | B  | C| D|

block_hash with block_size 32:
|  token   | 0-31 | 32-63   |
|----------|-----|------------|
| Hash    | AB | CD  |

Note: for non-hybrid model with different hidden size per layer is supported in https://github.com/vllm-project/vllm/issues/22432

## Test Plan
Set the kv_dtype of either sliding window attention or full attention to fp8 and run
```
test_correctness_sliding_window.py::test_sliding_window_retrieval[False-1-5-google/gemma-3-1b-it]
```

And also necessary unit tests.

## Test Result
Success

---
<details>

</details>


vllm/v1/core/kv_cache_utils.py
```diff
@@ -1342,4 +1381,60 @@ def get_kv_cache_configs(
         if len(kv_cache_config.kv_cache_groups) > 0:
             _report_kv_cache_config(vllm_config, kv_cache_config)
 
+    # TODO: Remove before merging. Only for testing.
+    print("kv_cache_configs", kv_cache_configs[0])
+
```
ivanium: Reminder: remove before merging

vllm/attention/layer.py
```diff
@@ -260,6 +260,12 @@ def __init__(
             kv_cache_dtype = "auto"
             block_size = 16
             calculate_kv_scales = False
+
+        # TODO: Remove before merging. Only for testing.
+        if sliding_window is None:
+            print("set kv_cache_dtype to fp8_e4m3 for layer", prefix)
+            kv_cache_dtype = "fp8_e4m3"
```
ivanium: Reminder: remove before merging

vllm/v1/core/block_pool.py
```diff
@@ -223,8 +227,15 @@ def cache_full_blocks(
             return
         new_full_blocks = blocks[num_cached_blocks:num_full_blocks]
         assert len(request.block_hashes) >= num_full_blocks
-        new_block_hashes = request.block_hashes[num_cached_blocks:]
+        if block_size == self.hash_block_size:
+            block_hashes: BlockHashList = request.block_hashes
+        else:
+            assert block_size % self.hash_block_size == 0
+            block_hashes = BlockHashListWithBlockSize(
```
heheda12345, line 243: Add a comment like this? You can also try to make it more clear. I feel it's hard for first-time reader to get what this piece of code is doing now.
```suggestion
            # `block_size` of `block_hashes` and `blocks` doesn't match. Generate `block_hashes` with  `block_size` of `blocks` from the original block_hashes.
            block_hashes = BlockHashListWithBlockSize(
```

vllm/v1/core/kv_cache_coordinator.py
```diff
@@ -414,25 +436,39 @@ def find_longest_cache_hit(
                 - The number of tokens of the longest cache hit.
         """
         # First, find the longest cache hit for full attention.
+        if self.full_attention_spec.block_size == self.hash_block_size:
+            full_attention_block_hashes: BlockHashList = block_hashes
+        else:
+            full_attention_block_hashes = BlockHashListWithBlockSize(
```
heheda12345, line 458:
```suggestion
            # Get the block_hash with full_attention_spec.block_size as block_size
            full_attention_block_hashes = BlockHashListWithBlockSize(
```

vllm/v1/core/kv_cache_coordinator.py
```diff
@@ -414,25 +436,39 @@ def find_longest_cache_hit(
...
+            alignment=self.lcm_block_size,
         )
         hit_length = len(hit_blocks_full_attn[0]) * self.full_attention_block_size
 
         # Next, find the cache hit for the other attention WITHIN
         # the cache hit of full attention.
+        if self.other_spec.block_size == self.hash_block_size:
+            other_block_hashes: BlockHashList = block_hashes
+        else:
+            other_block_hashes = BlockHashListWithBlockSize(
```
heheda12345, line 481:
```suggestion
            # Get the block_hash with other_spec.block_size as block_size
            other_block_hashes = BlockHashListWithBlockSize(
```

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -207,14 +207,15 @@ def get_num_common_prefix_blocks(self, running_request_id: str) -> int:
...
-        block_hashes: list[BlockHash],
+        block_hashes: BlockHashList,
         max_length: int,
         kv_cache_group_ids: list[int],
         block_pool: BlockPool,
         kv_cache_spec: KVCacheSpec,
         use_eagle: bool,
         dcp_world_size: int = 1,
         pcp_world_size: int = 1,
+        alignment: int = 1,
```
heheda12345:
```suggestion
        token_alignment: int = 1,
```

A bit confused about whether this alignment is for number of tokens or number of blocks. (please remember to update the comment)

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -405,6 +417,8 @@ def find_longest_cache_hit(
             ):
                 for computed, cached in zip(computed_blocks, cached_block):
                     computed[i] = cached
+                if num_contiguous_blocks == 0 and (i + 1) * block_size % alignment != 0:
```
heheda12345: should we move this before 
```
           for computed, cached in zip(computed_blocks, cached_block):
                    computed[i] = cached
```

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -636,6 +665,11 @@ def find_longest_cache_hit(
             if cached_block := block_pool.get_cached_block(
                 block_hashes[i], kv_cache_group_ids
             ):
+                # For Mamba, a cache hit only depends on finding the cached
+                # state for an exact position (i). Positions that are not
+                # properly aligned are ignored.
+                if (i + 1) % alignment != 0:
```
heheda12345: is it correct? alignment is number of token but i is number of block

vllm/v1/core/kv_cache_utils.py
```diff
@@ -1342,4 +1381,60 @@ def get_kv_cache_configs(
...
     return kv_cache_configs
+
+
+class BlockHashListWithBlockSize:
+    """
+    Convert the block hashes under hash_block_size to another target_block_size.
+    Only support scaling up the block size by an integer factor now. Implemented
+    by concatenating the block hashes under hash_block_size to form that of
+    target_block_size.
```
heheda12345: maybe we should add the example in the PR description to here? And I think it's worth mention that this is a "lazy" conversion.

heheda12345 (commented):
Thanks for your contribution!

vllm/v1/core/kv_cache_utils.py
ivanium: Great idea!

vllm/v1/core/single_type_kv_cache_manager.py
ivanium: I actually think your original logic is correct because `computed` is part of `computed_blocks` and will eventually be returned by this function. So we should still copy all `cached` blocks in prefix (L418:419), even if we will skip checking prefix match (L420) for some blocks.

vllm/v1/core/single_type_kv_cache_manager.py
ivanium: Nice catch! I think I misunderstood `block_size` for Mamba. I'll fix this and also double check this with you later.

ivanium (commented):
I left some comments but somehow they became part of a "review". Anyways I submit them here

vllm/v1/core/kv_cache_coordinator.py
```diff
@@ -414,25 +440,46 @@ def find_longest_cache_hit(
...
             block_pool=self.block_pool,
             kv_cache_spec=self.full_attention_spec,
             use_eagle=self.use_eagle,
+            alignment_tokens=self.lcm_block_size,
         )
         hit_length = len(hit_blocks_full_attn[0]) * self.full_attention_block_size
 
         # Next, find the cache hit for the other attention WITHIN
         # the cache hit of full attention.
+        if self.other_spec.block_size == self.hash_block_size:  # Common case.
```
heheda12345:
```suggestion
        if self.other_spec.block_size == self.hash_block_size:
            # Common case.
```

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -396,15 +407,25 @@ def find_longest_cache_hit(
...
         for i in range(max_num_blocks - 1, -1, -1):
             if cached_block := block_pool.get_cached_block(
                 block_hashes[i], kv_cache_group_ids
             ):
+                # Add the cached block to the computed blocks.
                 for computed, cached in zip(computed_blocks, cached_block):
                     computed[i] = cached
+                # Skip prefix matching check if the block is not aligned with
+                # `alignment_tokens`. But we still need to copy `cached` block above
+                # to keep its content.
```
heheda12345: let's sync offline

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -396,15 +407,25 @@ def find_longest_cache_hit(
...
+                if (
```
heheda12345, line 424:
```suggestion
                if (block_size != alignment_tokens and ....)
```
use this to avoid unnecessary overhead for common case?

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -421,7 +442,13 @@ def find_longest_cache_hit(
             # `num_contiguous_blocks < sliding_window_contiguous_blocks`.
             for computed in computed_blocks:
                 del computed[num_contiguous_blocks:]
+            while len(computed_blocks[0]) * block_size % alignment_tokens != 0:
```
heheda12345: same `block_size != alignment_tokens`

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -524,6 +556,10 @@ def find_longest_cache_hit(
         )
         assert dcp_world_size == 1, "DCP not support chunked local attn now."
         assert pcp_world_size == 1, "PCP not support chunked local attn now."
+        assert kv_cache_spec.block_size % alignment_tokens == 0, (
```
heheda12345:
```suggestion
        assert kv_cache_spec.block_size != alignment_tokens == 0, (
```
I think it should be !=

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -636,6 +673,11 @@ def find_longest_cache_hit(
             if cached_block := block_pool.get_cached_block(
                 block_hashes[i], kv_cache_group_ids
             ):
+                # For Mamba, a cache hit only depends on finding the cached
+                # state for an exact position (i). Positions that are not
+                # properly aligned are ignored.
+                if (i + 1) * kv_cache_spec.block_size % alignment_tokens != 0:
```
heheda12345: same `block_size != alignment_tokens`

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -421,7 +442,13 @@ def find_longest_cache_hit(
             # `num_contiguous_blocks < sliding_window_contiguous_blocks`.
             for computed in computed_blocks:
                 del computed[num_contiguous_blocks:]
+            while len(computed_blocks[0]) * block_size % alignment_tokens != 0:
+                for computed in computed_blocks:
+                    computed.pop()
         if use_eagle and computed_blocks[0]:
+            assert kv_cache_spec.block_size % alignment_tokens == 0, (
```
heheda12345:
```suggestion
            assert kv_cache_spec.block_size != alignment_tokens == 0, (
```

ivanium:
Made another pass. PTAL. Thanks!

heheda12345 (commented):
LGTM! Can you remove the debug code? I can make another pass after that.

ivanium:
Cleaned the code. PTAL. Thank you!

heheda12345 (approved):
LGTM! Thank you very much!

mgoin (approved):

PR #29144, opened 2025-11-21
[Fix] Add SM check to flashinfer MOE backend

jiahanc:
## Purpose
Add SM version check to flashinfer MOE backend because Flashinfer TRTLLM MOE only support SM100 and later.
## Test Plan
Not functional change, just a check and fallback so to test plan
## Test Result

---
<details>

</details>


mgoin (approved):

vllm/model_executor/layers/quantization/utils/flashinfer_utils.py
```diff
@@ -282,6 +282,15 @@ def get_flashinfer_moe_backend() -> FlashinferMoeBackend:
...
     flashinfer_moe_backend = envs.VLLM_FLASHINFER_MOE_BACKEND
     if flashinfer_moe_backend in backend_map:
+        if (
+            flashinfer_moe_backend == "latency"
+            and not current_platform.is_device_capability(100)
+        ):
+            logger.info_once(
+                "Flashinfer TRTLLM MOE backend is only supported on "
+                "SM100 and later, using CUTLASS backend instead"
+            )
```
yewentao256, line 293:
```suggestion
            logger.info_once(
                "Flashinfer TRTLLM MOE backend is only supported on "
                "SM100 and later, using CUTLASS backend instead",
                scope="local"
            )
```
To avoid duplicate logs across process

yewentao256 (commented):
Thanks for the work!

jiahanc:
> Does FlashInfer CUTLASS support GPU before Hopper?

Pre-Hopper is not supported.

yewentao256 (approved):

mgoin (approved):

vllm/model_executor/layers/quantization/utils/flashinfer_utils.py
```diff
@@ -282,6 +282,16 @@ def get_flashinfer_moe_backend() -> FlashinferMoeBackend:
 
     flashinfer_moe_backend = envs.VLLM_FLASHINFER_MOE_BACKEND
     if flashinfer_moe_backend in backend_map:
+        if (
+            flashinfer_moe_backend == "latency"
+            and not current_platform.is_device_capability(100)
+        ):
+            logger.info_once(
+                "Flashinfer TRTLLM MOE backend is only supported on "
+                "SM100 and later, using CUTLASS backend instead",
```
emirisman, line 291: `is_device_capability` checks if the CUDA CC is exactly SM100 and would fail if used with SM120, but the log message suggests it being supported on SM100 _and_ later?

jiahanc: Thanks! Fixed in https://github.com/vllm-project/vllm/pull/28238

PR #29160, opened 2025-11-21
Tool Call Parser logs should not contain user input / model output except on DEBUG

sfbemerk:
For data privacy reason, logs by vllm should not contain user inputs or model outputs by default, maybe except for DEBUG purposes. 
Since the logger in tool call parser implementations does not respect the `--no-enable-log-requests` startup argument, the parsers should not log any input or output on WARN level or higher. 

I propose replacing all logs that contain user input or model output in the tool call parser implementations with `logger.debug`. 
Alternatively, we could leave the logs on WARN and just drop the input/output data.


chaunceyjiang (approved):
Thanks~

PR #29161, opened 2025-11-21
[CI/Build] Enable eager mode for Samplers-Test group

qli88:
This PR is to enable eager mode for "Samplers Test" test group. For now model distilbert/distilgpt2 does not work correctly in graph mode and will break CI. We will fix this graph-mode problem ASAP. (The issue is #30801)


qli88:
@mgoin This issue happens only on ROCm platform. We will fix it asap and after that will reverse this change.

tests/samplers/test_no_bad_words.py
```diff
@@ -94,7 +94,7 @@ def setup_method(self, method):
         )[0]
 
     def test_two_token_bad_word(self, vllm_runner):
-        with vllm_runner(self.MODEL, dtype="half") as llm:
```
SageMoore, line 97: Can you change this so that we are only running eager on ROCm? Is there an issue tracking this failure? It would be good to have a link to an issue here.

qli88: Done. And also I created an issue (https://github.com/vllm-project/vllm/issues/30801) to track.

tests/samplers/test_logprobs.py
```diff
@@ -25,7 +26,12 @@ def test_ranks(
     flat_logprobs,
     example_prompts,
 ):
-    with vllm_runner(model, dtype=dtype, max_logprobs=MAX_LOGPROBS) as vllm_model:
+    # TODO: Remove once graph mode is fixed for distilbert/distilgpt2 on ROCm.
```
robertgshaw2-redhat, line 29: Can we open an issue to track this and link it here?

AndreasKaratzas: @robertgshaw2-redhat I think there is this one here: https://github.com/vllm-project/vllm/issues/30801 
But feel free to correct me if I got it wrong.

PR #29162, opened 2025-11-21
[BugFix] EPLB + B200 + DeepGEMM : Handle column-major scales tensor

varun-sundar-rabindranath:
## Purpose
On B200, the following command fails, 
```
VLLM_ALL2ALL_BACKEND=deepep_low_latency VLLM_USE_DEEP_GEMM=1  vllm serve  Qwen/Qwen3-30B-A3B-FP8 --trust-remote-code --tensor-parallel-size 1 --data-parallel-size 4 --enable-expert-parallel --no-enable-prefix-caching  --port 9010 --enable-eplb  --eplb-config '{"window_size":10,"step_interval":100,"num_redundant_experts":0,"log_balancedness":true}' --enforce-eager
```

On B200, when we use DeepGEMM, we transpose the MoE weight_scale tensors for efficient DeepGEMM matmuls. This in combination with EPLB fails the following assertion, 
```
        assert all(
            weight.is_contiguous()
            for name, weight in weights
            if not name.startswith("_shared_experts.")
        )
```
as the weight scale tensors are not contiguous.

Fix: This PR changes to view of the tensor so the `is_contiguous()` check passes. We also add a test to verify that this view update is safe.

## Test Plan
`tests/distributed/test_eplb_fused_moe_layer.py` passes

###  lm-eval test
```
server-command : VLLM_ALL2ALL_BACKEND=deepep_low_latency VLLM_USE_DEEP_GEMM=1  vllm serve  Qwen/Qwen3-30B-A3B-FP8 --trust-remote-code --tensor-parallel-size 1 --data-parallel-size 4 --enable-expert-parallel --no-enable-prefix-caching  --port 9010 
```
```
lm_eval --model local-completions --tasks gsm8k --model_args model=Qwen/Qwen3-30B-A3B-FP8,base_url=http://localhost:9010/v1/completions,num_concurrent=30,max_retries=3 --limit 100
```

## Test Result
server command + `--enable-eplb  --eplb-config '{"window_size":10,"step_interval":100,"num_redundant_experts":0,"log_balancedness":true}'`
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  | 0.88|±  |0.0327|
|     |       |strict-match    |     5|exact_match|↑  | 0.92|±  |0.0273|
```
server command (without eplb)
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  | 0.87|±  |0.0338|
|     |       |strict-match    |     5|exact_match|↑  | 0.89|±  |0.0314|
```

### DIFF
--- a/vllm/model_executor/layers/fused_moe/layer.py
+++ b/vllm/model_executor/layers/fused_moe/layer.py
@@ -1391,7 +1391,48 @@ class FusedMoE(CustomOp):
                     yield param_name
 
     def get_expert_weights(self) -> Iterable[torch.Tensor]:
+        def _maybe_make_contiguous(
+            name: str, p: torch.nn.Parameter
+        ) -> torch.nn.Parameter:
+            """
+            In some cases, the last 2 dimensions (the non-expert dimensions)
+            of the weight scale tensor are transposed. This function
+            transforms the tensor (view update) so the tensor is contiguous().
+            Example: A non-contiguous scale tensor,
+              `x` of shape (E, 32, 16) and stride (512, 1, 32) is transformed to
+              `x_` of shape (E, 16, 32) and stride (512, 32, 1).
+              Note that we specifically use torch.transpose() so `x_` refers
+              to the same underlying memory. The tensors `x` and `x_`, pointing
+              to the same underlying memory make this transformation safe in the
+              context of EPLB. i.e. It is the same memory and just the view
+              is different.
+            Note: This function handles the "weight_scale" tensors specifically.
+            This could however be generalized to handle similar tensors.
+            """
+            if p.ndim != 3:
+                return p
+            if p.is_contiguous():
+                # Already contiguous. do nothing.
+                return p
+            # p is non-contiguous. We only handle the case where the last 2
+            # dimensions of the scales tensor is transposed. We can handle
+            # other cases when they become relevant.
+            is_transposed_12 = p.stride(1) == 1 and p.stride(2) != 1
+            if "weight_scale" not in name or not is_transposed_12:
+                # do nothing.
+                return p
+
+            # Do not update the layer paramater as the layer's MoE operations would
+            # expect the parameter's tensor to the same shape / stride. Instead,
+            # make a new torch.nn.Parameter that is used just in the context of
+            # EPLB.
+            return torch.nn.Parameter(
+                torch.transpose(p.data, 1, 2), requires_grad=False
+            )
+
         weights = list(self.named_parameters())
+        weights = [(name, _maybe_make_contiguous(name, p)) for name, p in weights]
+
         assert all(
             weight.is_contiguous()
             for name, weight in weights
--- a/tests/distributed/eplb_utils.py
+++ b/tests/distributed/eplb_utils.py
@@ -1,0 +1,49 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+import os
+import random
+
+import torch
+import torch.multiprocessing as mp
+
+from vllm.distributed.parallel_state import (
+    init_distributed_environment,
+)
+from vllm.utils.system_utils import update_environment_variables
+
+mp.set_start_method("spawn", force=True)
+
+
+def distributed_run(fn, world_size, *args):
+    number_of_processes = world_size
+    processes: list[mp.Process] = []
+    for i in range(number_of_processes):
+        env: dict[str, str] = {}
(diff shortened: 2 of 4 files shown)


vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1225,7 +1225,40 @@ def load_weights(
...
+        def _maybe_make_contiguous(
+            name: str, p: torch.nn.Parameter
+        ) -> torch.nn.Parameter:
+            """
+            In some cases, the last 2 dimensions (the non-expert dimensions)
+            of the weight scale tensor are transposed. This function transposes
+            the tensor back so the tensor is contiguous().
+            Example: A scale tensor,
+              `x` of shape (E, 32, 16) and stride (512, 1, 32) is transposed to
+              `xt` of shape (E, 16, 32) and stride (512, 32, 1).
```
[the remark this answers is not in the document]
varun-sundar-rabindranath: reworded the comment.

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1225,7 +1225,36 @@ def load_weights(
...
+            Note: This function handles the "weight_scale" tensors specifically.
+            This could however be generalized to handle similar tensors.
+            """
+            # Check if the last 2 dimensions are trasposed
+            is_transposed = p.stride(1) == 1 and p.stride(2) != 1
+            if p.is_contiguous() or not is_transposed or "weight_scale" not in name:
+                # do nothing.
+                return p
+            assert p.ndim == 3
+            p.data = torch.transpose(p.data, 1, 2)
```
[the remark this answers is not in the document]
varun-sundar-rabindranath: Nice catch 🙌

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1391,7 +1391,42 @@ def load_weights(
...
+            # expect the parameter's tensor to the same shape / stride. Instead,
+            # make a new torch.nn.Parameter that is used just in the context of
+            # EPLB.
+            return torch.nn.Parameter(
+                torch.transpose(p.data, 1, 2), requires_grad=False
+            )
+
         weights = list(self.named_parameters())
+        weights = [(name, _maybe_make_contiguous(name, p)) for name, p in weights]
+
```
ilmarkov, line 1435: Probably, instead of `is_contiguous` check we need to check that we are row-major with `num_local_experts` in the first dimension. The fact that tensor is contiguous in other dimensions is not important as we flatten the view in those dimensions.

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1391,7 +1391,42 @@ def load_weights(
...
+              to the same underlying memory. The tensors `x` and `x_`, pointing
+              to the same underlying memory make this transformation safe in the
+              context of EPLB. i.e. It is the same memory and just the view
+              is different.
+            Note: This function handles the "weight_scale" tensors specifically.
+            This could however be generalized to handle similar tensors.
+            """
+            if p.ndim != 3:
+                return p
+            # Check if the last 2 dimensions are trasposed
```
tlrmchlsmth:
```suggestion
            # Check if the last 2 dimensions are transposed
```

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1391,7 +1391,42 @@ def load_weights(
...
+            is_transposed = p.stride(1) == 1 and p.stride(2) != 1
+            if p.is_contiguous() or not is_transposed or "weight_scale" not in name:
+                # do nothing.
+                return p
```
tlrmchlsmth: `p.is_contiguous()` should already cover the `is_transposed` check

```suggestion
            if p.is_contiguous() or "weight_scale" not in name:
                # do nothing.
                return p
```

tlrmchlsmth (approved):

varun-sundar-rabindranath:
Looks like pre-commit is failing on unrelated files in vllm PRs - Saw the same failure here https://github.com/vllm-project/vllm/pull/29188

vllm/model_executor/layers/fused_moe/layer.py
varun-sundar-rabindranath: This will catch the case where the tensor is non-contiguous and is transposed on dimensions other than [1, 2] (maybe [0, 1] / [0, 2]). I have refactored the checks to avoid confusion.

abmfy (approved):
The current fix LGTM. Thanks!

mgoin (approved):

PR #23673, opened 2025-08-26
[Core] Use sha256 bytes instead of BlockHash to reduce GC overhead

linzebing:
## Purpose

The current `BlockHash` design stores the hash value alongside a tuple of all token IDs and any extra keys to prevent collisions, and `BlockHashWithGroupId` layers a group identifier on top of that structure. Every block therefore retains a tuple of Python integers for its tokens, creating many short‑lived objects. This design reduces collision risk, but it can significantly increase memory pressure and trigger frequent garbage collection.

image

@Jialin's analysis: On average, there are floor(1500/16)=93 blocks per request when input prompt is 1500 and block_size is 16. As shown above, there are 4 objects to be collected per block, so there’re 372 objects to be collected per request. Given that, the default GC threshold is 700 (i.e. a GC collect might be kicked off when it collected more than 700 objects. **1.88**=700/372 requests might kick off a GC collect event which is too frequent.

Replacing these classes with bytes hashes (e.g., <sha256> and <sha256><group_id>) would eliminate the per-block token tuple and nested NamedTuples, greatly reducing object counts. A full 256‑bit SHA‑256 hexdigest makes collisions virtually impossible, so token IDs and extra keys no longer need to be embedded in the hash structure.

`bytes` digest is preferred over `str` as it avoids extra work to convert SHA-256 digest into a string while also being more memory efficient. Replacing hash_fn from `hash` to `sha256` incurs extra CPU overhead, but benchmarking shows that it's worth the effort as it greatly reduces number of short-lived objects and therefore reduces GC overheads.

### Performance (H100)
Used @Jialin's https://github.com/Jialin/vllm/pull/4 for GC analysis.
```
CUDA_VISIBLE_DEVICES=7 VLLM_USE_V1=1 HF_HUB_DISABLE_XET=1 python3 benchmarks/benchmark_throughput.py \
    --model facebook/opt-125m \
    --backend vllm \
    --input-len 1500 \
    --output-len 150 \
    --num-prompts 10000
```
```
CUDA_VISIBLE_DEVICES=7 VLLM_USE_V1=1 HF_HUB_DISABLE_XET=1 python3 benchmarks/benchmark_throughput.py \
    --model facebook/opt-125m \
    --backend vllm \
    --input-len 1500 \
    --output-len 1 \
    --num-prompts 20000
```

Hash methods benchmark: https://gist.github.com/linzebing/bd51f09a183bfd53c7b856ebf412fe68

### Eval
```
[...]

PR #17045, opened 2025-04-23
[Core] Prevent side-channel attacks via cache salting

dr75:
# Context
Prefix caching in vLLM improves inference performance by reusing KV blocks across requests. However, this reuse introduces a potential privacy risk in shared environments, where an attacker could infer prompt reuse via timing side channels as demonstrated in [Leaking Secrets from Prefix Caches](https://arxiv.org/html/2411.18191v1). 

To address this, we propose to isolate caches as described in an RFC: https://github.com/vllm-project/vllm/issues/16016

# Suggested change
This PR implements the single barrier approach from the RFC by adding support for an optional `cache_salt` field in the request schema. When present, the salt is injected into the hash of the first block, ensuring that only requests with the same salt can share cached blocks. This effectively segments cache reuse by salt and protects against timing-based attacks.

```json
{
  "messages": [
    {"role": "system", "content": "You are a helpful assistant."},
    {"role": "user", "content": "Here is a document with details about the world series: ..."},
    {"role": "user", "content": "Who won the world series in 2020?"}
  ],
  "cache_salt": "Z3V2bmV3aGxza3ZubGFoZ3Zud3V3ZWZ2bmd0b3V2bnZmc2xpZ3RoZ2x2aQ=="
}
```

The change is compatible with OpenAI requests as only an additional optional field is added. Users can still use the OpenAI client:
```python
response = client.chat.completions.create(
    model=model,
    messages=messages,
    extra_body={
        "cache_salt": "Z3V2bmV3aGxza3ZubGFoZ3Zud3V3ZWZ2bmd0b3V2bnZmc2xpZ3RoZ2x2aQ==",
    },
)
```

The scope of cache sharing can be configured per request as needed, e.g., full single user protection or cache sharing within a group of users.

The change is in line with cache protection applied by other providers such as OpenAI and provides a solution that allows for higher flexibility allowing for more fine-grained configuration.

PR #29163, opened 2025-11-21
[Core] Add xxHash as a high-performance hash option for accelerating prefix caching

LuminolT:
## Purpose

This PR introduces xxHash-based hash algorithms to vLLM’s prefix caching, providing a faster and collision-resistant alternative to the SHA-256 variants added in #23673.

**Why xxHash?** Prefix Caching does not require cryptographic security:

- It never stores secret data.
- It never leaves process boundaries.
- It is used exclusively to identify structural equivalence of prefix blocks.

Therefore, cryptographic collision resistance is unnecessary. 

xxHash strikes the right balance: Much better collision resistance compared to Python builtin `hash()`, which is salted and unstable across processes. Much faster than SHA-256 and SHA-256+CBOR used in current prefix caching.

This change is lightweight: it only introduces the xxhash dependency and adds minimal adapter code to integrate xxHash into the existing hashing interface. We added two small benchmarking scripts to validate correctness and performance, but these can be removed in follow-up cleanup if needed.

⚠️ Only problem is xxHash introduces a third-party dependency. SHA-256 is available via Python standard library [hashlib](https://docs.python.org/3/library/hashlib.html) but xxHash requires an additional Python dependency [xxhash](https://pypi.org/project/xxhash/).

As a result this PR adds xxHash as an optional high-performance hash. Users can opt in with:

```
--prefix-caching-hash-algo xxhash
--prefix-caching-hash-algo xxhash-cbor
```

### 1. Micro benchmark of hash functions

`benchmarks/benchmark_hash.py` is added for benchmarks for hash functions. This could be removed if PR is merged.

This evaluation is same as #23673, except for some configurations for xxHash.

```bash
python benchmarks/benchmark_hash.py --iterations 10000
```

This compares Python's builtin `hash()`, sha256 and xxHash.

### 2. Block-level hash throughput

`benchmarks/benchmark_prefix_block_hash.py` is added for block-level hash throughput evaluation  This could be removed if PR is merged.

This test is model-agnostic, so it reflects performance on larger models as well, rather than only on the small OPT-125M model we evaluated following.

```bash
python benchmarks/benchmark_prefix_block_hash.py --num-blocks 20000 --block-size 64
python benchmarks/benchmark_prefix_block_hash.py --num-blocks 50000 --block-size 16
```
This measures token hashing throughput for: `sha256`, `sha256_cbor`, `xxhash` and `xxhash_cbor`.

### 3. Prefix caching evaluation

This is the benchmark currently provided by vLLM. We tested it using different hash algorithms and compared their results.

Following evaluations are conducted on one-card A100 server.

```bash
python benchmarks/benchmark_prefix_caching.py \
  --model facebook/opt-125m \
  --enable-prefix-caching \
  --prefix-caching-hash-algo <algo> \
  --input-length-range "128:256" \
  --output-len 128 \
  --num-prompts 200 \
  --repeat-count 5 \
  --seed 42
```

### 4. Long-running throughput & GC behavior

Same as #23673. @Jialin 's [Jialin#4](https://github.com/Jialin/vllm/pull/4) is used for GC analysis. Since SHA-256 significantly reduces GC overhead, our goal is to match or further reduce GC time compared to SHA-256.

Input/Output configurations:
- input_len=1500, output_len=1
- input_len=1500, output_len=150

Each run generates logs:

```bash
throughput_sha256_out1_*.log
throughput_sha256_cbor_out1_*.log
throughput_xxhash_out1_*.log
throughput_xxhash_cbor_out1_*.log
...
```

### 1. Micro benchmark of hash functions

```bash
# python benchmarks/benchmark_hash.py --iterations 10000
============================================================
HASH FUNCTION MICRO BENCHMARK
============================================================
Test data: (32-byte bytes object, 32-int tuple)
Iterations: 10,000
============================================================

Results:
  SHA256 (pickle) :     1.45 ±   0.24 μs
  xxHash (pickle) :     0.99 ±   0.16 μs
  built-in hash() :     0.19 ±   0.07 μs

============================================================
SUMMARY (relative to built-in hash())
============================================================
• SHA256 (pickle) is 7.8x slower than built-in hash()
• xxHash (pickle) is 5.3x slower than built-in hash()
```

👉 Among non-builtin hashers, xxHash is ~1.5× faster than SHA-256.

### 2. Block-level hash throughput

20k × 64 tokens and 50k × 16 tokens.

```bash
# python benchmarks/benchmark_prefix_block_hash.py --num-blocks 20000 --block-size 64
Benchmarking 4 algorithms on 20000 blocks (block size=64).
sha256         avg: 0.044319s  best: 0.043920s  throughput: 29.14M tokens/s
sha256_cbor    avg: 0.195202s  best: 0.194645s  throughput: 6.58M tokens/s
xxhash         avg: 0.035225s  best: 0.035059s  throughput: 36.51M tokens/s
xxhash_cbor    avg: 0.184370s  best: 0.183651s  throughput: 6.97M tokens/s
```

👉 xxHash improves block hashing throughput by 25–40%.

### 3. Prefix caching evaluation

| Algo            | Time (s)  |
| --------------- | --------- |
| sha256          | ~4.17     |
| sha256_cbor     | ~4.25     |
| **xxhash**      | **~4.12** |
| **xxhash_cbor** | **~4.23** |

👉 Performance improves 1–2% with xxHash.

We should note that our experiments were conducted only on the small OPT-125M model. In larger production service, where prefix-hit rates are substantially higher, xxHash is expected to show even greater performance advantages.

#### output_len = 1 (prefill-dominant)

image

- sha256: 219.9 req/s
- xxhash: 223.4 req/s (+1.6%)

GC:

- Both ~300 events
- GC runtime ≈ 1.3–1.4% of wall clock
- Main GC distribution almost identical
- No new long-tail events from xxHash

#### output_len = 150 (decode-heavy)

image

- sha256: 77.9 req/s
- xxhash: 78.9 req/s (+1.2%)

GC:

- sha256: total GC ~1.92 s
- xxhash: total GC ~1.68 s (−12.8%)
- xxHash shows fewer and shorter >400ms GC outliers

#### CBOR variants

image

image

Similar behavior:

- xxHash_CBOR slightly faster than SHA256_CBOR
- GC profile remains ~1–2% of wall clock
- No regressions or long-tail amplification

👉 xxHash does not increase GC pressure and may even slightly reduce tail events.

---
<details>

</details>


benchmarks/benchmark_hash.py
ywang96, line 1: Maybe it's worth mentioning these scripts in the documentation.

ywang96 (commented):
@LuminolT Very much appreciate the contribution, especially all the detailed analysis you put out in this PR!

cc @russellb regarding any security concerns.

benchmarks/benchmark_hash.py
LuminolT: Thanks for pointing this out! It will be helpful for evaluating the hash algorithms used in prefix caching. I will update the relevant documentation accordingly.

russellb (changes requested):
First, I appreciate the PR, including the detailed PR description, including performance numbers.

> Why xxHash? Prefix Caching does not require cryptographic security:
> 
> It never stores secret data.
> It never leaves process boundaries.
> It is used exclusively to identify structural equivalence of prefix blocks.
> 
> Therefore, cryptographic collision resistance is unnecessary.

I disagree with this premise.

Prefix cache contents are secret data in a multi-tenant environment. A hash collision would cause undefined behavior and, at worst, leak private information. Consider this scenario:

- Company A sends a request
- Company B sends a request that encounters a hash collision, causing Company B's request to use cache based on Company A's request.

We had one security vulnerability where it was possible to get this to occur predictably: https://github.com/vllm-project/vllm/security/advisories/GHSA-rm76-4mrf-v9r8

This is a bit different, but it is another example of prefix cache security concerns. We had another case where someone demonstrated how to infer information about the contents of the prefix cache by using a timing attack to detect prefix cache hits. See this paper: https://arxiv.org/html/2411.18191v1, this PR: https://github.com/vllm-project/vllm/pull/17045, and this advisory: https://github.com/vllm-project/vllm/security/advisories/GHSA-4qjh-9fv9-r85r.

I'm not opposed to supporting this if it demonstrates a meaningful performance benefit, but I would also like to enhance our help text for the configuration to say that choosing something other than our current default of SHA256 is theoretically less secure in a multi-tenant environment and has a slightly increased risk of undefined behavior from hash collisions otherwise. Further, the code should include an additional comment clarifying the importance of the current default. I can post a separate PR adding this text, but I'd prefer it to be in place before we add other algorithms, as the tradeoffs are important.

requirements/common.txt
```diff
@@ -46,6 +46,7 @@ scipy # Required for phi-4-multimodal-instruct
 ninja # Required for xgrammar, rocm, tpu, xpu
 pybase64 # fast base64 implementation
 cbor2 # Required for cross-language serialization of hashable objects
+xxhash # Required for fast hashing for prefix caching
```
russellb: I would prefer this not be included in our main requirements file, given that it's optional. In fact, it MUST remain optional to avoid getting flagged in environments with strict security controls.

vllm/utils/hashing.py
```diff
@@ -10,6 +10,14 @@
 
 import cbor2
 
+try:
+    import xxhash as _xxhash
```
russellb, line 18: It would be helpful to add a comment here that it is important for this to remain an optional dependency, as it would be considered problematic to include at all in environments with strict security controls.

russellb (changes requested):

requirements/common.txt
russellb:
```suggestion
```

vllm/utils/hashing.py
russellb:
```suggestion
    # It is important that this remains an optional dependency.
    # It would not be allowed in environments with strict security controls,
    # so it's best not to have it installed when not in use.
    import xxhash as _xxhash
```

vllm/config/cache.py
```diff
@@ -79,7 +79,11 @@ class CacheConfig:
     """Set the hash algorithm for prefix caching:\n
     - "sha256" uses Pickle for object serialization before hashing.\n
     - "sha256_cbor" provides a reproducible, cross-language compatible hash. It
-    serializes objects using canonical CBOR and hashes them with SHA-256."""
+    serializes objects using canonical CBOR and hashes them with SHA-256.\n
+    - "xxhash" uses Pickle serialization with xxHash (128-bit) for faster,
+    non-cryptographic hashing. Requires the optional ``xxhash`` package.\n
+    - "xxhash_cbor" combines canonical CBOR serialization with xxHash for
+    reproducible hashing. Requires the optional ``xxhash`` package."""
```
russellb, line 94:
```suggestion
"""Set the hash algorithm for prefix caching:\n
   - "sha256" uses Pickle for object serialization before hashing. This is the
   current default, as SHA256 is the most secure choice to avoid potential
   hash collisions.\n
   - "sha256_cbor" provides a reproducible, cross-language compatible hash. It
   serializes objects using canonical CBOR and hashes them with SHA-256.\n
   - "xxhash" uses Pickle serialization with xxHash (128-bit) for faster,
   non-cryptographic hashing. Requires the optional ``xxhash`` package.
   IMPORTANT: Use of a hashing algorithm that is not considered 
   cryptographically secure theoretically increases the risk of hash collisions,
   which can cause undefined behavior or even leak private information in
   multi-tenant environments. Even if collisions are still very unlikely, it is
   important to consider your security risk tolerance against the performance
   benefits before turning this on.\n
   - "xxhash_cbor" combines canonical CBOR serialization with xxHash for
   reproducible hashing. Requires the optional ``xxhash`` package."""
```

russellb (changes requested):
Instead of a separate PR, I just included some suggested text additions here.

russellb (approved):
I'm happy with this now from a security perspective. Thank you for taking my documentation suggestions.

I would prefer at least one other review, though, to confirm we want to accept more hash options.

LuminolT:
Hi @ywang96, thanks again for the earlier review!

We have now updated the PR to address the security considerations especially the timing side-channel aspect that @russellb pointed out during the discussion. The revised text reflects the conclusions from our joint review, and the documentation/comments have been updated accordingly.

Whenever you have time, could you please take another look at the updated version? Thanks a lot for your time and help!

PR #29165, opened 2025-11-21
[Misc] Move dynamic seed initialization to `EngineArgs`

DarkLight1337:
## Purpose

This allows `ModelConfig.seed` to be statically type annotated as `int` instead of `int | None`.

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/config/model.py
+++ b/vllm/config/model.py
@@ -146,9 +146,12 @@ class ModelConfig:
     - "bfloat16" for a balance between precision and range.\n
     - "float" is shorthand for FP32 precision.\n
     - "float32" for FP32 precision."""
-    seed: int | None = None
-    """Random seed for reproducibility. Initialized to None in V0, but
-    initialized to 0 in V1."""
+    seed: int = 0
+    """Random seed for reproducibility.
+
+    We must set the global seed because otherwise,
+    different tensor parallel workers would sample different tokens,
+    leading to inconsistent results."""
     hf_config: PretrainedConfig = field(init=False)
     """The Hugging Face config of the model."""
     hf_text_config: PretrainedConfig = field(init=False)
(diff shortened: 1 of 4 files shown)


vllm/config/model.py
```diff
@@ -415,7 +418,7 @@ def _apply_dict_overrides(
     def __post_init__(
         self,
         # Multimodal config init vars
-        limit_mm_per_prompt: dict[str, int] | None,
+        limit_mm_per_prompt: dict[str, int | dict[str, int]] | None,
```
DarkLight1337, line 421: Found this type error while scanning for references to `.seed` so decided to fix this along the way

vllm/config/speculative.py
```diff
@@ -316,10 +315,6 @@ def __post_init__(self):
             self.prompt_lookup_min = 0
 
             if self.model is not None:
-                # TODO: Move this import to the top once `ModelConfig`
```
DarkLight1337, line 319: Found this TODO while scanning for references to `.seed` so decided to fix this along the way

vllm/engine/arg_utils.py
hmellor, line 1: What is the purpose of allowing it to be `None` in `EngineArgs`? I wonder if it could just be default 0 always?

DarkLight1337: It is to avoid breaking back-compatibility in case some people pass `seed=None` explicitly. We can deprecate that in another PR.

hmellor: Ok, we could add a validator that handles when None is passed similar to what we have for other fields which used to have delayed init. 

As you say, this can be a follow up.

hmellor (approved):

PR #24799, opened 2025-09-13
[Core] Async Scheduling X Spec Decoding Compatibility

Ronald1995:
## Purpose
PR #19970  implements async_scheduling, PR #23569 implement `prepare_input` overlap base on PR #19970.  RP #24539 refactor the logic of eagle spec_code， make it don't rely on cpu's sample_token_ids.

this PR is based on  #24539 , and aims to support spec decode with async_scheduling. when enable both async_scheduling and spec decode, we won't copy draft token ids to scheduler any more, but cache it in gpu_model_runner, and update the input_ids with the `_draft_token_ids`  directly for next step `execute_model`.

because ngram and medusa rely on cpu's sample_token_ids now, maybe we could refactor it in the future, but now this PR
only support eagle spec_decode with async_scheduling.
## Test Plan
we will make the e2e test.
- async_scheduling + EAGLE-LLaMA3-Instruct-8B draft model, make sure it works well.

Test config:
```python
# dataset is prm800k, read the jsonl and make prompts.
sampling_params = SamplingParams(temperature=0, max_tokens=1024)
llm = LLM(
    model="meta-llama/Meta-Llama-3-8B-Instruct",
    gpu_memory_utilization=0.9,
    tensor_parallel_size=1,
    max_model_len=2048,
    max_num_seqs=128,
    max_num_batched_tokens=4096,
    async_scheduling=True, 
    speculative_config={
            "model": "yuhuili/EAGLE-LLaMA3-Instruct-8B",
            "draft_tensor_parallel_size": 1,
            "num_speculative_tokens": 2,
            "method": "eagle",
        },
    seed=1234
)
```

test device: Nvidia A100

### performance
|num_prompts|async_scheduling(tps)|sync_scheduling(tps)|speedup|
|--|--|--|--|
|24|2356|2314|1.8%|
|48|3759|3539|6.2%|
|96|5110|4770|7.1%|

### precision
I compare the outputs of async_scheduling and sync_scheduling with speculative decoding, 
the outputs are actually the same. so the async_scheduling doesn't make precision problem.

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x] The test plan, such as providing test command.
- [x] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
[...]

PR #29184, opened 2025-11-21
[Core] NGram GPU Implementation compatible with Async Scheduler

PatchouliTIS:
## Purpose
This PR is based on PR https://github.com/vllm-project/vllm/pull/24799 aiming to implement GPU version of ngram speculative decoding and make it compatible with Async Scheduler.

## Test Plan

- Async Scheduler + NGram + Qwen3-1.7B
Test config: 
```
# dataset is CMU-DoG, which is an input-grounded dataset.
python3.12 -u -m vllm.entrypoints.openai.api_server \
--host 0.0.0.0 \
--port 8000 \
--max-num-seqs 128 \
--max-model-len 2048 \
--model Qwen/Qwen3-1.7B \
--tensor-parallel-size 1 \
--trust-remote-code \
--dtype bfloat16  \
--enable-chunked-prefill \
--disable-log-requests \
--async-scheduling \
--speculative_config '{"method": "ngram_gpu", "num_speculative_tokens": 3, "prompt_lookup_max": 2,"prompt_lookup_min": 2}'
```
Test Device: NVIDIA H20

## Test Result
Performance

num_prompts | async_ngram(tps) | sync_ngram(tps) | speedup
-- | -- | -- | --
2 | 466 | 357 | 30.5%
8 | 1378 | 988 | 39.4%
16 | 2082 | 1726 | 20.6%

---
<details>

</details>

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit f141cc118bb8a0dbe4035cffc995e19ee38ca91f. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Introduces a GPU path for N-gram speculative decoding and wires it into async scheduling.
> 
> - **New GPU proposer**: `v1/spec_decode/ngram_proposer_gpu.py` adds a fully vectorized, torch.compile-optimized N-gram kernel (`NgramGPUKernel`) and `NgramProposerGPU`; no CUDA Graph capture; includes warmup.
> - **Runner integration**: `gpu_model_runner.py` supports `ngram_gpu` end-to-end (GPU token/state buffers, incremental GPU updates, handling empty/invalid drafts, async output path, fallback when drafter can’t run).
> - **Config updates**: Adds `NgramGPUTypes`, `use_ngram_gpu()`, method normalization, and async scheduling validation for `ngram_gpu`; sets model to `ngram_gpu` when selected.
> - **Compilation cache**: Disables vLLM torch.compile cache when `ngram_gpu` is enabled to avoid cache errors (`compilation/backends.py`).
> - **I/O batch change**: `gpu_input_batch.py` stores `num_tokens_no_spec` as a pinned CPU torch tensor for faster transfers.
> - **Tests**: Adds `test_with_ngram_gpu_spec_decoding` and renames eagle3 test; increases `max_model_len` to 4096 and logs `spec_method` for clarity.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f141cc118bb8a0dbe4035cffc995e19ee38ca91f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit 1d94b706078fe8c1628575d901b7356e637ccb87. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Introduces a GPU path for N-gram speculative decoding and integrates it with async scheduling.
> 
> - New `v1/spec_decode/ngram_proposer_gpu.py` with `NgramGPUKernel` (torch.compile-optimized) and `NgramProposerGPU`
> - Runner integration in `v1/worker/gpu_model_runner.py`: enable `ngram_gpu`, maintain GPU `token_ids`/`num_tokens_no_spec`, incremental tensor updates after batch reordering, handle empty/invalid drafts, and async output/fallback paths
> - Config updates: add `NgramGPUTypes`, `use_ngram_gpu()`, method normalization, and async-scheduling validation for `ngram_gpu` in `config/speculative.py` and `config/vllm.py`
> - Disable vLLM torch.compile cache when `ngram_gpu` is enabled to avoid cache issues (`compilation/backends.py`)
> - I/O batching tweak: store `num_tokens_no_spec` as pinned CPU tensor in `v1/worker/gpu_input_batch.py` for faster transfers
> - Tests: add `test_with_ngram_gpu_spec_decoding`, rename eagle3 test function, log `spec_method`, and raise `max_model_len` to `4096` in tests
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1d94b706078fe8c1628575d901b7356e637ccb87. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit 93375ffaf1bf088c12bb25212072087da2db0a48. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Introduces a GPU path for N-gram speculative decoding and integrates it with async scheduling.
> 
> - New `v1/spec_decode/ngram_proposer_gpu.py` with `NgramGPUKernel` (torch.compile-optimized, no CUDA Graph) and `NgramProposerGPU`; includes warmup
> - Runner updates in `v1/worker/gpu_model_runner.py`: enable `ngram_gpu`, maintain GPU `token_ids`/`num_tokens_no_spec`, incremental tensor updates post-reorder, handle empty/invalid drafts, async output path, and fallback when drafter cannot run
> - Config changes: add `NgramGPUTypes`, `use_ngram_gpu()`, method normalization for `ngram_gpu`, and async-scheduling validation in `config/speculative.py` and `config/vllm.py`
> - Compilation: disable vLLM torch.compile cache when `ngram_gpu` is enabled to avoid cache errors (`compilation/backends.py`)
> - I/O batching: store `num_tokens_no_spec` as pinned CPU torch tensor for faster transfers (`v1/worker/gpu_input_batch.py`)
> - Tests: add `test_with_ngram_gpu_spec_decoding`, rename eagle3 test, log `spec_method`, and raise test `max_model_len` to `4096`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 93375ffaf1bf088c12bb25212072087da2db0a48. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Implements a GPU path for N-gram speculative decoding and integrates it with async scheduling.
> 
> - Adds `vllm/v1/spec_decode/ngram_proposer_gpu.py` with `NgramGPUKernel` (torch.compile-optimized) and `NgramProposerGPU`; warmup path and no CUDA Graph capture
> - Integrates `ngram_gpu` in `gpu_model_runner.py`: GPU `token_ids`/`num_tokens_no_spec` buffers, incremental updates after batch reorder, empty/invalid draft handling, async output/fallback paths
> - Updates configs: new `NgramGPUTypes`, `use_ngram_gpu()`, method normalization, and async-scheduling validation (`config/speculative.py`, `config/vllm.py`)
> - Disables vLLM torch.compile cache when `ngram_gpu` is enabled to avoid cache errors (`compilation/backends.py`)
> - Optimizes I/O batching by storing `num_tokens_no_spec` as pinned CPU torch tensor (`v1/worker/gpu_input_batch.py`)
> - Tests: adds `test_with_ngram_gpu_spec_decoding`, renames eagle3 test, logs `spec_method`, and raises test `max_model_len` to `4096`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit eac7085d83667315331637b87d079874978879d1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Introduces a GPU path for N-gram speculative decoding and integrates it end-to-end with async scheduling.
> 
> - New `ngram_proposer_gpu.py` with `NgramGPUKernel` (torch.compile, vectorized, no CUDA Graph) and `NgramProposerGPU`, including warmup
> - Runner integration: enable `ngram_gpu`, maintain GPU `token_ids`/`num_tokens_no_spec`, incremental tensor updates after reorder/condense, handle empty drafts, and async/fallback paths; safe copies of scheduler output
> - Config: add `NgramGPUTypes`, `use_ngram_gpu()`, normalize `ngram_gpu` method, and allow async scheduling with it; set model when selected
> - Compilation: disable vLLM torch.compile cache when `ngram_gpu` is enabled to avoid cache errors
> - I/O batching: store `num_tokens_no_spec` as pinned CPU torch tensor for faster transfers
> - Tests: add `test_with_ngram_gpu_spec_decoding`, rename eagle3 test, log `spec_method`, and raise test `max_model_len` to `4096`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6b4637233a16ec605c6fe7ae69d27c3e323f4ee2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Introduces a GPU-based N-gram proposer and integrates it end-to-end with async scheduling.
> 
> - New `vllm/v1/spec_decode/ngram_proposer_gpu.py`: vectorized `NgramGPUKernel` (torch.compile) and `NgramProposerGPU`, with warmup and no CUDA Graph capture
> - Runner integration in `vllm/v1/worker/gpu_model_runner.py`: enable `ngram_gpu`, maintain GPU `token_ids`/`num_tokens_no_spec`, incremental tensor updates post-reorder, handle empty drafts, safe scheduler output copy, and async/fallback paths
> - Config updates in `vllm/config/speculative.py` and `vllm/config/vllm.py`: add `NgramGPUTypes`, normalize `ngram_gpu`, `use_ngram_gpu()`, and allow async scheduling with it
> - Compilation change in `vllm/compilation/backends.py`: disable vLLM torch.compile cache when `ngram_gpu` is enabled
> - I/O batching tweak in `vllm/v1/worker/gpu_input_batch.py`: store `num_tokens_no_spec` as pinned CPU torch tensor
> - Tests (`tests/v1/e2e/test_async_scheduling.py`): add `test_with_ngram_gpu_spec_decoding`, rename eagle3 test, log `spec_method`, and raise `max_model_len` to 4096
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2a1460550171d3b06b9c11a54d7817f3a1c2aaba. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


ZJY0516:
cc @njhill

vllm/compilation/backends.py
```diff
@@ -649,6 +649,13 @@ def __call__(
         # Honors opt-outs such as CompilationMode.NONE or VLLM_DISABLE_COMPILE_CACHE.
         disable_cache = not is_compile_cache_enabled(self.inductor_config)
 
+        # TODO(patchy): ngram gpu kernel will cause vllm torch compile cache errors.
```
benchislett, line 910: Why? Can this be fixed?

vllm/v1/worker/gpu_input_batch.py
```diff
@@ -114,6 +114,9 @@ def __init__(
             pin_memory=False,
         )
         self.token_ids_cpu = self.token_ids_cpu_tensor.numpy()
+        self.token_ids_gpu_tensor = torch.zeros(
```
benchislett: This is a massive buffer, and can go up to 1GB of VRAM in normal use-cases. Is there anything that can be done about this?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -995,6 +1002,150 @@ def _update_states_after_model_execute(
...
+                    # Get prompt_token_ids + output_token_ids
+                    prompt_token_ids = (
+                        req_state.prompt_token_ids
+                        if req_state.prompt_token_ids is not None
+                        else []
+                    )
+                    all_token_ids = prompt_token_ids + req_state.output_token_ids
+                    num_tokens = len(all_token_ids)
+                    # Copy to GPU tensor
+                    self.input_batch.token_ids_gpu_tensor[idx, :num_tokens].copy_(
```
benchislett: It looks like this copy is copying _from a device tensor_, instead of a standard HtoD copy. Why is that?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -995,6 +1002,150 @@ def _update_states_after_model_execute(
...
+                        torch.tensor(
+                            all_token_ids, dtype=torch.int32, device=self.device
+                        ),
+                        non_blocking=True,
+                    )
+                    self.input_batch.num_tokens_no_spec_gpu[idx] = num_tokens
```
benchislett: Can this logic be integrated into `_update_states`, where `num_tokens_no_spec` (cpu) is maintained? That seems like it would be cleaner than recomputing it twice and copying over here. Also, we would not want to maintain two pieces of the same logic.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -995,6 +1002,150 @@ def _update_states_after_model_execute(
         for i, num_tokens in enumerate(num_accepted_tokens):
             self.input_batch.num_accepted_tokens_cpu[i] = num_tokens
 
+    def _update_ngram_gpu_tensors(self, scheduler_output: "SchedulerOutput") -> None:
```
benchislett: Is there any unique logic in here that is distinct from how we maintain token_ids_cpu_tensor and num_tokens_no_spec_cpu?

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,515 @@
...
+    CommonAttentionMetadata,
+)
+from vllm.v1.utils import CpuGpuBuffer
+from vllm.v1.worker.gpu_input_batch import CachedRequestState, InputBatch
+
+
+@support_torch_compile(
+    dynamic_arg_dims={
+        "num_tokens_no_spec": 0,
+        "token_ids_gpu": [0, 1],
```
benchislett: What happens when both dims are marked as dynamic? Does it recompile for every combination of BS and SeqLen? If so, isn't that _way too much_ compilation?

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,515 @@
...
+            & valid_mask
+            & (num_tokens < self.max_model_len)
+            & (num_tokens >= self.min_n)
+        )
+
+        for _ in range(3):
+            with set_forward_context(None, self.vllm_config):
+                _ = self.kernel(num_tokens, token_ids, combined_mask)
+
+    def _generate_dummy_data(
```
benchislett, line 276: Why does the dummy run need actual synthetic data? Can't it just use random tokens? Isn't the point of the dummy run just to initialize/record the shapes and buffers?

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,515 @@
...
+        assert token_ids_gpu.device == self.device
+        assert num_tokens_no_spec.device == self.device
+        assert sampled_flags.device == self.device
+        assert valid_mask.device == self.device
+
+        with set_forward_context(None, self.vllm_config):
+            combined_mask = (
+                sampled_flags
+                & valid_mask
+                & (num_tokens_no_spec < self.max_model_len)
```
benchislett: How could this happen?

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,515 @@
...
+                & (num_tokens_no_spec >= self.min_n)
+            )
+
+            return self.kernel(
+                num_tokens_no_spec,
+                token_ids_gpu,
+                combined_mask,
+            )
+
+    def prepare_next_token_ids_cpu(
```
benchislett: Please don't duplicate functions like this. Refactor and reuse the code.

benchislett: This function does not seem to be used.

benchislett (changes requested):
My only major concern with this PR right now is maintaining the code. Between this PR and the draft model support, we are duplicating/reusing many components of the original EAGLE implementation in a naive manner. We should carefully structure the code here to implement these drafters more cleanly

vllm/compilation/backends.py
PatchouliTIS: I enabled torch compile in the  ngram gpu kernel, the computational graph corresponding to ngram operator would hit a precompiled computational graph cache in the main model, leading to mismatched computational graph results. Therefore, I directly disabled the compile cache here. I tested this locally, and disabling the cache had no impact on performance.

vllm/v1/worker/gpu_input_batch.py
PatchouliTIS: both the ngram gpu computation and inputs preparation of ngram can benefit from the buffer, I think it is worth to maintain the buffer. BTW I'm confused about the size, considering set max_model_len as 128k, it will take approximately 1600 max_num_seqs to reach the VRAM size of 1GB, is it a normal use-cases? Besides users can specify the value of max_num_seqs and max_model_len as well.

vllm/v1/worker/gpu_model_runner.py
PatchouliTIS: This separate processing for the ngram GPU input avoids direct copying each time. It performs incremental updates to the GPU buffer based on the previous prev_req_id_to_index and the current self.input_batch.req_id_to_index, thereby preventing extensive copying operations.

vllm/compilation/backends.py
benchislett: I assume disabling the compile cache would lead to longer startup time? I'm not an expert here but maybe it's possible to add an identifier to the compile cache to avoid extraneous cache hits?

vllm/v1/worker/gpu_input_batch.py
benchislett: - Some models have larger max context length than 128K (Qwen3 has 256K, Llama4 has 1M+)
- Deployments often have max_num_seqs between 512 and 2048. I would consider this to be a normal range.

vllm/compilation/backends.py
PatchouliTIS: yes, the startup time will increase a little. I attempted to add additional input parameters and other member variables to the nn.Modules forward method decorated with @support_torch_compile to achieve cache isolation, but none of them worked. I suspect this might be related to the internal implementation of @support_torch_compile within vLLM. However, as things stand, disabling torch compile caching only impacts the performance of the entire inference service during the initial startup phase.

vllm/v1/worker/gpu_model_runner.py
njhill: Yes it looks like this would cause an unnecessary cpu->gpu sync. Probably better to create a pinned cpu tensor from the list and do the nonblocking cpu->gpu copy from that. May need to be careful about that CPU tensor being garbage collected though.

vllm/v1/worker/gpu_input_batch.py
PatchouliTIS: If this section requires more sophisticated logic to store token IDs for each request, it would necessitate significant modifications, including changes to how ngram_gpu handles token ID input. For this PR, I propose moving this buffer allocation logic to `gpu_model_runner`. This means buffer allocation would only occur when both `async_scheduler` and `ngram_gpu` are enabled. Users could also manually set `max_num_seqs` and `max_model_len` to reduce the VRAM footprint of this buffer. Further optimizations could be addressed in a separate PR.

vllm/v1/worker/gpu_model_runner.py
PatchouliTIS: I see, I'll fix that in the actual async implementation.

vllm/v1/worker/gpu_model_runner.py
PatchouliTIS: move the update of `token_ids_gpu_tensor` used for ngram gpu into `_update_states`, reuse the result of `num_tokens_no_spec` maintained in `input_batch`

vllm/v1/spec_decode/ngram_proposer_gpu.py
PatchouliTIS: I have removed the second dim flag for `token_ids_gpu`, now there is only `batch_size` dim of all inputs is marked as dynamic. The second dim is `seq_len` and since I use a fix size of buffer `token_ids_gpu_tensor`, the input always has fixed size in second dim.

vllm/v1/spec_decode/ngram_proposer_gpu.py
PatchouliTIS: Already fix it in the newest commit, removing irrelevant dummy data preparation.

vllm/v1/spec_decode/ngram_proposer_gpu.py
PatchouliTIS: Redundant state checks have been removed.

vllm/v1/spec_decode/ngram_proposer_gpu.py
PatchouliTIS: Redundant function has been removed.

PatchouliTIS: Refactored the function in `ngram_proposer_gpu.py` which is used for validate the sampling results.

PatchouliTIS (commented):
Change the codes according to comments.

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,465 @@
...
+from vllm.config import (
+    CompilationConfig,
+    VllmConfig,
+)
+from vllm.forward_context import set_forward_context
+from vllm.v1.worker.gpu_input_batch import InputBatch
+
+
+@support_torch_compile(
+    dynamic_arg_dims={
```
benchislett: Isn't the default behaviour to assume that each tensor has the dynamic dim = 0? Why do you need to specify dynamic_arg_dims explicitly?

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,465 @@
...
+        "token_ids_gpu": 0,
+        "combined_mask": 0,
+    }
+)
+class NgramGPUKernel(nn.Module):
+    """
+    GPU-accelerated N-gram proposer using fully async tensor operations.
+
+    Interface: All inputs are GPU tensors (no lists, no numpy arrays)
+
```
benchislett: Please avoid overly verbose summarization. It is clear that `forward()` only takes arguments of type `torch.Tensor` simply from the type annotations.

```suggestion
```

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,465 @@
...
+    PERFORMANCE OPTIMIZATION WITH TORCH.COMPILE:
+
+    1. Tensor Allocation Strategy:
+       - DO: Allocate tensors inside forward() - torch.compile will optimize this
+       - DON'T: Pre-allocate buffers as class attributes - breaks compilation
+       - WHY: torch.compile fuses allocations into the compiled graph for efficiency
+
+    2. Dynamic Shapes:
+       - Batch size (dim 0) and sequence length (dim 1) are marked as dynamic
```
benchislett: This does not appear to be reflected in the supports_torch_compile annotation. Is `dim 1` marked as dynamic somewhere else?

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,465 @@
...
+       - torch.compile generates specialized kernels for different shapes
+       - The first call with a new shape will trigger recompilation (cached)
+
+    3. Graph Compilation:
+       - Uses fullgraph=True mode for maximum optimization
+       - All operations are tensor-based (no Python loops or conditionals)
```
benchislett: Strictly speaking, loops and branches are permissible as long as they are consistently taken AFAIK. For instance, `_find_first_and_extract_all_n_parallel` loops from `self.min_n` to `self.max_n`, which is fine.
```suggestion
       - All operations are tensor-based (no Python loops or conditionals that depend on input values)
```

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,465 @@
...
+       - The entire forward pass is compiled into a single CUDA graph
+
+    4. Memory Efficiency:
+       - torch.compile's memory planning optimizes temporary allocations
+       - Fusion of operations reduces memory bandwidth requirements
+       - No manual memory management needed - compiler handles it
```
benchislett: While I appreciate the detailed documentation, this section feels more like a tutorial for torch.compile. We use torch.compile extensively in vLLM and need not document its features everywhere it gets used in the code.

```suggestion
```

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,465 @@
...
+        data: torch.Tensor,
+        seq_lengths: torch.Tensor,
+        min_pattern_len: int,
+        max_pattern_len: int,
+        result_len: int,
+    ) -> torch.Tensor:
+        """
+        Process all pattern lengths in parallel, selecting the longest match.
+        Completely free of data-dependent control flow, suitable for
+        torch.compile optimization.
```
benchislett:
```suggestion
```

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,465 @@
...
+
+        This function allows partial results when there are fewer than result_len
+        tokens available after the match. The remaining positions are filled with 0.
```
benchislett: Is zero a valid return value for this function, outside of padding? If so, should a different padding value be chosen?

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,465 @@
...
+            current_windows = all_windows[..., offset:]
+            matches = (current_windows == patterns.unsqueeze(1)).all(dim=-1)
+
+            # Validity check: ensure enough space for result extraction
+            max_valid_pattern_start = seq_lengths - pattern_len - 1
+            pattern_start_positions = window_starts + offset
+            valid_mask = pattern_start_positions <= max_valid_pattern_start.unsqueeze(1)
+            final_matches = matches & valid_mask
+
+            # Handle prefix positions that fall before the available windows
```
benchislett: Can you rephrase this comment? It looks like the branch is taken on every loop iteration except the final one, but I'm not sure what "before the available windows" is in reference to.

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,465 @@
...
+
+        self.min_n = vllm_config.speculative_config.prompt_lookup_min
+        self.max_n = vllm_config.speculative_config.prompt_lookup_max
+        self.k = vllm_config.speculative_config.num_speculative_tokens
+        self.max_model_len = vllm_config.model_config.max_model_len
+        self.max_num_seqs = vllm_config.scheduler_config.max_num_seqs
+        self.vocab_size = vllm_config.model_config.get_vocab_size()
+        self.device = device
+
+    def _find_first_and_extract_all_n_parallel(
```
benchislett, line 46: Could you give a brief overview of the algorithm employed here? It's hard to follow some of the terminology regarding "patterns", "windows" etc.

Also, can you rename `data` and annotate a reminder of the inputs' shapes in the docstring?

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,465 @@
...
+        pattern_lengths = torch.arange(
+            min_pattern_len, max_pattern_len + 1, device=device
+        )
+        batch_indices = torch.arange(batch_size, device=device)
+
+        all_first_matches = torch.full(
+            (batch_size, num_patterns), -1, dtype=torch.long, device=device
+        )
+
+        for i, pattern_len in enumerate(range(min_pattern_len, max_pattern_len + 1)):
```
benchislett: It seems like the number of operations is proportional to the bounds on the pattern window. Do you have any profiles showing how many kernels get launched when running this op? I worry that even with torch.compile it might be a lot.

benchislett: Maybe a triton kernel would be more effective here? Very curious how much fusion is happening with torch.compile.

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,465 @@
...
+        num_tokens_no_spec: torch.Tensor,
+        token_ids_gpu: torch.Tensor,
+        combined_mask: torch.Tensor,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        """
+        Forward pass for N-gram proposal using GPU tensor operations.
+
+        This is the core computation method that will be compiled by torch.compile
+        via the @support_torch_compile decorator.
+
```
benchislett:
```suggestion
```

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,465 @@
...
+            matches = (current_windows == patterns.unsqueeze(1)).all(dim=-1)
+
+            # Validity check: ensure enough space for result extraction
+            max_valid_pattern_start = seq_lengths - pattern_len - 1
+            pattern_start_positions = window_starts + offset
+            valid_mask = pattern_start_positions <= max_valid_pattern_start.unsqueeze(1)
+            final_matches = matches & valid_mask
+
+            # Handle prefix positions that fall before the available windows
+            if offset > 0:
```
benchislett: If I understand correctly, this code handles looking for matches in the first `offset` tokens of the prompt, since `all_windows[..., offset:]` is derived from `data.unfold(1, max_pattern_len, 1)` which expands the input tokens into windows of size `max_pattern_len` and therefore slicing can miss the first few patterns of size `pattern_len`.

I think it is much simpler and more efficient to just have `current_windows = data.unfold(1, pattern_len, 1)` inside the loop, which will automatically cover that edge-case that you currently are handling manually.

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,465 @@
...
+            num_tokens_no_spec,
+            min_pattern_len=self.min_n,
+            max_pattern_len=self.max_n,
+            result_len=self.k,
+        )
+
+        # Apply combined mask to results. Expand mask explicitly to avoid
+        # relying on broadcasting behavior that can confuse torch.compile.
+        mask = combined_mask.unsqueeze(1).expand(-1, self.k)
+        draft_tokens = torch.where(mask, results, draft_tokens)
```
benchislett: I think this is possible, did it cause issues with torch.compile?
```suggestion
        draft_tokens = torch.where(mask, results, -1)
```

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,465 @@
...
+            with set_forward_context(None, self.vllm_config):
+                _, _ = self.kernel(num_tokens, token_ids, combined_mask)
+
+    def _generate_dummy_data(
+        self,
+        batch_size: int,
+        max_seq_len: int,
+        vocab_size: int = 152064,
+        pattern_len: int = 3,
+        repetition_rate: float = 0.5,
```
benchislett:
```suggestion
        vocab_size: int,
        pattern_len: int,
        repetition_rate: float,
```
Don't provide defaults for these variables, they should always be provided based on configs. If a default is needed, the config can default to a certain value.

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,465 @@
...
+
+            draft_tokens, is_empty_draft_tokens = self.kernel(
+                num_tokens_no_spec,
+                token_ids_gpu,
+                combined_mask,
+            )
+
+            return draft_tokens, is_empty_draft_tokens
+
+    def update_token_ids_ngram(
```
benchislett, line 385: Is this different from the EAGLE one? If so, how?

Ideally we could reuse the implementation of this function.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -3325,15 +3503,87 @@ def propose_draft_token_ids(
         spec_config = self.speculative_config
         assert spec_config is not None
         if spec_config.method == "ngram":
-            assert isinstance(sampled_token_ids, list)
-            assert isinstance(self.drafter, NgramProposer)
-            draft_token_ids = self.drafter.propose(
+            # TODO:(patchy) NGram GPU proposal
+            if isinstance(self.drafter, NgramProposer):
+                assert isinstance(sampled_token_ids, list), (
+                    "sampled_token_ids should be a python list whenngram is used."
```
benchislett:
```suggestion
                    "sampled_token_ids should be a python list when ngram is used."
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -3325,15 +3503,87 @@ def propose_draft_token_ids(
...
+            max_new_tokens = valid_sampled_token_ids_gpu.shape[1]  # num_spec_tokens + 1
+
+            current_lens = self.num_tokens_no_spec_gpu[:batch_size]
+            offsets = torch.arange(max_new_tokens, device=self.device)
+
+            write_positions = current_lens.unsqueeze(1) + offsets.unsqueeze(0)
+            valid_write_mask = offsets.unsqueeze(
+                0
+            ) < valid_sampled_tokens_count.unsqueeze(1)
+            combined_mask = valid_write_mask & (valid_sampled_token_ids_gpu != -1)
```
benchislett: Can you move more of this logic into `propose()`? I don't see why you need to calculate `combined_mask` in `gpu_model_runner`, for instance.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -3160,6 +3309,11 @@ def propose_draft_token_ids(sampled_token_ids):
             and spec_config.use_eagle()
             and not spec_config.disable_padded_drafter_batch
         )
+        use_padded_batch_for_ngram = (
```
benchislett: Does this mean that ngram-gpu supports both padded and non-padded batch modes? Have you tested it in both cases?

benchislett (changes requested):
Some cleanup is required. I don't see very many structural issues, but there is a lot of messy/overcomplicated and inefficient code that needs to be improved.

vllm/v1/spec_decode/ngram_proposer_gpu.py
PatchouliTIS: Yes, there are key differences:

  1. Backup token source: ngram extracts backup tokens directly from cached `token_ids_gpu_tensor` and `num_tokens_no_spec_gpu` on GPU using vectorized gather, while eagle iterates through requests dict on CPU and calls request.get_token_id().
  2. CPU-GPU sync: the `update_token_ids_ngram` in ngram is fully running on GPU, while eagle has sync due to CPU-side iteration + copy_to_gpu().
  3. the `update_token_ids_ngram` return one more params `valid_sampled_token_ids_gpu` which is used in `propose_draft_token_ids` to update the cached `token_ids_gpu_tensor` with new valid sampled token ids.

vllm/v1/worker/gpu_model_runner.py
PatchouliTIS: In current version, ngram-gpu only supports padded batch modes since the ngram-gpu is only enabled in async-scheduling mode. If `disable_padded_drafter_batch` is `True`, it means the whole inference procedure is in sync mode and the ngram is back in vanilla CPU-implementation, which means it won't comes to the path of ngram-gpu in sync mode. This is different from Eagle for ngram-gpu is only for async-scheduling.

vllm/v1/spec_decode/ngram_proposer_gpu.py
PatchouliTIS: Verified that broadcasting works fine with torch.compile - no graph breaks or recompilation issues. Simplified the code.

vllm/v1/spec_decode/ngram_proposer_gpu.py
PatchouliTIS: I conducted nsys profiling on `NgramGPUKernel.forward()` using various `[min_ngram_len, max_ngram_len]` settings. Results confirm efficient kernel fusion via torch.compile. The profiling details are attached below:

For `min_ngram_len=1, max_ngram_len=5`, the gpu kernels in `NgramGPUKernel.forward()` without torch.compile:
Clipboard_Screenshot_1766044430

Same settings with torch.compile:
Clipboard_Screenshot_1766044456

The entire `forward()` method is fuesd into five triton ops.

The number of fused triton ops is fixed no matter how the pattern window changes:
`min_ngram_len=1, max_ngram_len=3` kernel:
Clipboard_Screenshot_1766044595

`min_ngram_len=1, max_ngram_len=7` kernel:
Clipboard_Screenshot_1766044608

Based on the profiling results above, I think current ngram-gpu implementations with torch.compile is efficient enough.

vllm/v1/spec_decode/ngram_proposer_gpu.py
PatchouliTIS: The previous implementation with all_windows[..., offset:] required complex manual prefix handling for positions that fell before the available windows (the 30+ lines of prefix_positions, gather_indices, prefix_matches, etc.).

I've already updated the code to use your suggested approach: token_ids.unfold(1, ngram_len, 1) is now called inside the loop for each n-gram length. The complex prefix handling block has been completely removed.

vllm/v1/worker/gpu_model_runner.py
PatchouliTIS: Fix typo in comments

vllm/v1/spec_decode/ngram_proposer_gpu.py
PatchouliTIS: Removed irrelevant input params for `_dummy_run`.

vllm/v1/worker/gpu_model_runner.py
PatchouliTIS: Move the preprocess of token_ids_gpu_tensor and mask tensor into `NgramProposerGPU.propose()`

vllm/v1/spec_decode/ngram_proposer_gpu.py
PatchouliTIS: Added more detailed comments.

vllm/v1/spec_decode/ngram_proposer_gpu.py
PatchouliTIS: Removed irrelevant documents about torch.compile

vllm/v1/spec_decode/ngram_proposer_gpu.py
PatchouliTIS: Thanks for pointing that out. Removed explicit `dynamic_arg_dims` specification.

vllm/v1/spec_decode/ngram_proposer_gpu.py
PatchouliTIS: Removed.

vllm/v1/spec_decode/ngram_proposer_gpu.py
PatchouliTIS: Rewrite the documents in `ngram_proposer_gpu.py`.

vllm/v1/spec_decode/ngram_proposer_gpu.py
PatchouliTIS: Change the paddding value from 0 to -1.

vllm/v1/spec_decode/ngram_proposer_gpu.py
benchislett: Thanks! A few quick questions:
- Is this before or after the `.unfold()` optimization discussed below?
- What batch_size is this? If BS=1, could you also show similar results for larger batch_sizes (32, 128, 512)?

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,500 @@
...
+            compile_sizes=[],
+            inductor_compile_config={
+                "enable_auto_functionalized_v2": False,
+                "max_autotune": True,
+                "aggressive_fusion": True,
+                "triton.autotune_pointwise": True,
+                "coordinate_descent_tuning": True,
+                "use_mixed_mm": False,
+            },
+            use_cudagraph=False,
```
benchislett: Why are cuda graphs disabled? The comments above specify that fullgraph should be enabled. If this is able to be turned on, please confirm via nsys profile that it is taking effect. If there is a blocker, please outline the problem.

vllm/v1/spec_decode/ngram_proposer_gpu.py
PatchouliTIS: 1. The profiling result is after the `.unfold()` optimization discussed before.
2. Profiling is conducted in BS=64, I will test it for larger batch_sizes.

PatchouliTIS: BS=256
Clipboard_Screenshot_1766130286


BS=128
Clipboard_Screenshot_1766130339

vllm/v1/spec_decode/ngram_proposer_gpu.py
PatchouliTIS: Sorry about this, I forget to change the documents. In profiling I notice that even I set the `cudagraph_mode=CUDAGraphMode.FULL` in NgramProposerGPU, the nsys profiling result shows that the forward method still contains five seperate triton ops and no cuda graph launch API exists, which is same as setting `cudagraph_mode=CUDAGraphMode.NONE` .

vllm/v1/spec_decode/ngram_proposer_gpu.py
benchislett: Thanks, this looks good

vllm/v1/spec_decode/ngram_proposer_gpu.py
benchislett: Is this because of graph breaks? Can you try explicitly using fullgraph mode to identify what the reason is?
https://docs.pytorch.org/docs/stable/compile/programming_model.error_on_graph_break.html

PatchouliTIS: I have conducted a series of benchmarks and functional tests regarding the integration of torch.compile. Here are the findings:

### 1. No Graph Breaks - `fullgraph=True` is working
The compilation of NgramGPUKernel.forward() completes without errors when decorated with @torch._dynamo.error_on_graph_break(True). Validation via torch._dynamo.explain() further confirms a clean graph capture:
  - `Graph count: 1`
  - `Graph break count: 0`

  The forward pass is compiled into a **single computation graph** with no graph breaks.

  ### 2. The 5 triton ops are expected behavior

  There's a key distinction:
  - **`torch.compile(fullgraph=True)`**: Ensures no Python-level graph breaks
  - **CUDA Graph**: Captures and replays a sequence of GPU kernel launches

  `fullgraph=True` does NOT fuse multiple tensor operations into a single kernel. Each tensor op (`unfold`, `gather`, `argmax`, etc.) still compiles to a separate triton kernel. This is expected.

  ### 3. Enable full cudaGraph in `NgramGPUKernel.fowrad()` 
  To enable full-graph in `NgramGPUKernel.fowrad()`, it requires:
  - Fixed memory address for foward inputs, which means maintaining a buffer in NgramGPUKernel;
  - Pre-capture various batch size during `_dummy_run()` just as main model did in `gpu_model_runner.py`. 
  This can be done in `CUDAGraphWrapper` decorators, but as I mentioned before, the performance of current implementation is efficient enough and the CUDA Graph feature maybe overkill for such simple scenario.

benchislett: Okay, this is fine. If cuda graphs are not in scope, there's no need to add any extra complexity to use them.

vllm/v1/worker/gpu_model_runner.py
benchislett: I don't see why ngram-gpu needs to be only used with async scheduling. Is this a concession of this initial PR, or are you designing it such that it will only ever be used with async scheduling?

PatchouliTIS: This is expected by design. If vLLM is running in sync-mode, the vanilla cpu version of ngram is efficient enough and  GPU Ngram implementations will introduce additional GPU-CPU data synchronization overhead.

benchislett: I don't expect there would be much "additional" synchronization in the sync-mode implementation since we're already operating synchronously. I believe there was some discussion in the community about poor performance of NGram-CPU at large batch sizes and I wonder if the GPU implementation might alleviate issues in that case. Do you have any benchmarks to suggest that ngram-cpu is better than ngram-gpu in the synchronous case?

Note also that asynchronous scheduling and padded drafter batch are distinct modes, and all combinations of enabling/disabling these features is possible.

benchislett: I believe that there might be significant advantages to using ngram on the GPU at large batch sizes, even if run eagerly. As such, I think it makes sense to support synchronous scheduling. At least, please investigate if there are any non-trivial blockers to enable support in these cases. If it is a simple change I think it would make sense to support these modes and let users opt-in by selecting ngram-gpu as the draft method.

If it would take significant effort to restructure in order to support these other modes, it is okay for it to be beyond the scope of this initial PR.

PatchouliTIS: I have updated the PR to support both async and sync scheduling, as well as padded and non-padded batch modes.
Here is a summary of the implementation details:
1. Decoupling from Scheduling Mode
ngram-gpu now functions correctly regardless of whether the engine is in async or sync mode.
2. Handling Irregular Batches (disable_padded_drafter_batch=True)
In sync mode with non-padded batches, sampled_token_ids arrives as a `list[list[int]]` with variable lengths (e.g., empty lists for prefill requests).
Padding Logic: I’ve implemented a lightweight padding mechanism to convert these irregular lists into a regular tensor, using -1 as a sentinel value.
Draft Token Filtering: The `take_draft_token_ids()` now correctly filters out these -1 values before passing them to the scheduler, ensuring only valid n-gram matches are proposed.
3. Architectural Simplification vs. Eagle
One key advantage of ngram-gpu over models like Eagle is its stateless nature regarding model activations:
No Dependency on Metadata: Unlike Eagle/Medusa, which depend on `hidden_states` and `attention_metadata` (which differ significantly between padded/non-padded paths), ngram-gpu only operates on `token_ids_gpu_tensor`.

I also conduct some benchmark about the performance of both cpu and gpu ngram kernel, and here is the result:

BS=128, Ngram CPU kernel e2e time cost:
Clipboard_Screenshot_1766493658


BS=128, Ngram GPU kernel e2e time cost:
Clipboard_Screenshot_1766493645

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,510 @@
...
+
+    2. Dynamic Shapes:
+       - Batch size (dim 0) is automatically marked as dynamic in support_torch_compile
+       - torch.compile generates specialized kernels for different shapes
+
+    3. Graph Compilation:
+       - Note: fullgraph=True does NOT mean the forward pass runs as a single
+         CUDA Graph. nsys profiling shows ~5 separate Triton kernel launches.
+         torch.compile with fullgraph=True ensures complete Dynamo tracing
+         without fallbacks, but Inductor still generates multiple kernels.
```
benchislett:
```suggestion
```
Can this be removed? I don't think "fullgraph=True" is specified anymore.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1053,6 +1087,51 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> None:
...
 
+            spec_token_ids = scheduler_output.scheduled_spec_decode_tokens.get(
+                req_id, []
+            )
+            num_spec_tokens = len(spec_token_ids)
+
+            is_empty_draft_tokens = (
+                self._is_empty_draft_tokens is not None
+                and req_index is not None
+                and self._is_empty_draft_tokens[req_index].item()
```
benchislett: This seems particularly inefficient. At least, copy the whole tensor and then access each value instead of calling .item() on each slot. 

When resolving "High Severity" bot comments, please at least leave a comment to indicate why you think it is not an issue.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1053,6 +1087,51 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> None:
...
+            )
+
+            # For async scheduling, token_ids_cpu assigned from
```
benchislett: I don't understand what any of this logic has to do with ngram-gpu. Is this just general bookkeeping for async scheduling? Don't we already have code to handle these cases?

vllm/v1/worker/gpu_model_runner.py
benchislett: In order to ensure full compatibility with async scheduling, there should be no synchronizations between iterations. If this is a target of the PR, please ensure via profile (nsys or torch) that this is not taking place. If this is being conceded in exchange for implementation simplicity, please make that clear.

benchislett (changes requested):
Left some comments. 

Besides the other comments, please make an effort to further reduce the impact on `gpu_model_runner.py`. ~300 lines is acceptable, but I would appreciate if a bit more code could be slimmed down and/or migrated into helper functions in `ngram_proposer_gpu.py` to lessen the impact on readability and maintenance of the core files and functions such as `_update_states`.

vllm/v1/spec_decode/ngram_proposer_gpu.py
PatchouliTIS: Removed fullgraph docs.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1050,6 +1084,51 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> None:
...
+            spec_token_ids = scheduler_output.scheduled_spec_decode_tokens.get(
+                req_id, []
+            )
+            num_spec_tokens = len(spec_token_ids)
+
+            is_empty_draft_tokens = (
+                self._is_empty_draft_tokens is not None
+                and req_index is not None
+                and self._is_empty_draft_tokens[req_index].item()
+            )
```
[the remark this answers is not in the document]
PatchouliTIS: Refactor those sync data access operations into async mode.

vllm/v1/worker/gpu_model_runner.py
PatchouliTIS: Refactor those sync data access operations into async mode.

vllm/v1/worker/gpu_model_runner.py
PatchouliTIS: Sorry, this code was left over from before, and I forgot to clean it up when handling new merge conflicts. Already removed this part.

vllm/v1/worker/gpu_model_runner.py
PatchouliTIS: Additionally, I moved several functions closely related to the ngram GPU implementation from `gpu_model_runner.py` to `ngram_proposer_gpu.py` to keep `gpu_model_runner` as concise as possible.

tools/vllm-rocm/pin_rocm_dependencies.py
```diff
@@ -11,10 +11,11 @@
 instead of allowing pip to download different versions from PyPI.
 """
 
-import re
 import sys
 from pathlib import Path
 
+import regex as re
+
```
benchislett, line 18: What is this?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -212,6 +218,7 @@ def __init__(
 
         # Initiate the copy on a separate stream, but do not synchronize it.
         default_stream = torch.cuda.current_stream()
+
```
benchislett: Remove empty line

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -647,6 +666,37 @@ def __init__(
...
+        # 3. Async Overlap: Using a dedicated CUDA stream for D2H copy allows the
+        #    transfer to overlap with GPU kernel execution, avoiding pipeline bubbles.
+        if (
+            self.speculative_config is not None
+            and self.speculative_config.use_ngram_gpu()
+        ):
+            self._is_empty_draft_tokens_cpu = torch.empty(
+                self.max_num_reqs, dtype=torch.bool, pin_memory=self.pin_memory
+            )
+            # CUDA event for synchronizing _is_empty_draft_tokens async copy
```
benchislett: Remove redundant comments

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -647,6 +666,37 @@ def __init__(
 
         # Cached outputs.
         self._draft_token_ids: list[list[int]] | torch.Tensor | None = None
+        # For ngram_gpu: indicates which requests have empty/invalid draft tokens
+        self._is_empty_draft_tokens: torch.Tensor | None = None
+        self._is_empty_draft_tokens_cpu: torch.Tensor | None = None
+        self._is_empty_draft_tokens_event: torch.cuda.Event | None = None
+        self._is_empty_draft_tokens_copy_stream: torch.cuda.Stream | None = None
+        # Async D2H copy mechanism for is_empty_draft_tokens (ngram_gpu only)
```
benchislett: This block comment is needlessly verbose. Simplify it to a few lines, if any

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -888,6 +938,15 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> None:
         for req_id in unscheduled_req_ids:
             self.input_batch.remove_request(req_id)
 
+        # Check if ngram_gpu mode is enabled for incremental GPU tensor updates
```
benchislett: Remove redundant comments

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1308,6 +1396,8 @@ def _prepare_input_ids(
         # because input_ids dtype is torch.int32,
         # so convert draft_token_ids to torch.int32 here.
         draft_token_ids = self._draft_token_ids.to(dtype=torch.int32)
+        self._draft_token_ids = None
```
benchislett: Is this safe for methods other than ngram-gpu? I'm not sure how this would affect async specdec (e.g. eagle, mtp)

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -3444,6 +3553,8 @@ def propose_draft_token_ids(sampled_token_ids):
                     spec_decode_metadata,
                     spec_decode_common_attn_metadata,
                 )
+                # DEBUG(patchy)
```
benchislett: Debug comment left in

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -3656,12 +3797,74 @@ def propose_draft_token_ids(
...
+                    sampled_token_ids,
+                    self.input_batch.num_tokens_no_spec,
+                    self.input_batch.token_ids_cpu,
+                )
+        elif spec_config.method == "ngram_gpu":
+            # GPU-accelerated ngram proposer
+            assert isinstance(self.drafter, NgramProposerGPU)
+            sampled_token_ids_gpu_tensor = sampled_token_ids
+            if isinstance(sampled_token_ids_gpu_tensor, list):
+                # When disable_padded_drafter_batch=True, sampled_token_ids is
```
benchislett: Could this be moved into update_token_ids_ngram?

vllm/v1/spec_decode/ngram_proposer_gpu.py
```diff
@@ -0,0 +1,654 @@
...
+
+    Args:
+        scheduler_output: The scheduler output to update.
+        req_id_to_index: A mapping from request IDs to their indices in the batch.
+    """
+    req_data = scheduler_output.scheduled_cached_reqs
+
+    # Sync the is_empty_draft_tokens copy (should be complete by now)
+    is_empty_draft_tokens_event.synchronize()
+    is_empty_draft_tokens_cpu = is_empty_draft_tokens_cpu
```
benchislett: Is this a no-op?

tools/vllm-rocm/pin_rocm_dependencies.py
PatchouliTIS: It is related to local pre-commit checks, asking me to change `import re` into `import regex as re`. I have bypassed it.

```
Enforce import regex as re..........................................................................Failed
- hook id: enforce-import-regex-instead-of-re
- exit code: 1

❌ tools/vllm-rocm/pin_rocm_dependencies.py:
  Line 14: import re

💡 Found 1 violation(s).
❌ Please replace 'import re' with 'import regex as re'
   Also replace 'from re import ...' with 'from regex import ...'
✅ Allowed imports:
   - import regex as re
   - import regex
```

vllm/v1/worker/gpu_model_runner.py
PatchouliTIS: Redundant comments removed

vllm/v1/worker/gpu_model_runner.py
PatchouliTIS: Simplified.

vllm/v1/worker/gpu_model_runner.py
PatchouliTIS: Sorry, this is a former code change and should be removed.

vllm/v1/worker/gpu_model_runner.py
PatchouliTIS: Removed.

vllm/v1/spec_decode/ngram_proposer_gpu.py
PatchouliTIS: Removed.

vllm/v1/worker/gpu_model_runner.py
PatchouliTIS: Done.

vllm/v1/worker/gpu_model_runner.py
PatchouliTIS: Removed

vllm/v1/worker/gpu_model_runner.py
PatchouliTIS: Removed.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -944,12 +987,28 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> None:
...
+            # Track new requests for ngram_gpu full tensor copy
+            if is_ngram_gpu:
+                ngram_gpu_new_reqs.append(req_state)
 
         # Update the states of the running/resumed requests.
         is_last_rank = get_pp_group().is_last_rank
         req_data = scheduler_output.scheduled_cached_reqs
         scheduled_spec_tokens = scheduler_output.scheduled_spec_decode_tokens
 
+        # Wait util _is_empty_draft_tokens async copy is done
```
benchislett:
```suggestion
        # Wait until _is_empty_draft_tokens async copy is done
```

PatchouliTIS: Typo errors fixed in `gpu_model_runner.py`

PatchouliTIS:
I've resolved the comments, removed redundant code comments, and fixed the typos. Could you please take another look? Thanks! @benchislett

benchislett (approved):
Nice work!

PR #29189, opened 2025-11-21
Fix: Resolve circular import in model_loader/utils.py

nandan2003:
## Purpose

Fixes a circular import error where `vllm.model_executor.model_loader.utils` could not be imported because it attempted to import from `models.adapters` at the top level, while `models.adapters` (indirectly) depends on `utils`.

Fixes \#29110

## Test Plan

I verified this locally by attempting to import the module in an isolated script to ensure the circular dependency chain is broken.

**Verification Script:**

```python
import torch

# Mock compile to bypass python 3.14+ restriction if needed for testing import logic only
def no_op_compile(*args, **kwargs):
    def decorator(func):
        return func
    return decorator
torch.compile = no_op_compile

try:
    from vllm.model_executor.model_loader.utils import process_weights_after_loading
    print("Success! Circular dependency is fixed.")
except ImportError as e:
    print(f"FAILURE: {e}")
```

## Test Result

  * **Before Fix:** `ImportError: cannot import name 'process_weights_after_loading' from 'vllm.model_executor.model_loader.utils'`
  * **After Fix:** `Success! Circular dependency is fixed.`

### Additional Context

The import of `as_embedding_model`, `as_reward_model`, `as_seq_cls_model`, and `try_create_mm_pooling_model_cls` was moved from the global scope of `utils.py` into `get_model_architecture()` where they are actually used. This defers the import execution and resolves the cycle.


DarkLight1337 (commented):
I don't think this is a problem when importing these modules from vLLM top level (otherwise vLLM would not be able to be launched at all)

nandan2003:
@DarkLight1337 You are correct that the standard vLLM entrypoints (like `vllm serve`) work fine, likely because the import order is different or the modules are cached in `sys.modules` earlier in the startup sequence.

However, this circular import specifically breaks downstream libraries (like VERL, as reported in #29110) that attempt to import utility functions directly.

**The Reproduction:**
If a user runs:
```python
from vllm.model_executor.model_loader.utils import process_weights_after_loading

DarkLight1337 (approved):
Got it, thanks for pointing that out!

Issue #29134, opened 2025-11-21
[Performance]: Fully Async Spec-Decoding | Make `seq_lens_cpu` in CommonAttentionMetadata optional


LucasWilkinson:
### Proposal to improve performance

Currently fully overlapping input-prep with model forward pass is blocked in the spec-decode case by the following Host<>GPU syncs:

1) `_get_valid_sampled_token_count` (ultimately needed to compute `seq_lens_cpu`): https://github.com/vllm-project/vllm/blob/fe25772aa97beb8bcb07ea49e06a2892b521a7ed/vllm/v1/worker/gpu_model_runner.py#L3109
2) in the `num_speculated_tokens > 1` case by needing to update `seq_lens_cpu` for attention metadata building (specifically for all the speculated tokens after the first one hence only impacting `num_speculated_tokens > 1` context: https://github.com/vllm-project/vllm/pull/26498)

Ultimately in-order to realize fully async spec decoding we need to build attention metadata without knowing `seq_lens_cpu` (using device `seq_lens` on device is fine since any metadata building GPU kernels will get queued to after this is updated by the CUDA driver).

This is currently not entirely possible for all backends (namely FlashInfer do to D2H or H2D, depending on if a host or device tensor is provided, inside the [plan function](https://github.com/flashinfer-ai/flashinfer/blob/cce4952fdd41b353325e11d99e1fc0b0737961ff/flashinfer/decode.py#L940-L955)). 

However, there are many (important/default) backends like TRTLLM (inside the FlashInfer backend), FlashAttn, FlashAttn-MLA (for pure decode), FlashMLA (for pure decode) etc. that could achieve this full overlap using only the device `seq_lens` tensor.

The proposal would be to update common attention metadata from: 

```
@dataclass
class CommonAttentionMetadata:
    ...
    seq_lens: torch.Tensor
    seq_lens_cpu: torch.Tensor
```

To something like:

```
@dataclass
class CommonAttentionMetadata:
    ...
    seq_lens: torch.Tensor
    _seq_lens_cpu: torch.Tensor | None

    # WARNING: using this property will mean spec-decode with async-scheduling will not achieve
    # full overlap due to Host<>GPU sync
    @property
    def seq_lens_cpu(self):
           # Potentially log warning here to encourage developers to avoid this property
           return self. _seq_lens_cpu or seq_lens.to("cpu")
```

[...]

PR #29194, opened 2025-11-21
[Model Runner V2] Change bookkeeping logic in preparation for spec decoding

WoosukKwon:
This PR updates the input-preparation and bookkeeping logic for the V2 model runner. The primary goal is to shift as much work as possible onto the GPU, laying the groundwork for async scheduling + speculative decoding.

### Key changes

* **`num_computed_tokens` is now GPU-only.**
  The CPU no longer tracks this value. Instead, it maintains **`num_computed_prefill_tokens`**, which is used to track the progress of chunked prefills.

* **Prefill and decode input preparation are now decoupled.**

  * **Prefill:** inputs and their statuses (e.g., `num_computed_prefill_tokens`) are read from NumPy arrays.
  * **Decode:** uses `last_sampled_tokens` and `num_computed_tokens` directly from the GPU.

* **Preparation for speculative decoding.**
  We now maintain `num_sampled_tokens` as GPU tensors for upcoming spec-decode support (where the number of accepted tokens are dynamic).


vllm/v1/worker/gpu/model_runner.py
```diff
@@ -420,61 +426,75 @@ def prepare_inputs(
...
         logits_indices = query_start_loc_gpu[1:] - 1
 
+        # Get num_computed_tokens.
+        # HACK(woosuk): Here, we use num_computed_tokens on GPU instead of
+        # num_computed_tokens_cpu. This works for most cases.
+        num_computed_tokens = self.req_states.num_computed_tokens[idx_mapping]
+        # HACK(woosuk): Only GPU has the exact seq_lens because at this point
+        # CPU does not know how many draft tokens are accepted/rejected in the
+        # previous step. Therefore, we use max_model_len to be safe.
+        seq_lens_np = np.full(num_reqs, self.max_model_len, dtype=np.int32)
```
WoosukKwon, line 485: @LucasWilkinson This is my current hack, which is totally undesirable. I plan to use a tighter upper bound for `seq_lens_np`, but I'd like to keep it if the refactoring may happen in the near future.

LucasWilkinson: I think we can refactor (I assume you are referring to: https://github.com/vllm-project/vllm/issues/29134) it soon-ish (next couple weeks) since its should be backwards compatible. 

You may already know this but for posterity/public-discussion/context: @benchislett and I did discuss making `seq_lens_cpu` an upper bound by a few tokens (i.e. would over estimate by num_rejected_tokens but the GPU tensor would be correct). Unfortunately, that doesn't really work for FlashInfer since it copies the CPU tensor to the GPU [inside the wrapper](https://github.com/flashinfer-ai/flashinfer/blob/cce4952fdd41b353325e11d99e1fc0b0737961ff/flashinfer/decode.py#L940-L955) and we end up attending to garbage tokens which leads to a small drop in acceptance rate (see: https://github.com/vllm-project/vllm/pull/26498) hence why we believe https://github.com/vllm-project/vllm/issues/29134 is a better/cleaner solution. This really only impacts the "classic" FlashInfer code paths, not the new TRTLLM code paths which I think are used in most cases anyways (although im not totally up to date on the FlashInfer backend). We still need to audit how the Mamba attention backends use `seq_lens_cpu`; @MatthewBonanni will help with this.

I do think it could be worth still providing exact `seq_lens_cpu` to the metadata builders when it is known (i.e. async scheduling no-spec-decode) because this means the FlashInfer "classic" wrappers can still achieve full overlap. If  FlashInfer "classic" is the only thing that benefits (after we audit Mamba and potentially other HW backends) I could see arguments for never providing it given that FlashInfer is moving more towards the TRTLLM kernels anyways.

WoosukKwon: Thanks for the explanation. As it turns out, this hack doesn’t work with FlashInfer correctly. For now, I’ve added a gate to raise an error and only allow the FA backend.

PR #29196, opened 2025-11-21
online fp8 quant with streaming weight post-processing

vkuzo:
Summary:

updates the fp8.py vllm online quant entry point to post-process weights as soon as they are ready, to prevent from holding the entire high precision model in memory.  Specifically:

1. during weight loading, keep track of how many elements we have loaded
2. when we have loaded all the elements, call post-processing

I'm putting up this PR to demonstrate the approach.  Since for now the # of use cases is 2 (linear and moe in fp8.py), keeping the logic simple.  If this is extended to other online quant providers such as compressed-tensors/torchao/etc in the future, might be good to make the code more reusable in future PRs.

Test Plan:

tested locally with `facebook/opt-125m` for linear and `Qwen/Qwen1.5-MoE-A2.7B` for MoE and `fp8` online quantization.  Verified (via prints) that `layer.process_weights_after_loading` is called immediately after the last weight chunk is loaded in both cases.

Reviewers:

Subscribers:

Tasks:

Tags:

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/layers/quantization/fp8.py
+++ b/vllm/model_executor/layers/quantization/fp8.py
@@ -461,6 +461,30 @@ class Fp8LinearMethod(LinearMethodBase):
                 output_size_per_partition, input_size_per_partition, weight_loader
             )
         else:
+
+            def patched_weight_loader(param, loaded_weight, *args, **kwargs):
+                # load the current weight chunk
+                res = weight_loader(param, loaded_weight, *args, **kwargs)  # type: ignore[misc]
+
+                # track how many elements we have updated
+                if not hasattr(layer, "_loaded_numel"):
+                    layer._loaded_numel = 0
+                layer._loaded_numel += loaded_weight.numel()
+
+                # if we have loaded all of the elements, call
+                # process_weights_after_loading
+                target_loaded_numel = layer.weight.numel()
+                if layer._loaded_numel == target_loaded_numel:
+                    self.process_weights_after_loading(layer)
+
+                    # Delete the bookkeeping
+                    del layer._loaded_numel
+                    # Prevent the usual `process_weights_after_loading` call from doing
+                    # anything
+                    layer._already_called_process_weights_after_loading = True
+
+                return res
+
             # For non-serialized checkpoints, use original dtype
             weight = ModelWeightParameter(
                 data=torch.empty(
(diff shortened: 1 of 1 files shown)


vllm/model_executor/layers/quantization/fp8.py
```diff
@@ -461,6 +461,33 @@ def create_weights(
                 output_size_per_partition, input_size_per_partition, weight_loader
             )
         else:
+
+            def patched_weight_loader(param, loaded_weight, *args, **kwargs):
+                # load the current weight chunk
+                res = weight_loader(param, loaded_weight, *args, **kwargs)  # type: ignore[misc]
+
+                # add a counter to track how many elements we have updated
+                if not hasattr(layer, "_target_loaded_numel"):
```
kylesayrs: Are all of these `hasattr` checks necessary? It's a little hard to read right now, I think the attribute is guaranteed to exist if  `not self.quant_config.is_checkpoint_fp8_serialized`

vllm/model_executor/layers/quantization/fp8.py
```diff
@@ -461,6 +461,33 @@ def create_weights(
...
+                    # for linear, the only weight we need to load is `layer.weight`
+                    layer._target_loaded_numel = layer.weight.numel()
```
kylesayrs: Nit: avoid extra flags by deleting `_target_loaded_numel` and just calling `layer.weight.numel()`

kylesayrs (approved):
Thanks!

vllm/model_executor/layers/quantization/fp8.py
vkuzo: makes sense, sg, we can delete the `_target_loaded_numel` var and just compute it every time

vllm/model_executor/layers/quantization/fp8.py
vkuzo: sg

vllm/model_executor/layers/quantization/fp8.py
kylesayrs: The same goes for `_loaded_numel`. If this attribute didn't exist, then something has gone terribly wrong and it might be appropriate to error.

robertgshaw2-redhat (approved):

PR #29197, opened 2025-11-21
[Frontend] Implement robust video frame recovery for corrupted videos

vSeamar:
## Purpose

This PR implements **optional** video frame recovery functionality for vLLM's multimodal video processing using a dynamic window forward-scan approach.

### Key Changes:

- **Video Frame Recovery System (Optional)**: Added intelligent recovery logic that attempts to recover missing/broken video frames when sequential reading fails. **Disabled by default** - only activates when `frame_recovery=True`.

- **Dynamic Window Forward-Scan Algorithm**: When a target frame fails to load, the system continues scanning forward and uses the next successfully grabbed frame (before the next target frame) as a substitute.

- **Comprehensive Testing**: Added 3 new unit/regression tests covering recovery functionality with simulated failures, real corrupted files, and both video backends.

### Technical Details:

- **Single-Pass Algorithm**: The recovery system scans through frames linearly, tracking failed frames in a queue. When a valid frame is found within the "recovery window" (between a failed frame and the next target), it's used to recover the oldest failure.

- **Recovery Window Concept**: Each failed frame has a window extending to the next target frame. This ensures recovered frames maintain temporal proximity to their intended position.

- **Temporal Safety**: Recovered frames are always from *after* the failed frame (forward scan), ensuring the model never sees temporal inversion.

- **Backward Compatible**: Default behavior unchanged - `frame_recovery=False` (default) disables recovery entirely, maintaining existing performance and behavior.

- **Opt-in Enhancement**: Users can enable recovery via CLI:

  ```bash
  --media-io-kwargs '{"video": {"frame_recovery": true}}'
  ```

- **Backend Support**: Recovery works for both `opencv` and `opencv_dynamic` video backends when enabled.

- **Logging**: Recovery attempts and results are logged at INFO level for monitoring.

### Algorithm Overview:

```
Target frames:    [0]          [10]         [20]         [30]
                   ↓            ↓            ↓            ↓
Video timeline:  [0][1][2]...[10][11]...[19][20]...[30]...
                              ╳
                           (failed)
                              
Recovery window for frame 10: [10 → 20)
Frame 11 grabbed successfully → used to recover frame 10
```

### Unit Tests Added:

- `test_video_recovery_simulated_failures`: Mocks `cv2.VideoCapture.grab()` to simulate failures on specific frames, verifies recovery successfully loads frames and preserves temporal order
- `test_video_recovery_with_corrupted_file`: Tests recovery using real corrupted video asset
- `test_video_recovery_dynamic_backend`: Tests recovery in dynamic frame sampling backend

### Test Coverage:

- Simulated frame grab failures
- Real corrupted video files
- Both video backend types (`opencv`, `opencv_dynamic`)
- Temporal ordering preservation
- Recovery metadata correctness

## Test Result

```
tests/multimodal/test_video.py - 14 passed
```

---
<details>

</details>


vllm/benchmarks/datasets.py
```diff
@@ -862,7 +862,7 @@ def generate_synthetic_video(
         fourcc = cv2.VideoWriter_fourcc(*"mp4v")
         fps = 30  # frames per second
 
-        with NamedTemporaryFile(suffix=".mp4", delete_on_close=False) as temp_file:
+        with NamedTemporaryFile(suffix=".mp4", delete=False) as temp_file:
```
vSeamar, line 870: benchmark tests would fail on python 3.10.16 (which should be supported):
`pytest tests/benchmarks/test_random_multimodal_dataset_video.py -v`

This change now passes all the tests on 3.10.16 and 3.13.6.

vllm/multimodal/video.py
```diff
@@ -63,12 +63,77 @@ def load_bytes(
...
+            # Seek to target frame
+            cap.set(cv2.CAP_PROP_POS_FRAMES, target_idx)
+            ret, frame = cap.read()
+
+            if not ret and recovery_offset > 0:
+                logger.debug(
+                    "Frame %d failed, attempting recovery (offset=%d)",
+                    target_idx,
+                    recovery_offset,
+                )
```
Isotr0py: Hmmm, I remember that "seek and read" can significantly slow down the video parsing performance, because seek frame is expensive...

vSeamar: You are correct. I commented below a benchmark I did with 100 videos <=30s, grab is faster on all of them; 76% faster on average. 

To help with performance I did these two things:
1. First, it will use grab as it did before, if this succeeds for all the frames we need, seek will not be used at all. Thus no performance lost. From a test I did with 1200 videos, this was the case for 1192 of them, so most of the time the performance should not decrease.
2. If grab fails, seek will begin at the frame that failed, saving time on the initial frames that succeeded. 

Also, this recovery mode is an optional argument, off by default. Users have control on how many frames to "seek" around the corrupted frames, letting them decide the performance trade-off for their use case 😃

vllm/multimodal/video.py
```diff
@@ -63,12 +63,77 @@ def load_bytes(
...
+                            recovered_frames[target_idx] = recovery_idx
+                            break
+                    if ret:
+                        break
+
+            if ret:
+                frames_list.append(cv2.cvtColor(frame, cv2.COLOR_BGR2RGB))
+                loaded_indices.append(target_idx)
+            else:
+                logger.debug("Frame %d failed to load after seeking", target_idx)
```
NickLucche: is this a logger.error too?

vllm/multimodal/video.py
```diff
@@ -63,12 +63,77 @@ def load_bytes(
...
+            if not ret and recovery_offset > 0:
+                logger.debug(
+                    "Frame %d failed, attempting recovery (offset=%d)",
+                    target_idx,
+                    recovery_offset,
+                )
+                # Try offsets: -1, +1, -2, +2, ...
+                for i in range(1, recovery_offset + 1):
+                    for direction in [-i, i]:
+                        recovery_idx = target_idx + direction
```
NickLucche: wouldn't it be simpler to keep the last frame successfully read and re-use it in case it's within `recovery_offset`? 
I feel like seeking forward as in here creates a pattern that might actually hurt performance

vllm/multimodal/video.py
vSeamar: I kept it as debug as this can get verbose. This is not expected to work all of the time, and if recovery does not recover all the necessary frames, line 297 below will warn the user with: "Recovery finished but video is still missing %d frames."

vllm/multimodal/video.py
vSeamar: Could you elaborate on both remarks?

Keeping last successful frame read could be helpful, but under certain conditions. Think about a 30 fps video sampled at 1 fps with recovery_offset=8. The recovery_offset is not large enough to ever overlap with a previous successful frame. This will probably be most cases, but if you think it is helpful I could add it.

I also am not quite sure what you mean by seeking forward, do you mean the approach of seeking X frames ahead of corrupted frame one by one is not a good idea?

vSeamar: Added a comment below explaining a change I made in response to your suggestion.

vllm/multimodal/video.py
```diff
@@ -63,12 +63,160 @@ def load_bytes(
...
+        frame_indices: list[int],
+        recovery_offset: int,
+        existing_frames: dict[int, npt.NDArray] | None = None,
+    ) -> tuple[npt.NDArray, list[int], dict[int, int]]:
+        """
+        Optimized Recovery Logic:
+        1. Cache Check (Zero I/O): Reuse frames from sequential read if close enough.
+        2. Forward Scan (Low I/O): Seek once to target+1, grab forward (handles gaps).
+        3. Backward Scan (Med I/O): Seek to target-offset, grab forward (truncation).
+        """
```
Isotr0py, line 105: Hmm, I feel current recovery implementation is a bit complicated. I wonder if we can recover frames from loaded numpy ndarray directly?

For example, given that `num_expected_framess=6` and `valid_frames=[0, 1, 3, 5]` with `broken_frames = [2, 4]`, we can know that frames used to fix are `[1, 3]` or `[3, 5]`. Then we recover frames with numpy like: 
```python3
num_expected_frames = 6
valid_frames = [0, 1, 3, 5]
invalid_frames = [2, 4]
...  # calculate frames_idx used to fix here
frames_used_to_recover = [1, 3]   # backward
frames_used_to_recover = [3, 5]   # forward
frames[invalid_frames] = frames[frames_used_to_recover]
```
I think this can avoid expensive OpenCV I/O operation, WDYT?

vSeamar: I like your approach as a direct replacement of Cache Check (Zero I/O), as it is much more efficient. But now that I am thinking about it, this step just duplicates the nearest valid frame to the corrupted position, so it does not add any value to the representation of the video. I think this step should not be added and we should simply skip a corrupted frame if a valid frame is "close enough".

The goal of this PR is to recover frames that are not too close to the nearest valid frame, say more than 2 seconds apart for example (since a lot can happen in 2 seconds). I now propose we could make it configurable to how far frames need to be to recover:

## Example: recovery_threshold_seconds=0.5, recovery_offset=5
## Video at 30 FPS → threshold = 0.5s × 30fps = 15 frames

### Dense sampling (every frame)
Failed: frame 80
Nearest valid: frame 79 (1 frame = 33ms)
33ms < 500ms → SKIP recovery (not worth it)

### Sparse sampling (every 40 frames)
Failed: frame 80  
Nearest valid: frame 40 (40 frames = 1.33s)
1.33s > 0.5s → RECOVER! Seek ±5 frames around 80 (still only 2 seeks but tries to recover nearby frames)

I think it should only be implemented when a uses FPS for sampling to avoid the added fps sampling overhead for static frame sampling.
If you think this is worth doing then I can implement it, otherwise I can close the PR and we can continue with the current culling approach where only the successful frames are returned. I personally think this could be useful in the rare case of failure, especially for very low sample rates.

vSeamar: @Isotr0py I see your thumbs up reaction above, what are your thoughts?

Isotr0py: I think we can try to add forward scan with a simplified implementation in this PR first, then consider remained strategy. Because we will need to refactor the loading logic before supporting all three or more recovery strategies, otherwise the video loading logic will become too complicated. 🤔 

We can simply implement forward scan like this, if we found broken frame during loading, we can keep trying loading next possible frame used to recovery until finding a valid ones:
```python3
        i = 0
        valid_frame_indices = []
        failed_frames_idx = []
        for idx in range(max_frame_idx + 1):
            is_target_frame = idx in frame_indices
            ok = cap.grab()
            if not ok:
                # Frame is broken/unreadable, log warning
                if is_target_frame:
                    failed_frames_idx.append(idx)
                continue
            if is_target_frame or (can_recover := can_use_to_recover_frame(idx, failed_frames_idx, ...)):
                ret, frame = cap.retrieve()
                if ret:
                    frames[i] = cv2.cvtColor(frame, cv2.COLOR_BGR2RGB)
                    valid_frame_indices.append(idx)
                    i += 1
                    if can_recover:
                        failed_frames_idx.pop(0)
                elif is_target_frame:
                    failed_frames_idx.append(idx)
```
Here, `can_use_to_recover_frame` should be a util function to check if current frame available for recovery. WDYT?

vSeamar: hm.. I cannot find a video example to test this, tried 10,000 videos and did not come across the initial corrupted frame problem. Without testing it I cannot say if this will be a good add or just code overhead. 
I think I will just close this PR and if I ever come across the issue, I can give your suggestion a try and reopen a new PR and ping you. Thoughts?

Isotr0py: I think this feature is important because we're ignoring corrupted frames when sampling videos, which can cause generation results degradation.

> I cannot find a video example to test this, tried 10,000 videos and did not come across the initial corrupted frame problem. Without testing it I cannot say if this will be a good add or just code overhead.

You can use this script to create corrupted videos: https://github.com/vllm-project/vllm/pull/29001#issuecomment-3553386368

vSeamar: Sounds good, I will test with the script you provided and report back.

vSeamar: @Isotr0py your suggestion worked great and is efficient. I tested it by simulating cap.grab() failing by returning false on certain frames, you can see the test [here](https://github.com/vllm-project/vllm/pull/29197/files#diff-653b66fd465290a62e806f0d7c818654d9822ec2edb99c881eabda02fadc086fR309), I added a small video for an integration test; Making sure the new frame recovery method retrieves more frames than the existing method. I made it an optional "frame_recovery".  Let me know what you think :)

vSeamar: Also updated the PR description.

vllm/multimodal/video.py
```diff
@@ -63,6 +66,126 @@ def load_bytes(
...
+            - valid_frame_indices: List of frame indices that were loaded
+            - recovered_map: Dict mapping recovered_idx -> source_idx
+        """
+        import cv2
+
+        width = int(cap.get(cv2.CAP_PROP_FRAME_WIDTH))
+        height = int(cap.get(cv2.CAP_PROP_FRAME_HEIGHT))
+
+        if width == 0 or height == 0:
+            width, height = 640, 480
```
Isotr0py: I think we can just assert width and height are positive here.

tests/multimodal/assets/simulate_corruption.mp4
Isotr0py, line 1: We have a corrupted video under `tests/multimodal/assets/corrupted.mp4`. Perhaps we can use it for test instead of uploading a duplicate ones?

Isotr0py (approved):
LGTM now, thanks for your patience! Just leave some nits.

vllm/multimodal/video.py
vSeamar: Good idea, will do.

tests/multimodal/assets/simulate_corruption.mp4
vSeamar: Done! We now use only corrupted.mp4 for all recovery tests (refactored to keep same testing functionality bu with only corrupted.mp4). The duplicate video has been removed.

vSeamar:
@Isotr0py 
buildkite/ci/pr seems to not get picked up even after a few hours. Is this normal?

vSeamar:
@Isotr0py Could I get some guidance on how the merging process is? I keep updating branch with main but the checks take very long and I usually have to update it again before they finish. 

Or since you enabled auto-merge do I not have to do anything anymore?

Isotr0py:
> Or since you enabled auto-merge do I not have to do anything anymore?

Usually it can automatically merge as long as there is no conflict with main, even if it's out of date. (Our CI takes a quite long time to run 😅 )

PR #29206, opened 2025-11-21
[Perf] Early return in KVCacheManager.allocate_slots

Jialin:
## Purpose
As the title, early return in KVCacheManager.allocate_slots to speed up scheduling, as most of the requests only allocate a block every 16 steps.

## Test Plan && Test Result
With the change, we could observed allocate slots cost is cut by half from 17.24% to 8.24% over the total runtime (with async scheduling enabled)
Screenshot 2025-11-21 at 1 36 40 PM

Total runtime distribution for allocate_slots per engine step
Screenshot 2025-11-21 at 1 35 50 PM

### Public benchmarks
- Before: 171.53s
- After: 165.90s (~3.3% total runtime reduction)

```
vllm bench latency --model facebook/opt-125m \
  --dtype bfloat16 \
  --batch-size 3000 \
  --input-len 128 \
  --output-len 1800 \
  -tp 1 \
  --num-iters-warmup 1 \
  --num-iters 3
```

---
<details>

</details>

### DIFF
--- a/vllm/v1/core/kv_cache_manager.py
+++ b/vllm/v1/core/kv_cache_manager.py
@@ -284,9 +284,11 @@ class KVCacheManager:
 
         if (
             num_blocks_to_allocate == 0
-            and new_computed_block_list is self.empty_kv_cache_blocks.blocks
+            and new_computed_blocks is None
+            and request.num_output_tokens > 0
         ):
-            # Early return: no new blocks needed to be allocated
+            # Early return: no new blocks needed to be allocated for decode steps
+            # (excluding the first decode steps).
             #
             # NOTE: This optimization may delay block cleanup (remove_skipped_blocks)
             # in rare edge cases, but the impact is negligible.
--- a/vllm/v1/core/kv_cache_manager.py
+++ b/vllm/v1/core/kv_cache_manager.py
@@ -267,16 +267,6 @@ class KVCacheManager:
         else:
             new_computed_block_list = self.empty_kv_cache_blocks.blocks
 
-        # Free the blocks that are skipped during the attention computation
-        # (e.g., tokens outside the sliding window).
-        # We can do this even if we cannot schedule this request due to
-        # insufficient free blocks.
-        # Should call this function before allocating new blocks to reduce
-        # the number of evicted blocks.
-        self.coordinator.remove_skipped_blocks(
-            request.request_id, request.num_computed_tokens
-        )
-
         # The number of computed tokens is the number of computed tokens plus
         # the new prefix caching hits
         num_computed_tokens = request.num_computed_tokens + num_new_computed_tokens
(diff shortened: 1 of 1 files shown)


vllm/v1/core/kv_cache_manager.py
```diff
@@ -292,6 +282,23 @@ def allocate_slots(
             num_encoder_tokens=num_encoder_tokens,
         )
 
+        if (
+            num_blocks_to_allocate == 0
+            and new_computed_block_list is self.empty_kv_cache_blocks.blocks
```
heheda12345:
```suggestion
            and new_computed_blocks is None
```
prefer this slightly.

And do we need to add `num_computed_tokens > request.num_prompt_tokens` to make `remove_skipped_blocks` to be called in the first decode step? This can help to free the prefill tokens used by the last prefill step but is outside the sliding window of the first decode step. Would be grateful if you can try gpt-oss and gemma3, two models with small sliding window size.

heheda12345: And can you add a comment for this may delay `remove_skipped_blocks` and `cache_blocks` and give some analysis about it is fine.

Jialin: Hmmm, great call. I didn't notice the cache_blocks logic. Let me think about the early return logic more :/

Jialin: - Updated per your suggestion
- Add request.num_output_tokens > 0 to ensure we DO NOT early return for prefill and first decode step

Do you feel if it's sufficient?

heheda12345 (approved):
LGTM! Great speedup!

vllm/v1/core/kv_cache_manager.py
```diff
@@ -292,6 +282,37 @@ def allocate_slots(
...
+            and request.num_output_tokens > 0
+        ):
+            # Early return: no new blocks needed to be allocated for decode steps
+            # (excluding the first decode steps). DO NOT early return for
+            # - prefill steps: to ensure cache_blocks are running
+            # - first decode steps: to ensure remove_skipped_blocks remove
+            #   prompt token blocks slided out of the context window
+            #
+            # NOTE: This optimization may delay block cleanup (remove_skipped_blocks)
+            # in rare edge cases, but the impact is negligible.
```
heheda12345, line 297: I think it also delays cache_blocks when num_lookahead_tokens > 0 (e.g., when spec decode is enabled)

Jialin: Hmmm, my understanding is that, if there's no blocks need to be allocated, then there should not be newly created full blocks that needed to be cached?

Did I miss any scenarios? @heheda12345

heheda12345: allocation is based on
```
num_tokens_need_slot = min(
            num_computed_tokens + num_new_tokens + num_lookahead_tokens,
            self.max_model_len,
        )
```
cache is based on 
```
num_tokens_to_cache = min(
            num_computed_tokens + num_new_tokens, request.num_tokens
        )
```
The diff is `num_lookahead_tokens`, which is the predicted tokens for speculative decoding. They need slot to save kv cache, but as the tokens are unverified, they shouldn't be cached.

vllm/v1/core/kv_cache_manager.py
```diff
@@ -292,6 +282,37 @@ def allocate_slots(
             num_encoder_tokens=num_encoder_tokens,
         )
 
+        if (
+            num_blocks_to_allocate == 0
+            and new_computed_blocks is None
+            and request.num_output_tokens > 0
```
heheda12345, line 288: BTW can you double-check whether this should be `request.num_output_tokens > 0` or `request.num_output_tokens > 1`? The last prefill step can compute the first output token so I guess num_output_tokens may be 1 in the first decode step.

vllm/v1/core/kv_cache_manager.py
heheda12345: For chunked local attention + spec decode, is this possible?
1. block X is allocated in step i
2. block X should be cached in step j but not cached due to early return
3. in step k, it is freed by remove_skipped_blocks and changed to null_block, so in cache_blocks, null_block is cached

To fix this, we need to change BlockPool.cache_full_blocks to 
```
for i, blk in enumerate(new_full_blocks):
        if blk is self.null_block:
            continue
```

But before this, can you check whether this will happen?

PR #29209, opened 2025-11-21
[Perf] Introduce a quick access of logprob without creating intermediate Logprob dictionary

Jialin:
## Purpose
Some clients might only need to access the logprob of the sampled token. Introduce FlatLogprob.get_logprob support which allow clients to do so without creating dictionaries.

Also incorporate further optimization suggested by @mgoin offline.

## Test Plan & Test Result
```
pytest -s -v tests/test_logprobs.py 
```

---
<details>

</details>

### DIFF
--- a/vllm/logprobs.py
+++ b/vllm/logprobs.py
@@ -83,19 +83,23 @@ class FlatLogprobs(MutableSequence[LogprobsOnePosition]):
         the intermediate logprob dictionary.
         """
         self.start_indices.append(len(self.logprobs))
-        for token_id, logprob, rank, decoded_token in zip(
-            token_ids, logprobs, ranks, decoded_tokens
-        ):
-            self.token_ids.append(token_id)
-            self.logprobs.append(logprob)
-            self.ranks.append(rank)
-            self.decoded_tokens.append(decoded_token)
+        self.token_ids.extend(token_ids)
+        self.logprobs.extend(logprobs)
+        self.ranks.extend(ranks)
+        self.decoded_tokens.extend(decoded_tokens)
         self.end_indices.append(len(self.logprobs))
 
     def extend(self, logprobs_multi_positions) -> None:
         """Extends the container with logprobs for the next multiple positions"""
         for logprobs_one_position in logprobs_multi_positions:
             self.append(logprobs_one_position)
+
+    def get_logprob(self, position: int, token_id: int) -> float | None:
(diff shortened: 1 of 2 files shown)


vllm/logprobs.py
```diff
@@ -83,20 +83,24 @@ def append_fast(
...
-            token_ids, logprobs, ranks, decoded_tokens
-        ):
-            self.token_ids.append(token_id)
-            self.logprobs.append(logprob)
-            self.ranks.append(rank)
-            self.decoded_tokens.append(decoded_token)
+        self.token_ids.extend(token_ids)
+        self.logprobs.extend(logprobs)
+        self.ranks.extend(ranks)
+        self.decoded_tokens.extend(decoded_tokens)
```
mgoin, line 89: I actually found this caused hangs for me, at least when using beam search. I have no clue why

Jialin: Interesting, does ci would capture the failure? Or there're customized commands to repro?

njhill, line 89: Is this measurably faster? It's not obvious to me that it would be. I guess may depend on how many tokens there are

Jialin: Yeah, probably not quite measurable, but much cleaner as well compared to the previous approach.

vllm/logprobs.py
```diff
@@ -83,20 +83,24 @@ def append_fast(
...
         """
         self.start_indices.append(len(self.logprobs))
-        for token_id, logprob, rank, decoded_token in zip(
-            token_ids, logprobs, ranks, decoded_tokens
-        ):
-            self.token_ids.append(token_id)
-            self.logprobs.append(logprob)
-            self.ranks.append(rank)
-            self.decoded_tokens.append(decoded_token)
+        self.token_ids.extend(token_ids)
```
huachenheli, line 86: Add an assert that all 4 lists should have the same length? (Can they ever differ?)

vllm/logprobs.py
```diff
@@ -83,20 +83,24 @@ def append_fast(
...
+        self.logprobs.extend(logprobs)
+        self.ranks.extend(ranks)
+        self.decoded_tokens.extend(decoded_tokens)
         self.end_indices.append(len(self.logprobs))
```
huachenheli, line 90: nit: Move this to l86 for better readability (so that it's clear start_indices and end_indices are always appended the same amount).

PR #27141, opened 2025-10-18
[MoE] CuteDSL MoE with Nvfp4 DeepEP dispatch 

wenscarl:
In light of. https://github.com/vllm-project/vllm/pull/25778/files 
## Purpose
Dispatch with nvfp4 DeepEP low latency mode. The dispatch is a fusion of quantization and dispatch. 
deps on https://github.com/deepseek-ai/DeepEP/pull/341 and https://github.com/vllm-project/vllm/pull/25990. Should rebase after 25990 is merged.

## Test Plan
NVFP4 dispatch:
VLLM_DEEPEPLL_NVFP4_DISPATCH=1 \
VLLM_USE_FLASHINFER_MOE_FP4=1 \
VLLM_USE_STANDALONE_COMPILE=0 \
VLLM_FLASHINFER_MOE_BACKEND="masked_gemm" \
VLLM_WORKER_MULTIPROC_METHOD=spawn \
VLLM_ALL2ALL_BACKEND="deepep_low_latency" \
lm_eval --model vllm --model_args pretrained=nvidia/DeepSeek-R1-0528-FP4,data_parallel_size=4,enable_expert_parallel=True,tensor_parallel_size=1,enforce_eager=True,max_model_len=2048 --trust_remote_code --tasks gsm8k --num_fewshot 5 --batch_size auto
## Test Result
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9484|±  |0.0061|
|     |       |strict-match    |     5|exact_match|↑  |0.9462|±  |0.0062|

BF16 dispatch:
with `VLLM_DEEPEPLL_NVFP4_DISPATCH=0`:
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9492|±  |0.0060|
|     |       |strict-match    |     5|exact_match|↑  |0.9439|±  |0.0063|
---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
[...]

PR #29211, opened 2025-11-21
Prototyping single batch overlapping for Deepseek EP

mxz297:
Single batch overlapping overlaps routed expert w2 gemm with deepep low latency combine phase 1 + shared expert compute. The cool thing about single batch overlap is the w2 gemm and deepep low latency combine shared a single tensor as a signal buffer. The w2 gemm kernel atomic_add to the signal to indicate the work of individual experts are done and the deepep ll combine kernel waits on the signal to proceed sending after the work is done. So this enables a fine-grain overlap between w2 gemm and combine.

The overlapping can be demonstrated below. Before SBO, we have sequential routed expert w2 gemm, combine phase 1, and then routed expert:

Screenshot 2025-11-21 at 1 30 28 PM

With SBO, we have routed expert w2 gemm and combine phase 1 start at the same time and combine phase 1 overlaps with w2 gemm and shared experts.

Screenshot 2025-11-21 at 10 35 43 AM

The current implementation is very hacky, just as a prototype. Enabling this with VLLM_EP_USE_SBO=1. Depends on https://github.com/vllm-project/vllm/pull/27141 to have flashinfer CuteDSL moe integrated.

Current CuteDSL MoE implementation only supports signaling with no more than 8 local experts, which means we need DP32EP. So we also need this flashinfer branch (https://github.com/mxz297/flashinfer/tree/sbo_extend_local_expert) to test on B200.

Also, we need this DeepEP branch: https://github.com/fzyzcjy/DeepEP/tree/gb200_blog_part_2

```
VLLM_EP_USE_SBO=1 NVSHMEM_IB_ENABLE_IBGDA=1 VLLM_DEEPEP_BUFFER_SIZE_MB=0  VLLM_DEEPEP_LOW_LATENCY_ALLOW_NVLINK=1  VLLM_DEEPEP_LOW_LATENCY_USE_MNNVL=1 VLLM_DEEPEPLL_NVFP4_DISPATCH=1 VLLM_DISABLE_COMPILE_CACHE=1 VLLM_USE_FLASHINFER_MOE_FP4=1 VLLM_USE_STANDALONE_COMPILE=0 VLLM_FLASHINFER_MOE_BACKEND="masked_gemm" VLLM_WORKER_MULTIPROC_METHOD=spawn VLLM_ALL2ALL_BACKEND="deepep_low_latency" VLLM_ATTENTION_BACKEND=FLASHINFER_MLA lm_eval --model vllm --model_args pretrained=/data/local/models/DeepSeek-R1-FP4/,data_parallel_size=8,enable_expert_parallel=True,tensor_parallel_size=1,max_model_len=32768 --trust_remote_code --tasks gsm8k --num_fewshot 5 --batch_size auto

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9409|±  |0.0065|
|     |       |strict-match    |     5|exact_match|↑  |0.9401|±  |0.0065|
```


yewentao256 (commented):
Thanks for the work! Could you also add metrics for performance improvement using `vllm bench`?
eg. `vllm bench serve --model deepseek-ai/DeepSeek-V3.1  --dataset-name random --host 127.0.0.1 --port 9256 --random-input-len 2 --random-output-len 100 --request-rate inf --num-prompts 1024`

hmellor:
Closing this as an abandoned prototype. It has been untouched since 2026-01-02, still depends on two unmerged forks (mxz297/flashinfer `sbo_extend_local_expert` and fzyzcjy/DeepEP `gb200_blog_part_2`) for the `src_signals`/`dst_signals` kernel arguments, and both files it edits have since been refactored into `fused_moe/prepare_finalize/deepep_ll.py` and `fused_moe/experts/flashinfer_cutedsl_moe.py`. The branch also carries a bad merge: committed conflict markers in `tests/kernels/moe/test_cutedsl_moe.py` and a re-added `VLLM_V0_USE_OUTLINES_CACHE`. Happy to see SBO return as a non-prototype PR once the FlashInfer and DeepEP pieces are upstream.

Issue #28427, opened 2025-11-10
[Feature][Kernel]: DeepSeek-R1 QKV Proj is Too Slow for TP


robertgshaw2-redhat:
### 🚀 The feature, motivation and pitch
TRT-LLM has a SwapAB kernel for QKV proj for DSR1. We should integrate this by collaborating with the FlashInfer team

Current situation: we run CUTLASS Block Fp8 for QKV proj because DeepGEMM upstream does not support it
- the CUTLASS Block Fp8 kernels are ~1/2 the speed of DeepGEMM for other Linear layers
- the CUTLASS Block Fp8 kernels have padding overhead

Example Trace:
<img width="1303" height="352" alt="Image" src="https://github.com/user-attachments/assets/dd3baaae-662a-48f6-8add-b99e17abb936" />
> note: in this example, we are using DeepGEMM for the Routed Expert. This will run with Triton / DeepGEMM Swap AB TEP in the final implementation

### Alternatives

- improve cutlass impl. IUUC in recent versions there is swap AB and an option for no padding

### Additional context

cc @mgoin @yewentao256 @LucasWilkinson @MatthewBonanni @alexm-redhat 

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

tunglinwood:
@robertgshaw2-redhat Hi Robert, just for my curiosity, how do I get the kernel progress chart like yours?

robertgshaw2-redhat:
very nice. Note that we also have https://github.com/vllm-project/vllm/pull/28687 which improves vs CUTLASS

Issue #28316, opened 2025-11-07
[Performance][Feature]: make DeepGEMM swapAB available for linear gemm


xuanzic:
### Proposal to improve performance

In flashinfer we already had fp8_gemm_kernel_swapAB kernel for optimizing Mixture of Experts (MOE) GEMM and Dense GEMM operations [reference 1](https://github.com/flashinfer-ai/flashinfer/tree/main/csrc/nv_internal/tensorrt_llm/deep_gemm), [reference 2](https://github.com/flashinfer-ai/flashinfer/blob/main/csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/fp8_blockscale_gemm/fp8_blockscale_gemm_kernel.cuh#L1366), and [reference 3](https://github.com/flashinfer-ai/flashinfer/blob/main/csrc/nv_internal/tensorrt_llm/kernels/cutlass_kernels/fp8_blockscale_gemm/fp8_blockscale_gemm_kernel.cuh#L1493).
This kernel improves performance in small batch scenarios by swapping the input order in matrix multiplication.

These kernels are currently used for:

- MoE operations (exposed via fused_moe module)

- Available in the codebase for Dense GEMM but not exposed for linear/dense layers

---

Proposed Solution:

#### Phase 1: FlashInfer

* **Add Python binding to expose linear operations**
    *  Can follow existing `fused_moe` pattern:
        * Extend current binding or create dedicated binding
    * Add JIT module generation in `flashinfer/jit/gemm/`
    * Expose API in `flashinfer/gemm.py`

#### Phase 2: vLLM Integration

* **Add FlashInfer linear GEMM backend option**
* **Add kernel selection heuristics (M threshold-based)**

### Your current environment (if you think it is necessary)

```text
The output of `python collect_env.py`
```

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

PR #29213, opened 2025-11-22
[Perf][Kernels] Enable FlashInfer DeepGEMM swapAB on SM90 (for W8A8 Linear Op)

katec846:
### **Purpose**

Per https://github.com/vllm-project/vllm/issues/28427, https://github.com/vllm-project/vllm/issues/28316, TRTLLM has `fp8_gemm_kernel_swapAB` kernel for blockscale FP8 gemm in linear layer. This PR https://github.com/vllm-project/vllm/pull/29213 (and FlashInfer PR https://github.com/flashinfer-ai/flashinfer/pull/2131) brings the gemm kernel and quant kernel (to get FP8 scaling) in VLLM.

Description:
M < 32 uses DeepGEMM swapAB from FlashInfer, M>=32 uses DeepGEMM from https://github.com/deepseek-ai/DeepGEMM. Any other fallback (FlashInfer not installed) would be the rest of the gemms (DeepGEMM/Cutlass/etc).

### **Benchmarking**
DeepSeek-V3.1 8xH200 TP8 @ conc=4
```
vllm serve $BASE_MODEL \
    --max-num-seqs 512 \
    --max-num-batched-tokens 4096 \
    --max-model-len 4096 \
    --host 0.0.0.0 \
    --port 8001 \
    --tensor-parallel-size 8 \
    --async-scheduling \
    --no-enable-prefix-caching > $LOGDIR/out.log &
```

```
vllm bench serve \
    --dataset-name random \
    --backend vllm \
    --model $BASE_MODEL \
    --served-model-name $BASE_MODEL \
    --host $HOST \
    --port $PORT \
    --seed $SEED \
    --num-prompts $((5*concurrency)) \
    --max-concurrency 4 \
    --ignore-eos \
    --random-range-ratio 0 \
    --random-input-len 1000 \
    --random-output-len 1000 \
    --random-prefix-len 0 > $LOGDIR/out.log
```

Main with `VLLM_USE_DEEP_GEMM=1`,`VLLM_MOE_USE_DEEP_GEMM=0`,`VLLM_USE_DEEP_GEMM_E8M0=1`
```
Successful requests:                     20        
Maximum request concurrency:             4         
Benchmark duration (s):                  63.55     
Total input tokens:                      19980     
Total generated tokens:                  20000     
Request throughput (req/s):              0.31      
Output token throughput (tok/s):         314.70    
Peak output token throughput (tok/s):    328.00    
Peak concurrent requests:                8.00      
Total token throughput (tok/s):          629.08    
Mean TTFT (ms):                          203.15    
Median TTFT (ms):                        229.50    
P99 TTFT (ms):                           264.69    
Mean TPOT (ms):                          12.52     
Median TPOT (ms):                        12.53     
P99 TPOT (ms):                           12.67     
Mean ITL (ms):                           12.52     
Median ITL (ms):                         12.49     
P99 ITL (ms):                            12.98     

```

This PR with `VLLM_USE_FLASHINFER_FP8_LINEAR=1,VLLM_USE_DEEP_GEMM=1`,`VLLM_MOE_USE_DEEP_GEMM=0`,`VLLM_USE_DEEP_GEMM_E8M0=1` **(1.11x more otps)**
```
Successful requests:                     20        
Maximum request concurrency:             4         
Benchmark duration (s):                  57.05     
Total input tokens:                      19980     
Total generated tokens:                  20000     
Request throughput (req/s):              0.35      
Output token throughput (tok/s):         350.57    
Peak output token throughput (tok/s):    372.00    
Peak concurrent requests:                8.00      
Total token throughput (tok/s):          700.78    
Mean TTFT (ms):                          203.37    
Median TTFT (ms):                        230.57    
P99 TTFT (ms):                           257.30    
Mean TPOT (ms):                          11.22     
Median TPOT (ms):                        11.22     
P99 TPOT (ms):                           11.39     
Mean ITL (ms):                           11.22     
Median ITL (ms):                         11.18     
P99 ITL (ms):                            11.72     
```

The rest of the sweep, since this affects M < 32 gemms.
image

Same sweeps but on Qwen3-235B-A22B-FP8, 4xH200.
image

image

OTPS/user Speedup Table

Concurrency | DS‑V3.1 TP8 | Qwen3‑235B‑A22B‑FP8 TP4 | Qwen3‑235B‑A22B‑FP8 TEP4 
-- | -- | -- | --
1 | 1.13 | 1.10 | 1.15
2 | 1.21 (may be node jitter) | 1.09 | 1.14
4 | 1.11 | 1.06 | 1.06
8 | 1.08 | 1.05 | 1.07
16 | 1.05 | 1.01 | 1.01
32 | 1.00 | 0.99 | 0.99
64 | 1.01 | 1.01 | 1.01
128 | 1.01 | 1.01 | 1.01

**Nsys**
DeepSeek-V3.1 TP8 @ conc=4
Main: running DeepGEMM, attention layer 89us, gen step 12.46ms
image

PR: running DeepGEMM swapAB, attention layer 75us (1.18x faster), gen step 11.28ms (1.10x faster)
image

### **Acc test**
```
VLLM_USE_DEEP_GEMM_E8M0=1 VLLM_USE_DEEP_GEMM=1 VLLM_MOE_USE_DEEP_GEMM=0 VLLM_USE_FLASHINFER_FP8_LINEAR=1 lm_eval --model vllm --model_args "pretrained=/scratch_omniml_data_2/models/DeepSeek-V3.1,tensor_parallel_size=8,max_model_len=32768" \
  --trust_remote_code --tasks gsm8k --num_fewshot 5 --batch_size 16
```
Main:
```
[0;36m(EngineCore_DP0 pid=43645)[0;0m INFO 12-19 09:54:30 [core.py:256] init engine (profile, create kv cache, warmup model) took 145.81 seconds
INFO 12-19 09:54:31 [llm.py:360] Supported tasks: ['generate']
vllm (pretrained=/scratch_omniml_data_2/models/DeepSeek-V3.1,tensor_parallel_size=8,max_model_len=32768,trust_remote_code=True), gen_kwargs: (None), limit: None, num_fewshot: 5, batch_size: 16
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9621|±  |0.0053|
|     |       |strict-match    |     5|exact_match|↑  |0.9598|±  |0.0054|
```
This PR:
```
[0;36m(EngineCore_DP0 pid=11123)[0;0m INFO 12-19 09:44:06 [core.py:256] init engine (profile, create kv cache, warmup model) took 647.95 seconds
INFO 12-19 09:44:08 [llm.py:360] Supported tasks: ['generate']
vllm (pretrained=/scratch_omniml_data_2/models/DeepSeek-V3.1,tensor_parallel_size=8,max_model_len=32768,trust_remote_code=True), gen_kwargs: (None), limit: None, num_fewshot: 5, batch_size: 16
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9553|±  |0.0057|
|     |       |strict-match    |     5|exact_match|↑  |0.9530|±  |0.0058|
```

Full Accuracy Test ran:
The accuracy test focuses on low batch size having no accuracy error, since it affects M<32, and high batch size staying the same (also having no accuracy error).

Batch size | Match mode | Qwen3‑235B‑A22B‑FP8 Main (DeepGEMM) | Qwen3‑235B‑A22B‑FP8 PR (DeepGEMM swapAB) | DeepSeek‑V3.1 Main (DeepGEMM) | DeepSeek‑V3.1 PR (DeepGEMM swapAB)
-- | -- | -- | -- | -- | --
bs=16 | flexible-extract | 0.8681 | 0.8658 | 0.9621 | 0.9553
bs=16 | strict-match | 0.8469 | 0.8552 | 0.9598 | 0.9530
bs=32 | flexible-extract | 0.8650 | 0.8643 | 0.9560 | 0.9598
bs=32 | strict-match | 0.8385 | 0.8408 | 0.9545 | 0.9606
bs=auto | flexible-extract | 0.8802 | 0.8688 | 0.9553 | 0.9568
bs=auto | strict-match | 0.8590 | 0.8446 | 0.9553 | 0.9575

**Note:** The reason of not using all FlashInfer Blockscale FP8 kernels for linear layer is we're seeing accuracy drop from 95% to 88% when running GSM8K. There's an internal logic in CPP that uses non-swapAB kernel for M>=32 gemms. Therefore an alternative way is only use swapAB in M<32 scenario, leveraging torch.cond to switch between FlashInfer and DeepGEMM while torch.compile captures the condition. The accuracy for the larger gemm in FlashInfer FP8 may be resolved in future PR since currently from acc benchmark there's no accuracy drop now.


yewentao256 (commented):
Thanks for the work!
A few thoughts:
1. has nothing to do with deepgemm, please update the title and description
2. Please benchmark and show the performance improvement using `vllm bench ...`
3. `lm_eval` to make sure acc is correct

vllm/model_executor/layers/quantization/utils/fp8_utils.py
```diff
@@ -226,6 +231,83 @@ def _triton_per_token_group_quant_fp8_fake(
 )
 
 
+def _flashinfer_fp8_blockscale_gemm_impl(
```
yewentao256, line 234: Could we split this function? I don't think it is a good idea to bind flashinfer deepgemm and real deepgemm together.

vllm/model_executor/layers/quantization/utils/fp8_utils.py
```diff
@@ -226,6 +231,83 @@ def _triton_per_token_group_quant_fp8_fake(
...
+    weight_scale: torch.Tensor,
+    group_size: int,
+    use_deep_gemm_e8m0: bool,
+) -> torch.Tensor:
+    return torch.empty(
+        input.shape[0], weight.shape[0], dtype=torch.bfloat16, device=input.device
+    )
+
+
+direct_register_custom_op(
```
yewentao256, line 333: Is this a must-have? If not compling related, we can directly call without registing a custom op

yewentao256 (commented):
Nice work! A few thoughts

Also, please update the PR title and description to Flashinfer DeepGEMM, which is not doing the swap with the DeepGEMM repo https://github.com/deepseek-ai/DeepGEMM

vllm/model_executor/layers/quantization/utils/fp8_utils.py
jhaotingc: This is a must have, for torch compile to do dummy run. 
I tried to use the plain function, but it broke at torch dynamo compile step.

vllm/model_executor/layers/quantization/utils/fp8_utils.py
jhaotingc: How would you like to split it?

The main issue is that we need `M<32` logic to switch between FlashInfer swapAB and original DeepGEMM to maintain accuracy (GSM8K 88 -> 95).

In order to keep this logic and still be capturable with torch compile, it needs to be in the same torch op, using `torch.cond` to switch according to the condition. Otherwise (using simple if else), I inspected nsys, torch compile failed to capture the M<32 (always uses original DeepGEMM) somehow. It may be because it's an input dependent operation and may be discarded during graph compilation.

When not using torch compile, (only CUDA graph), I tested a simple if condition is sufficient. However there's a hit on performance.

jhaotingc: And thanks for the review!!

jhaotingc:
> Nice work! A few thoughts
> 
> Also, please update the PR title and description to Flashinfer DeepGEMM, which is not doing the swap with the DeepGEMM repo https://github.com/deepseek-ai/DeepGEMM

Change the title and the description! Thanks!

vllm/model_executor/layers/quantization/utils/fp8_utils.py
```diff
@@ -226,6 +231,83 @@ def _triton_per_token_group_quant_fp8_fake(
...
+        )
+        return output
+
+    # there is only no benefit of using FlashInfer DeepGEMM for higher batch sizes since
+    # the swapAB optimization is only effective for small batch sizes.
+    # there is slight accuracy loss when using FlashInfer blockscale gemm for all batch
+    # sizes for DeepSeek-V3.
+    condition = input.shape[0] < 32
+
+    # torch.cond for torch compile compatibility
```
yewentao256: Please comment with more context as we discussed before, and it is needed to pack deepgemm func and flashinfer func together

vllm/model_executor/layers/quantization/utils/fp8_utils.py
```diff
@@ -226,6 +231,83 @@ def _triton_per_token_group_quant_fp8_fake(
...
+        weight_scale: torch.Tensor,
+    ) -> torch.Tensor:
+        return flashinfer_fp8_blockscale_gemm(
+            input=input,
+            weight=weight,
+            weight_scale=weight_scale,
+            out_dtype=torch.bfloat16,
+        )
+
+    def use_deepgemm(
```
yewentao256:
```suggestion
    def run_deepgemm(
```

vllm/model_executor/layers/quantization/utils/fp8_utils.py
```diff
@@ -226,6 +231,83 @@ def _triton_per_token_group_quant_fp8_fake(
...
 
 
+def _flashinfer_fp8_blockscale_gemm_impl(
+    input: torch.Tensor,
+    weight: torch.Tensor,
+    weight_scale: torch.Tensor,
+    group_size: int,
+    use_deep_gemm_e8m0: bool,
+) -> torch.Tensor:
+    def use_flashinfer_deepgemm_swapAB(
```
yewentao256:
```suggestion
    def run_flashinfer_deepgemm_swapAB(
```

yewentao256 (commented):
Thanks! A few more thoughts

And please do not force push, it might swallow previous commits and comments

vllm/model_executor/layers/quantization/utils/fp8_utils.py
jhaotingc: Left a more detailed comment as function docstring, in the comment as well. Thanks!

vllm/model_executor/layers/quantization/utils/fp8_utils.py
jhaotingc: Fixed.

vllm/model_executor/layers/quantization/utils/fp8_utils.py
jhaotingc: Fixed.

yewentao256 (approved):
LGTM, thanks for the work!

Maybe we can consider set `VLLM_BLOCKSCALE_FP8_GEMM_FLASHINFER` default to True as a following up PR (more bench results needed)

robertgshaw2-redhat (approved):

PR #29216, opened 2025-11-22
[BugFix] Fix returned logprobs with spec decode + prefill chunking

njhill:
The produced logprobs lists in this case were incorrect due incorrectly taking into account sampled token "discards" when computing offsets into the logprobs tensors.

Also:
- Fix error with bf16 models and "raw_logits" mode
- Use smaller model for test

Original PR which had this mistake: https://github.com/vllm-project/vllm/pull/26060

cc @TheEpicDolphin


DarkLight1337 (approved):
Thanks for fixing!

rasmith:
Yes, this works, thanks for fixing @njhill !

PR #29223, opened 2025-11-22
[Core] Support logprobs with spec decode + async scheduling 

njhill:


tests/v1/sample/test_logprobs.py
```diff
@@ -597,6 +600,8 @@ def test_spec_decode_logprobs(
     # Per-token logprobs are expected to be the same.
     assert len(ref_logprobs) == len(spec_logprobs)
     for ref_logprob, spec_logprob in zip(ref_logprobs, spec_logprobs):
-        assert math.isclose(ref_logprob.logprob, spec_logprob.logprob, abs_tol=1e-3)
+        assert math.isclose(
+            ref_logprob.logprob, spec_logprob.logprob, rel_tol=5e-2, abs_tol=1e-1
+        )
```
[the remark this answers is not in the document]
njhill: The test was previously less sensitive to batching variance because the prefill wasn't chunked.

vllm/v1/sample/rejection_sampler.py
```diff
@@ -220,10 +225,15 @@ def parse_output(
         valid_mask = (output_token_ids_np != PLACEHOLDER_TOKEN_ID) & (
             output_token_ids_np < vocab_size
         )
+        cu_num_tokens = None
+        if return_cu_num_tokens:
+            cu_num_tokens = [0] + valid_mask.sum(axis=1).cumsum().tolist()
+        if len(discard_req_indices) > 0:
```
benchislett, line 231: Why is this done _after_ computing `cu_num_tokens`?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -214,28 +214,29 @@ def get_output(self) -> ModelRunnerOutput:
 
         This function blocks until the copy is finished.
         """
+        max_gen_len = self.sampled_token_ids_cpu.shape[-1]
         self.async_copy_ready_event.synchronize()
 
         # Release the device tensors once the copy has completed.
         del self._logprobs_tensors
         del self._sampled_token_ids
-        max_gen_len = self.sampled_token_ids_cpu.shape[-1]
```
benchislett, line 222: why is this moved?

benchislett (commented):
Looks good overall

vllm/v1/worker/gpu_model_runner.py
njhill: Just in the interest of moving everything prior to the sync that can be :)

vllm/v1/sample/rejection_sampler.py
njhill: Because `cu_num_tokens` is used to index into the logprobs tensors that don't take the discarded indices into account, so doing it beforehand results in incorrect output.

Originally it was done before which was a bug, fixed by https://github.com/vllm-project/vllm/pull/29216.

benchislett (approved):

PR #29225, opened 2025-11-22
[CI/Build] Add dtype to LLM creation in v1/sample/test_logprobs.py

rasmith:
This PR adds the `dtype` to LLM creation in `v1/sample/test_logprobs.py` since the model will use `bfloat16 `if the dtype is not specified which leads to a `numpy` error, where `numpy` does not support `bfloat16 `for lists.

Adding the `dtype` allows all the tests to pass.

The test now returns:

23 passed, 7 skipped, 4 warnings

### DIFF
--- a/tests/v1/sample/test_logprobs.py
+++ b/tests/v1/sample/test_logprobs.py
@@ -556,6 +556,7 @@ def test_spec_decode_logprobs(
         seed=42,
         logprobs_mode=logprobs_mode,
         gpu_memory_utilization=0.4,
+        dtype=DTYPE,
     )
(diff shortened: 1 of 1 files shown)


DarkLight1337:
Why didn't this fail in the regular CI? I think using logprobs shouldn't depend on the dtype. cc @njhill

DarkLight1337:
Can you check if https://github.com/vllm-project/vllm/pull/29216 solves it without having to modify the tests?

njhill:
Agree with @DarkLight1337 we should fix the code rather than the tests. Should hopefully be fixed by https://github.com/vllm-project/vllm/pull/29216.

PR #29226, opened 2025-11-22
[CI/Build] Update requirements/rocm-test.txt with newer version of lm-eval so v1/sample/test_logprobs_e2e.py passes

rasmith:
This PR updates the the version of `lm-eval` used in  `requirements/rocm-test.txt` to prevent an error when `test_logprobs_e2e.py` tries to use `lm-eval`.  

The error arises when the older version tries to directly pass in `prompt_token_ids` to `llm.generate` which no longer takes that parameter and instead takes a `prompts: PromptType | Sequence[PromptType]` parameter.  The newer version of `lm-eval` correctly passes in this parameter which fixes the issue.

The test now passes with:

2 passed, 6 warnings


DarkLight1337 (approved):

requirements/rocm-test.txt
```diff
@@ -38,7 +38,7 @@ schemathesis==3.39.15
 mteb[bm25s]>=1.38.11, <2
 
 # Required for eval tests
-lm-eval[api] @ git+https://github.com/EleutherAI/lm-evaluation-harness.git@206b7722158f58c35b7ffcd53b035fdbdda5126d
+lm-eval[api] @ git+https://github.com/EleutherAI/lm-evaluation-harness.git@d021bf846218f1bb3bdc0603864789329476d464
```
mgoin, line 41: Can we move to a stable release now instead of a source commit?

DarkLight1337: There hasn't been a stable release since that commit hash

DarkLight1337 (approved):

DarkLight1337:
Actually wait, this doesn't make much sense, since the lm-eval does pass with the version `206b7722158f58c35b7ffcd53b035fdbdda5126d` using regular CI...

DarkLight1337 (changes requested):
Could you explain what is the difference in the AMD CI compared to CUDA CI?

PR #29236, opened 2025-11-22
Fix gpt oss tool parser v2

ShaikAbdulHafeez03:
Purpose
This PR fixes critical issues with tool call extraction in the OpenAIToolParser (Harmony/gpt-oss), specifically when using Custom Tools and MCP (Model Context Protocol).

Previously, the parser failed to correctly capture tool calls that resided in the active stream buffer (current_content) but hadn't yet been finalized into a message object. This led to two major issues:

Tool Call Failure: valid tool calls were dropped or ignored.

Channel Leakage/Hallucination: The model appeared to hallucinate or leak tool call structures into the final (user-facing) message channel instead of processing them as structured tool calls.

Changes Implemented:

Active Buffer Parsing: Added logic to inspect parser.current_content and parser.current_channel. If the active channel is commentary (used for functions in Harmony), the content is now correctly extracted as a tool call.

Tag Sanitization: Added a sanitization step to strip internal tags (e.g., <|channel|>) from function names.

Robust JSON Handling: Improved JSON parsing for tool arguments to handle partial or slightly malformed buffers.

Test Plan
Scenario: Run a request using a Custom Tool or MCP tool definition (e.g., web_search or a custom python function).

# Send a request that forces custom  tool call or try calling a custom mcp server (it fail's calling multiple tools because of parsing issue)
Test Result
Before: When using custom tools/MCP, the tool calling would frequently fail. The model output would often "hallucinate" or leak the raw tool call JSON into the final user message string, failing to trigger the actual tool execution on the client side.

After: Tool calls are cleanly extracted from the commentary channel. The final message content remains clean, and custom tools/MCP function correctly without leakage.

<details> <summary> Essential Elements of an Effective PR Description Checklist </summary>

[x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".

[x] The test plan, such as providing test command.

[x] The test results, such as pasting the results comparison before and after, or e2e results

[ ] (Optional) The necessary documentation update, such as updating supported_models.md and examples for a new model.

[ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).

</details>

BEFORE SUBMITTING, PLEASE READ https://docs.vllm.ai/en/latest/contributing (anything written below this line will be removed by GitHub Actions)


DarkLight1337:
Can you merge from main to fix pre-commit?

bbrowning:
> Previously, the parser failed to correctly capture tool calls that resided in the active stream buffer (current_content) but hadn't yet been finalized into a message object. This led to two major issues:

If there is content in current_content and not a finalized message, that means the model stopped mid-generation for some reason. Proper tool calls all end with a `<|call|>` token, which is considered a stop token and the tool call will end up in a message and not current_content of the parser.

The changes in #28729 will help many of the issues this proposes to fix in the parser. The root cause of many of these is we are prompting gpt-oss models wrong in many multi-turn scenarios, especially ones involving tool calling. So, what the changes in that PR do is fix how we're prompting the models which helps many of the situations where `<|channel|>` may leak into function recipients or we see other issues with models following their harmony format.

Some level of fuzzy parsing could be useful, but the changes here are large enough that I'd want to see what kind of impact this has on model accuracy and overall parsing success. Do you have any examples of requests that did not work before that this fixes? Have you run this against any tool calling benchmark suites to ensure it doesn't regress in other areas?

I'm pushing on the accuracy point here as I've spent a large amount of time benchmarking tool call accuracy in the gpt-oss models over the last few weeks to come up with the changes in #28729 and want to be careful about potentially regressing on any of that.

ApostaC:
cc @aarnphm @chanh

PR #29244, opened 2025-11-22
[ROCm] Fallback pytorch GELU with tanh approximation to GELU()

divakar-amd:
There is a bug when using native Pytorch's implementation of `F.gelu(..., approximate='tanh')`  with torch.compile on ROCm. Because of this, models such as _gemma3_ and _starcoder2_ produces garbage output.
This PR is a quick fix "band-aid" solution to avoid using tanh approximation and fallback to the general GELU() activation until the actual fix is found. 
Technically, tanh approximation for GELU is faster than normal GELU, hence, this PR might add a slight overhead in activation function. However, the correctness shouldn't be affected.
Tested using (torch.compile mode) :
`pytest -s -v test_correctness_sliding_window.py`


vllm/model_executor/layers/activation.py
```diff
@@ -209,6 +216,12 @@ def __init__(self, approximate: str = "none"):
                 self.op = torch.ops._C.gelu_and_mul
             elif approximate == "tanh":
                 self.op = torch.ops._C.gelu_tanh_and_mul
+        if current_platform.is_rocm() and approximate == "tanh":
+            logger.warning_once(
+                "[ROCm] PyTorch's native GELU with tanh approximation is unstable "
+                "with torch.compile. For native implementation, fallback to 'none' "
+                "approximation. The custom kernel implementation is unaffected."
+            )
```
mgoin, line 224: Why don't you set self.approximate here?

divakar-amd: So that we can still use the tanh approximation for the `forward_cuda` (customOp) path. The custom kernel `torch.ops._C.gelu_tanh_and_mul` works fine

tjtanaa (approved):
LGTM

PR #25431, opened 2025-09-22
[Core][Hybrid allocator + connector 2/n] Unify `remove_skipped_blocks` by `get_last_useful_token`

KuntaiDu:
## Purpose

This PR is separated from #23624 that aims to enable allocating tokens inside sliding window when long prefix is not cached in vLLM but cached in connector.

This PR unifies `remove_skipped_blocks` by adding a new function `get_last_useful_token`.

## Test Plan

`pytest -v -s tests/v1/core/test_single_type_kv_cache_manager.py`

## Test Result

image

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #29272, opened 2025-11-23
[V1] [Hybrid] Lighter Mamba Prefix Caching with standard memory layout

peakcrosser7:
#28176 with with standard memory layout

## Test Result

---
<details>

</details>


vllm/v1/attention/backends/gdn_attn.py
```diff
@@ -57,9 +58,18 @@ class GDNAttentionMetadata:
     batch_ptr: torch.Tensor | None = None
     token_chunk_offset_ptr: torch.Tensor | None = None
 
+def mamba_gather_indices(common_attn_metadata: CommonAttentionMetadata,
```
heheda12345: nit: Will it be faster & clearer to write a numba (cpu) / triton (gpu) kernel?

vllm/v1/core/sched/scheduler.py
```diff
@@ -224,35 +215,62 @@ def schedule(self) -> SchedulerOutput:
...
             )
 
             # Schedule encoder inputs.
             encoder_inputs_to_schedule = None
-            external_load_encoder_input: list[int] = []
             new_encoder_compute_budget = encoder_compute_budget
             if request.has_encoder_inputs:
                 (
                     encoder_inputs_to_schedule,
                     num_new_tokens,
```
heheda12345, line 359: reminder: num_new_tokens is updated here.

vllm/v1/core/sched/scheduler.py
```diff
@@ -224,35 +215,62 @@ def schedule(self) -> SchedulerOutput:
...
+                and self.cache_config.enable_prefix_caching
+                and self._has_mamba_spec()):
+                # To enable block-aligned caching of the Mamba state, `num_new_tokens`
+                # must be a multiple of `block_size`.
+                # As an exception, if `num_new_tokens` is less than `block_size`, the
+                # state is simply not cached, requiring no special handling.
+                # Additionally, when Eagle mode is enabled, FullAttn prunes the last
+                # matching block. To prevent this from causing a Mamba cache miss, the
+                # last chunk must be larger than `block_size`.
+                block_size = self.block_size
```
heheda12345: I can't understand this part of code. I thought we only need something like:
```
if request.num_output_tokens == 0: # prefill
    last_cache_position = request.num_prompt_tokens - request.num_prompt_tokens % block_size
    # eagle prune
    if self.use_eagle: last_cache_position  = max(last_cache_position - block_size, 0)
    num_computed_tokens_after_prefill = request.num_computed_tokens + num_new_tokens 
    if num_computed_tokens_after_prefill  < last_cache_position:
        num_new_tokens = num_new_tokens // block_size * block_size # align to block_size
    elif request.num_computed_tokens < last_cache_position and last_cache_position < num_computed_tokens_after_prefill:
        num_new_tokens = last_cache_position  -  request.num_computed_tokens # force to cache the last chunk
    else:
        pass # prefill the last few tokens
```

vllm/v1/core/sched/scheduler.py
```diff
@@ -270,73 +288,58 @@ def schedule(self) -> SchedulerOutput:
                 #    its max_total_tokens or max_model_len.
                 # 2. The encoder budget is exhausted.
                 # 3. The encoder cache is exhausted.
+                # 4. Insufficient budget for a block-aligned chunk in hybrid 
+                #    models with lighter mamba prefix caching.
```
heheda12345: in this case, should we allow the prefill of all scheduled tokens instead of forcing block-aligned chunk?

vllm/v1/core/sched/scheduler.py
heheda12345: `num_new_tokens = num_new_tokens // block_size * block_size` may not work if we don't force chunk align in this case
https://github.com/vllm-project/vllm/pull/29272/files#r2555167588

vllm/v1/core/sched/scheduler.py
```diff
@@ -505,14 +495,32 @@ def schedule(self) -> SchedulerOutput:
...
-                        not self.scheduler_config.enable_chunked_prefill
+                        not self.scheduler_config.chunked_prefill_enabled
                         and num_new_tokens > token_budget
                     ):
                         self.waiting.pop_request()
                         skipped_waiting_requests.prepend_request(request)
                         continue
 
-                    num_new_tokens = min(num_new_tokens, token_budget)
+                    if (envs.VLLM_USE_LIGHTER_MAMBA_CACHE
```
heheda12345: make this a util function to avoid code duplication of first prefill / chunked prefill?

vllm/v1/core/kv_cache_manager.py
```diff
@@ -304,7 +307,9 @@ def allocate_slots(
                 "Computed blocks should be empty when prefix caching is disabled"
             )
 
-        if new_computed_block_list is not self.empty_kv_cache_blocks.blocks:
+        if (envs.VLLM_USE_LIGHTER_MAMBA_CACHE
```
heheda12345: why do you need an if here?

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -647,6 +599,28 @@ def find_longest_cache_hit(
 
         return computed_blocks
 
+    def remove_skipped_blocks(self, request_id: str,
```
heheda12345: can you rebase the PR to include the recent changes like https://github.com/vllm-project/vllm/pull/25431?

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -662,28 +636,128 @@ def get_num_blocks_to_allocate(
...
+                    * self.kv_cache_spec.num_speculative_blocks
+                )
+            return super().get_num_blocks_to_allocate(
+                request_id, num_tokens, new_computed_blocks
             )
-        return super().get_num_blocks_to_allocate(
-            request_id, num_tokens, new_computed_blocks
-        )
+        else:
+            num_required_blocks = cdiv(num_tokens, self.block_size) + self.num_speculative_blocks
```
heheda12345: is it ok to always return `min(self.num_speculative_blocks + 1, super().get_num_blocks_to_allocate(...))` or 
```
if is_prefill: # I don't have a good idea on how to check is_prefill now
     return min(1, super().get_num_blocks_to_allocate(...))
else:
     return min(self.num_speculative_blocks + 1, super().get_num_blocks_to_allocate(...))
```

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -662,28 +636,128 @@ def get_num_blocks_to_allocate(
...
+            # free queue and ref_cnt == 0), it will be changed from a free block
+            # to a computed block when the request is allocated, so we also count
+            # it as needed to be allocated.
+            num_evictable_computed_blocks = sum(
+                blk.ref_cnt == 0 and not blk.is_null
+                for blk in new_computed_blocks)
+            
+            return num_new_alloc_blocks + num_evictable_computed_blocks
+
+    def save_new_computed_blocks(
```
heheda12345, line 826: remove this function?

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -662,28 +636,128 @@ def get_num_blocks_to_allocate(
...
+                        req_blocks = req_blocks[:-self.num_speculative_blocks]
+                        # TODO: reuse blocks 是否需要清除内存?尤其是decode
+                        reuse_blocks = req_blocks[-self.num_speculative_blocks:]
+                        new_blocks.extend(reuse_blocks)
+                new_alloc_blocks = self.block_pool.get_new_blocks(num_new_alloc_blocks)
+                new_blocks.extend(new_alloc_blocks)
+                req_blocks.extend(new_blocks)
+                return new_blocks
+            
+    def cache_blocks(self, request: Request, num_tokens: int) -> None:
```
heheda12345: remove this function?

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -662,28 +636,128 @@ def get_num_blocks_to_allocate(
...
-        return super().allocate_new_blocks(request_id, num_tokens)
+        if not envs.VLLM_USE_LIGHTER_MAMBA_CACHE:
+            if self.kv_cache_spec.num_speculative_blocks > 0:
+                num_tokens += (
+                    self.kv_cache_spec.block_size
+                    * self.kv_cache_spec.num_speculative_blocks
+                )
+            return super().allocate_new_blocks(request_id, num_tokens)
+        else:
+            req_blocks: list[KVCacheBlock] = self.req_to_blocks[request_id]
```
heheda12345, line 852: todo: will check this later

vllm/v1/attention/backends/gdn_attn.py
peakcrosser7: Yep, that's the plan. This is just a temporary helper function right now. It'll eventually be moved somewhere central so different Mamba variant metadata can all call it to get their state_indices.

vllm/v1/core/sched/scheduler.py
peakcrosser7: > I can't understand this part of code. I thought we only need something like:
> 
> ```
> if request.num_output_tokens == 0: # prefill
>     last_cache_position = request.num_prompt_tokens - request.num_prompt_tokens % block_size
>     # eagle prune
>     if self.use_eagle: last_cache_position  = max(last_cache_position - block_size, 0)
>     num_computed_tokens_after_prefill = request.num_computed_tokens + num_new_tokens 
>     if num_computed_tokens_after_prefill  < last_cache_position:
>         num_new_tokens = num_new_tokens // block_size * block_size # align to block_size
>     elif request.num_computed_tokens < last_cache_position and last_cache_position < num_computed_tokens_after_prefill:
>         num_new_tokens = last_cache_position  -  request.num_computed_tokens # force to cache the last chunk
>     else:
>         pass # prefill the last few tokens
> ```

Got it, your implementation is much more concise!
This part of your code should be executed after `num_new_tokens = min(num_new_tokens, token_budget)`.

peakcrosser7: > `num_new_tokens = num_new_tokens // block_size * block_size` may not work if we don't force chunk align in this case https://github.com/vllm-project/vllm/pull/29272/files#r2555167588

Yes, details in that comment.

vllm/v1/core/sched/scheduler.py
peakcrosser7: Thanks for the reminder! You're right, I missed the encoder case and will move the block-aligned logic after this section. 
By the way, does this block-aligned logic conflict with the encoder input?

vllm/v1/core/sched/scheduler.py
peakcrosser7: We can't do that. For a single prompt, if any intermediate chunk is not block-aligned, we can not bind the computed tokens to a block's hash in next chunks. 
And I think trying to re-align by adjusting subsequent chunk sizes would make the logic overly complex.

vllm/v1/core/sched/scheduler.py
peakcrosser7: Yep, I will do it

vllm/v1/core/kv_cache_manager.py
peakcrosser7: This is related to a previous version and no longer needed.

vllm/v1/core/single_type_kv_cache_manager.py
peakcrosser7: ok, I will do it

vllm/v1/core/single_type_kv_cache_manager.py
peakcrosser7: mistake, should call `super().save_new_computed_blocks()`

vllm/v1/core/single_type_kv_cache_manager.py
peakcrosser7: Let me think... If we can distinguish between the prefill and decode, we might not need to deal with the complex logic of reusing blocks.

vllm/v1/core/single_type_kv_cache_manager.py
peakcrosser7: mistake, same as `save_new_computed_blocks()`

vllm/v1/core/single_type_kv_cache_manager.py
peakcrosser7: I'm finding that the current design still needs `remove_skipped_blocks()` instead of just `get_num_skipped_tokens()`. 
The reason is that in `_preprocess_mamba()`, we copy the latest immutable block into a new allocated one, and that immutable block can only be freed in the next step. 
My plan is to use a dict `_req_to_last_computed` to track `last_computed_tokens` for each request. However, `get_num_skipped_tokens()` doesn't accept the `req_id` parameter, which prevents this.
Is there a better solution here?

vllm/v1/core/sched/scheduler.py
heheda12345: The aligned num_new_tokens can be computed with 
```
num_computed_tokens_after_prefill = num_computed_tokens_after_prefill // block_size * block_size
if num_computed_tokens_after_prefill > num_computed_tokens:
     num_new_tokens = num_computed_tokens_after_prefill - num_computed_tokens
else:
    # don't change
    pass
 ```

But I think it may also be fine to keep the current implementation

File: vllm/v1/core/sched/scheduler.py [hunk 12]
```diff
@@ -224,14 +262,27 @@ def schedule(self) -> SchedulerOutput:
...
-            if 0 < self.scheduler_config.long_prefill_token_threshold < num_new_tokens:
-                num_new_tokens = self.scheduler_config.long_prefill_token_threshold
-            num_new_tokens = min(num_new_tokens, token_budget)
+            logger.info(f'>>> [DEBUG] Scheduler: schedule RUNING: req_id={request.request_id}, '
+                        f'num_prompt_tokens={request.num_prompt_tokens=}')
+            # Ensure new tokens for a request in the prefill phase do not contain 
+            # sps tokens, especially in the last prefill chunk. For a hybrid-model, 
+            # extra sps tokens would corrupt the generated Mamba state.
+            # TODO: This logic does not yet handle resumed requests.
+            if request.num_computed_tokens < request.num_prompt_tokens:
```
heheda12345: What's the problem of num_tokens_with_spec > num_computed_tokens?

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -662,28 +726,113 @@ def get_num_blocks_to_allocate(
...
+            num_tokens -= (self.num_speculative_blocks
```
heheda12345: please note this:
```
num_tokens_need_slot = min(
            num_computed_tokens + num_new_tokens + num_lookahead_tokens,
            self.max_model_len,
        )
```
if `num_computed_tokens + num_new_tokens + num_lookahead_tokens > max_model_len`, we shouldn't do `-=num_speculative_blocks `

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -645,8 +684,33 @@ def find_longest_cache_hit(
...
+                              num_computed_tokens: int) -> None:
+        assert isinstance(self.kv_cache_spec, MambaSpec)
+        if not envs.VLLM_USE_LIGHTER_MAMBA_CACHE:
+            # Here unused blocks may be freed up for running requests.
+            # TODO(@s3woz) Free up all blocks that aren't needed by Mamba2
+            #  (for which find_longest_cache_hit returns block_pool.null_block)
+            pass
+        else:
+            blocks: list[KVCacheBlock] = self.req_to_blocks[request_id]
+            if request_id in self._req_to_last_computed:
```
heheda12345: pls note these corner cases when writing unit tests:
1. this seems fine
assume block_size=4
last step, we have
`block0=[*,*,A, sps1], block1=[sps2]`.
accept both sps1 and sps2
so we should free block0 here
2. this is not fine
last step, we have
`block0=[*,A, sps1,sps2], block1=[sps3]`
the running state is saved on block1
if accept no tokens, last_computed_tokens is on block0 and we can't free it

I think we should handle the two cases of remove skipped blocks seperately:
1. like case 1. A block is no longer needed as we've finished it's computation and future steps doesn't need it. It is similar to sliding window with window_size 1. We should find these blocks based on the num_computed_tokens of this step.
2. mamba prefill, we may have [*,*,*,A] NULL NULL [***B] in the prev step and need to free block [*,*,*,A]. we should find these blocks based on where the running state was saved.

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -662,28 +726,113 @@ def get_num_blocks_to_allocate(
...
+                    num_new_alloc_blocks = 1
+                    new_blocks.extend([self._null_block for _ in range(num_new_blocks - num_new_alloc_blocks)])
+                    if self.num_speculative_blocks > 0:
+                        # step i:            [0, 0, 0, a, sps_0, sps_1]
+                        # step i+1(i.e. j):  [0, 0, 0, a, 0, 0, 0, b, sps_0, sps_1]
+                        # reuse blocks sps_0 and sps_1
+                        # new_alloc_blocks: b
+                        # new_blocks: [0, 0, 0, b, sps_0, sps_1]
+                        # TODO: reuse blocks. if we need clean? especially in decode
+                        reuse_blocks = req_blocks[-self.num_speculative_blocks:]
```
heheda12345: not true. model runner is append only so we can't change the block_id of existing blocks unless we choose to change it to a null block.
```
# step i:            [0, 0, 0, a, sps_0, sps_1]
# step i+1(i.e. j):  [0, 0, 0, a,     0,b=sps_1, sps_0, sps1']
```

vllm/v1/core/single_type_kv_cache_manager.py
peakcrosser7: I think the current logic is able to handle both cases. The reason is that the  `num_computed_tokens` passed into `remove_skipped_blocks()` represents the actual number of tokens that have been computed. For the second case, the `target_idx` derived from `num_computed_tokens` is less than `block0`, so `block0` will not be freed.

vllm/v1/core/single_type_kv_cache_manager.py
peakcrosser7: yes, we should consider this

vllm/v1/core/sched/scheduler.py
peakcrosser7: the last prefill chunk will include extra spec tokens, causing the calculated mamba state to be incorrect.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -845,6 +846,22 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> None:
...
+                            self.kv_cache_config.kv_cache_groups
+                        ):
+                            kv_cache_spec = kv_cache_group.kv_cache_spec
+                            # NOTE(hhy): pop the last num_speculative_blocks (reuse blocks).
+                            #            maybe have better way to help copy blocks?
+                            if (isinstance(kv_cache_spec, MambaSpec)
+                                and new_block_ids[kv_cache_gid]
+                            ):
+                                block_ids = req_state.block_ids[kv_cache_gid]
+                                del block_ids[-kv_cache_spec.num_speculative_blocks:]
```
heheda12345: I think we shouldn't break the append-only assumption.

vllm/envs.py
```diff
@@ -1691,6 +1694,7 @@ def compile_factors() -> dict[str, object]:
         "LOCAL_RANK",
         "CUDA_VISIBLE_DEVICES",
         "NO_COLOR",
+        "VLLM_USE_LIGHTER_MAMBA_CACHE",
```
heheda12345, line 1698: use vllm_config.cache_config

vllm/model_executor/layers/mamba/abstract.py
```diff
@@ -48,7 +49,12 @@ def get_kv_cache_spec(self, vllm_config: VllmConfig) -> KVCacheSpec | None:
             raise NotImplementedError(
                 "Mamba with speculative decoding is not supported yet."
             )
-        mamba_block_size = vllm_config.cache_config.mamba_block_size
+        if not envs.VLLM_USE_LIGHTER_MAMBA_CACHE:
```
heheda12345: set correct mamba block size in vllm config resolution

vllm/model_executor/layers/mamba/abstract.py
```diff
@@ -61,6 +67,7 @@ def get_kv_cache_spec(self, vllm_config: VllmConfig) -> KVCacheSpec | None:
                 if vllm_config.speculative_config
                 else 0
             ),
+            enable_caching=vllm_config.cache_config.enable_prefix_caching,
```
heheda12345: remove this. it is not required.

vllm/engine/arg_utils.py
```diff
@@ -1356,6 +1356,13 @@ def create_engine_config(
             f"dcp_size={self.decode_context_parallel_size}."
         )
 
+        if (envs.VLLM_USE_LIGHTER_MAMBA_CACHE
+            and self.enable_prefix_caching 
+            and model_config.is_hybrid
+        ):
+            assert self.enable_chunked_prefill, (
+                "Prefix caching for hybrid models requires chunked prefill.")
```
heheda12345: update comment: only VLLM_USE_LIGHTER_MAMBA_CACHE needs chunked prefill.

vllm/model_executor/layers/mamba/mamba_mixer.py
```diff
@@ -240,7 +241,10 @@ def forward_cuda(self, hidden_states: torch.Tensor, output: torch.Tensor):
 
         assert self.cache_config is not None
         mamba_block_size = self.cache_config.mamba_block_size
-        prefix_caching_enabled = self.cache_config.enable_prefix_caching
+        prefix_caching_enabled = (
```
heheda12345: change prefix_caching_enabled to a proper name

vllm/model_executor/models/config.py
```diff
@@ -385,11 +393,16 @@ def verify_and_update_config(cls, vllm_config: "VllmConfig") -> None:
...
+            block_size = cache_config.block_size
+        else:
+            block_size = model_config.max_model_len
         # get mamba page size
         mamba_page_size = MambaSpec(
             shapes=model_cls.get_mamba_state_shape_from_config(vllm_config),
             dtypes=model_cls.get_mamba_state_dtype_from_config(vllm_config),
-            block_size=model_config.max_model_len,
+            block_size=block_size,
+            enable_caching=cache_config.enable_prefix_caching,
```
heheda12345: remove enable_caching from mamba_spec

vllm/model_executor/models/config.py
heheda12345, line 431: TODO: this function needs update (unrelated to this PR)

heheda12345, line 434: we need to adjust attention block_size for hybrid models.

vllm/model_executor/models/qwen3_next.py
```diff
@@ -1092,6 +1093,8 @@ def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
                         name.endswith(".bias") or name.endswith("_bias")
                     ) and name not in params_dict:
                         continue
+                    if name not in params_dict:
+                        continue
```
heheda12345, line 1096: todo: revert after https://github.com/vllm-project/vllm/pull/30433

vllm/v1/core/sched/scheduler.py
```diff
@@ -212,8 +212,71 @@ def __init__(
         )
         self.use_pp = self.parallel_config.pipeline_parallel_size > 1
         self.use_v2_model_runner = envs.VLLM_USE_V2_MODEL_RUNNER
+        print(
+            f">>> [DEBUG] Scheduler: init enable_prefix_caching={self.cache_config.enable_prefix_caching} block_size={self.block_size} kv_cache_config={self.kv_cache_config}"
+        )
+
+    def _has_mamba_spec(self) -> bool:
```
heheda12345:
```suggestion
    def has_mamba_layer(self) -> bool:
```
and move it to an attribute of scheduler

vllm/v1/core/sched/scheduler.py
```diff
@@ -212,8 +212,71 @@ def __init__(
...
+        has_mamba: bool = any(
+            isinstance(spec.kv_cache_spec, MambaSpec)
+            for spec in self.kv_cache_config.kv_cache_groups
+        )
+        assert not has_mamba or self.vllm_config.model_config.is_hybrid
+        return has_mamba
+
+    def _mamba_block_aligned_split(
```
heheda12345, line 231: change to:
```
if self.has_mamba_layer and self.cache_config.mamba_cache_mode == "align":
    num_new_tokens = self._mamba_block_aligned_split()
```

vllm/v1/core/sched/scheduler.py
heheda12345: split into a new PR and make it clear that this hack is only for spec decode + mamba?

vllm/v1/core/sched/scheduler.py
```diff
@@ -559,6 +659,16 @@ def schedule(self) -> SchedulerOutput:
                             # The request cannot be scheduled.
                             break
 
+                if envs.VLLM_USE_LIGHTER_MAMBA_CACHE:
```
heheda12345: update if

vllm/v1/core/block_pool.py
```diff
@@ -254,6 +296,10 @@ def cache_full_blocks(
             [] if self.enable_kv_cache_events else None
         )
         for i, blk in enumerate(new_full_blocks):
+            # NOTE: for mamba, full blocks includes the null block
+            #       and the null block should be skipped
```
heheda12345, line 258: null block is general and one case is mamba

vllm/v1/core/block_pool.py
```diff
@@ -270,6 +316,7 @@ def cache_full_blocks(
             if num_cached_blocks == 0:
                 parent_block_hash: ExternalBlockHash | None = None
             else:
+                # TODO(hhy): when LPS is enabled, parent_block maybe a null block
```
heheda12345: TODO: make the comment more clean

vllm/v1/core/kv_cache_manager.py
```diff
@@ -272,8 +273,9 @@ def allocate_slots(
         # The number of computed tokens is the number of computed tokens plus
         # the new prefix caching hits
         num_computed_tokens = request.num_computed_tokens + num_new_computed_tokens
+        num_tokens_target_model = num_computed_tokens + num_new_tokens
```
heheda12345: add comment.
"main model" may be a better name?
do we need min(max_model_len) here?

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -20,6 +21,30 @@
 from vllm.v1.request import Request
 
 
+def format_blocks(blocks: list[KVCacheBlock]):
```
heheda12345, line 24: remove these code

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -698,8 +749,31 @@ def find_longest_cache_hit(
...
 
+    def get_num_skipped_tokens(self, num_computed_tokens: int) -> int:
+        # TODO: merge https://github.com/vllm-project/vllm/pull/28047 first
+        return num_computed_tokens - 1
+
+    def remove_skipped_blocks(self, request_id: str, num_computed_tokens: int) -> None:
+        assert isinstance(self.kv_cache_spec, MambaSpec)
+        super().remove_skipped_blocks(request_id, num_computed_tokens)
+        if envs.VLLM_USE_LIGHTER_MAMBA_CACHE:
+            last_state_block_idx = self.last_state_block_idx.get(request_id)
```
heheda12345, line 753: add comment here

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -711,31 +785,151 @@ def get_num_blocks_to_allocate(
...
-        # Allocate extra `num_speculative_blocks` blocks for
-        # speculative decoding (MTP/EAGLE) with linear attention.
         assert isinstance(self.kv_cache_spec, MambaSpec)
-        if self.kv_cache_spec.num_speculative_blocks > 0:
-            num_tokens += (
-                self.kv_cache_spec.block_size
-                * self.kv_cache_spec.num_speculative_blocks
+        # mamba layers only exist in target model.
+        num_tokens = num_tokens_target_model
+        if not envs.VLLM_USE_LIGHTER_MAMBA_CACHE:
```
heheda12345: change to mamba_cache_mode=="xxx"

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -711,31 +785,151 @@ def get_num_blocks_to_allocate(
...
             )
-        return super().get_num_blocks_to_allocate(
-            request_id, num_tokens, new_computed_blocks
+            num_new_blocks = (
+                num_required_blocks
+                - len(new_computed_blocks)
+                - len(self.req_to_blocks[request_id])
+            )
+            if num_new_blocks > 0:
+                # (Chen): This may be possible. (block_size 4, 2 sps).
```
heheda12345: need better comment

vllm/v1/engine/core_client.py
```diff
@@ -269,6 +269,7 @@ def __init__(self, *args, **kwargs):
 
     def get_output(self) -> EngineCoreOutputs:
         outputs, _ = self.engine_core.step_fn()
+        self.engine_core.post_step(model_executed=True)
```
heheda12345, line 272: todo: merge https://github.com/vllm-project/vllm/pull/30319 first

vllm/v1/worker/block_table.py
```diff
@@ -113,6 +113,19 @@ def append_row(
         start = self.num_blocks_per_row[row_idx]
         self.num_blocks_per_row[row_idx] += num_blocks
         self.block_table.np[row_idx, start : start + num_blocks] = block_ids
+    
+    def pop_row(self, num_blocks: int, row_idx: int):
```
heheda12345: remove this

vllm/v1/worker/block_table.py
```diff
@@ -308,6 +325,10 @@ def append_row(self, block_ids: tuple[list[int], ...], row_idx: int) -> None:
         for i, block_table in enumerate(self.block_tables):
             block_table.append_row(block_ids[i], row_idx)
 
+    def pop_row(self, num_blocks: tuple[int, ...], row_idx: int) -> None:
```
heheda12345: remove this

vllm/v1/worker/block_table.py
```diff
@@ -291,6 +304,10 @@ def __init__(
             BlockTable(
                 block_size,
                 max_num_reqs,
+                # TODO: when prefix-caching and sps are both enable for 
+                #       mamba hybrid model, it will need 
+                #       `cdiv(max_model_len, block_size * total_cp_world_size) + num_speculative_tokens`
+                #       blocks for mamba groups
```
heheda12345, line 297: todo: need to address this

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1587,6 +1600,13 @@ def _get_block_table_and_slot_mapping(kv_cache_gid: int):
             slot_mapping[num_tokens:num_tokens_padded].fill_(-1)
             blk_table_tensor[num_reqs:num_reqs_padded].fill_(-1)
 
+            # if (envs.VLLM_USE_LIGHTER_MAMBA_CACHE
```
heheda12345: todo: remove

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2572,7 +2592,10 @@ def _sample(
             logits,
             sampling_metadata,
         )
-        self._update_states_after_model_execute(sampler_output.sampled_token_ids)
+        if is_global_first_rank():
```
heheda12345: wrong merge?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2902,6 +2925,162 @@ def _register_layerwise_nvtx_hooks(self) -> None:
                 pyt_hooks.register_hooks(self.model, self.model.__class__.__name__)
                 self.layerwise_nvtx_hooks_registered = True
 
+    def _mamba_copy_block(
```
heheda12345, line 2915: move these function to a utils_mamba.py?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -3030,6 +3209,13 @@ def execute_model(
 
                 pad_attn = cudagraph_mode == CUDAGraphMode.FULL
 
+                if (
```
heheda12345: todo: use a better condition

vllm/model_executor/models/config.py
heheda12345: chunk size is 64 for gdn

vllm/v1/core/block_pool.py
```diff
@@ -270,6 +316,7 @@ def cache_full_blocks(
...
                 parent_block = blocks[num_cached_blocks - 1]
                 assert parent_block.block_hash is not None
                 parent_block_hash = maybe_convert_block_hash(
```
heheda12345, line 279: fixing parent block may be null in https://github.com/vllm-project/vllm/pull/30544

examples/offline_inference/run.py
```diff
@@ -0,0 +1,56 @@
+# SPDX-License-Identifier: Apache-2.0
```
heheda12345, line 1: todo: remove this file

vllm/v1/core/sched/scheduler.py
```diff
@@ -213,6 +214,66 @@ def __init__(
...
+            and self.cache_config.mamba_cache_mode == "align"
+        )
+
+    def _mamba_block_aligned_split(
+        self,
+        request: Request,
+        num_new_tokens: int,
+        num_new_local_computed_tokens: int = 0,
+        num_external_computed_tokens: int = 0,
+    ) -> int:
```
heheda12345: @peakcrosser7 can you add some docstring to explain this function?

vllm/v1/core/sched/scheduler.py
```diff
@@ -213,6 +214,76 @@ def __init__(
...
+        assert num_external_computed_tokens == 0, (
+            "External KV connector is not verified yet"
+        )
+        if request.num_output_tokens == 0:  # prefill
+            # Ensure new tokens for a request in the prefill phase do not contain
+            # sps tokens, especially in the last prefill chunk. For a hybrid-model,
+            # extra sps tokens would corrupt the generated Mamba state.
+            # TODO: This logic does not yet handle resumed requests.
+            if request.num_computed_tokens < request.num_prompt_tokens:
```
heheda12345: I've moved this adjustment here.
Can you help to
1. add more comment to explain the problem
2. verify resumed requests

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -63,11 +89,16 @@ def __init__(
         self.kv_cache_group_id = kv_cache_group_id
         self._null_block = block_pool.null_block
 
+    def print(self, *args, **kwargs):
```
heheda12345: todo: remove this code

peakcrosser7:
@joennlae Thanks for following along! I'm focusing on mamba prefix caching optimizations and generalization this week. That will be followed by testing and bug fixes. I'm aiming to have this PR mostly complete by the end of the month. And #30877  is the cleaned-up version of this PR.

joennlae:
@peakcrosser7 That is fantastic :-) I had yoour last version running but had some issues with guided generation. I will try out the new PR just now.

PR #29278, opened 2025-11-23
[Core] Generalize Encoder Decoder arch to support variable encoder length beyond Whisper

ekagra-ranjan:
Encoder Decoder in vLLM is limited to Whisper style model which has fixed encoder len of 1500. 
This PR generalizes the Encoder decoder arch in vLLM to support variable len encoder inputs.

### Test
* WER: 
```
time pytest -sv tests/entrypoints/openai/correctness/test_transcription_api_correctness.py::test_wer_correctness
```

* offline inference: 
```
VLLM_USE_V1=1 VLLM_WORKER_MULTIPROC_METHOD='spawn' python3 examples/offline_inference/audio_language.py -m whisper
```
Output: `The first words I spoke in the original phonograph, a little piece of practical poetry. Mary had a little lamb, its streets were quite as slow, and everywhere that Mary went the lamb was sure to go.`

### DIFF
--- a/vllm/v1/attention/backends/utils.py
+++ b/vllm/v1/attention/backends/utils.py
@@ -92,6 +92,7 @@ class CommonAttentionMetadata:
     encoder_seq_lens: torch.Tensor | None = None
     encoder_seq_lens_cpu: np.ndarray | None = None
     max_encoder_seq_len: int | None = None
+    scheduled_encoder_req_index: list[int] | None = None
 
     dcp_local_seq_lens: torch.Tensor | None = None
     dcp_local_seq_lens_cpu: torch.Tensor | None = None
--- a/vllm/v1/attention/backends/utils.py
+++ b/vllm/v1/attention/backends/utils.py
@@ -89,7 +89,9 @@ class CommonAttentionMetadata:
     num_logits_indices: int | None = None
 
     # Needed by CrossAttentionBuilder
-    encoder_seq_lens: np.ndarray | None = None
+    encoder_seq_lens: torch.Tensor | None = None
+    encoder_seq_lens_cpu: np.ndarray | None = None
+    max_encoder_seq_len: int | None = None
 
     dcp_local_seq_lens: torch.Tensor | None = None
     dcp_local_seq_lens_cpu: torch.Tensor | None = None
--- a/vllm/v1/worker/gpu_model_runner.py
+++ b/vllm/v1/worker/gpu_model_runner.py
@@ -1188,13 +1188,6 @@ class GPUModelRunner(
             )
             self.encoder_seq_lens.np[req_index] = encoder_input_tokens
 
-        # for req_id, req_index in self.input_batch.req_id_to_index.items():
-        #     req_state = self.requests[req_id]
-        #     encoder_input_tokens = sum(
-        #         feature.mm_position.length for feature in req_state.mm_features
-        #     )
-        #     self.encoder_seq_lens.np[req_index] = encoder_input_tokens
-
         self.encoder_seq_lens.copy_to_gpu(num_reqs)
         encoder_seq_lens = self.encoder_seq_lens.gpu[:num_reqs]
         encoder_seq_lens_cpu = self.encoder_seq_lens.cpu[:num_reqs]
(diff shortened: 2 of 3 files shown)


DarkLight1337 (commented):
Looks like a duplicate of https://github.com/vllm-project/vllm/pull/29268?

ekagra-ranjan:
@NickLucche @DarkLight1337 - oh dang! The #29268 is tacking the same problem as this one. I actually worked on this PR a week ago on v11 branch but got time to upstreamed it just today along with some other PRs (#29282, https://github.com/vllm-project/vllm/pull/29279). What a coincidence that we were on the same thing at the same time! Sorry, I could've avoided this only if I upstreamed a little earlier.

I looked at #29268 just now and the main difference between #29268 and this PR is that this PR is using reusable buffers which avoids [these](https://github.com/vllm-project/vllm/pull/29268/files#diff-16828fcf07bae6a5517ec8599248470dd9d17dd394a00f3ed030b1d6f3d552ecR104-R109) allocation/deallocation of new cpu and gpu variables along with a cpu<->gpu sync in every fwd pass. 

Another difference is that #29268 finds the req_id of the scheduled/active encoder req using 2 passes: one using [np.where](https://github.com/vllm-project/vllm/pull/29268/files#diff-16828fcf07bae6a5517ec8599248470dd9d17dd394a00f3ed030b1d6f3d552ecR98-R100) and later `np.nonzero` in `_get_cross_slot_mapping()`. This PR directly finds it in a single pass `_get_encoder_seq_lens()` and carries it to CrossAttention via metadata. 

@NickLucche @DarkLight1337 - Please let me know what is the best way forward that from here. Happy to clarify anything about my approach if helpful.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -467,6 +467,7 @@ def __init__(
             self.max_num_reqs + 1, dtype=torch.int32
         )
         self.seq_lens = self._make_buffer(self.max_num_reqs, dtype=torch.int32)
+        self.encoder_seq_lens = self._make_buffer(self.max_num_reqs, dtype=torch.int32)
```
NickLucche, line 470: nice

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1167,20 +1168,49 @@ def _prepare_input_ids(
...
+            self.input_batch.req_id_to_index[req_id]
+            for req_id in scheduled_encoder_inputs
+        ]
+
+        return (
+            encoder_seq_lens,
+            encoder_seq_lens_cpu,
+            max_encoder_seq_len,
+            scheduled_encoder_req_index,
+        )
```
NickLucche, line 1213: I think I prefer keeping interfaces simple here if possible given how bloated the model_runner is. We can easily recover all other metadatas.

vllm/v1/attention/backends/utils.py
```diff
@@ -89,7 +89,10 @@ class CommonAttentionMetadata:
     num_logits_indices: int | None = None
 
     # Needed by CrossAttentionBuilder
-    encoder_seq_lens: np.ndarray | None = None
+    encoder_seq_lens: torch.Tensor | None = None
+    encoder_seq_lens_cpu: np.ndarray | None = None
+    max_encoder_seq_len: int | None = None
+    scheduled_encoder_req_index: list[int] | None = None
```
NickLucche, line 95: I think adding these 2 params feels like over-optmization, which hurts readability/maintainability in the long run

NickLucche (commented):
Thanks a lot for the work @ekagra-ranjan !
No worries these kind of clashes happen more often than one may think, let's just join efforts to get this landed. The PRs are already quite similar in impl.

I like the make_buffer optimization, while I feel like some other are maybe a bit too over-optimized and may hurt readability/maintainability of the code: enc-dec support is important but not core, hence I would prioritize keeping changes contained when possible as enc-dec have only barely made it to v1 thanks to whisper. 

If that's ok with you, I would add a commit with some of your optimizations to my PR (with you as author) and you can check it out. In case you don't like it, you can do the same with some of my changes in this PR.

PR #29288, opened 2025-11-24
[BugFix] Fix eagle async scheduling cpu race

njhill:
Need to protect reused cpu buffer.

### DIFF
--- a/vllm/v1/spec_decode/eagle.py
+++ b/vllm/v1/spec_decode/eagle.py
@@ -41,6 +41,7 @@ from vllm.v1.sample.sampler import _SAMPLING_EPS
 from vllm.v1.spec_decode.metadata import SpecDecodeMetadata
 from vllm.v1.utils import CpuGpuBuffer
 from vllm.v1.worker.gpu_input_batch import CachedRequestState, InputBatch
+from vllm.v1.worker.utils import async_barrier
 
 logger = init_logger(__name__)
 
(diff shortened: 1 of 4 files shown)


eble-amd:
@njhill, is this effort likely to be resumed?

I just spent some time directing (and directing, and directing ...) an LLM agent to guard any slightly suspicious object hand-off or reuse with events.  It suggested a number of changes, and my one test case that was failing before the many changes is passing now.  The agent identified this PR as partly overlapping with the things it found, and I am pinging you so I can get an early signal to help me allocate my time as I vet its suggestions.

In case this is useful, here's the agent's summary of its suggestions:

```
• If all GPU computation kernels are guaranteed on one stream, then these are probably
  unnecessary:

  - _draft_token_ids_ready_event
  - wait on _draft_token_ids_ready_event in _prepare_input_ids()
  - wait on _draft_token_ids_ready_event in _copy_draft_token_ids_to_cpu()
  - prev_sampled_token_ids_ready_event
  - wait on prev_sampled_token_ids_ready_event in _prepare_input_ids()
  - valid_sampled_token_count_gpu_ready_event
  - wait on valid_sampled_token_count_gpu_ready_event before
    update_num_computed_tokens_for_batch_change()

  Still useful / necessary for async copies or reused buffers:

  - record_stream() may still be useful if tensors are consumed by non-default copy streams or
    allocator lifetime is ambiguous.
  - valid_sampled_token_count_copy_stream.wait_event(...) or equivalent copy-stream wait.
  - draft_token_ids_copy_stream.wait_event(...) or wait_stream(...).
  - async_output_copy_stream.wait_stream(...) existing code.
  - prepare_inputs_event / async barrier for reused pinned CPU H2D buffers.
  - _backup_next_token_ids_done_event for persistent pinned CPU/GPU staging reuse.
  - discard_request_mask_done_event only if EAGLE’s consumer can overlap with the next input-
    prep H2D/copy. If single compute stream plus no overlap before next input prep, it may be
    redundant; if input prep can run ahead while previous EAGLE work is still queued/running,
    keep it.
```

PR #29291, opened 2025-11-24
[Misc] Suppress log outputs when constructing the default vllm config.

noooop:
## Purpose

Address: https://github.com/vllm-project/vllm/pull/26623#discussion_r2552773217

## Test Plan

vllm serve BAAI/bge-base-en

## Test Result

main:

A default vllm_config will be constructed here, including the creation of a default SchedulerConfig, and the following info log will be output, which is confusing.

vllm serve BAAI/bge-base-en

```
INFO 11-24 12:44:05 [scheduler.py:207] Chunked prefill is enabled with max_num_batched_tokens=2048.  <- confusing
(APIServer pid=7732) INFO 11-24 12:44:05 [api_server.py:2056] vLLM API server version 0.1.dev11596+g30854783a.d20251124
(APIServer pid=7732) INFO 11-24 12:44:05 [utils.py:253] non-default args: {'model_tag': 'BAAI/bge-base-en', 'model': 'BAAI/bge-base-en'}
(APIServer pid=7732) INFO 11-24 12:44:08 [config.py:998] Found sentence-transformers tokenize configuration.(APIServer pid=7732) INFO 11-24 12:44:09 [config.py:886] Found sentence-transformers modules configuration.
(APIServer pid=7732) INFO 11-24 12:44:09 [config.py:917] Found pooling configuration.
(APIServer pid=7732) INFO 11-24 12:44:09 [model.py:871] Resolved `--runner auto` to `--runner pooling`. Pass the value explicitly to silence this message.
```

BAAI/bge-base-en does not support chunked prefill, so chunked prefill is not used by default. This output in the first line of the log is confusing

this pr:

without this INFO log

---
<details>

</details>


DarkLight1337 (commented):
I think we should only do this for the configs in `EngineArgs` specifically to avoid accidentally hiding other errors.

vllm/logger.py
```diff
@@ -212,6 +213,13 @@ def init_logger(name: str) -> _VllmLogger:
     return cast(_VllmLogger, logger)
 
 
+@contextmanager
+def suppress_logging():
+    logging.disable(logging.INFO)
```
DarkLight1337: We should extract the previous setting and then revert it (in case the user manually sets the logging level), instead of always setting `NOTSET` at the end

vllm/logger.py
```diff
@@ -212,6 +213,14 @@ def init_logger(name: str) -> _VllmLogger:
     return cast(_VllmLogger, logger)
 
 
+@contextmanager
+def suppress_logging(level=logging.INFO):
```
DarkLight1337: Add type annotation

DarkLight1337 (approved):
LGTM, but would like @hmellor to confirm this is ok as well

vllm/engine/arg_utils.py
```diff
@@ -247,11 +247,13 @@ def _compute_kwargs(cls: ConfigType) -> dict[str, dict[str, Any]]:
...
-                default = (
-                    default.default
-                    if default.default_factory is None
-                    else default.default_factory()
-                )
+                if default.default_factory is None:
+                    default = default.default
+                else:
+                    # A default vllm_config will be constructed here.
+                    # Info logs will be output, which is confusing.
```
hmellor: I don't want to waste CI, but I'd update this comment as it's not technically correct

```suggestion
                    # VllmConfig's Fields have default_factory set to config classes.
                    # These could emit logs on init, which would be confusing.
```

hmellor (approved):
LGTM, just a nit about the comment explaining why we do this

DarkLight1337:
Going to force merge after pre-commit passes

PR #29300, opened 2025-11-24
[Model Runner V2] Implement Single-step Eagle 1

WoosukKwon:
This PR adds support for Eagle 1 with `num_speculative_steps=1`, implemented without introducing any CPU–GPU synchronization. It also ensures draft-token sampling to respect the request’s temperature, in contrast to the V1 model runner, which only supports argmax sampling for drafts.

### DIFF
--- a/vllm/v1/worker/gpu/spec_decode/eagle.py
+++ b/vllm/v1/worker/gpu/spec_decode/eagle.py
@@ -1,0 +1,197 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+import torch
+import torch.nn as nn
+
+from vllm.config import VllmConfig
+from vllm.config.compilation import CUDAGraphMode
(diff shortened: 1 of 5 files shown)


vllm/v1/worker/gpu/spec_decode/eagle.py
```diff
@@ -0,0 +1,197 @@
...
+        else:
+            last_hidden_states, hidden_states = ret_hidden_states
+        sample_hidden_states = last_hidden_states[last_token_indices]
+        logits = self.model.compute_logits(sample_hidden_states)
+
+        num_reqs = input_batch.num_reqs
+        cu_num_logits = input_batch.cu_num_logits[:num_reqs]
+        temperature = sampling_metadata.temperature[cu_num_logits]
+        seed = sampling_metadata.seeds[cu_num_logits]
+        # NOTE(woosuk): We must add 1 to the positions to match the Gumbel noise
```
benchislett, line 110: What's this?

PR #29305, opened 2025-11-24
[NIXL] Use config to enable telemetry + NIXL version bump

NickLucche:
This PR enables telemetry by default programmatically (instead of manually setting env_var), addressing a minor TODO, now possible since the dependency version bump.


markmc (approved):
Looks great to me

For reference, ai-dynamo/nixl#764 is the NIXL PR that introduced this API

PR #29308, opened 2025-11-24
[CI] Add batched audios Whisper test

NickLucche:
Add a basic 2-audio batched tests as I realized this is not a case we're currently covering.


DarkLight1337 (approved):

DarkLight1337:
Yeah probably we can't run both servers at the same time. Let's split them

PR #29313, opened 2025-11-24
[BugFix] bad_words filtering ineffective when n > 1

GOavi101:
- Use processed sampling_params (with bad_words_token_ids) for child requests
- Ensures child requests inherit bad_words_token_ids when n > 1
- Fixes issue where bad_words filtering failed for multiple completions

Fixes: #28861

## Test Result

---
<details>

</details>


njhill:
Thanks for this @GOavi101! I think it would be cleaner to do the following:
- Add a `@property` `params` method to `EngineCoreRequest` which returns `SamplingParams | PoolingParams` (based on which is non-None)
- Just add a line right after the call to `self.processor.process_inputs(`:
```python
    # We must use the cloned params that may have been updated in process_inputs()
    params = request.params
 ```
 
 I think that should cover it?

File: vllm/v1/engine/llm_engine.py [hunk 1]
```diff
@@ -245,6 +247,8 @@ def add_request(
                 trace_headers,
                 priority,
             )
+            # Use cloned params that may have been updated in process_inputs()
+            params = request.params
```
njhill: Can move this outside of the if/else (don't need in both branches)

njhill (commented):
Thanks @GOavi101 just one small thing

njhill:
Actually @GOavi101 could you also make the same change here: https://github.com/vllm-project/vllm/blob/b8328b49fb9954575bd1d7b30b22bb626ee47624/vllm/v1/engine/llm_engine.py#L252 :pray:

vllm/v1/engine/async_llm.py
```diff
@@ -321,14 +321,17 @@ async def add_request(
...
 
+        # Use cloned params that may have been updated in process_inputs()
+        params = request.params
+
         if is_pooling or params.n == 1:
             await self._add_request(request, prompt_text, None, 0, queue)
             return queue
 
         # Get the updated SamplingParams from the request, which
         # were cloned/updated in processor.process_inputs above.
```
njhill: Best to remove these comments now, since it's moved above

njhill (commented):
Thanks @GOavi101 just one more minor thing!

njhill (approved):
Thanks again @GOavi101

PR #29315, opened 2025-11-24
Support compressed-tensors W4A8 MoE checkpoints in GptOssModel weight loader for CPU

isharif168:
1. Add GptOssModel.load_per_expert_unfused_w4a8 helper to handle per-expert unfused MoE weights (gate_proj, up_proj, down_proj) in W4A8 checkpoints and map them into the fused FusedMoE layout (w13_* and w2_* parameters).
	•	Handles .weight, .weight_scale, .bias, and .input_scale suffixes.
	•	For biases, manually slices and writes into the appropriate columns of w13_bias (gate vs up) and w2_bias, supporting both 1D and 2D parameter layouts and using expert_id to pick the correct expert slice when the source tensor has an extra expert dimension.
	•	For weights/scales, delegates to a custom weight_loader when present, falling back to default_weight_loader otherwise, and surfaces whether the mapping was successfully handled.

2. Extend _load_weights_other to:
	•	Detect W4A8 (int4 weights, int8 activations) from self.config.quantization_config.config_groups["group_0"] and gate the new MoE path on is_w4a8.
	•	Precompute expert_params_mapping via FusedMoE.make_expert_params_mapping(...) for the MoE gate/up/down projections.
	•	For W4A8 models, first attempt load_per_expert_unfused_w4a8 for each weight tensor; if handled, mark the target param as loaded and skip the rest of the logic for that tensor. This allows loading checkpoints that store MoE weights per expert in an unfused form while still using FusedMoE fused storage internally.

3. Harden the generic stacking/renaming path in _load_weights_other:
	•	After substituting weight_name with param_name, skip if the resulting name is not in params_dict to avoid key errors on non-matching shards.
	•	Ensure loaded_params.add(name) is only called when a parameter is actually found and loaded (both in the stacked-path and the final “other weights” path).

4. Update GptOssForCausalLM metadata and mapping to match the new weight layout:
	•	Expand packed_modules_mapping to include:
	•	"qkv": ["q_proj", "k_proj", "v_proj"] for packed attention projections.
	•	"gate_up_proj": ["gate_proj", "up_proj"] for packed MoE gate+up projections used by compressed-tensors.
	•	Extend hf_to_vllm_mapper mappings:
	•	Map .self_attn. → .attn. (existing behavior).
	•	Map .qkv. → .qkv_proj. to align HF checkpoints that use a qkv naming convention with vLLM’s QKVParallelLinear.
	•	Map .mlp.experts.experts. → .mlp.experts. to flatten HF’s expert naming into the fused MoE layout.
	•	Add suffix mappings for MoE/compressed-tensors artifacts:
	•	Gate+up and down projection blocks/scales (MXFP4 and other formats) to w13_* and w2_*.
	•	Bias variants (.gate_up_proj_bias, .down_proj_bias) to w13_bias and w2_bias.

## Purpose
To use dynamic_int4_moe for gpt_oss model

## Test Plan
Tested with GPTOSS model quantized with llmcompresser

## Test Result
GPTOSS model passed

---
<details>

</details>


vllm/model_executor/models/gpt_oss.py
```diff
@@ -525,11 +644,36 @@ def _load_weights_other(
         tp_rank_start = tp_rank * per_rank_intermediate_size
         tp_rank_end = min((tp_rank + 1) * per_rank_intermediate_size, intermediate_size)
 
+        # W4A8 detection (int4 weights, int8 activations)
+        qc = getattr(self.config, "quantization_config", None)
+        group0 = (qc or {}).get("config_groups", {}).get("group_0", {})
+        w = group0.get("weights") or {}
+        ia = group0.get("input_activations") or {}
+        is_w4a8 = (w.get("num_bits") == 4) and (ia.get("num_bits") == 8)
```
nikhil-arm, line 652: We have an api to check this here: https://github.com/vllm-project/vllm/blob/70d5953f820ec528e2b6050a7969130009410d1e/vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py#L382

Will it be a good idea to re-use this api?

vllm/model_executor/models/gpt_oss.py
```diff
@@ -525,11 +644,36 @@ def _load_weights_other(
         tp_rank_start = tp_rank * per_rank_intermediate_size
         tp_rank_end = min((tp_rank + 1) * per_rank_intermediate_size, intermediate_size)
 
+        # W4A8 detection (int4 weights, int8 activations)
```
fadara01, line 647: generally speaking, can we apply the logic that this PR applies in `process_weights_after_loading` instead of needing to change the modeling file?, if not, then why not?

fadara01:
Thanks for your PR @isharif168 
The description currently assumes that the reviewer understands why the points you mention above are needed, could you please elaborate more on each of the points?
This also seems to depend on other PRs in llm_compressor, could you link this here? and generally explain how they're related and why these changes are needed.

vllm/model_executor/models/gpt_oss.py
```diff
@@ -495,6 +495,125 @@ def _load_weights_mxfp4(
             loaded_params.add(name)
         return loaded_params
 
+    def load_per_expert_unfused_w4a8(
```
fadara01, line 498: it would be nice for us not to specialize these functions for specific quantization schemes.
why can't the `w4a8` be an argument for this function, instead of baking it into the name/impl?

vllm/model_executor/models/gpt_oss.py
isharif168: > generally speaking, can we apply the logic that this PR applies in `process_weights_after_loading` instead of needing to change the modeling file?, if not, then why not?

No we cannot do in `process_weights_after_loading` because we need to first load the weights into w13_weights tensors and then it can be processed later, here we are getting all gate, up and down tensors as separate and then we are fusing them and loading the correct weights.

vllm/model_executor/models/gpt_oss.py
isharif168: > it would be nice for us not to specialize these functions for specific quantization schemes. why can't the `w4a8` be an argument for this function, instead of baking it into the name/impl?

I think its good to have different weight specific function and then we can call those specific function from load_weights_other, since all other combinations can have different weight loading scheme, we can do that if we have any in future

isharif168:
> Thanks for your PR @isharif168 The description currently assumes that the reviewer understands why the points you mention above are needed, could you please elaborate more on each of the points? This also seems to depend on other PRs in llm_compressor, could you link this here? and generally explain how they're related and why these changes are needed.

yes will make it more descriptive

vllm/model_executor/models/gpt_oss.py
isharif168: Good suggestion , but no we cant use that since it’s private and tied to CompressedTensorsConfig
Here we are just reading the flags from the config file and making the decision, to keep it simpler

fadara01 (approved):
Thanks for your PR :) 
Not very well-versed in this path, but the changes LGTM assuming everything passes for int4 and bf16

yewentao256 (commented):
Is this PR still needed? If so please also add metrics for E2E acc using `lm_eval...` and performance using `vllm bench...`

PR #29319, opened 2025-11-24
[BugFix] Fix initialization of draft model. 

halyavin:
This initialization is needed to make MTP for DeepSeek V3 work with high-throughput backend.

DeepSeek V3 draft model also has a MoE layer.  The method `DeviceCommunicationBase.prepare_communication_buffer_for_model` calls `FusedMoEMethodBase.init_prepare_finalize` method which in turn sets `fused_experts` field of the layer. During MTP calculation without this field `FusedMoE.forward_impl` method sees that `using_modular_kernel` property is false and sets `do_naive_dispatch_combine` flag and as a consequence calls `get_ep_group().dispatch()`. But `dispatch` method is not implemented in `DeepEPHTAll2AllManager` class which throws an exception.

Calling `prepare_communication_buffer_for_model` on a draft model makes this exception go away and makes MTP working.

### DIFF
--- a/vllm/v1/worker/gpu_model_runner.py
+++ b/vllm/v1/worker/gpu_model_runner.py
@@ -3345,7 +3345,9 @@ class GPUModelRunner(
             scope="local",
         )
         prepare_communication_buffer_for_model(self.model)
-        if (drafter := getattr(self, "drafter", None)) and (drafter_model := getattr(drafter, 'model', None)):
+        if (drafter := getattr(self, "drafter", None)) and (
+            drafter_model := getattr(drafter, "model", None)
+        ):
             prepare_communication_buffer_for_model(drafter_model)
         mm_config = self.model_config.multimodal_config
         self.is_multimodal_pruning_enabled = (
--- a/vllm/v1/worker/gpu_model_runner.py
+++ b/vllm/v1/worker/gpu_model_runner.py
@@ -3345,6 +3345,8 @@ class GPUModelRunner(
             scope="local",
         )
         prepare_communication_buffer_for_model(self.model)
+        if hasattr(self, "drafter") and hasattr(self.drafter, 'model'):
+            prepare_communication_buffer_for_model(self.drafter.model)
         mm_config = self.model_config.multimodal_config
         self.is_multimodal_pruning_enabled = (
             supports_multimodal_pruning(self.get_model())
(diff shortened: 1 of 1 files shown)


vllm/v1/worker/gpu_model_runner.py
```diff
@@ -3345,6 +3345,8 @@ def load_model(self, eep_scale_up: bool = False) -> None:
             scope="local",
         )
         prepare_communication_buffer_for_model(self.model)
+        if hasattr(self, "drafter") and hasattr(self.drafter, 'model'):
+            prepare_communication_buffer_for_model(self.drafter.model)
```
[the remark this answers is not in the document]
halyavin: I changed the code accordingly.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -3345,6 +3345,8 @@ def load_model(self, eep_scale_up: bool = False) -> None:
             scope="local",
         )
         prepare_communication_buffer_for_model(self.model)
+        if (drafter := getattr(self, "drafter", None)) and (drafter_model := getattr(drafter, 'model', None)):
+            prepare_communication_buffer_for_model(drafter_model)
```
tlrmchlsmth: I think we need a way to share the All2All state between `self.model` and the drafter -- this may duplicate state

tlrmchlsmth: @bnellnm @varun-sundar-rabindranath could you take a look?

tlrmchlsmth: From an offline conversation - the DeepEP buffers will be cached, so this won't involve any extra state for those All2All backends.  might not be the case for the FlashInfer All2Alls -- @pavanimajety would you know if this causes any overhead in that case?

I think we should go ahead and land this for now to get MTP working on main

varun-sundar-rabindranath: We cache the all2all handles here for deepep high-throughput here https://github.com/vllm-project/vllm/blob/0353d2e162cbda776d9dbfe026e65303204a7f1f/vllm/distributed/device_communicators/all2all.py#L370
 
This however, hashes on some model and DP/EP properties like hidden_size, num_local_experts and num_global_experts. This means that if the draft model's properties differ from the base-model, which is likely, we will create a new all2all handle. But this is necessary. 

I think this is good to land to unbreak `main` .

tlrmchlsmth (approved):

PR #29323, opened 2025-11-24
Scheduled removal of `CompilationConfig.use_inductor`

hmellor:
This field is scheduled for removal in v0.12.0, which will be the next release.


tests/utils_/test_argparse_utils.py
```diff
@@ -166,7 +166,6 @@ def test_dict_args(parser):
         "--hf-overrides.key2.key4",
         "val3",
         # Test compile config and compilation mode
-        "-O.use_inductor=true",
```
ProExpertProg, line 169: I think the idea was to test a nested bool, can you replace with `use_inductor_graph_partition` instead of removing?

ProExpertProg (approved):
Thanks!

yewentao256 (approved):
LGTM, thanks for the work!

tests/utils_/test_argparse_utils.py
hmellor: Good point, I'll do that

PR #29324, opened 2025-11-24
Scheduled removal of `ParallelConfig`'s direct child EPLB fields

hmellor:
These fields are scheduled for removal in v0.12.0, which will be the next release.

The config being removed now lives in `ParallelConfig.eplb_config`.


yewentao256 (approved):
LGTM, thanks for the work!

Issue #29317, opened 2025-11-24
[Bug]: triton 3.5 for gpt-oss fails on sm11.0a cu130


mcr-ksh:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
==============================
        System Info
==============================
OS                           : Ubuntu 24.04.3 LTS (aarch64)
GCC version                  : (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
Clang version                : Could not collect
CMake version                : version 4.1.0
Libc version                 : glibc-2.39

==============================
       PyTorch Info
==============================
PyTorch version              : 2.9.1+cu130
Is debug build               : False
CUDA used to build PyTorch   : 13.0
ROCM used to build PyTorch   : N/A

==============================
      Python Environment
==============================
Python version               : 3.12.11 | packaged by Anaconda, Inc. | (main, Jun  5 2025, 12:59:05) [GCC 11.2.0] (64-bit runtime)
Python platform              : Linux-6.8.12-tegra-aarch64-with-glibc2.39

==============================
       CUDA / GPU Info
==============================
Is CUDA available            : True
CUDA runtime version         : 13.0.88
CUDA_MODULE_LOADING set to   : 
GPU models and configuration : GPU 0: NVIDIA Thor
Nvidia driver version        : 580.00
cuDNN version                : Probably one of the following:
/usr/lib/aarch64-linux-gnu/libcudnn.so.9.16.0
/usr/lib/aarch64-linux-gnu/libcudnn_adv.so.9.16.0
/usr/lib/aarch64-linux-gnu/libcudnn_cnn.so.9.16.0
/usr/lib/aarch64-linux-gnu/libcudnn_engines_precompiled.so.9.16.0
/usr/lib/aarch64-linux-gnu/libcudnn_engines_runtime_compiled.so.9.16.0
/usr/lib/aarch64-linux-gnu/libcudnn_graph.so.9.16.0
/usr/lib/aarch64-linux-gnu/libcudnn_heuristic.so.9.16.0
/usr/lib/aarch64-linux-gnu/libcudnn_ops.so.9.16.0
HIP runtime version          : N/A
MIOpen runtime version       : N/A
Is XNNPACK available         : True

==============================
          CPU Info
==============================
Architecture:                         aarch64
CPU op-mode(s):                       64-bit
Byte Order:                           Little Endian
CPU(s):                               14
On-line CPU(s) list:                  0-13
[...]

PR #28788, opened 2025-11-15
[Build] Add OpenAI triton_kernels

varun-sundar-rabindranath:
## Purpose
This PR adds https://github.com/triton-lang/triton/tree/main/python/triton_kernels to vLLM. 
We can't install this package via pip. Please take a look at https://github.com/vllm-project/vllm/pull/27659 . As a result, this PR, injects the `triton_kernels` package directly into vLLM during build time, similar to the approach we take with `vllm_flash_attn`. Concretely, we just copy the entire `triton_kernels` folder in `<triton-root>/python/triton_kernels/triton_kernels` to `vllm/third_party/triton_kernels`during build time and add the module to the `sys.module["triton_kernel"]` during run-time. 

Fixes : https://github.com/vllm-project/vllm/issues/27672

## Test Plan
local-build testing on H100 : `python3 setup.py build_ext --inplace` / `uv pip install -vvv -e . --no-build-isolation`
package-build testing on H100 : `TORCH_CUDA_ARCH_LIST="9.0" python3 setup.py bdist_wheel --dist-dir=dist`

gpt-oss serve command: `vllm serve openai/gpt-oss-20b --tensor-parallel-size 2  --no-enable-prefix-caching  --port 9010 `
gpt-oss eval command : ` OPENAI_API_KEY=empty python -m gpt_oss.evals --model openai/gpt-oss-20b --eval gpqa --n-threads 128 --reasoning-effort low --base-url http://localhost:9010/v1 `

## Test Result
`local-build` : The build correctly copies `triton_kernels` into `vllm/third_party/triton_kernels`
`package-build` : The wheel when installed has `triton_kernels` in `<site-packages>/vllm/third_party/triton_kernels`
Both `local_build` and `package-build` obtains expected eval score of about `0.57`.

Need to perform further testing with CI wheels. 

Thanks @zyongye  @daniel-fahey for the insights.

PR #29339, opened 2025-11-24
[Bugfix] Only use triton_kernels for MXFP4 on SM90 and SM100

mgoin:
## Purpose

After we landed triton_kernels as a default dep in https://github.com/vllm-project/vllm/pull/28788, we started seeing some issues appear for gpt-oss users on sm110 and sm120 (see https://github.com/vllm-project/vllm/issues/29317 and https://vllm-dev.slack.com/archives/C0990U53QBV/p1764009918189409). 
This PR disables that path for now and uses Marlin for those GPUs.

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/layers/quantization/mxfp4.py
+++ b/vllm/model_executor/layers/quantization/mxfp4.py
@@ -132,10 +132,17 @@ def get_mxfp4_backend(with_lora_support: bool) -> Mxfp4Backend:
             )
 
         # If FlashInfer is not available, try either Marlin or Triton
+        triton_kernels_supported = (
+            has_triton_kernels()
+            and is_torch_equal_or_newer("2.8.0")
+            # NOTE: triton_kernels are only confirmed to work on SM90 and SM100
+            # SM110 fails with this error: https://github.com/vllm-project/vllm/issues/29317
+            # SM120 needs this fix: https://github.com/triton-lang/triton/pull/8498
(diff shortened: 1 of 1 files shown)


varun-sundar-rabindranath (approved):

vllm/model_executor/layers/quantization/mxfp4.py
```diff
@@ -132,12 +132,15 @@ def get_mxfp4_backend(with_lora_support: bool) -> Mxfp4Backend:
...
         # If FlashInfer is not available, try either Marlin or Triton
-        if (
-            envs.VLLM_MXFP4_USE_MARLIN
-            or current_platform.get_device_capability()[0] < 9
-            or not has_triton_kernels()
-            or not is_torch_equal_or_newer("2.8.0")
-        ):
+        triton_kernels_supported = (
+            has_triton_kernels()
+            and is_torch_equal_or_newer("2.8.0")
```
yewentao256, line 137: I think vllm requires torch 2.9.0 now, could we remove this?

yewentao256 (approved):
LGTM, thanks for the work!

vllm/model_executor/layers/quantization/mxfp4.py
mgoin: CUDA does, but other hardware does not. Even though this is in a CUDA branch, I think some people still build with their own versions of torch. So would like to keep for now

PR #29347, opened 2025-11-24
[Model Runner V2] Simplify Eagle bookkeeping with num_rejected

WoosukKwon:


vllm/v1/worker/gpu/spec_decode/rejection_sample.py
```diff
@@ -69,3 +69,15 @@ def rejection_sample(
...
     return sampled, num_sampled
+
+
+@torch.compile(dynamic=True)
+def get_num_rejected(
+    cu_num_logits: torch.Tensor,
+    num_sampled: torch.Tensor,
+) -> torch.Tensor:
+    num_logits = cu_num_logits[1:] - cu_num_logits[:-1]
+    num_rejected = num_logits - num_sampled
```
njhill, line 80: Why not this one in-place too? Or it doesn't make a difference for TC?

PR #29355, opened 2025-11-24
[BugFix] Fix duplicate req id tool-call race condition

njhill:
The front-end built-in tool calling code can make successive generate calls with the same top-level request id.

This can cause a problem with async scheduling since the prior request can still be present in the engine (in the case of stop/eos tokens for example, we generate one additional token).

This is a specific fix for the problematic "internal" request id duplication, but we need to fix things more generally. Our current duplicate id check is insufficient, especially when async scheduling is in use. Even if that's fixed to reject the successive requests in this case, it's still confusing for users who would expect that once a request has completed from their perspective, they should be able to make another one with the same id.

https://github.com/vllm-project/vllm/pull/27987 is related but I think we need something at the LLM/AsyncLLM level.

### DIFF
--- a/vllm/entrypoints/openai/serving_chat.py
+++ b/vllm/entrypoints/openai/serving_chat.py
@@ -273,6 +273,8 @@ class OpenAIServingChat(OpenAIServing):
         try:
             for i, engine_prompt in enumerate(engine_prompts):
                 prompt_text, _, _ = self._get_prompt_components(request_prompts[i])
+                # If we are creating sub requests for multiple prompts, ensure that they
+                # have unique request ids.
                 sub_request_id = (
                     request_id if len(engine_prompts) == 1 else f"{request_id}_{i}"
                 )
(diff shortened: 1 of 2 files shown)


vllm/entrypoints/openai/serving_chat.py
```diff
@@ -273,6 +273,9 @@ async def create_chat_completion(
         try:
             for i, engine_prompt in enumerate(engine_prompts):
                 prompt_text, _, _ = self._get_prompt_components(request_prompts[i])
+                sub_request_id = (
```
robertgshaw2-redhat, line 278: I think this warrants some more detailed comments

heheda12345 (approved):
LGTM! Agree with @robertgshaw2-redhat that we can add a bit more comments.

vllm/entrypoints/openai/serving_chat.py
njhill: Added

PR #29358, opened 2025-11-25
[ROCm][CI] Attempt to fix the failures under a subgroup of the e2e the test group

AndreasKaratzas:
This PR ensures that during `test_async_scheduling`, we utilize the `TRITON_ATTN` backend which is the default attention backend for ROCm.

Test used to verify functionality on ROCm: `pytest -v -s tests/v1/e2e/test_async_scheduling.py`


AndreasKaratzas:
I built the upstream dockers (both base and `.rocm`) and the issue you attached above did not appear on my MI325. However, accuracy issues appeared, and problems with bad word subgroup. So I refactored the solution. Let me know if you still get errors (I run the test 3 times before I upload it, passed all 3).

For the newly introduced modifications, here are some details:

### Platform-specific attention backends
- **ROCm:** `ROCM_AITER_FA` for non-spec-decoding, `TRITON_ATTN` for spec-decoding
- **Other platforms:** `FLEX_ATTENTION` (unchanged from original)

### Platform-specific dtype
- **ROCm:** `float16` for non-spec-decoding, `float32` for spec-decoding (TRITON_ATTN supports higher precision)
- **Other platforms:** `float32` always (original behavior)

### Relaxed tolerances for ROCm
- Logprob comparison uses `rel_tol=5e-2, abs_tol=1e-5` on ROCm since we are using fp16
- Other platforms use original strict tolerances `rel_tol=1e-3, abs_tol=1e-6`

### Reduced test configurations on ROCm
- Skip `chunk_prefill=True` configs (FP variance compounds across chunks)
- Only run `structured_outputs` tests (deterministic despite FP differences)

### Skip strict logprobs check for ROCm spec-decoding
- When `logprobs` are requested during spec-decoding tests on ROCm, skip strict comparison since values can differ slightly even when selected tokens match

I think I'm still respecting the purpose of this particular test with these new changes. I am open to feedback. Even for NVIDIA the only attention backend that works is `FLEX_ATTENTION`. So the peculiarities of this test as is render almost every attention backend incompatible with it.

requirements/rocm-test.txt
```diff
@@ -70,7 +70,7 @@ torchgeo==0.7.0
 mteb==2.1.2
 
 # Data processing
-xgrammar==0.1.27
+xgrammar @ git+https://github.com/divakar-amd/xgrammar@3272f7c520564858056a60480d5afdf69ae79c84
```
SageMoore, line 78: Is this intentional? Can you please explain why you are changing this to @divakar-amd's fork?

tests/v1/e2e/test_async_scheduling.py
```diff
@@ -231,6 +278,15 @@ def run_test(
...
     print(f"---- TESTING {test_str}: {test_config}")
     print("-" * 80)
+
+    # On ROCm: use float16 for first test (ROCM_AITER_FA), but float32 for
+    # spec decoding test (TRITON_ATTN) for better precision.
+    # On others: always use float32.
+    if current_platform.is_rocm() and not is_testing_with_spec_decoding:
+        dtype = "float16"
+    else:
+        dtype = "float32"
```
SageMoore, line 290: Why not use float32 for everything? Would you still have to update tolerances for rocm if everything was float32?

requirements/rocm-test.txt
AndreasKaratzas: Hi Sage, yes this is intentional. The problem is that in xgrammar there is a hard coded WARP SIZE: 
https://github.com/divakar-amd/xgrammar/commit/41a849f468a7091a4342dec5a156706302252799

So the pip package (or for that matter the upstream xgrammar repo) do not work correctly on ROCm. There is an open PR on xgrammar for this:
https://github.com/mlc-ai/xgrammar/pull/476

But as you can see this PR has been there for more than 2 weeks. So for the test to work for now, we are going with this solution. I am in contact with @divakar-amd and as soon as his PR gets merged, we are going to change the requirements to at least be in parity with CUDA.

tests/v1/e2e/test_async_scheduling.py
AndreasKaratzas: TL;DR
`ROCM_AITER_FA` which is the only one that can satisfy the that particular subgroup in the test for ROCm, does not support fp32.

I think that, even for NVIDIA, the only attention backend that is also accurate enough for this test is FLEX_ATTENTION. That's the only attention backend that is more numerically accurate than others. Unfortunately, FLEX_ATTENTION backend is not fully supported on ROCm yet. So all other backends are going to inject more numerical error on any platform. Other than that, our backends, especially the AITER ones, are tested on float16, bfloat16, fp8, and in some cases fp4. So for now and for ROCm we are going to go with the float16 precision. In the future we will introduce a more accurate backend to better serve fp32 workloads.

gshtras (approved):

SageMoore (approved):
This is a fine temporary solution, but let's push on https://github.com/mlc-ai/xgrammar/pull/476.

DarkLight1337 (approved):

PR #29381, opened 2025-11-25
[vLLM Benchmark Suite] Add default parameters section and update CPU benchmark cases

louie-tsai:
## Purpose

1. remove redundancy in vllm benchmark suite test cases by moving common arguments into default section (original format still works)
2. label each test with different tier. By default, only run base tier. only run full tier with "RUN_FULL_TESTS=1". 
3. cover all CPU supported models in full tier to make sure supported models work in new release
4. create a test mode, so benchmark suite only dump the run commands without running it. it helps users to understand the run BKMs quickly.

## Test Plan

manually tested on AWS R7i.96xlarge

## Test Result

WIP
---
<details>

</details>


.buildkite/performance-benchmarks/README.md
```diff
@@ -29,6 +29,8 @@ Runtime environment variables:
 - `THROUGHPUT_JSON`: JSON file to use for the throughout tests. Default value is empty string (use default file).
 - `REMOTE_HOST`: IP for the remote vLLM service to benchmark. Default value is empty string.
 - `REMOTE_PORT`: Port for the remote vLLM service to benchmark. Default value is empty string.
+- `TEST_MODE`: set the value to '1' to only dump run commands into result folder without benchmarking. Default value is empty string.
+- `RUN_FULL_TESTS`: set the value to '1' to run tests labeled as tier "full". Default value is empty string.
```
bigPYJ1151: Hi @louie-tsai I'm not sure whether we should add these two variables, as:
- For ```TEST_MODE```, I think all arguments are in JSON files, so looks like it is unnecessary.
- For ```RUN_FULL_TESTS```, I think all benchmark items in a JSON file should be executed. If we want multiple test suits, we can create different JSON file for it.

louie-tsai: @bigPYJ1151 
-  For TEST_MODE, the purpose of test mode is to generate exact run commands for test cases without running them, so users could quickly refer the run commands from benchmark suite.  e.g. one of the .commands file for a test case
image
   Therefore, we could ask user to run test mode to dump the up-to-date BKMs from benchmark suite.  I still got questions from users how to find BKMs from json file, so hopefully it can make it much eaiser.

- For RUN_FULL_TESTS, we host the full and base tier in one json file because we don't plan to host default section in two json file.  ideally, we only need to maintain one default section for one architecture to avoid duplication, so that is why I merge all CPU test cases into one json file under one default section.

bigPYJ1151: I see. I'm worried the script became more and more complex and hard to maintain. And we actually have more proper ways to achieve the goals:
- We can add an example about how to benchmark to the docs (like FAQ section in the CPU doc), just two commands. And users can refer detial arguments in the JSON files.
- We can maintain test cases with different tiers in different files (maybe is more clear compared with mix them in a single file) and merge them via ```jq add```, like:
```bash
jq -s 'add' file1.json file2.json > merged.json
```
How do you think about?

louie-tsai: @bigPYJ1151 

-  for FAQ section, I have added one section for getting the commands via TEST_MODE and how to refer to Perf dashboard.   [optimized runtime configs](https://github.com/intel-ai-tce/vllm/blob/test_mode_bench_suite/docs/getting_started/installation/cpu.md#how-to-find-optimized-runtime-configurations-for-supported-cpu-models)   good to know what your thoughts for it.  it might be easier to get the exact bash command for user than checking into json file.
- for merging two json files, do you expect that run script do it automatically when RUN_FULL_TEST=1 or do you expect users to do it manually?  I prefer not to ask users to manually merge it.

bigPYJ1151: I prefer to avoid introducing unnecessary complexity.
- For ```TEST_MODE```, I think it is unnecessary as the script is used for benchmark workflow rather than debugging or inspecting. Add some examples in the CPU doc and refer detail arguments in the JSON are enough for benchmark BKM.
- For ```RUN_FULL_TESTS```, I suppose you are using it in Pytorch CI? If so we can add a merge command to generate final JSON before benchmarking and avoid change the vLLM script. But why only execute partial of cases? For cases are not executed we should remove them.

louie-tsai: @bigPYJ1151 

-  For TEST_MODE, if some explanation is enough, we could remove that session in README. 
- For RUN_FULL_TESTS, we want to cover supported model in benchmark suite, so we could do a full test for each major releases on all models to make sure no perf regression. In the meantime, PyTorch CI need to finish the run in 1-2 hours, so we can't cover all models.  With all supported models in json, users can easily benchmark all supported models with right BKMs.

louie-tsai: @bigPYJ1151  remove test mode from cpu.md and README.md accordingly

bigPYJ1151: I still prefer to use ```jq add``` as it doesn't require to change the scripts:
- We can have 2 json ```serving-tests-cpu-base.json``` and ```serving-tests-cpu-extend.json```
- We can generate final test suit with:
```bash
cpu_test_suits=(serving-tests-cpu-base.json)
if $RUN_FULL_TESTS
  cpu_test_suits+=(serving-tests-cpu-extend.json)
jq -s "add" ${cpu_test_suits[@]} > serving-tests-cpu.json
```
Then we can have no change to the benchmark script.

.buildkite/performance-benchmarks/tests/serving-tests-cpu.json
```diff
@@ -1,276 +1,519 @@
...
+      "max_num_seqs": 256,
+      "load_format": "dummy"
+    },
+    "client_parameters": {
+      "model": "meta-llama/Llama-3.1-8B-Instruct",
+      "backend": "vllm",
+      "num_prompts": 200
+    }
+  },
+  "tests": [
```
bigPYJ1151, line 34: Looks like we just need this array. As other JSON files are all ```[{case 1}, {case 2},...]```

docs/getting_started/installation/cpu.md
```diff
@@ -125,6 +125,33 @@ vllm serve facebook/opt-125m --dtype=bfloat16
 
 Note, it is recommended to manually reserve 1 CPU for vLLM front-end process when `world_size == 1`.
 
+### What are supported models on CPU?
+
+For the full and up-to-date list of models validated on CPU platforms, please see the official documentation: [Supported Models on CPU](https://docs.vllm.ai/en/latest/models/hardware_supported_models/cpu)
+
+### How to Find Optimized Runtime Configurations for Supported CPU Models?
```
bigPYJ1151:
```suggestion
### How to find benchmark configuration examples for supported CPU models?
```
I think \`optimized\` depends on specific platforms.

bigPYJ1151: We can guide users find benchmark commands in https://docs.vllm.ai/en/latest/cli/?h=benchmark#bench

louie-tsai: those are general commands. we have more arguments and environment variables than what it has in the cli session

.buildkite/performance-benchmarks/tests/serving-tests-cpu.json
louie-tsai: the main point is to have default session, so we don't need to put same settings again and again in each test.

.buildkite/performance-benchmarks/README.md
louie-tsai: if we don't want to have tier implementation. I can make let all tests run in serving_test_cpu.json, but have less tests in CI  json file to finish CI in 1-2 hours.  however, all other platforms have one json file like serving-tests-hpu.json, so plan to also have one json file for CPU.

louie-tsai: @bigPYJ1151  
remove tier and RUN_FULL_TESTS implementations.
also reduce the new models test cases, so only 4 new test cases introduced.
hope it works for you.

.buildkite/performance-benchmarks/scripts/run-performance-benchmarks.sh
```diff
@@ -110,7 +110,11 @@ json2envs() {
 wait_for_server() {
   # wait for vllm server to start
   # return 1 if vllm server crashes
-  timeout 1200 bash -c '
+  local timeout_val="1200"
+  if [ "$TEST_MODE" == "1" ];then
```
bigPYJ1151: Please revert the related changes.

.buildkite/performance-benchmarks/README.md
```diff
@@ -108,6 +108,71 @@ The number of this test is less stable compared to the delay and latency benchma
 
 WARNING: The benchmarking script will save json results by itself, so please do not configure `--save-results` or other results-saving-related parameters in `serving-tests.json`.
 
+#### New JSON Format
```
bigPYJ1151: Maybe we can just name it as ```default parameters field```.

.buildkite/performance-benchmarks/README.md
```diff
@@ -108,6 +108,71 @@ The number of this test is less stable compared to the delay and latency benchma
...
+
+Serving tests now support an enhanced format that includes a default section.  
+Values defined in this section are applied globally to all tests unless explicitly overridden in a specific test case.  
+Each test may also specify a tier. Tests assigned to the "full" tier will only be executed when RUN_FULL_TESTS is set to 1.
```
bigPYJ1151:
```suggestion
We can specify default parameters in a JSON field with key `defaults`. Parameters defined in the field are applied globally to all serving tests, and can be overridden in test case fields. Here is an example:
```

.buildkite/performance-benchmarks/README.md
```diff
@@ -108,6 +108,71 @@ The number of this test is less stable compared to the delay and latency benchma
...
 WARNING: The benchmarking script will save json results by itself, so please do not configure `--save-results` or other results-saving-related parameters in `serving-tests.json`.
 
+#### New JSON Format
+
+Serving tests now support an enhanced format that includes a default section.  
+Values defined in this section are applied globally to all tests unless explicitly overridden in a specific test case.  
+Each test may also specify a tier. Tests assigned to the "full" tier will only be executed when RUN_FULL_TESTS is set to 1.  
+
+<details>
+<summary> An Example of New JSON Format </summary>
```
bigPYJ1151:
```suggestion
<summary> An Example of default parameters field </summary>
```

.buildkite/performance-benchmarks/README.md
```diff
@@ -108,6 +108,71 @@ The number of this test is less stable compared to the delay and latency benchma
...
+      "client_parameters": {
+        "model": "meta-llama/Llama-3.2-3B-Instruct",
+        "dataset_name": "random",
+        "random-input-len": 128,
+        "random-output-len": 128,
+        "ignore-eos": ""
+      }
+    }
+  ]
+}
```
bigPYJ1151, line 165:
```suggestion
{
  "defaults": {
    "qps_list": [
      "inf"
    ],
    "server_environment_variables": {
      "VLLM_ALLOW_LONG_MAX_MODEL_LEN": 1
    },
    "server_parameters": {
      "tensor_parallel_size": 1,
      "dtype": "bfloat16",
      "block_size": 128,
      "disable_log_stats": "",
      "load_format": "dummy"
    },
    "client_parameters": {
      "backend": "vllm",
      "dataset_name": "random",
      "random-input-len": 128,
      "random-output-len": 128,
      "num_prompts": 200,
      "ignore-eos": ""
    }
  },
  "tests": [
    {
      "test_name": "serving_llama3B_tp2_random_128_128",
      "server_parameters": {
        "model": "meta-llama/Llama-3.2-3B-Instruct",
        "tensor_parallel_size": 2,
      },
      "client_parameters": {
        "model": "meta-llama/Llama-3.2-3B-Instruct",
      }
    },
    {
      "test_name": "serving_qwen3_tp4_random_128_128",
      "server_parameters": {
        "model": "Qwen/Qwen3-14B",
        "tensor_parallel_size": 4,
      },
      "client_parameters": {
        "model": "Qwen/Qwen3-14B",
      }
    },
  ]
}
```

To simplify the example and make it more typical.

.buildkite/performance-benchmarks/scripts/run-performance-benchmarks.sh
```diff
@@ -316,12 +328,54 @@ run_throughput_tests() {
...
+  #          "test_name": "...",
+  #          "server_parameters": {...},
+  #          "client_parameters": {...},
+  #          "server_environment_variables": {...},
+  #          "qps_list": [...],
+  #          "max_concurrency_list": [...]
+  #        },
+  #        ...
+  #      ]
+  #    }
```
bigPYJ1151:
```suggestion
  # 1) Plain format: top-level array
  #    [ { "test_name": "...", "server_parameters": {...}, ... }, ... ]
  #
  # 2) Default parameters field + plain format tests
  #    {
  #      "defaults": { ... },
  #      "tests": [ { "test_name": "...", "server_parameters": {...}, ... }, ... ]
  #    }
```

.buildkite/performance-benchmarks/scripts/run-performance-benchmarks.sh
```diff
@@ -316,12 +328,54 @@ run_throughput_tests() {
...
 
   local serving_test_file
   serving_test_file=$1
 
   # Iterate over serving tests
-  jq -c '.[]' "$serving_test_file" | while read -r params; do
+  jq -c '
+    if type == "array" then
+      # Old format: each element is a test
```
bigPYJ1151:
```suggestion
      # Plain format: test cases array
```

.buildkite/performance-benchmarks/scripts/run-performance-benchmarks.sh
```diff
@@ -316,12 +328,54 @@ run_throughput_tests() {
...
+      .[]
+    elif (type == "object" and has("tests")) then
+      # New format: merge defaults into each test
```
bigPYJ1151:
```suggestion
      # merge the default parameters into each test cases
```

.buildkite/performance-benchmarks/scripts/run-performance-benchmarks.sh
```diff
@@ -335,20 +389,25 @@ run_serving_tests() {
       continue
     fi
 
-    # get client and server arguments
+    # get client and server arguments (after jq merge for new-format)
```
bigPYJ1151:
```suggestion
    # get client and server arguments (after merged the default parameters)
```

.buildkite/performance-benchmarks/scripts/run-performance-benchmarks.sh
```diff
@@ -450,7 +518,7 @@ run_serving_tests() {
     done
 
     # clean up
-    kill -9 $server_pid
+    kill -9 $server_pid 2>/dev/null || true
```
bigPYJ1151: Please remove the debug change

.buildkite/performance-benchmarks/tests/serving-tests-cpu.json
```diff
@@ -1,276 +1,277 @@
...
+      "VLLM_ALLOW_LONG_MAX_MODEL_LEN": 1,
+      "VLLM_ENGINE_ITERATION_TIMEOUT_S": 120,
+      "VLLM_CPU_SGL_KERNEL": 1,
+      "VLLM_CPU_KVCACHE_SPACE": 40
+    },
+    "server_parameters": {
+      "model": "meta-llama/Llama-3.1-8B-Instruct",
+      "tensor_parallel_size": 1,
+      "dtype": "bfloat16",
+      "distributed_executor_backend": "mp",
```
bigPYJ1151, line 18:
```suggestion
```
It is the unique supported backend and will be set by default.

.buildkite/performance-benchmarks/tests/serving-tests-cpu.json
```diff
@@ -1,276 +1,277 @@
...
+      "block_size": 128,
+      "trust_remote_code": "",
+      "disable_log_stats": "",
+      "enforce_eager": "",
```
bigPYJ1151, line 22: We shouldn't use ```--enforce-eager``` for most of models in benchmarks.

.buildkite/performance-benchmarks/tests/serving-tests-cpu.json
```diff
@@ -1,276 +1,277 @@
...
+      "test_name": "serving_llama8B_tp1_random_128_128",
+      "server_parameters": {
+        "tensor_parallel_size": 1,
+        "enable_chunked_prefill": ""
+      },
+      "client_parameters": {
+        "dataset_name": "random",
+        "random-input-len": 128,
+        "random-output-len": 128,
+        "ignore-eos": ""
```
bigPYJ1151: ```ignore-eos``` should be moved to the default parameters

.buildkite/performance-benchmarks/tests/serving-tests-cpu.json
```diff
@@ -1,276 +1,277 @@
...
+      "client_parameters": {
+        "dataset_name": "sharegpt",
+        "dataset_path": "./ShareGPT_V3_unfiltered_cleaned_split.json"
+      }
+    },
+    {
+      "test_name": "serving_llama8B_tp1_random_128_128",
+      "server_parameters": {
+        "tensor_parallel_size": 1,
+        "enable_chunked_prefill": ""
```
bigPYJ1151:
```suggestion
```
It is enabled by default.

.buildkite/performance-benchmarks/tests/serving-tests-cpu.json
```diff
@@ -1,276 +1,277 @@
-[
+{
+  "defaults": {
+    "qps_list": [
+      "inf"
+    ],
+    "max_concurrency_list": [12, 16, 24, 32, 64, 128, 200],
```
bigPYJ1151, line 6: I think we can reduce the concurrency cases to save benchmark time. I usually use ```[8,16,32,64]```. The latency may be too large and meaningless when the concurrency is too high.

.buildkite/performance-benchmarks/tests/serving-tests-cpu.json
```diff
@@ -1,276 +1,277 @@
...
+      "disable_log_stats": "",
+      "enforce_eager": "",
+      "max_num_batched_tokens": 2048,
+      "max_num_seqs": 256,
+      "load_format": "dummy"
+    },
+    "client_parameters": {
+      "model": "meta-llama/Llama-3.1-8B-Instruct",
+      "backend": "vllm",
+      "num_prompts": 200
```
bigPYJ1151, line 31: We can also reduce prompt num to save time. I usually set it as ```2 * concurrency```.

docs/getting_started/installation/cpu.md
```diff
@@ -125,6 +125,28 @@ vllm serve facebook/opt-125m --dtype=bfloat16
...
+To benchmark the supported models using these optimized settings, follow the steps in [running vLLM Benchmark Suite manually](https://docs.vllm.ai/en/latest/contributing/benchmarks/#manually-trigger-the-benchmark) and run the Benchmark Suite on a CPU environment.  
+
+Below is an example command to benchmark all CPU-supported models using optimized configurations.
+
+```bash
+ON_CPU=1 bash .buildkite/performance-benchmarks/scripts/run-performance-benchmarks.sh
+```
+
+When executed, the Benchmark Suite writes results to the benchmark/results/ directory.
+Within this directory, the generated .commands files contain all optimized server and client commands for the various supported models.
```
bigPYJ1151:
```suggestion
The benchmark results will be saved in `./benchmark/results/`.
In the directory, the generated `.commands` files contain all example commands for the benchmark.
```

.buildkite/performance-benchmarks/scripts/run-performance-benchmarks.sh
louie-tsai: addressed it accordingly.

.buildkite/performance-benchmarks/tests/serving-tests-cpu.json
louie-tsai: addressed it acordingly

.buildkite/performance-benchmarks/README.md
louie-tsai: addressed it accordingly

.buildkite/performance-benchmarks/tests/serving-tests-cpu.json
louie-tsai: according to AWS GNR benchmark. 8B with BF16 indeed might support max 64 users, but 3B, 1B and other smaller models indeed could support 128 or even 200 max users.
here is an AWS numbers for Qwen 1.7B  m8i.24x-94xlarge can support up to 200 users.
http://172.25.116.82/vLLM_Perf/AWS/GNR/autmoation_tools/Qwen_1.7B/perf_comparison_8x-96x.html

Therefore, plan to keep max 200 users.

.buildkite/performance-benchmarks/tests/serving-tests-cpu.json
louie-tsai: since we still want to support up to 200 concurrency, we might need 200 for 200 users case.

louie-tsai:
@mergify

.buildkite/performance-benchmarks/scripts/run-performance-benchmarks.sh
```diff
@@ -432,7 +480,11 @@ run_serving_tests() {
         echo "Running test case $test_name with qps $qps"
         echo "Client command: $client_command"
 
-        bash -c "$client_command"
+        if [ "$TEST_MODE" == "1" ]; then
+          echo "TEST MODE for '$client_command'"
+        else
```
bigPYJ1151: Please remove this

louie-tsai: addressed it accordingly

bigPYJ1151 (approved):

PR #29383, opened 2025-11-25
[responsesAPI][2] parse ResponseFunctionToolCallOutputItem

qandrew:
## Purpose

Pydantic is a bit confusing, if we curl the request, it enters as a dict, but if it's done via python, then the type is preserved. This will be needed in for supporting MCP tools for non harmony models (sneak preview https://github.com/qandrew/vllm/pull/10)

## Test Plan

added unit tests + e2e https://github.com/qandrew/vllm/pull/10


qandrew:
cc @22quinn @houseroad

houseroad (approved):
Looks good.

PR #29387, opened 2025-11-25
[BugFix] Avoid prefix cache hit in the same schedule step for mamba layers

heheda12345:
## Purpose
Mamba layers shouldn't treat blocks computed in the current step as a prefix cache hit:
Now suppose block_size = 1. We have two requests:
* req1: tokens = [A, B, C, D]
* req2: tokens = [A, B, C, E]

If these two requests are scheduled in the same step, then:
* The cache-hit length of the first request is 0, which is fine.
* The cache-hit length of the second request is 3, meaning that in this scheduling step, it is treated as if the three tokens A, B, C have already been computed, and we only need to compute E. However, computing E depends on the state of C. But when the model runner starts computing E, the state for C has not been computed yet, which will cause an error.

See test plan for the reproduce script

### Solution
If we detect a request needs a block computed in the current step, we will schedule this request into the next step. (Another solution is to treat it as cache miss, but I don't like it as it will add redundant computation)

**NOTE: the current implementation will schedule that request to the next step, and will also stop to schedule more waiting request to the current step.** This may be fine for chunked prefill as each step doesn't schedule too many waiting requests. For P/D prefill node, we need more discussion later.

**TODO: will add a test later**

## Test Plan

```

# Sample prompts.
prompts = [
    "hello what is one plus one what is one plus one what is one plus one the answer is",  # noqa: E501
    "hello what is one plus one what is one plus one what is one plus one the answer is",  # noqa: E501
]
# Create a sampling params object.
sampling_params = SamplingParams(temperature=0.8, top_p=0.95)


def main():
    # Create an LLM.
    llm = LLM(
        model="tiiuae/falcon-mamba-7b",
        enforce_eager=True,
        block_size=16,
        mamba_block_size=16,
        enable_prefix_caching=True,
    )
    # Generate texts from the prompts.
    # The output is a list of RequestOutput objects
    # that contain the prompt, generated text, and other information.
    outputs = llm.generate(prompts, sampling_params)
    # Print the outputs.
    print("\nGenerated Outputs:\n" + "-" * 60)
    for output in outputs:
        prompt = output.prompt
        generated_text = output.outputs[0].text
        print(f"Prompt:    {prompt!r}")
        print(f"Output:    {generated_text!r}")
        print("-" * 60)


if __name__ == "__main__":
    main()
```
```
VLLM_ENABLE_V1_MULTIPROCESSING=0 python3 examples/offline_inference/basic/basic.py
```
`VLLM_ENABLE_V1_MULTIPROCESSING=0` is required for reproducing this problem. It is for scheduling the two request in the same step.

## Test Result
Before this PR:
```
Generated Outputs:
Prompt:    'hello what is one plus one what is one plus one what is one plus one the answer is'
Output:    ' two yes two that is what you get when you add one plus one.\n'
Prompt:    'hello what is one plus one what is one plus one what is one plus one the answer is'
Output:    ' NO, since you only need to plug in $n$ values for '
```
It's clear that the second prompt didn't read the prefix

After this PR, thing looks better
```
Generated Outputs:
Prompt:    'hello what is one plus one what is one plus one what is one plus one the answer is'
Output:    ' two, correct?\n\nThe question here is a statement of the classic con'
Prompt:    'hello what is one plus one what is one plus one what is one plus one the answer is'
Output:    " two and what does two mean\n\nHey there! I'm here to"
```

---
<details>

</details>


vllm/v1/core/sched/scheduler.py
```diff
@@ -708,6 +710,7 @@ def schedule(self) -> SchedulerOutput:
             finished_req_ids=self.finished_req_ids,
             free_encoder_mm_hashes=self.encoder_cache_manager.get_freed_mm_hashes(),
         )
+        print("scheduler_output", scheduler_output)
```
heheda12345: todo: remove this

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -662,6 +672,16 @@ def get_num_blocks_to_allocate(
...
         assert isinstance(self.kv_cache_spec, MambaSpec)
+        if (
+            len(new_computed_blocks) > 0
+            and new_computed_blocks[-1].block_hash in self.cached_blocks_this_step
+        ):
+            # Mamba can't rely on blocks generated by other requests in the current step
+            # To put it in the next step, we return a large number of blocks so
+            # that kv_cache_manager will think there is no enough blocks to allocte now
+            # and don't schedule it in the current step.
+            print("!!!!!!!!!!!!!!!schedule into next step!!!!!!!!!!!!!!!")
```
heheda12345: todo: remove print

examples/offline_inference/basic/basic.py
```diff
@@ -5,18 +5,22 @@
 
 # Sample prompts.
 prompts = [
-    "Hello, my name is",
-    "The president of the United States is",
-    "The capital of France is",
-    "The future of AI is",
+    "hello what is one plus one what is one plus one what is one plus one the answer is",  # noqa: E501
```
heheda12345: todo revert

heheda12345:
@Josephasafg is right. This bug doesn't happen in attention.
Basically, attention has 2 steps:
1. calculate the kv cache of each token independently
2. with the calculated kv cache, compute the attention output
as the kv cache is computed in step 1, request 2 can have access to the computed kv of tokens ABC when doing attention in step 2.

vllm/v1/core/single_type_kv_cache_manager.py
```diff
@@ -662,6 +672,16 @@ def get_num_blocks_to_allocate(
...
+            return self.block_pool.num_gpu_blocks + 100000
```
Josephasafg: Should we maybe return -1 or `None` here and then check if we got either of them in `kv_cache_manager` to avoid magic numbers?

examples/offline_inference/basic/basic.py
```diff
@@ -5,18 +5,22 @@
...
 
 def main():
     # Create an LLM.
-    llm = LLM(model="facebook/opt-125m")
+    llm = LLM(
+        model="tiiuae/falcon-mamba-7b",
+        enforce_eager=True,
+        block_size=16,
+        mamba_block_size=16,
+        enable_prefix_caching=True,
```
tdoublep: Can we capture this scenario in the test_hybrid instead? I think it is useful to have a test for this.

vllm/v1/core/single_type_kv_cache_manager.py
tdoublep: I find the "returning big number" approach a bit hacky. I guess the suggestion from the AI isn't bad actually? I guess the clean way to do it would be to change the scheduler to add another check whether each request can be scheduled (that doesn't necessarily relate to the number of blocks), but we may not want to make changes in scheduler for this.

examples/offline_inference/basic/basic.py
heheda12345: I've added a new test

vllm/v1/core/single_type_kv_cache_manager.py
heheda12345: I don't want to change the scheduler. Modified to num_gpu_blocks+1

tdoublep (approved):
LGTM

PR #29388, opened 2025-11-25
[Platform] Make forward context manager pluggable for other device

shen-shanshan:
## Purpose

Currently, `set_forward_context` is directly used in some modeling files, such as `Qwen2.5-VL` (when process image/video input with ViT).

We want to use some custom forward context here (without modifying the modeling files in the plugin), such as https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/ascend_forward_context.py#L58.

Thus, this PR has made forward context manager pluggable for other device.

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/models/qwen2_5_vl.py
+++ b/vllm/model_executor/models/qwen2_5_vl.py
@@ -1246,6 +1246,10 @@ class Qwen2_5_VLForConditionalGeneration(
             self.language_model.make_empty_intermediate_tensors
         )
 
+        from vllm.platforms import current_platform
+
+        self.set_forward_context = current_platform.get_forward_context_manager()
+
     def set_aux_hidden_state_layers(self, layers: tuple[int, ...]) -> None:
         self.language_model.model.aux_hidden_state_layers = layers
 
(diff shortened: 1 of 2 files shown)


vllm/model_executor/models/qwen2_5_vl.py
```diff
@@ -1247,6 +1246,10 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = ""):
             self.language_model.make_empty_intermediate_tensors
         )
 
+        from vllm.platforms import current_platform
```
DarkLight1337: I think there is no need to import this lazily, just import it from top level

vllm/platforms/interface.py
```diff
@@ -653,6 +653,15 @@ def check_max_model_len(cls, max_model_len: int) -> int:
         """
         return max_model_len
 
+    @classmethod
+    def get_forward_context_manager(cls):
```
ProExpertProg: Can you add a type hint here?

vllm/model_executor/models/qwen2_5_vl.py
shen-shanshan: > I think there is no need to import this lazily, just import it from top level

done.

vllm/platforms/interface.py
shen-shanshan: > Can you add a type hint here?

done.

PR #29421, opened 2025-11-25
[Core] Whisper Enable Encoder Batching

NickLucche:
This PR addresses an important performance limitation of our current Whisper implementation, that is the encoder is only running one request at a time, instead of scheduling multiple audios and batching them in a single (encoder) forward. 
This is particularly bad at high-memory/high request rates deployments.  

To summarize changes with some examples:
```
# MAIN

(EngineCore_DP0 pid=2996843) INFO 11-25 18:02:30 [gpu_model_runner.py:4220] Encoder cache will be initialized with a budget of 1500 tokens, and profiled with 1 audio items of the maximum feature size.

# Extra-logging added for debugging 
***ENCODER INPUT HIDDEN STATES*** torch.Size([1500, 1280])

```
---
```
# PR

Encoder cache will be initialized with a budget of 32768 tokens, and profiled with 21 audio items of the maximum feature size.

# Extra-logging added for debugging 
***ENCODER INPUT HIDDEN STATES*** torch.Size([12, 1500, 1280])
```

These changes build on top of this PR https://github.com/vllm-project/vllm/pull/29268 to generalize seq_lens.

Changes here are confined to:
 
 Whisper: just batching hidden states properly before MHA
 
 Scheduling: avoid constraining `self.scheduler_config.max_num_encoder_input_tokens` to max size of one item. This effectively inhibits the EncoderCache from allowing multiple items in a scheduling step.
 
 ~~More scheduling: there's an issue related to skipping `check_and_update_cache` for requests with same input.  
 The problem is that `EncoderCacheManager.allocate()` always decrements slots, even when reusing an entry from freeable. This causes `num_freeable_slots` and `freeable` to get out of sync, as `check_and_update_cache()` is always skipped for encoder-decoder.~~
~~What I did in this PR is to have encoder-decoder models call `check_and_update_cache`, but stil schedule the encoder input even if cached. This allows to keep the cache state in sync while retaining previous behavior.~~
 Future work can focus on enabling the cache for Whisper, now that the flow is getting more and more aligned with MM models.

EDIT: I have instead provided a much simpler alternative cache that clearly highlights the workflow of enc-dec models (for scheduling only).

## Results

```
# MAIN


# 10
RESULTS SUMMARY
Total samples: 10
Successful: 10
Failed: 0
Total time: 0.39s
Average latency: 0.19s
Throughput: 25.96 requests/s

# 50

RESULTS SUMMARY
Total samples: 73
Successful: 73
Failed: 0
Total time: 1.35s
Average latency: 0.63s
Throughput: 54.27 requests/s

==========================================================================
```

```
# This PR

# 10
RESULTS SUMMARY
Total samples: 10
Successful: 10
Failed: 0
Total time: 0.35s
Average latency: 0.15s
Throughput: 28.83 requests/s

# 50
RESULTS SUMMARY
Total samples: 73
Successful: 73
Failed: 0
Total time: 0.98s
Average latency: 0.49s
Throughput: 74.32 requests/s
```

cc @DarkLight1337 @russellb

### DIFF
--- a/tests/entrypoints/openai/test_transcription_validation_whisper.py
+++ b/tests/entrypoints/openai/test_transcription_validation_whisper.py
@@ -49,6 +49,7 @@ async def test_basic_audio(whisper_client, mary_had_lamb):
 
 
 @pytest.mark.asyncio
+@pytest.mark.skip(reason="Skipping batched test due to interaction with other tests")
 async def test_basic_audio_batched(mary_had_lamb, winning_call, whisper_client):
     transcription = whisper_client.audio.transcriptions.create(
         model=MODEL_NAME,
--- a/tests/entrypoints/openai/test_transcription_validation_whisper.py
+++ b/tests/entrypoints/openai/test_transcription_validation_whisper.py
@@ -143,7 +143,6 @@ async def test_streaming_response(winning_call, whisper_client):
         language="en",
         temperature=0.0,
         stream=True,
-        timeout=30,
     )
     # Reconstruct from chunks and validate
     async for chunk in res:
@@ -162,7 +161,6 @@ async def test_stream_options(winning_call, whisper_client):
         temperature=0.0,
         stream=True,
         extra_body=dict(stream_include_usage=True, stream_continuous_usage_stats=True),
-        timeout=30,
     )
     final = False
     continuous = True
--- a/tests/entrypoints/openai/test_transcription_validation_whisper.py
+++ b/tests/entrypoints/openai/test_transcription_validation_whisper.py
@@ -220,6 +220,8 @@ async def test_audio_prompt(mary_had_lamb, whisper_client):
         temperature=0.0,
     )
     out = json.loads(transcription)["text"]
+    # Might cause timeout if the file is not seekable.
+    mary_had_lamb.seek(0)
     assert prefix in out
     transcription_wprompt = await whisper_client.audio.transcriptions.create(
         model=MODEL_NAME,
--- a/tests/entrypoints/openai/test_transcription_validation_whisper.py
+++ b/tests/entrypoints/openai/test_transcription_validation_whisper.py
@@ -49,7 +49,6 @@ async def test_basic_audio(whisper_client, mary_had_lamb):
 
 
 @pytest.mark.asyncio
-@pytest.mark.skip(reason="Skipping batched test due to interaction with other tests")
 async def test_basic_audio_batched(mary_had_lamb, winning_call, whisper_client):
     transcription = whisper_client.audio.transcriptions.create(
         model=MODEL_NAME,
--- a/tests/entrypoints/openai/test_transcription_validation_whisper.py
+++ b/tests/entrypoints/openai/test_transcription_validation_whisper.py
@@ -136,6 +136,7 @@ async def test_streaming_response(winning_call, whisper_client):
         language="en",
         temperature=0.0,
     )
+    winning_call.seek(0)
     res = await whisper_client.audio.transcriptions.create(
         model=MODEL_NAME,
         file=winning_call,
--- a/tests/entrypoints/openai/test_transcription_validation_whisper.py
+++ b/tests/entrypoints/openai/test_transcription_validation_whisper.py
@@ -136,13 +136,13 @@ async def test_streaming_response(winning_call, whisper_client):
         language="en",
         temperature=0.0,
     )
-    winning_call.seek(0)
     res = await whisper_client.audio.transcriptions.create(
         model=MODEL_NAME,
         file=winning_call,
         language="en",
         temperature=0.0,
         stream=True,
+        timeout=30,
     )
     # Reconstruct from chunks and validate
     async for chunk in res:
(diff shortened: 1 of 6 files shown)


DarkLight1337 (approved):
Thanks, LGTM

vllm/config/vllm.py
```diff
@@ -549,27 +549,16 @@ def __post_init__(self):
...
-                from vllm.multimodal import MULTIMODAL_REGISTRY
-
-                self.scheduler_config.max_num_encoder_input_tokens = (
-                    MULTIMODAL_REGISTRY.get_encdec_max_encoder_len(self.model_config)
-                )
-                logger.debug(
-                    "Encoder-decoder model detected: setting "
-                    "`max_num_encoder_input_tokens` to encoder length (%s)",
-                    self.scheduler_config.max_num_encoder_input_tokens,
+            if (
```
robertgshaw2-redhat: why not just set the multiproc method to spawn?

vllm/v1/core/sched/scheduler.py
```diff
@@ -892,15 +892,18 @@ def _try_schedule_encoder_inputs(
                 # in the decoder's KV cache.
                 continue
 
+            # Check and update encoder cache accounting.
```
robertgshaw2-redhat: This comment out of context is not super helpful, can you improve it?

vllm/config/vllm.py
NickLucche: not sure there was history here but should check with @russellb https://vllm-dev.slack.com/archives/C07QCGVDNUF/p1760579949992319

NickLucche:
Still blocked on CI here @DarkLight1337 :(

vllm/model_executor/models/whisper.py
```diff
@@ -530,7 +531,13 @@ def forward(self, input_features: torch.Tensor | list[torch.Tensor]):
...
             )
             hidden_states.append(embeds)
-        hidden_states = torch.cat(hidden_states)
+            input_is_batched = embeds.ndim > 2
+        # Input to MHA must be B x T x D
+        if input_is_batched:
+            # Models using WhisperEncoder may handle batching internally.
+            hidden_states = torch.cat(hidden_states)
+        else:
+            hidden_states = torch.stack(hidden_states, dim=0)
```
ekagra-ranjan, line 540: I was imagining the behavior in vLLM would be either be always batched (need cat) or not batched (needs stack). Wondering why we need to support both cases?

NickLucche: some architectures build on top of whisper and have been handling batching on their own (ie ultravox).
We certainly want to unify that behavior, but it's better to address that in a separate PR that doesn't focus on critical performance imo.

vllm/v1/core/sched/scheduler.py
```diff
@@ -936,15 +936,18 @@ def _try_schedule_encoder_inputs(
...
+            # This ensures proper tracking of freeable entries.
+            is_cached = self.encoder_cache_manager.check_and_update_cache(request, i)
             if not self.is_encoder_decoder:
-                # We are not using the encoder cache for encoder-decoder models,
-                # yet.
```
NickLucche, line 946: Earlier we weren't calling `check_and_update_cache` as cache size was 1 item, and at most 1 item was scheduled per scheduler step..

ekagra-ranjan:
> This needs some tweaking to the scheduling logic to avoid a hang with encoder cache starving

@NickLucche - which test surfaced this issue?

NickLucche:
@ekagra-ranjan the whole test_transcription_validation suite, but can repro by sending the same request until cache is full.
I am not sure we can have enc-dec share the cache logic nicely without actually using it (it works on main because it's constrained to a single item), because it breaks some assumptions related to how state is updated.

cc @DarkLight1337 in case you have some better ideas on how to handle it.

tests/entrypoints/openai/test_transcription_validation_whisper.py
```diff
@@ -49,6 +49,7 @@ async def test_basic_audio(whisper_client, mary_had_lamb):
 
 
 @pytest.mark.asyncio
+@pytest.mark.skip(reason="Skipping batched test due to interaction with other tests")
```
DarkLight1337: Is this intended? What do you mean by "interaction with other tests"?

NickLucche: nope I will revert this once scheduling is fixed, I initially believed this was an usual CI timeout

vllm/config/model.py
```diff
@@ -539,6 +539,11 @@ def __post_init__(
 
         self.original_max_model_len = self.max_model_len
         self.max_model_len = self.get_and_verify_max_len(self.max_model_len)
+
+        if self.is_encoder_decoder:
```
robertgshaw2-redhat, line 543: note: I dont think this actually is enough. I still need to specify this

DarkLight1337: This setting does not apply to the MM feature cache in the model runner.

PR #24662, opened 2025-09-11
[BUGFIX] Fix crash in Eagle Speculative Decoding models when exceedin…

AlonKejzman:
…g draft model length

## Purpose
Enable running Eagle Speculative Decoding in environments where the input may exceed the drafter model length but not the verifier's.

## Test Plan
Running with inputs that are below the drafter model length, and then between the drafter and verifier.

## Test Result
Successful.

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #29422, opened 2025-11-25
[Bugfix][Spec Decode] Fix spec decode hang with DP by unconditionally running drafter

MatthewBonanni:
**UPDATED 2026-04-22**

## Purpose
The `input_fits_in_drafter` check caused DP ranks to diverge when batches had different max sequence lengths -- some ranks would skip speculation while others ran it, leading to deadlocks. This PR clamps the drafter's max model length to no less than the target max model length, and runs the drafter unconditionally. At lengths greater than the drafter's `max_model_len`, the drafts may be low-quality, but there will be no hang. Also yields a nice LoC reduction in `GPUModelRunner`

User-specified `--speculative-config.max_model_len` is respected, even if less than the target's `max_model_len` and hangs may result (though a warning is printed in this case).

This approach is not possible with `rope_scaling`, so in that case, the target's `max_model_len` is capped at that of the drafter, with a logged warning.

## Test Plan
`yuhuili/EAGLE-LLaMA3-Instruct-8B` has `max_position_embeddings = 2048`
`meta-llama/Meta-Llama-3-8B-Instruct` has `max_position_embeddings = 8192`

```
vllm serve meta-llama/Meta-Llama-3-8B-Instruct \
    -dp 2 \
    --speculative-config '{"method": "eagle", "model": "yuhuili/EAGLE-LLaMA3-Instruct-8B", "num_speculative_tokens": 4}'
```

with 

```
lm_eval \
    --model_args base_url=http://localhost:8000/v1/chat/completions,model=meta-llama/Meta-Llama-3-8B-Instruct,num_concurrent=128,tokenized_requests=False \
    --tasks gsm8k \
    --apply_chat_template
```

## Test Result
No hangs. Works despite drafter's `max_model_len=2048`. Drafter hits typical acceptance rates:

`
(APIServer pid=2338266) INFO 04-22 20:06:41 [metrics.py:101] SpecDecoding metrics: Mean acceptance length: 2.45, Accepted throughput: 1096.17 tokens/s, Drafted throughput: 3015.65 tokens/s, Accepted: 10963 tokens, Drafted: 30160 tokens, Per-position acceptance rate: 0.712, 0.416, 0.216, 0.110, Avg Draft acceptance rate: 36.3%
`

```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.7748|±  |0.0115|
|     |       |strict-match    |     5|exact_match|↑  |0.6528|±  |0.0131|
```

---
<details>

</details>


njhill:
Thanks @MatthewBonanni it would be much better if we can remove this. If it's not required, I'm curious of the reason that it was added in the first place? (in https://github.com/vllm-project/vllm/pull/24662)

MatthewBonanni:
Updated on 2026-04-22, now addresses the repro from https://github.com/vllm-project/vllm/pull/24662 (although, as described above, user-specified --speculative-config.max_model_len is respected, even if less than the target's `max_model_len` and hangs may result, though a warning is printed in this case). Without the user-specified max, it works fine.

NaNAGISaSA:
Same problem for dp hangs, as this pr shown, `input_fits_in_drafter` is no use? @MatthewBonanni

MatthewBonanni:
closing in favor of #40589

PR #29430, opened 2025-11-25
[TPU] Fix tensor shape incompatibility in Transformers backend

SamuelMarks:
## Purpose

Here is the formatted Commit Message and generic PR description.

### Commit Message
```text
[TPU] Fix shape mismatch in Transformers backend

The HuggingFace Transformers backend passes 3D tensors (Batch, Seq, Hidden)
to layers, but vLLM's optimized kernels (linear, norm, etc.) on TPU explicitly
expect flattened 2D tensors (Tokens, Hidden). This mismatch causes ValueErrors
in JAX einsum operations.

This commit installs a forward-pass shim during the `recursive_replace` step
in `vllm/model_executor/models/transformers/base.py`. The shim automatically
flattens 3D inputs to 2D before execution and restores the 3D shape/batch
structure upon return. This ensures compatibility for models like Gemma3-MM
and PaliGemma on TPU without modifying core vLLM layers.
```

---

## Purpose
This PR resolves a critical tensor shape mismatch when running the HuggingFace Transformers backend on TPU. It introduces a compatiblity shim that adapts vLLM's optimized 2D kernels to accept the 3D tensor structure inherent to the Transformers architecture.

When using the Transformers backend (e.g., for **Gemma3-MM** or **PaliGemma**), tensor activations retain their 3D shape `(Batch, Seq, Hidden)`. However, vLLM's internal layers (Linear, RMSNorm) and specifically the TPU/JAX kernels, are optimized for flattened 2D inputs `(Num_Tokens, Hidden)`.

On TPU, this results in a dimension mismatch error:
```
ValueError: Einstein sum subscript 'mn' does not contain the correct number of indices for operand 0
```

Instead of modifying the core kernels (which would impact all models), this PR implements a strictly isolated fix within `vllm/model_executor/models/transformers/base.py`.

During `recursive_replace`, we inject a **Forward Shim** onto replaced modules (Linear, Norm, MoE):
1.  **Intercept:** Detects if the input tensor is 3D.
2.  **Flatten:** Reshapes `(B, S, H)` $\to$ `(B*S, H)` before calling the vLLM layer.
3.  **Execute:** Runs the efficient vLLM/TPU kernel on 2D data.
4.  **Restore:** Reshapes the output `(B*S, Out)` $\to$ `(B, S, Out)` before returning control to the HF model.

This also correctly handles vLLM layers that return `(output, bias)` tuples.

## Test Plan
TBD

## Test Result
TBD

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/models/transformers/base.py
+++ b/vllm/model_executor/models/transformers/base.py
@@ -18,6 +18,7 @@
 
 from collections.abc import Iterable
 from typing import TYPE_CHECKING
+import types
 
 import regex as re
 import torch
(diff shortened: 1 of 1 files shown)


hmellor:
I'm going to close this in favour of fixing the Jax in `tpu-inference`. Most of the Jax I've seen does seem compatible with an empty batch dimension, it's just one `einsum`

PR #29352, opened 2025-11-24
[Attention][CUDAGraph] Remove CG padding from attention backends

MatthewBonanni:
## Purpose
#28579 refactors the logic for cudagraph padding so that it happens prior to attention metadata building. This is a follow-up which removes remaining unnecessary padding calls from miscellaneous backends.

## Test Plan
`pytest tests/models/language/generation -m hybrid_model`

## Test Result
Should pass in CI: Language Models Tests (Hybrid)

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #29437, opened 2025-11-25
[Hybrid] Remove use of PAD_SLOT_ID for causal_conv_1d

tdoublep:
## Purpose

The causal conv kernels use (/misuse) the `PAD_SLOT_ID` to indicate where padding in the state indices tensor exists. This creates a headache when doing meta-data building for FCG because we need to ensure that the PAD_SLOT_IDs are inserted in the right place all the time. This shouldn't actually be necessary 

This PR:

- Strips out the use of `PAD_SLOT_ID` from the kernels, instead relying on the varlen representation to do the correct thing if some sequences in the batch have query length 0
- Ensures we use the varlen representation when calling the "update" kernel for decode from the Mamba2 mixer
- Removes some of the weirdness regarding the "metadata" is passed to the kernels

cc @LucasWilkinson 

TODO:
- Change other models that use these kernels 

## Test Plan

I've verified that the mamba2 tests and the kernel tests are passing

## Test Result

---
<details>

</details>

### DIFF
--- a/tests/kernels/mamba/test_causal_conv1d.py
+++ b/tests/kernels/mamba/test_causal_conv1d.py
@@ -7,12 +7,12 @@ import torch
 import torch.nn.functional as F
 from einops import rearrange
 
-from vllm.attention.backends.utils import PAD_SLOT_ID
 from vllm.model_executor.layers.mamba.ops.causal_conv1d import (
     causal_conv1d_fn,
     causal_conv1d_update,
 )
 from vllm.platforms import current_platform
+from vllm.v1.attention.backends.utils import compute_causal_conv1d_metadata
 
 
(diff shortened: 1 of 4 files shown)


hmellor:
Closing as obsolete: main has since moved the decode path onto the varlen representation (#33726), replaced `pad_slot_id` with `null_block_id` in the update kernel and rewrote `test_causal_conv1d.py` accordingly (#35431), and removed the CUDA graph `PAD_SLOT_ID` padding from `mamba2_attn.py` (#29352). All four files here conflict with main and every hunk targets code that has been rewritten. Worth a fresh PR against current main if the residual `pad_slot_id` in `causal_conv1d_fn` is still worth removing.

PR #29439, opened 2025-11-25
[Bugfix] Fix grouped_topk pytorch impl when num_experts can't be grouped properly

divakar-amd:
This pull request adds a safeguard to the `grouped_topk` function in `fused_moe.py` to ensure correct behavior when the number of experts does not meet the requirements for grouped top-k computation. If the conditions are not met, the function now logs a warning and falls back to the standard top-k computation.

Note:
- This PR is required to resolve the following test on ROCm:
    - `pytest -s -v tests/v1/e2e/test_spec_decode.py::test_mtp_correctness[deepseek]`
- The issue also occurs on cuda if  `envs.VLLM_USE_FUSED_MOE_GROUPED_TOPK` is not used

e.g. For the following model used in `tests/v1/e2e/test_spec_decode.py`
- `ZixiQi/DeepSeek-V3-4layers-MTP-FP8`
    - "n_group": 8,
    - "n_routed_experts": 8,

Without this PR we get the following error:
>   File "/usr/local/lib/python3.12/dist-packages/vllm/model_executor/layers/fused_moe/fused_moe.py", line 1209, in torch_dynamo_resume_in_grouped_topk_at_1205
    scores.view(num_token, num_expert_group, -1).topk(2, dim=-1)[0].sum(dim=-1)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: selected index k out of range


SageMoore (commented):
This looks reasonable to me, but I'm not seeing a log in the code. Am I just missing something obvious?

SageMoore (approved):

gshtras (approved):

robertgshaw2-redhat:
hey @divakar-amd - this PR changes the model definition (switch from using `grouped_topk` -> `fused_topk`).

I don't think this is a valid thing to do

Can you elaborate why this case occurs? There is 256 experts, 8 groups. So it should be fine ...

PR #29471, opened 2025-11-26
Remove upstream fa checks

Victor49152:
## Purpose
According to https://github.com/vllm-project/vllm/pull/28763, the vllm flash attention has supported all headsize for vit module, thus removing the upstream flash-attn checks as they are no longer necessary.

## Test Plan
Use one of impacted model qwen3-vl-235B as example to start the server 
`vllm serve RedHatAI/Qwen3-VL-235B-A22B-Instruct-NVFP4 -tp 4 -dp 1 --mm-encoder-tp-mode data --enable-expert-parallel --async-scheduling --max-num-seqs 1024`

## Test Result
Server started successfully 

---
<details>

</details>


vllm/attention/layer.py
```diff
@@ -53,53 +53,25 @@
...
             attn_backend = AttentionBackendEnum.ROCM_AITER_FA
-
-        elif (
-            check_upstream_fa_availability(torch.get_default_dtype())
-            and on_gfx9()
-            and attn_backend_override is None
-        ):
-            attn_backend = AttentionBackendEnum.FLASH_ATTN
-            use_upstream_fa = True
+        elif attn_backend_override is None \
```
tjtanaa: we need to add back the `on_gfx9()` condition here to differentiate between Radeon and Instinct GPUs.

On Radeon, only TORCH_SDPA is supported.

vllm/attention/layer.py
```diff
@@ -110,10 +82,7 @@ def maybe_get_vit_flash_attn_backend(
         if attn_backend == AttentionBackendEnum.ROCM_AITER_FA:
             from aiter import flash_attn_varlen_func
         else:
-            if use_upstream_fa:
-                from flash_attn import flash_attn_varlen_func
-            else:
-                from vllm.attention.utils.fa_utils import flash_attn_varlen_func
+            from vllm.attention.utils.fa_utils import flash_attn_varlen_func
```
tjtanaa, line 91: `vllm/attention/utils/fa_utils.py` does not have the logic for ROCm, `flash_attn_varlen_func` will be a None object if imported this way.

We can keep the import statement `from flash_attn import flash_attn_varlen_func` for now. Else we have to add this `from flash_attn import flash_attn_varlen_func` import statement into the `vllm/attention/utils/fa_utils.py` when platform is rocm.

vllm/attention/layer.py
Victor49152: Just pushed this changes, thanks and please comment if there is anything else you notice

vllm/attention/ops/vit_attn_wrappers.py
```diff
@@ -27,15 +27,11 @@ def flash_attn_maxseqlen_wrapper(
...
-    use_upstream_fa: bool,
 ) -> torch.Tensor:
     if is_rocm_aiter:
         from aiter import flash_attn_varlen_func
     else:
-        if use_upstream_fa:
-            from flash_attn import flash_attn_varlen_func
-        else:
-            from vllm.attention.utils.fa_utils import flash_attn_varlen_func
+        from vllm.attention.utils.fa_utils import flash_attn_varlen_func
```
tjtanaa, line 34: like wise, 

`vllm/attention/utils/fa_utils.py` does not have the logic for ROCm, `flash_attn_varlen_func` will be a None object if imported this way.

We can keep the import statement `from flash_attn import flash_attn_varlen_func` for now. Else we have to add this `from flash_attn import flash_attn_varlen_func` import statement into the `vllm/attention/utils/fa_utils.py`  when platform is rocm.

vllm/attention/layer.py
Victor49152: I added this import to fa_utils as it looks like the most simple way of it. And except message tells user to install upstream fa when import error is raised. Please check if that works, thanks!

ywang96 (approved):
I fixed the precommit error but otherwise LGTM

cc @tjtanaa for final check on the changes for resolving FA import on ROCM platform.

tjtanaa:
@ywang96 Thanks. LGTM.

It is using the flash attention and aiter flash attention.
And the code path on ROCm is working, ChartQA score of Qwen/Qwen3-VL-8B-Instruct when using both backends are 
```
Metrics:
{
    "explicit_prompt_relaxed_correctness": 0.7948,
    "anywhere_in_answer_relaxed_correctness": 0.7988
}
```

Issue #29369, opened 2025-11-25
[Bug]: nixl connector PD disagg stuck at INFO logging level


herotai214:
### Your current environment

/

### 🐛 Describe the bug

cc EPD authors @fake0fan @knlnguyen1802 @herotai214  @khuonglmhw 
cc @NickLucche 

Following #29248 having issue with the EPD disagg feature, I found that simply running PD disagg with nixl connector with `export VLLM_LOGGING_LEVEL=INFO` would also stuck at the decode instance when streaming for the 1st request.

`export VLLM_LOGGING_LEVEL=DEBUG` avoids the issue and works fine, but I think there are still some defects that worth paying attention...

My starting script for PD disagg with nixl connector using `/tests/v1/kv_connector/nixl_integration/toy_proxy_server.py`:

```shell
#!/bin/bash
set -xe

echo "🚧🚧 Running NIXL CONNECTOR!!!!!!!! 🚧🚧"

MODEL_NAME=<YOUR_MODEL_NAME_HERE>
LOG_PATH=<YOUR_LOG_PATH_HERE>
mkdir -p $LOG_PATH

# Cleanup function
cleanup() {
    echo "Stopping everything…"
    trap - INT TERM USR1   # prevent re-entrancy
    
    # Kill all tracked PIDs
    for pid in "${PIDS[@]}"; do
        if kill -0 "$pid" 2>/dev/null; then
            echo "Killing process $pid"
            kill "$pid" 2>/dev/null
        fi
    done
    
    # Wait a moment for graceful shutdown
    sleep 2
    
    # Force kill any remaining processes
    for pid in "${PIDS[@]}"; do
        if kill -0 "$pid" 2>/dev/null; then
            echo "Force killing process $pid"
            kill -9 "$pid" 2>/dev/null
        fi
    done
    
    # Kill the entire process group as backup
    kill -- -$$ 2>/dev/null
    
    echo "All processes stopped."
    exit 0
}

trap cleanup INT
trap cleanup USR1
trap cleanup TERM

export VLLM_HOST_IP=$(hostname -I | awk '{print $1}')

# a function that waits vLLM server to start
wait_for_server() {
  local port=$1
  timeout 12000 bash -c "
    until curl -s localhost:${port}/v1/chat/completions > /dev/null; do
      sleep 1
    done" && return 0 || return 1
}

### NIXL
A=2
B=3
C=6661
D=6662
PORT=9775

[...]

PR #29476, opened 2025-11-26
[BugFix] Add sleep to fix tight loop and release GIL

alec-flowers:
Potential Fix for - https://github.com/vllm-project/vllm/issues/29369

While not very elegant it does do the job of releasing the GIL.

## Test Result

---
<details>

</details>


vllm/v1/engine/core.py
```diff
@@ -895,6 +895,13 @@ def _process_engine_step(self) -> bool:
         # Post-step hook.
         self.post_step(model_executed)
 
+        # If no model execution happened but there are waiting requests
+        # (e.g., WAITING_FOR_REMOTE_KVS), yield the GIL briefly to allow
+        # background threads (like NIXL handshake) to make progress.
+        # Without this, the tight polling loop can starve background threads.
+        if not model_executed and self.scheduler.has_requests():
```
njhill: nit: this would be slightly better
```suggestion
        if not model_executed and self.scheduler.has_unfinished_requests():
```

njhill (commented):
Thanks @alec-flowers. I think this might actually be the best fix given the current design.

Ideally we should avoid a hot loop altogether in this scenario but that would require more significant rework.

njhill:
> I have a feeling this is an ARM vs x86 issue: #30228

The sched_yield thing only applies to multiproc executor case where we spin on the shm queue, this issue is kind of separate to that and would apply in uniproc case too.

njhill (approved):
Thanks @alec-flowers!

PR #29487, opened 2025-11-26
[Bugfix] Correct num_q_heads on DCP for Flashinfer backends 

gjc0824:
## Purpose
Fix num_q_heads on DCP for Flashinfer backends
In current DCP code for Flashinfer, the `num_qo_heads` is repeatedly multiplied by `dcp_world_size`.

## Test Result
Although it does not affect the precision of inference, sometimes it causes share memory issues. After fixing this bug, the issue can be mitigated.


LucasWilkinson (commented):
Thanks for the fix! Overall looks good to me but, do you know the current time for the current CP test; iirc they are already pretty long so im hesitant to add more cases till we can get the time under control; I think we can write simpler tests like `tests/v1/distributed/test_dbo.py` and limit cudagraph sizes to seed this up

gjc0824:
> Thanks for the fix! Overall looks good to me but, do you know the current time for the current CP test; iirc they are already pretty long so im hesitant to add more cases till we can get the time under control; I think we can write simpler tests like `tests/v1/distributed/test_dbo.py` and limit cudagraph sizes to seed this up

I'm sorry, forgive me for not fully understanding why the current CP test is so time-consuming. I see that `tests/v1/distriduted/test_dbo.py` uses part of the GSM8K data for quick evaluation, but the CP test only uses a single prompt [code](https://github.com/vllm-project/vllm/blob/6b691a9fdb5f60ae05d521e9ec1871023ce15466/tests/utils.py#L637) for testing.

LucasWilkinson (approved):
LGTM; we should address the test time issues ASAP

gjc0824:
> LGTM; we should address the test time issues ASAP

Sure. Thanks for your review.

PR #29488, opened 2025-11-26
Add SpecDec support to `selective_state_update`

roikoren755:
## Purpose
Add support in the `selective_state_update` kernel for speculative decoding.

This PR is the first step in enabling the use of Mamba-based models as target models in speculative decoding scenarios.
I've updated the `selective_state_update` kernel to more closely resemble the `fused_recurrent_gated_delta_rule` kernel used in Qwen3-Next's GDN module:
1. Accepting variable sequence lengths
2. Iteratively calculating SSM state and return values per sequence
3. Optionally storing intermediate SSM states according to `dst_state_batch_indices` tensor
4. Accepting a `num_accepted_tokens` tensor to signal which previous SSM state to use as the initial state

We used this kernel in an implementation of a Mamba2-based model with an MTP head, which follow-up PRs will add full support for, including full CUDA graphs and prefix-caching.

## Test Plan
All current `selective_state_update` tests pass.
Added tests for the new arguments to the `selective_state_update` kernel.

## Test Result
All current `selective_state_update` tests pass.
All new tests pass.

---
<details>

</details>

### DIFF
--- a/tests/kernels/mamba/test_mamba_ssm.py
+++ b/tests/kernels/mamba/test_mamba_ssm.py
@@ -973,7 +973,7 @@ def test_selective_state_update_with_num_accepted_tokens(
 @pytest.mark.parametrize("dstate", [16, 64])
 @pytest.mark.parametrize("dim", [2048, 4096])
 @pytest.mark.parametrize("max_seq_len", [2, 4])
-def test_selective_state_update_varlen_with_num_accepted_and_inplace(
+def test_selective_state_update_varlen_with_num_accepted(
     dim, dstate, has_z, itype, max_seq_len
 ):
     device = "cuda"
--- a/tests/kernels/mamba/test_mamba_ssm.py
+++ b/tests/kernels/mamba/test_mamba_ssm.py
@@ -425,6 +425,80 @@ def test_selective_state_update(dim, dstate, has_z, itype):
     assert torch.allclose(out, out_ref, rtol=rtol, atol=atol)
 
 
+@pytest.mark.parametrize("itype", [torch.float32, torch.bfloat16])
+@pytest.mark.parametrize("has_z", [False, True])
+@pytest.mark.parametrize("dstate", [16, 64])
+@pytest.mark.parametrize("dim", [2048, 2048 + 16, 4096])
+@pytest.mark.parametrize("max_seq_len", [1, 2, 4])
+def test_selective_state_update_varlen(dim, dstate, has_z, itype, max_seq_len):
+    device = "cuda"
+    rtol, atol = (3e-4, 1e-3) if itype == torch.float32 else (5e-3, 1e-2)
+    if itype == torch.bfloat16:
+        rtol, atol = 5e-2, 1.5e-1
+        if torch.version.hip:
+            atol *= 2
+    # set seed
+    current_platform.seed_everything(0)
+    batch_size = 4
+    token_counts = torch.randint(1, max_seq_len + 1, (batch_size,), device=device)
+    total_tokens = int(token_counts.sum().item())
+    cu_seqlens = torch.tensor(
+        [0] + torch.cumsum(token_counts, dim=0).tolist(),
+        dtype=torch.int32,
+        device=device,
+    )
+    state = torch.randn(batch_size, dim, dstate, dtype=itype, device=device)
+    x = torch.randn(total_tokens, dim, device=device, dtype=itype)
+    out = torch.empty_like(x)
+    dt = torch.randn(total_tokens, dim, device=device, dtype=itype)
+    dt_bias = torch.rand(dim, device=device) - 4.0
+    A = -torch.rand(dim, dstate, device=device) - 1.0
+    B = torch.randn(total_tokens, dstate, device=device)
+    C = torch.randn(total_tokens, dstate, device=device)
+    D = torch.randn(dim, device=device)
+    z = torch.randn_like(x) if has_z else None
+    state_ref = state.detach().clone()
+    selective_state_update(
+        state,
+        x,
+        dt,
+        A,
+        B,
+        C,
+        D=D,
+        z=z,
+        dt_bias=dt_bias,
+        dt_softplus=True,
+        out=out,
+        cu_seqlens=cu_seqlens,
+    )
+
+    out_ref_list = []
+    for seq_idx in range(batch_size):
+        start_idx = cu_seqlens[seq_idx].item()
+        end_idx = cu_seqlens[seq_idx + 1].item()
+        num_tokens = end_idx - start_idx
+        for token_idx in range(num_tokens):
+            idx = start_idx + token_idx
+            out_ref_list.append(
+                selective_state_update_ref(
+                    state_ref[seq_idx : seq_idx + 1],
+                    x[idx : idx + 1],
+                    dt[idx : idx + 1],
+                    A,
+                    B[idx : idx + 1],
+                    C[idx : idx + 1],
+                    D=D,
+                    z=z[idx : idx + 1] if has_z else None,
+                    dt_bias=dt_bias,
+                    dt_softplus=True,
+                )
+            )
+    out_ref = torch.cat(out_ref_list, dim=0)
+    assert torch.allclose(state, state_ref, rtol=rtol, atol=atol)
+    assert torch.allclose(out, out_ref, rtol=rtol, atol=atol)
+
+
 @pytest.mark.parametrize("wtype", [torch.float32])
 @pytest.mark.parametrize("itype", [torch.float32])
 @pytest.mark.parametrize("seqlen", [1, 256, 1024, 4096])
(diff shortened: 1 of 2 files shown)


vllm/model_executor/layers/mamba/ops/mamba_ssm.py
tlrmchlsmth, line 1: One thing I noticed is that it looks like the stride calculations might be happening in int32, which could be problematic due to overflows. Not a problem with this PR but something we might want to fix - WDYT @tdoublep?

tlrmchlsmth (commented):
Thanks for the contribution! Please add a unit test that provides coverage for the additional functionality.

How will this be used?

roikoren755:
@tlrmchlsmth updated the PR with tests for the new functionalities, would appreciate another look 🙏

vllm/model_executor/layers/mamba/ops/mamba_ssm.py
```diff
@@ -99,22 +114,39 @@ def _selective_scan_update_kernel(
...
+            dst_state_ptr = state_ptr + (
+                dst_state_batch_idx * stride_state_batch + pid_h * stride_state_head
+            )
+
+        state_batch_indices_ptr += (
+            pid_b * stride_state_indices_batch + init_token_idx * stride_state_indices_T
         )
-        state_batch_indices_ptr += pid_b
         state_batch_idx = tl.load(state_batch_indices_ptr).to(tl.int64)
         state_ptr += state_batch_idx * stride_state_batch + pid_h * stride_state_head
```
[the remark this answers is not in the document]
tlrmchlsmth: @roikoren755 could you please take a look at this comment to double check if this is an issue?

vllm/model_executor/layers/mamba/ops/mamba_ssm.py
```diff
@@ -99,22 +110,52 @@ def _selective_scan_update_kernel(
...
+        bos = tl.load(cu_seqlens_ptr + pid_b).to(tl.int64)
+        eos = tl.load(cu_seqlens_ptr + pid_b + 1).to(tl.int64)
+        seq_len = eos - bos
+
+        if seq_len == 0:
+            return
+    else:
+        bos = pid_b * T
+        eos = bos + T
+        seq_len = T
```
tlrmchlsmth: Could you explain this part a bit? I'm a bit confused about what's going on.

I thought that `cu_seqlens_ptr` would always be set now in vLLM. And what is T? 

My guess is that the speculative decoding changes to this kernel would only be used for pure decode cases, and T is the number of speculated tokens, but I'm not sure

tests/kernels/mamba/test_mamba_ssm.py
```diff
@@ -425,6 +425,80 @@ def test_selective_state_update(dim, dstate, has_z, itype):
     assert torch.allclose(out, out_ref, rtol=rtol, atol=atol)
 
 
+@pytest.mark.parametrize("itype", [torch.float32, torch.bfloat16])
+@pytest.mark.parametrize("has_z", [False, True])
+@pytest.mark.parametrize("dstate", [16, 64])
+@pytest.mark.parametrize("dim", [2048, 2048 + 16, 4096])
+@pytest.mark.parametrize("max_seq_len", [1, 2, 4])
```
tlrmchlsmth, line 432: These `max_seq_lens` seem very low - is that because they're the number of speculated tokens?

vllm/model_executor/layers/mamba/ops/mamba_ssm.py
roikoren755: The idea is that, in regular decoding cases, cu_seqlens won't be passed, and the kernel wrapper sets T as 1 (making these lines somewhat silly, come to think of it...). The T is the maximum sequence length, which in the non-varlen case is the uniform sequence length. In theory, the kernel can handle multiple tokens per-sequence in that case, but I don't think there will be a code flow that uses it.

In batched speculative decoding cases, the number of speculated tokens per request isn't always uniform, so T isn't relevant in those cases

tests/kernels/mamba/test_mamba_ssm.py
roikoren755: correct, maybe the parameter name isn't that clear here

vllm/model_executor/layers/mamba/ops/mamba_ssm.py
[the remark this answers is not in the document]
roikoren755: The comment isn't right here. The `num_accepted_tokens` tensor is provided in token verification forward passes, not in single token generation steps. And the `dst_state_batch_indices` in speculative decoding cases will (in a follow-up PR) hold the relevant Mamba speculative blocks.

I didn't want to add assertions that T isn't smaller than the maximum value in the `num_accepted_tokens` tensor, as it doesn't work well with CUDA graphs, but maybe it should be added? Or assert that if `num_accepted_tokens` is passed, then `state_batch_indices` is 2D?

vllm/model_executor/layers/mamba/ops/mamba_ssm.py
```diff
@@ -220,6 +287,9 @@ def selective_state_update(
     dst_state_batch_indices=None,
     pad_slot_id=PAD_SLOT_ID,
     out=None,
+    inplace_final_state=False,
+    num_accepted_tokens=None,
+    cu_seqlens=None,
```
tlrmchlsmth: Could you add a comment explaining these? Especially `inplace_final_state`, as I'm not clear on when to set it

tlrmchlsmth (commented):
Could you share lm_eval gsm8k results and some e2e performance results using `vllm serve` for a mamba2 model?

I'd especially like to make sure the changes to `mamba_ssm.py` aren't causing a performance regression

vllm/model_executor/layers/mamba/ops/mamba_ssm.py
tlrmchlsmth: Appreciate the explanation. That helps explain what's going on quite a bit.

Should we ditch the variable `T` in order to simplify the code?

vllm/model_executor/layers/mamba/ops/mamba_ssm.py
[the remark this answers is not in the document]
tlrmchlsmth: >Or assert that if num_accepted_tokens is passed, then state_batch_indices is 2D?

Thanks for checking. I'm in favor of adding that assert if it doesn't break CUDA Graphs - otherwise a comment would be good to add!

vllm/model_executor/layers/mamba/ops/mamba_ssm.py
roikoren755: Sounds good, removed it

vllm/model_executor/layers/mamba/ops/mamba_ssm.py
roikoren755: `inplace_final_state`, in our upcoming follow-up PRs, was only ever set to `True` in spec-decode cases. Because only in those cases `num_accepted_tokens` is always passed, I removed the argument altogether, and am relying on `IS_SPEC_DECODING` in the same way. Hopefully it'll be clearer now 👍

roikoren755:
> Could you share lm_eval gsm8k results and some e2e performance results using `vllm serve` for a mamba2 model?
> 
> I'd especially like to make sure the changes to `mamba_ssm.py` aren't causing a performance regression

Running the server on an H100 GPU with `vllm serve nvidia/NVIDIA-Nemotron-Nano-9B-v2 --trust-remote-code`, I got the following results.
GSM8K was ran with `lm_eval --model local-completions --tasks gsm8k --model_args "model=nvidia/NVIDIA-Nemotron-Nano-9B-v2,base_url=http://127.0.0.1:8000/v1/completions,num_concurrent=4,max_retries=3,tokenized_requests=False" --cache_requests true --batch_size 256`.
E2E with `vllm bench serve --backend vllm --model nvidia/NVIDIA-Nemotron-Nano-9B-v2 --endpoint /v1/completions --dataset-name random --num-prompts 10`.

**Branch:**
GSM8K:
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.6907|±  |0.0127|
|     |       |strict-match    |     5|exact_match|↑  |0.7309|±  |0.0122|
```

E2E:
```
Successful requests:                     10
Benchmark duration (s):                  1.51
Total input tokens:                      10230
Total generated tokens:                  1280
Request throughput (req/s):              6.61
Output token throughput (tok/s):         845.74
Peak output token throughput (tok/s):    749.00
Peak concurrent requests:                10.00
Total Token throughput (tok/s):          7605.09
Mean TTFT (ms):                          243.59
Median TTFT (ms):                        258.26
P99 TTFT (ms):                           292.79
Mean TPOT (ms):                          9.93
Median TPOT (ms):                        9.81
P99 TPOT (ms):                           11.28
Mean ITL (ms):                           9.93
Median ITL (ms):                         9.59
P99 ITL (ms):                            10.05
```

**Main:**
GSM8K:
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.6907|±  |0.0127|
|     |       |strict-match    |     5|exact_match|↑  |0.7309|±  |0.0122|
```

E2E:
```
Successful requests:                     10
Benchmark duration (s):                  1.53
Total input tokens:                      10230
Total generated tokens:                  1280
Request throughput (req/s):              6.55
Output token throughput (tok/s):         838.61
Peak output token throughput (tok/s):    739.00
Peak concurrent requests:                10.00
Total Token throughput (tok/s):          7540.93
Mean TTFT (ms):                          247.53
Median TTFT (ms):                        262.17
P99 TTFT (ms):                           297.03
Mean TPOT (ms):                          10.00
Median TPOT (ms):                        9.88
P99 TPOT (ms):                           11.37
Mean ITL (ms):                           10.00
Median ITL (ms):                         9.66
P99 ITL (ms):                            10.20
```

tlrmchlsmth (approved):

PR #29490, opened 2025-11-26
[compile] add sequence parallelism for moe

### LINKED ISSUE #29139: [Feature]: Optimize collectives in TP MoE case using torch.compile pass

To avoid redundant work in MoE models in the TP case, sequence parallelism was added to the Deepseek model definition in #24134 and expanded to other models in #24982. However, to avoid performing surgery on the linear layer, the current approach performs more communication than necessary. With a torch.compile custom pass, we can rewrite the graph to remove the redundant computation.

### More details

Before the SP optimization, the ops in the model were:
```
- o_proj:[num_tokens, ...] -> [num_tokens, ...] (incomplete results)
- all_reduce:[num_tokens, ...] -> [num_tokens, ...]
- rms_norm:[num_tokens, ...] -> [num_tokens, ...]
- router:[num_tokens, ...] -> [num_tokens, ...]
- experts:[num_tokens, ...] -> [num_tokens, ...]
- ...
```

With sequence parallel enabled, this becomes:
```
- o_proj: [num_tokens, ...] -> [num_tokens, ...] (incomplete results)
- all_reduce: [num_tokens, ...] -> [num_tokens, ...]
- rms_norm:[num_tokens, ...] -> [num_tokens, ...]
- chunk: [num_tokens, ...] -> [num_tokens/tp, ...]
- router: [num_tokens/tp, ...] -> [num_tokens/tp, ...]
- experts: [num_tokens/tp, ...] -> [num_tokens/tp, ...]
- all_gather: [num_tokens/tp, ...] -> [num_tokens, ...]
```

Additionally, experts now properly do the dp+tp<->ep dispatch instead of just the original replicated dp<->ep dispatch.

Notice that the `all_reduce` (and `rms_norm`) do redundant communication as each TP rank only requires partial results. With a compile pass, we can convert the `all_reduce` -> `rms_norm` -> `chunk` sequence into a `reduce_scatter` -> `rms_norm`:

```
- o_proj: [num_tokens, ...] -> [num_tokens, ...] (incomplete results)
- reduce_scatter: [num_tokens, ...] -> [num_tokens/tp, ...]
- router: [num_tokens/tp, ...] -> [num_tokens/tp, ...]
- experts: [num_tokens/tp, ...] -> [num_tokens/tp, ...]
- all_gather: [num_tokens/tp, ...] -> [num_tokens, ...]
```

We should create a new `SequenceParallelismMoEPass`, controlled by a new `PassConfig.enable_sp_moe` flag (following the new naming convention in #27995) so that it can be turned on independently of regular SP. We will likely need to pad the number of tokens to a multiple of TP size, although like described in #29136, there are alternatives.

#### SP
Additionally, to make SP-MoE compatible with AsyncTP, the SP pass should also replace the `all_gather` -> `rms_norm` sequence with `rms_norm` -> `all_gather`, such that the `all_gather` can be fused with the following QKV proj. We can see that in the SP case we perform an `all_gather` instead of `all_reduce` after the MLP layer.

### Alternatives

Alternatively, the original optimization could be done as a compile pass as well, which would significantly clean up the MoE model definitions. However, that would mean that `VLLM_COMPILE` compilation mode would be required for this optimization and if compilation is disabled, the optimization would be disabled as well. Generally we accept lower performance in eager mode as compilation is on by default, but I know there was a reason this was done this way (don't remember why).

We could also just let the SP pass first transform `all_reduce` -> `rms_norm` into `reduce_scatter` -> `rms_norm` -> `all_gather` and then eliminate the `all_gather` -> `chunk` combo. But then we have less control over SP after MoE (although it seems like all_reduce is not even used in that case).

### Additional context

Original proposal comment: https://github.com/vllm-project/vllm/pull/24982#pullrequestreview-3259494618

cc @tlrmchlsmth @bnellnm @robertgshaw2-redhat @alexm-redhat @zou3519 @nvpohanh @youkaichao 

anant37289:
## Purpose
Tranform a sequence of `allreduce->chunk` to `reduce_scatter` for moe models with `sequence_parallel_chunk` ops. Resolves #29139 
## Test Plan
`pytest -v tests/compile/distributed/test_sequence_parallelism_moe.py`
## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/compilation/sequence_parallelism_moe.py
+++ b/vllm/compilation/sequence_parallelism_moe.py
@@ -4,7 +4,6 @@
 import torch
 import torch._inductor.pattern_matcher as pm
 import torch.fx as fx
-import torch.nn as nn
 from torch._inductor.pattern_matcher import PatternMatcherPass
 
 from vllm.config import VllmConfig
@@ -61,11 +60,10 @@ class AllReduceMoePattern(_SequenceParallelMOEPatternHelper):
             return chunks
 
         def replacement(input: torch.Tensor):
-            seq_len = input.size(0)
-            pad_len = (-seq_len) % self.tp_size
-            y = nn.functional.pad(input, (0, 0, 0, pad_len))
+            ### no need to pad as we that the input shape is always padded
+            reduce_scatter = self._reduce_scatter(input)
 
-            return y
+            return reduce_scatter
 
         pm.register_replacement(
             pattern, replacement, self.get_inputs(), pm.fwd_only, pm_pass
@@ -95,19 +93,9 @@ class SequenceParallelismMoEPass(VllmPatternMatcherPass):
         self.dump_patterns(config, self.patterns)
 
     def is_applicable(self, shape: int | None) -> bool:
-        # When sequence parallelism is enabled, the residual tensor from RMSNorm
-        # needs to be split along the sequence dimension. However, this dimension
-        # is symbolic during piecewise compilation, and splitting symbolic shapes
-        # is not supported.
-        #
-        # This pass is therefore only applied when the sequence dimension is
-        # concrete:
-        # 1. In full-graph compilation mode (no Dynamo splitting ops are used).
-        #   For this case we always pad num_tokens to be a multiple of
-        #   tensor_parallel_size, so there's no need to check shape % tp_size == 0.
-        # 2. For specific shape provided during compilation (e.g., from
-        #    `compile_sizes`), which must be divisible by the tensor-parallel
-        #    size.
+        # This pass is only applied when the sequence dimension is
+        # concrete(In full graph compilation or specific shape
+        # provided during compilation)
         if (
             not self.compilation_config.splitting_ops
             or self.compilation_config.use_inductor_graph_partition
--- a/vllm/compilation/sequence_parallelism_moe.py
+++ b/vllm/compilation/sequence_parallelism_moe.py
@@ -62,15 +62,10 @@ class AllReduceMoePattern(_SequenceParallelMOEPatternHelper):
 
         def replacement(input: torch.Tensor):
             seq_len = input.size(0)
-            remainder = seq_len % self.tp_size
-            if remainder != 0:
-                pad_len = self.tp_size - remainder
-                y = nn.functional.pad(input, (0, 0, 0, pad_len))
-            else:
-                y = input
-            reduce_scatter = self._reduce_scatter(y)
+            pad_len = (-seq_len) % self.tp_size
+            y = nn.functional.pad(input, (0, 0, 0, pad_len))
 
-            return reduce_scatter
+            return y
 
         pm.register_replacement(
             pattern, replacement, self.get_inputs(), pm.fwd_only, pm_pass
--- a/vllm/compilation/sequence_parallelism_moe.py
+++ b/vllm/compilation/sequence_parallelism_moe.py
@@ -1,0 +1,128 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+import torch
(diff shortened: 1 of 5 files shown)


vllm/compilation/sequence_parallelism_moe.py
```diff
@@ -0,0 +1,128 @@
...
+            seq_len = input.size(0)
+            remainder = seq_len % self.tp_size
+            if remainder != 0:
+                pad_len = self.tp_size - remainder
+                y = nn.functional.pad(input, (0, 0, 0, pad_len))
+            else:
+                y = input
+            reduce_scatter = self._reduce_scatter(y)
+
+            return reduce_scatter
```
ProExpertProg: Try this:

```
pad_len = (-input.size(0)) % self.tp_size
y = nn.functional.pad(input, (0, 0, 0, pad_len))
reduce_scatter = self._reduce_scatter(y)
```

vllm/config/compilation.py
```diff
@@ -107,6 +107,8 @@ class PassConfig:
     """Whether to enable the custom no-op elimination pass."""
     enable_sequence_parallelism: bool = False
     """Whether to enable sequence parallelism."""
+    enable_sp_moe: bool = False
```
ZJY0516: Could we unify `enable_sp_moe` and `enable_sequence_parallelism` into a single `enable_sp` configuration? @ProExpertProg

ProExpertProg: I think no because MoE models have some dense layers as well, so the two flags do different things. With optimization levels that just landed, we will enable both by default anyway so having separate flags is better in my opinion to give users more control and for easier debugging & perf impact measurements.

ProExpertProg (commented):
Looks good overall! Can you:
1. Add a test case to `test_e2e_fusions.py`
2. Report lm_eval numbers?
3. Report performance numbers?

For perf and accuracy evals, I would look at the following models:
- Deepseek R1
- Qwen 3 30B
- Llama4

As well as the following deployment scenarios:
- TP=4
- DP=2 TP=2 `--enable-expert-parallelism`
- DP=2 TP=4 `--enable-expert-parallelism`
- DP=4 TP=2 `--enable-expert-parallelism`

anant37289:
Hi so i actually did not have the compute beyond 2 24 GB gpus to do this, so the earlier small models ran file but it was not able to support even the Qwen3-30b-A8b model. I tried renting 2xA40s, but I am running into a couple issues:
1. The sequence parallel chunking from earlier is only applied when TP size > 1 and DP size > 1, I do not see the reason for them both to be atleast 2. Am i missing something?
```
    @property
    def use_sequence_parallel_moe(self) -> bool:
        return (
            self.all2all_backend
            in (
                "allgather_reducescatter",
                "naive",
                "deepep_high_throughput",
                "deepep_low_latency",
            )
            and self.enable_expert_parallel
            and self.tensor_parallel_size > 1
            and self.data_parallel_size > 1
        )
```

2. I tried to make it work with TP=2 and DP=1. but it errors out, probably I am wrong in doing this(I am yet to test on TP=2, DP=2), so i wanted to confirm this once.

ProExpertProg:
Okay, sounds good, please test whatever you can on the hardware you have. Once we're ready to merge I can help run some tests (I am away for 2 weeks so will be after that).

I didn't realize we didn't do MoE SP unless DP>1 (TP>1 makes sense). Feel free to only test TP>1 & DP>1 in that case.

anant37289:
Hi @ProExpertProg 
> I didn't realize we didn't do MoE SP unless DP>1 (TP>1 makes sense). Feel free to only test TP>1 & DP>1 in that case.

About this. Is there a reason for doing this? After merging from main I was able to get sequence parallel chunking from earlier to work with TP=2 and DP=1(by just changing the above mention `use_sequence_parallel_moe` function(checking only for TP).

PR #29385, opened 2025-11-25
[CI] Add Async Eplb nightly CI tests

david6666666:
## Purpose
#22179 comment  Add Async Eplb nightly CI
  * DeepSeek V2-Lite Async EPLB Accuracy
  * Qwen3-Next-80B-A3B-Instruct MTP Async EPLB Accuracy
## Test Plan
```
bash .buildkite/scripts/scheduled_integration_test/deepseek_v2_lite_ep_async_eplb.sh 0.25 1319 8030
bash .buildkite/scripts/scheduled_integration_test/qwen3_next_mtp_async_eplb.sh 0.8 1319 8040
```
## Test Result
```
/tmp/vllm-scheduled/deepseek-ai_DeepSeek-V2-lite_deepep_high_throughput_async_eplb.json
{
  "accuracy": 0.3601213040181956,
  "invalid_rate": 0.006065200909780136,
  "latency": 85.7941705584526,
  "questions_per_second": 15.37400491681832,
  "total_output_tokens": 151616,
  "tokens_per_second": 1767.206314987359,
  "num_questions": 1319,
  "num_shots": 5,
  "max_tokens": 256,
  "timestamp": 1764056573.579663
}
/tmp/vllm-scheduled/deepseek-ai_DeepSeek-V2-lite_deepep_low_latency_async_eplb.json
{
  "accuracy": 0.3646702047005307,
  "invalid_rate": 0.006065200909780136,
  "latency": 73.81600442156196,
  "questions_per_second": 17.86875367118509,
  "total_output_tokens": 152375,
  "tokens_per_second": 2064.2542385495285,
  "num_questions": 1319,
  "num_shots": 5,
  "max_tokens": 256,
  "timestamp": 1764056759.8287513
}
```

```
/tmp/vllm-scheduled/Qwen__Qwe_3-Next-80B-A3B-I_struct_deepep_high_throughput.json
{
  "accuracy": 0.8658074298711145,
  "invalid_rate": 0.0,
  "latency": 132.54091829434037,
  "questions_per_second": 9.951643741224348,
  "total_output_tokens": 209727,
  "tokens_per_second": 1582.356623893676,
  "num_questions": 1319,
  "num_shots": 5,
  "max_tokens": 256,
  "timestamp": 1764058291.3716583
}
/tmp/vllm-scheduled/Qwen__Qwe_3-Next-80B-A3B-I_struct_deepep_low_latency.json
{
  "accuracy": 0.8605003790750568,
  "invalid_rate": 0.0,
[...]

PR #29499, opened 2025-11-26
[EPLB] Optimize EPLB with numpy

ilmarkov:
This PR adds following optimizations in EPLB algorithm which are applicable to sync and async mode:
- Removes multiple gpu-cpu transfers in `move_from_buffer`.
- Implements `get_ep_ranks_with_expert`, `move_to_buffer`, `move_from_buffer` in numpy which gives 20-25% on average for each primitive
- Adds `preserve_intragpu_slots` primitive that does post processing rebalance algo results. It ensures that the experts that are assigned to be moved within the same gpu, stay in their slots. It helps avoid unnecessary gpu memcopies.

Adds `log_balancedness_interval` config parameter to tune the frequency of balancedness logs and possible reduce number of collectives required for this logging.

## Purpose
Improves performance of sync and async eplb. Fixes bug in EPLB config post processing.

## Test Plan
Add tests for grouped layer weight exchange, tests for `preserve_intragpu_slots`. 

## Validation Result
Client 
`lm_eval --model local-completions --tasks gsm8k --model_args model=$model,base_url=http://0.0.0.0:$port/v1/completions,num_concurrent=50,max_retries=3,tokenized_requests=False`

Main
Qwen-30B-FP8 DP=8
```
vllm serve Qwen/Qwen3-30B-A3B-Instruct-2507-FP8 --disable-log-requests --no-enable-prefix-caching -tp 1 -dp 8 -enable-eplb --eplb-config.window_size 128 --eplb-config.step_interval 512 --eplb-config.num_redundant_experts 32 --eplb-config.use_async true
```

```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8939|±  |0.0085|
|     |       |strict-match    |     5|exact_match|↑  |0.8870|±  |0.0087|
```

Qwen-30B-FP8 DP=8 AsyncEPLB=True
```
vllm serve Qwen/Qwen3-30B-A3B-Instruct-2507-FP8 --disable-log-requests --no-enable-prefix-caching -tp 1 -dp 8 --max-num-seqs 256 --enable-expert-parallel --port $port --gpu-memory-utilization 0.8 --enable-eplb --eplb-config.window_size 128 --eplb-config.step_interval 512 --eplb-config.num_redundant_experts 32 --eplb-config.use_async true
```

```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8848|±  |0.0088|
|     |       |strict-match    |     5|exact_match|↑  |0.8764|±  |0.0091|
```

Qwen-30B-FP8 DP=8 AsyncEPLB=False
```
vllm serve Qwen/Qwen3-30B-A3B-Instruct-2507-FP8 --disable-log-requests --no-enable-prefix-caching -tp 1 -dp 8 --max-num-seqs 256 --enable-expert-parallel  --gpu-memory-utilization 0.8 --enable-eplb --eplb-config.window_size 128 --eplb-config.step_interval 512 --eplb-config.num_redundant_experts 32 --eplb-config.use_async false
```

```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8855|±  |0.0088|
|     |       |strict-match    |     5|exact_match|↑  |0.8719|±  |0.0092|
```

---

Main
DeepSeek-V2-Lite DP=4
`vllm serve deepseek-ai/DeepSeek-V2-Lite --disable-log-requests --no-enable-prefix-caching -tp 1 -dp 4 --enable-eplb --eplb-config.window_size 128 --eplb-config.step_interval 512 --eplb-config.num_redundant_experts 16 --eplb-config.use_async true`

```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.3813|±  |0.0134|
|     |       |strict-match    |     5|exact_match|↑  |0.3776|±  |0.0134|
```

DeepSeek-V2-Lite DP=4 AsyncEPLB=True
```
vllm serve deepseek-ai/DeepSeek-V2-Lite --disable-log-requests --no-enable-prefix-caching -tp 1 -dp 4 --max-num-seqs 256 --enable-expert-parallel --gpu-memory-utilization 0.9 --enable-eplb --eplb-config.window_size 128 --eplb-config.step_interval 512 --eplb-config.num_redundant_experts 16 --eplb-config.use_async true
```

```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.3813|±  |0.0134|
|     |       |strict-match    |     5|exact_match|↑  |0.3768|±  |0.0133|
```

DeepSeek-V2-Lite DP=4 AsyncEPLB=False
```
vllm serve deepseek-ai/DeepSeek-V2-Lite --disable-log-requests --no-enable-prefix-caching -tp 1 -dp 4 --max-num-seqs 256 --enable-expert-parallel --gpu-memory-utilization 0.9 --enable-eplb --eplb-config.window_size 128 --eplb-config.step_interval 512 --eplb-config.num_redundant_experts 16 --eplb-config.use_async false
```

```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.3783|±  |0.0134|
|     |       |strict-match    |     5|exact_match|↑  |0.3745|±  |0.0133|
```

## Benchmark results

Server:
```
vllm serve Qwen/Qwen3-Next-80B-A3B-Instruct --disable-log-requests --no-enable-prefix-caching -dp 8 -tp 1 --max-num-seqs 256 --enable-expert-parallel --gpu-memory-utilization 0.9 --enable-eplb --eplb-config.window_size 32 --eplb-config.step_interval 128 --eplb-config.num_redundant_experts 128 --eplb-config.use_async false --max-model-len 4096
```

We compare sync versions of EPLB. Profile the third EPLB call, to get more stable version.
In profile logs
Main: eplb_state.py: rearrange takes 1s370ms
PR: eplb_state.py: rearrange takes 1s77ms. ~27% speedup

Profile logs of move_to_buffer and move_from_buffer

Main:

Screenshot 2025-11-26 at 15 31 44

PR:
Screenshot 2025-11-26 at 15 32 12

 
---
<details>

</details>


david6666666:
LGTM.
EPLB config post processing were re-written by default values also occur in #29385, thx for fix and optimization.

vllm/engine/arg_utils.py
```diff
@@ -419,10 +419,10 @@ class EngineArgs:
...
     _api_process_count: int = ParallelConfig._api_process_count
     _api_process_rank: int = ParallelConfig._api_process_rank
-    num_redundant_experts: int = EPLBConfig.num_redundant_experts
-    eplb_window_size: int = EPLBConfig.window_size
-    eplb_step_interval: int = EPLBConfig.step_interval
-    eplb_log_balancedness: bool = EPLBConfig.log_balancedness
+    num_redundant_experts: int | None = None
+    eplb_window_size: int | None = None
+    eplb_step_interval: int | None = None
+    eplb_log_balancedness: bool | None = None
```
hmellor: Why change these defaults and types?

ilmarkov: Otherwise, eplb_config options are overwritten

hmellor: You're right, I now know that there was a bug and that shouldn't have been happening.

It's since been fixed on `main` so you should be ok to revert the changes to this file now.

vllm/distributed/eplb/rebalance_execute.py
```diff
@@ -237,38 +364,98 @@ def move_to_buffer(
...
     # wait for the communication to finish
-    return is_unchanged, is_received_locally, experts_recv_loc
+    return (
+        is_unchanged,
+        is_received_locally,
+        (recv_primary_mask, recv_counts, recv_expert_ids, recv_dst_rows),
+    )
 
 
 def move_from_buffer(
```
SageMoore, line 357: Let's get a doc string comment on this and `move_to_buffer`

vllm/distributed/eplb/rebalance_execute.py
```diff
@@ -281,7 +468,9 @@ async def transfer_layer(
     layer: int = 0,
     cuda_stream: torch.cuda.Stream | None = None,
     rank_mapping: dict[int, int] | None = None,
-) -> tuple[list[bool], list[bool], dict[int, int]]:
+) -> tuple[
```
SageMoore: Nit: Let's make a type annotation for this. We can also use it in `move_to_buffer`

vllm/distributed/eplb/rebalance_execute.py
```diff
@@ -237,38 +364,98 @@ def move_to_buffer(
...
-    expert_weights_buffer: list[torch.Tensor],
-    is_unchanged: list[bool],
-    is_received_locally: list[bool],
-    experts_recv_loc: dict[int, int],
-    new_indices: Sequence[int],
+    weights_group: Sequence[Iterable[torch.Tensor]],
+    buffers_group: Sequence[Sequence[torch.Tensor]],
+    is_unchanged: np.ndarray,
+    is_received_locally: np.ndarray,
+    recv_metadata: tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray],
```
SageMoore: Nit: Let's add a `@dataclass` for `recv_metadata`.

vllm/distributed/eplb/eplb_state.py
```diff
@@ -968,17 +981,21 @@ def move_to_workspace(
...
+                weights_group=weights_group,
+                buffers_group=buffers_group,
                 is_unchanged=model_state.is_unchanged,
                 is_received_locally=model_state.is_received_locally,
-                experts_recv_loc=model_state.experts_recv_loc,
-                new_indices=model_state.new_physical_to_logical_map[
-                    model_state.layer_to_transfer
-                ].tolist(),
+                recv_metadata=model_state.recv_metadata,
+                new_indices_group=model_state.new_physical_to_logical_map[
```
SageMoore: Nit: For readability, can you just create an local variable and pass it to this function?

SageMoore (commented):
Nice work, @ilmarkov. These speedups are quite exciting. I'm still going through the code, but here are some initial, largely cosmetic, comments.

vllm/distributed/eplb/policy/default.py
```diff
@@ -207,6 +207,110 @@ def inverse(perm: torch.Tensor) -> torch.Tensor:
...
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        """
+        Reorder the new mapping per GPU so that experts that remain on the same GPU
+        keep their previous slot positions when possible. Incoming experts to that GPU
+        fill any remaining available slots. This is applied only when the number of GPUs
+        is unchanged and the slots per GPU remain the same between
+        the old and new mappings.
+        """
+        device = phy2log.device
+        new_num_phy = phy2log.shape[1]
```
SageMoore: Nit: I assume the only time `old_num_phy` will be different then `new_num_phy` is when we are scaling up/down in elastic EP. Is that the case? If so can you add a brief comment explaining that.

SageMoore: Nit: Let's just call this `num_physical_experts`.

vllm/distributed/eplb/policy/default.py
```diff
@@ -207,6 +207,110 @@ def inverse(perm: torch.Tensor) -> torch.Tensor:
...
+            num_ranks <= 0
+            or new_num_phy % num_ranks != 0
+            or old_num_phy % num_ranks != 0
+            or (new_num_phy // num_ranks) != (old_num_phy // num_ranks)
+        ):
+            return phy2log, phyrank
+
+        # Move to CPU and convert to NumPy for processing
+        phy2log_np = phy2log.cpu().numpy()
+        phyrank_np = phyrank.cpu().numpy()
```
SageMoore: Not part of this PR, but I really feel like we need a better name for `phyrank`. Something like `replica_idxs`? Unless I'm misunderstanding it's purpose, it doesn't have anything to do with the physical experts or GPU ranks.

vllm/distributed/eplb/policy/default.py
```diff
@@ -207,6 +207,110 @@ def inverse(perm: torch.Tensor) -> torch.Tensor:
...
+        old_np = old_global_expert_indices.cpu().numpy()
```
SageMoore: Is this a logical->physical mapping or physical->logical mapping? It also needs a better name than `old_np` :)

vllm/distributed/eplb/policy/default.py
```diff
@@ -207,6 +207,110 @@ def inverse(perm: torch.Tensor) -> torch.Tensor:
...
+        slots_per_gpu = new_num_phy // num_ranks
+        num_layers = phy2log_np.shape[0]
+
+        post_phy2log_np = phy2log_np.copy()
+        post_phyrank_np = phyrank_np.copy()
+
+        for gpu_idx in range(num_ranks):
+            start = gpu_idx * slots_per_gpu
+            end = start + slots_per_gpu
+            # Segments across all layers for this GPU
```
SageMoore: I think that introducing the term `segment` is a bit confusing here. I think the prefix `local_` on the per-gpu slices would be a bit more readable? Let's also avoid shortening the dimension names in the comments.

vllm/distributed/eplb/policy/default.py
```diff
@@ -207,6 +207,110 @@ def inverse(perm: torch.Tensor) -> torch.Tensor:
...
+            old_seg = old_np[:, start:end]  # [L, S]
+            new_seg = phy2log_np[:, start:end]  # [L, S]
+            new_rnk = phyrank_np[:, start:end]  # [L, S]
+
+            used_new_indices = np.zeros((num_layers, slots_per_gpu), dtype=bool)
+            preserved_positions = np.zeros((num_layers, slots_per_gpu), dtype=bool)
+
+            # First pass: preserve same-logical experts in their previous slots
+            for pos in range(slots_per_gpu):
```
SageMoore: Nit: `slot_idx` instead of `pos`

SageMoore (commented):
I'm still going through the code, but here's another round of comments.

vllm/distributed/eplb/rebalance_execute.py
```diff
@@ -237,38 +342,95 @@ def move_to_buffer(
...
 
 
 def move_from_buffer(
     expert_weights: Iterable[torch.Tensor],
-    expert_weights_buffer: list[torch.Tensor],
-    is_unchanged: list[bool],
-    is_received_locally: list[bool],
-    experts_recv_loc: dict[int, int],
-    new_indices: Sequence[int],
+    expert_weights_buffers: list[torch.Tensor],
```
SageMoore, line 359: Nit: How do you feel about changing the name of this to `intermediate_expert_weights`?

vllm/distributed/eplb/rebalance_execute.py
```diff
@@ -237,38 +342,95 @@ def move_to_buffer(
...
+    recv_metadata: RecvMetadata,
+    new_indices: np.ndarray,
     ep_group: ProcessGroup,
 ) -> None:
+    """
+    Copies expert weights from communication buffers back to the target weight tensors
+    after EPLB rebalancing.
+
+    Args:
+        expert_weights: List of weight tensors for a MoE layer.
```
SageMoore: Nit: Let's make it clearer that these are references to the actual weights of the model.

`expert_weights: List of weight tensors used by the model during execution of MoE layers`

vllm/distributed/eplb/rebalance_execute.py
```diff
@@ -237,38 +342,95 @@ def move_to_buffer(
...
+        expert_weights_buffers: Intermediate buffers matching ``expert_weights``.
```
SageMoore: Nit: `expert_weights_buffers: Intermediate buffers containing weights that need to be transferred to expert_weights`

vllm/distributed/eplb/rebalance_execute.py
```diff
@@ -237,38 +342,95 @@ def move_to_buffer(
...
+        is_unchanged: (num_local_experts,), True where an expert row is unchanged.
+        is_received_locally: (num_local_experts,), True where a row is updated locally.
```
SageMoore, line 375: How does `is_received_locally` interact with your `preserve_intragpu_slots`? Doesn't `preserve_intragpu_slots` make `is_received_locally` redundant?

vllm/distributed/eplb/rebalance_execute.py
```diff
@@ -237,38 +342,95 @@ def move_to_buffer(
...
+        recv_metadata: RecvMetadata containing remote receive metadata.
+        new_indices: (num_experts_total,) mapping from local rows to desired
+            (possibly global) expert id, after rebalance.
+        ep_group: torch.distributed.ProcessGroup for expert parallel communication
```
SageMoore: Can we ditch this argument and just pass in the rank for simplicity?

vllm/distributed/eplb/rebalance_execute.py
```diff
@@ -237,38 +342,95 @@ def move_to_buffer(
...
+    # copy if locally received OR remote primary recv
+    copy_mask = np.logical_or(is_received_locally, recv_primary_mask)
+    dest_mask_np = np.logical_and(~is_unchanged, copy_mask)
+    if bool(dest_mask_np.any()):
+        dest_indices = np.nonzero(dest_mask_np)[0].tolist()
+        for dst in dest_indices:
+            for w, b in zip(expert_weights, expert_weights_buffers):
+                w[dst].copy_(b[dst], non_blocking=True)
+
+    # Duplicate remote received rows to non-primary duplicate dsts
```
SageMoore: I don't understand this comment. Isn't this essentially:
`# If this rank hasn't had any remote experts transferred to it, return`

Can we move this check to before the copy logic on line 390?

vllm/distributed/eplb/rebalance_execute.py
```diff
@@ -237,38 +342,95 @@ def move_to_buffer(
...
 
-    local2global = partial(
-        idx_local_to_global, local_cnt=num_local_experts, ep_rank=ep_rank
+    base = ep_rank * num_local_experts
+    local_experts = new_indices[base + np.arange(num_local_experts, dtype=np.int32)]
+    duplicate_mask = np.logical_and(
+        np.logical_and(~is_unchanged, ~is_received_locally),
+        np.logical_and(~recv_primary_mask, local_experts != -1),
     )
+    if not bool(duplicate_mask.any()):
```
SageMoore, line 408: A comment would be helpful here.  IIUC this is essentially handling the case where all physical experts on this rank have already been handled by previous logic and we can abort early.

vllm/distributed/eplb/rebalance_execute.py
ilmarkov: I would avoid the word weights as a main noun in this variable. I find `buffers` a better naming, we have matching funcs like move_to_buffer/move_from_buffer

vllm/distributed/eplb/rebalance_execute.py
ilmarkov: It still can be used in case of scale up/down and possible duplicates handling. For example, we can have an expert N on the rank before and two experts N on the same rank after the rebalance. `preserve_intragpu_slots` will ensure that the expert will be on the same position as before but we need to copy duplicate.

vllm/distributed/eplb/rebalance_execute.py
ilmarkov: The comment was in the wrong place

ilmarkov: We can't move it before the copy logic. `recv_count` represents the number of the experts we receive from remote. In the copy logic we also handle local experts

vllm/distributed/eplb/rebalance_execute.py
SageMoore: I see. Thanks for the explanation

vllm/distributed/eplb/policy/default.py
```diff
@@ -203,9 +203,109 @@ def inverse(perm: torch.Tensor) -> torch.Tensor:
...
+                        layer_indices, matched_new_positions
+                    ]
+                    post_phy_replicas_idx_np[layer_indices, start + slot_idx] = (
+                        new_ridx[layer_indices, matched_new_positions]
+                    )
+                    used_new_indices[layer_indices, matched_new_positions] = True
+                    preserved_positions[layer_indices, slot_idx] = True
+
+            # Second pass: fill remaining slots with remaining new experts
+            remaining_mask = ~used_new_indices  # [L, S]
```
SageMoore: Nit `[layers, slots]`

tests/distributed/test_eplb_algo.py
```diff
@@ -310,3 +310,136 @@ def test_additional_cases():
...
+    #   GPU0 -> [0, 1, 0, 2, 3]  (expert 0 duplicated)
+    #   GPU1 -> [4, 5, 6, 1, 2]
+    old_global_expert_indices = torch.tensor([[0, 1, 0, 2, 3, 4, 5, 6, 1, 2]])
+    # New mapping reorders within GPUs and moves some experts across GPUs,
+    # while still including duplicates:
+    #   GPU0 new -> [0, 5, 4, 0, 1]  (expert 0 duplicated, 4/5 incoming)
+    #   GPU1 new -> [6, 2, 3, 1, 2]  (expert 2 duplicated)
+    phy2log = torch.tensor([[0, 5, 4, 0, 1, 6, 2, 3, 1, 2]])
+    # Derive ranks so duplicates have ranks [0,1,...] by occurrence
+    phyrank = _make_phyrank_from_phy2log(phy2log)
```
SageMoore: Nit: lets update the names here.

tests/distributed/test_eplb_algo.py
```diff
@@ -310,3 +310,136 @@ def test_additional_cases():
...
+        phy2log,
+        phyrank,
+        post_phy2log,
+        post_phyrank,
+        num_ranks,
+        slots_per_gpu,
+    )
+
+
+def test_preserve_intragpu_slots_with_duplicates():
```
SageMoore: Can you add a test that's just one expert replicated 8 times across 2 GPUs?
```
#   GPU0 -> [0, 0, 0, 0, 0]
#   GPU1 -> [0, 0, 0, 0, 0]
```

SageMoore (commented):
Ok I think this is getting pretty close. Given the complex nature of this PR, it would be good to include lm_eval runs for both `Qwen/Qwen3-30B-A3B-Instruct-2507-FP8` and `deepseek-ai/DeepSeek-V2-Lite` with sync and async EPLB.

tests/distributed/test_eplb_algo.py
ilmarkov: We still need to have other experts at least a single instance of each on a third GPU. EP won't work otherwise

SageMoore (approved):

vllm/distributed/eplb/eplb_state.py
```diff
@@ -585,9 +598,10 @@ def step(
 
                 if ep_group.rank() == 0:
                     logger.info(
-                        "EPLB step: %d for model %s: avg_tokens=%.2f, "
+                        "EPLB step: %d/%d for model %s: avg_tokens=%.2f, "
                         "max_tokens=%d, balancedness=%.4f",
                         self.expert_rearrangement_step,
+                        self.expert_rearrangement_step_interval,
```
tlrmchlsmth: why log the interval here as well?

tlrmchlsmth (commented):
I am seeing a failure when running test_eplb_execute - @ilmarkov could you please take a look?

```
  File "/home/tms/vllm/tests/distributed/test_eplb_execute.py", line 308, in _test_async_transfer_layer_without_mtp_worker
    move_from_buffer(
TypeError: move_from_buffer() got an unexpected keyword argument 'ep_group'
[rank0]:[W106 22:51:45.365432802 ProcessGroupNCCL.cpp:1524] Warning: WARNING: destroy_process_group() was not called before program exit, which can leak resources. For more info, please see https://pytorch.org/docs/stable/distributed.html#shutdown (function operator())
=============================================== warnings summary ================================================
.venv/lib/python3.12/site-packages/schemathesis/generation/coverage.py:305
  /home/tms/vllm/.venv/lib/python3.12/site-packages/schemathesis/generation/coverage.py:305: DeprecationWarning: jsonschema.exceptions.RefResolutionError is deprecated as of version 4.18.0. If you wish to catch potential reference resolution errors, directly catch referencing.exceptions.Unresolvable.
    ref_error: type[Exception] = jsonschema.RefResolutionError,

============================================ short test summary info ============================================
FAILED tests/distributed/test_eplb_execute.py::test_async_transfer_layer_without_mtp[2-2-2-3] - AssertionError
```

vllm/distributed/eplb/rebalance_execute.py
```diff
@@ -18,213 +19,317 @@
...
+    """Metadata describing remote receives during EPLB rebalancing."""
 
+    recv_primary_mask: np.ndarray
+    """Mask of (num_local_experts,) indicating primary experts received."""
+    recv_count: int
+    """Number of received experts for the layer."""
+    recv_expert_ids: np.ndarray
+    """Expert ids (num_local_experts,) of remote primary experts."""
+    recv_dst_rows: np.ndarray
+    """Target expert indices (num_local_experts,) in local tensors to send."""
```
tlrmchlsmth, line 38: I think that having triple-quoted comments below the variables they are describing is pretty strange. Could you please change this to be more typical?
```
    # Mask of (num_local_experts,) indicating primary experts received.
    recv_primary_mask: np.ndarray
```

vllm/distributed/eplb/eplb_state.py
ilmarkov: So that it's clearer when the next round of EPLB is going to be.

vllm/distributed/eplb/rebalance_execute.py
ilmarkov: I believe that's a docstring formatting we use now in dataclass's field description. Check out [vllm config](https://github.com/vllm-project/vllm/blob/main/vllm/config/vllm.py#L176) or [multimodal processing](https://github.com/vllm-project/vllm/blob/main/vllm/multimodal/processing.py#L98)

vllm/distributed/eplb/eplb_state.py
tlrmchlsmth: I find this to be pretty confusing, especially since there's no explanation for what the first vs the second number means. Could you either revert this or make the log statement more clear?

vllm/distributed/eplb/rebalance_execute.py
tlrmchlsmth: OK, sounds good in that case

vllm/distributed/eplb/eplb_state.py
ilmarkov: Got it. Removed it and added info on the number of steps before the rearrangement

tlrmchlsmth (approved):
LGTM now, thank you!

PR #29503, opened 2025-11-26
[NIXL] Add compatibility checking to NIXL KV connector handshake

markmc:
Related to llm-d/llm-d#499

When prefill and decode instances have incompatible versions or configurations, it is possible that the handshake will fail with unclear errors, NIXL transfers will fail, or even server crashes might occur. This change adds explicit compatibility checks during handshake to fail fast with actionable error messages.
    
The compatibility check is based on the vLLM version, a NIXL connector version number, and a hash of relevant config. All of these must match exactly. We will likely evolve these compatibility factors to be more or less permissive based on feedback.
    
Compatibility metadata is checked before decoding NIXL agent metadata so that agent metadata schema incompatibilities are caught by the compat hash, rather than as a decoding error.
    
An escape hatch is evailable to disable compat hash checking:
    
```
--kv-transfer-config '{"kv_connector_extra_config": {"enforce_handshake_compat": false}}'
```

### DIFF
--- a/vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
+++ b/vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
@@ -59,6 +59,21 @@ Transfer = tuple[int, float]  # (xfer_handle, start_time)
 EngineId = str
 ReqId = str
 
+#
+# NIXL Connector Version
+#
+# Increment this version whenever there is an incompatible change to:
+#   - NixlAgentMetadata schema
+#   - kv_transfer_params schema or semantics
+#   - NIXL transfer protocol or wire format
+#   - KV cache memory layout or block organization
+#   - Any other change that breaks P/D interoperability
+#
+# Version History:
+#   1: Initial version with compatibility checking
+#
+NIXL_CONNECTOR_VERSION: int = 1
+
 GET_META_MSG = b"get_meta_msg"
 
 logger = init_logger(__name__)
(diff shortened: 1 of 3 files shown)


NickLucche (commented):
> we should have an escape hatch flag to disable this in case users have valid config mismatch use cases

Yes I think checking vLLM+nixl version is generally safe.
vllm_config check is too strict unless we take a subset of fields: eg parallel config can safely differ (heterogeneous deployments), as well as block_size (more compat checks are currently in `_validate_handshake`).

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1000,14 +1065,52 @@ def _nixl_handshake(
...
+
+            # Decode handshake payload to get compatibility hash
+            handshake_decoder = msgspec.msgpack.Decoder(NixlHandshakePayload)
+            try:
+                handshake_payload = handshake_decoder.decode(handshake_bytes)
+            except Exception as e:
+                raise RuntimeError(
+                    f"Failed to decode NixlHandshakePayload. This likely indicates "
+                    f"an incompatibility between connector version. Error: {e}"
+                ) from e
```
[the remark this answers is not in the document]
markmc: Fixed

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1000,14 +1065,52 @@ def _nixl_handshake(
...
+            metadata_decoder = msgspec.msgpack.Decoder(NixlAgentMetadata)
+            try:
+                metadata = metadata_decoder.decode(
+                    handshake_payload.agent_metadata_bytes
+                )
+            except Exception as e:
+                # This should not happen if hash matched
+                raise RuntimeError(
+                    f"Failed to decode NixlAgentMetadata. Error: {e}"
+                ) from e
```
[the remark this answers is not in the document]
markmc: Fixed

markmc:
Rebased to pick up #29619

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -97,18 +112,95 @@
...
+        "num_kv_heads": model_config.get_total_num_kv_heads(),
+        "head_size": model_config.get_head_size(),
+        "num_hidden_layers": model_config.get_total_num_hidden_layers(),
+        # Attention backend and KV cache dtype affect memory layout
+        "attn_backend_name": attn_backend_name,
+        "cache_dtype": str(cache_config.cache_dtype),
+    }
+
+    compat_hash = hash_factors(factors)
+    logger.info(
```
njhill, line 191: debug level?

njhill (approved):
Thanks @markmc, it would be good to get a final approval from @NickLucche too...

NickLucche (approved):
I like it, thanks for the work @markmc !
I am pushing the big green button now to unblock follow-up work, but pinging @xuechendi which is most affected by heterogeneous deployments.
We still have time to refine the hash factors in case they're too strict, but at least we have the machinery in place.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
NickLucche: yep this looks debug

xuechendi (approved):

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
markmc: Filed as #30137

PR #29507, opened 2025-11-26
fix hf vllm mapping for mixtral that was breaking due to v5 coming

3outeille:
referencing https://github.com/huggingface/transformers/pull/42429


hmellor:
Closing as obsolete. The hardcoded `block_sparse_moe` -> `mlp` mapping is superseded by #31545, which builds the Transformers modeling backend's `hf_to_vllm_mapper` generically from Transformers v5's own `WeightRenaming` conversion mappings, and Mixtral's `w1/w2/w3` expert naming is handled by `TransformersRoutedExperts.get_expert_mapping` since #47058. The second hunk reverts #28952, which is still the intended behaviour on main. The Transformers-side counterpart huggingface/transformers#42429 was also closed unmerged.

PR #29542, opened 2025-11-26
[BugFix] Fix spec decoding max_tokens scheduling perf issue

njhill:
This affected performance with async scheduling since requests could be unscheduled before they were finished and have to be scheduled again for the last 1-2 tokens.

Thanks @benchislett for finding.


vllm/v1/outputs.py
```diff
@@ -29,27 +29,15 @@ class LogprobsLists(NamedTuple):
...
-                cu_num - cu_num_offset
-                for cu_num in self.cu_num_generated_tokens[
-                    start_req_idx : end_req_idx + 1
-                ]
-            ]
-        else:
-            start = start_req_idx
-            end = end_req_idx
-            sliced_cu_num_generated_tokens = None
+    def slice_request(self, req_idx: int, num_positions: int):
```
njhill, line 32: Changing how logprobs are sliced to ensure we don't include extra positions beyond the number of output tokens.

DarkLight1337 (approved):

PR #29549, opened 2025-11-26
[responsesAPI] support input output messages for non harmony models

qandrew:
## Purpose

The responsesAPI reasoning / tool calling parser may not always be correct (ie Q1 in https://huggingface.co/moonshotai/Kimi-K2-Thinking/blob/main/docs/tool_call_guidance.md). We should provide support to show the raw token in / token out as optional additional payload in responsesAPI.

We already output the "token format" for GPTOSS harmony; this PR extends support for non harmony models, albeit in a more basic string / tokens format.

The structure will be a list of this, which can be extended to include other objects like tool calls, etc in the future

```
class ResponseRawMessageAndToken(OpenAIBaseModel):
    """Class to show the raw message.
    If message / tokens diverge, tokens is the source of truth"""

    message: str
    tokens: list[int]
    type: Literal["raw_message_tokens"] = "raw_message_tokens"
```

## Test Plan

```
vllm serve moonshotai/Kimi-K2-Thinking   --trust-remote-code   --tensor-parallel-size 8   --enable-auto-tool-choice   --max-num-batched-tokens 32768   --tool-call-parser kimi_k2   --reasoning-parser kimi_k2  
```

```
 curl -X POST "http://localhost:8000/v1/responses"   -H "Content-Type: application/json"   -H "Authorization: Bearer dummy-api-key"   -d '{
        "model": "MiniMaxAI/MiniMax-M2",
        "input": [{
          "role": "user",
          "content": "Hello."
          }], "enable_response_messages": true
      }' | jq
```

## Test Result

`
{"id":"resp_9f8e1028bc02e255","created_at":1764196749,"incomplete_details":null,"instructions":null,"metadata":null,"model":"MiniMaxAI/MiniMax-M2","object":"response","output":[{"id":"rs_b224d9f3c67292dc","summary":[],"type":"reasoning","content":[{"text":"The user just greeted me with \"Hello,\" which is a friendly start. I’ll keep my response straightforward and ask how I can assist them today. Since they didn't ask a specific question, I don't need to use any markdown headings or bullet points. I'll simply acknowledge their greeting and invite them to share their needs. This sets a positive tone and encourages them to engage further.\n","type":"reasoning_text"}],"encrypted_content":null,"status":null},{"id":"msg_9d7852c1f62721a7","content":[{"annotations":[],"text":"\n\nHello! How can I help you today?","type":"output_text","logprobs":null}],"role":"assistant","status":"completed","type":"message"}],"parallel_tool_calls":true,"temperature":1.0,"tool_choice":"auto","tools":[],"top_p":0.95,"background":false,"max_output_tokens":196584,"max_tool_calls":null,"previous_response_id":null,"prompt":null,"reasoning":null,"service_tier":"auto","status":"completed","text":null,"top_logprobs":null,"truncation":"disabled","usage":{"input_tokens":24,"input_tokens_details":{"cached_tokens":16,"input_tokens_per_turn":[],"cached_tokens_per_turn":[]},"output_tokens":87,"output_tokens_details":{"reasoning_tokens":0,"tool_output_tokens":0,"output_tokens_per_turn":[],"tool_output_tokens_per_turn":[]},"total_tokens":111},"user":null,"input_messages":{"message":["]~!b[]~b]system\nYou are a helpful assistant.[e~[\n]~b]user\nHello.[e~[\n]~b]ai\n<think>\n"],"tokens":[200034,200019,28463,10,2985,457,258,12473,23413,46,200020,10,200019,3995,10,19739,46,200020,10,200019,1361,10,200050,10]},"output_messages":{"message":["The user just greeted me with \"Hello,\" which is a friendly start. I’ll keep my response straightforward and ask how I can assist them today. Since they didn't ask a specific question, I don't need to use any markdown headings or bullet points. I'll simply acknowledge their greeting and invite them to share their needs. This sets a positive tone and encourages them to engage further.\n</think>\n\nHello! How can I help you today?"],"tokens":[758,3100,1127,71373,640,418,494,19739,2304,797,355,258,18903,1745,46,343,4713,2830,930,4108,29454,306,2874,1124,343,566,7202,1003,4380,46,8796,784,5775,2874,258,3246,3501,44,343,3110,1132,301,1223,949,97205,93065,436,29758,4790,46,13098,5636,27193,805,74664,306,25754,1003,301,4938,805,3887,46,1204,9363,258,5844,18619,306,34776,1003,301,14447,4052,320,200051,367,19739,33,2329,566,343,1576,390,4380,63,200020]}}
`


vllm/entrypoints/openai/protocol.py
```diff
@@ -2229,6 +2229,19 @@ def serialize_messages(msgs):
     return [serialize_message(msg) for msg in msgs] if msgs else None
 
 
+class ResponseRawMessageAndToken(OpenAIBaseModel):
+    """Class to show the raw message.
+    If message / tokens diverge, tokens is the source of truth"""
+
+    message: str
+    tokens: list[int]
```
qandrew: another thought was tokens: list[list[int]] in which we append each time we do a turn for MCP. But I think it's more logical to keep it flatter, clients downstream can parse the different turns.

vllm/entrypoints/context.py
```diff
@@ -155,6 +163,22 @@ def append_output(self, output) -> None:
...
+        if len(self.input_messages) == 0:
+            output_prompt = output.prompt or ""
+            output_prompt_token_ids = output.prompt_token_ids or []
+            self.input_messages.append(
+                ResponseRawMessageAndToken(
+                    message=output_prompt,
+                    tokens=output_prompt_token_ids,
+                )
+            )
+        self.output_messages.append(
```
alecsolder, line 184: If we are appending only the output of generation or tool calls for subsequent turns, I think this will actually not fully represent the tokens that are passed into the models for generation. For example I think we would miss [what is added with add_generation_prompt here](https://huggingface.co/MiniMaxAI/MiniMax-M2/blob/main/chat_template.jinja#L163)

alecsolder (commented):
I see the value in having the raw tokens here, but I don't know if it makes sense to tie the type to Responses API specifically. We will very likely be implementing this same thing for the Anthropic Messages API in the future for example.

> The structure will be a list of this, which can be extended to include other objects like tool calls, etc in the future

I think this points to the same thing, I don't think we want to have even more API specific types that still would need to be translated into the input for tokenization with `apply_hf_chat_template`.

vllm/entrypoints/context.py
qandrew: SimpleContext will never support multi turn (it's being built in a separate context) so this is fine for now.

alecsolder: When this is used by the new Context being introduced for non-harmony tool calling, how will this be handled?

qandrew: it's currently WIP, I will share more when I have updates

22quinn (approved):

PR #29557, opened 2025-11-27
[Frontend] Remap -O to -cc commandline flag

### LINKED ISSUE #27832: [RFC]: Remap `CompilationConfig` from `-O` to `-cc` in CLI
### Motivation.

With #20283 (and #26847), we're repurposing `-O0`/`-O1`/`-O2`/`-O3` to map to `optimization_level` instead of `CompilationConfig.level`/`CompilationConfig.mode`. This leaves us in a slightly confusing state where `-O` can refer to optimization level or compilation config depending on what follows it:
- `-O0` -> `optimization_level=0`
- `-O 3` -> `optimization_level=3`
- `-O {"cudagraph_mode": "NONE"}` -> `CompilationConfig(cudagraph_mode="NONE")`
- `-O.use_inductor=False` -> `CompilationConfig(use_inductor=False)`
- `--compilation-config.backend=eager` -> `CompilationConfig(backend="eager")`

This is bad UX, and we should fix it. However, a CLI shorthand for `CompilationConfig` is still needed so users can easily compose different properties.

### Proposed Change.

We should create a new shorthand for `CompilationConfig` should be `-cc`. Other options are `-c` and `-C`, but as discussed [here](https://github.com/vllm-project/vllm/pull/26847#discussion_r2439248068), single letters are not "pythonic" and capital letters are worse (extra `Shift` keystroke + less pythonic). However, the exact shorthand is up for discussion. React below to cast your vote.

Example changes:
- `-O0` -> `-O0` (unchanged)
- `-O 3` -> `-O 3` (unchanged)
- `-O {"cudagraph_mode": "NONE"}` -> `-cc {"cudagraph_mode": "NONE"}`
- `-O.use_inductor=False` -> `-cc.use_inductor=False`
- `--compilation-config.backend=eager` -> `--compilation-config.backend=eager` (unchanged)

### Feedback Period.

One week, 10/30 - 11/5

### CC List.

@hmellor @morrison-turnansky @zou3519

### Any Other Things.

Vote for your preferred shorthand:
- 👍 for `-cc`
- 👎 for `-O` (keep it the same)
- 🎉 for `-C`
- 🚀 for `-c`

gmagogsfm:
Users can now:
- Continue using `-O{1,2,3}`, which now means optimization level, not compilation mode any more.
- Continue using `-O.xxx` (deprecation warning until later time when we remove the support)
- Start using `-cc` (no warning, preferred)
- Use `--compilation-config` (no change)

Warning message: "The -O flag for --compilation-config is deprecated and will be removed in a later release. Please use -cc instead. Example: -cc.mode=0 instead of -O.mode=0"

Fixes #27832

### DIFF
--- a/tests/utils_/test_argparse_utils.py
+++ b/tests/utils_/test_argparse_utils.py
@@ -460,15 +460,23 @@ def test_flat_product():
     ]
 
 
-def test_o_legacy_syntax_deprecation():
-    """Test that -O.* dotted syntax converts correctly to -cc syntax."""
+def test_o_legacy_syntax_deprecation(caplog_vllm):
+    """Test that -O.* dotted syntax emits warnings and converts correctly to -cc syntax."""
     parser = FlexibleArgumentParser()
     parser.add_argument("-cc", "--compilation-config", type=json.loads)
 
-    # Test that -O.backend gets converted correctly
+    # Test that -O.backend gets converted correctly AND emits warning
     args = parser.parse_args(["-O.backend=eager"])
     assert args.compilation_config == {"backend": "eager"}
 
+    # Check that deprecation warning was logged
+    assert len(caplog_vllm.records) >= 1
+    assert (
+        "The -O.* dotted syntax for --compilation-config is deprecated"
+        in caplog_vllm.text
+    )
+
     # Test that -O.mode gets converted correctly
+    # Note: warning_once won't emit again in same session
     args = parser.parse_args(["-O.mode=2"])
     assert args.compilation_config == {"mode": 2}
--- a/tests/utils_/test_argparse_utils.py
+++ b/tests/utils_/test_argparse_utils.py
@@ -27,7 +27,7 @@ def parser():
     parser.add_argument("--batch-size", type=int)
     parser.add_argument("--enable-feature", action="store_true")
     parser.add_argument("--hf-overrides", type=json.loads)
-    parser.add_argument("-O", "--compilation-config", type=json.loads)
+    parser.add_argument("-cc", "--compilation-config", type=json.loads)
     parser.add_argument("--optimization-level", type=int)
     return parser
 
@@ -167,8 +167,8 @@ def test_dict_args(parser):
         "--hf-overrides.key2.key4",
         "val3",
         # Test compile config and compilation mode
-        "-O.use_inductor_graph_partition=true",
-        "-O.backend",
+        "-cc.use_inductor_graph_partition=true",
+        "-cc.backend",
         "custom",
         "-O1",
         # Test = sign
@@ -191,9 +191,9 @@ def test_dict_args(parser):
         "--hf_overrides.key14.key15",
         "-minus.and.dot",
         # Test array values
-        "-O.custom_ops+",
+        "-cc.custom_ops+",
         "-quant_fp8",
-        "-O.custom_ops+=+silu_mul,-rms_norm",
+        "-cc.custom_ops+=+silu_mul,-rms_norm",
     ]
     parsed_args = parser.parse_args(args)
     assert parsed_args.model_name == "something.something"
@@ -234,7 +234,7 @@ def test_duplicate_dict_args(caplog_vllm, parser):
         "--hf-overrides.key1",
         "val2",
         "-O1",
-        "-O.mode",
+        "-cc.mode",
         "2",
         "-O3",
     ]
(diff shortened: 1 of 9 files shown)


tests/utils_/test_argparse_utils.py
```diff
@@ -378,36 +378,46 @@ def test_load_config_file(tmp_path):
...
+    assert args.compilation_config == {"mode": 0}
+
+    args = parser.parse_args(["-O", "2"])
+    assert args.compilation_config == {"mode": 2}
+
+    args = parser.parse_args(["-O.mode", "0"])
+    assert args.compilation_config == {"mode": 0}
+
+    args = parser.parse_args(["-O.mode=NONE"])
+    assert args.compilation_config == {"mode": "NONE"}
```
ProExpertProg: Add a comment that we're planning to deprecate these in v0.13.0?

vllm/config/vllm.py
```diff
@@ -95,8 +95,10 @@ class VllmConfig:
     compilation_config: CompilationConfig = Field(default_factory=CompilationConfig)
     """`torch.compile` and cudagraph capture configuration for the model.
 
-    As a shorthand, one can append compilation arguments via 
-    -0.parameter=argument such as `-O.mode=3` (same as `-O='{"mode":3}'`).
+    As a shorthand, one can append compilation arguments via
+    -cc.parameter=argument such as `-cc.mode=3` (same as `-cc='{"mode":3}'`).
+    Optimization level shortcuts like `-O3` are also supported (equivalent to
+    -cc.mode=3).
```
ProExpertProg: This will no longer be true after #26247; `-O<n>` will map to optimization level.

vllm/engine/arg_utils.py
```diff
@@ -1105,7 +1121,11 @@ def add_cli_args(parser: FlexibleArgumentParser) -> FlexibleArgumentParser:
             "--ec-transfer-config", **vllm_kwargs["ec_transfer_config"]
         )
         vllm_group.add_argument(
-            "--compilation-config", "-O", **vllm_kwargs["compilation_config"]
+            "--compilation-config",
+            "-cc",
+            "-O",
```
ProExpertProg: Yeah let's remove the `-O` here completely and just remap any use of `-O.<...>`/`-O {...}` to `-cc` in `argparse_utils`

ProExpertProg (commented):
Thanks for implementing this! We should wait for #26847 to merge first. Can you make the deprecation comments and warnings mention the functionality will be removed in v0.13.0?

vllm/engine/arg_utils.py
```diff
@@ -105,6 +105,22 @@
...
+    def __call__(self, parser, namespace, values, option_string=None):
+        if option_string == "-O":
+            logger.warning_once(
+                "The -O.* dotted syntax for --compilation-config is deprecated and "
+                "will be removed in a future version. Please use -cc.* instead. "
+                "Example: -cc.mode=0 instead of -O.mode=0. Note: -O3 style "
+                "shortcuts remain supported.",
+                scope="deprecation-compilation-config-O-flag",
+            )
+        setattr(namespace, self.dest, values)
```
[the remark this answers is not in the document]
gmagogsfm: Good find, fixed and added test

vllm/config/vllm.py
gmagogsfm: OK, I will wait for #26847 to land and fix up the logic.

tests/utils_/test_argparse_utils.py
```diff
@@ -458,3 +458,17 @@ def test_flat_product():
...
+    """Test that -O.* dotted syntax converts correctly to -cc syntax."""
+    parser = FlexibleArgumentParser()
+    parser.add_argument("-cc", "--compilation-config", type=json.loads)
+
+    # Test that -O.backend gets converted correctly
+    args = parser.parse_args(["-O.backend=eager"])
+    assert args.compilation_config == {"backend": "eager"}
+
+    # Test that -O.mode gets converted correctly
+    args = parser.parse_args(["-O.mode=2"])
```
hmellor, line 481: Could you also [pytest.warns()](https://docs.pytest.org/en/stable/how-to/capture-warnings.html) so make sure that the deprecation warning is displayed?

vllm/utils/argparse_utils.py
```diff
@@ -257,6 +257,31 @@ def repl(match: re.Match) -> str:
             ):
                 # Convert -O <n> to --optimization-level <n>
                 processed_args.append("--optimization-level")
+            elif arg.startswith("-O") and arg[2] == ".":
```
hmellor:
```suggestion
            elif arg.startswith("-O."):
```

vllm/utils/argparse_utils.py
```diff
@@ -257,6 +257,31 @@ def repl(match: re.Match) -> str:
...
+                # If it has compilation config options, emit deprecation warning
+                if any(
+                    cc_option in arg
+                    for cc_option in [
+                        ".mode",
+                        ".backend",
+                        ".custom_option",
+                        ".enable_fusion",
+                        ".disable_custom_fusion",
+                    ]
```
hmellor: This list is not exhaustive and it shouldn't need to exist. Any use of `-O.` is deprecated

hmellor (changes requested):

vllm/utils/argparse_utils.py
gmagogsfm: Done

gmagogsfm: Done

tests/utils_/test_argparse_utils.py
gmagogsfm: added, not using pytest.warns() though as it doesn't support logging.warn

vllm/utils/argparse_utils.py
gmagogsfm: removed

hmellor (approved):
LGTM, thanks!

PR #29572, opened 2025-11-27
[Docs] Improve `priority` parameter documentation

maang-h:
## Purpose
Improved the documentation for the `priority` parameter in `LLM.generate()` method to make it clearer that it must be a list (not a single int) when provided.

### DIFF
--- a/vllm/entrypoints/llm.py
+++ b/vllm/entrypoints/llm.py
@@ -406,7 +406,7 @@ class LLM:
             priority: The priority of the requests, if any.
                 Only applicable when priority scheduling policy is enabled.
                 If provided, must be a list of integers matching the length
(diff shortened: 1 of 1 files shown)


vllm/entrypoints/llm.py
```diff
@@ -405,6 +405,9 @@ def generate(
             lora_request: LoRA request to use for generation, if any.
             priority: The priority of the requests, if any.
                 Only applicable when priority scheduling policy is enabled.
+                If provided, must be a list of integers matching the length
+                of `prompts`,where each priority value corresponds to the prompt
```
DarkLight1337:
```suggestion
                of `prompts`, where each priority value corresponds to the prompt
```

DarkLight1337 (approved):

PR #29575, opened 2025-11-27
CustomOp: grouped topk

xinyu-intel:
## Purpose

plugin can register grouped_topk op for deepseek_v2/3 workloads.

### DIFF
--- a/tests/kernels/moe/test_grouped_topk.py
+++ b/tests/kernels/moe/test_grouped_topk.py
@@ -9,8 +9,8 @@ import pytest
 import torch
 
 from vllm.model_executor.layers.fused_moe.fused_moe import (
+    GroupedTopk,
     fused_grouped_topk,
-    grouped_topk,
 )
 from vllm.platforms import current_platform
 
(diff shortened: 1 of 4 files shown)


Yikun:
Yep, agree.

xinyu-intel:
> cc @Yikun 
> 
> 
> 
> I think it might be a bit excessive to have a `has_...` plugin style for all ops not sure if this would be a good use for `CustomOP` (cc @ProExpertProg )

thx, let me take a try.

Yikun (commented):
Overall looks good, thanks.

@MengqingCao Do you have time to do a e2e confirm on next monday, thanks.

yewentao256 (commented):
CC @ProExpertProg

Yikun (approved):
LGTM, we'd better to land this in 2025 😁.

Release cut time: 12.15 (0.13.0) .

@LucasWilkinson @ProExpertProg Any concern?

tjtanaa:
@xinyu-intel can you follow up with a proper unit test for custom op GroupedTopk. Because we need to modify the `custom_op` list to properly trigger the platform dependent `forward_` functions. The unit tests in this PR does not address that. So it will always call `forward_native`.

PR #29600, opened 2025-11-27
[CPU]Parallelize over tokens in int4 moe

xiangze-arm:
## Purpose

This patch changes the parallelization strategy from over experts to over tokens.

When experts have very different numbers of tokens, the old approach could lead to thread imbalance. By iterating over total tokens instead of experts, the workload can be distributed more evenly. 

## Test Plan

Tested locally with Qwen3-30B-A3B w4a8 model quantized with llmcompressor.
```
vllm bench throughput --dtype=float32 --num-prompts 64 --seed 0 --dataset-name sharegpt --max-model-len=2048 --input_len=64 --model Qwen3-30B-A3B-w4a8-channelwise
```

## Test Result

With this PR, on 32 Neoverse-N2 cores, Qwen3-30B-A3B 4bit model gets 25% throughput improvement.


LucasWilkinson:
@bigPYJ1151 (Please redirect if you are not the correct person; not sure who the CPU codeowner is)

bigPYJ1151 (approved):

Issue #16313, opened 2025-04-09
[Feature]: Support structured output and tool call together


Wendong-Fan:
### Checklist

- [ ] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [ ] 2. Please use English, otherwise it will be closed.

### Motivation

as discussed in: https://www.reddit.com/r/LocalLLaMA/comments/1h2y7ys/why_can_you_not_use_structured_output_and_tool/

request support for tool execution followed by structured response output, similar to how OpenAI handles function calls (Tool Calls) and Schema outputs in the gpt-4o-mini API.

OpenAI supports this capability through a workflow where:

The model can first execute one or more tool calls to perform calculations or retrieve information
After receiving the results of these tool calls, the model can then produce a final structured response conforming to a predefined schema

**Example Implementation from OpenAI**

Input to OpenAI
User query: "current age of University of Oxford is 928"
Available tools: add, subtract, multiply, divide, round
Response format: Schema with current_age and calculated_age fields

Execution flow:
1. Model analyzes query and determines it needs to call the add function
2. Model executes: add(928, 1) to calculate the next year's age
3. Function returns: 929
4. Model produces final structured response:
   {"current_age":"928","calculated_age":"929"}

```
=================Input messages to OpenAI=================
[{'role': 'system', 'content': 'You are a helpful assistant. You must call tools'}, {'role': 'user', 'content': 'current age of University of Oxford is 928, '}]
=================Input config to OpenAI=================
[...]

PR #29612, opened 2025-11-27
[Feature]: Support structured output and tool call together

mladjan-gadzic:
## Purpose
This PR implements feature request https://github.com/vllm-project/vllm/issues/16313 to allow tools calling and structure output in a single request.

## Test Plan
Added test and example to reproduce behavior.

## Test Result
vLLM uses tools with structured output accordinglly.

Before: tools were not used.
After: tools are used.
---
<details>

- [ x The test plan, such as providing test command.
</details>


vllm/entrypoints/openai/serving_engine.py
```diff
@@ -1444,7 +1444,11 @@ def _parse_tool_calls_from_content(
                 FunctionCall(name=request.tool_choice.function.name, arguments=content)
             )
             content = None  # Clear content since tool is called.
-        elif request.tool_choice == "required":
+        elif request.tool_choice == "required" or (
+            request.tool_choice == "auto" and request.tools
```
chaunceyjiang, line 1448: I don't think this is a good modification.  

When a user sets both `tools` and `tool_choice="auto"`, but the user's messages are completely unrelated to the tools—for example, the tool is `get_weather`, and the message content is "What is 1+1?".

mladjan-gadzic: on the second thought i agree. i've tried numerous different things, but i could not make model use tools without this change. do you have any suggestion which change would be the best for this?

chaunceyjiang: To be honest, I still haven’t understood what problem you’re describing.

Also, the change you made looks more like a `required` choice.
Why not just use tool_choice = "required" instead?

mladjan-gadzic: there is example in the github issue https://github.com/vllm-project/vllm/issues/16313 where everything is explained in more details but i'll try to summarize it here.

it turns out that structured output and tool call is not supported. model always ignores tool call when structure output is in place. this is something that is supported by openai as stated in the issue

> The model can first execute one or more tool calls to perform calculations or retrieve information
After receiving the results of these tool calls, the model can then produce a final structured response conforming to a predefined schema

when tool_choice="required" it works as expected, when it does not work is when tool_choice="auto". even when prompted to use tools, those are ignored.

chaunceyjiang: Ohhh, thanks for the explanation, I understand now. Indeed, structured output + tool call is not supported yet.

Have you tested making a tool call where the first request uses only a tool call, and the second request uses structured output + tool call?

mladjan-gadzic: i tried making it two-phase request where first would use tool call and second structured output on top of that but that seemed like an overkill. that's why i went with current implementation.

there is test added to these changes and a example that help explain this further.

please have a look at test_tools_with_structured_output.py and example.py. before implementation it was failing. after implementation it was passing.

mladjan-gadzic: @chaunceyjiang any chance you had the time to look at this?

chaunceyjiang: https://github.com/vllm-project/vllm/pull/29612#discussion_r2600856500


I understand the problem you’re facing right now, but I still don’t think this is a good change, as I mentioned before. (At the moment, I also don’t know how to solve this issue. 😭)

PR #29621, opened 2025-11-27
[Multimodal][Core] Optimize multimodal preprocessing cache by hashing image bytes instead of pixel values

ImaGoodFella:
## Summary:
This PR optimizes the multimodal preprocessing cache by hashing image bytes instead of pixel values, significantly improving performance for both cached and uncached images, with the largest gains on high-resolution images.

## Problem:
The current implementation converts PIL Images to numpy arrays (np.asarray(obj)) for cache hashing, which requires decompressing images to full pixel values. For large images (e.g., 4K resolution with 16.8 megapixels), this involves hashing millions of pixel values, creating a significant performance bottleneck.

## Solution: 
Store original image bytes during loading and use them for hashing. This leverages the compressed image format (JPEG/PNG), which is typically orders of magnitude smaller than the uncompressed pixel data.

**Scenario 1**: Cached images (preprocessing already done)
This optimization dramatically reduces cache lookup time by eliminating the need to decode images to pixel values:

image

**Scenario 2**: Uncached images (first-time preprocessing)
Hash computation is faster since we hash compressed bytes instead of millions of pixels:

image

Reviewers:

Subscribers:

Tasks:

Tags:

## Test Result

---
<details>

</details>


DarkLight1337 (commented):
Depending on the raw bytes of the image without preprocessing will likely introduce security vulnerabilities where users cause prefix cache hits using unrelated images. I don't think this is a good idea. cc @ywang96 @Isotr0py @russellb

DarkLight1337:
Thanks for the explanation, I think it makes sense but I would like @russellb to verify this as I'm not a security expert at all.

Meanwhile I'll look at the other parts of the PR.

vllm/multimodal/hasher.py
```diff
@@ -33,7 +33,10 @@ def serialize_item(cls, obj: object) -> Iterable[bytes | memoryview]:
             ):
                 # If the image has exif ImageID tag, use that
                 return (exif[Image.ExifTags.Base.ImageID].bytes,)
-            data = {"mode": obj.mode, "data": np.asarray(obj)}
+            if hasattr(obj, "_image_bytes"):
```
DarkLight1337: I feel that this is too fragile (e.g. it assumes that the images will not be copied before this method is called), we should add a new interface to `MediaIO` to return the hash alongside the image and set `mm_hashes` in the input explicitly.

Isotr0py:
> Here we hash the encoded file itself, so format/metadata/compression differences are reflected in the hash and that class of collision cannot occur.
Given the current design, prefix-based cache attacks or unrelated-image collisions aren't feasible.

I wonder if we use encoded image for hasher, will it cause a mm-cache hit missing issue?

For example, given two images with same pixel values but in different encoding format (jpg/png) respectively, they will be treated as two different image and won't hit the cache?

vllm/multimodal/image.py
```diff
@@ -85,15 +85,17 @@ def _convert_image_mode(self, image: Image.Image) -> Image.Image:
 
     def load_bytes(self, data: bytes) -> Image.Image:
         image = Image.open(BytesIO(data))
-        image.load()
+        image._image_bytes = data  # Store original bytes for hashing
         return self._convert_image_mode(image)
```
nwaughachukwuma: Why not keep `load_bytes` pure without overriding/overloading private/internal field `_image_bytes`?

The caller should run/use the hashing.

```py
image = XMediaIO.load_file(...)
data = image.tobytes() # get the bytes and hash it.
```

vllm/multimodal/image.py
```diff
@@ -85,15 +85,17 @@ def _convert_image_mode(self, image: Image.Image) -> Image.Image:
...
         return self.load_bytes(pybase64.b64decode(data, validate=True))
 
     def load_file(self, filepath: Path) -> Image.Image:
-        image = Image.open(filepath)
-        image.load()
+        with open(filepath, "rb") as f:
+            data = f.read()
+        image = Image.open(BytesIO(data))
+        image._image_bytes = data  # Store original bytes for hashing
         return self._convert_image_mode(image)
```
nwaughachukwuma: 1. same as the above, i don't see the case for loading the bytes in `load_file` and overriding/overloading `_image_bytes`.
2. if you want the bytes in `load_file`, which I doubt you need, you could just do the following. but you really should just keep the function the same and call on get the byte at the call site
```py
# what you could do; but i doubt you need it
image = Image.open(filepath)
image.load()

data = image.tobytes()  # get the bytes here
```

```py
# how you should structure the code instead
image = XMediaIO.load_file(...)
data = image.tobytes() # get the bytes and hash it
```

nwaughachukwuma (changes requested):

vllm/multimodal/image.py
DarkLight1337: Wouldn't that cause unnecessary serialization?

vllm/multimodal/image.py
ImaGoodFella: > This is also fundamentally different from [CVE-2025-46722](https://advisories.gitlab.com/pkg/pypi/vllm/CVE-2025-46722/), which arose because vLLM previously hashed pixel bytes only, so two different files could produce the same hash if their decoded pixel arrays matched. Here we hash the encoded file itself, so format/metadata/compression differences are reflected in the hash and that class of collision cannot occur. Given the current design, prefix-based cache attacks or unrelated-image collisions aren't feasible.

Thanks for the feedback! I think there might be a misunderstanding about what image.tobytes() does vs. what we're trying to achieve here.

image.tobytes() returns the decoded pixel buffer (uncompressed RGB/RGBA values), not the original file bytes. This approach has two major problems:

1. Performance: It requires calling image.load() to decode the image first, which is exactly what we're trying to avoid for cache lookups. For a 4K image, we'd be decompressing and hashing ~31 MB of pixel data instead of just hashing the ~500 KB compressed JPEG bytes. This eliminates the 20-200x speedup this PR provides.

2. Security: Hashing decoded pixel bytes is precisely what caused [CVE-2025-46722](https://advisories.gitlab.com/pkg/pypi/vllm/CVE-2025-46722). When you hash only pixel values, two different files (different encodings, metadata, or maliciously crafted images) can produce identical hashes if their decoded pixels happen to match. By storing and hashing _image_bytes (the original compressed file data), we ensure the hash reflects the actual input file, preventing this entire class of collision attacks.

The _image_bytes approach keeps a reference to the compressed data we already read from disk/network, so there's no additional I/O overhead—we're just preserving what we already have for secure, efficient hashing.

Does that clarify the rationale?

vllm/multimodal/image.py
ImaGoodFella: > This is also fundamentally different from [CVE-2025-46722](https://advisories.gitlab.com/pkg/pypi/vllm/CVE-2025-46722/), which arose because vLLM previously hashed pixel bytes only, so two different files could produce the same hash if their decoded pixel arrays matched. Here we hash the encoded file itself, so format/metadata/compression differences are reflected in the hash and that class of collision cannot occur. Given the current design, prefix-based cache attacks or unrelated-image collisions aren't feasible.

Thanks for the feedback! I think there might be a misunderstanding about what image.tobytes() does vs. what we're trying to achieve here.

image.tobytes() returns the decoded pixel buffer (uncompressed RGB/RGBA values), not the original file bytes. This approach has two major problems:

1. Performance: It requires calling image.load() to decode the image first, which is exactly what we're trying to avoid for cache lookups. For a 4K image, we'd be decompressing and hashing ~31 MB of pixel data instead of just hashing the ~500 KB compressed JPEG bytes. This eliminates the 20-200x speedup this PR provides.

2. Security: Hashing decoded pixel bytes is precisely what caused [CVE-2025-46722](https://advisories.gitlab.com/pkg/pypi/vllm/CVE-2025-46722). When you hash only pixel values, two different files (different encodings, metadata, or maliciously crafted images) can produce identical hashes if their decoded pixels happen to match. By storing and hashing _image_bytes (the original compressed file data), we ensure the hash reflects the actual input file, preventing this entire class of collision attacks.

The _image_bytes approach keeps a reference to the compressed data we already read from disk/network, so there's no additional I/O overhead—we're just preserving what we already have for secure, efficient hashing.

Does that clarify the rationale?

nwaughachukwuma: Cool. Thanks for clarifying.

The goal is to a) keep the function pure b) not mutate and Image instance using a private field that doesn't exist in ImageFile or Image.Image.

Did you consider returning a tuple and handling the hashing at call site? You could also just use a wrapper like so

```py
class Wrapper:
   image: Image.Image
   _image_bytes: bytes
   
   def __init__(self, image: Image.Image, image_bytes: bytes):
       self.image = image
       self._image_bytes = image_bytes
    
# later
return Wrapper(self._convert_image_mode(image), data)
```

vllm/multimodal/image.py
nwaughachukwuma: Cool. Thanks for clarifying.

The goal is to a) keep the function pure b) not mutate and Image instance using a private field that doesn't exist in ImageFile or Image.Image.

Did you consider returning a tuple and handling the hashing at call site? You could also just use a wrapper like so

```py
class Wrapper:
   image: Image.Image
   _image_bytes: bytes
   
   def __init__(self, image: Image.Image, image_bytes: bytes):
       self.image = image
       self._image_bytes = image_bytes
    
# later
return Wrapper(self._convert_image_mode(image), data)
```

vllm/multimodal/hasher.py
ImaGoodFella: I agree, and I've updated my code to use a wrapper class that contains both the image and the original bytes. This approach is cleaner and addresses the fragility concern.

I didn't go with passing the hash directly through the MediaIO interface because the hashing actually requires hf_processor_mm_kwargs and tokenization_kwargs, which aren't available at the point where we're loading the image. By using a wrapper to carry the bytes alongside the image, we defer the hashing until we have all the necessary context, which I think is the simplest solution.

vllm/multimodal/image.py
ImaGoodFella: Thanks for the suggestions! I went with a very similar approach in my latest commit - I'm now using a wrapper class that contains both the image and the original bytes, just like you suggested. This keeps the loading functions pure and avoids mutating the PIL Image object with non-existent private fields.

vllm/multimodal/hasher.py
```diff
@@ -31,14 +33,33 @@ def serialize_item(cls, obj: object) -> Iterable[bytes | memoryview]:
...
+                exif[Image.ExifTags.Base.ImageID], uuid.UUID
+            ):
+                return (exif[Image.ExifTags.Base.ImageID].bytes,)
+
+            data = {"mode": obj.media.mode, "data": obj.original_bytes}
+            palette = obj.media.palette
+            if palette is not None:
+                data["palette"] = palette.palette
+                if palette.rawmode is not None:
+                    data["palette_rawmode"] = palette.rawmode
```
DarkLight1337: Why do we still need to include additional metadata if you say that the `original_bytes` is already sufficiently unique?

ImaGoodFella: You're right — if original_bytes already guarantees uniqueness, the extra palette metadata doesn’t add value and just inflates the hash input. I removed it so the hasher stays simpler and focused on the actual image bytes.

vllm/multimodal/parse.py
```diff
@@ -98,10 +105,18 @@ def get_passthrough_data(self) -> Mapping[str, object]:
...
+        """Extract media from wrapper if present."""
+        return item.media if isinstance(item, MediaWithBytes) else item
+
     def get_count(self) -> int:
         return len(self.data)
 
     def get(self, index: int) -> _T:
+        return self._unwrap(self.data[index])
+
+    def get_hash_item(self, index: int) -> _T | MediaWithBytes[_T]:
```
DarkLight1337: Rename to `get_item_for_hash`

ImaGoodFella: Sounds good!

DarkLight1337 (approved):
Thanks for your patience, looks good now!

PR #29624, opened 2025-11-27
[Attention] Make seq_lens_cpu optional in CommonAttentionMetadata to enable true async spec-decode

LucasWilkinson:
Implement: https://github.com/vllm-project/vllm/issues/29134


vllm/v1/attention/backends/utils.py
```diff
@@ -81,7 +76,7 @@ class CommonAttentionMetadata:
     max_query_len: int
     """Longest query in batch"""
     max_seq_len: int
-    """Longest context length in batch"""
+    """Longest context length (may be an upper bound when seq_lens_cpu is missing)"""
```
benchislett: I think we should treat this as an upper-bound everywhere, for safety. Unless you can think of a case where we need it to be exactly correct, I would prefer to design around the concept that max_seq_len is always an optimistic estimate.
```suggestion
    """Longest context length (may be an upper bound)"""
```

vllm/v1/attention/backends/utils.py
```diff
@@ -100,6 +95,34 @@ class CommonAttentionMetadata:
     dcp_local_seq_lens_cpu: torch.Tensor | None = None
     """Sequence lengths of the local rank in decode context parallelism world"""
 
+    _seq_lens_cpu: torch.Tensor | None = None
+    _num_computed_tokens_cpu: torch.Tensor | None = None
+
+    @property
+    def seq_lens_cpu(self) -> torch.Tensor:
+        """Lazy CPU copy of seq_lens. Accessing this may cause a sync.
```
benchislett: I'm not sure how I feel about this property definition. I think it should be the responsibility of the caller to serialize and reuse seq_lens_cpu where applicable, so that we don't run into consistency issues.

For instance, if EAGLE3 sets `_seq_lens_cpu` to None and only updates `seq_lens` each iteration, it must also take on the responsibility of re-setting `_seq_lens_cpu` to None every time since accessing this property can mutate the metadata which results in an out-of-date entry for `_seq_lens_cpu`.

vllm/v1/attention/backends/utils.py
```diff
@@ -100,6 +95,34 @@ class CommonAttentionMetadata:
...
+        return self._seq_lens_cpu
+
+    @property
+    def num_computed_tokens_cpu(self) -> torch.Tensor:
+        """Lazy CPU tensor for num computed tokens. Accessing may cause a sync.
+
+        .. deprecated::
+            Prefer using device seq_lens directly to avoid implicit H<>D sync.
+        """
+        if self._num_computed_tokens_cpu is None:
```
benchislett, line 125: Same comment here

vllm/v1/attention/backends/utils.py
benchislett: This would also allow us to make `seq_lens_cpu` the name of the class member instead of having to initialize the class with `_seq_lens_cpu=...`, which feels a bit messy

benchislett (changes requested):
I prefer a different style of implementation here, see comments. Open to discussion.

See also a few highlighted issues caught by the pre-commit

vllm/v1/attention/backends/utils.py
LucasWilkinson: > I'm not sure how I feel about this property definition. I think it should be the responsibility of the caller to serialize and reuse seq_lens_cpu where applicable, so that we don't run into consistency issues.

I fully agree; I think we just need to have a transition period for OOT plugins (and other backends in general) @Yikun can maybe weigh in here, this was just my idea for backwards compatibility for a few versions to allow people time to migrate 

> For instance, if EAGLE3 sets _seq_lens_cpu to None and only updates seq_lens each iteration, it must also take on the responsibility of re-setting _seq_lens_cpu to None every time since accessing this property can mutate the metadata which results in an out-of-date entry for _seq_lens_cpu.

good catch!

LucasWilkinson: (hence the `deprecated::` tag; we can make the wording stronger there though, like will be removed in v0.14; also want to read up on `from warnings import deprecated` to see if we can use that without too many issues)

benchislett: I see. This is acceptable I suppose, though I think I would still prefer for us to just make a big sweeping change to all the places that `seq_lens_cpu` is consumed in vLLM, and then let external plugins sort it out. In theory this will (for now) only actually ever be `None` in the case of EAGLE speculative decoding, right? In any case, I suspect ModelRunnerV2 is likely to deprecate these completely in a few months time.

Also you could add a `warning_once` in the copy pathway, to make it explicit to the end-user that this is not super efficient. Would that be appropriate, or too noisy?

Yikun: > I think we just need to have a transition period for OOT plugins (and other backends in general)

Thanks for notification,this will help plugin, but I guess some plugins will also be break when they use other workflow use the new var.

Any way, I think it's good start to care backard compatible.

Also cc @wangxiyuan

benchislett (approved):
LGTM. Let's remain vigilant after merging in case of any unexpected perf regressions

mgoin (approved):

PR #29626, opened 2025-11-27
[Feature]: Qwen3 Omni Transcriptions

sumitaryal:
Implementation for issue #29405 

## Test Result

---
<details>

</details>


NickLucche (commented):
Can we add tests or at least an example with running the model + related docs? Thanks for your work.

examples/online_serving/openai_qwen3_omni_transcription.py
```diff
@@ -0,0 +1,68 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""
+Minimal transcription example for Qwen3-Omni using the OpenAI-compatible API.
+
```
NickLucche: this is just a client it doesn't actually have anything to do with qwen other than model name. Let's stick with the client we already have and either 
- allow the user to pass in the model name
- automatically infer it by querying /v1/models

docs/serving/openai_compatible_server.md
```diff
@@ -372,6 +372,11 @@ you can use the [official OpenAI Python client](https://github.com/openai/openai
 
 Code example: [examples/online_serving/openai_transcription_client.py](../../examples/online_serving/openai_transcription_client.py)
 
+Qwen3-Omni quickstart (SupportsTranscription):
+
+- Start the server: `vllm serve Qwen/Qwen3-Omni-30B-A3B-Instruct --port 8000 --tensor-parallel-size 2` (adjust `--tensor-parallel-size` to match available GPUs)
+- Run the client: `python examples/online_serving/openai_qwen3_omni_transcription.py`
+
```
NickLucche: we should just add qwen among the supported transcription models, no need to provide a quickstart if the steps are the same as other models

NickLucche (commented):
what's the smallest qwen-omni size that we could use for the unit tests in test_transcription.py?

examples/online_serving/openai_qwen3_omni_transcription.py
sumitaryal: Do you mean modifying the current [openai_transcription_client.py](https://github.com/vllm-project/vllm/blob/main/examples/online_serving/openai_transcription_client.py) to accept a model argument through the cli instead of having a new example file for qwen3 omni?

sumitaryal:
> what's the smallest qwen-omni size that we could use for the unit tests in test_transcription.py?

Qwen has released 3 Qwen3-Omni models which are all 30B, I used the instruct one to test the example file.

examples/online_serving/openai_qwen3_omni_transcription.py
NickLucche: yep

PR #29627, opened 2025-11-27
[Attention] Cache attention metadata builds across hybrid KV-cache groups

LucasWilkinson:
Scaled back version of: https://github.com/vllm-project/vllm/pull/22788 and generalization of: https://github.com/vllm-project/vllm/pull/29444

Depends on https://github.com/vllm-project/vllm/pull/29628 land that first

## Test Plan

Decode-heavy workload with APC enabled => **latency reduction through this PR: 13.7%**

```
vllm bench latency --model ibm-granite/granite-4.0-tiny-preview --input-len 128  --output-len 2048 \
    --batch-size 8 --enable-prefix-caching --num-iters 10 --num-iters-warmup 2
```

Decode-heavy workload with APC disabled => **latency reduction through this PR: 2.4%**

```
vllm bench latency --model ibm-granite/granite-4.0-tiny-preview --input-len 128  --output-len 2048 \
    --batch-size 8 --num-iters 10 --num-iters-warmup 2
```

### APC enabled:

Before:
```
Avg latency: 17.656723083648830 seconds
10% percentile latency: 17.629154206998646 seconds
25% percentile latency: 17.638833652948960 seconds
50% percentile latency: 17.659772212151438 seconds
75% percentile latency: 17.681711453711614 seconds
90% percentile latency: 17.693322991393508 seconds
99% percentile latency: 17.694964875970037 seconds
```

This PR:
```
Avg latency: 15.236061835102737 seconds
10% percentile latency: 15.214241081103683 seconds
25% percentile latency: 15.218232876388356 seconds
50% percentile latency: 15.237253282219172 seconds
75% percentile latency: 15.245128114242107 seconds
90% percentile latency: 15.250385893788188 seconds
99% percentile latency: 15.278700626296922 seconds
```

### APC disabled:

Before:
```
Avg latency: 14.836663954611868 seconds
10% percentile latency: 14.762344243377447 seconds
25% percentile latency: 14.770147853763774 seconds
50% percentile latency: 14.822804148774594 seconds
75% percentile latency: 14.853949831798673 seconds
90% percentile latency: 14.937904387619346 seconds
99% percentile latency: 15.010869848066940 seconds
```

This PR:
```
Avg latency: 14.485000116657465 seconds
10% percentile latency: 14.469038359075785 seconds
25% percentile latency: 14.475818350212649 seconds
50% percentile latency: 14.479432722553610 seconds
75% percentile latency: 14.504760635318235 seconds
90% percentile latency: 14.514666751492769 seconds
[...]

PR #29628, opened 2025-11-27
[Core] Refactor `_build_attention_metadata`

LucasWilkinson:
Refactor to make it more concise and make the nested kv_cache_groups/attn_groups loops more evident (this is not a final refactor; just incremental)

Preparatory for https://github.com/vllm-project/vllm/pull/29627

### DIFF
--- a/vllm/v1/worker/gpu_model_runner.py
+++ b/vllm/v1/worker/gpu_model_runner.py
@@ -1530,6 +1530,10 @@ class GPUModelRunner(
         """
         :return: tuple[attn_metadata, spec_decode_common_attn_metadata]
         """
+        # Attention metadata is not needed for attention free models
+        if len(self.kv_cache_config.kv_cache_groups) == 0:
+            return {}, None
+
         num_tokens_padded = num_tokens_padded or num_tokens
         num_reqs_padded = num_reqs_padded or num_reqs
         assert num_reqs_padded is not None and num_tokens_padded is not None
--- a/vllm/v1/worker/gpu_model_runner.py
+++ b/vllm/v1/worker/gpu_model_runner.py
@@ -1542,17 +1542,6 @@ class GPUModelRunner(
                     logits_indices
                 )
 
-        # update seq_lens of decode reqs under DCP.
-        if self.dcp_world_size > 1:
-            self.dcp_local_seq_lens.cpu[:num_reqs] = get_dcp_local_seq_lens(
-                self.seq_lens.cpu[:num_reqs],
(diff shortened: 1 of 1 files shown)


heheda12345:
> ### 💡 Codex Review
> https://github.com/vllm-project/vllm/blob/7284d2044aca49ea76a4dbb083566b195e7aa91c/v1/worker/gpu_model_runner.py#L1642-L1649 **P1 Badge Slice spec decode masks for ubatches**
> When `use_spec_decode` is enabled and the batch is split into `ubatch_slices`, `_build_attn_group_metadata` passes `num_accepted_tokens`/`num_decode_draft_tokens_cpu` sliced to `num_reqs_padded` into `GDNAttentionMetadataBuilder` even though `common_attn_metadata` only covers a single ubatch. `GDNAttentionMetadataBuilder.build` masks `query_lens` with the provided tensors, so the full‑batch mask length no longer matches the ubatch query lengths and will raise a boolean indexing error as soon as spec decode runs with ubatching. The extra tensors need to be sliced to the same ubatch slice before calling `build`.
> ℹ️ About Codex in GitHub

CC @LucasWilkinson do you think this is a problem?

LucasWilkinson:
> > ### 💡 Codex Review
> > https://github.com/vllm-project/vllm/blob/7284d2044aca49ea76a4dbb083566b195e7aa91c/v1/worker/gpu_model_runner.py#L1642-L1649 **P1 Badge Slice spec decode masks for ubatches**
> > When `use_spec_decode` is enabled and the batch is split into `ubatch_slices`, `_build_attn_group_metadata` passes `num_accepted_tokens`/`num_decode_draft_tokens_cpu` sliced to `num_reqs_padded` into `GDNAttentionMetadataBuilder` even though `common_attn_metadata` only covers a single ubatch. `GDNAttentionMetadataBuilder.build` masks `query_lens` with the provided tensors, so the full‑batch mask length no longer matches the ubatch query lengths and will raise a boolean indexing error as soon as spec decode runs with ubatching. The extra tensors need to be sliced to the same ubatch slice before calling `build`.
> > ℹ️ About Codex in GitHub
> 
> CC @LucasWilkinson do you think this is a problem?

Its not really a problem since GDN + Spec + DBO isn't supported; currently on main `extra_attn_metadata_args` isn't even passed for the ubatch metadata building path.

But worth having an assert to make this explicit; added! thanks for calling this out

heheda12345 (approved):
LGTM!

mgoin (approved):

PR #29629, opened 2025-11-27
vLLM with MPS Support

ericcurtin:
For macOS/Apple/Metal

### DIFF
--- a/vllm/model_executor/custom_op.py
+++ b/vllm/model_executor/custom_op.py
@@ -75,6 +75,11 @@ class CustomOp(nn.Module):
         # NOTE(woosuk): This is a placeholder for future extensions.
         return self.forward_native(*args, **kwargs)
 
+    def forward_mps(self, *args, **kwargs):
+        # By default, we assume that MPS ops are compatible with the
+        # PyTorch-native implementation.
+        return self.forward_native(*args, **kwargs)
+
     def forward_oot(self, *args, **kwargs):
         # By default, we assume that OOT ops are compatible with the
         # PyTorch-native implementation.
@@ -101,6 +106,8 @@ class CustomOp(nn.Module):
             return self.forward_tpu
         elif current_platform.is_xpu():
             return self.forward_xpu
+        elif current_platform.is_mps():
+            return self.forward_mps
         elif current_platform.is_out_of_tree():
             return self.forward_oot
         else:
--- a/vllm/platforms/mps.py
+++ b/vllm/platforms/mps.py
@@ -1,0 +1,225 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""MPS (Metal Performance Shaders) platform support for Apple Silicon."""
+
+import subprocess
+from typing import TYPE_CHECKING
+
+import torch
+
+from vllm import envs
+from vllm.attention.backends.registry import AttentionBackendEnum
+from vllm.logger import init_logger
+
+from .interface import Platform, PlatformEnum
+
+logger = init_logger(__name__)
+
+if TYPE_CHECKING:
+    from vllm.config import VllmConfig
+else:
+    VllmConfig = None
+
+
+class MpsPlatform(Platform):
+    _enum = PlatformEnum.MPS
+    device_name: str = "mps"
+    device_type: str = "mps"
+    dispatch_key: str = "MPS"
+    dist_backend: str = "gloo"
+    device_control_env_var = "GPU_DEVICE_ORDINAL"
+
+    # MPS has limited operator support, use eager backend
+    simple_compile_backend: str = "eager"
+
+    @property
+    def supported_dtypes(self) -> list[torch.dtype]:
+        # MPS supports float16, bfloat16 (on newer Apple Silicon), and float32
+        # but does NOT support float64
+        if (
+            subprocess.check_output(
+                ["sysctl -n hw.optional.arm.FEAT_BF16"], shell=True
+            ).strip()
+            == b"1"
+        ):
+            return [torch.bfloat16, torch.float16, torch.float32]
+        return [torch.float16, torch.float32]
+
+    @classmethod
+    def get_device_name(cls, device_id: int = 0) -> str:
+        return (
+            torch.backends.mps.get_name()
+            if hasattr(torch.backends.mps, "get_name")
+            else "Apple MPS"
+        )
+
+    @classmethod
+    def get_attn_backend_cls(
+        cls,
+        selected_backend: "AttentionBackendEnum",
+        head_size: int,
+        dtype: torch.dtype,
+        kv_cache_dtype: str | None,
+        block_size: int,
+        use_mla: bool,
+        has_sink: bool,
+        use_sparse: bool,
+        attn_type: str | None = None,
+    ) -> str:
+        if use_mla:
+            raise NotImplementedError("MLA is not supported on MPS.")
+        if use_sparse:
+            raise NotImplementedError("Sparse Attention is not supported on MPS.")
+        # Use MPS attention backend
+        logger.info("Using MPS attention backend.")
+        return AttentionBackendEnum.MPS_ATTN.get_path()
+
+    @classmethod
+    def get_device_total_memory(cls, device_id: int = 0) -> int:
+        """Get total MPS device memory."""
+        import psutil
+
+        from vllm.utils.mem_constants import GiB_bytes
+
+        # Check for user-specified KV cache space first
+        kv_cache_space = envs.VLLM_CPU_KVCACHE_SPACE
+        if kv_cache_space is None:
+            # Use a portion of system memory for MPS
+            # Apple Silicon shares memory between CPU and GPU
+            total_memory = psutil.virtual_memory().total
+            # Use 50% of system memory for KV cache by default
+            # (same as CPU backend since MPS shares unified memory)
+            kv_cache_space = total_memory // 2
+            logger.warning_once(
+                "Environment variable VLLM_CPU_KVCACHE_SPACE (GiB) "
+                "for MPS backend is not set, using %.1f GiB by default.",
+                kv_cache_space / GiB_bytes,
+            )
+        else:
+            kv_cache_space = int(kv_cache_space * GiB_bytes)
+
+        return kv_cache_space
+
+    @classmethod
+    def set_device(cls, device: torch.device) -> None:
+        """Set the device for the current platform."""
+        # MPS only has one device, but set it anyway for consistency
+        # and to avoid potential side effects from not setting it
+        torch.mps._set_default_device(device)
+
+    @classmethod
+    def inference_mode(cls):
+        return torch.no_grad()
+
+    @classmethod
+    def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
+        import os
+
+        model_config = vllm_config.model_config
+
+        if model_config is not None:
+            model_config.disable_cascade_attn = True
+
+        cache_config = vllm_config.cache_config
+
+        if cache_config.block_size is None:
+            cache_config.block_size = 16  # Smaller block size for MPS
+
+        if cache_config.cache_dtype != "auto":
+            logger.warning(
+                "MPS backend doesn't support KV cache quantization, "
+                "falling back to auto."
+            )
+            cache_config.cache_dtype = "auto"
+
+        cache_config.cpu_kvcache_space_bytes = MpsPlatform.get_device_total_memory()
+
+        parallel_config = vllm_config.parallel_config
+
+        # MPS only supports single device
+        if parallel_config.world_size > 1:
+            raise ValueError("MPS backend only supports single device (world_size=1)")
+
+        # Force single process execution
+        if parallel_config.distributed_executor_backend is None:
+            parallel_config.distributed_executor_backend = "uni"
+
+        if parallel_config.worker_cls == "auto":
+            # Use MPS worker which handles macOS-specific initialization
+            parallel_config.worker_cls = "vllm.v1.worker.mps_worker.MPSWorker"
+
+        # Disable custom all reduce
+        parallel_config.disable_custom_all_reduce = True
+
+        # Disable DBO
+        if parallel_config.enable_dbo:
+            logger.warning("Dual-Batch Overlap is not supported on MPS, disabled.")
+            parallel_config.enable_dbo = False
+
+        # Disable CUDA graph equivalent features
+        from vllm.config import CompilationMode
+
+        vllm_config.compilation_config.cudagraph_capture_sizes = []
+
+        compilation_config = vllm_config.compilation_config
+        # Use eager mode for MPS - torch.compile has limited support
+        compilation_config.mode = CompilationMode.NONE
+        compilation_config.backend = "eager"
+
+        if vllm_config.lora_config is not None:
+            compilation_config.mode = CompilationMode.NONE
+
+        # MPS platform runs on MPS device
+        assert vllm_config.device_config.device_type == "mps"
+
+        # Environment variables for MPS executor
+        os.environ["VLLM_WORKER_MULTIPROC_METHOD"] = "spawn"
+
+        # Enable MPS fallback for unsupported ops
+        os.environ["PYTORCH_ENABLE_MPS_FALLBACK"] = "1"
+
+        if model_config is not None and model_config.use_mla:
+            logger.info(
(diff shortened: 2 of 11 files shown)


ericcurtin:
The reality is pytorch mps, needs more work

ericcurtin:
I'm more interested on getting upstream support with DGX Spark better right now

vllm/model_executor/custom_op.py
```diff
@@ -75,6 +75,11 @@ def forward_tpu(self, *args, **kwargs):
         # NOTE(woosuk): This is a placeholder for future extensions.
         return self.forward_native(*args, **kwargs)
 
+    def forward_mps(self, *args, **kwargs):
+        # By default, we assume that MPS ops are compatible with the
+        # PyTorch-native implementation.
+        return self.forward_native(*args, **kwargs)
```
hmellor, line 81: Does this work with `forward_cuda`? It's what the MacOS CPU install uses

hmellor: There may be no benefit, but it'd be neat if there was

ericcurtin: Didn't try tbh

vllm/platforms/mps.py
```diff
@@ -0,0 +1,209 @@
...
+    device_control_env_var = "GPU_DEVICE_ORDINAL"
+
+    # MPS has limited operator support, use eager backend
+    simple_compile_backend: str = "eager"
+
+    @property
+    def supported_dtypes(self) -> list[torch.dtype]:
+        # MPS supports float16, bfloat16 (on newer Apple Silicon), and float32
+        # but does NOT support float64
+        return [torch.float16, torch.float32]
```
hmellor, line 46: Same logic as on CPU to know if we can use bf16

```suggestion
        # MPS supports float16, bfloat16 (on newer Apple Silicon), and float32
        # but does NOT support float64
        if (
            subprocess.check_output(
                ["sysctl -n hw.optional.arm.FEAT_BF16"], shell=True
            ).strip()
            == b"1"
        ):
            return [torch.bfloat16, torch.float16, torch.float32]
        return [torch.float16, torch.float32]
```

vllm/platforms/mps.py
```diff
@@ -0,0 +1,209 @@
...
+
+        # Disable CUDA graph equivalent features
+        from vllm.config import CompilationMode
+
+        vllm_config.compilation_config.cudagraph_capture_sizes = []
+
+        compilation_config = vllm_config.compilation_config
+        # Use eager mode for MPS - torch.compile has limited support
+        compilation_config.mode = CompilationMode.NONE
+        compilation_config.backend = "eager"
```
hmellor, line 167: Can we let the user decide this? I just tried 

```py1
        compilation_config.mode = CompilationMode.DYNAMO_TRACE_ONCE
        compilation_config.backend = "inductor"
```

and got a ~40x throughput increase on Qwen3 0.6B

ericcurtin: The throughput increased for me, but it was junk:

```
$ ../lm-chat/lm-chat.py --url http://127.0.0.1:8000/v1 run HuggingFaceTB/SmolLM-135M-Instruct Write an essay
orgetownDefineCHNphaltAk'"
:""ageNGPublisherFlorFieldsadinicuserCompMexicoin� safelyPlaceAnalyPresentationwidAmericaHeartFields
:Hel________Dutch'"�:Tagnat
```

Maybe a follow on PR when we debug? Could we log an issue to track this?

hmellor: Ah, I didn't check the quality of the outputs, I just saw big numbers 😅

vllm/platforms/mps.py
ericcurtin: Applied

vllm/v1/attention/backends/mps_attn.py
```diff
@@ -0,0 +1,612 @@
...
+
+    accept_output_buffer: bool = True
+    supported_dtypes: ClassVar[list[torch.dtype]] = [
+        torch.float16,
+        torch.float32,
+    ]
+
+    @classmethod
+    def get_supported_dtypes(cls) -> list[torch.dtype]:
+        return [torch.float16, torch.float32]
```
hmellor: Can we add bf16 here too

vllm/platforms/mps.py
ericcurtin: It's actually important here, I often encountered moments when trying to get this working where I thought, eureka! Only to check the output was junk

ericcurtin: I would like to propose optimising after we merge this. Once we get this initial support in, it makes it possible for many to hone in and try and squeeze performance out of this from multiple perspectives.

hmellor: Definitely agree with you here, we've got to go from 0 to 1 first

vllm/v1/attention/backends/mps_attn.py
ericcurtin: Added

vllm/v1/worker/mps_model_runner.py
```diff
@@ -0,0 +1,146 @@
...
+        """Load the model onto MPS device."""
+        logger.info("Starting to load model %s...", self.model_config.model)
+
+        # Load model - it will be loaded on MPS due to device config
+        self.model = get_model(vllm_config=self.vllm_config)
+
+        # Ensure model is on MPS
+        if not all(p.device.type == "mps" for p in self.model.parameters() if p.numel() > 0):
+            logger.info("Moving model to MPS device...")
+            self.model = self.model.to("mps")
```
hmellor: Is this left over from debugging or does the context manager that's specifies device not always work for MPS?

vllm/v1/worker/mps_model_runner.py
```diff
@@ -0,0 +1,146 @@
...
+from vllm.v1.worker.gpu_model_runner import GPUModelRunner
+
+logger = init_logger(__name__)
+
+
+class MPSModelRunner(GPUModelRunner):
+    """Model runner for MPS (Metal Performance Shaders) on Apple Silicon.
+
+    This runner keeps the model and input tensors on MPS for GPU acceleration,
+    but allocates the KV cache on CPU for numerical stability. The attention
```
hmellor: This is contradicted by the docstring of `initialize_kv_cache_tensors`?

vllm/v1/worker/mps_model_runner.py
```diff
@@ -0,0 +1,146 @@
...
+        """Initialize the KV cache tensors on MPS.
+
+        Keep KV cache on MPS to avoid cross-device issues.
+        """
+        # Keep KV cache on MPS (same as model)
+        kv_caches = super().initialize_kv_cache_tensors(kv_cache_config, kernel_block_sizes)
+
+        # Log KV cache device info
+        for name, cache in kv_caches.items():
+            logger.info("KV cache %s allocated on %s", name, cache.device)
```
hmellor: Is this also a sanity check left over from debugging?

vllm/v1/worker/mps_model_runner.py
```diff
@@ -0,0 +1,146 @@
...
+        with _set_global_compilation_settings(self.vllm_config):
+            self._dummy_run(
+                min(
+                    max(16, self.max_num_reqs),
+                    self.scheduler_config.max_num_batched_tokens,
+                )
+            )
+        # Synchronize MPS
+        if hasattr(torch.mps, 'synchronize'):
+            torch.mps.synchronize()
```
hmellor: Is it important that this is synchronized here? We could remove this override if not.

vllm/v1/worker/mps_model_runner.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def _to_list(self, sampled_token_ids: torch.Tensor) -> list[list[int]]:
+        """Convert sampled token IDs from MPS tensor to list.
+
+        Override the parent class to use direct tolist() since MPS doesn't
+        support CUDA events. This is faster than sync + pinned memory copy
+        for small tensors.
+        """
+        # For MPS, direct tolist() is simpler and avoids extra sync overhead
+        # The tensor is small (batch_size x 1) so the copy is fast
```
hmellor: Looks like the `_torch_cuda_wrapper` should handle the CUDA stuff that's not compatible. If this override is still valid for performance reasons can we replace the explanation in the docstring with the one below?

vllm/v1/worker/mps_model_runner.py
```diff
@@ -0,0 +1,146 @@
...
+
+    cuda_event = torch.Event
+    cuda_stream = torch.cuda.Stream
+    try:
+        torch.Event = _EventPlaceholder
+        torch.cuda.Stream = _StreamPlaceholder
+        yield
+    finally:
+        torch.Event = cuda_event
+        torch.cuda.Stream = cuda_stream
```
hmellor: Could/should we import this from the cpu runner? It's probably safe to assume that any CUDA stuff that CPU can't do, MPS can't do it either. Then if the things are changed in the GPU runner and the version of this context manager from the CPU runner is updated, MPS will automatically receive the update too.

hmellor (commented):
This looks really cool, most of my comments are about making the code change as minimal as possible so that the future maintenance burden of supporting MPS is minimal

vllm/v1/worker/mps_model_runner.py
ericcurtin: Removed

vllm/v1/worker/mps_model_runner.py
ericcurtin: Removed

vllm/v1/worker/mps_model_runner.py
ericcurtin: Removed

vllm/v1/worker/mps_model_runner.py
ericcurtin: Replaced the explanation about CUDA events

vllm/v1/worker/mps_model_runner.py
ericcurtin: Now gone

vllm/v1/worker/mps_model_runner.py
ericcurtin: Imported _torch_cuda_wrapper from cpu_model_runner

vllm/platforms/mps.py
```diff
@@ -0,0 +1,222 @@
...
+        from vllm.utils.mem_constants import GiB_bytes
+
+        # Check for user-specified KV cache space first
+        kv_cache_space = envs.VLLM_CPU_KVCACHE_SPACE
+        if kv_cache_space is None:
+            # Use a portion of system memory for MPS
+            # Apple Silicon shares memory between CPU and GPU
+            total_memory = psutil.virtual_memory().total
+            # Use 25% of system memory for KV cache by default
+            kv_cache_space = total_memory // 4
```
hmellor: Out of curiosity, why 25%? The CPU version of MacOS would use 50%

vllm/platforms/mps.py
```diff
@@ -0,0 +1,222 @@
...
+        else:
+            kv_cache_space = int(kv_cache_space * GiB_bytes)
+
+        return kv_cache_space
+
+    @classmethod
+    def set_device(cls, device: torch.device) -> None:
+        """Set the device for the current platform."""
+        # MPS only has one device, no need to set
+        pass
```
hmellor: Even though there's only 1 device, should we set it anyway? Not doing so may have side effects

vllm/v1/attention/backends/mps_attn.py
```diff
@@ -0,0 +1,615 @@
...
+                attn_metadata.slot_mapping[:num_actual_tokens],
+            )
+
+        # Compute attention
+        # For prefill (max_query_len == max_seq_len): use original Q/K/V directly
+        # For decode (max_query_len == 1): read K/V from cache
+        max_query_len = attn_metadata.max_query_len
+        max_seq_len = attn_metadata.max_seq_len
+
+        if max_query_len == max_seq_len:
```
hmellor: Does this mean that we cannot prefill if the prompt is <`max_seq_len`?

vllm/v1/attention/backends/mps_attn.py
```diff
@@ -0,0 +1,615 @@
...
+            # Prefill: use original tensors directly for accuracy
+            return self._compute_attention_no_cache(
+                query[:num_actual_tokens],
+                key[:num_actual_tokens],
+                value[:num_actual_tokens],
+                output[:num_actual_tokens],
```
hmellor: The CPU version begins this slice at `num_decode_tokens` in prefill, why don't we do that here?

vllm/v1/attention/backends/mps_attn.py
```diff
@@ -0,0 +1,615 @@
...
+            and key is not None
+            and value is not None
+        ):
+            self._update_kv_cache_vectorized(
+                key[:num_actual_tokens],
+                value[:num_actual_tokens],
+                key_cache,
+                value_cache,
+                attn_metadata.slot_mapping[:num_actual_tokens],
+            )
```
hmellor, line 308: Given that the memory is unified, what's to stop us from using the CPU custom ops for KV cache manipulation?

vllm/platforms/mps.py
ericcurtin: Changed to 50%

vllm/platforms/mps.py
ericcurtin: Now calls torch.mps._set_default_device(device)

vllm/v1/attention/backends/mps_attn.py
ericcurtin: Changed from using max_query_len == max_seq_len condition to using split_decodes_and_prefills

vllm/v1/attention/backends/mps_attn.py
ericcurtin: Changed prefill to slice from num_decode_tokens to num_actual_tokens instead of 0 to num_actual_tokens

vllm/v1/attention/backends/mps_attn.py
ericcurtin: I thought running on GPU shader cores is preferred over CPU cores for parallelism

NickLucche (commented):
Hey @ericcurtin thanks a lot for the contribution here!

Given that you're already re-defining a new model-runner + worker, I would like to re-iterate on @hmellor point to support metal as an external plugin https://blog.vllm.ai/2025/05/12/hardware-plugin.html.
As of today, plugins are best practice and recommended way to add support for new hardware, allowing vllm to keep maintenance cost under control. 
There's many of them already supported for different hw platforms, and some major ones are linked in the blog post above.

ericcurtin:
I think monorepos have shown their value multiple times for long-term healthy projects: linux, freebsd, mesa, libcamera, git, systemd, kubernetes, llama.cpp, etc. It's a long list. I think the overhead of syncing versioning of changes between multiple smaller repos is even more overhead than maintenance. I do think vllm project is shooting themselves in the foot here long-term. Keep in mind this has a chance to become the number 1 desktop/laptop development platform for vLLM. An M4 Mac Mini with 16GB RAM is like 600 dollars these days, an Nvidia laptop/desktop with roughly the same amount of VRAM available is what, maybe 4,000 dollars? Nonetheless if a plugin is desired, could someone create this for us:

https://github.com/vllm-project/vllm-metal

ericcurtin:
@robertgshaw2-redhat @mgoin PTAL

aniolekx:
Can I already test these changes? Will they work on the Apple GPU?
I’m currently running continuous batching experiments on an M3 Ultra with 512 GB RAM. I’ve already evaluated both llama.cpp and mlx-lm, and continuous batching works well in both cases. For the 49-task Aider JavaScript benchmark, which normally takes about 26 minutes, I can reduce the runtime to around 10 minutes using Qwen3-30B-A3B-Instruct in Q6_K (6-bit) format.

mgoin:
I think this is a cool idea and would like to see the mac backend developed, but I agree with Nicolo that we want future hardware backends to go through the plugin interface. This is especially pertinent now as we are refactoring the gpu_model_runner to be very much hardcoded for cuda/rocm, so we are already planning to separate out of the CPU runner from being directly inheriting from the GPU runner. 

I've gone ahead and made the https://github.com/vllm-project/vllm-metal repo so we can start building the plugin. If anyone would like join in on the work, I'm happy to invite collaborators as needed. I also started a [#sig-metal](https://vllm-dev.slack.com/archives/C0A3B1RQYLA) channel in vllm slack where we can specifically discuss on the project.

PR #29646, opened 2025-11-28
[Core] Rename PassConfig flags as per RFC #27995

arpitkh101:
## Purpose
Implements the PassConfig flag renaming from issue #27995.

Changes:
- Renamed all PassConfig flags to be more descriptive (e.g., enable_fusion → fuse_norm_quant + fuse_act_quant)
- Added deprecation warnings for old flag names
- Old flags still work for backward compatibility
- Updated all internal code and tests to use new names


vllm/config/compilation.py
```diff
@@ -168,18 +186,77 @@ def compute_hash(self) -> str:
...
+            if not self.fuse_gemm_comms:
+                self.fuse_gemm_comms = self.enable_async_tp
+
+        if self.enable_fi_allreduce_fusion is not None:
+            logger.warning_once(
+                "enable_fi_allreduce_fusion is deprecated and will be "
+                "removed in a future release. Use fuse_allreduce_rms instead."
+            )
+            if not self.fuse_allreduce_rms:
+                self.fuse_allreduce_rms = self.enable_fi_allreduce_fusion
```
hmellor: Could we add a utility to `vllm/config/utils.py` that takes the old name, the new name and the removal version? We do this often enough that a utility like this would be nice to have.

i.e. in utills:

```py
def handle_deprecated(config: ConfigT, old_name, new_name, removal_version):
   old_val = getattr(config, old_name)
    if old_val is None:
        return
    logger.warning(
        "%s is deprecated and will be removed in %s. Use %s instead",
        old_name, removal_version, new_name,
    )
    setattr(config, new_name, old_val)
```

and then here:

```py
        handle_deprecated(self, "enable_noop", "eliminate_noops", "v0.13.0")
        ...
```

tests/compile/distributed/test_sequence_parallelism.py
```diff
@@ -153,7 +153,10 @@ def ops_in_model_before(self):
         ]
 
     def ops_in_model(self):
-        if self.vllm_config.compilation_config.pass_config.enable_fusion:
+        if (
+            self.vllm_config.compilation_config.pass_config.fuse_norm_quant
+            and self.vllm_config.compilation_config.pass_config.fuse_act_quant
+        ):
```
ProExpertProg: Just rmsnorm-quant here:
```suggestion
        if self.vllm_config.compilation_config.pass_config.fuse_norm_quant:
```

tests/compile/distributed/test_sequence_parallelism.py
```diff
@@ -183,7 +186,8 @@ def ops_in_model(self):
 @pytest.mark.parametrize("seq_len", [16])
 @pytest.mark.parametrize("hidden_size", [16])
 @pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16])
-@pytest.mark.parametrize("enable_fusion", [True, False])
+@pytest.mark.parametrize("fuse_norm_quant", [True, False])
+@pytest.mark.parametrize("fuse_act_quant", [True, False])
```
ProExpertProg: act quant shouldn't matter here

vllm/config/compilation.py
```diff
@@ -105,18 +105,36 @@ class PassConfig:
     improper state.
     """
 
+    # New flags
+    fuse_norm_quant: bool = Field(default=False)
+    """Whether to enable the custom RMSNorm+quant fusion pass."""
```
ProExpertProg: While we're at it, let's simplify the phrasing:
```suggestion
    """Fuse the custom RMSNorm + quant ops."""
```

vllm/config/compilation.py
```diff
@@ -105,18 +105,36 @@ class PassConfig:
...
+    eliminate_noops: bool = Field(default=False)
+    """Whether to enable the custom no-op elimination pass."""
+    enable_sp: bool = Field(default=False)
+    """Whether to enable sequence parallelism."""
+    fuse_gemm_comms: bool = Field(default=False)
+    """Whether to enable async TP."""
+    fuse_allreduce_rms: bool = Field(default=False)
+    """Whether to enable flashinfer allreduce fusion."""
+
+    # Deprecated flags
```
ProExpertProg, line 124: Add deprecation version

vllm/config/vllm.py
```diff
@@ -107,12 +108,13 @@ def enable_fusion(cfg: "VllmConfig") -> bool:
...
         "pass_config": {
-            "enable_noop": True,
-            "enable_fusion": enable_fusion,
-            "enable_fi_allreduce_fusion": False,
-            "enable_attn_fusion": False,
-            "enable_sequence_parallelism": False,
-            "enable_async_tp": False,
+            "eliminate_noops": True,
+            "fuse_norm_quant": enable_fusion,
+            "fuse_act_quant": enable_fusion,
```
ProExpertProg: need separate enabling logic

ProExpertProg (commented):
A few minor notes, looks ready otherwise! Could you also add a test that checks that specifying the old fields results in new fields set correctly and a warning (see #29557 for example)

vllm/config/vllm.py
arpitkh101: Could you confirm this once?
RMSNormQuantFusionPass will need both rms_norm and quant_fp8, whereas ActivationQuantFusionPass would need only quant_fp8.

vllm/config/compilation.py
arpitkh101: What exact version should I use?
Deprecated in: v0.12.0?
Will be removed in: v0.13.0?

vllm/config/vllm.py
ProExpertProg: `ActivationQuantFusionPass` also needs `silu_and_mul`

vllm/config/compilation.py
ProExpertProg: "Removed in v0.13.0 or v1.0.0, whichever is sooner"

tests/compile/distributed/test_sequence_parallelism.py
```diff
@@ -297,7 +301,7 @@ def sequence_parallelism_pass_on_test_model(
             sequence_parallelism_pass,
         ]
 
-        if enable_fusion:
+        if fuse_norm_quant or fuse_act_quant:
```
ProExpertProg:
```suggestion
        if fuse_norm_quant:
```

tests/compile/distributed/test_sequence_parallelism.py
```diff
@@ -193,7 +193,8 @@ def test_sequence_parallelism_pass(
     seq_len: int,
     hidden_size: int,
     dtype: torch.dtype,
-    enable_fusion: bool,
+    fuse_norm_quant: bool,
+    fuse_act_quant: bool,
```
ProExpertProg: Not relevant

tests/compile/test_config.py
```diff
@@ -375,3 +378,57 @@ def test_cudagraph_sizes_post_init(
             vllm_config.compilation_config.max_cudagraph_capture_size
             == expected_max_size
         )
+
+
+def test_pass_config_deprecation(caplog):
```
ProExpertProg: Use caplog_vllm fixture that handles propagation

vllm/config/compilation.py
```diff
@@ -190,18 +216,71 @@ def _skip_none_validation(cls, value: Any, handler: Callable) -> Any:
...
+            "enable_attn_fusion",
+            "fuse_attn_quant",
+            "v0.13.0",
+        )
+
+        handle_deprecated(
+            self,
+            "enable_sequence_parallelism",
+            "enable_sp",
+            "v0.13.0",
```
ProExpertProg: Extract to variable also
```suggestion
            "v0.13.0 or v1.0.0, whichever is sooner",
```

vllm/config/utils.py
```diff
@@ -293,3 +297,28 @@ def get_hash_factors(config: ConfigT, ignored_factors: set[str]) -> dict[str, ob
...
+    else:
+        new_names = new_name_or_names
+
+    msg = (
+        f"{old_name} is deprecated and will be removed in {removal_version}. "
+        f"Use {', '.join(new_names)} instead."
+    )
+    logger.warning(msg)
+
+    for new_name in new_names:
```
ProExpertProg, line 323: @hmellor should we delete these attrs, set them to none, or set them to false? It would be good to make sure no one is using the old attrs.

Also check in tests that old attribute is set to whatever we decide to do.
```suggestion
    delattr(config, old_name)

    for new_name in new_names:
```

vllm/config/vllm.py
```diff
@@ -83,22 +83,32 @@ class OptimizationLevel(IntEnum):
...
+    """Returns True if both RMS norm and quant FP8 custom ops are enabled.
+    RMSNormQuantFusionPass requires both."""
     return cfg.compilation_config.is_custom_op_enabled(
         "rms_norm"
-    ) or cfg.compilation_config.is_custom_op_enabled("quant_fp8")
+    ) and cfg.compilation_config.is_custom_op_enabled("quant_fp8")
+
+
+def enable_act_fusion(cfg: "VllmConfig") -> bool:
+    """Returns True if quant FP8 custom op is enabled.
```
ProExpertProg: Wrong comment. Also shorten no need to say "returns true" just describe why we enable the fusions if either custom op enabled (BCS if both are disabled, inductor can fuse for us)

tests/compile/test_config.py
```diff
@@ -219,15 +221,15 @@ def test_splitting_ops_dynamic():
         compilation_config=CompilationConfig(
             mode=CompilationMode.VLLM_COMPILE,
             use_inductor_graph_partition=True,
-            pass_config={"enable_attn_fusion": True, "enable_noop": True},
+            pass_config={"fuse_attn_quant": True, "eliminate_noops": True},
```
ProExpertProg: Use `PassConfig` instead of dicts for type checking

vllm/config/utils.py
hmellor: My understanding of deprecation is that the old way still works with a warning. Therefore, we should keep the old attrs around until they are deleted. This is the least disruptive way to deprecate things.

arpitkh101: I forced these variables to None in the `compilation.py` file. Do you want me to include that in this util?

hmellor: Leaving them as `None` by default is fine. I just mean that if a user explicitly passes the old name, we shouldn't delete it. They will get very confused 😅

ProExpertProg (approved):

vllm/config/compilation.py
```diff
@@ -190,18 +222,71 @@ def _skip_none_validation(cls, value: Any, handler: Callable) -> Any:
...
+            "v0.13.0 or v1.0.0, whichever is sooner",
+        )
+
+        # Force old flags to None to ensure they are not used
+        self.enable_fusion = None
+        self.enable_attn_fusion = None
+        self.enable_noop = None
+        self.enable_sequence_parallelism = None
+        self.enable_async_tp = None
+        self.enable_fi_allreduce_fusion = None
```
hmellor, line 276: We shouldn't really do this. This doesn't ensure they aren't used, just that they aren't what the user expected

vllm/config/compilation.py
```diff
@@ -105,18 +105,43 @@ class PassConfig:
...
+    fuse_attn_quant: bool = Field(default=None)
+    """Fuse the custom attention + quant ops."""
+    eliminate_noops: bool = Field(default=None)
+    """Eliminate no-op ops."""
+    enable_sp: bool = Field(default=None)
+    """Enable sequence parallelism."""
+    fuse_gemm_comms: bool = Field(default=None)
+    """Enable async TP."""
+    fuse_allreduce_rms: bool = Field(default=None)
+    """Enable flashinfer allreduce fusion."""
```
hmellor, line 122: Why do the new flags default to None? I know the old flags did, but that was confusing.

hmellor (commented):
Couple of nits, not necessary for the release but would be nice to clean up asap.

vllm/config/compilation.py
ProExpertProg: That's because the defaults depend on the optimization level, which was merged recently

arpitkh101: I didn't kept them None initially. But, the existing, older flags in the `PassConfig` and `compilation.py` file were also initialized to None. Also, there is an explicit unit test, `tests/test_config.test_vllm_config_defaults_are_none`, which requires these configuration fields to default to None. Changing the default to False would cause this test to fail.

vllm/config/compilation.py
ProExpertProg: What do you propose instead?

hmellor: Just leave them, for other deprecated fields we leave their value at whatever the user set

vllm/config/compilation.py
hmellor: Ok, that makes sense thanks for the explanation!

vllm/config/compilation.py
ProExpertProg: @arpitkh101 can you create a follow-up with this change?

arpitkh101: Created follow-up PR #29994 to address this!

PR #29653, opened 2025-11-28
fix potential object has no attribute 'bias' error

allerou4:
## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/layers/linear.py
+++ b/vllm/model_executor/layers/linear.py
@@ -577,7 +577,7 @@ class ColumnParallelLinear(LinearBase):
     def extra_repr(self) -> str:
         s = f"in_features={self.input_size}"
         s += f", output_features={self.output_size_per_partition}"
-        s += f", bias={self.bias is not None}"
+        s += f", bias={getattr(self, 'bias', None) is not None}"
         s += f", tp_size={self.tp_size}"
         s += f", gather_output={self.gather_output}"
         return s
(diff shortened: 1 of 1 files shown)


DarkLight1337:
In which case can this happen? cc @mgoin @Isotr0py @jeejeelee

allerou4:
> > I found this when I was debugging using pdb, I tried to print some values but this bug blocked printing.
> 
> Can you elaborate the bug? I think the root issue is that some layer's `extra_repr` implemented wrong. This fix looks unlike an ideal solution.

https://github.com/vllm-project/vllm/blob/5c2b5cb422182ae2a0c22a17532d465174396613/vllm/model_executor/layers/linear.py#L440

So originally bias is a bool，then if bias=True, it becomes a attribute. If bias = Flase, then if we try to print it, 
s += f", bias={self.bias is not None}" will throw an error. 

```python
    def __init__(
        self,
        input_size: int,
        output_size: int,
        bias: bool = True,
    )
        if bias:
            self.bias = Parameter(
                torch.empty(self.output_size_per_partition, dtype=params_dtype)
            )
            set_weight_attrs(
                self.bias,
                {
                    "output_dim": 0,
                    "weight_loader": self.weight_loader,
                },
            )
        else:
            self.register_parameter("bias", None)
```

Isotr0py:
Hmmm, hasn't `self.register_parameter("bias", None)` already set `bias` as attribute with None?
https://github.com/vllm-project/vllm/blob/953d9c820b0d8f9a55ee455814287536a498498c/vllm/model_executor/layers/linear.py#L509

allerou4:
> Hmmm, hasn't `self.register_parameter("bias", None)` already set `bias` as attribute with None?
> 
> https://github.com/vllm-project/vllm/blob/953d9c820b0d8f9a55ee455814287536a498498c/vllm/model_executor/layers/linear.py#L509

Yeah, right. I figured that this only happens when you try to print anything in PDB when you're at the "def __init__(", cause self.bias isn't initialized (I was trying to debug the "self.quant_method.create_weights()", self.bias initialization happens after this )

so we leave it be?

PR #29655, opened 2025-11-28
[Doc] Improve abnormal information string

maang-h:
## Purpose
`ValueError ` constructor only accepts one parameter, and when used in the current code, more than one parameter value is passed in, resulting in these parameter values being treated as a tuple and unable to visually indicate exception information.

#### Before: 
Inaccurate abnormal information before improvement, such as:
`"('Not enough resources to allocate %s DP ranks on DP master node %s, possible to fit %s DP ranks', 4, '192.168.1.100', 0)"`

#### After
After improvement, it can be displayed very intuitively, such as:
`"Not enough resources to allocate 4 DP ranks on DP master node 192.168.1.100, possible to fit 0 DP ranks."`

#### changes
- `vllm/v1/engine/utils.py` line 373-375: `assert`
- `vllm/v1/engine/utils.py` line 447-449: `assert`
- `vllm/v1/engine/utils.py` line 464-468: `ValueError`

## Test Plan
None

## Test Result
None


tests/v1/engine/test_utils.py
```diff
@@ -0,0 +1,162 @@
+# SPDX-License-Identifier: Apache-2.0
```
DarkLight1337: I think this test is not really necessary, it's just a UX issue

DarkLight1337 (approved):
Thanks!

PR #29661, opened 2025-11-28
Enable FlashInfer Hopper FP8 attention

nvpohanh:
## Purpose

Enable the best-performing Hopper FP8 attention kernels in FlashInfer backend.

In Prefill stage, the FlashInfer FA3 backend will be used. In Decode stage, the FlashInfer XQA backend (using trtllm interface) will be used.

However, things are complicated because different backends have different requirements. The most noticeable difference is that when kv-cache dtype is FP8, FA3 backend requires the query to be also in FP8, but XQA backend requires the query to be in FP16/BF16. Therefore, we cannot apply query quantization outside of the attention custom op. Instead, we must apply the query quantization inside the attention custom op's forward().

## Changes

In `vllm/utils/flashinfer.py`:

- Change `supports_trtllm_attention()` to `check_trtllm_attention_support()` which now returns:
  - True when we "must" use TRTLLM with the reason.
  - False when we "must not" use TRTLLM with the reason.
  - None when we can use TRTLLM but not a must.
- Change `use_trtllm_attention()` such that:
  - It decides whether to use TRTLLM based on `check_trtllm_attention_support()` and `force_use_trtllm_attention()`.
  - If the decision does not match `force_use_trtllm_attention()`, print a warning with a reason. Otherwise, print an info with a reason.

In `vllm/v1/attention/backends/flashinfer.py`:

- Change the decision of whether to use TRTLLM from runtime (in `build()`) to initialization time (in `__init__()`).
- Split `q_data_type` in metadata into `q_data_type_prefill` and `q_data_type_decode`.
  - Generally, these two will be the same, except on SM90 when FP8 kv-cache is used.
- Add a few classmethods to avoid duplicated code between `__init__()` and `get_cudagraph_support()` (the latter is a classmethod).
- Update args passed into FlashInfer interface to support out_dtype and pass in q_scale.

Requires https://github.com/flashinfer-ai/flashinfer/pull/2148 from FlashInfer.

## Test Plan

Run accuracy + performance tests on the cross-product of the following attributes:

- {H200, B200}
- {Qwen3-32B-FP8, GPT-OSS-120b}
- {TP1 conc16, TP1 conc512}
- {auto kv-cache dtype, fp8 kv-cache}
- {default attn backend, FLASHINFER attn backend, FLASHINFER attn backend + disable TRTLLM attn}

<details><summary>Script to run</summary>
<code>
TBA
</code>
</details>  

## Test Result

TBA

---
<details>

</details>

### DIFF
--- a/vllm/v1/attention/backends/flashinfer.py
+++ b/vllm/v1/attention/backends/flashinfer.py
@@ -18,6 +18,7 @@ from flashinfer.prefill import trtllm_batch_context_with_kv_cache
 from flashinfer.utils import FP4Tensor
 from typing_extensions import override
 
+from vllm import _custom_ops as custom_ops
 from vllm import envs
 from vllm.attention.backends.abstract import (
     AttentionBackend,
@@ -43,7 +44,8 @@ from vllm.platforms import current_platform
 from vllm.platforms.interface import DeviceCapability
 from vllm.triton_utils import tl, triton
 from vllm.utils.flashinfer import (
-    can_use_trtllm_attention,
+    check_trtllm_attention_support,
+    force_use_trtllm_attention,
     use_trtllm_attention,
 )
 from vllm.utils.math_utils import cdiv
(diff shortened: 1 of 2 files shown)


vllm/v1/attention/backends/flashinfer.py
```diff
@@ -594,12 +581,35 @@ def __init__(
...
+            num_kv_heads=self.num_kv_heads,
+            dcp_world_size=self.dcp_world_size,
+            kv_cache_dtype=self.kv_cache_dtype,
+            q_data_type=self.q_data_type_decode,
+            has_sinks=self.has_sinks,
+            has_spec=self.has_spec,
+        )
+
+        # Only TRTLLM attention supports spec-as-decode.
+        self._init_reorder_batch_threshold(
```
nvpohanh, line 609: @pavanimajety I am trying to move the `prefill_use_trtllm`/`decode_use_trtllm` decision logic to runtime (`build()`) but what about this `_init_reorder_batch_threshold()`? It needs to know whether we will use trtllm for decode... should we call it at runtime?

nvpohanh:
cc @elvischenv

PR #29666, opened 2025-11-28
[mypy] Pass type checking for `vllm/utils` and `vllm/v1/pool`

DarkLight1337:
## Purpose

Improve mypy coverage

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/utils/async_utils.py
+++ b/vllm/utils/async_utils.py
@@ -257,6 +257,10 @@ def in_loop(event_loop: AbstractEventLoop) -> bool:
         return False
 
 
+def _coro_anext(it: AsyncGenerator[T, None]):
+    return anext(it)
+
+
 async def merge_async_iterators(
     *iterators: AsyncGenerator[T, None],
 ) -> AsyncGenerator[tuple[int, T], None]:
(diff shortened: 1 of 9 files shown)


vllm/utils/async_utils.py
```diff
@@ -257,6 +257,13 @@ def in_loop(event_loop: AbstractEventLoop) -> bool:
         return False
 
 
+# A hack to pass mypy
+if TYPE_CHECKING:
+
+    def anext(it: AsyncGenerator[T, None]):
+        return it.__anext__()
```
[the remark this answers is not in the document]
DarkLight1337: The min version is Python 3.10 so it doesn't matter

Isotr0py (approved):

vllm/utils/async_utils.py
hmellor, line 264: I'm curious to know what error mypy was giving before this workaround?

DarkLight1337: It thinks that `anext(it)` returns `Awaitable[T]` instead of `Coroutine[T]`, the former cannot be passed to `loop.create_task`.

hmellor: According to [the docs](https://docs.python.org/3/library/functions.html#anext), it does return an awaitable. Perhaps we have a bug?

DarkLight1337: I think it's an issue with mypy. In VSCode (which uses pyright), the type is determined correctly.

PR #29667, opened 2025-11-28
[Bugfix] Fix O(n²) multimodal string prompt processing

mertunsall:
## Problem
`_apply_matches()` had O(n²) complexity - a 5000 token multimodal prompt took 2.6s for processing alone.

## Cause  
When no match found, the loop incremented `start_idx` but `_find_matches()` used `prev_end_idx` (unchanged), re-scanning the entire prompt each iteration.

## Fix
Exit loop when no matches remain instead of incrementing uselessly.

## Impact
- InternVL: 2.6s → <10ms for 5k tokens
- Models using token-based matching (Qwen3-VL) unaffected

## Test Plan
python -m pytest tests/multimodal/test_processing.py::TestApplyMatchesPerformance -v

## Test Result
Passed


vllm/multimodal/processing.py
```diff
@@ -777,7 +777,7 @@ def _apply_matches(
                 start_idx = prev_end_idx = match.end_idx
 
         if not found:
-            start_idx += 1
+            break  # No more matches to find, exit loop
```
DarkLight1337: This is incorrect, ~~your code assumes that multimodal placeholders are consecutive which isn't always true~~

mertunsall: could you please help write a test to demonstrate this?

DarkLight1337: Actually that is not the real problem. It will fail if the target for replacement don't occur at the very start of the string. So an example like

```
prompt = [0, image_token_id]
mm_prompt_updates = {"image": ResolvedPromptUpdate("image", 0, "insert", [image_token_id], PromptUpdateDetails([image_token_id] * 10))
```

already fails.

mertunsall: I added some tests and it seems to pass

mertunsall: hmm let me see this case

DarkLight1337: Sorry, I confused `start_idx` with `prev_end_idx`. It seems that we can actually replace `start_idx` with `prev_end_idx`.

mertunsall: are you sure about this? I created a test case for this and it passes.

DarkLight1337: I mean my previous statement

>  It will fail if the target for replacement don't occur at the very start of the string.

is actually incorrect, as I confused `start_idx` with `prev_end_idx`.

mertunsall: yepp exactly

mertunsall: do you think this is good to merge?

DarkLight1337: After thinking about it more, I think I understand your logic now. You can indeed return early if no match is found for the whole prompt since that implies that no match will be found for the partial prompt either.

DarkLight1337: I have another optimization in progress: https://github.com/vllm-project/vllm/pull/29668

It would be best to verify that the test that you newly added fails even after that is merged, before merging this PR.

DarkLight1337: And can you update this PR to get rid of `start_idx`?

mertunsall: Please see below my performance testing script - you can see the impact end-to-end directly on the internVL models

mertunsall: could you directly compare performance with that script? I think it's the best if we see the end-to-end performance instead of relying on the test (at least for the performance part)

DarkLight1337: The purpose is to ensure that the test that is going to run in CI actually does its job at preventing regressions. If after the other optimization, this test passes without this PR, then maybe the test isn't strict enough or a different input needs to be used.

mertunsall: let me ensure that that is the case

mertunsall: yep I think it is the case with current tests!

tests/multimodal/test_processing.py
```diff
@@ -1075,3 +1080,228 @@ def test_hf_processor_call_kwargs(
 
     result = ctx.call_hf_processor(processor, {}, inference_kwargs)
     assert result == expected_kwargs
+
+
+class TestApplyMatchesPerformance:
```
DarkLight1337: I think the correctness tests should be covered by the existing tests in this file already, can you keep only the ones pertaining to performance?

DarkLight1337: And also just put the test functions in global scope, no need for the test suite class

mertunsall: done!

ywang96 (approved):

mertunsall:
What’s the reason checks don’t pass?

mertunsall:
should we update? @ywang96

ywang96:
> should we update? @ywang96

Should be fine - the CI's still running and we don't need the PR to be completely aligned with main to be merged

PR #29671, opened 2025-11-28
hfrunner.classify should return list[list[float]] not list[str]

nwaughachukwuma:
## Purpose
Fix the return type of HfRunner.classify - should be list[list[float]] and not list[str]

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -478,6 +478,9 @@ class HfRunner:
         for inputs in all_inputs:
             output = self.model(**self.wrap_device(inputs))
             # NOTE: In practice, HF models return ModelOutput with a `logits` attribute (torch.Tensor).
+            if not hasattr(output, "logits"):
+                raise TypeError("Model output does not have a 'logits' attribute.")
+
             if not isinstance(output.logits, torch.Tensor):
                 raise TypeError(
                     f"'logits' is {type(output.logits)}, expected torch.Tensor"
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -467,7 +467,7 @@ class HfRunner:
 
         for inputs in all_inputs:
             output = self.model(**self.wrap_device(inputs))
-            # NOTE: In practice, HF models return ModelOutput with a `logits` attribute (torch.Tensor).
+            # NOTE: In practice, HF models return ModelOutput with a `logits` attribute (torch.Tensor)
             if not hasattr(output, "logits"):
                 raise TypeError("Model output does not have a 'logits' attribute.")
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -459,14 +459,33 @@ def get_prompt_embeddings(self, prompts: list[str]) -> list[torch.Tensor]:
...
+    def classify(self, prompts: list[str]) -> list[list[float]]:
+        from sentence_transformers import SentenceTransformer, CrossEncoder
+
+        # Prevent misuse with incompatible model types
+        if isinstance(self.model, (SentenceTransformer, CrossEncoder)):
+            raise TypeError(
+                "classify() is only supported for standard Hugging Face classification models "
+                "(e.g., AutoModelForSequenceClassification). "
+                "SentenceTransformer and CrossEncoder are not supported."
+            )
```
noooop: Thanks for your contribution.

The tests/conftest.py is only used by developers, so we should trust that they know what they're doing. There's no need for such excessive safeguards.

Please make DCO happy.

nwaughachukwuma: Thanks for the feedback, @noooop.

I've removed that code block. Should be fine now.

tests/conftest.py
```diff
@@ -459,14 +459,23 @@ def get_prompt_embeddings(self, prompts: list[str]) -> list[torch.Tensor]:
...
             output = self.model(**self.wrap_device(inputs))
+            # NOTE: In practice, HF models return ModelOutput with a `logits` attribute (torch.Tensor)
+            if not hasattr(output, "logits"):
+                raise TypeError("Model output does not have a 'logits' attribute.")
+
+            if not isinstance(output.logits, torch.Tensor):
+                raise TypeError(
+                    f"'logits' is {type(output.logits)}, expected torch.Tensor"
+                )
+
```
noooop: This check may not be necessary, as developers would ultimately need to debug the issue if the ModelOutput is missing a logits attribute.

nwaughachukwuma: Ok, I kept it there for type safety. Given developers would ultimately debug if there's an issue, I've replaced it with an assert statement. I hope it works better.

```py
# Before
if not hasattr(output, "logits"):
   raise TypeError("Model output does not have a 'logits' attribute.")

if not isinstance(output.logits, torch.Tensor):
   raise TypeError(f"'logits' is {type(output.logits)}, expected torch.Tensor")

# After
assert isinstance(output.logits, torch.Tensor)
```
@noooop

noooop (approved):

PR #29477, opened 2025-11-26
[Feature][Bench] Add pareto visualization

lengrongfu:
## Purpose

Current `vllm bench sweep` can running 

The current `vllm bench sweep` test performs multiple rounds of testing and yields a solution. While `vllm bench sweep plot` can show the relationships between metrics, it cannot plot Pareto Optimality because it requires some metric manipulation.

The current pull request provides a subcommand `vllm bench sweep plot_pareto` that can directly plot Pareto Optimality based on the results of the `sweep` test.

## Test Plan

Let's say we're doing a test on a machine with 8 GPUs.

1. Set serve param, perhaps this parameter combination is incorrect; I'm just explaining how to use it.
```
[
   {
       "tensor_parallel_size":1,
       "data_parallel_size": 8
   },
   {
       "tensor_parallel_size":2,
       "data_parallel_size": 4
   },
      {
       "tensor_parallel_size":4,
       "data_parallel_size": 2
   },
     {
       "tensor_parallel_size":8,
       "data_parallel_size": 1
   }
]
```

2. Set bench param
```
[
    {
        "max_concurrency": 5
    },
    {
        "max_concurrency": 10
    },
    {
        "max_concurrency": 15
    },
    {
        "max_concurrency": 20
    },
    {
        "max_concurrency": 25
    },
    {
        "max_concurrency": 30
    }
]
```

3. Start sweep, because i only one gpu, so not need serve param.
```
$ python3 -m vllm.benchmarks.sweep.serve \
    --serve-cmd 'vllm serve Qwen/Qwen3-0.6B --gpu-memory-utilization 0.2' \
    --bench-cmd 'vllm bench serve --model Qwen/Qwen3-0.6B --backend vllm --endpoint /v1/completions --dataset-name sharegpt --dataset-path target/lrf/benchmarks/ShareGPT_V3_unfiltered_cleaned_split.json' \
    --bench-params target/lrf/benchmarks/bench_params.json \
    -o target/lrf/benchmarks/results

$ python3 -m vllm.benchmarks.sweep.plot_pareto target/lrf/benchmarks/results/20251125_192250/
```

We can set `--label-by` to show the configuration used in each data point, default is `max_concurrency,gpu_count`,
image

but we can customer label, like this `--label-by max_concurrency,tp,pp,dp`, it plot is

image

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

[...]

PR #29689, opened 2025-11-28
[Docs] Add CLI reference doc for `vllm bench sweep plot_pareto`

hmellor:
This was missing from https://github.com/vllm-project/vllm/pull/29477


DarkLight1337 (approved):
Oops I forgot about that, thanks for adding this in!

PR #29691, opened 2025-11-28
[Kernel]Support W4A8 Grouped GEMM on Hopper

czhu-cohere:
## Purpose
As title; the benefit of W4A8 is it can use fp8 tensor cores while still maintaining the low memory footprint of W4A16 (with negligible quality loss). In addition there is no Machete-like impl in vLLM for W4A16 grouped gemm so the compute gains should be even larger compared to the current Marlin kernels. 

The CUTLASS kernel implementation follows [example 69](https://github.com/NVIDIA/cutlass/blob/main/examples/69_hopper_mixed_dtype_grouped_gemm/69_hopper_int4_fp8_grouped_gemm.cu) which uses a LUT-based method for fast INT4 -> FP8 conversion. Similarly to W4A8 dense, we also add per-channel/per-token epilogue.

We have uploaded a W4A8 quantized variant of [Qwen3-30B-A3B](https://huggingface.co/czhu-cohere/Qwen3-30B-A3B-quantized.w4a8) as an e2e sanity check.

### C++ changes
`csrc/quantization/cutlass_w4a8/w4a8_utils.cu`
- refactor int4 reordering/shuffling utilities to a common file that can be shared between dense and grouped gemm w4a8.

`csrc/quantization/cutlass_w4a8/get_group_starts.cuh`
- compute the pointers for each expert for grouped gemm. Most of the logic is the same as the w8a8 version, but w4a8 needs to account for slightly different input types, packed weights, and additional group-wise scales.

`csrc/quantization/cutlass_w4a8/w4a8_grouped_mm_entry.cu`
- main cutlass kernel implementation and dispatch.
- when encoding/shuffling the weight matrix (`encode_and_reorder_int4b`), we construct the layout object and serialize it to a torch tensor so that we can pass it into the grouped gemm at runtime. This is to avoid having to reconstruct the layout itself at runtime, which would incur significant overhead when the number of experts is large.
  - the `static_assert` and `layout_width` should guarantee that the layout can be serialized to the expected torch tensor dtype/size

`csrc/quantization/w8a8/cutlass/moe/moe_data.cu`
- W4A8 moe can re-use a lot of the helper functions/utilities of W8A8. However, the logic for `get_cutlass_moe_mm_problem_sizes` is coupled with `SwapAB`, so I added an argument to allow the user to explicitly specify SwapAB is true/false (for `RS` GEMM it is always true, since the argument to be dequantized - B - needs to be in the LHS)

### Python changes

`vllm/model_executor/layers/fused_moe/config.py`
`vllm/model_executor/layers/fused_moe/modular_kernel.py`
- I think the previous code assumes that group-wise and channel-wise scales are exclusive, which is not the case here. So I added a field to save the channel scales, in the case where original scales like `w1_scale` are used for the group-wise scales.

`vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors_moe.py`
- define weight loading and pre-processing. Some notable differences to FP8 cutlass moe:
  - uses different quant methods `FusedMoeWeightScaleSupported.GROUP.value` and `FusedMoeWeightScaleSupported.CHANNEL.value` to load group and channel scales respectively
  - stricter requirement on shapes due to limitation of `cutlass:reorder_tensor` explained above (In practice that means for small MoEs like Qwen 30B you may not be able to do TP2)
  - `s_strides1/2` which store strides for the group scales are stored as shape `[num_experts, 2]` and dtype `int64` since that is what the kernel expects
  - `b_strides1/2` is returned by the reordering op and saved to pass in at runtime

`vllm/model_executor/layers/fused_moe/cutlass_moe.py`
- The structure is similar to FP8 CUTLASS moe; main differences are 1) the extra arguments for scales, 2) different strides for each input, 3) enforcing `SwapAB` is true always.

### Other
For both w4a8 dense and grouped, do the fp8 scales conversion + int4b8 -> int4 conversion at runtime so that we can directly run W4A16 checkpoints without needing an offline conversion step. However, note that we still need to change the `config.json` quantization options to tell vllm that it should use the W4A8 path.

### Limitations
Have not implemented/checked compatibility with the different EP options other than default.

## Test Plan
kernel correctness test - `tests/kernels/quantization/test_cutlass_w4a8_moe.py` 
e2e eval - lm_eval gsm8k, compare qwen3-30b-a3b w4a16 and w4a8 variants

## Test Result
`tests/kernels/quantization/test_cutlass_w4a8_moe.py` - pass
lm_eval
```
czhu-cohere/Qwen3-30B-A3B-quantized.w4a8
|  Tasks  |Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|---------|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k_cot|      3|flexible-extract|     8|exact_match|↑  |0.9212|±  |0.0074|
|         |       |strict-match    |     8|exact_match|↑  |0.9007|±  |0.0082|

RedHatAI/Qwen3-30B-A3B-quantized.w4a16
|  Tasks  |Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|---------|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k_cot|      3|flexible-extract|     8|exact_match|↑  |0.9227|±  |0.0074|
|         |       |strict-match    |     8|exact_match|↑  |0.9052|±  |0.0081|
```

8k prefill for `Qwen3-30B-A3B` comparing w4a8 and w4a16
```
w4a8
Median TTFT (ms):                        163.17  
w4a16
Median TTFT (ms):                        246.70    
```
Note that the expert sizes for `Qwen3-30B-A3B` are quite small and it seems hard to approach peak FP8 TFLOPs with these shapes under the current schedule; larger experts can get higher flops. We leave investigation of this to future work.

---
<details>

</details>

### DIFF
--- a/csrc/quantization/cutlass_w4a8/get_group_starts.cuh
+++ b/csrc/quantization/cutlass_w4a8/get_group_starts.cuh
@@ -83,7 +83,6 @@ void run_get_group_gemm_starts(
   // logical k, n
   int64_t n = out_tensors.size(1);
   int64_t k = a_tensors.size(1);
-  printf("inside run_get_group_gemm_starts got n=%d, k=%d\n", n, k);
   int scale_k = cutlass::ceil_div(k, b_group_size);
 
   auto stream = at::cuda::getCurrentCUDAStream(a_tensors.device().index());
--- a/csrc/quantization/cutlass_w4a8/get_group_starts.cuh
+++ b/csrc/quantization/cutlass_w4a8/get_group_starts.cuh
@@ -12,51 +12,58 @@
 
 // TODO: maybe in general we can have better naming for the types?
 // e.g. ElementB is actually the underlying torch type int32 for 8x int4 packing
-// ElementGroupScale is also the cutlass packed type, packing 8 float8 scales to 1
-template <typename ElementA, typename ElementB, typename ElementC, typename ElementAccumulator, typename ElementGroupScale>
+// ElementGroupScale is also the cutlass packed type, packing 8 float8 scales to
+// 1
+template <typename ElementA, typename ElementB, typename ElementC,
+          typename ElementAccumulator, typename ElementGroupScale>
 __global__ void get_group_gemm_starts(
     int64_t* expert_offsets, ElementA** a_offsets, ElementB** b_offsets,
     ElementC** out_offsets, ElementAccumulator** a_scales_offsets,
-    ElementAccumulator** b_scales_offsets, ElementGroupScale** b_group_scales_offsets,
-    ElementA* a_base_as_int, ElementB* b_base_as_int, ElementC* out_base_as_int,
-    ElementAccumulator* a_scales_base_as_int, ElementAccumulator* b_scales_base_as_int,
-    ElementGroupScale* b_group_scales_base_as_int,
-    int64_t n, int64_t k, int scale_k) {
+    ElementAccumulator** b_scales_offsets,
+    ElementGroupScale** b_group_scales_offsets, ElementA* a_base_as_int,
+    ElementB* b_base_as_int, ElementC* out_base_as_int,
+    ElementAccumulator* a_scales_base_as_int,
+    ElementAccumulator* b_scales_base_as_int,
+    ElementGroupScale* b_group_scales_base_as_int, int64_t n, int64_t k,
+    int scale_k) {
   int expert_id = threadIdx.x;
 
   int64_t expert_offset = expert_offsets[expert_id];
-  
+
   // w8a8 code works for these
   a_offsets[expert_id] = a_base_as_int + expert_offset * k;
   out_offsets[expert_id] = out_base_as_int + expert_offset * n;
-  a_scales_offsets[expert_id] = a_scales_base_as_int +  expert_offset;
+  a_scales_offsets[expert_id] = a_scales_base_as_int + expert_offset;
   b_scales_offsets[expert_id] = b_scales_base_as_int + (n * expert_id);
 
   // these need some custom change for w4a8
-  constexpr int pack_factor = 8; // pack 8 int4 into int32
+  constexpr int pack_factor = 8;  // pack 8 int4 into int32
   b_offsets[expert_id] = b_base_as_int + (expert_id * k * n / pack_factor);
-  b_group_scales_offsets[expert_id] = b_group_scales_base_as_int + (expert_id * scale_k * n);
+  b_group_scales_offsets[expert_id] =
+      b_group_scales_base_as_int + (expert_id * scale_k * n);
 }
 
-
-#define __CALL_GET_STARTS_KERNEL(TENSOR_C_TYPE, C_TYPE)                    \
-  else if (out_tensors.dtype() == TENSOR_C_TYPE) {                         \
-    get_group_gemm_starts<cutlass::float_e4m3_t, int32_t, C_TYPE, float, cutlass::Array<cutlass::float_e4m3_t, 8>>            \
-        <<<1, num_experts, 0, stream>>>(                                   \
-            static_cast<int64_t*>(expert_offsets.data_ptr()),              \
-            static_cast<cutlass::float_e4m3_t**>(a_ptrs.data_ptr()),       \
-            static_cast<int32_t**>(b_ptrs.data_ptr()),                     \
-            static_cast<C_TYPE**>(out_ptrs.data_ptr()),                    \
-            static_cast<float**>(a_scales_ptrs.data_ptr()),                \
-            static_cast<float**>(b_scales_ptrs.data_ptr()),                \
-            static_cast<cutlass::Array<cutlass::float_e4m3_t, 8>**>(b_group_scales_ptrs.data_ptr()),          \
-            static_cast<cutlass::float_e4m3_t*>(a_tensors.data_ptr()),     \
-            static_cast<int32_t*>(b_tensors.data_ptr()),                   \
-            static_cast<C_TYPE*>(out_tensors.data_ptr()),                  \
-            static_cast<float*>(a_scales.data_ptr()),                      \
-            static_cast<float*>(b_scales.data_ptr()),                      \
-            static_cast<cutlass::Array<cutlass::float_e4m3_t, 8>*>(b_group_scales.data_ptr()),                \
-            n, k, scale_k);                                                \
+#define __CALL_GET_STARTS_KERNEL(TENSOR_C_TYPE, C_TYPE)                  \
+  else if (out_tensors.dtype() == TENSOR_C_TYPE) {                       \
+    get_group_gemm_starts<cutlass::float_e4m3_t, int32_t, C_TYPE, float, \
+                          cutlass::Array<cutlass::float_e4m3_t, 8>>      \
+        <<<1, num_experts, 0, stream>>>(                                 \
+            static_cast<int64_t*>(expert_offsets.data_ptr()),            \
+            static_cast<cutlass::float_e4m3_t**>(a_ptrs.data_ptr()),     \
+            static_cast<int32_t**>(b_ptrs.data_ptr()),                   \
+            static_cast<C_TYPE**>(out_ptrs.data_ptr()),                  \
+            static_cast<float**>(a_scales_ptrs.data_ptr()),              \
+            static_cast<float**>(b_scales_ptrs.data_ptr()),              \
+            static_cast<cutlass::Array<cutlass::float_e4m3_t, 8>**>(     \
+                b_group_scales_ptrs.data_ptr()),                         \
+            static_cast<cutlass::float_e4m3_t*>(a_tensors.data_ptr()),   \
+            static_cast<int32_t*>(b_tensors.data_ptr()),                 \
+            static_cast<C_TYPE*>(out_tensors.data_ptr()),                \
+            static_cast<float*>(a_scales.data_ptr()),                    \
+            static_cast<float*>(b_scales.data_ptr()),                    \
+            static_cast<cutlass::Array<cutlass::float_e4m3_t, 8>*>(      \
+                b_group_scales.data_ptr()),                              \
+            n, k, scale_k);                                              \
   }
 
 namespace {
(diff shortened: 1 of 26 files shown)


vllm/model_executor/layers/fused_moe/cutlass_moe.py
```diff
@@ -1050,3 +1096,402 @@ def run_cutlass_block_scaled_fused_experts(
...
+    assert w1_scale.dtype == torch.float8_e4m3fn
+    assert w2_scale.dtype == torch.float8_e4m3fn
+    assert w1.dtype == torch.int32
+    assert w2.dtype == torch.int32
+    assert w1_chan_scale.dtype == torch.float32
+    assert w2_chan_scale.dtype == torch.float32
+    assert w1.size(0) == w2.size(0), "Weights expert number mismatch"
+    assert a1q_scale is not None
+    assert a2_scale is None
+    assert out_dtype in [torch.bfloat16], f"Invalid output dtype: {out_dtype}"
```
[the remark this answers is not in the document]
czhu-cohere: the underlying grouped gemm kernel does not support fp16, that could be added in the future

csrc/quantization/cutlass_w4a8/get_group_starts.cuh
```diff
@@ -0,0 +1,104 @@
...
+    torch::Tensor const& a_scales, torch::Tensor const& b_scales,
+    torch::Tensor const& b_group_scales, const int64_t b_group_size) {
+  TORCH_CHECK(a_tensors.dtype() == torch::kFloat8_e4m3fn);
+  TORCH_CHECK(b_tensors.dtype() == torch::kInt32);  // int4 8x packed into int32
+  TORCH_CHECK(a_scales.dtype() == torch::kFloat32);
+  TORCH_CHECK(b_scales.dtype() == torch::kFloat32);
+  TORCH_CHECK(b_group_scales.dtype() ==
+              torch::kFloat8_e4m3fn);  // the underlying torch type is e4m3
+  TORCH_CHECK(out_tensors.dtype() ==
+              torch::kBFloat16);  // only support bf16 for now
```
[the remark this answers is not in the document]
czhu-cohere: the get group kernel supports fp16 but the grouped gemm kernel itself does not, we can remove this check when the underlying grouped gemm supports fp16

dsikka (commented):
The CT integration looks clean to me! Do we have a test model we can add?

csrc/quantization/cutlass_w4a8/w4a8_grouped_mm_entry.cu
```diff
@@ -0,0 +1,485 @@
...
+using namespace cute;
+
+// -------------------------------------------------------------------------------------
+// Static configuration shared across all instantiations
+// -------------------------------------------------------------------------------------
+using ProblemShape =
+    cutlass::gemm::GroupProblemShape<Shape<int, int, int>>;  // <M,N,K> per
+                                                             // group
+using MmaType = cutlass::float_e4m3_t;
+using QuantType = cutlass::int4b_t;
```
LucasWilkinson, line 39: any chance this could be extended to mxfp4 too? would be nice if we could make this compatible with gpt-oss (could be done in a future PR)

csrc/torch_bindings.cpp
```diff
@@ -342,6 +342,29 @@ TORCH_LIBRARY_EXPAND(TORCH_EXTENSION_NAME, ops) {
...
+      "   Tensor a_strides,"
+      "   Tensor b_strides,"
+      "   Tensor c_strides,"
+      "   Tensor group_scale_strides,"
+      "   str? maybe_schedule"
+      ") -> ()");
+  ops.def(
+      "cutlass_encode_and_reorder_int4b_grouped(Tensor b_tensors) -> (Tensor, "
+      "Tensor)");
+  // conditionally compiled so impl registration is in source file
```
LucasWilkinson, line 366: thank you!!!

LucasWilkinson (approved):
LGTM! Thanks; amazing work!

csrc/quantization/cutlass_w4a8/w4a8_grouped_mm_entry.cu
czhu-cohere: mxfp4 is an e8m0 scaling factor for every 32 elements? I think there is a group size limitation of 128 here though because the activation is 8 bits

csrc/quantization/cutlass_w4a8/w4a8_grouped_mm_entry.cu
```diff
@@ -0,0 +1,485 @@
...
+        const torch::Tensor& b_strides, const torch::Tensor& c_strides,
+        const torch::Tensor& group_scale_strides,
+        std::optional<std::string> maybe_schedule) {
+  // user has specified a schedule
+  if (maybe_schedule) {
+    mm_dispatch(out_tensors, a_tensors, b_tensors, a_scales, b_scales,
+                b_group_scales, b_group_size, expert_offsets, problem_sizes,
+                a_strides, b_strides, c_strides, group_scale_strides,
+                *maybe_schedule);
+  }
```
mgoin, line 381: Isn't this case missing an early return?

csrc/quantization/cutlass_w4a8/w4a8_grouped_mm_entry.cu
```diff
@@ -0,0 +1,485 @@
...
+      torch::Tensor& out_tensors, const torch::Tensor& a_tensors,
+      const torch::Tensor& b_tensors, const torch::Tensor& a_scales,
+      const torch::Tensor& b_scales, const torch::Tensor& b_group_scales,
+      const int64_t b_group_size, const torch::Tensor& expert_offsets,
+      const torch::Tensor& problem_sizes_torch, const torch::Tensor& a_strides,
+      const torch::Tensor& b_strides, const torch::Tensor& c_strides,
+      const torch::Tensor& group_scale_strides) {
+    // validation?
+
+    // TODO: cuda stream/guard
```
mgoin: I think this is out of date given the guard and stream right after :)

csrc/quantization/cutlass_w4a8/w4a8_grouped_mm_entry.cu
```diff
@@ -0,0 +1,485 @@
...
+
+  static void grouped_mm(
+      torch::Tensor& out_tensors, const torch::Tensor& a_tensors,
+      const torch::Tensor& b_tensors, const torch::Tensor& a_scales,
+      const torch::Tensor& b_scales, const torch::Tensor& b_group_scales,
+      const int64_t b_group_size, const torch::Tensor& expert_offsets,
+      const torch::Tensor& problem_sizes_torch, const torch::Tensor& a_strides,
+      const torch::Tensor& b_strides, const torch::Tensor& c_strides,
+      const torch::Tensor& group_scale_strides) {
+    // validation?
```
mgoin: nit: either add validation or leave a proper TODO

csrc/quantization/cutlass_w4a8/w4a8_grouped_mm_entry.cu
```diff
@@ -0,0 +1,485 @@
...
+            device_id)};
+
+    arguments = Args{cutlass::gemm::GemmUniversalMode::kGrouped, prob_shape,
+                     mainloop_arguments, epilogue_arguments, hw_info};
+
+    // Allocate workspace
+    size_t workspace_size = GemmShuffled::get_workspace_size(arguments);
+    torch::Tensor workspace =
+        torch::empty(workspace_size,
+                     torch::TensorOptions().dtype(torch::kU8).device(device));
```
mgoin, line 252: Could we allocate the workspace once instead of per-call? Not sure what it might be dependent on

vllm/model_executor/layers/fused_moe/config.py
```diff
@@ -152,6 +152,9 @@ class FusedMoEQuantDesc:
     # Biases for GPT triton MoE
     bias: torch.Tensor | None = None
 
+    # Per-channel scales when per-group scales are also used (W4A8)
+    chan_scale: torch.Tensor | None = None
```
mgoin: I don't understand why we need groupwise and channelwise scales for the weights. This point is rather confusing for me. Can we get good results without doing this? I would prefer to keep the format for w4a8 close to w4a16, at least with respect to the weight format.

vllm/model_executor/layers/fused_moe/cutlass_moe.py
```diff
@@ -1050,3 +1050,402 @@ def run_cutlass_block_scaled_fused_experts(
...
+    if expert_map is not None:
+        assert expert_num_tokens is None
+    assert not use_batched_format, "batched format not supported yet"
+    assert group_size == 128, "Only group size 128 supported"
```
mgoin: Can you add group_size to the msg?

vllm/model_executor/layers/fused_moe/cutlass_moe.py
```diff
@@ -1050,3 +1050,402 @@ def run_cutlass_block_scaled_fused_experts(
...
+
+    assert global_num_experts != -1
+    assert w1.size(2) * 8 == K, "w1 hidden size mismatch"
```
mgoin: Print the sizes please. Users might see this if they do too aggressive TP

vllm/model_executor/layers/fused_moe/cutlass_moe.py
```diff
@@ -1050,3 +1050,402 @@ def run_cutlass_block_scaled_fused_experts(
...
+
+    if expert_map is not None:
+        "Translate info from expert_map to topk_ids"
```
mgoin: Use a normal # comment

mgoin (changes requested):
Nice work! 
I have a fundamental concern about the weight format as I find the need for two sets of scales for the weights confusing, and particularly don't like the idea of needing a specific "chan_scale" attribute. We already have this for NVFP4 and would like to keep it just to that case if possible. So I'd like to discuss the format first before merging to see if we could get it closer to the same as w4a16, much like the recent w4a8 marlin support

csrc/quantization/cutlass_w4a8/w4a8_grouped_mm_entry.cu
czhu-cohere: @LucasWilkinson if I'm reading this correctly https://docs.nvidia.com/cuda/parallel-thread-execution/#asynchronous-warpgroup-level-matrix-shape it looks like the k-extent is always 32 for fp8 wgmma. The [implementation](https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/gemm/collective/sm90_mma_array_tma_gmma_rs_warpspecialized_mixed_input.hpp#L949) checks 
```
static_assert(K_BLOCK_MAX >= 4, "Consider increasing TileShapeK");
```
which means the tileshape K has to be at least 128. Then there is another constraint
```
args.chunk_size % size<2>(TileShape{})) == 0
```
i.e. TileShapeK must divide the group size => group size must be a multiple of 128

csrc/quantization/cutlass_w4a8/w4a8_grouped_mm_entry.cu
czhu-cohere: yes! good catch

csrc/quantization/cutlass_w4a8/w4a8_grouped_mm_entry.cu
czhu-cohere: good point, I see the [fp8 path](https://github.com/vllm-project/vllm/blob/main/csrc/quantization/w8a8/cutlass/moe/grouped_mm_c3x.cuh#L175) currently does this too 
I guess we need to somehow call `get_workspace_size` and pass the result to python, create the workspace tensor and pass it in as another argument

mgoin (approved):
Very nice work! Thanks for iterating

PR #29692, opened 2025-11-28
[Bugfix] Schedule failure due to wrong get_image_size_with_most_features

tomtomjhj:
## Purpose

The scheduler may fail to schedule a batch if `get_image_size_with_most_features` is wrong and `max_num_batched_tokens` is not big enough:
* Some models' `get_image_size_with_most_features` returns an image size that is encoded into a sequence shorter than the actual maximum. That is, `max_tokens_per_mm_item < num_mm_input_tokens` is possible.
* `encoder_compute_budget` is the max of the sequence length of an image with size `get_image_size_with_most_features` and `max_num_batched_tokens` (= `max_num_encoder_input_tokens`). <https://github.com/vllm-project/vllm/blob/0808eb813b6b219324092ab8cc25d3223e5ccb77/vllm/v1/core/encoder_cache_manager.py#L336-L338>
* Therefore, if `max_num_batched_tokens < num_mm_input_tokens`, it is possible that `max_num_encoder_input_tokens < num_mm_input_tokens`, which results in scheduling failure. <https://github.com/vllm-project/vllm/blob/0808eb813b6b219324092ab8cc25d3223e5ccb77/vllm/v1/core/encoder_cache_manager.py#L139-L141>

This PR fixes `get_image_size_with_most_features` of qwen2_vl and gemma3.
* qwen2_vl
    * Problem: An image with with most features is not a square if the number of features is not a squre.
    * Solution: Factorize the max number of features and contruct the image shape that is closest to square to meet the aspect ratio constraint.
* gemma3
    * Problem: The image is not big enough to trigger cropping.
    * Solution: Use the native image size.

## Test Plan
Apply the following patch
```diff
diff --git a/examples/offline_inference/vision_language.py b/examples/offline_inference/vision_language.py
index 8f72bf6f0..755aba99d 100755
--- a/examples/offline_inference/vision_language.py
+++ b/examples/offline_inference/vision_language.py
@@ -321,6 +321,7 @@ def run_gemma3(questions: list[str], modality: str) -> ModelRequestData:
         max_num_seqs=2,
         mm_processor_kwargs={"do_pan_and_scan": True},
         limit_mm_per_prompt={modality: 1},
+        max_num_batched_tokens=512,
     )
 
     prompts = [
@@ -1527,6 +1528,7 @@ def run_qwen2_5_vl(questions: list[str], modality: str) -> ModelRequestData:
             "fps": 1,
         },
         limit_mm_per_prompt={modality: 1},
+        max_num_batched_tokens=512,
     )
 
     if modality == "image":
```

Then run
```
python examples/offline_inference/vision_language.py --model-type qwen2_5_vl
```
and
```
python examples/offline_inference/vision_language.py --model-type gemma3
```

## Test Result
Without this PR's fix, vLLM will hang due to repeated scheduler failure. This PR fixes this problem.

---
<details>

</details>


DarkLight1337 (commented):
Thanks for fixing, can you add some regression tests using the image size that causes this failure?

tomtomjhj:
Added unit tests for `get_image_size_with_most_features`. These tests fail without my fix.

ywang96:
Looks like there are some test failures

```
unsupported operand type(s) for //: 'NoneType' and 'int'
```

vllm/model_executor/models/qwen2_vl.py
```diff
@@ -959,13 +960,29 @@ def get_num_video_tokens(
...
+            for d in range(math.isqrt(n), 0, -1):
+                if n % d == 0:
+                    return d, n // d
+            return 1, n
+
+        height_factor, width_factor = 1, max_seq_len
+        for seq_len in range(max_seq_len, 0, -1):
+            height_factor, width_factor = closest_factor_pair(seq_len)
+            if width_factor / height_factor <= 200:
+                break
```
ywang96, line 996: Thanks for the bugfix on Qwen2_VL but I don't think this is the right fix. IMO we should fix `_get_vision_info` itself instead.

I'm also curious if you can provide a repro example on when the current `_get_vision_info` impl is not accurate? Here we're already giving it a pretty big image size of `9999999, 9999999` so I'm actually a bit surprised that it doesn't give us the max result image size.

ywang96 (changes requested):

vllm/model_executor/models/qwen2_vl.py
tomtomjhj: `_get_vision_info` uses `smart_resize`.
https://github.com/huggingface/transformers/blob/cac0a28c83cf87b7a05495de3177099c635ba852/src/transformers/models/qwen2_vl/image_processing_qwen2_vl.py#L76

`smart_resize` ensures that the number of pixels in the resized image is less than or equal to `max_pixels`. If `max_pixels` is 1280 * 28 * 28, the max number of patches in the resize image is 1280. Since 1280 is not a square number, a square image cannot have the max number of patches regardless of the size. Specifically, the big square image is resized in to 1225 (35 * 35) patches. My patch fixes the issue by factoring 1280 into 32 * 40, and constructing an image size with 32 patches high and 40 patches wide.

ywang96: I think in this case we should be updating `_get_vision_info` instead since it's a "bug" of that function rather than `get_image_size_with_most_features`, correct?

tomtomjhj: I think `_get_vision_info` is doing its job correctly. Qwen2 processes huge square image into 35x35 patches when given pixel limit 1280x28x28, so `_get_vision_info` should follow that behavior.

ywang96: @tomtomjhj Thanks for your explaination and I think I understand what the problem is - here we're really trying to find an image size that would almost saturate the 1280x28x28 max pixels when an image of `999999x999999` cannot because it's square size. For that reason I think this fix makes sense.

I think it would be great if we can document the context here why we're not leveraging `_get_vision_info`!

tomtomjhj: Added!

ywang96 (approved):

PR #29716, opened 2025-11-29
Lazy loading to avoid importing all files

yongming-qin:
## Purpose

Introduce **lazy-loaded model config registry** for `vllm.transformers_utils.configs`, replacing the previous eager imports from many files, which is slow.
This significantly reduces import overhead and avoids loading all config submodules during startup.
Also preserves full backward compatibility with existing `configs.<ConfigClass>` usage.

* Adds lazy loading via `__getattr__` and a central `_CLASS_TO_MODULE` mapping
* Handles special case: `DeepseekV3Config` from `transformers`
* No behavior change to individual configs; only import mechanism changes

## Test Plan

Command examples:

```python
from vllm.transformers_utils import configs
print(configs.AfmoeConfig)
print(configs.DeepseekV3Config)
```

For loading time comparison:
```python
cmd = [
        "python",
        "-c",
        (
            "import time; "
            "t0 = time.perf_counter(); "
            "import vllm.transformers_utils.configs; "
            "t1 = time.perf_counter(); "
            "print(t1 - t0)"
        )
    ]

def measure_once() -> float:
    # Run a Python one-liner in a fresh process
    result = subprocess.run(cmd, capture_output=True, text=True)
    return float(result.stdout.strip())
```
For 10 times, the previous running time is 2.009404 ± 0.008547 s while the improved running time is 1.637713 ± 0.014081 s. 
The test disk is Samsung SSD 980 PRO 2TB.

## Test Result
* Verified dynamic import for each config class listed in `__all__`
* Ran basic vLLM startup to ensure no breakage
* Import time reduced (lazy import ensures submodules are not loaded upfront)
* No differences in model loading behavior observed


Jakub227 (approved):

yongming-qin:
Draft is complete — feedback welcome.

Isotr0py (approved):
LGTM

PR #29723, opened 2025-11-29
[V1][Spec Decode] Optimize Medusa proposer to avoid GPU-CPU sync

dongbo910220:
## Purpose
This PR optimizes the Medusa speculative decoding proposer by keeping draft token proposals on GPU instead of converting them to CPU lists, eliminating unnecessary GPU-CPU synchronization overhead. 

## Performance Impact

  Tested with **Mistral-7B-Instruct-v0.2** + **Medusa heads** on a single GPU:
  - Test configuration: batch_size=32, max_tokens=256, 64 diverse prompts
  - **A/B comparison by git checkout:**
    - Old version (`.tolist()` sync): 1184.2 tok/s
    - Optimized version (GPU tensor): **1195.0 tok/s**
    - **This optimization contributes ~1% throughput improvement** by avoiding GPU-CPU synchronization

  The optimization reduces latency by eliminating:
  - GPU-to-CPU memory transfer for each batch of draft tokens
  - Python list operations and transpose
  - Synchronization points between GPU and CPU

  Draft tokens now remain on GPU throughout the speculative decoding pipeline, which is especially beneficial for larger batch sizes and higher token generation rates.

## Test Result

---
<details>

</details>


dongbo910220:
Hi @DarkLight1337 , could you please help review this PR? Thanks

vllm/v1/spec_decode/medusa.py
```diff
@@ -38,16 +38,24 @@ def propose(
         self,
         target_hidden_states: torch.Tensor,
         sampling_metadata: SamplingMetadata,
-    ) -> list[list[int]]:
+    ) -> torch.Tensor:
```
benchislett, line 41: Are there no consequences to returning a different type from this function? If not, why?

dongbo910220: I've thoroughly audited  usages of MedusaProposer.propose() to ensure the type change is safe.
  - Single call site: only gpu_model_runner.py::propose_draft_token_ids() invokes it.
  - Downstream paths:
    1) GPU path in _prepare_input_ids : it asserts _draft_token_ids is a torch.Tensor and scatters it on GPU. With the old list return this code path couldn’t be taken; returning a tensor makes Medusa compatible with that optimization.
    2) Return path take_draft_token_ids : explicitly handles both tensor and list, converting tensor→list when needed. So external consumers still receive Python lists.

  Type pattern stays consistent: GPU-based proposers (Medusa/Eagle) return torch.Tensor; CPU-based proposers (ngram/suffix) return list[list[int]]; the union type reflects that.

dongbo910220:
Hi @benchislett, just a gentle ping on this. I believe I've addressed the concern regarding the return type safety above (it's consistent with other GPU-based proposers). Let me know if you have any other questions or if this looks good to go!

vllm/v1/spec_decode/medusa.py
```diff
@@ -38,16 +38,24 @@ def propose(
...
-        # Get draft tokens and transpose the result
-        # TODO(woosuk): OPTIMIZATION: Return GPU tensor without GPU-CPU
-        # synchronization.
-        draft_tokens = [logit.argmax(dim=-1).tolist() for logit in logits]
-        return [list(row) for row in zip(*draft_tokens)]
+        # Compute argmax for each Medusa head and stack into a single tensor
+        # Shape: [batch_size, num_heads]
+        draft_tokens = torch.stack([logit.argmax(dim=-1) for logit in logits], dim=1)
+
+        # Always validate shape (avoid assert being stripped by -O)
```
benchislett: Why is this check necessary? Is there a failure case here?

benchislett (approved):

vllm/v1/spec_decode/medusa.py
dongbo910220: Good catch! I've removed the shape validation. `torch.stack()` already validates dimensions and will raise a clear error if there's a mismatch, so the explicit check was indeed redundant.

PR #29748, opened 2025-11-30
[MoE-FP8-modelopt] Add FlashInfer alignment padding for intermediate dimensions

danielafrimi:
This PR adds support for FlashInfer MoE FP8 kernels that require the intermediate (gated) dimension to be aligned.
Some models break when the intermediate size isn’t divisible by 16, especially under different TP.

This PR introduces _maybe_pad_intermediate_for_flashinfer, which pads w13 and w2 along the intermediate dim so the weights meet FlashInfer’s alignment constraints.

If padding is needed, we zero-pad the up/gate and down projection weights and update intermediate_size_per_partition accordingly. This keeps the model numerically correct while avoiding kernel launch failures.


tlrmchlsmth (approved):
Looks good to me, thanks for the contribution!

PR #29755, opened 2025-11-30
Fix KV cache sync issue during CUDA graph replay

### LINKED ISSUE #29608: [Bug]: CUDA Graph Replay Skips KV Transfer Synchronization in Full Cache Hit Scenarios
### Your current environment

<details>

```text
```

</details>

### 🐛 Describe the bug

When using asynchronous KV transfer connectors with full cache hits, CUDA graph replay can skip the `wait_for_layer_load()` synchronization, leading to potential data races where attention operations use incompletely loaded KV caches.

- vLLM version: [v1]
- KV Connector: Asynchronous KV transfer （`wait_for_laeyr_load()` support)
- CUDA Graph mode: Enabled (FULL, FULL_AND_PIECEWISE, or FULL_DECODE_ONLY)

When a request has a full KV cache hit, connectors typically reserve 1 token for the scheduler (as 0-token scheduling is not allowed). 

In `gpu_model_runner.py:_determine_batch_execution_and_padding()`:
```
uniform_decode = (
    (max_num_scheduled_tokens == self.uniform_decode_query_len)  # 1 == 1 ✓
    and (num_tokens_padded == max_num_scheduled_tokens * num_reqs)  # 1 == 1×1 ✓
)
```

When CUDA graph replays (in `compilation/cuda_graph.py:207-208`):
```
entry.cudagraph.replay()
return entry.outputThe replay **directly executes the captured GPU operations** without calling `self.runnable(*args, **kwargs)`, which means:
- `unified_attention()` functions are not executed
- `@maybe_transfer_kv_layer` decorator is bypassed
- **`wait_for_layer_load()` is skipped**
```
For **asynchronous connectors** that rely on per-layer synchronization, this causes a data race where the attention kernel may read KV cache data that hasn't finished loading.

yashwantbezawada:
Fixes #29608

I ran into this while looking at the async KV transfer code. When using connectors like LMCache with CUDA graphs, there's a race condition during graph replay.

The issue is that `wait_for_layer_load()` calls happen inside the `@maybe_transfer_kv_layer` decorator on attention functions. During graph capture, these calls execute normally. But during replay, only the GPU operations are replayed - the Python decorator code gets skipped entirely. So the async KV loads might not be complete when the attention kernels run.

The fix adds a `wait_for_load()` method to the connector interface that waits for all pending loads to complete. I put the sync call right before `replay()` in CUDAGraphWrapper so it only runs during replay, not capture. During capture we still want the per-layer sync to work normally so saves happen correctly.

Changes:
- Added `wait_for_load()` to KVConnectorBase_V1 with a default no-op (connectors that don't do async loading are unaffected)
- Implemented it in LMCacheConnectorV1Impl to drain any remaining layers from the retrievers
- Added it to MultiConnector to delegate to sub-connectors
- Added the sync call in CUDAGraphWrapper before replay()

I couldn't test this end-to-end without an LMCache setup, but traced through the code paths carefully. The sync only happens when both KV transfer and CUDA graphs are enabled, and only during replay.

### DIFF
--- a/vllm/v1/worker/kv_connector_model_runner_mixin.py
+++ b/vllm/v1/worker/kv_connector_model_runner_mixin.py
@@ -14,7 +14,7 @@ from typing import (
 import torch
 
 from vllm.attention.backends.abstract import AttentionBackend
-from vllm.config import VllmConfig
+from vllm.config import CUDAGraphMode, VllmConfig
 from vllm.config.cache import CacheDType
 from vllm.distributed.kv_transfer import (
     ensure_kv_transfer_shutdown,
@@ -126,7 +126,15 @@ class KVConnectorModelRunnerMixin:
         # These transfers are designed to be async and the requests
         # involved may be disjoint from the running requests.
         # Do this here to save a collective_rpc.
-        kv_connector.start_load_kv(get_forward_context())
+        forward_ctx = get_forward_context()
+        kv_connector.start_load_kv(forward_ctx)
+
+        # Ensure all async KV cache loads are complete before CUDA graph
+        # capture/replay. During replay, the per-layer wait_for_layer_load()
+        # calls in the attention layer decorator are bypassed.
+        if forward_ctx.cudagraph_runtime_mode != CUDAGraphMode.NONE:
+            kv_connector.wait_for_load()
+
         try:
             yield output
         finally:
--- a/vllm/v1/worker/kv_connector_model_runner_mixin.py
+++ b/vllm/v1/worker/kv_connector_model_runner_mixin.py
@@ -14,7 +14,7 @@ from typing import (
 import torch
 
 from vllm.attention.backends.abstract import AttentionBackend
-from vllm.config import CUDAGraphMode, VllmConfig
+from vllm.config import VllmConfig
 from vllm.config.cache import CacheDType
 from vllm.distributed.kv_transfer import (
     ensure_kv_transfer_shutdown,
@@ -126,14 +126,7 @@ class KVConnectorModelRunnerMixin:
         # These transfers are designed to be async and the requests
         # involved may be disjoint from the running requests.
         # Do this here to save a collective_rpc.
-        forward_ctx = get_forward_context()
-        kv_connector.start_load_kv(forward_ctx)
-
-        # Ensure all async KV cache loads are complete before CUDA graph
-        # capture/replay. During replay, the per-layer wait_for_layer_load()
-        # calls in the attention layer decorator are bypassed.
-        if forward_ctx.cudagraph_runtime_mode.has_full_cudagraphs():
-            kv_connector.wait_for_load()
+        kv_connector.start_load_kv(get_forward_context())
 
         try:
             yield output
--- a/vllm/v1/worker/kv_connector_model_runner_mixin.py
+++ b/vllm/v1/worker/kv_connector_model_runner_mixin.py
@@ -132,7 +132,7 @@ class KVConnectorModelRunnerMixin:
         # Ensure all async KV cache loads are complete before CUDA graph
         # capture/replay. During replay, the per-layer wait_for_layer_load()
         # calls in the attention layer decorator are bypassed.
-        if forward_ctx.cudagraph_runtime_mode == CUDAGraphMode.FULL:
+        if forward_ctx.cudagraph_runtime_mode.has_full_cudagraphs():
             kv_connector.wait_for_load()
 
         try:
--- a/vllm/v1/worker/kv_connector_model_runner_mixin.py
+++ b/vllm/v1/worker/kv_connector_model_runner_mixin.py
@@ -132,7 +132,7 @@ class KVConnectorModelRunnerMixin:
         # Ensure all async KV cache loads are complete before CUDA graph
         # capture/replay. During replay, the per-layer wait_for_layer_load()
         # calls in the attention layer decorator are bypassed.
-        if forward_ctx.cudagraph_runtime_mode != CUDAGraphMode.NONE:
+        if forward_ctx.cudagraph_runtime_mode == CUDAGraphMode.FULL:
             kv_connector.wait_for_load()
 
         try:
--- a/vllm/distributed/kv_transfer/kv_connector/v1/base.py
+++ b/vllm/distributed/kv_transfer/kv_connector/v1/base.py
@@ -27,6 +27,8 @@ The class provides the following primitives:
     the Connector based on the metadata.
         start_load_kv() - starts loading all KVs (maybe async)
         wait_for_layer_load() - blocks until layer i load is done
+        wait_for_load() - blocks until all layer loads are done
+            (used before CUDA graph replay to ensure sync)
 
         save_kv_layer() - starts saving KV for layer i (maybe async)
         wait_for_save() - blocks until all saves are done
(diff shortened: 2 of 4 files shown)


vllm/compilation/cuda_graph.py
```diff
@@ -204,5 +208,11 @@ def __call__(self, *args, **kwargs):
                 f"got {new_input_addresses}"
             )
 
+        # Ensure all async KV cache loads are complete before replaying.
```
ProExpertProg: This should live in the GPU model runner; the cudagraph wrapper shouldn't need to worry about kv cache transfer

ProExpertProg (changes requested):
also cc @KuntaiDu @NickLucche

vllm/compilation/cuda_graph.py
yashwantbezawada: Done. Moved the wait_for_load() call to KVConnectorModelRunnerMixin._get_kv_connector_output(). The CUDAGraphWrapper no longer has any KV transfer dependencies

NickLucche (commented):
I think this may actually be related to the changes introduced in https://github.com/vllm-project/vllm/pull/28253 which intentionally skips the tracing.
There's a second PR to address that here https://github.com/vllm-project/vllm/pull/27026 by @sdavidbd .

Does that match your observations @yashwantbezawada ?

yashwantbezawada:
Thanks for the context!

Looked through #28253 and #27026 - those are fixing the capture/dummy run path (getting metadata set properly so KV connector works during capture).

This one's targeting a different scenario: during replay, `start_load_kv()` runs (it's in the model runner, outside the graph), but the decorator's `wait_for_layer_load()` doesn't execute since only captured GPU ops run. So no sync between the async loads and replayed attention.

Fix adds sync in the model runner after `start_load_kv()`, before the forward.

Think they should be complementary - #27026 for capture, this for replay. But happy to chat more or coordinate with @sdavidbd if there's overlap!

vllm/v1/worker/kv_connector_model_runner_mixin.py
```diff
@@ -126,7 +126,15 @@ def _get_kv_connector_output(
...
         # involved may be disjoint from the running requests.
         # Do this here to save a collective_rpc.
-        kv_connector.start_load_kv(get_forward_context())
+        forward_ctx = get_forward_context()
+        kv_connector.start_load_kv(forward_ctx)
+
+        # Ensure all async KV cache loads are complete before CUDA graph
+        # capture/replay. During replay, the per-layer wait_for_layer_load()
+        # calls in the attention layer decorator are bypassed.
+        if forward_ctx.cudagraph_runtime_mode != CUDAGraphMode.NONE:
```
ProExpertProg: I think piecewise is fine, so I think this is just needed for full cg.
```suggestion
        if forward_ctx.cudagraph_runtime_mode == CUDAGraphMode.FULL:
```

ProExpertProg (approved):
Approving to remove the "requested changes" mark while I'm gone for the next 2 weeks, please wait for @NickLucche's approval to merge!

vllm/v1/worker/kv_connector_model_runner_mixin.py
```diff
@@ -126,7 +126,15 @@ def _get_kv_connector_output(
...
         # involved may be disjoint from the running requests.
         # Do this here to save a collective_rpc.
-        kv_connector.start_load_kv(get_forward_context())
+        forward_ctx = get_forward_context()
+        kv_connector.start_load_kv(forward_ctx)
+
+        # Ensure all async KV cache loads are complete before CUDA graph
+        # capture/replay. During replay, the per-layer wait_for_layer_load()
+        # calls in the attention layer decorator are bypassed.
+        if forward_ctx.cudagraph_runtime_mode == CUDAGraphMode.FULL:
```
NickLucche:
```suggestion
        if forward_ctx.cudagraph_runtime_mode.has_full_cudagraphs():
```

NickLucche (commented):
Hey @yashwantbezawada sorry for the delay in getting back to this. 
I think there's some confusion around `start_load_kv` as you also noted in your comment, perhaps @ApostaC can chime in to clarify the original intent.
`start_load_kv` is supposed to start an (*async*) transfer of all layers. 
So far there hasn't been the need to explicitly sync on this transfer, otherwise we would lose the benefit of real async (wrt engine execution) transfer.
Now, if a sync was really needed for any reason at all, one could overload `wait_for_save`. If this was the case, I would agree resorting to your method would offer better semantics, but this hasn't been the case afaik.

There's also a second set of operations that can be used to move kv cache layer-by-laeyr and that naturally go together <save_kv_layer, wait_for_layer_load>.
If tracing of these ops is skipped, then they will be indeed skipped from replay. But I think this is a separate issue that we should address by either a) tracing the save/load b) only allow their usage in piecewise.

tl;dr: I think we're fixing a separate part of the interface to make up for a potential logic bug in how we're recording ops for kv cache save/load.

vllm/v1/worker/kv_connector_model_runner_mixin.py
```diff
@@ -126,7 +126,15 @@ def _get_kv_connector_output(
...
+            kv_connector.wait_for_load()
```
sdavidbd: I have a concern regarding the semantics implied by the new `wait_for_load()` API.

As currently described, it appears that wait_for_load() is intended to synchronize on all layer-level load operations initiated by `start_load_kv()`. However, I’m not sure this behavior is universally desirable across connector implementations. As @NickLucche noted, this seems to go against one of the core purposes of `start_load_kv()`: in synchronous-loading scenarios (i.e., when loading KV for the running requests in the current batch), `start_load_kv()` is meant to kick off layer-by-layer async loads in parallel with attention computation.

If `wait_for_load()` becomes a global synchronization point for everything triggered by `start_load_kv()`, we effectively eliminate that concurrency for connectors that rely on it.

In addition, since the `forward_ctx` is already passed into `start_load_kv()`, any connector that _does_ require full synchronization before CUDA graph replay could implement that behavior internally by making `start_load_kv()` block until its own load operations complete. That seems more flexible than introducing a global API that implies uniform synchronization semantics.

yashwantbezawada: Thanks for the detailed feedback.

You're right - `start_load_kv()` is designed for layer-by-layer async loading in parallel with attention, and adding `wait_for_load()` as a global sync would eliminate that concurrency benefit.

Your suggestion makes sense: if a connector needs full sync for CUDA graphs, it could handle that internally in `start_load_kv()` by checking `forward_ctx`. That keeps the decision with each connector rather than imposing uniform semantics via a new API.

Happy to close this PR or rework it based on your guidance. What would you recommend?

ProExpertProg: ++, @yashwantbezawada this seems easy enough to do in this PR

yashwantbezawada: Done! Moved the sync logic into LMCache's `start_load_kv()` - it now drains all layer retrievers at the end when CUDA graphs are enabled. Removed the `wait_for_load()` API entirely.

yashwantbezawada: Done! Moved the sync logic into LMCache's `start_load_kv()` - it now drains all layer retrievers at the end when CUDA graphs are enabled. Removed the `wait_for_load()` API entirely.

PR #29757, opened 2025-11-30
Add Mistral Large 3 and Ministral 3

juliendenize:
## Purpose

This PR adds support to Mistral-Large-3 and Ministral-3.

## Test Result

---
<details>

</details>


vllm/model_executor/layers/mla.py
```diff
@@ -159,6 +160,9 @@ def forward_native(
                 hidden_states, q_c, positions, self.indexer_rope_emb
             )
 
+        if llama_4_scaling is not None:
+            q *= llama_4_scaling
```
mgoin, line 164: Why not just put this in a rotary embedding layer?

mgoin (commented):
🥳 Make sure to update the supported model page https://github.com/vllm-project/vllm/blob/main/docs/models/supported_models.md and the testing registry https://github.com/vllm-project/vllm/blob/main/tests/models/registry.py with is_available_online=False for now

vllm/model_executor/models/mistral_large_3_eagle.py
```diff
@@ -0,0 +1,163 @@
...
+    def get_input_embeddings(
+        self,
+        input_ids: torch.Tensor,
+        multimodal_embeddings: MultiModalEmbeddings | None = None,
+        *,
+        is_multimodal: torch.Tensor | None = None,
+        handle_oov_mm_token: bool = False,
+    ) -> torch.Tensor:
+        from .utils import _merge_multimodal_embeddings
+
```
DarkLight1337: No need to lazy import

vllm/transformers_utils/configs/eagle.py
```diff
@@ -41,7 +41,9 @@ def __init__(
         # LlamaForCausalLM -> EagleLlamaForCausalLM
         # LlamaForCausalLM -> Eagle3LlamaForCausalLM
         # LlamaForCausalLMEagle3 -> LlamaForCausalLMEagle3
-        if method == "eagle":
+        if method is None:
```
DarkLight1337: Could you elaborate on why this change is needed?

vllm/model_executor/models/deepseek_v2.py
```diff
@@ -395,6 +395,16 @@ def yarn_get_mscale(scale: float = 1, mscale: float = 1) -> float:
     return 0.1 * mscale * math.log(scale) + 1.0
 
 
+def _get_llama_4_scaling(
```
benchislett, line 398: Do we have a plan to move these helper functions into a utility file?

vllm/transformers_utils/configs/eagle.py
benchislett: Also curious why `"method": None` is being observed here.

vllm/model_executor/layers/mla.py
juliendenize: Just a choice but no strong opinion about this that i can think of now. We also did in llama.py would it be necessary to refactor now or could it be in later pr ?

vllm/model_executor/models/deepseek_v2.py
juliendenize: I think this could done yes, we also have it in llama.py question would be do we do it now ? It is not exactly the same as the one in llama4.py (a different offset).

vllm/transformers_utils/configs/eagle.py
mickaelseznec: I run into a bug when disabling `--enforce-eager`.

vLLM tries to compute a hash of the config, and this `PretrainedConfig` is called from `transformers` with no arguments. From what I understand, it was trying to check the diff between the actual EAGLEConfig and an uninitialized one, which triggers the assert.

I believe the root cause is https://github.com/vllm-project/vllm/pull/26468, changing
- `hf_config_json = self.hf_config.to_json_string(use_diff=False)` https://github.com/vllm-project/vllm/pull/26468/files#diff-998c640befaf137b9af825f29f4e6e47d273caab1fd04093c97df24b18f5c417L345
- to `x.to_json_string()` (removing the `use_diff=False`, default is [True](https://github.com/huggingface/transformers/blob/main/src/transformers/configuration_utils.py#L920)) https://github.com/vllm-project/vllm/pull/26468/files#diff-f0679c95660e953a0b43d241ba6332d28cefd86ea565fc10c0d0cc57dc158cfcR257

Let me know if you need more investigation on my side

vllm/entrypoints/openai/tool_parsers/mistral_tool_parser.py
```diff
@@ -80,7 +80,7 @@ def __init__(self, tokenizer: TokenizerLike):
         self.tool_call_regex = re.compile(r"\[{.*}\]", re.DOTALL)
         if _is_fn_name_regex_support(self.model_tokenizer):
             self.fn_name_regex = re.compile(
-                r"([a-zA-Z0-9_-]+)(\{[\s\S]*?\})(?=\s*$|,|\s)", re.DOTALL
+                r"([a-zA-Z0-9_-]+)(\{[\s\S]*?\})(?=\s*$|,|\s)?", re.DOTALL
```
juliendenize, line 83: Had to add this to make sure tool calls worked as expected for some examples, will rerun some xp to see if i didn't break anything there as we didn't author this.

tests/models/registry.py
```diff
@@ -770,7 +770,13 @@ def check_available_online(
...
     "PixtralForConditionalGeneration": _HfExamplesInfo(
         "mistralai/Pixtral-12B-2409",
-        tokenizer_mode="mistral",
+        extras={
+            "mistral-large-3": "mistralai/Mistral-Large-3-675B-Instruct-2512-NVFP4",
+            "ministral-3": "mistralai/Ministral-3-3B-Instruct-2512",
+        },
+        tokenizer_mode="mistralai/Mistral-Large-3-675B-Instruct-2512-NVFP4",
+        # TODO: revert once Mistral-Large-3 and Ministral-3 are publicly available.
+        is_available_online=False,
```
juliendenize, line 782: lmk @mgoin is this is what you had in mind here.

vllm/transformers_utils/configs/eagle.py
DarkLight1337: cc @zou3519 @youkaichao @hmellor

tests/models/registry.py
```diff
@@ -770,7 +770,13 @@ def check_available_online(
     ),
     "PixtralForConditionalGeneration": _HfExamplesInfo(
         "mistralai/Pixtral-12B-2409",
-        tokenizer_mode="mistral",
+        extras={
+            "mistral-large-3": "mistralai/Mistral-Large-3-675B-Instruct-2512-NVFP4",
+            "ministral-3": "mistralai/Ministral-3-3B-Instruct-2512",
+        },
+        tokenizer_mode="mistralai/Mistral-Large-3-675B-Instruct-2512-NVFP4",
```
juliendenize:
```suggestion
        extras={
            "mistral-large-3": "mistralai/Mistral-Large-3-675B-Instruct-2512-NVFP4",
            "ministral-3": "mistralai/Ministral-3-3B-Instruct-2512",
        },
        tokenizer_mode="mistral",
```

vllm/tokenizers/mistral.py
```diff
@@ -139,6 +143,31 @@ def _prepare_apply_chat_template_tools_and_messages(
...
+                        if function_key not in function_fields:
+                            tool["function"].pop(function_key)
+                            logger.warning_once(
+                                f"'{function_key}' is not supported by mistral-common "
+                                "for function tools. It has been poped from the "
+                                "function definition."
+                            )
+                else:
+                    raise ValueError("mistral-common only supports function tools.")
+
```
juliendenize, line 170: as we got some requests we also give a bit more flexibility for tool definition. though extra args are not supported by mistral-common. this avoid error thrown for users that use multiple models with the same settings and would like to use our models without mistral-common throwing a tantrum.

vllm/transformers_utils/configs/eagle.py
zou3519: I think we can put the `use_diff=False` back, do you want to do this in this PR or should I do it in a separate one?

tests/models/registry.py
```diff
@@ -770,7 +770,13 @@ def check_available_online(
     ),
     "PixtralForConditionalGeneration": _HfExamplesInfo(
         "mistralai/Pixtral-12B-2409",
+        extras={
+            "mistral-large-3": "mistralai/Mistral-Large-3-675B-Instruct-2512-NVFP4",
+            "ministral-3": "mistralai/Ministral-3-3B-Instruct-2512",
+        },
         tokenizer_mode="mistral",
+        # TODO: revert once Mistral-Large-3 and Ministral-3 are publicly available.
+        is_available_online=False,
```
tlrmchlsmth, line 782: We should flip this before 0.12 goes out @khluu

vllm/tokenizers/mistral.py
```diff
@@ -97,6 +97,8 @@ def _prepare_apply_chat_template_tools_and_messages(
     continue_final_message: bool = False,
     add_generation_prompt: bool = False,
 ) -> tuple[list["ChatCompletionMessageParam"], list[dict[str, Any]] | None]:
+    from mistral_common.protocol.instruct.tool_calls import Function, Tool
```
patrickvonplaten, line 100: why dynamic import here?

vllm/transformers_utils/configs/eagle.py
patrickvonplaten: Should we just revert this to make sure the PR gets in?

vllm/tokenizers/mistral.py
juliendenize: it's the same we do for all functions in this file we import at use because some people don't want to install mistral-common.

vllm/transformers_utils/configs/eagle.py
zou3519: Might be too late to revert all of https://github.com/vllm-project/vllm/pull/26468, but I can put the use_diff=False back

tests/models/registry.py
ywang96: I think it's okay to leave this as is since we're cutting a branch today

vllm/model_executor/models/deepseek_v2.py
ywang96: Let's address this in a follow-up PR

vllm/entrypoints/openai/tool_parsers/mistral_tool_parser.py
juliendenize: Looks good on the examples I launched so this looks benign

vllm/transformers_utils/configs/eagle.py
ywang96: @zou3519 can you open a separate PR for this? then we can update this PR accordingly

patrickvonplaten: We reverted the change here, should be good

vllm/config/utils.py
```diff
@@ -254,7 +254,9 @@ def normalize_value(x):
 
     # PretrainedConfig
     if hasattr(x, "to_json_string") and callable(x.to_json_string):
-        return x.to_json_string()
+        # using `use_diff=False` to avoid initializing object with
+        # default arguments only
+        return x.to_json_string(use_diff=False)
```
patrickvonplaten: use_diff to make sure we don't have trouble with speculative decoding

vllm/transformers_utils/configs/eagle.py
patrickvonplaten: (use_diff=False) was added here: https://github.com/vllm-project/vllm/pull/29757/files#r2578969697

ywang96 (approved):
Amazing! Thanks for all the work and look forward to the release!

Approving as the rest of the comments can be addressed in follow-up PRs.

PR #29761, opened 2025-12-01
[BugFix] respect VLLM_LOGGING_LEVEL in logger

BoyuanFeng:
Issue: vLLM defaults logger vllm to be DEBUG (Line 65). `envs.VLLM_LOGGING_LEVEL` is used to update `DEFAULT_LOGGING_CONFIG["handlers"]["vllm"]` (aka `logging.StreamHandler`), but not `DEFAULT_LOGGING_CONFIG["loggers"]["vllm"]` (aka `logging.vllm`). `logging.vllm` is DEBUG and used as the parent of all other loggers, including execute_model.

https://github.com/vllm-project/vllm/blob/1336a1ea244fa8bfd7e72751cabbdb5b68a0c11a/vllm/logger.py#L40-L71

This PR fixed the issue.

### DIFF
--- a/vllm/v1/worker/gpu_model_runner.py
+++ b/vllm/v1/worker/gpu_model_runner.py
@@ -2928,15 +2928,6 @@ class GPUModelRunner(
                     use_cascade_attn=cascade_attn_prefix_lens is not None,
                 )
 
-                logger.debug(
-                    "Running batch with cudagraph_mode: %s, batch_descriptor: %s, "
-                    "ubatch_slices: %s, num_tokens_across_dp: %s",
-                    cudagraph_mode,
-                    batch_desc,
-                    ubatch_slices,
-                    num_tokens_across_dp,
-                )
-
                 num_tokens_padded = batch_desc.num_tokens
                 num_reqs_padded = (
                     batch_desc.num_reqs if batch_desc.num_reqs is not None else num_reqs
--- a/vllm/v1/worker/gpu_model_runner.py
+++ b/vllm/v1/worker/gpu_model_runner.py
@@ -2928,15 +2928,14 @@ class GPUModelRunner(
                     use_cascade_attn=cascade_attn_prefix_lens is not None,
                 )
 
-                if self.observability_config.log_runtime_cudagraph_metadata:
-                    logger.debug(
-                        "Running batch with cudagraph_mode: %s, batch_descriptor: %s, "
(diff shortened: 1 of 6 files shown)


vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2928,15 +2928,6 @@ def execute_model(
                     use_cascade_attn=cascade_attn_prefix_lens is not None,
                 )
 
-                logger.debug(
```
ProExpertProg, line 2931: What if we save a bool property on the object that controls this log and use an if statement on the hot path, perhaps that's better?

BoyuanFeng: yes added an observability_config to control that.

ProExpertProg: I was actually referring to using `if logger.isEnabledFor(logging.DEBUG):`, but I see that you're having issues with that

BoyuanFeng:
@njhill @ProExpertProg I found that vLLM defaults logger `vllm` to be `DEBUG` (Line 65). `envs.VLLM_LOGGING_LEVEL` is used to update `DEFAULT_LOGGING_CONFIG["handlers"]["vllm"]` (aka logging.StreamHandler), but not `DEFAULT_LOGGING_CONFIG["loggers"]["vllm"]` (aka logging.vllm). logging.vllm is `DEBUG` and used as the parent of all other loggers, including `execute_model`.

https://github.com/vllm-project/vllm/blob/1336a1ea244fa8bfd7e72751cabbdb5b68a0c11a/vllm/logger.py#L40-L71

vllm/v1/worker/gpu_model_runner.py
BoyuanFeng: `logger.isEnabledFor` still adds ~1 us latency so I kept the observability config
image

njhill (changes requested):
Please revert all the changes apart from those in logger.py. Then this looks good! (or open a separate PR for that as @ProExpertProg suggested)

njhill (approved):

njhill:
We can probably get around this quickly by adding `VLLM_LOGGING_LEVEL=DEBUG` to the tests in question in `test-pipeline.yaml` (looks like there's only two so far, can wait for the current CI run to finish first to make sure there aren't any more).

BoyuanFeng:
@njhill yeah fixed the tests. let's see if something is missed.

PR #29769, opened 2025-12-01
Refactor MLA attention: move prefill logic to layer.py

### LINKED ISSUE #26516: [Refactor][MLA]: Lift prefill/decode split into compiled region

therealnaveenkamal:
## Purpose
This PR refactors the MLA (Multi-head Latent Attention) implementation by moving the core prefill logic from vllm/v1/attention/backends/mla/common.py to vllm/attention/layer.py. This improves code organization and creates a cleaner separation between the attention layer and backend implementations.

This PR is part 1/3 of the work to "lift the MLA prefill/decode split into the compiled region" tracked in https://github.com/vllm-project/vllm/issues/26516 

---
<details>

</details>


vllm/v1/attention/backends/mla/common.py
```diff
@@ -1141,125 +1135,32 @@ def __init__(
...
-            f"{self.num_heads=}, "
-            f"{self.qk_nope_head_dim=}, "
-            f"{self.v_head_dim=}"
-        )
-        kv_b_proj_weight = kv_b_proj_weight.view(
-            self.kv_lora_rank,
-            self.num_heads,
-            self.qk_nope_head_dim + self.v_head_dim,
-        )
+        """Delegates to the layer's process_weights_after_loading method.
```
ProExpertProg: I think we can just remove this method as it will always get called on the layer first anyway!

vllm/v1/attention/backends/mla/common.py
```diff
@@ -1900,170 +1463,27 @@ def forward(
...
-        output = output[:num_actual_toks, ...]
-        q = q[:num_actual_toks, ...]
-        k_c_normed = k_c_normed[:num_actual_toks, ...]
-        k_pe = k_pe[:num_actual_toks, ...]
-
-        assert (
-            attn_metadata.num_decodes is not None
-            and attn_metadata.num_prefills is not None
-            and attn_metadata.num_decode_tokens is not None
+        """Forward method that delegates to the layer's forward_impl."""
```
ProExpertProg: Can we just remove this as well?

vllm/v1/attention/backends/mla/common.py
```diff
@@ -1141,125 +1135,32 @@ def __init__(
...
-            self.kv_lora_rank,
-            self.num_heads,
-            self.qk_nope_head_dim + self.v_head_dim,
-        )
+        """Delegates to the layer's process_weights_after_loading method.
 
-        W_UK, W_UV = kv_b_proj_weight.split(
-            [self.qk_nope_head_dim, self.v_head_dim], dim=-1
+        This method is kept for backward compatibility with tests that call
+        impl.process_weights_after_loading() directly. The actual implementation
```
ProExpertProg: Can we just change those tests to call the layer directly?

vllm/v1/attention/backends/mla/common.py
```diff
@@ -1141,125 +1135,32 @@ def __init__(
         self.is_aiter_triton_fp8_bmm_enabled = rocm_aiter_ops.is_fp8bmm_enabled()
 
     def process_weights_after_loading(self, act_dtype: torch.dtype):
```
ProExpertProg: Is there a difference between the sparse and dense weight processing?

ProExpertProg (commented):
Does this handle sparse MLA as well?

vllm/v1/attention/backends/mla/common.py
therealnaveenkamal: done

vllm/v1/attention/backends/mla/common.py
therealnaveenkamal: I don't think so, but I've moved.

vllm/v1/attention/backends/mla/common.py
therealnaveenkamal: done

vllm/attention/layer.py
```diff
@@ -647,6 +662,21 @@ def __init__(
...
+        # Use getattr since these exist on MLACommonImpl but not on abstract base
+        self.is_aiter_triton_fp8_bmm_enabled: bool = getattr(
+            self.impl, "is_aiter_triton_fp8_bmm_enabled", False
+        )
+        self.dcp_world_size: int | None = getattr(self.impl, "dcp_world_size", None)
+        self.chunked_prefill_workspace_size: int = getattr(
+            self.impl, "chunked_prefill_workspace_size", 0
+        )
+        self._pad_v: bool = getattr(self.impl, "_pad_v", False)
+        self._use_fi_prefill: bool = getattr(self.impl, "_use_fi_prefill", False)
```
ProExpertProg: Can we just copy over whatever logic is used inside common.py to initialize these?

vllm/v1/attention/backends/mla/common.py
```diff
@@ -1570,370 +1443,30 @@ def _run_prefill_context_chunk_trtllm_ragged(
...
-            self.num_heads,
-            self.qk_nope_head_dim + self.v_head_dim,
-        )
-
-        W_UK, W_UV = kv_b_proj_weight.split(
-            [self.qk_nope_head_dim, self.v_head_dim], dim=-1
-        )
+    def _v_up_proj(self, x: torch.Tensor, out: torch.Tensor):
+        # Convert from (B, N, L) to (N, B, L)
+        x = x.view(-1, self.num_heads, self.kv_lora_rank).transpose(0, 1)
```
ProExpertProg, line 1475: I thought this was moved to layer.py, if yes remove from common?

vllm/attention/backends/abstract.py
```diff
@@ -398,6 +399,19 @@ def process_weights_after_loading(self, act_dtype: torch.dtype):
 
 
 class MLAAttentionImpl(AttentionImpl[T], Generic[T]):
+    if TYPE_CHECKING:
+        W_K: torch.Tensor | None
```
ProExpertProg: Why not: 
```suggestion
    W_K: torch.Tensor | None = None
```

vllm/attention/backends/abstract.py
```diff
@@ -398,6 +399,19 @@ def process_weights_after_loading(self, act_dtype: torch.dtype):
...
+        W_K_scale: torch.Tensor | None
+        W_V: torch.Tensor | None
+        W_V_scale: torch.Tensor | None
+        W_UV: torch.Tensor | None
+        W_UK_T: torch.Tensor | None
+
+        # Optional methods for prefill operations
+        # These are set dynamically in MLACommonImpl.__init__()
+        _run_prefill_context_chunk: Callable | None
+        _run_prefill_new_tokens: Callable | None
```
ProExpertProg: I think these are for CP (context-parallel) - cc @MatthewBonanni @LucasWilkinson.

Let's make these abstract methods instead of members. We can keep the names as-is for now but we should really rename these to something clearer after this PR (and not starting with _ as they're now part of an interface contract between layer and impl and not just internal impl details)

MatthewBonanni: These are actually for all prefill generally. I agree it would be better to have them be abstract methods even though they're currently members on TOT. Eventually we should likely clean up the selection logic for the prefill backend.

vllm/attention/backends/abstract.py
therealnaveenkamal: Since they were initialized and used in layer.py, just added them only for TYPE_CHECKING. I've modified them to class attributes now.

vllm/attention/backends/abstract.py
therealnaveenkamal: _run_prefill_context_chunk and _run_prefill_new_tokens are now abstract methods.

common.py implements these methods - calls the appropriate prefill method initialized earlier in `MLACommonImpl.__init__`

vllm/attention/layer.py
therealnaveenkamal: This was done to avoid early CUDA initialization error: `RuntimeError: Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method`

This was due to one of the imports. I've added the imports inside __init__, used the same logic and removed it from common.py (whichever can be removed)

vllm/v1/attention/backends/mla/common.py
therealnaveenkamal: Yes, but both `flashmla_sparse.py` and `rocm_aiter_mla_sparse.py` use `_v_up_proj` in their decode path (`self._v_up_proj(attn_out, out=output[:num_actual_toks])`), and they inherit that implementation from `MLACommonImpl` in `common.py`, so I think keeping `_v_up_proj` there is necessary for now.

Maybe in the 2/2 PR, I could refactor things so that the common logic lives in one place by making the backends call `layer._v_up_proj` or define a new function for this. Any thoughts @ProExpertProg

ProExpertProg: Sounds good, thanks for clarifying! Can you add a comment next to this method with this explanation?

vllm/attention/backends/abstract.py
ProExpertProg: Oh, I see now - these are not dependent on the backend class but have their own dispatching logic. I did not realize this.

@MatthewBonanni do you think it makes sense to initialize a "sub"-backend where we follow GQA logic for backend init and use that backend for MLA prefill? Either way this is fine for now.

MatthewBonanni: I don't think so, because the logic isn't identical. The MLA prefills are all ragged so not all backends could be used for MLA prefill. It would also likely require another layer of abstraction lower than `AttentionBackend`. I'm not sure if it would be worth it. I think it would be better to just simplify the interface -- the current options (`disable_flashinfer_prefill`, `use_cudnn_prefill`, `use_trtllm_ragged_deepseek_prefill`) are confusing to navigate

ProExpertProg: Ok, we can keep the dispatching logic custom and clean it up later.

vllm/v1/attention/backends/mla/common.py
```diff
@@ -1158,157 +1155,31 @@ def __init__(
...
+        # Skip dense-only initialization for sparse backends
+        if use_sparse:
+            return
 
         if use_flashinfer_prefill():
             logger.debug_once("Using FlashInfer prefill for MLA")
-            self._run_prefill_context_chunk = self._run_prefill_context_chunk_fi
-            self._run_prefill_new_tokens = self._run_prefill_new_tokens_fi
+            self._prefill_context_chunk_impl = self._run_prefill_context_chunk_fi
+            self._prefill_new_tokens_impl = self._run_prefill_new_tokens_fi
```
LucasWilkinson: why rename?

ProExpertProg:
Closing in favor of #31919

PR #29782, opened 2025-12-01
[Hardware][AMD] Remove ROCm skip conditions for transformers backend tests

Abdennacer-Badaoui:
## Summary

Removes skip conditions for transformers backend tests on ROCm platform. These tests now pass successfully, so the skips are no longer needed. 

@hmellor

### DIFF
--- a/tests/models/test_transformers.py
+++ b/tests/models/test_transformers.py
@@ -6,8 +6,6 @@ from typing import Any
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


hmellor (approved):
LGTM! Good to see the Transformers modelling backend becoming more compatible with ROCm!

DarkLight1337 (approved):

PR #29784, opened 2025-12-01
[Doc] Update description disable_any_whitespace

FredericOdermatt:
## Purpose

Hi @russellb and @mgoin 

Thanks for your great work on structured outputs in VLLM.

I've been working with both xgrammar and llguidance through VLLM for a while with `json`, `regex` and `lark/grammar` and found the description for `disable_any_whitespace` to be quite general:

```
If `True`, the model will not generate any whitespace during structured
outputs. This is only supported for xgrammar and guidance backends.
```

Two goals:
- specify that this only applies to `json` output, which it does
- give a note of what the reason for the flag is

Proposal:

```python
disable_any_whitespace: bool = False
  """If `True`, guided_json output will always be compact without any
  whitespace. If `False`, the model may generate whitespace between JSON
  fields, which is still valid JSON. This is only supported for xgrammar
  and guidance backends."""
```

## Test Plan
None
## Test Result
None

### DIFF
--- a/vllm/config/structured_outputs.py
+++ b/vllm/config/structured_outputs.py
@@ -28,8 +28,10 @@ class StructuredOutputsConfig:
     disable_fallback: bool = False
     """If `True`, vLLM will not fallback to a different backend on error."""
     disable_any_whitespace: bool = False
-    """If `True`, the model will not generate any whitespace during structured
-    outputs. This is only supported for xgrammar and guidance backends."""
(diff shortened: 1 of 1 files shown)


russellb (approved):
Thank you for the doc improvement!

hmellor (approved):

yewentao256 (approved):
LGTM, thanks for the work!

jeejeelee (approved):

PR #29788, opened 2025-12-01
Use helper function instead of looping through attribute names

hmellor:
Since we have the helper function, it'd be better to use it so that the behaviour is consistent.


vllm/config/model.py
```diff
@@ -1314,14 +1311,10 @@ def get_total_num_kv_heads(self) -> int:
...
         ]
-        for attr in attributes:
-            num_kv_heads = getattr(self.hf_text_config, attr, None)
-            if num_kv_heads is not None:
-                return num_kv_heads
-
         # For non-grouped-query attention models, the number of KV heads is
         # equal to the number of attention heads.
-        return self.hf_text_config.num_attention_heads
+        default = self.hf_text_config.num_attention_heads
```
ProExpertProg: Does `self.hf_text_config.num_attention_heads` always exist? If not, existing code might not fail but current code certainly will. Could we pass a lambda default instead?

```suggestion
        default = lambda: self.hf_text_config.num_attention_heads
```

hmellor: Ah good point previously the default was evaluated lazily

ProExpertProg (approved):

PR #29789, opened 2025-12-01
[Bugfix] Handle layer name inconsistencies in pipeline parallel training

penfever:
## Purpose

Fix vLLM v1 engine to properly handle pipeline parallelism (PP) by correctly managing KV cache groups when layers are distributed across different ranks.

Currently, when using PP > 1, the KV cache configuration is created globally with all layer names, but each PP rank only has a subset of layers. This causes failures when:

  1. init_attn_backend tries to access layers that don't exist on the current rank
  2. _allocate_kv_cache creates tensors for layers that aren't local
  3. build_attn_metadata attempts to build metadata for non-existent layers

  This PR introduces:
  - resolve_layers_from_vllm_config() - returns both found layers AND missing layer names for better visibility
  - _prune_kv_cache_group_layers() - prunes KV cache groups to only include layers local to the current rank
  - Proper handling of None attention metadata builders for empty groups
  - Debug logging for skipped remote layers

## Test Plan

# Run the new unit test
  pytest tests/config/test_vllm_layers.py -v

  # Run with pipeline parallelism (requires multi-GPU)
  vllm serve <model> --pipeline-parallel-size 2

## Test Result

- Pre-commit checks pass (ruff, mypy, typos, clang-format)
- Unit tests pass for layer resolution logic

---
<details>

</details>

### DIFF
--- a/tests/config/test_vllm_layers.py
+++ b/tests/config/test_vllm_layers.py
@@ -1,0 +1,30 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+from vllm.config import (
+    VllmConfig,
+    get_layers_from_vllm_config,
+    resolve_layers_from_vllm_config,
+)
+
+
+class _DummyLayer:
+    pass
+
+
+def test_get_layers_from_vllm_config_skips_missing_entries():
+    cfg = VllmConfig()
+    cfg.compilation_config.static_forward_context = {
+        "layer_present": _DummyLayer(),
+    }
+
+    layers = get_layers_from_vllm_config(
+        cfg, _DummyLayer, ["layer_present", "layer_missing"]
+    )
+    assert list(layers.keys()) == ["layer_present"]
+
+    layers_with_missing, missing = resolve_layers_from_vllm_config(
+        cfg, _DummyLayer, ["layer_present", "layer_missing"]
+    )
+    assert list(layers_with_missing.keys()) == ["layer_present"]
+    assert missing == ["layer_missing"]
--- a/vllm/attention/layer.py
+++ b/vllm/attention/layer.py
@@ -820,6 +820,15 @@ def get_attention_context(
     forward_context: ForwardContext = get_forward_context()
     attn_metadata = forward_context.attn_metadata
     if isinstance(attn_metadata, dict):
+        if layer_name not in attn_metadata:
+            available_layers = list(attn_metadata.keys())[:10]
+            raise KeyError(
+                f"Layer '{layer_name}' not found in attn_metadata. "
+                f"Available layers (first 10): {available_layers}. "
+                f"Total layers in metadata: {len(attn_metadata)}. "
+                f"This may indicate a pipeline parallelism configuration issue "
+                f"where the KV cache config doesn't match the layers on this rank."
+            )
         attn_metadata = attn_metadata[layer_name]
     attn_layer: Attention | MLAAttention = forward_context.no_compile_layers[layer_name]
     kv_cache = attn_layer.kv_cache[forward_context.virtual_engine]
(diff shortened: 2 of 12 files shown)


heheda12345 (commented):
why do we need to do this. We have these logic to generate the kv_cache_config for each worker only with layers in that worker
https://github.com/vllm-project/vllm/blob/afb1e5b380ff623e478d19a246b42b2903b9331f/vllm/v1/core/kv_cache_utils.py#L1354-L1375

penfever:
> why do we need to do this. We have these logic to generate the kv_cache_config for each worker only with layers in that worker
> 
> https://github.com/vllm-project/vllm/blob/afb1e5b380ff623e478d19a246b42b2903b9331f/vllm/v1/core/kv_cache_utils.py#L1354-L1375

Because the runner sometimes mutates the per-worker KVCacheConfig after the config codepath in kv_cache_utils.py completes. The code you pointed at filters kv_cache_groups per worker correctly at the moment the config is produced, but once the config reaches the GPU runner we immediately append more “runner-only” layers: GPUModelRunner.may_add_encoder_only_layers_to_kv_cache_config(), GPUModelRunner.maybe_add_kv_sharing_layers_to_kv_cache_groups() ... those extra layer names never existed when the KV tensors were planned, so kv_cache_tensors still describe only the scheduler-visible layers. When _allocate_kv_cache_tensors() later asserts that every layer in every group has a tensor, the newly inserted runner-only layers make the set comparison fail (a “Some layers are not correctly initialized” assertion). 

Skipping self.runner_only_attn_layers on both sides lets the runner alias the real KV tensors later.

heheda12345:
Can you give us a command that we can reproduce this problem?

penfever:
> Can you give us a command that we can reproduce this problem?

This simulates the problem in vllm==0.11.2 -- reproductions in production are hard because the model is so large.

```
#!/usr/bin/env python3
"""
Reproduces the KeyError that the QuantTrio AWQ loader triggers in vLLM 0.11.2
without fetching a real model.

The new v1 engine records each attention module inside
`compilation_config.static_forward_context`. When a loader later swaps out the
module (as QuantTrio does) but leaves the registry untouched, the next call to
`get_layers_from_vllm_config` raises `KeyError`.

Run with:  python simulate_awq_swap_bug.py
Requires: pip install "vllm==0.11.2"
"""

from types import SimpleNamespace

from vllm.config import get_layers_from_vllm_config


def main() -> None:
    layer_name = "model.layers.69.self_attn.attn"

    # Minimal VllmConfig-like object: only the attribute used by
    # get_layers_from_vllm_config() matters.
    forward_context = {layer_name: object()}
    vllm_config = SimpleNamespace(
        compilation_config=SimpleNamespace(static_forward_context=forward_context)
    )

    # Sanity check: works while the registry still has the entry.
    get_layers_from_vllm_config(vllm_config, object, [layer_name])

    # Simulate an AWQ loader swapping out the attention module after
    # initialization without refreshing static_forward_context.
    forward_context.pop(layer_name)

    # The very next lookup now crashes exactly like the production stack.
    get_layers_from_vllm_config(vllm_config, object, [layer_name])


if __name__ == "__main__":
    main()

```

penfever:
vllm serve QuantTrio/GLM-4.6-AWQ \
  --backend ray \
  --ray-address auto \
  --tensor-parallel-size 1 \
  --pipeline-parallel-size 16 \
  --max-model-len 131072 \
  --max-num-seqs 16 \
  --swap-space 12 \
  --enable-expert-parallel \
  --enable-auto-tool-choice \
  --tool-call-parser glm45 \
  --reasoning-parser glm45 \
  --trust-remote-code \
  --dtype bfloat16 \
  --kv-cache-dtype fp8

on 16xGH200 nodes crashes with "layer not found" errors such as the ones produced by the simulation script.

PR #29790, opened 2025-12-01
Fix missing as_list() conversion in streaming chat completion

kennyu:
## Purpose

Fixes #29763
Add as_list() wrapper to output.token_ids in two code paths that were missing it:
- Line 1034: tool_choice_auto only path
- Line 1048: reasoning only path

This fixes GLM-4.5 reasoning parser streaming when no tools are included in the request. The token_ids field is a GenericSequence which may not be a list, causing issues with the `in` operator in parser methods.

### DIFF
--- a/vllm/entrypoints/openai/serving_chat.py
+++ b/vllm/entrypoints/openai/serving_chat.py
@@ -1031,7 +1031,7 @@ class OpenAIServingChat(OpenAIServing):
                             delta_text=delta_text,
                             previous_token_ids=previous_token_ids,
                             current_token_ids=current_token_ids,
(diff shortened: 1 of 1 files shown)


njhill:
@kennyu these abstract methods in the `ReasoningParser` interface declare the args as `Sequence[int]` so I'm not sure this is the "correct" fix. If a particular implementation requires it to be a list then better to add the `as_list()` within that impl.

However, `Sequence` should be sufficient to use the `in` operator, so are you sure this is actually needed? Could you elaborate on the specific error encountered?

njhill:
Feel free to reopen if needed

PR #29805, opened 2025-12-01
[Misc][Hybrid allocator + kv connector] Optionally enable hybrid allocator + KV cache connector

NickLucche:
This PR contains a simple toggle for enabling the experimental HMA + KVConnector integration, needed to build actual HMA support into existing connectors.
This feature is still under development so only the LMCache will actually work with it out of the box.

cc @KuntaiDu @heheda12345 @ivanium


vllm/config/vllm.py
```diff
@@ -872,19 +872,25 @@ def has_blocked_weights():
...
+                    logger.warning(
+                        "Turning off hybrid kv cache manager because "
+                        "`--kv-transfer-config` is set. This will reduce the "
+                        "performance of vLLM on LLMs with sliding window attention "
+                        "or Mamba attention. If you are a developer of kv connector"
+                        ", please consider supporting hybrid kv cache manager for "
+                        "your connector by making sure your connector is a subclass"
+                        " of `SupportsHMA` defined in kv_connector/v1/base.py."
+                    )
+                    self.scheduler_config.disable_hybrid_kv_cache_manager = True
```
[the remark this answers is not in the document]
NickLucche: that's intended, given that now there's no way to enable them in the first place

heheda12345: what about making disable_hybrid_kv_cache_manager a `bool | None` so that we can follow the user-specified config if it is not None and decide it automatically if it is None. Like this: 
```
num_gpu_blocks_override: int | None = None
```

vllm/config/vllm.py
```diff
@@ -928,6 +918,39 @@ def has_blocked_weights():
                     # local attention.
                     self.scheduler_config.disable_hybrid_kv_cache_manager = True
 
+            if (
+                prev_disable_hma is False
+                and self.scheduler_config.disable_hybrid_kv_cache_manager is True
+            ):
+                logger.info(
```
KuntaiDu: For this code path, maybe we can raise a `NotImplementedError` with some explanation instead of a warning? Because for me if the user intentionally set `--no-disable-hybrid-manager` they are not expecting vLLM falling back to non-hybrid-allocator code path.

Explanation example:
```
Hybrid KV cache manager is explicitly enabled, but currently `--kv-events-config` is set and KV events code path is not compatible with hybrid kv cache manager.
```

vllm/config/vllm.py
```diff
@@ -928,6 +918,39 @@ def has_blocked_weights():
...
+            # Disable HMA logic but only if the user didn't express a preference.
+            # NOTE(Kuntai): turn HMA off for connector unless specifically enabled.
+            # TODO(Kuntai): have a more elegent solution to check and
+            # turn off HMA for connector that does not support HMA.
+            logger.warning(
+                "Turning off hybrid kv cache manager because "
+                "`--kv-transfer-config` is set. This will reduce the "
+                "performance of vLLM on LLMs with sliding window attention "
+                "or Mamba attention. If you are a developer of kv connector"
+                ", please consider supporting hybrid kv cache manager for "
```
KuntaiDu: Would be great if we can also remind people here that they should set `--no-disable-hybrid-kv-cache-manager` if they want to develop HMA + connector.

KuntaiDu (commented):
Thanks for the contribution!
See comments.

vllm/config/vllm.py
NickLucche: I see your point, the idea was that I didn't want this PR to affect existing behavior. 
I felt like logging now and then swapping to a `NotImplementedError`  later on when we've given users time to omit `--no--disable-hybrid-manager`, would actually be a more desirable ux.

NickLucche:
Thanks for reviewing @KuntaiDu ! I've addressed your comments

vllm/config/vllm.py
ivanium: Would `logger.warning` better fit here?

ivanium (commented):
Overall the PR looks good to me too! I just have one question. Right now we still disable the hybrid allocator whenever `kv_transfer_config is not None`. When we eventually support hybrid allocation per KV connector, what’s the best practice you envision? For example, if we want hybrid allocation enabled only for a specific subset of connectors, will there be a mechanism for developers to enforce “hybrid allocator + chosen connector” together?

vllm/config/vllm.py
KuntaiDu: Oh I thought this code path will only be triggered when `--no-disable-hybrid-manager` is set. Let me double check.

NickLucche: chatted offline about this, we're raising exception on this PR

vllm/config/vllm.py
```diff
@@ -889,27 +889,16 @@ def has_blocked_weights():
         if not self.instance_id:
             self.instance_id = random_uuid()[:5]
 
+        # Runtime-dependent disable of hybrid kv cache manager logic.
         if not self.scheduler_config.disable_hybrid_kv_cache_manager:
+            prev_disable_hma = self.scheduler_config.disable_hybrid_kv_cache_manager
```
heheda12345: What about:
```
need_disable_hybrid_kv_cache_manager = False
if not current_platform.support_hybrid_kv_cache():
    need_disable_hybrid_kv_cache_manager = True
if ***: need_disable_hybrid_kv_cache_manager = True
...
if self.scheduler_config.disable_hybrid_kv_cache_manager  is None:
    if self.kv_transfer_config is not None:
         # Experimental feature. Default to disable but allow users to enable.
         need_disable_hybrid_kv_cache_manager = True
         logger.warning(***)
    self.scheduler_config.disable_hybrid_kv_cache_manager = need_disable_hybrid_kv_cache_manager
elif self.scheduler_config.disable_hybrid_kv_cache_manager == False:
    if need_disable_hybrid_kv_cache_manager: raise xxxx
```
I feel prev_disable_hma is a little bit hacky.

NickLucche: @heheda12345 I am sorry ~~but I don't see how checking whether a bool was flipped here is hacky~~.
EDIT: I think I see what you mean now, you're referring to clarity. Don't have a strong opinion on that, will check it out when I have the time.

I've separated features that do NOT work with HMA from features that _may_ work with HMA such as kv connector (a later check on supports_hma is carried out when attempting to create the actual connector).

Attempting to enable HMA explicitly on the former will crash the server. `prev_disable_hma` is just for checking whether the flag was set by the user without adding more lines such as `need_disable_hybrid_kv_cache_manager`.

heheda12345 (approved):
LGTM! Thank you very much.

PR #29813, opened 2025-12-01
[Core] Fix livelock in shm_broadcast under high concurrent load

kitaekatt:
## Summary

This PR fixes a livelock bug in the shared memory broadcast mechanism (`shm_broadcast.py`) that causes the V1 engine to freeze under sustained high concurrent load.

## Problem

When running with high concurrency settings (e.g., `max_num_seqs=62`) under sustained batch inference load, the V1 multiprocess executor freezes completely after processing ~492 items. All worker threads become blocked on `futex_wait_queue` with no error messages.

**Root cause**: The spin loops in `acquire_write()` and `acquire_read()` use pure `sched_yield()`, which can cause livelock when multiple processes are spinning simultaneously on shared memory. The OS scheduler can immediately reschedule the same process, creating a cycle where no process makes progress.

## Solution

Introduce a new `SpinBackoffTimer` class that adds periodic backoff sleeps (1ms every 1000 spins) to break potential livelock patterns. This is a minimal change that:

- Preserves low latency during normal operations (most spins still use `sched_yield()`)
- Breaks livelock patterns by ensuring processes yield long enough for others to acquire resources
- Is configurable via constructor parameters if tuning is needed

## Changes

- Add `SpinBackoffTimer` class with periodic backoff mechanism
- Replace `SpinTimer` with `SpinBackoffTimer` for the default (non-`VLLM_SLEEP_WHEN_IDLE`) case
- Add `_write_spin_timer` for `acquire_write()` spin loop

## Testing

**Before fix**: Server freezes at ~492 items (batch 41 of 12-concurrent request batches)

**After fix**: Successfully completed multiple runs of 120 batches × 12 concurrent requests (1440 items each) without freeze

Test configuration:
- Model: Qwen2.5-32B-Instruct-AWQ
- GPU: NVIDIA RTX 5090 (32GB)
- Settings: `max_num_seqs=62`, `max_model_len=4096`, `gpu_memory_utilization=0.68`

## Test script

```python
#!/usr/bin/env python3
"""Aggressive stress test - 120 batches to ensure freeze trigger."""
import asyncio
import aiohttp
import time

SERVER_URL = "http://127.0.0.1:8008"

async def make_request(session, i):
    try:
        async with session.post(
            f"{SERVER_URL}/v1/chat/completions",
            json={
                "model": "Qwen2.5-32B-Instruct-AWQ",
                "messages": [{"role": "user", "content": f"What is {i} + {i}?"}],
                "max_tokens": 20
            },
            timeout=aiohttp.ClientTimeout(total=60)
        ) as resp:
            return "ok" if resp.status == 200 else f"err:{resp.status}"
    except asyncio.TimeoutError:
        return "timeout"

async def run_batch(batch_num, concurrency=12):
    async with aiohttp.ClientSession() as session:
        tasks = [make_request(session, batch_num*concurrency + i) for i in range(concurrency)]
        return await asyncio.gather(*tasks)

async def main():
    for batch in range(120):
        results = await run_batch(batch, 12)
        ok_count = sum(1 for r in results if r == "ok")
        if ok_count == 0:
            print(f"FREEZE DETECTED at batch {batch+1}")
            return
    print("SUCCESS: Completed 120 batches without freeze!")

if __name__ == "__main__":
    asyncio.run(main())
```

### DIFF
--- a/vllm/distributed/device_communicators/shm_broadcast.py
+++ b/vllm/distributed/device_communicators/shm_broadcast.py
@@ -60,11 +60,49 @@ def long_wait_time_msg(threshold: int) -> str:
 
 
 class SpinTimer:
+    """Base spin timer that yields to the scheduler on each spin."""
+
     def record_activity(self):
         pass
 
     def spin(self):
         sched_yield()
+
+
+class SpinBackoffTimer(SpinTimer):
+    """
+    Spin timer with periodic backoff to prevent livelock under high contention.
+
+    Under sustained high load, pure sched_yield() can cause livelock when
+    multiple processes are spinning on shared memory. This timer adds a small
+    sleep every N spins to break potential livelock patterns.
+
+    The backoff interval is tuned to add ~1us of delay per spin on average,
+    which is enough to break livelocks without significantly impacting latency.
+    """
+
+    def __init__(self, backoff_interval: int = 1000, backoff_sleep_us: float = 1000):
+        """
+        Args:
+            backoff_interval: Number of spins between backoff sleeps
+            backoff_sleep_us: Backoff sleep duration in microseconds
+        """
+        self._spin_count = 0
+        self._backoff_interval = backoff_interval
+        self._backoff_sleep_s = backoff_sleep_us / 1_000_000
+
+    def record_activity(self):
+        # Reset spin count on activity to maintain low latency during normal ops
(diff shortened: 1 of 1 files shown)


njhill:
Thanks @kitaekatt!!

PR #29826, opened 2025-12-02
[Perf] Avoid pageable HtoD transfer in MinTokensLogitsProcessor

jthomson04:
## Purpose
The `MinTokensLogitsProcessor` introduces a pageable HtoD transfer here: https://github.com/vllm-project/vllm/blob/342c4f147266e7d01a8a6a64754df7e0f7f8991a/vllm/v1/sample/logits_processor/builtin.py#L232. This is problematic, since it is synchronizing wrt the device. When running with `--async-scheduling`, this synchronizing behavior causes the scheduler to no longer be overlapped with model execution. At low concurrencies, this leads to a significant (up to 30%) increase in TPOT when including min_tokens in the request.

I collected some nsys traces illustrating this.

Without min_tokens in request:
image

With min_tokens in request:
image

## Test Plan
Existing tests for `MinTokensLogitsProcessor` are sufficient; no further tests are required.

## Test Result

---
<details>

</details>


vllm/v1/sample/logits_processor/builtin.py
```diff
@@ -178,6 +178,13 @@ def __init__(
             self._device_tensor([], torch.int32),
         )
 
+        self.neg_inf_tensor = torch.full(
+            (),
+            -float("inf"),
+            dtype=torch.float32,
+            device=device,
+        )
```
njhill: Just use `torch.tensor(-float("inf"), ...`?

vllm/v1/sample/logits_processor/builtin.py
```diff
@@ -110,7 +110,7 @@ def apply(self, logits: torch.Tensor) -> torch.Tensor:
         # Identify valid tokens using threshold comparison
         invalid_token_mask = probability_values < adjusted_min_p
         # Apply mask using boolean indexing
-        logits[invalid_token_mask] = -float("inf")
+        logits[invalid_token_mask] = self.neg_inf_tensor
```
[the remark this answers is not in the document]
njhill: This appears to be correct - you need to create this in the init method of this class too

njhill (commented):
Great catch, thanks @jthomson04!

vllm/v1/sample/logits_processor/builtin.py
nvpohanh: could we try `logits[invalid_token_mask].fill_(-float("inf"))` so that we can avoid creating `neg_inf_tensor`?

nvpohanh (approved):

vllm/v1/sample/logits_processor/builtin.py
nvpohanh: I asked @jthomson04 to remove `neg_inf_tensor`

njhill (approved):

njhill:
The test failure looks likely related given it's to a logitsproc test, I can look into it soon.

PR #29829, opened 2025-12-02
[BugFix] add max-num-batched-token to scheduler hash

BoyuanFeng:
Before this PR, max-num-batched-token is not added to compute_hash by default, which decides whether recompile if this config changes.

However, max_num_batched_tokens need to be included in the hash due to two reasons:
1. LoRA creates static buffers based on max_num_batched_tokens. The tensor sizes and strides get captured in the torch.compile graph explicitly.
2. Inductor decides whether using 32-bit or 64-bit indexing integer based on the data sizes. `max_num_batched_tokens` has an impact on that. For more details, please check #29585

This PR fixed the issue.

cc @zou3519 

Close: #29585


njhill (approved):
Thanks @BoyuanFeng

ProExpertProg (approved):

PR #29844, opened 2025-12-02
[KV Transfer] Add optional int8 quantization for KV cache transfer to P2pNcclEngine

xbfs:
## Summary

This PR adds optional int8 quantization support for KV cache transfer in `P2pNcclEngine` to reduce network bandwidth usage and improve transfer throughput, especially in bandwidth-constrained scenarios.

## Motivation

When transferring KV cache across nodes in distributed inference setups, network bandwidth can become a bottleneck. By quantizing tensors from float16/float32 to int8 during transfer, we can reduce the data size by approximately 50-75%, significantly improving transfer speed in scenarios where network bandwidth is limited.

## Changes

- Added `quantize_kv_on_transfer` configuration option (boolean, default: `false`)
- Implemented symmetric int8 quantization with per-tensor scaling
- Added `_quantize_tensor_for_send()` and `_maybe_dequantize_tensor()` helper methods
- Integrated quantization into both PUT and GET transfer paths
- Updated logging to include quantization status

## Technical Details

- **Quantization method**: Symmetric quantization with per-tensor scale factor
  - Formula: `q = clamp(round(x / scale), -128, 127)`
  - Scale calculation: `scale = max(abs(x)) / 127.0`
- **Dequantization**: `x = q * scale`
- **Metadata transfer**: Quantization metadata (scale, original dtype) is transmitted via ZMQ messages
- **Compatibility**: Works with PUT, PUT_ASYNC, and GET transfer modes

## Usage

Enable quantization by setting `quantize_kv_on_transfer: true` in the KV connector extra config:

{
  "kv_connector": "P2pNcclConnector",
  "kv_connector_extra_config": {
    "quantize_kv_on_transfer": true
  }
}

## Benefits

- **Bandwidth reduction**: ~50-75% reduction in transfer data size for float16/float32 tensors
- **Improved throughput**: Faster transfer times in bandwidth-constrained environments
- **Zero breaking changes**: Default behavior unchanged (quantization disabled by default)
- **Backward compatible**: Existing configurations continue to work without modification

## Use Cases

This optimization is particularly beneficial for:
- Cross-node transfers (multi-machine, multi-GPU setups)
- Bandwidth-limited networks (standard Ethernet vs InfiniBand)
- Long sequence lengths (large KV cache sizes)
- High concurrency scenarios (multiple simultaneous transfers)

## Testing

The implementation has been tested with:
- PUT and PUT_ASYNC transfer modes
- Various tensor shapes and dtypes (float16, float32)
- Edge cases (zero tensors, very small/large values)

## Backward Compatibility

- Quantization is **disabled by default** (`quantize_kv_on_transfer: false`)
- No changes to existing APIs or behavior when quantization is disabled
- All existing configurations continue to work without modification

## Future Improvements

Potential enhancements for future PRs:
- Support for different quantization bit-widths (4-bit, etc.)
- Per-channel quantization for better accuracy
- Quantization-aware training integration


robertgshaw2-redhat:
@ApostaC - can you be on point for this one?

vllm/distributed/kv_transfer/kv_connector/v1/p2p/p2p_nccl_engine.py
```diff
@@ -586,6 +599,42 @@ def ping(self):
...
+            return tensor, None
+
+        abs_max = tensor.abs().max()
+        max_value = abs_max.item() if abs_max.numel() > 0 else 0.0
+        if max_value == 0:
+            scale = 1.0
+        else:
+            scale = float(max_value / 127.0)
+            if scale == 0:
+                scale = 1.0
```
ApostaC: Is there any case we have `scale == 0` here?  IIUC, it's already handled by the above `max_value == 0` branch.

ApostaC: It would be great if we could simplify the logic here for better readability.

ApostaC (commented):
Just some minor comments. Otherwise LGTM!

vllm/distributed/kv_transfer/kv_connector/v1/p2p/p2p_nccl_engine.py
xbfs: Thank you for the feedback! You're absolutely right. 

The `scale == 0` check on line 614-615 was indeed redundant since the `max_value == 0` condition is already handled above. I've simplified the logic to make it more readable:

**Before:**
if max_value == 0:
    scale = 1.0
else:
    scale = float(max_value / 127.0)
    if scale == 0:
        scale = 1.0
**After:**
scale = float(max_value / 127.0) if max_value > 0 else 1.0

This single-line ternary expression is clearer and eliminates the redundant check. Since we guarantee that `scale` is never 0 in the quantization function, I've also removed the unnecessary guard in the `_maybe_dequantize_tensor` function for consistency.

The changes have been pushed. Thanks again for helping improve the code quality! 🙏

xbfs:
Hi @ApostaC,
May I know if anything else is needed for this PR to be merged?  
I'd be happy to update it.

PR #29856, opened 2025-12-02
[Model] Add LoRA support for Whisper models

daje0601:
## Purpose

This PR enables Multi-LoRA support for Whisper speech-to-text models, allowing users to serve multiple fine-tuned Whisper adapters from a single base model.

### Background

Currently, vLLM's `WhisperForConditionalGeneration` does not implement the `SupportsLoRA` interface, preventing users from using LoRA adapters with Whisper models. This limitation requires users to deploy
 separate model instances for each fine-tuned variant, which is inefficient in terms of GPU memory usage.

### Changes

**1. `vllm/model_executor/models/whisper.py`**
- Add `SupportsLoRA` interface to `WhisperForConditionalGeneration`
- Add `embedding_modules` and `embedding_padding_modules` attributes required by LoRA
- Update `packed_modules_mapping` to use simplified keys (`qkv_proj`, `kv_proj`) for LoRA compatibility

**2. `vllm/lora/layers/column_parallel_linear.py`**
- Extend `MergedQKVParallelLinearWithLoRA` to support KV-only (2-slice) configurations
- This is necessary because Whisper's cross-attention layers (`encoder_attn.kv_proj`) only have K and V projections, not Q
- Update `can_replace_layer()` to accept both 2-module and 3-module configurations
- Refactor `slice_lora_a()` to dynamically handle variable number of slices

**3. `vllm/lora/worker_manager.py`**
- Add fallback to `max_target_positions` when `max_position_embeddings` is not available
- Whisper config uses `max_target_positions` instead of `max_position_embeddings`

**4. `examples/offline_inference/whisper_multilora_inference.py`**
- Add example script demonstrating Whisper Multi-LoRA inference

**5. `tests/lora/test_whisper_lora.py`**
- Add unit tests for Whisper LoRA interface compliance
- Add tests for KV-only configuration support
- Add tests for WorkerLoRAManager Whisper compatibility

## Test Plan

```bash
# Run unit tests
pytest tests/lora/test_whisper_lora.py -v

# Run example (requires LoRA adapter)
python examples/offline_inference/whisper_multilora_inference.py
```

Test Result(Unit Tests)
```
tests/lora/test_whisper_lora.py::TestWhisperLoRAInterface::test_supports_lora_attribute PASSED
tests/lora/test_whisper_lora.py::TestWhisperLoRAInterface::test_embedding_modules_defined PASSED
tests/lora/test_whisper_lora.py::TestWhisperLoRAInterface::test_embedding_padding_modules_defined PASSED
tests/lora/test_whisper_lora.py::TestWhisperLoRAInterface::test_packed_modules_mapping_format PASSED
tests/lora/test_whisper_lora.py::TestMergedQKVParallelLinearWithLoRAKVOnly::test_can_replace_layer_accepts_2_modules PASSED
tests/lora/test_whisper_lora.py::TestWorkerLoRAManagerWhisperCompat::test_max_position_embeddings_fallback PASSED
tests/lora/test_whisper_lora.py::TestWorkerLoRAManagerWhisperCompat::test_max_position_embeddings_priority PASSED
```

Manual Testing
Tested with openai/whisper-large-v3-turbo base model and custom LoRA adapters:
- Server startup with --enable-lora flag: ✅
- Single LoRA inference: ✅
- Multi-LoRA switching between requests: ✅
- Concurrent requests with different LoRAs: ✅

Example Usage
```
from vllm import LLM
from vllm.lora.request import LoRARequest

# Initialize with LoRA support
llm = LLM(
    model="openai/whisper-large-v3-turbo",
    enable_lora=True,
    max_loras=4,
    max_lora_rank=64,
)

# Use different LoRA adapters per request
outputs = llm.generate(
    inputs,
    lora_request=LoRARequest("my_whisper_lora", 1, "/path/to/lora")
)
```
or 
```
vllm serve yourname/yourmodel \
--host 0.0.0.0 \
--port 8181 \
--dtype bfloat16 \
--trust-remote-code \
--enable-lora \
--lora-modules  \
lora1=lora_module_path \
lora2=lora_module_path \
--max-lora-rank 32 \
--gpu-memory-utilization 0.7 \
--tensor-parallel-size 1
```
---
- The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- The test plan, such as providing test command.
- The test results, such as pasting the results comparison before and after, or e2e results
- (Optional) The necessary documentation update, such as updating supported_models.md and examples for a new model.
- (Optional) Release notes update. If your change is user facing, please update the release notes draft in the https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0.


jeejeelee:
Will look at this PR ASAP, also cc @NickLucche

vllm/model_executor/models/whisper.py
```diff
@@ -773,16 +778,16 @@ def _get_prompt_updates(
     dummy_inputs=WhisperDummyInputsBuilder,
 )
 class WhisperForConditionalGeneration(
-    nn.Module, SupportsTranscription, SupportsMultiModal
+    nn.Module, SupportsTranscription, SupportsMultiModal, SupportsLoRA
 ):
+    # LoRA-specific attributes
+    embedding_modules = {}
+    embedding_padding_modules: list[str] = []
```
jeejeelee: If the model inherits from SupportsLoRA, these two attributes are empty by default

examples/offline_inference/whisper_multilora_inference.py
```diff
@@ -0,0 +1,136 @@
+# SPDX-License-Identifier: Apache-2.0
```
jeejeelee: It looks like this example is similar to multilora_inference.py, so do we need to add this example?

vllm/lora/layers/column_parallel_linear.py
```diff
@@ -398,7 +403,11 @@ def can_replace_layer(
         packed_modules_list: list,
         model_config: PretrainedConfig | None = None,
     ) -> bool:
-        return type(source_layer) is QKVParallelLinear and len(packed_modules_list) == 3
```
jeejeelee, line 410: Can we use `MergedColumnParallelLinear` rather than  `QKVParallelLinear` in base model?

jeejeelee (commented):
Thank you for your contribution. The main concern is that maybe we should use MergedColumnParallelLinear rather than QKVLinear in the base model

vllm/model_executor/models/whisper.py
daje0601: Thank you, I'll remove these redundant.

examples/offline_inference/whisper_multilora_inference.py
daje0601: You're right - it's similar to the existing multilora_inference.py.
 I'll remove whisper_multilora_inference.py from this PR.

vllm/lora/layers/column_parallel_linear.py
daje0601: I will:
  - Revert my changes to MergedQKVParallelLinearWithLoRA in column_parallel_linear.py
  - Update whisper.py to use MergedColumnParallelLinear for the cross-attention's  kv_proj layer

I'll update the PR with these changes shortly. Thanks again for the review!

daje0601:
@jeejeelee Would you please let me know if there's any additional work? 
If not, I'm planning to work on implementing multi-LoRA support for TTS as Well. 
Thank you!

tests/lora/test_whisper_lora.py
```diff
@@ -0,0 +1,144 @@
+# SPDX-License-Identifier: Apache-2.0
```
jeejeelee: Could you please delete this test script, I think this test is unnecessary.

jeejeelee (approved):
After removing the above test, LGTM, thank you for contribution

daje0601:
> Fantastic work :-) What is the timeline for merging that?

I've been waiting too~ If there's anything else I need to do on my end, could you please let me know?

tests/lora/test_whisper_lora.py
jeejeelee: @daje0601  I think we should delete this test script

daje0601: I didn't see this post before this, but I do now, so I'll delete it and push it again.

vllm/lora/worker_manager.py
```diff
@@ -49,7 +49,12 @@ def __init__(
         # Use get_text_config() in case of multimodal models
         text_config = vllm_config.model_config.hf_config.get_text_config()
 
-        self.max_position_embeddings = text_config.max_position_embeddings
+        # Whisper uses max_target_positions instead of max_position_embeddings
+        self.max_position_embeddings = getattr(
+            text_config,
+            "max_position_embeddings",
+            getattr(text_config, "max_target_positions", None),
+        )
```
NickLucche: you should probably check if is_encoder_decoder with `vllm_config.model_config.is_encoder_decoder`

NickLucche: and add a TODO to generalize for OOT enc-dec models

NickLucche (commented):
@daje0601 Thanks for you work!

Given popularity of the model, I think we should really be adding tests with whisper+some lora adapter.

vllm/lora/worker_manager.py
daje0601: Thanks, I'll check it out tonight!

daje0601: Could you please trigger the buildkite/ci/pr check when you have a chance? Thank you!

NickLucche (approved):
Thanks for your work @daje0601 !

daje0601:
> Fantastic work :-) What is the timeline for merging that?

Thanks! It's finally  landed in v0.18.0. You can try it out now with `pip install vllm>=0.18.0`

PR #29858, opened 2025-12-02
[CI][DCP][Perf] reduce DCP CI execution time

pisceskkk:
Based on https://github.com/vllm-project/vllm/pull/29487#pullrequestreview-3513500797 , I've re-optimized the DCP UT to reduce CI execution time. Referring to `test_dbo.py`, I changed the comparison from DCP vs. TP to DCP vs. reference accuracy, and removed unnecessary tests (I'm not entirely sure about this—if needed, they can be added back).
Currently, each test takes less than two minutes:
`4 passed, 4 warnings in 409.86s (0:06:49)`

Additionally, I'm not entirely clear about the model naming conventions in the `tests/models/registry.py` and am unsure what the key for the newly added `Qwen/Qwen2.5-1.5B-Instruct` model should be.

CC @LucasWilkinson

### DIFF
--- a/tests/distributed/test_context_parallel.py
+++ b/tests/distributed/test_context_parallel.py
@@ -20,11 +20,30 @@ from vllm.config.model import RunnerOption
 from vllm.logger import init_logger
 
 from ..models.registry import HF_EXAMPLE_MODELS
-from ..utils import compare_two_settings, create_new_process_for_each_test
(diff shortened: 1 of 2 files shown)


LucasWilkinson (approved):
Amazing thanks for improving this! 35min -> 6min really helps us! Will get someone to to review registry.py

PR #29877, opened 2025-12-02
[KVConnector] Prevent async loads from exploding the KV Cache

orozery:
Previously, async loaded requests were unbounded,
both in terms of number of requests and number of tokens.
This PR accounts async requests for both `max_num_running_reqs` and `token_budget`.
This will prevent cases where async loaded requests will unnecessarily fill up the entire GPU KV cache.


orozery:
With the recent support for resuming preempted requests from CPU, it seems the case describe here is not so bad.
There are possible heuristic optimizations to balance off KV cache usage between async loads and running requests.
But I think we can defer this until we have a real justified use-case.

PR #29880, opened 2025-12-02
Improve and supplement LoRA tuning documentation

caozuoba:
## Purpose
Improve and supplement LoRA tuning documentation
## Test Result
bc182bf14be8ee10687ffecfe0dda875

## PR Content Description
This PR adds a standalone tuning script for Triton LoRA kernels, `vllm/benchmarks/kernels/benchmark_lora_tuning.py`.

**High‑level behavior**

- Performs a grid search over Triton kernel hyperparameters (e.g. `block_m/n/k`, `num_warps`, `num_stages`, `split_k`, etc.) for LoRA and fused MoE LoRA ops.
- Reuses the existing `benchmark_lora.bench_optype` harness to run the kernels and measure median latency per configuration.
- Optionally exports the best configuration(s) per setting to JSON files that are compatible with `get_lora_op_configs` (keyed by `max_loras`, `num_slices`, `m`, `k`, `n` and optionally `moe_intermediate_size` for fused MoE).

**Patching / config hijacking**

- The script introduces a small monkey‑patching layer around `get_lora_op_configs`:
  - `_CURRENT_LORA_KERNEL_CONFIG` holds the “current candidate” Triton config during tuning.
  - `_patched_get_lora_op_configs(...)` first checks `_CURRENT_LORA_KERNEL_CONFIG`:
    - If it is set, it returns this candidate config instead of running the original logic.
    - For fused MoE LoRA ops, it also maps the lower‑case search‑space keys (`block_m`, `num_warps`, `split_k`, …) to the macro‑style keys expected by the existing code (`BLOCK_SIZE_M`, `NUM_WARPS`, `SPLIT_K`, …).
    - If no candidate is set, it falls back to the original `_ORIGINAL_GET_LORA_OP_CONFIGS(...)`.
  - `_ensure_patched()` applies this patched function to all relevant modules (`utils`, `lora_shrink_op`, `lora_expand_op`, `fused_moe_lora_op`, and the `benchmark_lora` module itself) exactly once.
- `_benchmark_single_config(...)` simply sets `_CURRENT_LORA_KERNEL_CONFIG = cfg` before calling `bench_optype(...)`, so every benchmark run uses that specific kernel config via the patched `get_lora_op_configs`.
- If Triton reports `OutOfResources`, the script treats that configuration as invalid (marks status as `out_of_resources` and skips it when choosing the best config).
---
<details>

</details>

### DIFF
--- a/benchmarks/kernels/benchmark_lora.py
+++ b/benchmarks/kernels/benchmark_lora.py
@@ -867,7 +867,7 @@ class BenchmarkTensors:
             "expert_ids": expert_ids,
             "num_tokens_post_padded": num_tokens_post_padded,
             "top_k_num": ctx.top_k_num,
-             # Use lora_ids from LoRAKernelMeta and synthetic adapter_enabled.
+            # Use lora_ids from LoRAKernelMeta and synthetic adapter_enabled.
             "lora_ids": self.lora_kernel_meta.active_lora_ids,
             "adapter_enabled": adapter_enabled,
             "device": self.input.device,
@@ -1001,9 +1001,7 @@ class BenchmarkTensors:
         ref_output = self.output.clone()
 
         self.output.zero_()
-        op_type.bench_fn()(
-            **self.bench_fn_kwargs(ctx, op_type, expand_fn_add_inputs)
-        )
+        op_type.bench_fn()(**self.bench_fn_kwargs(ctx, op_type, expand_fn_add_inputs))
 
         op_type.run_ref_group_gemm(
             ref_output,
--- a/benchmarks/kernels/benchmark_lora.py
+++ b/benchmarks/kernels/benchmark_lora.py
@@ -652,7 +652,15 @@ class BenchmarkTensors:
             num_tokens_post_pad = torch.empty(
                 (max_loras), dtype=torch.int32, device=topk_ids.device
             )
-
+            # Newer ops.moe_lora_align_block_size signature requires
+            # adapter_enabled and lora_ids. For benchmarking we can use
+            # synthetic values that enable all loras.
+            adapter_enabled = torch.ones(
+                (max_loras + 1,), dtype=torch.int32, device=topk_ids.device
+            )
+            lora_ids = torch.arange(
+                max_loras + 1, dtype=torch.int32, device=topk_ids.device
+            )
             ops.moe_lora_align_block_size(
                 topk_ids,
                 token_lora_mapping,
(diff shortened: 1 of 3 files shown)


dcmaddix:
> Thank you for the code update and improvements! We found out if four fused_moe_lora kernels are using different `block_size_m` the illegal memory access will happen. This is because `block_size_m` is used in `moe_lora_alignment` to generate input blocks. Probably the best idea is to tune the four `fused_moe_lora` kernels together or at least make sure `block_size_m` is the same across four kernels per batch.

Yes, it is important that they use the same `block_size_m` which is also dependent on the `moe_lora_block_align` kernel. Otherwise LGTM, thanks!

caozuoba:
> Thank you for the code update and improvements! We found out if four fused_moe_lora kernels are using different `block_size_m` the illegal memory access will happen. This is because `block_size_m` is used in `moe_lora_alignment` to generate input blocks. Probably the best idea is to tune the four `fused_moe_lora` kernels together or at least make sure `block_size_m` is the same across four kernels per batch.

Thank you for pointing out the issue. I will make the necessary changes and update the relevant code.

benchmarks/kernels/benchmark_lora.py
```diff
@@ -652,7 +652,15 @@ def moe_lora_align_block_size(
             num_tokens_post_pad = torch.empty(
                 (max_loras), dtype=torch.int32, device=topk_ids.device
             )
-
+            # Newer ops.moe_lora_align_block_size signature requires
+            # adapter_enabled and lora_ids. For benchmarking we can use
+            # synthetic values that enable all loras.
+            adapter_enabled = torch.ones(
```
yugong333, line 659: Should be the last element in `adapter_enabled` 0, representing the base model?

caozuoba:
> > > Thank you for the code update and improvements! We found out if four fused_moe_lora kernels are using different `block_size_m` the illegal memory access will happen. This is because `block_size_m` is used in `moe_lora_alignment` to generate input blocks. Probably the best idea is to tune the four `fused_moe_lora` kernels together or at least make sure `block_size_m` is the same across four kernels per batch.
> > 
> > 
> > @yugong333 @dcmaddix Fixed! I've implemented joint tuning to ensure all four fused_moe_lora kernels share the same BLOCK_SIZE_M. This prevents the illegal memory access issue you mentioned. Please review when you have a chance.
> 
> Thank you very much! I also leave a small comment about `adapter_enabled` on the related code. Others LGTM! Thanks again.

@yugong333 Good point, thanks for catching that. I have fixed and submitted the new code.

PR #29893, opened 2025-12-02
[BugFix] Fix assert in `build_for_cudagraph_capture`

LucasWilkinson:
FIX: https://github.com/vllm-project/vllm/issues/29852

Make sure we only call `build_for_cudagraph_capture` if we are capturing cudagraphs not for DP dummy runs; this behavior; NOTE prior to https://github.com/vllm-project/vllm/pull/28579 we used always called `build_for_cudagraph_capture` for all dummy runs including DP dummy batches (this was a bit messy), this shouldn't be the case anymore since we build with padding which can cause asserts in `build_for_cudagraph_capture`

Signed-off-by: Lucas Wilkinson <lwilkins@redhat.com>


LucasWilkinson:
Confirmed fixed by @varun-sundar-rabindranath

varun-sundar-rabindranath:
I was hitting this assert https://github.com/vllm-project/vllm/issues/29852 on DP/EP on the Decode nodes in a PD setup. I no longer hit this assert with this PR.

tlrmchlsmth (approved):

PR #29897, opened 2025-12-02
[Compile] Add env `VLLM_FLOAT32_MATMUL_PRECISION` to fix torch warning `TensorFloat32 tensor cores for float32 matrix multiplication available but not enabled`

yewentao256:
## Purpose

```bash
/home/wentao/pytorch/torch/_inductor/compile_fx.py:321: UserWarning: TensorFloat32 tensor cores for float32 matrix multiplication available but not enabled. Consider setting `torch.set_float32_matmul_precision('high')` for better performance.
warnings.warn
(/home/wentao/pytorch/torch/_inductor/compile_fx.py:321: UserWarning: TensorFloat32 tensor cores for float32 matrix multiplication available but not enabled. Consider setting `torch.set_float32_matmul_precision('high')` for better performance.
warnings.warn
/home/wentao/pytorch/torch/_inductor/compile_fx.py:321: UserWarning: TensorFloat32 tensor cores for float32 matrix multiplication available but not enabled. Consider setting `torch.set_float32_matmul_precision('high')` for better performance.
warnings.warn
/home/wentao/pytorch/torch/_inductor/compile_fx.py:321: UserWarning: TensorFloat32 tensor cores for float32 matrix multiplication available but not enabled. Consider setting `torch.set_float32_matmul_precision('high')` for better performance.
warnings.warn
```

Now we explicitly set `torch.set_float32_matmul_precision('highest')` by default


vllm/v1/worker/gpu_worker.py
```diff
@@ -78,6 +78,9 @@ def __init__(
             is_driver_worker=is_driver_worker,
         )
 
+        # enable TF32 tensor cores when available (Ampere+ GPUs).
+        torch.set_float32_matmul_precision("high")
```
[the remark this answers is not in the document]
yewentao256: Nice suggestion

andy-neuma (approved):
cool

mgoin (changes requested):
This changes the precision for users who might rely on fp32 to actually be running at fp32. Blocking for careful testing and I likely think we should use the highest precision by default to maintain behavior
For instance, what does HF Transformers do?

vllm/envs.py
```diff
@@ -75,6 +75,7 @@
     VLLM_MM_INPUT_CACHE_GIB: int = 4
     VLLM_TARGET_DEVICE: str = "cuda"
     VLLM_MAIN_CUDA_VERSION: str = "12.9"
+    VLLM_FLOAT32_MATMUL_PRECISION: str = "high"
```
mgoin: Type should be literal and default should be highest

yewentao256: Nice catch, fixed

mgoin (approved):
Thanks LGTM!

PR #29900, opened 2025-12-02
Add DGX Spark compatibility to Dockerfile

ericcurtin:
So this Dockerfile can be build and used on DGX Spark

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -146,8 +146,16 @@ RUN --mount=type=cache,target=/root/.cache/uv \
 # can be useful for both `dev` and `test`
 # explicitly set the list to avoid issues with torch 2.2
 # see https://github.com/pytorch/pytorch/pull/123243
-ARG torch_cuda_arch_list='7.0 7.5 8.0 8.9 9.0 10.0 12.0'
-ENV TORCH_CUDA_ARCH_LIST=${torch_cuda_arch_list}
+# Support both x86_64 and aarch64 architectures, with defaults for x86_64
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -146,8 +146,16 @@ RUN --mount=type=cache,target=/root/.cache/uv \
 # can be useful for both `dev` and `test`
 # explicitly set the list to avoid issues with torch 2.2
 # see https://github.com/pytorch/pytorch/pull/123243
-ARG torch_cuda_arch_list='7.0 7.5 8.0 8.9 9.0 10.0 12.0'
-ENV TORCH_CUDA_ARCH_LIST=${torch_cuda_arch_list}
+# Support both x86_64 and aarch64 architectures, with defaults for x86_64
+ARG torch_cuda_arch_list_x86_64='7.0 7.5 8.0 8.9 9.0 10.0 12.0'
+ARG torch_cuda_arch_list_arm64='8.0 8.7 8.9 9.0+PTX 12.1a'
```
pavanimajety, line 151: The list looks incorrect, could you confirm why 10.0 is removed for arm64? What is the source for the list?

ericcurtin:
> Curious why these changes are needed? Internally, we are able to build with the existing Dockerfile for spark. I think this may mess up GB200 builds, I want to make sure we aren't introducing regressions.

I'm closing because there are conflicts now. But doesn't the working version extend this rather than use this Dockerfile? Would like to have this one working accelerated on Spark or move towards that if possible.

ericcurtin:
Any help with this would be appreciated, I'm throttled by the hardware I have available to iterate on this.

PR #29902, opened 2025-12-02
SigLIP example add chat_template

piood:
## Purpose
- Simple chat_template use, `--chat-template path` will auto check and redirect to build in chat-template when the path not found.
- SigLIP online serving need `--chat-template template_basic.jinja` to use correct chat_template, update the example.


examples/online_serving/pooling/openai_chat_embedding_client_for_multimodal.py
```diff
@@ -150,7 +150,8 @@ def run_siglip(client: OpenAI, model: str):
     Start the server using:
 
     vllm serve google/siglip-base-patch16-224 \
-        --runner pooling
+        --runner pooling \
+        --chat-template vllm/transformers_utils/chat_templates/template_basic.jinja
```
DarkLight1337: Can you try if simply passing `template_basic.jinja` works? I remember implementing something that allows auto redirecting to this directory if the chat template is not found via the direct path.

piood: I find only user not use `--chat-template` and model haven't chat-template will auto redirecting to CHAT_TEMPLATES_DIR. So i push new commit https://github.com/vllm-project/vllm/pull/29902/commits/efb79890c7d0e0919958f9b2cf9ecefa8369457d to do that, please review it.

DarkLight1337: I remembered wrong then. Thanks for implementing this!

DarkLight1337 (approved):

PR #29906, opened 2025-12-02
[Bugfix] Free requests to avoid a KV Cache exhaustion during VLLM_NIXL_ABORT_REQUEST_TIMEOUT

hasB4K:
## Purpose
We encountered a "bug" with P/D disaggregation in our pre-prod env a while ago. Here is what we think happened:

Some requests were canceled on Prefill by our router/proxy because of a Timeout (let's say 20s here). Usually a Timeout usually stop/abort a request entirely (which is what we want). But with P/D disagg, you can be unlucky, if you have a lot of requests that will take 20s to prefill if at the same time:

- At the same time:
  - The requests finish the Prefill in 20s
  - The request timeout after 20s! And you trigger an abort
- But (on main) the request is unfree-able because the NIXL KVConnector considers [that requests with FINISHED_LENGTH_CAPPED should be delayed](https://github.com/vllm-project/vllm/blob/5aa9b090407d5fb9b89c05d28fab808623e3070c/vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py#L635) (which is what we want in most cases, but not when we abort a request)
- => this request is now in an un-freeable state for 5 minutes (ie. the default [VLLM_NIXL_ABORT_REQUEST_TIMEOUT](https://github.com/vllm-project/vllm/blob/1bb17ecb396f911beaa26ab0d3926d46154c7155/vllm/envs.py#L1374))
- The requests will be retried on another instance, and create the same issues
- If you are unlucky with a lot of requests similar, you exhaust the available KV Cache, and you start rejecting requests which creates a denial of service for a time :confused: 

This patch, force the removal of a delayed request when an abort happen

(cc @NickLucche)

### DIFF
--- a/vllm/v1/core/sched/scheduler.py
+++ b/vllm/v1/core/sched/scheduler.py
@@ -1326,11 +1326,31 @@ class Scheduler(SchedulerInterface):
         waiting_requests_to_remove = []
         valid_requests = []
 
+        # this is only required only if we have a kv connector
+        should_force_abort = (
+            finished_status == RequestStatus.FINISHED_ABORTED
+            and self.get_kv_connector() is not None
+        )
+        forced_aborted_requests = []
+
         # First pass: collect requests to remove from queues
         for req_id in request_ids:
             request = self.requests.get(req_id)
-            if request is None or request.is_finished():
-                # Invalid request ID.
+            if request is None:
+                continue  # Invalid request ID.
+            elif request.is_finished():
+                if (
+                    should_force_abort
+                    and request.status == RequestStatus.FINISHED_LENGTH_CAPPED
+                ):
+                    # we need to force the status to FINISHED_ABORTED to avoid
+                    # the request being delayed freed. The kv_connector will
+                    # delay the free if it the status is FINISHED_LENGTH_CAPPED
+                    logger.info(
+                        "Request %s is finished but will get forced aborted.",
+                        req_id,
+                    )
+                    forced_aborted_requests.append(request)
                 continue
 
             valid_requests.append(request)
@@ -1347,6 +1367,11 @@ class Scheduler(SchedulerInterface):
 
         # Second pass: set status and free requests
         for request in valid_requests:
+            request.status = finished_status
+            self._free_request(request)
+
+        # Free the requests that are being delayed
+        for request in forced_aborted_requests:
             request.status = finished_status
             self._free_request(request)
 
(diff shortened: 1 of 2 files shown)


vllm/v1/core/sched/scheduler.py
```diff
@@ -1326,11 +1326,31 @@ def finish_requests(
...
         for req_id in request_ids:
             request = self.requests.get(req_id)
-            if request is None or request.is_finished():
-                # Invalid request ID.
+            if request is None:
+                continue  # Invalid request ID.
+            elif request.is_finished():
+                if (
+                    should_force_abort
+                    and request.status == RequestStatus.FINISHED_LENGTH_CAPPED
```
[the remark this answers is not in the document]
NickLucche: talked offline about this, not really a case for P (it outputs just one token)

vllm/v1/core/sched/scheduler.py
```diff
@@ -1326,11 +1326,31 @@ def finish_requests(
         waiting_requests_to_remove = []
         valid_requests = []
 
+        # this is only required only if we have a kv connector
+        should_force_abort = (
+            finished_status == RequestStatus.FINISHED_ABORTED
+            and self.get_kv_connector() is not None
+        )
+        forced_aborted_requests = []
```
NickLucche, line 1334: have you tried to set the timeout to a much lower value? if I understand correctly, this fix is freeing requests asap, which should be similar. Not sure if this is optimal in the case D is hogged as well

NickLucche (commented):
Looking into more structural changes right now. Related to "passive" https://github.com/vllm-project/vllm/issues/26400 or even "active" solution such as https://github.com/vllm-project/vllm/pull/26635 .

The former in particular would explain why this case is happening more frequently than anticipated.

vllm/v1/core/sched/scheduler.py
njhill: Another todo we have is to propagate the deadline from P to D to make smaller timeouts "safe". Otherwise D could pull invalid blocks in this case. This should be a pretty simple change

njhill:
Thanks for reporting @hasB4K. I think you are hitting the issue described in https://github.com/vllm-project/vllm/issues/26400, that if request cancellations happen during a forward pass, any requests that complete normally following that forward pass won't be handled as aborts (which will be all requests disagg prefill case, apart from chunks).

I can see the change in this PR is attempting to address this but I don't think it will actually work since the problem is that the scheduler `finish_requests` method will be called too late (only after the model output from the in-progress step has already been processed). I'm curious whether you were able to verify that this change actually fixes the issue (would be very surprised if it does).

I have in mind what's needed to fix this, will aim to open a PR for it today.

njhill:
Here you go, not yet tested :)  https://github.com/vllm-project/vllm/pull/29987

PR #29912, opened 2025-12-02
[Cleanup] Refactor profiling env vars into a CLI config

benchislett:
## Purpose

This PR refactors all profiling control environment variables into a CLI configuration object `profiler_config`. The previous environment variables are retained as deprecated flags which are parsed explicitly in the ProfilerConfig logic, and as such they all have their types changed to `str | None`.

### Note about release schedule

Because this PR deprecates a lot of the existing torch profiling env vars, I do not view it as a critical inclusion for the upcoming 0.12.0 release. I would not recommend rushing to include this before it has been fully reviewed.

## Testing

Tested the examples locally, as well as a few local runs to make sure the new flags work and the deprecated flags are still respected.

Updated the profiling iteration-counting logic test, which passes.

CI E2E should catch any stragglers regarding startup issues.

No plan to test cpu/xpu/tpu changes beyond CI coverage (diff is minimal but would be nice to have a quick review from codeowners there)


vllm/envs.py
hmellor, line 1: Could you group the deprecated vars together so that they'll be easier to delete in future?

vllm/config/profiler.py
```diff
@@ -0,0 +1,181 @@
...
+
+ProfilerKind = Literal["none", "torch", "cuda"]
+
+
+@config
+@dataclass
+class ProfilerConfig:
+    """Dataclass which contains profiler config for the engine."""
+
+    profiler: ProfilerKind = "none"
```
hmellor: How do we feel about this instead of a `"none"` string?

```suggestion
    profiler: ProfilerKind | None = None
```

benchmarks/benchmark_serving_structured_output.py
```diff
@@ -963,8 +964,7 @@ def create_argument_parser():
     parser.add_argument(
         "--profile",
         action="store_true",
-        help="Use Torch Profiler. The endpoint must be launched with "
-        "VLLM_TORCH_PROFILER_DIR to enable profiler.",
+        help="Use vLLM Profiling. profiler_config must be provided on the server side.",
```
hmellor:
```suggestion
        help="Use vLLM Profiling. --profiler-config must be provided on the server side.",
```

benchmarks/benchmark_serving_structured_output.py
```diff
@@ -45,6 +44,8 @@
 from tqdm.asyncio import tqdm
 from transformers import PreTrainedTokenizerBase
 
+import datasets
+
```
hmellor: Why was this moved? I'd expect isort to complain

docs/contributing/profiling.md
```diff
@@ -5,16 +5,15 @@
 
 ## Profile with PyTorch Profiler
 
-We support tracing vLLM workers using the `torch.profiler` module. You can enable tracing by setting the `VLLM_TORCH_PROFILER_DIR` environment variable to the directory where you want to save the traces: `VLLM_TORCH_PROFILER_DIR=/mnt/traces/`. Additionally, you can control the profiling content by specifying the following environment variables:
+We support tracing vLLM workers using the `torch.profiler` module. You can enable the torch profiler by setting `profiler_config`
```
hmellor:
```suggestion
We support tracing vLLM workers using the `torch.profiler` module. You can enable the torch profiler by setting `--profiler-config`
```

vllm/engine/arg_utils.py
```diff
@@ -732,6 +735,12 @@ def add_cli_args(parser: FlexibleArgumentParser) -> FlexibleArgumentParser:
             **structured_outputs_kwargs["reasoning_parser_plugin"],
         )
 
+        # Profiler arguments
+        parser.add_argument_group(
+            title="ProfilerConfig",
+            description=ProfilerConfig.__doc__,
+        )
+
```
hmellor: I'd remove this to avoid confusion since we're not adding individual profiler arguments to the CLI outside of `--profiler-config`

```suggestion
```

vllm/entrypoints/openai/api_server.py
```diff
@@ -513,6 +513,31 @@ async def get_tokenizer_info(raw_request: Request):
             )
 
 
+def maybe_register_profiler_endpoints(args):
+    """Conditionally add profiler start/stop endpoints if enabled."""
+    profiler_config = args.profiler_config
+    assert isinstance(profiler_config, ProfilerConfig)
+    if profiler_config.profiler != "none":
```
hmellor: To do with the other suggested change

```suggestion
    if profiler_config.profiler is not None:
```

vllm/entrypoints/openai/api_server.py
```diff
@@ -513,6 +513,31 @@ async def get_tokenizer_info(raw_request: Request):
...
+        logger.warning_once(
+            f"Profiler with mode '{profiler_config.profiler}' is enabled in the "
+            "API server. This should ONLY be used for local development!"
+        )
```
hmellor: Logging with f-string

```suggestion
        logger.warning_once(
            "Profiler with mode '%s' is enabled in the "
            "API server. This should ONLY be used for local development!", 
            profiler_config.profiler
        )
```

vllm/benchmarks/serve.py
```diff
@@ -1085,8 +1085,7 @@ def add_cli_args(parser: argparse.ArgumentParser):
     parser.add_argument(
         "--profile",
         action="store_true",
-        help="Use Torch Profiler. The endpoint must be launched with "
-        "VLLM_TORCH_PROFILER_DIR to enable profiler.",
+        help="Use vLLM Profiling. profiler_config must be provided on the server side.",
```
hmellor: Because this a tip for the server?

```suggestion
        help="Use vLLM Profiling. --profiler-config must be provided on the server side.",
```

vllm/benchmarks/throughput.py
```diff
@@ -651,8 +651,7 @@ def add_cli_args(parser: argparse.ArgumentParser):
         "--profile",
         action="store_true",
         default=False,
-        help="Use Torch Profiler. The env variable "
-        "VLLM_TORCH_PROFILER_DIR must be set to enable profiler.",
+        help="Use vLLM Profiling. profiler_config must be provided on the server side.",
```
hmellor: Because this a tip for the server?

```suggestion
        help="Use vLLM Profiling. --profiler-config must be provided on the server side.",
```

vllm/config/profiler.py
```diff
@@ -0,0 +1,181 @@
...
+
+        if self.profiler not in ("none", "torch", "cuda"):
+            raise ValueError(
+                f"Invalid profiler: {self.profiler} "
+                f"(choose from 'none', 'torch', 'cuda')"
+            )
+        if self.delay_iterations < 0:
+            raise ValueError("Profiler delay_iterations must be >= 0")
+        if self.max_iterations < 0:
+            raise ValueError("Profiler max_iterations must be >= 0")
```
hmellor: - Pydantic will validate `profiler` for us
- For the other two you can default them to `Field(default=0, ge=0)`

```suggestion
```

hmellor (commented):
Looks really nice, left a few comments about small things

vllm/config/profiler.py
```diff
@@ -0,0 +1,181 @@
...
+                "Environment variable VLLM_TORCH_CUDA_PROFILE is deprecated. "
+                "Please use profiler_config.profiler instead."
+            )
+        elif envs.VLLM_TORCH_PROFILER_DIR is not None:
+            self.profiler = "torch"
+            self.torch_profiler_dir = envs.VLLM_TORCH_PROFILER_DIR
+            logger.warning_once(
+                "Environment variable VLLM_TORCH_PROFILER_DIR is deprecated. "
+                "Please use profiler_config.torch_profiler_dir instead."
+            )
```
hmellor, line 104: Oh and in all the warning messages please state in which version whey'll be removed. You could take inspiration from https://github.com/vllm-project/vllm/pull/26315/files#diff-1b15c9b9ea534e3e02c27d1138d6fdc176eeedeab43244549f63ac380e62613cR73-R114 to reduce duplicated code in these warnings

Since you plan to merge this after 0.12.0, the removal version should be stated as 0.14.0 (so that the warnings are present for all of 0.13)

benchmarks/benchmark_serving_structured_output.py
benchislett: I think I've actually been running into this for a while, is it possible that isort just hasn't been run on this file for a while? Or maybe my local version is out of date. Isort is what moved this for me

vllm/entrypoints/openai/api_server.py
benchislett: just curious, what's wrong with logging with f strings?

vllm/benchmarks/serve.py
benchislett: oops, copy/paste issue

vllm/entrypoints/openai/api_server.py
hmellor: It's a performance thing.

With f-strings the substitution will always be evaluated, even if the statement is never logged (i.e. hidden behind `if` or log level is set to >warning in this case)

See https://docs.astral.sh/ruff/rules/logging-f-string/ for a full explanation

benchmarks/benchmark_serving_structured_output.py
hmellor: I don't think that's it, `pre-commit` is run on all files in CI.

You can see in the logs that it's actually trying to put it back https://github.com/vllm-project/vllm/actions/runs/19870981081/job/56946321252?pr=29912.

There must be something wrong with your local setup. The location it's putting it is where `isort` would put local imports. Do you have an uncommitted file/module called `datasets` anywhere?

vllm/config/profiler.py
```diff
@@ -0,0 +1,181 @@
...
+        factors: list[Any] = []
+        hash_str = safe_hash(str(factors).encode(), usedforsecurity=False).hexdigest()
+        return hash_str
+
+    @model_validator(mode="after")
+    def _validate_profiler_config(self) -> Self:
+        if envs.VLLM_TORCH_CUDA_PROFILE is not None:
+            self.profiler = "cuda" if envs.VLLM_TORCH_CUDA_PROFILE == "1" else "none"
+            logger.warning_once(
+                "Environment variable VLLM_TORCH_CUDA_PROFILE is deprecated. "
```
markmc: Consider unsetting these env variables once you've converted them to config? That way some stray usage of them doesn't come as a surprise when we remove them?

markmc (commented):
Beyond existing comments, this looks really excellent to me! Just one suggestion inline

vllm/config/profiler.py
benchislett: That's a good idea. Can you point me to a syntax example here? Will `del envs.XYZ` do what I want?

benchmarks/benchmark_serving_structured_output.py
benchislett: In fact I did! Great catch

vllm/config/profiler.py
benchislett: Done

vllm/envs.py
```diff
@@ -88,20 +88,22 @@
...
-    VLLM_TORCH_PROFILER_DISABLE_ASYNC_LLM: bool = False
+    VLLM_TORCH_PROFILER_RECORD_SHAPES: str | None = None
+    VLLM_TORCH_PROFILER_WITH_PROFILE_MEMORY: str | None = None
+    VLLM_TORCH_PROFILER_DISABLE_ASYNC_LLM: str | None = None
+    VLLM_TORCH_PROFILER_WITH_STACK: str | None = None
+    VLLM_TORCH_PROFILER_WITH_FLOPS: str | None = None
+    VLLM_TORCH_PROFILER_USE_GZIP: str | None = None
+    VLLM_TORCH_PROFILER_DUMP_CUDA_TIME_TOTAL: str | None = None
+    VLLM_PROFILER_DELAY_ITERS: str | None = None
+    VLLM_PROFILER_MAX_ITERS: str | None = None
```
hmellor, line 103:
```suggestion
    VLLM_PROFILER_MAX_ITERS: str | None = None
    # End of deprecated env variables for profiling
```

vllm/config/profiler.py
```diff
@@ -0,0 +1,198 @@
...
+
+@config
+@dataclass
+class ProfilerConfig:
+    """Dataclass which contains profiler config for the engine."""
+
+    profiler: ProfilerKind | None = None
+    """Which profiler to use. Defaults to None. Options are:
+    - 'torch': Use PyTorch profiler.
+    - 'cuda': Use CUDA profiler."""
```
hmellor: Helps out the `--help` rendering

```suggestion

    - 'torch': Use PyTorch profiler.\n
    - 'cuda': Use CUDA profiler."""
```

hmellor (approved):
LGTM!

Just a couple of nits and @markmc's comment

vllm/config/profiler.py
markmc: Something like this? (in `vllm/envs.py`)

```
def unset_envs(names: list[str]) -> None:
    """                                                                                                                                                                                      
    Unset environment variables, removing them from both os.environ and                                                                                                                      
    the cached values if caching is enabled.                                                                                                                                                 
    """
    for name in names:
        if name not in environment_variables:
            raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
        os.environ.pop(name, None)

    # functools.cache doesn't support invalidating individual entries,                                                                                                                       
    # so we clear the entire cache                                                                                                                                                           
    if hasattr(__getattr__, 'cache_clear'):
        __getattr__.cache_clear()
```

It might be even more useful to also remove the entry from `environment_variables` so any lingering usage triggers an exception?

benchislett: @markmc I have a prototype here but I am not sure it will work well, since it appears the config is initialized twice when starting up the engine, as pydantic parses the args. I'm not sure where I can trigger this clearing-out of env vars.

hmellor (approved):
LGTM once the last suggestion is added.

You mentioned you wanted second opinions so I won't try and merge immediately

vllm/config/profiler.py
markmc: Fair enough :+1:

jikunshang:
LGTM to xpu part. cc @zhangruoxu for awareness.

PR #29919, opened 2025-12-02
[Misc] Add a script to benchmark compilation time

desertfire:
Summary: Adds a benchmarking script for measuring torch.compile's cold and warm compilation times. For cold compilation, the script redirects both vLLM and torch.compile caching to a temporary directory on each iteration. For warm compilation, it uses the default caching directories.

The script sets VLLM_ENABLE_V1_MULTIPROCESSING=0 and runs each iteration in a separate subprocess to track the compilation time.

Usage: `vllm bench compilation --model YOUR_MODEL`


ProExpertProg (commented):
Thanks for doing this! A couple of thoughts:
- Let's rename the script from `bench compilation` to `bench startup`: we're measuring (and reporting) the total startup anyway. We also cache model info and we might cache other things in the future, so cold/warm distinction will apply to more than compilation as well.
- It would be nice if we can at least add some form of "total startup" reporting for all benchmarking scripts (can be warm-only). That's because devs often only report results from a single benchmark and I want to start tracking startup in PRs, but devs are less likely to run a separate script just for startup.
- Could you also add a `vllm bench sweep startup` command? I want to move more people over to the new `vllm bench sweep serve` command which sweeps over combinations of serving parameters and request rates, and reusing the  same `--serve-params` json file for startup comparison between configurations would be really helpful (this can be a follow-up as well).

ProExpertProg (approved):
Will try to review the follow-ups once I'm back! My definition of total startup time would be the time it takes for `vllm serve` to be ready to accept incoming requests.

vllm/benchmarks/startup.py
```diff
@@ -0,0 +1,326 @@
...
+
+        # Extract compilation time if available
+        compilation_time = 0.0
+        if hasattr(llm.llm_engine, "vllm_config"):
+            vllm_config = llm.llm_engine.vllm_config
+            if (
+                hasattr(vllm_config, "compilation_config")
+                and vllm_config.compilation_config is not None
+            ):
+                compilation_time = vllm_config.compilation_config.compilation_time
```
zou3519, line 85: related: I'm really not completely sure this time is correct. But it is the best thing we have. Also, how difficult would it be to have breakdowns by component?

(nothing needs to happen in this PR)

zou3519: See https://github.com/vllm-project/vllm/issues/27815

zou3519 (approved):

desertfire:
> Getting error, we should add this script to CI:

What was the commandline you used to trigger the error? I don't run into any problem locally.

ProExpertProg:
> > Getting error, we should add this script to CI:
> 
> What was the commandline you used to trigger the error? I don't run into any problem locally.

I believe it was fixed in #31348.

PR #29922, opened 2025-12-02
[Docs] Discuss api key limitations in security guide

russellb:
The use of `--api-key` only applies to the OpenAI-comaptible API
endpoints under `v1/`. It does not apply to other endpoints exposed
via the same HTTP server in vLLM. This doc update discusses why it is
important not to rely on `--api-key` exclusively for securing access
to vLLM.

Signed-off-by: Russell Bryant <rbryant@redhat.com>


aarnphm (approved):

chaunceyjiang (approved):
Thanks~

DarkLight1337 (approved):

PR #29923, opened 2025-12-02
[BUG] Make VLLM_COMPILE_CACHE_SAVE_FORMAT override CompilationConfig

hao-aaron:
## Purpose

Addressing https://github.com/vllm-project/vllm/pull/27616/files#r2488035326
and https://github.com/vllm-project/vllm/pull/27616#pullrequestreview-3391158198

### DIFF
--- a/vllm/compilation/backends.py
+++ b/vllm/compilation/backends.py
@@ -59,7 +59,8 @@ def make_compiler(compilation_config: CompilationConfig) -> CompilerInterface:
         ):
             logger.debug("Using InductorStandaloneAdaptor")
             return InductorStandaloneAdaptor(
-                compilation_config.compile_cache_save_format
+                envs.VLLM_COMPILE_CACHE_SAVE_FORMAT
+                or compilation_config.compile_cache_save_format
             )
         else:
             logger.debug("Using InductorAdaptor")
--- a/vllm/compilation/backends.py
+++ b/vllm/compilation/backends.py
@@ -48,7 +48,9 @@ from .pass_manager import PostGradPassManager
 logger = init_logger(__name__)
 
 
-def make_compiler(compilation_config: CompilationConfig) -> CompilerInterface:
+def make_compiler(
+    compilation_config: CompilationConfig, dp_rank: int = 0
+) -> CompilerInterface:
     if compilation_config.backend == "inductor":
         # Use standalone compile only if requested, version is new enough,
         # and the symbol actually exists in this PyTorch build.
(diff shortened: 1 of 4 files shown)


vllm/compilation/backends.py
```diff
@@ -59,7 +59,8 @@ def make_compiler(compilation_config: CompilationConfig) -> CompilerInterface:
         ):
             logger.debug("Using InductorStandaloneAdaptor")
             return InductorStandaloneAdaptor(
-                compilation_config.compile_cache_save_format
+                envs.VLLM_COMPILE_CACHE_SAVE_FORMAT
```
ProExpertProg: Can you move the override to `CompilationConfig.__post_init__`? That way when we print the full config the format that will end up used can be printed

zou3519 (commented):
@ProExpertProg @ahao-anyscale if I'm underestanding the proposal correctly, we want the envvar to override a command line flag?

If so, command line flags generally take precedence over envvars in many open source projects:

docker: https://docs.docker.com/reference/cli/docker/ "Command line options override environment variables and environment variables override properties you specify in a config.json file."

aws CLI: https://docs.aws.amazon.com/cli/v1/userguide/cli-chap-configure.html

git: https://git-scm.com/docs/gitcli#_options_trump_configuration_and_environment

For other envvars, has vLLM preferred envvar to config option? Is it too late for us to align our behavior with other projects?

vllm/envs.py
```diff
@@ -235,7 +235,7 @@
     VLLM_GC_DEBUG: str = ""
     VLLM_DISABLE_SHARED_EXPERTS_STREAM: bool = False
     VLLM_SHARED_EXPERTS_STREAM_TOKEN_THRESHOLD: int = 256
-    VLLM_COMPILE_CACHE_SAVE_FORMAT: Literal["binary", "unpacked"] = "binary"
+    VLLM_COMPILE_CACHE_SAVE_FORMAT: Literal["binary", "unpacked"] = ""
```
ProExpertProg, line 238: Let's keep this default so it's clearer to readers? Or add a comment that "binary" is the default if neither env nor cli are set?

vllm/compilation/backends.py
```diff
@@ -60,6 +62,8 @@ def make_compiler(compilation_config: CompilationConfig) -> CompilerInterface:
             logger.debug("Using InductorStandaloneAdaptor")
             return InductorStandaloneAdaptor(
                 compilation_config.compile_cache_save_format
+                if dp_rank == 0
```
ProExpertProg, line 65: Can we instead update this default during config init when rank is known?

ProExpertProg (commented):
I was not aware that conventionally, CLI args take precedence over env vars. So I think let's keep the precedence for env var < CLI, but let's still do the thing where DP rank 0 is unpacked and others are binary

PR #25986, opened 2025-09-30
[P/D] KVConnector for decode benchmarking

tlrmchlsmth:
KVConnector used for performance testing of the decode instance of vLLM, emulating the decode instance in a P/D disaggregated setting. This is advantageous in a resource-limited environment, if we want to use all available nodes to benchmark a WideEP decode.

On current main, the best we can do in this case is benchmark the decoder by setting the input sequence length (ISL) to 1. 
This PR lets us measure the effect of larger ISLs on the decoder's performance.

Testing with:
```
vllm serve Qwen/Qwen3-30B-A3B-FP8 --async-scheduling --port 8192 --kv-transfer-config '{"kv_connector":"DecodeBenchConnector", "kv_role":"kv_both"}'

vllm bench serve --base-url http://127.0.0.1:8192 --model Qwen/Qwen3-30B-A3B-FP8 --dataset-name random --random-input-len 40000 --random-output-len 100  --max-concurrency 10 --request-rate 4096 --seed $(date +%M%H%M%S) --num-prompts 10 --ignore-eos
```

I see:
```
============ Serving Benchmark Result ============
Successful requests:                     10
Maximum request concurrency:             10
Request rate configured (RPS):           4096.00
Benchmark duration (s):                  3.72
Total input tokens:                      400000
Total generated tokens:                  1000
Request throughput (req/s):              2.69
Output token throughput (tok/s):         268.80
Peak output token throughput (tok/s):    351.00
Peak concurrent requests:                10.00
Total Token throughput (tok/s):          107789.89
---------------Time to First Token----------------
Mean TTFT (ms):                          635.62
Median TTFT (ms):                        459.53
P99 TTFT (ms):                           2354.87
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          24.41
Median TPOT (ms):                        26.00
P99 TPOT (ms):                           26.23
---------------Inter-token Latency----------------
Mean ITL (ms):                           24.41
Median ITL (ms):                         25.98
P99 ITL (ms):                            36.63
==================================================
```
vs without the connector:
```
============ Serving Benchmark Result ============
Successful requests:                     10
Maximum request concurrency:             10
Request rate configured (RPS):           4096.00
Benchmark duration (s):                  18.71
Total input tokens:                      400000
Total generated tokens:                  1000
Request throughput (req/s):              0.53
Output token throughput (tok/s):         53.44
Peak output token throughput (tok/s):    284.00
Peak concurrent requests:                10.00
[...]

PR #29934, opened 2025-12-03
[bench] Support common prefix len config (for decode-only bench)

minosfuture:
## Purpose

This is to support decode-only bench for a quick understanding of decoding node performance. A convenient way to simulate D in P/D disagg.  NOTE: users should consider the gap between kv cache usage from the real case. Consider using https://github.com/vllm-project/vllm/pull/25986 for more realistic decode node simulation.

## Test Plan

`vllm bench serve --model Qwen/Qwen3-0.6B --port 8100  --dataset-name random --ignore-eos  --num-prompts 2048 --request-rate inf  --random-input-len 2  --random-output-len 1024 --max-concurrency 2048 --trust_remote_code --seed $RANDOM --ready-check-timeout-sec 0 --common-prefix-len 2048`

## Test Result

Performance measured with this is close to performance in P/D setup (using NixlConnector).

---
<details>

</details>


ywang96 (approved):

vllm/benchmarks/serve.py
```diff
@@ -1221,6 +1221,12 @@ def add_cli_args(parser: argparse.ArgumentParser):
         help="Repetition penalty sampling parameter. Only has effect on "
         "openai-compatible backends.",
     )
+    sampling_group.add_argument(
```
heheda12345, line 1224: Is this `--random-prefix-len`?

If it is the same, can you reuse it? If it is not the same, can you move it to https://github.com/vllm-project/vllm/blob/9c32df6101b24040c824dee5b0e5543a94dc6f45/vllm/benchmarks/datasets.py#L1317

This PR breaks this command
```
python3 examples/offline_inference/spec_decode.py --test
```

minosfuture: sorry that I missed this message. Lemme fix this.

PR #29941, opened 2025-12-03
[offloader] v2: Hide weight onloading latency via prefetching

minosfuture:
## Purpose

This PR adds CPU weight offloader that hides weight onloading latency by prefetching weights.  This saves the performance cost of zero-copy UVA access. This technique was first developed in SGLang for GB200: https://lmsys.org/blog/2025-09-25-gb200-part-2/, and now adapted to support torch.compile and CUDA graph within vLLM in this PR.

Also refactors the offloading to be extensible. 

Demonstrated in the trace:
* H2D is for prefetching next offloaded weights

image

## Test Plan

* deepseek fp4 in 2 GPUs with offloading
* deepseek fp4 in 4 GPUs without offloading (doesn't fit in 2 GPUs without offloading)

Example serving recipe for GB200:

```
HF_HUB_OFFLINE=1 NVIDIA_GDRCOPY=1 NVSHMEM_IB_ENABLE_IBGDA=1 VLLM_SKIP_P2P_CHECK=1 \
NCCL_CUMEM_ENABLE=1 NCCL_MNNVL_ENABLE=1 NCCL_NVLS_ENABLE=1 \
VLLM_RANDOMIZE_DP_DUMMY_INPUTS=1 VLLM_USE_FLASHINFER_MOE_FP4=1 \
VLLM_FLASHINFER_MOE_BACKEND=latency VLLM_ENABLE_FUSED_MOE_ACTIVATION_CHUNKING=0 \
VLLM_USE_NCCL_SYMM_MEM=1 VLLM_ENABLE_MOE_DP_CHUNK=0 numactl --cpunodebind=0 \
--membind=0 vllm serve \
nvidia/DeepSeek-R1-0528-NVFP4-v2 \
--attention-config.backend FLASHINFER_MLA \
--attention-config.use_trtllm_ragged_deepseek_prefill=true --kv-cache-dtype fp8 \
--enable-expert-parallel --data-parallel-rpc-port 13345 --max-model-len 2148 \
--disable-uvicorn-access-log --port 8000 --trust-remote-code --async-scheduling \
--disable_custom_all_reduce --disable_nccl_for_dp_synchronization \
--no-enable-prefix-caching --all2all-backend allgather_reducescatter \
--gpu-memory-utilization 0.85 --max-num-batched-tokens 65536 --max-num-seqs 1024 \
--swap-space 16 --data-parallel-size 2 --offload-group-size 2 \
--offload-num-in-group 1 --offload-prefetch-step 1 --offload-params w13_weight \
w2_weight --api-server-count 1 --profiler-config '{"profiler": "torch",
"torch_profiler_dir": "/tmp/traces"}' 2>&1 | tee /tmp/dsr1.log
```

### QPS per GPU 18.04 (2 GPUs)

this PR `--offload-group-size 2 --offload-num-in-group 1 --offload-prefetch-step 1`

```
Successful requests:                     1024
Maximum request concurrency:             1024
Benchmark duration (s):                  56.78
Total input tokens:                      2096128
Total generated tokens:                  1024
Request throughput (req/s):              18.04
Output token throughput (tok/s):         18.04
Peak output token throughput (tok/s):    63.00
Peak concurrent requests:                1024.00
Total Token throughput (tok/s):          36937.17
Mean TTFT (ms):                          30965.61
Median TTFT (ms):                        30190.67
P99 TTFT (ms):                           56685.92
Mean TPOT (ms):                          0.00
Median TPOT (ms):                        0.00
P99 TPOT (ms):                           0.00
Mean ITL (ms):                           0.00
Median ITL (ms):                         0.00
P99 ITL (ms):                            0.00
```

### QPS per GPU 16.9 (4 GPUs)

without `--offload-group-size 2 --offload-num-in-group 1 --offload-prefetch-step 1`

```
Successful requests:                     1024
Maximum request concurrency:             1024
Benchmark duration (s):                  30.28
Total input tokens:                      2096128
Total generated tokens:                  1024
Request throughput (req/s):              33.82
Output token throughput (tok/s):         33.82
Peak output token throughput (tok/s):    64.00
Peak concurrent requests:                1024.00
Total Token throughput (tok/s):          69259.31
Mean TTFT (ms):                          17335.01
Median TTFT (ms):                        17177.41
P99 TTFT (ms):                           30175.24
Mean TPOT (ms):                          0.00
Median TPOT (ms):                        0.00
P99 TPOT (ms):                           0.00
Mean ITL (ms):                           0.00
Median ITL (ms):                         0.00
P99 ITL (ms):                            0.00
```

### Accuracy:

local-completions (model=nvidia/DeepSeek-R1-0528-FP4-v2,base_url=http://127.0.0.1:8000/v1/completions,num_concurrent=32), gen_kwargs: (None), limit: 100.0, num_fewshot: None, batch_size: 1
|gsm8k|      3|flexible-extract|     5|exact_match|↑  | 0.98|±  |0.0141|
|     |       |strict-match    |     5|exact_match|↑  | 0.98|±  |0.0141|

---
<details>

</details>


vllm/config/cache.py
```diff
@@ -89,6 +89,16 @@ class CacheConfig:
...
+    offload_group_size: int = Field(default=0, ge=0)
+    """Advanced CPU offloading: Group every N layers together. Offload last
+    `offload_num_in_group` layers of each group. Default is 0 (disabled).
+    Example: group_size=8, num_in_group=2 offloads layers 6,7,14,15,22,23,...
+    """
+    offload_num_in_group: int = Field(default=1, ge=1)
+    """Advanced CPU offloading: Number of layers to offload per group. Default is 1."""
+    offload_prefetch_step: int = Field(default=1, ge=0)
+    """Advanced CPU offloading: Number of layers to prefetch ahead. Higher values hide
+    more latency but use more GPU memory. Default is 1."""
```
hmellor: This is not configuration for the KV cache?

minosfuture: had the same thought, but followed --cpu-offload-gb here.  Let me find a better place for them.

ywang96 (commented):
IIUC this is mostly a RL feature, correct? Maybe @youkaichao can take a look?

elvircrn:
Why is TPOT and ITL 0 in the benchmark result?

vllm/model_executor/offloader/uva.py
```diff
@@ -0,0 +1,110 @@
...
+            return module
+
+        if self.cpu_offload_bytes >= self.cpu_offload_max_bytes:
+            return module
+
+        pin_memory = is_pin_memory_available()
+        uva_available = is_uva_available()
+
+        assert uva_available, "V1 CPU offloading requires uva (pin memory) support"
+        uva_offloading = True
```
Liccol: uva_offloading seems to be always True? making the conditional checks redundant

minosfuture: Good catch. Let me remove this.

vllm/model_executor/models/deepseek_v2.py
```diff
@@ -1275,6 +1275,31 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = ""):
                 vllm_config, prefix, topk_indices_buffer=topk_indices_buffer
             ),
             prefix=f"{prefix}.layers",
+            offloader_kwargs=dict(
+                # Extract the MLP submodule - for MoE layers, go deeper to the experts
+                submodule_accessor=lambda layer: (
+                    layer.mlp.experts
+                    if isinstance(layer.mlp, DeepseekV2MoE)
+                    else layer.mlp
+                ),
```
mgoin: Maybe we could generalize this, at least for the moe, by searching for an attr that isinstance of FusedMoE

vllm/model_executor/models/deepseek_v2.py
```diff
@@ -1275,6 +1275,31 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = ""):
...
+                    [
+                        # Core MoE expert weights
+                        "w13_weight",
+                        "w2_weight",
+                        # NVFP4 quantization scales (if present)
+                        *(
+                            ["w13_blockscale_swizzled", "w2_blockscale_swizzled"]
+                            if hasattr(module, "w13_blockscale_swizzled")
+                            else []
+                        ),
```
mgoin: Is there harm in offloading "too many" params? I feel like I'd prefer a general "w13" or "w2" match rather than specializing on specific quant methods

vllm/config/cache.py
```diff
@@ -89,6 +89,16 @@ class CacheConfig:
     Note that this requires fast CPU-GPU interconnect, as part of the model is
     loaded from CPU memory to GPU memory on the fly in each model forward pass.
     """
+    offload_group_size: int = Field(default=0, ge=0)
+    """Advanced CPU offloading: Group every N layers together. Offload last
+    `offload_num_in_group` layers of each group. Default is 0 (disabled).
+    Example: group_size=8, num_in_group=2 offloads layers 6,7,14,15,22,23,...
+    """
+    offload_num_in_group: int = Field(default=1, ge=1)
+    """Advanced CPU offloading: Number of layers to offload per group. Default is 1."""
```
mgoin: I didn't see any validation that that offload_num_in_group <= offload_group_size

tests/basic_correctness/test_v2_offload.py
```diff
@@ -0,0 +1,31 @@
...
+    """Test V2 CPU offloading with DeepSeek-V2-Lite.
+
+    Compares outputs between:
+    1. Baseline (no offloading)
+    2. V2 offloading (group_size=8, num_in_group=2, prefetch_step=1)
+
+    This tests the advanced offloading with prefetching on a MoE model.
+    """
+    compare_two_settings(
+        "deepseek-ai/DeepSeek-V2-Lite",
```
mgoin: Could use an fp8 model here to make it faster like RedHatAI/DeepSeek-Coder-V2-Lite-Instruct-FP8

vllm/model_executor/offloader/v2.py
```diff
@@ -0,0 +1,273 @@
...
+            self._device_tensors = {
+                name: offloader.create_device_tensor()
+                for name, offloader in self._param_offloaders.items()
+            }
+            self._load_event = torch.cuda.Event()
+            self._load_event.record()
+
+    def offload(self):
+        """Free device tensors (offload from GPU memory)."""
+        self._device_tensors = None
```
mgoin: Should we be calling torch.cuda.empty_cache() occasionally here as well? Not sure if this is sufficient to free

vllm/model_executor/offloader/v2.py
```diff
@@ -0,0 +1,273 @@
...
+    def post_init(self):
+        """No-op: offloading already done in __init__."""
+        pass
+
+    def create_device_tensor(self) -> torch.Tensor:
+        """Load from CPU to GPU (async if pinned).
+
+        Returns a CUDA copy of the parameter (which has CPU data).
+        """
+        return self._param.to("cuda", non_blocking=True)
```
mgoin: Will this fail on multi-gpu? I feel we should store the original device

tests/basic_correctness/test_v2_offload.py
minosfuture: serving "RedHatAI/DeepSeek-Coder-V2-Lite-Instruct-FP8" failed at flashinfer autotuning stage on GB200. 😿

vllm/config/cache.py
minosfuture: thanks. added.

vllm/config/cache.py
minosfuture: moved to a new OffloadConfig

vllm/model_executor/models/deepseek_v2.py
minosfuture: To avoid overheads, onloading latency needs to be hidden by overlap with computation.  This requires careful check on latencies of H2D ops and layer latency.  We probably shouldn't blindly offload too many weights. I removed quantization scales that are not needed.

vllm/model_executor/offloader/v2.py
minosfuture: From experiment, it is freed.  It's actually also good if this is only returned to torch mem pool, so next alloc is fast.

vllm/model_executor/offloader/v2.py
minosfuture: I thought we have only one GPU per worker process?

vllm/model_executor/offloader/uva.py
minosfuture: cleaned up

vllm/model_executor/models/deepseek_v2.py
minosfuture: done

vllm/model_executor/models/deepseek_v2.py
```diff
@@ -1274,6 +1277,21 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = ""):
...
+                    [
+                        # Core MoE expert weights
+                        "w13_weight",
+                        "w2_weight",
+                    ]
+                    # Only offload from MoE experts (SharedFusedMoE/FusedMoE)
+                    if hasattr(module, "w13_weight")
+                    else []
+                ),
+            ),
```
youkaichao: do we need to do it for every model? this looks intrusive 🤔

youkaichao (commented):
mainly thinking about how to make it work for `torch.compile` + cudagraph.

I think it's possible to have a static buffer `workspace` (rough size `prefetch_layers * params_offloaded_per_layer)`, and then after offloading:

```python
cpu_tensor = w
w = workspace[start:end].reshape(xxx)
w.cpu_tensor = cpu_tensor
```

then we insert one pytorch custom op `prefetch(layer_name)` in every layer that contains offloaded parameters, which copies cpu weights to the slice of the workspace buffer. it seems to me this can be compatible with `torch.compile`. However, it might be dangerous that `torch.compile` might rearrange the op while it actually inplace mutates many weight tensors. cc @zou3519

youkaichao:
> However, it might be dangerous that torch.compile might rearrange the op while it actually inplace mutates many weight tensors.

traditional compilers have the concept of "volatile" that prevents the compiler doing certain optimizations. Not sure if `torch.compile` has similar tools.

vllm/model_executor/models/deepseek_v2.py
minosfuture: This is highly targeted optimization as it requires carefully overlapping weight onloading (memcpy) and computation.  So it is impacted by model weight size, CPU<>GPU bandwidth, computation latency (per batch size).  So it should be configured at least model level.

Or, maybe this should be configurable via cli args. WDYT?

vllm/config/cache.py
hmellor, line 1: When deprecating parameters you must specify which version they will be removed https://docs.vllm.ai/en/latest/contributing/deprecation_policy/

vllm/config/offload.py
```diff
@@ -0,0 +1,94 @@
...
+"""Configuration for model weight offloading."""
+
+from pydantic import Field, model_validator
+from pydantic.dataclasses import dataclass
+
+from vllm.config.utils import config
+
+
+@config
+@dataclass
```
hmellor: `config` is a `dataclass_transform` now

```suggestion
@config
```

vllm/model_executor/models/deepseek_v2.py
wzhao18: In the current offloading, we use `cpu_offload_params` CLI arg to support offloading MoE weights only. It is not perfect (as we need to figure out parameter names which differ across models), but it stays non-intrusive and can apply to all models without code modification. Could you see if that could be applied similarly for the V2?

vllm/model_executor/offloader/v2_ops.py
```diff
@@ -0,0 +1,102 @@
...
+    get_offloader()._start_prefetch(layer_idx)
+    return output_tensor
+
+
+def _start_prefetch_fake(
+    output_tensor: torch.Tensor,
+    layer_idx: int,
+) -> torch.Tensor:
+    """Fake implementation for torch.compile tracing."""
+    return output_tensor
```
zou3519: bad things will happen if you return the tensor directly like this, can you try a custom operator that returns nothing?

It should still maintain the data dependence (since you marked the custom operator as mutating output_tensor).

ditto for wait_prefetch

vllm/model_executor/offloader/v2_ops.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def register_v2_offloader_ops() -> None:
```
zou3519: I'm trying to understand what is going on for cudagraphs. Are these operations cudagraph safe? Because they are CPU->GPU transfers I think the answer is no, and that instead you'd need to exclude them from the cudagraph regions. Does that sound right?

vllm/model_executor/offloader/prefetch_ops.py
```diff
@@ -0,0 +1,102 @@
...
+            dependency. This prevents torch.compile from reordering
+            this op before the computation that produces output_tensor.
+        layer_idx: Index of the layer to prefetch.
+
+    Returns:
+        output_tensor unchanged, creating data dependency for torch.compile.
+    """
+    get_offloader()._start_prefetch(layer_idx)
+    return output_tensor
+
```
zou3519, line 62: btw I tried to run this with `CUDA_VISIBLE_DEVICES=6 TORCH_TRACE=./trace/offload vllm
serve meta-llama/Llama-3.2-1B-Instruct   --offload-group-size 4   --offload-num-in-group 1   --offload-prefetch-step 2   --max-model-len 1024 --port 9023` and it crashed - is this feature only for the specific deepseek model?

vllm/model_executor/offloader/v2_ops.py
zou3519: I think I see what's going on, I'm wrong about the CPU->GPU transfers. we can indeed cudagraph-capture the H2D as long as the addresses are static

vllm/model_executor/offloader/prefetch_ops.py
minosfuture: only configured for deepseek for now, but it shouldn't crash for other models.  It should be no-op if offloading args are not provided.  Lemme check on this crash.

vllm/model_executor/offloader/prefetch.py
```diff
@@ -0,0 +1,686 @@
...
+
+        Call this after the model forward pass but before CUDA graph capture
+        ends. This ensures copy_stream is rejoined for any prefetches started
+        during the forward pass.
+
+        We join ALL layers that have _prefetch_in_capture=True, meaning their
+        prefetch was started during capture but not yet waited on (joined).
+        This handles both full and piecewise cudagraph modes correctly:
+        - Full mode: joins layers 0..prefetch_step-1 (prefetched by last layers)
+        - Piecewise mode: joins only layers prefetched by THIS subgraph's layers
```
zou3519, line 300: Does this feature offer any value for piecewise cudagraphs? for full cudagraphs (and without cudagraphs), there is overlap happening. (I'm not very familiar with how h2d and cudagraphs work, so sorry if this is a bad question).

For example, let's say we have [layer1, layer2, layer3, layer4], and at the end of layer2 we prefetch the weights for layer4, so we are overlapping the prefetch of layer4 weights with the execution of layer3.

When we do piecewise cudagraphs, the cudagraphs are:

[layer1_pre_attention, layer1_post_attention_to_layer2_pre_attention, layer2_post_attention_to_layer3_pre_attention, layer3_post_attention_to_layer4_pre_attention, layer4_post_attention].

The layer2_post_attention_to_layer3_pre_attention will prefetch the layer4 weights. But do the cudagraph constraints mean that the layer4 weight prefetch must finish in this cudagraph?

If so then we're not actually overlapping the prefetch with anything.

vllm/compilation/cuda_graph.py
```diff
@@ -265,6 +266,11 @@ def __call__(self, *args: Any, **kwargs: Any) -> Any | None:
                     set_graph_pool_id(self.graph_pool)
                 else:
                     set_graph_pool_id(current_platform.graph_pool_handle())
+
+                # Sync offloader's copy stream before capture.
+                # Ensure any pre-capture prefetches from offloader are complete.
+                get_offloader().sync_prev_onload()
```
zou3519, line 272: @eellison and/or @BoyuanFeng, could you take a look at this too please? These look reasonable to me but I'm new to this

tests/basic_correctness/test_v2_offload.py
minosfuture: switched to llama as we support any model now.

vllm/model_executor/models/deepseek_v2.py
minosfuture: Updated as @wzhao18 suggested. Feeling much better about this now. cc @mgoin

minosfuture:
@mgoin Hi Michael, could you help review for another round? I hope I've addressed your earlier concerns by moving param config to cli so we can run this with any models now. cc @tlrmchlsmth

vllm/model_executor/offloader/v2_ops.py
minosfuture: updated!

vllm/model_executor/offloader/prefetch.py
minosfuture: that's correct. This support of CUDA graph is to reduce crash with default config (full_and_piecewise). It would not benefit the perf much when CUDA graph is in use.
Batch size of tokens should be small for cases when CUDA graph makes sense. With small batch, layer latency would be too short to hide weight onloading latency anyway. In other words, weight offloading is not useful when CUDA graph is.
In actual usage, to perfectly hide onloading overheads, weight offloading should be used with large batch size, for which CUDA graph is not captured by default.

vllm/model_executor/offloader/prefetch_ops.py
minosfuture: fixed. also switched to this model for unit test as it's much faster.

vllm/compilation/cuda_graph.py
```diff
@@ -273,6 +279,11 @@ def __call__(self, *args: Any, **kwargs: Any) -> Any | None:
                 ):
                     # `output` is managed by pytorch's cudagraph pool
                     output = self.runnable(*args, **kwargs)
+                    # Join offloader's copy stream after forward to avoid
+                    # unjoined stream error. The last layer's start_prefetch
+                    # forks copy_stream, but wait_prefetch only happens in
+                    # the next forward pass.
+                    get_offloader().join_after_forward()
```
BoyuanFeng, line 286: where is the `copy_stream` branched from the current stream?

CUDAGraph requires side stream `copy_stream` to branch off and join back to cudagraph capture stream within the cudagraph capture region:
```python
with torch.cuda.graph(g):
    copy_stream.wait_stream(torch.cuda.current_stream())    # branch the copy_stream from the current_stream
    with torch.cuda.stream(copy_stream):
        x_cpu.copy_(x_cuda, non_blocking=True)

    # any computation
    out = y_cuda + 1

    torch.cuda.current_stream().wait_stream(copy_stream)  # join the copy_stream
```

vllm/compilation/cuda_graph.py
eellison: Just to validate: the same sync, occurs at runtime, prior to graph replay ?

vllm/compilation/cuda_graph.py
minosfuture: hi Boyuan, it's in `start_onload_to_static`, which records `_copy_done_event` that this `join_after_forward` waits on before finishing capture.

BoyuanFeng: thanks. for future-proof, could we expose an api to call it in `cuda_graph.py` for symmetry?

vllm/model_executor/offloader/prefetch.py
```diff
@@ -0,0 +1,691 @@
...
+        self.module = module
+        self.device = next(module.parameters()).device
+        self.copy_stream = copy_stream
+        self.layer_idx = layer_idx
+        self.offloaded_bytes = 0
+
+        # Event to signal when H2D copy to static buffer is complete.
+        # Used for CUDA graph compatible synchronization during capture.
+        # Outside capture, we use wait_stream instead (more robust).
+        self._copy_done_event = torch.cuda.Event()
```
eellison: For my own understanding, why not use streams inside of cudagraphs? there is a constraint that in cudagraphs, any stream forks back to the main stream during capture. but can we enforce that the fork/join loading of weights is always in the prior layer ?

vllm/compilation/cuda_graph.py
```diff
@@ -305,5 +316,8 @@ def __call__(self, *args: Any, **kwargs: Any) -> Any | None:
                 f"got {new_input_addresses}"
             )
 
+        # Sync offloader before replay - ensures any external dependencies
+        # from pre-capture prefetches are satisfied.
+        get_offloader().sync_prev_onload()
```
eellison, line 321: note: this is runtime overhead on hot path. when offloader is not on, worth making this cheaper ?

vllm/model_executor/offloader/prefetch.py
```diff
@@ -0,0 +1,691 @@
...
+        original_forward = module.forward
+
+        def forward(*args, **kwargs):
+            # Temporarily restore original forward to avoid recursion
+            module.forward = original_forward
+
+            # Wait for this layer's prefetch to complete
+            # mutates_args on input_tensor creates data dependency for torch.compile
+            input_tensor = args[0] if args else kwargs.get("hidden_states")
+            torch.ops.vllm.wait_prefetch(input_tensor, index)
```
eellison, line 220: cc @zou3519 , will this result in an unnecessary clone at runtime? do we need an alternative api to declare dependencies ? e.g. effects / control deps etc ?

eellison (commented):
couple comments

tests/basic_correctness/test_prefetch_offload.py
```diff
@@ -0,0 +1,33 @@
...
+            "2",
+            "--offload-prefetch-step",
+            "1",
+            # Selective offloading: only MLP weights
+            "--offload-params",
+            "gate_up_proj",
+            "down_proj",
+        ],
+        [],  # Baseline: no offloading
+    )
```
mgoin, line 33: We should also consider adding a nightly test for running a quantized MoE model e2e on Hopper or Blackwell

vllm/model_executor/offloader/base.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+# Global singleton offloader instance
+_instance: BaseOffloader | None = NoopOffloader()
+
+
+def get_offloader() -> BaseOffloader:
+    """Get the global offloader instance."""
+    assert _instance is not None, "Offloader instance is None"
+    return _instance
```
mgoin: It seems like `assert _instance is not None` will never be hit since the default value is `NoopOffloader()`. I actually think this is bad practice since if this is ever checked before `set_offloader()` is reached they will silently get NoopOffloader instead of failing. We should start as None

vllm/model_executor/offloader/prefetch.py
```diff
@@ -0,0 +1,691 @@
...
+        """
+        return (self.name, self.shape, self.stride, self.dtype)
+
+    @property
+    def num_bytes(self) -> int:
+        """Size in bytes."""
+        numel = 1
+        for dim in self.shape:
+            numel *= dim
+        return numel * torch.tensor([], dtype=self.dtype).element_size()
```
mgoin: Don't make a tensor just to check size? How about `torch.finfo(self.dtype).bits // 8` instead

vllm/entrypoints/llm.py
```diff
@@ -169,6 +169,14 @@ class LLM:
...
             you can use to hold the model weights, at the cost of CPU-GPU data
             transfer for every forward pass.
+        offload_group_size: Advanced CPU offloading: Group every N layers
+            together. Offload last `offload_num_in_group` layers of each group.
+            Default is 0 (disabled).
+        offload_num_in_group: Advanced CPU offloading: Number of layers to
+            offload per group. Default is 1.
+        offload_prefetch_step: Advanced CPU offloading: Number of layers to
+            prefetch ahead. Higher values hide more latency but use more GPU
+            memory. Default is 1.
```
mgoin: Looks like we are missing `offload_params` here in the LLM interface

vllm/model_executor/offloader/prefetch.py
```diff
@@ -0,0 +1,691 @@
...
+        if device is None:
+            # No modules to offload
+            return
+
+        # Allocate static buffer pool
+        self.buffer_pool = StaticBufferPool(
+            param_infos=param_infos,
+            slot_capacity=self.prefetch_step,
+            device=device,
+        )
```
mgoin, line 343: Shouldn't this be `prefetch_step + 1` for true double buffering? If `prefetch_step=1` and there's only 1 buffer slot, layer 0 reads from slot 0 and the prefetch for layer 1 writes to slot 1

vllm/model_executor/offloader/prefetch.py
minosfuture: it's using event to only wait for the weights it needs when prefetch_step is larger than 1.

vllm/compilation/cuda_graph.py
minosfuture: it's no op when offload is not configured. see BaseOffloader and NoopOffloader.

vllm/model_executor/offloader/prefetch.py
zou3519: as long as reinplacing works correctly we should be good

vllm/model_executor/offloader/prefetch.py
minosfuture: prefetch starts at the end of current layer, illustrated in the figure below to compare prefetch_step==1  vs. prefetch_step==2 (double buffering).
image

vllm/compilation/cuda_graph.py
minosfuture: It's asymmetric with the current implementation: events are recorded at per-layer level, and then waited on by the later layers that need the weights.  But unfortunately, CUDA graph require events to be joined before capture ends, and this is why we need `join_after_forward` that waits on any open events that would've been sync'ed by following layers outside the current CUDA graph capture scope.
It is indeed confusing that we only have the sync here. Lemme at least clarify a bit more in the comments for now.

tests/basic_correctness/test_prefetch_offload.py
minosfuture: added deepseek lite test on Hopper

vllm/model_executor/offloader/base.py
minosfuture: Updated. This is one of those defensive code LLM generated that slipped away.  Thanks for the catch.

vllm/model_executor/offloader/prefetch.py
minosfuture: updated

vllm/entrypoints/llm.py
minosfuture: added.  Thanks!

vllm/compilation/cuda_graph.py
minosfuture: Yes. I updated the comment in `run_fullgraph` to clarify this a bit more.

The `sync_prev_onload` before replay is only needed for when vllm transitions from non-cuda-graph execution to cuda-graph execution.
1. non-cg -> non-cg: first layers of next forward would wait on correct copy events
2. non-cg -> cg: first layers of next forward doesn't wait as event cannot cross capture boundary, so we need this sync
3. cg -> non-cg: we have called`join_after_forward` where event sync is captured (copy stream is joined)
4. cg -> cg: same, we have called `join _after_forward` where event sync is captured (copy stream is joined)

vllm/model_executor/offloader/prefetch.py
minosfuture: confirmed from the trace that it didn't introduce memcpy

.buildkite/test_areas/e2e_integration.yaml
```diff
@@ -28,3 +28,12 @@ steps:
   working_dir: "/vllm-workspace"
   commands:
   - bash .buildkite/scripts/scheduled_integration_test/qwen30b_a3b_fp8_block_ep_eplb.sh 0.8 200 8020 2 1
+
+- label: DeepSeek V2-Lite Prefetch Offload Accuracy
```
mgoin:
```suggestion
- label: DeepSeek V2-Lite Prefetch Offload Accuracy (H100)
```

mgoin (approved):
LGTM, thanks for the changes. I think it is obvious the `get_offloader().sync_prev_onload()` and `get_offloader().join_after_forward()` insertions in the model runner are fragile and prone to failure in the future, but I'm not sure how to better structure this. Maybe @benchislett or @LucasWilkinson have strong opinions against this, but we do need this performant feature to land sooner or later

zou3519 (approved):
compile pieces lgtm

PR #29952, opened 2025-12-03
[PCP&DCP] move CUDAGraph check for PCP&DCP to the check func of platforms

pisceskkk:
Since NPU already provides FullGraph support for both PCP and DCP, I believe we should relocate the graph support determination logic from the generic code to platform-specific judgment functions.
Currently, CUDA only supports PIECEWISE mode, but it's unclear whether ROCM behaves consistently with CUDA. Additionally, based on the graph mode determination logic for each platform, backends other than CUDA and ROCM should not require additional handling. If my understanding is incorrect, please feel free to point it out.

CC @LucasWilkinson @FENP @zhenwenqi2024


LucasWilkinson (approved):
LGTM; we should really try to turn on DCP with full-CGs on CUDA; do you know what the blocker is here?

PR #29957, opened 2025-12-03
[Perf][Async] Implement zero-bubble async speculative decoding

izhuhaoran:
## Related Issue

image

#28947 mention: Improved overlapping beyond drafting time (todo, discussion https://github.com/vllm-project/vllm/pull/24799#issuecomment-3528663588)

## Problem

Previously, in async + spec mode, the `_get_valid_sampled_token_count` call in `_update_states` required CPU-GPU synchronization to determine how many tokens were accepted by the target model's rejection sampler. As a result, the prepare CPU work for the current step could only overlap with the draft model’s GPU execution from the previous step. This limited overlap created bubbles (idle periods) in the GPU pipeline, significantly reducing throughput during asynchronous execution.

## Solution
This PR introduces a zero-bubble async execution strategy that:

1. Enables non-blocking CPU execution by optimistically assuming all tokens from the previous speculation round were accepted, allowing `_update_states` and `_prepare_inputs` to proceed without waiting for GPU synchronization
2. Introduces a new `_maybe_adjust_inputs_on_gpu` method to correct any incorrect assumptions entirely on the GPU
3. Defers CPU-side state correction to `_update_states_after_model_execute`, which updates request states and input batch metadata after the target model's forward has been launched

## Test Result

Target model: Qwen3-235B-A22B-Thinking-2507-FP8
Draft Model: lmsys-Qwen3-235B-A22B-EAGLE3
Nvidia H20 GPUs

### bench serve

TP=4, num_speculative_tokens: 1

bench serve --dataset-name spec_bench --spec-bench-output-len 1024 --num-prompts 256

max concurrency | Main Mean TPOT | This PR Mean TPOT | speedup
-- | -- | -- | --
1 | 14.217 | 13.92 | 1.02133620689655
2 | 14.947 | 14.623 | 1.022156876154
4 | 18.157 | 17.82 | 1.018911335578
8 | 20.553 | 20.11 | 1.02202884137245
16 | 23.897 | 23.293 | 1.02593053707122
32 | 23.85 | 23.287 | 1.02417657920728
64 | 23.853 | 23.297 | 1.02386573378547
128 | 23.867 | 23.293 | 1.02464259648822

### GPU Trace

- Main
image

- This PR **(No Bubble)**
image

### lm_eval

- Main
```
local-completions (model=/mnt/data/nas/models/Qwen3-235B-A22B-Thinking-2507-FP8,base_url=http://localhost:8000/v1/completions,max_retries=3), gen_kwargs: (None), limit: None, num_fewshot: None, batch_size: 128
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.6513|±  |0.0131|
|     |       |strict-match    |     5|exact_match|↑  |0.6118|±  |0.0134|
```

- This PR
```
local-completions (model=/mnt/data/nas/models/Qwen3-235B-A22B-Thinking-2507-FP8,base_url=http://localhost:8000/v1/completions,max_retries=3), gen_kwargs: (None), limit: None, num_fewshot: None, batch_size: 128
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.6535|±  |0.0131|
|     |       |strict-match    |     5|exact_match|↑  |0.6194|±  |0.0134|
```

### DIFF
--- a/vllm/config/speculative.py
+++ b/vllm/config/speculative.py
@@ -103,6 +103,12 @@ class SpeculativeConfig:
     speculative input batches can contain sequences of different lengths,
     which may only be supported by certain attention backends. This currently
     only affects the EAGLE method of speculation."""
+    async_spec_zero_bubble_mode: bool = False
+    """Enable async speculative decoding zero bubble mode. If set to True,
+    optimistically assumes all pre-step speculated tokens were accepted to skip the
+    cpu sync in _update_states, and correct the inputs based on GPU operations. It
+    allowing current-step CPU preparation to fully overlap with previous-step target
+    model GPU execution, eliminating gpu timeline bubbles."""
 
     # Ngram proposer configuration
     prompt_lookup_max: int | None = Field(default=None, ge=1)
--- a/vllm/v1/spec_decode/eagle.py
+++ b/vllm/v1/spec_decode/eagle.py
@@ -604,11 +604,18 @@ class EagleProposer:
         """
         # Precompute get_token_id for when there is no valid next token
         num_reqs = gpu_input_batch.num_reqs
+        num_computed_tokens_cpu = gpu_input_batch.num_computed_tokens_cpu_tensor[
+            :num_reqs
+        ]
+        query_lens_cpu = (
+            common_attn_metadata.query_start_loc_cpu[1 : num_reqs + 1]
+            - common_attn_metadata.query_start_loc_cpu[:num_reqs]
+        )
+        seq_lens_cpu = num_computed_tokens_cpu + query_lens_cpu
+        seq_lens_list = seq_lens_cpu.tolist()
         self.backup_next_token_ids.np[:num_reqs] = np.array(
             [
-                requests[gpu_input_batch.req_ids[i]].get_token_id(
-                    common_attn_metadata.seq_lens_cpu[i].item()
-                )
+                requests[gpu_input_batch.req_ids[i]].get_token_id(seq_lens_list[i])
                 for i in range(num_reqs)
             ],
             dtype=np.int32,
@@ -693,7 +700,7 @@ class EagleProposer:
             num_reqs=common_attn_metadata.num_reqs,
             num_actual_tokens=total_num_tokens,
             max_query_len=new_query_len_per_req.max().item(),
-            max_seq_len=common_attn_metadata.seq_lens_cpu.max().item(),
+            max_seq_len=common_attn_metadata.max_seq_len,
             block_table_tensor=common_attn_metadata.block_table_tensor,
             slot_mapping=common_attn_metadata.slot_mapping[:total_num_tokens],
             causal=True,
--- a/vllm/v1/worker/gpu_model_runner.py
+++ b/vllm/v1/worker/gpu_model_runner.py
@@ -1792,14 +1792,21 @@ class GPUModelRunner(
         block_table_gid_0, slot_mapping_gid_0 = _get_block_table_and_slot_mapping(0)
         if self.model_config.enable_return_routed_experts:
             self.slot_mapping = slot_mapping_gid_0[:num_tokens].cpu().numpy()
+
+        if self.async_spec_zero_bubble_mode:
+            seq_lens_cpu = None
+            num_computed_tokens_cpu = None
+        else:
+            seq_lens_cpu = self.seq_lens.cpu[:num_reqs_padded]
+            num_computed_tokens_cpu = self.input_batch.num_computed_tokens_cpu_tensor[
+                :num_reqs_padded
+            ]
         cm_base = CommonAttentionMetadata(
             query_start_loc=self.query_start_loc.gpu[: num_reqs_padded + 1],
             query_start_loc_cpu=self.query_start_loc.cpu[: num_reqs_padded + 1],
             seq_lens=self.seq_lens.gpu[:num_reqs_padded],
-            _seq_lens_cpu=self.seq_lens.cpu[:num_reqs_padded],
-            _num_computed_tokens_cpu=self.input_batch.num_computed_tokens_cpu_tensor[
-                :num_reqs_padded
-            ],
+            _seq_lens_cpu=seq_lens_cpu,
+            _num_computed_tokens_cpu=num_computed_tokens_cpu,
             num_reqs=num_reqs_padded,
             num_actual_tokens=num_tokens_padded,
             max_query_len=max_query_len,
--- a/vllm/v1/worker/gpu_model_runner.py
+++ b/vllm/v1/worker/gpu_model_runner.py
@@ -1418,49 +1418,8 @@ class GPUModelRunner(
         are accepted in update_states to avoid cpu sync, and will adjust the
         inputs correctly based on the actual accepted token count in GPU."""
 
-        if self.valid_sampled_token_count_gpu is None:
+        if not self.async_spec_zero_bubble_mode:
             return
-
-        prev_req_id_to_index = self.input_batch.prev_req_id_to_index
-        if not prev_req_id_to_index or not self.async_spec_reqs_to_fix:
-            return
-
-        req_indices_to_correct = []
-        prev_req_indices = []
-        prev_num_draft_lens = []
-        for i, req_id in enumerate(self.input_batch.req_ids):
-            if req_id not in self.async_spec_reqs_to_fix:
-                continue
-            num_draft_len = self.async_spec_reqs_to_fix[req_id]
-            prev_index = prev_req_id_to_index.get(req_id)
-            assert prev_index is not None
-            req_indices_to_correct.append(i)
-            prev_req_indices.append(prev_index)
-            prev_num_draft_lens.append(num_draft_len)
-
-        if not req_indices_to_correct:
-            return
-
-        req_indices_tensor = torch.tensor(
-            req_indices_to_correct, dtype=torch.int64, pin_memory=self.pin_memory
-        ).to(self.device, non_blocking=True)
-        prev_req_indices_tensor = torch.tensor(
-            prev_req_indices, dtype=torch.int64, pin_memory=self.pin_memory
-        ).to(self.device, non_blocking=True)
-
-        # Compute the diff between optimistic and actual valid tokens cnt.
-        prev_num_draft_len_tensor = torch.tensor(
-            prev_num_draft_lens, dtype=torch.int64, pin_memory=self.pin_memory
-        ).to(self.device, non_blocking=True)
-
-        diff = (
-            prev_num_draft_len_tensor
-            + 1
-            - self.valid_sampled_token_count_gpu[prev_req_indices_tensor].to(
-                torch.int64
-            )
-        )
-        self.seq_lens.gpu[req_indices_tensor] -= diff.int()
 
         num_reqs = self.input_batch.num_reqs
         req_indices_flat_cpu = np.repeat(
@@ -1470,23 +1429,66 @@ class GPUModelRunner(
             self.device, non_blocking=True
         )
 
-        diff_all_reqs = torch.zeros(num_reqs, dtype=torch.int64, device=self.device)
-        diff_all_reqs[req_indices_tensor] = diff
+        req_indices_to_correct = []
+        prev_req_indices = []
+        prev_num_draft_lens = []
+        prev_req_id_to_index = self.input_batch.prev_req_id_to_index
+        if (
+            self.valid_sampled_token_count_gpu is not None
+            and prev_req_id_to_index
+            and self.async_spec_reqs_to_fix
+        ):
+            for i, req_id in enumerate(self.input_batch.req_ids):
+                if req_id not in self.async_spec_reqs_to_fix:
+                    continue
+                num_draft_len = self.async_spec_reqs_to_fix[req_id]
+                prev_index = prev_req_id_to_index.get(req_id)
+                assert prev_index is not None
+                req_indices_to_correct.append(i)
+                prev_req_indices.append(prev_index)
+                prev_num_draft_lens.append(num_draft_len)
 
-        diff_per_token = diff_all_reqs[req_indices_flat_gpu]
-        self.positions.gpu[: req_indices_flat_gpu.shape[0]] -= diff_per_token
+        if req_indices_to_correct:
+            assert self.valid_sampled_token_count_gpu is not None
+            req_indices_tensor = torch.tensor(
+                req_indices_to_correct, dtype=torch.int64, pin_memory=self.pin_memory
+            ).to(self.device, non_blocking=True)
+            prev_req_indices_tensor = torch.tensor(
+                prev_req_indices, dtype=torch.int64, pin_memory=self.pin_memory
+            ).to(self.device, non_blocking=True)
 
-        if self.uses_mrope:
-            for dim in range(3):
-                self.mrope_positions.gpu[dim, : req_indices_flat_gpu.shape[0]] -= (
-                    diff_per_token
+            # Compute the diff between optimistic and actual valid tokens cnt.
+            prev_num_draft_len_tensor = torch.tensor(
+                prev_num_draft_lens, dtype=torch.int64, pin_memory=self.pin_memory
+            ).to(self.device, non_blocking=True)
+
+            # fix seq_lens on gpu
+            diff = (
+                prev_num_draft_len_tensor
+                + 1
+                - self.valid_sampled_token_count_gpu[prev_req_indices_tensor].to(
+                    torch.int64
                 )
-        if self.uses_xdrope_dim > 0:
-            for dim in range(self.uses_xdrope_dim):
-                self.xdrope_positions.gpu[dim, : req_indices_flat_gpu.shape[0]] -= (
-                    diff_per_token
-                )
+            )
+            self.seq_lens.gpu[req_indices_tensor] -= diff.int()
 
+            # fix positions on gpu
+            diff_all_reqs = torch.zeros(num_reqs, dtype=torch.int64, device=self.device)
+            diff_all_reqs[req_indices_tensor] = diff
+            diff_per_token = diff_all_reqs[req_indices_flat_gpu]
+            self.positions.gpu[: req_indices_flat_gpu.shape[0]] -= diff_per_token
+            if self.uses_mrope:
+                for dim in range(3):
+                    self.mrope_positions.gpu[dim, : req_indices_flat_gpu.shape[0]] -= (
+                        diff_per_token
+                    )
+            if self.uses_xdrope_dim > 0:
+                for dim in range(self.uses_xdrope_dim):
+                    self.xdrope_positions.gpu[dim, : req_indices_flat_gpu.shape[0]] -= (
+                        diff_per_token
+                    )
+
+        # Compute slot mapping only on GPU in async_spec_zero_bubble_mode
         self.input_batch.block_table.compute_slot_mapping_gpu(
             req_indices_flat_gpu, self.positions.gpu[: req_indices_flat_gpu.shape[0]]
         )
@@ -1604,8 +1606,9 @@ class GPUModelRunner(
 
                 output_idx += num_sched
 
-        self.input_batch.block_table.compute_slot_mapping(req_indices, positions_np)
-        self.input_batch.block_table.commit_slot_mapping(total_num_scheduled_tokens)
+        if not self.async_spec_zero_bubble_mode:
+            self.input_batch.block_table.compute_slot_mapping(req_indices, positions_np)
+            self.input_batch.block_table.commit_slot_mapping(total_num_scheduled_tokens)
 
         # Prepare the attention metadata.
         self.query_start_loc.np[0] = 0
--- a/vllm/v1/worker/gpu_model_runner.py
+++ b/vllm/v1/worker/gpu_model_runner.py
@@ -456,6 +456,9 @@ class GPUModelRunner(
             self.rejection_sampler = RejectionSampler(self.sampler)
 
         self.num_spec_tokens = 0
+        self.async_spec_zero_bubble_mode = False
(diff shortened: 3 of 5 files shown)


izhuhaoran:
@njhill I've already add the perf bench result in PR description, PTAL, also cc @benchislett 

BTW, the actual perf improvement isn't as large as shown in Nsight Systems, possibly because profiling itself exacerbates the bubble issue (just a guess). Regardless, this PR overall should be more beneficial for async + spec ?

robertgshaw2-redhat:
really nice work

vllm/envs.py
```diff
@@ -236,6 +236,7 @@
     VLLM_SHARED_EXPERTS_STREAM_TOKEN_THRESHOLD: int = 256
     VLLM_COMPILE_CACHE_SAVE_FORMAT: Literal["binary", "unpacked"] = "binary"
     VLLM_USE_V2_MODEL_RUNNER: bool = False
+    VLLM_ASYNC_SPS_ZERO_BUBBLE_MODE: bool = False
```
robertgshaw2-redhat: Rather than an enviornment variable, lets add this to the speculative decoding config

izhuhaoran: Thanks for this suggestion, done.

benchislett (changes requested):
I would prefer to implement this based on #29624 instead of letting seq_lens_cpu decay when unused and gating at the attention-backend level

benchislett:
now that 29624 is merged, can you update this to utilize this? I think it might help simplify some of the logic

vllm/config/speculative.py
```diff
@@ -101,6 +101,12 @@ class SpeculativeConfig:
     speculative input batches can contain sequences of different lengths,
     which may only be supported by certain attention backends. This currently
     only affects the EAGLE method of speculation."""
+    async_spec_zero_bubble_mode: bool = False
```
MatthewBonanni, line 106: Could we eliminate this option and just have it on by default? Seems the only consequence will be that async + spec decode won't be compatible with cascade attention, which isn't a big deal

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -648,6 +652,14 @@ def __init__(
             pin_memory=self.pin_memory,
         )
 
+        self.async_spec_zero_bubble_mode = (
+            async_spec_zero_bubble_mode
+            and self.use_async_scheduling
+            and self.num_spec_tokens > 0
+            and not self.cascade_attn_enabled
+        )
+        self.async_spec_reqs_to_fix: set[str] = set()
```
MatthewBonanni: Instead of a `set[str]`, could we make this a `dict[str, int]` where the key is `req_id` and the value is `req_state.prev_num_draft_len`? Then you could get rid of `pending_prev_num_draft_len` and `pre_valid_sampled_token_count`

vllm/v1/attention/backends/utils.py
```diff
@@ -314,6 +314,7 @@ class AttentionMetadataBuilder(abc.ABC, Generic[M]):
     # Does this backend/builder support CUDA Graphs for attention (default: no).
     # Do not access directly. Call get_cudagraph_support() instead.
     _cudagraph_support: ClassVar[AttentionCGSupport] = AttentionCGSupport.NEVER
+
```
MatthewBonanni: nit: undo this change if there are no other changes to this file

vllm/v1/worker/gpu_model_runner.py
MatthewBonanni:
```suggestion
if not self.async_spec_reqs_to_fix:
    self.input_batch.block_table.compute_slot_mapping(req_indices, positions_np)
    self.input_batch.block_table.commit_slot_mapping(total_num_scheduled_tokens)
```

MatthewBonanni: We shouldn't run these at all in async mode, right?

vllm/v1/worker/gpu_input_batch.py
```diff
@@ -49,6 +49,7 @@ class CachedRequestState:
 
     # Used when both async_scheduling and spec_decode are enabled.
     prev_num_draft_len: int = 0
+    pending_prev_num_draft_len: int = 0
```
MatthewBonanni: See https://github.com/vllm-project/vllm/pull/29957/changes#r2666380423
```suggestion
```

vllm/v1/worker/gpu_input_batch.py
```diff
@@ -267,6 +268,7 @@ def __init__(
         # Cached reference to the GPU tensor of previously sampled tokens
         self.prev_sampled_token_ids: torch.Tensor | None = None
         self.prev_req_id_to_index: dict[str, int] | None = None
+        self.pre_valid_sampled_token_count: torch.Tensor | None = None
```
MatthewBonanni: See https://github.com/vllm-project/vllm/pull/29957/changes#r2666380423
```suggestion
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -648,6 +652,14 @@ def __init__(
             pin_memory=self.pin_memory,
         )
 
+        self.async_spec_zero_bubble_mode = (
+            async_spec_zero_bubble_mode
+            and self.use_async_scheduling
+            and self.num_spec_tokens > 0
+            and not self.cascade_attn_enabled
+        )
```
MatthewBonanni: We can make a property `self.use_async_spec_decode` and use this throughout
```suggestion
        self.use_async_spec_decode = (
            self.use_async_scheduling and self.num_spec_tokens > 0
        )
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1040,7 +1064,14 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> None:
...
             if self.use_async_scheduling:
-                req_state.prev_num_draft_len = num_spec_tokens
+                if (
+                    self.async_spec_zero_bubble_mode
+                    and req_id in self.async_spec_reqs_to_fix
+                ):
+                    req_state.pending_prev_num_draft_len = num_spec_tokens
+                else:
+                    req_state.prev_num_draft_len = num_spec_tokens
+                    req_state.pending_prev_num_draft_len = 0
```
MatthewBonanni: You don't need this change if you do https://github.com/vllm-project/vllm/pull/29957/changes#r2666380423
```suggestion
                req_state.prev_num_draft_len = num_spec_tokens
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1095,6 +1131,59 @@ def _update_states_after_model_execute(
         for i, num_tokens in enumerate(num_accepted_tokens):
             self.input_batch.num_accepted_tokens_cpu[i] = num_tokens
 
+    def _finalize_async_spec_cpu_state(self) -> None:
+        """Synchronize CPU metadata after deferred speculative acceptance."""
+
+        if not self.async_spec_zero_bubble_mode:
```
MatthewBonanni:
```suggestion
        if not self.async_spec_reqs_to_fix:
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1095,6 +1131,59 @@ def _update_states_after_model_execute(
...
+        req_ids_to_fix = list(self.async_spec_reqs_to_fix)
+
+        if not req_ids_to_fix:
+            return
+
+        valid_sampled_token_count = self._get_valid_sampled_token_count()
+        if not valid_sampled_token_count:
+            return
+
+        for req_id in req_ids_to_fix:
```
MatthewBonanni:
```suggestion
        for req_id, prev_draft_len in self.async_spec_reqs_to_fix.items():
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1333,6 +1422,87 @@ def _get_encoder_seq_lens(
...
+        scheduler_output: "SchedulerOutput",
+        num_scheduled_tokens: np.ndarray,
+    ) -> None:
+        """adjust inputs correctly on gpu for async scheduling with spec decode.
+
+        In async scheduling with spec decode, we assume all speculated tokens
+        are accepted in update_states to avoid cpu sync, and will adjust the
+        inputs correctly based on the actual accepted token count in GPU."""
+
+        if not self.async_spec_zero_bubble_mode:
```
MatthewBonanni, line 1421: I think we should just always update on GPU when doing async spec decoding. This would mean getting rid of the early returns (and managing these cases some other way, perhaps with nested if statements?) so that [line 1502](https://github.com/vllm-project/vllm/pull/29957/changes#diff-80ee7e2a62f9dcfbb8a312dc4e3948557e97ef187290daebbcae1e28596bda29R1502) always runs

vllm/v1/worker/gpu_model_runner.py
MatthewBonanni: If we plan on always updating the slot mapping on GPU, we'll need to make sure these don't run in `use_async_spec_decode` mode

```suggestion

        if not self.use_async_spec_decode:
            self.input_batch.block_table.compute_slot_mapping(req_indices, positions_np)
            self.input_batch.block_table.commit_slot_mapping(total_num_scheduled_tokens)
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1629,14 +1804,25 @@ def _get_block_table_and_slot_mapping(kv_cache_gid: int):
             return blk_table_tensor, slot_mapping
 
         block_table_gid_0, slot_mapping_gid_0 = _get_block_table_and_slot_mapping(0)
+
+        # NOTE: In async spec zero bubble mode, we keep _seq_lens_cpu and
+        # _num_computed_tokens_cpu as None. Attn Backends needing these
+        # values will auto-compute them via seq_lens.to("cpu")
+        if self.async_spec_zero_bubble_mode:
```
MatthewBonanni, line 1796:
```suggestion
        if self.use_async_spec_decode:
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -3559,6 +3747,7 @@ def _copy_valid_sampled_token_count(
             counts_cpu[: counts.shape[0]].copy_(counts, non_blocking=True)
             self.valid_sampled_token_count_event.record()
 
+        self.input_batch.pre_valid_sampled_token_count = valid_sampled_tokens_count
```
MatthewBonanni:
```suggestion
        if self.use_async_spec_decode:
            self.valid_sampled_token_count_gpu = valid_sampled_tokens_count
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -3400,6 +3586,7 @@ def sample_tokens(
             sampler_output = self._sample(logits, spec_decode_metadata)
 
         self.input_batch.prev_sampled_token_ids = None
+        self.input_batch.pre_valid_sampled_token_count = None
```
MatthewBonanni:
```suggestion
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1333,6 +1422,87 @@ def _get_encoder_seq_lens(
...
+
+        # Compute the diff between optimistic and actual valid tokens cnt.
+        prev_num_draft_len_tensor = torch.tensor(
+            prev_num_draft_lens, dtype=torch.int64, pin_memory=self.pin_memory
+        ).to(self.device, non_blocking=True)
+
+        diff = (
+            prev_num_draft_len_tensor
+            + 1
+            - pre_valid_sampled_token_count[prev_req_indices_tensor].to(torch.int64)
```
MatthewBonanni: Per https://github.com/vllm-project/vllm/pull/29957/changes#r2677405792, you can use `self.valid_sampled_token_count_gpu[prev_req_indices_tensor]` here

MatthewBonanni (commented):
Thanks for this contribution! Just a few comments. As is, it looks like this would sometimes update the slot mapping on CPU and sometimes on GPU (depending on whether or not there are requests to adjust). I think we should just always do it on GPU when we're doing async spec decoding

LucasWilkinson (changes requested):
hey @izhuhaoran thanks for the contribution! apologies for the very delayed reviews was following up with @MatthewBonanni 's [comment](https://github.com/vllm-project/vllm/pull/29957#pullrequestreview-3632600156) offline i agree with @MatthewBonanni i think we should explore just making `async_spec_zero_bubble_mode=True` be default and just eliminate that flag completely to simplify the code

vllm/config/speculative.py
izhuhaoran: Since `async_spec_zero_bubble_mode` introduces additional minor GPU/CPU overhead, I’d prefer not to enable it by default. It’s mainly beneficial in the specific case where async+spec has bubbles (i.e., the previous step’s draft GPU time is smaller than the current step’s CPU time). Otherwise, the original async+spec path is more efficient.

vllm/v1/worker/gpu_model_runner.py
izhuhaoran: Good idea — I’ve updated it to use a dict

vllm/v1/attention/backends/utils.py
izhuhaoran: Sorry for missing that — I’ve removed this.

vllm/v1/worker/gpu_input_batch.py
izhuhaoran: Done.

vllm/v1/worker/gpu_model_runner.py
izhuhaoran: I’ve cleaned this up and moved the async_spec_zero_bubble_mode gating logic into VllmConfig.post_init()

vllm/v1/worker/gpu_model_runner.py
izhuhaoran: Since the current code path already supports async+spec (with bubbles in some cases, as described in the PR), use_async_spec_decode feels a bit misleading here — it would read as “async+spec is enabled”, but this PR is really introducing an alternative implementation (the zero-bubble variant). So, keep the name or other better suggestion?

vllm/v1/worker/gpu_model_runner.py
izhuhaoran: nice suggestion, done.

vllm/v1/worker/gpu_model_runner.py
izhuhaoran: done.

vllm/v1/worker/gpu_model_runner.py
izhuhaoran: Done.

vllm/v1/worker/gpu_model_runner.py
izhuhaoran: Done.

vllm/v1/worker/gpu_model_runner.py
izhuhaoran: also see https://github.com/vllm-project/vllm/pull/29957#discussion_r2695281552

vllm/v1/worker/gpu_model_runner.py
izhuhaoran: Nice catch — WIP ,  I’ll try this so that under `async_spec_zero_bubble_mode` we always use `compute_slot_mapping_gpu()`, and skip `compute_slot_mapping()` / `commit_slot_mapping()`.

vllm/v1/worker/gpu_model_runner.py
izhuhaoran: same https://github.com/vllm-project/vllm/pull/29957#discussion_r2695368650

vllm/v1/worker/gpu_model_runner.py
MatthewBonanni: I made this suggestion based on making this zero-bubble variant non-optional (eliminating `async_spec_zero_bubble_mode`), in which case no distinction would be necessary. I think `async_spec_zero_bubble_mode` is a fine name if we leave it optional

vllm/config/speculative.py
MatthewBonanni: See discussion below https://github.com/vllm-project/vllm/pull/29957#issuecomment-3756348391

vllm/v1/worker/gpu_model_runner.py
izhuhaoran: Done. I've apply this change

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1102,6 +1129,48 @@ def _update_states_after_model_execute(
         for i, num_tokens in enumerate(num_accepted_tokens):
             self.input_batch.num_accepted_tokens_cpu[i] = num_tokens
 
+    def _finalize_async_spec_cpu_state(self) -> None:
```
MatthewBonanni, line 1132: I think that this method can be eliminated entirely and we can just let the CPU's `num_computed_tokens` drift. We'd just need to modify EAGLE's `prepare_next_token_ids_padded`

vllm/v1/spec_decode/eagle.py
MatthewBonanni, line 627: Pairing with the above change, you could replace 607-627 with the following:

```
        batch_size, num_tokens = sampled_token_ids.shape
        device = sampled_token_ids.device
        prev_sampled = gpu_input_batch.prev_sampled_token_ids
        prev_req_id_to_index = gpu_input_batch.prev_req_id_to_index
        if (
            prev_sampled is not None
            and prev_req_id_to_index
            and prev_sampled.shape[0] > 0
        ):
            # Build index mapping: current index -> previous index (-1 if new)
            prev_indices = [
                prev_req_id_to_index.get(req_id, -1)
                for req_id in gpu_input_batch.req_ids[:num_reqs]
            ]
            prev_indices_tensor = torch.tensor(
                prev_indices, dtype=torch.int64, device=device
            )
            valid_mask = prev_indices_tensor >= 0
            # Clamp to 0 for gather (masked out below), gather backup tokens
            clamped_indices = prev_indices_tensor.clamp(min=0)
            backup_tokens_gpu = torch.where(
                valid_mask,
                prev_sampled[clamped_indices, 0].to(torch.int32),
                torch.zeros(1, dtype=torch.int32, device=device),
            )
        else:
            backup_tokens_gpu = torch.zeros(
                num_reqs, dtype=torch.int32, device=device
            )
```

Then this function won't need `common_attn_metadata` and `requests` as arguments

vllm/v1/worker/gpu_model_runner.py
izhuhaoran: Thanks for the suggestion. But I don’t think we can eliminate this method safely.

If we let the CPU-side `num_computed_tokens` drift (accumulate error across steps), later steps typically only adjust by the current step’s rejected-token count. That won’t match the accumulated drift, so `_maybe_adjust_inputs_on_gpu` can end up using an incorrect offset and produce wrong GPU-side inputs.

The only way to make “drift” safe would be to call `_get_valid_sampled_token_count` elsewhere and track/apply the cumulative reject count when adjusting—but that still requires the same sync point. Profiling shows this sync is the dominant cost in `_finalize_async_spec_cpu_state`; the remaining adjustments are negligible.

vllm/v1/spec_decode/eagle.py
izhuhaoran: Nice idea overall, but I think there may be correctness issues.
With this approach, if a req wasn’t present in the previous batch, `prev_req_id_to_index` returns -1 and the code falls back to token 0 via torch.where. This could cause a req that would otherwise get the correct backup token to instead receive 0, which can lead to incorrect behavior. In the previous logic the special/missing case was -1, but here it always becomes 0, which can also be problematic.
Also, given we still need `_finalize_async_spec_cpu_state` to prevent CPU/GPU state drift, keeping the current logic that derives the backup token from the request state (via seq_lens) seems safer/consistent.

MatthewBonanni: The backup token only matters during chunked prefill, though, right? I don't think it should make any difference what token we pad with during spec decoding

vllm/v1/worker/gpu_model_runner.py
MatthewBonanni: I've been playing around locally with the idea of just computing `_prepare_inputs` on the GPU, so the GPU is always the default source of truth, rather than the CPU. This could add some overhead but I understand from @LucasWilkinson that this seems to be what model runner V2 is doing. I'll share some code tomorrow once I get this a bit more ironed out

izhuhaoran: Yep, that was my initial goal for this PR—to move computed_tokens and seq_len operations entirely to the GPU. But the changes felt substantial and would benefit from community discussion first. I also discovered that model runner V2 is already doing this. So to minimize changes, I went with this approach: preserving the existing CPU path while adding GPU-side corrections. Moving everything to GPU would definitely be cleaner, and I'd be happy to collaborate on that!

vllm/v1/spec_decode/eagle.py
izhuhaoran: My concern isn’t mainly about padding with 0 vs -1 (though -1 can sometimes make it easier to see it’s invalid, but 0 also no wrong ). What I’m worried about is: the backup token is for next_token_ids, which is for drafer's input_ids. If a request is not in the previous batch but is in the current batch, then with the suggested code the backup would become a pad token, whereas the previous logic would fetch a normal backup token from `prompt_token_ids` or `output_token_ids`. I’m not sure whether this could lead to a potential drop in acceptance rate.

izhuhaoran: But, if we decide to remove `num_computed_tokens_cpu`, we’d need to further modify this code—maybe use `prev_sampled_token_ids` or some other approach (e.g., model runner v2 UVA tensors). We probably need more discussion on this.  Also cc @LucasWilkinson @benchislett

PR #28922, opened 2025-11-18
feat: Add NCCL batch optimization for P2P KV transfer

xbfs:
# Add NCCL Batch Optimization for P2P KV Transfer

## Summary

This PR adds NCCL batch optimization for P2P KV cache transfer to improve performance when transferring multiple tensors (e.g., multiple layers) to the same remote worker. The optimization reduces NCCL launch overhead and improves GPU pipeline utilization.

**Target Branch**: `releases/v0.10.1`

## Changes

- **Batch sending**: Implemented `send_sync_batch` method that groups multiple tensors with the same `remote_address` and sends them using `ncclGroupStart/End`
- **PUT_BATCH protocol**: Added support for batch tensor transfer via new `PUT_BATCH` command in `listen_for_requests`
- **Async stream operations**: Replaced blocking `stream.synchronize()` with async `wait_stream()` operations in `send` and `recv` methods
- **Queue batching**: Modified `send_async` to collect consecutive items with the same remote address into batches

## Performance Impact

**Expected improvements for 32-layer model:**
- Before: 32 ZMQ messages + 32 NCCL launches + 32 synchronizations
- After: 1 ZMQ message + 1 grouped NCCL launch + async operations
- **Estimated speedup**: 2-5x for multi-layer transfers

## Backward Compatibility

✅ **Fully backward compatible**
- Single tensor transfers use existing `PUT` command (unchanged)
- Batch optimization only activates for multiple tensors to same address
- No breaking changes to public APIs

## Files Changed

- `vllm/distributed/kv_transfer/kv_connector/v1/p2p/p2p_nccl_engine.py`

## Testing

origin-nccl-kv-transfer:
nccl
improved-nccl-kv-transfer:
groupnccl

---

**Type**: Performance Optimization  
**Target Version**: v0.10.1

PR #29961, opened 2025-12-03
[Feature] Integrate NCCL batch optimization for P2P KV transfer

xbfs:
## Summary

This PR integrates the NCCL batch optimization for P2P KV cache transfer from PR #28922 (https://github.com/vllm-project/vllm/pull/28922) into the latest main branch.

Adds NCCL batch optimization to improve performance when transferring multiple tensors (e.g., multiple layers) to the same remote worker. The optimization reduces NCCL launch overhead and improves GPU pipeline utilization.

## Changes

* **Batch sending**: Implemented `send_sync_batch` method that groups multiple tensors with the same `remote_address` and sends them using `ncclGroupStart/ncclGroupEnd`
* **PUT_BATCH protocol**: Added support for batch tensor transfer via new `PUT_BATCH` command in `listen_for_requests`
* **Queue batching**: Modified `send_async` to collect consecutive items with the same remote address into batches
* **Helper method**: Added `_finalize_received_tensor` method for code reuse

## Performance Impact

**Tested with 128 concurrent requests:**
Before optimization and After optimization:
  

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| TTFT Mean | 2681.78ms | 497.92ms | **-81.4%** ⬇️ |
| TTFT Median | 2999.17ms | 504.38ms | **-83.2%** ⬇️ |
| TTFT P90 | 3011.02ms | 585.05ms | **-80.6%** ⬇️ |
| Request throughput | 4.06 req/s | 4.39 req/s | **+8.1%** ⬆️ |
| Output token throughput | 727.82 tok/s | 787.16 tok/s | **+8.2%** ⬆️ |

The improvement is especially significant in high concurrency scenarios.

## Backward Compatibility

✅ **Fully backward compatible**

* Single tensor transfers use existing `PUT` command (unchanged)
* Batch optimization only activates for multiple tensors to same address
* No breaking changes to public APIs

## Testing

Thoroughly tested under high concurrency scenarios (128 concurrent requests) with verification of:
- ✅ Batch transfer correctness
- ✅ Single tensor transfer compatibility
- ✅ Memory management and error handling
- ✅ NCCL Group API synchronization


vllm/distributed/kv_transfer/kv_connector/v1/p2p/p2p_nccl_engine.py
```diff
@@ -433,6 +420,74 @@ def listen_for_requests(self):
                     self.have_received_tensor_id(tensor_id)
                     self.recv_store_cv.notify()
 
+            elif data["cmd"] == "PUT_BATCH":
```
ywang96, line 425: Could you please also update our documentation regarding this new cmd?
https://docs.vllm.ai/en/latest/design/p2p_nccl_connector/?h=put#kv-cache-transfer-methods

ywang96 (commented):
@ApostaC @NickLucche If either of you can review this PR it would be great!

vllm/distributed/kv_transfer/kv_connector/v1/p2p/p2p_nccl_engine.py
xbfs: @ywang96 Thanks for the review!
This is a great suggestion, but the `PUT_BATCH` command is an **internal optimization that automatically activates in `PUT_ASYNC` mode** (which is the default mode).
**How it works:**
The optimization is in the `send_async` thread (L534-548):
- **PUT_ASYNC mode** (default): The async sending thread automatically detects when multiple tensors are queued for the same destination and batches them using `PUT_BATCH` command with NCCL Group API
- **Single tensor**: Uses existing `PUT` command
- **Multiple tensors to same destination**: Automatically uses `PUT_BATCH` 
- **PUT/GET modes**: Continue to work as before without batching

Should I add a design note explaining this automatic batching in the default `PUT_ASYNC` mode, or keep the docs as-is since no user configuration is needed?

vllm/distributed/kv_transfer/kv_connector/v1/p2p/p2p_nccl_engine.py
```diff
@@ -433,6 +420,74 @@ def listen_for_requests(self):
...
+                            )
+                    finally:
+                        self.nccl.ncclGroupEnd()
+
+                self.recv_stream.synchronize()
+
+                tensors_to_store: list[tuple[str, Any]] = []
+                for tensor_meta, tensor in allocated:
+                    tensor_id = tensor_meta["tensor_id"]
+                    stored_tensor = self._finalize_received_tensor(
```
ApostaC, line 482: The function `_finalize_received_tensor` has a branch where it will return `(addr, tensor.dtype, tensor.shape)` rather than a single tensor. 

How does that work here (as well as at the other place we call `_finalize_received_tensor` ) above?

ApostaC (commented):
Otherwise LGTM!

vllm/distributed/kv_transfer/kv_connector/v1/p2p/p2p_nccl_engine.py
xbfs: @ApostaC Good catch! 

Yeah, `_finalize_received_tensor` can return either a tensor or a tuple `(addr, dtype, shape)` when the buffer is full. This is actually the same pattern used in the existing PUT command.

Both types get stored in `recv_store[tensor_id]` (which is `dict[str, Any]`), and `recv_tensor` already handles this with the `isinstance(tensor, tuple)` check at L321-325. So the batch path just reuses the same logic - we're processing multiple tensors in a loop instead of one, but the storage/retrieval pattern is identical.

The existing PUT command does the same thing (L405-407), so this is consistent with the codebase.

xbfs:
Hi, @ApostaC, Thank you again for pointing out the detail earlier.
I've rechecked the logic to ensure the batch path is fully consistent with the existing `PUT` handling.
If there's anything else that I might have overlooked, I'm more than happy to update the PR. Otherwise, please let me know if it looks good to you now. Thanks for your time and guidance!

vllm/distributed/kv_transfer/kv_connector/v1/p2p/p2p_nccl_engine.py
ApostaC: I see. Then it could be a bit confusing to call it a tensor.

Additionally, if the value type of `recv_store` is either torch.Tensor or tuple[addr, dtype, shape], we should give it a clear type hint with `Union[torch.Tensor, tuple[xxx, xxx, xxx]]` rather than `Any`.

vllm/distributed/kv_transfer/kv_connector/v1/p2p/p2p_nccl_engine.py
```diff
@@ -624,6 +754,28 @@ def recv(self, comm, tensor: torch.Tensor, src: int, stream=None):
             )
         stream.synchronize()
 
+    def _finalize_received_tensor(
+        self,
+        tensor_id: str,
+        tensor: torch.Tensor,
+        remote_address: str,
+        metadata: dict[str, Any],
+    ) -> Any:
```
ApostaC: nit: have a clear type hint using `Union[...]` rather than `Any`

ApostaC (commented):
Thanks for the update. Left some comments about the clarity (and also replied you above).

vllm/distributed/kv_transfer/kv_connector/v1/p2p/p2p_nccl_engine.py
xbfs: Hi @ApostaC, thanks for the feedback. I've updated the type hints to use Union instead of Any for both `_finalize_received_tensor` and `recv_store`, so it's clear they can return either a tensor or a tuple (addr, dtype, shape) when the buffer is full, plus None for recv_store in OOM cases. This should make the code clearer. Let me know if anything else needs tweaking.

PR #29975, opened 2025-12-03
[Fix] Support TP for ModelOpt NVFP4 by adding dynamic padding

danielafrimi:
## Summary
This PR adds dynamic padding support for NVFP4-quantized linear layers to handle tensor parallelism (TP) cases where partitioned weights don't meet kernel alignment requirements.

## Problem
When using TP > 2 with NVFP4 quantization, weight partitions may not satisfy the alignment constraints required by FlashInfer and Cutlass kernels (e.g., 32-row tile alignment, group-size block alignment). This caused failures or incorrect results.

### DIFF
--- a/vllm/model_executor/layers/quantization/modelopt.py
+++ b/vllm/model_executor/layers/quantization/modelopt.py
@@ -948,6 +948,7 @@ class ModelOptNvFp4LinearMethod(LinearMethodBase):
                 "NVFP4 quantization was selected, "
                 " dynamic quantization is not supported."
             )
+        # [1024, 1024, 256, 256, 16] is the output_partition_sizes for the modeland sum(output_partition_sizes) is 2576
         output_size_per_partition = sum(output_partition_sizes)
         weight_loader = extra_weight_attrs.get("weight_loader")
         layer.logical_widths = output_partition_sizes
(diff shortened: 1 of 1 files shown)


vllm/model_executor/layers/quantization/modelopt.py
```diff
@@ -1055,9 +1056,91 @@ def process_weights_after_loading(self, layer: Module) -> None:
...
+            # Swizzle block scales and then pad the packed NVFP4 weights so that
+            # both N (rows) and K (columns) satisfy the alignment constraints
+            # implied by the FlashInfer / Cutlass kernels.
+            swizzled_weight_scale = swizzle_blockscale(layer.weight_scale)  # [2576, 168] -> [2688, 168]
             layer.weight_scale = Parameter(swizzled_weight_scale, requires_grad=False)
-            layer.weight = Parameter(layer.weight.data, requires_grad=False)
+            weight = layer.weight.data
+
+            # Zero-pad weights in N dim if swizzled block scales have more rows
+            # than the original weights (e.g. to satisfy 32-row tile alignment).
```
pavanimajety, line 1067: Could you please share the error log where the kernels require this?

hmellor:
Closing as superseded by #31099, which landed the same NVFP4 TP padding (`pad_nvfp4_weight_for_cutlass`, `pad_nvfp4_activation_for_cutlass`, `slice_nvfp4_output`). The code this draft patches has also since been refactored into `vllm/model_executor/kernels/linear/nvfp4/`.

PR #29978, opened 2025-12-03
[Frontend][Model] Add 'float16' to possible mamba cache dtype values, override mamba SSM cache dtype value for NemotronH

amitz-nv:
## Purpose

1. Add `float16` to the set of valid values of mamba cache dtype.
2. For NemotronH models only, override the value of `mamba_ssm_cache_dtype` when it wasn't explicitly set in the CLI (has the value `auto`) with the value of `mamba_ssm_cache_dtype` field in the model's HF config, or if not set there, use float16. Or more simply, its value would be determined by the following order of precedence:
  2.1. CLI (`--mamba-ssm-cache-dtype`).
  2.2. `mamba_ssm_cache_dtype` field value in model's HF config.
  2.3. float16

## Test Plan

1. `mamba-ssm-cache-dtype` wasn't set in CLI and wasn't set in model's HF config -> it's value should be updated to "float16".
2. `mamba-ssm-cache-dtype` wasn't set in CLI and was set in model's HF config -> it's value should be updated to the value in the model's HF config.
3. `mamba-ssm-cache-dtype` was set in CLI and wasn't set in model's HF config -> it's value should be as set in the CLI.
4. `mamba-ssm-cache-dtype` was set in CLI and was set in model's HF config -> it's value should be as set in the CLI.

## Test Result

All 4 cases worked as expected.

---
<details>

</details>

### DIFF
--- a/vllm/config/cache.py
+++ b/vllm/config/cache.py
@@ -29,7 +29,7 @@ CacheDType = Literal[
     "fp8_inc",
     "fp8_ds_mla",
 ]
-MambaDType = Literal["auto", "float32", "half"]
+MambaDType = Literal["auto", "float32", "float16"]
 PrefixCachingHashAlgo = Literal["sha256", "sha256_cbor", "xxhash", "xxhash_cbor"]
 KVOffloadingBackend = Literal["native", "lmcache"]
 
--- a/vllm/config/cache.py
+++ b/vllm/config/cache.py
@@ -29,7 +29,7 @@ CacheDType = Literal[
     "fp8_inc",
     "fp8_ds_mla",
 ]
-MambaDType = Literal["auto", "float32"]
+MambaDType = Literal["auto", "float32", "half"]
 PrefixCachingHashAlgo = Literal["sha256", "sha256_cbor", "xxhash", "xxhash_cbor"]
 KVOffloadingBackend = Literal["native", "lmcache"]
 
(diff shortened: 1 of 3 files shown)


vllm/config/cache.py
```diff
@@ -29,7 +29,7 @@
     "fp8_inc",
     "fp8_ds_mla",
 ]
-MambaDType = Literal["auto", "float32"]
+MambaDType = Literal["auto", "float32", "half"]
```
hmellor: Why `half` and not `float16`?

amitz-nv: Changed to `float16`

amitz-nv: Wait, I dropped the commit that changes to `float16`, as the dict that converts `str` to `torch.dtype` contains `half` but doesn't contain `float16`: https://github.com/vllm-project/vllm/blob/9998ea5b576972a508c854227f57829dd4bca940/vllm/utils/torch_utils.py#L28

So I'll add `float16` to that dict as well and test it locally

amitz-nv: Done and verified that it works as expected @hmellor

hmellor: Thanks!

hmellor (approved):
The code change LGTM.

I'd like confirmation that the new behaviour is expected, cc @tdoublep / @benchislett maybe?

yewentao256 (approved):
LGTM, thanks for the work!

PR #29980, opened 2025-12-03
[Fix]Load kv-cache dtype from hf_quant_config.json automatically

danielafrimi:
HF checkpoint `hf_quant_config.json` (modelopt) file specifies "kv_cache_quant_algo": "FP8", but vLLM loads the unquantized version of kv-cache.

setting the vLLM server argument explicitly as below:
 "--kv-cache-dtype fp8"  makes it worked.

this PR fetched the kv_cache_quant_algo from the config itself without specifying it as args (while there is `hf_quant_config.json`).


ywang96 (commented):
PR itself makes sense to me but I'm not sure whether this should be a hf problem or vllm problem so @hmellor @mgoin please take a look

vllm/utils/torch_utils.py
```diff
@@ -194,12 +194,33 @@ def get_kv_cache_torch_dtype(
...
+            return STR_DTYPE_TO_TORCH_DTYPE[kv_algo.lower()]
+    return None
+
+
 def kv_cache_dtype_str_to_dtype(
     kv_cache_dtype: str, model_config: ModelConfig
 ) -> torch.dtype:
+    dtype = model_config.dtype if model_config else torch.half
     if kv_cache_dtype == "auto":
         # Model config may not be specified for unit tests, default to float16
```
mgoin: nit: move this comment to the top of the function

mgoin (approved):
This change is good with me, thank you. I've wanted to change the default of "auto" to be unquantized unless the model checkpoint has calibrated scales, which this achieves for modelopt. So in line with what I want!

vllm/utils/torch_utils.py
danielafrimi: done, thanks

mgoin (approved):

PR #29984, opened 2025-12-03
[ROCm][CI][Bugfix] Fix `Multi-Modal Models Test (Extended) 3` group

AndreasKaratzas:
**1. GLM-4V (`zai-org/glm-4v-9b`) - ROCm compatibility fixes in `glm4v_patch_hf_runner`:**
- Added `num_hidden_layers` config patch: `ChatGLMConfig` uses `num_layers` instead of `num_hidden_layers`, which newer transformers versions expect for `DynamicCache` initialization
- Disabled `DynamicCache` by patching `_prepare_cache_for_generation`: GLM-4V's modeling code expects the old tuple-based cache format
- Added `output_hidden_states=True` to generate and forward calls to ensure hidden states are returned for logprobs comparison

**2. GLM-4.1V (`zai-org/GLM-4.1V-9B-Thinking`) - Prompt format fix:**
- Added `[gMASK]` prefix to `prompt_formatter`: GLM-4 family models require this token at the start of prompts, otherwise they produce degenerate/repetitive output

**3. GLM-4.1V Video custom inputs - Two fixes:**
- Fixed `glm4_1v_patch_hf_runner` to filter out `do_sample_frames` from `VideoMetadata` (not a valid argument)
- Fixed `video_with_metadata_glm4_1v()` to include `[gMASK]` prefix in generated prompts

#### Root Causes
- ROCm + newer transformers versions trigger errors on several outdated models. We either skip those models, or patched them depending on how much they depended on older transformer API utilities
- GLM-4 family models require `[gMASK]` BOS token prefix for proper prompt recognition

Test command:
```bash
pip install git+https://github.com/TIGER-AI-Lab/Mantis.git && pip freeze | grep -E 'torch' && pytest -v -s models/multimodal/generation/test_common.py -m 'split(group=1) and not core_model'
```


tests/models/multimodal/generation/vlm_utils/model_utils.py
```diff
@@ -366,6 +367,40 @@ def gemma3_vllm_to_hf_output(vllm_output: RunnerOutput, model: str) -> RunnerOut
...
+            return original_generate(*args, **kwargs)
+
+        hf_model.model.generate = patched_generate
+        original_forward = hf_model.model.forward
+
+        def patched_forward(*args, **kwargs):
+            kwargs["output_hidden_states"] = True
+            return original_forward(*args, **kwargs)
+
+        hf_model.model.forward = patched_forward
```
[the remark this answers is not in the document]
AndreasKaratzas: This is a patch for a HuggingFace test to run. I don't think we should emphasize on our platform specific patch, which has nothing to do with vLLM, but only the transformers API and for that specific model.

tests/models/multimodal/generation/vlm_utils/model_utils.py
```diff
@@ -366,6 +367,40 @@ def gemma3_vllm_to_hf_output(vllm_output: RunnerOutput, model: str) -> RunnerOut
...
+            kwargs["output_hidden_states"] = True
+            kwargs["return_dict_in_generate"] = True
+            return original_generate(*args, **kwargs)
+
+        hf_model.model.generate = patched_generate
+        original_forward = hf_model.model.forward
+
+        def patched_forward(*args, **kwargs):
+            kwargs["output_hidden_states"] = True
+            return original_forward(*args, **kwargs)
```
Isotr0py, line 400: I think these `kwargs` can be directly passed through `hf_model_kwargs={...}` in `VLMTestInfo` to avoid adding these monkey patches.

AndreasKaratzas: I think, only `output_hidden_states=True` could be moved to `hf_model_kwargs` (as a config-level setting). I think the other patches require runtime modification:

- `_prepare_cache_for_generation` override to disable caching
- `generate`/`forward` patches to inject kwargs at call time (not construction time)

Also, these patches are workarounds for the HuggingFace transformers API on ROCm, not vLLM code. And since the purpose of this test is to verify vLLM functionality (using HF as a reference), I'd prefer not to invest too much effort into refactoring the HF runner workarounds. I kept them grouped together for clarity since they're platform-specific and outside the main scope of what we're testing.

PR #29986, opened 2025-12-03
[CI/Build] Add MM code path to Examples Test

zhewenl:
## Purpose
#29621 broke Examples Test on CI, and we don't have MM code path currently so the test suite is fragile - this PR adds MM code path coverage to the test suite, so that changes can be caught

### DIFF
--- a/.buildkite/test-pipeline.yaml
+++ b/.buildkite/test-pipeline.yaml
@@ -387,6 +387,7 @@ steps:
   working_dir: "/vllm-workspace/examples"
   source_file_dependencies:
   - vllm/entrypoints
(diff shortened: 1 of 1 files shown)


.buildkite/test-pipeline.yaml
zhewenl, line 401: I feel like maybe we also considering spitting MM to something like Examples Test(Multimodal)

DarkLight1337 (approved):

.buildkite/test-pipeline.yaml
DarkLight1337: I think right now the test isn't long enough to justify splitting it

PR #29987, opened 2025-12-03
[BugFix] Eagerly abort cancelled final-step requests

njhill:
Currently, when requests are cancelled while executing their final step, "completion" of those requests is subsequently handled based on normal stop processing (e.g. length or stop token), and so the abort essentially has no effect.

This is typically not a problem since the final output would be ignored/discarded in this case anyhow. When a kv connector is involved however, it means that the connector will think the request completed successfully rather than being aborted.

This has turned out to be problematic for disaggregated prefill which will free the kv cache blocks if the request was aborted but not if it thinks the request has completed successfully. Since the top-level request was cancelled, it will never be sent to the decode side and so the kv cache blocks remain pinned unnecessarily until the fall-back timeout expires.

The problem is exacerbated when a large number of requests are cancelled and/or there are large prefills whose forward pass takes a long time (since the window for this to occur is bigger).

---

This PR fixes the problem by explicitly processing any pending aborts immediately prior to processing the model output each step. We process only the aborts and not new requests since it's still preferable for latency reasons to process the model outputs before new incoming requests.

Fixes https://github.com/vllm-project/vllm/issues/26400.


robertgshaw2-redhat:
Could you provide some more detailed explanation about what was happening before + why this fixes it?

This is pretty complicated logic so I think we will value the posterity

vllm/v1/worker/gpu_worker.py
```diff
@@ -51,14 +51,14 @@
...
 from vllm.v1.utils import report_usage_stats
-from vllm.v1.worker.gpu_model_runner import GPUModelRunner
 from vllm.v1.worker.utils import is_residual_scattered_for_sp
 from vllm.v1.worker.worker_base import WorkerBase
 
 logger = init_logger(__name__)
 
 if TYPE_CHECKING:
     from vllm.model_executor.model_loader.tensorizer import TensorizerConfig
+    from vllm.v1.worker.gpu_model_runner import GPUModelRunner
```
njhill, line 61: The changes in this file avoid cuda being initialized at import time (lazy import of `GPUModelRunner`)

hasB4K:
Hello @njhill! I tried your PR and it seems to work as expected. Now that I think about it my PR was more a patch than a proper fix. Yours is the proper fix. I'll let you know if we encounter this issue again but hopefully not 😉 

Any chance to see this PR merged quickly?

vllm/v1/engine/core.py
```diff
@@ -871,9 +891,13 @@ def _process_input_queue(self):
             and not self.scheduler.has_requests()
             and not self.batch_queue
         ):
-            if logger.isEnabledFor(DEBUG) and self.input_queue.empty():
-                logger.debug("EngineCore waiting for work.")
-                waited = True
+            if self.input_queue.empty():
+                # Drain aborts queue; all aborts are also processed via input_queue.
+                with self.aborts_queue.mutex:
+                    self.aborts_queue.queue.clear()
```
markmc, line 897: This all makes me a bit nervous - is there a race here where we could completely lose an abort ? (No, I don't think so, but only after some thought)

And the `Queue.deque` thing doesn't seem to be a public, stable API?

Would it really be that much overhead to just process each abort a second time, since's it's just a `Scheduler.requests` lookup?

(I'm torn)

tests/v1/engine/test_abort_final_step.py
```diff
@@ -0,0 +1,311 @@
...
+                        for line in status_lines
+                        if line and line.startswith("FINISHED_")
+                    ]
+
+                assert len(captured_statuses) >= 1, (
+                    f"Expected at least 1 captured finish status, got "
+                    f"{len(captured_statuses)}. File content: {status_lines}"
+                )
+
+                assert "FINISHED_ABORTED" in captured_statuses, (
```
markmc, line 292: Not a big deal, but why not assert that `FINISHED_ABORTED` is the only one received?

markmc (approved):
Thanks, Nick. Looks good to me.

As discussed, I was trying to quantify the size of the window this closes, versus the window remaining

Here's claude's take on that, FWIW:

- Window CLOSED by this fix (Large) - Chunked prefill: Hundreds of milliseconds to several seconds
- Window REMAINING (Tiny) - Duration: Pure Python operations, no GPU wait, sub-millisecond (microseconds in practice)

vllm/v1/engine/core.py
njhill: I don't think we can lose an abort since the aborts are also added to the input queue. I avoided only adding to abort queue since then we might process and abort out of order if it happened immediately after an add request.

I think it may actually be safer to drain the abort queue here since we could otherwise in theory abort a new request with the same id that was received _after_ the abort was issued.

PR #29990, opened 2025-12-03
[Feature] Add Layer-wise NVTX Support

maxyanghu:
## Purpose

This PR adds an optional server arg `--enable-layerwise-nvtx-tracing`. When enabled it uses forward hooks to trace detailed Pytorch module information such as input/output shapes, module name etc. This is especially useful when the users want to correspond CUDA kernels with Pytorch modules/model layers in Nsight systems.

The majority of the implementation is borrowed from this [PR](https://github.com/sgl-project/sglang/pull/11870). With some tweaks for it to support vLLM's torch compiled modules.

## Modifications
- Add pre/post forward hooks to a model to trace modules
- Add nvtx push/pop before torch.compile wrapper to trace torch-compiled modules
- This feature doesn't work for regions captured with CUDA graphs
- This feature doesn't work for **STOCK_TORCH_COMPILE** mode as nvtx operations cause graph break
- Torch compiled regions appear as a single module

## Usage
```
VLLM_TORCH_CUDA_PROFILE=1 nsys profile --wait primary -f true -o qwen3_0.6B_profile_mode_0_cg_0 --trace-fork-before-exec=true --cuda-graph-trace=node --capture-range=cudaProfilerApi --capture-range-end repeat --trace cuda,nvtx vllm serve Qwen/Qwen3-0.6B --compilation-config '{"mode": 0, "cudagraph_mode": 0}' --enable-layerwise-nvtx-tracing --host 0.0.0.0 --port 56563 --async-scheduling --max-num-seqs 1024
```
and then:
```
vllm bench serve --profile --backend openai-chat --base-url http://0.0.0.0:56563 --endpoint /v1/chat/completions --model Qwen/Qwen3-0.6B --dataset-name random --random-input-len 10 --random-output-len 10 --num-prompts 1
```
This generates a nested nvtx module stack with detailed module infos:
image

This also supports Compiled mode **DYNAMO_TRACE_ONCE** and **VLLM_COMPILE**, that is when `--compilation-config '{"mode": 2 or 3, "cudagraph_mode": 0}'`. This yields something like:
image

---
<details>

</details>


vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2998,6 +3000,36 @@ def execute_model(
             # Mark KV scales as calculated after the first forward pass
             self.calculate_kv_scales = False
 
+        if (
+            self.vllm_config.observability_config.enable_layerwise_nvtx_tracing
+            and not self.layerwise_nvtx_hooks_registered
+        ):
+            if cudagraph_mode != CUDAGraphMode.NONE:
+                logger.debug(
```
benchislett: Is `debug_once` an option here?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2998,6 +3000,36 @@ def execute_model(
             # Mark KV scales as calculated after the first forward pass
             self.calculate_kv_scales = False
 
+        if (
```
benchislett: Please refactor this out of `execute_model` and into a helper function. We want to keep critical code pathways brief and readable.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2998,6 +3000,36 @@ def execute_model(
...
+            ):
+                logger.debug(
+                    "layerwise NVTX tracing is not supported when "
+                    "CompilationMode is STOCK_TORCH_COMPILE, skipping "
+                    "function hooks registration"
+                )
+            else:
+                pyt_hooks = PytHooks()
+                pyt_hooks.register_hooks(self.model, self.model.__class__.__name__)
+                self.layerwise_nvtx_hooks_registered = True
```
benchislett: Why not just run this code once at the end of `load_model`? I feel like there has to be a better way to do this than having a dedicated flag for everything we preload based on the model

vllm/utils/nvtx_pytorch_hooks.py
```diff
@@ -0,0 +1,251 @@
...
+                if tensor_list:
+                    marker_dict[key] = tensor_list
+
+    param_info = process_layer_params(module_obj)
+    if param_info:
+        marker_dict["StaticParams"] = param_info
+    return marker_dict
+
+
+def nvtx_range_push(marker_dict):
```
benchislett: I think this could be implemented more cleanly than having these trivial helper functions

benchislett (commented):
A few comments, in general we should avoid adding diffs to execute_model and other core functions unless strictly necessary. I think this feature should be refactored to be more self-contained.

vllm/v1/worker/gpu_model_runner.py
maxyanghu: The main reason is nvtx operations inside hook functions trigger graph breaks if we do the registration in `load_model()` before dynamo traces the model.

vllm/v1/worker/gpu_model_runner.py
maxyanghu: fixed

vllm/v1/worker/gpu_model_runner.py
maxyanghu: fixed

vllm/v1/worker/gpu_model_runner.py
maxyanghu: I've moved the registration inside `_dummy_run()`

vllm/utils/nvtx_pytorch_hooks.py
maxyanghu: modified implementation with a context manager to achieve more encapsulation.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -4125,6 +4158,18 @@ def _dummy_run(
                     is_graph_capturing=is_graph_capturing,
                 )
 
+        # Register layerwise NVTX hooks if --enable-layerwise-nvtx-tracing is enabled.
```
benchislett: Please move this into the docstring for `_register_layerwise_nvtx_hooks`.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -598,6 +599,7 @@ def __init__(
         # Ephemeral state transferred between execute_model() and sample_tokens().
         self.execute_model_state: ExecuteModelState | None = None
         self.kv_connector_output: KVConnectorOutput | None = None
+        self.layerwise_nvtx_hooks_registered = False
```
benchislett, line 603: Is this guard still needed? Is the `dummy_run` ever run more than once?

maxyanghu: yes, it also runs in [this warmup function](https://github.com/vllm-project/vllm/blob/c8ab988b15af5e30e87c6eb27a0ededf0377ac9e/vllm/v1/worker/gpu_worker.py#L405). Without this guard multiple hooks could be registered to the module and results in an incorrect profile

vllm/v1/worker/gpu_model_runner.py
maxyanghu: moved

benchislett (approved):
LGTM

ywang96:
@benchislett Let's not forget to add ready label and automerge next time after you approve a PR - thanks 😄

PR #29991, opened 2025-12-03
[Frontend] Remove deprecated -O.xx flag

gmagogsfm:
Remove deprecated -O.* dotted syntax for --compilation-config. Users now an error directing them to use -cc.* instead.

- -O{0,1,2,3} style flags are still supported.

### DIFF
--- a/vllm/utils/argparse_utils.py
+++ b/vllm/utils/argparse_utils.py
@@ -244,9 +244,15 @@ class FlexibleArgumentParser(ArgumentParser):
                 else:
                     key = pattern.sub(repl, arg, count=1)
                     processed_args.append(key)
-            elif arg.startswith("-O") and arg != "-O" and arg[2] != ".":
+            elif arg.startswith("-O."):
+                # Provide clear error for deprecated -O.* syntax
+                raise ValueError(
+                    f"The -O.* syntax is no longer supported. "
+                    f"Please use -cc.* instead. "
+                    f"For example, replace '{arg}' with '{arg.replace('-O', '-cc', 1)}'"
+                )
+            elif arg.startswith("-O") and arg != "-O":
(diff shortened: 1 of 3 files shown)


vllm/utils/argparse_utils.py
```diff
@@ -244,9 +244,15 @@ def repl(match: re.Match) -> str:
                 else:
                     key = pattern.sub(repl, arg, count=1)
                     processed_args.append(key)
-            elif arg.startswith("-O") and arg != "-O" and arg[2] != ".":
+            elif arg.startswith("-O."):
```
ProExpertProg, line 247: @hmellor do we want to keep this around or just remove it completely? If we do, we should plan to completely rip it out at some point to avoid cruft staying in the code

ProExpertProg (approved):
Looks good overall, just a comment for the detailed error message

vllm/utils/argparse_utils.py
gmagogsfm: @hmellor Would appreciate your opinion here. There is a bit of urgency since the previous deprecation warning said the -O.xx behavior would be removed in v0.13, which I believe will be cut in about 10 days. Thanks.

hmellor: I'm conflicted. I've always thought it'd be nice to leave an error like this when removing deprecations. And I agree that they shouldn't stay around forever if they do exist.

In this case let's just rip it out because it's too similar to `-On` which might confuse people. e.g. if a user does `-O.2` thinking they're setting optimisation level but then they see an error suggesting they do `-cc.2` which also won't work.

gmagogsfm: Looks like this PR is already merged, so I made another one for removing the error: 

https://github.com/vllm-project/vllm/pull/30169

PR #29995, opened 2025-12-03
[CI/Build][AMD] Skip test on test_hybrid_attention_mamba_tensor_shapes on ROCm, requires FLASHINFER

rasmith:
This PR skips `test_hybrid_attention_mamba_tensor_shapes`  in `test_gpu_model_runner.py` since it uses `FLASHINFER `backend which is not supported on ROCm.


tjtanaa (approved):
LGTM, but we have to come back again to properly enable the test for this backend.

PR #28774, opened 2025-11-15
[BugFix] Corner case that could cause out-of-sync with external launcher mode and dp >1

bangshengtang:
Summary:
we observed a comms issue when both async scheduling and DP>1 are enabled

when async scheduling is enabled, num_scheduled_tokens can become zero, and DP>1 requires coordinate_batch_across_dp to be called into where an all_reduce across dp ranks is required. returning early would cause dp ranks go out-of-sync

basically we got
```
terminate called after throwing an instance of 'gloo::EnforceNotMet'
  what():  [enforce fail at /mnt/code/pytorch/third_party/gloo/gloo/transport/tcp/pair.cc:454] op.preamble.length <= op.nbytes. 8 vs 4
I1114 16:43:23 native_sampler.py:491 1199692:MainThread] rank=2: has_unfinished_requests
W1114 16:43:23.344583 1233050 ExceptionTracer.cpp:193] Invalid trace stack for exception of type: gloo::EnforceNotMet
terminate called after throwing an instance of 'gloo::EnforceNotMet'
  what():  [enforce fail at /mnt/code/pytorch/third_party/gloo/gloo/transport/tcp/pair.cc:454] op.preamble.length <= op.nbytes. 8 vs 4
```
when dp_rank 0 is calling into has_unfinished_requests(), while dp_rank 1 is waiting at coordinate_batch_across_dp()

and log indicate that this happens in the previous step dp_rank 0 returned early with num_scheduled_tokens=0

Test Plan:
with the change, our workload runs for multiple ours without triggering this issue.

**before**
```
INFO 11-14 19:28:01 [gpu_model_runner.py:2533] rank=2, num_scheduled_tokens=0
INFO 11-14 19:28:01 [gpu_model_runner.py:2533] rank=3, num_scheduled_tokens=0
[...]

PR #29996, opened 2025-12-03
[BugFix] Adding env variable to disable async grammar compilation

alecsolder:
## Purpose
This PR provides good context on the situation where this issue may come up [#28774](https://github.com/vllm-project/vllm/pull/28774).

From what I understand, when using external launcher mode and async scheduling, dp ranks can get out of sync. This required the introduction of a dummy run to get them back in sync. 

By default, grammar compilation for guided decoding is done async for vLLM. When grammar compilation is happening, the request is in [WAITING_FOR_FSM state](https://github.com/xieyangxu/vllm/blob/457ff3a8dd18d441ef0015ac4859689df4596432/vllm/v1/core/sched/scheduler.py#L424)  which will cause the scheduler to ignore the request until compilation is finished. While waiting for it to finish, num_scheduled_tokens would be 0. This causes inconsistencies with which ranks trigger a dummy run, and can cause deadlocks waiting for an allreduce. 

This PR has no impact on vLLM unless the env variable is set.

## Test Plan
Tested in external launcher mode with different parallelism configs in internal runs. Test cases added to verify env variable functions properly.

## Test Result
Without this change runs deadlock consistently on the first request.
With this change runs no longer deadlock at all.


alecsolder:
> Is there a configuration where we would recommend turning this on? Is that something that should be documented?
> 
> Even better, is this something we can detect? If we know we're in a config where it should be on, we could just do it instead of requiring the env var.
> 
> Or is this really a workaround for a bug we haven't figured out a complete fix for, so this is a temporary (we hope) workaround?

IMO I think we don't know enough yet to have this be configured automatically. We could consider making this part of the guided decoding configurations instead of an env variable, but I think we would want to confirm that this is a general issue first in that case.

vllm/v1/structured_output/__init__.py
```diff
@@ -137,11 +138,15 @@ def grammar_init(self, request: Request) -> None:
                 )
             else:
                 raise ValueError(f"Unsupported structured output backend: {backend}")
+        # Async grammar creation can cause deadlocks due to grammar compilation
+        # impacting scheduling across ranks from RequestStatus.WAITING_FOR_FSM state
+        if envs.VLLM_ASYNC_CREATE_GRAMMAR:
+            grammar = self.executor.submit(self._create_grammar, request)
+            request.structured_output_request.grammar = grammar  # type: ignore[assignment]
+        else:
+            request.structured_output_request.grammar = self._create_grammar(request)
```
22quinn:
```suggestion
        else:
            grammar = self._create_grammar(request)
        request.structured_output_request.grammar = grammar  # type: ignore[assignment]
```

zhuohan123 (approved):

njhill (commented):
Thanks @alecsolder I understand now. I kind of agree with @russellb though that it would then be better for this to be controlled based on whether external launcher is in use (along with similar explanation in the code), rather than introducing another env var.

vllm/v1/structured_output/__init__.py
```diff
@@ -137,11 +141,18 @@ def grammar_init(self, request: Request) -> None:
                 )
             else:
                 raise ValueError(f"Unsupported structured output backend: {backend}")
-
-        grammar = self.executor.submit(self._async_create_grammar, request)
+        # When in external_launcher mode, async grammar compilation causes deadlocks
+        # due to external_launcher mode having a scheduler for each TP rank.
+        # Async grammar compilation causes the WAITING_FOR_FSM → WAITING transition to
+        # happen at different times on different TP ranks,
+        # breaking the determinism assumption that external_launcher relies on.
```
njhill: nit: I think it makes more sense to move this comment up to where `self_use_async_grammar_compilation` is set based on the executor backend

njhill (commented):
Thanks @alecsolder .. one small thing then lgtm

PR #30001, opened 2025-12-03
[Rocm][CI] Fix test_speculator_eagle3 by skipping the CompressedTensorw4a16 Model

charlifu:
The unit test of nm-testing/Speculator-Qwen3-8B-Eagle3-converted-071-quantized-w4a16 is failing on Rocm.

By digging into the error, we found it is caused by the illegal memory access from CompressedTensorsWNA16, which calls the mixed_precision_gemm from conch.

And we also found the [unit test for CompressedTensorsWNA16](https://github.com/vllm-project/vllm/blob/main/tests/quantization/test_compressed_tensors.py#L283) is skipping all non-cuda platform, so we suppose this model should also be skipped for rocm.
{D33E1910-92E3-434A-90C0-EDB6EDA86B10}


tjtanaa (approved):
LGTM

PR #17010, opened 2025-04-22
[V1][Metrics] Add API for accessing in-memory Prometheus metrics

markmc:
The V0 LLM offline inference API exposes per-request metrics via `RequestOutput.RequestMetrics`. In V1, so far we have chosen to not track per-request metrics or implement this API.

All recent work implementing EAGLE have been using `examples/offline_inference/eagle.py` which depends on these metrics to report an aggregated mean acceptance length number.

See e.g.  the EAGLE3 PR #16937 which used a WIP implementation of the per-request metrics #16367

The proposal in this PR is to achieve the same aggregated view by using the Prometheus metrics already implemented for the online serving case. This will mean we automatically gain new spec decoding metrics in #16665 for both offline and online inferencing

This does not preclude us from implementing per-request metrics in future in V1 if that proves to be important.

See also [the spec decoding metrics design doc](https://docs.google.com/document/d/1xoxP9X9Zt6wkpIhErUcI7g0XtiFtKkvu1Kjaxi_uJqU).

PR #30028, opened 2025-12-04
[docs] Remove _total from counter metrics names

googs1025:
related pr: #17010

In Prometheus Counters always expose their actual numeric value with a metric name that ends in `_total`. We should document the base name, as this what appears in the `get_metrics()` API.


markmc (approved):
Thanks!

PR #30033, opened 2025-12-04
[Bugfix] Fix the issue with interleaved thinking when using streaming

chaunceyjiang:
## Purpose
Fix the issue with interleaved thinking when using streaming.

```
｜begin▁of▁sentence｜>

## Tools
You have access to a set of tools you can use to answer the user's question.

<think>...thinking about results</think>

The current date is: 2024-09-30<｜User｜> What is the weather like in 
Beijing today? And tomorrow? <｜Assistant｜><think>


```

When the user's prompt contains something like `<think>...</think> ... <think>`, `is_reasoning_end` incorrectly determines that reasoning has already ended. In reality, because the last token is `<think>`, the reasoning has just begun. This causes reasoning to fail to work properly.

Especially in interleaved thinking mode, where the chat_template renders a `<think>...</think> ... <think>` prompt with 100% consistency.

## Test Result

---
<details>

</details>

### DIFF
--- a/tests/reasoning/test_base_thinking_reasoning_parser.py
+++ b/tests/reasoning/test_base_thinking_reasoning_parser.py
@@ -112,7 +112,7 @@ class TestBaseThinkingReasoningParserMethods:
         """Test the is_reasoning_end method."""
         parser = TestThinkingReasoningParser(test_tokenizer)
         end_token_id = parser.end_token_id
-
+        start_token_id = parser.start_token_id
         # Test with end token present
         assert parser.is_reasoning_end([1, 2, end_token_id, 4]) is True
 
@@ -121,6 +121,16 @@ class TestBaseThinkingReasoningParserMethods:
 
         # Test with empty list
         assert parser.is_reasoning_end([]) is False
+
+        # Test with interleaved thinking
+        assert parser.is_reasoning_end([1, start_token_id, 2, end_token_id]) is True
+        assert parser.is_reasoning_end([1, start_token_id, 2, 3]) is False
+        assert (
+            parser.is_reasoning_end(
+                [1, start_token_id, 2, end_token_id, 2, 2, start_token_id]
+            )
+            is True
+        )
 
     def test_extract_content_ids(self, test_tokenizer):
         """Test the extract_content_ids method."""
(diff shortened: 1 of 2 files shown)


vllm/reasoning/basic_parsers.py
```diff
@@ -64,8 +64,12 @@ def __init__(self, tokenizer: TokenizerLike, *args, **kwargs):
...
 
     def is_reasoning_end(self, input_ids: list[int]) -> bool:
-        end_token_id = self.end_token_id
-        return any(input_id == end_token_id for input_id in reversed(input_ids))
+        for i in range(len(input_ids) - 1, -1, -1):
+            if input_ids[i] == self.start_token_id:
+                return False
+            if input_ids[i] == self.end_token_id:
+                return True
+        return False
```
DarkLight1337, line 75:
```suggestion
        start_token_id = self.start_token_id
        end_token_id = self.end_token_id

        for i in range(len(input_ids) - 1, -1, -1):
            if input_ids[i] == start_token_id:
                return False
            if input_ids[i] == end_token_id:
                return True
        return False
```

Avoid repeating accessing `self`

chaunceyjiang: Done.

DarkLight1337 (approved):

tests/reasoning/test_base_thinking_reasoning_parser.py
```diff
@@ -122,6 +122,16 @@ def test_is_reasoning_end(self, test_tokenizer):
...
         assert parser.is_reasoning_end([]) is False
 
+        # Test with interleaved thinking
+        assert parser.is_reasoning_end([1, start_token_id, 2, end_token_id]) is True
+        assert parser.is_reasoning_end([1, start_token_id, 2, 3]) is False
+        assert (
+            parser.is_reasoning_end(
+                [1, start_token_id, 2, end_token_id, 2, 2, start_token_id]
+            )
+            is True
```
DarkLight1337: Actually shouldn't this be false?

chaunceyjiang: good catch!!!!

PR #18354, opened 2025-05-19
[V1][Metrics] Deprecate metrics with gpu_ prefix for non GPU specific metrics.

sahelib25:
This PR deprecates metrics with `gpu_` prefix for existing non-GPU specific metrics-

- `gpu_cache_usage`
-  `gpu_prefix_cache_queries` 
- `gpu_prefix_cache_hits`

and, introduce new metrics after renaming-

- `kv_cache_usage`
- `prefix_cache_queries`
- `prefix_cache_hits`

PR #30041, opened 2025-12-04
docs: update metrics design doc to use new vllm:kv_cache_usage_perc

haitwang-cloud:
In #18354, the `vllm:gpu_cache_usage_perc` metric was deprecated and `vllm:kv_cache_usage_perc` was introduced as its replacement.

See also #27133


markmc (approved):
Thank you

PR #30049, opened 2025-12-04
Delete HF version of Phi 4 MM

hmellor:
Phi 4 multimodal is currently compatible with vLLM in 2 ways:

- `Phi4MMForCausalLM` - reimplemented in vLLM based on Microsoft's custom code from https://huggingface.co/microsoft/Phi-4-multimodal-instruct
- `Phi4MultimodalForCausalLM` - reimplemented in vLLM based on Huigging Face Transformers (requires https://huggingface.co/microsoft/Phi-4-multimodal-instruct/discussions/70)

Given that:

- Microsoft appears to have no intention of merging the PR
- The PR for the Transformers version is no longer compatible with vLLM on the latest Transformers (v5)

We have decided to drop support for the Transformers implementation in vLLM.

This means that:

- Using Phi 4 multimodal in vLLM will always require `--trust-remote-code`
- Phi 4 multimodal will eventually become incompatible with vLLM as its custom code rots


DarkLight1337 (approved):
cc @Isotr0py

tests/models/registry.py
```diff
@@ -767,10 +767,6 @@ def check_available_online(
     "Phi4MMForCausalLM": _HfExamplesInfo(
         "microsoft/Phi-4-multimodal-instruct", trust_remote_code=True
     ),
-    "Phi4MultimodalForCausalLM": _HfExamplesInfo(
-        "microsoft/Phi-4-multimodal-instruct",
-        revision="refs/pr/70",
-    ),
```
Isotr0py, line 773: Let's also remove `tests/models/multimodal/generation/test_phi4_multimodal.py`

Isotr0py (approved):
BTW, I wonder if we can create a separated community repo like `mistral-community/pixtral-12b`/`deepseek-community/deepseek-vl-7b-chat` to hold HF format model instead of converting existing repo?

PR #30050, opened 2025-12-04
[Misc][PCP&DCP] relocate PCP feature check

pisceskkk:
Recently, vllm-ascend has been advancing the main-to-main adaptation work with the vllm repository. We have now fully adapted PCP and other features on vllm-ascend, so we believe it is necessary to add essential feature adaptation flags for the attention backend and modify the logic of the PCP feature check code to ensure that the functional switches for GPU and NPU do not interfere with each other.

Modifications:​
- Consolidate all PCP-related feature check code in the same location as DCP checks. As backend functionalities are progressively completed, the relevant code can be deleted based on the situation.
- Add feature adaptation flags for the attention backend. Subsequent PRs for backend adaptation can enable or disable the corresponding flags based on development progress.

CC @LucasWilkinson @FENP @LookAround0301 @zhenwenqi2024


vllm/v1/worker/gpu_model_runner.py
```diff
@@ -5296,19 +5299,50 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
...
+                    ):
+                        self.scheduler_config.enable_chunked_prefill = False
+                        self.cache_config.enable_prefix_caching = False
+                        logger.warning_once(
+                            "Prefill context parallel (PCP) is enabled, "
+                            "which is incompatible with chunked prefill "
+                            f" and prefix caching in {layer_impl.__class__.__name__}. "
+                            "Disabling chunked prefill and prefix caching.",
+                            scope="local",
+                        )
```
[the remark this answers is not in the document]
pisceskkk: fixed

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -5296,19 +5299,50 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
...
+                        "PCP requires attention impls' support, "
+                        f"but the impl {layer_impl.__class__.__name__} "
+                        "does not support PCP."
+                    )
+                    if (
+                        self.speculative_config is not None 
+                        and self.parallel_config.cp_kv_cache_interleave_size > 1
+                    ):
+                        assert layer_impl.support_mtp_for_non_token_interleave, (
+                            "MTP with cp_kv_cache_interleave_size > 1 is not "
```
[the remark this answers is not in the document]
pisceskkk: fixed

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -5296,19 +5299,48 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
                 kv_transfer_group.register_kv_caches(kv_caches)
             kv_transfer_group.set_host_xfer_buffer_ops(copy_kv_blocks)
 
-        if self.dcp_world_size > 1:
+        if self.pcp_world_size * self.dcp_world_size > 1:
```
LucasWilkinson: These checks have gotten very long and are starting to pollute `initialize_kv_cache`; these should be moved into a separate `_check_attention_cp_compatibility` function or something like that

vllm/attention/backends/abstract.py
```diff
@@ -285,6 +285,12 @@ class AttentionImpl(ABC, Generic[T]):
     # Some features like decode context parallelism require the softmax lse.
     can_return_lse_for_decode: bool = False
 
+    # Whether the attention impl supports Prefill Context Parallelism
+    # and is compatible with other features.
+    support_pcp: bool = False
+    support_chunked_prefill_for_pcp: bool = False
+    support_mtp_for_non_token_interleave: bool = False
```
LucasWilkinson: please add docs since theres no example usage here; example I have no idea what `support_mtp_for_non_token_interleave` means

prefer `supports` to `support` e.g. `supports_pcp` v.s. `support_pcp`

do we really need `support_chunked_prefill_for_pcp`? I think we should consider forcing all backends that support PCP to support this given its needed for prefix-caching and chunked-prefill both things that are first class citizens in vLLM

LucasWilkinson (changes requested):

vllm/v1/worker/gpu_model_runner.py
pisceskkk: Yes, I moved this code into a new function `_check_attention_cp_compatibility` and placed it in `WorkerWrapperBase` because I believe this check should be platform-independent.  
However, I noticed that `WorkerWrapperBase` is relatively lean—I’m not sure if placing this larger function here is appropriate. Do you have any suggestions on this?

vllm/attention/backends/abstract.py
pisceskkk: add docs and rename these flags.
And I think that forcing all backends with PCP support prefix-caching and chunked-prefill makes sense, so remove this flag.

vllm/v1/worker/worker_base.py
```diff
@@ -319,6 +321,41 @@ def initialize_from_config(self, kv_cache_configs: list[Any]) -> None:
         with set_current_vllm_config(self.vllm_config):
             self.worker.initialize_from_config(kv_cache_config)  # type: ignore
 
+    def _check_attention_cp_compatibility(self) -> None:
```
LucasWilkinson: we should move this to `vllm/v1/worker/cp_utils.py` (you can create it); and call it from the model runner :+1:

vllm/attention/backends/abstract.py
```diff
@@ -285,6 +285,12 @@ class AttentionImpl(ABC, Generic[T]):
     # Some features like decode context parallelism require the softmax lse.
     can_return_lse_for_decode: bool = False
 
+    # Whether the attention impl supports Prefill Context Parallelism.
+    supports_pcp: bool = False
+    # Whether the attention impl(or ops) supports MTP
+    # when cp_kv_cache_interleave_size > 1
+    supports_mtp_for_non_token_interleave: bool = False
```
LucasWilkinson: Thanks for adding the the comments!

nit: `supports_mtp_with_cp_non_trivial_interleave_size` is probably a clearer name

vllm/v1/worker/worker_base.py
pisceskkk: thanks! I create `cp_utils.py` and call it in `GPUModelRunner.initialize_attn_backend()`.

vllm/attention/backends/abstract.py
pisceskkk: replace with the clearer name

LucasWilkinson (approved):
LGTM

PR #30051, opened 2025-12-04
[bugfix] fix type[AttentionBackend] bug in kv_connector_base_v1

HF-001:
## Purpose
fix the bug in KVConnectorBase_v1, change type[AttentionBackend] to  type["AttentionBackend"]


LucasWilkinson (approved):
LGTM

PR #25735, opened 2025-09-26
 [Frontend] Improve the performance of `is_reasoning_end`

chaunceyjiang:
## Purpose
~~`is_reasoning_end` is executed once for every token generated during the reasoning phase, which results in poor performance.~~
~~Optimize it to perform incremental checks instead.~~

Currently, `is_reasoning_end` cannot support incremental checking for the following reason:

When `stream=true`, `is_reasoning_end` checks `output.token_ids` and `res.prompt_token_ids` separately. This leads to non-deterministic inputs to `is_reasoning_end` within the same request, making incremental checking impossible.

Therefore, the only feasible optimization at present is to modify `is_reasoning_end` to search backward for the `end_token` from the end of the token sequence.

## Test Plan
```
vllm serve /home/jovyan/qwen3-8b  --reasoning-parser qwen3 --guided-decoding-backend xgrammar --enable-auto-tool-choice --tool-call-parser hermes --no-enable-prefix-caching
```
```
vllm bench serve \   
  --backend vllm \
  --model /home/jovyan/qwen3-8b \
  --served-model-name /home/jovyan/qwen3-8b \
  --endpoint /v1/completions \
  --dataset-name random \
  --random-input 2048 \
  --random-output 1024 \
  --max-concurrency 10 \
  --num-prompt 100
```
main
```
============ Serving Benchmark Result ============
Successful requests:                     100       
Maximum request concurrency:             10        
Benchmark duration (s):                  95.70     
Total input tokens:                      204800    
Total generated tokens:                  96051     
Request throughput (req/s):              1.04      
Output token throughput (tok/s):         1003.69   
Peak output token throughput (tok/s):    1150.00   
Peak concurrent requests:                20.00     
Total Token throughput (tok/s):          3143.74   
---------------Time to First Token----------------
Mean TTFT (ms):                          311.72    
Median TTFT (ms):                        282.02    
P99 TTFT (ms):                           525.12    
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          9.56      
Median TPOT (ms):                        9.17      
P99 TPOT (ms):                           11.46     
---------------Inter-token Latency----------------
Mean ITL (ms):                           9.13      
Median ITL (ms):                         8.88      
P99 ITL (ms):                            10.55     
==================================================
```

this pr
```
============ Serving Benchmark Result ============
Successful requests:                     100       
Maximum request concurrency:             10        
Benchmark duration (s):                  95.65     
Total input tokens:                      204800    
[...]

PR #30056, opened 2025-12-04
[Structured Output][Reasoning] Improves decoding throughput for models using single-token reasoning endings.

hdlj-h:
## Purpose

This PR introduces `is_reasoning_end_streaming` to the `ReasoningParser` interface and implements it in all reasoning parsers and uses it in the StructuredOuputManager. 

Previously, `is_reasoning_end` checked for the reasoning end token at every decode step by scanning 
all `input_ids` for models that relies on a single token for end reasoning. 

```python
def is_reasoning_end(self, input_ids: list[int]) -> bool:
    end_token_id = self.end_token_id
    return any(input_id == end_token_id for input_id in reversed(input_ids))
```

This was called at every decoding step in the StructuredOutputManager and could be inefficient for models that end reasoning with a single token (long reasoning + structured output)

The new method checks only the last token of the current decoding step (**BaseThinkingReasoningParser**)

```python
def is_reasoning_end_streaming(
        self, input_ids: list[int], delta_ids: list[int]
    ) -> bool:
        # We only check the end token beause interleaved reasoning and content is not
        # compatible with the current structured output manager.
        end_token_id = self.end_token_id
        return end_token_id in delta_ids
```

For multi tokens end reasoning (like gptoss), this PRs preserve the previous behaviour by calling `is_reasoning_end` and doesn't bring improvements. 

## Additional benefits from [30033](https://github.com/vllm-project/vllm/pull/30033)

During the implementation of this PR, [#30033][Fix the issue with interleaved thinking when using streaming](https://github.com/vllm-project/vllm/pull/30033)￼ was merged, which already brought a significant improvement in output token throughput.

Even though is_reasoning_end still checks the full input, handling interleaved reasoning requires returning False if the last reasoning token is **\<think\>**. Many chat templates end with **\<think\>**, which is why [#30033](https://github.com/vllm-project/vllm/pull/30033)￼ improved the previous behavior by checking only the model output instead of the full prompt.

## Test Plan

1. Add test for this new method interface in `tests/reasoning/test_base_thinking_reasoning_parser.py`

```bash
pytest tests/reasoning/test_base_thinking_reasoning_parser.py 
```

2. Benchmark vLLM on Holo2 8B with and without the change on a scenario with long reasoning and structured output. We use `extra-body` to actvate the structured output manager and add a logit bias on </think> to force the model to only think. 

```
MODEL=holo2-8b vllm bench serve \
    --dataset-name 'random' \
    --backend 'openai-chat' \
    --tokenizer 'Qwen/Qwen2-7B-Instruct' \
    --endpoint '/v1/chat/completions' \
    --ignore-eos \
    --metric-percentiles '25,50,75,90,95,99' \
    --percentile-metrics 'ttft,tpot,itl,e2el' \
    --goodput 'e2el:10000' \
    --save-detailed \
    --save-result \
    --extra-body '{ "guided_json": { "properties": {"city": {"title": "City", "type": "string"}}, "required": ["city"], "title": "Output", "type": "object" }, "logit_bias": { "151668": -100 } }' \
    --model $MODEL \
    --base-url http://$MODEL:8000 \
    --seed '1764348242' \
    --request-rate 'inf' \
    --random-input-len '7000' \
    --random-output-len '3000' \
    --max-concurrency '20' \
    --num-prompts '100'
```

## Test Result
1.  BaseThinkingReasoningParser tests are passing

```
================================================================================== test session starts ===================================================================================
platform darwin -- Python 3.12.11, pytest-8.4.2, pluggy-1.6.0
rootdir: /Users/hubert/code/hdlj-vllm
configfile: pyproject.toml
plugins: anyio-4.11.0, asyncio-1.3.0
asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 22 items                                                                                                                                                                       

tests/reasoning/test_base_thinking_reasoning_parser.py ......................      
```

### Holo2 8B bench result on H100

**main** before this PR: [Fix the issue with interleaved thinking when using streaming](https://github.com/vllm-project/vllm/pull/30033)

```
Successful requests:                     100
Maximum request concurrency:             20
Benchmark duration (s):                  521.35
Total input tokens:                      700000
Total generated tokens:                  300000
Request throughput (req/s):              0.19
Request goodput (req/s):                 0.00
Output token throughput (tok/s):         575.43
Peak output token throughput (tok/s):    700.00
Peak concurrent requests:                26.00
Total Token throughput (tok/s):          1918.09
Mean TTFT (ms):                          1562.85
Median TTFT (ms):                        1099.16
P25 TTFT (ms):                           1062.06
P50 TTFT (ms):                           1099.16
P75 TTFT (ms):                           1379.54
P90 TTFT (ms):                           3577.44
P95 TTFT (ms):                           4618.21
P99 TTFT (ms):                           5682.78
Mean TPOT (ms):                          34.23
Median TPOT (ms):                        34.34
P25 TPOT (ms):                           34.04
P50 TPOT (ms):                           34.34
P75 TPOT (ms):                           34.45
P90 TPOT (ms):                           34.50
P95 TPOT (ms):                           34.59
P99 TPOT (ms):                           34.69
Mean ITL (ms):                           34.25
Median ITL (ms):                         33.13
P25 ITL (ms):                            30.91
P50 ITL (ms):                            33.13
P75 ITL (ms):                            35.33
P90 ITL (ms):                            36.58
P95 ITL (ms):                            37.27
P99 ITL (ms):                            42.25
Mean E2EL (ms):                          104217.78
Median E2EL (ms):                        104375.43
P25 E2EL (ms):                           104054.62
P50 E2EL (ms):                           104375.43
P75 E2EL (ms):                           104696.93
P90 E2EL (ms):                           105622.35
P95 E2EL (ms):                           106714.51
P99 E2EL (ms):                           107775.61
```

**[main] After the fix in the interleaved reasoning**

```
Successful requests:                     100
Maximum request concurrency:             20
Benchmark duration (s):                  315.18
Total input tokens:                      700000
Total generated tokens:                  300000
Request throughput (req/s):              0.32
Request goodput (req/s):                 0.00
Output token throughput (tok/s):         951.83
Peak output token throughput (tok/s):    1280.00
Peak concurrent requests:                26.00
Total Token throughput (tok/s):          3172.77
Mean TTFT (ms):                          940.90
Median TTFT (ms):                        703.91
P50 TTFT (ms):                           703.91
P75 TTFT (ms):                           720.16
P95 TTFT (ms):                           2910.81
P99 TTFT (ms):                           3820.84
Mean TPOT (ms):                          20.70
Median TPOT (ms):                        20.76
P50 TPOT (ms):                           20.76
P75 TPOT (ms):                           20.86
P95 TPOT (ms):                           20.97
P99 TPOT (ms):                           21.10
Mean ITL (ms):                           20.70
Median ITL (ms):                         19.72
P50 ITL (ms):                            19.72
P75 ITL (ms):                            21.90
P95 ITL (ms):                            23.73
P99 ITL (ms):                            24.81
Mean E2EL (ms):                          63011.14
Median E2EL (ms):                        63034.41
P50 E2EL (ms):                           63034.41
P75 E2EL (ms):                           63570.68
P95 E2EL (ms):                           65097.66
P99 E2EL (ms):                           66037.71
```

**This PR:** (Checking only the delta)

```
Successful requests:                     100
Maximum request concurrency:             20
Benchmark duration (s):                  277.61
Total input tokens:                      700000
Total generated tokens:                  300000
Request throughput (req/s):              0.36
Request goodput (req/s):                 0.00
Output token throughput (tok/s):         1080.66
Peak output token throughput (tok/s):    1280.00
Peak concurrent requests:                27.00
Total Token throughput (tok/s):          3602.20
Mean TTFT (ms):                          919.02
Median TTFT (ms):                        691.28
P50 TTFT (ms):                           691.28
P75 TTFT (ms):                           702.26
P95 TTFT (ms):                           2860.22
P99 TTFT (ms):                           3766.02
Mean TPOT (ms):                          18.20
Median TPOT (ms):                        18.27
P50 TPOT (ms):                           18.27
P75 TPOT (ms):                           18.31
P95 TPOT (ms):                           18.41
P99 TPOT (ms):                           18.46
Mean ITL (ms):                           18.30
Median ITL (ms):                         17.28
P50 ITL (ms):                            17.28
P75 ITL (ms):                            18.13
P95 ITL (ms):                            18.91
P99 ITL (ms):                            32.93
Mean E2EL (ms):                          55499.46
Median E2EL (ms):                        55505.07
P50 E2EL (ms):                           55505.07
P75 E2EL (ms):                           55611.81
P95 E2EL (ms):                           57556.07
P99 E2EL (ms):                           58462.71
```

### Holo2 30B A3B bench result on H100

**main** before this PR: [Fix the issue with interleaved thinking when using streaming](https://github.com/vllm-project/vllm/pull/30033)

```
Successful requests:                     100
Maximum request concurrency:             20
Benchmark duration (s):                  399.75
Total input tokens:                      700000
Total generated tokens:                  300000
Request throughput (req/s):              0.25
Request goodput (req/s):                 0.00
Output token throughput (tok/s):         750.47
Peak output token throughput (tok/s):    900.00
Peak concurrent requests:                33.00
Total Token throughput (tok/s):          2501.57
Mean TTFT (ms):                          504.10
Median TTFT (ms):                        362.88
P50 TTFT (ms):                           362.88
P75 TTFT (ms):                           391.95
P95 TTFT (ms):                           1434.96
P99 TTFT (ms):                           1797.39
Mean TPOT (ms):                          26.48
Median TPOT (ms):                        26.56
P50 TPOT (ms):                           26.56
P75 TPOT (ms):                           26.60
P95 TPOT (ms):                           26.96
P99 TPOT (ms):                           26.97
Mean ITL (ms):                           27.50
Median ITL (ms):                         26.06
P50 ITL (ms):                            26.06
P75 ITL (ms):                            27.68
P95 ITL (ms):                            30.50
P99 ITL (ms):                            56.67
Mean E2EL (ms):                          79921.00
Median E2EL (ms):                        80023.61
P50 E2EL (ms):                           80023.61
P75 E2EL (ms):                           80104.15
P95 E2EL (ms):                           82260.51
P99 E2EL (ms):                           82625.40
```

**[main] After the fix in the interleaved reasoning**

```
Successful requests:                     100
Maximum request concurrency:             20
Benchmark duration (s):                  202.75
Total input tokens:                      700000
Total generated tokens:                  300000
Request throughput (req/s):              0.49
Request goodput (req/s):                 0.00
Output token throughput (tok/s):         1479.67
Peak output token throughput (tok/s):    2039.00
Peak concurrent requests:                37.00
Total Token throughput (tok/s):          4932.22
Mean TTFT (ms):                          489.01
Median TTFT (ms):                        341.32
P50 TTFT (ms):                           341.32
P75 TTFT (ms):                           428.19
P95 TTFT (ms):                           1345.94
P99 TTFT (ms):                           1581.73
Mean TPOT (ms):                          13.35
Median TPOT (ms):                        13.45
P50 TPOT (ms):                           13.45
P75 TPOT (ms):                           13.58
P95 TPOT (ms):                           13.73
P99 TPOT (ms):                           13.76
Mean ITL (ms):                           13.69
Median ITL (ms):                         12.98
P50 ITL (ms):                            12.98
P75 ITL (ms):                            14.31
P95 ITL (ms):                            16.06
P99 ITL (ms):                            28.01
Mean E2EL (ms):                          40531.71
Median E2EL (ms):                        40744.67
P50 E2EL (ms):                           40744.67
P75 E2EL (ms):                           41468.42
P95 E2EL (ms):                           41804.52
P99 E2EL (ms):                           42048.05
```

**This PR:** (Checking only the delta)

```
Successful requests:                     100
Maximum request concurrency:             20
Benchmark duration (s):                  161.63
Total input tokens:                      700000
Total generated tokens:                  300000
Request throughput (req/s):              0.62
Request goodput (req/s):                 0.00
Output token throughput (tok/s):         1856.05
Peak output token throughput (tok/s):    2134.00
Peak concurrent requests:                37.00
Total Token throughput (tok/s):          6186.84
Mean TTFT (ms):                          532.56
Median TTFT (ms):                        411.96
P50 TTFT (ms):                           411.96
P75 TTFT (ms):                           432.54
P95 TTFT (ms):                           1402.83
P99 TTFT (ms):                           1722.41
Mean TPOT (ms):                          10.60
Median TPOT (ms):                        10.49
P50 TPOT (ms):                           10.49
P75 TPOT (ms):                           10.67
P95 TPOT (ms):                           11.05
P99 TPOT (ms):                           11.11
Mean ITL (ms):                           11.22
Median ITL (ms):                         10.15
P50 ITL (ms):                            10.15
P75 ITL (ms):                            10.46
P95 ITL (ms):                            20.04
P99 ITL (ms):                            20.83
Mean E2EL (ms):                          32310.77
Median E2EL (ms):                        31802.08
P50 E2EL (ms):                           31802.08
P75 E2EL (ms):                           32415.42
P95 E2EL (ms):                           34495.17
P99 E2EL (ms):                           34809.45
```

---
<details>

</details>

### DIFF
--- a/docs/features/reasoning_outputs.md
+++ b/docs/features/reasoning_outputs.md
@@ -299,6 +299,9 @@ Additionally, to enable structured output, you'll need to create a new `Reasoner
 
         def is_reasoning_end(self, input_ids: list[int]) -> bool:
             return self.end_token_id in input_ids
+
+        def is_reasoning_end_streaming(self, input_ids: list[int], delta_ids: list[int]) -> bool:
+            return self.end_token_id in delta_token_ids
         ...
     ```
 
--- a/docs/features/reasoning_outputs.md
+++ b/docs/features/reasoning_outputs.md
@@ -299,9 +299,6 @@ Additionally, to enable structured output, you'll need to create a new `Reasoner
 
         def is_reasoning_end(self, input_ids: list[int]) -> bool:
             return self.end_token_id in input_ids
-
-        def is_reasoning_end_on_decode_step(self, input_ids: list[int]) -> bool:
-            return len(input_ids) > 0 and self.end_token_id == input_ids[-1]
         ...
     ```
 
--- a/docs/features/reasoning_outputs.md
+++ b/docs/features/reasoning_outputs.md
@@ -299,6 +299,9 @@ Additionally, to enable structured output, you'll need to create a new `Reasoner
 
         def is_reasoning_end(self, input_ids: list[int]) -> bool:
             return self.end_token_id in input_ids
+
+        def is_reasoning_end_on_decode_step(self, input_ids: list[int]) -> bool:
+            return len(input_ids) > 0 and self.end_token_id == input_ids[-1]
         ...
     ```
 
--- a/vllm/reasoning/basic_parsers.py
+++ b/vllm/reasoning/basic_parsers.py
@@ -77,7 +77,10 @@ class BaseThinkingReasoningParser(ReasoningParser):
     def is_reasoning_end_streaming(
         self, input_ids: list[int], delta_ids: list[int]
     ) -> bool:
-        return self.is_reasoning_end(delta_ids)
+        # We only check the end token beause interleaved reasoning and content is not
+        # compatible with the current structured output manager.
+        end_token_id = self.end_token_id
+        return end_token_id in delta_ids
 
     def extract_content_ids(self, input_ids: list[int]) -> list[int]:
         """
--- a/vllm/reasoning/basic_parsers.py
+++ b/vllm/reasoning/basic_parsers.py
@@ -74,9 +74,6 @@ class BaseThinkingReasoningParser(ReasoningParser):
                 return True
         return False
 
-    def is_reasoning_end_on_decode_step(self, input_ids: list[int]) -> bool:
-        return len(input_ids) > 0 and self.end_token_id == input_ids[-1]
-
     def extract_content_ids(self, input_ids: list[int]) -> list[int]:
         """
         Extract the content after the end tokens
--- a/vllm/reasoning/basic_parsers.py
+++ b/vllm/reasoning/basic_parsers.py
@@ -74,6 +74,9 @@ class BaseThinkingReasoningParser(ReasoningParser):
                 return True
         return False
 
+    def is_reasoning_end_on_decode_step(self, input_ids: list[int]) -> bool:
+        return len(input_ids) > 0 and self.end_token_id == input_ids[-1]
+
     def extract_content_ids(self, input_ids: list[int]) -> list[int]:
         """
         Extract the content after the end tokens
--- a/vllm/reasoning/basic_parsers.py
+++ b/vllm/reasoning/basic_parsers.py
@@ -77,8 +77,6 @@ class BaseThinkingReasoningParser(ReasoningParser):
     def is_reasoning_end_streaming(
         self, input_ids: list[int], delta_ids: list[int]
     ) -> bool:
-        # We only check the end token beause interleaved reasoning and content is not
-        # compatible with the current structured output manager.
         end_token_id = self.end_token_id
         return end_token_id in delta_ids
 
--- a/vllm/reasoning/basic_parsers.py
+++ b/vllm/reasoning/basic_parsers.py
@@ -74,6 +74,11 @@ class BaseThinkingReasoningParser(ReasoningParser):
                 return True
         return False
 
+    def is_reasoning_end_streaming(
+        self, input_ids: list[int], delta_ids: list[int]
+    ) -> bool:
+        return self.is_reasoning_end(delta_ids)
+
     def extract_content_ids(self, input_ids: list[int]) -> list[int]:
         """
         Extract the content after the end tokens
--- a/vllm/v1/structured_output/__init__.py
+++ b/vllm/v1/structured_output/__init__.py
@@ -339,7 +339,9 @@ class StructuredOutputManager:
             return True
 
         # Check if reasoning ends in *this* step
-        if self.reasoner.is_reasoning_end_on_decode_step(request.all_token_ids):
+        if self.reasoner.is_reasoning_end(
+            request.all_token_ids[request.num_computed_tokens :]
+        ):
             # Reasoning just ended, so we shouldn't advance til
             # next pass
             structured_req.reasoning_ended = True
--- a/vllm/v1/structured_output/__init__.py
+++ b/vllm/v1/structured_output/__init__.py
@@ -339,8 +339,8 @@ class StructuredOutputManager:
             return True
 
         # Check if reasoning ends in *this* step
-        if self.reasoner.is_reasoning_end(
-            request.all_token_ids[request.num_computed_tokens :]
+        if self.reasoner.is_reasoning_end_streaming(
+            request.all_token_ids, request.all_token_ids[request.num_computed_tokens :]
         ):
             # Reasoning just ended, so we shouldn't advance til
             # next pass
--- a/vllm/v1/structured_output/__init__.py
+++ b/vllm/v1/structured_output/__init__.py
@@ -339,7 +339,7 @@ class StructuredOutputManager:
             return True
 
         # Check if reasoning ends in *this* step
-        if self.reasoner.is_reasoning_end(request.all_token_ids):
+        if self.reasoner.is_reasoning_end_on_decode_step(request.all_token_ids):
             # Reasoning just ended, so we shouldn't advance til
             # next pass
             structured_req.reasoning_ended = True
--- a/tests/reasoning/test_base_thinking_reasoning_parser.py
+++ b/tests/reasoning/test_base_thinking_reasoning_parser.py
@@ -132,6 +132,20 @@ class TestBaseThinkingReasoningParserMethods:
             is False
         )
 
+    def test_is_reasoning_end_on_decode_step(self, test_tokenizer):
+        """Test the is_reasoning_end_on_decode_step method."""
+        parser = TestThinkingReasoningParser(test_tokenizer)
+        end_token_id = parser.end_token_id
+
+        assert parser.is_reasoning_end_on_decode_step([]) is False
+        assert parser.is_reasoning_end_on_decode_step([1, 2, 3, 4]) is False
+        assert parser.is_reasoning_end_on_decode_step([1, 2, 3, end_token_id]) is True
+        # Even if the end token is present, the function can return False if it is not
+        # the current decode step that is ending the reasoning content.
+        assert (
+            parser.is_reasoning_end_on_decode_step([1, 2, 3, end_token_id, 4]) is False
+        )
+
     def test_extract_content_ids(self, test_tokenizer):
         """Test the extract_content_ids method."""
         parser = TestThinkingReasoningParser(test_tokenizer)
--- a/tests/reasoning/test_base_thinking_reasoning_parser.py
+++ b/tests/reasoning/test_base_thinking_reasoning_parser.py
@@ -155,8 +155,8 @@ class TestBaseThinkingReasoningParserMethods:
         )
         assert (
             parser.is_reasoning_end_streaming(
-                [1, start_token_id, 2, end_token_id, 2, start_token_id],
-                [end_token_id, 2, start_token_id],
+                [1, start_token_id, 2, end_token_id, 2, start_token_id, 2],
+                [2],
             )
             is False
         )
--- a/tests/reasoning/test_base_thinking_reasoning_parser.py
+++ b/tests/reasoning/test_base_thinking_reasoning_parser.py
@@ -132,6 +132,41 @@ class TestBaseThinkingReasoningParserMethods:
             is False
         )
 
+    def test_is_reasoning_end_streaming(self, test_tokenizer):
+        """Test the is_reasoning_end_streaming method."""
+        parser = TestThinkingReasoningParser(test_tokenizer)
+        end_token_id = parser.end_token_id
+        start_token_id = parser.start_token_id
+
+        assert (
+            parser.is_reasoning_end_streaming([1, 2, end_token_id], [end_token_id])
+            is True
+        )
+        assert parser.is_reasoning_end_streaming([1, 2, 3, 4], [4]) is False
+        assert parser.is_reasoning_end_streaming([], []) is False
+        assert (
+            parser.is_reasoning_end_streaming(
+                [1, start_token_id, 2, end_token_id], [end_token_id]
+            )
+            is True
+        )
+        assert (
+            parser.is_reasoning_end_streaming([1, start_token_id, 2, 3], [3]) is False
+        )
+        assert (
+            parser.is_reasoning_end_streaming(
+                [1, start_token_id, 2, end_token_id, 2, start_token_id],
+                [end_token_id, 2, start_token_id],
+            )
+            is False
+        )
+        assert (
+            parser.is_reasoning_end_streaming(
+                [1, start_token_id, 2, end_token_id, 2, 2], [2]
+            )
+            is False
+        )
+
     def test_extract_content_ids(self, test_tokenizer):
         """Test the extract_content_ids method."""
         parser = TestThinkingReasoningParser(test_tokenizer)
--- a/tests/reasoning/test_base_thinking_reasoning_parser.py
+++ b/tests/reasoning/test_base_thinking_reasoning_parser.py
@@ -132,20 +132,6 @@ class TestBaseThinkingReasoningParserMethods:
             is False
         )
 
-    def test_is_reasoning_end_on_decode_step(self, test_tokenizer):
-        """Test the is_reasoning_end_on_decode_step method."""
-        parser = TestThinkingReasoningParser(test_tokenizer)
-        end_token_id = parser.end_token_id
-
-        assert parser.is_reasoning_end_on_decode_step([]) is False
-        assert parser.is_reasoning_end_on_decode_step([1, 2, 3, 4]) is False
-        assert parser.is_reasoning_end_on_decode_step([1, 2, 3, end_token_id]) is True
-        # Even if the end token is present, the function can return False if it is not
-        # the current decode step that is ending the reasoning content.
-        assert (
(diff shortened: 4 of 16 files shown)


vllm/v1/structured_output/__init__.py
```diff
@@ -326,7 +326,7 @@ def should_advance(self, request: Request) -> bool:
             return True
 
         # Check if reasoning ends in *this* step
-        if self.reasoner.is_reasoning_end(request.all_token_ids):
+        if self.reasoner.is_reasoning_end_on_decode_step(request.all_token_ids):
             # Reasoning just ended, so we shouldn't advance til
```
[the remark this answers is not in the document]
hdlj-h: I have added the function to this reasoning parser.

docs/features/reasoning_outputs.md
```diff
@@ -297,6 +297,9 @@ Additionally, to enable structured output, you'll need to create a new `Reasoner
 
         def is_reasoning_end(self, input_ids: list[int]) -> bool:
             return self.end_token_id in input_ids
+
+        def is_reasoning_end_on_decode_step(self, input_ids: list[int]) -> bool:
```
chaunceyjiang: I suggest adding a `**kwargs` parameter to `is_reasoning_end`. What do you think?

vllm/v1/structured_output/__init__.py
```diff
@@ -326,7 +326,7 @@ def should_advance(self, request: Request) -> bool:
             return True
 
         # Check if reasoning ends in *this* step
-        if self.reasoner.is_reasoning_end(request.all_token_ids):
+        if self.reasoner.is_reasoning_end_on_decode_step(request.all_token_ids):
```
chaunceyjiang:
```suggestion
        if self.reasoner.is_reasoning_end(request.all_token_ids, setp="decode"):
```

WDYT?

docs/features/reasoning_outputs.md
chaunceyjiang: It's currently known that there are some reasoning_parser plugins outside of vLLM. I'm concerned that this change might be too aggressive.

vllm/v1/structured_output/__init__.py
hdlj-h: Thanks for the suggestion ! If plugins can use this interface outside, I agree that adding a new method breaks them. 
I have hesitating with your approach initially. 

Either we provide more information to the `is_reasoning_end` function to be retro-compatible. However, by looking to the code, the same class is used for 2 different purposes: at the frontend level (extract_... + is_reasoning_ends) but also in the EngineCore process. Maybe another approach could be to have a dedicated class for the structured output (initiated per request) that let us do the same approach for multi token end reasoning too. WDYT ? 

On the engine core side, we initiate the ReasoningParser -> get_structured_output_reasoning_checker() with a retro-compatibility check if the function is not implemented by the external plugins. 

I will update my PR with your suggested changes.

chaunceyjiang: > Either we provide more information to the is_reasoning_end function to be retro-compatible. However, by looking to the code, the same class is used for 2 different purposes: at the frontend level (extract_... + is_reasoning_ends) but also in the EngineCore process. Maybe another approach could be to have a dedicated class for the structured output (initiated per request) that let us do the same approach for multi token end reasoning too. WDYT ?


Yes, I completely agree with your point. In fact, I have previously tried to optimize `is_reasoning_end`, but I encountered a difficult issue: as you mentioned, `is_reasoning_end` is used in multiple places, especially in the frontend + streaming scenario. See https://github.com/vllm-project/vllm/pull/25735#issue-3455641009.

Given that `is_reasoning_end` now has such a significant impact on performance, I believe it's time to address this issue.

As I just mentioned, I'm concerned that overly aggressive changes could break downstream dependencies. I hope we can find a more backward-compatible way to implement the modifications.

hdlj-h: Thanks for your answer, this design suggestion is more for a future version but indeed it will be breaking. 

However if I add `kwargs` argument to the `is_reasoning_end`, external plugins will still be broken no ? 

Otherway, we do a default implementation:

```
@abstractmethod
 def is_reasoning_end_on_decode_step(self, input_ids: list[int]) -> bool:
        """
        Check if the reasoning content ends in the input_ids on a
        decode step.

        It is used in structured engines like `xgrammar` to check if the
        reasoning content ends in the model output before applying the
        structured output.

        Notes:
            - The first time the reasoning content ends during a decode step, this
            method returns True. StructuredOutputManager then caches the result.
            - Subsequent decode steps for the same reasoning segment can return
            False or True.

        Parameters:
        input_ids: list[int]
            The input_ids of the model output at the current decode step.

        Returns:
        bool
            True if the reasoning content ends in the input_ids on a
            decode step.
        """
        return self.is_reasoning_end(input_ids)
```
WDYT ?

chaunceyjiang: go ahead

chaunceyjiang:
```suggestion
        if self.reasoner.is_reasoning_end(request.all_token_ids[request.num_computed_tokens:]):
```

chaunceyjiang: @hdlj-h 🤔，Perhaps this should be handled this way. WDYT?

chaunceyjiang: We only need to compute the newly generated tokens, not the tokens that have already been computed.

hdlj-h: Indeed, that is also an option, but some reasoning parsers rely on multi-token end markers, so the assumption that all parsers are single-token–based does not always hold.

For example:
• OLMo3 operates in string space rather than token space. It may not be compatible with structured output for this reason.
• Granite also works in string space and does not implement is_reasoning_end, meaning it could crash if used with structured output.
• Hunyuan maintains an internal state machine for reasoning, which likely makes it incompatible with structured output + reasoning.

num_computed_tokens can be greater than 1 in the V1 ? or is it the output tokens counter ? If it is the case my PR is not bullet proof because I take the assumption that a decoding step == 1 generated token.

chaunceyjiang: > num_computed_tokens can be greater than 1 in the V1 ?

Yes.

chaunceyjiang: OLMo3, Granite, and Hunyuan — I think we can set these aside for now. They are inherently incompatible with structured output to begin with.

Therefore, we should ignore them for the time being and optimize them separately later.

hdlj-h: > > num_computed_tokens can be greater than 1 in the V1 ?
> 
> Yes.

Ah I didn't take in account the speculative decoding case. So my PR is not compatible in such situation when num_computed_tokens > 1. 

So I will revert my change and apply your approach instead that is a smaller change.

chaunceyjiang (commented):
Thanks~

Can you help test the benchmark again?

vllm/v1/structured_output/__init__.py
hdlj-h: @chaunceyjiang  num_computed_tokens, I don't find the related docs, is it the number of computed tokens before the step or is it the number of computed tokens during the step ?

vllm/v1/structured_output/__init__.py
```diff
@@ -326,7 +326,9 @@ def should_advance(self, request: Request) -> bool:
             return True
 
         # Check if reasoning ends in *this* step
-        if self.reasoner.is_reasoning_end(request.all_token_ids):
+        if self.reasoner.is_reasoning_end(
+            request.all_token_ids[request.num_computed_tokens :]
+        ):
```
chaunceyjiang, line 344: The e2e test failed because gpt-oss uses multiple tokens in `reasoning_end_token_ids` for judgment.

This approach doesn't seem to work well either.

Currently, there are two solutions. Which one do you prefer?  
1. Special handling for gpt-oss?  
2. Use your initial approach—add an `is_reasoning_end_streaming` method?

hdlj-h: Thanks for detecting the issue gpt-oss. If we do a special handle for gpt-oss we could have the situation with other parsers that are multi tokens based. 

For the 2nd approach, I see 2 possible interface with default implementation for backward compatibility:

```
is_reasoning_end_streaming(input_ids:list[int], diff: list[int])
```

```
is_reasoning_end_streaming(input_ids:list[int], num_computed_tokens: list[int])
```

The first signature is better I think because it could be re-usable for frontend purposes, but the second one avoid a list copy at each decode step but maybe it is negligeable.  WDYT ?

hdlj-h: new implementation integrated.

vllm/reasoning/basic_parsers.py
```diff
@@ -74,6 +74,14 @@ def is_reasoning_end(self, input_ids: list[int]) -> bool:
                 return True
         return False
 
+    def is_reasoning_end_streaming(
+        self, input_ids: list[int], delta_ids: list[int]
+    ) -> bool:
+        # We only check the end token beause interleaved reasoning and content is not
+        # compatible with the current structured output manager.
```
chaunceyjiang: I don't quite understand this comment. I think it should be compatible, right?

chaunceyjiang (commented):
LGTM.

I sincerely thank you for the effort and time you have put into this. @hdlj-h

vllm/reasoning/basic_parsers.py
hdlj-h: The Structured Output manager tracks if the model output has the \</think\> token and then marks the request as reasoning ended to start enforcing the structured output grammar. After this point, the `is_reasoning_end_streaming` is not called anymore. Because of this, if the model outputs \<think\> again, we cannot go back to reasoning tracking again to support interleaved reasoning and content in the model output. 

It is my understanding from looking at the code. Unless, we reset the `.reasoning_ended` flag somewhere else but I don't find it in the current code base. However, interleaved reasoning/content is supported at the prompt level. 

That is why, in this function, we can check only `</think>` on the delta. 

I can edit my comment or remove it if you think it is unclear or not relevant to specify this.

chaunceyjiang: Okay, I understand.
My suggestion is to remove it, because vLLM currently supports interleaved reasoning/content at the prompt level, and keeping it may cause confusion.

chaunceyjiang (approved):
Thanks~ @hdlj-h

Issue #30052, opened 2025-12-04
[Bug]: An error will occur when using a quantization configuration with modules_to_not_convert for the DeepSeek-V3.2 model on non-NVIDIA GPUs


tom-zju:
### Your current environment

vllm_version:
- release/v0.11.1

DeepSeek-V3.2-Exp-AWQ:
https://huggingface.co/QuantTrio/DeepSeek-V3.2-Exp-AWQ/blob/main/config.json
<html>
<body>

"quantization_config": {
--
"quant_method": "awq",
"bits": 4,
"group_size": 128,
"version": "gemm",
"zero_point": true,
"modules_to_not_convert": ["mlp.gate.", "indexer", "model.layers.0.", "model.layers.1.", "model.layers.2.", "model.layers.60.", "model.layers.61."]
},

</body>
</html>

server launch code
-  python -m vllm.entrypoints.openai.api_server   --port 8000   --model=/home/models/DeepSeek-V3.2-Exp-AWQ   --tensor-parallel-size=8   --disable-log-requests   --gpu-memory-utilization=0.8   --block-size=64   --dtype=auto   --max-model-len=10   --max-num-batched-tokens 10   --max-num-seqs 1   --trust-remote-code   --enforce-eager --quantization=awq

error:
(Worker_TP0 pid=10292) ERROR 12-04 10:31:42 [multiproc_executor.py:743]   File "/usr/local/lib/python3.10/dist-packages/vllm/model_executor/models/deepseek_v2.py", line 1124, in __init__
(Worker_TP0 pid=10292) ERROR 12-04 10:31:42 [multiproc_executor.py:743]     self.mlp = DeepseekV2MoE(
(Worker_TP0 pid=10292) ERROR 12-04 10:31:42 [multiproc_executor.py:743]   File "/usr/local/lib/python3.10/dist-packages/vllm/model_executor/models/deepseek_v2.py", line 310, in __init__
(Worker_TP0 pid=10292) ERROR 12-04 10:31:42 [multiproc_executor.py:743]     self.experts = SharedFusedMoE(
(Worker_TP0 pid=10292) ERROR 12-04 10:31:42 [multiproc_executor.py:743]   File "/usr/local/lib/python3.10/dist-packages/vllm/model_executor/layers/fused_moe/shared_fused_moe.py", line 28, in __init__
(Worker_TP0 pid=10292) ERROR 12-04 10:31:42 [multiproc_executor.py:743]     super().__init__(**kwargs)
[...]

PR #30057, opened 2025-12-04
[Bugfix] fix _get_quant_method of FusedMoE for deepseekV3.2 on non-NV…

tom-zju:
Purpose: to fix this bug: 
https://github.com/vllm-project/vllm/issues/30052


vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -584,10 +584,11 @@ def _get_quant_method() -> FusedMoEMethodBase:
             Helper method to ensure self.quant_method is never None and
             of the proper type.
             """
+            from vllm.model_executor.layers.linear import LinearMethodBase
             quant_method = None
             if self.quant_config is not None:
                 quant_method = self.quant_config.get_quant_method(self, prefix)
-            if quant_method is None:
+            if quant_method is None or isinstance(quant_method, LinearMethodBase)::
```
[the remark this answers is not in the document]
tom-zju: > critical
> 
> This line contains a syntax error. The `::` at the end of the line should be a single colon `:`. This will cause a `SyntaxError` and prevent the code from running.

finished

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -584,12 +584,15 @@ def _get_quant_method() -> FusedMoEMethodBase:
             Helper method to ensure self.quant_method is never None and
             of the proper type.
             """
+            from vllm.model_executor.layers.linear import LinearMethodBase
+
             quant_method = None
             if self.quant_config is not None:
                 quant_method = self.quant_config.get_quant_method(self, prefix)
-            if quant_method is None:
+            if quant_method is None or isinstance(quant_method, LinearMethodBase):
```
LucasWilkinson: should this be `FusedMoEMethodBase`?

tom-zju: yes, this should be FusedMoEMethodBase,  https://github.com/vllm-project/vllm/issues/30052, but may be return UnquantizedLinearMethod() and cause to error

tom-zju:
> cc @mgoin

is any question about this patch?

vllm/model_executor/layers/fused_moe/layer.py
pavanimajety: The excluded layers should be excluded from the corresponding quantization file's get_quant_method. Excluding from fused_moe/layer.py seems slightly hacky. Eg: Check out - https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/quantization/fp8.py#L314-L341

https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/quantization/awq.py#L95-L140

pavanimajety (changes requested):

vllm/model_executor/layers/fused_moe/layer.py
tom-zju: OK， you are right， i has finished as your code review comments

pavanimajety:
@tom-zju  Could you please post evals for the model?

pavanimajety (approved):
Thanks for the PR!

tom-zju:
> Thanks for the PR!

CI is throwing an error, but this error is unrelated to the code I submitted. How can I resolve this?

PR #30072, opened 2025-12-04
[Core] Whisper enable `FULL_DECODE_ONLY` CudaGraph 

NickLucche:
## Overview

Whisper currently only supports PIECEWISE cudagraph mode, but it does NOT support torch.compile, so in practice this results in running in eager mode (see profiler below, `cudagraphlaunch` query). 

Pasted image 20251202144814

This PR addresses another important performance limitation by adding support for cuda graph through the `FULL_DECODE_ONLY` mode.  
This guarantees that all decode steps after the first one will just replay the graph.
Mind that the first decode step is actually branching in flow due to `encoder_outputs` being present for populating the crossattn kv cache. Hence we only GC the steps that follow.

New profile(r) pic:
Pasted image 20251204180234

Full explicit command:
```
vllm serve openai/whisper-large-v3-turbo --max-num-batched-tokens 32874 -cc.cudagraph_mode=FULL_DECODE_ONLY
```

New default:
```
vllm serve openai/whisper-large-v3-turbo
...
INFO 12-04 17:44:48 [vllm.py:704] Encoder-decoder models do not support FULL_AND_PIECEWISE. Overriding cudagraph_mode to FULL_DECODE_ONLY.
```

## Benchmark

Pre
```
RESULTS SUMMARY
Total samples: 50
Successful: 50
Failed: 0
Total time: 1.36s
Average latency: 0.88s
Throughput: 36.82 requests/s
```

Post 
```
RESULTS SUMMARY
Total samples: 50
Successful: 50
Failed: 0
Total time: 1.02s
Average latency: 0.61s
Throughput: 49.11 requests/s
```

cc @ProExpertProg @LucasWilkinson @robertgshaw2-redhat

### DIFF
--- a/tests/models/multimodal/generation/test_whisper.py
+++ b/tests/models/multimodal/generation/test_whisper.py
@@ -102,6 +102,7 @@ def run_test(
         max_model_len=448,
         tensor_parallel_size=tensor_parallel_size,
         distributed_executor_backend=distributed_executor_backend,
+        enforce_eager=True,
     ) as vllm_model:
         llm = vllm_model.llm
 
--- a/tests/models/multimodal/generation/test_whisper.py
+++ b/tests/models/multimodal/generation/test_whisper.py
@@ -102,6 +102,7 @@ def run_test(
         max_model_len=448,
         tensor_parallel_size=tensor_parallel_size,
         distributed_executor_backend=distributed_executor_backend,
+        # TODO (NickLucche) figure out output differences with non-eager and re-enable
         enforce_eager=True,
     ) as vllm_model:
         llm = vllm_model.llm
--- a/vllm/config/vllm.py
+++ b/vllm/config/vllm.py
@@ -666,17 +666,17 @@ class VllmConfig:
 
         default_config = OPTIMIZATION_LEVEL_TO_CONFIG[self.optimization_level]
         self._apply_optimization_level_defaults(default_config)
-        if (
-            self.compilation_config.cudagraph_mode != CUDAGraphMode.NONE
-            and self.compilation_config.mode != CompilationMode.VLLM_COMPILE
-        ):
-            logger.info(
-                "Cudagraph mode %s is not compatible with compilation mode %s."
-                "Overriding to NONE.",
-                self.compilation_config.cudagraph_mode,
-                self.compilation_config.mode,
-            )
-            self.compilation_config.cudagraph_mode = CUDAGraphMode.NONE
+        # if (
+        #     self.compilation_config.cudagraph_mode != CUDAGraphMode.PIECEWISE OR FULL_AND_PIECEWISE
+        #     and self.compilation_config.mode != CompilationMode.VLLM_COMPILE
+        # ):
+        #     logger.info(
+        #         "Cudagraph mode %s is not compatible with compilation mode %s."
+        #         "Overriding to NONE.",
+        #         self.compilation_config.cudagraph_mode,
+        #         self.compilation_config.mode,
+        #     )
+        #     self.compilation_config.cudagraph_mode = CUDAGraphMode.NONE
 
         # async tp is built on top of sequence parallelism
         # and requires it to be enabled.
@@ -703,11 +703,12 @@ class VllmConfig:
                     )
                     self.compilation_config.cudagraph_mode = CUDAGraphMode.PIECEWISE
                 elif self.model_config.is_encoder_decoder:
-                    logger.warning_once(
-                        "Encoder-decoder models do not support full cudagraphs. "
-                        "Overriding cudagraph_mode to PIECEWISE."
-                    )
-                    self.compilation_config.cudagraph_mode = CUDAGraphMode.PIECEWISE
+                    pass
+                    # logger.warning_once(
+                    #     "Encoder-decoder models do not support full cudagraphs. "
+                    #     "Overriding cudagraph_mode to PIECEWISE."
+                    # )
+                    # self.compilation_config.cudagraph_mode = CUDAGraphMode.PIECEWISE
 
             # disable cudagraph when enforce eager execution
             if self.model_config is not None and self.model_config.enforce_eager:
--- a/vllm/config/vllm.py
+++ b/vllm/config/vllm.py
@@ -666,9 +666,9 @@ class VllmConfig:
 
         default_config = OPTIMIZATION_LEVEL_TO_CONFIG[self.optimization_level]
         self._apply_optimization_level_defaults(default_config)
+
         if (
-            self.compilation_config.cudagraph_mode
-            in (CUDAGraphMode.PIECEWISE, CUDAGraphMode.FULL_AND_PIECEWISE)
+            self.compilation_config.cudagraph_mode.requires_piecewise_compilation()
             and self.compilation_config.mode != CompilationMode.VLLM_COMPILE
         ):
             logger.info(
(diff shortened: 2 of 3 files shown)


vllm/config/vllm.py
```diff
@@ -646,7 +646,8 @@ def has_blocked_weights():
         default_config = OPTIMIZATION_LEVEL_TO_CONFIG[self.optimization_level]
         self._apply_optimization_level_defaults(default_config)
         if (
-            self.compilation_config.cudagraph_mode != CUDAGraphMode.NONE
+            self.compilation_config.cudagraph_mode
+            in (CUDAGraphMode.PIECEWISE, CUDAGraphMode.FULL_AND_PIECEWISE)
```
NickLucche: @ProExpertProg need your eyes on this bit

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2771,6 +2774,12 @@ def _determine_batch_execution_and_padding(
             if force_uniform_decode is None
             else force_uniform_decode
         )
+        # Encoder-decoder models only support GC for decoder_step > 0 (no enc_output
```
ProExpertProg: Spell out cudagraphs?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -4035,6 +4045,7 @@ def _dummy_run(
             assert num_tokens_padded <= self.max_num_tokens
             model_kwargs = self._init_model_kwargs(num_tokens_padded)
             if self.supports_mm_inputs and not self.model_config.is_encoder_decoder:
+                # NOT ENC-DEC
```
ProExpertProg: Cruft?

vllm/config/vllm.py
ProExpertProg:
```suggestion            self.compilation_config.cudagraph_mode.requires_piecewise_compilation()
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2756,6 +2774,12 @@ def _determine_batch_execution_and_padding(
             if force_uniform_decode is None
             else force_uniform_decode
         )
+        # Encoder-decoder models only support GC for decoder_step > 0 (no enc_output
+        # is present). Also, chunked-prefill is disabled, so batch are uniform.
+        has_encoder_output = (
```
LucasWilkinson, line 2788: I think we should rename `use_cascade_attn` parameter to `self.cudagraph_dispatcher.dispatch` to `piecewise_or_eager_only` (I have a WIP PR that will do this anyways) and then do:

```
self.cudagraph_dispatcher.dispatch(
    num_tokens=num_tokens,
    has_lora=has_lora,
    uniform_decode=uniform_decode,
    piecewise_or_eager_only=use_cascade_attn or has_encoder_output,
)
```

instead of changing the semantic meaning of `uniform_decode`

vllm/v1/worker/gpu_model_runner.py
NickLucche: yeah I tried

NickLucche:
Thanks for reviewing @ProExpertProg @LucasWilkinson , I hope I've addressed your comments

vllm/v1/cudagraph_dispatcher.py
```diff
@@ -145,7 +145,7 @@ def dispatch(
         num_tokens: int,
         uniform_decode: bool,
         has_lora: bool,
-        use_cascade_attn: bool = False,
+        piecewise_or_eager_only: bool = False,
```
ProExpertProg: `skip_attention` or `attention_cg_unsupported`?

vllm/config/vllm.py
```diff
@@ -681,12 +682,19 @@ def has_blocked_weights():
...
                     self.compilation_config.cudagraph_mode = CUDAGraphMode.PIECEWISE
-                elif self.model_config.is_encoder_decoder:
-                    logger.warning_once(
-                        "Encoder-decoder models do not support full cudagraphs. "
-                        "Overriding cudagraph_mode to PIECEWISE."
+                elif (
+                    self.model_config.is_encoder_decoder
+                    and self.compilation_config.cudagraph_mode
+                    not in (CUDAGraphMode.NONE, CUDAGraphMode.FULL_DECODE_ONLY)
+                ):
```
ProExpertProg, line 710: if somebody sets mode=piecewise, we don't handle it here, but we should:

```
            # check support based on model type
            if self.model_config is not None:
                if self.model_config.pooler_config is not None and self.compilation_config.cudagraph_mode.has_full_cudagraphs():
                    ...
                elif (
                    self.model_config.is_encoder_decoder
                    and self.compilation_config.cudagraph_mode
                    not in (CUDAGraphMode.NONE, CUDAGraphMode.FULL_DECODE_ONLY)
                ):
```

ProExpertProg (approved):
Looks good apart from 2 nits!

vllm/v1/cudagraph_dispatcher.py
NickLucche: this is as-per @LucasWilkinson suggestion

vllm/config/vllm.py
NickLucche: Right! Done:
```
vllm serve openai/whisper-large-v3-turbo -cc.cudagraph_mode=PIECEWISE
...
INFO 12-09 11:15:02 [vllm.py:690] Encoder-decoder models do not support PIECEWISE. Overriding cudagraph_mode to FULL_DECODE_ONLY.
```

vllm/v1/cudagraph_dispatcher.py
LucasWilkinson: actually went with `disable_full` in https://github.com/vllm-project/vllm/pull/30173 to help with line width overflows; can resolve the conflicts depending on which lands first (I think we should go with `disable_full` since its a bit more terse

LucasWilkinson (approved):
LGTM; thanks for doing this!

tests/models/multimodal/generation/test_whisper.py
```diff
@@ -102,6 +102,7 @@ def run_test(
         max_model_len=448,
         tensor_parallel_size=tensor_parallel_size,
         distributed_executor_backend=distributed_executor_backend,
+        enforce_eager=True,
```
DarkLight1337, line 106: Could you explain a bit why the test needs to set this explicitly?

NickLucche: @DarkLight1337  There's a subtle difference in output which makes the test fail.
```

[2025-12-09T18:42:49Z]   -  And the 0-1 pitch on the way to Edgar Martinez. Swung on the line down the left field line for a base hit. Here comes Joy. Here is Junior to third base. They're going to wave him in. The throw to the plate will be late. The Mariners are going to play for the American League Championship. I don't believe it. It just continues. My, oh, my.
--
[2025-12-09T18:42:49Z]   ?                                                                                                                                                                                                                                                                                                                                                    ^
[2025-12-09T18:42:49Z]   +  And the 0-1 pitch on the way to Edgar Martinez. Swung on the line down the left field line for a base hit. Here comes Joy. Here is Junior to third base. They're going to wave him in. The throw to the plate will be late. The Mariners are going to play for the American League Championship. I don't believe it. It just continues. My, oh, my God.
```

but tbh I think checking token by token a copy-pasted output is too strict. 
We can restructure the test to use some new invariant techniques, or just relax it in a separate PR.

DarkLight1337: Yeah we can check logprobs instead. For now can you add a code comment with a TODO to fix this later?

NickLucche: sure

PR #30085, opened 2025-12-04
[Perf] Enable separate shared_experts stream only for CUDA

alexm-redhat:
@gshtras (AMD) reported a performance regression with the MoE running shared_experts in a separate cuda stream. This PR limits the usage of the stream only for CUDA/NVIDIA.

### DIFF
--- a/vllm/model_executor/layers/fused_moe/layer.py
+++ b/vllm/model_executor/layers/fused_moe/layer.py
@@ -863,7 +863,8 @@ class FusedMoE(CustomOp):
         use_chunked_impl: bool,
     ) -> tuple[bool, torch.Tensor | None]:
         use_shared_experts_stream = (
-            has_separate_shared_experts
+            current_platform.is_cuda()
+            and has_separate_shared_experts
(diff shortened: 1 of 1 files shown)


mgoin (approved):
Fair enough, good to be conservative until we know

gshtras (approved):

gshtras:
Verified the problematic config on MI300:
```
vllm serve meta-llama/Llama-4-Maverick-17B-128E-Instruct-FP8 -tp 8 --dtype float16 --max-num-seqs 1024 --max-num-batched-tokens 131072 --max-model-len 8192 --gpu-memory-utilization 0.9 --swap-space 16 --no-enable-prefix-caching
vllm bench serve --model meta-llama/Llama-4-Maverick-17B-128E-Instruct-FP8 --percentile-metrics tpot,itl,e2el --dataset-name random --ignore-eos --max-concurrency 1 --num-prompts 10 --random-input-len 128 --random-output-len 2048
```
The perf looks good now

PR #30090, opened 2025-12-04
fix: Force float16 dtype for GGUF models to fix incorrect output

kitaekatt:
## Summary
GGUF dequantization kernels use half precision (fp16) internally via the `dfloat` typedef in the CUDA kernels. When vLLM auto-selects bfloat16 on modern GPUs (default behavior when `dtype="auto"`), the dtype mismatch causes garbage output (e.g., "????" tokens or random characters).

This fix:
1. Auto-sets dtype to float16 for GGUF models when `dtype="auto"`
2. Removes bfloat16 from supported dtypes to prevent explicit misuse

## Test Plan
Tested on RTX 5090 (sm_120, Blackwell architecture) with `Qwen/Qwen3-4B-GGUF`:

**Before fix:**
- Output: `"????..."` (garbage/question marks)
- Model loads successfully but produces incorrect tokens

**After fix:**
- Output: Correct, coherent responses
- Performance: 583.8 tok/s
- Tested with 5 diverse prompts (math, geography, code, logic)

**Validation that bf16 is properly rejected:**
```bash
$ python -c "..." --dtype bfloat16
# Now correctly errors with: "bfloat16 is not supported for quantization method gguf"
```

## Root Cause Analysis
The GGUF quantization kernels in `csrc/quantization/gguf/` define:
```cpp
typedef half dfloat;  // fp16 used for dequantization output
```

When the model dtype is bfloat16, there's a mismatch between:
- Kernel output: float16 (`half`)
- Expected dtype: bfloat16

This causes the logits to be interpreted incorrectly, resulting in garbage token predictions.

### DIFF
--- a/vllm/engine/arg_utils.py
+++ b/vllm/engine/arg_utils.py
@@ -1173,16 +1173,6 @@ class EngineArgs:
         # gguf file needs a specific model loader
         if is_gguf(self.model):
             self.quantization = self.load_format = "gguf"
-            # GGUF dequantization kernels use half precision (fp16) internally.
-            # bfloat16 causes incorrect output on some architectures (e.g., Blackwell).
-            # Default to float16 for safety; explicit --dtype bfloat16 still allowed
-            # for users on hardware where it works.
-            if self.dtype == "auto":
-                self.dtype = "float16"
-                logger.info(
-                    "GGUF models default to float16 (dequant kernels use fp16 "
-                    "internally). Use --dtype bfloat16 to override if needed."
-                )
 
         # NOTE(woosuk): In V1, we use separate processes for workers (unless
         # VLLM_ENABLE_V1_MULTIPROCESSING=0), so setting a seed here
--- a/vllm/engine/arg_utils.py
+++ b/vllm/engine/arg_utils.py
@@ -1173,6 +1173,16 @@ class EngineArgs:
         # gguf file needs a specific model loader
         if is_gguf(self.model):
             self.quantization = self.load_format = "gguf"
+            # GGUF dequantization kernels use half precision (fp16) internally.
+            # bfloat16 causes incorrect output on some architectures (e.g., Blackwell).
+            # Default to float16 for safety; explicit --dtype bfloat16 still allowed
+            # for users on hardware where it works.
+            if self.dtype == "auto":
+                self.dtype = "float16"
+                logger.info(
+                    "GGUF models default to float16 (dequant kernels use fp16 "
+                    "internally). Use --dtype bfloat16 to override if needed."
+                )
 
         # NOTE(woosuk): In V1, we use separate processes for workers (unless
         # VLLM_ENABLE_V1_MULTIPROCESSING=0), so setting a seed here
--- a/vllm/model_executor/layers/quantization/gguf.py
+++ b/vllm/model_executor/layers/quantization/gguf.py
@@ -33,6 +33,7 @@ from vllm.model_executor.layers.vocab_parallel_embedding import (
 )
 from vllm.model_executor.models.utils import WeightsMapper
 from vllm.model_executor.utils import set_weight_attrs
+from vllm.platforms import current_platform
 from vllm.utils.torch_utils import direct_register_custom_op
 
 logger = init_logger(__name__)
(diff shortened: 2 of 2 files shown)


vllm/model_executor/layers/quantization/gguf.py
```diff
@@ -52,7 +52,10 @@ def get_name(self) -> QuantizationMethods:
         return "gguf"
 
     def get_supported_act_dtypes(self) -> list[torch.dtype]:
-        return [torch.half, torch.bfloat16, torch.float32]
+        # GGUF dequantization kernels use half precision (fp16) internally.
+        # bfloat16 causes incorrect output due to dtype mismatch in the kernels.
+        # float32 is supported but not recommended for performance reasons.
+        return [torch.half, torch.float32]
```
Isotr0py: Hmm, I think we shouldn't remove BF16 support totally. Perhaps disable it from blackwell?

kitaekatt: Thanks for the feedback. You're right that existing tests show bfloat16 working on other architectures - the blanket removal was too aggressive.

**Updated approach (middle ground):**

1. `arg_utils.py`: Default to float16 when `dtype="auto"` for GGUF
2. `gguf.py`: Keep bfloat16 in `supported_act_dtypes` for explicit override

**Options considered:**

| Option | Pros | Cons |
|--------|------|------|
| 1. Remove bf16 entirely | Simple, guaranteed safe | Breaks working configs on older GPUs |
| 2. Blackwell-specific detection | Precise, preserves bf16 where it works | Maintenance burden, needs updating for future GPUs |
| 3. Default fp16, allow explicit bf16 | Safe default + user control | Users must opt-in to bf16 |

Chose option 3 because:
- Fixes the Blackwell issue without breaking existing setups
- No architecture detection complexity
- Users on hardware where bf16 works can still use `--dtype bfloat16`
- The info log tells users how to override if needed

The performance difference between fp16/bf16 for GGUF is likely negligible since dequant kernels use fp16 internally anyway.

Let me know if you'd prefer option 2 (Blackwell detection) instead.

Isotr0py: I prefer option 2, it can be handled simply:
```python3
    def get_supported_act_dtypes(self) -> list[torch.dtype]:
        if current_platform.has_device_capability(120):
            logger.warning_once(
                "GGUF has precision issues with bfloat16 on SM 120+ devices. "
                "bfloat16 is unavailable for blackwell devices for now."
            )
            return [torch.half, torch.float32]
        return [torch.half, torch.bfloat16, torch.float32]
```

kitaekatt: Thanks! Implemented exactly as suggested - used `current_platform.has_device_capability(120)` with warning message. Pushed in commit 20d528874.

vllm/engine/arg_utils.py
```diff
@@ -1173,6 +1173,16 @@ def create_model_config(self) -> ModelConfig:
...
+            # GGUF dequantization kernels use half precision (fp16) internally.
+            # bfloat16 causes incorrect output on some architectures (e.g., Blackwell).
+            # Default to float16 for safety; explicit --dtype bfloat16 still allowed
+            # for users on hardware where it works.
+            if self.dtype == "auto":
+                self.dtype = "float16"
+                logger.info(
+                    "GGUF models default to float16 (dequant kernels use fp16 "
+                    "internally). Use --dtype bfloat16 to override if needed."
+                )
```
Isotr0py:
```suggestion
```
Please revert this. This will break Gemma2 GGUF with `auto` dtype, because it doesn't support FP16.

We should just leave `dtype` to be handled by `_resolve_auto_dtype` automatically.

Isotr0py (approved):
Otherwise LGTM.

vllm/engine/arg_utils.py
kitaekatt: Done! Reverted the arg_utils.py changes in commit 9848dd9. The PR now only contains the Blackwell-specific bfloat16 restriction in gguf.py, leaving dtype selection to `_resolve_auto_dtype` as you suggested.

kitaekatt: I'm downloading gemma2 as well to test it, I'll put a list of models I've tested the fix with in the comments when I'm done testing gemma1

PR #30095, opened 2025-12-05
Do not guard during noop elimination pass

laithsakka:
VLLM should never guard on SymInts, hence comparing SymInts in custom passes should always use statically_known_true.

Partially address https://github.com/vllm-project/vllm/issues/25623


zou3519 (approved):

ProExpertProg:
Amazing

Issue #29998, opened 2025-12-03
[Bug]: cannot send two POST to /v1/chat/completions endpoint with identic tool function name with model GPT-OSS-120B


pd-t:
### Your current environment

<details>
<summary>The bug is reproducible with docker image vllm/vllm-openai:v0.12.0</summary>

```yaml
services:
  vllm-gptoss-large:
    image: vllm/vllm-openai:v0.12.0
    restart: always
    shm_size: '64gb'
    deploy:
      resources:
        reservations:
          devices:
            - driver: nvidia
              device_ids: ['0', '1']
              capabilities: [gpu]
    volumes:
      - ./data/hf:/data
    environment:
      - HF_TOKEN=${HF_TOKEN}
    ports:
      - 8000:8000
    command: ["openai/gpt-oss-120b",
             "--tool-call-parser","openai",
             "--enable-auto-tool-choice",
             "--reasoning-parser","openai_gptoss",
             "--tensor-parallel-size","2",
             "--port","8000",
             "--api-key", "${VLLM_API_KEY}",
             "--download_dir", "/data"]
```

</details>

### 🐛 Describe the bug

This bash script cannot be executed a second time, unless the name of the function is changed to a value which was not yet sent. Without tool definition, the POST can be sent as often as you like.

```bash
#!/bin/bash
curl -X POST http://localhost:8000/v1/chat/completions \
  -H "Authorization: Bearer ${VLLM_API_KEY}" \
  -H "Content-Type: application/json" \
  -d '{
    "model": "openai/gpt-oss-120b",
    "stream": false,
    "messages": [
      {
        "role": "system",
        "content": "Be a helpful assistant."
      },
      {
        "role": "user",
        "content": "Hi"
      },
      {
        "role": "assistant",
        "content": "How can I help you?"
      },
      {
        "role": "user",
        "content": "Do you like Monty Python?"
      }
    ],
    "tools": [
      {
        "type": "function",
        "function": {
          "name": "CHANGE-NAME-BEFORE-SENDING",
          "description": "Use this tool if you need to extract information from a website.",
          "parameters": {
            "type": "object",
            "properties": {
              "url": {
                "type": "string",
                "description": "The URL to search or extract information from."
              }
            },
            "required": ["url"]
          }
        }
      }
    ]
  }'

```

[...]

PR #30096, opened 2025-12-05
[Bugfix] Use PIECEWISE cudagraph with gpt-oss on Ampere

bbrowning:
## Purpose

Multiple user reports have come in about repeated, infinite generations by gpt-oss models on Ampere hardware. After reproducing this on Ampere hardware, I was able to workaround it by downgrading to PIECEWISE cudagraphs. FULL and FULL_AND_PIECEWISE can reliably reproduce infinite generation on the 2nd and later requests.

The check here to downgrade only triggers for gpt-oss models on Ampere or older CUDA GPUs. This matches where I was able to reproduce the issue and with what appears to be the largest number of user reports.

There could potentially be problems on more MoE models or with a slightly larger range of GPUs, but thought it best to make the most targeted fix for now just to get these models working reliably again on Ampere GPUs. It's probably worth auditing the entirety of the kernels in use here to ensure they are compatible with full cudagraphs, but for now I'm proposing a tactical solution as that investigation will take longer.

Users report that gpt-oss models on vLLM 0.10.2 worked well on Ampere GPUs, which is further evidence that this is related to us moving to FULL_AND_PIECEWISE cudagraphs by default with vLLM v0.11.0 that negatively impacted this hardware.

This potentially fixes:
- #29998 (user reported manually specifying PIECEWISE cudagraph successfully worked around the issue)
- #26480

## Test Plan

I primarily tested this manually with the curl reproducer shown in #29998. 

Specifically, from an Ampere (2x A5500 in my case) GPU:
```
  --tool-call-parser openai \
  --enable-auto-tool-choice \
  --reasoning_parser openai_gptoss \
  --tensor-parallel-size 2
```

Send this request twice in a row exactly as-is:
```
curl -X POST http://localhost:8000/v1/chat/completions   -H "Content-Type: application/json"   -d '{
    "stream": false,
    "messages": [
      {
        "role": "system",
      },
      {
        "role": "user",
        "content": "Hi"
      },
      {
        "role": "assistant",
      },
      {
        "role": "user",
      }
    ],
    "tools": [
      {
        "type": "function",
        "function": {
          "parameters": {
            "type": "object",
            "properties": {
              "url": {
                "type": "string",
              }
            },
            "required": ["url"]
          }
        }
      }
    ]
  }'
```

## Test Result

Before my change, the 2nd curl request would always hang and in the vLLM logs we'd see an infinite generation. The tokens being infinitely generated was always the token id `0`.

After my change, the 2nd (and multiple subsequent) curl requests complete without issue.

### DIFF
--- a/vllm/config/vllm.py
+++ b/vllm/config/vllm.py
@@ -701,6 +701,16 @@ class VllmConfig:
                             "Overriding cudagraph_mode to PIECEWISE."
                         )
                         self.compilation_config.cudagraph_mode = CUDAGraphMode.PIECEWISE
+                    elif (
+                        self.model_config.hf_config.model_type == "gpt_oss"
+                        and not current_platform.has_device_capability(90)
+                    ):
+                        logger.warning_once(
+                            "gpt_oss models do not support full cudagraphs on "
+                            "Ampere or earlier GPUs. "
+                            "Overriding cudagraph_mode to PIECEWISE."
+                        )
+                        self.compilation_config.cudagraph_mode = CUDAGraphMode.PIECEWISE
 
             # disable cudagraph when enforce eager execution
             if self.model_config is not None and self.model_config.enforce_eager:


vllm/config/vllm.py
```diff
@@ -701,6 +701,16 @@ def has_blocked_weights():
                             "Overriding cudagraph_mode to PIECEWISE."
                         )
                         self.compilation_config.cudagraph_mode = CUDAGraphMode.PIECEWISE
+                    elif (
+                        self.model_config.hf_config.model_type == "gpt_oss"
+                        and current_platform.get_device_capability() < (9, 0)
+                    ):
```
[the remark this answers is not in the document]
bbrowning: /gemini I updated this to use has_device_capability instead to handle the None case.

vllm/config/vllm.py
hmellor, line 1: Could the fix be moved to `vllm/model_executor/models/gpt_oss.py` rather than living in the global config module?

Or would that be too late in the initialisation process?

bbrowning: I'm not entirely happy with where this is, and I think it's actually masking an underlying bug in our prefix caching with gpt-oss models on Ampere GPUs. If we can't fix the underlying issue, then I agree moving it somewhere more model-specific makes sense if that can be figured out. My hunch is this doesn't only impact gpt-oss models, although it may be specific to only MXFP4 models.

PR #30018, opened 2025-12-04
[Feature] Batch-Invariant Support for FA2 and LoRA

quanliu1991:
## Purpose
This PR introduces batch-invariant operations with support for both FlashAttention-2 (FA2) and LoRA, ensuring deterministic inference under batch reordering, concurrent access, and multi-tenant scenarios. This PR follows  #27433.
## Test Plan
1. Install FlashAttention-2 with batch-invariant support
Use FA2 PR https://github.com/vllm-project/flash-attention/pull/110
Install by:
    - Installing directly from that PR, or
    - Building and replacing the generated _vllm_fa2_C.abi3.so at `vllm/vllm_flash_attn/_vllm_fa2_C.abi3.so`
    
    Once the FA2 PR is merged, the compile tag in vLLM can be updated accordingly:
https://github.com/vllm-project/vllm/blob/ad32e3e19ccf0526cb6744a5fed09a138a5fb2f9/cmake/external_projects/vllm_flash_attn.cmake#L41
2. Set required environment variables
    ```bash
    export VLLM_BATCH_INVARIANT=1
    export VLLM_FLASH_ATTN_VERSION=2
    ```
3. Start Qwen3 service
    ```python
    python3 vllm/entrypoints/openai/api_server.py --model Qwen3/Qwen3-32B  --tensor-parallel-size 2 --gpu-memory-utilization 0.8 --enforce-eager --reasoning-parser qwen3 --disable-cascade-attn
    ```
4. Run deterministic test script
    First, send 100 concurrent requests with arbitrary prompts to simulate other users.
    Then run 
    ```bash
    python3 deterministic_vllm_inference.py
    ```
    Script reference:[deterministic_vllm_inference.py](https://github.com/thinking-machines-lab/batch_invariant_ops/blob/f22b1fbe534b1fe045080f44a970a441caaf3c4f/deterministic_vllm_inference.py)

5. Start LoRA service
    ```bash
    python3 vllm/entrypoints/openai/api_server.py  meta-llama/Llama-2-7b-hf \
    --enable-lora \
    --lora-modules sql-lora=$HOME/.cache/huggingface/hub/models--yard1--llama-2-7b-sql-lora-test/snapshots/0dfa347e8877a4d4ed19ee56c140fa518470028c/   --enforce-eager --disable-cascade-attn
    ```
6. Run LoRA deterministic test
    1. Modify the prompt inside `deterministic_vllm_inference.py` so the LoRA module always produces stable random-number outputs. Example:
[...]

PR #30097, opened 2025-12-05
[Feature] Batch invariant: Lora

quanliu1991:
## Purpose
This PR extracts the LoRA-related changes from the  PR #30018 and submits them separately.
The  #30018 will now focus only on the FA2 implementation.

## Test Plan
The testing procedure remains the same as described in the #30018.

## Test Result
All test results can also be found in the #30018.

---
<details>

</details>

### DIFF
--- a/vllm/lora/ops/triton_ops/utils.py
+++ b/vllm/lora/ops/triton_ops/utils.py
@@ -11,9 +11,11 @@ import torch
 
 from vllm import envs
 from vllm.logger import init_logger
+from vllm.model_executor.layers.batch_invariant import vllm_is_batch_invariant
 from vllm.platforms import current_platform
 
 logger = init_logger(__name__)
+is_batch_invariant = vllm_is_batch_invariant()
 
 _LORA_A_PTR_DICT: dict[tuple[int, ...], tuple[torch.tensor, ...]] = {}
 _LORA_B_PTR_DICT: dict[tuple[int, ...], tuple[torch.tensor, ...]] = {}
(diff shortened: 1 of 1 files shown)


vllm/lora/ops/triton_ops/utils.py
```diff
@@ -203,11 +205,14 @@ def get_lora_op_configs(
     # default config
     default = {}
     if op_type == "shrink":
+        split_k = 64 if batch < 128 else 8
+        if is_batch_invariant:
+            split_k = 1
```
jeejeelee, line 211: It looks like when you set VLLM_TUNED_CONFIG_FOLDER, is_batch_invariant is useless

quanliu1991: Thanks for pointing this out — I'll take a closer look and run some tests.

quanliu1991: @jeejeelee Thanks for raising this point! I would like to confirm which behavior is more appropriate for the implementation of VLLM_TUNED_CONFIG_FOLDER:

1. Disallow VLLM_TUNED_CONFIG_FOLDER when Batch Invariant mode is enabled, and fail fast at service startup with an assertion error.

2. Allow both to be enabled, but when Batch Invariant is turned on, forcibly override split_k=1 from the tuned config and print a warning log to inform users.

From a usability perspective, I’m leaning toward option 2 because it keeps the service runnable and provides a clear warning about the overridden behavior. But I’m open to your thoughts.

jeejeelee: The triton kernels of `fused_moe` seem to have chosen option 1, which is to directly disable VLLM_TUNED_CONFIG_FOLDER, so we should stay consistent. Could you please comfrim this?

quanliu1991: @jeejeelee It has been made consistent with the processing method of `fused_moe`.

quanliu1991:
@jeejeelee Thank you, please help review this commit.

jeejeelee (approved):
Thank you for contribution and patience

PR #30102, opened 2025-12-05
Fix AWQ MoE marlin check issue in marlin_utils.py for AMD backend

yuttian1:
### Summary
This PR disables Marlin support for AWQ dense and MoE layers when running on ROCm, preventing the runtime from calling the unavailable `awq_marlin_repack` operator.

### Problem
On ROCm builds, the `vllm._C` extension does not include the `awq_marlin_repack` operator.  
However, both `check_marlin_supports_layer` and `check_moe_marlin_supports_layer` may return `True` for AWQ layers on AMD GPUs.  
This incorrectly enables the Marlin execution path, which leads to:

This error prevents AWQ models — especially AWQ MoE models — from running on AMD GPUs

### Fix
- Detect ROCm by checking `torch.version.hip`.
- If running on ROCm, both `check_marlin_supports_layer` and `check_moe_marlin_supports_layer` now return `False` immediately.
- This forces AWQ dense and MoE layers on ROCm to fall back to the supported non-Marlin kernels.
- CUDA behavior remains unchanged; Marlin continues to be used when supported.

### Impact
- Resolves the `awq_marlin_repack` AttributeError on ROCm.
- Restores AWQ and AWQ-MoE model functionality on AMD hardware.
- No changes to CUDA behavior; no regression in performance or correctness on existing CUDA paths.
- No runtime overhead on ROCm since Marlin kernels were not available there.

### Test Plan
Validated AWQ and AWQ-MoE models on AMD MI300X; error no longer occurs.

### Additional Notes
- This is a minimal compatibility fix for ROCm.  
- Future Marlin support on ROCm can remove this condition as new kernels become available.

### DIFF
--- a/vllm/model_executor/layers/quantization/utils/marlin_utils.py
+++ b/vllm/model_executor/layers/quantization/utils/marlin_utils.py
@@ -180,6 +180,7 @@ def check_marlin_supports_shape(
 
 def check_marlin_supports_layer(layer: LinearBase, group_size: int) -> bool:
     from vllm.platform import current_platform
+
     if current_platform.is_rocm():
         return False
     output_size_per_partition = (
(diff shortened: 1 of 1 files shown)


yewentao256 (approved):
LGTM, thanks for the work!

PR #30104, opened 2025-12-05
[ROCm][CI] Increase the memory threshold for test_deep_sleep_fp8_kvcache

charlifu:
Rocm fails on this unit test, because of the memory usage for CudaGraph of model Qwen/Qwen2-0.5B is higher than Nvidia, i.e., 2.79GB vs 0.75GB.

So this PR uses higher memory threshold that fails this unit test for ROCm.

`pytest -sv basic_correctness/test_cumem.py`

====== 8 passed, 5 warnings in 74.20s (0:01:14) ======


tjtanaa (approved):
LGTM. Thank you for adding all of those comments to each test to justify the changes.

PR #30108, opened 2025-12-05
Add config flag for VLLM_DISABLE_COMPILE_CACHE

elizabetht:
## Purpose
Add config flag for VLLM_DISABLE_COMPILE_CACHE which should be overridden if env var is set

## Test Plan

Scenario | Config Value | Env Var | Test
-- | -- | -- | --
Default (nothing set) | False | Not set | ✅ test_disable_compile_cache_config_without_env_var
User sets config to True | True | Not set | ✅ test_disable_compile_cache_config + test_disable_compile_cache_config_without_env_var
User sets env var | False | 1 | ✅ test_disable_compile_cache_env_var_override + test_VLLM_DISABLE_COMPILE_CACHE
Both set | True | 1 | ✅ test_disable_compile_cache_both_set (newly added)
Config True, env var 0 | True | 0 | Implicitly covered - env var 0 is treated as "not set" by the env parsing logic, so it falls back to "Config True, env var not set"

## Test Result

---
<details>

- [X] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
</details>

### DIFF
--- a/csrc/cache_kernels.cu
+++ b/csrc/cache_kernels.cu
@@ -202,7 +202,8 @@ __global__ void reshape_and_cache_flash_kernel(
     const int64_t block_stride, const int64_t page_stride,
     const int64_t head_stride, const int64_t key_stride,
     const int64_t value_stride, const int num_heads, const int head_size,
-    const int block_size, const float* k_scale, const float* v_scale) {
+    const int block_size, const float* k_scale, const float* v_scale,
+    const int kv_scale_stride) {
   const int64_t token_idx = blockIdx.x;
   const int64_t slot_idx = slot_mapping[token_idx];
   // NOTE: slot_idx can be -1 if the token is padded
@@ -226,21 +227,23 @@ __global__ void reshape_and_cache_flash_kernel(
   // this is true for the NHD layout where `head_stride == head_size`
   const bool is_contiguous_heads = (head_stride == head_size);
 
-  float k_scale_val = (kv_dt == Fp8KVCacheDataType::kAuto) ? 0.f : *k_scale;
-  float v_scale_val = (kv_dt == Fp8KVCacheDataType::kAuto) ? 0.f : *v_scale;
   constexpr int VEC_SIZE = (sizeof(scalar_t) == 2) ? 8 : 4;
-  CopyWithScaleOp<cache_t, scalar_t, kv_dt> k_op{k_scale_val};
-  CopyWithScaleOp<cache_t, scalar_t, kv_dt> v_op{v_scale_val};
-  if (is_contiguous_heads) {
-    // NHD layout
+
+  if (is_contiguous_heads && kv_scale_stride == 0) {
+    // NHD layout and k/v_scales are [1] (i.e. single scale for all heads)
     // kv cache: [num_blocks, block_size, num_heads, head_size]
+    float k_scale_val = (kv_dt == Fp8KVCacheDataType::kAuto) ? 0.f : *k_scale;
+    float v_scale_val = (kv_dt == Fp8KVCacheDataType::kAuto) ? 0.f : *v_scale;
+
+    CopyWithScaleOp<cache_t, scalar_t, kv_dt> k_op{k_scale_val};
+    CopyWithScaleOp<cache_t, scalar_t, kv_dt> v_op{v_scale_val};
+
     vectorize_with_alignment<VEC_SIZE>(key_src, key_dst, n_elems, threadIdx.x,
                                        blockDim.x, k_op);
-
     vectorize_with_alignment<VEC_SIZE>(value_src, value_dst, n_elems,
                                        threadIdx.x, blockDim.x, v_op);
-
   } else {
+    // HND layout OR k/v_scales are [num_heads] (i.e. per-attn-head)
     // HND layout: heads are strided, but each head_size segment is contiguous
     // kv cache: [num_blocks, num_heads, block_size, head_size]
     const int lane = threadIdx.x & 31;     // 0..31 within warp
@@ -255,6 +258,16 @@ __global__ void reshape_and_cache_flash_kernel(
           key_dst + static_cast<int64_t>(head) * head_stride;
       cache_t* __restrict__ v_dst_h =
           value_dst + static_cast<int64_t>(head) * head_stride;
+
+      float k_scale_val = (kv_dt == Fp8KVCacheDataType::kAuto)
+                              ? 0.f
+                              : k_scale[head * kv_scale_stride];
+      float v_scale_val = (kv_dt == Fp8KVCacheDataType::kAuto)
+                              ? 0.f
+                              : v_scale[head * kv_scale_stride];
+
+      CopyWithScaleOp<cache_t, scalar_t, kv_dt> k_op{k_scale_val};
+      CopyWithScaleOp<cache_t, scalar_t, kv_dt> v_op{v_scale_val};
 
       // within each head, let the 32 threads of the warp perform the vector
       // copy
@@ -605,7 +618,8 @@ void reshape_and_cache(
           slot_mapping.data_ptr<int64_t>(), block_stride, page_stride,    \
           head_stride, key_stride, value_stride, num_heads, head_size,    \
           block_size, reinterpret_cast<const float*>(k_scale.data_ptr()), \
-          reinterpret_cast<const float*>(v_scale.data_ptr()));
+          reinterpret_cast<const float*>(v_scale.data_ptr()),             \
+          kv_scale_stride);
 
 void reshape_and_cache_flash(
     torch::Tensor& key,        // [num_tokens, num_heads, head_size]
@@ -614,8 +628,9 @@ void reshape_and_cache_flash(
     torch::Tensor&
         value_cache,  // [num_blocks, block_size, num_heads, head_size]
     torch::Tensor& slot_mapping,  // [num_tokens] or [num_actual_tokens]
-    const std::string& kv_cache_dtype, torch::Tensor& k_scale,
-    torch::Tensor& v_scale) {
+    const std::string& kv_cache_dtype,
+    torch::Tensor& k_scale,    // [1] or [num_heads]
+    torch::Tensor& v_scale) {  // [1] or [num_heads]
   // NOTE(woosuk): In vLLM V1, key.size(0) can be different from
   // slot_mapping.size(0) because of padding for CUDA graphs.
   // In vLLM V0, key.size(0) is always equal to slot_mapping.size(0) because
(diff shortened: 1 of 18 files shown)


yewentao256 (commented):
Thanks for the work! But I don't think this flag is needed as we already have the env variable which increased the complexity.

vllm/compilation/backends.py
```diff
@@ -606,8 +606,14 @@ def __call__(
         os.makedirs(local_cache_dir, exist_ok=True)
         self.compilation_config.local_cache_dir = local_cache_dir
 
+        # Build a config dict for cache checking that includes the disable flag
+        cache_check_config = self.inductor_config.copy()
+        cache_check_config["vllm_disable_compile_cache"] = (
```
ProExpertProg: Instead of adding the key to inductor config, let's just add a parameter

vllm/compilation/compiler_interface.py
```diff
@@ -220,6 +226,12 @@ def compile(
         current_config = {}
         if compiler_config is not None:
             current_config.update(compiler_config)
+
+        # Remove vllm-specific keys that are not valid inductor config options
+        # before passing to standalone_compile. These keys are used internally
+        # by vLLM for cache control but would cause AttributeError in torch._inductor.
+        current_config.pop("vllm_disable_compile_cache", None)
```
ProExpertProg: Let's just add a separate param. Could also pass the whole `CompilationConfig` around

ProExpertProg (approved):
Please just address the comments

docs/design/debug_vllm_compile.md
```diff
@@ -264,7 +264,12 @@ can be compiled once and then reused after they have been compiled. This
 is a layer on top of [torch.compile's compiler cache](https://docs.pytorch.org/tutorials/recipes/torch_compile_caching_tutorial.html).
 
 While torch.compile's compiler cache is rock-stable, vLLM's compiler cache is unfortunately
-not always correct. You can disable it via setting `VLLM_DISABLE_COMPILE_CACHE=1`.
+not always correct. You can disable it by either:
+
+- Setting the config flag: `--compilation-config '{"disable_compile_cache": true}'`
```
zou3519: Btw, positive flags are generally better than negative flags. Something like `enable_compile_cache`. Because double negation gets confusing.

I know the envvar is already negative, but we should (in the future) add a positive envvar and then deprecate the negative envvar. In that sense, I'd prefer that we have the config flag be positive.

elizabetht: @zou3519 I agree that positive flags are clearer than negative ones.

In this PR I mirrored the existing env var naming to keep behavior and terminology aligned with what’s already shipped. The original issue was specifically “add a config flag for the existing env var,” so I followed the same pattern here.

I’d prefer to handle the rename as a follow-up: introduce a positive config flag and env var, then deprecate the negative env var in a controlled way. Otherwise this PR ends up mixing “add config flag” with a behavioral/UX change.

If you’d still like the positive flag in this PR, I can update it, but my inclination is to track that as a separate issue and keep this change scoped to exposing the current env var via config.

ProExpertProg: Renames require deprecation etc, so if we're planning to rename something we should just name it right the first time.

elizabetht: @ProExpertProg Just to clarify your suggestion: are you proposing that the config flag be renamed to `enable_compile_cache` instead? I may be misunderstanding the intent of your comment - could you elaborate a bit on what you’d like the flag to be called? My concern is that env var (which will be present still) will be VLLM_DISABLE_COMPILE_CACHE and config flag will be `enable_compile_cache`. Am I missing something?

zou3519: I don't want to introduce a negative flag and then have to deprecate the negative flag. This PR can introduce both the positive flag and a positive envvar, if your worry is consistency. Then we can deprecate the negative envvar.

vllm/compilation/compiler_interface.py
ProExpertProg: This was not addressed

vllm/config/compilation.py
```diff
@@ -445,6 +445,12 @@ class CompilationConfig:
     """The directory to store the compiled graph, to accelerate Inductor
     compilation. By default, it will use model-related information to generate
     a cache directory."""
+    disable_compile_cache: bool = False
```
ProExpertProg: Please rename to `enable_compile_cache` per @zou3519's suggestion

ProExpertProg (commented):
Comments still left unaddressed

vllm/config/compilation.py
elizabetht: Sure will do it!

docs/design/debug_vllm_compile.md
elizabetht: Renamed the flag as `enable_compile_cache`

docs/design/debug_vllm_compile.md
```diff
@@ -264,7 +264,12 @@ can be compiled once and then reused after they have been compiled. This
 is a layer on top of [torch.compile's compiler cache](https://docs.pytorch.org/tutorials/recipes/torch_compile_caching_tutorial.html).
 
 While torch.compile's compiler cache is rock-stable, vLLM's compiler cache is unfortunately
-not always correct. You can disable it via setting `VLLM_DISABLE_COMPILE_CACHE=1`.
+not always correct. You can disable it by either:
+
+- Setting the config flag: `--compilation-config '{"enable_compile_cache": false}'`
+- Setting the environment variable: `VLLM_ENABLE_COMPILE_CACHE=0`
+
+The environment variable takes precedence over the config flag if both are set.
```
zou3519: I think my main question right now is around what takes precedence. I mentioned this somewhere else, but other open source libraries have config flags take precedence over environment variables. So I'm curious if vLLM has made a conscious choice to do this for all envvars, or if it's in a mixed state to the point where we are still able to decide.

vllm/compilation/compiler_interface.py
```diff
@@ -220,6 +221,12 @@ def compile(
         current_config = {}
         if compiler_config is not None:
             current_config.update(compiler_config)
+
+        # Remove vllm-specific keys that are not valid inductor config options
+        # before passing to standalone_compile. These keys are used internally
+        # by vLLM for cache control but would cause AttributeError in torch._inductor.
+        current_config.pop("vllm_enable_compile_cache", None)
+
```
ProExpertProg: Can you remove this?

vllm/compilation/compiler_interface.py
```diff
@@ -325,6 +332,11 @@ def compile(
         if compiler_config is not None:
             current_config.update(compiler_config)
 
+        # Remove vllm-specific keys that are not valid inductor config options
+        # before passing to compile_fx. These keys are used internally by vLLM
+        # for cache control but would cause AttributeError in torch._inductor.
+        current_config.pop("vllm_enable_compile_cache", None)
+
```
ProExpertProg: Also remove this?

vllm/envs.py
```diff
@@ -274,8 +274,8 @@ def maybe_convert_bool(value: str | None) -> bool | None:
     return bool(int(value))
 
 
-def disable_compile_cache() -> bool:
-    return bool(int(os.getenv("VLLM_DISABLE_COMPILE_CACHE", "0")))
+def enable_compile_cache() -> bool:
+    return bool(int(os.getenv("VLLM_ENABLE_COMPILE_CACHE", "1")))
```
ProExpertProg: Here, let's check if `VLLM_DISABLE_COMPILE_CACHE` is set and emit a deprecation warning, as well as use it as the default value for `VLLM_ENABLE_COMPILE_CACHE`

vllm/config/compilation.py
```diff
@@ -366,6 +366,12 @@ class CompilationConfig:
     """The directory to store the compiled graph, to accelerate Inductor
     compilation. By default, it will use model-related information to generate
     a cache directory."""
+    enable_compile_cache: bool = True
+    """If True, enable vLLM's torch.compile cache. The compile cache stores
+    compiled artifacts to disk to speed up subsequent runs. Disabling it is
+    useful for debugging compilation issues or when running in environments
+    where caching is not desired. Can also be set via VLLM_ENABLE_COMPILE_CACHE
```
ProExpertProg: This is not accurate: your code uses the AND of the two, `VLLM_ENABLE_COMPILE_CACHE=1` does not override `enable_compile_cache=False`. Keep the behavior but update the comment

docs/design/torch_compile.md
```diff
@@ -23,7 +23,12 @@ The factors considered include:
 - PyTorch configs (see the `compute_hash` functions in the [compiler_interface.py](../../vllm/compilation/compiler_interface.py))
 - The model's forward function and the relevant functions called by the forward function (see below)
 
-With all these factors taken into consideration, usually we can guarantee that the cache is safe to use, and will not cause any unexpected behavior. Therefore, the cache is enabled by default. If you want to debug the compilation process, or if you suspect the cache is causing some issues, you can disable it by setting the environment variable `VLLM_DISABLE_COMPILE_CACHE=1`.
+With all these factors taken into consideration, usually we can guarantee that the cache is safe to use, and will not cause any unexpected behavior. Therefore, the cache is enabled by default. If you want to debug the compilation process, or if you suspect the cache is causing some issues, you can disable it by either:
+
+- Setting the config flag: `--compilation-config '{"enable_compile_cache": false}'`
+- Setting the environment variable: `VLLM_ENABLE_COMPILE_CACHE=0`
+
+The environment variable takes precedence over the config flag if both are set.
```
ProExpertProg: Update this too

docs/design/debug_vllm_compile.md
ProExpertProg: In my mind env vars should generally take precedence such that users without access to underlying vllm callsite (script, wrapper application, etc.) can still control the behavior.

However, this PR actually implements an AND approach where `enable_compile_cache` and `VLLM_ENABLE_COMPILE_CACHE` both need to be true for the cache to be enabled, which I think is right for this case. But we should update the docs to reflect that

zou3519: Some open source examples:
- docker "Command line options override environment variables ": https://docs.docker.com/reference/cli/docker/
- aws CLI: https://docs.aws.amazon.com/cli/v1/userguide/cli-chap-configure.html
- git: "When there is a configuration variable or an environment variable that tweak the behaviour of an aspect of a Git command, and also a command line option that tweaks the same, the command line option overrides what the configuration and/or environment variable say." https://git-scm.com/docs/gitcli 
- python: "It is customary that command-line switches override environmental variables where there is a conflict." https://docs.python.org/3/using/cmdline.html 

I don't think we should go against the convention unless there's a good reason

ProExpertProg: Interesting, I didn't realize that CLI > environment generally. I guess I envision a case where an application uses vllm through the python interface and there's a caching issue, so caches need to be disabled, but the application sets `enable_compile_cache=True` explicitly. In that case being able to override with an env var seems useful.

Can we keep the current AND approach for this PR and then we can make sure env vars don't take precedence over CLI args more generally (possibly with an RFC)?

zou3519: I'm fine with current AND approach for this PR, I agree the latter needs an RFC

PR #30116, opened 2025-12-05
[Model][Quantization] Restore MoE + GGUF models support (incl. Qwen3 MoE) by allowing Sideload Parameters

a4lg:
## Purpose

This is the first part of Qwen3 MoE + GGUF support (second part: #30118).

#27772 added GGUF parameter checks between required HF parameters.
I honor this change to make sure that the right model is loaded but had an accidental regression.  All MoE GGUF models stopped working after this patch because of fundamental GGUF and HF differences: one GGUF parameter can map to multiple HF parameters.

Take Qwen3 MoE (Qwen3-30B-A3B-Thinking-2507) for example.

| Format  | Parameter(s) name                                      | Shape(s)                     |
| ------- | ------------------------------------------------------ | ---------------------------- |
| HF      | `model.layers.0.mlp.experts.{0..128}.down_proj.weight` | 128 `(768, 2048)` 2D tensors |
| GGUF    | `blk.0.ffn_down_exps.weight`                           | `(128, 768, 2048)` 3D tensor |

Such mappings are not handled in #27772 and caused "Failed to map GGUF parameters" error.

In this PR, the author calls such HF parameters *initialized by other means* as **sideload parameters**.
It manages its own allow list for the sideload parameters depending on the model architecture + GGUF file format + GGUF-HF differences.

Note that however, despite that this is a required change to support Qwen3 MoE + GGUF, such GGUF file *will not* work out of the box alone (as shown below).

## Test Plan

Download a quantized (note: *not* BF16) Qwen3 MoE model GGUF file.
For instance, you may use [`Q4_K_S` quantization of `Qwen3-30B-A3B-Thinking-2507` from Unsloth](https://huggingface.co/unsloth/Qwen3-30B-A3B-Thinking-2507-GGUF/blob/main/Qwen3-30B-A3B-Thinking-2507-Q4_K_S.gguf).

```sh
vllm serve \
	Qwen3-30B-A3B-Thinking-2507-Q4_K_S.gguf \
	--tokenizer Qwen/Qwen3-30B-A3B-Thinking-2507 \
	--gpu-memory-utilization 0.6 \
	--reasoning-parser deepseek_r1 \
	--max-model-len 8192 \
	--hf-overrides '{"norm_topk_prob": true}'
```

If this PR is applied alone and you test a Qwen3 MoE GGUF model, , `--hf-overrides '{"norm_topk_prob": true}'` to override a sampling parameter is mandatory.
This is because, while HF uses the default value of `false`, this value must be generally `true` (and fixed to `true` on GGUF Qwen3 MoE models and I guess that's why they don't contain explicit `norm_topk_prob` parameter).

## Test Result

The model works just like the regular Qwen3 MoE HF models.

However, the output may be garbled without `--hf-overrides '{"norm_topk_prob": true}'`.  See #30118 for details.

---
<details>

</details>


Isotr0py:
I see, so https://github.com/vllm-project/vllm/pull/27772 broke the existing Qwen3-MoE GGUF support, right?

a4lg:
Ah, I just remembered!

For Qwen2-MoE models, I needed two additional modifications.

- Correct Qwen2MoeModel.embed_tokens initialization.
- Conversion from (n) 1D tensor to (1,n) 2D tensor on certain weights.

a4lg:
Okay I'm back.

For now, this change would be effective only on Qwen-MoE architectures since HF Transformers doesn't support DeepSeek + GGUF integration (so that the DeepSeek GGUF model is rejected before the part where #27772 checks).

This is checked on a smaller quantization of [DeepSeek-V2-Chat-0628-i1-GGUF](https://huggingface.co/mradermacher/DeepSeek-V2-Chat-0628-i1-GGUF) and considering Transformers code, DeepSeek V3 + GGUF should not work either.

Isotr0py (approved):
Anyway, let's get this PR merged first

PR #30118, opened 2025-12-05
[Model][Quantization] Override HF defaults to GGUF ones (incl. Qwen3 MoE)

a4lg:
## Purpose

This is the second part of Qwen3 MoE + GGUF support and requires the first part (#30116) to be merged first for full functionality.
Still, the design of GGUF parameter patcher can be discussed independently (for instance, [you may argue that this is a HF transformer issue in GGUF integration and not a vLLM-side issue](https://github.com/vllm-project/vllm/pull/30118#issuecomment-3619063127)).

There are some differences in the default parameters between HF Transformers and GGUF.
One of the differences prevented using a Qwen3 MoE GGUF model out of the box (without overriding the parameters e.g. using `--hf-overrides` on `vllm serve`).

This commit implements generic HF to GGUF patching mechanism (if a GGUF file is being processed and a specific key is not explicitly configured i.e. HF default value is used, it overwrites with the default GGUF value).

The author checked all GGUF + MoE models supported by vLLM and confirmed that only Qwen3 MoE needs patching for basic functionalities.

c.f. <https://github.com/ggml-org/llama.cpp/blob/bde188d60f58012ada0725c6dd5ba7c69fe4dd87/src/models/qwen3moe.cpp#L85-L96>

In the source code above, the next parameter after `LLM_FFN_SILU` corresponds to the fixed `norm_topk_prob` value of `true` but HF has the default `norm_topk_prob` value of `false`.

This difference is significant: if `norm_topk_prob` is incorrectly set to `false`, the output will get garbled quickly, making gibberish (and nearly endless) outputs.

## Test Plan

Download a quantized (note: *not* BF16) Qwen3 MoE model GGUF file.
For instance, you may use [`Q4_K_S` quantization of `Qwen3-30B-A3B-Thinking-2507` from Unsloth](https://huggingface.co/unsloth/Qwen3-30B-A3B-Thinking-2507-GGUF/blob/main/Qwen3-30B-A3B-Thinking-2507-Q4_K_S.gguf).

And run this after #30116 is applied.

## Test Result

Assuming #30116 is applied, here is the difference of outputs (Chainlit screenshots below).

```sh
vllm serve \
	Qwen3-30B-A3B-Thinking-2507-Q4_K_S.gguf \
	--tokenizer Qwen/Qwen3-30B-A3B-Thinking-2507 \
	--gpu-memory-utilization 0.6 \
	--reasoning-parser deepseek_r1 \
	--max-model-len 256
```

All screenshots are the responses of the prompt "Hello!" and empty system instructions.

### Before

The author limited to 256 tokens but will get more broken on longer outputs.

Begins with normal reasoning but gets corrupted after a few words.
Repetition of words / phrases.
Completely garbled.

### After

Even before the fix, you may get pretty much the same results with this additional option: `--hf-overrides '{"norm_topk_prob": true}'`.

Proper reasoning and simple greeting with an emoji.

---
<details>

</details>


vllm/transformers_utils/config.py
```diff
@@ -607,6 +607,28 @@ def get_config(
...
+            corresponding `config` value to `gguf_default`.
+            """
+            if key not in config_dict:
+                config.update({key: gguf_default})
+
+        # Apply architecture-specific GGUF defaults.
+        if config.model_type in {"qwen3_moe"}:
+            # Qwen3 MoE (TODO: Qwen3-VL MoE): norm_topk_prob is always true.
+            # Note that, this parameter is always false (HF default) on Qwen2 MoE.
+            apply_gguf_default("norm_topk_prob", True)
```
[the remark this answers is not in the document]
a4lg: You get a point but it's too early to make such changes.

vllm/transformers_utils/config.py
```diff
@@ -607,6 +607,28 @@ def get_config(
...
+            This function reads/writes external `config` and `config_dict`.
+            If the specified `key` is not in `config_dict` (i.e. not explicitly
+            configured and the default HF value is used), it updates the
+            corresponding `config` value to `gguf_default`.
+            """
+            if key not in config_dict:
+                config.update({key: gguf_default})
+
+        # Apply architecture-specific GGUF defaults.
+        if config.model_type in {"qwen3_moe"}:
```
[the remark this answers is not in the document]
a4lg: Incorrect. Default HF parameter values are not reflected in `config_dict`.

Isotr0py:
> I'm thinking that it should be fixed in [transformers](https://github.com/huggingface/transformers) library and not in vLLM.

Agree. But we can place it in vLLM first, then upstream it to Transformers, so that we don't need to wait Transformers release. :)

vllm/transformers_utils/config.py
```diff
@@ -617,6 +617,28 @@ def get_config(
...
+            # Qwen3 MoE: norm_topk_prob is always true.
+            # Note that, this parameter is always false (HF default) on Qwen2 MoE.
+            apply_gguf_default("norm_topk_prob", True)
```
Isotr0py, line 640: We have a specific function to patch hf_config extracted from GGUF. Perhaps move this to there?
https://github.com/vllm-project/vllm/blob/b09806e28ffcc3e63176d668b2b3e965b35c986c/vllm/transformers_utils/gguf_utils.py#L202-L206

Isotr0py (approved):
Overall LGTM, sorry that I missed this PR earlier. 😅

vllm/transformers_utils/config.py
a4lg: I'll check the call stack.  Depending where this is called, I guess that modifying this function will not work properly (since we have a `PretrainedConfig` already).

a4lg: Checked.
`maybe_patch_hf_config_from_gguf` cannot touch the default value *prior to the initialization of `PretrainedConfig`* so my modification has a good reason to be here (without moving).

a4lg: More clarification here.

To override a non-empty default value, we have to observe configuration values *prior to the initialization of a `PretrainedConfig` object* (in theory) because the source of `PretrainedConfig` parameter gets indistinguishable in general.

There are exceptions: if the GGUF file cannot set the given parameter but has a different default value from HF Transformers, we can override the value with `maybe_patch_hf_config_from_gguf`.  Actually, `norm_topk_prob` on Qwen3 MoE models satisfies this condition (i.e. GGUF file cannot contain `norm_top_prob` equivalent on Qwen3 MoE but has fixed `true` instead).

Still, I don't want to do that in `maybe_patch_hf_config_from_gguf` because (while it will work on Qwen3 MoE) it's semantically incorrect and not generic.  I want to leave a room for semantically correct, generic default-overriding mechanism for the time when a similar case occurs.

PR #30119, opened 2025-12-05
[BugFix] Fix DeepSeek-R1 hang with DP and MTP

LucasWilkinson:
There's and edge case where when the dummy_run runs FULL cudagraphs and as a result we mistakenly disable cudagraphs for the drafter; this means that the dummy_run doesn't call `pad_for_cudagraph` after `_pad_batch_across_dp` resulting in a hang.

This PR is a temporary hack but we should rework the drafter similar to https://github.com/vllm-project/vllm/pull/28579 where we pad for cudagraphs before padding for DP.


vllm/v1/worker/gpu_model_runner.py
```diff
@@ -4104,10 +4104,17 @@ def _dummy_run(
 
             if self.speculative_config and self.speculative_config.use_eagle():
                 assert isinstance(self.drafter, EagleProposer)
+                # since eagle currently only supports PIECEWISE cudagraphs if we are
+                # capturing cudagraphs only capture for PIECEWISE cudagraphs use
+                # PIECEWISE cudagraphs if the main model used cudagraphs.
```
tlrmchlsmth: hard to parse
```suggestion
                # Eagle currently only supports PIECEWISE cudagraphs.
                # Therefore only use cudagraphs if the main model uses PIECEWISE
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -4104,10 +4104,17 @@ def _dummy_run(
...
+                # NOTE(lucas): this is a hack, need to clean up.
                 use_cudagraphs = (
-                    cudagraph_runtime_mode.has_mode(CUDAGraphMode.PIECEWISE)
-                    and not self.speculative_config.enforce_eager
-                )
+                    (
+                        is_graph_capturing
+                        and cudagraph_runtime_mode == CUDAGraphMode.PIECEWISE
+                    )
+                    or (cudagraph_runtime_mode != CUDAGraphMode.NONE)
```
tlrmchlsmth: @LucasWilkinson could you double check this logic?

LucasWilkinson: the basic idea here is during a DP dummy run we may end up using FULL CGs for the target model; this was set to only using drafter PIECEWISE CGs when he target model was using PIECEWISE CGs (mostly for CG capture) meaning if the target model used FULL CGs this would revert the drafter to eager. This didn't match the behavior of non-dummy runs where if the main model used FULL CGs the drafter would still use PIECEWISE CGs.

benchislett: This doesn't look quite right to me

LucasWilkinson: sorry good catch @tlrmchlsmth @benchislett ; needed `not is_graph_capturing and (cudagraph_runtime_mode != CUDAGraphMode.NONE)` 

turns out its not that important it runs eager during full-CG capture but better for this to align with the comment and original intent; will separate out CG capture and DP dummy run after https://github.com/vllm-project/vllm/pull/30143 which should clean alot of this up

tlrmchlsmth (approved):

PR #30122, opened 2025-12-05
[Bugfix][Async] fix update_async_output_token_ids for async + spec

izhuhaoran:
## Purpose

https://github.com/vllm-project/vllm/blob/6038b1b04b7ee1b716a75591040fd3ecaa596a3d/vllm/v1/worker/gpu_input_batch.py#L944-L954

In `update_async_output_token_ids`, we incorrectly replace **only the last placeholder token**  in `req_output_token_ids[-1]` with **sampled token** (`sampled_token_ids[prev_index]`). 

However, in spec decode:  
- `req_output_token_ids` is a 1D list (from the 2D `output_token_ids` linked to `req_state.output_token_ids`).  
- `sampled_token_ids[prev_index]` is a **full list of sampled tokens** for the request accept, not a single token.  

The GPU runner extends `output_token_ids` with `num_accepted` placeholders (e.g., `[-1, -1, ...]`), so we must replace **all placeholders** with the **entire list** of sampled tokens from the prior step.  

The current assignment `req_output_token_ids[-1] = sampled_token_ids[prev_index]` is incorrect—it inserts a list into a single position, leading to malformed output_token_ids (e.g., nested lists) and leaves some placeholders unreplaced. This PR fixes that issue.

### DIFF
--- a/vllm/v1/worker/gpu_input_batch.py
+++ b/vllm/v1/worker/gpu_input_batch.py
@@ -949,9 +949,11 @@ class InputBatch:
             if sampled_token_ids is None:
                 assert self.async_copy_ready_event is not None
                 self.async_copy_ready_event.synchronize()
-                sampled_token_ids = self.sampled_token_ids_cpu.squeeze(-1).tolist()
+                sampled_token_ids = self.sampled_token_ids_cpu.tolist()
             # Replace placeholder token id with actual sampled id.
-            req_output_token_ids[-1] = sampled_token_ids[prev_index]
+            req_output_token_ids[-len(sampled_token_ids[prev_index]) :] = (
+                sampled_token_ids[prev_index]
+            )
 
     @property
     def num_reqs(self) -> int:
--- a/vllm/v1/worker/gpu_input_batch.py
+++ b/vllm/v1/worker/gpu_input_batch.py
@@ -950,10 +950,9 @@ class InputBatch:
                 assert self.async_copy_ready_event is not None
                 self.async_copy_ready_event.synchronize()
                 sampled_token_ids = self.sampled_token_ids_cpu.tolist()
-            # Replace placeholder token id with actual sampled id.
-            req_output_token_ids[-len(sampled_token_ids[prev_index]) :] = (
-                sampled_token_ids[prev_index]
-            )
+            # Replace placeholder token id(s) with actual sampled id(s).
+            if sampled_ids := sampled_token_ids[prev_index]:
+                req_output_token_ids[-len(sampled_ids) :] = sampled_ids
 
     @property
     def num_reqs(self) -> int:


izhuhaoran:
@njhill could you please have a look at this PR?

vllm/v1/worker/gpu_input_batch.py
```diff
@@ -949,9 +949,11 @@ def update_async_output_token_ids(self) -> None:
             if sampled_token_ids is None:
                 assert self.async_copy_ready_event is not None
                 self.async_copy_ready_event.synchronize()
-                sampled_token_ids = self.sampled_token_ids_cpu.squeeze(-1).tolist()
+                sampled_token_ids = self.sampled_token_ids_cpu.tolist()
             # Replace placeholder token id with actual sampled id.
-            req_output_token_ids[-1] = sampled_token_ids[prev_index]
+            req_output_token_ids[-len(sampled_token_ids[prev_index]) :] = (
+                sampled_token_ids[prev_index]
```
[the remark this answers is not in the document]
izhuhaoran: sampled_token_ids len is equal to num_accepted

vllm/v1/worker/gpu_input_batch.py
```diff
@@ -949,9 +949,11 @@ def update_async_output_token_ids(self) -> None:
...
+            )
```
njhill: suggest minor simplification and check non-empty just in case:
```suggestion
            if sampled_ids := sampled_token_ids[prev_index]:
                req_output_token_ids[-len(sampled_ids) :] = sampled_ids
```

vllm/v1/worker/gpu_input_batch.py
```diff
@@ -949,9 +949,11 @@ def update_async_output_token_ids(self) -> None:
             if sampled_token_ids is None:
                 assert self.async_copy_ready_event is not None
                 self.async_copy_ready_event.synchronize()
-                sampled_token_ids = self.sampled_token_ids_cpu.squeeze(-1).tolist()
+                sampled_token_ids = self.sampled_token_ids_cpu.tolist()
             # Replace placeholder token id with actual sampled id.
```
njhill:
```suggestion
            # Replace placeholder token id(s) with actual sampled id(s).
```

njhill (commented):
Thanks @izhuhaoran for this!

I'm curious though, did you actually hit it in real usage?

AFAIK it shouldn't be possible currently, since this path is only taken when there are requests in the batch which require the output ids during sampling - specifically when there are penalties or "bad_words" specified, or possibly for custom logits processors.

But we don't yet support any of these for the async scheduling + spec decoding combo. We reject any requests with the corresponding parameters when such config is active and so it should never actually reach here.

---

The reason we don't yet support them however is just because we hadn't yet looked into what additional changes are needed for the combination. It's quite possible that your change here is all that's needed which would be awesome! And we could then remove the validation that blocks those kinds of requests.

(your change here is certainly necessary, we just need to check whether it's sufficient... I'll try to test that soon)

izhuhaoran:
@njhill Thanks for your time and detailed explanation!  

> specifically when there are penalties or "bad_words" specified  

Yes—I hit this while testing async scheduling with `penalty` enabled, and find the `update_async_output_token_ids` bug.  

> Unfortunately from some cursory tests it seems it's not sufficient  

I initially thought this was the only blocker, but agree further debugging is likely needed to fully support penalties/bad_words.

vllm/v1/worker/gpu_input_batch.py
izhuhaoran: Thanks for suggestion, done.

vllm/v1/worker/gpu_input_batch.py
izhuhaoran: Done.

PR #30134, opened 2025-12-05
[OpenAI] Add parameter metadata to validation errors

R3hankhan123:
## Purpose
This PR adds VLLMValidationError exception class to provide detailed parameter information in OpenAI API error responses, improving API compatibility and debugging experience.

**Changes:**
- Add `VLLMValidationError` class with `parameter` and `value` attributes in `protocol.py`
- Convert 30+ validation errors across all OpenAI endpoints to use `VLLMValidationError`
- Update exception handlers to extract and propagate parameter metadata
- Add `param` field to error responses matching OpenAI API specification

## Test Plan

1. Build vLLM image and deploy a model
```
docker build -t local -f docker/Dockerfile . 
docker run --rm --gpus 1 -p 8000:8000 local --model=gpt2 --port=8000
```

2. Test validation errors with parameter metadata:
```
# Test 1: Invalid model
curl -X POST http://localhost:8000/v1/completions \
  -d '{"model": "fake-model", "prompt": "test"}'

# Test 2: Too many input tokens
curl -X POST http://localhost:8000/v1/completions \
  -d '{"model": "gpt2", "prompt": "'$(python -c 'print("test " * 1000)')'"}'

# Test 3: max_tokens too large
curl -X POST http://localhost:8000/v1/completions \
  -d '{"model": "gpt2", "prompt": "test", "max_tokens": 10000}'
```

## Test Result

All error responses now include proper `param` field

Test 1 - Invalid model:
```
{
  "error": {
    "message": "The model `fake-model` does not exist.",
    "type": "NotFoundError",
    "param": "model",
    "code": 404
  }
}
```

Test 2 - Too many input tokens:
```
{
  "error": {
    "message": "This model's maximum context length is 1019 tokens. However, your request has 2001 input tokens...",
    "type": "BadRequestError",
    "param": "input_tokens",
    "code": 400
  }
}
```

Test 3 - max_tokens too large:
```
{
  "error": {
    "message": "'max_tokens' (10000) cannot be greater than the model's maximum context length (1024).",
    "type": "BadRequestError",
    "param": "max_tokens",
    "code": 400
  }
}
```

---
<details>

</details>

### DIFF
--- a/vllm/entrypoints/openai/serving_chat.py
+++ b/vllm/entrypoints/openai/serving_chat.py
@@ -417,8 +417,7 @@ class OpenAIServingChat(OpenAIServing):
 
                 generators.append(generator)
         except ValueError as e:
-            # TODO: Use a vllm-specific Validation Error
-            return self.create_error_response(str(e))
+            return self.create_error_response(e)
 
         assert len(generators) == 1
         (result_generator,) = generators
@@ -448,8 +447,7 @@ class OpenAIServingChat(OpenAIServing):
         except GenerationError as e:
             return self._convert_generation_error_to_response(e)
         except ValueError as e:
-            # TODO: Use a vllm-specific Validation Error
-            return self.create_error_response(str(e))
+            return self.create_error_response(e)
 
     def get_chat_request_role(self, request: ChatCompletionRequest) -> str:
         if request.add_generation_prompt:
@@ -682,7 +680,7 @@ class OpenAIServingChat(OpenAIServing):
                 tool_parsers = [None] * num_choices
         except Exception as e:
             logger.exception("Error in tool parser creation.")
-            data = self.create_streaming_error_response(str(e))
+            data = self.create_streaming_error_response(e)
             yield f"data: {data}\n\n"
             yield "data: [DONE]\n\n"
             return
@@ -1328,9 +1326,8 @@ class OpenAIServingChat(OpenAIServing):
         except GenerationError as e:
             yield f"data: {self._convert_generation_error_to_streaming_response(e)}\n\n"
         except Exception as e:
-            # TODO: Use a vllm-specific Validation Error
             logger.exception("Error in chat completion stream generator.")
-            data = self.create_streaming_error_response(str(e))
+            data = self.create_streaming_error_response(e)
             yield f"data: {data}\n\n"
         # Send the final done message after all response.n are finished
         yield "data: [DONE]\n\n"
@@ -1354,8 +1351,7 @@ class OpenAIServingChat(OpenAIServing):
         except asyncio.CancelledError:
             return self.create_error_response("Client disconnected")
         except ValueError as e:
-            # TODO: Use a vllm-specific Validation Error
-            return self.create_error_response(str(e))
+            return self.create_error_response(e)
 
         assert final_res is not None
 
--- a/vllm/entrypoints/openai/serving_completion.py
+++ b/vllm/entrypoints/openai/serving_completion.py
@@ -23,6 +23,7 @@ from vllm.entrypoints.openai.protocol import (
     PromptTokenUsageInfo,
     RequestResponseMetadata,
     UsageInfo,
+    VLLMValidationError,
 )
 from vllm.entrypoints.openai.serving_engine import (
     GenerationError,
@@ -247,8 +248,7 @@ class OpenAIServingCompletion(OpenAIServing):
 
                 generators.append(generator)
         except ValueError as e:
-            # TODO: Use a vllm-specific Validation Error
-            return self.create_error_response(str(e))
+            return self.create_error_response(e)
 
         result_generator = merge_async_iterators(*generators)
 
@@ -308,8 +308,7 @@ class OpenAIServingCompletion(OpenAIServing):
         except GenerationError as e:
             return self._convert_generation_error_to_response(e)
         except ValueError as e:
-            # TODO: Use a vllm-specific Validation Error
-            return self.create_error_response(str(e))
+            return self.create_error_response(e)
 
         # When user requests streaming but we don't stream, we still need to
         # return a streaming response with a single event.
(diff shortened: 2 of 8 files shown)


vllm/entrypoints/openai/serving_completion.py
```diff
@@ -137,6 +138,9 @@ async def create_completion(
                 prompt_embeds=request.prompt_embeds,
                 config=self._build_render_config(request),
             )
+        except VLLMValidationError as e:
+            logger.exception("Error in preprocessing prompt inputs")
+            return self.create_error_response(str(e), param=e.parameter)
```
DarkLight1337: To avoid code duplication, we can allow `create_error_response` (and the streaming variant) to accept the raw exception and, inside that method, distinguish between `VLLMValidationError` and other exceptions

vllm/entrypoints/openai/serving_chat.py
```diff
@@ -347,8 +348,9 @@ async def create_chat_completion(
                     )
 
                 generators.append(generator)
+        except VLLMValidationError as e:
+            return self.create_error_response(e)
```
DarkLight1337, line 420: We can make `VLLMValidationError` a subclass of `ValueError` and always pass the exception itself to `create_error_response`

R3hankhan123: I don't think that possible since `ValueError` is a Built-in Exception in python

DarkLight1337: I mean that we can modify the `create_error_response` method to check `isinstance(e, VLLMValidationError)` and construct the error message from there, instead of having to add a separate `except` block everywhere `create_error_response` is used.

R3hankhan123: So instead like this
```
except VLLMValidationError as e:
    return self.create_error_response(e)
except ValueError as e: 
    return self.create_error_response(str(e))
 ```
 we can directly call
 ```
 except Exception as e:
    return self.create_error_response(e)
 ```

Comment (DarkLight1337) [reply vllm/entrypoints/openai/serving_chat.py:420]: Yes

Comment (DarkLight1337) [reply vllm/entrypoints/openai/serving_chat.py:420]: We can still catch `ValueError` since `VLLMValidationError` is a subclass of `ValueError`

Comment (R3hankhan123) [reply vllm/entrypoints/openai/serving_chat.py:420]: @DarkLight1337 I already have implemented it based on your initial review

**Review by DarkLight1337 — APPROVED** [review]
Thanks, LGTM now

File: vllm/entrypoints/openai/serving_chat.py [hunk 3]
```diff
@@ -1348,9 +1345,8 @@ async def chat_completion_full_generator(
                 final_res = res
         except asyncio.CancelledError:
             return self.create_error_response("Client disconnected")
-        except ValueError as e:
-            # TODO: Use a vllm-specific Validation Error
-            return self.create_error_response(str(e))
+        except Exception as e:
```
DarkLight1337: We should still only catch `ValueError` to avoid leaking internal server errors

R3hankhan123: so call `VLLMValidationError` instead of `Exception` ?

DarkLight1337: `ValueError` should be fine, since `VLLMValidationError` is a subclass of that

R3hankhan123: So for my clarification, `ValueError` should be in `protocol.py` only or in the rest of the files also?

DarkLight1337: The point is. we should avoid modifying the `except` condition

DarkLight1337: If the original condition already catches `Exception` then no need to change that. But we should not change `except ValueError` to `except Exception`

vllm/entrypoints/openai/serving_chat.py
```diff
@@ -680,9 +678,9 @@ async def chat_completion_stream_generator(
                 ] * num_choices
             else:
                 tool_parsers = [None] * num_choices
-        except Exception as e:
+        except ValueError as e:
```
DarkLight1337: This should be kept as exception

vllm/entrypoints/openai/serving_chat.py
```diff
@@ -1322,10 +1320,9 @@ async def chat_completion_stream_generator(
 
         except GenerationError as e:
             yield f"data: {self._convert_generation_error_to_streaming_response(e)}\n\n"
-        except Exception as e:
-            # TODO: Use a vllm-specific Validation Error
+        except ValueError as e:
```
DarkLight1337: Same

vllm/entrypoints/openai/serving_completion.py
```diff
@@ -307,9 +307,8 @@ async def create_completion(
             return self.create_error_response("Client disconnected")
         except GenerationError as e:
             return self._convert_generation_error_to_response(e)
-        except ValueError as e:
-            # TODO: Use a vllm-specific Validation Error
-            return self.create_error_response(str(e))
+        except VLLMValidationError as e:
```
DarkLight1337: Here it should be kept as `ValueError`

vllm/entrypoints/openai/serving_completion.py
```diff
@@ -509,10 +508,9 @@ async def completion_stream_generator(
 
         except GenerationError as e:
             yield f"data: {self._convert_generation_error_to_streaming_response(e)}\n\n"
-        except Exception as e:
-            # TODO: Use a vllm-specific Validation Error
+        except ValueError as e:
```
DarkLight1337: Here it should be kept as `Exception`

DarkLight1337: and so on...

vllm/entrypoints/openai/speech_to_text.py
```diff
@@ -410,9 +421,9 @@ async def _create_speech_to_text(
                 audio_data=audio_data,
             )
 
-        except ValueError as e:
+        except VLLMValidationError as e:
```
DarkLight1337: Revert here as well

vllm/entrypoints/openai/speech_to_text.py
```diff
@@ -447,9 +458,8 @@ async def _create_speech_to_text(
                 )
                 for i, prompt in enumerate(prompts)
             ]
-        except ValueError as e:
-            # TODO: Use a vllm-specific Validation Error
-            return self.create_error_response(str(e))
+        except VLLMValidationError as e:
```
DarkLight1337: Please check the rest of your code. You missed this and some other places

DarkLight1337 (approved):
LGTM now, thanks for your patience!

PR #30141, opened 2025-12-05
Add llmcompressor fp8 kv-cache quant (per-tensor and per-attn_head)

eldarkurtic:
**TLDR**: this PR adds support to load and run `llm-compressor` models with FP8 KV-cache and attention quantization. In addition to the standard "per-tensor" quantization, it adds support for "per-attention-head" quantization.

## Summary
1. enable using the existing pathway of "per-tensor" KV-cache (and query) FP8 quantization with scales calibrated through `llm-compressor`
2. Flash Attention v3 backend supports "finer-grained" scales, i.e. one scale per attention head. This is currently not supported, and this PR enables it in the following way:

2.1 for query quantization: 
- expand `QuantFP8` to support per-channel static quantization (queries are of the shape `num_tokens x hidden_size` so we expand the per-attention-head scales to accommodate per-channel scaling 
- this is enabled further by expanding the `static_scaled_fp8_quant` kernel to work with an array of scales

2.2 for kv-quantization:
- expand the `reshape_and_cache_flash` kernel by adding support for an array of `k/v_scale`
- it also covers both cache layouts (NHD and HND)
3. reorganized some things around KV-cache in `vllm/attention/layer.py` as so far it has been hardcoded for `calculate_kv_scales` pathway only
4. expands tests to cover all kernel-related updates 
5. update documentation on the newly supported KV-cache and attention quantization technique in vLLM

## Tests
To confirm that the existing pathway with `calculate_kv_scales=True/False` isn't affected, I ran GSM8k evals on the following models: `Llama-2-7b-chat-hf` (MHA), `Llama-3.1-8B-Instruct` (GQA), `Qwen/Qwen3-8B` (different model family), and got the same results before and after the PR. They are as follows:
- Model = `meta-llama/Llama-2-7b-chat-hf`
```
unquantized baseline
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.2509|±  |0.0119|
|     |       |strict-match    |     5|exact_match|↑  |0.1933|±  |0.0109|

calculate_kv_scales=False
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.2563|±  |0.0120|
|     |       |strict-match    |     5|exact_match|↑  |0.1911|±  |0.0108|

calculate_kv_scales=True
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.0053|±  | 0.002|
|     |       |strict-match    |     5|exact_match|↑  |0.0000|±  | 0.000|
``` 
**Note:** The score of `0` in `calculate_kv_scales=True` is already present on `main`; this PR does not introduce it.

- Model = `meta-llama/Llama-3.1-8B-Instruct`
```
unquantized baseline
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8522|±  |0.0098|
|     |       |strict-match    |     5|exact_match|↑  |0.8249|±  |0.0105|


calculate_kv_scales=False
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8302|±  |0.0103|
|     |       |strict-match    |     5|exact_match|↑  |0.8036|±  |0.0109|


calculate_kv_scales=True
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8324|±  |0.0103|
|     |       |strict-match    |     5|exact_match|↑  |0.8127|±  |0.0107|
```

- Model = `Qwen/Qwen3-8B`
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8810|±  |0.0089|
|     |       |strict-match    |     5|exact_match|↑  |0.8764|±  |0.0091|

calculate_kv_scales=True
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8787|±  |0.0090|
|     |       |strict-match    |     5|exact_match|↑  |0.8734|±  |0.0092|


calculate_kv_scales=False
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8741|±  |0.0091|
|     |       |strict-match    |     5|exact_match|↑  |0.8688|±  |0.0093|
```

And to confirm that the new support for `llm-compressor` models with both, `per-tensor` and `per-attention-head` scales is working correctly, I ran the same models from above with both configurations and observed the expected results:
- Model = `meta-llama/Llama-2-7b-chat-hf`
```
unquantized baseline
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.2509|±  |0.0119|
|     |       |strict-match    |     5|exact_match|↑  |0.1933|±  |0.0109|

per-tensor
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.2555|±  |0.0120|
|     |       |strict-match    |     5|exact_match|↑  |0.1835|±  |0.0107|

per-attn-head
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.2441|±  |0.0118|
|     |       |strict-match    |     5|exact_match|↑  |0.1736|±  |0.0104|
``` 

- Model = `meta-llama/Llama-3.1-8B-Instruct`
```
unquantized baseline
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8522|±  |0.0098|
|     |       |strict-match    |     5|exact_match|↑  |0.8249|±  |0.0105|


per-tensor
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8362|±  |0.0102|
|     |       |strict-match    |     5|exact_match|↑  |0.8021|±  |0.0110|

per-attn-head
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8393|±  |0.0101|
|     |       |strict-match    |     5|exact_match|↑  |0.8112|±  |0.0108|
```

- Model = `Qwen/Qwen3-8B`
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8810|±  |0.0089|
|     |       |strict-match    |     5|exact_match|↑  |0.8764|±  |0.0091|

per-tensor
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8810|±  |0.0089|
|     |       |strict-match    |     5|exact_match|↑  |0.8757|±  |0.0091|

per-attn-head
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8787|±  |0.0090|
|     |       |strict-match    |     5|exact_match|↑  |0.8757|±  |0.0091
```
The `llm-compressor` code to produce these models is available in diffs of `docs/features/quantization/quantized_kvcache.md`. I haven't done any tuning of the calibration parameters for the testing purposes, just ran the defaults so better results are expected with better tuning.
I've also verified that changes are working with both, LLM class and vllm serve. 

Note: some model implementations have remapping of scales guarded by something like this: `if "scale" in name` so I had to expand it to `if "scale" in name or "zero_point" in name:` to support loading of zero_points which are present in `llm-compressor` checkpoints.


vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
```diff
@@ -178,13 +178,24 @@ def _add_fused_moe_to_target_scheme_map(self):
...
     def from_config(cls, config: dict[str, Any]) -> "CompressedTensorsConfig":
+        # We keep only config groups which are not doing Attention quantization
+        # because Attention quantization on its own is not supported by vLLM.
+        # It is coupled with KV-cache quantization, and if scales are present in the
+        # checkpoint, they will be used properly.
+        config["config_groups"] = {
+            k: v
+            for k, v in config["config_groups"].items()
+            if "Attention" not in v["targets"][0]
+        }
```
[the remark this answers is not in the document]
eldarkurtic: This is a temporary and hacky way to drop config-groups where targets are `LlamaAttention`/`Qwen3Attention`/etc in `llm-compressor` checkpoints because this config doesn't have any meaning for vllm. My assumption here is that targets will never be an empty list.

vllm/v1/attention/backends/flash_attn.py
```diff
@@ -730,9 +738,9 @@ def forward(
             fa_version=self.vllm_flash_attn_version,
             prefix_scheduler_metadata=attn_metadata.prefix_scheduler_metadata,
             suffix_scheduler_metadata=attn_metadata.scheduler_metadata,
-            q_descale=layer._q_scale,
-            k_descale=layer._k_scale,
-            v_descale=layer._v_scale,
+            q_descale=q_descale,
+            k_descale=k_descale,
+            v_descale=v_descale,
```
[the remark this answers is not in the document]
eldarkurtic: resolved with the latest commit

vllm/model_executor/layers/quantization/input_quant_fp8.py
```diff
@@ -54,10 +54,15 @@ def __init__(
             assert not static, "Group quantization only supports dynamic mode"
             self.group_size = group_shape.col
         else:
-            assert group_shape in {GroupShape.PER_TOKEN, GroupShape.PER_TENSOR}
-            assert not static or group_shape == GroupShape.PER_TENSOR, (
-                "Only per-tensor scales supported for static quantization."
-            )
+            assert group_shape in {
```
ProExpertProg: Dynamic quant doesn't support per-channel

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
```diff
@@ -963,3 +979,101 @@ def validate_kv_cache_scheme(kv_cache_scheme: dict[str, Any] | None):
...
+
+        shape_kv: tuple[int, ...]
+        shape_q: tuple[int, ...]
+        if strategy == "tensor":
+            shape_kv = (1,)
+            shape_q = (1,)
+        else:  # strategy == "attn_head"
+            from vllm.v1.attention.backends.flash_attn import FlashAttentionImpl
+
+            assert (
```
ProExpertProg: Add a property that attention backend supports per-channel scales (default false), see CG support or query quant support

eldarkurtic: resolved now

vllm/attention/layer.py
```diff
@@ -70,21 +68,26 @@ def _init_kv_cache_quant(
...
     # kv-cache in fp8_e5m2. For kv-cache in fp8_e4m3, we
     # expect the pre-quantized k/v_scale to be loaded along
     # with the model weights.
-    layer.kv_cache_dtype = kv_cache_dtype
-    layer.calculate_kv_scales = calculate_kv_scales
     layer._k_scale = torch.tensor(1.0, dtype=torch.float32)
     layer._v_scale = torch.tensor(1.0, dtype=torch.float32)
     layer._q_scale = torch.tensor(1.0, dtype=torch.float32)
+    # defaults to layer._q_scale when per-attention-head KV-cache quant is not used
+    layer._q_scale_for_fa = layer._q_scale
```
LucasWilkinson: I think we should really try to avoid having backend specific naming here; is there reason we can't use `_q_scale` here? we may need to update https://github.com/vllm-project/vllm/blob/ae2e503dda693b6b7ab9052ec61e012a3c730f2f/vllm/attention/backends/abstract.py#L128-L134 to support quantization schemes; I think we can just store `layer._q_scale_group_shape` to indicate the format for `layer._q_scale`

I think we could extend `GroupShape` to support per-head by either making it:
1) 3D, i.e. for `(num_tokens, num_heads, hidden_dim)` we'd have 
     - per-tensor: (-1, -1, -1)
     - per-head: (-1, 1, -1)
2) 2D but make the hidden dim group shape head_size; i.e.
     - per-tensor: (-1, -1)
     - per-head: (-1, 128) "for head_size 128"

1 is probably cleaner; but 2 probably fits nicer into the existing format / tooling

eldarkurtic: the reason I added `q_scale_for_fa` is because I need it in two different shapes: 
1. query quantization goes through `QuantFP8` and there I needed `q_scale` to be expanded into per-channel scales since queries are of the shape `[num_tokens, head_size x num_heads]`. So `q_scale` is kept as an array of `[head_size x num_heads]` elements for channel-wise scaling. 
2. on the other hand, FlashAttention expects `q_scale` to have a very unusual shape: `[num_kv_heads]` (not `[num_heads]`) so that is the reason I introduced `q_scale_for_fa` which keeps this unusual shape only for the FA, and it defaults to `q_scale` for everything else.

eldarkurtic: btw @LucasWilkinson , is my understanding correct that we need to have scales in both target shapes already created before the compilation happens (i.e. we can't reshape scales on-the-fly)?
Because I assumed this is the case, I had two create and maintain two of them: `_q_scale` and `_q_scale_for_fa`

tests/kernels/quantization/test_fp8_quant.py
```diff
@@ -96,6 +97,61 @@ def test_dynamic_per_tensor_fp8_quant(
     opcheck_fp8_quant(ops_out, x)
 
 
+@pytest.mark.parametrize("num_tokens", NUM_TOKENS)
+@pytest.mark.parametrize("hidden_size", HIDDEN_SIZES)
+@pytest.mark.parametrize("dtype", DTYPES)
+@pytest.mark.parametrize("seed", SEEDS)
+@pytest.mark.parametrize("per_channel", [True, False])
+@torch.inference_mode()
+def test_static_fp8_quant(
```
MatthewBonanni: Could you add a test that covers the composition of per-channel and per-head quantization?

vllm/attention/layer.py
ProExpertProg: +1 for avoiding backend-specific naming, I would vote for the second option, quantization generally only works on 2D tensors anyway.

csrc/quantization/w8a8/fp8/common.cu
```diff
@@ -28,6 +28,24 @@ __global__ void scaled_fp8_quant_kernel_strided(
...
+    fp8_type* __restrict__ out, const scalar_t* __restrict__ input,
+    const float* __restrict__ scale, int hidden_size, int64_t in_row_stride,
+    int64_t out_row_stride) {
+  const int64_t token_idx = blockIdx.x;
+  const int tid = threadIdx.x;
+
+  const scalar_t* token_in = input + token_idx * in_row_stride;
+  fp8_type* token_out = out + token_idx * out_row_stride;
+
+  for (int i = tid; i < hidden_size; i += blockDim.x) {
```
MatthewBonanni: Could you use `vectorize_with_alignment` here like `scaled_fp8_quant_kernel_strided` above? Would require extending `vectorize_with_alignment` with a variant that can take a second array input, but should be beneficial.

csrc/quantization/w8a8/fp8/common.cu
```diff
@@ -400,4 +400,4 @@ void dynamic_per_token_scaled_fp8_quant(
                   hidden_size, in_row_stride, out_row_stride);
             });
       });
-}
+}
```
MatthewBonanni: Nit: clean up this change

vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
```diff
@@ -980,3 +996,133 @@ def validate_kv_cache_scheme(kv_cache_scheme: dict[str, Any] | None):
...
+        if strategy == "attn_head":
+
+            def _tp_aware_loader(
+                param: torch.Tensor,
+                loaded_weight: torch.Tensor,
+                kind: Literal["q", "k", "v"],
+                param_type: Literal["scale", "zero_point"],
+            ):
+                # Zero-points are not used as vLLM only supports symmetric quantization
+                if param_type == "zero_point":
```
MatthewBonanni, line 1058: Is it worth adding a warning that we're ignoring the zero point?

docs/features/quantization/quantized_kvcache.md
```diff
@@ -1,162 +1,187 @@
...
+   Scales are estimated using a curated calibration dataset for maximum accuracy.  
+   This requires the [llm-compressor](https://github.com/vllm-project/llm-compressor) library.  
+   _See example below!_
 
-### Performance Impact
+#### Additional `kv_cache_dtype` Options
 
-The current FP8 KV cache implementation primarily benefits throughput by allowing approximately double the amount of space for KV cache allocation. This enables either:
+- `kv_cache_dtype="auto"`: Use the model's default data type
+- `kv_cache_dtype="fp8_e4m3"`: Supported on CUDA 11.8+ and ROCm (AMD GPUs)
```
MatthewBonanni, line 49: The support matrix for kv cache dtypes is a bit more complicated than this. I'd recommend leaving out any comment about support for now, and when #32477 lands we can point to that. This section is also a bit misleading and makes it seem like `fp8` is an option which is distinct from `fp8_e4m3` and `fp8_e5m2` (when it actually acts as effectively an alias for `fp8_e4m3`). We should make that behavior clear. Also, can we use `calculate_kv_scales` with `fp8_e5m2`?

vllm/v1/attention/backend.py
```diff
@@ -610,6 +610,7 @@ class AttentionImpl(ABC, Generic[T]):
     # TODO add support to more backends:
     # https://github.com/vllm-project/vllm/issues/25584
     supports_quant_query_input: bool = False
+    supports_per_head_quant_scales: bool = False
```
MatthewBonanni, line 623: Can we add this to `validate_configuration` so that the automatic backend selector can filter on it?

csrc/quantization/w8a8/fp8/common.cu
LucasWilkinson: remove unrelated change please

tests/quantization/test_compressed_tensors.py
```diff
@@ -32,6 +33,35 @@
...
+from vllm.v1.attention.backends.fa_utils import get_flash_attn_version
+
+
+def _kv_cache_fp8_per_attn_head_skip_reason() -> str | None:
+    """Per attn_head FP8 KV cache is only supported with FlashAttention v3+"""
+    if not current_platform.is_cuda():
+        return "This test is skipped on non-CUDA platform."
+
+    if (
+        envs.VLLM_ATTENTION_BACKEND is not None
```
MatthewBonanni: This will be deprecated very soon (#32812), can you get the backend from the `AttentionConfig` instead?

MatthewBonanni (commented):
Thanks for updating the PR! Just a few more small comments

vllm/v1/attention/backend.py
LucasWilkinson: +1 i think we can do this in follow-up PR though in the interest of getting this one landed (and avoiding rebase thrash)

tests/kernels/attention/test_cache.py
```diff
@@ -230,15 +240,50 @@ def permute_and_compact(x):
...
+                for h in range(num_heads):
+                    input_slice = input[:, h : h + 1, :, :].contiguous()
+                    output_slice = torch.empty_like(input_slice, dtype=output.dtype)
+                    ops.convert_fp8(
+                        output_slice,
+                        input_slice,
+                        scale[h].item(),
+                        kv_dtype,
+                    )
+                    output[:, h : h + 1, :, :].copy_(output_slice)
```
LucasWilkinson: maybe replace with
```
    def convert_fp8_local(output, input, scale, kv_dtype):
        fp8_input = input.view(torch.float8_e4m3fn)
        if scale.numel() == 1:  # per-tensor
            result = scaled_dequantize(
                fp8_input.flatten(0, 2), scale, group_shape=None,
                out_dtype=output.dtype
            ).reshape(*input.shape)
        else:  # per-head: broadcast scale along the head dimension
            # Original code uses dim 2 for NHD, dim 1 for HND
            if kv_cache_layout == "NHD":
                result = fp8_input.to(output.dtype) * scale.view(1, 1, -1, 1)
            else:
                result = fp8_input.to(output.dtype) * scale.view(1, -1, 1, 1)
        output.copy_(result)
```

LucasWilkinson (approved):
overall LGTM once the comments are addressed; thanks for all the hard work! and apologies for the long delay

tests/kernels/attention/test_cache.py
mawong-amd: It should be `current_platform.fp8_dtype()` in place of `torch.float8_e4m3fn` instead, but nice and compact implementation. Tiny PR to fix: https://github.com/vllm-project/vllm/pull/32904.

PR #30143, opened 2025-12-05
[Misc] Remove pad_for_cudagraphs from config

LucasWilkinson:
With https://github.com/vllm-project/vllm/pull/28579 we can strip `pad_for_cudagraph` from the config forcing cudagraph padding through the dispatch consistently making it easier to debug and enable support for separate sizes for FULL and PIECEWISE graphs


ProExpertProg (approved):
Nice cleanup!

PR #30009, opened 2025-12-04
[Bugfix]: Fix `TokenizerLike` interface

Rohan138:
## Purpose

Corrected+rebased version of #22121 to fix #22013. Note that this function is only called from `RandomDataset.sample`, which calls `tokenizer.encode(prompt, add_special_tokens=False)` (https://github.com/vllm-project/vllm/blob/main/vllm/benchmarks/datasets.py#L407). Hence, the tokenizer only adds the bos token; not eos,\<INST\>,\</INST\>.

Changes:
- Add `num_special_tokens_to_add` to MistralTokenizer
- update type hint and functions from PretrainedTokenizerBase -> TokenizerLike across all files in vllm/benchmark/*.py
- Fix RandomDatasetForReranking.sample() and specify allowed_tokens

## Test Plan
Currently, with `vllm/vllm-openai:nightly`:

```
vllm bench serve --model mistralai/Mixtral-8x22B-Instruct-v0.1 --percentile-metrics tpot,itl,e2el --dataset-name random --ignore-eos --max-concurrency 1 --num-prompts 10 --random-input-len 128 --random-output-len 128 --trust-remote-code --save-result --result-filename pyt_vllm_mixtral-8x22b_serving.json
```

## Test Result
```
[...]

PR #30150, opened 2025-12-05
refactoring of the benchmark dataset capabilities and added a random …

jdebache:
## Purpose

- Added a new dataset option which: 
  - Downloads/reads a txt file
  - Tokenizes it
  - Generates samples by randomly slicing the resulting token sequence

The goal of this dataset is to provide a way to generate a benchmarking dataset with sequences that are more realistic than the standard random option. This is useful when benchmarking expert parallel and speculative decoding based deployments.

I took the opportunity to refactor the datasets as this single file was growing huge and becoming hard to maintain. I also fixed type hinting issues

TODO: add some more unit tests.

Note that I would like to merge this after https://github.com/vllm-project/vllm/pull/30009.


hypdeb:
This is too large.

PR #30151, opened 2025-12-05
[CI/Build][AMD][Quantization] Fix test_int8_kernel.py by updating int8_utils to use hip.libdevice.round

rasmith:
kernels/quantization/test_int8_kernel.py test fails due to a Triton compile error.

Previously,` tl.extra.hip.libdevice.round `was not available, so I added a `core.extern` function to make this work. Now, this `core.extern` function fails to compile with Triton, however,` tl.extra.hip.libdevice.round` is now available!  I updated the `round_int8 `function to use the `tl.extra.hip.libdevice.round` and everything works again.


yewentao256 (approved):
LGTM, thanks for the work!

PR #30152, opened 2025-12-05
let draft model follow target model's config_format

bangshengtang:
Summary: internally we use a custom plugin to handle our customized config format, without this change, draft model config will be default to "auto", this has caused mismatches

Test Plan: CI

Differential Revision: D88501036


zhuohan123 (approved):

yewentao256 (approved):
LGTM, thanks for the work!

PR #30156, opened 2025-12-05
feat: add TxtSlicesDataset to allow sampling slices from txt file for benchmarking

jdebache:
## Purpose

Sampling randomly directly from a tokenizer for benchmarking creates data that is not ideal to benchmark when using speculative decoding or expert parallelism.

On the other hand, random datasets are very flexible and offer complete control on the input and output sequence lengths, which is desirable to create reproducible benchmarks.

This PR introduces a new type of benchmarking dataset called `TxtSlicesDataset` which offers a compromise between the flexibility of a random dataset and the fidelity of a real dataset. It allows sampling slices from a user-provided txt file.

## Content
- The implementation of `TxtSlicesDataset`
- Fixes to typing in `datasets.py`
- A unit test for the new dataset type


jdebache:
Thanks for the review @DarkLight1337.

The main benefit is that you don't need to create the dataset in `jsonl` separately for each combination of ISL/OSL and number of prompts. While custom dataset is for the use case where you care about the content of your prompts specifically.

jdebache:
While I have you here, could I also bring your attention to PR #39281? It removes some deep dependencies for the datasets module, which have a tendency to break documentation generation.

DarkLight1337:
> The main benefit is that you don't need to create the dataset in `jsonl` separately for each combination of ISL/OSL and number of prompts. While custom dataset is for the use case where you care about the content of your prompts specifically.

I think it would be simpler to have a helper script to generate the JSONL dataset from these settings, instead of having to modify the benchmark code in vLLM.

jdebache:
I think that wouldn't be a problem.

However, I would advise against it. I find that it's not great for discoverability. It adds a new concept of "a script that generates a dataset that I can than pass to `vllm bench serve`, instead of reusing the existing entrypoint, which already takes care of obtaining the data in all other cases. What do you think?

DarkLight1337:
I feel that the current changes are too intrusive for a single dataset type that's relatively niche. I could accept this PR if you make the `TxtSlicesDataset` a wrapper over the JSONL dataset without having to change `sample`.

jdebache:
Okay, I will write a separate script to generate JSONL. Do you mind if I still do the two following changes:
- Support separate range ratios for ISL and OSL, which is useful for all random datasets
- Pull out `get_sampling_params`, so that I can reuse it in the generation script for consistency between the new script and `vllm bench serve`?

DarkLight1337:
Yes, that's fine

jdebache:
Hey @DarkLight1337: I have adjusted the PR as discussed.

tests/benchmarks/test_sampling_params.py
```diff
@@ -0,0 +1,250 @@
...
+        )
+        # real_input_len = 100 - 4 = 96, range_ratio 0 → all 96
+        assert np.all(input_lens == 96)
+
+    def test_special_tokens_not_subtracted_from_output(self):
+        rng = np.random.default_rng(7)
+        _, output_lens, _ = get_sampling_params(
+            rng, 10, 0.0, 0.0, 100, 50, self._tok(num_special=4)
+        )
+        assert np.all(output_lens == 50)
```
DarkLight1337: Merge these two tests together

tests/benchmarks/test_txt_slices_dataset.py
```diff
@@ -0,0 +1,77 @@
...
+incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud
+exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
+Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat
+nulla pariatur. Excepteur sint occaecat cupidatat non proident,
+sunt in culpa qui officia deserunt mollit anim id est laborum.
+"""
+
+
+@pytest.mark.benchmark
+def test_create_txt_slices_jsonl(hf_tokenizer: PreTrainedTokenizerBase) -> None:
```
DarkLight1337: Use `tmp_path` fixture so we don't need try-finally to ensure cleanup

vllm/benchmarks/datasets/utils.py
```diff
@@ -0,0 +1,73 @@
+# SPDX-License-Identifier: Apache-2.0
```
DarkLight1337, line 1: Let's create a new directory `vllm/benchmarks/datasets` and put all the related files inside.

vllm/benchmarks/datasets.py
```diff
@@ -60,20 +60,16 @@
 
 DEFAULT_NUM_PROMPTS = 1000
 
-# -----------------------------------------------------------------------------
-# Data Classes
-# -----------------------------------------------------------------------------
-
 
-@dataclass
+@dataclass(frozen=True)
```
DarkLight1337: Is there a need to change this class?

vllm/benchmarks/datasets.py
```diff
@@ -1651,9 +1661,27 @@ def add_random_dataset_base_args(
...
         default=0.0,
         help="Range ratio for sampling input/output length, "
-        "used only for random sampling. Must be in the range [0, 1) to define "
-        "a symmetric sampling range"
-        "[length * (1 - range_ratio), length * (1 + range_ratio)].",
+        "used only for random sampling. Sets both input and output range "
+        "ratios unless overridden by --random-input-range-ratio or "
+        "--random-output-range-ratio. Must be in [0, 1).",
+    )
+    parser_or_group.add_argument(
```
DarkLight1337: I think it would be simpler to make `--random-range-ratio` accept a JSON dictionary (which is parsed in Python) so we don't need to add two extra arguments

vllm/benchmarks/datasets/datasets.py
```diff
@@ -2340,9 +2374,32 @@ def load_data(self) -> None:
         if not getattr(self, "disable_shuffle", False):
             random.shuffle(self.data)
 
-    def sample(self, **kwargs) -> list:
+    def sample(
```
DarkLight1337, line 2344: Unnecessary change

vllm/benchmarks/datasets.py
jdebache: It is never mutated and there is no a-priori reason it should be mutable, which is why I chose to make it `frozen`. The advantages of choosing immutable types by default are, among others:
- predictability and easier reasoning about the code
- removing possibility of hidden aliasing
- intent signalling

vllm/benchmarks/datasets.py
jdebache: Just to clarify: it should accept both a single value and a dictionary correct? Otherwise, it would break backwards compatibility.

vllm/benchmarks/datasets/datasets.py
jdebache: It is true that it is not strictly necessary. However, this will give more information to the type checker for this function, which will allow to:
- catch more potential bugs in future changes
- shifting verification of correctness left and reducing effort in code-review, as typing issues can be caught by pre-commit hook

jdebache:
Thanks again for the review. I have responded to some of the comments and would appreciate your feedback on those. In the meantime, I will address the other comments and update the changes.

vllm/benchmarks/datasets.py
DarkLight1337: Yes

vllm/benchmarks/datasets.py
DarkLight1337: I think that's outside the scope of this PR

vllm/benchmarks/datasets/datasets.py
DarkLight1337: Again, outside the scope of this PR

tests/benchmarks/test_sampling_params.py
jdebache: Done.

vllm/benchmarks/datasets/utils.py
jdebache: Done.

vllm/benchmarks/datasets.py
jdebache: Reverted.

vllm/benchmarks/datasets/datasets.py
jdebache: Reverted.

vllm/benchmarks/datasets.py
jdebache: Done.

tests/benchmarks/test_txt_slices_dataset.py
jdebache: Done.

jdebache:
Okay, I have now addressed all review comments.

vllm/benchmarks/datasets/utils.py
```diff
@@ -0,0 +1,101 @@
...
```
DarkLight1337, line 1: Call this `utils.py`

vllm/benchmarks/throughput.py
```diff
@@ -357,6 +357,12 @@ def get_requests(args, tokenizer):
         and args.dataset_name not in {"prefix_repetition", "random-mm", "random-rerank"}
     ):
         sample_kwargs["range_ratio"] = args.random_range_ratio
+        sample_kwargs["input_range_ratio"] = getattr(
```
DarkLight1337: Can revert these changes

jdebache: Oh, forgot this file, sorry. Will do.

vllm/benchmarks/datasets/datasets.py
```diff
@@ -110,7 +110,7 @@ def __init__(
         # default seed.
         self.random_seed = random_seed if random_seed is not None else self.DEFAULT_SEED
         self.disable_shuffle = disable_shuffle
-        self.data = None
+        self.data: Any
```
DarkLight1337: We should still default to `None`

DarkLight1337 (approved):
LGTM now, thanks for your patience!

jdebache:
Thanks for your thorough review!

PR #30164, opened 2025-12-06
Nvidia ModelOpt workaround for issue 28072

### LINKED ISSUE #28072: [Bug]: WeightsMapper used in QuantizationConfig does not handle regex/wildcard paths

shengliangxu:
## Purpose

This is a workaround for issue https://github.com/vllm-project/vllm/issues/28072

FIX #28072

**The Core Problem**
vLLM currently fails to correctly apply `WeightsMapper` remapping logic to quantization configurations that utilize regex patterns or wildcards (e.g., in `ignore` or `exclude_modules` lists).

* **Intended Behavior:** When vLLM loads a model, it often remaps HuggingFace model structures to match vLLM's internal structure (e.g., remapping `vision_tower.` -> `vision.`).
* **Actual Behavior:** The `WeightsMapper` assumes it is working with explicit, single weight names. It cannot parse or transform regex strings (e.g., `re:vision_tower.*`) or wildcards used in quantization configs. Consequently, if a module name in the config needs remapping but is wrapped in a regex pattern, the remapping fails, and the module might be incorrectly quantized or processed.

Before this issue gets officially resolved, this workaround can make ModelOpt quantized models loading work without issues.

## Test Plan

Tested using nvidia/Qwen2.5-VL-7B-Instruct-NVFP4

## Test Result

vllm serve succeed


yewentao256 (commented):
Thanks for the work!
Could you add more context in the PR description?
And could you also add a `lm_eval` for the model to make sure it has the correct accuracy?

shengliangxu:
@yewentao256 Here's one result following[ this doc ](https://docs.vllm.ai/projects/recipes/en/latest/Llama/Llama4-Scout.html#verify-accuracy):

Without the change, accuracies are:


|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8741|±  |0.0091|
|     |       |strict-match    |     5|exact_match|↑  |0.8567|±  |0.0097|


With this change, it's:

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8757|±  |0.0091|
|     |       |strict-match    |     5|exact_match|↑  |0.8575|±  |0.0096|


Both run are on B200.

So accuracy is not impacted.

pavanimajety (approved):
Thanks for the PR, it looks good to me

mgoin (approved):

PR #30173, opened 2025-12-06
[BugFix] Fix `assert  batch_descriptor.num_tokens == num_tokens_padded`

LucasWilkinson:
There's a bug with `FULL_DECODE_ONLY` (`FULL_AND_PIECEWISE` works fine) with DP. There is an edge case where one rank runs eager but all other ranks want to run with cudagraphs, so now we synchronize the cudagraph mode each rank wants to run as across all ranks. Since currently PIECEWISE can be treated as eager (valid in all the same situations) it is sufficient to just disable full cudagraphs if all ranks want to; we make want to pass an explicit list of valid modes in the  future.

FIXES: https://github.com/vllm-project/vllm/pull/28579#issuecomment-3619429461


SageMoore (approved):
Thanks for the fix, @LucasWilkinson!

tlrmchlsmth (approved):
makes sense, LGTM

Issue #28407, opened 2025-11-10
[Feature]: Improve DCP error messages


LucasWilkinson:
### 🚀 The feature, motivation and pitch

Currently if a backend doesn't support DCP we get the following error message 

```
AssertionError: DCP requires attention impls to return the softmax lse for decode, but the impl FlashInferImpl does not return the softmax lse for decode.
```

It would good to suggest to the user to try an alternative backend using `VLLM_ATTENTION_BACKEND`

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

WorldExplored:
Working on it.

DraconicLegend:
Hi! First-time contributor here, so please correct me if I'm wrong. Is this issue still open, stale, or just a duplicate of #43053?

linshi-w:
I'd like to work on this.

wak327:
Hi Like to work on this can you assign this Issue to me?

luyixiao12:
Hi is this still open? I would like to work on this.

LucasWilkinson:
Unfortunately this ends up being pretty complicated and messy; the error messaging has been improved a but closing this as it is now outdated that `VLLM_ATTENTION_BACKEND` has been deprecated

PR #30174, opened 2025-12-06
[Bugfix] Improve DCP error message with backend hint

GeoffreyWang1117:
## Summary
- Add a helpful hint to the DCP error message suggesting users try setting `VLLM_ATTENTION_BACKEND` to a compatible backend
- Before: `AssertionError: DCP requires attention impls to return the softmax lse for decode, but the impl FlashInferImpl does not return the softmax lse for decode.`
- After: Same message + `Try setting VLLM_ATTENTION_BACKEND to a compatible backend such as FLASH_ATTN or FLASHINFER.`

## Test plan
- [x] Verified error message contains the new hint text
- [x] Verified pre-commit checks pass

Fixes #28407

Signed-off-by: GeoffreyWang1117 <173976389+GeoffreyWang1117@users.noreply.github.com>

### DIFF
--- a/vllm/v1/worker/gpu_model_runner.py
+++ b/vllm/v1/worker/gpu_model_runner.py
@@ -5241,7 +5241,9 @@ class GPUModelRunner(
                     "DCP requires attention impls to return"
                     " the softmax lse for decode, but the impl "
                     f"{layer_impl.__class__.__name__} "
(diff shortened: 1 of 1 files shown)


ApostaC (approved):
LGTM!

PR #24252, opened 2025-09-04
[Compile] Conditional compilation. Introduce compile_ranges

ilmarkov:
Second part of splitting https://github.com/vllm-project/vllm/pull/22086

Dynamic Graph dispatch via compile_ranges: Introduces a new configuration option, compile_ranges, as an alternative to compile_sizes. This enables dynamic dispatch to different compiled graphs based on the input batch size.
Now with this approach, when allreduce fusion is enabled, vllm adds additional compile range split point in order to separate the graphs: 1. One with fused allreduce for small-middle shape inputs. 2 One with nccl based allreduce for large shape inputs

The existing compile_sizes feature is extended and generalized with compile_ranges. Defined by split points, these ranges allow vllm to dynamically dispatch requests to specific, pre-compiled graphs based on input batch size. For example, a configuration of (32, 64) defines three distinct ranges: [1, 32], [33, 64], and [65, max_num_batched_tokens). This provides granular control, allowing developers to statically enable or disable fusions within each graph to optimize performance for different batch sizes.

All the compilation now is going through `piecewise_backend.py`. All compilations will now be done in the bounds on certain compile range, dynamic shape compilation is removed.

## Purpose
Corresponding RFC: https://github.com/vllm-project/vllm/issues/23113
The primary motivation for these changes is to enhance vllm's performance and adaptability for diverse workloads. By supporting allreduce fusion without custom ops and introducing dynamic graph dispatch, we empower users to fine-tune vllm for more efficient and scalable inference.

## Test Plan
Added test `test_compile_ranges.py`

## Follow ups

- [ ] Deal with sharing shape env for all graphs which could lead to the situation when one compilation constraints SymInts for the other compilations. Might need support from torch.compile, e.g. `shapenv.assume_ranges, shapenv.do_error_at_specialize`.
- [ ] Put fusions under O3 level of compilations.
- [ ] Sharing an range info with the inductor for the SimInt. [comment](https://github.com/vllm-project/vllm/pull/24252#discussion_r2514839468)

## Performance benchmarks:

**Server**:
```
 VLLM_ALLREDUCE_USE_SYMM_MEM=1  vllm serve {{model}} 
        --disable-log-requests --no-enable-prefix-caching -tp {{tp}} -dp 1 --max-num-seqs 256
```
To enable allreduce fusions:
[...]

PR #30178, opened 2025-12-06
[Bugfix] fix fuse_allreduce_rms when tp =1

ZJY0516:
## Purpose
Fix https://github.com/vllm-project/vllm/pull/24252#issuecomment-3619468884

cc @ProExpertProg @hjjq 

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/compilation/collective_fusion.py
+++ b/vllm/compilation/collective_fusion.py
@@ -1188,6 +1188,18 @@ class AllReduceFusionPass(VllmPatternMatcherPass):
         self.disabled = False
 
     def is_applicable_for_range(self, compile_range: Range) -> bool:
+        if self.disabled:
+            logger.warning_once(
+                "AllReduce fusion pass is disabled for tensor parallel size %s; "
+                "skipping",
+                self.tp_size,
+            )
+            return False
+        if not hasattr(self, "max_token_num"):
+            logger.warning_once(
+                "AllReduce fusion pass missing max token bound; skipping",
+            )
+            return False
         return compile_range.end <= self.max_token_num
 
     @VllmInductorPass.time_and_log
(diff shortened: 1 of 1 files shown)


vllm/compilation/collective_fusion.py
```diff
@@ -1188,6 +1188,13 @@ def register_patterns(self):
         self.disabled = False
 
     def is_applicable_for_range(self, compile_range: Range) -> bool:
+        if self.disabled:
```
ProExpertProg, line 1195: The pass might also be disabled for another reason (failure in init), could you either log the reason there or save the reason as a string and log it here? Just don't want the message to confuse people (disabled for TP=2 might imply it's not supported for TP=2)

ProExpertProg (approved):

vllm/compilation/collective_fusion.py
ZJY0516: > The pass might also be disabled for another reason (failure in init)

You mean it will fail in `register_patterns`? Could we just log `AllReduce fusion pass is disabled` without specific reasons?

ProExpertProg: yes that's fine as well; maybe add a warning in the constructor to the cases that don't currently warn?

ZJY0516: Add a warning at the end of `AllReduceFusionPass.__init__` if `self.disabled == True`?

ProExpertProg: No, just add warnings to lines 1079 and 1084 (there are existing warnings on 1090 and 1100)

ZJY0516: make sense

PR #30184, opened 2025-12-06
[Feature] Add offline FastAPI documentation support for air-gapped environments

### LINKED ISSUE #29847: [Feature]: enable FastAPI documention (`/docs`) for air-gaped host

When used on a air-gaped system, FastAPI documentation route (`/docs`) is not accessible because it requires pulling static content from internet.
This route is very useful because it is both a documentation and a simple way to send requests to the server, easing a lot when integrating vllm to a system.

One simple way to enable this route is to use the [fastapi_offline](https://github.com/turettn/fastapi_offline) package  and edit the way we call the app object :
```
from fastapi import FastAPI
app = FastAPI()
```
by 
```
from fastapi_offline import FastAPIOffline
app = FastAPIOffline()
```

### Alternatives

_No response_

### Additional context

_No response_

ricky-chaoju:
## Summary

This PR adds support for offline FastAPI documentation (`/docs`) in air-gapped environments by integrating the `fastapi-offline` package.

When used on an air-gapped system, FastAPI documentation route (`/docs`) is not accessible because it requires pulling static content from the internet. This route is very useful because it is both documentation and a simple way to send requests to the server, easing integration of vLLM into systems.

This PR adds the `--enable-offline-docs` CLI flag which uses `fastapi-offline` to bundle static assets locally.

Fixes #29847

## Changes

- Add `--enable-offline-docs` CLI flag to `FrontendArgs`
- Use `fastapi-offline` for air-gapped environments when enabled
- Add `fastapi-offline` as optional dependency in `pyproject.toml`
- Fallback to standard FastAPI if package not installed with warning message


vllm/entrypoints/openai/api_server.py
```diff
@@ -26,6 +26,14 @@
 from fastapi.exceptions import RequestValidationError
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse, Response, StreamingResponse
+
+try:
+    from fastapi_offline import FastAPIOffline
```
chaunceyjiang: It’s just a question: does FastAPI have an official solution for this?

ricky-chaoju: Hi @chaunceyjiang! 

Thanks for reviewing PR #30184. Regarding your question about FastAPI's official solution:

FastAPI does have an official self-hosting method documented here: https://fastapi.tiangolo.com/how-to/custom-docs-ui-assets/

It requires manually:
1. Downloading static files (swagger-ui-bundle.js, swagger-ui.css, redoc.standalone.js)
2. Setting up StaticFiles mounting
3. Creating custom route handlers

The `fastapi-offline` package essentially automates this official process.

I can implement either approach:
- **Option A**: Keep `fastapi-offline` (simpler, 1-line change, auto-updates)
- **Option B**: Manual implementation following FastAPI's official docs (no dependencies, ~2MB static files in repo)

Which would you prefer for vLLM? Happy to adjust based on your preference!

chaunceyjiang: /cc @hmellor PTAL.

chaunceyjiang (commented):
Just a question: will the vLLM Docker image include these JS/CSS files?

Can you help confirm this further?

vllm/entrypoints/openai/api_server.py
```diff
@@ -957,11 +962,46 @@ def _log_non_streaming_response(response_body: list) -> None:
         logger.info("response_body={<binary_data>}")
 
 
+def _setup_offline_docs(app: FastAPI) -> None:
```
chaunceyjiang: I would rather recommend placing this endpoint in `vllm/entrypoints/serve/instrumentator` and also putting the `static` files in this directory.

vllm/entrypoints/openai/api_server.py
```diff
@@ -962,6 +962,13 @@ def build_app(args: Namespace) -> FastAPI:
         app = FastAPI(
             openapi_url=None, docs_url=None, redoc_url=None, lifespan=lifespan
         )
+    elif args.enable_offline_docs:
```
chaunceyjiang, line 873: Can you move it to `register_vllm_serve_api_routers()`?

chaunceyjiang (commented):
LGTM.

@hmellor PTAL.

hmellor (approved):
I just updated the `js` to the latest from FastAPI's CDN so that I know with certainty that that's where it came from.

This PR LGTM, thanks for improving the UX for air-gapped users!

PR #30190, opened 2025-12-06
[grpc] Support gRPC server entrypoint

CatherineSue:
## Purpose

Add gRPC server support to vLLM, enabling the community to integrate vLLM via gRPC protocol for any upstream application or routing layer. 

**Key Benefits:**

1. **Native gRPC Protocol Support**
   - Enables upstream applications to connect via gRPC/Protobuf instead of HTTP/JSON
   - Binary protocol reduces serialization overhead
   - HTTP/2 multiplexing improves connection efficiency
   - Expands vLLM's integration options beyond HTTP/REST APIs

2. **Integration with `sgl-model-gateway`**
   - Enables vLLM workers to operate as gRPC backends
   - Bypasses Python GIL bottleneck by moving tokenization logic to Rust
   - Provides production-grade features: advanced routing, secured mcp and database management, responses api
   - Measured performance gains at high concurrency (see Test Results)  

### Changed Files

**Protocol & Codegen:**

- `vllm_scheduler.proto` - Protocol buffer definition (source)
- `vllm_scheduler_pb2.py` - Generated protobuf messages (auto-generated)
- `vllm_scheduler_pb2_grpc.py` - Generated gRPC service (auto-generated)
- `compile_protos.py` - Script to compile proto files
- `__init__.py` - Module initialization

**Server Implementation:**

- `vllm/grpc/grpc_request_manager.py` - Request manager (GrpcRequestManager class)
- `vllm/entrypoints/grpc_server.py` - Server entrypoint (VllmSchedulerServicer + main)

#### Compilation

To regenerate the Python code from the `.proto` file:

```bash
python vllm/grpc/compile_protos.py
```

**Requirements**: `pip install grpcio-tools`

This generates:

- `vllm_scheduler_pb2.py` - Message classes
- `vllm_scheduler_pb2_grpc.py` - Service stubs and servicers

## Test Plan

Run the gRPC server with a `Llama-3.1-8B-Instruct`:

```bash
  python3 -m vllm.entrypoints.grpc_server \
    --model meta-llama/Llama-3.1-8B-Instruct \
    --port 8080 \
    --host 0.0.0.0 \
    --tensor-parallel-size 1
```

Test with gateway integration with `sgl-model-gateway`

Verify:
  1. Health check endpoint responds correctly
  3. Streaming generation returns token IDs (not text)
  4. gRPC reflection is available for introspection
  5. Request abort/cancellation works properly
  6. `GetModelInfo` and `GetServerInfo` return correct metadata

## Test Result

We use `genai-bench` to measure the http_server vs (grpc_server + `sgl-model-gateway`) with `Llama-3.3-70B-Instruct` on 4xH100.

**Performance Results (Llama-3.3-70B, D100_100, Concurrency 256):**

At high concurrency, gRPC demonstrates superior production characteristics:

| Metric | gRPC | HTTP | Improvement |
|--------|------|------|-------------|
| **Throughput** | 9,068 tok/s | 6,629 tok/s | **+37%** |
| **Requests/sec** | 45.7 | 33.4 | **+37%** |
| **p99 TTFT** | 1,792ms | 2,434ms | **-26%** |
| **p90 TTFT** | 1,728ms | 2,188ms | **-21%** |
| **TTFT Variance (stddev)** | 428ms | 651ms | **-34%** |

**Key Value Proposition:**

- Processes **39% more requests** in same time with **26% better tail latency**
- **34% more consistent** performance (lower variance)

D100_100_group_by_server_version_combined_plots_1x4

D100_1000_group_by_server_version_combined_plots_1x4

---
<details>

</details>


njhill:
Thanks @CatherineSue! I think this is definitely something worthwhile to add.

It would be interesting to see how the performance compares in the same tests with e.g. `--api-server-scaleout=4`.

Also just for reference there here's a gRPC wrapper we had used in the past in IBM/Red Hat, probably quite similar but might be useful to compare https://github.com/opendatahub-io/vllm-tgis-adapter.

.github/CODEOWNERS
```diff
@@ -158,3 +158,7 @@ mkdocs.yaml @hmellor
 /docs/usage/security.md @russellb
 /SECURITY.md @russellb
 /docs/contributing/vulnerability_management.md @russellb
+
+# gRPC server
+/vllm/grpc @CatherineSue @slin1237
+/vllm/entrypoints/grpc_server.py @CatherineSue @slin1237
```
simon-mo: Unfortunately, github only allows members with write permission to be added to codeowners. While the write permission is granted via a separate nomination process based on stewardship, my recommendation is to add this to mergify

> For features that you intend to maintain, please feel free to add yourself in `[mergify.yml](https://github.com/vllm-project/vllm/blob/main/.github/mergify.yml)` to receive notifications and auto-assignment when the PRs touching the feature you are maintaining. Over time, the ownership will be evaluated and updated through the committers nomination and voting process.

docker/Dockerfile
```diff
@@ -580,7 +580,7 @@ RUN --mount=type=cache,target=/root/.cache/uv \
     else \
         BITSANDBYTES_VERSION="0.46.1"; \
     fi; \
-    uv pip install --system accelerate hf_transfer modelscope "bitsandbytes>=${BITSANDBYTES_VERSION}" 'timm>=1.0.17' 'runai-model-streamer[s3,gcs]>=0.15.3'
+    uv pip install --system accelerate hf_transfer modelscope "bitsandbytes>=${BITSANDBYTES_VERSION}" 'timm>=1.0.17' 'runai-model-streamer[s3,gcs]>=0.15.3' 'grpcio>=1.75.1' 'grpcio-reflection'
```
simon-mo: please feel free to add this to core `requirements.txt`

vllm/grpc/vllm_engine.proto
simon-mo, line 1: cc @robertgshaw2-redhat for the protocol to share with llm-d folks. I'm not expecting a lot of changes to this btw, but if there's some standard we can follow that will be useful as well.

vllm/grpc/vllm_engine.proto
```diff
@@ -0,0 +1,218 @@
...
+  // Get server information
+  rpc GetServerInfo(GetServerInfoRequest) returns (GetServerInfoResponse);
+}
+
+// =====================
+// Common Types
+// =====================
+
+// Sampling parameters for text generation
+message SamplingParams {
```
simon-mo, line 33: Given this is not touched exactly as part of request, can we use protobuf wildcard for this? https://protobuf.dev/programming-guides/proto3/#any or embed a JSON? 

I'm just thinking about the complexity as we evolve SamplingParams itself.

simon-mo (commented):
LGTM on the first pass, pending @njhill @robertgshaw2-redhat quick skims on `grpc_request_manager.py` and `grpc_server.py` for correct usage of `AsyncLLM`.

Another question is whether we should just enable this by default so at `vllm serve` we also start gRPC server at another port?

vllm/grpc/compile_protos.py
hmellor, line 1: Can this be moved to `tools/pre_commit` and a hook added to `.pre-commit-config.yaml`?

If `vllm_engine.proto` is not changed (i.e. it's not passed as an argument to the hook, it should exit early). See the other hook scripts in that directory for examples of this.

vllm/grpc/vllm_engine.proto
```diff
@@ -0,0 +1,218 @@
...
+message TokenizedInput {
+  string original_text = 1;  // For reference/debugging
+  repeated uint32 input_ids = 2;  // Actual token IDs to process
+}
+
+// =====================
+// Generate Request
+// =====================
+
+message GenerateRequest {
```
smarterclayton, line 90: @NickLucche as we discussed, we should review this as it is probably the better approach in general for high efficiency integration for both disaggregation, the coordinator, and broader more decoupled efforts.

The biggest question in both the http and grpc version of this is "how do we allow for reasonable custom fields for things like the coordinator, or to disaggregate how the openai response is generated".

I don't think proto is worse than http, but I do think we should be deliberate in the http schema evolution to avoid creating a mismatch between HTTP and gRPC, especially if we think most people would prefer to use the gRPC endpoint.

vllm/grpc/grpc_request_manager.py
```diff
@@ -0,0 +1,320 @@
...
+
+        Args:
+            async_llm: The AsyncLLM engine instance to submit requests to
+        """
+        self.async_llm = async_llm
+        self.rid_to_collector: dict[str, RequestOutputCollector] = {}
+
+        logger.info("GrpcRequestManager initialized")
+
+    async def generate(
```
njhill: @CatherineSue any reason for not using the `AsyncLLM.generate()` method directly?

I don't immediately see any difference/downside in doing that.

vllm/grpc/vllm_engine.proto
smarterclayton: I think there is a significant cost to use wildcard or json in protobuf, for both implementors and frontends.

I think as SamplingParams evolves, at least proto3 will at least indicate when a field was specified that is no longer supported (and the backend can report an error).

Since the gRPC endpoint is intended to be used by middleware layers (whether that's a future disaggregation coordinator process, the sgl-model-gateway proxy, or the vllm frontend directly), any parameter a client needs to send to the backend benefits from being a protobuf field, and those middle layers may need to deal with multiple deployed versions of the proto schema present in backends during rollouts safely if we move to more disaggregation.  Using fields for now avoids the need to implement that safety checking logic in multiple forms (not just proto, but nested json).

From a Kube perspective (i wrote the protobuf backend), it was a mistake to nest objects opaquely and contributed to a lot of friction from protobuf consumers - the point where many of them were forced to use JSON (since kube was "json first").  The decision to embed sub protocols like Any or JSON should not be taken lightly, and I recommend trying to avoid it as long as possible.

vllm/grpc/grpc_request_manager.py
```diff
@@ -0,0 +1,319 @@
...
+        else None,
+        seed=proto_params.seed if proto_params.HasField("seed") else None,
+        include_stop_str_in_output=proto_params.include_stop_str_in_output,
+        logit_bias=logit_bias,
+        truncate_prompt_tokens=proto_params.truncate_prompt_tokens
+        if proto_params.HasField("truncate_prompt_tokens")
+        else None,
+        structured_outputs=structured_outputs,
+        detokenize=False,
+        output_kind=RequestOutputKind.DELTA if stream else RequestOutputKind.CUMULATIVE,
```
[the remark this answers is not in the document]
njhill: This is a legit comment, stop strings (`stop` param) aren't supported with `detokenize=False`.

vllm/grpc/grpc_request_manager.py
```diff
@@ -0,0 +1,320 @@
...
```
njhill: Actually both `RequestOutputKind.CUMULATIVE` and `RequestOutputKind.DELTA` are streaming, for non-streaming we would want to use `RequestOutputKind.FINAL_ONLY`.

vllm/grpc/vllm_engine.proto
```diff
@@ -0,0 +1,218 @@
...
+  string request_id = 1;
+
+  // Pre-tokenized input (required)
+  TokenizedInput tokenized = 2;
```
njhill: I guess there would be no harm in supporting either text or token ids input?

vllm/grpc/vllm_engine.proto
wseaton: There are some service endpoints that are missing (like `/scale_elastic_ep`), is the goal going to be 100% complete coverage?

vllm/grpc/vllm_engine_pb2_grpc.py
```diff
@@ -0,0 +1,355 @@
+# SPDX-License-Identifier: Apache-2.0
```
dtrifiro: I highly discourage adding generated code to this repository. 

As @njhill  pointed out, in `vllm-tgis-adapter`, we generate the code when building the wheel. See https://github.com/opendatahub-io/vllm-tgis-adapter/blob/main/setup.py#L10-L40

vllm/grpc/compile_protos.py
dtrifiro: Rather than doing this, it would be better to have a a build class that compiles the protos, such as what's done here: https://github.com/opendatahub-io/vllm-tgis-adapter/blob/main/setup.py#L10-L40

vllm/entrypoints/grpc_server.py
```diff
@@ -0,0 +1,520 @@
...
+        Initialize the servicer.
+
+        Args:
+            request_manager: The GrpcRequestManager instance
+        """
+        self.request_manager = request_manager
+        logger.info("VllmEngineServicer initialized")
+
+    async def Generate(
+        self,
```
NickLucche, line 71: dumb q: why the capitalized method names here?

vllm/entrypoints/grpc_server.py
```diff
@@ -0,0 +1,520 @@
...
+    async_llm = AsyncLLM.from_vllm_config(
+        vllm_config=vllm_config,
+        usage_context=UsageContext.OPENAI_API_SERVER,
+        enable_log_requests=not args.disable_log_requests_server,
+        disable_log_stats=args.disable_log_stats_server,
+    )
+
+    logger.info("Model: %s", vllm_config.model_config.model)
+    logger.info("Max model len: %s", vllm_config.model_config.max_model_len)
+    logger.info("Vocab size: %s", vllm_config.model_config.get_vocab_size())
```
NickLucche: should we keep logging fairly consistent with http server?

vllm/entrypoints/grpc_server.py
```diff
@@ -0,0 +1,520 @@
...
+        type=int,
+        default=50051,
+        help="Port to bind gRPC server to",
+    )
+    parser.add_argument(
+        "--disable-log-requests-server",
+        action="store_true",
+        help="Disable request logging on server side",
+    )
+    parser.add_argument(
```
NickLucche, line 511: don't we already have cil args that we can re-use for these?

vllm/grpc/grpc_request_manager.py
```diff
@@ -0,0 +1,320 @@
...
+            # Add request to output processor
+            # Use None for prompt since we have pre-tokenized input
+            # TODO: Support sampling_params.n > 1 (parallel sampling)
+            # When n > 1, we need to:
+            # 1. Create a ParentRequest to track all child requests
+            # 2. Fan out multiple child EngineCoreRequests with different
+            #    request_index values
+            # 3. Aggregate outputs from all children
+            # For now, we only support n=1, so parent_req=None and
+            # request_index=0
```
NickLucche: plan looks good

vllm/grpc/vllm_engine.proto
NickLucche: this is good for /generate

vllm/grpc/vllm_engine.proto
NickLucche: Can we start by centralizing the message definition st http/grpc interface definition is shared? 
re: custom field protobuf side, would we just bite the bullet and use a Any field?

NickLucche (commented):
> Another question is whether we should just enable this by default so at vllm serve we also start gRPC server at another port

@simon-mo I would be in favor of opt-in first (enable with flag, or explicitly start entrypoint) and start as default in a later release.

vllm/grpc/vllm_engine.proto
CatherineSue: that would require grpc server to include tokenizer
this can be added in the future, its currently not supported (one of the reasons to have grpc is to have oai server and other related component to be fully written in rust or other languages which is "more" production ready)

vllm/entrypoints/grpc_server.py
CatherineSue: it's a gRPC convention. The method names in the .proto file are defined in PascalCase.
When grpc_tools.protoc generates the Python code, it preserves these names exactly. The generated _pb2_grpc.py file creates an abstract servicer class with PascalCase methods.
gRPC chose PascalCase as a neutral convention that works across all languages. Python implementations must follow it to match the generated base class, even though it violates PEP 8

vllm/grpc/vllm_engine_pb2_grpc.py
CatherineSue: addressed, thanks for the sample code

vllm/entrypoints/grpc_server.py
CatherineSue: FrontendArgs in `cli_args.py:67-70` defines `host: str | None = None` and `port: int = 8000`
gRPC server uses different defaults: `host = "0.0.0.0"` and `port = 50051` (standard gRPC port)
`--disable-log-requests-server` is redundant
removed

vllm/entrypoints/grpc_server.py
CatherineSue: addressed

vllm/grpc/grpc_request_manager.py
CatherineSue: thanks for pointing that out
i think we eventually should, for now, we should eventually wrap around `AsyncLLM.generate()`, for now, for now, i will keep it that way
there are few follow up PRs
1. pd
2. embeddings
3. classify
grpc server would also need to handle few other rpc, including
1. fetch tokenizer and related files (py code), because a lot of times gateway, or any openai server will be running in cpu instead of gpu nodes, thus it would need to fetch tokenizer either from HF or grpc server, and grpc server is usually desired in production env
2. lora placement
all of those would be follow up PRs

njhill:
Thanks @CatherineSue, the updates look great!

The main concern I have now is related to [this](https://github.com/vllm-project/vllm/pull/30190#discussion_r2604157646) -  the "grpc request manager" is mostly duplicate logic and adds unnecessarily tight coupling to AsyncLLM internals. I've made some changes in another branch to show how much simpler it can be: https://github.com/njhill/vllm/commit/d23b707b3a5cf90c4895b5924427509a18b13fb3. If you're ok with those I can push that commit to this branch.

### Cancellation / aborts

Both vLLM and gRPC have cancellation properly integrated into the asyncio mechanics, so no extra checking/handling is required when wiring these together. When the grpc request is cancelled/closed on the client side, it will automatically propagate to the server and the request will be aborted internally.

Related to this, I don't think there should be an `Abort` protobuf rpc. The API is stateless, you abort a request by just closing the corresponding connection (there is a standard way of doing this in gRPC client libraries).

### Error handling

I noticed that you have defined custom error message types rather than making use of gRPC's native request status propagation. Was just wondering whether this was deliberate and if so the reasoning behind it.

A more grpc native approach would be to close the call with an appropriate status code and message. This can be done in python via the `ServicerContext`, for example:
```python
            await context.abort(StatusCode.INVALID_ARGUMENT, "Missing tokenized input")
```
It's also good to map python `ValueError`s to 400-type errors, I included this change in the commit mentioned above.

vllm/grpc/vllm_engine.proto
njhill: You can also pass a string prompt to `AsyncLLM.generate()` so it would be trivial to expose this as an option so that the gRPC API could also be used in a standalone manner if desired (albeit probably not recommended for performance).

njhill (approved):

njhill (approved):

PR #30195, opened 2025-12-07
[BugFix][DeepSeek-V3.2] Fix backend selection logic for Blackwell

LucasWilkinson:
Fix
```
ValueError: Selected backend AttentionBackendEnum.CUTLASS_MLA is not valid for this configuration. Reason: ['kv_cache_dtype not supported', 'sparse not supported']
```

When running
```
vllm serve deepseek-ai/DeepSeek-V3.2 
```

on 8xB200


houseroad (approved):
Thanks for the fix.

mgoin (approved):

PR #30197, opened 2025-12-07
[Build] Use pip constraints to pin dependency versions

### LINKED ISSUE #28071: [RFC]: Pin all dependencies
### Motivation.

Our builds should be as deterministic as possible. Building an image from the same commit should yield the same result. Otherwise we are in the wild west and things can break arbitrarily when things change underneath us, turning our main branch red.

### Proposed Change.

- We pin all dependencies to precise versions
- Have a way of annotating them as "hard pinned" or not
- We have an automated job that:
   - Compares all deps with latest available versions
   - If there are any with new minor versions, opens a new PR updating the requirements.txt files with them. "hard pinned" dependencies are excluded.
   - The PR text can also include notification of any new major versions
- This job runs automatically at some regular interval (suggest 1-2 times per week), but can also be run manually

### Feedback Period.

_No response_

### CC List.

_No response_

### Any Other Things.

_No response_

njhill:
Used claude code for this.

I think an alternative option could be to use `.in` files like we already do for the test dependencies.

Resolves https://github.com/vllm-project/vllm/issues/28071.


requirements/constraints/README.md
```diff
@@ -0,0 +1,204 @@
...
+### Refreshing All Pins to Latest Compatible
+
+To update all pins to the latest versions that satisfy the requirements:
+
+```bash
+cd requirements
+./compile_constraints.sh
+```
+
+This will resolve all dependencies to their latest compatible versions available on PyPI.
```
markmc, line 127: Given that we're trying to insulate ourselves from package versions that break us, we are expecting this updated package set to break with some regularity ... and yet we do want to update to the latest version of dependencies as quickly as possible

In #28071 you're suggesting (I guess) running `compile_constraints.sh` in an updated job that will submit a PR.

When that fails, someone will narrow down which package broke us, and submit a separate PR with a version block in `requirements/*.txt` e.g. `requests!=2.16.1,>= 2.26.0` ?

njhill: @markmc yes exactly, or adjust code accordingly. But as also discussed in that issue, perhaps dependabot can be used to automate opening the PRs rather than a job that runs the script.

hmellor:
Closing as obsolete: the requirements layout this compiles from was restructured in #39024, and the checked-in constraint files are a Dec 2025 resolution snapshot that would need full regeneration. The underlying goal in #28071 is still worth doing, so please reopen or refile against the current `requirements/` tree if you want to pick it back up.

PR #30201, opened 2025-12-07
kv_transfer: Rename the shared storage connectors

orozery:
This PR renames the {EC}SharedStorageConnectors to {EC}ExampleConnectors. This makes it clear that these connectors should be treated as examples for connector developers, and not as actual production-worthy shared storage connectors.


markmc (approved):
lgtm, thanks

PR #30207, opened 2025-12-07
Enable Cross layers KV cache layout at NIXL Connector

liranschour:
## Purpose
Enable NIXL Connector to us the new continuous cross layer KV cache layout described in RFC [#27742](https://github.com/vllm-project/vllm/issues/27742) and implemented in [#27743](https://github.com/vllm-project/vllm/pull/27743).
Demonstrate performance improvement of more the 2x in Tok/sec and TTFT due to dramatic reduction of fragmentation of transfer buffers.

Tested with P!=D with run_accuracy_test.sh P=1 D=2

Tested on H100 1P1D Qwen/Qwen3-0.6B:

| branch           | num reqs | input len | TTFT      | ITL  | tok/s      | Desc/transfer |
|------------------|----------|-----------|-----------|------|------------|---------------|
| main             | 1000     | 16        | 18756.42  | 5.35 | 5288.41    | 56            |
| kv_cross_layers  | 1000     | 16        | 8494.44   | 8.74 | 8572.56    | 1             |
| main             | 128      | 1024      | 1660.84   | 9.40 | 37945.20   | 3528          |
| kv_cross_layers  | 128      | 1024      | 686.98    | 9.26 | 55418.76   | 1             |
| main             | 128      | 10240     | 11140.52  | 42.78| 62339.74   | 34000         |
| kv_cross_layers  | 128      | 10240     | 5226.71   | 14.41|117631.48   | 422           |


vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -348,6 +351,17 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
...
+    def register_cross_layers_kv_cache(
+        self, kv_cache: torch.Tensor, attn_backend: type[AttentionBackend]
+    ):
+        assert self.connector_worker is not None
+
+        cross_layer_name = "ALL_LAYERS"
+
+        kv_caches = {cross_layer_name: kv_cache}
+        self.connector_worker.cross_layers = True
+        self.connector_worker.register_kv_caches(kv_caches)
```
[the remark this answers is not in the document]
liranschour: In cross layers cache.shape[0] holds the number of Blocks.
And num_layers should be 1.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -783,6 +798,7 @@ def __init__(self, vllm_config: VllmConfig, engine_id: str):
         if vllm_config.kv_transfer_config is None:
             raise ValueError("kv_transfer_config must be set for NixlConnector")
         self.kv_transfer_config = vllm_config.kv_transfer_config
+        self.cross_layers = False
```
orozery: I think that we should move this variable into `TpKVTopology`.
Then `TpKVTopology.split_k_and_v` can return False if we are in a cross layers case.
This should also address the bug that gemini pointed out in the current revision of this PR.

This change means we should defer the initialization of `TpKVTopology` from `NixlConnectorWorker.__init__` to `NixlConnectorWorker.register_kv_caches`.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1202,7 +1218,7 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
 
         # TODO (NickLucche): Get kernel_block_size in a cleaner way
         # NHD default "view" for non-MLA cache
-        if self.device_type == "cpu":
+        if self.device_type == "cpu" or self.cross_layers:
             block_size_position = -2
         else:
             block_size_position = -2 if self.use_mla else -3
```
orozery: As I suggested to eliminate the `self.cross_layers` variable,
I suggest that we move this entire logic behind a new property `TpKVTopology.block_size_position`.
cc @NickLucche

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -186,6 +187,7 @@ def compute_nixl_compatibility_hash(
         # Attention backend and KV cache dtype affect memory layout
         "attn_backend_name": attn_backend_name,
         "cache_dtype": str(cache_config.cache_dtype),
+        "cross_layers": CROSS_LAYERS,
```
orozery: An option to consider:
Defer the computation of `self.compat_hash` to `NixlConnectorWorker.register_kv_caches`.
This will allow you to conditionally set this new `cross_layers` compatibility constraint only if cross layers is actually being used.

orozery (changes requested):
Looks good overall. Thanks!

Also wanted to share a summary I prepared on how this change effects the I/O size per different models:

| Model | Old I/O size | New I/O size |
|-------|--------------|--------------|
| deepseek-ai/DeepSeek-R1-Distill-Qwen-32B (tensor_parallel_size=2) | 16 KB | 2 MB |
| deepseek-ai/DeepSeek-V2-Lite-Chat (GPU block size=64) | 72 KB | 1.9 MB |
| meta-llama/Llama-3.1-8B-Instruct | 32 KB | 2 MB |
| meta-llama/Llama-3.2-1B-Instruct | 16 KB | 0.5 MB |
| meta-llama/Llama-3.1-70B-Instruct | 8 KB | 1.25 MB |
| mistralai/Mistral-7B-Instruct-v0.2 | 32 KB | 2 MB |
| mistralai/Mistral-Small-24B-Instruct-2501 | 32 KB | 2.5 MB |
| Qwen/Qwen2.5-3B-Instruct | 8 KB | 0.56 MB |
| Qwen/Qwen3-0.6B | 32 KB | 1.75 MB |
| Qwen/Qwen2.5-7B-Instruct | 16 KB | 0.87 MB |
| Qwen/Qwen3-4B-Instruct-2507 | 32 KB | 2.25 MB |
| Qwen/Qwen2.5-1.5B-Instruct | 8 KB | 0.44 MB |
| Qwen/Qwen3-8B | 28 KB | 1.97 MB |
| Qwen/Qwen3-1.7B | 32 KB | 1.75 MB |
| Qwen/Qwen3-32B (tensor_parallel_size=2) | 16 KB | 2 MB |

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
liranschour: Done. Moved this logic under TpKVToplogy

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
liranschour: Done. Moved this logic under TpKVToplogy

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
liranschour: I deferred the compatibility computation to NixlConnectorWorker.register_kv_caches

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -74,6 +74,7 @@
 #   2: Add remote_request_id to kv_transfer_params
 #
 NIXL_CONNECTOR_VERSION: int = 2
+CROSS_LAYERS: bool = True
```
NickLucche: what's this for?

NickLucche (commented):
Thanks for the work @liranschour and @orozery !
Planning to review/run some benchmarks asap, in the meantime can we run tp_config_sweep.ch to verify heteroTP correctness with this layout? Thanks!

Also cc @xuechendi

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
liranschour: I will remove this. It is a leftover from previous version of the code.

liranschour: Removed @NickLucche

vllm/distributed/kv_transfer/kv_connector/utils.py
```diff
@@ -200,6 +200,7 @@ class TpKVTopology:
     attn_backend: type[AttentionBackend]
     engine_id: str
     remote_block_size: dict[str, int]
+    cross_layers: bool
```
orozery: nit: maybe change to `cross_layers_blocks` / `cross_layers_kv_cache` or something else that is a bit more self-explanatory? 
Also applies for the other `cross_layers` introduced in this PR.

vllm/distributed/kv_transfer/kv_connector/utils.py
```diff
@@ -233,6 +239,14 @@ def tp_size(self) -> int:
     def block_size(self) -> int:
         return self.remote_block_size[self.engine_id]
 
+    def block_size_position(self, device_type: str) -> int:
+        if device_type == "cpu" or self.cross_layers:
```
orozery: I think that `device_type` is nixl-connector specific. Can you move it out from here to the nixl-connector?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -930,30 +943,13 @@ def __init__(self, vllm_config: VllmConfig, engine_id: str):
         logger.debug("Detected attention backend %s", self.backend_name)
         logger.debug("Detected kv cache layout %s", self.kv_cache_layout)
 
-        self.compat_hash = compute_nixl_compatibility_hash(
```
orozery, line 987: I think it's better to leave initialization to None here, so that all fields are declared in `__init__`.
See for example how it's done in `gpu_model_runner.py`:
```
        # Lazy initializations
        # self.model: nn.Module  # Set after load_model
        # Initialize in initialize_kv_cache
        self.kv_caches: list[torch.Tensor] = []
        # Initialize in initialize_kv_cache_tensors
        self.cross_layers_kv_cache: torch.Tensor | None = None
        self.cross_layers_attn_backend: type[AttentionBackend] | None = None
```

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1163,6 +1159,33 @@ def request_ready(f: Future[Any], entry=(req_id, meta)):
     def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
         """Register the KV Cache data in nixl."""
 
+        backend = get_attn_backend(
```
orozery: nit: introduce `self.attn_backend` in `__init__` and re-use it here?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1163,6 +1159,33 @@ def request_ready(f: Future[Any], entry=(req_id, meta)):
...
+
+        self.kv_topo = TpKVTopology(
+            tp_rank=self.tp_rank,
+            engine_id=self.engine_id,
+            remote_tp_size=self._tp_size,  # shared state
+            remote_block_size=self._block_size,  # shared state
+            is_mla=self.use_mla,
+            total_num_kv_heads=self.model_config.get_total_num_kv_heads(),
+            attn_backend=backend,
+            cross_layers=next(iter(kv_caches)) == "ALL_LAYERS",
```
orozery: There is a small edge-case if a model has a layer named `ALL_LAYERS` :)
In the offloading connector we detect a cross layers kv cache differently:
https://github.com/vllm-project/vllm/blob/33278073d68940dcaff579ab2dc316700e1db87a/vllm/v1/kv_offload/worker/cpu_gpu.py#L197-L205

Actually, maybe you can introduce `TpKVTopology.tensor_shape` initialized from the `kv_caches` tensor shape, and then set `_cross_layers` in `TpKVTopology.__post_init__` ?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1163,6 +1159,33 @@ def request_ready(f: Future[Any], entry=(req_id, meta)):
...
+        )
+        self._use_pallas = self.kv_topo._use_pallas
+
+        self.compat_hash = compute_nixl_compatibility_hash(
+            self.vllm_config, self.backend_name, self.kv_topo.cross_layers
+        )
+        self.enforce_compat_hash = self.kv_transfer_config.get_from_extra_config(
+            "enforce_handshake_compat", True
+        )
```
orozery: This can be moved back to `__init__`.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1163,6 +1159,33 @@ def request_ready(f: Future[Any], entry=(req_id, meta)):
...
+            tp_rank=self.tp_rank,
+            engine_id=self.engine_id,
+            remote_tp_size=self._tp_size,  # shared state
+            remote_block_size=self._block_size,  # shared state
+            is_mla=self.use_mla,
+            total_num_kv_heads=self.model_config.get_total_num_kv_heads(),
+            attn_backend=backend,
+            cross_layers=next(iter(kv_caches)) == "ALL_LAYERS",
+        )
+        self._use_pallas = self.kv_topo._use_pallas
```
orozery: We can either move this back to `__init__` (setting it using `attn_backend == AttentionBackendEnum.PALLAS`),
or remove it all together and switch `self._use_pallas` to `self.kv_topo.use_pallas` (new property).

orozery (changes requested):

vllm/distributed/kv_transfer/kv_connector/utils.py
liranschour: Done

vllm/distributed/kv_transfer/kv_connector/utils.py
liranschour: Done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
liranschour: Done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
liranschour: Done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
liranschour: Done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
liranschour: Done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
liranschour: Done

vllm/distributed/kv_transfer/kv_connector/utils.py
```diff
@@ -216,14 +217,24 @@ def __post_init__(self):
         attn_backend = AttentionBackendEnum[self.attn_backend.get_name()]
         self._use_pallas = attn_backend == AttentionBackendEnum.PALLAS
 
+        test_shape = self.attn_backend.get_kv_cache_shape(
+            num_blocks=1234, block_size=16, num_kv_heads=8, head_size=256)
+
+        self._cross_layers_blocks = len(self.tensor_shape) != len(test_shape)
```
orozery:
```suggestion
        self._cross_layers_blocks = len(self.tensor_shape) != len(test_shape)
        if self._cross_layers_blocks:
            # expect one additional dimension (num_layers)
            assert len(self.tensor_shape) == len(test_shape) + 1
```

vllm/distributed/kv_transfer/kv_connector/utils.py
```diff
@@ -233,6 +244,17 @@ def tp_size(self) -> int:
...
 
+    @property
+    def use_pallas(self) -> bool:
+        return self._use_pallas
+
+    @property
+    def cross_layers_blocks(self) -> bool:
+        return self._cross_layers_blocks
+
+    def block_size_position(self, device_type: str) -> int:
```
orozery:
```suggestion
    @property
    def block_size_position(self) -> int:
```

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -917,25 +930,20 @@ def __init__(self, vllm_config: VllmConfig, engine_id: str):
...
         logger.debug("Detected attention backend %s", self.backend_name)
         logger.debug("Detected kv cache layout %s", self.kv_cache_layout)
 
-        self.compat_hash = compute_nixl_compatibility_hash(
-            self.vllm_config, self.backend_name
-        )
-        self.enforce_compat_hash = self.kv_transfer_config.get_from_extra_config(
-            "enforce_handshake_compat", True
-        )
+        self.compat_hash: Optional[str] = None
```
orozery:
```suggestion
        self.compat_hash: str | None = None
```

orozery: What about defining `self.kv_topo`?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -917,25 +930,20 @@ def __init__(self, vllm_config: VllmConfig, engine_id: str):
...
         logger.debug("Detected attention backend %s", self.backend_name)
         logger.debug("Detected kv cache layout %s", self.kv_cache_layout)
 
-        self.compat_hash = compute_nixl_compatibility_hash(
-            self.vllm_config, self.backend_name
-        )
-        self.enforce_compat_hash = self.kv_transfer_config.get_from_extra_config(
-            "enforce_handshake_compat", True
-        )
+        self.compat_hash: str | None = None
```
orozery, line 1014: what about declaring `self.kv_topo`?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1197,29 +1214,27 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
...
-        split_k_and_v = self.kv_topo.split_k_and_v
         tensor_size_bytes = None
 
         # TODO (NickLucche): Get kernel_block_size in a cleaner way
         # NHD default "view" for non-MLA cache
-        if self.device_type == "cpu":
-            block_size_position = -2
-        else:
-            block_size_position = -2 if self.use_mla else -3
+        block_size_position = self.kv_topo.block_size_position(self.device_type)
```
orozery:
```suggestion
        if self.device_type == "cpu":
            block_size_position = -2
        else:
            block_size_position = self.kv_topo.block_size_position
```

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1197,29 +1214,27 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
...
 
         # Enable different block lengths for different layers when MLA is used.
         self.block_len_per_layer = list[int]()
         self.slot_size_per_layer = list[int]()  # HD bytes in kv terms
         for layer_name, cache_or_caches in xfer_buffers.items():
-            cache_list = cache_or_caches if split_k_and_v else [cache_or_caches]
-
+            cache_list = (
+                cache_or_caches
+                if not self.kv_topo.cross_layers_blocks and self.kv_topo.split_k_and_v
```
orozery:
```suggestion
                if self.kv_topo.split_k_and_v
```

vllm/distributed/kv_transfer/kv_connector/utils.py
```diff
@@ -198,6 +198,7 @@ class TpKVTopology:
     is_mla: bool
     total_num_kv_heads: int
     attn_backend: type[AttentionBackend]
+    tensor_shape: torch.Size
```
orozery: This will break other users of `TpKVTopology` (MooncakeConnector).
So either adapt their code, or move this field to the bottom and add a default (which should be checked before being used below):
```
tensor_shape: torch.Size | None = None
```

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -186,6 +186,7 @@ def compute_nixl_compatibility_hash(
         # Attention backend and KV cache dtype affect memory layout
         "attn_backend_name": attn_backend_name,
         "cache_dtype": str(cache_config.cache_dtype),
+        "cross_layers": cross_layers,
```
orozery: nit:
```suggestion
        "cross_layers_blocks": cross_layers_blocks,
```

orozery (changes requested):

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
liranschour: @NickLucche I am able to run successfully tp_config_sweep_accuracy_test.sh only on Qwen/Qwen3-0.6B:

- "GPU_MEMORY_UTILIZATION=0.6 PREFILLER_TP_SIZE=2 DECODER_TP_SIZE=2"
- "GPU_MEMORY_UTILIZATION=0.6 PREFILLER_TP_SIZE=1 DECODER_TP_SIZE=2"
- 
DeppSeek models are failing on my environment also on the main branch. So probably I have some error in the environment.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
liranschour: Done

liranschour: Done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
liranschour: Done

vllm/distributed/kv_transfer/kv_connector/utils.py
liranschour: Done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1723,7 +1742,9 @@ def _process_local_gt_remote(blocks_to_update, block_size_ratio):
         if len(self.device_kv_caches) == 0:
             return
         split_k_and_v = not (
-            self.use_mla or self._use_pallas or self.kv_topo.is_kv_layout_blocks_first
+            self.use_mla
+            or self.kv_topo.use_pallas
+            or self.kv_topo.is_kv_layout_blocks_first
         )
```
orozery: Can we remove this variable and replace its uses with `self.kv_topo.split_k_and_v` ?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
orozery: I think you should assert `is not None` in any function that use `self.compat_hash` and `self.kv_topo` ?

vllm/distributed/kv_transfer/kv_connector/utils.py
```diff
@@ -216,14 +217,29 @@ def __post_init__(self):
         attn_backend = AttentionBackendEnum[self.attn_backend.get_name()]
         self._use_pallas = attn_backend == AttentionBackendEnum.PALLAS
 
+        test_shape = self.attn_backend.get_kv_cache_shape(
+            num_blocks=1234, block_size=16, num_kv_heads=8, head_size=256
+        )
+
+        if self.tensor_shape is not None:
```
orozery, line 335: You should define self._cross_layers_blocks (and set to False) even if self.tensor_shape is None.

orozery (changes requested):

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
liranschour: Done

vllm/distributed/kv_transfer/kv_connector/utils.py
liranschour: Done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1163,6 +1169,23 @@ def request_ready(f: Future[Any], entry=(req_id, meta)):
...
+            remote_tp_size=self._tp_size,  # shared state
+            remote_block_size=self._block_size,  # shared state
+            is_mla=self.use_mla,
+            total_num_kv_heads=self.model_config.get_total_num_kv_heads(),
+            attn_backend=self.attn_backend,
+            tensor_shape=next(iter(kv_caches.values())).shape,
+        )
+
+        assert self.kv_topo is not None, "kv_topo is not initialized"
+
```
orozery: You can remove this check as it is immediately after initialization

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -967,6 +970,8 @@ def _nixl_handshake(
 
         start_time = time.perf_counter()
 
+        assert self.kv_topo is not None, "kv_topo is not initialized"
+
```
orozery: nit: I would move this assert to directly precede the line that accesses `self.kv_topo`.
Also, I think we can keep all the added asserts without any text message (e.g. just `assert self.kv_topo is not None`).

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
liranschour: Done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
liranschour: Done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -917,25 +930,21 @@ def __init__(self, vllm_config: VllmConfig, engine_id: str):
...
         logger.debug("Detected attention backend %s", self.backend_name)
         logger.debug("Detected kv cache layout %s", self.kv_cache_layout)
 
-        self.compat_hash = compute_nixl_compatibility_hash(
-            self.vllm_config, self.backend_name
-        )
-        self.enforce_compat_hash = self.kv_transfer_config.get_from_extra_config(
-            "enforce_handshake_compat", True
-        )
+        self.compat_hash: str | None = None
```
orozery, line 1014: one last nit: can you add a comment above this line like:
`# lazy initialized in register_kv_caches`

liranschour: Done

orozery (approved):
Thanks @liranschour !
@NickLucche Can you please review?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1301,29 +1334,28 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
...
-        split_k_and_v = self.kv_topo.split_k_and_v
         tensor_size_bytes = None
 
         # TODO (NickLucche): Get kernel_block_size in a cleaner way
         # NHD default "view" for non-MLA cache
         if self.device_type == "cpu":
             block_size_position = -2
         else:
-            block_size_position = -2 if self.use_mla else -3
+            block_size_position = self.kv_topo.block_size_position
```
NickLucche: we can probably move the device_type check inside`kv_topo.block_size_position` too.

vllm/distributed/kv_transfer/kv_connector/utils.py
```diff
@@ -329,14 +330,27 @@ def __post_init__(self):
             len(kv_cache_shape) == 5 and kv_cache_shape[0] == 1
         )
 
+        test_shape = self.attn_backend.get_kv_cache_shape(
+            num_blocks=1234, block_size=16, num_kv_heads=8, head_size=256
+        )
```
NickLucche: we already define a "test_shape" at L324 with `kv_cache_shape`

vllm/distributed/kv_transfer/kv_connector/utils.py
```diff
@@ -329,14 +330,27 @@ def __post_init__(self):
...
+
+        self._cross_layers_blocks = False
+        if self.tensor_shape is not None:
+            self._cross_layers_blocks = len(self.tensor_shape) != len(test_shape)
+            if self._cross_layers_blocks:
+                # expect one additional dimension (num_layers)
+                assert len(self.tensor_shape) == len(test_shape) + 1
```
NickLucche: I wonder whether we should simplify to 

```suggestion
            self._cross_layers_blocks = len(self.tensor_shape) == len(test_shape) + 1
```

NickLucche (commented):
@liranschour @orozery  Only left a few comments, ptal and thanks for the work on this important feature.
Overall this looks good. 
My only real concern is whether we should enable this feature optionally, at least for the first 1/2 releases to give the community a window to report incompatibilities/bugs. Also we definitely need a way to opt out of it regardless eg if the user wants to run a model with unsupported feature (see below).
Therefore my proposal is to make `prefer_cross_layer_blocks` dependent on config:
user can opt-in by passing a specific kv connector config eg 
```
'{"kv_connector":"NixlConnector","kv_role":"kv_both", "enable_cross_layer_kv_layout": "True"}`
```

Second point, in terms of compatibility checking, we should probably raise an unsupported error/assert, or even better verify if a simple tweak is needed, the following features:
- remote block_size != local block_size (block_size_ratio) 
- different local/remote layout (`enable_permute_local_kv`)
- cpu-backed transfers (`kv_buffer_device: "cpu"`) 

The check can be done in `_validate_remote_agent_handshake`

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
orozery: I actually suggested that we don't, since I think this logic is specific to the nixl connector.

vllm/distributed/kv_transfer/kv_connector/utils.py
orozery: Agreed, we should remove this.

tests/v1/kv_connector/unit/test_nixl_connector.py
```diff
@@ -366,7 +369,22 @@ def test_kv_transfer_handshake(dist_init):
...
+            is_mla=decode_worker.use_mla,
+            total_num_kv_heads=decode_worker.model_config.get_total_num_kv_heads(),
+            attn_backend=decode_worker.attn_backend,
+            tensor_shape=next(iter(kv_caches.values())).shape,
+        )
+        decode_worker.compat_hash = compute_nixl_compatibility_hash(
+            decode_worker.vllm_config,
+            decode_worker.backend_name,
+            decode_worker.kv_topo.cross_layers_blocks,
+        )
```
orozery: Can we call `decode_connector.register_kv_caches` instead?
And in other places you changed in the test file as well.

vllm/distributed/kv_transfer/kv_connector/utils.py
liranschour: Removed

vllm/distributed/kv_transfer/kv_connector/utils.py
liranschour: Done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
liranschour: Done

tests/v1/kv_connector/unit/test_nixl_connector.py
liranschour: Done

docs/features/nixl_connector_usage.md
```diff
@@ -184,6 +184,13 @@ Support use case: Prefill with 'HND' and decode with 'NHD' with experimental con
 --kv-transfer-config '{..., "enable_permute_local_kv":"True"}'
 ```
 
+### Cross layers blocks
+By default, this feature is enabled. On backends that support this feature, each logical block is contiguous in physical memory. This reduces the number of buffers that need to be transferred.
```
orozery:
```suggestion
By default, this feature is enabled. On attention backends that support this feature, each logical block is contiguous in physical memory. This reduces the number of buffers that need to be transferred.
```

docs/features/nixl_connector_usage.md
```diff
@@ -184,6 +184,13 @@ Support use case: Prefill with 'HND' and decode with 'NHD' with experimental con
...
+To disable this feature:
```
orozery:
```suggestion
You can disable this feature using:
```

docs/features/nixl_connector_usage.md
```diff
@@ -184,6 +184,13 @@ Support use case: Prefill with 'HND' and decode with 'NHD' with experimental con
...
+```
+--kv-transfer-config '{..., "kv_connector_extra_config":{"enable_cross_layers_block": "False"}}'
```
orozery:
```suggestion
--kv-transfer-config '{..., "kv_connector_extra_config": {"enable_cross_layers_block": "False"}}'
```

Also, let's change to enable_cross_layers_blocks (blocks instead of block)?

tests/v1/kv_connector/unit/test_nixl_connector.py
```diff
@@ -396,6 +403,93 @@ class FakeNixlConnectorWorker(NixlConnectorWorker):
...
+            total_num_kv_heads=self.model_config.get_total_num_kv_heads(),
+            attn_backend=self.attn_backend,
+            tensor_shape=test_shape,
+        )
+
+        self.compat_hash = compute_nixl_compatibility_hash(
+            self.vllm_config, self.backend_name, self.kv_topo.cross_layers_blocks
+        )
+
+    def _nixl_handshake(
```
orozery, line 431: this looks like some kind of merge/rebase error? (this is not code you added)

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -298,6 +299,20 @@ def add_new_req_to_recv(
 
 
 class NixlConnector(KVConnectorBase_V1):
+    @property
+    def prefer_cross_layer_blocks(self) -> bool:
+        backend = get_current_attn_backend(self._vllm_config)
+        if backend().get_name() not in [
+            "FLASH_ATTN",
+            "FLASHINFER",
+        ]:
```
orozery: nit:
```suggestion
        if backend().get_name() not in (
            "FLASH_ATTN",
            "FLASHINFER",
        ):
```

tests/v1/kv_connector/unit/test_nixl_connector.py
liranschour: Done

docs/features/nixl_connector_usage.md
liranschour: Done

docs/features/nixl_connector_usage.md
liranschour: Done

docs/features/nixl_connector_usage.md
liranschour: Done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
liranschour: Done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1301,29 +1344,21 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
         # (roughly 8KB vs 5KB).
         # Conversely for FlashInfer, K and V are registered in the same region
         # to better exploit the memory layout (ie num_blocks is the first dim).
-        split_k_and_v = self.kv_topo.split_k_and_v
         tensor_size_bytes = None
 
-        # TODO (NickLucche): Get kernel_block_size in a cleaner way
-        # NHD default "view" for non-MLA cache
-        if self.device_type == "cpu":
-            block_size_position = -2
```
orozery, line 1310: I think we should not touch this. Just the `else`.
@NickLucche how can we be sure that we haven't broken anything with `kv_buffer_device="cpu"`?
Maybe `run_accuracy_test.sh --kv_buffer_device=cpu`?

NickLucche: yes there's an env var for run_accuracy_test, I don't think it's an argument though

docs/features/nixl_connector_usage.md
```diff
@@ -184,6 +184,15 @@ Support use case: Prefill with 'HND' and decode with 'NHD' with experimental con
 --kv-transfer-config '{..., "enable_permute_local_kv":"True"}'
 ```
 
+### Cross layers blocks
+
+By default, this feature is enabled. On attention backends that support this feature, each logical block is contiguous in physical memory. This reduces the number of buffers that need to be transferred.
+You can disable this feature:
+
+```bash
+--kv-transfer-config '{..., "kv_connector_extra_config": {"enable_cross_layers_blocks": "False"}}'
```
NickLucche: *is disabled

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -298,6 +299,20 @@ def add_new_req_to_recv(
 
 
 class NixlConnector(KVConnectorBase_V1):
+    @property
+    def prefer_cross_layer_blocks(self) -> bool:
+        backend = get_current_attn_backend(self._vllm_config)
+        if backend().get_name() not in (
```
NickLucche, line 305: note to self: for tests where attn context isn't set, this defaults to another way to get the attn backend. Let's double check the backend is consistent.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1005,10 +1028,14 @@ def __init__(self, vllm_config: VllmConfig, engine_id: str):
...
             is_mla=self.use_mla,
             total_num_kv_heads=self.model_config.get_total_num_kv_heads(),
-            attn_backend=backend,
+            attn_backend=self.attn_backend,
         )
         self._physical_blocks_per_logical_kv_block = 1
 
+        self.enforce_compat_hash = self.kv_transfer_config.get_from_extra_config(
+            "enforce_handshake_compat", True
+        )
```
NickLucche: cruft, enforce_compat_hash and kv_topo are re-defined in register_kv_caches (weird mypy doesnt complain either..)

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1968,6 +2012,7 @@ def _get_new_notifs(self) -> set[str]:
 
                 # NOTE: `tp_ratio` is the opposite when swapping local<>remote
                 n_consumers = int(tp_size)
+                assert self.kv_topo is not None
```
NickLucche: nit: probably cleaner to move these asserts at the top of the affected functions so you can chain `assert __ and __`

vllm/distributed/kv_transfer/kv_connector/utils.py
```diff
@@ -316,27 +316,61 @@ class TpKVTopology:
...
-            num_blocks=1, block_size=16, num_kv_heads=1, head_size=1
+            num_blocks=1, block_size=16, num_kv_heads=4, head_size=1
         )
         # Non-MLA backends caches have 5 dims [2, num_blocks, H,N,D],
         # we just mock num_blocks to 1 for the dimension check below.
         self._is_kv_layout_blocks_first = (
             len(kv_cache_shape) == 5 and kv_cache_shape[0] == 1
         )
 
+        self._kv_heads_position: int | None = None
```
NickLucche, line 333: cruft? This is unused (is pre-commit working?)

vllm/distributed/kv_transfer/kv_connector/utils.py
```diff
@@ -346,6 +380,19 @@ def tp_size(self) -> int:
...
+        return self._cross_layers_blocks
+
+    @property
+    def block_size_position(self) -> int:
+        return self._physical_block_size_position
+
+    @property
+    def physical_kv_heads_position(self) -> int:
+        assert self._physical_kv_heads_position is not None
+        return self._physical_kv_heads_position
```
NickLucche: this property is also unused so we shouldnt be adding that

vllm/distributed/kv_transfer/kv_connector/utils.py
NickLucche: ok I see, rename to `self._physical_kv_heads_position`

vllm/distributed/kv_transfer/kv_connector/utils.py
NickLucche: I wonder why we're computing `self._physical_kv_heads_position` at this point 🤔

vllm/distributed/kv_transfer/kv_connector/utils.py
```diff
@@ -316,27 +316,61 @@ class TpKVTopology:
     attn_backend: type[AttentionBackend]
     engine_id: EngineId
     remote_block_size: dict[EngineId, int]
+    tensor_shape: torch.Size | None = None
+    device_type: str = "cuda"
```
NickLucche: cruft? this `device_type` also looks unused..

NickLucche (approved):
Thanks for the work and the patience with this PR @liranschour @orozery !

Accepting to unblock this PR and speed the process up now that this feature is being added as opt-in to start with!

Left some comments, I find that there's still quite a bit of cruft that I am frankly surprised the linter didn't catch..

docs/features/nixl_connector_usage.md
liranschour: Fixed

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
liranschour: Done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
liranschour: Moved asserts to the top of affected functions

vllm/distributed/kv_transfer/kv_connector/utils.py
liranschour: Removed

vllm/distributed/kv_transfer/kv_connector/utils.py
liranschour: Removed

vllm/distributed/kv_transfer/kv_connector/utils.py
liranschour: Removed

vllm/distributed/kv_transfer/kv_connector/utils.py
```diff
@@ -346,6 +375,14 @@ def tp_size(self) -> int:
     def block_size(self) -> int:
         return self.remote_block_size[self.engine_id]
 
+    @property
+    def cross_layers_blocks(self) -> bool:
+        return self._cross_layers_blocks
+
+    @property
+    def block_size_position(self) -> int:
+        return self._physical_block_size_position
```
NickLucche, line 384: note: this is not fully dependent on providing a tensor shape, we can likely refactor later

PR #30215, opened 2025-12-07
[Feature] Auto-calculate num_redundant_experts for EPLB (#30075)

### LINKED ISSUE #30075: [Feature]: Default eplb num_redundant_experts to the lowest valid value if unspecified

EPLB requires the number of experts to be chosen up front and there is a known minimum valid value that can be derived from the vllm startup configuration.  Since extra EPLB experts trades kv cache memory for potential performance improvements, but that is not guaranteed to pay off, having the EPLB value default to the minimum valid value would reduce friction on enabling EPLB the first time until users are ready to tune.

As a consequence, it would also streamline templating the same config to work across multiple EP sizes for the default case.

### Alternatives

_No response_

### Additional context

_No response_

parlakisik:
## Purpose:
Users had to manually calculate num_redundant_experts when enabling EPLB, which required understanding the formula and their model's configuration. This created friction for adoption and was error-prone.

## Solution:
- Auto-calculate num_redundant_experts when set to -1 (new default)
- Formula: (ep_size - (num_logical_experts % ep_size)) % ep_size
- Ensures minimum redundancy for even expert distribution across EP ranks
- Users can still set 0 (no redundancy) or positive values (manual control)
- Logs calculated value for transparency

## Test Plan
- Auto-calculation with various expert/EP size combinations (8 scenarios)
- Manual override preservation (explicit values honored)
- Explicit zero preservation (0 not auto-calculated)
- Validation when EPLB disabled with positive num_redundant_experts
- Formula correctness verification (ensures even distribution)

Fixes: https://github.com/vllm-project/vllm/issues/30075

### DIFF
--- a/vllm/config/parallel.py
+++ b/vllm/config/parallel.py
@@ -61,8 +61,15 @@ class EPLBConfig:
     of the last `lb_window_size` steps will be used for rearranging experts.
     """
 
-    num_redundant_experts: int = Field(default=0, ge=0)
-    """Number of redundant experts to use for expert parallelism."""
+    num_redundant_experts: int = Field(default=-1, ge=-1)
+    """Number of redundant experts to use for expert parallelism.
+
+    If set to -1 (default), the minimum valid value will be automatically
+    calculated based on the model's expert configuration and EP size to
+    ensure even distribution across ranks.
+    Set to 0 for no redundancy, or a higher value to improve load balancing
+    at the cost of additional KV cache memory.
+    """
 
     log_balancedness: bool = False
     """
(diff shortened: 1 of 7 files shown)


vllm/config/vllm.py
```diff
@@ -360,6 +360,52 @@ def enable_trace_function_call_for_thread(self) -> None:
...
+        # Update config value
+        parallel_config.eplb_config.num_redundant_experts = min_redundant_experts
+
+        logger.info(
+            "Auto-calculated num_redundant_experts=%d for EPLB "
+            "(num_logical_experts=%d, ep_size=%d)",
+            min_redundant_experts,
+            num_logical_experts,
+            ep_size,
+        )
```
hmellor: Could this logic be moved to `EPLBConfig` in a method that has `num_logical_experts` as an argument?

You can then call it from `VllmConfig.__post_init__` like 

```py
self.parallel_config.eplb_config.auto_set_num_redundant_experts(
    self.model_config.get_num_experts()
)
```

parlakisik:
@codex review

vllm/config/parallel.py
```diff
@@ -294,28 +347,42 @@ def _validate_parallel_config(self) -> Self:
...
+            if self.enable_eplb:
+                raise ValueError("enable_expert_parallel must be True to use EPLB.")
+            if self.eplb_config.num_redundant_experts > 0:
+                raise ValueError(
+                    "num_redundant_experts can only be set when "
+                    "enable_expert_parallel is True"
+                )
+            elif self.eplb_config.num_redundant_experts == -1:
+                self.eplb_config.num_redundant_experts = 0
+            return
```
njhill:
```suggestion
            return self
```
I think this is also the cause of the docs failure

parlakisik: ahh good catch .

parlakisik:
@codex review

PR #30220, opened 2025-12-07
[BugFix] Unblock use of LoRA with data parallel mode

njhill:
The metrics initialization code currently includes a check for LoRA + DP which will hard fail. However this combination should work fine. The problem is that the lora prometheus metrics might be incorrect in this case.

Change the check from raising an exception to logging a warning about the metric.

cc @mgoin


mgoin (approved):
Thanks for looking into this!

PR #30223, opened 2025-12-08
[BugFix] Streaming response output field empty in response.completed event

penfree:
# Summary

  Fixed a bug where the output field in response.completed event was empty when using streaming mode with non-Harmony
  models (e.g., Qwen), even though the streaming delta events contained correct content.

# Problem

  When using the Responses API with stream=True, the streaming events (response.output_text.delta) correctly returned
  text deltas, but the final response.completed event had an empty output array:
```json
  {
    "type": "response.completed",
    "response": {
      "output": [],  // Should contain the message with full text
      "status": "completed"
    }
  }
```

# Root Cause

  In _process_simple_streaming_events, each iteration overwrites context.last_output with only the latest incremental
  output (delta). When responses_full_generator builds the final response, it reads context.last_output.outputs[0].text
  which only contains the last token's text instead of the accumulated full text.

# Fix

  At the end of _process_simple_streaming_events, save the accumulated previous_text and previous_token_ids back to
  context.last_output:
```python
  # Update context.last_output with accumulated text and token_ids
  # so that responses_full_generator can build the correct final response
  if (
      isinstance(context, SimpleContext)
      and context.last_output is not None
      and context.last_output.outputs
  ):
      context.last_output.outputs[0].text = previous_text
      context.last_output.outputs[0].token_ids = tuple(previous_token_ids)
```

# Test Plan

- Added test_streaming_output_consistency unit test that verifies:
  - Concatenated text from all response.output_text.delta events
  - Matches output_text in the final response.completed event
- Manual testing with Qwen/Qwen2-0.5B-Instruct model confirmed the fix

# Files Changed

- vllm/entrypoints/openai/serving_responses.py - Bug fix
- tests/entrypoints/openai/test_response_api_simple.py - Added unit test

### DIFF
--- a/benchmarks/fused_kernels/layernorm_rms_benchmarks.py
+++ b/benchmarks/fused_kernels/layernorm_rms_benchmarks.py
@@ -14,6 +14,9 @@ from tqdm import tqdm
 
 import vllm._custom_ops as ops
 from vllm.model_executor.layers.layernorm import RMSNorm
+from vllm.model_executor.layers.quantization.utils.fp8_utils import (
+    per_token_group_quant_fp8,
+)
 
 
 @dataclass
@@ -22,6 +25,7 @@ class bench_params_t:
     hidden_size: int
     add_residual: bool
     dtype: torch.dtype
+    group_size: list[int]
 
     def description(self):
         return (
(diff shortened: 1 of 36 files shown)


chaunceyjiang (commented):
Could you provide an example of a response returned by the OpenAI online service?

PR #30242, opened 2025-12-08
[BugFix] skip language model in Encoder

Bounty-hunter:
## Purpose
As describe in https://github.com/vllm-project/vllm/issues/30231, our solution follows the approach used for pooling models in `adapters.py`

The model adaptation as follows:

(1) implement the class method get_language_model_spec for vl model.
(2) launch with --convert 'vision_only'

## Usage
```
CUDA_VISIBLE_DEVICES="$GPU_E" vllm serve "$MODEL" \
    --gpu-memory-utilization 0.01 \
    --port "$ENCODE_PORT" \
    --enforce-eager \
    --enable-request-id-headers \
    --no-enable-prefix-caching \
    --max-num-batched-tokens 114688 \
    --max-num-seqs 128 \
    --conver "vision_only" \
    --allowed-local-media-path ${GIT_ROOT}/tests/v1/ec_connector/integration \
    --ec-transfer-config '{
        "ec_connector": "ECExampleConnector",
        "ec_role": "ec_producer",
        "ec_connector_extra_config": {
            "shared_storage_path": "'"$EC_SHARED_STORAGE_PATH"'"
        }
    }' \
    >"${ENC_LOG}" 2>&1 &
```

## Test Plan
Run:
```
vllm/examples/online_serving/disaggregated_encoder/disagg_1e1pd_example.sh
```

### Qwen2.5 vl

Check the result:

```
{"id":"chatcmpl-544bbd22-f1ea-41eb-98ec-f2273aa521b6","object":"chat.completion","created":1765963033,"model":"Qwen/Qwen2.5-VL-3B-Instruct","choices":[{"index":0,"message":{"role":"assistant","content":"The image shows a street scene with several elements:\n\n1. **Foreground**: There are three pigeons on the ground, one of which is prominently in the foreground.\n2. **Background**: Two people are walking away from the camera along a sidewalk. They are dressed in formal attire, suggesting they might be business professionals or students.\n3. **Sidewalk**: The sidewalk is made of small stones and has some fallen leaves scattered around.\n4. **Wall**: A brick wall runs parallel to the sidewalk.\n5. **Trees**: There are trees with green and yellow leaves, indicating it might be autumn.\n6. **Buildings**: In the background, there are buildings, including a tall structure that appears to be a residential building.\n\nThe overall setting seems to be an urban area during the fall season.","refusal":null,"annotations":null,"audio":null,"function_call":null,"tool_calls":[],"reasoning":null,"reasoning_content":null},"logprobs":null,"finish_reason":"stop","stop_reason":null,"token_ids":null}],"service_tier":null,"system_fingerprint":null,"usage":{"prompt_tokens":8057,"total_tokens":8220,"completion_tokens":163,"prompt_tokens_details":null},"prompt_logprobs":null,"prompt_token_ids":null,"kv_transfer_params":null}
```

Check the device mem usage:
```
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 530.30.02              Driver Version: 530.30.02    CUDA Version: 12.1     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                  Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf            Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA A100-SXM4-80GB           Off| 00000000:07:00.0 Off |                    0 |
| N/A   39C    P0               67W / 400W|   3708MiB / 81920MiB |      0%      Default |
|                                         |                      |             Disabled |
+-----------------------------------------+----------------------+----------------------+
|   1  NVIDIA A100-SXM4-80GB           Off| 00000000:0A:00.0 Off |                    0 |
| N/A   33C    P0               65W / 400W|  44766MiB / 81920MiB |      0%      Default |
|                                         |                      |             Disabled |

```

### qwen3 vl
Check the result:
```
{"id":"chatcmpl-be0c8768-f0cb-49ba-836f-6c3e7754e89c","object":"chat.completion","created":1765965713,"model":"Qwen/Qwen3-VL-8B-Instruct","choices":[{"index":0,"message":{"role":"assistant","content":"This image captures a quiet, autumnal street scene, likely in an urban area. Here's a breakdown of what's visible:\n\n*   **Foreground:** A pigeon is the main subject, positioned prominently in the foreground on a gravel path. Its feathers are a mix of grey and iridescent purple/blue, and it appears to be walking away from the camera. The camera's shallow depth of field keeps this bird in sharp focus while blurring the rest of the scene.\n\n*   **Midground:** The path continues into the distance, lined with more pigeons walking away from the viewer. Scattered yellow leaves, likely from ginkgo trees, are on the ground.\n\n*   **Background:** Two people in dark coats are walking away down the path, also slightly out of focus. The trees lining the path have leaves that are turning yellow, indicating it is autumn. Buildings are visible in the distance.\n\n*   **Right Side:** A prominent red brick wall runs along the right edge of the path.\n\nThe overall mood is calm and slightly melancholic, enhanced by the autumn colors and the low-angle perspective. The image uses motion blur and selective focus to draw the viewer's attention to the pigeon in the foreground.","refusal":null,"annotations":null,"audio":null,"function_call":null,"tool_calls":[],"reasoning":null,"reasoning_content":null},"logprobs":null,"finish_reason":"stop","stop_reason":null,"token_ids":null}],"service_tier":null,"system_fingerprint":null,"usage":{"prompt_tokens":6171,"total_tokens":6417,"completion_tokens":246,"prompt_tokens_details":null},"prompt_logprobs":null,"prompt_token_ids":null,"kv_transfer_params":null}
```
Check the device mem usage:

```
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 530.30.02              Driver Version: 530.30.02    CUDA Version: 12.1     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                  Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf            Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA A100-SXM4-80GB           Off| 00000000:07:00.0 Off |                    0 |
| N/A   36C    P0               66W / 400W|   5255MiB / 81920MiB |      0%      Default |
|                                         |                      |             Disabled |
+-----------------------------------------+----------------------+----------------------+
|   1  NVIDIA A100-SXM4-80GB           Off| 00000000:0A:00.0 Off |                    0 |
| N/A   33C    P0               65W / 400W|  64161MiB / 81920MiB |      0%      Default |
|                                         |                      |             Disabled |

```

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/models/adapters.py
+++ b/vllm/model_executor/models/adapters.py
@@ -520,3 +520,64 @@ def seq_cls_model_loader(model, weights: Iterable[tuple[str, torch.Tensor]]):
     method = getattr(text_config, "method", None)
     assert method in SEQ_CLS_LOAD_METHODS, f"method {method} not supported"
     return SEQ_CLS_LOAD_METHODS[method](model, weights)
+
+
+def as_mm_encoder_only_model(cls: _T) -> _T:
+    """
+    Subclass an existing vLLM vl model to support mm encoder only for
+    EPD encoder instances.
+    """
+    if not hasattr(cls, "embed_multimodal"):
+        # Submodel case: return the original class.
+        return cls
+
+    if not hasattr(cls, "get_language_model_spec"):
+        raise TypeError(f"{cls} need to implement `get_language_model_spec` method.")
+
+    lm_model_cls, lm_attr = cls.get_language_model_spec()
+
+    if lm_model_cls is None or lm_attr is None:
+        raise TypeError(
+            f"{cls}.get_language_model_spec() must return (lm_model_cls, lm_attr)"
+        )
+
+    class DummyLM(nn.Module):
+        def __init__(self, *args, **kwargs):
+            self.make_empty_intermediate_tensors = None
+
+    class ModelForMMEncoderOnly(cls):
+        def __init__(
+            self,
+            *,
+            vllm_config: "VllmConfig",
+            prefix: str = "",
+            **kwargs: Any,
+        ) -> None:
+            self.is_mm_encoder_only_model = True
+            origin_init = lm_model_cls.__init__
+            try:
+                lm_model_cls.__init__ = DummyLM.__init__
+                super().__init__(vllm_config=vllm_config, prefix=prefix, **kwargs)
+
+                if hasattr(self, lm_attr):
+                    delattr(self, lm_attr)
+            finally:
+                lm_model_cls.__init__ = origin_init
+
+        def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
+            from .utils import AutoWeightsLoader
+
+            origin_init_ = AutoWeightsLoader.__init__
+
+            def _new_init_(self, *args, **kwargs):
+                origin_init_(self, *args, **kwargs)
+                self.skip_prefixes = (self.skip_prefixes or []) + [f"{lm_attr}."]
+
+            try:
+                AutoWeightsLoader.__init__ = _new_init_
+                result = super().load_weights(weights)
+            finally:
+                AutoWeightsLoader.__init__ = origin_init_
+            return result
+
+    return ModelForMMEncoderOnly  # type: ignore
--- a/vllm/model_executor/models/interfaces.py
+++ b/vllm/model_executor/models/interfaces.py
@@ -141,6 +141,14 @@ class SupportsMultiModal(Protocol):
         """
         ...
 
+    @classmethod
+    def get_language_model_spec(cls) -> tuple[nn.Module | None, str | None]:
+        """
+        Return the language model spec:
+        (language model class, language model attr)
+        """
+        return None, None
+
     @overload
     def embed_input_ids(self, input_ids: Tensor) -> Tensor: ...
 
@@ -300,6 +308,10 @@ def supports_multimodal_raw_input_only(model: type[object] | object) -> bool:
 
 def supports_multimodal_encoder_tp_data(model: type[object] | object) -> bool:
     return getattr(model, "supports_encoder_tp_data", False)
+
+
+def supports_mm_encoder_only(model: type[object] | object) -> bool:
+    return getattr(model, "is_mm_encoder_only_model", False)
 
 
 @overload
--- a/vllm/model_executor/models/qwen2_5_vl.py
+++ b/vllm/model_executor/models/qwen2_5_vl.py
@@ -34,7 +34,7 @@ import einops
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
-from transformers import BatchFeature
+from transformers import BatchFeature, Qwen2ForCausalLM
 from transformers.models.qwen2_5_vl import Qwen2_5_VLProcessor
 from transformers.models.qwen2_5_vl.configuration_qwen2_5_vl import (
     Qwen2_5_VLConfig,
@@ -1567,3 +1567,11 @@ class Qwen2_5_VLForConditionalGeneration(
             connector="visual.merger.",
             tower_model="visual.",
         )
+
+    @classmethod
+    def get_language_model_spec(cls) -> tuple[nn.Module | None, str | None]:
+        """
+        Return the language model spec:
+        (language model class, language model attr)
+        """
+        return Qwen2ForCausalLM, "language_model"
--- a/vllm/v1/worker/gpu_model_runner.py
+++ b/vllm/v1/worker/gpu_model_runner.py
@@ -66,6 +66,7 @@ from vllm.model_executor.models.interfaces import (
     SupportsXDRoPE,
     is_mixture_of_experts,
     supports_eagle3,
+    supports_mm_encoder_only,
     supports_mrope,
     supports_multimodal_pruning,
     supports_transcription,
@@ -4067,6 +4068,11 @@ class GPUModelRunner(
             remove_lora: If False, dummy LoRAs are not destroyed after the run
             activate_lora: If False, dummy_run is performed without LoRAs.
         """
+        if supports_mm_encoder_only(self.model):
+            # The current dummy run only covers LM execution, so we can skip it.
+            # mm encoder dummy run may need to add in the future.
+            return torch.tensor([]), torch.tensor([])
+
         assert (
             cudagraph_runtime_mode is None
             or cudagraph_runtime_mode.valid_runtime_modes()
@@ -4344,6 +4350,11 @@ class GPUModelRunner(
         # The dummy hidden states may contain special values,
         # like `inf` or `nan`.
         # To avoid breaking the sampler, we use a random tensor here instead.
+
+        if supports_mm_encoder_only(self.model):
+            # MM Encoder only model no need to run sampler.
+            return torch.tensor([])
+
         hidden_states = torch.rand_like(hidden_states)
 
         logits = self.model.compute_logits(hidden_states)
@@ -4472,6 +4483,10 @@ class GPUModelRunner(
         self,
         hidden_states: torch.Tensor,
     ) -> PoolerOutput:
+        if supports_mm_encoder_only(self.model):
+            # MM Encoder only model not need to run pooler.
+            return torch.tensor([])
+
         # Find the task that has the largest output for subsequent steps
         supported_pooling_tasks = self.get_supported_pooling_tasks()
 
--- a/examples/online_serving/disaggregated_encoder/README.md
+++ b/examples/online_serving/disaggregated_encoder/README.md
@@ -38,6 +38,8 @@ Encoder engines should be launched with the following flags:
 
 - `--max-num-batched-tokens=<large value>` **(default: 2048)** – This flag controls the token scheduling budget per decoding step and is irrelevant to encoder-only instances. **Set it to a very high value (effectively unlimited) to bypass scheduler limitations.** The actual token budget is managed by the encoder cache manager.
 
+- `--convert "mm_encoder_only"` **(Optional)** - The language model is skipped during initialization to reduce device memory usage. **Models using this option must implement the `get_language_model_spec` interface.**
+
 ## Local media inputs
 
 To support local image inputs (from your ```MEDIA_PATH``` directory), add the following flag to the encoder instance:
--- a/vllm/config/model.py
+++ b/vllm/config/model.py
@@ -71,7 +71,7 @@ else:
 logger = init_logger(__name__)
 
 RunnerOption = Literal["auto", RunnerType]
-ConvertType = Literal["none", "embed", "classify", "reward"]
+ConvertType = Literal["none", "embed", "classify", "reward", "mm_encoder_only"]
 ConvertOption = Literal["auto", ConvertType]
 TokenizerMode = Literal["auto", "hf", "slow", "mistral", "deepseek_v32"]
 ModelDType = Literal["auto", "half", "float16", "bfloat16", "float", "float32"]
--- a/vllm/model_executor/model_loader/utils.py
+++ b/vllm/model_executor/model_loader/utils.py
@@ -189,7 +189,9 @@ def _get_model_architecture(model_config: ModelConfig) -> tuple[type[nn.Module],
             )
 
     convert_type = model_config.convert_type
-    if convert_type != "none" and supports_multimodal(model_cls):
+    if convert_type not in ["none", "mm_encoder_only"] and supports_multimodal(
+        model_cls
+    ):
         logger.debug_once("Detected conversion of Multi Modal model.")
         converted = try_create_mm_pooling_model_cls(model_cls)
         if converted is not None:
@@ -200,6 +202,11 @@ def _get_model_architecture(model_config: ModelConfig) -> tuple[type[nn.Module],
 
     if convert_type == "none":
         pass
+    elif convert_type == "mm_encoder_only":
+        logger.debug_once("Converting to mm encoder only model.")
+        from vllm.model_executor.models.adapters import as_mm_encoder_only_model
+
+        model_cls = as_mm_encoder_only_model(model_cls)
     elif convert_type == "embed":
         logger.debug_once("Converting to embedding model.")
         model_cls = as_embedding_model(model_cls)
--- a/vllm/model_executor/models/qwen3_vl.py
+++ b/vllm/model_executor/models/qwen3_vl.py
@@ -2090,3 +2090,11 @@ class Qwen3VLForConditionalGeneration(
             connector="visual.merger",
             tower_model="visual.",
         )
+
+    @classmethod
+    def get_language_model_spec(cls) -> tuple[nn.Module | None, str | None]:
+        """
+        Return the language model spec:
+        (language model class, language model attr)
+        """
+        return Qwen3LLMForCausalLM, "language_model"


vllm/model_executor/models/utils.py
```diff
@@ -107,6 +108,26 @@ def apply_dict(self, values: dict[str, Any]) -> dict[str, Any]:
...
+def maybe_init_language_model(init_fn):
+    if has_ec_transfer() and get_ec_transfer().is_producer:
+        return None
+    return init_fn()
+
+
+# Skiped language model prefix
+def maybe_skip_language_model_prefix(
+    module: nn.Module, skip_prefixes: list[str]
+) -> str:
```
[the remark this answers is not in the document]
Bounty-hunter: resolved

vllm/model_executor/models/utils.py
```diff
@@ -107,6 +108,28 @@ def apply_dict(self, values: dict[str, Any]) -> dict[str, Any]:
         }
 
 
+# Skip language model in Encoder instance
+def maybe_init_language_model(init_fn):
+    if has_ec_transfer() and get_ec_transfer().is_producer:
+        return None
```
NickLucche: we may also just want to add a cli option or better an env var to avoid loading the LM. This could be at the very least useful for testing.

NickLucche (commented):
@DarkLight1337 @ywang96  have more context on the undergoing efforts to generalize this piece, so we definitely need their feedback.

From my own perspective, this change looks reasonable, thanks for the work @fake0fan !

DarkLight1337 (commented):
Possibly stupid question: Can we simply set the architecture name to the LM backbone via `--hf-overrides`?

Bounty-hunter:
> Possibly stupid question: Can we simply set the architecture name to the LM backbone via `--hf-overrides`?

This approach also introduces additional modifications：
First, we need to register the vision model in the registry.py — for example, the vision model in Qwen2.5-VL-7B-Instruct is  Qwen2_5_VisionTransformer.

Second, simply modifying the architecture via --hf-overrides is not enough. The vision model Qwen2_5_VisionTransformer still cannot correctly parse the config.json of Qwen2.5-VL-7B-Instruct, because its related information is located under config.vision_config

```
"use_cache": true,
  "use_sliding_window": false,
  "vision_config": {
    "depth": 32,
    "hidden_act": "silu",
    "hidden_size": 1280,
    "intermediate_size": 3420,
    "num_heads": 16,
    "in_chans": 3,
    "out_hidden_size": 3584,
    "patch_size": 14,
    "spatial_merge_size": 2,
    "spatial_patch_size": 14,
    "window_size": 112,
    "fullatt_block_indexes": [

```


Finally, model.safetensors.index.json also needs to be adapted accordingly.

```
    "visual.blocks.8.attn.qkv.weight": "model-00001-of-00005.safetensors",
    "visual.blocks.8.mlp.down_proj.bias": "model-00001-of-00005.safetensors",
    "visual.blocks.8.mlp.down_proj.weight": "model-00001-of-00005.safetensors",
    "visual.blocks.8.mlp.gate_proj.bias": "model-00001-of-00005.safetensors",
    "visual.blocks.8.mlp.gate_proj.weight": "model-00001-of-00005.safetensors",
    "visual.blocks.8.mlp.up_proj.bias": "model-00001-of-00005.safetensors",
    "visual.blocks.8.mlp.up_proj.weight": "model-00001-of-00005.safetensors",
    "visual.blocks.8.norm1.weight": "model-00001-of-00005.safetensors",
    "visual.blocks.8.norm2.weight": "model-00001-of-00005.safetensors",
```

DarkLight1337:
I see. In order to avoid having to modify each model individually, can we use the adapter approach (`adapters.py`) similar to pooling models? Then the user only needs to pass `--convert lm_only` and the adapter can handle the automatic conversion.

Bounty-hunter:
> I see. In order to avoid having to modify each model individually, can we use the adapter approach (`adapters.py`) similar to pooling models? Then the user only needs to pass `--convert lm_only` and the adapter can handle the automatic conversio

i will try it

vllm/model_executor/models/adapters.py
```diff
@@ -520,3 +520,49 @@ def seq_cls_model_loader(model, weights: Iterable[tuple[str, torch.Tensor]]):
...
     return SEQ_CLS_LOAD_METHODS[method](model, weights)
+
+
+def as_vision_only_model(cls: _T) -> _T:
+    """
+    Subclass an existing vLLM vl model to support vision-only for
+    EPD encoder instances.
+    """
+
+    if not hasattr(cls, "get_multimodal_embeddings"):
```
DarkLight1337: No need to soft ignore, we should raise an error explicitly in this case

vllm/model_executor/models/interfaces.py
```diff
@@ -141,6 +141,14 @@ def get_language_model(self) -> VllmModel:
         """
         ...
 
+    @classmethod
+    def get_model_spec(cls) -> tuple[nn.module | None, str | None]:
```
DarkLight1337:
```suggestion
    def get_model_spec(cls) -> tuple[nn.module | None, str | None]:
```
```suggestion
    def get_model_spec(cls) -> tuple[nn.Module | None, str | None]:
```

vllm/model_executor/models/qwen2_5_vl.py
```diff
@@ -1561,3 +1561,10 @@ def get_mm_mapping(self) -> MultiModelKeys:
             connector="visual.merger.",
             tower_model="visual.",
         )
+
+    def get_model_spec(cls) -> tuple[nn.module | None, str | None, Callable | None]:
```
DarkLight1337:
```suggestion
    @classmethod
    def get_model_spec(cls) -> tuple[nn.Module | None, str | None]:
```

vllm/model_executor/models/adapters.py
```diff
@@ -520,3 +520,58 @@ def seq_cls_model_loader(model, weights: Iterable[tuple[str, torch.Tensor]]):
...
+        return cls  # only for multimodal models
+
+    if not hasattr(cls, "get_model_spec"):
+        raise TypeError(f"{cls} need to implement `get_model_spec` method.")
+
+    lm_model, lm_attr = cls.get_model_spec()
```
DarkLight1337: also check that it's not None

vllm/model_executor/models/interfaces.py
```diff
@@ -141,6 +141,14 @@ def get_language_model(self) -> VllmModel:
         """
         ...
 
+    @classmethod
+    def get_model_spec(cls) -> tuple[nn.Module | None, str | None]:
```
DarkLight1337: Call this `get_language_model_spec`?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -4004,6 +4004,11 @@ def _dummy_run(
             remove_lora: If False, dummy LoRAs are not destroyed after the run
             activate_lora: If False, dummy_run is performed without LoRAs.
         """
+        if getattr(self.model, "is_vision_only_model", False):
```
DarkLight1337: Let's add a function to `interfaces.py` so we can avoid having `getattr` in multiple places

vllm/model_executor/models/adapters.py
Bounty-hunter: submodel also reach hear, e.g., when qwen detect "Qwen2ForCausalLM"
```
self.language_model = init_vllm_registered_model(
                vllm_config=vllm_config,
                prefix=maybe_prefix(prefix, "language_model"),
                architectures=["Qwen2ForCausalLM"],
            )
```

vllm/model_executor/models/qwen2_5_vl.py
Bounty-hunter: modified

vllm/model_executor/models/adapters.py
DarkLight1337: Got it, thanks for explaining

DarkLight1337 (commented):
cc @NickLucche I think it makes sense to just skip the LM part in profile run / dummy run of Encoder process regardless of `supports_vision_only`

vllm/model_executor/models/utils.py
Bounty-hunter: we add --convert "vision_only" to identify whether to load lm weight.

vllm/model_executor/models/interfaces.py
Bounty-hunter: modified,

vllm/model_executor/models/utils.py
DarkLight1337: Maybe we can auto-select the converter if `ec_transfer_config` is provided

Bounty-hunter: I think @NickLucche 's intention is to preserve the scenario where LM weights are loaded, so we need this param.

DarkLight1337: Alright let's keep this then

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -4272,6 +4278,11 @@ def _dummy_sampler_run(
         # The dummy hidden states may contain special values,
         # like `inf` or `nan`.
         # To avoid breaking the sampler, we use a random tensor here instead.
+
```
DarkLight1337, line 4353: It would be better to avoid calling this entirely by modifying `profile_run` to return early so that `_dummy_run`, `_dummy_pooler_run` and `_dummy_sampler_run` are all skipped

vllm/model_executor/models/qwen2_5_vl.py
```diff
@@ -1567,3 +1561,11 @@ def get_mm_mapping(self) -> MultiModelKeys:
...
             tower_model="visual.",
         )
+
+    @classmethod
+    def get_language_model_spec(cls) -> tuple[nn.Module | None, str | None]:
+        """
+        Return the language model spec:
+        (language model class, language model attr)
+        """
+        return Qwen2ForCausalLM, "language_model"
```
NickLucche, line 1577: I think this is fine, but just curious how many MM models are *not* using `language_model` as attr name? I feel this would be a simple rule to standardize.

In that case, we could have `get_language_model_spec` just return the class type and we could do something like
``` 
        self.language_model = init_vllm_registered_model(
            vllm_config=vllm_config,
            prefix=maybe_prefix(prefix, "language_model"),
            architectures=[self.get_language_model_spec()],
        )
```
or even wrap this boilerplate into another function, in order to make the integration of a new model a bit more guided.

vllm/model_executor/models/adapters.py
```diff
@@ -520,3 +520,67 @@ def seq_cls_model_loader(model, weights: Iterable[tuple[str, torch.Tensor]]):
...
+        raise TypeError(f"{cls} need to implement `get_language_model_spec` method.")
+
+    lm_model, lm_attr = cls.get_language_model_spec()
+
+    if lm_model is None or lm_attr is None:
+        raise TypeError(
+            f"{cls}.get_language_model_spec() must return (lm_model, lm_attr)"
+        )
+
+    class FakeLM(nn.Module):
```
NickLucche: nit: DummyLM?

vllm/model_executor/models/adapters.py
```diff
@@ -520,3 +520,67 @@ def seq_cls_model_loader(model, weights: Iterable[tuple[str, torch.Tensor]]):
...
+            prefix: str = "",
+            **kwargs: Any,
+        ) -> None:
+            self.is_vision_only_model = True
+            origin_init = lm_model.__init__
+            lm_model.__init__ = FakeLM.__init__
+
+            super().__init__(vllm_config=vllm_config, prefix=prefix, **kwargs)
+
+            lm_model.__init__ = origin_init
```
NickLucche: should the old init be kept around? I wonder whether it can interact in some unintended way with features that refresh the deployment (reload weights, sleep mode..)

vllm/v1/worker/gpu_model_runner.py
NickLucche: agree!

NickLucche (commented):
This approach is much cleaner thanks @Bounty-hunter and @DarkLight1337 !

I only have one remark regarding the `VisionOnly` naming we're using: I think our specific purpose in this PR is to enable EPD deployments, therefore the encoder(s) we're isolating could be an audio tower, a vision tower or even both (eg gemma3n).
What about using `EncoderOnly` or something similar, so that it's clear the model must not be a ViT but a more generic MMEncoder?

> I think it makes sense to just skip the LM part in profile run / dummy run of Encoder process regardless of supports_vision_only

@DarkLight1337 I agree, as long as the encoder part is profiled and compiled, we can discard the rest. 
I also think a lot of the init args we pass to vllm will just be noop for now (eg CG, max_model_len), while more importantly we should disable some other overhead when this mode is detected (prefix-caching for one).

vllm/v1/worker/gpu_model_runner.py
Bounty-hunter: If we skip it at the upper layer, we also need to modify compile_or_warm_up_model and profile_run accordingly. In addition, dummy_run also needs to be updated, considering the DP (data parallel) scenario.  Which way do you think would be better?

DarkLight1337: Ok in that case I'm fine with putting this logic inside the individual methods

vllm/model_executor/models/qwen2_5_vl.py
Bounty-hunter: There are still some existing models that use different attribute names, so I think it’s better to keep returning the attribute for better compatibility.
image
image

vllm/model_executor/models/adapters.py
Bounty-hunter: Acturally, the vl model become ModelForVisionOnly， and the self.language_model att will be deleted (The model never contain the lm model info), so it ok to be refresh the deployment


```
lm_model.__init__ = origin_init
if hasattr(self, lm_attr):
    delattr(self, lm_attr)
```

Bounty-hunter:
> This approach is much cleaner thanks @Bounty-hunter and @DarkLight1337 !
> 
> I only have one remark regarding the `VisionOnly` naming we're using: I think our specific purpose in this PR is to enable EPD deployments, therefore the encoder(s) we're isolating could be an audio tower, a vision tower or even both (eg gemma3n). What about using `EncoderOnly` or something similar, so that it's clear the model must not be a ViT but a more generic MMEncoder?
> 
> > I think it makes sense to just skip the LM part in profile run / dummy run of Encoder process regardless of supports_vision_only
> 
> @DarkLight1337 I agree, as long as the encoder part is profiled and compiled, we can discard the rest. I also think a lot of the init args we pass to vllm will just be noop for now (eg CG, max_model_len), while more importantly we should disable some other overhead when this mode is detected (prefix-caching for one).

Changed to MMEncoderOnly

vllm/model_executor/models/adapters.py
Bounty-hunter: changed to DummyLM

vllm/model_executor/models/adapters.py
DarkLight1337: Maybe we can add `try`-f`inally` block to ensure `lm_model.__init__` is always reset even if an error occurs

DarkLight1337: Also we can rename `lm_model` to `lm_model_cls` to improve readability

DarkLight1337: Btw I just found `init_on_device_without_buffers` from Transformers backend which lets you initialize the model without allocating memory. Maybe we can use that to avoid patching the LM backbone

Bounty-hunter: > init_on_device_without_buffers
init_on_device_without_buffers only skip buffer, but parameter still need loaded. I think hook in higher level would be more clearer.

DarkLight1337: We can use `"meta"` device to avoid allocating memory

DarkLight1337: https://docs.pytorch.org/docs/stable/meta.html

DarkLight1337: As discussed offline, that would make the logic too complicated as we still need to distinguish between LM and MM modules, so let's just keep this

Bounty-hunter: > Also we can rename `lm_model` to `lm_model_cls` to improve readability

modified

NickLucche: nice optimization with meta here

DarkLight1337 (approved):
Thanks, LGTM now

NickLucche (approved):
LGTM thank you @Bounty-hunter !

PR #27883, opened 2025-10-31
[Performance] Fused blockwise quant RMS norm

ElizaWszola:
CUDA kernel and fusion code for Fused Groupwise FP8-Quantized RMS Norm. This code allows to fuse RMS Norm + FP8 Quantization of the RMS Norm's output when `enable_fusion==True`.

#### Test fused op
```pytest tests/kernels/core/test_fused_quant_layernorm.py```

#### Test fusion
```pytest tests/compile/test_fusion.py```

(tested with both `VLLM_USE_DEEP_GEMM=1` and `VLLM_USE_DEEP_GEMM=0`)

#### Offline inference
Run with
```
llm = LLM(model="Qwen/Qwen3-30B-A3B-FP8",
        compilation_config=CompilationConfig(
            pass_config=PassConfig(
                enable_fusion=True,
                enable_noop=True)))
```

(tested with both `VLLM_USE_DEEP_GEMM=1` and `VLLM_USE_DEEP_GEMM=0`, verified that the fused kernel is being produced)

#### Microbenchmark isolated op:
```python benchmarks/fused_kernels/layernorm_rms_benchmarks.py```

<details>
<summary>Results on H100 (click to show)</summary>

```text
[-------------------------------------------- rms-norm-dynamic-per-token-quant --------------------------------------------]
                                                                  |  unfused_groupwise_fp8_impl  |  fused_groupwise_fp8_impl
1 threads: -----------------------------------------------------------------------------------------------------------------
      N 1 x D 1024 x R True x DT torch.bfloat16x GS [1, 64]       |             36.2             |            24.7          
      N 1 x D 1024 x R True x DT torch.bfloat16x GS [1, 128]      |             36.1             |            25.1          
      N 1 x D 1024 x R True x DT torch.float32x GS [1, 64]        |             35.9             |            24.7          
      N 1 x D 1024 x R True x DT torch.float32x GS [1, 128]       |             34.9             |            23.7          
      N 1 x D 1024 x R False x DT torch.bfloat16x GS [1, 64]      |             39.2             |            24.9          
      N 1 x D 1024 x R False x DT torch.bfloat16x GS [1, 128]     |             39.5             |            24.5          
      N 1 x D 1024 x R False x DT torch.float32x GS [1, 64]       |             38.2             |            24.2          
      N 1 x D 1024 x R False x DT torch.float32x GS [1, 128]      |             38.7             |            24.9          
[...]

PR #30244, opened 2025-12-08
[Bugfix] Fix fusion for VL models

ElizaWszola:
A fix for #27883 so the fusion code doesn't break VL models.

### Testing:
All tests have been run on Hopper GPU with both `VLLM_USE_DEEP_GEMM=0` and `VLLM_USE_DEEP_GEMM=1`. Note that DeepGemm runs currently require changes from https://github.com/vllm-project/vllm/pull/30336 to run.

#### E2E tests:
- `Qwen/Qwen3-30B-A3B-FP8`
- `Qwen/Qwen3-VL-4B-Instruct`
- `Qwen/Qwen3-VL-2B-Instruct-FP8`
- `Qwen/Qwen3-VL-30B-A3B-Instruct-FP8`

All FP8 models have been manually verified to produce the fused group quant rms norm kernel during compilation.

#### Unit test:
`tests/compile/test_fusion.py`


ProExpertProg (commented):
Can we instead just create multiple patterns, one for column scales and one for row scales?

ElizaWszola:
@cjackal Looking into this now. In the meantime, you can also try disabling the group quant rms norm fusion altogether: https://github.com/vllm-project/vllm/pull/30273

ElizaWszola:
@cjackal This looks like an unrelated issue, can you try applying changes from this PR on the top of the current one? https://github.com/vllm-project/vllm/pull/30336

ElizaWszola:
@cjackal I've been seeing the same error, thanks for identifying the root PR!

cjackal:
> @cjackal I've been seeing the same error, thanks for identifying the root PR!
> 
> It looks like I'm able to run the model with deepgemm when I replace `m.weight_scale` with `m.weight_scale_inv` in `_extract_data_from_linear_base_module()` in `deep_gemm_warmup.py`, but I'm not sure atp if this is a safe solution to the problem

Fortunately there's ongoing bugfix at #30399 :rocket:

tests/compile/distributed/test_fusions_e2e.py
```diff
@@ -474,3 +476,79 @@ def run_model(compile_config: int | CompilationConfig, model: str, **model_kwarg
...
+CUSTOM_OPS_QUANT_RMS_NORM = ["+quant_fp8,+rms_norm"]
+
+
+@pytest.mark.parametrize(
+    "model_name, model_kwargs, backend, matches, custom_ops",
+    # Test rms norm+group quant_fp8 fusion
+    list[tuple[Any, ...]](flat_product(MODELS_GROUP_FP8, CUSTOM_OPS_QUANT_RMS_NORM)),
+)
+@pytest.mark.parametrize("inductor_graph_partition", [True, False])
+def test_rms_group_quant(
```
ProExpertProg, line 502: We don't need a new test, just enable rmsnorm+quant fusion in the other tests!

ProExpertProg (approved):

tests/compile/distributed/test_fusions_e2e.py
ElizaWszola: My reasoning was to test a block-quant model separately, so I don't have to figure out the number of fusions for all models (with the current regex-based counting, block and non-blocks quant rms fusions are counted together), and also so I don't have to make the code inside tests more complicated (block quant rms fusions are supported only when fp8 is enabled).

yewentao256 (approved):
LGTM, thanks for the work!

tests/compile/distributed/test_fusions_e2e.py
ProExpertProg: We should address this in a follow up also to reduce the total test running times

PR #29124, opened 2025-11-20
[AITER] [ROCm] Fix crash when loading llama4 model with old aiter version installed, fallback to forward_native implementation

xli:
## Purpose

Catch importing aiter.ops.sampling failure and fallback to forward_native implementation. So that running vllm without aiter sampling on AMD works.

## Test Plan

In my local test env, aiter old version is installed, started the server with llama4 model

## Test Result

In the server start up log, will get "aiter.ops.sampling is not available on ROCm..." warning log instead of crash:

```
INFO 11-20 14:02:11 [parallel_state.py:1425] rank 3 in world size 8 is assigned as DP rank 0, PP rank 0, PCP rank 0, TP rank 3, EP rank 3
INFO 11-20 14:02:11 [parallel_state.py:1425] rank 2 in world size 8 is assigned as DP rank 0, PP rank 0, PCP rank 0, TP rank 2, EP rank 2
WARNING 11-20 14:02:12 [topk_topp_sampler.py:72] aiter.ops.sampling is not available on ROCm. Falling back to forward_native implementation.
WARNING 11-20 14:02:12 [topk_topp_sampler.py:72] aiter.ops.sampling is not available on ROCm. Falling back to forward_native implementation.
WARNING 11-20 14:02:12 [topk_topp_sampler.py:72] aiter.ops.sampling is not available on ROCm. Falling back to forward_native implementation.
WARNING 11-20 14:02:12 [topk_topp_sampler.py:72] aiter.ops.sampling is not available on ROCm. Falling back to forward_native implementation.
WARNING 11-20 14:02:12 [topk_topp_sampler.py:72] aiter.ops.sampling is not available on ROCm. Falling back to forward_native implementation.
WARNING 11-20 14:02:12 [topk_topp_sampler.py:72] aiter.ops.sampling is not available on ROCm. Falling back to forward_native implementation.
WARNING 11-20 14:02:12 [topk_topp_sampler.py:72] aiter.ops.sampling is not available on ROCm. Falling back to forward_native implementation.
WARNING 11-20 14:02:12 [topk_topp_sampler.py:72] aiter.ops.sampling is not available on ROCm. Falling back to forward_native implementation.
^[[0;36m(Worker_TP0 pid=3978)^[[0;0m INFO 11-20 14:02:16 [gpu_model_runner.py:3260] Starting to load model /tmp/checkpoints...
[...]

PR #30255, opened 2025-12-08
[ROCm] Use aiter.topk_sigmoid in llama4

tpopp:
This fuses a set of kernels that are bound by launch latency and is used whenever aiter is available.

Test Plan:
Server Launch:
`VLLM_ROCM_USE_AITER_MOE=1 VLLM_ROCM_USE_AITER=1 vllm serve ${MODEL}  --port ${PORT}   --swap-space 32  --max-model-len ${MAX_MODEL_LEN}    --tensor-parallel-size ${TP}   --max-num-seqs ${MAX_NUM_SEQS}   --gpu-memory-utilization 0.93 --kv-cache-dtype fp8   --max-num-batched-tokens ${MAX_NUM_BATCHED_TOKENS} --compilation-config "{\"custom_ops\": [\"-rms_norm\", \"-quant_fp8\", \"-silu_and_mul\"]  }"     --no-enable-prefix-caching     --async-scheduling`

Example benchmark command:
`vllm bench serve --model meta-llama/Llama-4-Maverick-17B-128E-Instruct-FP8 --host localhost --dataset-name random --random-input-len 1024 --random-output-len 1024 --max-concurrency 4 --num-prompts 48 --ignore-eos`

Example correctness check:
`lm_eval   --model local-completions   --tasks gsm8k   --model_args base_url=http://0.0.0.0:${PORT}/v1/completions,model=meta-llama/Llama-4-Maverick-17B-128E-Instruct-FP8,tokenized_requests=False,tokenizer_backend=None,num_concurrent=128,timeout=120,max_retries=5`

Test Result (using mi325x8):

Throughput increase by about 4% and accuracy comparable/better.

Before:
```
Successful requests:                     48
Maximum request concurrency:             4
Benchmark duration (s):                  117.04
Total input tokens:                      49104
Total generated tokens:                  49152
Request throughput (req/s):              0.41
Output token throughput (tok/s):         419.97
Peak output token throughput (tok/s):    436.00
Peak concurrent requests:                8.00
Total token throughput (tok/s):          839.53
Mean TTFT (ms):                          130.65
Median TTFT (ms):                        119.85
P99 TTFT (ms):                           203.61
Mean TPOT (ms):                          9.40
Median TPOT (ms):                        9.42
P99 TPOT (ms):                           9.48
Mean ITL (ms):                           9.41
Median ITL (ms):                         9.30
P99 ITL (ms):                            13.74

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8961|±  |0.0084|
|     |       |strict-match    |     5|exact_match|↑  |0.8961|±  |0.0084|
```

After:
```
Successful requests:                     48
Maximum request concurrency:             4
Benchmark duration (s):                  113.55
Total input tokens:                      49104
Total generated tokens:                  49152
Request throughput (req/s):              0.42
Output token throughput (tok/s):         432.87
Peak output token throughput (tok/s):    448.00
Peak concurrent requests:                8.00
Total token throughput (tok/s):          865.31
Mean TTFT (ms):                          131.95
Median TTFT (ms):                        120.20
P99 TTFT (ms):                           197.02
Mean TPOT (ms):                          9.12
Median TPOT (ms):                        9.12
P99 TPOT (ms):                           9.21
Mean ITL (ms):                           9.12
Median ITL (ms):                         9.02
P99 ITL (ms):                            13.55

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8946|±  |0.0085|
|     |       |strict-match    |     5|exact_match|↑  |0.9007|±  |0.0082|
```

### DIFF
--- a/vllm/_aiter_ops.py
+++ b/vllm/_aiter_ops.py
@@ -1070,6 +1070,7 @@ class rocm_aiter_ops:
                 fake_impl=_rocm_aiter_rmsnorm_fused_dynamic_quant_fake,
                 dispatch_key=current_platform.dispatch_key,
             )
+
             direct_register_custom_op(
                 op_name="rocm_aiter_rmsnorm_fused_add_dynamic_quant",
                 op_func=_rocm_aiter_rmsnorm_fused_add_dynamic_quant_impl,
@@ -1115,6 +1116,7 @@ class rocm_aiter_ops:
                 fake_impl=_rocm_aiter_per_token_quant_fake,
                 dispatch_key=current_platform.dispatch_key,
             )
+
             if AITER_TOPK_SIGMOID_FOUND:
                 direct_register_custom_op(
                     op_name="rocm_aiter_topk_sigmoid",
--- a/vllm/_aiter_ops.py
+++ b/vllm/_aiter_ops.py
@@ -7,15 +7,10 @@ import torch
 from torch._ops import OpOverload
 
 import vllm.envs as envs
-from vllm.logger import init_logger
 from vllm.platforms import current_platform
 from vllm.utils.torch_utils import direct_register_custom_op, is_torch_equal_or_newer
 
-<<<<<<< HEAD
 _FP8_DTYPE = current_platform.fp8_dtype()
-=======
-logger = init_logger(__name__)
->>>>>>> 1b6821f1d (Gate aiter's topk_sigmoid usage on it being available)
 
 
 def is_aiter_found() -> bool:
@@ -55,16 +50,18 @@ def if_aiter_supported(func: Callable) -> Callable:
 
     return wrapper
 
+
 def is_aiter_function_found(function_name: str) -> bool:
     """Check if aiter module exists and has the specified function."""
     # First check if module exists
     if IS_AITER_FOUND:
         import aiter
+
         return hasattr(aiter, function_name)
     return False
 
-AITER_TOPK_SIGMOID_FOUND = is_aiter_function_found("topk_sigmoid")
 
+AITER_TOPK_SIGMOID_FOUND = is_aiter_function_found("topk_sigmoid")
 
 
 # Can't use dtypes.fp8 directly inside an op
@@ -766,18 +763,7 @@ def _rocm_aiter_act_mul_and_fp8_group_quant_fake(
     return x_fp8, out_bs
 
 
-try:
-    from aiter import topk_sigmoid  # noqa: F401
-
-    _rocm_aiter_topk_sigmoid_available = True
-except ImportError:
-    logger.warning_once(
-        "torch.ops.vllm.rocm_aiter_topk_sigmoid is not available. "
-        "Using generic torch functions. Update aiter for this optimization."
-    )
-    _rocm_aiter_topk_sigmoid_available = False
-
-if _rocm_aiter_topk_sigmoid_available:
+if AITER_TOPK_SIGMOID_FOUND:
 
     def _rocm_aiter_topk_sigmoid_impl(
         gating_output: torch.Tensor, topk: int
@@ -1071,24 +1057,12 @@ class rocm_aiter_ops:
                 dispatch_key=current_platform.dispatch_key,
             )
 
-<<<<<<< HEAD
             direct_register_custom_op(
                 op_name="rocm_aiter_rmsnorm_fused_dynamic_quant",
                 op_func=_rocm_aiter_rmsnorm_fused_dynamic_quant_impl,
                 fake_impl=_rocm_aiter_rmsnorm_fused_dynamic_quant_fake,
                 dispatch_key=current_platform.dispatch_key,
             )
-=======
-            if _rocm_aiter_topk_sigmoid_available:
-                direct_register_custom_op(
-                    op_name="rocm_aiter_topk_sigmoid",
-                    op_func=_rocm_aiter_topk_sigmoid_impl,
-                    mutates_args=[],
-                    fake_impl=_rocm_aiter_topk_sigmoid_fake,
-                    dispatch_key=current_platform.dispatch_key,
-                )
->>>>>>> 1b6821f1d (Gate aiter's topk_sigmoid usage on it being available)
-
             direct_register_custom_op(
                 op_name="rocm_aiter_rmsnorm_fused_add_dynamic_quant",
                 op_func=_rocm_aiter_rmsnorm_fused_add_dynamic_quant_impl,
@@ -1134,14 +1108,14 @@ class rocm_aiter_ops:
                 fake_impl=_rocm_aiter_per_token_quant_fake,
                 dispatch_key=current_platform.dispatch_key,
             )
-			if AITER_TOPK_SIGMOID_FOUND:
-				direct_register_custom_op(
-					op_name="rocm_aiter_topk_sigmoid",
-					op_func=_rocm_aiter_topk_sigmoid_impl,
-					mutates_args=[],
-					fake_impl=_rocm_aiter_topk_sigmoid_fake,
-					dispatch_key=current_platform.dispatch_key,
-				)
+            if AITER_TOPK_SIGMOID_FOUND:
+                direct_register_custom_op(
+                    op_name="rocm_aiter_topk_sigmoid",
+                    op_func=_rocm_aiter_topk_sigmoid_impl,
+                    mutates_args=[],
+                    fake_impl=_rocm_aiter_topk_sigmoid_fake,
+                    dispatch_key=current_platform.dispatch_key,
+                )
 
             _OPS_REGISTERED = True
 
@@ -1501,7 +1475,7 @@ class rocm_aiter_ops:
         assert group_size == 128, "Group size must be 128"
         return torch.ops.vllm.rocm_aiter_group_fp8_quant(input_2d, group_size)
 
-	@staticmethod
+    @staticmethod
     def topk_sigmoid(
         gating_output: torch.Tensor,
         topk: int,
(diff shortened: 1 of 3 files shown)


tjtanaa (approved):
LGTM

vllm/model_executor/models/llama4.py
```diff
@@ -70,6 +71,8 @@ def custom_routing_function(
         topk: int,
         renormalize: bool,
     ) -> tuple[torch.Tensor, torch.Tensor]:
+        if rocm_aiter_ops.is_enabled():
```
tjtanaa: NITS: Since this is related to MoE, let's change to use `rocm_aiter_ops.is_fused_moe_enabled()`

tjtanaa:
@houseroad If you are free could you take a quick look. Thank you.

vllm/model_executor/models/llama4.py
yuguo68: Meta's internal aiter version does not have topk_sigmoid operator yet, i.e., from aiter import topk_sigmoid will fail. Hence we need to guard the usage of it.

tpopp: @yuguo68 I assume Meta has a patching process to handle when it uses different dependency versions? Otherwise, how would this be handled?

yuguo68: > @yuguo68 I assume Meta has a patching process to handle when it uses different dependency versions? Otherwise, how would this be handled?

Nope, we don't have that! https://github.com/vllm-project/vllm/pull/29124/files is an example how we handled a previous breakage. Basically we check whether aiter has topk_sigmoid first.

tpopp: `rocm_aiter_ops.is_fused_moe_enabled()` is now used, and I've added conditional logic to ensure this functionality is available in the aiter module used.

vllm/_aiter_ops.py
```diff
@@ -438,6 +441,47 @@ def _rocm_aiter_rmsnorm2d_fwd_with_add_fake(
     return torch.empty_like(x), torch.empty_like(residual)
 
 
+try:
```
vllmellm: instead of try catch block , could you please use a function like `IS_AITER_FOUND = is_aiter_found()` that checks topk_sigmoid exist in `aiter` package. you can name the function to `is_aiter_topk_sigmoid _found` to create a global variable AITER_TOPK_SIGMOID_FOUND
so change `_rocm_aiter_topk_sigmoid_available` to `AITER_TOPK_SIGMOID_FOUND`

you can use a the function below:

```
def is_aiter_function_found(function_name: str) -> bool:
    """Check if aiter module exists and has the specified function."""
    # First check if module exists
    if IS_AITER_FOUND:
        import aiter
        return hasattr(aiter, function_name)
    return False

AITER_TOPK_SIGMOID_FOUND = is_aiter_function_found("topk_sigmoid")

```

vllm/_aiter_ops.py
```diff
@@ -636,6 +680,15 @@ def register_ops_once() -> None:
                 dispatch_key=current_platform.dispatch_key,
             )
 
+            if _rocm_aiter_topk_sigmoid_available:
```
vllmellm: then here would need to use if `AITER_TOPK_SIGMOID_FOUND`.
also we you need to create a `@staticmethod` in `rocm_aiter_ops` singleton class to call your registered kernel from raw `torch.ops.vllm_your_registered_kernel_name`.
kindly look into other methods in `rocm_aiter_ops` singleton class to write your entry kernel call that makes call to `torch.ops.vllm_your_registered_kernel_name`.

you would need like

```
    @staticmethod
    def topk_sigmoid(
        gating_output: torch.Tensor,
        topk: int,
        return torch.ops.vllm.rocm_aiter_topk_sigmoid(gating_output, topk)
```

vllm/model_executor/models/llama4.py
```diff
@@ -59,6 +60,13 @@
     is_pp_missing_parameter,
 )
 
+try:
```
vllmellm: this is try catch block is redundant, after the above mentioned changes in aiter_ops.py you can import the global variable `AITER_TOPK_SIGMOID_FOUND` as follow
 from vllm._aiter_ops import rocm_aiter_ops, AITER_TOPK_SIGMOID_FOUND

vllm/model_executor/models/llama4.py
```diff
@@ -70,6 +78,8 @@ def custom_routing_function(
         topk: int,
         renormalize: bool,
     ) -> tuple[torch.Tensor, torch.Tensor]:
+        if _rocm_aiter_topk_sigmoid_available and rocm_aiter_ops.is_fused_moe_enabled():
```
vllmellm: change _rocm_aiter_topk_sigmoid_available to `AITER_TOPK_SIGMOID_FOUND`

vllm/model_executor/models/llama4.py
```diff
@@ -70,6 +78,8 @@ def custom_routing_function(
...
+            return torch.ops.vllm.rocm_aiter_topk_sigmoid(gating_output, topk)
```
vllmellm: as mentioned above as well you need to create an `@staticmethod` in `rocm_aiter_ops` singelton class that the body that function calls torch.ops.vllm.rocm_aiter_topk_sigmoid(gating_output, topk). 
you may name the function to `topk_sigmoid` with the required signature then change this line to
`rocm_aiter_ops.topk_sigmoid(gating_output, topk)`

OR

an optional suggestion would be creating a dispatcher function for this custom_routing_fucntion in `vllm/vllm/model_executor/models/utils.py`  so that in future if other models would be able to add support to use aiter topk and reuse the code.
by placing a dispatcher function in `vllm/vllm/model_executor/models/utils.py`  as following:

```
from collections.abc import Callable
from vllm._aiter_ops import rocm_aiter_ops, AITER_TOPK_SIGMOID_FOUND

def torch_routing_func(
    gating_output: torch.Tensor,
    topk: int,
    router_scores, router_indices = fast_topk(gating_output, topk, dim=-1)
    router_scores = torch.sigmoid(router_scores.float())
    return (router_scores, router_indices.to(torch.int32))

def dispatch_rounting_function(
        fallback_func: Callable[[torch.Tensor, int], tuple[torch.Tensor, torch.Tensor]] | None = None
) -> Callable[[torch.Tensor, int], tuple[torch.Tensor, torch.Tensor]]:
    if AITER_TOPK_SIGMOID_FOUND and rocm_aiter_ops.is_fused_moe_enabled():
        return rocm_aiter_ops.topk_sigmoid
    if fallback_func is not None:
        return fallback_func

    # torch implementation
    return torch_routing_func
```

then in llama4.py

```
from .utils import (
    dispatch_rounting_function
)

class Llama4MoE(nn.Module):
    @staticmethod
    def custom_routing_function(
        hidden_states: torch.Tensor,
        gating_output: torch.Tensor,
        topk: int,
        renormalize: bool,
        routing_func = dispatch_rounting_function()

        return routing_func(gating_output, topk)
```

vllm/_aiter_ops.py
tpopp: Done. Thank you for the detailed examples :)

I saw some discussions previously that python developers prefer the try-catch approach, but I also prefer the hasattr form.

vllm/_aiter_ops.py
tpopp: Done. For my own knowledge, why does vLLM have these `@staticmethod`s to wrap all of the `torch` ops? Is it to make the types more easily searchable? Otherwise, it seems like unneeded abstraction.

vllm/model_executor/models/llama4.py
tpopp: Done

vllm/model_executor/models/llama4.py
tpopp: I chose a different name, with `topk_sigmoid_routing_func` in the name as `routing_func` seemed too vague. I implemented the `utils.py` based suggestion though to avoid needing to replicate this call in future places if it comes up (though, I'd like to move this to a compiler/inductor based pattern in the future when I've figured that out)

vllm/_aiter_ops.py
```diff
@@ -1020,7 +1070,6 @@ def register_ops_once() -> None:
                 fake_impl=_rocm_aiter_rmsnorm_fused_dynamic_quant_fake,
                 dispatch_key=current_platform.dispatch_key,
             )
-
```
tjtanaa, line 1023: NITs, can you remove this changes?

tpopp: Done. I also added a newline in the new code to maintain consistency.

tpopp:
@vllmellm PTAL if you have time. I've updated the inital PR description with up to date numbers and environment variables required and confirmed the new function is called with the environment variables and that nothing fails when this branch is not taken.

vllm/model_executor/models/llama4.py
```diff
@@ -70,10 +70,9 @@ def custom_routing_function(
         topk: int,
         renormalize: bool,
     ) -> tuple[torch.Tensor, torch.Tensor]:
-        router_scores, router_indices = fast_topk(gating_output, topk, dim=-1)
-        # pseudo-standard is that the router scores are floats
-        router_scores = torch.sigmoid(router_scores.float())
-        return (router_scores, router_indices.to(torch.int32))
+        topk_sigmoid_func = dispatch_topk_sigmoid_routing_func()
```
tjtanaa, line 73: @tpopp can you rebase and use the latest abstraction?

They have defined topk_sigmoid_routing function in this path.
https://github.com/vllm-project/vllm/blob/a1946570d80c1bef78063e84b097951d8e8d4e6a/vllm/model_executor/layers/fused_moe/router/fused_topk_router.py#L61

After you rebasing, can you update the lmeval score? Thanks.

Issue #30167, opened 2025-12-06
[Bug][ROCm]: `vision_embeddings` in transformers inaccurate without math SDP


AndreasKaratzas:
```
==============================
        System Info
==============================
OS                           : Ubuntu 22.04.5 LTS (x86_64)
GCC version                  : (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0
Clang version                : 20.0.0git (https://github.com/RadeonOpenCompute/llvm-project roc-7.1.0 25425 1b0eada6b0ee93e2e694c8c146d23fca90bc11c5)
CMake version                : version 3.31.10
Libc version                 : glibc-2.35

==============================
       PyTorch Info
==============================
PyTorch version              : 2.9.0a0+git1c57644
Is debug build               : False
CUDA used to build PyTorch   : N/A
ROCM used to build PyTorch   : 7.1.25424-4179531dcd

==============================
      Python Environment
==============================
Python version               : 3.12.12 (main, Oct 10 2025, 08:52:57) [GCC 11.4.0] (64-bit runtime)
Python platform              : Linux-5.15.0-161-generic-x86_64-with-glibc2.35

==============================
       CUDA / GPU Info
==============================
Is CUDA available            : True
CUDA runtime version         : Could not collect
CUDA_MODULE_LOADING set to   :
GPU models and configuration :  (gfx942:sramecc+:xnack-)
Nvidia driver version        : Could not collect
cuDNN version                : Could not collect
HIP runtime version          : 7.1.25424
MIOpen runtime version       : 3.5.1
Is XNNPACK available         : True

==============================
          CPU Info
==============================
Architecture:                            x86_64
CPU op-mode(s):                          32-bit, 64-bit
Address sizes:                           52 bits physical, 57 bits virtual
Byte Order:                              Little Endian
CPU(s):                                  256
On-line CPU(s) list:                     0-255
Vendor ID:                               AuthenticAMD
Model name:                              AMD EPYC 9575F 64-Core Processor
CPU family:                              26
Model:                                   2
Thread(s) per core:                      2
Core(s) per socket:                      64
Socket(s):                               2
Stepping:                                1
Frequency boost:                         enabled
CPU max MHz:                             5008.0068
CPU min MHz:                             1500.0000
BogoMIPS:                                6600.02
[...]

PR #30270, opened 2025-12-08
[ROCm][CI][Bugfix] Multi-Modal Model Support Fixes and Attention Backend Improvements

AndreasKaratzas:
This PR addresses several ROCm-specific issues with multi-modal/vision-language models and improves attention backend dispatching for encoder-only self-attention models. It renders green the following test groups on ROCm:
- `Multi-Modal Models Test (Standard)`
- `Multi-Modal Models Test (Extended) 1`
- `Multi-Modal Models Test (Extended) 2`
- `Multi-Modal Models Test (Extended) 3`

#### Key Changes

**Attention Backend Selection (`vllm/platforms/rocm.py`):**
- Added dtype validation (fp16/bf16 only) for Flash Attention backend selection
- Added automatic FlexAttention fallback for `ENCODER_ONLY` attention types
- Implemented `VLLM_ATTENTION_BACKEND` environment variable override mechanism for model-specific backend selection
- Improved logging messages with "V1 engine" suffix for clarity

**Qwen3-VL Support (`vllm/model_executor/models/qwen3_vl.py`):**
- Added ROCm-specific attention backend override to use `ROCM_AITER_FA` on gfx9 architectures
- Sets `VLLM_ATTENTION_BACKEND` env var to propagate backend choice to platform layer

**Transformers Multimodal Mixin (`vllm/model_executor/models/transformers/multimodal.py`):**
- Force `MATH` SDP backend for vision encoder on ROCm to avoid accuracy issues with `flash_sdp` and `mem_efficient_sdp`
- Patches issue [#30167](https://github.com/vllm-project/vllm/issues/30167)

**SigLIP2 NaViT (`vllm/model_executor/models/siglip2navit.py`):**
- Disabled `vllm_flash_attn` rotary embedding on ROCm

**Test Updates:**
- Moved `conftest.py` up one level to share ROCm SDP workarounds across multimodal tests
- Use `float16` dtype instead of `float` for CLIP/SigLIP pooling tests on ROCm
- Added ROCm-specific processor kwargs for Qwen2.5-VL tests
- Skip MiniMax-VL-01 on ROCm due to GPU OOM and pickle issues with spawn + tp>1
- Modified CI timeouts for multi-modal test suites for precision

**CI Configuration (`.buildkite/test-amd.yaml`):**
- Extended timeouts for Multi-Modal Models Test suites (Standard: 100min, Extended 2: 120min, Accuracy Eval: 240min)


vllm/platforms/rocm.py
```diff
@@ -241,26 +246,34 @@ def get_attn_backend_cls(
                 )
             if selected_backend == AttentionBackendEnum.TRITON_MLA:
                 if block_size != 1:
-                    logger.info_once("Using Triton MLA backend.")
+                    logger.info_once("Using Triton MLA backend on V1 engine.")
```
hmellor: There is no V0 anymore so mentioning V1 creates confusion

vllm/platforms/rocm.py
```diff
@@ -313,6 +326,11 @@ def get_attn_backend_cls(
                 logger.info("Using Aiter Flash Attention backend on V1 engine.")
                 return AttentionBackendEnum.ROCM_AITER_FA.get_path()
 
+            # Priority 5: If model is Encoder-only self-attention type
+            if attn_type is not None and attn_type in (AttentionType.ENCODER_ONLY):
+                logger.info("Using FlexAttention backend  on V1 engine.")
+                return AttentionBackendEnum.FLEX_ATTENTION.get_path()
```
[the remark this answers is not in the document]
AndreasKaratzas: Comment addressed in [9c9f225](https://github.com/vllm-project/vllm/pull/30270/commits/9c9f225e8e5cd9b264d78756276d2d019558e406)

vllm/platforms/rocm.py
AndreasKaratzas: Thank you for your Review Harry, let me know if I addressed your comment in [33f9bff](https://github.com/vllm-project/vllm/pull/30270/commits/33f9bff6f61ae65c1fedd4989990df5372a8e038)

hmellor: All good, thanks!

vllm/model_executor/models/qwen3_vl.py
```diff
@@ -370,6 +371,28 @@ def __init__(
...
+                # and test accuracy on ROCm with `Multi-Modal Models Test (Standard)`
+                # group.
+                from vllm._aiter_ops import IS_AITER_FOUND
+                from vllm.platforms.rocm import on_gfx9
+
+                if on_gfx9() and IS_AITER_FOUND:
+                    attn_backend_override = AttentionBackendEnum.ROCM_AITER_FA
+                    # Set also env variable for platform rocm to use ROCM_AITER_FA
+                    # for `selected_backend` in attention backend getter.
+                    os.environ["VLLM_ATTENTION_BACKEND"] = "ROCM_AITER_FA"
```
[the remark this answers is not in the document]
AndreasKaratzas: There is no concurrency that involves this part of code. This behavior is ROCm specific. This is critical for this model for now. Nonetheless, we are going to add the support necessary for the rest of the ROCm specific attention backends and make this mini-patch obsolete in the future.

vllm/transformers_utils/config.py
```diff
@@ -242,6 +242,22 @@ def get_config_parser(config_format: str) -> ConfigParserBase:
...
+    ``transformers.PretrainedConfig`` instances.
+
+    Note:
+        Duplicated from ``vllm.config.utils`` to avoid circular import
+        (vllm.transformers_utils.config <-> vllm.config.model).
+    """
+    for name in names:
+        if hasattr(obj, name):
+            return getattr(obj, name)
+    return default
```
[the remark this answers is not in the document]
AndreasKaratzas: This was a standard function. It serves very specific action and would be highly unlikely to be needed any "maintenance" or modifications at all in the future. Therefore, the duplication only serves the resolution of the circular import bug.

vllm/model_executor/models/qwen3_vl.py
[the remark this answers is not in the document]
tjtanaa: @AndreasKaratzas  vLLM has already clean up all these overriding logic in all model definition files. They are hard to track and trace.

Please perform overriding in the unit tests themselves through controlling the environment variables or using this new attention config interface to select the attention backend has changed 5 days ago in this PR https://github.com/vllm-project/vllm/pull/26315 . Please find out more details about it here.

If it is a platform specific default behaviour, the logic should be specified here

https://github.com/vllm-project/vllm/blob/d6464f267979946a1c2d9c6029ef2007be73ca09/vllm/platforms/rocm.py#L191-L206

vllm/platforms/rocm.py
```diff
@@ -259,21 +264,29 @@ def get_attn_backend_cls(
                 f"is not MLA type while requested for MLA backend."
             )
 
+        attn_backend_override = os.environ.get("VLLM_ATTENTION_BACKEND")
```
tjtanaa: We don't need to handle the attention overriding logic in this platform class. 
The overriding logic happens before `get_attn_backend_cls` is called.

There is a new attention config interface for overriding the attention to select the attention backend has changed 5 days ago in this PR https://github.com/vllm-project/vllm/pull/26315 . Please find out more details about it here.

tests/models/multimodal/pooling/test_clip.py
```diff
@@ -70,7 +72,9 @@ def _run_test(
 
 
 @pytest.mark.parametrize("model", MODELS)
-@pytest.mark.parametrize("dtype", ["float"])
+@pytest.mark.parametrize(
+    "dtype", ["float"] if not current_platform.is_rocm() else ["float16"]
```
DarkLight1337: Add a comment explaining why this is necessary

vllm/transformers_utils/config.py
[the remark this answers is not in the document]
DarkLight1337: Is it just me or do I not see this function used anywhere via this import?

vllm/model_executor/models/qwen3_vl.py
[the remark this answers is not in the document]
AndreasKaratzas: This logic is part of the attention dispatching for multi-modal models. I don't see another way to declare it—using the utilities proposed in #26315 wouldn't limit this logic to multi-modal models only. Also, I'd rather avoid setting environment variables in the test as a workaround. I think we should resolve this permanently, not just for the test file.

Edit: For some reason the conversation shows your review pointing in the rocm.py file. I am going to take up on your comment in qwen.

vllm/transformers_utils/config.py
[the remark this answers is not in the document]
AndreasKaratzas: You are absolutely right. There was a circular import when I was first working on this, but maybe it got resolved later by an intermediate PR. I've removed that function.

vllm/model_executor/models/qwen3_vl.py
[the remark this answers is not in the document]
AndreasKaratzas: @tjtanaa Thank you for the review. I removed that part from the qwen3-vl code and I am using attention config inside the test to do the dispatching.

vllm/platforms/rocm.py
AndreasKaratzas: Thank you for pointing out this. I resolved it in my recent commit.

tests/models/multimodal/pooling/test_clip.py
AndreasKaratzas: Thank you for pointing this out. I forgot to remove this. I had written this part of code before I add the check in the vit_backend selection mechanism:
```python
        if (
            on_gfx9()
            and find_spec("flash_attn") is not None
            and (dtype == torch.float16 or dtype == torch.bfloat16)
        ):
            logger.info("Using Flash Attention backend for ViT model.")
            return AttentionBackendEnum.FLASH_ATTN
```

But now that the above is in place, I removed this bit of code.

vllm/model_executor/models/siglip2navit.py
```diff
@@ -190,7 +190,10 @@ def apply_rotary_pos_emb(
 ) -> tuple[torch.Tensor, torch.Tensor]:
     cos = cos.chunk(2, dim=-1)[0].contiguous()
     sin = sin.chunk(2, dim=-1)[0].contiguous()
-    if is_flash_attn_backend and not current_platform.is_xpu():
+    if is_flash_attn_backend and not (
+        current_platform.is_xpu() or current_platform.is_rocm()
```
tjtanaa: We only need to check `current_platform.is_cuda()` , `vllm.vllm_flash_attn` only exists on CUDA platform.

```
if is_flash_attn_backend and current_platform.is_cuda():
```

vllm/platforms/rocm.py
```diff
@@ -193,16 +194,26 @@ def get_vit_attn_backend(
...
-        if rocm_aiter_ops.is_mha_enabled():
-            # Note: AITER FA is only supported for Qwen-VL models.
-            # TODO: Add support for other VL models in their model class.
+        @if_aiter_supported
+        def _get_aiter_backend():
             return AttentionBackendEnum.ROCM_AITER_FA
 
-        if on_gfx9() and find_spec("flash_attn") is not None:
+        aiter_backend = _get_aiter_backend()
+        if aiter_backend is not None and envs.VLLM_ROCM_USE_AITER:
```
tjtanaa: May I know the reason behind the needs to change the behaviour? 

We should be able to keep using `rocm_aiter_ops.is_mha_enabled()` as it includes the checks for all of the following conditions
1. `on_gfx9()` (handle the case for Radeon GPU, right now the docker image for Radeon GPU and Instinct GPU share the same docker image, so Radeon Image has `aiter` installed. So we need to explicitly blocked Radeon from using AITER)
2. `IS_AITER_FOUND` (check if `aiter` package is installed)
3. `current_platform.is_rocm()`
4. `envs.VLLM_ROCM_USE_AITER` (default is `False`)
5. `envs.VLLM_ROCM_USE_AITER_MHA` (default is `True`)

AndreasKaratzas: I originally used `is_mha_enabled` to check this, but I thought that it depended on MHA, which has nothing to do with ViT attention. So I migrated to a more tailored check, through `if_aiter_supported`. I can revert back the is mha, but I think that it is not exactly what we want for this check in ViT attention. Also if the default for MHA env changes, we would have to refactor this as well, and I don't see why they should be correlated like that.

micah-wil: It might be nice to add some indicator into `_aiter_ops.py` like `IS_AITER_ENABLED` that basically combines `IS_AITER_FOUND`, `if_aiter_supported`, and `_AITER_ENABLED` (i.e. it evaluates to true if all of those other three things are true). Is that the type of behavior we are looking for here? Might be nice to have something like that for other parts of the codebase as well

vllm/model_executor/models/siglip2navit.py
AndreasKaratzas: I don't know if that is true but, when I run `tests/models/multimodal/generation/test_common.py::test_single_image_models[ovis2_5-test_case6]`, it triggered:

```
(EngineCore_DP0 pid=825) ERROR 12-12 19:10:03 [core.py:866]   File "/usr/local/lib/python3.12/dist-packages/vllm/model_executor/models/siglip2navit.py", line 306, in forward
(EngineCore_DP0 pid=825) ERROR 12-12 19:10:03 [core.py:866]     queries, keys = apply_rotary_pos_emb(
(EngineCore_DP0 pid=825) ERROR 12-12 19:10:03 [core.py:866]                     ^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=825) ERROR 12-12 19:10:03 [core.py:866]   File "/usr/local/lib/python3.12/dist-packages/vllm/model_executor/models/siglip2navit.py", line 204, in apply_rotary_pos_emb
(EngineCore_DP0 pid=825) ERROR 12-12 19:10:03 [core.py:866]     from vllm.vllm_flash_attn.layers.rotary import apply_rotary_emb
(EngineCore_DP0 pid=825) ERROR 12-12 19:10:03 [core.py:866] ModuleNotFoundError: No module named 'vllm.vllm_flash_attn.layers'
```

Because:
```
(EngineCore_DP0 pid=825) INFO 12-12 19:09:20 [vllm.py:722] Cudagraph is disabled under eager mode
(EngineCore_DP0 pid=825) INFO 12-12 19:09:20 [rocm.py:333] Using Triton Attention backend.
(EngineCore_DP0 pid=825) INFO 12-12 19:09:20 [rocm.py:213] Using Flash Attention backend for ViT model.
```

So this part of code resolves that error.

vllm/platforms/rocm.py
tjtanaa: I understand the situation now, so basically, when we first introduced the `VLLM_ROCM_USE_AITER_MHA` it is to indicate that that is to use AITER's MHA function, which include all of those `flash_attn_varlen` function. Like how `VLLM_ROCM_USE_AITER_LINEAR` is for all GEMMS ops (quantized, unquantized, etc). The whole intention is to classify the AITER OPs and avoid introducing new environment variables. Ultimately, we will be removing all of them and only use `rocm_aiter_ops.is_enabled()`.

Another thing, correct me if I am wrong, we want the `test_common.py` to pass by using AITER's `flash_attn_varlen`? If so, a better approach is making sure we can enable the flag `VLLM_ROCM_USE_AITER` in the `test_common.py` by using `monkeypatch`, then set the environment variable `VLLM_ROCM_USE_AITER=1` if the unit tests is running on ROCm.

vllm/model_executor/models/siglip2navit.py
tjtanaa: @AndreasKaratzas can you try finding out why when we are on ROCm, `current_platform.is_cuda()` condition is True?

vllm/platforms/rocm.py
tjtanaa: > It might be nice to add some indicator into `_aiter_ops.py` like `IS_AITER_ENABLED` that basically combines `IS_AITER_FOUND`, `if_aiter_supported`, and `_AITER_ENABLED` (i.e. it evaluates to true if all of those other three things are true). Is that the type of behavior we are looking for here? Might be nice to have something like that for other parts of the codebase as well

Great suggestions. Will follow up on this.

AndreasKaratzas: I just pushed the proposed function. Hope this resolves this issue.

vllm/model_executor/models/siglip2navit.py
AndreasKaratzas: But it is not true. I just put a `assert current_platform.is_cuda(), "This is a test to fail if not on CUDA platform."` line on top of `vllm/tests/models/multimodal/generation/test_common.py`, and it asserted. Let me know if there is anything else unanswered for this.

vllm/platforms/rocm.py
AndreasKaratzas: > I understand the situation now, so basically, when we first introduced the `VLLM_ROCM_USE_AITER_MHA` it is to indicate that that is to use AITER's MHA function, which include all of those `flash_attn_varlen` function. Like how `VLLM_ROCM_USE_AITER_LINEAR` is for all GEMMS ops (quantized, unquantized, etc). The whole intention is to classify the AITER OPs and avoid introducing new environment variables. Ultimately, we will be removing all of them and only use `rocm_aiter_ops.is_enabled()`.
> 
> Another thing, correct me if I am wrong, we want the `test_common.py` to pass by using AITER's `flash_attn_varlen`? If so, a better approach is making sure we can enable the flag `VLLM_ROCM_USE_AITER` in the `test_common.py` by using `monkeypatch`, then set the environment variable `VLLM_ROCM_USE_AITER=1` if the unit tests is running on ROCm.

I am not sure if I understand the last point. I hope that the implementation after the recent commits is satisfactory enough for this PR to be approved.

vllm/_aiter_ops.py
```diff
@@ -54,6 +54,28 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+def is_aiter_enabled() -> bool:
```
tjtanaa: This has already been implemented by the class method `is_enabled()`.

Please use this in the `rocm.py`

```python
from vllm._aiter_ops import rocm_aiter_ops

if rocm_aiter_ops.is_enabled():
   ...
```

vllm/model_executor/models/siglip2navit.py
tjtanaa: > We only need to check `current_platform.is_cuda()` , `vllm.vllm_flash_attn` only exists on CUDA platform.
> 
> ```
> if is_flash_attn_backend and current_platform.is_cuda():
> ```

I have validated this condition to be working for `tests/models/multimodal/generation/test_common.py::test_single_image_models[ovis2_5-test_case6]`. Let's change the condition to `if is_flash_attn_backend and current_platform.is_cuda():`
Because `vllm.vllm_flash_attn` is a CUDA only module.

Like you mentioned, `current_platform.is_cuda()` should be `False`, so it won't go into this path to import `from vllm.vllm_flash_attn.layers.rotary import apply_rotary_emb`.

vllm/platforms/rocm.py
```diff
@@ -193,16 +194,21 @@ def get_vit_attn_backend(
     ) -> AttentionBackendEnum:
         from importlib.util import find_spec
 
-        from vllm._aiter_ops import rocm_aiter_ops
+        from vllm._aiter_ops import is_aiter_enabled
 
-        if rocm_aiter_ops.is_mha_enabled():
-            # Note: AITER FA is only supported for Qwen-VL models.
-            # TODO: Add support for other VL models in their model class.
+        if is_aiter_enabled():
```
tjtanaa: This has already been implemented by the class method `is_enabled()`.

Please use this in the `rocm.py`

```python
from vllm._aiter_ops import rocm_aiter_ops

if rocm_aiter_ops.is_enabled():
   ...
```

However, I think we should retain `rocm_aiter_ops.is_mha_enabled()` condition as it is not affecting any unit tests and the intention of the flag is to control all mha related ops and backends, including `flash_attn_varlen`, like documented in this PR https://github.com/vllm-project/vllm/pull/30586 . (I have opened a draft PR https://github.com/vllm-project/vllm/pull/30586 that will add usage documentation to the `_aiter_ops.py`.)
```
    # _MHA_ENABLED is intended to control MHA Ops/Backends
    # including flash_attn_varlen
    _MHA_ENABLED = envs.VLLM_ROCM_USE_AITER_MHA
```
 But if you still think we should change the default behaviour, please use `if rocm_aiter_ops.is_enabled()` .

tjtanaa (changes requested):
@AndreasKaratzas Thank you for going through this process with me. I have left comments.

vllm/platforms/rocm.py
AndreasKaratzas: Great :) Thank you very much for all these comments. Thank you for bringing up to my attention `rocm_aiter_ops.is_enabled()`. Regarding your last point I still think it's more precise to go for `rocm_aiter_ops.is_enabled()` instead of the MHA. I realize that MHA flag is true by default, but I don't think it's good to be a control knob for the activation of this particular path.

tjtanaa (approved):
LGTM

.buildkite/test-amd.yaml
```diff
@@ -1523,6 +1526,29 @@ steps:
...
 
+- label: Blackwell LM Eval Small Models
+  timeout_in_minutes: 120
+  gpu: b200
+  optional: true # run on nightlies
+  source_file_dependencies:
+  - csrc/
+  - vllm/model_executor/layers/quantization
+  commands:
+  - pytest -s -v evals/gsm8k/test_gsm8k_correctness.py --config-list-file=configs/models-blackwell.txt --tp-size=1
```
tjtanaa: can you check if this command works. I remember `--tp-size=1` is causing an issue and this PR fixes it `https://github.com/vllm-project/vllm/pull/30911`

.buildkite/test-amd.yaml
```diff
@@ -1523,6 +1526,29 @@ steps:
...
+  mirror_hardwares: [amdexperimental, amdproduction]
+  agent_pool: mi325_1
+  # grade: Blocking
+  working_dir: "/vllm-workspace/.buildkite/lm-eval-harness"
+  source_file_dependencies:
+  - vllm/multimodal/
+  - vllm/inputs/
+  - vllm/v1/core/
+  commands:
+  - pytest -s -v test_lm_eval_correctness.py --config-list-file=configs/models-mm-small.txt --tp-size=1
```
tjtanaa: can you check if this command works. I remember `--tp-size=1` is causing an issue and this PR fixes it `https://github.com/vllm-project/vllm/pull/30911`

.buildkite/test-amd.yaml
```diff
@@ -989,6 +989,8 @@ steps:
   - vllm/inputs/
   - vllm/v1/core/
   commands:
+  - export MIOPEN_DEBUG_CONV_DIRECT=0
```
tjtanaa, line 994: @AndreasKaratzas Wow this is major improvement! Could you provide more context on how these environment variable settings come about?

.buildkite/test-amd.yaml
tjtanaa: can you check if this command works. I remember `--tp-size=1` is causing an issue and this PR fixes it `https://github.com/vllm-project/vllm/pull/30911`

.buildkite/test-amd.yaml
micah-wil: Context for that here: https://github.com/pytorch/pytorch/issues/169857

.buildkite/test-amd.yaml
AndreasKaratzas: I'm gonna try it and update you, sure.

.buildkite/test-amd.yaml
AndreasKaratzas: This does not have any other issues. The env var update is for performance purposes.

.buildkite/test-amd.yaml
AndreasKaratzas: Is this duplicate comment?

.buildkite/test-amd.yaml
AndreasKaratzas: Ok, I see what has happened here, I merged and duplicated some tests. I am amending this immediately.

.buildkite/test-amd.yaml
AndreasKaratzas: Ok, I see what has happened here, I merged and duplicated some tests. I am amending this immediately.

AndreasKaratzas: Ok, I see what has happened here, I merged and duplicated some tests. I am amending this immediately.

AndreasKaratzas: @tjtanaa Thank you for pointing out this. Let me know if my most recent commit reflects the changes you suggested.

.buildkite/test-amd.yaml
AndreasKaratzas: @tjtanaa Thank you for pointing out this. Let me know if my most recent commit reflects the changes you suggested.

.buildkite/test-amd.yaml
AndreasKaratzas: @tjtanaa Thank you for pointing out this. Let me know if my most recent commit reflects the changes you suggested.

tjtanaa (approved):
Thank you for following up.

PR #26575, opened 2025-10-10
[ROCm][FEAT] Support AITER RMSNorm quantization fusion pass 

vllmellm:
## Purpose

This PR supports fusion pass for ROCM AITER by fusing `+rms_norm`, aiter rmsnorm ops, and `+quant_fp8`, vllm quantization custom ops.

**Benchmark Result**

| Metric                              | Without Fusion Pass | With Fusion Pass   |
|--------------------------------------|---------------------|-------------------|
| Successful requests                  | 500                 | 500               |
| Benchmark duration (s)               | 173.76              | 170.31            |
| Total input tokens                   | 520,558             | 520,558           |
| Total generated tokens               | 456,122             | 456,834           |
| Request throughput (req/s)           | 2.88                | 2.94              |
| Output token throughput (tok/s)      | 2,625.02            | 2,682.37          |
| Peak output token throughput (tok/s) | 7,924.00            | 7,410.00          |
| Peak concurrent requests             | 500.00              | 500.00            |
| Total token throughput (tok/s)       | 5,620.87            | 5,738.91          |
| Mean TTFT (ms)                       | 35,048.69           | 34,637.29         |
| Median TTFT (ms)                     | 28,413.54           | 28,841.97         |
| P99 TTFT (ms)                        | 91,625.66           | 90,845.96         |
| Mean TPOT (ms)                       | 167.64              | 170.28            |
| Median TPOT (ms)                     | 117.48              | 118.62            |
| P99 TPOT (ms)                        | 881.79              | 913.72            |
| Mean ITL (ms)                        | 114.17              | 113.99            |
| Median ITL (ms)                      | 57.83               | 61.15             |
| P99 ITL (ms)                         | 2,111.96            | 2,086.31          |

### benchmark setting
`
vllm bench serve \
--backend vllm \
--model "RedHatAI/Qwen3-14B-FP8-dynamic" \
--dataset-name random \
--num-prompts 500 \
--random-input-len 1000 \
--random-output-len 1000 \
--endpoint /v1/completions \
--random-range-ratio 0.9 \
`

## Test Plan

- Unit test has been added in `vllm/tests/compile/test_rocm_aiter_fusion.py` that verifies accuracy, replacement of the ops in cuda graph.
- end to end test using `RedHatAI/Qwen3-14B-FP8-dynamic` model

**environment setting**
Step 1: run vllm serve
`
vllm serve RedHatAI/Qwen3-14B-FP8-dynamic \
[...]

PR #25693, opened 2025-09-25
[Rocm][torch.compile] Adding layernorm + fp8 block quant and silu + fp8 block quant for Aiter

charlifu:
This PR adds a few fusion passes for Aiter to fusion layernorm + fp8 block quant and silu + fp8 block quant.

PR #30274, opened 2025-12-08
[AMD] Amd/deepseek aiter fusions

k50112113:
This PR is a draft PR for adding triton fusion optimization for DS FP4 and FP8 on ROCm devices


hmellor:
Closing as obsolete. The AITER fusion passes here landed separately in #25693 and #26575, and `vllm/compilation` has since been reorganised (#33731) and rewritten for vLLM IR (#33825, #36823), so every module this draft imports from is gone. The remaining DeepSeek triton shared-expert and FP4 pieces would need a fresh PR against current main.

PR #30276, opened 2025-12-08
[ROCM][CI] Fix AMD Examples Test Group

Concurrensee:
This PR fixes AMD Examples Test Group

### DIFF
--- a/.buildkite/test-amd.yaml
+++ b/.buildkite/test-amd.yaml
@@ -435,7 +435,7 @@ steps:
 
 - label: Examples Test # 30min
   timeout_in_minutes: 45
-  mirror_hardwares: [amdexperimental]
+  mirror_hardwares: [amdexperimental, amdproduction]
   agent_pool: mi325_1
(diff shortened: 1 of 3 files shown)


requirements/rocm-test.txt
```diff
@@ -88,3 +88,6 @@ pqdm==0.2.0
 arctic-inference == 0.1.1
 # Required for Nemotron test
 open-clip-torch==2.32.0
+
+# Required for Examples test
+vllm[audio]==0.11.0
```
tjtanaa: Is this safe? will it override the `vllm` version?

I suggest following `test.txt` by specifying the packages directly, avoid using `vllm[audio]==0.11.0`

Concurrensee: This used to address an error cause by incorrect file names. But that error has been fixed, and this no longer needed. Its also actually not good to directly use vllm[audio]

gshtras (approved):

PR #30277, opened 2025-12-08
[BugFix] Fix non detected failing tests

ilmarkov:
## Purpose
Compilation tests were using the find command, which does not propagate errors, so the CI passes even when tests fail.
Fix compilation tests find commands.

Fixes tests that were not previously detected by CI:

1. `compile/test_pass_manager.py::test_pass_manager_uuid`
Failures were caused by usage of PassManager.uuid without pass_context. 

2. 
```
compile/test_compile_ranges.py::test_compile_ranges
compile/test_compile_ranges.py::test_compile_config_get_compile_ranges
compile/test_compile_ranges.py::test_inductor_cache_compile_ranges
```

Test failures were caused by enforced non-default fields in SchedulerConfig.

3. `compile/fullgraph/test_multiple_graphs.py`
Fixes circular import

4. `compile/fullgraph/test_multimodal_compile.py::test_qwen2_5_vl_compilation`
Increases upper bound of the last compile range in case of encoder blocks compilation.
Probably, needs a general way to detect encoder compilation in PiecewiseBackend.

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/compilation/piecewise_backend.py
+++ b/vllm/compilation/piecewise_backend.py
@@ -53,8 +53,27 @@ class PiecewiseBackend:
         self.is_last_graph = piecewise_compile_index == total_piecewise_compiles - 1
 
         self.is_full_graph = total_piecewise_compiles == 1
+        # TODO: we need to generalize encoder compilation to other models
+        self.is_encoder_compilation = vllm_backend.prefix in [
+            "Qwen2_5_VisionPatchEmbed",
+            "Qwen2_5_VisionPatchMerger",
+            "Qwen2_5_VisionBlock",
+        ]
 
         self.compile_ranges = self.compilation_config.get_compile_ranges()
+        if self.is_encoder_compilation:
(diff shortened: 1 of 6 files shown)


zou3519 (approved):
thank you

zou3519 (approved):

vllm/compilation/piecewise_backend.py
```diff
@@ -53,8 +53,27 @@ def __init__(
         self.is_last_graph = piecewise_compile_index == total_piecewise_compiles - 1
 
         self.is_full_graph = total_piecewise_compiles == 1
+        # TODO: we need to generalize encoder compilation to other models
```
ProExpertProg, line 56: Could we move this to config init ewhere we init the range in the first place?

ilmarkov: We init compilation config once. But we need to use different ranges for encoder and decoder compilations. We should mark the type of compilation somehow in a follow-up PR.

PR #29999, opened 2025-12-03
[Bug] Fix vLLM config is not set error

yewentao256:
## Purpose

`parallel_config = get_current_vllm_config().parallel_config` will try to get the vllm config that is not set, logging like

```bash
(Worker_TP0_EP0 pid=2902226) WARNING 12-03 13:43:52 [vllm.py:1322] Current vLLM config is not set.
(Worker_TP0_EP0 pid=2902226) INFO 12-03 13:43:52 [scheduler.py:228] Chunked prefill is enabled with max_num_batched_tokens=2048.
(Worker_TP3_EP3 pid=2902229) WARNING 12-03 13:43:52 [vllm.py:1322] Current vLLM config is not set.
(Worker_TP3_EP3 pid=2902229) INFO 12-03 13:43:52 [scheduler.py:228] Chunked prefill is enabled with max_num_batched_tokens=2048.
(Worker_TP1_EP1 pid=2902227) WARNING 12-03 13:43:52 [vllm.py:1322] Current vLLM config is not set.
(Worker_TP1_EP1 pid=2902227) INFO 12-03 13:43:52 [scheduler.py:228] Chunked prefill is enabled with max_num_batched_tokens=2048.
(Worker_TP2_EP2 pid=2902228) WARNING 12-03 13:43:52 [vllm.py:1322] Current vLLM config is not set.
(Worker_TP2_EP2 pid=2902228) INFO 12-03 13:43:52 [scheduler.py:228] Chunked prefill is enabled with max_num_batched_tokens=2048.
(Worker_TP0_EP0 pid=2902226) WARNING 12-03 13:43:52 [vllm.py:1322] Current vLLM config is not set.
(Worker_TP0_EP0 pid=2902226) INFO 12-03 13:43:52 [scheduler.py:228] Chunked prefill is enabled with max_num_batched_tokens=2048.
(Worker_TP3_EP3 pid=2902229) WARNING 12-03 13:43:52 [vllm.py:1322] Current vLLM config is not set.
(Worker_TP3_EP3 pid=2902229) INFO 12-03 13:43:52 [scheduler.py:228] Chunked prefill is enabled with max_num_batched_tokens=2048.
(Worker_TP1_EP1 pid=2902227) WARNING 12-03 13:43:52 [vllm.py:1322] Current vLLM config is not set.
(Worker_TP1_EP1 pid=2902227) INFO 12-03 13:43:52 [scheduler.py:228] Chunked prefill is enabled with max_num_batched_tokens=2048.
[...]

PR #30282, opened 2025-12-08
[Feat] Refactor for `parallel_config` in `FusedMoEModularKernel`

yewentao256:
## Purpose

Intend to fix https://github.com/vllm-project/vllm/pull/29999#discussion_r2593802287

and https://github.com/vllm-project/vllm/pull/29999#discussion_r2604562114

But after discussion with @ProExpertProg and @bnellnm  , we found that require a must-pass param needs to update everything especially for tests, we decide to instead using a moe_parallel_config

This PR could also fix current issue in main

```bash
(Worker_TP2_EP2 pid=248418) WARNING 12-10 13:35:58 [vllm.py:1394] Current vLLM config is not set.
(Worker_TP2_EP2 pid=248418) INFO 12-10 13:35:58 [scheduler.py:228] Chunked prefill is enabled with max_num_batched_tokens=2048.
(Worker_TP4_EP4 pid=248420) WARNING 12-10 13:35:58 [vllm.py:1394] Current vLLM config is not set.
(Worker_TP4_EP4 pid=248420) INFO 12-10 13:35:58 [scheduler.py:228] Chunked prefill is enabled with max_num_batched_tokens=2048.
(Worker_TP3_EP3 pid=248419) WARNING 12-10 13:35:58 [vllm.py:1394] Current vLLM config is not set.
(Worker_TP3_EP3 pid=248419) INFO 12-10 13:35:58 [scheduler.py:228] Chunked prefill is enabled with max_num_batched_tokens=2048.
(Worker_TP5_EP5 pid=248421) WARNING 12-10 13:35:58 [vllm.py:1394] Current vLLM config is not set.
(Worker_TP5_EP5 pid=248421) INFO 12-10 13:35:58 [scheduler.py:228] Chunked prefill is enabled with max_num_batched_tokens=2048.
(Worker_TP0_EP0 pid=248416) WARNING 12-10 13:35:58 [vllm.py:1394] Current vLLM config is not set.
(Worker_TP0_EP0 pid=248416) INFO 12-10 13:35:58 [scheduler.py:228] Chunked prefill is enabled with max_num_batched_tokens=2048.
(Worker_TP6_EP6 pid=248422) WARNING 12-10 13:35:58 [vllm.py:1394] Current vLLM config is not set.
(Worker_TP6_EP6 pid=248422) INFO 12-10 13:35:58 [scheduler.py:228] Chunked prefill is enabled with max_num_batched_tokens=2
```


vllm/model_executor/layers/fused_moe/cutlass_moe.py
```diff
@@ -460,7 +460,7 @@ def cutlass_moe_fp8(
     expert_map: torch.Tensor | None = None,
     apply_router_weight_on_input: bool = False,
     global_num_experts: int = -1,
-    parallel_config=None,
+    moe_parallel_config=None,
```
bnellnm: I don't think these extra parameters are necessary.  This version of the cutlass kernels is never used in an EP context.

vllm/model_executor/layers/quantization/utils/flashinfer_utils.py
```diff
@@ -247,9 +247,9 @@ def flashinfer_cutlass_moe_fp8(
     assert quant_config is not None
 
     # Construct modular kernel with block-scale support when requested.
-    parallel_config = getattr(
-        getattr(layer, "vllm_config", None),
-        "parallel_config",
+    moe_parallel_config = getattr(
+        moe if moe is not None else layer,
+        "moe_parallel_config",
```
bnellnm: The `layer` should always have a `moe_parallel_config` (it should match the one from `moe` if any)

bnellnm:
I think you should get rid of all the extra `moe_parallel_config` arguments except for the ones in `fused_moe_modular_kernel.py` and `flashinfer_utils.py`.  Most of these are only used in non-EP situations.

The following test files run in a parallel context so might optionally need a parallel context
- tests/kernels/moe/test_pplx_moe.py
- tests/kernels/moe/modular_kernel_tools/common.py
- tests/kernels/moe/test_deepep_moe.py
- tests/kernels/moe/test_pplx_cutlass_moe.py
- tests/kernels/moe/test_deepep_deepgemm_moe.py

vllm/model_executor/layers/fused_moe/cutlass_moe.py
yewentao256: Let's have it here as we don't know whether it will be expanded? Passing a param won't cause perf loss

yewentao256:
> I think you should get rid of all the extra `moe_parallel_config` arguments except for the ones in `fused_moe_modular_kernel.py` and `flashinfer_utils.py`. Most of these are only used in non-EP situations.
> 
> The following test files run in a parallel context so might optionally need a parallel context
> 
> * tests/kernels/moe/test_pplx_moe.py
> * tests/kernels/moe/modular_kernel_tools/common.py
> * tests/kernels/moe/test_deepep_moe.py
> * tests/kernels/moe/test_pplx_cutlass_moe.py
> * tests/kernels/moe/test_deepep_deepgemm_moe.py

Thanks for the reply! As far as I know, the `deep_gemm_moe` could be called in ep cases as well, so to make sure we cover everything and make sure it could be supported automatically if we expand, I think we can leave as it is since passing a param won't cause additional trouble.

Test fixed, thanks for pointing out!

vllm/model_executor/layers/quantization/utils/flashinfer_utils.py
yewentao256: Fixed

vllm/model_executor/layers/quantization/utils/flashinfer_utils.py
```diff
@@ -262,7 +257,7 @@ def flashinfer_cutlass_moe_fp8(
             out_dtype=hidden_states.dtype,
             use_deepseek_fp8_block_scale=use_deepseek_fp8_block_scale,
         ),
-        parallel_config=parallel_config,
+        moe_parallel_config=layer.moe_parallel_config,
```
pavanimajety, line 260: is parallel_config a property of both FusedMoEConfig and layer?

yewentao256: The object that owns a parallel_config field is `VllmConfig`.

Both the MoE layer (FusedMoE) and FusedMoEConfig instead carry a moe_parallel_config: `FusedMoEParallelConfig`, which is constructed from `vllm_config.parallel_config` (see `FusedMoE.__init__` and FusedMoEConfig).

vllm/model_executor/layers/fused_moe/cutlass_moe.py
bnellnm: This function is specifically intended for the non-EP case, i.e. it uses `MoEPrepareAndFinalizeNoEP` for constructing the modular kernel.  Similarly with other use sites.

bnellnm:
> > I think you should get rid of all the extra `moe_parallel_config` arguments except for the ones in `fused_moe_modular_kernel.py` and `flashinfer_utils.py`. Most of these are only used in non-EP situations.
> > The following test files run in a parallel context so might optionally need a parallel context
> > 
> > * tests/kernels/moe/test_pplx_moe.py
> > * tests/kernels/moe/modular_kernel_tools/common.py
> > * tests/kernels/moe/test_deepep_moe.py
> > * tests/kernels/moe/test_pplx_cutlass_moe.py
> > * tests/kernels/moe/test_deepep_deepgemm_moe.py
> 
> Thanks for the reply! As far as I know, the `deep_gemm_moe` could be called in ep cases as well, so to make sure we cover everything and make sure it could be supported automatically if we expand, I think we can leave as it is since passing a param won't cause additional trouble.
> 
> Test fixed, thanks for pointing out

Like `cutlass_moe_fp8`, `deep_gemm_moe` is specifically for the non-EP case.  All the EP kernels are constructed and called by `FusedMoEModularMethod`.

vllm/model_executor/layers/fused_moe/modular_kernel.py
```diff
@@ -716,20 +719,30 @@ def __init__(
...
+        # current vLLM config.
+        if moe_parallel_config is None:
+            vllm_config = get_current_vllm_config()
+            parallel_config = vllm_config.parallel_config
+            moe_parallel_config = FusedMoEParallelConfig.make(
+                tp_size_=parallel_config.tensor_parallel_size,
+                pcp_size_=parallel_config.pipeline_parallel_size,
+                dp_size_=parallel_config.data_parallel_size,
+                vllm_parallel_config=parallel_config,
+            )
```
bnellnm: I think we should just assume that this is a non-EP case if no `moe_parallel_config` is passed.  It will avoid the call to `get_current_vllm_config` which is causing all the warnings/logs.

vllm/model_executor/layers/fused_moe/cutlass_moe.py
yewentao256: Fixed

vllm/model_executor/layers/fused_moe/modular_kernel.py
yewentao256: Fixed

bnellnm (approved):
LGTM. Thanks for the fix!

jiahanc (commented):
Are MOE backend not `FusedMoEModularKernel` tested not affected? For example, Flashinfer TRTLLM MOE

robertgshaw2-redhat (approved):

robertgshaw2-redhat:
merged main to re-run failing test

PR #30287, opened 2025-12-09
[Bugfix] Fix compressed-tensors models failing to load with transformers backend

mgoin:
## Purpose

The transformers backend's `WeightsMapper` uses a broad prefix mapping (`"" -> "model."`) to transform weight paths. This mapper is also applied to quantization config targets via `apply_vllm_mapper`. However, compressed-tensors targets can be module class names (e.g., `"Linear"`) or regex patterns (e.g., `"re:.*proj"`), not just layer paths.

When `"Linear"` is transformed to `"model.Linear"`, the target matching fails because the module class name check looks for `"Linear"` in `"model.Linear"` (substring match), which fails.

The fix filters which targets get transformed: only layer paths (containing `.` and not starting with `re:`) are mapped. Class names and regex patterns are preserved.

## Test Plan

```bash
# Previously failing, now passing - compressed-tensors with transformers backend
vllm serve RedHatAI/Qwen3-0.6B-FP8-BLOCK --model-impl transformers

# Verify native backend still works
vllm serve RedHatAI/Qwen3-0.6B-FP8-BLOCK

# Verify fp8 (non-compressed-tensors) still works with transformers
vllm serve Qwen/Qwen3-0.6B-FP8 --model-impl transformers
```

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
+++ b/vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py
@@ -136,10 +136,10 @@ class CompressedTensorsConfig(QuantizationConfig):
             return target
 
         def _apply_dict(d: dict) -> dict:
-            return {k: v for t, v in d.items() if (k := _map_target(t))}
+            return {k: v for t, v in d.items() if (k := _map_target(t)) is not None}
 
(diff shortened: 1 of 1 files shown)


Isotr0py (approved):
LGTM

PR #30292, opened 2025-12-09
[CI/Build][Kernel][BugFix][AMD] Fix per_token_group_quant_fp8 to use correct fp8 min/max values and update atol/rtol in test_quantfp8_group_functionality 

rasmith:
This PR updates per_token_group_quant_fp8 to use `-224.0/224.0` values for fp8 min/max values when the platform uses the `torch.float8_e4m3fnuz` fp8 dtype.  Without this, torch returns `-240.0/240.0` which results in accuracy issues and can result in mismatches between implementations since the `-224.0/224.0` values are returned.  I also modified the `atol/rtol `values in `test_quantfp8_group_functionality `since the failures were just barely out of the range of the original ones.


vllm/model_executor/layers/quantization/utils/fp8_utils.py
```diff
@@ -723,8 +723,8 @@ def per_token_group_quant_fp8(
     assert x.stride(-1) == 1, "`x` groups must be contiguous"
 
     finfo = torch.finfo(dtype)
-    fp8_min = finfo.min
-    fp8_max = finfo.max
+    fp8_min = -224.0 if current_platform.is_fp8_fnuz() else finfo.min
+    fp8_max = 224.0 if current_platform.is_fp8_fnuz() else finfo.max
```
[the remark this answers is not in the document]
rasmith: This is a redundant check.

vllm/model_executor/layers/quantization/utils/fp8_utils.py
```diff
@@ -723,8 +723,8 @@ def per_token_group_quant_fp8(
     assert x.stride(-1) == 1, "`x` groups must be contiguous"
 
     finfo = torch.finfo(dtype)
-    fp8_min = finfo.min
-    fp8_max = finfo.max
+    fp8_min = -224.0 if current_platform.is_fp8_fnuz() else finfo.min
```
gshtras, line 737: Since this is a copy-paste from input_quant_fp8.py, please either copy over the comment explaining this too, or create a function

rasmith: I added a comment.

SageMoore (approved):

rasmith:
@ghstras Please take another look

vllm/model_executor/layers/quantization/utils/fp8_utils.py
```diff
@@ -730,9 +730,11 @@ def per_token_group_quant_fp8(
     )
     assert x.stride(-1) == 1, "`x` groups must be contiguous"
 
+    # Using the default value (240.0) from pytorch will cause accuracy
+    # issue on dynamic quantization models. Here use 224.0 for fnuz on ROCm.
```
yewentao256: Is 224.0 only needed for Rocm? Then please update the code to `if rocm()` so we don't affect cuda

yewentao256 (commented):
Thanks for the work!

vllm/model_executor/layers/quantization/utils/fp8_utils.py
rasmith: It uses `current_platform.is_fp8_fnuz` since `is_rocm `won't work for MI355, since that one will return `True `for `is_rocm` but `False `for `is_fp8_fnuz`, since MI355 uses the `torch.float8_e4m3fn` dtype for fp8

rasmith: @yewentao256  I updated the comment to clarify.

rasmith: @yewentao256  It's using `current_platform.is_fp8_fnuz()` which shouldn't affect CUDA platforms.

yewentao256 (approved):
LGTM, thanks for the work!

PR #30294, opened 2025-12-09
[Doc] update Intel GPU MM status in Feature x Hardware matrix

faaany:
## Purpose
All the MM models in https://docs.vllm.ai/en/stable/features/multimodal_inputs.html are now in parity on XPU to status of CUDA.

More details see https://github.com/vllm-project/vllm/issues/26965


DarkLight1337 (approved):
Thanks for updating!

PR #30299, opened 2025-12-09
[Bugfix] Update WSL detection to check for WSL1 compatibility as WSL2…

HoneyBerries:
Fixed the memory pinning being disabled on WSL2 instances.
## Purpose
The purpose of this PR is to fix a bug where on WSL2 memory-pinning would be disabled.  However, memory-pinning actually works on WSL2, but not on WSL1.  The changes currently check for WSL1 and not any WSL distro.
## Test Plan
There was not any tests for the memory pinning logic nor the WSL1/WSL2 detection in the codebase. If you want, I can help you add a unit test that verifies the correct behavior for different environments.
## Test Result
Success (nothing else changed)

This is just a trivial fix and did not change much of the codebase, other than a few lines in 1 file.

---

> [!NOTE]
> Enables pinned memory on WSL2 by refining WSL detection to target WSL1 only.
> 
> - Updates `in_wsl()` to detect WSL1 (`"microsoft" in uname` and not `"wsl2"`)
> - Adjusts `Platform.is_pin_memory_available()` to only disable pin memory on WSL1 and updates warning message
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 29efefa42d0b5ebb7c1468d5dfbc70f8df84e6b8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


vllm/platforms/interface.py
```diff
@@ -31,8 +31,9 @@
 
 
 def in_wsl() -> bool:
-    # Reference: https://github.com/microsoft/WSL/issues/4071
-    return "microsoft" in " ".join(platform.uname()).lower()
+    # (Old) Reference: https://github.com/microsoft/WSL/issues/4071
+    # (New) Now checks only for WSL1 as WSL2 has full Linux kernel support.
+    return "wsl1" in " ".join(platform.uname()).lower()
```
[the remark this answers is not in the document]
HoneyBerries: then whould it be more reliable to check for both microsoft and wsl1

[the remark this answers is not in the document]
mgoin: I think it would be best to check multiple signals for wsl1, seems like some research needs to be done if the assistants think there are issues

HoneyBerries:
@vllm-bot  add-label

vllm/platforms/interface.py
a4lg: Confirmed that this is *NOT* working.

I investigated both WSL1 and WSL2 and found some key differences.

## Proposal

1.  Detect WSL by querying like `"-microsoft" in platform.release().lower()` or some other means.
2.  Identify the version of WSL (defaulting to WSL2 might be useful today).

## WSL1/2 Identification Ideas (some differences)

### Presence / Absence of `/proc` files

WSL1: Lacks both `/proc/[PID]/io` and `/proc/config.gz` (effectively hardcoded)
WSL2: Has both `/proc/[PID]/io` and `/proc/config.gz`

If we assume the default WSL2 Linux kernel, we are safe to assume that they are one of the key differences.

### `platform.release()` (case-sensitive)

WSL1: contains (ends with) `-Microsoft` (hardcoded in `lxcore.sys`, a key system file in the WSL1 runtime)
WSL2: contains (ends with) `-microsoft-standard-WSL2` (used in default WSL2 kernels; not hardcoded; with small exceptions)

Checking WSL2 by `"-microsoft-standard-WSL2" in platform.release()` (not with `.endswith(...)` seems marginally better considering the fact that the user might rebuild their own custom kernel.
On the other hand, `platform.release()` on WSL1 cannot be customized and `platform.release().endswith("-Microsoft")` seems safer.

Technically, very early versions of the official WSL 2 Linux kernel (Linux 4.19-based ones) did not have `-WSL2` suffix  (ends with `-microsoft-standard`) but it's very unlikely that the user depends on such outdated kernel (they are also WSLg-incompatible).

### `/proc/cmdline`

Both are possibly set by user land components of WSL1/WSL2 and probably not hardcoded.

WSL1: begins with `BOOT_IMAGE=`
WSL2: begins with `initrd=`

a4lg (changes requested):

vllm/platforms/interface.py
a4lg: Okay, I might submit a separate PR but at least there's no reason to object.

a4lg (approved):

vllm/platforms/interface.py
HoneyBerries: thx

a4lg:
Technically, very early versions of the official WSL 2 Linux kernel (Linux 4.19-based ones) did not have `-WSL2` suffix but it's very unlikely that the user depends on such outdated kernel (they are also WSLg-incompatible).

PR #30302, opened 2025-12-09
[Misc] Pass reasoning to deepseekV32 tokenizer

kingsmad:
## Purpose
Currently the thinking mode is passed by "chat_template_kwargs":{"thinking":true}} to the chatCompletion from client. In addition, this PR adds the capability to pass the reasoning field from responsesRequest to the deepseekV32 tokenizer thus when reasoning effort is set, we will be able to enable thinking. 

## Test Plan
Run local server, with/without setting reasoning_effort, checked responses.

## Test Result
When passed reasoning_effort, we do see \<think\> tag in the answer

 "We are asked: \"what's 1+1\". This is a simple arithmetic question. The answer is 2. However, the user might be testing if the model responds correctly to a trivial question. So we answer straightforwardly. \</think\> 1 + 1 = 2."

---
<details>

</details>

### DIFF
--- a/vllm/tokenizers/deepseekv32.py
+++ b/vllm/tokenizers/deepseekv32.py
@@ -43,6 +43,10 @@ class DeepseekV32Tokenizer(HfTokenizer):
         thinking_mode = "thinking"
         if not thinking:
             thinking_mode = "chat"
+        reasoning = kwargs.get("reasoning")
+        if reasoning is not None:
+            thinking_mode = "thinking"
+
         conversation = kwargs.get("conversation", messages)
         messages = conversation.copy()
         drop_thinking = True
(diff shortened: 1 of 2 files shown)


vllm/tokenizers/deepseekv32.py
```diff
@@ -43,6 +43,10 @@ def apply_chat_template(self, messages, tools=None, **kwargs):
         thinking_mode = "thinking"
         if not thinking:
             thinking_mode = "chat"
+        reasoning = kwargs.get("reasoning")
```
Liccol, line 46: The relationship between thinking and reasoning parameters is confusing

vllm/tokenizers/deepseekv32.py
```diff
@@ -43,6 +43,10 @@ def apply_chat_template(self, messages, tools=None, **kwargs):
...
+        if reasoning is not None:
+            thinking_mode = "thinking"
```
Liccol, line 48: maybe use Enum for the possible values of thinking_mode, will introduce other values in the future?

PR #30310, opened 2025-12-09
[Misc][Quantization] Clarify the intent of GGUF `FusedMoE` weight materialization

a4lg:
## Purpose

This is a refactoring PR with no functional changes (unless the GGUF file is broken).

In the process of `FusedMoE` weight data materialization from GGUF files, there is a magic number and some intents are not clear enough.

This commit clarifies some of them:

1.  GGUF (currently) requires 3D tensor(s) (i.e. `full_load`) for `FusedMoE` layer weights.
2.  `w1` and `w3` are merged per expert, i.e. the next dimension after the expert ID is to be doubled to store both `w1` and `w3`.
    *   Expert ID is the first dimension (as in the code right after the `if is_gguf_weight...` block).
    *   That means, the second dimension's size (`final_shape[1]`) should be doubled for `w1` and `w3`, improving the clarity.

... and makes some minor adjustments.

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/layers/fused_moe/layer.py
+++ b/vllm/model_executor/layers/fused_moe/layer.py
@@ -1200,10 +1200,14 @@ class FusedMoE(CustomOp):
         if full_load:
             shard_dim += 1
 
-        # Materialize GGUF UninitializedParameter
+        # Materialize GGUF UninitializedParameter accounting merged weights
         if is_gguf_weight and isinstance(param, UninitializedParameter):
+            # To materialize a tensor, we must have full shape including
+            # number of experts, making this portion to require `full_load`.
+            assert full_load
             final_shape = list(loaded_weight.shape)
-            if shard_id in ["w1", "w3"]:
+            # w1 and w3 are merged per expert.
+            if shard_id in {"w1", "w3"}:
                 final_shape[1] *= 2
             final_shape[shard_dim] = final_shape[shard_dim] // self.tp_size
             param.materialize(final_shape, dtype=loaded_weight.dtype)


vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1207,10 +1207,13 @@ def weight_loader(
         if full_load:
             shard_dim += 1
 
-        # Materialize GGUF UninitializedParameter
+        # Materialize GGUF UninitializedParameter accounting merged weights
         if is_gguf_weight and isinstance(param, UninitializedParameter):
+            # GGUF currently requires full load (3D tensors).
+            assert full_load
```
Isotr0py, line 1207: Perhaps add a message to clarify what's happening when assertion is false?

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1207,10 +1207,13 @@ def weight_loader(
...
             final_shape = list(loaded_weight.shape)
-            if shard_id in ["w1", "w3"]:
+            # w1 and w3 are merged per expert.
+            if shard_id in {"w1", "w3"}:
```
Isotr0py, line 1210: Is this necessary to switch list to set here?

a4lg: When an object is tested whether it is in a collection, I prefer sets (unless the collection is a short and dynamic list).

vllm/model_executor/layers/fused_moe/layer.py
a4lg: Hmm, rather, it would be better to improve the comment (i.e. what is truly necessary).  I'll consider changing around this line.

vllm/model_executor/layers/fused_moe/layer.py
a4lg: While it's not functionally necessary, it's a good time to improve minor things.

Isotr0py (approved):

vllm/model_executor/layers/fused_moe/layer.py
a4lg: I changed the comment to note why `full_load` is required.

Before:

```py
# GGUF currently requires full load (3D tensors).
```

After:

```py
# To materialize a tensor, we must have full shape including
# number of experts, making this portion to require `full_load`.
```

PR #30316, opened 2025-12-09
[Core] Allow users to modify the scheduler configuration online in dev mode.

noooop:
## Purpose

The pull request introduces functionality to dynamically reconfigure scheduler parameters (`max_num_seqs` and `max_num_batched_tokens`) online. 

This is a valuable addition for benchmarking and dynamic resource management. 

e.g. 

vllm bench sweep only needs to be started once 

cc @DarkLight1337 @lengrongfu 

## Test Plan

benchmark demo:
offline: https://github.com/noooop/snippet/blob/main/benchmarks/embed5/v1_offline.py
online: https://github.com/noooop/snippet/blob/main/benchmarks/embed5/v1_online.py

## Test Result

nan

---
<details>

</details>


noooop:
@DarkLight1337 @lengrongfu

Quick review

vllm/engine/protocol.py
```diff
@@ -188,3 +188,13 @@ async def collective_rpc(
     async def get_supported_tasks(self) -> tuple[SupportedTask, ...]:
         """Get supported tasks"""
         raise NotImplementedError
+
+    def reconfigure(
+        self, max_num_seqs: int | None, max_num_batched_tokens: int | None
```
lengrongfu, line 196: Could this parameter be designed to be more universal? If other parameters need to be modified in the future, there's no need to add new parameters. It's suggested to pass a structure or dic instead.

noooop: I also agree that passing a structure or dict would be better. Initially, the first version created a structure, but since there are only two parameters that can be modified, adding a structure just for that would make the code overly verbose. That’s why it was changed to the current approach.

PR #30331, opened 2025-12-09
[Bugfix] tpu_model_runner: set vllm config context when calling reset_dynamo_cache()

dtrifiro:
The switch to `TorchCompileWithNoGuardsWrapper.__init__(compiled_model)` in #25110 causes a startup crash
due to an attempt to access `VllmConfig` too early.

See https://github.com/vllm-project/vllm/pull/25110/files#r2593461144

### DIFF
--- a/vllm/v1/worker/tpu_worker.py
+++ b/vllm/v1/worker/tpu_worker.py
@@ -10,7 +10,7 @@ import torch
 import torch.nn as nn
 
 import vllm.envs as envs
-from vllm.config import VllmConfig
+from vllm.config import VllmConfig, set_current_vllm_config
 from vllm.distributed import (
     ensure_model_parallel_initialized,
     init_distributed_environment,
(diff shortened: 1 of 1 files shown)


NickLucche (commented):
Thanks for the fix @dtrifiro , this makes sense to me!

One suggestion: what about moving the `set_current_vllm_config` call to tpu_worker.py, so that it better matches the use of it we're doing in gpu_worker.py?

NickLucche (approved):
LGTM

ProExpertProg (approved):

dtrifiro:
Test failures look unrelated, can this be merged? 


Failures:
- ```[2025-12-09T18:25:18Z] E   ModuleNotFoundError: No module named 'lmcache.v1.plugin.plugin_launcher' ```
- `test_suffix_decoding_acceptance` (?)

PR #29038, opened 2025-11-19
[Bug] Fix torch dynamo warning Dynamo detected a call to a `functools.lru_cache`

yewentao256:
## Purpose

```bash
(EngineCore_DP0 pid=2798273) /home/wentao/.venv/lib/python3.12/site-packages/torch/_dynamo/variables/functions.py:1692: UserWarning: Dynamo detected a call to a `functools.lru_cache`-wrapped function. Dynamo ignores the cache wrapper and directly traces the wrapped function. Silent incorrectness is only a *potential* risk, not something we have observed. Enable TORCH_LOGS="+dynamo" for a DEBUG stack trace.
(EngineCore_DP0 pid=2798273)   torch._dynamo.utils.warn_once(msg)
(EngineCore_DP1 pid=2798274) /home/wentao/.venv/lib/python3.12/site-packages/torch/_dynamo/variables/functions.py:1692: UserWarning: Dynamo detected a call to a `functools.lru_cache`-wrapped function. Dynamo ignores the cache wrapper and directly traces the wrapped function. Silent incorrectness is only a *potential* risk, not something we have observed. Enable TORCH_LOGS="+dynamo" for a DEBUG stack trace.
(EngineCore_DP1 pid=2798274)   torch._dynamo.utils.warn_once(msg)
(EngineCore_DP6 pid=2798279) /home/wentao/.venv/lib/python3.12/site-packages/torch/_dynamo/variables/functions.py:1692: UserWarning: Dynamo detected a call to a `functools.lru_cache`-wrapped function. Dynamo ignores the cache wrapper and directly traces the wrapped function. Silent incorrectness is only a *potential* risk, not something we have observed. Enable TORCH_LOGS="+dynamo" for a DEBUG stack trace.
(EngineCore_DP6 pid=2798279)   torch._dynamo.utils.warn_once(msg)
(EngineCore_DP2 pid=2798275) /home/wentao/.venv/lib/python3.12/site-packages/torch/_dynamo/variables/functions.py:1692: UserWarning: Dynamo detected a call to a `functools.lru_cache`-wrapped function. Dynamo ignores the cache wrapper and directly traces the wrapped function. Silent incorrectness is only a *potential* risk, not something we have observed. Enable TORCH_LOGS="+dynamo" for a DEBUG stack trace.
(EngineCore_DP2 pid=2798275)   torch._dynamo.utils.warn_once(msg)
[...]

PR #26197, opened 2025-10-03
[Feature] Enable E8M0 by Default on Hopper for DeepGEMM, 5% E2E throughput improvement

yewentao256:
## Purpose

1. Reduce the complexity
2. Speed up!

### Unit Test

`~/vllm-source/tests/kernels/moe$ pytest test_deepgemm.py `

```bash
collected 12 items                                                                                                                     

test_deepgemm.py ............                                                                                   [100%]

============================================ 12 passed in 96.91s (0:01:36) ============================================
```

```bash
collected 32 items                                                                                                    

test_batched_deepgemm.py ................................                                                       [100%]

============================================ 32 passed in 79.44s (0:01:19) ============================================
```

### Acc

`lm_eval   --model vllm   --model_args "pretrained=Qwen/Qwen3-30B-A3B-FP8,max_model_len=32768,enforce_eager=True"   --trust_remote_code   --tasks gsm8k   --num_fewshot 5   --batch_size auto`

```bash
# now
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8560|±  |0.0097|
|     |       |strict-match    |     5|exact_match|↑  |0.8939|±  |0.0085|

# main
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8097|±  |0.0108|
|     |       |strict-match    |     5|exact_match|↑  |0.8870|±  |0.0087|
```

### Perf

`vllm bench throughput --model Qwen/Qwen3-30B-A3B-FP8 --load-format dummy --input-len 1000 --output-len 100 --trust_remote_code --enable-expert-parallel`

```bash
# now
Throughput: 35.84 requests/s, 39427.04 total tokens/s, 3584.28 output tokens/s
# main
Throughput: 33.89 requests/s, 37275.64 total tokens/s, 3388.69 output tokens/s
```

PR #30336, opened 2025-12-09
[Bugfix] Fix fp8 DeepGemm compilation issues

ElizaWszola:
`is_deep_gemm_e8m0_used()` and `current_platform.is_device_capability()` are not compatible with Dynamo, causing failed compilations. This PR intends to fix this problem.

### Testing
Run inference on `Qwen/Qwen3-30B-A3B-FP8` (one of the models affected) with `VLLM_USE_DEEP_GEMM=1`.

### DIFF
--- a/vllm/model_executor/layers/quantization/utils/fp8_utils.py
+++ b/vllm/model_executor/layers/quantization/utils/fp8_utils.py
@@ -30,7 +30,6 @@ from vllm.model_executor.parameter import (
 from vllm.platforms import current_platform
 from vllm.triton_utils import tl, triton
 from vllm.utils.deep_gemm import (
-    DeepGemmQuantScaleFMT,
     fp8_gemm_nt,
     is_deep_gemm_e8m0_used,
     is_deep_gemm_supported,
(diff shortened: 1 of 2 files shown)


yewentao256 (commented):
Why `is_deepgemm_e8m0` could not be used? If it is because of the `@cache`, I am thinking we should have something like https://github.com/vllm-project/vllm/pull/29038 instead of hardcode for is_blackwell.

bnellnm (approved):

mgoin (approved):

vllm/model_executor/layers/quantization/utils/fp8_utils.py
```diff
@@ -269,7 +269,7 @@ def _run_deepgemm(
         weight: torch.Tensor,
         weight_scale: torch.Tensor,
     ) -> torch.Tensor:
-        if DeepGemmQuantScaleFMT.from_oracle() == DeepGemmQuantScaleFMT.UE8M0:
+        if self.use_deep_gemm_e8m0 and self.is_blackwell:
```
jhaotingc, line 272: https://github.com/vllm-project/vllm/pull/26197

Shouldn't e8m0 also compatible with hopper?

jhaotingc: cc @yewentao256 , thanks!

mgoin: IIUC this is specifically for the case where e8m0 scales need to be *packed*, which is a Blackwell only case

mgoin (approved):

vllm/model_executor/layers/quantization/utils/fp8_utils.py
jhaotingc: Thank you @mgoin for the explanation!

PR #30125, opened 2025-12-05
[CustomOp][MM] Extract MMEncoderAttention as CustomOp and replace the backend of QwenVisionAttention with it.

shen-shanshan:
## Purpose

To avoid maintaining a variety of modeling files in vllm-ascend, we propose to remove all files in `models` dir in vllm-ascend. After this, the only thing a vllm plugin need to do is just registering their custom device-specific OOT ops to vllm when adding a new model. To achieve this, there are some refactors need to be done both in vllm and vllm-ascend, such as extracting some general layers as CustomOp, find more details at https://github.com/vllm-project/vllm-ascend/issues/4084.

Following https://github.com/vllm-project/vllm/pull/27919 and https://github.com/vllm-project/vllm/pull/27147, this PR has unified the getting logic of `vit_attn_backend` and extracted `MMEncoderAttention` as a CustomOp.

To be specific, vision attention backend should only be checked and overwritten in the platform-specific implementation. We should not overwrite this logic in any other places, such as `model_executor/models/<model_name>.py`. In addition, I have moved scattered forward dispatch logic into this CustomOp to avoid verification for `current_platform` in any other places.

To minimize the influence, I only replaced the backend of `QwenVisionAttention` with this CustomOp and have tested this PR both on Ascend A2 NPU and NVIDIA A100 GPU (TODO). I will modify other modeling files and delete the old `MultiHeadAttention` in the future if this PR could be merged.

## Test Plan

- [x] Test this PR together with https://github.com/vllm-project/vllm-ascend/pull/4750 on Ascend A2 NPU.

### ✅ Ascend A2 NPU

Run Qwen2.5-VL:

```bash
vllm serve /root/.cache/modelscope/hub/models/Qwen/Qwen2.5-VL-7B-Instruct \
--max_model_len 16384 \
--max-num-batched-tokens 16384 \
--tensor-parallel-size 2 \
--enforce-eager
```

Output:

```
[...]

PR #30344, opened 2025-12-09
[Bugfix] Fix HunyuanOCR cross-image contamination in batch processing

anker-c2:
This fixes a bug where differently sized images in a batch would have their features / chunks mixed.

The fix processes each image independently one by one through the transformer layers by splitting, processing, and concatenating per layer.

Fixes #30342

## Purpose
- Fix https://github.com/vllm-project/vllm/issues/30342
- Changes the VisionTransformer such that each image in a bad gets processed independent per Layer
- Done like this instead of implementing a mask like discussed in https://github.com/Tencent-Hunyuan/HunyuanOCR/issues/60 for simplicity. This way no change to attention needed. 
- Split / merge kept inside loop for consistency, could be moved outside. 

## Test Plan
- Ran the changes on a batch of ~900 independent images. 
- Same isolated environment reproduced the same issue in the same resulting chunks three times. 
- Additional test on a minimal example would be appreciated, since i was not yet able to reproduce the behavior on public files. 

## Test Result
- I was not able to reproduce the issue after the fix. All inspected chunks were extracted as expected / found no loss in precision
- Speed seems to be slightly improved as a bonus in some cases (depending on the batch configuration) since less attention calculations should be required at times
---
<details>

</details>

### DIFF
--- a/vllm/model_executor/models/hunyuan_vision.py
+++ b/vllm/model_executor/models/hunyuan_vision.py
@@ -502,6 +502,7 @@ class HunYuanVisionTransformer(nn.Module):
         cu_seqlens: list = [0]
 
         hidden_states = x.to(device=self.device, dtype=self.dtype)
+        # embeddings = patch_embeds + patch_pos_embed
         hidden_states = self.embeddings(hidden_states, grid_thw)
 
         for t, h, w in grid_thw:
@@ -515,8 +516,14 @@ class HunYuanVisionTransformer(nn.Module):
 
         hidden_states = hidden_states.reshape(seq_len, -1)
         hidden_states = hidden_states.unsqueeze(0)
-        for layer_num, layer in enumerate(self.layers):
-            hidden_states = layer(hidden_states)
+
+        # build per-image lengths once
+        split_lengths = [int(h) * int(w) for (_, h, w) in grid_thw]
+        for layer in self.layers:
+            # hidden_states: (1, T_total, D)
+            parts = hidden_states.split(split_lengths, dim=1)  # list of (1, L_i, D)
+            parts = [layer(p) for p in parts]
+            hidden_states = torch.cat(parts, dim=1)
 
         # adapter
         split_lengths = (cu_seqlens[1:] - cu_seqlens[:-1]).tolist()


vllm/model_executor/models/hunyuan_vision.py
```diff
@@ -515,8 +516,14 @@ def forward(
...
-        for layer_num, layer in enumerate(self.layers):
-            hidden_states = layer(hidden_states)
+
+        # build per-image lengths once
+        split_lengths = [int(h) * int(w) for (_, h, w) in grid_thw]
+        for layer in self.layers:
+            # hidden_states: (1, T_total, D)
+            parts = hidden_states.split(split_lengths, dim=1)  # list of (1, L_i, D)
+            parts = [layer(p) for p in parts]
+            hidden_states = torch.cat(parts, dim=1)
```
Isotr0py, line 526: Hmmm, IIUC, the root issue is that we need a docment attention mask for HunyuanOCR's ViT, right?
image-29

anker-c2: Right, the root issue seems to be that the image patches get concat together without regard for the size (and therefore no way to separate after) and then get falsely altered across img boundaries during attention. 
A document mask would fix the issue, but i am not sure if it's needed since it would require a lot more changes. 

Imo just processing the images independently per layer like it's done here, instead of building one big attention matrix should achieve the same effect. 

I am not sure about the efficiency of a mask, since i am not well versed in GPU magic: If the attention matrix gets masked, do the extra (useless) computations of the masked fields get optimized away? If so it should be equivalent, but if not shouldn't there be a performance gain computing over multiple (potentionally small) attention matrices depending on the composition of the batch (in regards to the sizes of the included images)?

Either way it should work.

Isotr0py: > If the attention matrix gets masked, do the extra (useless) computations of the masked fields get optimized away?

Yes, at least for FlashAttention, it only calculates blocks with valid attention mask (red and yellow area), and skip calculation for invalid mask (blue area). 

95c1afb773cd933c999b53a73ccc8a7f


> but if not shouldn't there be a performance gain computing over multiple (potentionally small) attention matrices depending on the composition of the batch (in regards to the sizes of the included images)?

Not really, launching CUDA kernel can also have overheads, so if there are lots of images, processing them one by one can cause significant overheads for kernel launching as well.

BTW, we're refactoring `MultiHeadAttention`/`MMEncoderAttention` to support passing attention mask. (https://github.com/vllm-project/vllm/pull/30125/files/626d3bba4f8fec51dafe34c0a1089f4691f44d98..628dbfabfac4e978ccf374a544e841cbfef77359?notification_referrer_id=NT_kwDOAncmpLQyMDk3Mzk3NzMzMDo0MTM2MzEwOA), I think we can fix this issue after that PR through passing correct attention mask. :)

anker-c2: I see, in that case disregard. Thanks a lot for taking the time to explain it! 

Is there a rough guess on the timeline regarding Mask support feature? I can't estimate the current progress. 
Asking in case it might make sense to employ this as a workaround so that the model at least isn't broken before it's being fixed thoroughly with masking. The current behavior is failing silently, which is a bit of a pain. 
Speedwise it's still slightly faster than the current implementation.

Cheers.

vince62s: Hello,
I think this is not the only issue.
The model uses xdrope on the first step=0 pass in the language model.
But this part of the code is broken, it does not support batches:
https://github.com/vllm-project/vllm/blame/main/vllm/transformers_utils/processors/hunyuan_vl.py#L92-L132

Isotr0py (approved):
> Hello,
I think this is not the only issue.
The model uses xdrope on the first step=0 pass in the language model.
But this part of the code is broken, it does not support batches:
https://github.com/vllm-project/vllm/blame/main/vllm/transformers_utils/processors/hunyuan_vl.py#L92-L132

I see, let's merge this one as workaround for now then.

PR #30349, opened 2025-12-09
[BugFix] Fix minimax m2 model rope_parameters

esmeetu:
Currently Minimax-M2 output is garbled, and i found this issue was introduced by #29966.

This PR might be a trick solution, while i checked `https://huggingface.co/MiniMaxAI/MiniMax-M2/blob/main/configuration_minimax_m2.py` that `partial_rotary_factor` should be `0.5`. However, m2 need to set `1.0` to work. It makes me confused.

cc @hmellor

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/models/minimax_m2.py
+++ b/vllm/model_executor/models/minimax_m2.py
@@ -199,6 +199,9 @@ class MiniMaxM2Attention(nn.Module):
             prefix=f"{prefix}.o_proj",
         )
 
+        if rope_parameters is not None:
+            rope_parameters["partial_rotary_factor"] = 1.0
+
         self.rotary_emb = get_rope(
             self.head_dim,
             rotary_dim=rotary_dim,


hmellor (changes requested):
I don't think this is a vLLM bug.

In https://huggingface.co/MiniMaxAI/MiniMax-M2/blob/main/configuration_minimax_m2.py#L188-L189, the `partial_rotary_factor` is set to `self.rotary_dim / self.head_dim`, i.e. `64 / 128 = 0.5`.

vLLM is simply reading this config and using it. This should be fixed in the model checkpoint.

hmellor:
Good catch @rogeryoungh!

Yes I was aware of the two ways this could be done and chose to always scale inside `get_rope` because some models were doing both `A` _and_ `B`, so the `rotary_dim` was being double scaled.

Since most checkpoints do not store, `rotary_dim` in their configs, I'll make a PR to refactor the `rotary_dim` argument out of `get_rope` to avoid confusion in future.

PR #30353, opened 2025-12-09
[Fix] Handle multiple tool calls in Qwen3-MTP tool parser

### LINKED ISSUE #30229: [Bug]: Qwen3-235B-Instruct-MTP fails to parse the response while a chunk includes multiple tokens
### 🐛 Describe the bug

Because the MTP model has an extra token, and when CPU usage is high, the post-processing after FunctionCalling is asynchronous, it can lead to a scenario where multiple tokens appear in a single chunk.
My bad case is  as follows:
'''
<tool_call>\n{"name": "get_nums", "arguments": {"grade": ["一年级", "二年级", "三年级"]}}\n</tool_call>
<tool_call>
\n{"name": "send_email", "arguments": {"userInput": "ALAN的ID是123456，详细信息是：ALAN，中国人，英语
数学很牛"}}\n</tool_call>
<tool_call>
\n{"name": "get_nums", "arguments": {"grade": ["一年级", "二年级", "三年级"]}}\n
</tool_call>
'''
Currently, the community's solutions do not accommodate these kind of scenario.

ArkVex:
## Description
Fixes #30229.

### The Issue
When using Qwen3-MTP (Multi-Token Prediction) or under high load, the model may output multiple `<tool_call>` blocks in a single streaming chunk (e.g., `<tool_call>A</tool_call><tool_call>B</tool_call>`).

The current `StreamingXMLToolCallParser` passes this entire chunk to `xml.parsers.expat`. Since `expat` expects a single root element, it fails when encountering the second `<tool_call>` tag, interpreting it as "junk after document element," causing the tool calls to be dropped or the parser to error out.

### The Fix
I updated `extract_tool_calls` in `Qwen3XMLToolParser` to:
1.  Use `re.finditer` to identify and isolate individual `<tool_call>...</tool_call>` blocks from the `model_output`.
2.  Iteratively parse each isolated block as a separate XML document using the existing streaming parser (resetting state between calls).
3.  Aggregate the results from all blocks into the final `ExtractedToolCallInformation`.

### Verification
This change ensures that chunks containing batched tool calls are parsed correctly. If no tool calls are found via regex (standard text generation), it falls back to the original parsing logic to capture content.

**Credit:** Issue reported and reproduction provided by @PHOEBEMOON0802.


ArkVex:
@njhill need someone for the review plz

PR #28495, opened 2025-11-11
2.9.1 PyTorch release update

atalman:
Update to 2.9.1 PyTorch release and triton 3.5.1

CI Test: https://buildkite.com/vllm/ci/builds/42137

PR #30354, opened 2025-12-09
[WIP][Core] Update PyTorch to 2.9.1 generally

orionr:
## Purpose
Try to upgrade vLLM to PyTorch 2.9.1 generally across CUDA, ROCm, etc.

## Test Plan

Work with the vLLM maintainers to test and certify both from a correctness and performance perspective that things are good with the PyTorch 2.9.1 point release.

## Test Result

N/A currently

cc @khluu  @zhewenl @huydhn @atalman


requirements/test.txt
```diff
@@ -1206,7 +1206,7 @@ transformers==4.57.3
     #   transformers-stream-generator
 transformers-stream-generator==0.0.5
     # via -r requirements/test.in
-triton==3.5.0
+triton==3.5.1
```
orionr, line 1209: Is it okay if we upgrade triton here? This was added as a part of the dependencies generation.

orionr:
Might already be covered by https://github.com/vllm-project/vllm/pull/28495/ - stay tuned

PR #29008, opened 2025-11-19
[ROCm][Quantization] GPT_OSS in amd-quark format model loading and emulations 

xuebwang-amd:
## Purpose

This PR aims for:
- [x] quark model loading, combined with `mxfp4` loading [function ](https://github.com/vllm-project/vllm/blob/256a33ecb4923a4bfa6e1a9adbc69f5255a5aa6d/vllm/model_executor/models/gpt_oss.py#L309) for original openai/gpt-oss-20b & openai/gpt-oss-120b
- [x] OCPMX_W4A16, OCPMX_W4AFP8 MoE scheme and emulation forward, unified into class `QuarkOCP_MX_MoEMethod`

## Test Plan

- Models:
  - [x] GPT_OSS_20B
  - [x] GPT_OSS_120B
- Quantization schemes:
  - [x] W: MXFP4, A: BF16, (optional) KV: FP8
  - [x] W: MXFP4, A: FP8, (optional) KV: FP8
  - [x] W: MXFP4, A: MXFP4, (optional) KV: FP8
- TP:
  - [x] TP1
  - [x] TP2
  - [x] TP4
  - [x] TP8

See results below.

## Test Result
image

## (Sub)-tasks

- [x] unify `mxfp4` format model loading with `quark` format model loading, specifically `FIXME` [here](https://github.com/vllm-project/vllm/blob/23a1946e3b1b2dd45a98c3dfd86556563543c91b/vllm/model_executor/layers/fused_moe/layer.py#L1120)
- [x] fix derived issue due to issues [#30621](https://github.com/vllm-project/vllm/issues/30621) , c.c. @fxmarty-amd @BowenBao : done in commit c4a05ed90c992d4f9062a1e9342e38e3787e8363
- [x] refactor `quark_moe.py` , specially modify `QuarkOCP_MX_MoEMethod` to be compatible with W4A16 `Quark_OCPMX_W4A16_MoEMethod` and `Quark_OCPMX_FP8_MoEMethod` as @fxmarty-amd also pointed out
- [ ] new scheme: `W-MXFP4-A-MXFP8`: _not_ in the PR scope as @BowenBao had clarified https://github.com/vllm-project/vllm/pull/29008#discussion_r2636702801
- [x] `W-FP8-AFP8` quantization for gpt_oss
  - [x] model loading
  - [x] model forward --> fix in PR#31676  https://github.com/vllm-project/vllm/pull/31676
- [x] add a test script
- [x] fix pre-commit issues

## TODO
[...]

PR #30357, opened 2025-12-09
[ROCm][Quantization] GPT OSS Upstream MoE wmxfp4_afp8 with static scales

maleksan85:
# gpt-oss120b-w-mxfp4-a-fp8

    "model": "/data/models/gpt-oss120b-w-mxfp4-a-fp8",
    "prompt": "San Francisco is a",
    "max_tokens": 16,
    "temperature": 0
}'

---

```
Successful requests:                     640
Maximum request concurrency:             64
Benchmark duration (s):                  113.83
Total input tokens:                      655360
Total generated tokens:                  655360
Request throughput (req/s):              5.62
Output token throughput (tok/s):         5757.52
Peak output token throughput (tok/s):    6464.00
Peak concurrent requests:                128.00
Total token throughput (tok/s):          11515.03
Mean TTFT (ms):                          249.81
Median TTFT (ms):                        214.82
P99 TTFT (ms):                           692.63
Mean TPOT (ms):                          10.88
Median TPOT (ms):                        10.90
P99 TPOT (ms):                           11.23
Mean ITL (ms):                           10.88
Median ITL (ms):                         10.52
P99 ITL (ms):                            11.14
Mean E2EL (ms):                          11378.74
Median E2EL (ms):                        11378.59
P99 E2EL (ms):                           11614.36
```
---

```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.912|±  |0.0180|
|     |       |strict-match    |     5|exact_match|↑  |0.180|±  |0.0243|
```
---
# Regression check  gpt-oss-120b

"model": "/data/models/gpt-oss-120b",
"prompt": "San Francisco is a",
"max_tokens": 16,
"temperature": 0
}'


vllm/_aiter_ops.py
```diff
@@ -708,6 +708,7 @@ class rocm_aiter_ops:
     _LINEAR_ENABLED = envs.VLLM_ROCM_USE_AITER_LINEAR
     _RMSNORM_ENABLED = envs.VLLM_ROCM_USE_AITER_RMSNORM
     _FMOE_ENABLED = envs.VLLM_ROCM_USE_AITER_MOE
+    _CK_MXFP4_MOE = envs.VLLM_ROCM_USE_CK_MXFP4_MOE
```
tjtanaa: Please reuse `VLLM_ROCM_USE_AITER_MOE` for this feature and not introduce another FLAG `VLLM_ROCM_USE_CK_MXFP4_MOE` . It should be dispatched automatically based on the model config.

vllm/model_executor/models/gpt_oss.py
```diff
@@ -496,6 +497,274 @@ def _load_weights_mxfp4(
...
+                    # (only load on rank 0 to avoid duplication)
+                    if tp_rank != 0:
+                        loaded_weight.zero_()
+                weight_loader(param,
+                              loaded_weight,
+                              weight_name=name,
+                              shard_id=None,
+                              expert_id=expert_id)
+                loaded_params.add(name)
+                continue
```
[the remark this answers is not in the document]
maleksan85: https://github.com/vllm-project/vllm/pull/29008 is for load refactoring

vllm/_aiter_ops.py
```diff
@@ -831,6 +831,7 @@ class rocm_aiter_ops:
...
```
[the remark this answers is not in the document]
maleksan85: removed

vllm/_aiter_ops.py
maleksan85: thanks, cleaned

vllm/envs.py
```diff
@@ -115,6 +115,7 @@
     VLLM_ROCM_USE_AITER_PAGED_ATTN: bool = False
     VLLM_ROCM_USE_AITER_LINEAR: bool = True
     VLLM_ROCM_USE_AITER_MOE: bool = True
+    VLLM_ROCM_USE_AITER_FUSED_MOE_A16W4: bool = True
```
tjtanaa: Let use the VLLM_ROCM_USE_AITER_MOE flag. The conditions whether to use the a16w4 should be determined based in the quantization config.

maleksan85: Removed option VLLM_ROCM_USE_AITER_FUSED_MOE_A16W4 for now as per discsussion.

vllm/model_executor/layers/quantization/quark/quark_moe.py
```diff
@@ -48,7 +50,13 @@
 
 logger = init_logger(__name__)
 
-__all__ = ["QuarkMoEMethod", "QuarkW8A8Fp8MoEMethod", "QuarkOCP_MX_MoEMethod"]
+__all__ = [
+    "QuarkMoEMethod",
+    "QuarkW8A8Fp8MoEMethod",
+    "QuarkW4MXFp4MoEMethod_OSS",
+    "QuarkW4MXFp4MoEMethod",
```
BowenBao: No need to add these two. Both are supported under `QuarkOCP_MX_MoEMethod`.

Both wmxfp4_amxfp4 and wmxfp4_afp8 are covered under `QuarkOCP_MX_MoEMethod`.

vllm/model_executor/layers/quantization/quark/quark_moe.py
BowenBao: For wmxfp4_afp8, just need to add the triton kernel dispatch at here.

If there are some triton kernel specific setups needed like pre-shuffling, kwargs setting etc, please do it in process_weights_after_loading.

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
```diff
@@ -299,6 +299,182 @@ def triton_kernel_fused_experts(
...
+            w2,
+            quant_config.w2_bias,
+            routing_data,
+            scatter_indx=scatter_indx,
+            precision_config=quant_config.w2_precision,
+            gammas=None if apply_router_weight_on_input else gammas,
+            y=output_tensor,
+        )
+
+    elif quant_config.use_mxfp4_w4a4:
```
BowenBao: why is it `use_mxfp4_w4a4` in config but `a8w4` kernel

BowenBao (changes requested):

vllm/model_executor/layers/quantization/quark/quark_moe.py
maleksan85: inherited QuarkOCP_MX_MoEMethod_OSS from QuarkOCP_MX_MoEMethod this achieves better reusability and less branching.

vllm/model_executor/layers/quantization/quark/quark_moe.py
maleksan85: yep, thanks, removed `QuarkW4MXFp4MoEMethod` as it shows same perf

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
maleksan85: there is `downcast_to_static_fp8`, i.e. conversion from afp4 to afp8

vllm/model_executor/layers/quantization/quark/quark_moe.py
```diff
@@ -1031,3 +1047,133 @@ def apply(
                 expert_map=layer.expert_map,
                 quant_config=self.moe_quant_config,
             )
+
+
+class QuarkOCP_MX_MoEMethod_OSS(QuarkOCP_MX_MoEMethod):
```
BowenBao, line 1050: I think this can be merged with `QuarkOCP_MX_MoEMethod`. But probably it's okay to keep separate in this PR.

There will be a larger refactor to abstract out kernel dependent logic for `process_weights_after_loading` and align with modular kernel refactors. That will do the merge anyways.

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
BowenBao: ok ithink we should add a flag `use_mxfp4_w4a8` and use that, 'afp4' was not involved in this case, the inputs are fp16 before we quant to fp8

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
```diff
@@ -304,6 +304,182 @@ def triton_kernel_fused_experts(
...
+
+        assert quant_config.w1_precision is not None, (
+            "w1_precision in quant config can't be None"
+        )
+        assert quant_config.w2_precision is not None, (
+            "w2_precision in quant config can't be None"
+        )
+
+        hidden_states = downcast_to_static_fp8(
+            hidden_states, quant_config.w1_precision.flex_ctx.lhs_data.scale
```
BowenBao: ah i see how input scale is passed, through within w1_precision.

I think by using `mxfp4_w4a8_moe_quant_config`, we can use `quant_config.a1_scale` here instead

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -609,6 +611,7 @@ def _get_quant_method() -> FusedMoEMethodBase:
         moe_quant_params = {
             "num_experts": self.local_num_experts,
             "hidden_size": hidden_size,
+            "unpadded_hidden_size": unpadded_hidden_size,
```
BowenBao, line 615: fyi: no action for this pr, i have a PR #34285 refactoring how hidden_size is padded. we can land this first and I'll pull in the update.

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -1060,16 +1063,43 @@ def weight_loader(
...
+                        shard_dim, shard_size * self.tp_rank, shard_size
+                    )
+                if "bias" in weight_name:
+                    dim1 = loaded_weight.shape[0]
+                    expert_data.data[:dim1].copy_(loaded_weight)
+                else:
+                    dim1 = loaded_weight.shape[0]
+                    dim2 = loaded_weight.shape[1]
+                    expert_data.data[:dim1, :dim2].copy_(loaded_weight)
+                return True if return_success else None
```
BowenBao: are these needed? I thought Xuebin's earlier PR that enables w4a8 emulation should support all the necessary scales loading?

vllm/model_executor/layers/quantization/quark/quark_moe.py
```diff
@@ -71,14 +77,31 @@ def get_moe_method(
...
         elif quant_config._is_fp8_w8a8(weight_config, input_config):
             return QuarkW8A8Fp8MoEMethod(weight_config, input_config, module.moe_config)
         elif quant_config._is_w_ocp_mx_a_x(weight_config, input_config):
-            return QuarkOCP_MX_MoEMethod(weight_config, input_config, module.moe_config)
+            from vllm.config import get_current_vllm_config
+
+            model_type = getattr(
+                get_current_vllm_config().model_config.hf_config, "model_type", None
+            )
+            if model_type == "gpt_oss" and not emulate:
```
BowenBao: best to avoid conditioning on model type, ideally something like `if self.ocp_mx_scheme == "w_mxfp4_a_fp8" and not emulate:`

vllm/model_executor/layers/quantization/quark/quark_moe.py
```diff
@@ -1031,3 +1047,133 @@ def apply(
...
+            )
+
+    def get_fused_moe_quant_config(
+        self, layer: torch.nn.Module
+    ) -> FusedMoEQuantConfig | None:
+        w1_scale = self.w13_precision_config
+        w2_scale = self.w2_precision_config
+
+        # TODO: how to set scale?
+        return mxfp4_w4a4_moe_quant_config(
```
BowenBao: as commented elsewhere, suggest return `mxfp4_w4a8_moe_quant_config`

vllm/model_executor/layers/fused_moe/layer.py
maleksan85: thanks!

vllm/model_executor/layers/fused_moe/layer.py
maleksan85: thank you for pointing out! removed!

vllm/model_executor/layers/quantization/quark/quark_moe.py
maleksan85: also don't like model specific things, however this PR in scope of upstreaming which used to be `355_wip` branch, so kept logic close to original. Anyway your option is better

vllm/model_executor/layers/quantization/quark/quark_moe.py
maleksan85: well I tried to put into base class, however it becomes pretty messy with conditions, so easier to reuse common and extend with specifics.

vllm/model_executor/layers/quantization/quark/quark_moe.py
```diff
@@ -71,14 +76,30 @@ def get_moe_method(
...
+
         if quant_config._is_fp8_w4a8(weight_config, input_config):
             return QuarkW4A8Fp8MoEMethod(weight_config, input_config, module.moe_config)
         elif quant_config._is_fp8_w8a8(weight_config, input_config):
             return QuarkW8A8Fp8MoEMethod(weight_config, input_config, module.moe_config)
         elif quant_config._is_w_ocp_mx_a_x(weight_config, input_config):
-            return QuarkOCP_MX_MoEMethod(weight_config, input_config, module.moe_config)
+            emulate = not current_platform.supports_mx() or not (
+                rocm_aiter_ops.is_fused_moe_enabled()
+            )
```
BowenBao, line 90: nit: for `QuarkOCP_MX_MoEMethod_OSS` purpose should we just check `emulate = not rocm_aiter_ops.is_fused_moe_enabled()`? IIUC both mxfp4_w4a16 and mxfp4_w4a8 kernels could run on mi300x

BowenBao (approved):
LGTM, thank you!

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
```diff
@@ -304,6 +304,182 @@ def triton_kernel_fused_experts(
     return output_tensor
 
 
+def triton_kernel_moe_oss_forward(
```
Rohan138: can we unify this with [`triton_kernel_moe_forward`](https://github.com/ROCm/vllm/blob/0b21db603eae8077c9a25ce00cb2af922fae19a9/vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py#L169)?

Currently for this function, we always have `quant_config.use_mxfp4_w4a8 = True`; but the w4a16 section here would actually fail on our current Docker images with Triton < 3.6 because it's not calling `legacy_routing`.

vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py
```diff
@@ -304,6 +304,182 @@ def triton_kernel_fused_experts(
...
+        expert_map=expert_map,
+        unpadded_N_w1=unpadded_N_w1,
+        unpadded_K_w1=unpadded_K_w1,
+        unpadded_N_w2=unpadded_N_w2,
+        unpadded_K_w2=unpadded_K_w2,
+    )
+
+
+# This is a triton implementation of the fused_experts function
+def triton_kernel_fused_oss_experts(
```
Rohan138: Same comment here, can we reuse [triton_kernel_fused_experts](https://github.com/ROCm/vllm/blob/0b21db603eae8077c9a25ce00cb2af922fae19a9/vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py#L206) and add an `elif quant_config.use_mxfp4_w4a8 and rocm_aiter_ops.is_enabled()` for aiter triton with w4a8 moe?

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -525,16 +525,18 @@ def __init__(
 
         # Round up hidden size before creating moe_config.
         # This way moe_config is created with the correct hidden_size from the start.
+        unpadded_hidden_size = hidden_size
```
Rohan138, line 528: @BowenBao #34285 is removing the `maybe_roundup_hidden_size` from this file entirely ... which PR should go in first?

vllm/model_executor/layers/quantization/quark/quark_moe.py
BowenBao: synced offline, thanks @maleksan85 for clarifying these kernels indeed requires gfx950.

gshtras (approved):

vllm/model_executor/layers/fused_moe/layer.py
BowenBao: plan to land this one first and rebase #34285

PR #30364, opened 2025-12-09
[Bugfix] awq_gemm: fix argument order swap

mgehre-amd:
Swap the order of `scales`, `zeros` arguments in `_custom_ops.awq_gemm` to be consistent with
- how the caller passes those
- how the callees (`awq_gemm_triton`, `awq_gemm`) expect those
- how CUDA [awq_gemm](https://github.com/vllm-project/vllm/blob/9db78f34dce03d149f3571d45a2d2f259bdc7d15/csrc/ops.h#L191) orders those

Due to the double swap in argument declaration and passing to callees, the functionality was correct but it made the code harder to read.

## Test Plan
On gfx1151:
`pytest tests/kernels/quantization/test_awq_triton.py tests/models/quantization/test_awq.py`

## Test Result
Passed


yewentao256 (approved):
LGTM, thanks for the work!

yewentao256 (commented):
Let's try again before force merge

PR #30093, opened 2025-12-04
fix#30092 Kimi-Linear model loading failure with missing indexer_rotary_emb

baonudesifeizhai:
## Purpose
fix #30092 
## Test Plan
vllm serve moonshotai/Kimi-Linear-48B-A3B-Instruct \
  --port 8000 \
  --tensor-parallel-size 8 \
  --max-model-len 1048576 \
  --trust-remote-code \
  --gpu-memory-utilization 0.95  
## Test Result
image

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #30366, opened 2025-12-10
[Bug Fix] Fix Kimi-Linear model initialization crash due to missing 'indexer_rotary_emb' arg

yonasTMC:
### Description
This PR fixes a regression in `vllm/model_executor/models/kimi_linear.py` where the initialization of `MLAModules` was missing the required `indexer_rotary_emb` argument. This caused the engine to crash immediately upon loading the `moonshotai/Kimi-Linear-48B-A3B-Instruct` model.

### Issue
The `MLAModules` class signature was recently updated to require `indexer_rotary_emb`, but the Kimi model definition was not updated to reflect this change.

**Error Log:**
> TypeError: MLAModules.__init__() missing 1 required positional argument: 'indexer_rotary_emb'

### Fix
Added `indexer_rotary_emb=None` to the `MLAModules` initialization call in `kimi_linear.py`.

### Verification
- Tested locally with `moonshotai/Kimi-Linear-48B-A3B-Instruct`.
- Verified that the server starts up successfully and the error is resolved.

### DIFF
--- a/vllm/model_executor/models/kimi_linear.py
+++ b/vllm/model_executor/models/kimi_linear.py
@@ -248,6 +248,7 @@ class KimiMLAAttention(nn.Module):
             kv_a_layernorm=self.kv_a_layernorm,
             kv_b_proj=self.kv_b_proj,
             rotary_emb=None,
+            indexer_rotary_emb=None,
             o_proj=self.o_proj,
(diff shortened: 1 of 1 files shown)


jeejeelee:
I think this bug has been fixed by #30093

PR #30373, opened 2025-12-10
Implement LMDB-based multi-modal cache

petersalas:
## Purpose

This implements an LMDB-based multi-modal item cache which supports LRU-based eviction, multiple API server workers, and/or multiple Engine processes.

- `BaseMultiModalProcessorCache`/`BaseMultiModalReceiverCache` now include a `begin()` method which (for the LMDB implementation) uses a transaction. In the sender-side cache, writes are queued outside of the transaction scope so that processing/serialization all occur outside of the scope of the write transaction (LMDB has single writer semantics).
- Objects are split into to ~4KB chunks to avoid pathological LMDB free list fragmentation (at the expense of cache locality and copy overhead).
- A single engine process effectively owns the cache: it starts an evictor process and handles resets. The evictor starts operating at 50% utilization and ramps up its aggressiveness (as measured by % of time holding the write lock) as the utilization approaches 100%.

One caveat with this implementation is that any item that hasn't been used within a fixed time window (`--mm-lmdb-cache-min-eviction-age`, defaults to 600 seconds) may be evicted, even if there are queued requests still depending on those items (the worker's `execute_model` will raise in that case). A future improvement could be to instead track the oldest active request in each of the frontends (`OutputProcessor` seemingly already has this) and use that instead.

## Test Plan

Tested across permutations of API/Tensor/Data parallelism.

(Happy to run any suggested benchmarks as well!)

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/envs.py
+++ b/vllm/envs.py
@@ -82,6 +82,7 @@ if TYPE_CHECKING:
     VLLM_MAX_IMAGE_PIXELS: int = 178_956_970
     VLLM_VIDEO_LOADER_BACKEND: str = "opencv"
     VLLM_MEDIA_CONNECTOR: str = "http"
+    VLLM_MM_LMDB_CACHE_ID: str | None = None
     VLLM_MM_HASHER_ALGORITHM: str = "blake3"
     VLLM_TARGET_DEVICE: str = "cuda"
     VLLM_MAIN_CUDA_VERSION: str = "13.0"
@@ -995,6 +996,9 @@ environment_variables: dict[str, Callable[[], Any]] = {
     # imported at runtime.
     # If a non-existing backend is used, an AssertionError will be thrown.
     "VLLM_MEDIA_CONNECTOR": lambda: os.getenv("VLLM_MEDIA_CONNECTOR", "http"),
+    # The LMDB multimodal cache ID to use.
+    # N.B. This is automatically assigned on configuration load if not set.
+    "VLLM_MM_LMDB_CACHE_ID": lambda: os.getenv("VLLM_MM_LMDB_CACHE_ID"),
     # Hash algorithm for multimodal content hashing.
     # - "blake3": Default, fast cryptographic hash (not FIPS 140-3 compliant)
     # - "sha256": FIPS 140-3 compliant, widely supported
@@ -2133,6 +2137,7 @@ def compile_factors() -> dict[str, object]:
         "VLLM_OBJECT_STORAGE_SHM_BUFFER_NAME",
         "VLLM_ASSETS_CACHE",
         "VLLM_ASSETS_CACHE_MODEL_CLEAN",
+        "VLLM_MM_LMDB_CACHE_ID",
         "VLLM_WORKER_MULTIPROC_METHOD",
         "VLLM_ENABLE_V1_MULTIPROCESSING",
         "VLLM_V1_OUTPUT_PROC_CHUNK_SIZE",
--- a/vllm/multimodal/cache.py
+++ b/vllm/multimodal/cache.py
@@ -1,9 +1,11 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+import contextlib
 import operator
 import sys
 from abc import ABC, abstractmethod
 from collections.abc import Mapping, Sequence
+from contextlib import AbstractContextManager
 from multiprocessing.synchronize import Lock as LockType
 from typing import TYPE_CHECKING, Generic, TypeAlias, TypeVar, cast
 
@@ -17,6 +19,11 @@ from vllm.distributed.device_communicators.shm_object_storage import (
     SingleWriterShmRingBuffer,
 )
 from vllm.logger import init_logger
+from vllm.multimodal.lmdb_cache import (
+    LmdbMultiModalCache,
+    LmdbReadTransaction,
+    LmdbWriteTransaction,
+)
 from vllm.utils.cache import CacheInfo, LRUCache
 from vllm.utils.jsontree import json_count_leaves, json_map_leaves, json_reduce_leaves
 from vllm.utils.mem_constants import GiB_bytes, MiB_bytes
@@ -263,6 +270,10 @@ class BaseMultiModalProcessorCache(
     BaseMultiModalCache[MultiModalProcessorCacheInItem, MultiModalProcessorCacheOutItem]
 ):
     """The required interface for caches on P0."""
+
+    def begin(self) -> "AbstractContextManager[BaseMultiModalProcessorCache]":
+        """Context manager for batch cache operations."""
+        return contextlib.nullcontext(self)
 
     @abstractmethod
     def is_cached_item(self, mm_hash: str) -> bool:
@@ -586,6 +597,10 @@ class BaseMultiModalReceiverCache(
 ):
     """The required interface for caches on P1."""
 
+    def begin(self) -> "AbstractContextManager[BaseMultiModalReceiverCache]":
+        """Context manager for batch cache operations."""
+        return contextlib.nullcontext(self)
+
     def get_and_update_features(
         self,
         mm_features: list["MultiModalFeatureSpec"],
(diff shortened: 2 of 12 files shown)


vllm/multimodal/lmdb_cache.py
```diff
@@ -0,0 +1,674 @@
...
+    ) -> None:
+        super().__init__()
+        os.makedirs(cache_dir, exist_ok=True)
+
+        self._cache_dir = cache_dir
+        self._cache_size = cache_size
+        self._min_eviction_age = min_eviction_age
+        self._max_object_size = max_object_size
+
+        with LmdbMultiModalCache.OPEN_ENVS_LOCK:
```
DarkLight1337: Can replace `LmdbMultiModalCache` with `self` for simplicity

vllm/multimodal/lmdb_cache.py
```diff
@@ -0,0 +1,674 @@
...
+from vllm.distributed.device_communicators.shm_object_storage import MsgpackSerde
+from vllm.logger import init_logger
+from vllm.multimodal.processing import ResolvedPromptUpdate
+from vllm.utils.mem_constants import GiB_bytes, MiB_bytes
+from vllm.utils.system_utils import decorate_logs, get_mp_context, set_process_title
+
+from .inputs import MultiModalBatchedField, MultiModalFieldElem, MultiModalKwargsItem
+
+if TYPE_CHECKING:
+    from multiprocessing.synchronize import Event
```
DarkLight1337, line 28: Why does this need to be put under `TYPE_CHECKING`?

vllm/multimodal/lmdb_cache.py
```diff
@@ -0,0 +1,674 @@
...
+        with LmdbMultiModalCache.OPEN_ENVS_LOCK:
+            # A given LMDB environment can only be opened once per process,
+            # and not across forks.
+            if existing_env := LmdbMultiModalCache.OPEN_ENVS.get(cache_dir):
+                logger.debug("Reusing existing LMDB environment at %s", cache_dir)
+                self.lmdb_env = existing_env
+            else:
+                self.lmdb_env = lmdb.Environment(
+                    path=cache_dir,
+                    map_size=max(self._cache_size * 2, GiB_bytes),
```
DarkLight1337: This seems a bit arbitrary, can you add a comment explaining why this size is chosen?

vllm/multimodal/lmdb_cache.py
```diff
@@ -0,0 +1,674 @@
...
+
+    from .cache import MultiModalProcessorCacheInItem, MultiModalProcessorCacheOutItem
+
+logger = init_logger(__name__)
+
+
+class LmdbMultiModalCache:
+    """LMDB-based multi-modal processor cache."""
+
+    VLLM_LMDB_CACHE_ID_ENV = "VLLM_LMDB_CACHE_ID"
```
DarkLight1337: Add this to `vllm.envs`

vllm/multimodal/lmdb_cache.py
```diff
@@ -0,0 +1,674 @@
...
+
+        return cache_id
+
+    @classmethod
+    def from_vllm_config(cls, vllm_config: "VllmConfig") -> "LmdbMultiModalCache":
+        # The cache ID must be set by now.
+        cache_id = os.environ[cls.VLLM_LMDB_CACHE_ID_ENV]
+
+        mm_config = cast(
+            "MultiModalConfig", vllm_config.model_config.get_multimodal_config()
```
DarkLight1337: Why is `cast` necessary here?

vllm/v1/serial_utils.py
```diff
@@ -164,9 +164,10 @@ def enc_hook(self, obj: Any) -> Any:
 
         if isinstance(obj, slice):
             # We are assuming only int-based values will be used here.
-            return tuple(
-                int(v) if v is not None else None
-                for v in (obj.start, obj.stop, obj.step)
+            return (
```
DarkLight1337, line 201: Seems this is functionally the same as the original code?

vllm/multimodal/lmdb_cache.py
```diff
@@ -0,0 +1,674 @@
...
+        while True:
+            if os.getppid() == 1:
+                # Parent process has exited.
+                break
+
+            if got_signal or time.monotonic() >= next_reader_check:
+                stale_readers = cache.lmdb_env.reader_check()
+                if stale_readers > 0:
+                    logger.warning("Removed %d stale LMDB readers.", stale_readers)
+                next_reader_check = time.monotonic() + 60.0
```
DarkLight1337: Use a constant instead of magic number

DarkLight1337 (commented):
Some very initial comments

vllm/multimodal/lmdb_cache.py
```diff
@@ -0,0 +1,674 @@
...
+                        while evicted_items < batch_size:
+                            combined_key = cursor.key()
+                            if not combined_key:
+                                # No more items to evict.
+                                break
+
+                            timestamp = self.bytes2int(combined_key[0 : self.INT_SIZE])
+
+                            # Don't evict any items newer than the minimum eviction age.
+                            if current_timestamp - timestamp < self._min_eviction_age:
```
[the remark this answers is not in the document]
petersalas: Switched to `time.time`, although in practice this isn't really intended for cross-instance usage (in particular because the cache is currently always cleared on startup).

vllm/multimodal/lmdb_cache.py
petersalas: The correct way to instantiate an event is `multiprocessing.Event()` but that is in fact a function not a type. So I put it in `TYPE_CHECKING` to prevent instantiation (though to be fair, there's no instantiation in this file anyway).

vllm/multimodal/lmdb_cache.py
petersalas: It's not really, I did it for autocomplete purposes because for some reason (my?) VS Code doesn't recognize `VllmConfig` properly. Removed!

vllm/v1/serial_utils.py
petersalas: Yup -- at one point over the course of this work this line showed up in a profile as a hot spot, and it was easy to reduce.

```
>>> import timeit
>>> obj = slice(0, 10, 1)
>>> timeit.timeit(lambda: tuple(int(v) if v is not None else None for v in (obj.start, obj.stop, obj.step)))
1.0410025781020522
>>> timeit.timeit(lambda: (int(obj.start) if obj.start is not None else None, int(obj.stop) if obj.stop is not None else None, int(obj.step) if obj.step is not None else None))
0.473489907104522
```

The actual problem I was debugging turned out to be GC-related, but it also seemed like an easy, if small, win. But I'm happy to revert if you'd prefer :)

DarkLight1337: Let's use your optimized version then, thanks!

vllm/multimodal/lmdb_cache.py
```diff
@@ -0,0 +1,657 @@
...
+                    max_dbs=4,
+                    writemap=True,
+                    map_async=True,
+                )
+
+                self.OPEN_ENVS[cache_dir] = self.lmdb_env
+
+                # Ensure that the LMDB environment is closed/removed on fork.
+                os.register_at_fork(
+                    after_in_child=lambda: self.OPEN_ENVS.pop(cache_dir).close()
```
petersalas: Replace this with a single registration that clears the dictionary (otherwise the same cache_dir can be registered multiple times in successive forks).

vllm/multimodal/processing.py
```diff
@@ -1899,55 +1897,56 @@ def _cached_apply_hf_processor(
...
-            prompt_updates=mm_prompt_updates,
-        )
+            mm_info = MultiModalProcessingInfo(
+                kwargs=mm_kwargs,
+                hashes=mm_hashes,
+                prompt_updates=mm_prompt_updates,
+            )
 
-        return prompt_ids, mm_info, is_update_applied
+            return prompt_ids, mm_info, is_update_applied
```
DarkLight1337: This part can be outside the transaction

vllm/envs.py
```diff
@@ -1563,6 +1564,9 @@ def get_vllm_port() -> int | None:
     "VLLM_USE_V2_MODEL_RUNNER": lambda: bool(
         int(os.getenv("VLLM_USE_V2_MODEL_RUNNER", "0"))
     ),
+    # The LMDB multimodal cache ID to use.
```
DarkLight1337: Would be nice to put this with the other multimodal related envs

vllm/multimodal/lmdb_cache.py
```diff
@@ -0,0 +1,657 @@
...
+    ) -> None:
+        super().__init__()
+        os.makedirs(cache_dir, exist_ok=True)
+
+        self._cache_dir = cache_dir
+        self._cache_size = cache_size
+        self._min_eviction_age = min_eviction_age
+        self._max_object_size = max_object_size
+
+        with self.OPEN_ENVS_LOCK:
```
DarkLight1337: This part seems relatively self.contained, maybe we can extract the part of getting a `lmdb_env` into a global function which would enable unit tests for it.

petersalas: Good call!

vllm/multimodal/lmdb_cache.py
```diff
@@ -0,0 +1,702 @@
...
+    def is_cached_item(self, mm_hash: str) -> bool:
+        mm_hash_key = mm_hash.encode("utf-8")
+
+        return (
+            self._read_txn.get(mm_hash_key, db=self._cache.hash_to_timestamp)
+            is not None
+        )
+
+    def get_and_update_item(
+        self, modality: str, mm_item: "MultiModalProcessorCacheInItem", mm_hash: str
```
reaganjlee: > get_and_update_item now includes modality to spare the SHM/LMDB implementations from also needing to cache the modality. (Every caller already has it.)

We only need the information for the SHM implementation right? Doesn't look like LMDB uses the modality info

tests/multimodal/test_cache.py
```diff
@@ -276,7 +283,7 @@ def _run_test_cache_eviction_lru(
         item_tuple = (request1_items[h], []) if not is_cached else None
         print(f"Request 1: key={h} | cached={is_cached}")
 
-        p0_cache.get_and_update_item(item_tuple, h)
+        p0_cache.get_and_update_item(h, item_tuple, h)
```
reaganjlee: I don't understand the purpose of this, unless the purpose is just as a placeholder? If so, maybe `"fake_modality"` could be more clear

vllm/multimodal/lmdb_cache.py
petersalas: As of #33331 neither of them needs it! :)

tests/multimodal/test_cache.py
petersalas: Elsewhere the hash was used as the modality, but yet another piece of awkwardness fixed by #33331!

vllm/multimodal/cache.py
```diff
@@ -723,3 +738,194 @@ def touch_receiver_cache_item(
...
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self._txn.__exit__(exc_type, exc_value, traceback)
+
+    @override
+    def is_cached_item(self, mm_hash: str) -> bool:
+        return self._txn.is_cached_item(mm_hash)
+
+    @override
+    def get_and_update_item(self, mm_item, mm_hash) -> MultiModalProcessorCacheOutItem:
```
DarkLight1337: Can you add type annotations?

DarkLight1337 (commented):
Asked @ywang96 to take a look since I did a pass before already

ywang96:
Bumping this so it doesn't get stale - I actually still very much want it in but will ask folks to clean it up. (Thanks @petersalas for your patience and we will take it over from here)

cc @BugenZhao on your thoughts on this feature!

petersalas:
@ywang96 SGTM, feel free to tag me if you guys have any questions

gty111:
@petersalas Hi Peter, could you please resolve the conflicts and rebase your branch onto the latest main? Alternatively, a merge would also work. Thanks!

gty111:
@petersalas  Thanks for the great work! Regarding the eviction caveat (items may be evicted while queued requests still depend on them, causing execute_model to raise), I think we should address this before merging since it can crash the whole engine (not just fail one request), and it's most likely exactly in the batch / API-server scale-out scenarios this feature targets.

Proposed approach:

Oldest-active-request watermark (primary fix). Each frontend publishes the arrival_time of its oldest active request (OutputProcessor already tracks this) to a shared slot; the evictor takes the global min and only evicts items whose last_used < min_active_arrival. This is provably safe (any item an active request depends on was touched at/after that request's arrival, so last_used >= min_active_arrival) and cheap — O(1) per frontend per step, no per-item refcounting write contention. Add a heartbeat so a dead/stalled frontend can't pin the cache forever, with min_eviction_age/a large max_eviction_age kept as a safety floor.

Graceful fallback on receiver miss (defense-in-depth). Catch the miss in _apply_mm_cache and fail just that request (or trigger a resend/recompute) instead of letting it propagate and take down the engine.

Happy to take a stab at implementing this if that sounds good.

petersalas:
@gty111 Yup, that is sensible and sounds very much like what I had prototyped before. The main reason I didn’t include such an approach in this PR was to minimize integration sprawl on an approach I wanted to vet 😀 (And also our scenario is exclusively realtime audio inference so our requests time out long before the cache gets under eviction pressure.)

Burhan1189 (approved):

PR #25084, opened 2025-09-17
[Core] Refactor self.model() to call a helper for subclassing.

patrick-toulme:
## Purpose
Meta MTIA subclasses GPUModelRunner. We need to be able to replace the self.model() call with some of our custom logic. This very small PR just refactors the execute_model call into self.model() to use a helper method called self.model_forward. Subclasses then can replace that method as they want.

Signed-off-by: Patrick C. Toulme <ptoulme@meta.com>

## Test Plan
None as this is refactoring code.

## Test Result
None
---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x] The test plan, such as providing test command.
- [x] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #30385, opened 2025-12-10
[Core] Whisper support `torch.compile`

NickLucche:
This PR is yet another Whisper performance optimization, adding support for torch.compile during *decoding step*.
It follows a very similar approach to https://github.com/vllm-project/vllm/pull/30072 (and should also land after that to ensure best defaults) in which only the 2nd decoder steps onward are compiled.
This is due to the fact that step0 in enc-dec models computes and caches crossattn KVs, requiring encoder_output as additional input and hence generating a different graph from the other steps.

Updated profiling:
image

## Considerations

I've attempted to add the "2nd decoder step" selection logic directly in the model runner (`_model_forward`).   
I am aware that `_model_forward` is currently used by OOT runners (https://github.com/vllm-project/vllm/pull/25084), although no "official" runner interface contract is maintained to ensure compatibility (unlike connectors to name one), which makes maintaining these kinds of methods without breaking external usage quite hairy.
As this change may affect those runners I am also pinging @patrick-toulme .

Happy to change to a less invasive logic if we find a cleaner way to do it @LucasWilkinson . 
Other options are adding the flag to attn_metadata and then retrieving the metadata from the support_torch_compile wrapper OR hacking the WhisperDecoder `__call__` (definitely not nice).

UPDATE: 
Following @ProExpertProg suggestion, I've moved to implementing the alternative option in which the `skip_compiled` logic is generically handled inside the compile decorator.
So no change to `_model_forward` is actually needed.

Related PRs https://github.com/vllm-project/vllm/pull/29421 https://github.com/vllm-project/vllm/pull/30072

### Test with

Compilation is enabled by default:
```
vllm serve openai/whisper-large-v3-turbo
```

cc @DarkLight1337  @robertgshaw2-redhat

### DIFF
--- a/tests/entrypoints/openai/correctness/test_transcription_api_correctness.py
+++ b/tests/entrypoints/openai/correctness/test_transcription_api_correctness.py
@@ -156,7 +156,9 @@ def test_wer_correctness(
     model_name, dataset_repo, expected_wer, n_examples=-1, max_concurrent_request=None
 ):
     # TODO refactor to use `ASRDataset`
-    with RemoteOpenAIServer(model_name, ["--enforce-eager"]) as remote_server:
+    with RemoteOpenAIServer(
+        model_name, ["--enforce-eager"], max_wait_seconds=480
+    ) as remote_server:
         dataset = load_hf_dataset(dataset_repo)
 
         if not max_concurrent_request:
--- a/vllm/model_executor/models/whisper.py
+++ b/vllm/model_executor/models/whisper.py
@@ -22,6 +22,7 @@ from vllm.attention.layer import Attention
 from vllm.compilation.decorators import support_torch_compile
 from vllm.config import CacheConfig, ModelConfig, SpeechToTextConfig, VllmConfig
 from vllm.config.multimodal import BaseDummyOptions
+from vllm.distributed import get_tensor_model_parallel_world_size
 from vllm.inputs.data import PromptType
 from vllm.logger import init_logger
 from vllm.model_executor.layers.activation import get_act_fn
--- a/vllm/model_executor/models/whisper.py
+++ b/vllm/model_executor/models/whisper.py
@@ -19,9 +19,9 @@ from transformers import (
 from transformers.models.whisper.modeling_whisper import sinusoids
 
 from vllm.attention.layer import Attention
+from vllm.compilation.decorators import support_torch_compile
 from vllm.config import CacheConfig, ModelConfig, SpeechToTextConfig, VllmConfig
 from vllm.config.multimodal import BaseDummyOptions
-from vllm.distributed import get_tensor_model_parallel_world_size
 from vllm.inputs.data import PromptType
 from vllm.logger import init_logger
 from vllm.model_executor.layers.activation import get_act_fn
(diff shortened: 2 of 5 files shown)


DarkLight1337 (commented):
I think we should have at least one test that uses Whisper with CUDA graph

patrick-toulme:
Changes look fine to me. All you are doing is adding a gated option to run eager mode in model_forward. Any downstream consumers who are subclassing just have to add that variable now. LGTM

jikunshang:
a noob question: I noticed that this only enable Decoder part. Is there any blocker to enable torch.compile + Encoder part? just like mm + compile suport  https://docs.vllm.ai/en/latest/design/torch_compile_multimodal/

NickLucche:
@jikunshang https://github.com/vllm-project/vllm/pull/30549

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2919,6 +2920,17 @@ def _model_forward(
         Returns:
             Model output tensor
         """
+
+        if force_eager:
```
ProExpertProg: Let's just add force_eager to `ForwardContext` and read that in the compile decorator?

vllm/model_executor/models/whisper.py
```diff
@@ -561,6 +562,7 @@ def forward(self, input_features: torch.Tensor | list[torch.Tensor]):
         return self.forward_layers(hidden_states)
 
 
+@support_torch_compile(dynamic_arg_dims={"input_ids": 0, "positions": -1})
```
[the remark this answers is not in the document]
ProExpertProg: This seems right, how does it still work?

vllm/v1/worker/gpu_model_runner.py
Lucaskabela: There is an option for `enable_if` in the support_torch_compile decorator - perhaps we can leverage that?

see https://github.com/vllm-project/vllm/blob/3a612322eb9351594bc0c35fe5b85426931be36e/vllm/model_executor/models/qwen2_5_vl.py#L522 for example usage

vllm/model_executor/models/whisper.py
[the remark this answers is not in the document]
NickLucche: This looks right to me too, thanks for raising it, will look into it.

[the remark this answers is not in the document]
NickLucche: smooth @cursoragent

vllm/v1/worker/gpu_model_runner.py
NickLucche: that flag is for optional compilation, here I need to always compile, but optionally do eager (aka not call the compiled graph)

ProExpertProg (approved):

tests/entrypoints/openai/correctness/test_transcription_api_correctness.py
```diff
@@ -156,7 +156,9 @@ def test_wer_correctness(
     model_name, dataset_repo, expected_wer, n_examples=-1, max_concurrent_request=None
 ):
     # TODO refactor to use `ASRDataset`
-    with RemoteOpenAIServer(model_name, ["--enforce-eager"]) as remote_server:
+    with RemoteOpenAIServer(
```
ProExpertProg, line 159: `-cc.mode=NONE` if you want to just disable compilation but keep cuda graphs

NickLucche: will follow up with more cg+compilation tests (which is the default `vllm serve` setup)

PR #30388, opened 2025-12-10
[Docs] Generate full list of metrics in user docs

markmc:
(Generate script produced by Claude)

We used to have a snippet of Python code in https://docs.vllm.ai/en/latest/usage/metrics/ that served as a list of available metrics, but metrics are now a bit more scattered, and "read the code" is hardly super-friendly for users.

Let's generate tables of metrics by parsing the relevant source files

Also, update the metrics design doc to make it clear it's not a complete or canonical list

### DIFF
--- a/docs/mkdocs/hooks/generate_metrics.py
+++ b/docs/mkdocs/hooks/generate_metrics.py
@@ -1,0 +1,157 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+import ast
+import logging
+from pathlib import Path
+from typing import Literal
+
+logger = logging.getLogger("mkdocs")
+
+ROOT_DIR = Path(__file__).parent.parent.parent.parent
+DOCS_DIR = ROOT_DIR / "docs"
+GENERATED_METRICS_DIR = DOCS_DIR / "generated" / "metrics"
+
+# Files to scan for metric definitions - each will generate a separate table
+METRIC_SOURCE_FILES = [
+    {"path": "vllm/v1/metrics/loggers.py", "output": "general.md", "name": "general"},
+    {
+        "path": "vllm/v1/spec_decode/metrics.py",
+        "output": "spec_decode.md",
+        "name": "spec_decode",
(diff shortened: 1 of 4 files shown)


docs/mkdocs/hooks/generate_metrics.py
```diff
@@ -0,0 +1,157 @@
...
+    total_metrics = 0
+    for source_config in METRIC_SOURCE_FILES:
+        source_path = source_config["path"]
+        output_file = source_config["output"]
+        source_name = source_config["name"]
+
+        filepath = ROOT_DIR / source_path
+        if not filepath.exists():
+            logger.warning("Metrics source file not found: %s", filepath)
+            continue
```
hmellor: It'd be better if we raised here instead so that the metrics docs don't go missing silently

docs/mkdocs/hooks/generate_metrics.py
```diff
@@ -0,0 +1,157 @@
...
+        with open(filepath, encoding="utf-8") as f:
+            source = f.read()
+
+        tree = ast.parse(source, filename=str(filepath))
+        extractor = MetricExtractor(str(filepath))
+        extractor.visit(tree)
+        return extractor.metrics
+    except Exception as e:
+        logger.warning("Failed to parse %s: %s", filepath, e)
+        return []
```
hmellor: Same here with the warning

docs/mkdocs/hooks/generate_metrics.py
```diff
@@ -0,0 +1,157 @@
...
+            name = self._extract_kwarg(node, "name")
+            documentation = self._extract_kwarg(node, "documentation")
+
+            if name:
+                self.metrics.append(
+                    {
+                        "name": name,
+                        "type": metric_type,
+                        "documentation": documentation or "",
+                        "file": self.filepath,
```
hmellor: This is unused and could probablyt be removed from `MetricExtractor`

hmellor (commented):
The output looks very nice, just a few nits about the hook

hmellor (approved):

PR #30390, opened 2025-12-10
fix: Update json features supported by xGrammar

johannesflommersfeld:
Update check for json features unsupported by xGrammar to be aligned with xGrammar after https://github.com/mlc-ai/xgrammar/pull/361.

## Purpose
Avoids unnecessary fallback to llguidance, which comes with its own set of unsupported keywords. This results in improved support for json schemas.

## Test Plan
Run `pytest -s -v tests/v1/entrypoints/llm/test_struct_output_generate.py` with the updated sample schema that contains the newly supported schema keywords.


aarnphm (approved):

johannesflommersfeld:
It appears as if xgrammar does still not properly handle regex based pattern matching. I will see if this can be fixed on the xgrammar side but for now we should keep  "propertyNames" and "patternProperties" classified as unsupported.

russellb (approved):
Can you file a bug against xgrammar for the items not working properly?

PR #30409, opened 2025-12-10
[BugFix] Lazy tokenizer init in StructuredOutputManager to prevent GGUF semaphore leak

kitaekatt:
## Summary

Fixes semaphore exhaustion when repeatedly loading GGUF models by deferring tokenizer initialization.

## Changes

- Lazy initialization of tokenizer in StructuredOutputManager
- Prevents resource leak during model loading/unloading cycles

## Testing

Tested with repeated model loads - no more semaphore exhaustion.


kitaekatt:
Closing in favor of a fresh PR rebased on current main. The original lazy-init approach is preserved but rewritten against the current StructuredOutputManager structure which has changed significantly.

PR #30411, opened 2025-12-10
fix(gguf): Ensure Gemma2 configs have hidden_act for backward compatibility

kitaekatt:
## Summary

Fixes `AttributeError: 'Gemma2Config' object has no attribute 'hidden_act'` when loading Gemma2 GGUF models.

## Changes

- In `ModelConfig.__init__`, if model_type is "gemma2" and config has `hidden_activation` but not `hidden_act`, copy the value

## Root Cause

- Transformers' `Gemma2Config` only defines `hidden_activation`, not `hidden_act`
- vLLM's `gemma2.py` directly accesses `config.hidden_act` without fallback

## Testing

Tested with `bartowski/gemma-2-2b-it-GGUF` - model resolves architecture correctly.

### DIFF
--- a/vllm/config/model.py
+++ b/vllm/config/model.py
@@ -484,6 +484,17 @@ class ModelConfig:
         )
 
         self.hf_config = hf_config
+
+        # Ensure Gemma2 configs have hidden_act for backward compatibility.
+        # GGUF configs may only have hidden_activation; model code expects both.
+        if (
+            hasattr(hf_config, "model_type")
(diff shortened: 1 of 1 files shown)


yewentao256 (approved):
LGTM, thanks for the work!

hmellor (changes requested):
- Model specific edge cases in config classes should be avoided if possible
- Why can't we just update `gemma2.py` to access `hidden_activation`?

PR #30412, opened 2025-12-10
fix(gguf): Skip lm_head mapping for models with tied word embeddings

kitaekatt:
## Summary

Fixes `RuntimeError: Failed to map GGUF parameters (1): ['lm_head.weight']` for models using tied word embeddings.

## Changes

For models like Gemma2 that use `tie_word_embeddings=True`, add `lm_head.weight` to `sideload_params` to allow GGUF loading to succeed.

## Root Cause

When `tie_word_embeddings=True`, model shares weights between input embeddings and output projection. GGUF files don't contain separate `lm_head.weight` tensor.

## Testing

Tested with `bartowski/gemma-2-2b-it-GGUF` - model loads without parameter mapping error.


kitaekatt:
**Testing performed:**

Tested with GGUF models on RTX 5090 (32GB, Blackwell architecture):
- bartowski/NousResearch_Hermes-4-14B-GGUF
- tensorblock/Qwen_Qwen3-30B-A3B-Instruct-2507-GGUF

Ran models through a local benchmark runner for HumanEval and GSM8K. Verified the semaphore leak fix by running repeated model load/unload cycles without resource exhaustion.

**Related PRs:**

This is part of a series of GGUF pipeline fixes for Blackwell GPU compatibility:
- #30410: Fix GGUF dtype selection for Blackwell compatibility
- #30412 (this PR): Fix tokenizer semaphore leak
- #30434: Shared memory barrier fix

kitaekatt:
Hi @22quinn @Isotr0py — this has been sitting without review for a while. Just rebased on latest `upstream/main` (merge commit), so the branch should now be mergeable. Quick summary:  Skip lm_head mapping for models with tied word embeddings:small fix in gguf_loader.py — tied-embedding models were failing when lm_head mapping was attempted. Rebased on latest main.

Would appreciate a look when you have cycles. Thanks!

ErenAta16:
Went to check `_get_gguf_weights_map` against current `main` before reviewing the logic, and `vllm/model_executor/model_loader/gguf_loader.py` no longer exists there at all, the whole file is gone from `vllm/model_executor/model_loader/` (checked the directory listing directly, no gguf_loader.py, and no other file in that directory references GGUF weight mapping either). Whatever the fix here was meant to do, the target it's patching has since been removed or restructured, so this can't apply as-is regardless of whether the tied-embeddings logic itself is correct.

Might be worth confirming whether GGUF loading was dropped/relocated and, if the underlying bug is still relevant, re-filing against wherever that logic lives now rather than rebasing this diff.

PR #30417, opened 2025-12-10
[CI/Build][AMD] Skip tests in test_fusions_e2e and test_dbo_dp_ep_gsm8k that require non-existing imports for ROCm 

rasmith:
The test` test_fusions_e2e.py::test_attn_quant` is unable to run if `inductor_graph_partition` is `True` since the ROCm installation of pytorch 2.9 does NOT have `torch._inductor.utils.CUDAGraphWrapperMetadata`.  I checked in` dist-packages` and confirmed that the class was not available in our installation.

The other test that is unable to run is `test_dbo.py::test_dbo_dp_ep_gsm8k` since it requires `deep_ep `and ROCm does not have the `deep_ep `module available.

This change enables our` Distributed Tests (H200)`  test group to run successfully.


SageMoore (approved):

gshtras (approved):

tests/compile/distributed/test_fusions_e2e.py
```diff
@@ -138,14 +138,38 @@ class ModelBackendTestCase(NamedTuple):
...
 
 
+def has_cuda_graph_wrapper_metadata() -> bool:
+    from importlib import import_module
+
+    try:
+        module = import_module("torch._inductor.utils")
+        module.CUDAGraphWrapperMetadata  # noqa B018
+    except AttributeError:
+        return False
```
gshtras, line 148: Let's take a mental note to make sure this gets enabled and passes once we switch from 2.9.0a to 2.9.0 on ROCm

ProExpertProg (approved):
This is a better fix, thanks!

Issue #30358, opened 2025-12-09
[Bug]: Prefill scheduled num_block mismatch at update_state_after_alloc and request_finished


xuechendi:
### Your current environment

vllm-commit-id: 73a484caa1ad320d6e695f098c25c479a71e6774

Tested with A100

### 🐛 Describe the bug

How to reproduce
```
PREFILL_BLOCK_SIZE=16 DECODE_BLOCK_SIZE=16 bash tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh
```

---

What is the issue

I found that the num_blocks sent to `update_state_after_alloc` and `request_finished` sometimes is not match. 

`update_state_after_alloc` => this function is scheduled by `scheduler.schedule` to update req_to_save and req_to_receive list, and block_ids passed by the method will indicate which blocks belong to one request.

`request_finished` => this function is called also in `scheduler._connector_finished` to send completed request block_ids list to create a new metadata for decoder.

However, based print logs, sometimes, block_ids in `scheduler.schedule` `update_state_after_alloc` is shorter than `scheduler._connector_finished` `request_finished`  sometimes.

Example as below

```

📊 Found 1320 unique Request IDs.

FINAL SUMMARY
✅ Consistent Requests : 1085  => num_blocks are same at `update_state_after_alloc` and `request_finished` 
❌ Mismatched Requests : 235 => num_blocks is less in `update_state_after_alloc` than `request_finished` 
```

```
================================================================================
🔴 MISMATCH DETECTED: cmpl-25c7397c-5686-4b70-a569-29ef04c7b4f9-0
   First Block Count: 44
   Last Block Count : 71
   --- Raw Lines for Context ---
   [0;36m(EngineCore_DP0 pid=417455)[0;0m update_state_after_alloc req_id="request.request_id='cmpl-25c7397c-5686-4b70-a569-29ef04c7b4f9-0'" num_tokens=1121 len(block_ids)=44 block_ids=[162, 163, 164, 165, 166, 167, 168, 169, 170, 171, 172, 173, 174, 175, 176, 177, 178, 179, 180, 181, 182, 183, 184, 185, 186, 187, 188, 189, 190, 191, 192, 193, 194, 195, 196, 197, 198, 199, 200, 201, 202, 203, 204, 205]
[...]

PR #29870, opened 2025-12-02
[KVConnector] OffloadingConnector: Fix bug in handling of preemptions

orozery:
This PR fixes the OffloadingConnector to fail any stores for a preempted request, as well as allowing to store the result of the request re-computed blocks.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit 77bab86e5a415ec6ae3be7ffa56aa6f3fe24c149. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Ensures safe handling of preempted requests and completes/flushes in-flight KV offload transfers.
> 
> - Adds `handle_preemptions()` to KV connector base; OffloadingConnector worker now waits on pending store job IDs for preempted requests, and scheduler completes stores for preempted IDs in `build_connector_meta()`
> - Calls `handle_preemptions()` from `gpu_model_runner` before blocks are overwritten
> - Extends `OffloadingHandler`/`OffloadingWorker` with `wait(job_ids)`; CPU↔GPU handler tracks CUDA events per job and implements blocking `wait()`
> - Worker-side OffloadingConnector defers store submission to next step but can now force-wait on preemption; finished-getters unchanged besides plumbing
> - Tests: refactor mocks to track job/spec state and flushed jobs; add `test_request_preemption` validating flushed block indexes and re-load/store behavior; minor test harness updates
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3f960add3709a8ba8bbf258c9039e58ce19eb698. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit 5aa68e443d78f25f0bbd8d5fc6672cc6945a0191. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Fixes offload preemption handling and adds blocking wait support across KV transfer.
> 
> - Add `handle_preemptions()` to KV connector base; invoke from `gpu_model_runner` before blocks are overwritten
> - Scheduler side: in `OffloadingConnectorScheduler.build_connector_meta()`, mark in-flight stores complete for `preempted_req_ids`
[...]

PR #30419, opened 2025-12-10
[NIXL][BUG FIX] Fix both failing issue and accuracy issue with nixl + host_buffer on CUDA

xuechendi:
## Purpose

This PR is fixed upon https://github.com/vllm-project/vllm/pull/30420
Should get that one merged firstly

Two issue detected and resolved in this PR

1. Fix a bug after #29665 for running PD with cpu host buffer
2. Fix accuracy issue for running PD with cpu host buffer, described in https://github.com/vllm-project/vllm/issues/30358

## Test Plan

```
PREFILL_BLOCK_SIZE=16 DECODE_BLOCK_SIZE=16 bash tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh --kv_buffer_device cpu
```

Qwen3-0.6B
Before: accuracy is ~0.3
Now: Accuracy is = 0.4109

## Root Cause and proposed change

Previous:

when scheduler call `schedule`, only brand new request will gets `self.connector.update_state_after_alloc`

=> However, if one prefill request is chunked into small request, the `self.connector.update_state_after_alloc` only registered partial of block_ids(first request of the prefill) into nixl_metadata

--

Solution:

Add another call to `self.connector.update_state_after_alloc` at running queue process, so if the following chunked gets scheduled, it will continue to update block_ids to nixl metadata.

## Test Result

---
<details>

</details>


vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -238,7 +238,7 @@ def add_new_req(
             local_physical_block_ids=local_block_ids,
             remote_block_ids=kv_transfer_params["remote_block_ids"],
             remote_engine_id=kv_transfer_params["remote_engine_id"],
-            remote_request_id=kv_transfer_params["remote_request_id"],
+            remote_request_id=kv_transfer_params.get("remote_request_id", ""),
             remote_host=kv_transfer_params["remote_host"],
             remote_port=kv_transfer_params["remote_port"],
```
markmc: I don't think we should require `remote_engine_id`, `remote_block_ids`, `remote_request_id`, `remote_host`, or `remote_port` for a prefill request - it just happens that existing proxy servers set those fields, but I don't think there's any reason to require them

Let's use `.get(..., None)` for all of them - AFAICT all the other fields are None in a prefill request currently

markmc (dismissed):
Thank you 

This is only an issue with `--kv_buffer_device cpu` which I guess we don't test in CI

I don't think `remote_request_id` should be particularly special here - a prefill request should be able to just contain `kv_transfer_params = {'do_remote_decode': True}` IMO

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
markmc: Let's deal with this part in #30420

xuechendi: thanks, @markmc , I removed the same commit in #30420, Was trying to fix the issue in one PR
Now this PR is cleaned and only focus on the scheduler change, please help to check

orozery:
@markmc @njhill I don't think this change is necessary as `block_ids` of subsequent allocations are available to connectors via `SchedulerOutput.scheduled_cached_reqs.new_block_ids` which is passed in `build_connector_meta`.

@xuechendi you can look at the offloading connector for a reference:
https://github.com/vllm-project/vllm/blob/783644e4ac7d4ee324a1817bbe199fc5b557bc7d/vllm/distributed/kv_transfer/kv_connector/v1/offloading_connector.py#L521

Now that I think of it, I think there's also another bug in the nixl connector triggered when a chunked-prefill request gets preempted.
Currently, looks like the nixl connector does detect that.
For the offloading connector we also have the same bug, which should be solved by #29870.
Basically, we detect preemptions via `SchedulerOutput.preempted_req_ids` in `build_connector_meta`.

xuechendi:
@orozery 
Oh, I see, so in `offloadingConnectorMetadata`, it leverage `scheduler_output` to obtain the `new_block_ids`.
I can do that in nixl_connector as well.
It introduces a bit refactoring to current nixl_connetor impl, will draft a seperate PR see @NickLucche is OK with that or not.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -674,23 +676,46 @@ def build_connector_meta(
...
+
+        added_reqs = []
+        for req_id, (req, _) in self._reqs_need_save.items():
             assert req.kv_transfer_params is not None
+            # don't use the block_ids from _reqs_need_save since it may be
+            # incomplete for partial prefill.
+            new_block_ids = self._request_block_ids.get(req_id, [])
+            is_partial = req.num_tokens > (len(new_block_ids) * self.block_size)
+            if is_partial:
+                continue
```
orozery: I was wondering if it's not better to save what we have, instead of waiting for the entire request to be available.
Then I see that `self.copy_blocks` is blocking so I guess it does not matter.
@NickLucche your thoughts?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -472,6 +472,8 @@ def __init__(self, vllm_config: VllmConfig, engine_id: str):
         # Reqs to remove from processed set because they're not to send after
         # remote prefill or aborted.
         self._reqs_not_processed: set[ReqId] = set()
+        # list of GPU block IDs per request
+        self._request_block_ids: dict[ReqId, list[int]] = {}
```
orozery: Instead of tracking  the block IDs of all requests, let's just track block IDs of requests that needs saving.
You can discard this new variable, and use `self._reqs_need_save` to track the block IDs.

NickLucche: agree we can re-use the existing container

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -2292,6 +2317,26 @@ def zmq_ctx(socket_type: Any, addr: str) -> Iterator[zmq.Socket]:
             ctx.destroy(linger=0)
 
 
+def yield_req_data(
+    scheduler_output,
+) -> Iterator[tuple[str, tuple[list[int], ...], bool]]:
```
NickLucche: I don't see harm in moving this shared function in utils

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -674,23 +676,46 @@ def build_connector_meta(
...
                 request_id=req_id,
-                local_block_ids=block_ids,
+                local_block_ids=new_block_ids,
                 kv_transfer_params=req.kv_transfer_params,
                 load_remote_cache=False,
                 save_to_host=True,
             )
+            added_reqs.append(req_id)
+        # NOTE: for partial prefill, we will keep saving until the full
+        # prefill is done. Clear to avoid double saving.
```
NickLucche: do you mean

```suggestion
        # NOTE: for partial prefill, we do not save until the full
        # prefill is done. Clear once transfer on full prefill is started.
```

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
NickLucche, line 691: maybe clearer if we ditch added_reqs?
```suggestion
            # Don't use the block_ids from _reqs_need_save since it may be
            # incomplete for partial prefill.
            new_block_ids = self._request_block_ids.get(req_id, [])
            is_partial = req.num_tokens > (len(new_block_ids) * self.block_size)
            if not is_partial:
                # Prefill request fully processed
                meta.add_new_req(
                    request_id=req_id,
                    local_block_ids=new_block_ids,
                    kv_transfer_params=req.kv_transfer_params,
                    load_remote_cache=False,
                    save_to_host=True,
                )
                self._reqs_need_save.pop(req_id, None)
                self._request_block_ids.pop(req_id, None)```

File: vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py [hunk 6]
```diff
@@ -674,23 +676,46 @@ def build_connector_meta(
...
+                )
+                self._request_block_ids[req_id] = new_block_ids
+
+        added_reqs = []
+        for req_id, (req, _) in self._reqs_need_save.items():
             assert req.kv_transfer_params is not None
+            # don't use the block_ids from _reqs_need_save since it may be
+            # incomplete for partial prefill.
+            new_block_ids = self._request_block_ids.get(req_id, [])
+            is_partial = req.num_tokens > (len(new_block_ids) * self.block_size)
```
NickLucche: can you double check this is working instead
```suggestion
            is_partial = request.num_computed_tokens < request.num_prompt_tokens
```
it's better once we switch to HMA to avoid counting blocks manually.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
NickLucche: I think there's interesting developments for copying chunks, but for now I would treat this PR as a bug fix and prioritize getting the feature in a working state. 
We can leave this optimization for future work.

NickLucche (commented):
Thanks for the work @xuechendi and the great review @orozery !
Left a few comments but things look good overall.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: @orozery  done. Change back to copy immediately

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: done. 
`request.num_computed_tokens < request.num_prompt_tokens` is not working.
So I also add with `num_scheduled_tokens` from `scheduler_output`.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py

xuechendi:
@NickLucche @orozery 
I've resolved all comments, and I now switched back to do "copy immediately" instead of "waiting all blocks then copy".

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -689,21 +689,39 @@ def build_connector_meta(
                 kv_transfer_params=req.kv_transfer_params,
             )
 
-        for req_id, (req, block_ids) in self._reqs_need_save.items():
+        # NOTE: For prefill side, there might be chance that early added
+        # request is chunked prefill, so we need to check if new blocks are added
```
orozery: nit:
```suggestion
        # NOTE: For the prefill side, there might be a chance that an early added
        # request is a chunked prefill, so we need to check if new blocks are added
```

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -689,21 +689,39 @@ def build_connector_meta(
...
+        for req_id, new_block_id_groups, preempted in yield_req_data(scheduler_output):
+            if req_id not in self._reqs_need_save:
+                continue
```
orozery: Let's switch `self._reqs_need_save` to req_id -> Request.
Then, avoid the additional lookup on line 702.
```suggestion
            req = self._reqs_need_save.get(req_id)
            if req is None:
                continue
```

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -689,21 +689,39 @@ def build_connector_meta(
...
+            if preempted:
+                self._reqs_need_save.pop(req_id)
```
orozery: This will cause preempted requests to be excluded from sending.
I think this should be removed:
```suggestion
```

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -689,21 +689,39 @@ def build_connector_meta(
...
+                local_block_ids=new_block_id_groups[0],
                 kv_transfer_params=req.kv_transfer_params,
             )
+            # NOTE: for partial prefill, we will keep the req_id until the full
+            # prefill is done. Clear once transfer on full prefill is started.
+            assert scheduler_output.num_scheduled_tokens is not None
+            num_scheduled_tokens = scheduler_output.num_scheduled_tokens[req_id]
+            is_partial = (
+                req.num_computed_tokens + num_scheduled_tokens
+            ) < req.num_prompt_tokens
```
orozery, line 704: This might work...
But I think it's safer to just move `self._reqs_need_save.pop(req_id)` to `request_finished`.
@NickLucche WDYT?

xuechendi: `self._reqs_need_save.clear()` was originally in this function, so I prefer to keep `self._reqs_need_save.pop(req_id)` here instead of changing the logic more?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: Wondering if request was scheduled as preempted before finishing prefill, should we still let Decoder reading this?

orozery: The decoder starts pulling data only after the prefiller finished with the request, right?
In that case, preemption on the profiller should not matter for the decoder.
It just means the request gets re-allocated inside the prefiller.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: Oh, do you mean updating the `self._reqs_need_save` definition to dict[ReqId, Request] ? 
@NickLucche , would like to know your comments. That might introduce more changes to other functions

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: Done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: Makes sense, I was still thinking about my old implementation that copy to host happens when entire request completes.  For the copy immediately strategy, you're right, if request gets preempted or not should not impact here,
I have removed above if statement, Thanks!

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: I updated the codes to skip the second lookup, thanks for the suggestion
Should we update the `self._reqs_need_save`, I will still wait for Nick's confirmation before doing that

vllm/distributed/kv_transfer/kv_connector/utils.py
```diff
@@ -201,6 +202,26 @@ def copy_kv_blocks(
         copy_fn(src_tensor, dst_tensor, src_indices, dst_indices)
 
 
+def yield_req_data(
+    scheduler_output,
+) -> Iterator[tuple[str, tuple[list[int], ...], bool]]:
+    """
```
NickLucche, line 210: I meant we should pull this function out of the offloading_connector and share its implementation between the 2 connectors here

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -689,21 +689,38 @@ def build_connector_meta(
...
 
-        for req_id, (req, block_ids) in self._reqs_need_save.items():
+        # NOTE: For the prefill side, there might be a chance that an early added
+        # request is a chunked prefill, so we need to check if new blocks are added
+        for req_id, new_block_id_groups, preempted in yield_req_data(scheduler_output):
+            req_tuple = self._reqs_need_save.get(req_id)
+            if req_tuple is None:
+                continue
+            if new_block_id_groups is None:
+                continue
```
NickLucche: nit: 
```suggestion
            if req_tuple is None or new_block_id_groups is None:
                continue
```

vllm/distributed/kv_transfer/kv_connector/utils.py
NickLucche: *assuming function is the same in offloading_connector

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -689,21 +689,38 @@ def build_connector_meta(
...
+            req = req_tuple[0]
             assert req.kv_transfer_params is not None
             meta.add_new_req_to_save(
                 request_id=req_id,
-                local_block_ids=block_ids,
+                local_block_ids=new_block_id_groups[0],
                 kv_transfer_params=req.kv_transfer_params,
             )
+            # NOTE: for partial prefill, we will keep the req_id until the full
+            # prefill is done. Clear once transfer on full prefill is started.
```
NickLucche: I didn't get this note on req_id sorry

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -689,21 +689,38 @@ def build_connector_meta(
...
+            assert scheduler_output.num_scheduled_tokens is not None
+            num_scheduled_tokens = scheduler_output.num_scheduled_tokens[req_id]
+            is_partial = (
+                req.num_computed_tokens + num_scheduled_tokens
+            ) < req.num_prompt_tokens
+            if not is_partial:
+                self._reqs_need_save.pop(req_id)
```
NickLucche, line 711: I don't get it, aren't we missing the last chunk like this?

iiuc, right now we get 
| chunk0 | chunk1 | chunk2
chunk0-1 are saved because partial, but chunk2 evaluates to NOT partial because it's the terminal chunk

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
NickLucche: yep let's keep this here this is a tmp buffer, its function terminates here once metadata are built

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -689,21 +689,38 @@ def build_connector_meta(
...
-                local_block_ids=block_ids,
+                local_block_ids=new_block_id_groups[0],
                 kv_transfer_params=req.kv_transfer_params,
             )
+            # NOTE: for partial prefill, we will keep the req_id until the full
+            # prefill is done. Clear once transfer on full prefill is started.
+            assert scheduler_output.num_scheduled_tokens is not None
+            num_scheduled_tokens = scheduler_output.num_scheduled_tokens[req_id]
+            is_partial = (
+                req.num_computed_tokens + num_scheduled_tokens
```
NickLucche, line 703: nice that we can express it without blocks :)

NickLucche (commented):
Left some comments @xuechendi , not sure I fully understand the new logic to copy each chunk immediately, it feels to me we might be skipping the terminal chunk now..?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
NickLucche: > do you mean updating the self._reqs_need_save definition to dict[ReqId, Request] ?

Agree with @orozery , if we don't need the block_ids in `self._reqs_need_save`, no reason to keep them around anymore.

NickLucche: @xuechendi

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: It will not, because the last chunk has been submitted into meta already
The logic here is that:
1. `scheduler.schedule() running queue for loop` calls `connector_scheduler`.`update_state_after_alloc`  to add req to `_reqs_need_save`
2. `scheduler.schedule()` calls `connector_scheduler`.`build_connector_meta`  to **create new `req_meta`** according to `_reqs_need_save`, then decide if we need to clean up this `req_id `from `_reqs_need_save` (line 715/716)
3. `model runner` calls `connector_worker`.`wait_for_save` to use `req_meta` created in `build_connector_meta` for copying
4. `scheduler.schedule()` -> `connector_scheduler`.`finished_request` # prepare request meta to remote   

--

For the func above, originally, we don't assume partial prefill requests, so we clear entire `_reqs_need_save` once we create `req_meta` from it. 
Now, because some of the requests might be partial prefill, we will not delete these `req_id` from `_reqs_need_save`.
And once the last chunk comes in, we can clean it from `_reqs_need_save`.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: Hmm, I found the note is little redundant and confusing, removed

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py

vllm/distributed/kv_transfer/kv_connector/utils.py
xuechendi: done

xuechendi:
> Left some comments @xuechendi , not sure I fully understand the new logic to copy each chunk immediately, it feels to me we might be skipping the terminal chunk now..?

Oh... I think I should not say "immediately". The logic is same as before actually. 
Here is my definition

"copy at once": when partial prefill comes in, we keep collecting block_ids of this request in `_req_to_save` , hold on `req_meta` creation until last chunk comes in. Once we have all chunks, we create `req_meta` with all block ids of this request.

"copy immediately": when partial prefill comes in, we create `req_meta` for this partial prefill req immediately, once model_runner completes forward and call wait_for_save, this partial will be copied to host_buffer. And rest chunks will do the same. `_req_to_save` will be used to keep monitoring this partial prefill request until all chunks scheduled

And now I am using "copy immediately" strategy, so it is same as before

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: Done

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: Further relief is that I verified accuracy locally (with partial prefill), so I can confirm we didn't miss last chunk

NickLucche (approved):
Thanks for the breakdown @xuechendi , I missed this part my bad
> the last chunk has been submitted into meta already

so this is working as intended.

Could you possibly distill your breakdown in a comment 
```
            if not is_partial:
                # HERE
                self._reqs_need_save.pop(req_id)
``` 
that explains why we're holding onto the req ids? I think this could be useful for context in some time.
Other than that, this is LGTM!

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
orozery: @NickLucche @xuechendi I think you are missing an edge case:
https://github.com/vllm-project/vllm/blob/811cdf5197acb4d6ab42250a5b0f822887d1190a/vllm/v1/core/sched/scheduler.py#L1353-L1362

When the scheduler finish requests before they finish processing, you will not clear `self._reqs_need_save`.

NickLucche: Which is every time a request is aborted too, nice catch @orozery !

xuechendi: Thanks, @orozery ! @NickLucche , I've updated the PR with clean up in `request_finished`, validated locally

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -751,6 +760,9 @@ def request_finished(
             self._reqs_not_processed.add(request.request_id)
             return False, None
 
+        # Clear _reqs_need_save if a request is aborted as partial prefill.
+        self._reqs_need_save.pop(request.request_id, None)
+
```
orozery: I think you need to move this upward a few lines, at least before `if request.status != RequestStatus.FINISHED_LENGTH_CAPPED:`

xuechendi: Oh, I see, thanks, updated

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -749,6 +758,8 @@ def request_finished(
             # Also include the case of a P/D Prefill request with immediate
             # block free (eg abort). Stop tracking this request.
             self._reqs_not_processed.add(request.request_id)
+            # Clear _reqs_need_save if a request is aborted as partial prefill.
+            self._reqs_need_save.pop(request.request_id, None)
```
orozery, line 765: I think this should work, but it seems more fragile to me.
I would go further and put this pop right at the beginning.
Then, you could also remove the entire `is_partial` check from `build_connector_meta`.

xuechendi: Since `_reqs_need_save` originally is used to buffer which requests should create req_meta for saving to host. So the life cyle is from "scheduled" to "all request metadata are created".

If we go with your proposal, the life cycle becomes from "scheduled" to "request ends"
This changes the original design. 

@NickLucche, do you think we should do that?

I assume the fix here is to just handle a corner case when request was aborted ?

NickLucche: I think @orozery is proposing to only clear the id on request finished, so either terminal block was processed or abort/error. 
I think it makes sense but I don't necessarily think the current implementation is fragile.

Hence I don't have a strong opinion here, this could also be done in a separate PR, as long as we maximize clarity for these cases.

xuechendi: @NickLucche @orozery , let's do that in separate PR, since other queues `_reqs_to_recv` `_reqs_need_send` and etc are being cleared in `build_connector_meta`, I would prefer not adding refactor scope into this PR (originally for for accuracy fix). And once this one merged, I can open a new PR and we can have better design there.

DarkLight1337 (approved):

PR #30420, opened 2025-12-10
[NIXL][BUG FIX] Fix a bug for PD with host_buffer after merging 29665

xuechendi:
## Purpose

After https://github.com/vllm-project/vllm/pull/29665 merged, the PR added `remote_request_id=kv_transfer_params["remote_request_id"],` to `add_new_req`

However, in prefill side, there is no `remote_request_id` in kv_transfer_params

=> This issue only occurs when running PD with host_buffer

## Test Plan

```
PREFILL_BLOCK_SIZE=16 DECODE_BLOCK_SIZE=16 bash tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh --kv_buffer_device cpu
```

without the PR, test will fail immediately
With this PR, there will still be an accuracy issue, which fix is proposed at https://github.com/vllm-project/vllm/pull/30419

## Test Result

---
<details>

</details>


vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -238,7 +238,7 @@ def add_new_req(
             local_physical_block_ids=local_block_ids,
             remote_block_ids=kv_transfer_params["remote_block_ids"],
             remote_engine_id=kv_transfer_params["remote_engine_id"],
-            remote_request_id=kv_transfer_params["remote_request_id"],
+            remote_request_id=kv_transfer_params.get("remote_request_id", ""),
```
markmc: I don't think we should require `remote_engine_id`, `remote_block_ids`, `remote_request_id`, `remote_host`, or `remote_port` for a prefill request - it just happens that existing proxy servers set those fields, but I don't think there's any reason to require them

Let's use `.get(..., None)` for all of them - AFAICT all the other fields are None in a prefill request currently

markmc (changes requested):
Thank you

This is only an issue with `--kv_buffer_device cpu` which I guess we don't test in CI

I don't think `remote_request_id` should be particularly special here - a prefill request should be able to just contain `kv_transfer_params = {'do_remote_decode': True}` IMO

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: Ok, will do. 
`remote_request_id=kv_transfer_params.get("remote_request_id", None)` => Failed pre-commit mypy, so I used `remote_request_id=kv_transfer_params.get("remote_request_id", "")` instead

xuechendi: Done, @markmc , please help to review again.

markmc: Sorry to be a pain, but I'm pretty sure the proxies are sending `None` for these values currently - if that's true (please double-check I'm not mistaken!) then we should update the `ReqMeta` type hints to allow these to be None

xuechendi: Yes, proxies are sending None! I tested with current default("" or 0), it also works, but I can update the datatype in ReqMeta to make it as None as well

xuechendi: @markmc 
By adding Optional of None for ReqMeta, the mypy complains more now
Wondering if I can ease the pain by simply providing a default value(""  or 0 or [])?

```
vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py:1129: error: Argument 2 to "submit" of "Executor" has incompatible type "str | None"; expected "str"  [arg-type]                                     
vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py:1130: error: Argument 3 to "submit" of "Executor" has incompatible type "int | None"; expected "int"  [arg-type]                                     
vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py:1636: error: Argument 2 to "map" has incompatible type "list[int] | None"; expected "Iterable[int]"  [arg-type]                                      
vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py:1771: error: Need type annotation for "block_ids_to_permute" (hint: "block_ids_to_permute: list[<type>] = ...")  [var-annotated]                     
vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py:1780: error: Argument 1 to "__iadd__" of "list" has incompatible type "list[int] | None"; expected "Iterable[Any]"  [arg-type]                       
vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py:1920: error: Argument 1 to "_logical_to_kernel_block_ids" of "NixlConnectorWorker" has incompatible type "list[int] | None"; expected "list[int]"  [arg-type]
vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py:1937: error: Argument 2 to "_background_nixl_handshake" of "NixlConnectorWorker" has incompatible type "str | None"; expected "str"  [arg-type]      
vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py:1975: error: Argument "dst_engine_id" to "_read_blocks" of "NixlConnectorWorker" has incompatible type "str | None"; expected "str"  [arg-type]      
vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py:1976: error: Argument "remote_request_id" to "_read_blocks" of "NixlConnectorWorker" has incompatible type "str | None"; expected "str"  [arg-type]  
vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py:1977: error: Argument "local_block_ids" to "_read_blocks" of "NixlConnectorWorker" has incompatible type "list[int] | None"; expected "list[int]"  [arg-type]
vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py:1978: error: Argument "remote_block_ids" to "_read_blocks" of "NixlConnectorWorker" has incompatible type "list[int] | None"; expected "list[int]"  [arg-type]
```

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -236,11 +236,11 @@ def add_new_req(
...
-            remote_block_ids=kv_transfer_params["remote_block_ids"],
-            remote_engine_id=kv_transfer_params["remote_engine_id"],
-            remote_request_id=kv_transfer_params["remote_request_id"],
-            remote_host=kv_transfer_params["remote_host"],
-            remote_port=kv_transfer_params["remote_port"],
+            remote_block_ids=kv_transfer_params.get("remote_block_ids"),  # type: ignore
+            remote_engine_id=kv_transfer_params.get("remote_engine_id"),  # type: ignore
+            remote_request_id=kv_transfer_params.get("remote_request_id"),  # type: ignore
+            remote_host=kv_transfer_params.get("remote_host"),  # type: ignore
+            remote_port=kv_transfer_params.get("remote_port"),  # type: ignore
```
xuechendi: @markmc , do you think this works?
I tried updating ReqMeta with optional, then the mypy check exploded. => Since we only want to align with toy_proxy, so I am hoping to use `# type ignore` , in that case, we can still use None and make mypy happy

markmc:
The mypy errors are a sign that we can do better here - the current code basically disables type checking

```
   def _add_new_req(
         ...
         kv_transfer_params: dict[str, Any],
         _req = ReqMeta(
             ...
             remote_engine_id=kv_transfer_params["remote_engine_id"]
```

Because the value returned by the dict lookup is `Any`, mypy does no type checking - even though it could be None (and we know it definitely is None on the decode side runtime), mypy doesn't complain that we are assigning it to a `str` field. And what's worse, it's a really subtle way of disabling type checking ... it's not at all obvious to the reader, so someone reading this code might be very surprised to find these fields are `None` at runtime

I've sent you xuechendi/vllm-fork#15 which will enable proper type checking. Please do check it over and make sure it makes sense to you :+1:

markmc (approved):
@NickLucche PTAL, since I've done a bunch of refactoring in the latest version

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -202,17 +202,22 @@ def compute_nixl_compatibility_hash(
     return compat_hash
 
 
+@dataclass
```
njhill, line 205: nit: perf benefit
```suggestion
@dataclass(slots=True)
```

can add it to other dataclasses here too!

Edit: can do this in a follow-on

njhill (approved):
Thanks @xuechendi @markmc! The RemoteMeta refactoring seems much cleaner!

PR #30427, opened 2025-12-10
fix(gguf): Extract attn_logit_softcapping from GGUF metadata

kitaekatt:
## Summary
Fixes garbage output from Gemma2 GGUF models by extracting the `attn_logit_softcapping` parameter from GGUF metadata and patching it onto the HuggingFace config.

## Root Cause
GGUF models store softcap in metadata with arch-specific keys (e.g., `gemma2.attn_logit_softcapping`), but this wasn't being extracted and applied to the HF config. Without softcap, the V1 FlashAttention backend uses `softcap=0` (disabled), causing numerical instability and garbage output.

The attention backends ALREADY support softcap correctly - this fix is in the GGUF config mapping layer, not the attention kernels.

## Changes
- Add `extract_softcap_from_gguf()` to read softcap from GGUF metadata
- Update `maybe_patch_hf_config_from_gguf()` to apply softcap values
- Support both `attn_logit_softcapping` and `final_logit_softcapping`

## Testing
Tested with: `google/gemma-2-2b-it:Q4_K_M` - produces coherent output instead of garbage

## Related
This is part of a series of GGUF fixes for Blackwell (SM 120+) compatibility.

### DIFF
--- a/vllm/config/model.py
+++ b/vllm/config/model.py
@@ -470,6 +470,17 @@ class ModelConfig:
         )
 
         self.hf_config = hf_config
+
+        # Ensure Gemma2 configs have hidden_act for backward compatibility.
+        # GGUF configs may only have hidden_activation; model code expects both.
+        if (
+            hasattr(hf_config, "model_type")
+            and hf_config.model_type == "gemma2"
+            and not hasattr(hf_config, "hidden_act")
+            and hasattr(hf_config, "hidden_activation")
+        ):
+            hf_config.hidden_act = hf_config.hidden_activation
+
         if dict_overrides:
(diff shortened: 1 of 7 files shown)


kitaekatt:
**Update**: Upstream Transformers fix submitted: huggingface/transformers#42881

Once merged, this vLLM workaround can be removed as Transformers will properly extract `attn_logit_softcapping` from GGUF metadata.

kitaekatt:
Addressed @gemini-code-assist feedback in 3297f677d:
- Changed `logger.debug` to `logger.warning` 
- Added model path to log message for easier debugging

Thanks for the suggestion!

PR #30443, opened 2025-12-11
[CI][torch nightlies] Use main Dockerfile with flags for nightly torch tests

orionr:
Use standard Docker image instead of torch_nightly image for PyTorch nightlies testing and CI runs.

Moving this from https://github.com/vllm-project/ci-infra/pull/239 to a branch on upstream for testing purposes outlined at https://github.com/vllm-project/ci-infra?tab=readme-ov-file#how-to-test-changes-in-this-repo

Tests to confirm:
1. Baseline (my `vllm` fork matching HEAD, no `ci-infra` changes) at https://buildkite.com/vllm/ci/builds/42874/steps/canvas. Allowed 5 test runs to move forward. -> Seems like PT nightlies build itself failed on installing `flashinfer` so all tests failed afterwards.
2. New (my `vllm` changes at https://github.com/vllm-project/vllm/pull/30443, my `ci-infra` changes at https://github.com/vllm-project/ci-infra/pull/244) with a successful build at https://buildkite.com/vllm/ci/builds/45736/steps/canvas?sid=019b9459-43ce-46d3-99c2-c10a1a8ce96c. One downstream test is failing, but that looks real and something we will investigate.

We will remove https://github.com/vllm-project/vllm/blob/main/docker/Dockerfile.nightly_torch in a separate commit.

Looking for review and landing with help from @khluu, @amrmahdi, @atalman, @huydhn . Thanks!

After we land this PR, here are some thoughts on next steps:
1. https://github.com/vllm-project/vllm/pull/32721 : We should explore moving all torch dep requirements into requirements/torch.txt (torch==...) and requirements/torchlibs.txt (torchaudio==..., torchtext==...) and cpu.txt, xpu.txt, test.in, etc would all -r that file. If we do that the use_existing_torch.py file would just need to clear or override that file to make sure we are referencing nightlies (update to torch_lib_versions.txt) or using the existing torch (make it a blank file). I might try and put up a separate PR around this with my ideas.
2. There would still be some pip install flags adjusted for torch nightly vs release (--pre, --index-url vs --extra-index-url) but we could maybe get rid of the if [ "${PYTORCH_NIGHTLY}" = "1" ]; then in most cases and especially outside of the two base images. This would be good! I don't think we would need to use --constraint if the requirements/ are updated correctly, but we should check.
3. We should look at caching torch nightly builds same as standard builds, but be careful about cache location and collisions with the torch release builds.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit f8f0e53f6e9e547196cd1cc11218cef56c07851f. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Moves PyTorch nightly support into the main `docker/Dockerfile` via a `PYTORCH_NIGHTLY=1` flag and enforces consistent torch lib versions across all build/test stages.
> 
> - Adds `PYTORCH_NIGHTLY` ARG to install nightly `torch/torchvision/torchaudio` and switches indexes accordingly
> - Introduces `use_existing_torch.py` to strip torch deps from `requirements/*` and `pyproject.toml` (prefix-aware)
> - Records installed torch lib versions in `torch_lib_versions.txt` and validates them in `csrc-build`, `build`, `vllm-base`, and `test` stages
> - Updates dev/test installs to pin/compile requirements against recorded torch versions when using nightlies
> - Marks `docker/Dockerfile.nightly_torch` as deprecated; use the main Dockerfile with `PYTORCH_NIGHTLY=1`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f8f0e53f6e9e547196cd1cc11218cef56c07851f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit c00a0691687a77994bd613ce4e507f5541096c60. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Moves PyTorch nightly support into the main `docker/Dockerfile` behind `PYTORCH_NIGHTLY=1`, with consistent torch lib versioning across all stages.
> 
> - Adds `PYTORCH_NIGHTLY` ARG and switches installs to nightly indexes when set; otherwise keeps release flow
> - Introduces `use_existing_torch.py` to strip `torch*` deps from `requirements/*` and `pyproject.toml` (prefix-aware)
> - Records installed `torch/torchvision/torchaudio` into `torch_lib_versions.txt` and reuses them in `csrc-build`, `build`, `dev`, `vllm-base`, and `test` stages to avoid version drift
> - Dev/test installs now compile/pin requirements against recorded torch versions when using nightlies
> - Marks `docker/Dockerfile.nightly_torch` as deprecated in favor of the main Dockerfile + flag
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c00a0691687a77994bd613ce4e507f5541096c60. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


atalman (approved):
lgtm

orionr:
> @orionr What is the main motivation for this change? I'm worried it will impact the cache invalidation negatively. For starters, can you add a dummy commit and see how long your rebuild takes and how it affects the cache hits?

If you look at https://github.com/vllm-project/vllm/blob/main/docker/Dockerfile.nightly_torch it has diverged heavily from the core Dockerfile and no longer functional - it's challenging to keep them in sync and provide the benefits added to the core Dockerfile. Generally, I think having these be hardware-specific (Nvidia, AMD, CPU, etc) makes sense, but splitting on library version less so. Also, this sets the foundation to simplify cross-repo releases between PyTorch and vLLM, where there is even more usage of torch.compile and other bits for performance.

I have the same concerns around the caching, which is why I tagged you here and on the other PR. Let me start a chat with you about concerns here as well as on the ci-infra PR.

docker/Dockerfile
```diff
@@ -131,12 +131,36 @@ ARG PYTORCH_CUDA_INDEX_BASE_URL
 
 WORKDIR /workspace
 
-# install build and runtime dependencies
+# We can specify the standard or nightly build of PyTorch
+ARG PYTORCH_NIGHTLY
```
andy-neuma, line 152: cool. how does this impact the image caching?

andy-neuma (approved):
thanks

docker/Dockerfile
orionr: Seems to be unchanged. The non-torch nightly builds look to be caching as expected per @amrmahdi with the rebuild times at ~12 minutes. Unfortunately the torch nightly build artifacts aren't cached at all, but that matches to what we had with the separate Dockerfile (before it stopped working), so this is strictly better. I'd like to see if we can enable caching across all the docker builds (AMD, torch nightly, etc) but that can be separate from this.

docker/Dockerfile
```diff
@@ -131,12 +131,36 @@ ARG PYTORCH_CUDA_INDEX_BASE_URL
...
+        --index-url ${PYTORCH_CUDA_INDEX_BASE_URL}/nightly/cu$(echo $CUDA_VERSION | cut -d. -f1,2 | tr -d '.') \
+        && echo "Installing other requirements..." \
+        && /opt/venv/bin/python3 use_existing_torch.py --prefix \
+        && uv pip install --python /opt/venv/bin/python3 -r requirements/cuda.txt \
+        --extra-index-url ${PYTORCH_CUDA_INDEX_BASE_URL}/nightly/cu$(echo $CUDA_VERSION | cut -d. -f1,2 | tr -d '.'); \
+    else \
+        uv pip install --python /opt/venv/bin/python3 -r requirements/cuda.txt \
+        --extra-index-url ${PYTORCH_CUDA_INDEX_BASE_URL}/cu$(echo $CUDA_VERSION | cut -d. -f1,2 | tr -d '.'); \
+    fi
+
```
amrmahdi, line 173: @orionr Thanks for your patience on this, I've been reviewing the changes, and one idea that might let us simplify the changes is to install torch once per base and use a constraints file to prevent drift, instead of the repeated version checks.

the Dockerfile has two roots:
  - Build base (base)
  - Runtime base (vllm-base):

The idea is:
In base, install torch (nightly or stable) and capture the versions:
```
RUN if [ "${PYTORCH_NIGHTLY}" = "1" ]; then \
        uv pip install torch torchvision torchaudio --pre --index-url .../nightly/...; \
    else \
        uv pip install -r requirements/cuda.txt --extra-index-url ...; \
    fi && \
    uv pip freeze | grep -E '^(torch|torchvision|torchaudio)==' > /opt/torch_constraints.txt
```
In vllm-base, grab the constraints and install the same torch:
```
COPY --from=base /opt/torch_constraints.txt /opt/torch_constraints.txt
RUN if [ "${PYTORCH_NIGHTLY}" = "1" ]; then \
        uv pip install $(cat /opt/torch_constraints.txt | xargs) --pre --index-url .../nightly/...; \
    else \
        uv pip install $(cat /opt/torch_constraints.txt | xargs) --extra-index-url ...; \
    fi
```
Then everywhere else we don't need conditionals and version checks, just:
`RUN uv pip install -c /opt/torch_constraints.txt -r requirements/build.txt`

The constraints file basically becomes the "contract" between build and runtime — guaranteeing they use the same torch.

We could even go a step further and encapsulate the conditional in a small script making the Dockerfile itself completely conditional free.

orionr: Good feedback. FYI that `/opt/torch_constraints.txt` is basically the same as `torch_lib_versions.txt` so we have it, but you are looking to leverage the constraints flag instead of a separate torch lib version check.

I'm actually wondering, given that I'm now appending torch versions to test.in (see `uv pip compile requirements/test.in` below) this might actually be covering the biggest issue I saw previously, and I should just remove the torch version sections to minimize changes. Let me see if that has tests passing. If so, I'll just do that. If not, I'll see if we can use `-c` and constraints.

amrmahdi: > FYI that /opt/torch_constraints.txt is basically the same as torch_lib_versions.txt

Yup! 

> I'm actually wondering, given that I'm now appending torch versions to test.in (see uv pip compile requirements/test.in below) this might actually be covering the biggest issue I saw previously, and I should just remove the torch version sections to minimize changes. Let me see if that has tests passing. If so, I'll just do that. If not, I'll see if we can use -c and constraints.

What about other steps that don't have a `.in` file ?

orionr: Might need the constraints, but let me check.

orionr: I did some digging on the weekend and I feel like a pip constraints check isn't required anywhere (surprise) after I did that test.in modification and looked closely at all the other install cases. Also ran the tests and didn't see any skew with the current bits.

With that, I think we're good with just removing the checks, which I've done. Everything built and test failures (a small number, not about version skew) are the same as they were with the checks.

amrmahdi: @orionr the constraints approach could still simplify things. Right now the PR has many conditionals and multiple copies of use_existing_torch.py / torch_lib_versions.txt across stages.

But stages that inherit from base already have torch installed, they don't need conditionals. With constraints they'd just be:

  `RUN uv pip install -c /opt/torch_constraints.txt -r requirements/build.txt`

Anything I'm missing?

khluu:
cc @Harry-Chen

orionr:
Heads up that we'd like to land this after I do some final rebuild and cache testing, which should be done by tomorrow. @amrmahdi and I had a good chat today and I'll also add some notes here about directions for standardization and improvements in the future.

docker/Dockerfile
Harry-Chen: Great work! I think caching should be considered to make the nightly tests practical. The fresh build time of vllm and all dependencies can easily go over 6 hours on our build agents, so we can hardly afford a fresh rebuild for any PR updates. For example, #30908 would very much need the nightly tests.

orionr: Working on caching for these is absolutely a next step. I just added some notes at https://github.com/vllm-project/vllm/pull/30443#issuecomment-3756334148 around things to work on once this lands.

Harry-Chen (approved):

PR #30448, opened 2025-12-11
[NIXL] Heterogeneous KV Layout and block_size - prefill NHD and nP > nD support

xuechendi:
## Purpose

Proposal codes for #26744

This PR is to support non-host-buffer path Heterogeneous KV Layout and block_size - prefill NHD and nP > nD
Only support when prefix_cache is disabled

This PR includes two bug fix PR, pending on them merged firstly
https://github.com/vllm-project/vllm/pull/30420
https://github.com/vllm-project/vllm/pull/30419

## Test Plan

1P 1D with nP > nD + Prefill with NHD
```
DISABLE_PREFIX_CACHE=true AGREED_BLOCK_SIZE=16  PREFILL_KV_LAYOUT=NHD PREFILL_BLOCK_SIZE=64 DECODE_BLOCK_SIZE=16 bash tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh
```
=> Accuracy passed Qwen3-0.6= ~0.41

P1D2 + prefill with NHD
```
DECODER_TP_SIZE=2 DISABLE_PREFIX_CACHE=true AGREED_BLOCK_SIZE=16  PREFILL_KV_LAYOUT=NHD PREFILL_BLOCK_SIZE=64 DECODE_BLOCK_SIZE=16 bash tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh
```
Accuracy passed Qwen3-0.6= ~0.41

## Design

This PR is to propose update KV_layout and block_size at prefill side after request complete prefill fwd

case 1, very naive case, the proposal can work properly
```
fwd_prefill => KV update(NHD to HND and blk_size = 16) 
=> send finish to decoder => decoder copy data correctly
```

Case 2, with chunked prefill

```
fwd partial prefill => temp save block_ids to nixl_connector_worker
=> second partial fwd => KV update(NHD to HND and blk_size = 16) 
=> send finish to decoder => decoder copy data correctly
```

## Changes proposed in this PR:
1. add new metadata `block_size_after_save` and `kv_layout_after_save` => If prefill is using different setting to decode, the `block_size_after_save` and `kv_layout_after_save` will use agreed block_size and 'HND' respectively.
2. This feature will only get enabled when prefix_cache is disabled, and `enable_permute_local_kv` is on.
3. Add a new `kv_caches_postprocess` function, and called at `wait_for_save`

---
<details>

</details>

### DIFF
--- a/vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
+++ b/vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
@@ -29,7 +29,10 @@ from vllm.distributed.kv_transfer.kv_connector.utils import (
     get_current_attn_backend,
     kv_postprocess_blksize_and_layout_on_receive,
     kv_postprocess_blksize_on_receive,
+    kv_postprocess_blksize_on_save,
+    kv_postprocess_layout_and_blksize_on_save,
     kv_postprocess_layout_on_receive,
+    kv_postprocess_layout_on_save,
     yield_req_data,
 )
 from vllm.distributed.kv_transfer.kv_connector.v1.base import (
@@ -250,6 +253,66 @@ class ReqMeta:
     remote: RemoteMeta | None = None
 
 
+def if_postprocess_kvcache_on_save(
+    vllm_config, current_block_size, current_kv_cache_layout
+):
+    assert vllm_config.kv_transfer_config.enable_permute_local_kv
+    if vllm_config.cache_config.enable_prefix_caching:
+        logger.warning_once(
+            "KV cache postprocess is not compatible with prefix caching."
+        )
+        return False, current_kv_cache_layout, current_block_size
+    postprocess_kv_caches_on_save = False
+    kv_cache_layout_on_save = "HND"
+    agreed_block_size = int(
+        vllm_config.kv_transfer_config.get_from_extra_config(
+            "agreed_block_size", current_block_size
+        )
+    )
+    # Only allow save to smaller block size (larger required additional allocation)
+    block_size_on_save = (
+        agreed_block_size
+        if agreed_block_size <= current_block_size
+        else current_block_size
+    )
+    if (
+        kv_cache_layout_on_save != current_kv_cache_layout
+        or block_size_on_save != current_block_size
+    ):
+        postprocess_kv_caches_on_save = True
+        logger.info(
+            "KV cache postprocess on save is enabled. "
+            "Local kv cache layout: %s -> %s, "
+            "block size: %d -> %d",
+            current_kv_cache_layout,
+            kv_cache_layout_on_save,
+            current_block_size,
+            block_size_on_save,
+        )
+    return postprocess_kv_caches_on_save, kv_cache_layout_on_save, block_size_on_save
+
+
+def get_mapped_blocks(block_ids, block_size_ratio, num_blocks):
+    """
+        Calculates the new set of block IDs by mapping every element
+        in the (potentially sparse) input array.
+        Example: block_ids=[0, 2], block_size_ratio=2
+    get_mapped_blocks    0     1     [2     3]     4     5
+            # remote is |h0-b0|h1-b0||h0-b1|h1-b1||h0-b1|h1-b1||
+            # local is  |h0-b0......||h1-b0......||h2-b0........
+    local_block_ids         0           [1]           2
+    """
+    if block_ids.size == 0:
+        return []
+
+    start_ids = block_ids * block_size_ratio
+    offsets = np.arange(block_size_ratio)
+    mapped_2d = start_ids[:, None] + offsets[None, :]
+    ret = mapped_2d.flatten().tolist()[:num_blocks]
+
+    return ret
+
+
 class NixlConnectorMetadata(KVConnectorMetadata):
     def __init__(self):
         self.reqs_to_recv: dict[ReqId, ReqMeta] = {}
@@ -460,6 +523,8 @@ class NixlConnector(KVConnectorBase_V1):
         assert isinstance(self._connector_metadata, NixlConnectorMetadata)
         if self.connector_worker.use_host_buffer and self.connector_worker.copy_blocks:
             self.connector_worker.save_kv_to_host(self._connector_metadata)
+        elif self.connector_worker.postprocess_kv_caches_on_save:
+            self.connector_worker.kv_caches_postprocess(self._connector_metadata)
 
     def shutdown(self):
         if self.connector_worker is not None:
@@ -487,6 +552,7 @@ class NixlConnectorScheduler:
     def __init__(self, vllm_config: VllmConfig, engine_id: str):
         self.vllm_config = vllm_config
         self.block_size = vllm_config.cache_config.block_size
+        self.kv_cache_layout = get_kv_cache_layout()
         self.engine_id: EngineId = engine_id
         self.side_channel_host = envs.VLLM_NIXL_SIDE_CHANNEL_HOST
         self.side_channel_port = (
@@ -501,6 +567,21 @@ class NixlConnectorScheduler:
                 vllm_config.kv_transfer_config.kv_buffer_device == "cpu"
             )
 
+        self.postprocess_kv_caches_on_save = False
+        self.kv_cache_layout_on_save = self.kv_cache_layout
+        self.block_size_on_save = self.block_size
+
+        # list of chunked prefill partials
+        self.partial_reqs: dict[ReqId, list] = {}
+
+        if vllm_config.kv_transfer_config.enable_permute_local_kv:
+            (
+                self.postprocess_kv_caches_on_save,
+                self.kv_cache_layout_on_save,
+                self.block_size_on_save,
+            ) = if_postprocess_kvcache_on_save(
+                self.vllm_config, self.block_size, self.kv_cache_layout
+            )
         logger.info("Initializing NIXL Scheduler %s", engine_id)
 
         # Background thread for handling new handshake requests.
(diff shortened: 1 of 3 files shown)


xuechendi:
> @xuechendi @markmc Other than verifying functional correctness of the logic introduced here, I think it's also worthwhile to also check the e2e performance implications of it. P/D is all about improving performance, so in theory some P->D flows are better to be handshake-failed instead of proceeding.

@orozery, thanks for sharing your thought.
This path is for our heterogeneous arch proposal for Intel Gaudi + cuda. The specific PR is for prefill with Gaudi and decode with CUDA. (where Gaudi only support NHD and more performant with block_size=128)
Our performance team reported such setting bring similar performance comparing to H200 to H200 with way better TCO story.
For H200 to H200, it is not necessary to do hetero block_size or kv_layout, so we explicitly asked for "enable_permute_local_kv=true" to set in kv_connector_config to enable this feature.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -460,6 +523,8 @@ def wait_for_save(self):
         assert isinstance(self._connector_metadata, NixlConnectorMetadata)
         if self.connector_worker.use_host_buffer and self.connector_worker.copy_blocks:
             self.connector_worker.save_kv_to_host(self._connector_metadata)
+        elif self.connector_worker.postprocess_kv_caches_on_save:
+            self.connector_worker.kv_caches_postprocess(self._connector_metadata)
```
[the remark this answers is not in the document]
xuechendi: This is intentional in this PR

xuechendi:
@NickLucche , may you help to review this PR to support nP > nD case

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -250,6 +253,66 @@ class ReqMeta:
...
 
+def if_postprocess_kvcache_on_save(
+    vllm_config, current_block_size, current_kv_cache_layout
+):
+    assert vllm_config.kv_transfer_config.enable_permute_local_kv
+    if vllm_config.cache_config.enable_prefix_caching:
+        logger.warning_once(
+            "KV cache postprocess is not compatible with prefix caching."
+        )
+        return False, current_kv_cache_layout, current_block_size
```
NickLucche, line 264: shouldn't we check this during handshake validation? I think we should crash somewhere if we're in hetero-block size scenario with unsupported config

NickLucche: same for these other constraints you listed 
> prefill NHD and nP > nD

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1348,12 +1470,22 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
                     "All kv cache tensors must have the same number of blocks"
                 )
 
+                block_size_ratio_on_save = self.block_size // self.block_size_on_save
```
NickLucche: this is constant and should be computed once after this loop detectes any logical<>kernel block size mismatch

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1348,12 +1470,22 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
...
                 self.block_len_per_layer.append(
                     curr_tensor_size_bytes // self.num_blocks
                 )
                 self.slot_size_per_layer.append(
                     self.block_len_per_layer[-1] // self.block_size
                 )
+                block_len_per_layer_on_save.append(
+                    curr_tensor_size_bytes
+                    // self.num_blocks
+                    // block_size_ratio_on_save
```
NickLucche: this is basically `block_len_per_layer // block_size_ratio_on_save` where denominator is some constant. 
We don't need a separate bookkeeping struct for it, just compute it on the fly from `block_len_per_layer`

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1348,12 +1470,22 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
...
+                )
+                num_blocks_on_save = (
+                    curr_tensor_size_bytes // block_len_per_layer_on_save[-1]
+                )
```
NickLucche: ditto, this should not be computed in a for loop. 
It can also be simplified to 
```
num_blocks = curr_tensor_size_bytes / N 
num_blocks_on_save = curr_tensor_size_bytes / (N / block_size_ratio_on_save)
=>num_blocks_on_save = curr_tensor_size_bytes/N *block_size_ratio_on_save

==>num_blocks_on_save = num_blocks * block_size_ratio_on_save

```

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -1400,9 +1532,10 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
 
         # Register local/src descr for NIXL xfer.
         self.seen_base_addresses = seen_base_addresses
-        self.src_xfer_handles_by_block_size[self.block_size], self.src_blocks_data = (
-            self.register_local_xfer_handler(self.block_size)
-        )
+        (
+            self.src_xfer_handles_by_block_size[self.block_size_on_save],
+            self.src_blocks_data,
+        ) = self.register_local_xfer_handler(self.block_size_on_save)
```
NickLucche, line 1539: is this pre-commit or..?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -990,6 +1101,16 @@ def __init__(self, vllm_config: VllmConfig, engine_id: str):
         self.enforce_compat_hash = self.kv_transfer_config.get_from_extra_config(
             "enforce_handshake_compat", True
         )
+        self.kv_cache_layout_on_save = self.kv_cache_layout
+        self.block_size_on_save = self.block_size
+        if self.kv_transfer_config.enable_permute_local_kv:
+            (
+                self.postprocess_kv_caches_on_save,
+                self.kv_cache_layout_on_save,
+                self.block_size_on_save,
```
NickLucche, line 1110: some of this should go into the compat check right?

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -250,6 +253,66 @@ class ReqMeta:
...
+            "KV cache postprocess is not compatible with prefix caching."
+        )
+        return False, current_kv_cache_layout, current_block_size
+    postprocess_kv_caches_on_save = False
+    kv_cache_layout_on_save = "HND"
+    agreed_block_size = int(
+        vllm_config.kv_transfer_config.get_from_extra_config(
+            "agreed_block_size", current_block_size
+        )
+    )
```
NickLucche, line 271: do I get this right, this is an agreement through config *before* handshake, which is actually breaking the current flow as per all similar logic is discovery-based: configs are matched in between instances and evaluated at runtime. 

do we have alternatives to this pre-agreement? It may also end up looking as "custom" wrt current ux @xuechendi 
cc @markmc

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -250,6 +253,66 @@ class ReqMeta:
     remote: RemoteMeta | None = None
 
 
+def if_postprocess_kvcache_on_save(
```
NickLucche: not a great function name, better to change to `should_postprocess_kvcache_on_save` or "needs" or
`should_transform_kv_for_transfer` or similar

NickLucche (commented):
Thanks for the work @xuechendi !

I currently have 2 main issues with current implementation:
- This is making the scheduler kv cache dependent. In its current form, the scheduler is only operating on the "logical" plane on block ids and it is agnostic to layout/block_size and other physical configs. This is a separation of concerns which I quite like and would like to keep unless really necessary. One alternative I would propose for block_size mismatch is to group the changes to the scheduler you're proposing into a Mixin class.
- The "AGREED_BLOCK_SIZE" design is breaking the current workflow which is handshake based: currently 2 peer instances exchange info at runtime and match their config/logic accordingly *after* discovery. Allowing a top-bottom agreement through config may result in feeling "custom" or unordinary. Let's discuss what other options we have to maintain the discovery-based paradigm.

Also, this PR will clash with https://github.com/vllm-project/vllm/pull/32204. Therefore I would like to land it after the latter to address potential conflicts with this feature in this PR rather than on the hma one.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
markmc: @xuechendi could we handle this with an additional factor in `compute_nixl_compatibility_hash()` ?

xuechendi: Yes, @NickLucche , `agreed_block_size` is breaking the current discovery-based design that prefill side will use `agreed_block_size` as its block_size when communicate to decoders.

I tried to search for possible alternatives in my mind, but I really can't come up with any. If we go with my previous PR that let decoder side to permute for large block_size, decoder will end up with allocating temp memory to hold a block_size= 128 buffer when it only has block_size=16* num_blocks=4 buffers.

vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
xuechendi: yes, because I updated `self.block_size` to `self.block_size_on_save`

PR #30449, opened 2025-12-11
[Core] Add SLA-tiered scheduling (opt-in)

ProdByBuddha:
## Purpose
Add opt-in SLA-tiered scheduling to vLLM:
- introduce `SamplingParams.sla_tier` (interactive/batch/background) with validation and default batching to “batch”
- enable SLA-aware scheduling via `SchedulerConfig.sla_tier_enabled` and an interactive batch token cap
- order queue/preemption by SLA tier before priority; add unit coverage and docs for the new behavior

## Test Plan
- python -m pytest tests/v1/core/test_scheduler.py

## Test Result
- Passed locally (CPU-only environment; no GPU available)

## Notes
- Requests can set SamplingParams.sla_tier (defaults to batch if unset)
- When SchedulerConfig.sla_tier_enabled is true, the queue orders by SLA tier first, then priority/arrival
- Preemption respects the tier ordering, and an interactive batch token cap is enforced for interactive requests

Screenshot 2025-12-10 142236
Screenshot 2025-12-10 142210
image


ProdByBuddha:
There's a pesky pre commit error I will resolve in due time

hmellor (changes requested):
Something appears to have gone wrong where diffs are appearing much larger than they should. Have you changed the file permissions or line endings by accident?

PR #30452, opened 2025-12-11
[Core] Optimize encoder cache with mask-based storage

sunYtokki:
## Purpose
Optimizes encoder cache memory usage by storing only actual embeddings instead of full placeholder tensors using `is_embed` masks. This addresses the memory inefficiency identified in https://github.com/vllm-project/vllm/issues/25903

**Key improvements:**
- Stores only embedding positions marked in `is_embed` mask (not full scattered tensors)
- Adds `PlaceholderRange.get_num_embeds()` and `extract_embeds_range()` methods
- Simplifies encoder output gathering logic in `gpu_model_runner.py`
- Achieves significant memory savings (up to 12.5x demonstrated in tests)

RESOLVE #25903

## Test Plan
- Unit test added
```bash
pytest tests/multimodal/test_utils.py::test_placeholder_range_get_num_embeds
pytest tests/multimodal/test_utils.py::test_placeholder_range_extract_embeds_range
pytest tests/v1/core/test_encoder_cache_manager.py::test_encoder_cache_with_is_embed_mask
pytest tests/v1/core/test_encoder_cache_manager.py::test_encoder_cache_mask_based_retrieval
pytest tests/v1/worker/test_gpu_model_runner.py::test_encoder_cache_storage_format
```
- regression check
```bash
python examples/offline_inference/vision_language.py --model-type internvl_chat --model internvl_chat --modality image
python examples/offline_inference/vision_language.py --model-type internvl_chat --model qwen2_vl --modality image
```

- inference benchmark

## Test Result
All unit tests passed. Tests validate:

Correct embedding counting with various mask configurations
Proper range extraction for masked positions
Memory savings calculation (12.5x reduction demonstrated)
Cache storage format consistency (raw tensors, not scattered placeholders)

All regression check on relevant models passed

### Inference benchmark
before change
```sh
(EngineCore_DP0 pid=140497) INFO 12-07 07:34:59 [gpu_worker.py:364] Available KV cache memory: 3.93 GiB
(EngineCore_DP0 pid=140497) INFO 12-07 07:35:00 [kv_cache_utils.py:1287] GPU KV cache size: 28,592 tokens
EngineCore_DP0 pid=140497) INFO 12-07 07:35:00 [kv_cache_utils.py:1292] Maximum concurrency for 4,096 tokens per request: 6.98x
```

after change
```sh
(EngineCore_DP0 pid=142442) INFO 12-07 07:38:32 [gpu_worker.py:364] Available KV cache memory: 8.54 GiB
(EngineCore_DP0 pid=142442) INFO 12-07 07:38:33 [kv_cache_utils.py:1287] GPU KV cache size: 62,192 tokens
(EngineCore_DP0 pid=142442) INFO 12-07 07:38:33 [kv_cache_utils.py:1292] Maximum concurrency for 4,096 tokens per request: 15.18x
```

---
<details>

- [x] The purpose of the PR, such as "Fix issue https://github.com/vllm-project/vllm/issues/25903". 
</details>

### DIFF
--- a/tests/v1/worker/test_gpu_model_runner.py
+++ b/tests/v1/worker/test_gpu_model_runner.py
@@ -1110,17 +1110,3 @@ def test_hybrid_cache_integration(model_runner, dist_init):
     runner._update_states(scheduler_output)
     assert _is_req_scheduled(runner, req_id)
     assert _is_req_state_block_table_match(runner, req_id)
-
-
-def test_encoder_cache_storage_format():
-    encoder_cache = {}
-
-    mock_encoder_output = torch.randn(8, 768)
-
-    mm_hash = "test_image_hash"
-    encoder_cache[mm_hash] = mock_encoder_output
-
-    assert isinstance(encoder_cache[mm_hash], torch.Tensor)
-    assert encoder_cache[mm_hash].shape == (8, 768)
-    assert not isinstance(encoder_cache[mm_hash], dict)
-    assert torch.equal(encoder_cache[mm_hash], mock_encoder_output)
--- a/tests/v1/worker/test_gpu_model_runner.py
+++ b/tests/v1/worker/test_gpu_model_runner.py
@@ -1110,3 +1110,17 @@ def test_hybrid_cache_integration(model_runner, dist_init):
     runner._update_states(scheduler_output)
     assert _is_req_scheduled(runner, req_id)
     assert _is_req_state_block_table_match(runner, req_id)
+
+
+def test_encoder_cache_storage_format():
+    encoder_cache = {}
+
+    mock_encoder_output = torch.randn(8, 768)
+
+    mm_hash = "test_image_hash"
+    encoder_cache[mm_hash] = mock_encoder_output
+
+    assert isinstance(encoder_cache[mm_hash], torch.Tensor)
+    assert encoder_cache[mm_hash].shape == (8, 768)
+    assert not isinstance(encoder_cache[mm_hash], dict)
+    assert torch.equal(encoder_cache[mm_hash], mock_encoder_output)
--- a/vllm/multimodal/inputs.py
+++ b/vllm/multimodal/inputs.py
@@ -188,8 +188,6 @@ class PlaceholderRange:
             Returns full placeholder range if `is_embed` is `None`.
         """
         if self.is_embed is None:
-            if self.length == 0:
-                return []
             return [(self.offset, self.offset + self.length - 1)]
 
         mask_i = self.is_embed.int()
--- a/vllm/multimodal/inputs.py
+++ b/vllm/multimodal/inputs.py
@@ -5,7 +5,7 @@ from abc import ABC, abstractmethod
 from collections import UserDict, defaultdict
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
-from functools import partial
+from functools import cached_property, partial
 from itertools import accumulate
 from typing import (
     TYPE_CHECKING,
@@ -174,6 +174,48 @@ class PlaceholderRange:
             return self.length
 
         return int(self.is_embed.sum().item())
+
+    @cached_property
+    def embeds_cumsum(self) -> Optional["torch.Tensor"]:
+        """Cumulative sum of is_embed mask for efficient range queries.
+
+        Returns:
+            Cumulative sum tensor if is_embed mask exists, None otherwise.
+        """
+        if self.is_embed is None:
+            return None
+
+        return self.is_embed.cumsum(dim=0)
+
+    def get_embeds_indices_in_range(
+        self, start_idx: int, end_idx: int
+    ) -> tuple[int, int]:
+        """Map token range [start_idx, end_idx) to embedding indices.
+
+        Uses the cumulative sum of the is_embed mask to efficiently determine
+        which embeddings from the encoder output correspond to a given token
+        range within this placeholder.
+
+        Args:
+            start_idx: Start of token range (relative to placeholder offset).
+            end_idx: End of token range (exclusive, relative to placeholder offset).
+
+        Returns:
+            Tuple of (embeds_start_idx, embeds_end_idx) representing the
+            range of embeddings needed from encoder output.
+        """
+        assert start_idx <= end_idx
+
+        if self.embeds_cumsum is None:
+            # No sparse mask - all tokens are embeddings
+            return start_idx, end_idx
+
+        embeds_start_idx = (
+            int(self.embeds_cumsum[start_idx - 1]) if start_idx > 0 else 0
+        )
+        embeds_end_idx = int(self.embeds_cumsum[end_idx - 1])
+
+        return embeds_start_idx, embeds_end_idx
 
     def extract_embeds_range(self) -> list[tuple[int, int]]:
         """Extract the start and end indices of the embedded region in prompt.
--- a/vllm/multimodal/inputs.py
+++ b/vllm/multimodal/inputs.py
@@ -188,7 +188,9 @@ class PlaceholderRange:
             Returns full placeholder range if `is_embed` is `None`.
         """
         if self.is_embed is None:
-            return [(self.offset, self.offset + self.length)]
+            if self.length == 0:
+                return []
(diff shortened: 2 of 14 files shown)


tests/v1/worker/test_gpu_model_runner.py
```diff
@@ -1110,3 +1110,17 @@ def test_hybrid_cache_integration(model_runner, dist_init):
...
     assert _is_req_state_block_table_match(runner, req_id)
+
+
+def test_encoder_cache_storage_format():
+    encoder_cache = {}
+
+    mock_encoder_output = torch.randn(8, 768)
+
+    mm_hash = "test_image_hash"
+    encoder_cache[mm_hash] = mock_encoder_output
```
DarkLight1337: I think this test is rather redundant

DarkLight1337 (commented):
Otherwise the PR looks good. Can you run some evals to verify the correctness? I am not sure whether we have evals for video inputs, so maybe you can just use an image eval on a relevant model like InternVL.

cc @ywang96

liangwen12year (approved):

vllm/multimodal/inputs.py
```diff
@@ -188,7 +188,9 @@ def extract_embeds_range(self) -> list[tuple[int, int]]:
             Returns full placeholder range if `is_embed` is `None`.
         """
         if self.is_embed is None:
-            return [(self.offset, self.offset + self.length)]
+            if self.length == 0:
+                return []
```
ywang96, line 234: I don't think we need to worry about the case when length is 0, but thank you for fixing this.

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -166,9 +166,7 @@
     MultiModalBudget,
     add_kv_sharing_layers_to_kv_cache_groups,
     bind_kv_cache,
-    gather_mm_placeholders,
     sanity_check_mm_encoder_outputs,
-    scatter_mm_placeholders,
```
ywang96, line 174: We can deprecate these two methods now

vllm/multimodal/inputs.py
Isotr0py, line 234: I think we can move this out of `if self.is_embed is None:`:
```python3
if self.length == 0:
    return []

if self.is_embed is None:
    ...
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2245,17 +2240,29 @@ def _gather_mm_embeddings(
...
+                    # Calculate the number of embeddings before start_idx using the mask
+                    num_embeds_before = (
+                        is_embed[:start_idx].sum().item() if start_idx > 0 else 0
+                    )
+                    num_embeds_in_range = is_embed[start_idx:end_idx].sum().item()
+
+                    # Slice the cached encoder output using the actual embedding indices
+                    mm_embeds_item = encoder_output[
+                        num_embeds_before : num_embeds_before + num_embeds_in_range
+                    ]
```
ywang96, line 2254: IMO we should create a cumulative sum of embeddings per position as an attribute of `mm_position`/`PlaceholderRange` in its `__post_init__` instead of computing it at runtime/enginecore
```python
class PlaceholderRange:
    def __post_init__(self):
        if is_embed is not None:
            object.__setattr__(self, "embed_cumsum", is_embed.cumsum(dim=0))
        else:
            object.__setattr__(self, "embed_cumsum", None)
...
```
then here we can simply do something like

```python
embed_cumsum = pos_info.embed_cumsum
num_embeds_before = int(embed_cumsum[start_idx - 1]) if start_idx > 0 else 0
num_embeds_in_range = int(embed_cumsum[end_idx - 1]) - num_embeds_before 
mm_embeds_item = encoder_output[
    num_embeds_before : num_embeds_before + num_embeds_in_range
]
```
We should also make this code snippet a method of `PlaceholderRange` instead of having it inside model runner.

ywang96 (changes requested):
Thanks for the contribution - I left some comments and please take a look.

Could you also run some benchmarks on models to ensure this does not cause performance regression?

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2245,17 +2240,29 @@ def _gather_mm_embeddings(
...
+                    is_embed_slice = is_embed[start_idx:end_idx]
+                else:
+                    # No mask: all positions are embeddings
+                    mm_embeds_item = encoder_output[start_idx:end_idx]
+                    is_embed_slice = None
 
                 req_start_pos = req_start_idx + start_pos - num_computed_tokens
                 is_mm_embed[req_start_pos + start_idx : req_start_pos + end_idx] = (
-                    True if is_embed is None else is_embed
+                    True if is_embed_slice is None else is_embed_slice
```
ywang96: Why did we need to rename this to `is_embed_slice`?

vllm/v1/worker/gpu_model_runner.py
DarkLight1337: We probably don't want to construct this on P0, let's make it a cached property and only access it in P1. (Not sure how would this interact with serialization/deserialization though)

ywang96: > We probably don't want to construct this on P0

Why so?

DarkLight1337: It introduces additional serialization/deserialization cost. After thinking a bit more, I think we should run some benchmarks to compare the two approaches. Maybe this can be done in another PR?

ywang96: Let me open a separate PR for that approach then

vllm/multimodal/inputs.py
sunYtokki: make sense will address this in the next revision

vllm/v1/worker/gpu_model_runner.py
sunYtokki: got it. I will remove it

vllm/v1/worker/gpu_model_runner.py
sunYtokki: good catch. will use `is_embed` in the next revision

tests/v1/worker/test_gpu_model_runner.py
sunYtokki: agreed. will remove in the next revision

vllm/v1/worker/gpu_model_runner.py
ryansherby: Hey @ywang96, I built a version of this using embed_cumsum as a cached property.

I'll commit to that new PR and run some tests to see if performance improves 👍

ywang96: @ryansherby Hey I actually opened #30475 already for this - see if it works for you

sunYtokki:
> I just realized this PR still does not align encoder cache budget with the actual space...

I had a seperate PR to align encoder cache bedet https://github.com/vllm-project/vllm/pull/30536

vllm/v1/worker/gpu_model_runner.py
sunYtokki: I had to restore those utility functions with a note because XPU and HPU still depend on them, and removing them causes import errors. It looks like this requires coordination with external packages. I’ll open an issue in https://github.com/vllm-project/vllm-gaudi once this change is merged.

DarkLight1337: Let's add a deprecation to notify them that these methods will be removed soon

sunYtokki: Updated!

vllm/v1/worker/utils.py
```diff
@@ -215,6 +216,14 @@ def scatter_mm_placeholders(
             tokens need to be filled with multimodal embeddings.
             Shape: `(num_placeholders, num_embeds)`
     """
+    warnings.warn(
```
DarkLight1337, line 226: Use `typing_extensions.deprecated`

sunYtokki: updated!

PR #30456, opened 2025-12-11
feat: support video list inference

LiuLi1998:
## Purpose

Support video frame input for multimodal model

### DIFF
--- a/vllm/entrypoints/chat_utils.py
+++ b/vllm/entrypoints/chat_utils.py
@@ -2,6 +2,7 @@
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 
 import asyncio
+import base64
 import inspect
 import json
 from abc import ABC, abstractmethod
@@ -1305,6 +1306,7 @@ MM_PARSER_MAP: dict[
     "input_audio": lambda part: _InputAudioParser(part).get("input_audio", None),
     "refusal": lambda part: _RefusalParser(part).get("refusal", None),
     "video_url": lambda part: _VideoParser(part).get("video_url", {}).get("url", None),
+    "video": lambda part: _VideoParser(part).get("video", []),
 }
 
 
@@ -1516,6 +1518,18 @@ def _parse_chat_message_content_part(
         modality = "audio"
     elif part_type == "video_url":
         str_content = cast(str, content)
+        mm_parser.parse_video(str_content, uuid)
+        modality = "video"
+    elif part_type == "video":
+        str_content = "data:video/jpeg;base64,"
+        list_content = cast(list[str], content)
+        base64_list = []
+        for content in list_content:
+            with open(content, "rb") as f:
+                data = f.read()
+                base64_list.append(base64.b64encode(data).decode('utf-8'))
+        str_content += ",".join(base64_list)
+
         mm_parser.parse_video(str_content, uuid)
         modality = "video"
     else:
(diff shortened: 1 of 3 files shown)


vllm/entrypoints/chat_utils.py
```diff
@@ -1516,6 +1518,18 @@ def _parse_chat_message_content_part(
...
+        mm_parser.parse_video(str_content, uuid)
+        modality = "video"
+    elif part_type == "video":
+        str_content = "data:video/jpeg;base64,"
+        list_content = cast(list[str], content)
+        base64_list = []
+        for content in list_content:
+            with open(content, "rb") as f:
+                data = f.read()
+                base64_list.append(base64.b64encode(data).decode('utf-8'))
```
jeremyteboul: can you use pybase64 which is 5-10x faster , especially with long video ?

vllm/multimodal/video.py
```diff
@@ -303,7 +303,12 @@ def load_base64(
 
             return np.stack(
                 [np.asarray(load_frame(frame_data)) for frame_data in data.split(",")]
-            ), {}
+            ), {
+                "total_num_frames": len(data.split(",")),
+                "frames_indices": list(range(len(data.split(",")))),
+                "fps": 24,
+                "duration": len(data.split(",")) // 24,
```
jeremyteboul: There is no reason to have a default to 24 fps

vllm/entrypoints/chat_utils.py
```diff
@@ -1305,6 +1306,7 @@ def _get_full_multimodal_text_prompt(
     "input_audio": lambda part: _InputAudioParser(part).get("input_audio", None),
     "refusal": lambda part: _RefusalParser(part).get("refusal", None),
     "video_url": lambda part: _VideoParser(part).get("video_url", {}).get("url", None),
+    "video": lambda part: _VideoParser(part).get("video", []),
```
jeremyteboul: why only video if other modality use input_image or input_audio ?

LucasWilkinson:
cc @ywang96 

@LiuLi1998 please fix the pre-commit

vllm/entrypoints/chat_utils.py
LiuLi1998: Thx for comments, fixed

vllm/entrypoints/chat_utils.py
LiuLi1998: I think 'video_frames' would be a better name. I'm also confused about why 'input_image' and 'input_audio' are used.

vllm/multimodal/video.py
LiuLi1998: I’ll try to find a better way to set the FPS. Any suggestions would be appreciated!

vllm/entrypoints/chat_utils.py
```diff
@@ -1516,6 +1518,18 @@ def _parse_chat_message_content_part(
...
     elif part_type == "video_url":
         str_content = cast(str, content)
+        mm_parser.parse_video(str_content, uuid)
+        modality = "video"
+    elif part_type == "video":
+        str_content = "data:video/jpeg;base64,"
+        list_content = cast(list[str], content)
+        base64_list = []
+        for content in list_content:
+            with open(content, "rb") as f:
```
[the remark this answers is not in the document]
DarkLight1337: Can you address this?

DarkLight1337 (commented):
Sorry for the delayed review! Can you also add a simple test and update the docs to demonstrate how to use this?

PR #30459, opened 2025-12-11
set assume_32bit_indexing and pass unbacked hints

laithsakka:
1. Those are needed for unbacked perf along with the changes on top of https://github.com/pytorch/pytorch/pull/170060 . With both changes unbaked perf for ```Qwen/Qwen2-1.5B-Instruct``` matches that of backed.


vllm/compilation/decorators.py
```diff
@@ -488,6 +500,12 @@ def patched_inline_call(self_):
         if ds_type == DynamicShapesType.BACKED_SIZE_OBLIVIOUS:
             fx_config_patches["backed_size_oblivious"] = True
 
+        # Prepare inductor config patches
+        # assume_32bit_indexing is only available in torch 2.10.0.dev+
+        inductor_config_patches = {}
+        if is_torch_equal_or_newer("2.10.0.dev"):
+            inductor_config_patches["assume_32bit_indexing"] = True
```
zou3519, line 507: does this add a guard or assertion of any sort?
we also know what the max size of the symbolic dimension is, it is equal to max_num_batched_tokens. Can we just check if this value is less than the 32bit int max, and if so, set it?

laithsakka: for Q1: oh yeh this still guards actually let me update the summary , but for unbacked it does not guard but add runtime assertions. without assume_32bit_indexing for unbacked we would use 64 bit indexing. 
maybe we need assume_32bit_indexing_no_guard ? as well in the future if we want to freeze shape env and throw errors at specializations in inductor. 

for Q2: the problem is that we know the max input size, but we then we have intermediate tensors that have sizes that are different are expressions of those sizes so we do not know what we want to assert on . does that make sense?

zou3519: okay I see

I'm happy as long as there are runtime assertions, so that we know we messed up

zou3519 (approved):

PR #30478, opened 2025-12-11
fix gme model do not use mrope

zhuikefeng986285005-byte:
## Purpose
Fix the inconsistency between vLLM and Transformers when running the GME model. 

When executing the GME model with image inputs using Transformers, its behavior diverges from vLLM due to inconsistent handling of M-RoPE. Specifically, the GME model does not explicitly specify the use of M-RoPE; 
image

however, vLLM currently assumes M-RoPE should be used based on its internal logic, leading to this inconsistency. 
image

This commit resolves the issue by detecting the presence of the "GME" keyword via string matching during model loading and forcibly removing any M-RoPE–related fields, thereby ensuring M-RoPE is definitively disabled.
## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/models/qwen2_vl.py
+++ b/vllm/model_executor/models/qwen2_vl.py
@@ -1122,6 +1122,12 @@ class Qwen2VLMultiModalProcessor(BaseMultiModalProcessor[Qwen2VLProcessingInfo])
             self.info.get_hf_config().vision_config.spatial_merge_size
         )(hf_inputs)
 
+def disable_gme_mrope(config):
+    contains_gme = any('gme' in str(arch).lower() for arch in config.architectures)
+    if contains_gme:
+        if hasattr(config, 'rope_scaling') and config.rope_scaling is not None:
+            if "mrope_section" in config.rope_scaling:
+                del config.rope_scaling["mrope_section"]
 
 @MULTIMODAL_REGISTRY.register_processor(
     Qwen2VLMultiModalProcessor,
@@ -1276,6 +1282,7 @@ class Qwen2VLForConditionalGeneration(
         quant_config = vllm_config.quant_config
         multimodal_config = vllm_config.model_config.multimodal_config
 
+        disable_gme_mrope(config)
         self.use_data_parallel = multimodal_config.mm_encoder_tp_mode == "data"
         self.config = config
         self.multimodal_config = multimodal_config


hmellor:
Can you please help me understand, if this model doesn't use `mrope`, why does the config specify that it should?

zhuikefeng986285005-byte:
> If the model shouldn't actually be using `mrope` should we make a PR on the model repo itself?

You're absolutely right. After taking a closer look at the Transformers code, I realized that it doesn't apply M-RoPE to position_ids. Instead, it applies M-RoPE later to position_embeddings. However, vLLM directly applies M-RoPE to position_ids, which causes the values of position_ids to differ between the two implementations, ultimately leading to misalignment in the final output consistency.

In the figure, the first red box does not apply M-RoPE to position_ids, while the second red box applies M-RoPE to position_embeddings.
image

Below is the code executed in decoder_layer().
image

zhuikefeng986285005-byte:
> @zhuikefeng986285005-byte hey, can you share a minimal code of how you got the position ids? In qwen models the position ids are supposed to be a bit "duplicated", because each grid in an image shares the same position. That is what 3D RoPE means and looks like vLLM got it right. For a better ex with visual see the [doc here](https://github.com/huggingface/transformers/blob/1dc69bd6f7f5dcc1a32467be505dd873d573846e/src/transformers/models/qwen2_vl/modeling_qwen2_vl.py#L971-L980)
> 
> So can you share the code and I can tell if that is expected or not? Could be that you are feeding the inputs differently to transformers and to vLLM

[get_rope_index](https://github.com/huggingface/transformers/blob/1dc69bd6f7f5dcc1a32467be505dd873d573846e/src/transformers/models/qwen2_vl/modeling_qwen2_vl.py#L851-L854) is only used when there is no attention mask and no position ids. So in my example, get_rope_index was not executed. 

Below is my test code, and the inputs are identical. I used PDB to print the position_ids in the program rather than obtaining them through code.

```python
from transformers import AutoModel
import os
os.environ["TOKENIZERS_PARALLELISM"] = "false"
from PIL import Image
from vllm import LLM
import torch

model_path='path/to/gme-Qwen2-VL-2B-Instruct'
prompt = 'Represent the given text.'
images = ['./6921031838579065050.jpg']

#加载GME
gme = AutoModel.from_pretrained(model_path,torch_dtype="float16", device_map='cuda', trust_remote_code=True)
model = LLM(model=model_path, task="embed",dtype='float16',trust_remote_code=True)

import torch.nn.functional as F
from collections.abc import Sequence
def check_embeddings_close(
    *,
    embeddings_0_lst: Sequence[list[float]],
    embeddings_1_lst: Sequence[list[float]],
    name_0: str,
    name_1: str,
    tol: float = 1e-3,
) -> None:
    assert len(embeddings_0_lst) == len(embeddings_1_lst)

    for prompt_idx, (embeddings_0, embeddings_1) in enumerate(
            zip(embeddings_0_lst, embeddings_1_lst)):
        assert len(embeddings_0) == len(embeddings_1), (
            f"Length mismatch: {len(embeddings_0)} vs. {len(embeddings_1)}")

        sim = F.cosine_similarity(torch.tensor(embeddings_0),
                                  torch.tensor(embeddings_1),
                                  dim=0)

        fail_msg = (f"\nTest{prompt_idx}:"
                    f"\nsim:\t{sim}"
                    f"\n{name_0}:\t{embeddings_0[:16]!r}"
                    f"\n{name_1}:\t{embeddings_1[:16]!r}")

        assert sim >= 1 - tol, fail_msg

def comp_data(a, b):
    print(torch.equal(a, b))           # False（浮点不严格相等）
    print(torch.allclose(a, b, atol=1e-3, rtol=1e-3))       # True（在容忍范围内）
    print(torch.all(a == b).item())    # False（逐元素比较失败）
    diff = a - b
    abs_diff = diff.abs()
    print(diff)
    print("最大绝对误差:", abs_diff.max().item())
    print("均方误差 (MSE):", (diff ** 2).mean().item())    

def comp_img():
    e_image = gme.get_image_embeddings(images=images).squeeze()

    instruction = 'You are a helpful assistant.'
    input_str = '<|vision_start|><|image_pad|><|vision_end|>'
    true_prompt = f'<|im_start|>system\n{instruction}<|im_end|>\n<|im_start|>user\n{input_str}<|im_end|>\n<|im_start|>assistant\n<|endoftext|>'
    mm_data = {}
    image = Image.open(images[0]).convert('RGB')
    mm_data["image"] = image
    
    outputs = model.embed({
        "prompt": true_prompt,
        "multi_modal_data": mm_data,
    })
    embeds = outputs[0].outputs.embedding
    vllm_output = torch.tensor(embeds, dtype=torch.float16)

    #对比
    comp_data(e_image, vllm_output)
    check_embeddings_close(
        embeddings_0_lst=[e_image],
        embeddings_1_lst=[vllm_output],
        name_0="hf",
        name_1="vllm",
    )
comp_img()

```

zucchini-nlp:
Thanks @hmellor !

I see that the authors of GME recommend to use an older version on qwen2, where `Qwen2VLModel` was a language model backbone. In other words, it is not supposed to prepare mrope position or handle multimodal data. In newer version of transformers we have refactored it to be more intuitive and `Qwen2VLModel` is a base multimodal model. 

So, what GME authors either didn't know and assumed that calling a `Qwen2VLModel` just works out of the box (which is not correct), or they did it on purpose and decided to skip multimodal position id preparation step. In any case, we have to ask the authors to be sure. @zhuikefeng986285005-byte I suggest to open an issue in GME rope page on the hub

vLLM implementation uses mRoPE and creates correct position ids, as outlines in the official qwen paper.

PR #30490, opened 2025-12-11
[DeepSeek V3.2] Proper drop_thinking logic

vladnosiv:
## Purpose

DeepSeek's [technical report](https://arxiv.org/pdf/2512.02556) describes the context management logic, which describes a mechanism for discarding thinking blocks when receiving a new message with the user role.

The current vLLM code never drops thinking blocks if the tools are passed, which leads not only to inefficiency in tokens, but also to a drop in FC metrics on tau^2 airline bench:

## Test Plan

Run Tau^2 Airline bench.

## Test Result

| Setup | Tau² Airline Pass^1 |
|-------|-------------------|
| drop_thinking=False | 57.5% |
| drop_thinking=last_is_user | 62.3% |

image


chaunceyjiang (approved):
Thanks~ @vladnosiv

PR #30513, opened 2025-12-11
[CI] Speed up fusion tests

LucasWilkinson:
Speed-up fusion tests by trimming model size and using dummy weights

Based off: https://github.com/vllm-project/vllm/pull/32489 merge that first

Main
image

PR
image


ProExpertProg (commented):
Can we leave at least one real model in with real weights and layers? The whole point of these tests is to confirm this works E2E. It can be limited to nightly only and just the tp2 are fusion

LucasWilkinson:
I think it would make sense to add separate nightly tests that actually test accuracy with a full model (and maybe a more limited set of fusions); but I think that would be a separate PR

PR #30518, opened 2025-12-11
Don't compile vision encoder for Transformers backend

hmellor:
vLLM doesn't compile the vision encoder for multimodal models, but when models are loaded with the Transformers modeling backend, the entire model is decorated for compilation.

This PR fixes this so that:

- for text only models, only the base model is decorated for compilation (i.e. the language modeling head is excluded)
- for vision models, the text model is always compiled and the vision model is only compiled when `-cc.compile-mm-encoder true` is passed

This will make using vision models with vLLM via the Transformers modeling backend behave as they would if they were loaded with the vLLM modeling backend.

---

Other useful PRs:

- https://github.com/vllm-project/vllm/pull/37345 - means that we don't need to also decorate the encoder with `set_model_tag`
- https://github.com/vllm-project/vllm/pull/37234 - fixes an issue dynamo has with `@can_return_tuple` from Transformers


vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -292,6 +293,9 @@ class MultiModalMixin(SupportsMultiModal, SupportsMRoPE):
     def __init__(self, *, vllm_config: "VllmConfig", prefix: str = ""):
         # Skip SupportsMRoPE.__init__ and call the next class in MRO
         super(SupportsMRoPE, self).__init__(vllm_config=vllm_config, prefix=prefix)
+        # Ensure encoder is not compiled
+        encoder = self.model.get_encoder()
+        encoder.__class__ = ignore_torch_compile(type(encoder))
```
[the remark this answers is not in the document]
hmellor: This was intentional, as I can only make the modification after init

vllm/model_executor/models/transformers/base.py
```diff
@@ -198,6 +205,60 @@ def __init__(self, *, vllm_config: "VllmConfig", prefix: str = ""):
...
+
+        logger.debug("Decorating `%s` for torch compile", cls.__name__)
+
+        # Decorate the cls for torch compile
+        @support_torch_compile(dynamic_arg_dims=dynamic_arg_dims, enable_if=enable_if)
+        class SupportTorchCompileWrapper(cls): ...
+
+        # Patch the class in its module
+        module = sys.modules[cls.__module__]
+        setattr(module, cls.__name__, SupportTorchCompileWrapper)
```
Isotr0py: Perhaps update tests under `tests/compile/fullgraph/test_multimodal_compile.py` to include  Transformers backend?

Isotr0py (approved):
Overall LGTM, but would like to let @ProExpertProg double check the compilation side as well.

vllm/model_executor/models/transformers/base.py
hmellor: Good idea, I'll look at adding this test

Lucaskabela (approved):
This LGTM - thanks for the support and wiring through the dynamic dims

Issue #28408, opened 2025-11-10
[Feature]: Generalize RoutingMethodType for broader MoE routing control


mgoin:
### 🚀 The feature, motivation and pitch

PR #27492 introduced `RoutingMethodType` to support different routing methods for FP8 flashinfer TRTLLM MOE (DeepSeekV3, Llama4, Renormalize, etc.).
While this was implemented to support Qwen3 and Qwen3-next models, the review discussion revealed opportunities to use this more broadly across the
codebase to simplify MoE routing configuration.

### Motivation:

Currently, MoE routing behavior is controlled through multiple fragmented parameters (scoring_func, renormalize, use_grouped_topk, custom routing
functions, etc.). This creates several issues:

1. Lack of clarity: The routing method isn't explicitly defined in one place
2. Code duplication: Each model must explicitly specify routing parameters
3. Maintenance burden: Adding new routing methods requires updates across multiple locations
4. Tight coupling: Current implementation is tied to flashinfer's specific enum values

As noted by @mgoin:
"I like the idea of having a routing method type so we can reduce the need for hacks like checking the llama 4 custom routing function within the 
quant method... I think if we do this right, we can actually remove other arguments we have in FusedMoE such as renormalize."

### Proposed improvements:

1. Auto-derive routing type: Instead of requiring each model to explicitly set routing_method_type, automatically derive it from existing parameters
(`scoring_func`, `renormalize`, `use_grouped_topk`, `top_k`, etc.) within `FusedMoE.__init__`
1. Decouple from flashinfer: Make RoutingMethodType a vLLM-native abstraction that works across all fused MoE backends (not just flashinfer TRTLLM),
with backend-specific mapping happening at the kernel level
1. Simplify FusedMoE API: Remove redundant parameters like `renormalize` and potentially apply_router_weight_on_input by folding them into the routing
type
1. Support explicit override: Allow models to explicitly specify routing type when auto-derivation isn't sufficient
2. Router abstraction: Consider implementing router objects/functions that can be passed directly (as suggested by @bnellnm)

### Alternatives

Keep the current approach of using multiple discrete parameters (`scoring_func`, `renormalize`, etc.), but this requires ongoing maintenance of mapping
logic scattered across quant methods and model code.

### Additional context

Related PR: #27492 - Initial implementation of `RoutingMethodType`

Code locations that would benefit:
- `vllm/model_executor/layers/fused_moe/config.py:RoutingMethodType` - Make backend-agnostic
[...]

PR #30519, opened 2025-12-11
[Misc][Refactor] Add FusedMoERouter object

bnellnm:
## Purpose

Add abstract `FusedMoERouter` class that provides `select_experts`.  There's a concrete subclass that wraps `FusedMoE._select_experts` that is passed to all the quantization methods instead of calling `FusedMoE._select_experts` directly.  This is a step on teasing out the routing logic from `FusedMoE.apply`.

See https://github.com/vllm-project/vllm/issues/28408

## Test Plan
Existing CI

## Test Result

cc @robertgshaw2-redhat 

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/layers/fused_moe/layer.py
+++ b/vllm/model_executor/layers/fused_moe/layer.py
@@ -31,6 +31,7 @@ from vllm.model_executor.layers.fused_moe.config import (
     FusedMoEQuantConfig,
     RoutingMethodType,
 )
+from vllm.model_executor.layers.fused_moe.fused_moe_router import FusedMoERouter
 from vllm.model_executor.layers.fused_moe.rocm_aiter_fused_moe import (
     init_aiter_topK_meta_data,
 )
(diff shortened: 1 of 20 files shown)


vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -293,6 +294,23 @@ def maybe_roundup_hidden_size(
     return hidden_size
 
 
+class FusedMoERouterImpl(FusedMoERouter):
```
robertgshaw2-redhat, line 288: Is the idea that eventually we will have N of these, 1 for each `routing_method_type`?

And then that `FusedMoERouter` would not have the layer as an attr?

bnellnm: Yeah, see https://github.com/vllm-project/vllm/pull/30623

bnellnm: And also https://github.com/vllm-project/vllm/pull/30573 (which comes before #30623)

robertgshaw2-redhat (approved):

Issue #29580, opened 2025-11-27
[Feature]: [CPU Backend] Grouped GEMM kernel for CPU backend


bigPYJ1151:
### 🚀 The feature, motivation and pitch

Implement Grouped GEMM kernel to optimize MOE for CPU backend

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

ayushsatyam146:
Hi @bigPYJ1151 can I work on this issue? If possible can you please assign me?

fadara01:
Hi @bigPYJ1151 - Thanks for taking the initiative on this.

I see that you copied the GEMM kernels we use for CPU Attention to https://github.com/vllm-project/vllm/tree/main/csrc/cpu/micro_gemm

Is the plan to rely on these for GroupedGEMM or are you planning to use oneDNN's brgemm (I remember you mentioning that before)?
It's not a good idea to have the same GEMM kernels existing in 2 different places in vLLM, is there a plan to unify them and share them between Attention and GroupedGEMM?

Issue #30487, opened 2025-12-11
[Bug] [CPU Backend]: Wrong L2 cache size on Arm CPU for Attention tiles


Radu2k:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
Your output of `python collect_env.py` here
```

</details>

### 🐛 Describe the bug

On Arm-based instances (e.g. AWS c8g), the CPU attention implementation in vLLM uses:

`long l2_size = sysconf(_SC_LEVEL2_CACHE_SIZE);`

inside csrc/cpu/cpu_attn_impl.hpp to estimate the L2 cache size and derive tiling parameters for the attention kernel.

sysconf(_SC_LEVEL2_CACHE_SIZE) returns 0, which effectively means:
	•	The L2 cache size is treated as 0.
	•	The tiling logic falls back to minimal tile sizes or tile size changes have no effect on performance.
	•	The cache aware code is performing poorly due to missing cache size information.

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

fadara01:
#30531 now uses PyTorch utility to query for L2 cache size, let's see if that fixes this issue.

PR #28681, opened 2025-11-13
[CPU][Bugfix] Fix Apple Silicon M1 compilation failure

mgoin:
## Purpose

The CPU attention backend refactor introduced unconditional use of BF16Vec16, which is only available on ARM platforms with native BF16 support (ARMv8.6-A extension). M1 chips and other older ARM processors lack this hardware capability.

Additionally, macOS doesn't support `_SC_LEVEL2_CACHE_SIZE` sysconf, causing L2 cache detection to fail.

Errors found:
- `error: no type named 'BF16Vec16' in namespace 'vec_op'` in `cpu_attn_impl.hpp:821`
- `error: use of undeclared identifier '_SC_LEVEL2_CACHE_SIZE'` in `cpu_attn_impl.hpp:744`

cc @bigPYJ1151 @fadara01

## Test Plan

Build on M1 Mac: `uv pip install -e .`

## Test Result

Build succeeds, vLLM imports successfully on M1.

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #30531, opened 2025-12-12
[CPU] Refactor CPU fused MOE

bigPYJ1151:
## Purpose

Refactor CPU fused MOE by optimizing tile schedule and enable torch compile

Part of #29580

main:
```log
Successful requests:                     16        
Maximum request concurrency:             8         
Benchmark duration (s):                  274.87    
Total input tokens:                      16384     
Total generated tokens:                  16384     
Request throughput (req/s):              0.06      
Output token throughput (tok/s):         59.61     
Peak output token throughput (tok/s):    88.00     
Peak concurrent requests:                15.00     
Total token throughput (tok/s):          119.21    
Mean TTFT (ms):                          3477.60   
Median TTFT (ms):                        3591.83   
P99 TTFT (ms):                           4792.09   
Mean TPOT (ms):                          130.91    
Median TPOT (ms):                        130.83    
P99 TPOT (ms):                           133.00    
Mean ITL (ms):                           130.91    
Median ITL (ms):                         130.06    
P99 ITL (ms):                            145.10    
```

this:
```log
Successful requests:                     16        
Maximum request concurrency:             8         
Benchmark duration (s):                  156.29    
Total input tokens:                      16384     
Total generated tokens:                  16384     
Request throughput (req/s):              0.10      
Output token throughput (tok/s):         104.83    
Peak output token throughput (tok/s):    144.00    
Peak concurrent requests:                16.00     
Total token throughput (tok/s):          209.67    
Mean TTFT (ms):                          1381.60   
Median TTFT (ms):                        1432.94   
P99 TTFT (ms):                           1965.64   
Mean TPOT (ms):                          75.00     
Median TPOT (ms):                        75.35     
P99 TPOT (ms):                           76.97     
Mean ITL (ms):                           75.00     
Median ITL (ms):                         74.75     
P99 ITL (ms):                            85.00     
```

## Test Plan

CI tests

## Test Result

---
<details>

</details>


csrc/cpu/cpu_arch_macros.h
```diff
@@ -26,7 +26,7 @@
           _mm512_castsi512_ps(_mm512_set1_epi32(0x42b17218));                  \
       const __m512i vec_127 = _mm512_set1_epi32(0x0000007f);                   \
       const int n_mantissa_bits = 23;                                          \
-      auto fast_exp = [&](vec_op::FP32Vec16& vec) __attribute__((              \
+      auto fast_exp = [&](const vec_op::FP32Vec16& vec) __attribute__((        \
```
fadara01, line 29: would you be open to making the same change for AArch64 part too for consistency?

.buildkite/release-pipeline.yaml
```diff
@@ -71,6 +71,20 @@ steps:
     env:
       DOCKER_BUILDKIT: "1"
 
+  # x86 CPU wheel build
+  - label: "Build x86 CPU wheel"
```
fadara01, line 75: This is not related to CPU MoE, let's leave it for another PR to make history tractable.
E.g. if we ever have to revert this PR, the "Build x86 CPU wheels" will get reverted too.

.buildkite/scripts/hardware_ci/run-cpu-test.sh
```diff
@@ -50,6 +50,7 @@ function cpu_tests() {
   docker exec cpu-test-"$NUMA_NODE" bash -c "
     set -e
     pytest -x -v -s tests/kernels/attention/test_cpu_attn.py
+    pytest -x -v -s tests/kernels/moe/test_cpu_fused_moe.py
```
fadara01, line 53: Given you enabled this through ISA::Vec which is CPU agnostic, would it be a good idea to enable this for all run-cpu-tests, i.e:
- .buildkite/scripts/hardware_ci/run-cpu-test-arm.sh
- .buildkite/scripts/hardware_ci/run-cpu-test-ppc64le.sh
- .buildkite/scripts/hardware_ci/run-cpu-test-s390x.sh

csrc/cpu/cpu_fused_moe.cpp
```diff
@@ -0,0 +1,726 @@
+#include "cpu/cpu_types.hpp"
+#include "cpu/utils.hpp"
+#include "cpu/scratchpad_manager.h"
+#include "cpu/micro_gemm/cpu_micro_gemm_vec.hpp"
+#include "cpu/cpu_attn_macros.h"
```
fadara01: It'd be a good idea to change the name of this `cpu/cpu_attn_macros.h` file.
Maybe something like `cpu_exp_impl` or something, because it doesn't make sense to include `cpu_attn_macros.h` in a FusedMoE kernel.

requirements/cpu-build.txt
```diff
@@ -1,7 +1,7 @@
 cmake>=3.26.1
 ninja
 packaging>=24.2
-setuptools>=77.0.3,<81.0.0
+setuptools==77.0.3 # this version can reuse CMake build dir
```
aditew01, line 4: The changes are unrelated to the description?

csrc/cpu/micro_gemm/cpu_micro_gemm_vec.hpp
```diff
@@ -109,6 +109,23 @@ class MicroGemm<cpu_utils::ISA::VEC, scalar_t> {
   void gemm(DEFINE_CPU_MICRO_GEMM_PARAMS) {
     TileGemm82<scalar_t>::gemm(CPU_MICRO_GEMM_PARAMS);
   }
+
```
fadara01, line 112: These kernels are exactly the same as those in vllm/csrc/cpu/cpu_attn_vec.hpp
From a maintainability point of view, It's doesn't seem like a good idea to have the same kernel copied in 2 different places, can we unify them?

csrc/cpu/torch_bindings.cpp
```diff
@@ -296,6 +307,19 @@ TORCH_LIBRARY_EXPAND(TORCH_EXTENSION_NAME, ops) {
       "pack_factor, str isa_hint) -> ()");
   ops.impl("cpu_gemm_wna16", torch::kCPU, &cpu_gemm_wna16);
 #endif
+
+  // fused moe
+#if defined(__AVX512F__)
```
fadara01, line 312: If `isa::vec` is the intended cross-arch abstraction (which is currently the case in cpu attention), it’d be nice if this wasn’t hard-wired to AVX512. If we only want to expose this on x86 for now, could we push that policy into Python based on `CpuArchEnum` instead?

fadara01 (changes requested):
Great work! Thank you :)

I added some initial comments, sorry for the NITs

csrc/cpu/micro_gemm/cpu_micro_gemm_vec.hpp
```diff
@@ -109,6 +109,23 @@ class MicroGemm<cpu_utils::ISA::VEC, scalar_t> {
...
+  static void pack_weight(const scalar_t* __restrict__ weight,
```
fadara01, line 115: can you say in the comments the format we're packing the weights to

vllm/model_executor/layers/fused_moe/cpu_fused_moe.py
```diff
@@ -174,8 +184,105 @@ def __call__(
...
+            topk_ids,
+            activation,
+            global_num_experts,
+        )
+
+    def check_grouped_gemm(
+        self,
+        layer: torch.nn.Module,
+    ) -> tuple[bool, str]:
+        if not hasattr(torch.ops._C, "prepack_moe_weight"):
```
fadara01, line 245: I think it's better to enable/disable CPU grouped gemm based on `CpuArchEnum`, rather than the existence of `prepack_moe_weight` which is currently determined by an AVX512 ifdef.

IMO, `prepack_moe_weight` should exist for all architectures given it relies on `isa::vec`.

I'm under the impression that `isa::vec` is meant to be Architecture agnostic (since this is the case in attention)

vllm/model_executor/layers/fused_moe/cpu_fused_moe.py
```diff
@@ -174,8 +184,105 @@ def __call__(
...
+
+        if not (w13_output_size % 32 == 0 and w2_output_size % 32 == 0):
+            return False, "none"
+
+        supports_amx = torch._C._cpu._is_amx_tile_supported()
+
+        if (
+            supports_amx
+            and dtype == torch.bfloat16
+            and w13_input_size % 32 == 0
```
fadara01, line 262: `w13` and `w2` sizes are guaranteed to be a multiple of 32 by:
```
        if not (w13_output_size % 32 == 0 and w2_output_size % 32 == 0):
            return False, "none"
```

let's remove these redundant checks.

csrc/cpu/cpu_types_x86.hpp
```diff
@@ -352,6 +352,10 @@ struct FP32Vec16 : public Vec<FP32Vec16> {
   explicit FP32Vec16(bool, void* ptr)
       : reg((__m512)_mm512_stream_load_si512(ptr)) {}
 
+  // strided load
+  explicit FP32Vec16(const float* ptr, INT32Vec16 idx)
```
fadara01, line 356: in my opinion, we should aim to keep the vectorizer APIs consistent across CPUs supported in vLLM - similar to what we do in PyTorch vectorizer classes. 

I don't think vLLM vectorizers are currently consistent for different CPUs, but we should aim not to make them diverge further.

I.e. if we are to introduce a new API to any vectorizer, IMO we should make sure that it's supported by the other platforms (even if that support is through a slow/scalar impl), otherwise we'll end with a lot of ifdefs and unmaintainable code in implementations using vLLM's vec abstractions.

For this particular case, one should be able to use `explicit FP32Vec16(const float* ptr, INT32Vec16 idx)` on any CPU vLLM supports. I don't think this is the case, but please correct me if I'm wrong.

Please let me know if there's a good reason as to why this (generally speaking) is not currently the case and/or shouldn't be the case in the future.

csrc/cpu/cpu_fused_moe.cpp
```diff
@@ -0,0 +1,726 @@
...
+  // Weighted sum thread buffer
+  const int32_t ws_output_buffer_size =
+      cpu_utils::round_up<64>(output_size_2 * sizeof(float));
+  const int32_t ws_output_buffer_offset = ws_thread_buffer_offset;
+  ws_thread_buffer_offset += ws_output_buffer_size;
+
+  const int32_t buffer_size =
+      common_buffer_offset +
+      std::max(w13_thread_buffer_offset, ws_thread_buffer_offset) * thread_num;
+  DNNLScratchPadManager::get_dnnl_scratchpad_manager()->realloc(buffer_size);
```
fadara01: this implementation has nothing to do with oneDNN… Maybe we should rename this to CPUScratchPadManager?
we have the same smell in CPU attention which confused me when I was reviewing it :smile:

cmake/cpu_extension.cmake
```diff
@@ -365,6 +365,7 @@ if (AVX512_FOUND AND NOT AVX512_DISABLED)
     set(VLLM_EXT_SRC
         "csrc/cpu/shm.cpp"
         "csrc/cpu/cpu_wna16.cpp"
+        "csrc/cpu/cpu_fused_moe.cpp"
```
fadara01, line 367: See my comments above about `isa::vec` being platform agnostic.

tests/kernels/moe/test_cpu_fused_moe.py
```diff
@@ -0,0 +1,172 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
```
fadara01, line 2: let's try to unify this with CPU MoE test in https://github.com/vllm-project/vllm/blob/main/tests/kernels/moe/test_moe.py#L1062

fadara01 (changes requested):

csrc/cpu/utils.hpp
```diff
@@ -15,6 +15,16 @@
 namespace cpu_utils {
 enum class ISA { AMX, VEC };
 
+inline ISA get_isa(const std::string& isa) {
```
fadara01, line 13: We have code that does the same thing in attention:
- https://github.com/vllm-project/vllm/blob/main/csrc/cpu/cpu_attn.cpp#L82
- https://github.com/vllm-project/vllm/blob/main/csrc/cpu/cpu_attn_impl.hpp#L17

Given that you're (re)introducing this `get_isa` function and `ISA` enums at this high level in `cpu/utils.hpp`, can we try to unify it with what we already have in attention.

csrc/cpu/cpu_fused_moe.cpp
```diff
@@ -0,0 +1,726 @@
...
+  }
+}
+
+template <typename scalar_t>
+void swigluoai_and_mul(float* __restrict__ input, scalar_t* __restrict__ output,
+                       const int32_t m_size, const int32_t n_size,
+                       const int32_t input_stride,
+                       const int32_t output_stride) {
+  using scalar_vec_t = typename cpu_utils::VecTypeTrait<scalar_t>::vec_t;
+  alignas(64) static int32_t index[16] = {0,  2,  4,  6,  8,  10, 12, 14,
```
fadara01, line 52: let's add a comment saying why this is needed - i.e. because gpt-oss interleaves gate-up weights.
I expressed worries before about hard-coding memory layout information in activation func implementations (see: https://github.com/vllm-project/vllm/pull/27024#issuecomment-3567879307) but I guess we're stuck with this now.

csrc/cpu/utils.hpp
```diff
@@ -68,6 +78,35 @@ inline int64_t get_l2_size() {
   }();
   return size;
 }
+
+template <int32_t alignment_v, typename T>
```
fadara01, line 64: Can we not reuse PyTorch ATen utilities? - https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/ceil_div.h

csrc/cpu/cpu_fused_moe.cpp
```diff
@@ -0,0 +1,726 @@
...
+  TORCH_CHECK_EQ(output_size_2 % gemm_n_tile_size, 0);
+  TORCH_CHECK_EQ(output_size_13 / 2, input_size_2);
+
+  const int32_t thread_num = omp_get_max_threads();
+
+  const int32_t w13_input_buffer_size = cpu_utils::round_up<64>(
+      gemm_m_tile_size * input_size_13 * sizeof(scalar_t));
+
+  const int32_t w13_n_tile_size = [&]() {
+    const int64_t cache_size = cpu_utils::get_l2_size();
```
fadara01: This `cpu_utils::get_l2_size()` function turned out to be problematic.
See https://github.com/vllm-project/vllm/pull/28681/files where it failed on Apple Silicon and https://github.com/vllm-project/vllm/issues/30487 where it returns zero on Arm CPUs.

I wonder if we could rely on something more portable (and battle-tested) like cpuinfo from pytorch?

fadara01 (commented):
btw, should we trim the SGLang CPU MoE kernel path? Is there any reason as to why it needs to co-exist with the new fused MoE impl?  https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/fused_moe/unquantized_fused_moe_method.py#L247

csrc/cpu/cpu_fused_moe.cpp
bigPYJ1151: good idea :)

requirements/cpu-build.txt
bigPYJ1151: It is a small change. I'd like to add it at one😂

.buildkite/release-pipeline.yaml
bigPYJ1151: It is an accidental addition.

.buildkite/scripts/hardware_ci/run-cpu-test.sh
bigPYJ1151: Not exactly. This kerenl is only avaliable on AVX512 because of some new vec_op. Other platforms need a bit further change and verification.

csrc/cpu/micro_gemm/cpu_micro_gemm_vec.hpp
bigPYJ1151: Yes, they are almost same but with some difference, will unify them in another PR.

csrc/cpu/micro_gemm/cpu_micro_gemm_vec.hpp
bigPYJ1151: Updated. For this weight ```[oc, ic]``` should be packed as ```[oc // 16, ic, 16]```. In general weight should be packed as ```oc // 16``` blocks with logical shape ```[16, ic]```. The layout of blocks is ISA-specific.

vllm/model_executor/layers/fused_moe/cpu_fused_moe.py
bigPYJ1151: Here checks input_size rather than output_size.

csrc/cpu/cpu_types_x86.hpp
bigPYJ1151: For these fundmental operations I think throwing error explicitly is more helpful for finding missed features.

tests/kernels/moe/test_cpu_fused_moe.py
bigPYJ1151: The numeric difference from dtype conversion is too large in ```torch_experts```.

csrc/cpu/utils.hpp
bigPYJ1151: I did't see ```round_down``` in it 🤔

csrc/cpu/cpu_types_x86.hpp
fadara01: I guess this depends on what we want the vectorizer classes in vLLM to be.
In PyTorch, the vectorizer class is an architecture agnostic interface for a bunch of ops with architecture specific implementations (e.g. Arm uses SVE, x86 uses AVX, etc). The PyTorch vectorizer does not allow you to define/expose a new op without providing a reference impl (usually scalar impl) avaliable for all vectrorizer implementations.

In my opinion the PyTorch vectorizer approach is a lot more elegant than what we currently have in vLLM.

We can ignore this comment for the purposes of this PR, but should discuss/agree later on what we want vLLM vectorizers to be :)

tests/kernels/moe/test_cpu_fused_moe.py
fadara01: sorry, I don't fully get you. I'm just proposing to have both CPU MoE tests in the same file.

csrc/cpu/utils.hpp
fadara01: ohh I assumed pytorch would have a corresponding `floor_div.h` but they don't seem to have that.

csrc/cpu/micro_gemm/cpu_micro_gemm_vec.hpp
fadara01: ACK, looking forward to that PR.

csrc/cpu/micro_gemm/cpu_micro_gemm_vec.hpp
fadara01: Thanks!

vllm/model_executor/layers/fused_moe/cpu_fused_moe.py
fadara01: ACK

tests/kernels/moe/test_cpu_fused_moe.py
bigPYJ1151: Oh, I thought you suggest to use ```torch_experts``` as reference impl likes test cases in ```test_moe.py```.

I think it is not needed to put them togther as there is nothing to reuse.

csrc/cpu/cpu_types_x86.hpp
bigPYJ1151: Yes, in fact the vec_op needs to refactor and clean up as many operations are no longer be used.

tests/kernels/moe/test_cpu_fused_moe.py
fadara01: I'm just trying to say that it's a good idea to have 1 test file to test all CPU FusedMoE impls

jikunshang (approved):

DarkLight1337:
This PR is causing Apple Silicon test to fail on main: https://github.com/vllm-project/vllm/actions/runs/20328342556/job/58398002846

PR #30535, opened 2025-12-12
[Core] Add repetitive token detection for hallucination prevention

jeremyteboul:
This PR adds a new feature to detect and stop generation when repetitive token patterns are detected, which is a common indicator of model hallucination.

## Changes

1. Added  parameter to :
   - Set to 0 (default) to disable
   - Set to N to stop generation when N consecutive identical tokens are detected

2. Added  status and :
   - New request status for when repetition detection triggers
   - New finish reason exposed in API responses

3. Implemented O(1) fast-path repetition detection in :
   - Only performs full check when last two tokens are identical
   - Avoids performance impact in common case (no repetition)

4. Added comprehensive unit tests covering:
   - Edge cases (empty list, disabled detection, thresholds)
   - Integration with check_stop()
   - Performance characteristics

## Usage

```python
from vllm import SamplingParams

# Enable repetition detection with threshold of 4
params = SamplingParams(
    max_consecutive_repeats=4,  # Stop if 4+ identical tokens in a row
    max_tokens=1000,
)
```

## Motivation

Some multimodal models (e.g., Qwen3 VL) can get stuck outputting repetitive tokens, which wastes compute and produces unusable output. This feature allows early termination of such requests.


jeremyteboul:
> Hi @jeremyteboul, the pre-commit checks have failed. Please run:
> 
> ```shell
> uv pip install pre-commit
> pre-commit install
> pre-commit run --all-files
> ```
> 
> Then, commit the changes and push to your branch.
> 
> For future commits, `pre-commit` will run automatically on changed files before each commit.
> 
> Tip
> 
> Is `mypy` or `markdownlint` failing?

resolved

PR #30537, opened 2025-12-12
Filter safetensors files to download if .safetensors.index.json exists

mgoin:
## Purpose

I noticed when running `vllm serve openai/gpt-oss-20b` on a new machine that it not only downloaded the base model files, but also included the `original/` directory that I assumed was picked up because it has files ending in `*.safetensors`

```
model-00000-of-00002.safetensors: 100%|██████████████████████████████████████████████████████████████████████████| 4.79G/4.79G [06:24<00:00, 12.5MB/s]
model-00001-of-00002.safetensors: 100%|██████████████████████████████████████████████████████████████████████████| 4.80G/4.80G [06:10<00:00, 13.0MB/s]
model-00002-of-00002.safetensors: 100%|██████████████████████████████████████████████████████████████████████████| 4.17G/4.17G [05:03<00:00, 13.7MB/s]
original/model.safetensors:  76%|█████████████████████████████████████████████████████████████                   | 10.5G/13.8G [14:48<04:34, 11.8MB/s]
```

Since the model has a [`model.safetensors.index.json`](https://huggingface.co/openai/gpt-oss-20b/blob/main/model.safetensors.index.json), I think we should use this to filter the glob if we can find one

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/model_loader/weight_utils.py
+++ b/vllm/model_executor/model_loader/weight_utils.py
@@ -462,7 +462,12 @@ def download_weights_from_hf(
                 )
                 with open(index_path) as f:
                     weight_map = json.load(f)["weight_map"]
-                allow_patterns = list(set(weight_map.values()))
+                if weight_map:
+                    # Extra [] so that weight_map files are treated as a
+                    # single allow_pattern in the loop below
+                    allow_patterns = [list(set(weight_map.values()))]  # type: ignore[list-item]
+                else:
+                    allow_patterns = ["*.safetensors"]
             else:
                 # Use the first pattern found in the HF repo's files.
                 for pattern in allow_patterns:
@@ -471,7 +476,9 @@ def download_weights_from_hf(
                         break
         except Exception as e:
             logger.warning(
-                "Failed to get file list for '%s': %s",
+                "Failed to get file list for '%s'. Trying each pattern in "
+                "allow_patterns individually until weights have been "
+                "downloaded. Error: %s",
                 model_name_or_path,
                 e,
             )
@@ -481,15 +488,23 @@ def download_weights_from_hf(
     # downloading the same model weights at the same time.
     with get_lock(model_name_or_path, cache_dir):
         start_time = time.perf_counter()
-        hf_folder = snapshot_download(
-            model_name_or_path,
-            allow_patterns=allow_patterns,
-            ignore_patterns=ignore_patterns,
-            cache_dir=cache_dir,
-            tqdm_class=DisabledTqdm,
-            revision=revision,
-            local_files_only=local_only,
-        )
+        for allow_pattern in allow_patterns:
+            hf_folder = snapshot_download(
+                model_name_or_path,
+                allow_patterns=allow_pattern,
+                ignore_patterns=ignore_patterns,
+                cache_dir=cache_dir,
+                tqdm_class=DisabledTqdm,
+                revision=revision,
+                local_files_only=local_only,
+            )
+            # If we have downloaded weights for this allow_pattern,
+            # we don't need to check the rest.
+            # allow_pattern can be a list (from weight_map) or str (glob)
+            if isinstance(allow_pattern, list):
+                break
+            if any(Path(hf_folder).glob(allow_pattern)):
+                break
         time_taken = time.perf_counter() - start_time
         if time_taken > 0.5:
             logger.info(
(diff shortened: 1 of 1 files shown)


vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -468,20 +481,15 @@ def download_weights_from_hf(
...
-                revision=revision,
-                local_files_only=local_only,
-            )
-            # If we have downloaded weights for this allow_pattern,
-            # we don't need to check the rest.
-            if any(Path(hf_folder).glob(allow_pattern)):
-                break
+        hf_folder = snapshot_download(
+            model_name_or_path,
+            allow_patterns=allow_patterns,
```
hmellor: If you pass all the `allow_patterns` at once, then it will download all matching files.

The loop that this replaces was deliberate. It handles the case where the file list fetching fails and the checkpoint contains files which match multiple patterns (i.e. duplicated weights in multiple formats). With the change in this PR, the fallback would result in all formats being downloaded.

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -448,17 +449,29 @@ def download_weights_from_hf(
...
+                for pattern in allow_patterns:
+                    if fnmatch.filter(file_list, pattern):
+                        allow_patterns = [pattern]
+                        break
         except Exception as e:
             logger.warning(
-                "Failed to get file list for '%s'. Trying each pattern in "
-                "allow_patterns individually until weights have been "
-                "downloaded. Error: %s",
+                "Failed to get file list for '%s': %s",
```
hmellor: For the reason in my first comment, this should be reverted

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -468,20 +481,15 @@ def download_weights_from_hf(
...
-                break
+        hf_folder = snapshot_download(
+            model_name_or_path,
+            allow_patterns=allow_patterns,
+            ignore_patterns=ignore_patterns,
+            cache_dir=cache_dir,
+            tqdm_class=DisabledTqdm,
+            revision=revision,
+            local_files_only=local_only,
+        )
```
hmellor: For the reason in my first comment, this should be reverted

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -448,17 +448,29 @@ def download_weights_from_hf(
...
+            if "*.safetensors" in allow_patterns and index_file in file_list:
+                index_path = hf_hub_download(
+                    repo_id=model_name_or_path,
+                    filename="model.safetensors.index.json",
+                    cache_dir=cache_dir,
+                    revision=revision,
+                )
+                with open(index_path) as f:
+                    weight_map = json.load(f)["weight_map"]
+                allow_patterns = list(set(weight_map.values()))
```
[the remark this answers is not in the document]
hmellor: This seems like a valid concern that we could be more defensive about. 

i.e.

```
if weight_map:
    allow_patterns = [list(set(weight_map.values()))]
else:
    allow_patterns = ["*.safetensors"]
```

(The extra `[]` is explained in another comment)

vllm/model_executor/model_loader/weight_utils.py
```diff
@@ -448,17 +449,29 @@ def download_weights_from_hf(
...
+            if "*.safetensors" in allow_patterns and index_file in file_list:
+                index_path = hf_hub_download(
+                    repo_id=model_name_or_path,
+                    filename=SAFE_WEIGHTS_INDEX_NAME,
+                    cache_dir=cache_dir,
+                    revision=revision,
+                )
+                with open(index_path) as f:
+                    weight_map = json.load(f)["weight_map"]
+                allow_patterns = list(set(weight_map.values()))
```
hmellor: This change should allow the loop below to be reverted, fixing the fallback case when the file list fails to be read.

```suggestion
                # Extra [] so that weight_map all lives in a single allow_pattern in the loop below
                allow_patterns = [list(set(weight_map.values()))]
```

hmellor (changes requested):

mgoin:
@hmellor thanks a lot for clearing up my mistaken assumptions, PTAL at the new state!

hmellor (approved):
LGTM thanks for the fix!

PR #30544, opened 2025-12-12
[KVEvent] User request.block_hash for parent block_hash

heheda12345:
## Purpose

Parent block can be a null block:

* mamba: the block_table will be like [null_block, null_block, ..., null_block, normal_block] as we only need one block per decode step.      
* sliding window + kv cache connector, assume block_size 1, window size 2, hit 3 local tokens + 3 external tokens, the block table will be [NULL, NULL, NULL, NULL, 4, 5]
we will do:
```
# first allocation
allocate_slots(delay_cache_blocks=True):
    save_new_computed_blocks() caches the first 3 blocks
# after kv cache transfer
allocate_slots(delay_cache_blocks=False):
    cache_blocks() caches the first 6 blocks, parent block is block 2 (null_block)
```

So we extract parent block hash from request.block_hashes instead of null_block

## Test Result

---
<details>

</details>

### DIFF
--- a/tests/v1/core/test_prefix_caching.py
+++ b/tests/v1/core/test_prefix_caching.py
@@ -1356,6 +1356,69 @@ def test_kv_cache_events(blocks_to_cache: int):
     assert len(manager.block_pool.cached_block_hash_to_block) == 0
 
 
+def test_null_parent_block_hash():
+    block_size = 1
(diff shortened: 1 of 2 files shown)


mgoin (approved):
Seems reasonable to me. Should we add a unit test if this failed in some case?

PR #29873, opened 2025-12-02
[CustomOp] Extract ApplyRotaryEmb as CustomOp and unify the dispatch logic

shen-shanshan:
## Purpose

1. In some modeling files, there are direct calling of `apply_rotary_emb` function by using pre-computed cos/sin cache, like: https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/qwen2_5_vl.py#L383-L385. This is just a function, not an operation, and cannot be overwritten by some plugins (e.g., vllm-ascend). By extracting it as an CustomOp, we can just extend this class and implement our `forward_oot()` function, after which we register it to easily replace this op with other op of OOT device.
2. There are several dispatch function for `apply_rotary_emb` which may make users or developers confused, such as https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/rotary_embedding/common.py#L56-L70 and https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/rotary_embedding/common.py#L73-L97. This PR has unified these separate dispatch logic into one CustomOp to make it clearer.
3. There are also some redundant definitions of some functions, such as `rotate_half()` and `apply_rotary_emb_torch()`. This PR has removed these replicated definitions in different modeling files and just remained one replica in https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/rotary_embedding/common.py.

---
**Recent updates:**

1. Move `find_spec("flash_attn")` to `__init__()` of `ApplyRotaryEmb`.
2. Add `enable_fp32_compute` param to `ApplyRotaryEmb` to determine whether convert input to float and recover it after computation inside the OP.
3. Correct forward dispatch of `ApplyRotaryEmb` in `XxxRope`. Make sure `forward_native()` -> `forward_native()`, `forward_cuda() / forward_hip()` -> `forward()`.
4. Support object-level enable for `ApplyRotaryEmb` in ViT part to avoid extra configs when lauching the server. Find more details at https://github.com/vllm-project/vllm/pull/30547.

## Test Plan

- [x] Functional test on Ascend A2 NPU.
- [x] Functional test on NVIDIA A100 GPU.
- [x] Benchmark on NVIDIA A100 GPU.
- [x] Add UT to make sure the dispatching logic is correct.

### ✅ Functional test on Ascend A2 NPU

I have tested this PR on Ascend A2 NPU together with https://github.com/vllm-project/vllm-ascend/pull/4667.

Run:

```bash
[...]

PR #30547, opened 2025-12-12
[CustomOp] Support object-level enable for CustomOp

shen-shanshan:
## Purpose

Add `enforce_enable` param to CustomOp to control whether enable a CustomOp at object-level. This is useful to enable some device-specific kernels in ViT models when enabling graph mode. By default, it will still follow the `compilation_config` to determine whether enable itself.

This is can resolve the issue in https://github.com/vllm-project/vllm/pull/29873#issuecomment-3636102452.

Like:

```python
class XxxVisionAttention:

    def __init__(...):
        apply_rotary_emb = ApplyRotaryEmb(enforce_enable=True)
```

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/custom_op.py
+++ b/vllm/model_executor/custom_op.py
@@ -40,8 +40,8 @@ class CustomOp(nn.Module):
 
     def __init__(self, enforce_enable: bool = False):
         super().__init__()
-        self._forward_method = self.dispatch_forward()
         self._enforce_enable = enforce_enable
+        self._forward_method = self.dispatch_forward()
 
     def forward(self, *args, **kwargs):
         return self._forward_method(*args, **kwargs)
--- a/vllm/model_executor/custom_op.py
+++ b/vllm/model_executor/custom_op.py
@@ -38,9 +38,10 @@ class CustomOp(nn.Module):
             )
         return super().__new__(op_cls_to_instantiate)
 
-    def __init__(self):
+    def __init__(self, enforce_enable: bool = False):
         super().__init__()
         self._forward_method = self.dispatch_forward()
+        self._enforce_enable = enforce_enable
 
     def forward(self, *args, **kwargs):
         return self._forward_method(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


vllm/model_executor/custom_op.py
```diff
@@ -84,7 +85,11 @@ def dispatch_forward(self):
         # NOTE(woosuk): Here we assume that vLLM was built for only one
         # specific backend. Currently, we do not support dynamic dispatching.
         compilation_config = get_cached_compilation_config()
-        enabled = self.enabled()
+
+        # CustomOp object can be enforce enabled, e.g., enable device-specific
+        # kernels in ViT models when enabling graph mode. By default, it will
+        # follow the compilation_config to determine whether enable itself.
+        enabled = self._enforce_enable or self.enabled()
```
Isotr0py, line 92: I think this will also force enable custom_ops on other layers, because `compilation_config.enabled_custom_ops` will be updated in following lines:
```python3
        if enabled:
            compilation_config.enabled_custom_ops.update([self.__class__.name])
        else:
            compilation_config.disabled_custom_ops.update([self.__class__.name])
```

IMO, `_enforce_enable` should not affect global compilation config:
```python3
        enabled = self.enabled()
        if enabled:
            compilation_config.enabled_custom_ops.update([self.__class__.name])
        else:
            compilation_config.disabled_custom_ops.update([self.__class__.name])

        if not enabled and not self._enforce_enable:
            return self.forward_native
        ...
```

Isotr0py (commented):
Also cc @tjtanaa and @ProExpertProg

vllm/model_executor/custom_op.py
ProExpertProg: I disagree here, the global config should reflect that this op is force enabled.

Isotr0py: NVM, I mistake `enabled_custom_ops` and `custom_ops`, so this change LGTM.

ProExpertProg (commented):
Why is this necessary? Are these ops that don't have a `forward_native` implementation? Or we just want to enable the custom impl by default.

If it's the former, this makes sense. If it's the latter, we should make this a `default_on` parameter and allow overriding from the compilation config

tjtanaa (approved):
LGTM

vllm/model_executor/custom_op.py
gcanlin: I’m still a bit confused about this. With this change, even if the user explicitly specifies disabled custom ops like `--compilation_config.custom_ops '["-mm_encoder_attn"]'`, once it is enabled in the code via `_enforce_enable`, the argument in user-face would be ignored. Would this lead to a mismatch between user expectations and actual behavior?

shen-shanshan:
> @shen-shanshan honestly I think we should create a whole separate instance of mm_compilation_config instead. We can default it to no compilation for now but add separate logic in the future. Cc @Lucaskabela also I remember you were working on compiling the MM part of models

@ProExpertProg I think it's a good idea to add a separate `mm_compilation_config` for ViT part. Maybe we could do this later? Currently, we plan to make sure https://github.com/vllm-project/vllm/pull/29873 and https://github.com/vllm-project/vllm/pull/30125 could be merged before the last release this year.

PR #30551, opened 2025-12-12
docs: Clarify block_quant_to_tensor_quant docstring (fixes #30098)

yurekami:
## Summary
- Clarifies the misleading docstring for `block_quant_to_tensor_quant` function in `fp8_utils.py`
- The function converts block-scale FP8 quantized data to tensor-scale format, but does NOT perform full dequantization
- Added warning to docstring to prevent confusion about the function's output

Fixes #30098

## Changes
- Updated docstring to clarify that the function "converts" rather than "dequantizes"
- Added explicit warning that output is still quantized (not original precision values)
- Improved parameter and return value documentation

## Test plan
- [x] Docstring-only change, no functional changes
- [x] Verified the fix addresses the confusion reported in issue #30098

### DIFF
--- a/vllm/model_executor/layers/quantization/utils/fp8_utils.py
+++ b/vllm/model_executor/layers/quantization/utils/fp8_utils.py
@@ -18,6 +18,7 @@ from vllm.model_executor.layers.quantization.input_quant_fp8 import QuantFP8
 from vllm.model_executor.layers.quantization.utils.quant_utils import (
     GroupShape,
     group_broadcast,
+    scaled_dequantize,
 )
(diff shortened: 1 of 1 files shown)


vllm/model_executor/layers/quantization/utils/fp8_utils.py
```diff
@@ -466,15 +467,50 @@ def input_to_float8(
 def block_quant_to_tensor_quant(
```
pavanimajety, line 467: Could we add a unit test for this function? It seems like this function doesn't have any call sites right now.

mgoin:
This function is removed by https://github.com/vllm-project/vllm/pull/30551

PR #30553, opened 2025-12-12
[Bug][CPU Backend]: Improve L2 cache size detection and usage on aarch64

Radu2k:
## Purpose
- Add /sys/devices/system/cpu/cpu0/cache/index2/size parsing for L2 cache on aarch64
- Fallback to 1MB with warning if the sysfs file cannot be opened
- Use 70% of detected L2 cache for attention scheduling to leave enough room for sys processes

## Test Plan
`l2_cache_size` gets the right value.

## Test Result
`l2_cache_size` is set to 2048*1024*0.7 bytes on c8g instance type which is the correct value give that it has 2MB of L2 cache per core and we want to use 70%.

Solves: #30487 

cc: @fadara01  @aditew01

### DIFF
--- a/csrc/cpu/cpu_attn_impl.hpp
+++ b/csrc/cpu/cpu_attn_impl.hpp
@@ -3,6 +3,7 @@
 
 #include <type_traits>
 #include <cstddef>
+#include <fstream>
 
 #if defined(__APPLE__)
   #include <sys/sysctl.h>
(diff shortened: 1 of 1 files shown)


csrc/cpu/cpu_attn_impl.hpp
```diff
@@ -761,6 +762,39 @@ class AttentionScheduler {
       }
       // Fallback if sysctlbyname fails
       return 128LL * 1024 >> 1;  // use 50% of 128KB
+#elif defined(__aarch64__)
```
aditew01, line 765: Nice, I like the logic! 
Although, for reusability, can we use something like [pytorch/cpuinfo](https://github.com/pytorch/cpuinfo?tab=readme-ov-file#examples) to get the data?  Unless it's not possible?

aditew01 (changes requested):

csrc/cpu/cpu_attn_impl.hpp
```diff
@@ -761,6 +762,39 @@ class AttentionScheduler {
...
+      std::ifstream l2_cache_file(
+          "/sys/devices/system/cpu/cpu0/cache/index2/size");
+      if (!l2_cache_file.is_open()) {
+        std::printf(
```
fadara01, line 769: Let's use `TORCH_WARN`

csrc/cpu/cpu_attn_impl.hpp
```diff
@@ -761,6 +762,39 @@ class AttentionScheduler {
...
+        return static_cast<long>(1024 * 1024);
+      }
+      std::string size_str;
+      std::getline(l2_cache_file, size_str);
+      l2_cache_file.close();
+      // Parse size based on suffix
+      char suffix = size_str.back();
+      size_str.pop_back();
+      long l2_cache_size = std::stol(size_str);
+      switch (suffix) {
```
fadara01, line 782: Could we explain what these mean?

csrc/cpu/cpu_attn_impl.hpp
fadara01: +1 to using pytorch/cpuinfo, great suggestion @aditew01

csrc/cpu/cpu_attn_impl.hpp
Radu2k: Keeping in mind that the usual format is `<value><unit(K/M)>` we get the unit in suffix, pop it form the back, and then convert the string(value) to a long type.

csrc/cpu/cpu_attn_impl.hpp
Radu2k: I agree, good idea.

Rohanjames1997:
Thanks @Radu2k 

I've raised one here https://github.com/pytorch/cpuinfo/pull/372

PR #30557, opened 2025-12-12
[GPT OSS] Fix tool_choice required

southfreebird:


southfreebird:
Hey @bbrowning 
Sorry, I missed your reply. Yeah, it turned out this approach is not reliable, and the model can easily ignore these instructions. 

> If you want to force a tool call, the proper way to do that would be with forcing the model via structured decoding, right?

I agree here and want to find a way to do so. Apparently, I found that the json schema is only applicable for the "final" channel, and I cannot find where this logic is located. It means that some logic tracks the channel and only when it sees the "final" channel it decides to start to apply Json shema. The problem here is that we need to track the channel too, and force the model to "commentary" when tool_choise set to "require". Any chance you have any idea how it's possible to implement it?

eto-gyu:
@southfreebird 
In order to force a particular channel through structured decoding you need to enforce structure over the whole output. In doing so you do not need to track anything about the channel so long as the grammar is sound. You can do this in the StructuredOutputManager's method should_fill_bitmask and return `True` if tool_choice is `"required"`. And of course you would then have to create a grammar for either `.*<|channel|>commentary` or `[grammar definition for valid analysis messages].*<|channel|>commentary` (I would recommend the latter). That means you would need to compose that grammar with the jsonschema for the tool definitions. I have been able to do this before with the xgrammar backend using the grammar concat feature, but given the existence of llguidance you should use that instead as it allows you to specify tokens rather than text so the model can't put the literal text `<|channel|>commentary` to satisfy your grammar. There is also a feature to arbitrarily nest grammars in llguidance. Here is a relevant harmony chat template grammar I have written before to help you get started. It is in xgrammar, so you will have to translate them to llguidance and also translate the text into literal tokens for the special tokens like `<|channel|>`:
```ebnf
root ::= "<|channel|>" "analysis" "<|message|>" ([^<]* "<" [^|])* [^<]* ("<")? "<|end|>" "<|start|>assistant" "<|channel|>" "commentary" "<|message|>"
```
The above does not contain the actual message which would be the json tool call contents and does not include `<|call|>` at the end (there is no `<|end|>` if it was a function call, just `<|call|>`). You should also replace the `([^<]* "<" [^|])* [^<]* ("<")?` section with `[not any special tokens]*`.
You will also have to force something similar to `<|channel|>commentary to=functions.get_weather <|constrain|>json` and make sure whatever comes after `to` is one of the function definitions sent over the wire.

etog-yu:
@southfreebird Funny thing to note about this:
If you do intend to start applying structured decoding for tool_choice="required", there is currently a bug in vLLM causing bitmasks of CFGs and schemas to be applied to the whole of gpt-oss outputs.
https://github.com/vllm-project/vllm/issues/23120#issuecomment-3731726666
The very last comment in this issue describes the bug.
In my testing through the chat completions entrypoint, I am seeing the same here, which actually only occurs during multiturn:
If your messages array contains 1 system message and 1 user message, the schema/bitmask is applied only to the final channel.
If your messages array contains 1 system message and some number of user and assistant turns, the schema is instead applied to the entire message object and harmony parsing fails, resulting in content=null for streamed and non-streamed completions. I get these results from a vllm/vllm-openai:v0.13.0 container launched with gpt-oss-20b and --tool-call-parser openai --reasoning-parser openai_gptoss --enable-auto-tool-choice
If you inspect the tokens being emitted in serving_chat.py by logging output.text at line 718, you'll be able to see the same. And this little test script should reproduce it.

```python
# testing issue from https://github.com/vllm-project/vllm/issues/23120#issuecomment-3731726666
import requests

ENDPOINT = "http://[ENDPOINT]/v1"
API_KEY = "none"

MODEL = "openai/gpt-oss-20b"

schema = {
    "type": "object",
    "properties": {
        "id": {"type": "string"},
        "sender": {"type": "string"},
    },
    "required": ["id", "sender"]
}

full_resp_cfg = """root ::= pre-message-with-header? (end start-assistant message-with-header)* call?

start-assistant ::= "<|start|>assistant"
pre-message-with-header ::= recipient-tag? "<|channel|>" channel-id recipient-tag? data-format? "<|message|>" pre-message-body
message-with-header ::= recipient-tag? "<|channel|>" channel-id recipient-tag? data-format? "<|message|>" message-body


channel-id ::= ("analysis" | "commentary" | "final")
recipient-tag ::= (" to=" [0-9a-zA-Z._-]+)
data-format ::= " " ("<|constrain|>")? "json"

end ::= "<|end|>"
call ::= "<|call|>"

pre-message-body ::= ([^<]* "<" [^|])* [^<]* ("<")?
message-body ::= result

result ::= "{ \\\"id\\\": \\\"2021-B782\\\", \\\"sender\\\": \\\"Maria\\\" }"

"""

def test_schema():
    messages = [
        {"role": "system", "content": "You are a helpful assistant that extracts information and returns it in a JSON format that adheres to the provided schema."},
        {"role": "user", "content": "Given this NLP content, extract the following information and return it in a JSON format that adheres to this schema: " + str(schema) + "\n\nNLP Content:\nHi, I'm emailing in regards to order id 1037-A459. I wanted to check the status of my shipment.\nThanks! - Alex"},
    ]

    payload = {
        "model": MODEL,
        "messages": messages,
        "response_format": {
            "type": "json_schema",
            "json_schema": {
                "name": "extracted_info",
                "strict": True,
                "schema": schema
            }
        }
    }
    headers = {
        "Authorization": f"Bearer {API_KEY}",
        "Content-Type": "application/json"
    }
    response = requests.post(f"{ENDPOINT}/chat/completions", headers=headers, json=payload)
    response.raise_for_status()
    data = response.json()
    print("This should succeed, using JSON with one turn:")
    print("Response JSON:", data["choices"][0]["message"]["content"])

    messages.extend([
        {"role": "assistant", "content": '{"id":"1037-A459","sender":"Alex"}' },
        {"role": "user", "content": "Given this NLP content, extract the following information and return it in a JSON format that adheres to this schema: " + str(schema) + "\n\nNLP Content:\nHello, my name is Maria. I am reaching out about task id 2021-B782 to get an update on its progress."},
    ])
    payload["messages"] = messages
    response = requests.post(f"{ENDPOINT}/chat/completions", headers=headers, json=payload)
    response.raise_for_status()
    data = response.json()
    print("This should fail, using JSON with multiple turns:")
    print("Response JSON:", data["choices"][0]["message"]["content"])

    payload = {
        "model": MODEL,
        "messages": messages,
        "extra_body": {
            "guided_grammar": full_resp_cfg
        }
    }

    response = requests.post(f"{ENDPOINT}/chat/completions", headers=headers, json=payload)
    response.raise_for_status()
    data = response.json()
    print("This should succeed with guided grammar, showing application to the full response regardless of reasoning parser:")
    print("Response JSON:", data["choices"][0]["message"]["content"])


if __name__ == "__main__":
    test_schema()
```

hmellor:
Closing as obsolete. The approach here (dropping the `final` channel from the Harmony system message when `tool_choice="required"`) was acknowledged as unreliable in-thread, and the structured-decoding fix that reviewers pointed to has since landed in #45560 (merged 2026-08-01), refined by #52222. `get_harmony_structural_tag` in `vllm/parser/harmony.py` now only admits the `final` channel when `tool_choice == "auto"`, so `required` is enforced by the grammar. Four of the five files touched here no longer exist on main after the entrypoints refactor.

PR #30559, opened 2025-12-12
[Feat] Enable eplb with default all2all backend

yewentao256:
## Purpose

A follow up PR for https://github.com/vllm-project/vllm/pull/28377

EPLB can't be compatible with deepep or pplx since we are using the wrong physical/logical mapping, but for default all2all it is supported.

## Test

`export MODEL="deepseek-ai/DeepSeek-V2-lite"`

`vllm serve $MODEL -dp 2  --port 9256 --enable-expert-parallel --enable-eplb --eplb-config '{"window_size":100,"step_interval":100}' `

`lm_eval --model local-completions --model_args "base_url=http://127.0.0.1:9256/v1/completions,model=$MODEL,num_concurrent=1024" --tasks gsm8k`

```bash
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.3798|±  |0.0134|
|     |       |strict-match    |     5|exact_match|↑  |0.3738|±  |0.0133|
```


mgoin (approved):

USTCKAY:
Hi, sorry to comment on a closed PR. You mentioned that EPLB can't be compatible with deepep or pplx, does it mean that we can't enable EPLB when running dp+ep with deepep?  Or just cannot enale EPLB when overlapping shared experts? And why the physical/logical mapping is wrong? It would be very thankful to receive your reply!

yewentao256:
> Hi, sorry to comment on a closed PR. You mentioned that EPLB can't be compatible with deepep or pplx, does it mean that we can't enable EPLB when running dp+ep with deepep? Or just cannot enale EPLB when overlapping shared experts? And why the physical/logical mapping is wrong? It would be very thankful to receive your reply!

Just for overlap with shared experts with EPLB in deepep/pplx. We can't pass the shared experts overlap related logic to deepep / pplx so it will cause acc issue.

USTCKAY:
> > Hi, sorry to comment on a closed PR. You mentioned that EPLB can't be compatible with deepep or pplx, does it mean that we can't enable EPLB when running dp+ep with deepep? Or just cannot enale EPLB when overlapping shared experts? And why the physical/logical mapping is wrong? It would be very thankful to receive your reply!
> 
> Just for overlap with shared experts with EPLB in deepep/pplx. We can't pass the shared experts overlap related logic to deepep / pplx so it will cause acc issue.

Get it! Thanks for your reply!

PR #30588, opened 2025-12-13
Fix edge case Mistral tool parser

joa-stdn:
- Fixing an edge case in Mistral tool parser without streaming where content before [TOOL_CALLS] contains {
- Returning partial json instead of [TOOL_CALLS] when json is wrong, like we do on API (otherwise messes up follow up turn)


vllm/tool_parsers/mistral_tool_parser.py
```diff
@@ -146,7 +146,7 @@ def extract_tool_calls(
             try:
                 if not self._is_pre_v11:
                     function_call_arr = []
-                    for single_tool_content in model_output.split(self.bot_token):
+                    for single_tool_content in model_output.split(self.bot_token)[1:]:
```
joa-stdn: We should not extract anything before the first `[TOOL_CALLS]`

juliendenize: Is it possible to refactor a bit here and what is at line 180 ?
```python
raw_tool_call = self.tool_call_regex.findall(tool_content)[0]
```
Think we could have one line before try-except blocks to separate content and tool calls before entering in the actual tool call parsing logic.

vllm/tool_parsers/mistral_tool_parser.py
```diff
@@ -146,7 +156,7 @@ def extract_tool_calls(
             try:
                 if not self._is_pre_v11:
                     function_call_arr = []
-                    for single_tool_content in model_output.split(self.bot_token):
```
juliendenize, line 152: Line 141 there is this docstring :
```
        """
        Extract the tool calls from a complete model response. Requires
        find-and-replacing single quotes with double quotes for JSON parsing,
        make sure your tool call arguments don't ever include quotes!
        """
```
Does it still makes sense with the recent refactoring ?

vllm/tool_parsers/mistral_tool_parser.py
```diff
@@ -158,18 +168,18 @@ def extract_tool_calls(
 
                         # fn_name is encoded outside serialized json dump
                         # only arguments are serialized
-                        function_call_arr.append(
-                            {"name": fn_name, "arguments": json.loads(args)}
-                        )
+                        function_call_arr.append({"name": fn_name, "arguments": args})
```
juliendenize: Based on slack discussion, i understand here that you avoid loading args because some models like Devstral can sometimes have wrong json generations as they can be complicated without guidance. Is the the right understanding to have ?

However i think it would still be quite nice to load them as a dict when we can to have a best behavior situation. Is it possible to do another try except here to decode json and if it fails return strings ?

I feel like the all try-except blocks may not completely have sense now with your recent refactoring, which is probably a good thing but deserves some clean up. Is it possible for you to work on that or do you need help ?

juliendenize (commented):
Left some comments, nice bug catch on the content before the tool calls.

However, I'm not sure why we'd let the model return tool calls as wrongly parsed JSON as a valid behavior. How would the user be able to use it then ? IMO raising an error might have more sense then outputting something unusuable.

vllm/tool_parsers/mistral_tool_parser.py
joa-stdn: Indeed removing

vllm/tool_parsers/mistral_tool_parser.py
joa-stdn: > load them as a dict when we can to have a best behavior situation

We could do this but since this is not done in streaming mode, not sure if this is adapted?
Ie we could do something like the following but I'm not sure what would be the use case (no change from end user / scaffold perspective and different from streaming mode)
```
def _dump_post_v11_arguments(name: str, args: str) -> dict:
    try:
        normalized_args = json.dumps(json.loads(args), ensure_ascii=False)
    except json.JSONDecodeError:
        normalized_args = args
    return {"name": name, "arguments": normalized_args}
```

PR #30613, opened 2025-12-13
[Bugfix] Add validation for tool requests when tool_parser is unavailable

### LINKED ISSUE #29432: [Bug]: Add validation for tool requests that the tool is available
### 🐛 Describe the bug

Currently there is no check for this, vLLM will just respond sub-optimally in a "silent" way. We should fail the individual request in this case with appropriate error message.

This is a problem in particular for example when using the demo tool server without the `gpt-oss` package installed.

See discussion here: https://github.com/vllm-project/vllm/pull/29336#issuecomment-3573053391

majiayu000:
## Summary

When `tool_choice` is set to `"required"` or a named tool (e.g., `{"type": "function", "function": {"name": "get_weather"}}`), but the server was not started with `--enable-auto-tool-choice` and `--tool-call-parser`, vLLM would silently degrade instead of returning a clear error.

This PR adds validation to return an appropriate error message indicating that these flags are required for tool parsing.

## Changes

- Added validation in `serving_chat.py` to check if `tool_parser` is available when `tool_choice` requires tool parsing
- Added test case covering both `tool_choice="required"` and named tool scenarios

## Test Plan

- [x] Added unit test `test_tool_choice_validation_without_parser`
- [x] Test verifies error response contains guidance about `--enable-auto-tool-choice`

Fixes #29432


vllm/entrypoints/openai/serving_chat.py
```diff
@@ -204,18 +204,31 @@ async def create_chat_completion(
...
+                    # --enable-auto-tool-choice and --tool-call-parser
+                    return self.create_error_response(
+                        '"auto" tool choice requires '
+                        "--enable-auto-tool-choice and --tool-call-parser to be set"
+                    )
+                elif request.tool_choice != "auto":
+                    # "required" or named tool requires tool parser
+                    return self.create_error_response(
+                        f'tool_choice="{request.tool_choice}" requires '
+                        "--enable-auto-tool-choice and --tool-call-parser to be set"
```
DarkLight1337:
```suggestion
                        "--tool-call-parser to be set"
```

DarkLight1337 (approved):
Thanks for improving the UX!

majiayu000:
Done, thanks for the suggestion!

DarkLight1337:
PTAL at the failing test

DarkLight1337:
PTAL at the failing test: https://buildkite.com/vllm/ci/builds/43870/steps/canvas?jid=019b2a35-d04e-415c-a37a-0ad599cb3110#019b2a35-d04e-415c-a37a-0ad599cb3110

DarkLight1337 (approved):

PR #30614, opened 2025-12-13
[Feature] Default EPLB num_redundant_experts to minimum valid value

### LINKED ISSUE #30075: [Feature]: Default eplb num_redundant_experts to the lowest valid value if unspecified

majiayu000:
## Summary

EPLB requires enough physical experts for every expert-parallel rank. This PR
lets `num_redundant_experts` default to `None`, then normalizes it during
`VllmConfig` initialization to the minimum non-negative value that makes the
total number of physical experts divisible by the effective EP size.

## Changes

- default `EPLBConfig.num_redundant_experts` to `None`
- compute the effective EP size as `TP × PCP × DP`
- compute the default as:

  ```text
  (ep_size - num_logical_experts % ep_size) % ep_size
  ```

- normalize disabled EPLB to zero while preserving explicit non-negative values
- reject explicitly negative values
- keep downstream consumers typed against the post-initialization `int` invariant

## Tests

- EPLB enabled with the omitted default
- EPLB disabled normalization
- explicit values remain unchanged
- negative-value validation
- TP / PCP / DP combinations, including PCP values greater than one
- focused Ruff and mypy validation for the touched configuration contract

The full Python 3.10–3.13 pre-commit matrix, DCO, and Buildkite checks on the
current PR head are the final verification gates.

## Scope and provenance

- Fixes #30075.
- GitHub search and the issue history identify this as the only open PR covering
  #30075; maintainers closed the competing duplicates in favor of this PR.
- AI assistance was used to inspect reviewer feedback, verify the Pydantic/mypy
  boundary, and run the focused checks. The amended commit records that
  assistance with an `Assisted-by` trailer.
- Model evaluation is not applicable because this changes configuration
  normalization only; it does not change kernels, model weights, or inference
  outputs.


ApostaC:
@WoosukKwon @youkaichao Can you help take a look at this PR? Thanks!

vllm/config/model.py
```diff
@@ -1040,6 +1040,31 @@ def verify_with_parallel_config(
...
+            if parallel_config.enable_eplb:
+                num_logical_experts = self.get_num_experts()
+                # EP size is TP * DP for EPLB
+                ep_size = (
+                    parallel_config.tensor_parallel_size
+                    * parallel_config.data_parallel_size
+                )
+                # Minimum value ensures at least 1 local physical expert per rank:
+                # (num_logical_experts + num_redundant_experts) / ep_size >= 1
+                min_redundant = max(0, ep_size - num_logical_experts)
```
ilmarkov: Would it make sense to set the `min_redundant` to a value closest divisible by ep_size. So that we would automatically could support unusual ep_sizes? Something like `(ep_size - num_logical_experts % ep_size) % ep_size`

vllm/config/parallel.py
```diff
@@ -53,8 +53,19 @@ class EPLBConfig:
...
     """
 
-    num_redundant_experts: int = Field(default=0, ge=0)
-    """Number of redundant experts to use for expert parallelism."""
+    num_redundant_experts: int | None = None
+    """Number of redundant experts to use for expert parallelism.
+    If None (default), the minimum valid value will be computed automatically
+    based on the number of logical experts and the expert parallel size."""
+
+    @model_validator(mode="after")
```
ilmarkov: I think it has to be either in `_validate_parallel_config` or `_validate_eplb_config` not in separate method just for one config parameter.

tests/test_config.py
```diff
@@ -1050,3 +1050,70 @@ def test_scheduler_config_init():
...
+
+    The minimum valid value formula is:
+        min_redundant = max(0, ep_size - num_logical_experts)
+
+    This ensures at least 1 local physical expert per EP rank:
+        (num_logical_experts + num_redundant_experts) / ep_size >= 1
+    """
+
+    # Test the formula logic directly
+    def compute_min_redundant(num_logical_experts: int, ep_size: int) -> int:
```
ilmarkov: I believe we want to test the executed code in the main codebase not the replicated formula.

ilmarkov (changes requested):
Thank you for you work! Left some comments

tests/test_config.py
majiayu000: Thanks! Updated per your feedback:
 - Formula now uses (ep_size - num_experts % ep_size) % ep_size for non-divisible cases
 - Moved validation to _validate_parallel_config
 - Tests now use parametrize without duplicating formula

tests/test_config.py
```diff
@@ -1050,3 +1050,67 @@ def test_scheduler_config_init():
...
+@pytest.mark.parametrize("num_experts,ep_size,expected", [
+    (8, 8, 0),    # Divisible: 8 % 8 = 0
+    (8, 16, 8),   # ep_size > experts: (16 - 8%16) % 16 = 8
+    (8, 6, 4),    # Non-divisible: (6 - 8%6) % 6 = (6-2)%6 = 4
+    (16, 8, 0),   # Divisible: 16 % 8 = 0
+    (10, 8, 6),   # Non-divisible: (8 - 10%8) % 8 = (8-2)%8 = 6
+    (7, 4, 1),    # Non-divisible: (4 - 7%4) % 4 = (4-3)%4 = 1
+    (1, 4, 3),    # Single expert: (4 - 1%4) % 4 = 3
+])
+def test_eplb_num_redundant_experts_auto_computation(num_experts, ep_size, expected):
```
ilmarkov: In this test you want to check `parallel_config.eplb_config.num_redundant_experts` is computed correctly at the initialization not the formula that you replicate here.

vllm/config/model.py
```diff
@@ -1040,6 +1040,31 @@ def verify_with_parallel_config(
         if parallel_config.enable_expert_parallel:
             self._verify_with_expert_parallelism()
 
+        # Compute num_redundant_experts if not specified
```
ilmarkov: I believe the updating config logic shouldn't be in the function named `verify_`. I would put it in post init of eplb_config.

ilmarkov (commented):
Thanks for the update! Added short comments on the test and logic location. Also, please, fix the pre-commit.

@tlrmchlsmth I recall, you mentioned that such automatic choice of `num_redundant_experts` would be helpful

vllm/config/model.py
majiayu000: The challenge with putting this logic in EPLBConfig's post_init is that it requires `num_logical_experts` which only becomes available from ModelConfig after it's loaded. The EPLBConfig doesn't have access to this information at initialization time.

Current flow:
1. ParallelConfig initialized (EPLBConfig.num_redundant_experts = None)
2. ModelConfig loaded (has num_experts info)
3. VllmConfig calls compute_eplb_num_redundant_experts(num_experts)

This is similar to how other cross-config computations work in VllmConfig's __post_init__.

Would it be acceptable to keep the current approach, or would you prefer I move the call site? Happy to make changes if you have a preferred location in mind.

tests/test_config.py
majiayu000: Updated the tests to check the computed value directly without replicating the formula. The test now uses hardcoded expected values in the parametrize decorator:

```python
@pytest.mark.parametrize(
    ("num_experts", "tp_size", "dp_size", "expected"),
    [
        (8, 2, 2, 0),   # ep_size=4, divisible: 8 % 4 = 0
        (8, 8, 2, 8),   # ep_size=16, need 8 redundant
        (8, 2, 3, 4),   # ep_size=6, need 4 redundant
        ...
    ],
)
def test_eplb_num_redundant_experts_auto_computation(...):
    ...
    assert parallel_config.eplb_config.num_redundant_experts == expected
```

The test now verifies the actual computed value matches the expected value, rather than computing the expected value using the same formula.

vllm/config/parallel.py
```diff
@@ -314,17 +316,63 @@ def _validate_parallel_config(self) -> Self:
...
+            return
+
+        if self.enable_eplb:
+            # EP size is TP * DP for EPLB
+            ep_size = self.tensor_parallel_size * self.data_parallel_size
+            # Ensure (num_logical_experts + num_redundant_experts) is
+            # divisible by ep_size, supporting non-standard ep_size values
+            min_redundant = (ep_size - num_logical_experts % ep_size) % ep_size
+            self.eplb_config.num_redundant_experts = min_redundant
+            logger.info(
```
ilmarkov: Either `info_once` or log only on the root rank.

vllm/config/vllm.py
```diff
@@ -543,6 +543,11 @@ def __post_init__(self):
         self.try_verify_and_update_config()
 
         if self.model_config is not None:
+            # Compute EPLB num_redundant_experts before verification
+            num_experts = self.model_config.get_num_experts()
```
ilmarkov: Use method `is_moe()`.

vllm/config/parallel.py
```diff
@@ -314,17 +316,63 @@ def _validate_parallel_config(self) -> Self:
...
-                    f"{self.eplb_config.num_redundant_experts} but EPLB is not "
-                    "enabled. Either enable EPLB or unset "
-                    "num_redundant_experts."
-                )
+        # Validate num_redundant_experts is non-negative if explicitly set
+        if (
+            self.eplb_config.num_redundant_experts is not None
+            and self.eplb_config.num_redundant_experts < 0
+        ):
+            raise ValueError(
```
ilmarkov: Use `Field(default=None, ge=0)` to avoid adding this validations.

vllm/config/parallel.py
```diff
@@ -61,8 +61,10 @@ class EPLBConfig:
     of the last `lb_window_size` steps will be used for rearranging experts.
     """
 
-    num_redundant_experts: int = Field(default=0, ge=0)
-    """Number of redundant experts to use for expert parallelism."""
+    num_redundant_experts: int | None = None
+    """Number of redundant experts to use for expert parallelism.
+    If None (default), the minimum valid value will be computed automatically
```
ilmarkov, line 74: Needs an explanation that the experts will be rounded to the nearest divisible by EP size

ilmarkov (dismissed):
Sorry for the delay. 
Thank you for the updates! Please sync with the latest main. I also left some minor comments.

vllm/config/parallel.py
```diff
@@ -79,6 +86,14 @@ class EPLBConfig:
...
+    def _validate_eplb_config(self) -> Self:
+        if self.use_async and self.policy != "default":
+            raise ValueError("Async EPLB is only supported with the default policy.")
+        if self.log_balancedness and self.log_balancedness_interval <= 0:
+            raise ValueError("log_balancedness_interval must be greater than 0.")
+        return self
+
 
 @config
 @dataclass
```
ilmarkov: I believe `@config` includes dataclass transform now, so no need to add this decorator.

vllm/config/parallel.py
ilmarkov: Please, address this

ilmarkov (commented):
Thank you for the update. Overall LGTM. Cosmetic nits are left.

majiayu000:
Updated the PR:

- Addressed the remaining cosmetic nits: `logger.info_once` and removal of the extra `@dataclass`.
- Fixed the Ruff/pre-commit issue by moving the EPLB config tests away from the latest-main test insertion point and wrapping the long patch lines.
- Verified the branch is mergeable with latest `origin/main` via `git merge-tree --write-tree HEAD origin/main`.

Local checks run:

```bash
uv run --no-project --with pre-commit pre-commit run --files tests/test_config.py vllm/config/parallel.py vllm/config/vllm.py
/tmp/vllm-pr30614-venv/bin/python -m pytest --noconftest tests/test_config.py -k eplb_num_redundant_experts
```

The targeted EPLB tests passed: `10 passed, 123 deselected`.

majiayu000:
The code update is pushed and the PR is now mergeable. The current `pre-run-check` failure is the repo gate requiring either a `ready`/`verified` label or 4+ previously merged PRs by the author. I tried to add `ready`, but I do not have label permissions on this repo. Could a maintainer please add `ready` or `verified` to let pre-commit run?

vllm/config/vllm.py
```diff
@@ -577,6 +577,12 @@ def __post_init__(self):
         self.try_verify_and_update_config()
 
         if self.model_config is not None:
+            # Compute EPLB num_redundant_experts before verification
+            if self.model_config.is_moe:
+                num_experts = self.model_config.get_num_experts()
```
ilmarkov: Does `get_num_experts` actually returns None?

vllm/config/parallel.py
```diff
@@ -61,8 +61,12 @@ class EPLBConfig:
     of the last `lb_window_size` steps will be used for rearranging experts.
     """
 
-    num_redundant_experts: int = Field(default=0, ge=0)
-    """Number of redundant experts to use for expert parallelism."""
+    num_redundant_experts: int | None = Field(default=None, ge=0)
+    """Number of redundant experts to use for expert parallelism.
+    If None (default), the minimum valid value will be computed automatically
+    based on the number of logical experts and the expert parallel size.
+    The value will be rounded to the nearest multiple of the expert parallel size.
```
ilmarkov: nit: `rounded up`

ilmarkov (commented):
Just small comments.

vllm/config/parallel.py
majiayu000: Done in 0b1e9dc62. The docstring now says `rounded up`.

majiayu000:
Addressed the latest small comments in 0b1e9dc62:

- Removed the redundant `None` guard after `is_moe`; `is_moe` already requires `get_num_experts() > 0`.
- Updated the EPLB docstring wording to say the value is `rounded up`.

Local checks run:

```bash
python3 -m py_compile tests/test_config.py vllm/config/parallel.py vllm/config/vllm.py
uv run --no-project --with pre-commit pre-commit run --files tests/test_config.py vllm/config/parallel.py vllm/config/vllm.py
/tmp/vllm-pr30614-venv/bin/python -m pytest --noconftest tests/test_config.py -k eplb_num_redundant_experts
git merge-tree --write-tree HEAD origin/main
```

The targeted pytest result was `10 passed, 123 deselected`. The branch remains mergeable with latest `origin/main`. The only current CI blocker is still the repo label gate: `pre-run-check` requires `ready`/`verified` or 4+ merged PRs; the author currently has 3.

tlrmchlsmth (approved):
Looks good to me, thanks for the contribution and will improve UX!

tlrmchlsmth:
tests are failing:
```
Value error, World size (4) is larger than the number of available GPUs (1) in this node. If this is intentional and you are using:
--
  | - ray, set '--distributed-executor-backend ray'.
```

@majiayu000 do you think we can mock the tests to not require GPUs?

majiayu000:
Pushed a follow-up fix for the new Buildkite failures. Both failing multimodal shards hit the same root cause: Qwen3.5 loads a FusedMoE expert mapping even though `num_redundant_experts` was still `None`, which caused `num_experts + num_redundant_experts` to raise `TypeError`.

The latest commit normalizes `num_redundant_experts` in `VllmConfig` for every model config, not only when `model_config.is_moe` is true. This preserves the old runtime invariant that disabled EPLB uses `0`, while still allowing `None` as the input sentinel for automatic computation when EPLB is enabled.

Local checks run:
- `python3 -m py_compile vllm/config/vllm.py tests/test_config.py`
- `/tmp/vllm-pr30614-venv/bin/python -m pytest --noconftest tests/test_config.py -k eplb_num_redundant_experts`
- `uv run --no-project --with "pre-commit>=4.5.1" pre-commit run --files tests/test_config.py vllm/config/vllm.py vllm/config/parallel.py`

majiayu000:
Synced with latest `main` and re-ran the full pre-commit suite on the changed files locally — all hooks pass now (ruff check, ruff format, typos, mypy, SPDX, config-docstring validation, etc.).

One additional correctness fix in this update: the default `num_redundant_experts` was computed with `ep_size = TP * DP`, but the EP group actually spans `TP x PCP x DP` (see `init_model_parallel` and the EPLB validation message in `ParallelConfig`). With `prefill_context_parallel_size > 1` the computed default could leave the physical expert count indivisible by the real EP group size. `compute_eplb_num_redundant_experts` now uses `TP * PCP * DP`, and `test_eplb_num_redundant_experts_auto_computation` gained PCP>1 cases.

Could someone please re-trigger CI? Thanks!

hmellor:
Please fix DCO and `pre-commit`

vllm/config/parallel.py
```diff
@@ -69,8 +69,13 @@ class EPLBConfig:
     of the last `lb_window_size` steps will be used for rearranging experts.
     """
 
-    num_redundant_experts: int = Field(default=0, ge=0)
-    """Number of redundant experts to use for expert parallelism."""
+    num_redundant_experts: int | None = Field(default=None, ge=0)
```
hmellor: By default Pydantic won't validate the default. And this tells the rest of vLLM that it can always expect this field to be `int` by the time it is read

```suggestion
    num_redundant_experts: int = Field(default=None, ge=0)
```

tests/test_config.py
```diff
@@ -1050,3 +1050,36 @@ def test_scheduler_config_init():
...
+    # Should not raise - None is allowed when EPLB is disabled
+    assert parallel_config.eplb_config.num_redundant_experts is None
+
+    # Test ParallelConfig validation - EPLB disabled with non-zero value
+    with pytest.raises(ValueError, match="EPLB is not enabled"):
+        ParallelConfig(
+            enable_eplb=False,
+            enable_expert_parallel=False,
+            eplb_config=EPLBConfig(num_redundant_experts=4),
+        )
```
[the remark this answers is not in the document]
majiayu000: Addressed by the current test suite and validated on head 906df9e. The implementation has since moved from ModelConfig.verify_with_parallel_config to ParallelConfig.compute_eplb_num_redundant_experts, invoked during VllmConfig.__post_init__. The parameterized test executes that production method across 10 TP/PCP/DP cases and asserts the computed value and divisibility; separate tests cover disabled normalization to 0, VllmConfig runtime normalization, explicit values, and negative validation. Current-head DCO and the full pre-commit/mypy matrix are green.

majiayu000:
The current head is synchronized with upstream main and all current-head DCO and pre-commit/mypy checks are green. I also updated the PR description to the final TP × PCP × DP formula and current test coverage, and resolved the now-covered test thread with exact evidence. Since the previous approval predates the runtime normalization, PCP fix, and upstream synchronization, could a maintainer please perform a fresh current-head review?

PR #30626, opened 2025-12-13
[docker] Restructure Dockerfile for more efficient and cache-friendly builds

amrmahdi:
## Approach:
- Pre-install PyTorch, FlashInfer, and other slow-changing dependencies in vllm-base before installing the vLLM wheel for better layer caching
- Add parallel extensions-build stage for DeepGEMM and EP kernels
- Move stable packages (accelerate, bitsandbytes, etc.) earlier in build

This allows incremental builds with Python-only changes to skip the expensive dependency installation layers.

## Performance:
Incremental builds with Python-only changes now complete in ~16 minutes (previously 35+ minutes).

## Future work:
We considered building these base stages as separate images that could be built independently and baked into CI AMIs for maximum cache reuse. However, this introduces maintenance burden of extra pipelines and update strategies. The inline approach is simpler and can be optimized later by baking the main build image into AMIs daily to maximize layer cache reuse.


docker/Dockerfile
```diff
@@ -107,32 +105,31 @@ RUN echo 'tzdata tzdata/Areas select America' | debconf-set-selections \
...
 RUN ldconfig /usr/local/cuda-$(echo $CUDA_VERSION | cut -d. -f1,2)/compat/
 
+# ============================================================
+# SLOW-CHANGING DEPENDENCIES BELOW
+# These are the expensive layers that we want to cache
+# ============================================================
+
+# Install PyTorch and core CUDA dependencies
+# This is ~2GB and rarely changes
+ARG PYTORCH_VERSION=2.6.0
```
mgoin: This should be 2.9.0? But also doesn't look like it is actually used

docker/Dockerfile
```diff
@@ -408,62 +425,107 @@ RUN echo 'tzdata tzdata/Areas select America' | debconf-set-selections \
...
-    uv pip install --system dist/*.whl --verbose \
-        --extra-index-url ${PYTORCH_CUDA_INDEX_BASE_URL}/cu$(echo $CUDA_VERSION | cut -d. -f1,2 | tr -d '.')
+# ============================================================
+# SLOW-CHANGING DEPENDENCIES BELOW
+# These are the expensive layers that we want to cache
+# ============================================================
+
+# Install PyTorch and core CUDA dependencies
+# This is ~2GB and rarely changes
+ARG PYTORCH_VERSION=2.6.0
```
mgoin: Ditto, what is PYTORCH_VERSION used for?

docker/Dockerfile
```diff
@@ -142,13 +139,15 @@ RUN --mount=type=cache,target=/root/.cache/uv \
...
-# see https://github.com/pytorch/pytorch/pull/123243
+# CUDA arch list used by torch
+# Explicitly set the list to avoid issues with torch 2.2
+# See https://github.com/pytorch/pytorch/pull/123243
 ARG torch_cuda_arch_list='7.0 7.5 8.0 8.9 9.0 10.0 12.0'
 ENV TORCH_CUDA_ARCH_LIST=${torch_cuda_arch_list}
-#################### BASE BUILD IMAGE ####################
+
+# Set library path for CUDA
+ENV LD_LIBRARY_PATH=/usr/local/cuda/lib64:/usr/local/nvidia/lib64:$LD_LIBRARY_PATH
```
mgoin: I see LD_LIBRARY_PATH set several times now, can we consolidate this?

mgoin (commented):
LGTM overall, great work. Just a few quick nits to cleanup

mgoin (approved):
LGTM, thanks for the nice work!

PR #30650, opened 2025-12-14
[Bugfix] CustomAR + TritonAttn[AMPERE] + FULL_CG - gpt-oss

bbrowning:
## Purpose

Full CUDA graph capture + replay (the default since vLLM v0.11.0) could lead to memory correctness issues in the custom_all_reduce implementation due to previous logic that skipped running the actual all_reduce operation during the compile and warmup phase. This resulted in compile and warmup seeing a memory allocation, but not observing the custom op call or the actual reduce operation which could lead to optmizations being taken assuming purely functional code with no side-effects when in fact the reduce operation inherently has side effects.

To fix this, instead of allocating an empty tensor to mimic the reduce operation, we now call the actual all_reduce during compile and warmup as well.

Multiple users reported an issue that tracks back to this root cause on Ampere hardware, which is one platform where we use custom_all_reduce by default if tensor parallelism is used and NVLink is available.

Examples:
- https://github.com/vllm-project/vllm/issues/26480
- https://github.com/vllm-project/vllm/issues/29998
- https://github.com/vllm-project/vllm/issues/30498

Note that until I rebased this on top of a very recent main, I also had to tag the all_reduce custom op registration in parallel_state.py with `torch.Tag.maybe_aliasing_or_mutating` to fix the reproducer described in #29998. The tag is no longer required on latest main, although I have not identified exactly why or what changed that fixed that.

## Test Plan

On Ampere hardware with NVLink enabled (2x A5500 in my case), start gpt-oss-20b with tensor-parallel-size 2:

```
  --tool-call-parser openai \
  --enable-auto-tool-choice \
  --tensor-parallel-size 2
```

Then, execute the curl reproducer from #29998 in a loop:

```
while true; do curl -X POST http://localhost:8000/v1/chat/completions  -H "Content-Type: application/json"   -d '{
    "stream": false,
    "messages": [
      {
        "role": "system",
      },
      {
        "role": "user",
        "content": "Hi"
      },
      {
        "role": "assistant",
      },
      {
        "role": "user",
      }
    ],
    "tools": [
      {
        "type": "function",
        "function": {
          "parameters": {
            "type": "object",
            "properties": {
              "url": {
                "type": "string",
              }
            },
            "required": ["url"]
          }
        }
      }
    ]
  }'; done
```

I also exercised the reproducer from #30498, slightly modified to point to gpt-oss-20b (instead of 120b) and to raise the max_tokens to 8192 as it was generating valid but verbose output for the first few iterations of the loop.

## Test Result

Before this change, the 2nd time the curl loop from #29998 executed it always hung and the vLLM continued to generated token id 0 indefinitely. After this change, the 2nd and all subsequent curl requests succeed as expected.

Before this change, the the python requests loop from #30498 always hung by the 2nd or 3rd request. After this change, it continues to work for multiple subsequent requests.

I was able to successfully reproduce the test cases in both of those issues on my hardware and this change fixes both of those.

### DIFF
--- a/vllm/distributed/device_communicators/custom_all_reduce.py
+++ b/vllm/distributed/device_communicators/custom_all_reduce.py
@@ -273,9 +273,10 @@ class CustomAllreduce:
             if torch.cuda.is_current_stream_capturing():
                 return self.all_reduce(input, registered=True)
             else:
-                # If warm up, mimic the allocation pattern since custom
-                # allreduce is out-of-place.
-                return torch.empty_like(input)
+                # During warmup, we need to run the all_reduce operation so
+                # that the allocation, copy, and collective behaviors
+                # are consistent between warmup, capture, and replay.
+                return self.all_reduce(input, registered=False)
         else:
             # Note: outside of cuda graph context, custom allreduce incurs a
             # cost of cudaMemcpy, which should be small (<=1% of overall


bbrowning:
While all user reports of this that I've come across have been on Ampere hardware (A5500, A6000, A1000 80GB), I have also been able to reproduce this on H100 GPUs when Triton ATTN is in use. After reproducing the same issue on an H100 (by setting `VLLM_ATTENTION_BACKEND=TRITON_ATTN`), I can confirm this fixes that issue on Hopper GPUs as well when Triton ATTN is in use.

xyang16:
> While all user reports of this that I've come across have been on Ampere hardware (A5500, A6000, A1000 80GB), I have also been able to reproduce this on H100 GPUs when Triton ATTN is in use. After reproducing the same issue on an H100 (by setting `VLLM_ATTENTION_BACKEND=TRITON_ATTN`), I can confirm this fixes that issue on Hopper GPUs as well when Triton ATTN is in use.

Thanks for the fix! Curious why skipping all_reduce operation would only cause problem for Triton ATTN on H100?

bbrowning:
> > While all user reports of this that I've come across have been on Ampere hardware (A5500, A6000, A1000 80GB), I have also been able to reproduce this on H100 GPUs when Triton ATTN is in use. After reproducing the same issue on an H100 (by setting `VLLM_ATTENTION_BACKEND=TRITON_ATTN`), I can confirm this fixes that issue on Hopper GPUs as well when Triton ATTN is in use.
> 
> Thanks for the fix! Curious why skipping all_reduce operation would only cause problem for Triton ATTN on H100?

I'm not 100% sure that doesn't cause problems for other attention backends. However, the reproducers we've had reported from users all had Triton ATTN in common, and were all on Ampere hardware (where we use Triton ATTN by default). By manually forcing Triton ATTN on H100 hardware, I was also able to reproduce it there.

There could theoretically be correctness issues with other attention backends in certain scenarios that this fixes, but I have no direct evidence of that myself.

ApostaC:
Hey @mgoin , could you please help review this PR? Thanks!

bbRLdev:
I have been reliably producing this issue on 4x 40 GB A100 but not on 2x 80 GB H100 with the current master branch. I will let you know if this PR fixes this issue.

vllm/distributed/device_communicators/custom_all_reduce.py
```diff
@@ -273,9 +273,10 @@ def custom_all_reduce(self, input: torch.Tensor) -> torch.Tensor | None:
             if torch.cuda.is_current_stream_capturing():
                 return self.all_reduce(input, registered=True)
             else:
-                # If warm up, mimic the allocation pattern since custom
-                # allreduce is out-of-place.
-                return torch.empty_like(input)
```
bbrowning, line 278: This could maybe accomplish the same thing if it was `torch.zeros_like(input)`? Initial tests seem promising, but I'm not sure how to interpret what that means as far as the root issue here and what this actually fixes.

bbrowning:
With the merging of #30887, I believe this fix is no longer needed.

bbrowning:
> So I was able to test this PR, and while it does prevent GPT OSS from simply generating nothing now, it appears that some bugs that were fixed in #28729 are back, particularly the one that I described where no final answer is produced [#28729 (comment)](https://github.com/vllm-project/vllm/pull/28729#issuecomment-3550411048)
> 
> ...

@bbRLdev I am looking into this using the sample request you gave to see if I can reproduce it on top of the latest main. I appreciate the detailed report, and it might be best to pull your comment out to a new issue so that we can properly track finding the root cause and solving what you're seeing. I believe you've hit on a separate real bug that we need to fix, as with your reproducer even in non-streaming mode I see the model is generating tool calls but we're not properly getting those back to the caller for whatever reason.

bbrowning:
@bbRLdev To fully track this down, I'll need to find the full parameters Cline is sending to vLLM here, including the `tool_choice` and `tools` parameters. The request body you gave had no tools, but the model is generating a tool call internally in response to this request.

PR #30652, opened 2025-12-14
Strengthen input validation and tests for 'parse_raw_prompts’.

mivehk:
## Purpose
The ‘parse_raw_prompts’ can benefit from the stricter input validation in the array-of-arrays input path. This is for when input is a list of lists; however, we verify the nested inner list is a non-null list of integers beyond the first inner list verification.

## Test Plan
Added a pytest coverage for invalid and edge-case inputs of mixed-type nested lists for parse_raw_prompts. Updated the array-of-arrays input condition of parse_raw_prompts to strictly check empty prompt lists within list and mixed-type nested lists. Validated these invalid inputs by running  “pytest tests/test_inputs.py” to confirm correct exception types are raised for malformed inputs.

## Test Result
Before this change, mixed-type nested prompt inputs were not checked for rejection and could lead to ambiguous behavior. After applying strict validation, the new and existing tests pass, and malformed inputs deterministically raise TypeError or ValueError as expected. 

---
<details>

</details>

### DIFF
--- a/tests/test_inputs.py
+++ b/tests/test_inputs.py
@@ -34,6 +34,13 @@ INPUTS_SLICES = [
 ]
 
 
+# Test that a nested mixed-type list of lists raises a TypeError.
+@pytest.mark.parametrize("invalid_input", [[[1, 2], ["foo", "bar"]]])
+def test_invalid_input_raise_type_error(invalid_input):
+    with pytest.raises(TypeError):
+        parse_raw_prompts(invalid_input)
+
+
 def test_parse_raw_single_batch_empty():
     with pytest.raises(ValueError, match="at least one prompt"):
         parse_raw_prompts([])
--- a/tests/test_inputs.py
+++ b/tests/test_inputs.py
@@ -33,11 +33,11 @@ INPUTS_SLICES = [
     slice(None, None, -2),
 ]
 
+# List of lists that prompt[0] passes test case#4, but no TypeError will be raised.
 @pytest.mark.parametrize(
     "invalid_input",
     [
-        ["foo", 1], # mixed of string and token
-        [["foo"], ["bar"]] # list of list of strings
+        [[1, 2], ["foo", "bar"]] 
     ]
 )
 def test_invalid_input_raise_type_error(invalid_input):
--- a/tests/test_inputs.py
+++ b/tests/test_inputs.py
@@ -34,13 +34,6 @@ INPUTS_SLICES = [
 ]
 
 
-# Test that a nested mixed-type list of lists raises a TypeError.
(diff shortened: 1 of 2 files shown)


DarkLight1337 (approved):
Thanks

DarkLight1337:
PTAL at the failing test

vllm/inputs/parse.py
```diff
@@ -45,16 +45,23 @@ def parse_raw_prompts(
...
+            for elem in prompt:
+                if not isinstance(elem, list):
+                    raise TypeError(
+                        "prompt must be a list of lists, but found a non-list element."
+                    )
+                if not elem:
+                    raise ValueError(
+                        "Empty lists of tokens are not allowed in prompts."
+                    )
+                if not is_list_of(elem, int, check="all"):
```
DarkLight1337: Let's avoid using "all" check as it's quite expensive for long prompts

mivehk: Parse_raw_prompts is updated to avoid expensive asymptotic check all.

tests/test_inputs.py
```diff
@@ -34,13 +34,6 @@
 ]
 
 
-# Test that a nested mixed-type list of lists raises a TypeError.
```
DarkLight1337, line 37: I think this test case should still be valid?

mivehk: The test was meant to enforce thorough validation across all nested lists, which goes beyond the intended scope of parse_raw_prompts. That function performs lightweight structural checks, avoiding asymptotically expensive ‘check=all’ validation.

DarkLight1337: We still check the first item of each prompt (`1` and `foo` in this case) which should still trigger the error

mivehk: The newly added test is not merely a counterexample for a previously undetected edge case. It simulates a realistic integration scenario where raw (string-based) prompts are inadvertently mixed with tokenized (integer-based) prompts within a batched input. The implementation continues to rely on is_list_of for lightweight validation, while additionally ensuring that the first element of each inner list is validated, which strengthens correctness without introducing deep expensive validation.

tests/test_inputs.py
```diff
@@ -34,8 +34,8 @@
...
-@pytest.mark.parametrize("invalid_input", [[[1, 2], ["foo", "bar"]]])
+# Per-prompt validation: mixed batched raw-string & tokenized prompts raise TypeError.
+@pytest.mark.parametrize("invalid_input", [[["key", "cat"], [2, 3]]])
```
DarkLight1337: Just use the original test case? Why does it need to be changed?

mivehk: I reverted to the original test case. Thanks

PR #30653, opened 2025-12-14
Revert "[Fix]Load kv-cache dtype from hf_quant_config.json automatically"

robertgshaw2-redhat:
Reverts vllm-project/vllm#29980

This breaks the CI: https://buildkite.com/vllm/ci/builds/43393#019b176c-8881-47f4-aaf9-5dac38740906

The reason is that CUTLASS MLA does not support FP8 kv cache

### DIFF
--- a/vllm/utils/torch_utils.py
+++ b/vllm/utils/torch_utils.py
@@ -194,33 +194,12 @@ def get_kv_cache_torch_dtype(
     return torch_dtype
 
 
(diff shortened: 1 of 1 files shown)


robertgshaw2-redhat:
I opened up this issue: https://github.com/vllm-project/vllm/issues/30654

mgoin (approved):

PR #30657, opened 2025-12-14
[Log] Skip piecewise cudagraph warn when using full cudagraph

BoyuanFeng:
When using full cudagraph `vllm bench latency -cc.cudagraph_mode=FULL -cc.splitting_ops=[]`, log still shows 

```
[compilation.py:935] Using piecewise compilation with empty splitting_ops
```

We should only have this log if we are using piecewise cudagraph and splitting_ops=[]. This PR improves the log.

### DIFF
--- a/vllm/config/compilation.py
+++ b/vllm/config/compilation.py
@@ -932,9 +932,13 @@ class CompilationConfig:
                 self.splitting_ops = list(self._attention_ops)
                 added_default_splitting_ops = True
             elif len(self.splitting_ops) == 0:
-                logger.warning_once(
-                    "Using piecewise compilation with empty splitting_ops"
-                )
+                if (
+                    self.cudagraph_mode == CUDAGraphMode.PIECEWISE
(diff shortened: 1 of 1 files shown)


robertgshaw2-redhat (approved):

vllm/config/compilation.py
```diff
@@ -932,9 +932,13 @@ def set_splitting_ops_for_v1(
...
             elif len(self.splitting_ops) == 0:
-                logger.warning_once(
-                    "Using piecewise compilation with empty splitting_ops"
-                )
+                if (
+                    self.cudagraph_mode == CUDAGraphMode.PIECEWISE
+                    or self.cudagraph_mode == CUDAGraphMode.FULL_AND_PIECEWISE
+                ):
+                    logger.warning_once(
+                        "Using piecewise compilation with empty splitting_ops"
```
ProExpertProg, line 940: piecewise cudagraphs?

BoyuanFeng: yes updated in #30728.

PR #30680, opened 2025-12-15
[Model] Add video input support for transformers modeling backend

ch3nku1:
I am a developer from Cybercore. We are developing a multimodal model named Leum and plan to deploy it using the vLLM transformers modeling backend. We noticed that the current implementation does not support video input, which is a necessary feature for our model. This pull request introduces the required changes to enable video input processing.

Key changes:
- Extended multimodal classes (`MultiModalProcessingInfo`, `MultiModalProcessor`, `MultiModalMixin`) to handle video-specific logic, including token calculation, dummy data generation, and embedding.
- Corrected the frame size extraction for video frames in `vllm/multimodal/parse.py`.
- Updated documentation to reflect video support.
- Fixed a potential OOM issue in the dummy batch generator for multimodal models.

Thank you for considering our contribution!

---

Edits from maintainers.

Merge after:

- [ ] https://github.com/huggingface/transformers/pull/42919

### DIFF
--- a/vllm/model_executor/models/transformers/multimodal.py
+++ b/vllm/model_executor/models/transformers/multimodal.py
@@ -33,7 +33,11 @@ from vllm.multimodal.inputs import (
     MultiModalUUIDDict,
     PlaceholderRange,
 )
-from vllm.multimodal.parse import ImageProcessorItems, MultiModalDataItems
+from vllm.multimodal.parse import (
+    ImageProcessorItems,
+    MultiModalDataItems,
+    VideoProcessorItems,
+)
 from vllm.multimodal.processing import BaseMultiModalProcessor, BaseProcessingInfo
 from vllm.multimodal.profiling import BaseDummyInputsBuilder
 from vllm.sequence import IntermediateTensors
@@ -55,10 +59,13 @@ DYNAMIC_ARG_DIMS = {
 
 class MultiModalProcessingInfo(BaseProcessingInfo):
     def get_supported_mm_limits(self):
-        return {"image": None}
+        return {"image": None, "video": None}
 
     def get_mm_max_tokens_per_item(self, seq_len, mm_counts):
-        return {"image": self.get_max_image_tokens()}
+        return {
+            "image": self.get_max_image_tokens(),
+            "video": self.get_max_video_tokens(seq_len),
+        }
 
     def get_max_image_tokens(self) -> int:
         width, height = self.get_max_image_size()
@@ -71,20 +78,53 @@ class MultiModalProcessingInfo(BaseProcessingInfo):
         image_tokens = mm_tokens["num_image_tokens"][0]
         return image_tokens
 
+    def _get_video_tokens(self, num_frames, width, height) -> int:
+        processor = self.get_hf_processor()
+        multimodal_config = self.ctx.model_config.multimodal_config
+        mm_processor_kwargs = multimodal_config.mm_processor_kwargs or {}
+        mm_tokens = processor._get_num_multimodal_tokens(
+            video_sizes=([num_frames, height, width],), **mm_processor_kwargs
+        )
+        video_tokens = mm_tokens["num_video_tokens"][0]
+        return video_tokens
+
+    def get_max_video_tokens(self, seq_len: int) -> int:
+        width, height = self.get_max_image_size()
+        num_frames = self.get_max_video_frames(seq_len)
+        return self._get_video_tokens(num_frames, width, height)
+
     def get_max_image_size(self):
         return 10_000, 10_000  # hardcode for arbitrary very large size
+
+    def get_max_video_frames(self, seq_len: int) -> int:
+        width, height = self.get_max_image_size()
+
+        max_num_frames = 1
+
+        while True:
+            next_num_frames = max_num_frames + 1
+            video_tokens = self._get_video_tokens(next_num_frames, width, height)
+            if video_tokens > seq_len:
+                break
+
+            max_num_frames = next_num_frames
+
+        return max_num_frames
 
 
 class MultiModalDummyInputsBuilder(BaseDummyInputsBuilder[MultiModalProcessingInfo]):
     def get_dummy_text(self, mm_counts: Mapping[str, int]) -> str:
         num_images = mm_counts.get("image", 0)
+        num_videos = mm_counts.get("video", 0)
 
         processor = self.info.get_hf_processor()
         if "gemma3" in processor.__class__.__name__.lower():
             image_token = processor.boi_token
+            video_token = ""
         else:
             image_token = getattr(processor, "image_token", "")
-        return image_token * num_images
+            video_token = getattr(processor, "video_token", "")
+        return image_token * num_images + video_token * num_videos
 
     def get_dummy_mm_data(
         self,
@@ -93,8 +133,11 @@ class MultiModalDummyInputsBuilder(BaseDummyInputsBuilder[MultiModalProcessingIn
         mm_options: Mapping[str, "BaseDummyOptions"] | None = None,
     ) -> MultiModalDataDict:
         num_images = mm_counts.get("image", 0)
+        num_videos = mm_counts.get("video", 0)
 
         target_width, target_height = self.info.get_max_image_size()
+        max_total_frames = self.info.get_max_video_frames(seq_len)
+        target_num_frames = max_total_frames // max(num_videos, 1)
 
         image_overrides = mm_options.get("image") if mm_options else None
 
@@ -104,6 +147,12 @@ class MultiModalDummyInputsBuilder(BaseDummyInputsBuilder[MultiModalProcessingIn
                 height=target_height,
                 num_images=num_images,
                 overrides=image_overrides,
+            ),
+            "video": self._get_dummy_videos(
+                width=target_width,
+                height=target_height,
+                num_frames=target_num_frames,
+                num_videos=num_videos,
             ),
         }
 
@@ -148,8 +197,18 @@ class MultiModalProcessor(BaseMultiModalProcessor[MultiModalProcessingInfo]):
 
         # Keep these as batched, as they always have batch size as first dim
         mm_fields["image_grid_thw"] = MultiModalFieldConfig.batched("image")
-        mm_fields["video_grid_thw"] = MultiModalFieldConfig.batched("image")
         mm_fields["num_image_patches"] = MultiModalFieldConfig.batched("image")
+
+        video_grid_thw = hf_inputs.get("video_grid_thw", torch.empty((0, 3)))
+        video_grid_sizes = video_grid_thw.prod(-1)
+        mm_fields["pixel_values_videos"] = MultiModalFieldConfig.flat_from_sizes(
+            "video", video_grid_sizes
+        )
+        mm_fields["video_embeds"] = MultiModalFieldConfig.flat_from_sizes(
+            "video", video_grid_sizes
+        )
+        mm_fields["video_grid_thw"] = MultiModalFieldConfig.batched("video")
+        mm_fields["num_video_patches"] = MultiModalFieldConfig.batched("video")
         return mm_fields
 
     def _get_hf_mm_data(
@@ -211,24 +270,36 @@ class MultiModalProcessor(BaseMultiModalProcessor[MultiModalProcessingInfo]):
 
         # We can infer vLLM style placeholder from token type ids, if we split
         # it for each input `mm_data`.
-        mm_positions = torch.where(mm_token_type_ids == 1)[1]
-        images = mm_items.get_items("image", ImageProcessorItems)
+        image_sizes = []
+        if "image" in mm_items:
+            images = mm_items.get_items("image", ImageProcessorItems)
+            for item_idx in range(len(images)):
+                image_size = images.get_image_size(item_idx)
+                image_sizes.append((image_size.height, image_size.width))
+
+        video_sizes = []
+        if "video" in mm_items:
+            videos = mm_items.get_items("video", VideoProcessorItems)
+            for item_idx in range(len(videos)):
+                video_size = videos.get_frame_size(item_idx)
+                num_frames = videos.get_num_frames(item_idx)
+                video_sizes.append((num_frames, video_size.height, video_size.width))
+
         multimodal_config = self.info.ctx.model_config.multimodal_config
         mm_processor_kwargs = multimodal_config.mm_processor_kwargs or {}
-        image_sizes = []
-        for item_idx in range(len(images)):
-            image_size = images.get_image_size(item_idx)
-            image_sizes.append((image_size.height, image_size.width))
 
         mm_tokens_per_modality = hf_processor._get_num_multimodal_tokens(
-            image_sizes=image_sizes, **mm_processor_kwargs
+            image_sizes=image_sizes, video_sizes=video_sizes, **mm_processor_kwargs
         )
 
         mm_placeholders = {}
+
+        # image_token_ids
+        mm_positions = torch.where(mm_token_type_ids == 1)[1]
         split_sizes = mm_tokens_per_modality["num_image_tokens"]
         if split_sizes:
             chunked_mm_positions = torch.split(mm_positions, split_sizes)
-            mm_tokens = torch.tensor(prompt_ids)[mm_token_type_ids[0].bool()]
+            mm_tokens = torch.tensor(prompt_ids)[mm_token_type_ids[0] == 1]
             chunked_mm_tokens = torch.split(mm_tokens, split_sizes)
             ranges = [
                 PlaceholderRange(
@@ -238,11 +309,34 @@ class MultiModalProcessor(BaseMultiModalProcessor[MultiModalProcessingInfo]):
                 )
                 for positions, mm_tokens in zip(chunked_mm_positions, chunked_mm_tokens)
             ]
-            mm_placeholders = {"image": ranges}
+            mm_placeholders["image"] = ranges
 
         processed_data["num_image_patches"] = torch.tensor(
             mm_tokens_per_modality["num_image_patches"]
         )
+
+        # video_token_ids
+        mm_positions = torch.where(mm_token_type_ids == 2)[1]
+
+        split_sizes = mm_tokens_per_modality["num_video_tokens"]
+        if split_sizes:
+            chunked_mm_positions = torch.split(mm_positions, split_sizes)
+            mm_tokens = torch.tensor(prompt_ids)[mm_token_type_ids[0] == 2]
+            chunked_mm_tokens = torch.split(mm_tokens, split_sizes)
+            ranges = [
+                PlaceholderRange(
+                    offset=positions[0].item(),
+                    length=positions.shape[0],
+                    is_embed=(mm_tokens == hf_processor.video_token_id).bool(),
+                )
+                for positions, mm_tokens in zip(chunked_mm_positions, chunked_mm_tokens)
+            ]
+            mm_placeholders["video"] = ranges
+
+        processed_data["num_video_patches"] = torch.tensor(
+            mm_tokens_per_modality["num_video_patches"]
+        )
+
         mm_kwargs = MultiModalKwargsItems.from_hf_inputs(
             processed_data,
             self._get_mm_fields_config(processed_data, hf_processor_mm_kwargs),
@@ -333,21 +427,29 @@ class MultiModalMixin(SupportsMultiModal, SupportsMRoPE):
     def embed_multimodal(self, **kwargs):
         pixel_values: torch.Tensor | None = kwargs.pop("pixel_values", None)
         image_embeds: torch.Tensor | None = kwargs.pop("image_embeds", None)
+        pixel_values_videos: torch.Tensor | None = kwargs.pop(
+            "pixel_values_videos", None
+        )
+        video_embeds: torch.Tensor | None = kwargs.pop("video_embeds", None)
+
         # Model might use `image_patches` instead of `pixel_values`
         if pixel_values is None:
             pixel_values = kwargs.pop("image_patches", None)
 
+        multimodal_embeddings: tuple[torch.Tensor, ...] = ()
+
         if image_embeds is not None:
-            return image_embeds
+            multimodal_embeddings += tuple(image_embeds)
 
-        if pixel_values is None:
-            return None
-
-        num_image_patches = kwargs.pop("num_image_patches")
         kwargs.pop("token_type_ids", None)  # used only in `forward`
+
         if pixel_values is not None:
+            num_image_patches = kwargs.pop("num_image_patches")
             vision_embeddings = self.model.get_image_features(pixel_values, **kwargs)
 
+            if isinstance(vision_embeddings, tuple):
+                # For qwen3 vl, The deepstack visual features are also returned
+                vision_embeddings = vision_embeddings[0]
             if isinstance(vision_embeddings, torch.Tensor):
                 if vision_embeddings.ndim == 2:
                     vision_embeddings = vision_embeddings.unsqueeze(0)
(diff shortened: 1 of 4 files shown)


vllm/v1/worker/gpu_model_runner.py
```diff
@@ -1118,7 +1118,18 @@ def _dummy_mm_kwargs(self, num_seqs: int) -> BatchedTensorInputs:
...
+        # The mm_dummy_batch below is only retrieved when
+        # supports_multimodal_raw_input_only is True.
+        # Currently, only the transform modeling backend and terratorch have
+        # supports_multimodal_raw_input_only as True.
+        # When testing the transform modeling backend, it was found that
+        # if num_seqs (usually the default 256) is passed in here,
+        # an OOM error occurs.
+        # It needs to be confirmed what value should be passed in here,
+        # for now it is fixed to 1.
+        return self._get_mm_dummy_batch(dummy_modality, 1)
```
hmellor, line 1132: I was recently working in this area and noticed this too. It causes 256 x 100MP images (the max image size defined in `vllm/model_executor/models/transformers/multimodal.py`) to be materialised on the GPU.

@ywang96 what do you think should be the correct behaviour here?

vllm/multimodal/parse.py
```diff
@@ -280,7 +280,7 @@ def get_frame_size(self, item_idx: int) -> ImageSize:
         if isinstance(image, PILImage.Image):
             return ImageSize(*image.size)
         if isinstance(image, (np.ndarray, torch.Tensor)):
-            _, h, w = image.shape
+            w, h, _ = image.shape
```
hmellor, line 283: I doubt that this was wrong for all models already in vLLM, why does this need changing here?

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -71,20 +78,53 @@ def get_max_image_tokens(self) -> int:
...
 
 class MultiModalDummyInputsBuilder(BaseDummyInputsBuilder[MultiModalProcessingInfo]):
     def get_dummy_text(self, mm_counts: Mapping[str, int]) -> str:
         num_images = mm_counts.get("image", 0)
+        num_videos = mm_counts.get("video", 0)
 
         processor = self.info.get_hf_processor()
         if "gemma3" in processor.__class__.__name__.lower():
             image_token = processor.boi_token
+            video_token = ""
```
hmellor:
```suggestion
```

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -71,20 +78,53 @@ def get_max_image_tokens(self) -> int:
...
         else:
             image_token = getattr(processor, "image_token", "")
-        return image_token * num_images
+            video_token = getattr(processor, "video_token", "")
```
hmellor: Given that Gemma3 will have no `video_token` this should be fine, right?

```suggestion
        video_token = getattr(processor, "video_token", "")
```

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -105,6 +148,12 @@ def get_dummy_mm_data(
                 num_images=num_images,
                 overrides=image_overrides,
             ),
+            "video": self._get_dummy_videos(
+                width=target_width,
+                height=target_height,
+                num_frames=target_num_frames,
+                num_videos=num_videos,
+            ),
```
hmellor, line 161: We should also provide video overrides, right?

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -238,11 +309,34 @@ def apply(
...
+                    length=positions.shape[0],
+                    is_embed=(mm_tokens == hf_processor.video_token_id).bool(),
+                )
+                for positions, mm_tokens in zip(chunked_mm_positions, chunked_mm_tokens)
+            ]
+            mm_placeholders["video"] = ranges
+
+        processed_data["num_video_patches"] = torch.tensor(
+            mm_tokens_per_modality["num_video_patches"]
+        )
```
hmellor, line 343: This code is _almost_ identical to the video code, could it be abstracted to a method?

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -148,8 +197,18 @@ def _get_mm_fields_config(
...
+        video_grid_thw = hf_inputs.get("video_grid_thw", torch.empty((0, 3)))
+        video_grid_sizes = video_grid_thw.prod(-1)
+        mm_fields["pixel_values_videos"] = MultiModalFieldConfig.flat_from_sizes(
+            "video", video_grid_sizes
+        )
+        mm_fields["video_embeds"] = MultiModalFieldConfig.flat_from_sizes(
+            "video", video_grid_sizes
+        )
+        mm_fields["video_grid_thw"] = MultiModalFieldConfig.batched("video")
+        mm_fields["num_video_patches"] = MultiModalFieldConfig.batched("video")
```
hmellor, line 216: This is a place where the name of something on the Transformers side is assumed. We'd need to make sure that the Transformers team is happy with this as the standard name which will be propagated to all models in Transformers.

cc @zucchini-nlp

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -55,10 +59,13 @@
...
 class MultiModalProcessingInfo(BaseProcessingInfo):
     def get_supported_mm_limits(self):
-        return {"image": None}
+        return {"image": None, "video": None}
 
     def get_mm_max_tokens_per_item(self, seq_len, mm_counts):
-        return {"image": self.get_max_image_tokens()}
+        return {
+            "image": self.get_max_image_tokens(),
+            "video": self.get_max_video_tokens(seq_len),
```
zucchini-nlp, line 71: I remember that vLLM chooses only one modality when profling, whichever has more tokens. So I think we will need to safe-get `mm_tokens["num_video_tokens"]` and otherwise set to `0` because not all models support videos

vllm/model_executor/models/transformers/multimodal.py
zucchini-nlp: yeah, this is the part which might fails for some models. There are models that treat videos as a sequence of images and thus don't have a specific video token 🥲  It made me crazy trying to make them work

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -148,8 +197,18 @@ def _get_mm_fields_config(
 
         # Keep these as batched, as they always have batch size as first dim
         mm_fields["image_grid_thw"] = MultiModalFieldConfig.batched("image")
-        mm_fields["video_grid_thw"] = MultiModalFieldConfig.batched("image")
         mm_fields["num_image_patches"] = MultiModalFieldConfig.batched("image")
+
+        video_grid_thw = hf_inputs.get("video_grid_thw", torch.empty((0, 3)))
+        video_grid_sizes = video_grid_thw.prod(-1)
```
zucchini-nlp, line 208: not super generalizable, only qwen-like models have a THW tensor. Ideally we need to use `num_video_patches` as in image modality

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -211,24 +270,36 @@ def apply(
...
         multimodal_config = self.info.ctx.model_config.multimodal_config
         mm_processor_kwargs = multimodal_config.mm_processor_kwargs or {}
-        image_sizes = []
-        for item_idx in range(len(images)):
-            image_size = images.get_image_size(item_idx)
-            image_sizes.append((image_size.height, image_size.width))
 
         mm_tokens_per_modality = hf_processor._get_num_multimodal_tokens(
-            image_sizes=image_sizes, **mm_processor_kwargs
+            image_sizes=image_sizes, video_sizes=video_sizes, **mm_processor_kwargs
```
zucchini-nlp, line 297: I don't think all processor expect `video_sizes`, though it will be swallowed by `kwargs` and shouldn't raise issues

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -238,11 +309,34 @@ def apply(
...
+        if split_sizes:
+            chunked_mm_positions = torch.split(mm_positions, split_sizes)
+            mm_tokens = torch.tensor(prompt_ids)[mm_token_type_ids[0] == 2]
+            chunked_mm_tokens = torch.split(mm_tokens, split_sizes)
+            ranges = [
+                PlaceholderRange(
+                    offset=positions[0].item(),
+                    length=positions.shape[0],
+                    is_embed=(mm_tokens == hf_processor.video_token_id).bool(),
+                )
```
zucchini-nlp, line 336: the hard part here are models which add timestamps between each video frame and the timestamps are encoded differently, they have no special tokens. So inferring the ranges is not as easy as with images

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -333,21 +427,29 @@ def __init__(self, multimodal_model):
...
+        )
+        video_embeds: torch.Tensor | None = kwargs.pop("video_embeds", None)
+
         # Model might use `image_patches` instead of `pixel_values`
         if pixel_values is None:
             pixel_values = kwargs.pop("image_patches", None)
 
-        if image_embeds is not None:
-            return image_embeds
+        multimodal_embeddings: tuple[torch.Tensor, ...] = ()
```
hmellor: Since tuples are immutable, would we be better accumilating a list and then casting it to a tuple at the end?

(also can we type hint `embed_multimodal`)

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -238,11 +309,34 @@ def apply(
...
+                    offset=positions[0].item(),
+                    length=positions.shape[0],
+                    is_embed=(mm_tokens == hf_processor.video_token_id).bool(),
+                )
+                for positions, mm_tokens in zip(chunked_mm_positions, chunked_mm_tokens)
+            ]
+            mm_placeholders["video"] = ranges
+
+        processed_data["num_video_patches"] = torch.tensor(
+            mm_tokens_per_modality["num_video_patches"]
```
zucchini-nlp, line 342: `mm_tokens_per_modality` isn't guaranteed to contain `num_video_patches` if the model doesn't support video modality. Same for `key=num_video_tokens`

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -333,21 +427,29 @@ def __init__(self, multimodal_model):
...
-        if pixel_values is None:
-            return None
+        if image_embeds is not None:
+            multimodal_embeddings += tuple(image_embeds)
 
-        num_image_patches = kwargs.pop("num_image_patches")
         kwargs.pop("token_type_ids", None)  # used only in `forward`
+
         if pixel_values is not None:
+            num_image_patches = kwargs.pop("num_image_patches")
```
hmellor: I think this was outside the if block so that it was always popped from `kwargs` regardless of if we used it. The same shold probably be done for `num_video_patches`

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -333,21 +427,29 @@ def __init__(self, multimodal_model):
...
             vision_embeddings = self.model.get_image_features(pixel_values, **kwargs)
 
+            if isinstance(vision_embeddings, tuple):
+                # For qwen3 vl, The deepstack visual features are also returned
```
zucchini-nlp, line 457: right, though qwen3-vl will give bad performance without deepstack visual features, Currently vLLM doesn't support models like Qwen3-VL and Ovis2 on purpose

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -333,21 +427,29 @@ def __init__(self, multimodal_model):
...
+                vision_embeddings = vision_embeddings[0]
```
hmellor, line 458: This might not be intentional. @zucchini-nlp should the return value of Qwen3 VL have been changed to this?

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -362,8 +464,37 @@ def embed_multimodal(self, **kwargs):
...
+                # but transformers returns concat tensors if each patch
+                # is of different size. We split it back to make vLLM happy
+                vision_embeddings = torch.split(
+                    vision_embeddings, num_video_patches.flatten().tolist()
+                )
+                vision_embeddings = [
+                    embed.flatten(start_dim=0, end_dim=-2)
+                    for embed in vision_embeddings
+                ]
+            multimodal_embeddings += tuple(vision_embeddings)
```
hmellor: Similar comment to above, this is very similar to the image code. Could they be deduplicated?

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -386,18 +517,14 @@ def get_mrope_input_positions(
             if k not in {"image_grid_thw", "video_grid_thw"}
         ):
             raise NotImplementedError(
-                "Transformers modeling backend only supports images."
+                "Transformers modeling backend only supports images and videos."
             )
 
         image_grid_thw = kwargs.get("image_grid_thw", [])
         video_grid_thw = kwargs.get("video_grid_thw", [])
```
hmellor: Now that we're not trying to create empty tensors with these

```suggestion
        image_grid_thw = kwargs.get("image_grid_thw", None)
        video_grid_thw = kwargs.get("video_grid_thw", None)
```

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -386,18 +517,14 @@ def get_mrope_input_positions(
...
 
-        image_grid_thw = (torch.stack if image_grid_thw else torch.tensor)(
-            image_grid_thw
-        )
-        video_grid_thw = (torch.stack if video_grid_thw else torch.tensor)(
-            video_grid_thw
-        )
+        image_grid_thw = torch.stack(image_grid_thw) if image_grid_thw else None
+        video_grid_thw = torch.stack(video_grid_thw) if video_grid_thw else None
```
hmellor, line 532: @zucchini-nlp was it important that vLLM passed empty grids rather than `None` to `get_rope_index`?

hmellor (commented):
Awesome PR! Thank you for using and contributing to the Transformers modelling backend!

I've left a few comments and looped in @zucchini-nlp. I want to make sure that the standards we're defining here for video model interfaces align with how Transformers would like to standardise all the video models in the Transformers library.

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -362,8 +464,37 @@ def embed_multimodal(self, **kwargs):
...
+                if vision_embeddings.ndim == 2:
+                    vision_embeddings = vision_embeddings.unsqueeze(0)
+
+                # Embeddings have to be 2D tensors of length `num_images`
+                # but transformers returns concat tensors if each patch
+                # is of different size. We split it back to make vLLM happy
+                vision_embeddings = torch.split(
+                    vision_embeddings, num_video_patches.flatten().tolist()
+                )
+                vision_embeddings = [
```
zucchini-nlp, line 496: I am not 100% sure, though for video modality this might not be needed. Mostly videos don't have several patches per one frame

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -362,8 +464,37 @@ def embed_multimodal(self, **kwargs):
...
+                    embed.flatten(start_dim=0, end_dim=-2)
+                    for embed in vision_embeddings
+                ]
+            multimodal_embeddings += tuple(vision_embeddings)
 
-            return vision_embeddings
+        return multimodal_embeddings
```
zucchini-nlp, line 502: @hmellor does vLLM know how to handle the tuple output here, and are there any assumptions made?

vllm/model_executor/models/transformers/multimodal.py
```diff
@@ -211,24 +270,36 @@ def apply(
...
 
         mm_placeholders = {}
+
+        # image_token_ids
+        mm_positions = torch.where(mm_token_type_ids == 1)[1]
         split_sizes = mm_tokens_per_modality["num_image_tokens"]
         if split_sizes:
             chunked_mm_positions = torch.split(mm_positions, split_sizes)
-            mm_tokens = torch.tensor(prompt_ids)[mm_token_type_ids[0].bool()]
+            mm_tokens = torch.tensor(prompt_ids)[mm_token_type_ids[0] == 1]
```
DarkLight1337, line 307: Avoid computing `mm_token_type_ids == 1` twice

DarkLight1337: Also is there somewhere in Transformers where the token type ID is defined by modality? I don't like this magic number

zucchini-nlp (commented):
Left a few comments from my experience trying to make videos compatible with the backend. IMO this PR has to wait until necessary changes are made in transformers side for video LLMs. I am quite sure we can't support all of them, but we could try to support 70%

There aren't many models, around 10 I think in total with explicit video support.

vllm/v1/worker/gpu_model_runner.py
DarkLight1337: I think here the profiling is actually correct for models that have `is_multimodal_raw_input_only_model`, since the maximum number of videos possible during inference is indeed based on `max_num_seqs` (and can be actually more than that if the model accepts more than 1 video per prompt).

DarkLight1337: So I prefer patching this on the model side. If removing `supports_multimodal_raw_input_only=True` is not feasible, then maybe you can override the `mm_counts` in `get_dummy_mm_data` to use the same value as for regular MM models.

vllm/model_executor/models/transformers/multimodal.py
hmellor: Qwen2-VL would be a good example of this with `get_num_frames_with_most_features`. But it gets its `mm_counts` from the limits in `get_supported_mm_limits` unless the user specified otherwise. 

Ideally there would be a way to set `get_supported_mm_limits()["video"]` to zero when the model doesn't support video.

vllm/model_executor/models/transformers/multimodal.py
hmellor: Yeah, this method is supposed to return 
```py
MultiModalEmbeddings: TypeAlias = list[Tensor] | Tensor | tuple[Tensor, ...]
```

vllm/model_executor/models/transformers/multimodal.py
zucchini-nlp: We never supportted qwen3-vl in vLLM because of this 😿  The model has to propagate `deepstack visual features` down to the LM's forward, which we could theoretically overcome

We are currently doing standardization of `get_image_features` in transformers side and I think the best would be to always ask for a dict output here (i.e. we get all the outputs from vision encoder) and pass it over to LM, let it handle encoder outputs whichever way it pleases

vllm/model_executor/models/transformers/multimodal.py
zucchini-nlp: should be `None` if the modality is not present

vllm/model_executor/models/transformers/multimodal.py
zucchini-nlp: Yeah, that would be great. I even can say how we infer if model supports videos from model class 😄 It has a class attribute `model.input_modalities`

vllm/model_executor/models/transformers/multimodal.py
hmellor: Do you mean in vLLM with the Transformers modelling backend? vLLM does support this model natively.

Yeah a standardisation on the Transformers side would be great.

zucchini-nlp: Yes, with the backend the model was not supported when released

vllm/model_executor/models/transformers/multimodal.py
hmellor: Oh ok, are these class attributes or instance attributes?

vllm/model_executor/models/transformers/multimodal.py
hmellor: To clarify, it should be `None` if:
- the modality is not present in the individual request? (i.e. an image model but no image has been passed)
- the modelity is not present in the model? (i.e. a model which doesn't support video)

vllm/v1/worker/gpu_model_runner.py
hmellor: I forget the exact reason we chose to use `supports_multimodal_raw_input_only=True`. IIRC it was because it reduced the amount of necessary monkey patching because the processor already exists on the Transformers side so it's needlessly complicated to monkey patch the processing on the vLLM side.

@DarkLight1337 do you know more about the key differences between `supports_multimodal_raw_input_only` being `True`/`False`?

hmellor: > since the maximum number of videos possible during inference is indeed based on `max_num_seqs`

Why is this not also the case when `is_multimodal_raw_input_only_model=False`?

vllm/model_executor/models/transformers/multimodal.py
zucchini-nlp: class attributes available from v5 and on

vllm/v1/worker/gpu_model_runner.py
DarkLight1337: > I forget the exact reason we chose to use `supports_multimodal_raw_input_only=True`

cc @Isotr0py do you remember anything about this?

vllm/multimodal/parse.py
ch3nku1: I'm also very confused about this. I'm not sure why other models (like minicpmv) works.
I see that `_get_dummy_videos` uses a shape of (num_frames, width, height, 3)
https://github.com/vllm-project/vllm/blob/0d0c929f2360cde5bae6817ad0f555641329e79d/vllm/multimodal/profiling.py#L225
while `VideoLoader._read_frames` uses (num_frames, height, width, 3)
https://github.com/vllm-project/vllm/blob/0d0c929f2360cde5bae6817ad0f555641329e79d/vllm/multimodal/video.py#L77
so I think it should be w, h, c or h, w, c here.

DarkLight1337: For profiling purposes, the width and height are interchangable in general since most transformations don't care about that

vllm/v1/worker/gpu_model_runner.py
DarkLight1337: > the key differences between supports_multimodal_raw_input_only being True/False?

IIRC this is to pass `kwargs` to `forward` method instead of `embed_multimodal`.

@christian-pinto maybe you can explain more about this.

christian-pinto: Back then we have introduced `supports_multimodal_raw_input_only` because some models would not process multimodal embeddings but instead they would need to process the raw multimodal data. 

So setting `supports_multimodal_raw_input_only=True` will pass the actual multi-modal data to the model via kwargs. This means that during the warmup phase vLLM will get  256 (`max_num_seq`) dummy multimodal data from your model and load it to GPU. The amount of data loaded (e.g., 100MP) depends on how your model class builds the dummy data.

vllm/multimodal/parse.py
DarkLight1337: Strictly speaking the profiling code is wrong, so feel free to change that.

vllm/model_executor/models/transformers/multimodal.py
ch3nku1: I made this change because `mm_token_type_ids[0].bool()` cannot distinguish between image and video, which causes issues when testing with simultaneous image and video inputs.

vllm/v1/worker/gpu_model_runner.py
zucchini-nlp: FWIW `supports_multimodal_raw_input_only` is needed for the backend only with gemma3 models because we want to pass over some processing outputs to model's forward. We still do the whole `mm_embedding` part and manually pass over extra kwargs 

I am quite sure that is not how `supports_multimodal_raw_input_only` was designed but I cannot find a better way for vLLM to pass some processor output directly to model

vllm/model_executor/models/transformers/multimodal.py
zucchini-nlp: This part is called only with qwen-like models as I see, so it's when modality isn't present in request, I can't think of any model with mrope that doesn't support videos, we're safe

vllm/v1/worker/gpu_model_runner.py
DarkLight1337: Maybe you can handle it in a similar way as `visual_token_mask` in `InternVLChatModel`. Since we assume that `embed_multimodal` is called before `forward` in every batch, you can store the inputs to `embed_multimodal` temporarily and then use them inside `forward`.

christian-pinto: We use, and have introduced multimodal_raw_input processing, when adding the models in the `Terratorch` class.

Please, see my comment above for the rationale behind that field.

vllm/model_executor/models/transformers/multimodal.py
ch3nku1: > We are currently doing standardization of `get_image_features` in transformers side and I think the best would be to always ask for a dict output here (i.e. we get all the outputs from vision encoder) and pass it over to LM, let it handle encoder outputs whichever way it pleases

@zucchini-nlp Is there a rough timeline?
If there’s an issue or PR tracking it, please point me to it, I’d be happy to help move this forward.

zucchini-nlp: The PR is drafted in https://github.com/huggingface/transformers/pull/42564, though there is no specific deadline. Should be merged before v5 definitely

vllm/model_executor/models/transformers/multimodal.py
ch3nku1: > right, though qwen3-vl will give bad performance without deepstack visual features, Currently vLLM doesn't support models like Qwen3-VL and Ovis2 on purpose

Is there any plan to support Qwen3-VL?

zucchini-nlp: not yet. It is doable, and we would need some changes in transformers. There is already some PRs in progress. Right now the code assumes that there are no extra inputs except for input embeddings

Issue #30679, opened 2025-12-15
[RFC]: Replace `torch.cuda` API with `torch.accelerator` for better hardware compatiblity.


jikunshang:
### Motivation.

vLLM is a framework support multi hardware backend. while there are some torch.cuda hard code call. this is unfriendly to non-cuda compatible device. Fortunately, there is a new set of `torch.accelerator` [API](https://docs.pytorch.org/docs/stable/accelerator.html) in pytorch which can dispatch based on platform.

Meanwhile, we should add some lint tool to avoid add more `torch.cuda` call in new added code.

### Proposed Change.

torch accelerator API status on torch-2.9.0:

|cuda API name| unified torch API name| torch API Status | vLLM replace status| 
|-------|-------|-------|-------|
|torch.cuda.Event| torch.Event | | |
|torch.cuda.Stream| torch.Stream | | |
|torch.cuda.device_count |torch.accelerator.device_count  | | |
|torch.cuda.is_available |torch.accelerator.is_available | | |
|torch.cuda.synchronize | torch.accelerator.synchronize| | |
|torch.cuda.set_stream | torch.accelerator.set_stream| | |
|torch.cuda.current_device | torch.accelerator.current_accelerator| | |
|torch.cuda.current_stream |torch.accelerator.current_stream | | |
|torch.cuda.empty_cache | torch.accelerator.empty_cache | | https://github.com/vllm-project/vllm/pull/30681 |
|torch.cuda.max_memory_allocated | torch.accelerator.max_memory_allocated | | |
|torch.cuda.max_memory_reserved | torch.accelerator.max_memory_reserved  | | |
|torch.cuda.memory_allocated | torch.accelerator.memory_allocated | | |
|torch.cuda.memory_reserved | torch.accelerator.memory_reserved  | | |
|torch.cuda.memory_stats | torch.accelerator.memory_stats | | |
|torch.cuda.reset_peak_memory_stats | torch.accelerator.reset_peak_memory_stats | | |

### CC List.

@youkaichao @simon-mo @WoosukKwon @zhuohan123 

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

hongxiayang:
This is doable, the accelerator APIs are stable.

baonudesifeizhai:
could i  help with it ?

[...]

PR #10477, opened 2024-11-20
[Core] Free CPU pinned memory on environment cleanup

janimo:
When using pinned memory, such as with CPU offload, the physical RAM will not be freed on shutdown, only when the process exits or at some unknown time in the future. To permit subsequent runs not OOMing, we can free it using a call introduced in Pytorch 2.5.

vLLM depends on Pytorch 2.5 but I put in the check defensively.

Context:
https://github.com/pytorch/pytorch/issues/134332

PR #30681, opened 2025-12-15
[Hardware] Replace `torch.cuda.empty_cache` with `torch.accelerator.empty_cache`

jikunshang:
## Purpose
vLLM is a framework support multi hardware backend. while there are some torch.cuda hard call. this is unfriendly to non-cuda compatible device. fortunately, there is a new set of `torch.accelerator` API which can dispatch based on platform.
I will try to create a series of PR to address this issue, start from `empty_cache` API.

[1/N] of https://github.com/vllm-project/vllm/issues/30679 

## Test Plan
CI.

## Test Result

---
<details>

</details>


.pre-commit-config.yaml
```diff
@@ -136,6 +136,14 @@ repos:
     types: [python]
     pass_filenames: false
     additional_dependencies: [regex]
+  # forbid use torch.cuda APIs
+  - id: forbid-direct-torch-cuda
+    name: "Forbid direct 'torch.cuda' APIs"
+    entry: python tools/pre_commit/check_torch_cuda.py
+    language: python
+    types: [python]
+    pass_filenames: false
```
hmellor: We should pass file names, otherwise this will run on all files every commit

```suggestion
```

tools/pre_commit/check_torch_cuda.py
hmellor, line 1: The hook doesn't need to work out which files it should run on, pre-commit already does that for us. Please look at `tools/pre_commit/check_pickle_imports.py` to see how this should be done

hmellor (dismissed):

tools/pre_commit/check_torch_cuda.py
jikunshang: thanks for review. I rewrite the check_torch_cuda.py file.

hmellor: Thanks, the tool looks much nicer now!

tools/pre_commit/check_torch_cuda.py
```diff
@@ -0,0 +1,39 @@
...
+def scan_file(path: str) -> int:
+    with open(path, encoding="utf-8") as f:
+        for i, line in enumerate(f, 1):
+            if _TORCH_CUDA_RE.search(line):
+                print(
+                    f"{path}:{i}: "
+                    "\033[91merror:\033[0m "  # red color
+                    "Found torch.cuda API call"
+                )
+                return 1
```
hmellor:
```suggestion
    with open(path, encoding="utf-8") as f:
        content = f.read()
    for match in re.finditer(_TORCH_CUDA_RE, content, re.MULTILINE):
        # Calculate line number from match position
        line_num = content[: match.start() + 1].count("\n") + 1
        print(
            f"{path}:{line_num}: "
            "\033[91merror:\033[0m "  # red color
            "Found torch.cuda API call"
        )
        return 1
```

tools/pre_commit/check_torch_cuda.py
```diff
@@ -0,0 +1,39 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+import sys
+
+import regex as re
+
+# --------------------------------------------------------------------------- #
+# Regex: match `torch.cuda.xxx` but allow `torch.accelerator.xxx`
+# --------------------------------------------------------------------------- #
+_TORCH_CUDA_RE = re.compile(r"\btorch\.cuda\.empty_cache\b")
```
hmellor:
```suggestion
_TORCH_CUDA_RE = r"\btorch\.cuda\.empty_cache\b"
```

tools/pre_commit/check_torch_cuda.py
```diff
@@ -0,0 +1,41 @@
...
+    for match in re.finditer(_TORCH_CUDA_RE, content, re.MULTILINE):
+        # Calculate line number from match position
+        line_num = content[: match.start() + 1].count("\n") + 1
+        print(
+            f"{path}:{line_num}: "
+            "\033[91merror:\033[0m "  # red color
+            "Found torch.cuda API call"
+        )
+        return 1
+    return 0
```
hmellor, line 30:
```suggestion
    with open(path, encoding="utf-8") as f:
        content = f.read()
    return_code = 0
    for match in re.finditer(_TORCH_CUDA_RE, content, re.MULTILINE):
        # Calculate line number from match position
        line_num = content[: match.start() + 1].count("\n") + 1
        print(
            f"{path}:{line_num}: "
            "\033[91merror:\033[0m "  # red color
            "Found torch.cuda API call"
        )
        return_code = 1
    return return_code
```

hmellor:
Please let me know when this is ready to come out of draft

jikunshang:
> Please let me know when this is ready to come out of draft

Thanks for reviewing this. I will be OOO for next 2 weeks. I will revisit this when come back from holiday with vllm-ascend folks. cc @Yikun

xinyu-intel (approved):

hmellor (changes requested):
I like the change, I'd just like to request that the pre-commit hook be updated ot use `re.finditer` on the entire file instead of `re.match` on each line. Since `regex` is compiled, this will run significantly faster than the Python loop and `re.match`. I think this is worth doing for DX.

You can follow the example in `tools/pre_commit/check_forbidden_imports.py`

vllm/distributed/parallel_state.py
hmellor, line 1: Is it safe to replace `torch._C._host_emptyCache()` with `torch.accelerator.empty_cache()`? Are they the same method?

jikunshang: thanks for pointing this, I have misunderstanding for `torch._C._host_emptyCache()`  before. 
it's added in https://github.com/vllm-project/vllm/pull/10477, we should keep it.

vllm/distributed/parallel_state.py
```diff
@@ -1732,14 +1732,14 @@ def cleanup_dist_env_and_memory(shutdown_ray: bool = False):
...
+    if not current_platform.is_cpu():
+        torch.accelerator.empty_cache()
+        try:
             torch._C._host_emptyCache()
-    except AttributeError:
-        logger.warning("torch._C._host_emptyCache() only available in Pytorch >=2.5")
+        except AttributeError:
+            logger.warning(
+                "torch._C._host_emptyCache() only available in Pytorch >=2.5"
+            )
```
hmellor, line 1926: This block was not confitional on `if not current_platform.is_cpu()` before?

```suggestion
    try:
        torch._C._host_emptyCache()
    except AttributeError:
        logger.warning("torch._C._host_emptyCache() only available in Pytorch >=2.5")
```

vllm/distributed/parallel_state.py
```diff
@@ -1732,14 +1732,14 @@ def cleanup_dist_env_and_memory(shutdown_ray: bool = False):
     gc.collect()
     from vllm.platforms import current_platform
 
-    empty_cache = current_platform.empty_cache
-    if empty_cache is not None:
-        empty_cache()
-    try:
-        if not current_platform.is_cpu():
+    if not current_platform.is_cpu():
+        torch.accelerator.empty_cache()
```
hmellor, line 1920: If the call is no longer CUDA specific does it need to be guarded?

```suggestion
    torch.accelerator.empty_cache()
```

jikunshang: my understanding is: cpu is not an accelerator, so it will throw error if I force call such API on cpu. see failed ci case here https://buildkite.com/vllm/ci/builds/53084#019c9394-6f03-4f78-9ed6-d5db707646d5

vllm/distributed/parallel_state.py
jikunshang: `empty_cache = current_platform.empty_cache` this will be emtpy method for cpu platform.

vllm/distributed/parallel_state.py
jikunshang: cc @bigPYJ1151  for cpu part.

vllm/distributed/parallel_state.py
hmellor: This doesn't call `empty_cache`, it calls `torch._C._host_emptyCache()`

vllm/distributed/parallel_state.py
```diff
@@ -1732,14 +1732,14 @@ def cleanup_dist_env_and_memory(shutdown_ray: bool = False):
...
+        try:
             torch._C._host_emptyCache()
-    except AttributeError:
-        logger.warning("torch._C._host_emptyCache() only available in Pytorch >=2.5")
```
jikunshang, line 1926: @hmellor 
before code logic is:
```
call empty_cache for all platform (if possible, actually do nothing on cpu)
then for non-cpu platform, call torch._C._host_emptyCache()
```

after:
```
for non-cpu platform,
    call empty_cache 
    then call torch._C._host_emptyCache()
(do nothing here for cpu platform)
```
I think they are equal, anything I miss?

hmellor: Thanks for the explanation, I see this makes sense now

hmellor (approved):
LGTM! Let's see what CI says

PR #30686, opened 2025-12-15
Exclude .git from Docker build to Improve layer cache hits

### LINKED ISSUE #23588: [CI]: Reduce docker build time with caching

The build image stage currently takes ~40 minutes and most other tests can't start until this is complete.

We should ensure rarely-changing parts/layers are cached, with changes to the layer arrangement to accommodate this if necessary.

In particular if there are no changes to dependencies or C++/cuda code (e.g. if the changes are only docs and/or python files), the update should be equivalent to overlaying just the changed files within the file structure in the container which should in theory take seconds.

The vast majority of PRs are in this category - we should get to the point where the image setup time is < 1min for these.

### Alternatives

_No response_

### Additional context

_No response_

wzshiming:
Fixed #23588

## Purpose

Preventing the vllm wheel build layer from being invalidated when only .git metadata changes

Now all the cache can be hit, It only takes 1 minute.

This is the details of all cache hits: https://buildkite.com/vllm/ci/builds/44332/steps/canvas?sid=019b3610-dd64-46fd-aad0-4a85752c9d6b

## Test Result

---
<details>

</details>

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -31,3 +31,5 @@ share/python-wheels/
 .installed.cfg
 *.egg
 MANIFEST
+docker
+.deps
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,3 +1,5 @@
+# syntax=docker/dockerfile:1
+
 # The vLLM Dockerfile is used to construct vLLM image that can be directly used
 # to run the OpenAI compatible server.
 
@@ -394,11 +396,8 @@ WORKDIR /workspace
 
 # Copy pre-built csrc wheel directly
 COPY --from=csrc-build /workspace/dist /precompiled-wheels
-COPY . .
 
-ARG GIT_REPO_CHECK=0
-RUN --mount=type=bind,source=.git,target=.git \
-    if [ "$GIT_REPO_CHECK" != "0" ]; then bash tools/check_repo.sh ; fi
+COPY --exclude=.git . .
 
 ARG vllm_target_device="cuda"
 ENV VLLM_TARGET_DEVICE=${vllm_target_device}
@@ -413,8 +412,8 @@ RUN --mount=type=cache,target=/root/.cache/uv \
     fi
 
 # Build the vLLM wheel
+ARG VLLM_VERSION_OVERRIDE="0.0.0+src.build"
 RUN --mount=type=cache,target=/root/.cache/uv \
-    --mount=type=bind,source=.git,target=.git \
     if [ "${vllm_target_device}" = "cuda" ]; then \
         export VLLM_PRECOMPILED_WHEEL_LOCATION=$(ls /precompiled-wheels/*.whl); \
     fi && \
--- a/.buildkite/release-pipeline.yaml
+++ b/.buildkite/release-pipeline.yaml
@@ -16,7 +16,7 @@ steps:
         commands:
           # #NOTE: torch_cuda_arch_list is derived from upstream PyTorch build files here:
           # https://github.com/pytorch/pytorch/blob/main/.ci/aarch64_linux/aarch64_ci_build.sh#L7
-          - "DOCKER_BUILDKIT=1 docker build --build-arg max_jobs=16 --build-arg USE_SCCACHE=1 --build-arg GIT_REPO_CHECK=1 --build-arg CUDA_VERSION=12.9.1 --build-arg torch_cuda_arch_list='8.7 8.9 9.0 10.0+PTX 12.0' --tag vllm-ci:build-image --target build --progress plain -f docker/Dockerfile ."
+          - "DOCKER_BUILDKIT=1 docker build --build-arg max_jobs=16 --build-arg USE_SCCACHE=1 --build-arg CUDA_VERSION=12.9.1 --build-arg torch_cuda_arch_list='8.7 8.9 9.0 10.0+PTX 12.0' --tag vllm-ci:build-image --target build --progress plain -f docker/Dockerfile ."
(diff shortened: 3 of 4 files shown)


docker/Dockerfile
```diff
@@ -307,18 +307,14 @@ COPY --from=csrc-build /workspace/dist /precompiled-wheels
...
-RUN --mount=type=bind,source=.git,target=.git \
-    if [ "$GIT_REPO_CHECK" != "0" ]; then bash tools/check_repo.sh ; fi
-
 ARG vllm_target_device="cuda"
 ENV VLLM_TARGET_DEVICE=${vllm_target_device}
 
 # Skip adding +precompiled suffix to version (preserves git-derived version)
 ENV VLLM_SKIP_PRECOMPILED_VERSION_SUFFIX=1
 
+ARG VLLM_VERSION_OVERRIDE="0.0.0+src.build"
```
wzshiming, line 415: Defined in 
https://github.com/vllm-project/vllm/blob/ac1c934276167085dc7d9e5adb8c2089de0460f9/setup.py#L647-L653

docker/Dockerfile
```diff
@@ -307,18 +307,14 @@ COPY --from=csrc-build /workspace/dist /precompiled-wheels
 
 COPY . .
 
-ARG GIT_REPO_CHECK=0
-RUN --mount=type=bind,source=.git,target=.git \
-    if [ "$GIT_REPO_CHECK" != "0" ]; then bash tools/check_repo.sh ; fi
-
```
wzshiming, line 402: As each commit is different, the .git file will inevitably cause the Docker build cache to be missing.

It has been tested that if one layer is missing, all subsequent layers will also be missing.

Maybe we should move it to run before docker build?

docker/Dockerfile
wzshiming: The parameter needs to be added when releasing the image.
Adding this parameter to the test will also cause the cache to miss.
```
--build-arg VLLM_VERSION_OVERRIDE=$(python3 -m setuptools_scm -f plain) 
```

.dockerignore
```diff
@@ -31,3 +32,5 @@ share/python-wheels/
 .installed.cfg
 *.egg
 MANIFEST
+docker
+.deps
```
wzshiming, line 35: It's very strange that `.deps` also causes cache misses, It took a long time to locate it.

I'm guessing it's because there are residual `.deps` in the runner environment, and they're not the same on every runner.

docker/Dockerfile
wzshiming, line 415: > Also built image will no longer have a git-derived version string. What version will vllm --version report with this change?

@amrmahdi

It will show what is filled in here.

amrmahdi: Can you include the release pipeline change in this PR so the picture full picture is clear?

amrmahdi: Thanks @wzshiming

I think with this change we probably don't need the following step in the tests stage right ?
```
RUN echo 'tzdata tzdata/Areas select America' | debconf-set-selections \
    && echo 'tzdata tzdata/Zones/America select Los_Angeles' | debconf-set-selections \
    && apt-get update -y \
    && apt-get install -y git
```

amrmahdi (approved):
LGTM -- I'm ok with the trade-off of CI builds having a dummy version -- others maybe know of other use cases that I'm not aware of where this is needed.

docker/Dockerfile
wzshiming: > Thanks [@wzshiming](https://github.com/wzshiming)
> 
> I think with this change we probably don't need the following step in the tests stage right ?
> 
> ```
> RUN echo 'tzdata tzdata/Areas select America' | debconf-set-selections \
>     && echo 'tzdata tzdata/Zones/America select Los_Angeles' | debconf-set-selections \
>     && apt-get update -y \
>     && apt-get install -y git
> ```

I've tested it and it doesn't work, the next step requires git

amrmahdi: yeah just noticed in the tests, thanks

amrmahdi:
> I've changed it to only exclude the .git directory before the build so that the logic doesn't change too much afterward.

@wzshiming I ran a build with an empty commit on top of your changes, I see the same cache misses happening, and no improvement in build times 

https://buildkite.com/vllm/ci/builds/47249/steps/canvas?sid=019bc2be-dacb-42f5-909f-6cbca0587a4e

amrmahdi:
> I've noticed that there seems to be a new problem that causes the cache to get lost before the exclude .git step, which is very strange.

 Hey @wzshiming, I've been investigating the same cache invalidation issue and I think the remaining cache misses you were seeing are likely caused by the bake config file leaking into the Docker build context (fixed in https://github.com/vllm-project/vllm/pull/34569).

Can we switch the approach to add the .git to .dockerignore instead of using COPY --exclude=.git. It keeps .git out of the entire build context in one place (no per-stage --exclude flags, no # syntax=docker/dockerfile:1 directive needed), and avoids sending .git to the BuildKit daemon at all.

One thing I noticed is missing: tests/standalone_tests/python_only_compile.sh runs git merge-base HEAD origin/main inside the container, which will fail without .git. It needs a fallback to a VLLM_MERGE_BASE_COMMIT env var passed as a build arg.

I have some additional caching improvements (narrowing the build stage COPY . . to specific directories, and splitting vllm-base into cacheable stages) that I can send as a follow-up once your PR lands. Let  me know if you'd like help with any of the above.

PR #28677, opened 2025-11-13
[Bugfix][Nixl] Fix kernel physical<>logical block_size issue 

NickLucche:
Fix https://github.com/vllm-project/vllm/issues/28661.
PD deployments with block_size=128  *on FA* (as well as FI) are currently broken on main.

This is pretty bad because 128 is the suggested `block_size` as it maximizes the xfer window. 
This is due to the changes introduced in https://github.com/vllm-project/vllm/pull/24486/, which adjust `block_size`at runtime based on Backend constraints.
This update value is not picked by NixlConnector and it ends up using the one defined by the user (to be precise, the updated value it's not stored anywhere).

Example
```
kv_cache.shape=torch.Size([3636, ->64, 8, 128])
nixl_worker.block_size = ->128
```

Furthermore, `kv_cache_manager.get_block_ids()` returns manager (logical) block IDs, but the connector (particularly nixl_connector) needs kernel-space block IDs.
Currently there's not a clean accessible way to map from logical->physical block_ids from anywhere in code. Also it is not clear in terms of Connector *interface contract*, whether blocks supplied by the Scheduler should be logical or already physical.
For the sake of getting this bug fixed, I have implemented the mapping logic within the connector worker. 

Bug was spotted after https://github.com/vllm-project/vllm/pull/27753/ landed, as it changed the default FA backend to not support 128 anymore.

## Test with

Same procedure as described on issue https://github.com/vllm-project/vllm/issues/28661:

```
bash tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh
...
(EngineCore_DP0 pid=74640) INFO 11-13 17:19:11 [distributed/.../v1/nixl_connector.py:1142] User-specified logical block size (128) does not match physical kernel block size (64). Using the latter.
...
(APIServer pid=497251) INFO:     Application shutdown complete.
+ echo 'All tests completed!'
All tests completed!
++ jobs -pr
+ kill 498809
```

cc @heheda12345 @njhill @robertgshaw2-redhat

PR #30692, opened 2025-12-15
OffloadingConnector: Support kernel_block_size != block_size

orozery:
This PR enables the offloading connector to work for cases where the kernel block size is different than vLLM's logical block size.

To efficiently copy blocks when the logical block size is greater
than the block size, we add an additional parameter to the swap_blocks
function that determines the size per each contiguous copy.

This fix addresses the same issue that was fixed in the NixlConnector in #28677.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit d8e386b4561e2c7af91a7d5de588874e400c49d9. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Enables offloading when `kernel_block_size` differs from the logical block size by making block copies explicitly size-driven.
> 
> - Add `int block_size_in_bytes` to `swap_blocks` (C++/CUDA), remove internal stride-based size; update op schema, registration, and Python wrapper
> - Refactor CPU↔GPU offloading handler to normalize block-size factors, compute per-tensor `block_size_in_bytes`, and call new `swap_blocks`; handle split K/V tensors via unbinding
> - Revise kv-offload tests to parameterize kernel vs logical block sizes and verify mappings at kernel-block granularity
> - Update attention cache tests (`test_swap_blocks`, MLA variants) to pass `block_size_in_bytes`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d8e386b4561e2c7af91a7d5de588874e400c49d9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


vllm/v1/kv_offload/worker/cpu_gpu.py
```diff
@@ -96,6 +96,7 @@ def __init__(
         self.src_block_size_factor: int = src_block_size_factor
         self.dst_block_size_factor: int = dst_block_size_factor
         self.priority = priority
+        self.block_size_in_bytes = src_tensors[0].element_size() * src_tensors[0].stride(0) * src_block_size_factor
```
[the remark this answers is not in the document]
orozery: I actually thinks my logic is correct.
@NickLucche @zhiyuan1i your thoughts?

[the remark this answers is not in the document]
orozery: Indeed there was a bug. The unit test caught it.
Should be fixed now.

vllm/_custom_ops.py
```diff
@@ -2355,9 +2355,12 @@ def copy_blocks_mla(kv_caches: list[torch.Tensor], block_mapping: torch.Tensor)
 
 
 def swap_blocks(
-    src: torch.Tensor, dst: torch.Tensor, block_mapping: torch.Tensor
+    src: torch.Tensor,
+    dst: torch.Tensor,
+    block_size_in_bytes: int,
+    block_mapping: torch.Tensor,
```
ApostaC, line 2461: QQ: Was this function called by other places in the code?

ApostaC (commented):
LGTM overall!

vllm/_custom_ops.py
orozery: It was called in v0, but as you know v0 is gone.
As far as I could check, it is currently only used by the OffloadingConnector.

NickLucche: would other OOT offloading connectors still use it? @sdavidbd

tests/v1/kv_offload/test_cpu_gpu.py
```diff
@@ -24,7 +24,8 @@
 
 NUM_GPU_BLOCKS = [64]
 NUM_CPU_BLOCKS = [256]
-GPU_BLOCK_SIZES = [16]
+KERNEL_BLOCK_SIZES = [16]
+GPU_BLOCK_SIZES = [16, 32]
```
NickLucche: nit: I prefer
```suggestion
LOGICAL_BLOCK_SIZES = [16, 32]
```
since this is what the scheduler+internals use. The actual gpu tensor only has one physical shape.

NickLucche (commented):
Overall this looks good, but admittedly due to my poor familiarity with `cache_kernels.cu` I don't understand why `block_size_in_bytes` has been externalized, despite knowing what we're trying to achieve here.
Perhaps a few comments or PR description would go a long way in helping out the next reviewer/dev :)

orozery:
> Perhaps a few comments or PR description would go a long way in helping out the next reviewer/dev :)

I've added a comment to the commit + PR. Let me know if it's enough :)
Also did some refactoring to simplify the code (get rid of `kv_dim_before_num_blocks`).

csrc/cache_kernels.cu
```diff
@@ -49,10 +50,6 @@ void swap_blocks(torch::Tensor& src, torch::Tensor& dst,
   char* src_ptr = static_cast<char*>(src.data_ptr());
   char* dst_ptr = static_cast<char*>(dst.data_ptr());
 
-  // We use the stride instead of numel in case the cache is padded for memory
-  // alignment reasons, we assume the blocks data (inclusive of any padding)
-  // is contiguous in memory
-  const int64_t block_size_in_bytes = src.element_size() * src.stride(0);
```
NickLucche, line 55: just curious, was this assuming kv cache shape was [num_blocks, 2, N, H, D] , specifically with first dim being num_block?

NickLucche (commented):
@orozery  As far as I understand you externalized `block_size_in_bytes` in order to simulate logical->physical block expansion using the same blocks.
I think we should add a short docstring to the swap method describing what values `block_size_in_bytes` is expected to take in different situations.

What I was expecting to find was a simpler approach:
- expand logical to physical blocks
- copy this new block table, using previous `src.element_size() * src.stride(0)` which is still (afaik) identifying correctly a physical block, given it's calculated based on the actual kv tensor.

I assume this second approach would've been suboptimal as the region are contiguous, so the read can be coalesced by using a bigger `block_size_in_bytes * physical_logical_ratio`.

csrc/cache_kernels.cu
orozery: It was just assuming `[num_blocks, ...]`

orozery:
> @orozery As far as I understand you externalized `block_size_in_bytes` in order to simulate logical->physical block expansion using the same blocks.

Correct.

> I think we should add a short docstring to the swap method describing what values `block_size_in_bytes` is expected to take in different situations.

Done.

> I assume this second approach would've been suboptimal as the region are contiguous, so the read can be coalesced by using a bigger `block_size_in_bytes * physical_logical_ratio`.

Correct :)

NickLucche (approved):
Thanks for confirming @orozery , I think the discussion in this PR can provide useful context for those interested in these changes. 
 Let's get this merged

PR #30500, opened 2025-12-11
feat(gguf): Extract HF config from GGUF metadata for repos without config.json

kitaekatt:
## Summary

This PR enables vLLM to load GGUF models from repositories that don't include `config.json` (e.g., bartowski repos) by extracting the configuration values directly from GGUF metadata.

### Changes:
- Add `GGUF_ARCH_TO_HF_MODEL_TYPE` mapping for architecture name translation (including MoE variants like `qwen3moe` → `qwen3_moe`)
- Add `extract_hf_config_from_gguf()` function that reads GGUF metadata and constructs a HuggingFace-compatible config dictionary
- Add `GGUFConfigParser` class that uses the extraction function
- Register "gguf" format in the config parser system
- Update auto-detection to use GGUF parser for local GGUF files without `config.json`

### Extracted metadata fields:
- Architecture (model_type) with proper normalization for MoE variants
- hidden_size, intermediate_size, num_hidden_layers
- num_attention_heads, num_key_value_heads
- max_position_embeddings, rope_theta, rms_norm_eps
- sliding_window, vocab_size
- bos_token_id, eos_token_id
- attn_logit_softcapping, final_logit_softcapping (for Gemma2)

### Model type mapping
GGUF metadata uses different naming conventions than HuggingFace (e.g., no underscores in MoE architecture names):
- `qwen3moe` → `qwen3_moe`
- `qwen2moe` → `qwen2_moe`
- `gemma3` → `gemma3`

## Test plan
- [ ] Test with bartowski Phi-3.5-mini-instruct-GGUF
- [ ] Test with MoE GGUF models (Qwen3-MoE)
- [ ] Test with Gemma3 GGUF models
- [ ] Verify existing GGUF models with config.json continue to work

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #30702, opened 2025-12-15
[Bugfix] Handle missing config.json in speculator probe for GGUF models

kitaekatt:
## Summary

GGUF models from repositories without `config.json` (e.g., bartowski repos) fail during engine initialization because `maybe_override_with_speculators()` calls `PretrainedConfig.get_config_dict()` before GGUF-specific handling runs.

This is a more targeted fix for the issue identified in #30500, following @Isotr0py's feedback that Transformers already handles GGUF config extraction via `PretrainedConfig.get_config_dict()`.

## Root Cause

The speculator probe in `maybe_override_with_speculators()` always attempts to load `config.json` to check for speculators config. For GGUF repos that only contain `.gguf` files (no `config.json`), this raises an `OSError` before the GGUF metadata path can run.

## Changes

- Wrap the `get_config_dict()` call in a try-except for `OSError`
- For GGUF models (where `gguf_model_repo is not None`), return early since speculators config is defined in `config.json` and cannot exist for these models
- Re-raise the error for non-GGUF models to preserve existing behavior

## Test Plan

- [ ] Test with bartowski Phi-3.5-mini-instruct-GGUF (no config.json)
- [ ] Test with GGUF models that have config.json (should still work)
- [ ] Test with non-GGUF models (behavior unchanged)

## Related

- Supersedes #30500 (which attempted to implement a duplicate GGUF config parser)
- cc @Isotr0py - This is the targeted fix you suggested


kitaekatt:
## Validation Results

| vLLM | transformers | Cherry-picked PRs | HumanEval | IFEval |
|------|-------------|-------------------|-----------|--------|
| HEAD | 5.x | #30410, #30411, #30412, #30413, #30424, #30434, #30699, #30702, #31464, #33846 | gem2-2b-gguf (42.1%), gemma3-1b (26.8%) | gem2-2b-gguf (65.6%) |
| HEAD | 4.x | #30410, #30411, #30412, #30413, #30424, #30434, #30699, #30702, #31464, #33846 | q3-moe-gguf (83.5%) | q3-moe-gguf (85.4%) |

Tested on RTX 5090 (Blackwell, SM 120) with all listed PRs cherry-picked together; models listed under each benchmark passed that benchmark in the given environment, while the same models crash or fail without these PRs applied.

Converting from draft to open. Without this fix, GGUF models without a standard config.json path crash during speculator probe at startup. Validated on RTX 5090 (Blackwell, SM 120).

kitaekatt:
Hi @Isotr0py @mgoin — this has been sitting without review for a while. Just rebased on latest `upstream/main` (merge commit), so the branch should now be mergeable. Quick summary: speculator probe tries to load config.json, which doesn't exist for GGUF models served by HF blob path. Tightens OSError handling so GGUF loading proceeds. Rebased on latest main.

Would appreciate a look when you have cycles. Thanks!

PR #30561, opened 2025-12-12
feat(serve): add warmup support for consistent first-request performance

TheCodeWrangler:
### Problem
vLLM exhibits slow performance on initial requests due to JIT compilation, memory allocation patterns, CUDA kernel warmup, and other first-run costs. In production environments with health checks and load balancers, this can cause issues when traffic is routed to a freshly started instance before it's fully "warmed up."

### Solution
Add a `--warmup-config` CLI option to `vllm serve` that runs configurable warmup requests **before** the health endpoint reports ready. This ensures consistent performance from the very first production request.

### Features
- **New `--warmup-config` CLI argument** - Points to a JSON configuration file
- **Configurable concurrency** - Run warmup requests in parallel for faster startup
- **Comprehensive endpoint support**:
  - `/v1/chat/completions` - Chat completions (LLMs)
  - `/v1/completions` - Text completions (LLMs)
  - `/v1/embeddings` - Embeddings (embedding models)
  - `/pooling` - Pooling (pooling models)
  - `/classify` - Classification (classification models)
  - `/score`, `/v1/score` - Scoring (cross-encoder models)
  - `/rerank`, `/v1/rerank`, `/v2/rerank` - Re-ranking (cross-encoder models)
- **Health endpoint integration** - Returns 503 during warmup, 200 after completion (works with K8s readiness probes)
- **Traceable warmup requests** - Uses `warmup-{idx}` request IDs for easier debugging
- **Failure tolerance** - Warmup failures are logged but don't block server startup

### Configuration Format

```json
{
  "concurrency": 4,
  "requests": [
    {
      "endpoint": "/v1/chat/completions",
      "payload": {
        "messages": [{"role": "user", "content": "Hello"}],
        "max_tokens": 32
      },
      "count": 5
    },
    {
      "endpoint": "/v1/completions",
      "payload": {
        "prompt": "The quick brown fox",
        "max_tokens": 16
      },
      "count": 3
    }
  ]
}
```

| Field | Required | Default | Description |
|-------|----------|---------|-------------|
| `concurrency` | No | 1 | Maximum number of parallel warmup requests |
| `requests` | Yes | - | List of warmup request configurations |
| `requests[].endpoint` | Yes | - | API endpoint path |
| `requests[].payload` | Yes | - | Request body (model is auto-filled if not specified) |
| `requests[].count` | No | 1 | Number of times to repeat this request |

### Usage

```bash
vllm serve meta-llama/Llama-3.1-8B-Instruct --warmup-config warmup.json
[...]

PR #30706, opened 2025-12-15
fix: add warmup for audio preprocessing

TheCodeWrangler:
# Audio Preprocessing Warmup for Whisper Models

## Purpose

Fixes first-request latency issue for Whisper transcription models where the first request can take significantly longer (10-20x) than subsequent requests.

**Problem**: The first Whisper transcription request experiences substantial latency overhead due to lazy initialization, causing poor user experience and potential timeout issues in production.

**Root Cause**: Investigation revealed two lazy initialization bottlenecks:
1. **Librosa/audio preprocessing**: First call to `librosa.load()` and mel-spectrogram functions triggers FFTW plan initialization, numba JIT compilation, and audio codec loading
2. **Input processor**: First call to `input_processor.process_inputs()` with multimodal audio triggers WhisperProcessor initialization and tensor operation JIT compilation

**Solution**: Add warmup methods that run during server initialization to eliminate first-request overhead:
- `_warmup_audio_preprocessing()`: Warms up librosa functions with dummy audio
- `_warmup_input_processor()`: Warms up multimodal processing pipeline with dummy request

### Setup
```bash
# Start vLLM server with Whisper model
vllm serve openai/whisper-large-v3 --dtype bfloat16

# Run test client that makes 10 consecutive transcription requests
python testing/timed_requests_whisper.py
```

### Test Script
Sends 10 requests with the same 60-second audio file and measures client-side latency for each.

### Example Results

**Test Configuration:**
- Model: `openai/whisper-large-v3`
- Hardware: H100 GPU
- Audio: 60-second MP3 file
- Test: 10 consecutive transcription requests

#### Before Fix
```
Request 0: 15.56s  ← First request (18x slower!)
Request 1-9: ~0.85s each
```

**Server-side breakdown (first request)**:
- Audio preprocessing (librosa): **7.67s** (lazy initialization)
- Input processor (multimodal): **2.53s** (lazy initialization)
- Model execution: ~0.02s (already warmed up)
- Text generation: ~5.3s

#### After Fix
```
Request 0: 1.08s  ← First request (now similar to subsequent!)
Request 1-9: ~0.85s each
```

**Server-side breakdown (first request)**:
- Audio preprocessing: **0.08s** (warmed up during startup)
- Input processor: **0.007s** (warmed up during startup)
- Model execution: ~0.02s
- Text generation: ~0.5s

#### Impact Summary

| Component | Before | After | Improvement |
|-----------|--------|-------|-------------|
| Audio preprocessing | ~7-8s | ~0.04-0.08s | ~100-200x faster |
| Input processor | ~2-3s | ~0.002-0.007s | ~400-1500x faster |
| **First request** | ~15-20s | ~1-1.5s | **~10-20x faster** |
| Subsequent requests | ~0.8-1s | ~0.8-1s | No change |
| **Server startup** | Baseline | +10-12s | One-time cost |

**Note**: Exact timings vary based on model size, hardware, audio duration, and system configuration. The key improvement is eliminating the large first-request overhead caused by lazy initialization.

### Trade-off Analysis
- **Added startup time**: ~10-12s (warmup runs once during initialization)
- **Saved per first-request**: ~10-15s
- **Break-even**: After just 1 first request
- **User experience**: Consistent latency across all requests

### Changes Made

**File**: `vllm/entrypoints/openai/speech_to_text.py` (+95 lines)

1. **`_warmup_audio_preprocessing()`**
   - Creates 1-second dummy audio (silence at target sample rate)
   - Calls `librosa.get_duration()` to initialize librosa
   - Calls `librosa.feature.melspectrogram()` with Whisper-specific parameters
   - Initializes FFTW plans, numba JIT, audio codecs
   - Non-fatal: logs warning if warmup fails

2. **`_warmup_input_processor()`**
   - Creates dummy audio and generates proper prompt using `model_cls.get_generation_prompt()`
   - Creates minimal `SamplingParams`
   - Calls `input_processor.process_inputs()` to process dummy request
   - Triggers WhisperProcessor initialization, multimodal hashing, tensor operations
   - Non-fatal: logs warning if warmup fails

3. **Called in `__init__()`**
   - Both warmups run after audio config setup
   - Executed before server starts accepting requests
   - Logs clearly indicate warmup progress

### Why This Approach

- **Targeted**: Only warms up components that showed first-call overhead
- **Safe**: Error handling ensures server starts even if warmup fails
- **Observable**: Clear logging shows warmup completion and timing
- **Efficient**: Uses minimal resources (1-second dummy audio)
- **Complete**: Warms up entire pipeline end-to-end

### Alternate Approaches Considered

1. **Lazy warmup on first request**: Rejected - still causes bad UX for first user
2. **Background warmup**: Rejected - adds complexity, potential race conditions
3. **Only librosa warmup**: Insufficient - still left 2.5s input processor delay
4. **Disable JIT**: Rejected - would hurt performance on all requests

## Documentation Update

None required - this is an internal optimization with no API changes.

## Release Notes

**Performance Improvement**: Whisper transcription models now warm up audio preprocessing and multimodal input processing during server initialization, eliminating 10+ seconds of first-request latency. First requests are now 14x faster (15.5s → 1.08s), providing consistent latency for all users.

## Related Issues

This addresses first-request latency for audio/speech-to-text models, particularly Whisper. The issue manifests as:
- First transcription request timing out or taking 15+ seconds
- Poor user experience for cold-start scenarios
- Inconsistent latency patterns

## Testing Checklist

- [x] Tested with `openai/whisper-large-v3` model
- [x] Verified warmup completes successfully during server startup
- [x] Confirmed first request latency reduced from 15.5s to 1.08s
- [x] Verified subsequent requests unchanged at ~0.85s
- [x] Tested with 60-second audio files
- [x] Verified warmup logs appear during initialization
- [x] Confirmed non-fatal error handling (server starts even if warmup fails)
- [x] No functional changes to transcription accuracy or behavior
- [x] No changes to API surface or response format

## Additional Notes

- First request may still be slightly slower than subsequent requests (~10-30% overhead) due to first HTTP connection, first batch through scheduler, and first encoder hash computation - this is acceptable and unavoidable without significantly more complex warmup
- Warmup is automatic and requires no configuration changes
- Applies to all speech-to-text models using the OpenAI-compatible transcription API (Whisper, Voxtral, etc.)
- Server startup time increase is a one-time cost that provides immediate value on first user request
- Warmup uses minimal resources (1 second of silence) and completes in the background during initialization


TheCodeWrangler:
@LucasWilkinson @robertgshaw2-redhat

Related to prior warmup PR: https://github.com/vllm-project/vllm/pull/30561

But targeted to audio transcription models.

DarkLight1337 (commented):
We should avoid calling these methods unless the model actually supports it

vllm/entrypoints/openai/speech_to_text.py
```diff
@@ -112,6 +112,117 @@ def __init__(
...
+
+    def _warmup_audio_preprocessing(self) -> None:
+        """Warm up audio processing libraries to avoid first-request latency.
+
+        The first call to librosa functions (load, get_duration, mel-spectrogram)
+        triggers JIT compilation and library initialization which can take ~7s.
+        This method warms up these operations during server initialization.
+        """
+        try:
+            warmup_start = time.perf_counter()
```
DarkLight1337, line 136: I was thinking more along the lines of skipping the whole warmup if the model is not multimodal and supports audio inputs, especially considering `librosa` is an optional dependency and we don't want to pollute the startup log with unnecessary warnings.

DarkLight1337: Even if `librosa` is installed, for models that don't use this endpoint, this would also unnecessarily add to the startup time which is quite long already.

vllm/entrypoints/openai/speech_to_text.py
```diff
@@ -112,6 +112,133 @@ def __init__(
...
+        The first call to librosa functions (load, get_duration, mel-spectrogram)
+        triggers JIT compilation and library initialization which can take ~7s.
+        This method warms up these operations during server initialization.
+        """
+        # Skip warmup if librosa is not installed (optional dependency)
+        if isinstance(librosa, PlaceholderModule):
+            return
+
+        # Skip warmup if model doesn't support multimodal/audio inputs
+        if not self.model_config.is_multimodal_model:
```
DarkLight1337: Use `supports_transcription`

DarkLight1337 (approved):
Thanks!

vllm/entrypoints/openai/speech_to_text.py
```diff
@@ -233,7 +233,7 @@ def _warmup_input_processor(self) -> None:
             logger.info("Input processor warmup completed in %.2fs", warmup_elapsed)
         except Exception as e:
             # Don't fail initialization if warmup fails - log warning and continue
-            logger.warning(
+            logger.exception(
```
DarkLight1337, line 235: In that case there is no need to log `e` since `exc_info=True` automatically

PR #29901, opened 2025-12-02
[Kernel][Quantization][MoE] add marlin kernel support for turing (sm75)

jinzhen-lin:
This PR add marlin kernel support for turing (sm75) (e.g. 2080ti / T4).

- The Turing architecture lacks the `cp.async` instruction, but we can still use synchronous instructions to read from global memory and write to shared memory.
- The Turing architecture does not have the `m16n8k16` MMA instruction, but it does have the `m16n8k8` instruction. We only need to stack the instruction twice to achieve the same effect as `m16n8k16`.
- The throughput of the MMA instruction on the Turing architecture is only half when using an FP32 accumulator compared to an FP16 accumulator (In fact, this is not only for Turing but also for architectures like Ada Lovelace). I also believe that if the final result will not overflow, the probability of the intermediate results overflowing is also very low. Therefore, I have temporarily changed the implementation on Turing to use an FP16 accumulator, while leaving the other architectures unchanged.
- Supported Weights: GPTQ, AWQ, FP8, NVFP4 (MXFP4 is not supported since it requires BF16 activation)
- Supported Activation: FP16, INT8 (only for GPTQ and AWQ)

## Kernel Benchmark

2080ti + Dense Marlin + GPTQ Channelwise

```
   n     k     m    torch-fp16    marlin-w4a16    marlin-w4a8-int8
----  ----  ----  ------------  --------------  ------------------
1024  2048     1    0.0116937       0.00575462          0.00530092
1024  2048     2    0.0141614       0.00586464          0.00534524
1024  2048     4    0.0132746       0.00596898          0.00544628
1024  2048     8    0.0133064       0.00612971          0.0055609
1024  2048    16    0.013491        0.00690344          0.00596592
1024  2048    32    0.0136423       0.00926577          0.00758568
1024  2048    64    0.0158824       0.0149638           0.0111064
1024  2048   128    0.023759        0.0209582           0.0143538
1024  2048   256    0.0349035       0.0301925           0.0181232
1024  2048   512    0.0500498       0.044179            0.0295197
1024  2048  1024    0.0899547       0.077805            0.0546717
1024  2048  2048    0.157824        0.14364             0.099661
1024  2048  4096    0.313458        0.269902            0.183839
2048  1024     1    0.00999344      0.00581668          0.00541013
2048  1024     2    0.00840535      0.00590156          0.00544462
2048  1024     4    0.00863134      0.00599312          0.0055447
2048  1024     8    0.0113632       0.00618215          0.00561357
[...]

PR #30719, opened 2025-12-15
[ROCm][GPTQ][Bugfix] Fix GPTQ GEMM kernel output zeroing race condition

AndreasKaratzas:
## Summary

Fixes a race condition in the GPTQ GEMM kernels that caused incorrect results when `input_size > BLOCK_KN_SIZE (128)`.

## Problem

The GPTQ GEMM kernels use multiple thread blocks along the k-dimension (`gridDim.z > 1`) that accumulate partial results via `atomicAdd`. The output tensor was being zeroed inside the kernel by `blockIdx.z == 0`:
```cpp
if (blockIdx.z == 0) {
  for (int m = 0; m < m_count; m++)
    *((uint64_t*)c_.item_ptr(offset_m + m, n)) = 0;
}
__syncthreads();  // Only syncs within block, not across blocks!
```

Since `__syncthreads()` only synchronizes threads within the same block, not across different blocks, this creates a race condition where:
- Blocks with `z > 0` may `atomicAdd` their results before block `z=0` finishes zeroing
- Block `z=0` may overwrite results that other blocks have already added

This caused numerical errors up to 45x the expected values, particularly when:
- `input_size > 128` (triggers multiple k-blocks)
- Errors concentrated in specific rows at m-block boundaries

## Solution

1. Pre-zero the output tensor using `torch::zeros()` instead of `torch::empty()`
2. Remove the in-kernel zeroing logic from all GPTQ GEMM kernel variants


AndreasKaratzas:
> maybe we should just deprecate this kernel...

For now, I think we can merge this PR though, since it resolves the GPTQ test bug on ROCm.

mgoin (approved):
This makes sense and simplifies the kernel, LGTM! I do think we were planning to deprecate this kernel now that @jinzhen-lin added SM75 support for Marlin https://github.com/vllm-project/vllm/pull/29901, but if ROCm needs this kernel we can keep it for now. I would highly recommend the ROCm team investigating if they can reuse the Marlin kernels

AndreasKaratzas:
cc @tjtanaa Can we merge this one too? The failing tests are known to be problematic.

PR #30724, opened 2025-12-15
Fix edge case Mistral tool parser

joa-stdn:
New version of https://github.com/vllm-project/vllm/pull/30588
- Fixing an edge case in Mistral tool parser without streaming where content before [TOOL_CALLS] contains {
- Returning partial json instead of [TOOL_CALLS] when json is wrong, like we do on API (otherwise messes up follow up turn)


vllm/tool_parsers/mistral_tool_parser.py
```diff
@@ -131,78 +131,102 @@ def extract_tool_calls(
...
-        except Exception:
-            logger.exception("Error in extracting tool call from response.")
-            # return information to just treat the tool call as regular JSON
-            return ExtractedToolCallInformation(
-                tools_called=False, tool_calls=[], content=tool_content
-            )
+        return ExtractedToolCallInformation(
+            tools_called=True,
+            tool_calls=mistral_tool_calls,
+            content=content if len(content) > 0 else None,
```
patrickvonplaten, line 231:
```suggestion
            content=content
```
(nit) isn't an empty string also ok?

vllm/tool_parsers/mistral_tool_parser.py
```diff
@@ -131,78 +131,102 @@ def extract_tool_calls(
...
-                        # function call args are JSON but as a string
-                        arguments=json.dumps(
-                            raw_function_call["arguments"], ensure_ascii=False
+                try:
+                    raw_tool_call = self.tool_call_regex.findall(
+                        stringified_tool_calls
+                    )[0]
+                    tool_calls = json.loads(raw_tool_call)
+                except (IndexError, json.JSONDecodeError):
+                    logger.exception("Error in extracting tool call from response.")
```
patrickvonplaten:
```suggestion
                except (IndexError, json.JSONDecodeError) as e:
                    logger.exception("Error in extracting tool call from response: {e}")
```

vllm/tool_parsers/mistral_tool_parser.py
```diff
@@ -131,78 +131,102 @@ def extract_tool_calls(
...
+                    raw_tool_call[end_name:],
+                )
+
+                tool_calls.append({"name": tool_name, "arguments": args})
 
-        try:
+        # < v11: content[BOT] [{tool_call1},{tool_call2}]
+        else:
+            if len(raw_tool_calls) != 1:
+                raise ValueError("Only one BOT token should have been outputted.")
```
patrickvonplaten:
```suggestion
                raise ValueError(f"Only one BOT token should have been outputted, but got {model_output}")
```

patrickvonplaten (approved):
Left some smaller comments

vllm/tool_parsers/mistral_tool_parser.py
juliendenize: it's ok for mistral-common but it matches what other parsers do. Let's leave it that way.

DarkLight1337 (approved):

PR #30729, opened 2025-12-16
[Perf] enable flashinfer rotary_embedding custom ops in DeepSeek rotary

jiahanc:
## Purpose
- Enable flashinfer rotary_embedding custom ops in DeepSeek
**To use the custom op, add `--compilation_config.custom_ops+=+rotary_embedding` in vllm config**
## Test Plan
```
VLLM_USE_FLASHINFER_MOE_FP4=1 python3 -m vllm.entrypoints.openai.api_server --model nvidia/DeepSeek-R1-0528-FP4-v2 --tokenizer nvidia/DeepSeek-R1-0528-FP4-v2 --dtype auto --kv-cache-dtype fp8 --tensor-parallel-size 1 --pipeline-parallel-size 1 --data-parallel-size 4 --enable-expert-parallel --swap-space 16 --max-num-seqs 1024 --trust-remote-code --max-model-len 4096 --gpu-memory-utilization 0.8 --max-num-batched-tokens 16384 --no-enable-prefix-caching --compilation-config '{"cudagraph_mode":"NONE"}' --compilation_config.pass_config.enable_fi_allreduce_fusion true --compilation_config.pass_config.enable_attn_fusion true --compilation_config.pass_config.enable_noop true --compilation_config.custom_ops+=+quant_fp8,+rms_norm,+rotary_embedding --compilation_config.max_cudagraph_capture_size 2048
```
```
lm_eval --model local-completions --tasks gsm8k --model_args model=nvidia/DeepSeek-R1-0528-FP4-v2,base_url=http://0.0.0.0:8000/v1/completions,max_retries=3,tokenized_requests=False,timeout=1200,max_gen_toks=2048,max_length=8192 --batch_size 2048 --trust_remote_code --limit 0.5
```
## Test Result
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9636|±  |0.0073|
|     |       |strict-match    |     5|exact_match|↑  |0.9636|±  |0.0073|
```
---
<details>

</details>

### DIFF
--- a/vllm/model_executor/layers/rotary_embedding/deepseek_scaling_rope.py
+++ b/vllm/model_executor/layers/rotary_embedding/deepseek_scaling_rope.py
@@ -6,6 +6,7 @@ import math
 import torch
 
 from vllm.platforms import current_platform
+from vllm.utils.flashinfer import has_flashinfer
 
 from .base import RotaryEmbeddingBase
 from .common import (
@@ -55,6 +56,13 @@ class DeepseekScalingRotaryEmbedding(RotaryEmbeddingBase):
             yarn_get_mscale(self.scaling_factor, float(mscale))
             / yarn_get_mscale(self.scaling_factor, float(mscale_all_dim))
             * attn_factor
+        )
+        self.use_flashinfer = (
+            self.enabled()
+            and dtype in (torch.float16, torch.bfloat16)
+            and current_platform.is_cuda()
+            and has_flashinfer()
+            and head_size in [64, 128, 256, 512]
         )
         super().__init__(
             head_size, rotary_dim, max_position_embeddings, base, is_neox_style, dtype
@@ -162,4 +170,15 @@ class DeepseekScalingRotaryEmbedding(RotaryEmbeddingBase):
         key: torch.Tensor | None = None,
         offsets: torch.Tensor | None = None,
     ) -> tuple[torch.Tensor, torch.Tensor | None]:
-        return self.forward_native(positions, query, key, offsets)
+        if self.use_flashinfer:
+            torch.ops.vllm.flashinfer_rotary_embedding(
+                positions,
+                query,
+                key,
+                self.head_size,
+                self.cos_sin_cache,
+                self.is_neox_style,
+            )
+            return query, key
+        else:
+            return self.forward_native(positions, query, key, offsets)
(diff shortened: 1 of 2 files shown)


vllm/model_executor/layers/rotary_embedding/deepseek_scaling_rope.py
```diff
@@ -114,6 +114,18 @@ def forward_native(
     ) -> tuple[torch.Tensor, torch.Tensor | None]:
         """PyTorch-native implementation equivalent to forward()."""
         assert key is not None
+
+        if self.use_flashinfer:
```
pavanimajety: we should enable this path for `forward_cuda` instead.

jiahanc: Yeah I am ok with it. The current `forward_cuda` just calls the `forward_native`

vllm/model_executor/layers/rotary_embedding/deepseek_scaling_rope.py
```diff
@@ -162,4 +163,15 @@ def forward_cuda(
         key: torch.Tensor | None = None,
         offsets: torch.Tensor | None = None,
     ) -> tuple[torch.Tensor, torch.Tensor | None]:
-        return self.forward_native(positions, query, key, offsets)
+        if self.use_flashinfer:
+            torch.ops.vllm.flashinfer_rotary_embedding(
+                positions,
```
pavanimajety: do we need to add positions to offsets? We are ignoring that argument when using Flashinfer.

jiahanc: the offset is always none because in `deepseek_v2.py`, offset is not passed https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/deepseek_v2.py#L180

pavanimajety:
@jiahanc Could you also share the baseline eval scores?

vllm/model_executor/layers/rotary_embedding/deepseek_scaling_rope.py
pavanimajety: I think we still need to add offsets IMO to maintain consistency with how it is used in forward_native - https://github.com/jiahanc/vllm/blob/e5925488e6c6ee929aef70681757ef507793425a/vllm/model_executor/layers/rotary_embedding/deepseek_scaling_rope.py#L132-L134

If the same deepseek style yarn based rope is used for a different model which uses offsets, we will run into errors.

jiahanc: yeah makse sense. Updated

pavanimajety:
@mgoin @alexm-redhat Could you take a look?

mgoin (approved):
LGTM, thank you

PR #30745, opened 2025-12-16
[BugFix]Reclaim resources to prevent memory leaks when use LMCacheMPConnector

wz1qqx:
## Purpose
This is from Novita.AI Team

Do clean up to prevent memory leak
## Test Result

---
<details>

</details>


vllm/distributed/kv_transfer/kv_connector/v1/lmcache_mp_connector.py
```diff
@@ -748,6 +748,11 @@ def request_finished(
             Optional KVTransferParams to be included in the request outputs
             returned by the engine.
         """
+        # Clean up request tracker to prevent memory leak
+        # Note: We clean up here because the request is finished
+        # The store operation will use the tracker info that was already
+        # captured in the metadata during previous build_connector_meta calls
```
ApostaC: nit: we don't need to have this note here.

ApostaC (commented):
Otherwise LGTM!

vllm/distributed/kv_transfer/kv_connector/v1/lmcache_integration/multi_process_adapter.py
```diff
@@ -147,6 +147,14 @@ def num_blocks_per_chunk(self) -> int:
...
         return self.blocks_in_chunk
 
+    def cleanup_request(self, request_id: str) -> None:
+        """
+        Clean up lookup future for a finished request to prevent memory leak.
+        Args:
+            request_id: The ID of the finished request.
+        """
+        self.lookup_futures.pop(request_id, None)
+
```
ApostaC: Can we rename this function to `cleanup_lookup_result`?

ApostaC: Actually I realized a problem, the look up result should be cleared once we are sure that the cooresponding `retrieve` request will be issued.
This is because `lookup` will "lock" the object and the corresponding `retrieve` will unlock them upon finished. 
Therefore, if we keep the lookup result longer, the blocks may never be retrieved again, especially when there are preempted requests.

wz1qqx: Yes, I think the timing of releasing lookup_futures  can be considered together with the subsequent pr for preemption.

vllm/distributed/kv_transfer/kv_connector/v1/lmcache_mp_connector.py
```diff
@@ -893,3 +895,17 @@ def _get_or_create_request_tracker(
...
+        This should be called when a request is finished to prevent memory leak.
+        """
+        # Clean up request tracker
+        if self.request_trackers.pop(request_id, None):
+            logger.debug(
+                "[KVConnector] Cleaned up request_tracker for request %s",
+                request_id,
+            )
+        # Clean up lookup future in scheduler adapter
+        self.scheduler_adapter.cleanup_lookup_result(request_id)
```
ApostaC: Can you move this function call to `update_state_after_alloc`? This should be called when the request's state changes from `PREFETCHING` to other states

ApostaC (commented):
See the comment please. Otherwise LGTM

vllm/distributed/kv_transfer/kv_connector/v1/lmcache_mp_connector.py
wz1qqx: i think, maybe move to check_lookup_result better?

ApostaC (approved):
LGTM!

PR #30747, opened 2025-12-16
[Fix] uniform decode batch check

Jialin:
## Purpose
Correctly tag batch with the right uniform_decode value. Before #28579, unpadded scheduled tokens are used to check uniform_decode.

In this PR, we simply revert it back to the original behaviour and move the logic to a dedicated function so we could add unit test easily.

*Before*
Screenshot 2025-12-15 at 10 30 33 PM

*Current*
Screenshot 2025-12-15 at 10 32 00 PM

## Test Plan & Test Result
```
pytest -s -v tests/v1/worker/test_gpu_model_runner.py
```

---
<details>

</details>

### DIFF
--- a/vllm/v1/worker/gpu_model_runner.py
+++ b/vllm/v1/worker/gpu_model_runner.py
@@ -2798,14 +2798,6 @@ class GPUModelRunner(
         CUDAGraphStat | None,
     ]:
         num_tokens_padded = self._pad_for_sequence_parallelism(num_tokens)
-        uniform_decode = (
-            (
-                (max_num_scheduled_tokens == self.uniform_decode_query_len)
-                and (num_tokens == max_num_scheduled_tokens * num_reqs)
-            )
-            if force_uniform_decode is None
-            else force_uniform_decode
-        )
         # Encoder-decoder models only support CG for decoder_step > 0 (no enc_output
         # is present). Also, chunked-prefill is disabled, so batch are uniform.
         has_encoder_output = (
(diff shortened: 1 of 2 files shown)


vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2797,6 +2810,14 @@ def _determine_batch_execution_and_padding(
             else force_has_lora
         )
 
+        uniform_decode = self._is_uniform_decode(
```
LucasWilkinson: why move this? we should do it before any padding we are not going use the `sequence_parallelism` padded token count; so its clear the padding order:

1) detect uniform decode
2) pad for sequence_parallelism
3) pad for CGs
4) pad for DP

LucasWilkinson (changes requested):
This was done to make sure `sequence_parallelism` was accounted for; the idea was once you padded for sequence parallelism you are sometimes no-longer uniform decode, i.e. `uniform_decode_query_len = 3` then if you bad to 8 for `sequence_parallelism` your batch is effectively `[3, 3, 2 (padding)]`. But because of https://github.com/vllm-project/vllm/blob/66c3537e5df215d8095d7042b8e7abd51260393f/vllm/config/compilation.py#L1094 I think this can be fine we should just make the order clearly 

```
1) detect uniform decode
2) pad for sequence_parallelism
3) pad for CGs
4) pad for DP
5) assert sequence parallelism padding requirements is still respected < please add
```

so its obvious the uniform decode detection is happening on the unpadded batch

Were you seeing issues when enabling `sequence_parallelism`?

LucasWilkinson (approved):
LGTM; thanks for digging into this!

PR #30753, opened 2025-12-16
[Core][KVConnector] Propagate block hashes in SchedulerOutput

qierli:
### What it Does
Propagate the block hashes within SchedulerOutput, on each schedule() finish.
- prompt_block_hashes for new request, and the 1st iteration of Chunked Prefill
- all_block_hashes for resumed preemptions
- running_block_hashes for >1 iterations of Chunked Prefill

Separate data types to kv_cache_types.py to avoid circular reference.

### It enables
KVConnector, as external prefix caching, to **persist NO requests** across multiple schedule() cycles. It can simply get the necessary hashes of KVs to offload from the build_connector_meta(SchedulerOutput).

One dangerous design of current **offloading_connector** is that it persists requests (and theirs hashes) on get_num_new_matched_tokens(), and then try to maintain & clean up across multiple schedule(). This causes dependencies on the scheduling, and is **reported to have multiple data correctness issues** on edge cases like Chunked Prefill, Preemption->Resumption etc.

#### Take Chunked Prefill as an example:
Scheduler only calls get_num_new_matched_tokens() on the 1st iteration, yet call build_connector_meta(SchedulerOutput) on every iteration. This requires complicated idx computations by the KVConnector, if without this PR.

### Overhead
Minimum

## Test Plan
Propagation UTs added in test_scheduler.py

## Test Result
Passed

---
<details>

</details>


vllm/v1/core/sched/scheduler.py
```diff
@@ -848,8 +852,19 @@ def _make_cached_request_data(
...
                 resumed_req_ids.add(req_id)
-            if not scheduled_in_prev_step:
+            if scheduled_in_prev_step:
+                num_tokens = num_scheduled_tokens[req_id]
+                start_token = req.num_computed_tokens
+                end_token = start_token + num_tokens
+                start_block = start_token // self.block_size
+                end_block = (end_token + self.block_size - 1) // self.block_size
+                running_token_ids.append(req.all_token_ids[start_token:end_token])
+                running_block_hashes.append(req.block_hashes[start_block:end_block])
```
zhuohan123, line 867: will this cause we send repetitive block hashes?

orozery:
I'm trying to understand the big picture here.

I like your motivation for making KV connectors stateless.
IMO The KV Connector API as it is today is low-level and complex.
It is indeed hard (or impossible? :) ) to write a bug-free connector.

However, I don't see how this specific change you're suggesting here (propagating block hashes) solves it.
Or is it part of a larger set of changes you're planning?
It would be easier to understand if you can give a high-level flow of how a new connector could employ this.
If you already have a POC code this would even be better.

cc @njhill

QierLi:
> It is indeed hard (or impossible? :) ) to write a bug-free connector.

Totally agree - we've also encountered misses of caching, or even data corruptions during our implementation.

We just happened to adopt a different initial design pattern: 
- All transient metadata in connector will be cleared per each schedule() cycle
- To decide KVs to offload, we mostly depend on **SchedulerOutput (blocks & hashes for computation on this step)**

This appears to be quite robust, and not depends on the complicated scheduling logics. KVs are always correctly offloaded based on SchedulerOutput, no matter preemption->resumption, chunked prefill or aync scheduling happens.

> IMO The KV Connector API as it is today is low-level and complex.

Yeah I believe core APIs will iterate... For example, we should have a dedicated scheduler->connector dataclass instead of extending this SchedulerOutput shared with model runner. But at this point, this PR with no lazy propagation but low code complexity seems to be a balanced choice.

orozery:
> * To decide KVs to offload, we mostly depend on **SchedulerOutput (blocks & hashes for computation on this step)**

OK I think I'm starting to get it.
Today, block hashes are available via `get_num_new_matched_tokens`, which is only called on a new request or a request resumed from preemption.
To keep tracking blocks throughout the request life-cycle you need to keep a state mapping req_id -> Req.
Once you have this state, each step you can get the new block hashes by getting the req_id's from SchedulerOutput, then using the stateful req_id -> Req mapping.
With the change you're proposing, you won't have to keep this state, at least not for getting the block hashes.

My problem is just with the fact that you add fields to SchedulerOutput as you need them on the scheduler-side connector.
You don't actually need it on the worker-side as well, right?
If so, we can save the unnecessary pass of information scheduler->worker by explicitly propagating the block hashes on each step, either in `build_connector_meta`, or some new scheduler-side connector API.
WDYT?

QierLi:
> If so, we can save the unnecessary pass of information scheduler->worker by explicitly propagating the block hashes on each step, either in `build_connector_meta`, or some new scheduler-side connector API. WDYT?

This is similar with the **lazy propagation** I mentioned, in the sacrifice of code complexity.

Given this PR just increases the proto by a bit with no extra scheduler-worker events or blocking actions, and this SchedulerOutput has already been extended for partial use cases (e.g. async scheduling, v2 model runner), I personally incline to have a **one-off whole refactoring / API iteration** in the future instead.

But I'll defer the decision to @zhuohan123 and @njhill here.

orozery:
> This is similar with the **lazy propagation** I mentioned, in the sacrifice of code complexity.

Can we go get more concrete?
In the current suggested code changes, you add fields to `SchedulerOutput`.
Who will consume those fields? I assume `build_connector_meta` on the KV connector. Right? No other direct consumers, right?
If that so, what difference does it make if it's consumed from `SchedulerOutput` or from an additional param of `build_connector_meta`?

More generally, I think `SchedulerOutput` should contain just fields that are of interest to the worker.

QierLi:
> More generally, I think `SchedulerOutput` should contain just fields that are of interest to the worker.

I understand your proposal here, but at least for now, SchedulerOutput **by design** is the [only param](https://github.com/vllm-project/vllm/blob/ec965569d94c09eb1c85d235319b24d1b795d048/vllm/distributed/kv_transfer/kv_connector/v1/base.py#L464) of `build_connector_meta()` and does contain information to the scheduler side - not worker - connector.

If we want to create a new dataclass or an API, I'll expect a broader discussion to include and achieve multiple optimizations at once, but not just saving this negilible overhead of one PR.

QierLi:
>  this means the running set of requests will occupy the entire 400K token capacity, yielding 400K/(16=block_size)=25K 
> This means you're adding 25K block hashes passed in over IPC at each engine step.

It sounds exaggerated, because obviously only **Prefill's scheduled tokens in this step** will be passed. The number is constrained by {max_num_batched_tokens x max_batched_size}. And you're counting the **total tokens the Paged KV can hold**.

Plus, the added fields in this PR takes a fairly **small ratio** of NewRequestData and CachedRequestData.

Again, I agree the motivation to squeeze every bit of performance, but we also can't deprecate API - `build_connector_meta(SchedulerOutput)` - arbitrarily. So I proposed this solution:
- For this PR, let's make new fields optional and do lazy propagation with `(scheduler) self.connector is not None`, so at least engine without KVConnector won't be impacted.
- In the future, let's correct the design of sharing `SchedulerOutput` for both (1) worker-side model_runner, and (2) scheduler-side KVConnector. We need to have a dedicated context proto for the latter.

WDYT?

PR #30575, opened 2025-12-12
[Bugfix] Pass FA version in `MultiHeadAttention`

MatthewBonanni:
## Purpose
The call to `flash_attn_varlen_func` in `MultiHeadAttention` does not pass `fa_version`. This causes multimodal models to use a mix of FA2 (default) and FA3. This PR fixes this.

## Test Plan
```
vllm serve openai/whisper-large-v3
```
```
curl -X POST http://localhost:8000/v1/audio/transcriptions \
    -H "Content-Type: multipart/form-data" \
    -F "file=@mary_had_lamb.ogg" \
    -F "model=openai/whisper-large-v3"
```

## Test Result
Putting a simple print statement in `vllm.vllm_flash_attn.flash_attn_interface.flash_attn_varlen_func` shows it's being called correctly with FA3 on Hopper.

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x] The test plan, such as providing test command.
- [x] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #30756, opened 2025-12-16
[MM] Pass FA version in ViT Attn

NickLucche:
Same rationale as https://github.com/vllm-project/vllm/pull/30575/ but for non-causal ViT attn.

Test with `pytest -v -s tests/models/multimodal/generation/test_vit_backend_functionality.py`

### DIFF
--- a/vllm/attention/ops/vit_attn_wrappers.py
+++ b/vllm/attention/ops/vit_attn_wrappers.py
@@ -58,6 +58,7 @@ def flash_attn_maxseqlen_wrapper_fake(
     max_seqlen: torch.Tensor,
     batch_size: int,
     is_rocm_aiter: bool,
+    fa_version: int,
 ) -> torch.Tensor:
     return torch.empty_like(q)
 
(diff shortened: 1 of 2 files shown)


DarkLight1337:
`test_vit_backend_functionality.py` was failing on main so let's see if it passes with this PR

vllm/attention/ops/vit_attn_wrappers.py
```diff
@@ -43,6 +44,7 @@ def flash_attn_maxseqlen_wrapper(
         max_seqlen_k=max_seqlen.item(),
         dropout_p=0.0,
         causal=False,
+        fa_version=fa_version,
```
Isotr0py: I remember that aiter's `flash_attn_varlen_func` doesn't accept `fa_version`.

cjackal: Right, that is why I put `fa_version` as an optional kwargs in #29889

NickLucche: Thanks!

Isotr0py (approved):

PR #30761, opened 2025-12-16
[KVConnector]: Enable Cross-layers KV cache layout for MultiConnector

kfirtoledo:
Enable MultiConnector to use the new continuous cross-layer KV cache layout 
described in RFC #27742 and implemented in #27743.
The MultiConnector aggregates cross-layer support across all underlying connectors 
and returns true only if all connectors support it.


NickLucche (commented):
Thanks for doing this @kfirtoledo! 
And yes the property change is fine here @orozery , given this feature was just introduced. 

Would you mind adding a small unit test to keep this feature in check? It looks trivial, but we had similar issues with MC before..
Other than that this is lgtm.

kfirtoledo:
@NickLucche  and @orozery, I added the unit tests, PTAL

vllm/distributed/kv_transfer/kv_connector/v1/multi_connector.py
```diff
@@ -138,6 +138,14 @@ def __init__(
         # Propagated from scheduler to worker side via the connector metadata.
         self._extra_async_saves: dict[str, int] = {}
 
+    @property
+    def prefer_cross_layer_blocks(self) -> bool:
+        if not self._connectors:
+            return False
+        return all(
+            getattr(c, "prefer_cross_layer_blocks", False) for c in self._connectors
+        )
```
orozery:
```suggestion
        return all(c.prefer_cross_layer_blocks for c in self._connectors)
```

vllm/distributed/kv_transfer/kv_connector/v1/base.py
```diff
@@ -152,7 +152,9 @@ class KVConnectorBase_V1(ABC):
             Defaults to False.
     """
 
-    prefer_cross_layer_blocks: ClassVar[bool] = False
+    @property
```
orozery, line 149: Can you move the docstring from the class definition over here?

orozery (changes requested):

vllm/distributed/kv_transfer/kv_connector/v1/multi_connector.py
kfirtoledo: Done

vllm/distributed/kv_transfer/kv_connector/v1/base.py
kfirtoledo: Done

orozery (approved):
lgtm. Thanks!

NickLucche (approved):
Thanks @kfirtoledo !

PR #30765, opened 2025-12-16
[Doc][CPU] Update CPU doc

bigPYJ1151:
## Purpose

Update CPU doc for wheel installation

## Test Result

---
<details>

</details>

### DIFF
--- a/docs/getting_started/installation/cpu.x86.inc.md
+++ b/docs/getting_started/installation/cpu.x86.inc.md
@@ -141,7 +141,7 @@ uv pip install dist/*.whl
     IOMP_PATH=...
 
     # add them to LD_PRELOAD
-    export LD_PRELOAD="$TC_PATH:IOMP_PATH:$LD_PRELOAD"
+    export LD_PRELOAD="$TC_PATH:$IOMP_PATH:$LD_PRELOAD"
     ```
 
 !!! example "Troubleshooting"
--- a/docs/getting_started/installation/cpu.x86.inc.md
+++ b/docs/getting_started/installation/cpu.x86.inc.md
@@ -43,7 +43,7 @@ uv pip install vllm --extra-index-url https://wheels.vllm.ai/${VLLM_VERSION}/cpu
     IOMP_PATH=...
 
     # add them to LD_PRELOAD
-    export LD_PRELOAD="$TC_PATH:IOMP_PATH:$LD_PRELOAD"
+    export LD_PRELOAD="$TC_PATH:$IOMP_PATH:$LD_PRELOAD"
     ```
 
 **Install the latest code**
--- a/docs/getting_started/installation/cpu.x86.inc.md
+++ b/docs/getting_started/installation/cpu.x86.inc.md
@@ -70,7 +70,7 @@ Install recommended compiler. We recommend to use `gcc/g++ >= 12.3.0` as the def
 
 ```bash
 sudo apt-get update -y
-sudo apt-get install -y gcc-12 g++-12 libnuma-dev python3-dev
+sudo apt-get install -y gcc-12 g++-12 libnuma-dev
 sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-12 10 --slave /usr/bin/g++ g++ /usr/bin/g++-12
 ```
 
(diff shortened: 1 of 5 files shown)


docs/getting_started/installation/cpu.x86.inc.md
```diff
@@ -17,7 +17,51 @@ vLLM supports basic model inferencing and serving on x86 CPU platform, with data
...
+# use uv
+uv pip install vllm --extra-index-url https://wheels.vllm.ai/${VLLM_VERSION}/cpu --index-strategy first-index --torch-backend cpu
+```
+??? console "pip"
+    ```bash
+    # use pip
+    pip install vllm==${VLLM_VERSION}+cpu --extra-index-url https://wheels.vllm.ai/${VLLM_VERSION}/cpu --extra-index-url https://download.pytorch.org/whl/cpu
+    ```
+!!! warning "set `LD_PRELOAD`"
+    Before use vLLM CPU installed via wheels, make sure TCMalloc and Intel OpenMP are installed and added to `LD_PRELOAD`:
```
fadara01, line 34: out of curiosity, these are just needed for optimal performance, but not absolutely required, is this correct?
if that's the case we should say that it's "recommended for optimal perf" rather than give the user the impressions that it's absolutely necessary.

docs/getting_started/installation/cpu.x86.inc.md
```diff
@@ -17,7 +17,51 @@ vLLM supports basic model inferencing and serving on x86 CPU platform, with data
...
 # --8<-- [start:pre-built-wheels]
 
-Currently, there are no pre-built x86 CPU wheels.
+Pre-built vLLM wheels for x86 with AVX512 are available since version 0.13.0. To install release wheels:
+
+```bash
+export VLLM_VERSION=$(curl -s https://api.github.com/repos/vllm-project/vllm/releases/latest | jq -r .tag_name | sed 's/^v//')
+
+# use uv
+uv pip install vllm --extra-index-url https://wheels.vllm.ai/${VLLM_VERSION}/cpu --index-strategy first-index --torch-backend cpu
```
fadara01, line 26: `--index-strategy fiest-index` seems redundant here as it's the [default with uv](https://docs.astral.sh/uv/concepts/indexes/#pinning-a-package-to-an-index:~:text=By%20default%2C%20uv%20will%20stop%20at%20the%20first%20index%20on%20which%20a%20given%20package%20is%20available%2C%20and%20limit%20resolutions%20to%20those%20present%20on%20that%20first%20index%20(first%2Dindex).)

docs/getting_started/installation/cpu.x86.inc.md
bigPYJ1151: In fact I think ```TCMalloc``` is necessary, without it performance will be half.

docs/getting_started/installation/cpu.x86.inc.md
bigPYJ1151: Yes, just for some environments may change the default to ```unsafe-best-match``` to get same behaviour with ```pip```.

docs/getting_started/installation/cpu.x86.inc.md
fadara01: yeah this is the case for Arm too, but i was hesitant to add it to the docs. I think we should add it there too.

docs/getting_started/installation/cpu.x86.inc.md
fadara01: >Yes, just for some environments may change the default to unsafe-best-match to get same behaviour with pip.

Yeah I assumed it's on the user if they chose to do this.

If you plan to go ahead with this it would be nice to make the same change for Arm to keep CPU installation docs consistent: https://docs.vllm.ai/en/latest/getting_started/installation/cpu/#pre-built-wheels

bigPYJ1151: updated

docs/getting_started/installation/cpu.x86.inc.md
bigPYJ1151: updated

fadara01 (approved):
LGTM, thank you!

docs/getting_started/installation/cpu.x86.inc.md
fadara01: Thank you!

docs/getting_started/installation/cpu.x86.inc.md
fadara01: Thank you

jikunshang (approved):

PR #30773, opened 2025-12-16
[BugFix] Fix initialization bug in openpangu.py

JeffLee1874:
…1 required positional argument: 'indexer_rotary_emb'

## Purpose
Fix initialization error "TypeError: MLAModules.__init__() missing 1 required positional argument: 'indexer_rotary_emb'" in openpangu.py

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/model_executor/models/openpangu.py
+++ b/vllm/model_executor/models/openpangu.py
@@ -372,6 +372,7 @@ class OpenPanguMLAAttention(nn.Module):
             q_b_proj=self.q_b_proj if self.q_lora_rank is not None else None,
             q_proj=self.q_proj if self.q_lora_rank is None else None,
             indexer=None,
+            indexer_rotary_emb=None,
             is_sparse=False,
             topk_indices_buffer=None,
         )


vllm/model_executor/models/openpangu.py
```diff
@@ -372,6 +372,7 @@ def __init__(
             q_b_proj=self.q_b_proj if self.q_lora_rank is not None else None,
             q_proj=self.q_proj if self.q_lora_rank is None else None,
             indexer=None,
+            indexer_rotary_emb=None,
```
DarkLight1337, line 375: Hmm, there is a default value of `None` already, why is this needed?

JeffLee1874: In the `main` branch, the `MLAModules` class assigns default values (e.g., `None`) to parameters such as `indexer_rotary_emb`. However, older versions—such as v0.12.0—do not provide these defaults, requiring explicit values to be passed during initialization.

For backward compatibility, should we:
- Modify the `modeling.py` to handle missing defaults, **or**
- Update the `MLAModules` class in v0.12.0 to include default values?

DarkLight1337: I don't quite understand why this is a problem of backward compatibility. From your wording, it seems that the old version of `MLAModules` is more restrictive than the new version. Therefore, simply upgrading vLLM to a newer version should not break code that relies on the old version of `MLAModules` since those code should always pass `indexer_rotary_emb`.

JeffLee1874: My mention of "backward compatibility" might have caused some confusion. What I actually meant is that users may have specific version requirements for vLLM—for example, they might not upgrade to the latest version but instead use an older one like `v0.12.0`.

In `v0.12.0`, the `MLAModules` class requires `indexer_rotary_emb` to be explicitly passed during initialization. However, `openpangu.py` does not pass this argument, which leads to a error when initializing the model.

In contrast, starting from `v0.13.0` (and in the `main` branch), `MLAModules` assigns default values (e.g., `None`) to parameters like `indexer_rotary_emb`, so no error occurs.

It’s possible that my PR was targeting the wrong `vllm` version. But should we consider backporting a fix to `v0.12.0`—specifically, updating `openpangu.py` to pass `indexer_rotary_emb=None`—to ensure it works in that release?

DarkLight1337: I don't think that's possible since we don't have a policy of releasing patches for previous releases. I suggest to just update your vLLM version, as even if this PR is merged, it will only be applied to future releases. Hope you understand.

JeffLee1874: Thanks for your reply. Since the issue in the previous releases won't be fixed, I'll close this PR.

PR #30780, opened 2025-12-16
Optimize workspace memory in DeepGEMM.

halyavin:
The tensor workspace13 is used only with one shape.

This optimization is similar to #28254.

### DIFF
--- a/vllm/model_executor/layers/fused_moe/batched_deep_gemm_moe.py
+++ b/vllm/model_executor/layers/fused_moe/batched_deep_gemm_moe.py
@@ -314,7 +314,7 @@ class BatchedDeepGemmExperts(mk.FusedMoEPermuteExpertsUnpermute):
         num_experts = local_num_experts
         max_num_tokens = M if self.max_num_tokens is None else self.max_num_tokens
         activation_out_dim = self.adjust_N_for_activation(N, activation)
-        workspace13 = (num_experts, max_num_tokens * num_dispatchers, max(K, N))
+        workspace13 = (num_experts, max_num_tokens * num_dispatchers, N)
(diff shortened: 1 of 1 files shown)


vllm/model_executor/layers/fused_moe/batched_deep_gemm_moe.py
halyavin, line 382: This is where it is used.

halyavin:
> can you please provide some details about the memory savings you are seeing (and the workload/config/setup?)

I mainly see improvements in stability. I don't have concrete numbers for memory savings. 

I am running DeepSeek V3 in multihost configuration with 1­­-4 single-host prefills and single 2-host decode. DeepSeek model occupies almost all memory on prefills making them prone to OOM errors.

varun-sundar-rabindranath:
> > The change looks harmless but I believe, we might end up allocating more memory for the layer. look here,
> > https://github.com/vllm-project/vllm/blob/38c361f99dffcf2b83725e81a1b5ed949eef43ab/vllm/model_executor/layers/fused_moe/modular_kernel.py#L839
> > 
> > After we get the workspace shapes, we try to re-use workspace13 for the output buffer.
> 
> This got me thinking. Why do we check `prod(workspace13_shape) >= prod(fused_out_shape)` in the first place? If it is the other way around, we can instead reuse `fused_out` buffer as `workspace13` buffer. Since all buffers are allocated together, it is only the total size that matters. This way, the `workspace_shapes` function doesn't need to "lie" about how much space `workspace13` buffer needs.
> 
> This `modular_kernel.py` improvement has to be a separate change that should be committed first.

Yes, I agree. given the correct workspace and fused_out shapes, it'd be better to have `modular_kernel.py` look at the shapes and decide when to re-use and when not to.

PR #30784, opened 2025-12-16
[Improvement] Persist CUDA compat libraries paths to prevent reset on `apt-get`

emricksini-h:
Currently, the Dockerfile registers CUDA compatibility libraries using a transient `RUN ldconfig /path/to/compat` command. This updates the cache but does not persist the configuration.

If a user extends this image or runs a debug container and executes `apt-get` install (which triggers a default `ldconfig`), the custom compatibility path is wiped from the cache. This causes the container to silently fall back to the host driver's native CUDA version (e.g., 12.4) instead of the container's optimized version (12.9), potentially degrading performance or raising version compatibility mismatch errors.

This PR make this more robust by writing the path to `/etc/ld.so.conf.d/00-cuda-compat.conf` before running `ldconfig`. This ensures the compatibility layer persists regardless of future package installations or cache rebuilds.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -118,7 +118,7 @@ ENV UV_LINK_MODE=copy
 RUN gcc --version
 
 # Ensure CUDA compatibility library is loaded
-RUN echo "/usr/local/cuda-$(echo $CUDA_VERSION | cut -d. -f1,2)/compat/" > /etc/ld.so.conf.d/00-cuda-compat.conf && ldconfig
+RUN echo "/usr/local/cuda-$(echo "$CUDA_VERSION" | cut -d. -f1,2)/compat/" > /etc/ld.so.conf.d/00-cuda-compat.conf && ldconfig
 
 # ============================================================
 # SLOW-CHANGING DEPENDENCIES BELOW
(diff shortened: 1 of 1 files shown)


wangshangsam (commented):
Looks reasonable, but, in the PR description, could you include a concrete example of where the existing code fails? Cuz

> If a user extends this image or runs a debug container and executes apt-get install (which triggers a default ldconfig), the custom compatibility path is wiped from the cache. 

This I can understand. But

> This causes the container to silently fall back to the host driver's native CUDA version (e.g., 12.4) instead of the container's optimized version (12.9), potentially degrading performance or raising version compatibility mismatch errors.

This I don't quite understand. I thought that, the container's CUDA version is just the container's CUDA version, and the whole `/usr/local/cuda-$(echo $CUDA_VERSION | cut -d. -f1,2)/compat/` thing is just to enable compatibility mode (i.e., you can run a later CUDA version on a older driver version)?

mgoin (approved):
I see, this makes sense to me as a fragile configuration right now. But I agree with @wangshangsam to clarify

wangshangsam (approved):

PR #30785, opened 2025-12-16
[Fix]Load kv-cache dtype from hf_quant_config.json automatically (fix for reverted PR)

danielafrimi:
HF checkpoint hf_quant_config.json (modelopt) file specifies "kv_cache_quant_algo": "FP8", but vLLM loads the unquantized version of kv-cache.

setting the vLLM server argument explicitly as below:
"--kv-cache-dtype fp8" makes it worked.

this PR fetched the kv_cache_quant_algo from the config itself without specifying it as args (while there is hf_quant_config.json).

The key fix was doing this conversion early when creating `CacheConfig` in `arg_utils.py`, so the resolved value is seen everywhere in the codebase. this way it works exactly like you explicitly passed `--kv-cache-dtype`

The reverted [PR](https://github.com/vllm-project/vllm/pull/29980) that should supposed to supported this feat had a bug, [CI](https://buildkite.com/vllm/ci/builds/43393/steps/canvas?jid=019b176c-8881-47f4-aaf9-5dac38740906) failed on some tests (managed to pass them locally on B200)

### DIFF
--- a/vllm/engine/arg_utils.py
+++ b/vllm/engine/arg_utils.py
@@ -93,6 +93,7 @@ from vllm.transformers_utils.utils import is_cloud_storage
 from vllm.utils.argparse_utils import FlexibleArgumentParser
 from vllm.utils.mem_constants import GiB_bytes
 from vllm.utils.network_utils import get_ip
+from vllm.utils.torch_utils import resolve_kv_cache_dtype_string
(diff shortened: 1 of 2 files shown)


robertgshaw2-redhat (approved):

robertgshaw2-redhat:
Thanks for the update!

PR #30800, opened 2025-12-16
[PERF] Add interleaved memory allocation to NUMA module

skaraban3807:
## Purpose
This PR adds support for interleaved memory allocation when the system is configured with more than one NUMA node. Enabling interleaved allocation helps distribute memory traffic across nodes more evenly, which is particularly beneficial for memory-bound workloads.
In summary, this PR:
- Introduces NUMA-aware interleaved memory allocation when multiple NUMA nodes are present.
- Delivers around **30%** performance uplift on tested AMD EPYC systems for memory-bound workloads.

## Test Plan
export VLLM_CPU_KVCACHE_SPACE=100
export VLLM_CPU_OMP_THREADS_BIND=0-95
export LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libtcmalloc_minimal.so.4:$LD_PRELOAD
export LD_PRELOAD=<env_path>/lib/libiomp5.so:$LD_PRELOAD

vllm bench throughput --input-len 128 --output-len 1024 --model <model_name> --num-prompts 96 --dtype bfloat16 --max-num-seqs 32
## Test Result
I have validated this change on AMD EPYC servers with NPS4-NUMA configurations. Across a range of memory-intensive benchmarks, I observed an average performance improvement of approximately **30%** compared to the baseline (non-interleaved) configuration.
image

---
<details>

</details>

### DIFF
--- a/csrc/cpu/utils.cpp
+++ b/csrc/cpu/utils.cpp
@@ -24,6 +24,8 @@ std::string init_cpu_threads_env(const std::string& cpu_ids) {
 #ifndef VLLM_NUMA_DISABLED
 std::string init_cpu_threads_env(const std::string& cpu_ids) {
   bitmask* omp_cpu_mask = numa_parse_cpustring_all(cpu_ids.c_str());
+  TORCH_CHECK(omp_cpu_mask != nullptr,
+              "Failed to parse CPU string: " + cpu_ids);
   TORCH_CHECK(omp_cpu_mask->size > 0);
   std::vector<int> omp_cpu_ids;
   omp_cpu_ids.reserve(omp_cpu_mask->size);
@@ -44,19 +46,11 @@ std::string init_cpu_threads_env(const std::string& cpu_ids) {
 
   // Memory node binding
   if (numa_available() != -1) {
-    int mem_node_id = numa_node_of_cpu(omp_cpu_ids.front());
     std::set<int> node_ids;
     for (const auto& cpu_id : omp_cpu_ids) {
       int node_id = numa_node_of_cpu(cpu_id);
       if (node_id != -1) {
         node_ids.insert(node_id);
-      }
-      if (node_id != mem_node_id) {
-        TORCH_WARN("CPU ", cpu_id, " is on NUMA node ", node_id, ", but CPU ",
-                   omp_cpu_ids.front(), " is on NUMA node ", mem_node_id,
-                   ". All CPUs should be on the same NUMA node for optimal "
-                   "performance. Memory will be bound to NUMA node ",
-                   mem_node_id, ".");
       }
     }
     // Concatenate all node_ids into a single comma-separated string
@@ -70,7 +64,7 @@ std::string init_cpu_threads_env(const std::string& cpu_ids) {
       }
 
       bitmask* mask = numa_parse_nodestring(node_ids_str.c_str());
-      bitmask* src_mask = numa_get_membind();
+      bitmask* src_mask = numa_get_mems_allowed();
 
       int pid = getpid();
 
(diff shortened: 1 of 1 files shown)


csrc/cpu/utils.cpp
```diff
@@ -83,8 +81,26 @@ std::string init_cpu_threads_env(const std::string& cpu_ids) {
...
+        // Restrict memory allocation to the selected NUMA node(s).
+        // Enhances memory locality for the threads bound to those NUMA CPUs.
+        if (node_ids.size() > 1) {
+          std::stringstream interleave_msg;
+          interleave_msg << "\nNUMA binding: Using INTERLEAVE policy for memory "
+                         << "allocation across multiple NUMA nodes (nodes: "
+                         << node_ids_str << "). Memory allocations will be "
+                         << "interleaved across the specified NUMA nodes.\n";
+          std::cout << interleave_msg.str();
+          numa_set_interleave_mask(mask);
```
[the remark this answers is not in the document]
skaraban3807: numa_set_interleave_mask(mask) functions return void not integer. So this suggestion is not valid

csrc/cpu/utils.cpp
```diff
@@ -83,8 +81,26 @@ std::string init_cpu_threads_env(const std::string& cpu_ids) {
...
+        } else {
+          std::stringstream membind_msg;
+          membind_msg << "\nNUMA binding: Using MEMBIND policy for memory "
+                    << "allocation on a single NUMA node ("
+                    << node_ids_str << "). Memory allocations will be "
+                    << "strictly bound to this NUMA node.\n";
+          std::cout << membind_msg.str();
+          numa_set_membind(mask);
```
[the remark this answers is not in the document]
skaraban3807: numa_set_membind(mask) functions return void not integer. So this suggestion is not valid

csrc/cpu/utils.cpp
```diff
@@ -83,8 +81,27 @@ std::string init_cpu_threads_env(const std::string& cpu_ids) {
...
+        // Restrict memory allocation to the selected NUMA node(s).
+        // Enhances memory locality for the threads bound to those NUMA CPUs.
+        if (node_ids.size() > 1) {
+          std::stringstream interleave_msg;
+          interleave_msg
+              << "\nNUMA binding: Using INTERLEAVE policy for memory "
+              << "allocation across multiple NUMA nodes (nodes: "
+              << node_ids_str << "). Memory allocations will be "
+              << "interleaved across the specified NUMA nodes.\n";
+          std::cout << interleave_msg.str();
```
bigPYJ1151: Please use ```LOG(INFO)``` from ```c10/util/Logging.h```

csrc/cpu/utils.cpp
```diff
@@ -70,7 +63,12 @@ std::string init_cpu_threads_env(const std::string& cpu_ids) {
       }
 
       bitmask* mask = numa_parse_nodestring(node_ids_str.c_str());
-      bitmask* src_mask = numa_get_membind();
+      bitmask* src_mask;
+      if (node_ids.size() > 1) {
+        src_mask = numa_get_interleave_mask();
```
bigPYJ1151: Here we want to retrival allowed memory nodes of current process for the page migration. By default Linux uses ```localalloc```, then ```numa_get_interleave_mask``` will return a empty mask. So we should always ues ```numa_get_membind``` here to get correct memory nodes.

csrc/cpu/utils.cpp
skaraban3807: Hi @bigPYJ1151, Thank you for the comment.
`LOG(INFO)` is controlled by the `TORCH_CPP_LOG_LEVEL` environment variable. In practice, this means the message is only emitted when `TORCH_CPP_LOG_LEVEL=INFO` is explicitly set. If users do not configure this environment variable, the log will never be visible, which significantly limits its utility for diagnosing configuration or runtime issues. To ensure that the message is visible under default settings and provides actionable feedback in typical deployments, I have changed it to `TORCH_WARN`. 

Please let me know if `LOG(INFO)` is a requirement for vLLM or if the move to `TORCH_WARN` is acceptable.

csrc/cpu/utils.cpp
skaraban3807: numa_get_run_node_mask() returns the effective set of NUMA nodes the process can currently allocate memory on, regardless of which memory policy is active. Since we don't know what policy the process was started with, numa_get_run_node_mask() can be a reliable choice.

| Function | What it returns | When it works correctly |
|----------|-----------------|-------------------------|
| `numa_get_membind()` | Nodes from MEMBIND policy | Only if MEMBIND is active |
| `numa_get_interleave_mask()` | Nodes from INTERLEAVE policy | Only if INTERLEAVE is active |
| `numa_get_run_node_mask()` | Current effective node mask | **Always** (any policy) |

@bigPYJ1151 please let me know if you have any other view on this.

csrc/cpu/utils.cpp
```diff
@@ -70,7 +64,7 @@ std::string init_cpu_threads_env(const std::string& cpu_ids) {
       }
 
       bitmask* mask = numa_parse_nodestring(node_ids_str.c_str());
-      bitmask* src_mask = numa_get_membind();
+      bitmask* src_mask = numa_get_run_node_mask();
```
bigPYJ1151: Looks like ```numa_get_run_node_mask``` returns a mask of CPUs on which the current task is allowed to run.

Perhaps ```numa_get_mems_allowed``` is a more proper usage? We can get all avaliable mem nodes as source for the page migration.

skaraban3807: Hi @bigPYJ1151, I have updated the code according to your suggestions. Please let me know if you have any further questions or concerns.

bigPYJ1151 (approved):
LGTM 
Verified locally, both of allocation strategy work as expected.

PR #30803, opened 2025-12-16
RayLLM Bugfix - Preserve obj store URL for multi engine_config creation

omer-dayan:
Continueing the discussion from this PR: https://github.com/vllm-project/vllm/pull/30154

We suggest in here a solution that works for both RayLLM + vllm serve.

The root cause of the bug is that object store path is cloned to local dir, and `self.model` is changed to that local dir while `self.model_weights` contains the object store path - unless you run create_engine_config once, and then re-run it and now `self.model` equal to local dir and `self.model_weights` is not preserved.

The solution proposing a new field in arg_utils of model_weight with such that even after multiple calls to `create_engine_config` the values of `self.model` and `self.model_weights` both in `ModelConfig` and in `EngineArgs` are preserved.

@jiangwu300
I managed to reproduce and made sure it fix the bug, can you check it yourself and approve?

@DarkLight1337 @Isotr0py  Adding you both as of the interest you had in the original PR and the Streamer


jiangwu300:
@Isotr0py are you able to take a look at this PR? I'll scrap mine afterwards if this is good to go.

Isotr0py (approved):
Thanks, this solution looks more generic!

yewentao256 (approved):
LGTM, thanks for the work!

vllm/config/model.py
```diff
@@ -102,6 +102,10 @@ class ModelConfig:
     """Name or path of the Hugging Face model to use. It is also used as the
     content for `model_name` tag in metrics output when `served_model_name` is
     not specified."""
+    model_weights: str | None = None
```
Isotr0py:
```suggestion
    model_weights: str = ""
```
Seems default NoneType will break existing model loading.

vllm/config/model.py
```diff
@@ -688,6 +692,10 @@ def maybe_pull_model_tokenizer_for_runai(self, model: str, tokenizer: str) -> No
             tokenizer: Tokenizer name or path
         """
 
+        # Skip if model_weights is already set (model already pulled)
+        if self.model_weights is not None:
```
Isotr0py:
```suggestion
        if self.model_weights:
```

vllm/engine/arg_utils.py
```diff
@@ -356,6 +356,7 @@ class EngineArgs:
     """Arguments for vLLM engine."""
 
     model: str = ModelConfig.model
+    model_weights: str | None = ModelConfig.model_weights
```
Isotr0py:
```suggestion
    model_weights: str = ModelConfig.model_weights
```

PR #30804, opened 2025-12-16
[CI] Skip ci failure test

yewentao256:
## Purpose

Skip for https://buildkite.com/vllm/ci/builds/43505#019b1d08-18c9-453e-8104-83e07006c4e9

We will find the root cause and fix accordingly


ProExpertProg (approved):

DarkLight1337:
Do we need to set both `fuse_act_quant` and `fuse_norm_quant` here? See https://github.com/vllm-project/vllm/pull/30787

PR #30808, opened 2025-12-16
[Refactor][TPU] Remove torch_xla path and use tpu-inference

weiyu0824:
## Purpose
Removes torch_xla related code paths as this backend is now deprecated. To run vLLM on TPU, users should now install and use tpu-inference.

## Test Plan
- Triggered tpu-inference CI/CD pipeline.
- Verified that removal does not impact non-TPU backends.

## Test Result
- tpu-inference CI/CD: https://buildkite.com/tpu-commons/tpu-inference-ci/builds/71

<details>
<summary> What has been removed </summary>

- Classes previously migrated to tpu-inference, this is guarded by USE_TPU_INFERENCE before.
- Functions explicitly used torch_xla.
- All TPU-related test files (migrated to tpu-inference repo)
</details>

<details>
<summary> Pending Removal </summary>

- Update README.md
- Update Dockerfile.tpu
</details>


vllm/v1/worker/tpu_worker.py
```diff
@@ -2,350 +2,15 @@
...
-    def shutdown(self) -> None:
-        self.model_runner.ensure_kv_transfer_shutdown()
-
-    def apply_model(self, fn: Callable[[nn.Module], _R]) -> _R:
-        """Apply a function on the model inside this worker."""
-        return fn(self.get_model())
-
-
 if USE_TPU_INFERENCE:
     from tpu_inference.worker.tpu_worker import TPUWorker as TpuInferenceWorker
```
yaochengji, line 16: Could we add a TODO to use the hardware plugin in an official way? After that, we don't need such a tpu_workder.py file.

vllm/model_executor/model_loader/default_loader.py
```diff
@@ -242,20 +242,7 @@ def _get_weights_iterator(
                 )
 
         if current_platform.is_tpu():
```
yaochengji: We can move this two lines totally.

yaochengji (commented):
Thanks for the cleanup! Could you please fix the pre-commit?

vllm/model_executor/model_loader/default_loader.py
weiyu0824: Fixed

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -68,21 +68,15 @@ def _eplb_map_to_physical_and_record(
...
-else:
-    fused_moe_pallas = None  # type: ignore
-
 from vllm.model_executor.layers.fused_moe.fused_moe_method_base import (
     FusedMoEMethodBase,
 )
 from vllm.model_executor.layers.fused_moe.fused_moe_modular_method import (
     FusedMoEModularMethod,
 )
+from vllm.model_executor.layers.fused_moe.rocm_aiter_fused_moe import (  # noqa: E501
```
yaochengji: Why did we add rocm related code?

yaochengji (approved):
LGTM, thanks!

vllm/model_executor/layers/fused_moe/layer.py
weiyu0824: This line: 
from vllm.model_executor.layers.fused_moe.rocm_aiter_fused_moe import (  # noqa: E501
    rocm_aiter_grouped_topk,
)
is moved by `pre-commit run all-files`

weiyu0824:
Affected by https://github.com/vllm-project/vllm/issues/31044#issue

PR #30718, opened 2025-12-15
[Perf] Cache for deepgemm moe, 2.1% Throuput improvement, 2.2% TTFT improvement

yewentao256:
## Purpose

A follow up PR for https://github.com/vllm-project/vllm/pull/30282#issuecomment-3648318869

We can optimize the performance through building the kernel cache

## Test

`export MODEL="deepseek-ai/DeepSeek-V3.2"`

### Acc

`lm_eval --model local-completions --model_args "base_url=http://127.0.0.1:9256/v1/completions,model=$MODEL,num_concurrent=1024" --tasks gsm8k`

```bash
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9507|±  | 0.006|
|     |       |strict-match    |     5|exact_match|↑  |0.9492|±  | 0.006|
```

### Performance

`vllm bench serve --model $MODEL  --dataset-name random --host 127.0.0.1 --port 9256 --random-input-len 2 --random-output-len 16 --request-rate inf --num-prompts 1024 --max-concurrency 128`

```bash
main
============ Serving Benchmark Result ============
Successful requests:                     1024      
Failed requests:                         0         
Maximum request concurrency:             128       
Benchmark duration (s):                  19.65     
Total input tokens:                      2048      
Total generated tokens:                  16384     
Request throughput (req/s):              52.12     
Output token throughput (tok/s):         833.99    
Peak output token throughput (tok/s):    896.00    
Peak concurrent requests:                256.00    
Total token throughput (tok/s):          938.24    
---------------Time to First Token----------------
Mean TTFT (ms):                          295.75    
Median TTFT (ms):                        289.03    
P99 TTFT (ms):                           352.55    
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          143.67    
Median TPOT (ms):                        143.59    
P99 TPOT (ms):                           144.50    
---------------Inter-token Latency----------------
Mean ITL (ms):                           143.67    
Median ITL (ms):                         142.93    
P99 ITL (ms):                            157.59    
==================================================

now
============ Serving Benchmark Result ============
Successful requests:                     1024      
Failed requests:                         0         
Maximum request concurrency:             128       
Benchmark duration (s):                  19.24     
Total input tokens:                      2048      
Total generated tokens:                  16384     
[...]

PR #30820, opened 2025-12-16
[Bug] Fix compressed tensor not using deepgemm

yewentao256:
## Purpose

A fix for https://github.com/vllm-project/vllm/pull/30718#discussion_r2624920272

CT should use deepgemm when available, this PR fixes that

## Test

`export MODEL="RedHatAI/Qwen3-30B-A3B-FP8-block"`

`vllm serve $MODEL -tp 2 --port 9256  --enable-expert-parallel `

`lm_eval --model local-completions --model_args "base_url=http://127.0.0.1:9256/v1/completions,model=$MODEL,num_concurrent=1024" --tasks gsm8k`

```bash
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8590|±  |0.0096|
|     |       |strict-match    |     5|exact_match|↑  |0.8886|±  |0.0087|
```

Locally assertion confirms that deepgemm moe was called


mgoin (commented):
LGTM, did you test with a model that it works? We should check TP and EP. This model might work https://huggingface.co/RedHatAI/Qwen3-30B-A3B-FP8-block

mgoin (approved):

PR #30822, opened 2025-12-16
[Bugfix][torch2.10] Fix test_qwen2_5_vl_compilation with 2.10 RC

Lucaskabela:
## Purpose
See https://github.com/pytorch/pytorch/issues/170568 

In torch 2.10, we will rely more heavily on aot_precompile; part of this upgrade leads to a successful load and re-instantion of the `VLLMBackend` in caching.py; however, for multimodal encoders that previously relied on the `set_model_tag` context manager to forward `is_encoder` information to compile ranges, this field is no longer set (since aot loading happens at callsite)

Therefore, we need to serialize/deserialize the `is_encoder` information in addition to the `prefix`

## Test Plan
```
pytest tests/compile/fullgraph/test_multimodal_compile.py::test_qwen2_5_vl_compilation
```
## Test Result
```
1 passed, 18 warnings in 31.14s 
```

---
<details>

</details>


vllm/compilation/piecewise_backend.py
```diff
@@ -170,8 +170,7 @@ def __call__(self, *args) -> Any:
         range_entry = self._find_range_for_shape(runtime_shape)
 
         assert range_entry is not None, (
-            f"Shape out of considered range: {runtime_shape} "
-            "[1, max_num_batched_tokens]"
+            f"Shape: {runtime_shape} out of considered ranges: {self.compile_ranges}"
```
Lucaskabela, line 173: Update this error message, as for multimodal models, [1, max_num_batched_tokens]` is not the right compile range

vllm/compilation/caching.py
```diff
@@ -105,7 +108,10 @@ def deserialize_compile_artifacts(cls, data: bytes) -> "VllmSerializableFunction
         fake_mode = FakeTensorMode(shape_env=ShapeEnv())
         state["graph_module"] = GraphPickler.loads(state["graph_module"], fake_mode)
         state["example_inputs"] = GraphPickler.loads(state["example_inputs"], fake_mode)
-        vllm_backend = VllmBackend(get_current_vllm_config(), state["prefix"])
+        is_encoder = state.get("is_encoder", False)
+        vllm_backend = VllmBackend(
+            get_current_vllm_config(), state["prefix"], is_encoder
+        )
```
zou3519, line 115: I'm trying to get a high-level overview of what is going on. @zhxchen17 @Lucaskabela.
Are we saying that aot_compile now works on image encoders? And that we were missing serialization of part of the backend?

zou3519 (approved):

vllm/compilation/caching.py
Lucaskabela: Yes we missed serializing the (very recently added) `is_encoder` field from the VllmBackend.  This is needed to set the correct `compile_ranges` in PiecewiseBackend when recreated in `__call__` from loading the function

In regards to "aot_compile now works with image encoders", if the encoder is using torch.compile integration then when we `__call__` it will go through the aot_load logic; there is no special gating

See:
https://github.com/vllm-project/vllm/blob/2497228ad4427310bc55427f6db404a00de4fd78/vllm/compilation/decorators.py#L381

PR #30833, opened 2025-12-17
[Quant] Make static quant support all group shapes

LucasWilkinson:
Preparatory pr for https://github.com/vllm-project/vllm/pull/30141 (per-head quant)

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit c82c073a6134ce25dbd7234699e8e6e66021dc4f. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Adds flexible static FP8 quantization with explicit group shapes and multi-shape scale support.
> 
> - New CUDA kernel `scaled_fp8_quant_kernel_strided_group_shape` with stride-based dispatch, vectorized paths, and group-aware indexing (per-tensor/channel/token and arbitrary 2D groups)
> - Updates C++ op: `static_scaled_fp8_quant(result, input, scale, group_shape?)` to infer/validate group sizes from 0D/1D/2D scales; binding schema adjusted in `torch_bindings.cpp`
> - Python API: `_custom_ops.scaled_fp8_quant` accepts `group_shape`; `QuantFP8` passes `group_shape` for static mode and refines dynamic/group logic
> - Tests: add coverage for 2D group scales and 1D scales (per-token/channel) using `scaled_quantize`; opcheck updated to include `group_shape`
> - Utils: `group_broadcast` supports tensors with fewer dims; `scaled_quantize` gets `compute_dtype` for higher precision
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c82c073a6134ce25dbd7234699e8e6e66021dc4f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/csrc/quantization/w8a8/fp8/common.cu
+++ b/csrc/quantization/w8a8/fp8/common.cu
@@ -4,6 +4,7 @@
 #include "quantization/vectorization_utils.cuh"
 #include <c10/cuda/CUDAGuard.h>
 #include <ATen/cuda/Exceptions.h>
+#include <tuple>
 
 namespace vllm {
 
@@ -175,11 +176,11 @@ __global__ void dynamic_per_token_scaled_fp8_quant_kernel_strided(
 }  // namespace vllm
 
 void static_scaled_fp8_quant(
-    torch::Tensor& out,                  // [..., d]
-    torch::Tensor const& input,          // [..., d]
-    torch::Tensor const& scale,          // various shapes
-    std::optional<int64_t> opt_group_m,  // optional explicit group_m
-    std::optional<int64_t> opt_group_n)  // optional explicit group_n
+    torch::Tensor& out,          // [..., d]
+    torch::Tensor const& input,  // [..., d]
+    torch::Tensor const& scale,  // various shapes
+    std::optional<std::tuple<int64_t, int64_t>>
+        opt_group_shape)  // optional explicit (group_m, group_n)
 {
   TORCH_CHECK(input.stride(-1) == 1,
               "last dimension of input must be contiguous");
@@ -195,10 +196,6 @@ void static_scaled_fp8_quant(
   int group_m, group_n;
   int64_t scale_stride_0, scale_stride_1;
 
-  // If explicit group shape is provided, use it
-  const bool has_explicit_group =
-      opt_group_m.has_value() && opt_group_n.has_value();
-
   if (scale.dim() == 0 || scale.numel() == 1) {
     // Per-tensor: one scale for the entire tensor
     group_m = num_tokens;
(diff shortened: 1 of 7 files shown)


csrc/quantization/w8a8/fp8/common.cu
```diff
@@ -7,25 +7,66 @@
...
+          blockDim.x, [=] __device__(fp8_type & dst, const scalar_t& src) {
+            dst = scaled_fp8_conversion<true, fp8_type>(static_cast<float>(src),
+                                                        inv_scale);
+          });
+    }
+  } else {
+    // Scalar path for small column groups (group_n < VEC_SIZE)
+    for (int n = tid; n < hidden_size; n += blockDim.x) {
+      const int gj = n / group_n;
+      const float inv_scale =
```
MatthewBonanni: We could save on `div`s by only computing `inv_scale` when `n % group_n == 0` (or `n == gj * group_n` to avoid the `rem`)

csrc/quantization/w8a8/fp8/common.cu
```diff
@@ -7,25 +7,66 @@
...
+              : 1.0f / scale[gi * scale_stride_0 + gj * scale_stride_1];
+
+      vectorize_with_alignment<VEC_SIZE>(
+          token_in + gj * group_n, token_out + gj * group_n, group_n, tid,
+          blockDim.x, [=] __device__(fp8_type & dst, const scalar_t& src) {
+            dst = scaled_fp8_conversion<true, fp8_type>(static_cast<float>(src),
+                                                        inv_scale);
+          });
+    }
+  } else {
```
MatthewBonanni, line 70: Should we add a specialization for `group_n = 1` and use vectorized loads for both input and scale?

LucasWilkinson: I think it's fine to leave it for now given this isn't a used pathway yet; only per-tensor is currently used and this is mainly targeting per-head in which case group_n > 1. But good idea! if we see a PR that would use something like this we can try it and benchmark it then :+1:

csrc/quantization/w8a8/fp8/common.cu
LucasWilkinson: done

csrc/quantization/w8a8/fp8/common.cu
```diff
@@ -153,17 +301,32 @@ void static_scaled_fp8_quant(torch::Tensor& out,          // [..., d]
 
   const at::cuda::OptionalCUDAGuard device_guard(device_of(input));
   const cudaStream_t stream = at::cuda::getCurrentCUDAStream();
+
+// Dispatch to template-specialized kernel based on stride pattern
+#define LAUNCH_KERNEL(S0_ZERO, S1_ZERO)                                        \
```
ProExpertProg: Should be able to use `DISPATCH_BOOL` macro here

csrc/quantization/w8a8/fp8/common.cu
```diff
@@ -133,17 +182,116 @@ __global__ void dynamic_per_token_scaled_fp8_quant_kernel_strided(
...
+    const int64_t expected_scale_n = hidden_size / group_n;
+    const int64_t expected_scale_numel = expected_scale_m * expected_scale_n;
+
+    TORCH_CHECK(scale_len == expected_scale_numel, "1D scale length (",
+                scale_len, ") does not match expected size (",
+                expected_scale_numel, ") for group_shape (", opt_group_m, ", ",
+                opt_group_n, ") with input shape (", num_tokens, ", ",
+                hidden_size, ")");
+
+    // For 1D scale, determine strides based on which dim is trivial
```
ProExpertProg, line 237: It seems to me that this logic here is similar to `group_broadcast` - could we extract into a utility function & op and test that they stay in sync? Can be a follow-up

ProExpertProg (approved):

mgoin (approved):

PR #28775, opened 2025-11-15
[Model] Add support for openPangu moe model

yt0428:
## Purpose
This PR adds support for openpangu moe model, which characterizes by its different kv head size and sink kv in attention.

The model has two new features that have not been supported:
1. Different key head size and value head size.
Although flash_attn kernel can handle the different kv head size, current vllm framwork does not make it optional choice. In the implemented `FlashSinkAttentionBackend`, the kv_cache_shape is defined to be of shape `[num_blocks, block_size, num_kv_heads, head_size_k + head_size_v]`. The corresponding kv_cache update function `reshape_and_cache_kernel_flash_diffkv` is implemented by triton.
2. `sink_key` and `sink_value` in attention
The attention module of the model receive two more arguments `sink_key` and `sink_value`, which is learned durining the training and shared by all inputs. In this initial implementation, I store them in the first blocks in the block pool and remove these blocks from the free blocks, so they can not be scheduled to avoid overwriting. During the forward of `FlashSinkAttentionBackend`, block_ids of `sink_key` and `sink_value` are concated to the normal block_table to calculate attention correctly.

11.26 Update:
We refactor the code to seperate two features above, specifically:
1. We rename the `FlashSinkAttentionBackend` to `FlashDiffkvAttentionBackend`, which is modified from `FlashAttentionBackend` to support different head_size for key and value.
2. We move the `sink_key` related logic to `GPUModelRunner`, in which we primarily do two thing. First, we store the sink_key and sink_value to kv_caches during the initialization of kv_caches. This is implemented by adding a function `prepare_sink_kv_cache` and called in `initializa_kv_cache`. Second, we modify the blk_table_tensor and seq_lens in-place in `_build_attention_metadata`, so that attention backends will know that there are sink_key and sink_value in kv_caches

**The current implementation is just functional corrently but not so elegant. And suggestions from those who are familiar with these parts of vllm are very appreciated. Many thanks!!**

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
[...]

PR #30836, opened 2025-12-17
[Model] Add MiMo-V2-Flash support

Abatom:
## Purpose
Add support for MiMo-V2-Flash.
### Example 1
```
vllm serve XiaomiMiMo/MiMo-V2-Flash \
    --host 0.0.0.0 \
    --port 9001 \
    --seed 1024 \
    --served-model-name base_model \
    --tensor-parallel-size 4 \
    --trust-remote-code \
    --generation-config vllm
```
### Example 2
```
vllm serve XiaomiMiMo/MiMo-V2-Flash \
    --host 0.0.0.0 \
    --port 9001 \
    --seed 1024 \
    --served-model-name base_model \
    --data-parallel-size 2 \
    --tensor-parallel-size 4 \
    --enable-expert-parallel \
    --trust-remote-code \
    --generation-config vllm
```
## Accuracy
GSM8K
```
lm_eval \
    --tasks gsm8k \
    --num_fewshot 5 \
    --apply_chat_template \
    --model_args model=base_model,base_url=http://0.0.0.0:9001/v1/chat/completions,num_concurrent=100,max_retries=20,tokenized_requests=False,tokenizer_backend=none,max_gen_toks=256
```
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9128|±  |0.0078|
|     |       |strict-match    |     5|exact_match|↑  |0.9075|±  |0.0080|
```


vllm/model_executor/layers/linear.py
```diff
@@ -906,9 +935,11 @@ def __init__(
         *,
         return_bias: bool = True,
         disable_tp: bool = False,
+        v_head_size: Optional[int] = None,
```
jeejeelee:
```suggestion
        v_head_size: int | None = None,
```

vllm/model_executor/models/mimo_v2_flash.py
```diff
@@ -0,0 +1,738 @@
...
+        self.n_local_physical_experts = self.n_physical_experts // self.ep_size
+
+        self.physical_expert_start = self.ep_rank * self.n_local_physical_experts
+        self.physical_expert_end = (
+            self.physical_expert_start + self.n_local_physical_experts
+        )
+
+        dtype = getattr(config, "moe_router_dtype", "float32")
+        if dtype == "auto":
+            self.gate_dtype = None  # 使用默认 dtype
```
jeejeelee:
```suggestion
            self.gate_dtype = None  
```

vllm/model_executor/models/mimo_v2_flash.py
```diff
@@ -0,0 +1,738 @@
...
+
+
+    def compute_logits(
+        self,
+        hidden_states: torch.Tensor,
+    ) -> torch.Tensor | None:
+        logits = self.logits_processor(self.lm_head, hidden_states)
+        return logits
+
+    def get_expert_mapping(self) -> list[tuple[str, str, int, str]]:
```
jeejeelee, line 502: We need  to move this function to `MiMoV2Model`, you can refer to https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/qwen3_moe.py#L750

vllm/model_executor/models/mimo_v2_flash.py
```diff
@@ -0,0 +1,738 @@
...
+        # (param_name, weight_name, expert_id, shard_id)
+        return FusedMoE.make_expert_params_mapping(
+            ckpt_gate_proj_name="gate_proj",
+            ckpt_down_proj_name="down_proj",
+            ckpt_up_proj_name="up_proj",
+            num_experts=self.config.n_routed_experts,
+            num_redundant_experts=self.num_redundant_experts,
+        )
+
+    def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
```
jeejeelee, line 513: DITTO

jeejeelee (commented):
Some init comments, thank you for contribution

jeejeelee:
To add a new LLM model, you also need to :
- Add the model to the [documentation](https://github.com/vllm-project/vllm/blob/main/docs/models/supported_models.md#generative-models)
- Register the model in [test registry](https://github.com/vllm-project/vllm/blob/main/tests/models/registry.py#L174)

vllm/model_executor/models/mimo_v2_flash.py
```diff
@@ -0,0 +1,731 @@
...
+
+        dtype = getattr(config, "moe_router_dtype", "float32")
+        if dtype == "auto":
+            self.gate_dtype = None
+        elif dtype == "float" or dtype == "float32":
+            self.gate_dtype = torch.float32
+        elif dtype == "bfloat16":
+            self.gate_dtype = torch.bfloat16
+        elif dtype == "half" or dtype == "float16":
+            self.gate_dtype = torch.float16
```
Isotr0py: Perhaps use `_STR_DTYPE_TO_TORCH_DTYPE`?
https://github.com/vllm-project/vllm/blob/a9e15c21efbbc5b4a7a1e69e40378fdfe1acdcb7/vllm/config/model.py#L1844-L1850

vllm/model_executor/models/mimo_v2_flash.py
```diff
@@ -0,0 +1,731 @@
...
+            v_head_size=self.v_head_dim,
+        )
+
+        o_proj_prefix = f"{prefix}.o_proj"
+        normalized_name = o_proj_prefix.lstrip(".")
+        o_proj_quant_config = quant_config
+        if quant_config is not None:
+            ignored = getattr(quant_config, "ignored_layers", None)
+            if ignored is not None and normalized_name in ignored:
+                o_proj_quant_config = None
```
Isotr0py: I think quantization config should handle layer skipping automatically?

cinjon:
Will this support MTP and the tool-call-parser? Like in the sglang instructions at https://huggingface.co/XiaomiMiMo/MiMo-V2-Flash :

```
# Launch server
python3 -m sglang.launch_server \
        --model-path XiaomiMiMo/MiMo-V2-Flash \
        --served-model-name mimo-v2-flash \
        --pp-size 1 \
        --dp-size 2 \
        --enable-dp-attention \
        --tp-size 8 \
        --moe-a2a-backend deepep \
        --page-size 1 \
        --host 0.0.0.0 \
        --port 9001 \
        --trust-remote-code \
        --mem-fraction-static 0.75 \
        --max-running-requests 128 \
        --chunked-prefill-size 16384 \
        --reasoning-parser qwen3 \
        --tool-call-parser mimo \
        --context-length 262144 \
        --attention-backend fa3 \
        --speculative-algorithm EAGLE \
        --speculative-num-steps 3 \
        --speculative-eagle-topk 1 \
        --speculative-num-draft-tokens 4 \
        --enable-mtp
```

vllm/model_executor/models/mimo_v2_flash.py
Abatom: Done!

vllm/model_executor/models/mimo_v2_flash.py
Abatom: @Isotr0py  Done!

vllm/model_executor/models/mimo_v2_flash.py
```diff
@@ -0,0 +1,727 @@
...
+        self.qkv_proj = QKVParallelLinear(
+            hidden_size,
+            self.head_dim,
+            self.total_num_heads,
+            self.total_num_kv_heads,
+            bias=attention_bias,
+            quant_config=quant_config,
+            prefix=f"{prefix}.qkv_proj",
+            v_head_size=self.v_head_dim,
+        )
```
Isotr0py, line 257:
```suggestion
        self.qkv_proj = MergedColumnParallelLinear(
            input_size=hidden_size,
            output_sizes=[
                self.q_size * tp_size,
                self.k_size * tp_size,
                self.v_size * tp_size,
            ],
            bias=attention_bias,
            quant_config=quant_config,
            prefix=f"{prefix}.qkv_proj",
        )
```
I think we can use `MergedColumnParallelLinear` like https://github.com/vllm-project/vllm/pull/28775 for flexibility, so that we don't need to introduce new arguments for `QKVParallelLinear`.

vllm/model_executor/models/mimo_v2_flash.py
```diff
@@ -0,0 +1,727 @@
...
+            num_experts=self.config.n_routed_experts,
+            num_redundant_experts=self.num_redundant_experts,
+        )
+
+    def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
+        stacked_params_mapping = [
+            # (param_name, shard_name, shard_id)
+            ("qkv_proj", "q_proj", "q"),
+            ("qkv_proj", "k_proj", "k"),
+            ("qkv_proj", "v_proj", "v"),
```
Isotr0py, line 518:
```suggestion
            ("qkv_proj", "q_proj", 0),
            ("qkv_proj", "k_proj", 1),
            ("qkv_proj", "v_proj", 2),
```
If use `MergedColumnParallelLinear`, we need to also update `stacked_params_mapping`.

vllm/model_executor/layers/quantization/utils/fp8_utils.py
```diff
@@ -1252,6 +1252,14 @@ def validate_fp8_block_shape(
     """Validate block quantization shapes for tensor parallelism."""
     from vllm.distributed import get_tensor_model_parallel_world_size
 
+    if getattr(layer, "allow_fp8_block_shape_mismatch", False):
+        logger.debug(
+            "Skipping FP8 block shape validation for layer %s due to detected"
+            " mismatch allowance.",
+            getattr(layer, "prefix", "<unknown>"),
+        )
+        return
```
Isotr0py, line 1261: I'm a bit worried that this will cause unexpected behavior for FP8 kernel if we disabled block shape check. 

Perhaps we should improve the block shape check for Mimo-V2's edge case instead of just skipping it.

Perhaps @mgoin can give more insights?

vllm/model_executor/models/mimo_v2_flash.py
```diff
@@ -0,0 +1,727 @@
...
+
+        self.physical_expert_start = self.ep_rank * self.n_local_physical_experts
+        self.physical_expert_end = (
+            self.physical_expert_start + self.n_local_physical_experts
+        )
+
+        dtype = getattr(config, "moe_router_dtype", "float32")
+        self.gate_dtype = str_dtype_to_torch_dtype(dtype)
+
+        self.gate = ReplicatedLinear(
```
jeejeelee: Maybe we can implement gate like [glm4_moe gate](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/glm4_moe.py#L142) does

vllm/model_executor/models/mimo_v2_flash.py
```diff
@@ -0,0 +1,727 @@
...
+    def forward(
+        self,
+        positions: torch.Tensor,
+        hidden_states: torch.Tensor,
+    ) -> torch.Tensor:
+        qkv, _ = self.qkv_proj(hidden_states)
+        q, k, v = qkv.split([self.q_size, self.k_size, self.v_size], dim=-1)
+        q, k = self.rotary_emb(positions, q, k)
+        if self.v_scale is not None:
+            v = v * self.v_scale
```
jeejeelee: Mabye we can fuse this multiplication into v's weight In following PR

jeejeelee:
Here are the GSM8K results from my local testing

```
local-chat-completions (model=XiaomiMiMo/MiMo-V2-Flash,base_url=http://0.0.0.0:8000/v1/chat/completions,num_concurrent=100,max_retries=20,tokenized_requests=False,tokenizer_backend=none,max_gen_toks=256), gen_kwargs: (None), limit: None, num_fewshot: 5, batch_size: 1
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9075|±  |0.0080|
|     |       |strict-match    |     5|exact_match|↑  |0.9022|±  |0.0082|

```

vllm/model_executor/layers/quantization/utils/fp8_utils.py
Abatom: @Isotr0py We tried removing the code above and got the following error.
```
ValueError: Weight output_partition_size = 192 is not divisible by weight quantization block_n = 128.
```

vllm/model_executor/models/mimo_v2_flash.py
Abatom: Done!

vllm/model_executor/models/mimo_v2_flash.py
Abatom: @Isotr0py We attempted a modification and found it affected both accuracy and performance.

vllm/model_executor/models/mimo_v2_flash.py
Abatom: Done!

vllm/model_executor/models/mimo_v2_flash.py
Isotr0py: Oh, it's because mimo-v2 has `num_key_value_heads=4`, so using `MergedColumnParallelLinear` will hit edge case with large tp size. Let's keep `QKVParallelLinear` here then.

jeejeelee (approved):
Thank you for contribution, let's land this model first and continue improving it in subsequent PRs

Isotr0py (approved):
Unlock quantization tests CI to see if other models are fine with the temporary hacky fp8 patch.

vllm/model_executor/layers/quantization/utils/fp8_utils.py
mgoin: Hmm we support weights that aren't divisible by 128 for other block fp8 models fine, such as kv_a_proj in deepseek, I wonder if it is a specific fused layer

vllm/model_executor/models/mimo_v2_flash.py
```diff
@@ -0,0 +1,720 @@
...
+                if param_name == "qkv_proj" and shard_id == "v":
+                    v_scale = (
+                        self.v_scale
+                        if self.v_scale is not None
+                        else getattr(self.config, "attention_value_scale", None)
+                    )
+                    if v_scale is not None and (
+                        name.endswith("weight_scale_inv") or name.endswith(".bias")
+                    ):
+                        loaded_weight *= float(v_scale)
```
mgoin, line 618: I don't think this is valid. When I revert this to apply `v = v * self.v_scale` before attention in forward pass, I see my gsm8k eval score improve from ~74% to 78%

PR #30846, opened 2025-12-17
[BugFix] Fix logprobs with spec decode and modified logits

njhill:
"raw" logprobs/logits mode does not currently work properly with spec decoding.

The logits need to be cloned in this case before being modified in-place by sampling parameters, but this wasn't being done in the rejection sampler.


tests/v1/sample/test_logprobs.py
```diff
@@ -558,14 +565,17 @@ def test_spec_decode_logprobs(
         seed=42,
         logprobs_mode=logprobs_mode,
         gpu_memory_utilization=0.4,
+        enable_prefix_caching=False,
```
benchislett, line 568: Why disable prefix caching for this test?

benchislett (approved):
LGTM

tests/v1/sample/test_logprobs.py
njhill: well spotted :) .. prefix caching seems to change the numerics such that the logprobs vary beyond the tolerance in the test. I started to investigate this but found that setting the dtype to float32 recovers things. So I think it's just down to precision-related batch invariance.

micah-wil: Hi @njhill, actually I noticed that setting `enable_prefix_caching=False` is causing this test to fail on ROCm now. I think your speculation about precision-related batch invariance is probably right. I set `dtype="float32"` and it passes on ROCm (also passes if I re-enable prefix caching). Do you think we can run this test in float32 or can I re-enable prefix caching conditioned on if platform is ROCm? Or do you think I need to dig deeper into what is really going on here that's causing the disparity between CUDA and ROCm? Would appreciate your thoughts, thanks!

njhill: Thanks @micah-wil, that's pretty odd that it would fail the opposite way. I guess it would be interesting to understand really what's going on with the difference, but I think it should also be fine to change this test to use fp32.

PR #30849, opened 2025-12-17
[Misc][Benchmark]  use pid tracking, replaced pkill commands

RuixiangMa:
## Purpose
Problem ：pkill commands could accidentally kill other vllm services
Solution: pid-based process tracking and safe termination，which is zero risk to other services

## Test Result
截屏2025-12-17 15 52 24


RuixiangMa:
> FYI we now have `vllm bench sweep serve_sla` which should be more powerful than this script, feel free to try it out!
> 
> https://docs.vllm.ai/en/latest/benchmarking/sweeps/#sla-auto-tuner

Got it—ths for the heads-up, will try it out

PR #30856, opened 2025-12-17
Fix changes the behavior from overriding kv_connector_extra_config to updating

akakakakakaa:
When using pooling models together with LMCache, the system crashes due to a missing guard in the LMCache implementation: it does not check whether sampling_params is None. Pooling models do not use sampling parameters, so this leads to a runtime error. ([LMCache implementation](https://github.com/LMCache/LMCache/blob/b1b05c60c0df8b2899ed3923163d0769bf0befd3/lmcache/integration/vllm/vllm_v1_adapter.py#L1541), [vLLM native implementation (maybe works correctly)](https://github.com/vllm-project/vllm/blob/0cd5353644d3d045ab33c7e8e19c182bfd7db911/vllm/distributed/kv_transfer/kv_connector/v1/lmcache_integration/vllm_v1_adapter.py#L1164))

To work around this, I attempted to use vLLM’s native LMCache implementation, which already handles this case correctly. However, there is a configuration issue that prevents enabling the native implementation because of [overriding](https://github.com/vllm-project/vllm/blob/0cd5353644d3d045ab33c7e8e19c182bfd7db911/vllm/config/vllm.py#L505).

So, the fix changes the behavior from overriding kv_connector_extra_config to updating it, ensuring that user-provided options such as use_native are preserved.


yewentao256 (approved):
LGTM, thanks for the work!

yewentao256 (commented):
Please merge from main and fix the DCO issue

benchislett (approved):

PR #30864, opened 2025-12-17
[Model] Nemotron Parse 1.1 Support

### LINKED ISSUE #29292: [Feature]: support Nemotron parse-v1.1

support [Nemotron parse-v1.1](https://huggingface.co/nvidia/NVIDIA-Nemotron-Parse-v1.1)

### Alternatives

_No response_

### Additional context

_No response_

amitz-nv:
## Purpose

- Add support for NVIDIA Nemotron Parse 1.1 (HF name: `nvidia/NVIDIA-Nemotron-Parse-v1.1`)
  - Adapted from https://github.com/amalad/vllm/blob/nemotron_parse/vllm/model_executor/models/nemotron_parse.py that's based on https://huggingface.co/nvidia/NVIDIA-Nemotron-Parse-v1.1/blob/main/hf_nemotron_parse_modeling.py
  - Bart classes based on old vLLM codebase: https://github.com/vllm-project/vllm/blob/v0.10.2/vllm/model_executor/models/bart.py
- Updated RADIO model
  -  Changed its `forward` method return value from `features: Tensor` to a tuple of two `Tensor`s `(summary, features)` like its HF implementation.
  - Added some configuration values to `RadioConfig` that exist in its HF config and are required for Nemotron Parse.
  - Renamed some configuration values in `RadioConfig` to match its HF config.
- Updated RADIO model test and all of its uses accordingly (only NanoNemotronVL uses it).

CLOSE #29292

## Test Plan
Added `tests/models/multimodal/generation/test_nemotron_parse.py` that verifies HF output is the same as VLLM output.

## Test Result

---
<details>

</details>

### DIFF
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -52,6 +52,3 @@ openai-harmony >= 0.0.3  # Required for gpt-oss
 anthropic == 0.71.0
 model-hosting-container-standards >= 0.1.10, < 1.0.0
 mcp
-albumentations # Required for NemotronParse
-timm # Required for NemotronParse
-open_clip_torch # Required for NemotronParse
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -52,3 +52,6 @@ openai-harmony >= 0.0.3  # Required for gpt-oss
 anthropic == 0.71.0
 model-hosting-container-standards >= 0.1.10, < 1.0.0
 mcp
+albumentations # Required for NemotronParse
+timm # Required for NemotronParse
+open_clip_torch # Required for NemotronParse
--- a/vllm/model_executor/models/nemotron_parse.py
+++ b/vllm/model_executor/models/nemotron_parse.py
@@ -162,10 +162,8 @@ class BartDecoderLayer(nn.Module):
     ) -> torch.Tensor:
         r"""
         Args:
-            decoder_hidden_states
-                torch.Tensor of *decoder* input embeddings.
-            encoder_hidden_states
-                torch.Tensor of *encoder* input embeddings.
+            decoder_hidden_states: torch.Tensor of *decoder* input embeddings.
+            encoder_hidden_states: torch.Tensor of *encoder* input embeddings.
         Returns:
             Decoder layer output torch.Tensor
         """
@@ -297,12 +295,9 @@ class MBartDecoderNoPos(nn.Module):
     ) -> torch.Tensor:
         r"""
         Args:
-            decoder_input_ids
-                Indices of *decoder* input sequence tokens in the vocabulary.
-                Padding will be ignored by default should you
-                provide it.
-            encoder_hidden_states:
-                Tensor of encoder output embeddings
+            decoder_input_ids: Indices of *decoder* input sequence tokens in the
+                vocabulary. Padding will be ignored by default should you provide it.
+            encoder_hidden_states: Tensor of encoder output embeddings
         Returns:
             Decoder output torch.Tensor
         """
@@ -494,7 +489,6 @@ class NemotronParseImageProcessor:
 
         Args:
             images: Input image(s)
-            return_tensors: Type of tensors to return
         """
         # Ensure images is a list
         if not isinstance(images, list):
@@ -902,12 +896,9 @@ class NemotronParseForConditionalGeneration(nn.Module, SupportsMultiModal):
     ) -> torch.Tensor:
         r"""
         Args:
-            input_ids
-                torch.Tensor of *decoder* input token ids.
-            positions
-                torch.Tensor of *decoder* position indices.
-            encoder_outputs
-                List of encoder output tensors (vision embeddings).
+            input_ids: torch.Tensor of *decoder* input token ids.
+            positions: torch.Tensor of *decoder* position indices.
+            encoder_outputs: List of encoder output tensors (vision embeddings).
                 During profiling, this may be None or empty.
         Returns:
             Output torch.Tensor
--- a/vllm/model_executor/models/nemotron_parse.py
+++ b/vllm/model_executor/models/nemotron_parse.py
@@ -31,6 +31,7 @@ from vllm.attention.backends.abstract import AttentionType
 from vllm.config import CacheConfig, VllmConfig
 from vllm.config.lora import LoRAConfig
 from vllm.config.multimodal import BaseDummyOptions
+from vllm.logger import init_logger
 from vllm.model_executor.layers.activation import get_act_fn
 from vllm.model_executor.layers.linear import ColumnParallelLinear, RowParallelLinear
 from vllm.model_executor.layers.logits_processor import LogitsProcessor
@@ -64,6 +65,7 @@ from vllm.transformers_utils.configs.radio import RadioConfig
 from vllm.transformers_utils.tokenizer import AnyTokenizer
 from vllm.utils.tensor_schema import TensorSchema, TensorShape
 
+logger = init_logger(__name__)
 DEFAULT_FINAL_IMAGE_SIZE = (2048, 1648)
 
 
@@ -145,12 +147,14 @@ class BartDecoderLayer(nn.Module):
             ffn_intermediate_size,
             bias=ffn_has_bias,
             quant_config=quant_config,
+            prefix=f"{prefix}.fc1",
         )
         self.fc2 = RowParallelLinear(
             ffn_intermediate_size,
             ffn_hidden_size,
             bias=ffn_has_bias,
             quant_config=quant_config,
+            prefix=f"{prefix}.fc2",
         )
 
         self.final_layer_norm = nn.LayerNorm(self.embed_dim)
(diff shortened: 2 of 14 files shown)


requirements/common.txt
```diff
@@ -52,3 +52,6 @@ openai-harmony >= 0.0.3  # Required for gpt-oss
 anthropic == 0.71.0
 model-hosting-container-standards >= 0.1.10, < 1.0.0
 mcp
+albumentations # Required for NemotronParse
```
DarkLight1337: We shouldn't include model-specific dependencies in vLLM. HF will tell the user to install them anyway

vllm/model_executor/models/registry.py
```diff
@@ -418,6 +418,10 @@
     "VoxtralForConditionalGeneration": ("voxtral", "VoxtralForConditionalGeneration"),  # noqa: E501
     # [Encoder-decoder]
     "WhisperForConditionalGeneration": ("whisper", "WhisperForConditionalGeneration"),  # noqa: E501
+    "NemotronParseForConditionalGeneration": (
```
DarkLight1337, line 431: Please keep this in alphabetical order

DarkLight1337 (commented):
And tbh I am not sure about whether we can accept this PR. @WoosukKwon said before that we should drop all encoder-decoder models except for Whisper. Maybe it's better to keep this as OOT model.

cc @ywang96 @NickLucche

vllm/model_executor/models/registry.py
amitz-nv: Fixed within the "[Encoder-decoder]" section

requirements/common.txt
amitz-nv: Removed

vllm/model_executor/models/nemotron_parse.py
```diff
@@ -0,0 +1,952 @@
...
+        self.encoder_attn_layer_norm = nn.LayerNorm(self.embed_dim)
+
+        ffn_hidden_size = self.embed_dim
+        ffn_intermediate_size = config.encoder_ffn_dim
+        ffn_has_bias = True
+        self.fc1 = ColumnParallelLinear(
+            ffn_hidden_size,
+            ffn_intermediate_size,
+            bias=ffn_has_bias,
+            quant_config=quant_config,
```
mgoin, line 148: Make sure to add the prefix

vllm/model_executor/models/nemotron_parse.py
```diff
@@ -0,0 +1,952 @@
...
+        )
+        self.fc2 = RowParallelLinear(
+            ffn_intermediate_size,
+            ffn_hidden_size,
+            bias=ffn_has_bias,
+            quant_config=quant_config,
```
mgoin, line 155: ditto

vllm/model_executor/models/nemotron_parse.py
```diff
@@ -0,0 +1,961 @@
...
+                if pixel_values_tensor.shape[0] == 1:
+                    pixel_values_tensor = pixel_values_tensor.expand(3, -1, -1)
+
+                pixel_values.append(pixel_values_tensor)
+
+            except Exception as e:
+                print(f"Warning: Error processing image: {e}")
+                # Fallback: create a dummy tensor
+                dummy_tensor = torch.zeros((3, self.target_height, self.target_width))
+                pixel_values.append(dummy_tensor)
```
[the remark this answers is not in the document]
mgoin: Ditto, should we fail here instead? At least we should use logger.warning instead of print

vllm/model_executor/models/nemotron_parse.py
```diff
@@ -0,0 +1,952 @@
...
+        super().__init__()
+        self.config = config.encoder
+
+        self.model_encoder = self.get_vit_model_from_radio_config(
+            config, quant_config=quant_config
+        )
+
+        # Neck components
+        last_hidden_state = 1024
+        self.conv1 = nn.Conv1d(1280, last_hidden_state, 1)
```
mgoin, line 725: Can we get these values from the config?

vllm/model_executor/models/nemotron_parse.py
```diff
@@ -0,0 +1,952 @@
...
+                encoder_weights.append((".".join(name.split(".")[1:]), w))
+            elif is_decoder(name):
+                decoder_weights.append((".".join(name.split(".")[1:]), w))
+            elif is_lm_head(name):
+                trimmed_name = ".".join(name.split(".")[1:])
+                param = lm_head_dict[trimmed_name]
+                with torch.no_grad():
+                    default_weight_loader(param, w)
+            else:
+                print("Found unexpected weight: ", name)
```
mgoin: Use logger instead of print

vllm/model_executor/models/nemotron_parse.py
```diff
@@ -0,0 +1,952 @@
...
+            )
+        ]
+
+
+class RadioWithNeck(nn.Module):
+    """Vision encoder using RADIO model with custom neck."""
+
+    def __init__(
+        self,
+        config,
```
mgoin: nit: missing type

vllm/model_executor/models/nemotron_parse.py
```diff
@@ -0,0 +1,952 @@
...
+            prefix=f"{prefix}.sum_proj",
+        )
+        self.layer_norm3 = nn.LayerNorm(
+            last_hidden_state, eps=1e-06, elementwise_affine=True
+        )
+
+    def get_vit_model_from_radio_config(
+        self,
+        hf_config,
+        quant_config=None,
```
mgoin: nit: missing type

vllm/model_executor/models/nemotron_parse.py
```diff
@@ -0,0 +1,952 @@
...
+
+        radio_config = RadioConfig(
+            model_name=model_name,
+            image_size=hf_config.image_size,
+            **hf_config_vision.args,
+        )
+
+        return RadioModel(config=radio_config, quant_config=quant_config)
+
+    def forward(self, pixel_values, **kwargs):
```
mgoin: nit: missing type

mgoin (commented):
It seems this PR has sat long enough and not been blocked by others. I think the changes are fairly separable from the rest of the code, only touching sections that are also used by nano_nemotron_vl.py, which the NVIDIA team also manages. Given this, I think it is a valid change. Please just address the nits I have first and then I'll make CI ready

vllm/model_executor/models/nemotron_parse.py
amitz-nv: I don't think so, I'm relying on:
- HF config: https://huggingface.co/nvidia/NVIDIA-Nemotron-Parse-v1.1/blob/main/config.json
- HF implementation: https://huggingface.co/nvidia/NVIDIA-Nemotron-Parse-v1.1/blob/main/hf_nemotron_parse_modeling.py

In the HF implementation these values are also hardcoded, and when looking into the config:
-  I don't see `1280` at all
- I don't think the keys that have `1024` as a value are related (`decoder.d_model`, `decoder. hidden_size` and `encoder.args.teachers[3].input_size`)

vllm/model_executor/models/nemotron_parse.py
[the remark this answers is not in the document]
amitz-nv: I think you're right, removed the try-except from here

vllm/model_executor/models/nemotron_parse.py
amitz-nv: Done

vllm/model_executor/models/nemotron_parse.py
amitz-nv: Done

vllm/model_executor/models/nemotron_parse.py
amitz-nv: Done

vllm/model_executor/models/nemotron_parse.py
amitz-nv: Done

vllm/model_executor/models/nemotron_parse.py
amitz-nv: Done

vllm/model_executor/models/nemotron_parse.py
amitz-nv: Done

amitz-nv:
@mgoin Thanks, after locally adding the entry to `tests/models/registry.py` the test fails on `import albumentations as A` (in `vllm/model_executor/models/nemotron_parse.py`). I originally added this dependency to vllm in this PR but I was asked to remove it as it's specific for this model.
Should I move the failing import line to inside the code (like inside `NemotronParseImageProcessor` class), so loading the model class won't fail when `albumentations` is not installed? or do you have a different suggestion?

mgoin:
@amitz-nv If we aren't adding the dep by default, then our usual pattern is to lazily import it when required and raise an error if not found for the user to install the dep. For example look at this
https://github.com/vllm-project/vllm/blob/0d4044edd85de30d7d4558aeea4d1e95c7c556d6/vllm/model_executor/layers/quantization/utils/mxfp4_utils.py#L122-L129

mgoin (approved):
Thanks for the nice work!

PR #30870, opened 2025-12-17
Remove obsolete token chunking in fused MoE kernel

KonstGolfi:
**### Fixes #30620 Removing obsolete token chunking in fused MoE kernel**

- [ ] The chunking logic (~65k tokens) in fused_experts_impl was introduced to avoid IMA issues prior to chunked prefill. With chunked prefill now enforced upstream, this code path is no longer reachable. Removing it eliminates dead code and simplifies the fused MoE execution path without changing behavior.
- [ ] Test plan commands;
1) vllm bench throughput
2) pytest -s -v vllm/model_executor/layers/fused_moe/fused_moe.py
- [ ] Test Results

Signed-off-by: KonstGolfi <konstgolfi@gmail.com>

### DIFF
--- a/vllm/model_executor/layers/fused_moe/fused_moe.py
+++ b/vllm/model_executor/layers/fused_moe/fused_moe.py
@@ -1855,8 +1855,17 @@ def fused_experts_impl(
     top_k_num = topk_ids.size(1)
     # We execute the fused_moe kernel in chunks to circumvent this issue:
     # https://github.com/vllm-project/vllm/issues/5938
-    CHUNK_SIZE = envs.VLLM_FUSED_MOE_CHUNK_SIZE
-    M = min(num_tokens, CHUNK_SIZE)
+    
+
+    max_tokens = envs.VLLM_FUSED_MOE_CHUNK_SIZE
+    if num_tokens > max_tokens:
+        raise RuntimeError(
+            f"fused_experts_impl received {num_tokens} tokens, which exceeds the supported limit of {max_tokens}. Internal chunking has been removed. "
+            "Ensure inputs are chunked upstream (e.g. via chunked prefill)."
+        )
+
+    M = num_tokens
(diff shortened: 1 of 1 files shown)


vllm/model_executor/layers/fused_moe/fused_moe.py
```diff
@@ -1855,8 +1855,17 @@ def fused_experts_impl(
...
     # https://github.com/vllm-project/vllm/issues/5938
-    CHUNK_SIZE = envs.VLLM_FUSED_MOE_CHUNK_SIZE
-    M = min(num_tokens, CHUNK_SIZE)
+    
+
+    max_tokens = envs.VLLM_FUSED_MOE_CHUNK_SIZE
+    if num_tokens > max_tokens:
+        raise RuntimeError(
+            f"fused_experts_impl received {num_tokens} tokens, which exceeds the supported limit of {max_tokens}. Internal chunking has been removed. "
+            "Ensure inputs are chunked upstream (e.g. via chunked prefill)."
```
pavanimajety, line 1864:
```suggestion
            "Ensure inputs are chunked upstream (e.g. via chunked prefill). Use `max-num-batched-tokens` parameter to modify the default chunked prefill size."
```

vllm/model_executor/layers/fused_moe/fused_moe.py
```diff
@@ -1855,8 +1855,17 @@ def fused_experts_impl(
     top_k_num = topk_ids.size(1)
     # We execute the fused_moe kernel in chunks to circumvent this issue:
     # https://github.com/vllm-project/vllm/issues/5938
-    CHUNK_SIZE = envs.VLLM_FUSED_MOE_CHUNK_SIZE
-    M = min(num_tokens, CHUNK_SIZE)
+
```
pavanimajety, line 1858: There are a few variables in `envs.py` that are relevant to MOE Chunking which may need to be removed as well.

pavanimajety (changes requested):
Thanks for the PR!

PR #30874, opened 2025-12-17
[CI] Fix mypy for `vllm/lora`

hmellor:
This change brings us closer to not needing a custom mypy check. Part of #26533


vllm/config/lora.py
```diff
@@ -40,7 +41,7 @@ class LoRAConfig:
     parallelism. Enabling this will use the fully sharded layers. At high
     sequence length, max rank or tensor parallel size, this is likely faster.
     """
-    max_cpu_loras: int | None = None
+    max_cpu_loras: int = None
```
DarkLight1337: How does this pass mypy?

hmellor: Not sure tbh... Maybe a rule was changed the last time we updated mypy.

It might mean that the `Field(default=None)` stuff we had been doing elsewhere is unnecessary.

vllm/lora/model_manager.py
```diff
@@ -52,7 +65,7 @@ class LoRAModelManager:
 
     def __init__(
         self,
-        model: SupportsLoRA,
+        model: SupportsLoRAModel | SupportsMultiModalLoRAModel,
```
jeejeelee: QQ: Why do we need SupportsMultiModalLoRAModel?

hmellor: Because of the following code below:

```py
    def _filter_unsupported_mm_module(self, module_name: str) -> bool:
        """
        Regarding multimodal models, vLLM currently only supports adding LoRA to
        language model. LoRA for other modules, such as the vision tower, will
        be filtered out.
        """
        if self.supports_mm:
            module_mapping: MultiModelKeys = self.model.get_mm_mapping()
        ...
```

`supports_mm` does not prove to the type checker that `get_mm_mapping` exists. Hinting the model as `SupportsMultiModalLoRAModel` does.

hmellor: I'm open to alternative solutions, but this one seemed the most explicit

jeejeelee: Feels like there's no need to add `SupportsMultiModalLoRAModel `, it's somewhat over-designed

hmellor: Ok, so shall I just explicitly check that `get_mm_mapping` exists before calling it?

jeejeelee: I think it's okay, but `support_mm` needs to meet the following conditions. Can't the type checker detect it?
```python
        # Used to indicate whether the model is a multimodal model
        self.supports_mm: bool = (
            supports_multimodal(self.model)
            # In case the model only supports LoRA for
            # text modules (e.g. ChatGLM)
            and hasattr(self.model, "get_mm_mapping")
        )
```

hmellor: > Can't the type checker detect it?

Unfortunately, no

yewentao256 (commented):
Please fix the conflicts so that we can get this landed

vllm/lora/model_manager.py
jeejeelee: > Ok, so shall I just explicitly check that `get_mm_mapping` exists before calling it?

I think this is reasonable.

hmellor:
Should be good now, I've started CI

vllm/lora/model_manager.py
```diff
@@ -580,7 +586,12 @@ def get_adapter(self, adapter_id: int) -> LoRAModel | None:
         return self._registered_adapters.get(adapter_id)
 
 
-class LoRALRUCache(AdapterLRUCache[LoRAModel]):
+class LoRAModelLRUCache(AdapterLRUCache[LoRAModel]):
+    def __init__(self, capacity: int, deactivate_lora_fn: Callable[[int], bool]):
+        super().__init__(capacity, deactivate_lora_fn)
+
+
+class NoneLRUCache(AdapterLRUCache[None]):
```
jeejeelee: Maybe naming it `LoRASlotsCache` is better?
```suggestion
class LoRASlotsCache(AdapterLRUCache[None]):
```

yewentao256 (commented):
What's the current status? Could you solve the conflicts so that we can land this?

hmellor:
Superseded by https://github.com/vllm-project/vllm/pull/41722

Issue #30776, opened 2025-12-16
[Usage]: Qwen3-omni's offline usage


Auraithm:
### Your current environment

I used the code below in vllm==0.12.0, but failed.
```
import os
import torch

from vllm import LLM, SamplingParams
from transformers import Qwen3OmniMoeProcessor
from qwen_omni_utils import process_mm_info

def build_input(processor, messages, use_audio_in_video):
    text = processor.apply_chat_template(
        messages,
        tokenize=False,
        add_generation_prompt=True,
    )
    # print(text[0])
    # print(len(text[0]))
    audios, images, videos = process_mm_info(messages, use_audio_in_video=use_audio_in_video)

    inputs = {
        'prompt': text,
        'multi_modal_data': {},
        "mm_processor_kwargs": {
            "use_audio_in_video": use_audio_in_video,
        },
    }

    if images is not None:
        inputs['multi_modal_data']['image'] = images
    if videos is not None:
        inputs['multi_modal_data']['video'] = videos
    if audios is not None:
        inputs['multi_modal_data']['audio'] = audios
    
    return inputs

if __name__ == '__main__':
    # vLLM engine v1 not supported yet
    os.environ['VLLM_USE_V1'] = '1'
    os.environ['CUDA_DEVICES'] = '0,1,2,3,4,5,6,7'

    MODEL_PATH = "Qwen3-Omni-30B-A3B-Instruct"
    llm = LLM(
            model=MODEL_PATH, trust_remote_code=True, gpu_memory_utilization=0.95,
            tensor_parallel_size=1,
            limit_mm_per_prompt={'image': 3, 'video': 3, 'audio': 3},
            max_num_seqs=8,
            max_model_len=32768,
            seed=17114,
    )

    sampling_params = SamplingParams(
        temperature=0.6,
        top_p=0.95,
        top_k=20,
        max_tokens=16384,
    )

    processor = Qwen3OmniMoeProcessor.from_pretrained(MODEL_PATH)

    conversation1 = [
        {
            "role": "user",
            "content": [
                {
                    "type": "video",
                    "video": "1.mp4",
                    "fps": 6,
                }
            ],
        }
    ]
    
    USE_AUDIO_IN_VIDEO = True

    # Combine messages for batch processing
    conversations = [conversation1]
    inputs = [build_input(processor, messages, USE_AUDIO_IN_VIDEO) for messages in conversations]
    # print(inputs[0])
[...]

PR #30884, opened 2025-12-17
[BugFix] Fix offline inference of Qwen3 omni with use_audio_in_video=True

Li-dongyang:
## Purpose

During offline inference with Qwen3-omni using vLLM 0.12, enabling use_audio_in_video leads to nonsensical generations:
 - garbled outputs
 - repetitive tokens

Investigation suggests the cause might be related to the cache logic of processor. A temporary workaround has been proposed [here](https://github.com/vllm-project/vllm/issues/30776#issuecomment-3665853090).

For background information regarding bug reproducibility, please refer to [this related issue](https://github.com/vllm-project/vllm/issues/30776).

## Current Status
developing a new example.

### DIFF
--- a/vllm/model_executor/models/qwen3_omni_moe_thinker.py
+++ b/vllm/model_executor/models/qwen3_omni_moe_thinker.py
@@ -814,7 +814,7 @@ class Qwen3OmniMoeThinkerMultiModalProcessor(
                     use_audio_in_video = False
 
         # normal case with `use_audio_in_video=False`
-        if is_update_applied:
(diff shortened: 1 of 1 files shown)


DarkLight1337 (commented):
Thanks, can you also apply https://github.com/vllm-project/vllm/issues/30776#issuecomment-3665900723?

Li-dongyang:
> Thanks, can you also apply [#30776 (comment)](https://github.com/vllm-project/vllm/issues/30776#issuecomment-3665900723)?

Certainly, I have intended to do so❤️
will add it after work.

Issue #29539, opened 2025-11-26
[Bug]: FULL_AND_PIECEWISE cudagraph mode leading to !!! in generated text


xyang16:
### Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
Collecting environment information...
==============================
        System Info
==============================
OS                           : Ubuntu 22.04.5 LTS (x86_64)
GCC version                  : (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Clang version                : Could not collect
CMake version                : version 4.1.3
Libc version                 : glibc-2.35

==============================
       PyTorch Info
==============================
PyTorch version              : 2.9.0+cu129
Is debug build               : False
CUDA used to build PyTorch   : 12.9
ROCM used to build PyTorch   : N/A

==============================
      Python Environment
==============================
Python version               : 3.12.12 (main, Oct 10 2025, 08:52:57) [GCC 11.4.0] (64-bit runtime)
Python platform              : Linux-6.8.0-1021-aws-x86_64-with-glibc2.35

==============================
       CUDA / GPU Info
==============================
Is CUDA available            : True
CUDA runtime version         : 12.9.86
CUDA_MODULE_LOADING set to   : 
GPU models and configuration : 
GPU 0: NVIDIA H200
GPU 1: NVIDIA H200
GPU 2: NVIDIA H200
GPU 3: NVIDIA H200
GPU 4: NVIDIA H200
GPU 5: NVIDIA H200
GPU 6: NVIDIA H200
GPU 7: NVIDIA H200

Nvidia driver version        : 550.144.03
cuDNN version                : Could not collect
HIP runtime version          : N/A
MIOpen runtime version       : N/A
Is XNNPACK available         : True

==============================
          CPU Info
==============================
Architecture:                         x86_64
CPU op-mode(s):                       32-bit, 64-bit
Address sizes:                        48 bits physical, 48 bits virtual
Byte Order:                           Little Endian
CPU(s):                               192
On-line CPU(s) list:                  0-191
Vendor ID:                            AuthenticAMD
Model name:                           AMD EPYC 7R13 Processor
CPU family:                           25
Model:                                1
Thread(s) per core:                   2
Core(s) per socket:                   48
Socket(s):                            2
Stepping:                             1
BogoMIPS:                             5300.00
[...]

PR #30887, opened 2025-12-17
[Bugfix] [Kernel] Triton attention kernels: mask out V blocks that fall outside sliding window

tdoublep:
## Purpose

There is currently a bug in the Triton attention kernels where we don't correctly mask out V blocks that fall out of the sliding window. On main, we can be reading garbage blocks (that may even contain NaN values) which will corrupt the output. This PR resolves it

Potentially fix:
- https://github.com/vllm-project/vllm/issues/29998
- https://github.com/vllm-project/vllm/issues/26480

## Test Plan

Server:
```
VLLM_ATTENTION_BACKEND=TRITON_ATTN vllm serve openai/gpt-oss-20b   --tool-call-parser openai   --enable-auto-tool-choice   --tensor-parallel-size 2 --max-num-seqs 1
```

Client:
```
while true; do curl -X POST http://localhost:8000/v1/chat/completions  -H "Content-Type: application/json"   -d '{
    "stream": false,
    "messages": [
      {
        "role": "system",
      },
      {
        "role": "user",
        "content": "Hi"
      },
      {
        "role": "assistant",
      },
      {
        "role": "user",
      }
    ],
    "tools": [
      {
        "type": "function",
        "function": {
          "parameters": {
            "type": "object",
            "properties": {
              "url": {
                "type": "string",
              }
            },
            "required": ["url"]
          }
        }
      }
    ]
  }'; done
```

On main the above hangs after the first request.

## Test Result

After, the PR, the test does not hang anymore. 

---
<details>

</details>


bbrowning:
While this does fix that curl request for me, I'm still getting a lot of repeated "!!!!" generations (token id 0) using the gsm8k lm_eval dataset on my A5500. So, I believe this helps, but does not fix all cases where this kind of things has been reported on Triton attention.

If you want to reproduce what I'm seeing (which copies what was reported in #29539), [install lm_eval](https://github.com/EleutherAI/lm-evaluation-harness?tab=readme-ov-file#install), then spin up gpt-oss-20b and run the gsm8k dataset test:

**Serve gpt-oss-20b in vLLM with TRITON_ATTN**
```
VLLM_ATTENTION_BACKEND=TRITON_ATTN \
  --tensor-parallel-size 2 \
  --max-num-seqs 16
```

**Run lm_eval**
From your locally cloned lm-evaluation-harness directory:

```
python -m lm_eval \
  --model local-completions \
  --model_args model=openai/gpt-oss-20b,base_url=http://127.0.0.1:8000/v1/completions,num_concurrent=16 \
  --tasks gsm8k \
  --log_samples --output_path /tmp/lm_eval/
```

Grep for repeated `!!!` in your generated samples - ie in my case:

```
grep \!\!\! /tmp/lm_eval/samples_gsm8k_2025-12-16T13-48-03.250538.jsonl
```

I had 56 different samples with this infinite token id 0 repeated generation when testing this change. For reference, I get zero when testing the change from #30650

tdoublep:
@Isotr0py Could you help review this since you've been working on these kernels recently?

Isotr0py (approved):
LGTM

tdoublep:
Which attention backend are you using? This fix relates specifically to Triton

PR #30890, opened 2025-12-17
[UX] Add `-ep` shorthand for `--enable-expert-parallel`

mgoin:
## Purpose

Typing `--enable-expert-parallel` is verbose and `-ep` is much better, however we do diverge from the other tp/pp/dp args that require numbers instead of just the flag. I think we can expand this in the future to optionally take an int though.

Example Usage:
```
# Before (still works)
vllm serve model -tp 2 -dp 2 --enable-expert-parallel

# After
vllm serve model -tp 2 -dp 2 -ep
```

---
<details>

</details>

### DIFF
--- a/vllm/engine/arg_utils.py
+++ b/vllm/engine/arg_utils.py
@@ -838,7 +838,9 @@ class EngineArgs:
             **parallel_kwargs["data_parallel_external_lb"],
         )
         parallel_group.add_argument(
-            "--enable-expert-parallel", **parallel_kwargs["enable_expert_parallel"]
+            "--enable-expert-parallel",
+            "-ep",
+            **parallel_kwargs["enable_expert_parallel"],
         )
         parallel_group.add_argument(
             "--all2all-backend", **parallel_kwargs["all2all_backend"]


vllm/engine/arg_utils.py
```diff
@@ -838,7 +838,9 @@ def add_cli_args(parser: FlexibleArgumentParser) -> FlexibleArgumentParser:
             **parallel_kwargs["data_parallel_external_lb"],
         )
         parallel_group.add_argument(
-            "--enable-expert-parallel", **parallel_kwargs["enable_expert_parallel"]
+            "--enable-expert-parallel",
+            "-ep",
```
chaunceyjiang: It feels a bit confusing.

I have `-tp 8` and `-dp 8`, and `-ep` might be mistakenly interpreted as also needing to be set to `-ep 8`.

chaunceyjiang: How about `--enable-ep`?

mgoin: I think consensus seemed like people want a proper expert-parallel-size now, so updated to that

yewentao256 (commented):
When I used SGL, `-ep 4` is extremely confusing especially with tp, that's one of the reason I don't like SGL

I strongly recommend not adding this flag and keep what we already have.

For some special case that people wants to adjust the size, I think we can have an env to adjust.

vllm/config/parallel.py
```diff
@@ -296,6 +301,23 @@ def _validate_parallel_config(self) -> Self:
                 "data_parallel_external_lb can only be set when data_parallel_size > 1"
             )
 
+        # Validate expert_parallel_size
+        expected_ep_size = self.tensor_parallel_size * self.data_parallel_size
```
pavanimajety: is the user still expected to give `tp` size?  I think if the user sets `ep 4 dp 4` we should be able to infer `tp` size too

tlrmchlsmth: I think we want to leave the door open to using both TP and EP in the MoE layers, in which case we’d need to specify at least three of EP, DP, MoE TP, Attn TP

tlrmchlsmth: And we’d need the user to explicitly specify TP (== Attn TP) today to give us that interface flexibility in the future

mgoin: tensor_parallel_size defaults to 1, so this will be valid for `ep 4 dp 4`

yewentao256 (changes requested):
Let's block this for a little while before we fully understand the benefits, I still don't like the SGL confusing style command config, it hurts user experience a lot since not every user can fully understand the ep size and make use of them.

yewentao256 (approved):
LGTM, thanks for the work!

PR #30894, opened 2025-12-17
Improve DCP error message with actionable guidance

Dhruv-80:
## Purpose

Improve the error message shown when Decode Context Parallelism (DCP) is enabled
with an attention backend that does not support returning softmax LSE values
during decode.

Previously, this case raised a generic assertion error that did not provide
actionable guidance to users. This change replaces the assertion with a
user-facing exception that clearly explains the limitation and suggests how to
resolve it (e.g., by selecting a different attention backend or disabling DCP).

Fixes: #28407

## Test Plan

- Reviewed the updated error message logic in
  `vllm/v1/worker/cp_utils.py`.
- Verified that the exception is raised with a clear, actionable message when an
  unsupported attention backend is used with DCP enabled.

## Test Result

- The error message now clearly indicates why DCP is not supported for the
  selected attention backend.
- The message provides actionable guidance on how to resolve the issue by
  changing `VLLM_ATTENTION_BACKEND` or disabling DCP.

### DIFF
--- a/docs/usage/README.md
+++ b/docs/usage/README.md
@@ -7,3 +7,42 @@ Then, vLLM supports the following usage patterns:
 - [Inference and Serving](../serving/offline_inference.md): Run a single instance of a model.
 - [Deployment](../deployment/docker.md): Scale up model instances for production.
 - [Training](../training/rlhf.md): Train or fine-tune a model.
+
+## OpenAI-Compatible API Usage
+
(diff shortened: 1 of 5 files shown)


Dhruv-80:
It looks like this change has already been addressed upstream.
Closing this PR to avoid duplication.

Thanks for the review and feedback!

Issue #26946, opened 2025-10-15
[RFC]: Enable libtorch-ABI-stable vLLM cuda wheels


janeyx99:
### Motivation.

PyTorch is building out a limited ABI through which custom ops writers can build extensions that build with one torch version and run on another. vLLM is an important part of the ecosystem that depends on torch. We'd like to help enable building vLLM wheels that would not be tied to a specific torch version to promote build system simplicity and general developer experience.

### Proposed Change.

As PyTorch increases our ABI surface, we'd love vLLM's collaboration in testing out our stable APIs. We've gone through the repo and gathered the following torch APIs that we are working to make stable. The hope is that once we have the stable versions of these ops, we can build vLLM cuda wheels in a libtorch stable manner, similar to what we did in FA3 here: https://github.com/Dao-AILab/flash-attention/pull/1791

The APIs:
- [ ] Various TORCH_LIBRARY registrations [[link](https://github.com/vllm-project/vllm/blob/main/csrc/torch_bindings.cpp)]
- [ ] at::Tag [[link](https://github.com/vllm-project/vllm/blob/1c63a16b653d1f3f4260e862f01eaffed283c4c8/csrc/torch_bindings.cpp#L30)]
- [ ] Device stuff
    - [ ] torch::kDevice
    - [ ] at::Device
    - [ ] t.device().type()
    - [ ] Get_device
- [ ] SymInt [[link](https://github.com/vllm-project/vllm/blob/main/csrc/torch_bindings.cpp#L304)]
    - We're wondering if we can replace these with ints and use Python meta functions to support compile
- [ ] Tensor[] [[link](https://github.com/vllm-project/vllm/blob/main/csrc/torch_bindings.cpp#L780)] and Int[] [[link](https://github.com/vllm-project/vllm/blob/main/csrc/torch_bindings.cpp#L747)]
- [ ] Various AT_DISPATCH macros [[link](https://github.com/vllm-project/vllm/blob/main/csrc/dispatch_utils.h)]
    - [ ] AT_DISPATCH_SWITCH, AT_DISPATCH_CASE, AT_DISPATCH_BYTE_CASE
- [ ] at::get_num_threads
- [ ] at::cuda::OptionalCUDAGuard
- [ ] at::cuda::getCurrentCUDAStream()
- [ ] torch::from_blob
- [ ] torch::full
- [ ] tensor.index(torch::indexing::Slice(...))
- [ ] torch::empty(..., TensorOptions.device(...).dtype(...))
[...]

PR #30908, opened 2025-12-17
[1/n] Migrate activation kernels to libtorch stable ABI

mikaylagawarecki:
## Purpose

**This change requires 2.10+ to build**

First step towards migration of the cuda wheel to the libtorch stable ABI see https://github.com/vllm-project/vllm/issues/26946. **The benefit of migrating to the libtorch stable ABI is that the stable binary will be able to be built once (with torch 2.10 or greater) and can run in all versions of torch >= 2.10.**

1. `activation_kernels.cu` is migrated to use the stable ABI/API
2. Creates `csrc/torch_bindings_stable.cpp` that registers kernels in activation_kernels.cu via `TORCH_STABLE_LIBRARY_FRAGMENT` to the `_C` namespace
3. Creates a new `_C_stable.so` in addition to the existing `_C.so` that has sources `csrc/activation_kernels.cu`
  `csrc/torch_bindings_stable.cpp` 
  
**I am looking for feedback on the following**

1. The PR separates the stable kernels into a new `_C_stable` target, is this ok?
         a. The rationale is that this extension is built with -DTORCH_TARGET_VERSION=... (ensuring only stable APIs are used)
         b. Kernels in this extension are still registered to the _C namespace for backward compat.
2. The PR requires torch version >=2.10 in order to build `_C_stable`, I'm wondering whether that is alright
          a.  It can be built once with any version >= 2.10 and will be able to run in all versions of torch >= 2.10

## Test Plan
`pytest tests/kernels/core/test_activation.py` 

## Test Result

Screenshot 2025-12-17 at 5 47 19 PM
---
<details>

</details>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -288,7 +288,6 @@ set(VLLM_EXT_SRC
   "csrc/attention/merge_attn_states.cu"
   "csrc/attention/vertical_slash_index.cu"
   "csrc/pos_encoding_kernels.cu"
-  "csrc/activation_kernels.cu"
   "csrc/layernorm_kernels.cu"
   "csrc/fused_qknorm_rope_kernel.cu"
   "csrc/layernorm_quant_kernels.cu"
@@ -959,6 +958,42 @@ define_extension_target(
 # driver API. This causes problems when linking with earlier versions of CUDA.
 # Setting this variable sidesteps the issue by calling the driver directly.
 target_compile_definitions(_C PRIVATE CUTLASS_ENABLE_DIRECT_CUDA_DRIVER_CALL=1)
+
+
+if(VLLM_GPU_LANG STREQUAL "CUDA" OR VLLM_GPU_LANG STREQUAL "HIP")
+  #
+  # _C_stable_libtorch extension (ops registered via STABLE_TORCH_LIBRARY)
+  #
+  set(VLLM_STABLE_EXT_SRC
+    "csrc/stable/activation_kernels.cu"
+    "csrc/stable/torch_bindings.cpp")
+
+  if(VLLM_GPU_LANG STREQUAL "CUDA")
+    set_gencode_flags_for_srcs(
+      SRCS "${VLLM_STABLE_EXT_SRC}"
+      CUDA_ARCHS "${CUDA_ARCHS}")
+  endif()
+
+  message(STATUS "Enabling C_stable extension.")
+  define_extension_target(
+    _C_stable_libtorch
+    DESTINATION vllm
+    LANGUAGE ${VLLM_GPU_LANG}
+    SOURCES ${VLLM_STABLE_EXT_SRC}
+    COMPILE_FLAGS ${VLLM_GPU_FLAGS}
+    ARCHITECTURES ${VLLM_GPU_ARCHES}
+    USE_SABI 3
+    WITH_SOABI)
+
+  # Set TORCH_TARGET_VERSION for stable ABI compatibility.
+  # This ensures we only use C-shim APIs available in PyTorch 2.10+.
+  target_compile_definitions(_C_stable_libtorch PRIVATE
+    TORCH_TARGET_VERSION=0x020A000000000000ULL)
+
+  # Needed to use cuda APIs from C-shim
+  target_compile_definitions(_C_stable_libtorch PRIVATE
+    USE_CUDA)
+endif()
 
 #
 # _moe_C extension
--- a/cmake/utils.cmake
+++ b/cmake/utils.cmake
@@ -541,7 +541,25 @@ function (define_extension_target MOD_NAME)
   if (ARG_LANGUAGE STREQUAL "CUDA")
     target_link_libraries(${MOD_NAME} PRIVATE torch CUDA::cudart CUDA::cuda_driver ${ARG_LIBRARIES})
   else()
-    target_link_libraries(${MOD_NAME} PRIVATE torch ${TORCH_LIBRARIES} ${ARG_LIBRARIES})
+    # Link against PyTorch's bundled libtorch_hip.so (for DeviceGuard registration)
+    # and libamdhip64.so (to share a single HIP runtime with PyTorch).
+    find_library(TORCH_HIP_LIBRARY torch_hip PATHS "${TORCH_INSTALL_PREFIX}/lib" NO_DEFAULT_PATH)
+    find_library(TORCH_AMDHIP64_LIBRARY amdhip64 PATHS "${TORCH_INSTALL_PREFIX}/lib" NO_DEFAULT_PATH)
+
+    set(_hip_libs)
+    if (TORCH_HIP_LIBRARY)
+      list(APPEND _hip_libs ${TORCH_HIP_LIBRARY})
+    endif()
+    if (TORCH_AMDHIP64_LIBRARY)
+      list(APPEND _hip_libs ${TORCH_AMDHIP64_LIBRARY})
+      # Ensure PyTorch's bundled libamdhip64.so is found at runtime, not system ROCm's.
+      set(_torch_lib_dir "${TORCH_INSTALL_PREFIX}/lib")
+      set_target_properties(${MOD_NAME} PROPERTIES
+        BUILD_RPATH "${_torch_lib_dir}"
+        INSTALL_RPATH "${_torch_lib_dir}")
+    endif()
+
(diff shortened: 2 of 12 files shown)


csrc/ops.h
```diff
@@ -142,8 +142,6 @@ void rotary_embedding(torch::Tensor& positions, torch::Tensor& query,
                       std::optional<torch::Tensor> key, int64_t head_size,
                       torch::Tensor& cos_sin_cache, bool is_neox);
 
-void silu_and_mul(torch::Tensor& out, torch::Tensor& input);
-
 void silu_and_mul_quant(torch::Tensor& out, torch::Tensor& input,
```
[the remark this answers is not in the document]
mikaylagawarecki: Included stable_ops.h in the relevant file

mikaylagawarecki:
cc @youkaichao @mgoin @FlamingoPg @simon-mo

Harry-Chen:
@mikaylagawarecki 

> Indeed, the final switch of building a libtorch ABI stable wheel would have to wait till all the relevant files are only using stable APIs, but the progress does not need to be all or nothing. Does it make sense to you if I continue this enablement with a stack that migrates files one by one?

Yes this totally makes sense. So our target is to move all c++ compilation units to `VLLM_STABLE_EXT_SRC`. Do you have any evaluation on this, e.g. can we finish this with PyTorch 2.10, or will we be blocked by some APIs that do not exist yet?

> To clarify what I meant here, I meant that the stable library in `C_stable` registers a `TORCH_LIBRARY_FRAGMENT` with the namespace `_C`. This ensures that the function is still callable from python as `torch._C.silu_and_mul`.
> 
> I don't think there are duplicate symbols because I removed the registrations from the respective TORCH_LIBRARY in the `_C` target

I got it now. Thank you for explaining!

> > We could do this after the release of PyTorch 2.10. I think that is about ~1 month later. And during this period we could run build and test on nightly pytorch
> 
> That sounds good! Is there a guide on how to enable this in CI?

Maybe you can refer to <https://docs.vllm.ai/en/latest/contributing/ci/update_pytorch_version/#test-pytorch-release-candidates-rcs>. And CC @youkaichao and @khluu for more ideas here.

csrc/ops.h
```diff
@@ -399,6 +380,15 @@ void free_shared_buffer(int64_t buffer);
 
 torch::Tensor hadacore_transform(torch::Tensor& x, bool inplace);
 
+// CPU activation function declarations
+// These use regular torch::Tensor (not stable ABI) for CPU implementations
+void silu_and_mul(torch::Tensor& out, torch::Tensor& input);
+void gelu_and_mul(torch::Tensor& out, torch::Tensor& input);
```
ProExpertProg: Missing `mul_and_silu`? Also why not moving CPU as well?

ProExpertProg (commented):
I think instead of having a bunch of `*_stable.*` files, we should just start a new stable directory.

I see the original high-level RFC, but now that we're doing this, could you actually create a tracking issue with an overall plan? That would make it easier to discuss some of these small (but impactful) decisions.

ProExpertProg:
I just worry that instead of migrating everything and keeping cpu separate will increase complexity more than the extra work necessary to just do both. But I'll defer to you. 

I think the stable folder would be cleaner, even if just for the headers. Could you try moving the activation files and see if GitHub deals with it well (I feel like it will)?

cmake/utils.cmake
```diff
@@ -541,7 +541,25 @@ function (define_extension_target MOD_NAME)
   if (ARG_LANGUAGE STREQUAL "CUDA")
     target_link_libraries(${MOD_NAME} PRIVATE torch CUDA::cudart CUDA::cuda_driver ${ARG_LIBRARIES})
   else()
-    target_link_libraries(${MOD_NAME} PRIVATE torch ${TORCH_LIBRARIES} ${ARG_LIBRARIES})
+    # Link against PyTorch's bundled libtorch_hip.so (for DeviceGuard registration)
```
mikaylagawarecki, line 544: On a RoCM machine, I found that I needed these changes to make sure that vllm was linking to these two .so from torch (in particular the libamd.so hip headers packaged by torch) otherwise the stable DeviceGuard would not work correctly

It seemed that 2 separate hipContexts were created by the raw hip calls that vllm did (which were from a hip header from elsewhere) and the libtorch shims that called raw hip APIs, which used the hip headers packaged by torch

csrc/ops.h
```diff
@@ -165,8 +165,6 @@ void persistent_masked_m_silu_mul_quant(
     at::Tensor& y_s,           // (E, T, H//group_size) [OUT]
     bool use_ue8m0);
 
-void mul_and_silu(torch::Tensor& out, torch::Tensor& input);
```
mikaylagawarecki, line 168: Since it might be confusing why this is deleted -- I kept the other declarations because the cpu torch_bindings.cpp includes ops.h. All the other ops are also defined for CPU as well, but this op isn't defined for CPU so its declaration is deleted

CMakeLists.txt
```diff
@@ -959,6 +959,42 @@ define_extension_target(
...
+    SOURCES ${VLLM_STABLE_EXT_SRC}
+    COMPILE_FLAGS ${VLLM_GPU_FLAGS}
+    ARCHITECTURES ${VLLM_GPU_ARCHES}
+    USE_SABI 3
+    WITH_SOABI)
+
+  # Set TORCH_TARGET_VERSION for stable ABI compatibility.
+  # This ensures we only use C-shim APIs available in PyTorch 2.10+.
+  target_compile_definitions(_C_stable_libtorch PRIVATE
+    TORCH_TARGET_VERSION=0x020A000000000000ULL)
```
zou3519, line 991: nit: comment that "_C_stable_libtorch is abi compatible with PyTorch >= TORCH_TARGET_VERSION which is currently set to 2.10". (explicitly state what we get)

cmake/utils.cmake
zou3519: @mikaylagawarecki I'm not sure what's going on here. Do all extensions need this code? Or is vLLM doing something weird?

csrc/stable/dispatch_utils.h
```diff
@@ -0,0 +1,25 @@
+/*
+ * Stable ABI compatible dispatch utilities for vLLM.
```
zou3519, line 2: nit: Whenever you add "stable abi" in a comment you probably want to call it "libtorch stable abi" to clarify

csrc/stable/torch_bindings.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+#include <torch/csrc/stable/library.h>
+
+// Register ops using STABLE_TORCH_LIBRARY for stable ABI compatibility.
+// Note: We register under namespace "_C" so ops are accessible as
+// torch.ops._C.<op_name> for compatibility with existing code.
+STABLE_TORCH_LIBRARY_FRAGMENT(_C, m) {
+  // Activation ops
+  // Activation function used in SwiGLU.
+  m.def("silu_and_mul(Tensor! result, Tensor input) -> ()");
```
zou3519, line 12: Just to check... you are able to add tags to operators in the stable abi?

csrc/stable/torch_utils.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include <torch/csrc/inductor/aoti_torch/c/shim.h>
+#include <torch/headeronly/util/Exception.h>
+
+#include <cuda_runtime.h>
+#include <deque>
+#include <mutex>
+#include <string>
+#include <vector>
+
```
zou3519, line 11: there's no namespacing at all, is that intentional?

zou3519: looks like [vllm](https://github.com/mikaylagawarecki/vllm/tree/e8d195772b11503ff2b447b53fe636910f25013b)/[csrc](https://github.com/mikaylagawarecki/vllm/tree/e8d195772b11503ff2b447b53fe636910f25013b/csrc)
/ops.h has no namespacing in the first place lol

csrc/stable/torch_bindings.cpp
mikaylagawarecki: Do you mean  adding `at::Tag` to the registration? that is currently not ABI stable (but I don't believe at::Tag is used within the repo atm)

Harry-Chen:
Superseded by newer PRs.

PR #29558, opened 2025-11-27
[Perf] Enable cuda graph for deepepHT, 5.3% throughput improvement, 4.4% TTFT improvement

yewentao256:
## Purpose

We cancel the update in https://github.com/vllm-project/vllm/pull/25093 because of the cache hit issue, and seems already be fixed in main, now we can enable cudagraph as default for deepEPHT with the split of moe ops.

## Test

`vllm serve deepseek-ai/DeepSeek-V3.1 -dp 8  --enable-expert-parallel --port 9256`

### Acc

`lm_eval --model local-completions --model_args "base_url=http://127.0.0.1:9256/v1/completions,model=deepseek-ai/DeepSeek-V3.1,num_concurrent=1024" --tasks gsm8k`

```bash
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9568|±  |0.0056|
|     |       |strict-match    |     5|exact_match|↑  |0.9575|±  |0.0056|
```

### Perf

`vllm bench serve --model deepseek-ai/DeepSeek-V3.1  --dataset-name random --host 127.0.0.1 --port 9256 --random-input-len 2 --random-output-len 256 --request-rate inf --num-prompts 1024`

```bash
# now
============ Serving Benchmark Result ============
Successful requests:                     1024      
Failed requests:                         0         
Benchmark duration (s):                  34.20     
Total input tokens:                      1024      
Total generated tokens:                  262144    
Request throughput (req/s):              29.94     
Output token throughput (tok/s):         7665.18   
Peak output token throughput (tok/s):    8192.00   
Peak concurrent requests:                1024.00   
Total Token throughput (tok/s):          7695.12   
---------------Time to First Token----------------
Mean TTFT (ms):                          1195.77   
Median TTFT (ms):                        1248.90   
P99 TTFT (ms):                           1272.42   
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          129.00    
Median TPOT (ms):                        128.99    
P99 TPOT (ms):                           129.19    
---------------Inter-token Latency----------------
Mean ITL (ms):                           129.00    
Median ITL (ms):                         129.15    
P99 ITL (ms):                            167.09    
==================================================

# main (cuda graph be set to None)
============ Serving Benchmark Result ============
Successful requests:                     1024      
Failed requests:                         0         
[...]

PR #30910, opened 2025-12-17
[BugFix] Partial revert of #29558 (DeepEP HT + PIECEWISE CG support)

LucasWilkinson:
Partially revert https://github.com/vllm-project/vllm/pull/29558 as this broke H200 tests

https://buildkite.com/vllm/ci/builds/43863#019b29e9-5c1b-4eff-83f7-c8304f774aa7

i.e.

```
VLLM_ALL2ALL_BACKEND=deepep_high_throughput VLLM_USE_DEEP_GEMM=1 VLLM_LOGGING_LEVEL=DEBUG python3 examples/offline_inference/data_parallel.py --model Qwen/Qwen1.5-MoE-A2.7B --tp-size=1 --dp-size=2 --max-model-len 2048
```

There seems to be multiple issues here so we will try to follow up with a proper fix to restore the PIECEWISE CG support,

1) torch.compile does not support Size as output meaning the pattern:

```
orig_shape = hidden_states.shape
...
final_hidden_states = self.experts(              <=== Splitting op!!!!
   hidden_states=hidden_states, router_logits=router_logits
)
...
return final_hidden_states.view(orig_shape)
```

breaks torch.compile and is common in many MoE model definitions

doing:
```
final_hidden_states = self.experts(              <=== Splitting op!!!!
   hidden_states=hidden_states, router_logits=router_logits
)
...
orig_shape = hidden_states.shape
return final_hidden_states.view(orig_shape)
```
can fix this but requires updating all the MoE definitions (and creates a footgun)

2) the outputs do not seem to have consistent addresses leading to garbage outputs


BoyuanFeng (approved):

yewentao256 (approved):
Thanks for catching this! We can land this first to unblock CI, and I can fix this issue thoroughly in https://github.com/vllm-project/vllm/pull/30914 later

PR #30723, opened 2025-12-15
[CI] Generalize gsm8k test args and add Qwen3-Next MTP B200 test

mgoin:
## Purpose

Expands the gsm8k test so we can save any server args into the config to go beyond just TP, and adds a Qwen3-Next MTP B200 with EP=2

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #30911, opened 2025-12-17
[AMD][CI] fix lm eval ci arg

divakar-amd:
Fixing args for AMD-CI run in accordance with https://github.com/vllm-project/vllm/pull/30723 

This fixes the following tests for AMD-CI:

> LM Eval Small Models
> LM Eval Small Models (1 Card)


tjtanaa (approved):
LGTM

PR #30915, opened 2025-12-18
[Fix][FlexAttention] return max logical block index to handle reused blocks

ivanium:
## Purpose

For FlexAttention, we need to build `physical_to_logical_mapping` to reversely map physical block ids to logical block ids. This process previously assumes logical block ids are always unique, which is not true for some attention types such as sliding window attention, where some blocks may be released and reused later, causing the same physical block id to appear multiple times in a row of `block_table` at different logical block indices. As a result, before this PR, the physical to logical mapping will be non-deterministic (most times return the first matched logical block index) and lead to errornous results as shown in the following example.

<details><summary>test_gemma3_reuse.py</summary>
<p>

```python

# Initialize LLM with LMCache configuration
# Adjust gpu_memory_utilization based on your GPU memory
# Parameters below are for 80GB GPUs
llm = LLM(
    model="google/gemma-3-4b-it",
    max_model_len=75000,
    gpu_memory_utilization=0.28,  # 80GB*0.28 = 22.4GB
    max_num_seqs=16,
    enforce_eager=True,
)

sampling_params = SamplingParams(temperature=0, top_p=0.95, max_tokens=10)

outputs = llm.generate("hi" * 70000 + "\nhow are you?", sampling_params)
generated_text = outputs[0].outputs[0].text
print(f"Generated text: {generated_text!r}")
# Wrong results: 'hihihihihihihihihihi'
# Correct results: '\nI am fine, thank you for asking!'
```

</p>
</details> 

This PR fixes this issue by always returning the latest (maximum) logical block index during this building process, given the fact that the latest/newly allocated block must be the one in use while previous ones are stale.

## Test Plan

The standard flex attention tests should pass. Additionally, this PR adds a new test case with reused blocks.

```shell
pytest -q tests/kernels/test_flex_attention.py
```

The above example `test_gemma3_reuse.py` should also pass.

## Test Result

Passed.

---
<details>

</details>

### DIFF
--- a/tests/kernels/test_flex_attention.py
+++ b/tests/kernels/test_flex_attention.py
@@ -15,7 +15,10 @@ from tests.v1.attention.utils import (
     create_standard_kv_cache_spec,
     create_vllm_config,
 )
-from vllm.v1.attention.backends.flex_attention import FlexAttentionMetadataBuilder
+from vllm.v1.attention.backends.flex_attention import (
+    FlexAttentionMetadataBuilder,
+    physical_to_logical_mapping,
+)
 
 from ..models.utils import check_embeddings_close, check_logprobs_close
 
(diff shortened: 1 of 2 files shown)


Isotr0py (approved):
LGTM, also cc @drisspg

drisspg (commented):
Looks great, thanks for fixing

PR #30917, opened 2025-12-18
[Feature] Support using FlexKV as anothor KV Cache Offloading option.

axxx03:
# Description
FlexKV is a distributed KV store and multi-level cache management system developed by Tencent Cloud's TACO team in collaboration with the community, designed for large-scale LLM inference scenarios. FlexKV leverages multi-level caching to enable inference engines to achieve higher throughput and lower latency.

In our case, when intergated with FlexKV, we can achieve the following improvement:

- ISL=21K, OSL=1K, batch_size=8: TTFT decreases by 60%, TPOT increases by 13%, and QPM increases by 16%.

### DIFF
--- a/vllm/distributed/kv_transfer/kv_connector/v1/flexkv_connector.py
+++ b/vllm/distributed/kv_transfer/kv_connector/v1/flexkv_connector.py
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
-from typing import TYPE_CHECKING, Any, Optional
+from typing import TYPE_CHECKING, Any, Iterable, Optional
 
 import torch
 from flexkv.integration.vllm.vllm_v1_adapter import FlexKVConnectorV1Impl
@@ -14,6 +14,7 @@ from vllm.v1.outputs import KVConnectorOutput
 
 if TYPE_CHECKING:
     from vllm.attention.backends.abstract import AttentionMetadata
+    from vllm.distributed.kv_events import KVCacheEvent
     from vllm.forward_context import ForwardContext
     from vllm.v1.core.kv_cache_manager import KVCacheBlocks
     from vllm.v1.request import Request
@@ -194,3 +195,17 @@ class FlexKVConnectorV1(KVConnectorBase_V1):
             returned by the engine.
         """
         return self._flexkv_connector.request_finished(request, block_ids)
+
+
+class FlexKVDynamoConnectorV1(FlexKVConnectorV1):
+    '''
+    Use this connector for combined use of Dynamo, vLLM and FlexKV.
+    '''
+    def take_events(self) -> Iterable["KVCacheEvent"]:
+        '''
+        Collect buffered KV cache events.
+
+        Returns:
+            New KV cache events since the last call.
+        '''
+        return self._flexkv_connector.take_events()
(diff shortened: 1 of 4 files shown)


vllm/distributed/kv_transfer/kv_connector/v1/flexkv_connector.py
```diff
@@ -0,0 +1,191 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+from typing import TYPE_CHECKING, Any, Optional
+
+import torch
+from flexkv.integration.vllm.vllm_v1_adapter import FlexKVConnectorV1Impl
```
chaunceyjiang, line 6: Just a question: is flexkv an open-source project?

axxx03: Yes, our repository is available here: https://github.com/taco-project/FlexKV￼

KuntaiDu: I clicked this link and it gives me a 404

KuntaiDu (changes requested):
Ideally we shouldn't change the scheduler when adding new connector.

vllm/v1/core/sched/scheduler.py
```diff
@@ -164,6 +157,7 @@ def __init__(
         # KV Connector: requests in process of async KV loading or recving
         self.finished_recving_kv_req_ids: set[str] = set()
         self.failed_recving_kv_req_ids: set[str] = set()
+        self.sending_kv_reqs: dict[str, Request] = {}
```
KuntaiDu: Just wondering the motivation of adding this `sending_kv_reqs`

vllm/v1/core/sched/scheduler.py
```diff
@@ -1649,6 +1580,10 @@ def _update_from_kv_xfer_finished(self, kv_connector_output: KVConnectorOutput):
             schedule the request during the next step.
         """
 
+        # avoid busy checking
+        if len(self.running) == 0:
+            time.sleep(0.01)
```
KuntaiDu: Just wondering why busy checking is not ideal here? (scheduler process don't have any other things to do anyway)

vllm/distributed/kv_transfer/kv_connector/v1/flexkv_connector.py
axxx03: Sorry, the link above contains an extra Unicode character. Please use this:
https://github.com/taco-project/FlexKV

vllm/v1/core/sched/scheduler.py
axxx03: Thanks, I'll remove this.

vllm/v1/core/sched/scheduler.py
axxx03: Thanks for the review.

Here is the motivation: In KV cache offloading scenario, requests are offloaded only after decode finishes. Therefore, for the final request, when it start offloading KV cache, the scheduler may think there’s no ongoing task and switch to idle state. So, `update_from_output` wouldn’t be called for that task to refresh its state. 

After review the code, we find it’s OK to remove `sending_kv_reqs ` and let the task state be refreshed when the next request arrives.

Therefore, we will remove this.

PR #30936, opened 2025-12-18
[v1][CP] Improve DCP/PCP/MTP error messages with actionable guidance

yurekami:
## Summary

This PR improves error messages when context parallel features (DCP/PCP/MTP) are used with incompatible attention backends.

**Problem (Fixes #28407):**
- Current error messages use raw `AssertionError` with technical jargon
- No explanation of what DCP (Decode Context Parallel), PCP (Prefill Context Parallel), or MTP are
- No list of compatible backends
- No actionable fix suggestions

**Solution:**
Replace `assert` statements with informative `RuntimeError` exceptions that include:
- Clear explanation of the feature requirement
- List of compatible attention backends
- Environment variable instructions (`VLLM_ATTENTION_BACKEND`)
- Documentation links

## Changes

- `vllm/v1/worker/cp_utils.py`: Replace 3 assertion errors with descriptive RuntimeErrors

## Example of improved error message

Before:
```
AssertionError: DCP requires attention impls to return the softmax lse for decode, but the impl FlashAttentionImpl does not return the softmax lse for decode.
```

After:
```
RuntimeError: Decode Context Parallel (DCP) requires an attention backend that supports returning softmax LSE (log-sum-exp) for decode operations. The current backend 'FlashAttentionImpl' does not support this feature.

To resolve this issue, try one of the following:
  1. Use a compatible attention backend by setting:
     export VLLM_ATTENTION_BACKEND=<backend>
     Compatible backends: FLASH_ATTN, FLASHINFER, TRITON_MLA, FLASH_MLA, FLASH_ATTN_MLA, CUTLASS_MLA
  2. Disable DCP by removing the --decode-context-parallel-size flag

For more information, see:
  https://docs.vllm.ai/en/latest/serving/distributed_serving.html
```

## Test plan

- [x] Syntax validation passes
- [x] Import verification passes
- [ ] CI tests should pass (no behavior change for valid configurations)

### DIFF
--- a/vllm/v1/worker/cp_utils.py
+++ b/vllm/v1/worker/cp_utils.py
@@ -23,20 +23,25 @@ def check_attention_cp_compatibility(vllm_config: VllmConfig) -> None:
                 continue
             if vllm_config.speculative_config is not None and interleave_size > 1:
                 assert layer_impl.supports_mtp_with_cp_non_trivial_interleave_size, (
-                    "MTP with cp_kv_cache_interleave_size > 1 is not "
-                    f"supported in {layer_impl.__class__.__name__}."
+                    "MTP with cp_kv_cache_interleave_size > 1 requires an "
+                    "attention backend that supports this configuration, but "
+                    f"{layer_impl.__class__.__name__} does not. "
+                    "Try setting VLLM_ATTENTION_BACKEND to a compatible backend."
                 )
             if dcp_size > 1:
                 assert layer_impl.need_to_return_lse_for_decode, (
-                    "DCP requires attention impls to return"
-                    " the softmax lse for decode, but the impl "
-                    f"{layer_impl.__class__.__name__} "
-                    "does not return the softmax lse for decode."
+                    "Decode Context Parallelism (DCP) requires an attention "
+                    "backend that returns softmax LSE for decode, but "
+                    f"{layer_impl.__class__.__name__} does not. "
+                    "Set VLLM_ATTENTION_BACKEND to a DCP-compatible backend "
+                    "such as FLASH_ATTN or FLASHINFER, and retry."
                 )
 
             if pcp_size > 1:
                 assert layer_impl.supports_pcp, (
-                    "PCP requires attention impls' support, "
-                    f"but the impl {layer_impl.__class__.__name__} "
-                    "does not support PCP."
+                    "Prefill Context Parallelism (PCP) requires an attention "
+                    "backend with PCP support, but "
+                    f"{layer_impl.__class__.__name__} does not. "
+                    "Set VLLM_ATTENTION_BACKEND to a PCP-compatible backend "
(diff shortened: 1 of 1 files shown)


vllm/v1/worker/cp_utils.py
```diff
@@ -21,22 +21,56 @@ def check_attention_cp_compatibility(vllm_config: VllmConfig) -> None:
...
+                and not supports_mtp
+            ):
+                raise RuntimeError(
+                    f"Multi-Token Prediction (MTP) with "
+                    f"cp_kv_cache_interleave_size > 1 is not supported by the "
+                    f"current attention backend "
+                    f"'{layer_impl.__class__.__name__}'.\n\n"
+                    f"To resolve this issue, try one of the following:\n"
+                    f"  1. Use a different attention backend by setting:\n"
+                    f"     export VLLM_ATTENTION_BACKEND=<backend>\n"
```
LucasWilkinson: please point them to `--attention-backend` instead; we are trying to deprecate `VLLM_ATTENTION_BACKEND`

vllm/v1/worker/cp_utils.py
```diff
@@ -21,22 +21,56 @@ def check_attention_cp_compatibility(vllm_config: VllmConfig) -> None:
...
-                    "does not support PCP."
+            if pcp_size > 1 and not layer_impl.supports_pcp:
+                raise RuntimeError(
+                    f"Prefill Context Parallel (PCP) requires an attention "
+                    f"backend that supports PCP. The current backend "
+                    f"'{layer_impl.__class__.__name__}' does not support this "
+                    f"feature.\n\n"
+                    f"To resolve this issue, try one of the following:\n"
+                    f"  1. Use a compatible attention backend by setting:\n"
+                    f"     export VLLM_ATTENTION_BACKEND=<backend>\n"
```
LucasWilkinson: ditto: https://github.com/vllm-project/vllm/pull/30936/changes#r2636536227

LucasWilkinson (changes requested):
Overall looks pretty good to me, one nit; thank you for the contributions!

yurekami:
## Alternative Simpler Approach

I've pushed a simpler alternative to my fork branch `feature/improve-dcp-error-messages` that keeps the existing assertion pattern while improving the error messages:

**Key differences from this PR:**
- Keeps `assert` statements (less invasive change)
- Shorter, more concise messages
- Still includes `VLLM_ATTENTION_BACKEND` suggestion with FLASH_ATTN/FLASHINFER examples

Example:
```
Decode Context Parallelism (DCP) requires an attention backend that returns softmax LSE for decode, but FooImpl does not. Set VLLM_ATTENTION_BACKEND to a DCP-compatible backend such as FLASH_ATTN or FLASHINFER, and retry.
```

If maintainers prefer the simpler approach, I can open a separate PR. Otherwise, happy to close my branch in favor of this more comprehensive solution.

PR #30946, opened 2025-12-18
[Core] Improve DP synchronization error messages

yurekami:
## Summary

Replace bare `assert` statements with informative `RuntimeError` exceptions in `dp_utils.py` to improve debugging experience for Data Parallel users.

**Changes:**
- Replace `assert num_tokens_padded >= num_tokens_unpadded` with descriptive error showing actual values
- Replace `assert should_attempt_dp_padding == should_dp_pad` with error explaining DP rank synchronization mismatch
- Replace `assert uniform_decode is not None` with error explaining microbatching parameter requirement

**Each error now includes:**
- Clear explanation of what went wrong
- Current values for debugging (where applicable)
- Link to GitHub issues for reporting internal errors

## Background

This follows the same pattern as PR #30936 which improved error messages in `cp_utils.py` for DCP/PCP/MTP. Users encountering DP synchronization issues will now get actionable error messages instead of bare `AssertionError`.

## Test Plan

- [x] Syntax verification passes (`python -m py_compile`)
- [ ] Existing DP tests pass (CI will verify)
- Error paths only modified - no behavioral changes to success paths

### DIFF
--- a/vllm/v1/worker/dp_utils.py
+++ b/vllm/v1/worker/dp_utils.py
@@ -126,14 +126,7 @@ def _synchronize_dp_ranks(
     ]
 
     """
-    if num_tokens_padded < num_tokens_unpadded:
-        raise RuntimeError(
-            f"Data Parallel synchronization error: padded token count "
-            f"({num_tokens_padded}) is less than unpadded token count "
-            f"({num_tokens_unpadded}). This indicates an internal error in "
-            f"token padding calculation. Please report this issue at "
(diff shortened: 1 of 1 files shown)


LucasWilkinson (changes requested):
I dont not find these very actionable or informative (just adds lines bloating the files) given it does not give the user any recommendations other then reporting the error, maybe we can just have global "Please report this issue at https://github.com/vllm-project/vllm/issues" for assertion errors given assertions imply something went wrong internally

PR #30948, opened 2025-12-18
fix: Suppress torch.frombuffer UserWarning for non-writable buffers

yurekami:
## Summary
Fixes #26781 - Suppress UserWarning from `torch.frombuffer()` on non-writable buffers

## Problem
`torch.frombuffer()` emits a `UserWarning` when given a non-writable buffer (e.g., from `pickle.dumps()` or shared memory):
```
UserWarning: The given buffer is not writable, and PyTorch does not support non-writable tensors.
This means you can write to the underlying (supposedly non-writable) buffer using the tensor...
```

This warning clutters logs but is harmless in our use cases since we either clone the tensor immediately, don't mutate it, or know the buffer will be properly handled.

## Solution
Add `frombuffer_with_writable_warning_suppressed()` helper function to `vllm/utils/torch_utils.py` and use it at affected call sites.

This approach was approved by @zou3519 in the issue discussion.

## Changes
- `vllm/utils/torch_utils.py`: Add helper function that wraps `torch.frombuffer()` with warning suppression
- `vllm/distributed/parallel_state.py`: Use helper for pickle serialization (line 653)
- `vllm/distributed/device_communicators/shm_broadcast.py`: Use helper for shared memory initialization (line 203)

## Testing
- The warning suppression is targeted specifically to the "buffer is not writable" message
- No functional changes to tensor creation behavior
- Existing tests should pass unchanged

### DIFF
--- a/vllm/utils/torch_utils.py
+++ b/vllm/utils/torch_utils.py
@@ -4,6 +4,7 @@ import contextlib
 import importlib.metadata
 import os
 import threading
+import warnings
 from collections.abc import Callable, Collection
 from functools import lru_cache
 from typing import TYPE_CHECKING, Any, TypeVar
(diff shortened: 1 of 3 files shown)


LucasWilkinson (approved):
LGTM; please fix the pre-commit

vllm/utils/torch_utils.py
```diff
@@ -721,3 +722,24 @@ def direct_register_custom_op(
     my_lib.impl(op_name, op_func, dispatch_key=dispatch_key)
     if fake_impl is not None:
         my_lib._register_fake(op_name, fake_impl)
+
+
+def frombuffer_with_writable_warning_suppressed(
```
zou3519, line 727: is it possible to add a test somewhere? PyTorch could change the error message and then we wouldn't catch it. The test can just  call this function and then assert that no warnings were raised.

PR #30738, opened 2025-12-16
[Metrics] Model FLOPs Utilization estimation

SungMinCho:
Signed-off-by: SungMinCho <tjdals4565@gmail.com>

## TL;DR

This PR implements optional "MFU stats logging", which appends "per-GPU flops/bandwidth" information to the existing periodic logs, reporting the average compute/memory performance of GPUs in each Engine for the duration of that log interval. These stats are calculated with minimal overhead by feeding in SchedulerOutput to the analytic config-based perf calculator at every iteration.

## How to use

Set `--enable-mfu-metrics` when launching the vLLM server to enable MFU logging.

`VLLM_DEBUG_MFU_METRICS=1` with `VLLM_LOGGING_LEVEL=DEBUG` is for debugging purposes for developers working on the metrics calculations.

## Purpose

To track the hardware performance over the course of vLLM serving.

## Examples

(for each example, please find the MFU stats at the end of each logged line)
(all examples were gathered on B200 GPUs)

### GPT-OSS 120B TP=8 BatchSize=256 Input=6K Output=3K NumBatch=1

**with `VLLM_MFU_LOGGING_LEVEL=1`**
```
INFO 12-15 17:38:18 [loggers.py:259] Engine 000: Avg prompt throughput: 3599.9 tokens/s, Avg generation throughput: 199.6 tokens/s, Running: 6 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.1%, Prefix cache hit rate: 0.0%, MFU: 6.6 TF/s/GPU 105.0 GB/s/GPU
INFO 12-15 17:38:28 [loggers.py:259] Engine 000: Avg prompt throughput: 150003.2 tokens/s, Avg generation throughput: 938.8 tokens/s, Running: 256 reqs, Waiting: 0 reqs, GPU KV cache usage: 2.4%, Prefix cache hit rate: 0.0%, MFU: 259.9 TF/s/GPU 491.7 GB/s/GPU
INFO 12-15 17:38:38 [loggers.py:259] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 18609.5 tokens/s, Running: 256 reqs, Waiting: 0 reqs, GPU KV cache usage: 2.7%, Prefix cache hit rate: 0.0%, MFU: 32.6 TF/s/GPU 1138.1 GB/s/GPU
[...]

PR #30950, opened 2025-12-18
[Metrics] Add Prometheus counters for Model FLOPs Utilization (MFU)

markmc:
See #30738 - this is a follow-on to export these metrics via Prometheus in addition to the console logging

The metrics are only calculated and available with `--enable-mfu-metrics`

### DIFF
--- a/docs/mkdocs/hooks/generate_metrics.py
+++ b/docs/mkdocs/hooks/generate_metrics.py
@@ -22,6 +22,7 @@ METRIC_SOURCE_FILES = [
         "path": "vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py",
         "output": "nixl_connector.inc.md",
     },
+    {"path": "vllm/v1/metrics/perf.py", "output": "perf.md"},
 ]
 
 
--- a/docs/mkdocs/hooks/generate_metrics.py
+++ b/docs/mkdocs/hooks/generate_metrics.py
@@ -22,7 +22,7 @@ METRIC_SOURCE_FILES = [
         "path": "vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py",
         "output": "nixl_connector.inc.md",
     },
-    {"path": "vllm/v1/metrics/perf.py", "output": "perf.md"},
+    {"path": "vllm/v1/metrics/perf.py", "output": "perf.inc.md"},
 ]
 
 
--- a/docs/usage/metrics.md
+++ b/docs/usage/metrics.md
@@ -49,7 +49,7 @@ The following metrics are exposed:
 
 These metrics are available via `--enable-mfu-metrics`:
 
---8<-- "docs/generated/metrics/perf.md"
+--8<-- "docs/generated/metrics/perf.inc.md"
 
 ## Deprecation Policy
 
--- a/docs/usage/metrics.md
+++ b/docs/usage/metrics.md
@@ -45,6 +45,12 @@ The following metrics are exposed:
 
 --8<-- "docs/generated/metrics/nixl_connector.inc.md"
 
+## Model Flops Utilization (MFU) Performance Metrics
+
+These metrics are available via `--enable-mfu-metrics`:
+
+--8<-- "docs/generated/metrics/perf.md"
+
 ## Deprecation Policy
 
 Note: when metrics are deprecated in version `X.Y`, they are hidden in version `X.Y+1`
--- a/vllm/v1/metrics/loggers.py
+++ b/vllm/v1/metrics/loggers.py
@@ -19,7 +19,7 @@ from vllm.distributed.kv_transfer.kv_connector.v1.metrics import (
 from vllm.logger import init_logger
 from vllm.plugins import STAT_LOGGER_PLUGINS_GROUP, load_plugins_by_group
 from vllm.v1.engine import FinishReason
-from vllm.v1.metrics.perf import PerfMetricsLogging
+from vllm.v1.metrics.perf import PerfMetricsLogging, PerfMetricsProm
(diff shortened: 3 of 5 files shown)


vllm/v1/metrics/perf.py
```diff
@@ -1225,6 +1226,87 @@ def log(self, log_fn=logger.info, log_prefix: str = "") -> None:
...
+
+
+def make_per_engine(
+    counter: prometheus_client.Counter, per_engine_labelvalues: dict[int, list[str]]
+):
+    """Create a counter for each label value."""
+    return {
+        idx: counter.labels(*labelvalues)
+        for idx, labelvalues in per_engine_labelvalues.items()
+    }
```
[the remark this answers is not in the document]
markmc: It's a pretty trivial copy, this PR just follows example of other PRs, and I like that the logic isn't abstracted away - it's important to think about which labels are relevant when adding new metrics

hmellor:
It looks like all the tests passed but it's not been reviewed

markmc:
> It looks like all the tests passed but it's not been reviewed

Yes, this is ready for review/merge

> I don't see anything that adds `--enable-mfu-metrics` though

This was added by #30738 which just added console logging of these metrics. This PR follows-on to add Prometheus support

vllm/v1/metrics/loggers.py
```diff
@@ -390,6 +390,7 @@ class PrometheusStatLogger(AggregateStatLoggerBase):
     _histogram_cls = Histogram
     _spec_decoding_cls = SpecDecodingProm
     _kv_connector_cls = KVConnectorPrometheus
+    _perf_metrics_cls = PerfMetricsProm
```
hickeyma, line 395: Small nit. It would be better for consistency if the suffix of the class is `Prom` or `Prometheus` but not both. It seems to be `Prom` but line above is using `Prometheus`  for `KVConnectorPrometheus`? Personally, I'd prefer `Prometheus` for clarity when reading the code.

vllm/v1/metrics/perf.py
```diff
@@ -1225,6 +1226,87 @@ def log(self, log_fn=logger.info, log_prefix: str = "") -> None:
...
+
+def make_per_engine(
+    counter: prometheus_client.Counter, per_engine_labelvalues: dict[int, list[object]]
+):
+    """Create a counter for each label value."""
+    return {
+        idx: counter.labels(*labelvalues)
+        for idx, labelvalues in per_engine_labelvalues.items()
+    }
+
```
hickeyma, line 1316: This looks like its a duplicate of `make_per_engine()` in `vllm/v1/spec_decode/metrics.py`. If so, do you mind making it a utility function? Also, suggestions on naming to improve clarity as follows:

- `make_per_engine` --> `create_counter_per_engine`
- `per_engine_labelvalues` --> `per_engine_label_values`

hickeyma (changes requested):
Thanks @markmc, it looks good bar a few small nits. Should unit test be added for `PerfMetricsProm` class?

vllm/v1/metrics/loggers.py
markmc: `KVConnectorPrometheus` is really the odd one out. There's over 50 places where `Prom` is used. I'd consider a follow-on PR to fix the inconsistency, but I consider `PerfMetricsProm` consistent with the current codebase

vllm/v1/metrics/perf.py
markmc: Yes, there's 3 other identically named helper functions. I did consider consolidating them in this PR, but I have mixed feelings - it's important to understand the metric creation logic with labels, etc., and I'd almost prefer to inline this code than have a far-away helper

I'd reconsider in a follow-up, but again ... a consolidation of the other helpers with this one would be beyond the scope of this PR

hickeyma: ok, thats fair enough

hickeyma (approved):
LGTM, thanks @markmc

vllm/v1/metrics/loggers.py
hickeyma: @markmc FYI, pushed #33898 to handle inconsistency.

vllm/v1/metrics/perf.py
hickeyma: @markmc FYI, pushed #33898 to handle helper function.

zhuohan123 (approved):

docs/mkdocs/hooks/generate_metrics.py
```diff
@@ -22,6 +22,7 @@
         "path": "vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py",
         "output": "nixl_connector.inc.md",
     },
+    {"path": "vllm/v1/metrics/perf.py", "output": "perf.md"},
```
hmellor:
```suggestion
    {"path": "vllm/v1/metrics/perf.py", "output": "perf.inc.md"},
```

docs/usage/metrics.md
```diff
@@ -45,6 +45,12 @@ The following metrics are exposed:
 
 --8<-- "docs/generated/metrics/nixl_connector.inc.md"
 
+## Model Flops Utilization (MFU) Performance Metrics
+
+These metrics are available via `--enable-mfu-metrics`:
+
+--8<-- "docs/generated/metrics/perf.md"
```
hmellor:
```suggestion
--8<-- "docs/generated/metrics/perf.inc.md"
```

hmellor (commented):
Otherwise the docs build will complain that `perf.md` can't be navigated to

PR #30647, opened 2025-12-14
[Perf] Eliminate padding and slicing op for GPT-OSS with Flashinfer MXFP4 MXFP8 MoE

elvischenv:
## Purpose

- Depends on Flashinfer update #30993
- Eliminated padding op before the MoE: by setting the alignment in flashinfer mxfp8 quant, the output quantized tensor will be padded.
- Eliminated slicing op after the MoE: by passing the output tensor with unpadded hidden size to MoE kernel, this depends on a Flashinfer PR:
  - https://github.com/flashinfer-ai/flashinfer/pull/2217
  - This will also resolve the previous AR+Norm fusion broken by slice op issue: #28841
- ~~Cleaned up the padding logic: for mxfp4 quant, the padded hidden size is calculated in `create_weights()`, the `maybe_roundup_hidden_size()` in `vllm/model_executor/layers/fused_moe/layer.py` seems like a dup.~~

### Accuracy

PR:
```
[{'eval_name': 'gpqa', 'model_name': 'gpt-oss-120b-high_temp1.0_20251213_233136', 'metric': 0.7803030303030303}]
[{'eval_name': 'aime25', 'model_name': 'gpt-oss-120b-high_temp1.0_20251213_234320', 'metric': 0.8875}]
```

main:
```
[{'eval_name': 'gpqa', 'model_name': 'gpt-oss-120b-high_temp1.0_20251214_002509', 'metric': 0.7891414141414141}]
[{'eval_name': 'aime25', 'model_name': 'gpt-oss-120b-high_temp1.0_20251214_001505', 'metric': 0.8875}]
```

### Kernel

PR:
```
void cublasLt::splitKreduce_kernel                          2.400 μs
void tensorrt_llm::kernels::quantize_with_block_size        2.944 μs
void moe::dev::routing::routingRenormalize                  5.216 μs
bmm_MxE4m3_MxE2m1MxE4m3_Fp32_t128x16x256u2                 13.216 μs
bmm_Bfloat16_MxE2m1MxE4m3_Fp32_t128x16x256u2                9.504 μs
void moe::dev::finalize::finalizeKernel                     3.168 μs
void flashinfer::trtllm_allreduce_fusion                    7.744 μs (ar+norm)
nvjet_tst_32x64_64x16_4x1_v_bz_splitK_TNN                   4.448 μs
```

main:
```
void cublasLt::splitKreduce_kernel                          2.048 μs
triton_poi_fused_constant_pad_nd_moe_forward_0              1.407 μs (pad)
[...]

PR #30976, opened 2025-12-18
Use aiter triton fused_add_rmsnorm_pad for gpt-oss

Rohan138:
## Purpose

Adds fused padding op before router GEMM on ROCm, eliminating this unfused pad after the GEMM before the fused_moe: https://github.com/ROCm/vllm/blob/main/vllm/model_executor/layers/fused_moe/layer.py#1603

Before:
image
After:
image

~Follow-up/alternate possibility is to replace this with a single `F.pad` before the router, then add a fusion pass to fuse AITER CK rmsnorm and pad to PassManager similar to https://github.com/vllm-project/vllm/pull/25693.~ Done

See also #30357 (gpt-oss quark w4a8 enablement) and #30647 (eliminate padding op on NV w4a8 gpt-oss)

## Test Result

---
<details>

</details>


vllm/model_executor/models/gpt_oss.py
```diff
@@ -233,7 +244,18 @@ def forward(
...
+                self.post_attention_layernorm.variance_epsilon,
+                residual,
+                x_pad_to_multiple=256,
+            )
+        else:
+            hidden_states, residual = self.post_attention_layernorm(
+                hidden_states, residual
+            )
         output = self.mlp(hidden_states)
         return output, residual
```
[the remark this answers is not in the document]
Rohan138: fused_add_rmsnorm_pad only pads the output tensor, not the residual. See https://github.com/ROCm/aiter/blob/main/aiter/ops/triton/fused_add_rmsnorm_pad.py and https://github.com/ROCm/aiter/blob/main/aiter/ops/triton/_triton_kernels/fused_add_rmsnorm_pad.py

vllm/model_executor/models/gpt_oss.py
```diff
@@ -49,6 +50,9 @@
     maybe_prefix,
 )
 
+if current_platform.is_rocm() and rocm_aiter_ops.is_enabled():
```
gshtras: is_rocm is already a part of rocm_aiter_ops.is_enabled through @if_aiter_supported

vllm/model_executor/models/gpt_oss.py
```diff
@@ -233,7 +237,18 @@ def forward(
         hidden_states = self.attn(hidden_states, positions)
 
         # Fully Connected
-        hidden_states, residual = self.post_attention_layernorm(hidden_states, residual)
+        if current_platform.is_rocm() and rocm_aiter_ops.is_enabled():
```
gshtras: Same as above. Also it's better to cache the aiter check in a class var

ProExpertProg (changes requested):
Let's do this via compile pass instead of platform-specific model definition changes

vllm/compilation/pass_manager.py
```diff
@@ -18,8 +18,9 @@
 
 if rocm_aiter_ops.is_enabled():
     from vllm.compilation.rocm_aiter_fusion import (
-        RocmAiterRMSNormFusionPass,
+        RocmAiterRMSNormQuantFusionPass,
```
Rohan138, line 21: Renamed to `RocmAiterRMSNormQuantFusionPass` for clarity

vllm/compilation/rocm_aiter_fusion.py
```diff
@@ -399,3 +399,72 @@ def uuid(self) -> str:
...
+
+class AddAiterRMSNormPadPattern:
+    """
+    This pattern replaces aiter_rmsnorm_with_add & a pad op with a aiter_triton_add_rmsnorm_pad op.
+    """
+
+    AITER_TRITON_ADD_RMSNORM_PAD_OP = (
+        rocm_aiter_ops.get_aiter_triton_add_rmsnorm_pad_op()
+    )
+    PAD_SIZE = 192
```
Rohan138: This is currently hardcoded for the gpt-oss pad i.e. pad(2880, 256) = 2880 + 192 = 3072.

I can generalize this in a follow up, possibly in #32307 which will switch MI300 padding from 256 (back) to 128.

vllm/envs.py
```diff
@@ -969,6 +970,11 @@ def _get_or_set_default() -> str:
     "VLLM_ROCM_USE_AITER_RMSNORM": lambda: (
         os.getenv("VLLM_ROCM_USE_AITER_RMSNORM", "True").lower() in ("true", "1")
     ),
+    # Whether to use fused add+rmsnorm+pad kernel for gpt-oss
+    "VLLM_ROCM_USE_AITER_TRITON_FUSED_ADD_RMSNORM_PAD": lambda: (
```
Rohan138: Currently using the env var for testing, but I can drop it and keep it as default/stick it inside a pass_config pass instead.

vllm/compilation/pass_manager.py
```diff
@@ -123,13 +124,16 @@ def configure(self, config: VllmConfig) -> None:
...
                     self.passes += [
-                        RocmAiterRMSNormFusionPass(config),
+                        RocmAiterRMSNormQuantFusionPass(config),
                     ]
             if self.pass_config.fuse_act_quant:
                 self.passes += [ActivationQuantFusionPass(config)]
                 if rocm_aiter_ops.is_enabled():
                     self.passes += [RocmAiterSiluMulFp8GroupQuantFusionPass(config)]
 
+            if rocm_aiter_ops.is_enabled():
```
ProExpertProg: Let's add a flag here (can be enabled by default in optimization level 1+)

vllm/compilation/rocm_aiter_fusion.py
ProExpertProg: Can you make a pad size param like epsilon is?

vllm/model_executor/models/gpt_oss.py
```diff
@@ -182,12 +182,21 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
             x = sequence_parallel_chunk(x)
 
         if current_platform.is_rocm():
+            # Pad before router GEMM to fuse with preceding aiter_rmsnorm_with_add
```
ProExpertProg: Instead of changing the model definition, the transformation should just transform `rms -> gemm -> pad` into `rms+pad -> gemm`.

Also cc @tjtanaa could we unify this path in general so that we can call `self.router` on ROCm as well?

vllm/compilation/rocm_aiter_fusion.py
Rohan138: I made both `hidden_size` and `x_pad_to_multiple` params, then specified them in the [pass](https://github.com/ROCm/vllm/blob/6adba53bd0703ee2ce5333c3d99a0c9c4d9808cb/vllm/compilation/rocm_aiter_fusion.py#L463).

Alternatively, I could hardcode `hidden_size` in a get_inputs function instead:
```
    def get_inputs(self) -> list[torch.Tensor]:
        M, N = 5, 2880
        input = self.rmsnorm_matcher.empty(M, N) if self.rmsnorm_matcher.enabled else self.rmsnorm_matcher.empty_f32(M, N)
        weight = self.rmsnorm_matcher.empty(N)
        residual = self.rmsnorm_matcher.empty(M, N)
        return [input, weight, residual]

    def register(self, pm_pass: PatternMatcherPass) -> None:
        def pattern(
            input: torch.Tensor, weight: torch.Tensor, residual: torch.Tensor
            _, N = input.shape
            pad_size = self.x_pad_to_multiple - N % self.x_pad_to_multiple
            result_rms, residual_out = self.rmsnorm_matcher(input, weight, residual)
            result = torch.nn.functional.pad(
                result_rms, (0, pad_size), mode="constant", value=0.0
            )
            return result, residual_out
```

vllm/compilation/pass_manager.py
Rohan138: Added `fuse_act_padding`, please check if the conditions in `vllm/config/vllm.py` look good to you

vllm/model_executor/models/gpt_oss.py
Rohan138: @ProExpertProg can I put both of those changes in a follow-up PR?
- Unify `rocm_unquantized_gemm` with `self.router` and remove the `F.pad` introduced here
- Benchmark `wvSplitKrc` vs `gemm_a16w16`
- switch the fusion to an `rmsnorm -> router_gemm -> pad` fusion pass

This PR is intended to be used with the AITER Triton GEMM (`VLLM_ROCM_USE_AITER_TRITON_GEMM=1`), we need some internal discussion+benchmarking to extend it to the splitK and splitKrc (#29843) GEMM paths in [rocm_unquantized_gemm_impl](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/utils.py#L129)

Currently those calls reshape the input `(N,K) -> `, which results in this extra copy+fill between the rmsnorm and the GEMM:
image

For comparison, this is current main (AITER rmsnorm + fill + GEMM):
image

cc @dllehr-amd

ProExpertProg: Yeah I think the router unification should definitely be a follow-up. But let's fix the pattern now so we don't have to modify the model definition at all

Rohan138: @ProExpertProg put `rocm_unquantized_gemm` into the pattern`, PTAL

vllm/compilation/pass_manager.py
Rohan138: cc @gshtras for viz on the new flag+conditions for it in `vllm/config/vllm.py`

tests/compile/test_fuse_act_padding.py
```diff
@@ -0,0 +1,112 @@
...
+    ):
+        super().__init__()
+        self.norm = RMSNorm(hidden_size, eps=1e-5)
+        self.router = torch.nn.Linear(hidden_size, num_local_experts)
+        self.x_pad_to_multiple = x_pad_to_multiple
+        self.pad_dim = x_pad_to_multiple - (hidden_size % x_pad_to_multiple)
+
+    def forward(self, x, residual):
+        x, residual = self.norm(x, residual)
+        router_logits = rocm_unquantized_gemm(
```
ProExpertProg: Let's replicate a few layers here to simulate residual reuse, as that can sometimes cause matching bugs.

ProExpertProg (approved):
LGTM, nice work! Just one comment about improving the test

tests/compile/test_fuse_act_padding.py
Rohan138: Done, checking CI failures

PR #30983, opened 2025-12-18
Check for truthy `rope_parameters` not the existence of it

hmellor:
Fixes the following situation:

- Model explicitly sets `rope_parameters=None` (new name, Transfomers v5) or `rope_scaling=None` (old name, Transformers v4)
- User is using Transformers v4 with vLLM


mgoin (approved):
This fixes `vllm serve google/functiongemma-270m-it`, thanks

PR #30991, opened 2025-12-18
[ROCm][CI/Build] Update ROCm dockerfiles

gshtras:
Dockerfile.rocm:
Override torch profiler setting to allow full traces
Write extra info to versions.txt

Dockerfile.rocm_base:
Revert to ROCm7.0 due to a performance regression in 7.1 that is investigated internally
Update torch commit to ROCm 2.9.0 release
Update triton commit to match the new torch commit
Update AITER to a more recent commit

### DIFF
--- a/docker/Dockerfile.rocm
+++ b/docker/Dockerfile.rocm
@@ -130,6 +130,7 @@ RUN --mount=type=bind,from=export_vllm,src=/,target=/install \
     && uv pip install --system *.whl
 
 ARG COMMON_WORKDIR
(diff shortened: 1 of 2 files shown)


tjtanaa (approved):
LGTM, if this has been validated internally by AMD on AMD CI unit tests as well. The AMD CI runs is not using this environment (the `rocm/vllm-dev:base` is still using rocm7.1), so we are not able to use that as reference.

tjtanaa:
@gshtras I will leave the merge to you.

PR #30739, opened 2025-12-16
[BugFix] Support online dense model DP without overhead

njhill:
Currently, there's unnecessary overhead when running non-MoE models in a data parallel configuration because the steps across the ranks are synchronized with redundant all-reduce ops and coordination is done to ensure "idle" ranks perform dummy forward passes.

This PR changes the parallel config at the worker level to be equivalent to DP=1 for non-MoE models, so each rank operates independently. When internal load-balancing is used, the DP coordinator still runs to propagate stats back from the engines for load balancing purposes, but the step/wave synchronization logic is disabled.

Fixes https://github.com/vllm-project/vllm/issues/24461.
Fixes https://github.com/vllm-project/vllm/issues/30655.

This is supported in the online / AsyncLLM case only.

The [offline DP](https://github.com/vllm-project/vllm/blob/main/examples/offline_inference/data_parallel.py) will now fail during startup for non-MoE models (it really makes no sense to use it in that configuration).

Benchmark on 4xH100:
```
vllm serve Qwen/Qwen3-8B --data-parallel-size 4 --uvicorn-log-level=error
```
```
vllm bench serve \
    --backend vllm \
    --model Qwen/Qwen3-8B \
    --dataset-name random \
    --random-input-len 128 \
    --random-output-len 512 \
    --ignore-eos \
    --port 8033 \
    --num-prompts 4000 \
    --max-concurrency 200 \
    --seed 42
```

#### Before
```
============ Serving Benchmark Result ============
Successful requests:                     4000      
Failed requests:                         0         
Maximum request concurrency:             200       
Benchmark duration (s):                  104.41    
Total input tokens:                      512000    
Total generated tokens:                  2048000   
Request throughput (req/s):              38.31     
Output token throughput (tok/s):         19615.26  
Peak output token throughput (tok/s):    21597.00  
Peak concurrent requests:                400.00    
Total token throughput (tok/s):          24519.08  
---------------Time to First Token----------------
Mean TTFT (ms):                          131.78    
Median TTFT (ms):                        124.31    
P99 TTFT (ms):                           404.36    
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          9.93      
Median TPOT (ms):                        9.95      
P99 TPOT (ms):                           10.08     
---------------Inter-token Latency----------------
Mean ITL (ms):                           9.93      
Median ITL (ms):                         9.79      
P99 ITL (ms):                            15.93     
==================================================
```

#### After
```
============ Serving Benchmark Result ============
[...]

PR #26012, opened 2025-10-01
[Bugfix] Fix `_reqs_to_process` leak on abort

NickLucche:
Fix a small leak that can happen whenever a request is aborted while prefilling on P, causing its id to not be removed from the set.
I've added a unit test to clarify the scenario.
The test should fail on main with

```
>       assert req.request_id not in connector.connector_worker._reqs_to_process
E       AssertionError: assert 'id-1' not in {'id-1'}
E        +  where 'id-1' = <vllm.v1.request.Request object at 0x7f58fc128a10>.request_id
E        +  and   {'id-1'} = <tests.v1.kv_connector.unit.test_nixl_connector.FakeNixlConnectorWorker object at 0x7f58fc1289b0>._reqs_to_process
E        +    where <tests.v1.kv_connector.unit.test_nixl_connector.FakeNixlConnectorWorker object at 0x7f58fc1289b0> = <vllm.distributed.kv_transfer.kv_connector.v1.nixl_connector.NixlConnector object at 0x7f5901481760>.connector_worker
```

PR #31034, opened 2025-12-19
[P/D] rework mooncake connector and introduce its bootstrap server

dtcccc:
## Purpose

Rework mooncake connector to achieve better performance and prepare for more features in future.
Introduce a central bootstrap server on P.

init phase:
All P workers register their info (dp/tp/pp rank, zmq worker addr) with the bootstrap server.
After all P workers finished registering, proxy and D workers can query when they meet a new engine_id.

Note:
(deprecated)
> Since #30739 `data_parallel_size` and `data_parallel_size_local` are stick to 1 for non-Moe models. So `origin_data_parallel_size` and `origin_data_parallel_size_local` are introduced to get the raw value.
> For non-Moe model with dp_size > 1, all its engines by dp have the same engine_id and unable to distinguish dp_ranks and workers. So `[engine_id][dp_rank]` is used. See comments in `MooncakeBootstrapServer` for detail.
> 
> After startup, the proxy will send reqs to both P and D concurrently. Due to #27987 P and D cannot know the exact request_id of each other. #32630 seems to be a solution, and I just try to drop the last 9 chars as workaround here. See comments in `TruncatingDict`.

Thanks to #33037 and #32937 we can drop all workarounds now.

This design is partially inspired by sglang, aiming to improve Time To First Token (TTFT) performance and to lay the groundwork for a future layerwise transfer feature.

With random dataset and max-concurrency = 1, TTFT on two A10 machines running Qwen2.5-7B-Instruct is improved:
random-input-len 128: 83.41ms -> 77.53ms
random-input-len 1024: 252.35ms ->246.25ms
This result shows TTFT win about 6ms from running in P and D simultaneously.

Other highlights of this PR:

- Introduced the proxy and example scripts for the new bootstrap server architecture, along with updated documentation to facilitate learning and testing.
- Prepare the future introduction of heterogeneous TP. For now, heterogeneous TP and PP is explicitly rejected.
- Pick a bugfix from https://github.com/vllm-project/vllm/pull/26012

## Test Result

---
<details>

</details>


examples/online_serving/disaggregated_serving/mooncake_connector/mooncake_connector_proxy.py
```diff
@@ -0,0 +1,344 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
```
robertgshaw2-redhat, line 3: why do we need a separate proxy for mooncake?

dtcccc: From what I've seen so far, the proxies used by vLLM's various connectors aren't unified. For example, disagg_proxy_demo.py doesn't support the nixl-specific "do_remote_decode" and "do_remote_prefill" parameters, which is why nixl has its own toy_proxy_server.py. So, in my view, it's common practice for each connector to have its own dedicated proxy.

To get back to the main point, we are using this specific proxy because it aligns with the central bootstrap server architecture we want to introduce. This is a prerequisite for supporting layerwise transfer in the future. (The other existing proxies only send data to D after receiving the full result from P, which is too late for our use case.)

vllm/distributed/kv_transfer/kv_connector/v1/mooncake_connector.py
```diff
@@ -59,56 +63,100 @@
 
 logger = init_logger(__name__)
 
+http_log_level = logger.getEffectiveLevel()
+# INFO logs of http are too noisy. Silence them.
+# Setting vllm log level to DEBUG if we really want to see.
+if http_log_level == logging.INFO:
+    http_log_level = logging.WARNING
+logging.getLogger("httpx").setLevel(http_log_level)
```
NickLucche: this is a bit arbitrary, we should probably either do it in the mooncakeconnector init or push a separate global change cc @markmc 

Either ways, we should log that we're "silencing" logs.

vllm/distributed/kv_transfer/kv_connector/v1/mooncake_connector.py
```diff
@@ -59,56 +63,100 @@
...
+
+
+class QueryRequestsResponse(BaseModel):
+    results: dict[ReqId, tuple[str, list[str]]]
 
 
 class MooncakeConnectorMetadata(KVConnectorMetadata):
     def __init__(self):
-        self.reqs_to_recv: dict[ReqId, RecvReqMeta] = {}
+        self.reqs_to_recv: list[PullReqMeta] = []
```
NickLucche: why did we switch to a list here? 
If we want to group the reqs by remote boostrap server, can't we just use a dict[server_addr, list[meta]] ?

vllm/distributed/kv_transfer/kv_connector/v1/mooncake_connector.py
```diff
@@ -311,8 +394,36 @@ def update_state_after_alloc(
...
+            logger.debug("Registered request %s with bootstrap server", req_id)
+        except Exception as e:
+            err_msg = (
+                e.response.text if isinstance(e, httpx.HTTPStatusError) else str(e)
+            )
+            logger.error(
+                "Failed to register request %s with bootstrap server: %s",
+                req_id,
+                err_msg,
+            )
```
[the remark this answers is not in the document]
NickLucche: I think this is a valid concern, we should at least add a TODO.
Re: disconnects I am assuming the proxy forwards to both instances

vllm/distributed/kv_transfer/kv_connector/v1/mooncake_connector.py
```diff
@@ -905,10 +1283,223 @@ def group_concurrent_contiguous(
...
+
+
+# ####################################################################
+# ## Mooncake Bootstrap Server
+# ####################################################################
+
+
+class MooncakeBootstrapServer:
+    """
+    A centralized server running on the global rank 0 prefiller worker.
```
NickLucche: nit: could probably live in a separate mooncake_utils file

vllm/distributed/kv_transfer/kv_connector/v1/mooncake_connector.py
```diff
@@ -905,10 +1283,223 @@ def group_concurrent_contiguous(
...
+        # We only support homogeneous TP now.
+        if self.tp_size != payload.tp_size:
+            raise HTTPException(
+                status_code=400,
+                detail=(
+                    "heterogeneous TP is not supported yet. TP size mismatch: "
+                    f"expected {self.tp_size}, got {payload.tp_size}"
+                ),
+            )
+
```
NickLucche: logger debug with source payload info could help here

NickLucche (commented):
Thanks for the work once again @dtcccc !

After some thinking, I believe all in all these changes are already fine.

My main concern is the added complexity that this design brings in: we have a longer "failure-chain" to propagate through and effectively an extra point of failure in the side fastapi server which needs handling of failures for the registering/querying endpoints .
In particular, I feel like the extra "registering" call that each P worker now has to perform in its flow is the weakest link, so we really have to be sure the benefits from having D be aware of the req at step 0 outweigh this overhead.
To that extent it would be nice to have a broader benchmark sweep to confirm TTFT gains, possibly comparing with a prev version + "Refactored the sender thread using async coroutines" (in hindsight, this should've been a separate PR to help the review process here).

But I also understand that the alternative to the point above would be to "push" the dp_rank request-selection at the proxy level (or some Coordinator in front of the DP instances @robertgshaw2-redhat ) which would take away control from the connector and/or require more invasive changes. Therefore I am overall ok with this, but just wanted to bring up a few points for discussion. 


One qq, what is currently stopping D from running `query_requests` before P registers the request (3/4)?


cc @wseaton for changes that are very similar to a past work of yours (+opinion on future failure handling?)

vllm/distributed/kv_transfer/kv_connector/v1/mooncake/mooncake_connector.py
```diff
@@ -52,46 +60,75 @@
...
 
 @dataclass
 class SendBlockMeta:
+    p_req_id: ReqId
+    transfer_id: TransferId
     local_block_ids: list[int]
     ready: asyncio.Event
     expire_time: float = float("inf")
+    need_send: int = 0
+    sended: int = 0
```
NickLucche: nit: 
```suggestion
    sent: int = 0
```

vllm/distributed/kv_transfer/kv_connector/v1/mooncake/mooncake_connector.py
```diff
@@ -294,17 +348,25 @@ def update_state_after_alloc(
...
         elif params.get("do_remote_decode"):
-            # Add an empty list to worker to create event.
-            self._reqs_need_send[request.request_id] = []
+            assert not self.is_kv_consumer
+            if not params.get("transfer_id"):
+                logger.warning(
+                    "Got invalid KVTransferParams: %s. This "
+                    "request will not utilize KVTransfer",
+                    params,
+                )
```
NickLucche: nit: we can probably indicate transfer_Id is missing from router/lb

vllm/distributed/kv_transfer/kv_connector/v1/mooncake/mooncake_connector.py
```diff
@@ -409,33 +484,35 @@ def __init__(self, vllm_config: VllmConfig, engine_id: str):
...
+        self.pp_rank = get_pp_group().rank_in_group
 
         self.kv_caches_base_addr: list[int] = []
         self.device_kv_caches: dict[str, torch.Tensor] = {}
-        self.reqs_need_send: dict[ReqId, SendBlockMeta] = {}
+        self.reqs_need_send: dict[TransferId, SendBlockMeta] = {}
+
+        # Only used by prefillers.
+        host, port = get_mooncake_bootstrap_addr(vllm_config)
+        self.bootstrap_addr = make_zmq_path("http", host, port)
```
NickLucche: nit: probably this doesnt need to be an instance attribute since its only used once?

vllm/distributed/kv_transfer/kv_connector/v1/mooncake/mooncake_connector.py
```diff
@@ -549,76 +669,192 @@ async def _sender_worker(self, sock: zmq.asyncio.Socket):
...
+                status=response_status,
+                ok_reqs=[d_req_id for d_req_id, _ in ready_reqs],
+            )
+            await sock.send_multipart((identity, self._encoder.encode(response)))
+
+    def resolve_need_send(self, send_meta: SendBlockMeta, remote_tp_ranks: list[int]):
+        # Prepare for heterogeneous TP (one P pairs to multiple D)
+        send_meta.need_send = len(remote_tp_ranks)
+        if send_meta.need_send != 1:
+            logger.error("Mooncake: Heterogeneous TP is not supported yet.")
```
NickLucche, line 825: I think we should likely crash here, something's very wrong

vllm/distributed/kv_transfer/kv_connector/v1/mooncake/mooncake_connector.py
```diff
@@ -549,76 +669,192 @@ async def _sender_worker(self, sock: zmq.asyncio.Socket):
...
-            lengths,
-        )
+    async def send_kv_to_decode(
+        self, identity: bytes, sock: zmq.asyncio.Socket, meta: MooncakeXferMetadata
+    ):
+        pending_reqs: dict[ReqId, SendBlockMeta] = {}
+        remote_tp_ranks = self.kv_topo.get_target_remote_ranks(meta.remote_tp_size)
+        if self.tp_rank not in remote_tp_ranks:
+            # This D worker does not pair with the P worker.
+            msg = f"This P tp_rank {self.tp_rank} not in remote D target ranks {remote_tp_ranks}"  # noqa: E501
```
NickLucche, line 690: you could probably bundle this check with the heteogenous TP one

vllm/distributed/kv_transfer/kv_connector/v1/mooncake/mooncake_connector.py
```diff
@@ -549,76 +669,192 @@ async def _sender_worker(self, sock: zmq.asyncio.Socket):
...
+            done, pending = await asyncio.wait(
+                wait_tasks,
+                timeout=envs.VLLM_MOONCAKE_ABORT_REQUEST_TIMEOUT,
+                return_when=asyncio.FIRST_COMPLETED,
+            )
+
+            if not done:
+                # Timeout, abort all pending requests.
+                for task in wait_tasks:
+                    task.cancel()
```
NickLucche, line 731: failure should be logged

vllm/distributed/kv_transfer/kv_connector/v1/mooncake/mooncake_connector.py
```diff
@@ -549,76 +669,192 @@ async def _sender_worker(self, sock: zmq.asyncio.Socket):
...
+                        err_reqs=err_reqs,
+                        err_msg=f"Mooncake transfer engine returned {ret_value}",
+                    )
+                    await sock.send_multipart(
+                        (identity, self._encoder.encode(response))
+                    )
+                    continue
+
+            for d_req_id, send_meta in ready_reqs:
+                # Todo: for heterogeneous TP (one P pairs to multiple D),
```
NickLucche: nit to make match search easier
```suggestion
                # TODO: for heterogeneous TP (one P pairs to multiple D),
```

vllm/distributed/kv_transfer/kv_connector/v1/mooncake/mooncake_connector.py
```diff
@@ -549,76 +669,192 @@ async def _sender_worker(self, sock: zmq.asyncio.Socket):
...
+                d_req_id, send_meta = task.result()
+                del pending_reqs[d_req_id]
+                # Do we still in reqs_need_send (not expired)?
+                if send_meta.transfer_id in self.reqs_need_send:
+                    # Mark it sending to avoid expiration.
+                    send_meta.sending += 1
+                    if not send_meta.need_send:
+                        self.resolve_need_send(send_meta, remote_tp_ranks)
+                    ready_reqs.append((d_req_id, send_meta))
+                # Otherwise (expired, very unlikely), forget it. Do not let D retry.
```
NickLucche: is this a case to log?

vllm/distributed/kv_transfer/kv_connector/v1/mooncake/mooncake_connector.py
```diff
@@ -795,69 +1032,172 @@ def get_finished(self) -> tuple[set[str] | None, set[str] | None]:
...
+    async def _start_load_kv(
+        self, reqs_to_recv: dict[EngineId, dict[ReqId, PullReqMeta]]
+    ):
+        for remote_engine_id, pull_metas in reqs_to_recv.items():
+            if remote_engine_id not in self._remote_agents:
+                asyncio.create_task(
+                    self.handle_new_engine_id(remote_engine_id, pull_metas)
+                )
+                continue
+            self.receive_kv(remote_engine_id, pull_metas)
```
NickLucche: nit:
```suggestion
            else:
                   self.receive_kv(remote_engine_id, pull_metas)
```

NickLucche (commented):
Looking good now, thanks for addressing previous concerns @dtcccc ! 
I only left a few very minor comments.
We should be looking to add a few tests to make the connector more maintainable though.

vllm/distributed/kv_transfer/kv_connector/v1/mooncake/mooncake_connector.py
dtcccc: In future when we support heteogenous TP, this check is still need. While the check in resolve_need_send() will be removed.

NickLucche (approved):

PR #31048, opened 2025-12-19
[NIXL] delay req_id clean for _req_to_save to finished_request 

xuechendi:
## Purpose

Follow up on https://github.com/vllm-project/vllm/pull/30419#discussion_r2631500643 to clean _req_to_save in `request_finished` instead of `build_connecter_meta`
@orozery @NickLucche, would like to know your thoughts.

accuracy is verified locally. I verified all requests gets cleaned by adding req_id print out locally.

## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
+++ b/vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
@@ -697,18 +697,6 @@ class NixlConnectorScheduler:
                 local_block_ids=new_block_id_groups[0],
                 kv_transfer_params=req.kv_transfer_params,
             )
-            assert scheduler_output.num_scheduled_tokens is not None
-            num_scheduled_tokens = scheduler_output.num_scheduled_tokens[req_id]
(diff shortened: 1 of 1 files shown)


vllm/distributed/kv_transfer/kv_connector/v1/nixl_connector.py
```diff
@@ -744,6 +732,9 @@ def request_finished(
         if not params:
             return False, None
 
+        # request is finished, now remove it from self._reqs_need_save
+        self._reqs_need_save.pop(request.request_id, None)
```
orozery, line 736: I've been trying to think if a possible future flow where request.kv_transfer_params are cleared during the life-cycle of a request.
For example, a future flow where the request transfer is handled by some other flow, which clears `kv_transfer_params` to indicate no further transfer-related action is needed (including delaying freeing of blocks).
In that case, we will not clear the request of `self._reqs_need_save`.

I think it would be safer to move this pop before the `if not params` check.

PR #30993, opened 2025-12-18
Bump Flashinfer to v0.6.1

elvischenv:
## Purpose
Bump Flashinfer to v0.6.1 when it is released.
API change: argument `tile_tokens_dim` has been removed from all TRTLLM MoE kernels.

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit a5e35ec5cd52bd4ca0c9e5a6fbb3a3b491e21ffa. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> **Upgrade FlashInfer to v0.6.0**
> 
> - Bumps FlashInfer to `0.6.0` in `docker/Dockerfile`, `Dockerfile.nightly_torch` (source build pinned to `v0.6.0`), and `requirements/cuda.txt`.
> 
> **API updates for MoE kernels**
> 
> - Removes `tile_tokens_dim` from all TRTLLM MoE call sites and related helpers; deletes `calculate_tile_tokens_dim` and associated imports/usages across `flashinfer_trtllm_moe.py`, `trtllm_moe.py`, `mxfp4.py`, `flashinfer_fp4_moe.py`, tests.
> 
> **Attention backend adjustments**
> 
> - Passes `o_data_type` through FlashInfer prefill/decode wrappers; updates fast-plan call to handle backend-specific arg lists (adds conditional args for `fa2`).
> 
> **Tests**
> 
> - Updates MXFP4 MoE tests to align with new FlashInfer interfaces (removes tile sizing logic).
> 
[...]

PR #31049, opened 2025-12-19
[CI] Add Qwen3-Next-FP8 to Blackwell model tests

vadiklyutiy:
Add Qwen3-Next-FP8 to Blackwell model tests. 

By default it's run on nightly testing only.


vllm/model_executor/layers/quantization/utils/flashinfer_fp4_moe.py
```diff
@@ -325,6 +325,7 @@ def flashinfer_trtllm_fp4_moe(
         local_expert_offset=layer.ep_rank * layer.local_num_experts,
         local_num_experts=layer.local_num_experts,
         routed_scaling_factor=None,
+        tile_tokens_dim=None,
```
vadiklyutiy, line 328: "Blackwell LM Eval Small Models" in nightly fails due to that.

https://buildkite.com/vllm/ci/builds/44620/steps/canvas?sid=019b4814-a0e0-493b-8805-1ef9ca77aebe

vadiklyutiy:
`buildkite/ci/pr/async-engine-inputs-utils-worker-config-test` fails on ToT

vllm/model_executor/layers/quantization/utils/flashinfer_fp4_moe.py
pavanimajety: you can rebase now to get it to work. Flashinfer 0.6.0 removes this parameter.

vadiklyutiy: But we haven't promoted to 0.6.0 yet. 
#30993 haven't merged. 
And Flashinfer haven't released 0.6.0 yet.

mgoin (approved):
Thanks!

PR #31052, opened 2025-12-20
[MoE Refactor][9/N] Use modular kernel for unquantized Triton MoE

zyongye:
Use `TritonExperts` modular kernel methods for unquantized moe method on GPUs.


vllm/model_executor/layers/fused_moe/unquantized_fused_moe_method.py
```diff
@@ -322,15 +326,20 @@ def forward_cuda(
                 apply_router_weight_on_input=layer.apply_router_weight_on_input,
             )
         else:
-            result = fused_experts(
+            self.kernel = mk.FusedMoEModularKernel(
```
robertgshaw2-redhat: we shouldnt initialize this during the forward pass, we should just call it during the forward pass

yewentao256 (commented):
CC @robertgshaw2-redhat

robertgshaw2-redhat (approved):

PR #31053, opened 2025-12-20
[CI] FIx `fixture 'siglip_attention_config' not found`

LucasWilkinson:
https://github.com/vllm-project/vllm/pull/30270 broke "Multi-Modal Models Test (Extended) 1" see: https://buildkite.com/vllm/ci/builds/44399/steps/canvas?sid=019b38a1-7696-47cf-b0fc-7151b865df52 with
```
[2025-12-19T23:01:06Z] E       fixture 'siglip_attention_config' not found
```

add it back


tjtanaa (approved):
Thank you @LucasWilkinson for the fix.
This test is not covered in PR CI. It is set as an optional test, so it was not detected.

PR #31057, opened 2025-12-20
[KVConnector] Auto-downgrade to PIECEWISE cudagraph mode for layerwise async ops

### LINKED ISSUE #29608: [Bug]: CUDA Graph Replay Skips KV Transfer Synchronization in Full Cache Hit Scenarios

yashwantbezawada:
## Purpose

Fixes #29608

When using async KV connectors with layerwise operations (`wait_for_layer_load`, `save_kv_layer`), these calls are made inside the `@maybe_transfer_kv_layer` decorator. During CUDA graph replay, only the captured GPU ops run - the decorator's Python code is skipped entirely. This means `wait_for_layer_load()` never executes, causing data races where attention kernels read KV cache data that hasn't finished loading.

This is a follow-up to #29755 based on @NickLucche's feedback to implement approach (b): restricting layerwise ops to piecewise mode rather than trying to trace them into CUDA graphs.

### Changes

1. Added `requires_piecewise_for_cudagraph(cls, extra_config)` class method to `KVConnectorBase_V1`
   - Returns `False` by default (most connectors don't use layerwise async ops)
   - Connectors override this if they need PIECEWISE mode

2. `LMCacheConnectorV1` overrides to return `True` when `use_layerwise=True`
   - Other connectors (Nixl, P2P, etc.) have no-op layerwise implementations, so they're fine with FULL mode

3. `VllmConfig.__post_init__()` checks this method and auto-downgrades to PIECEWISE when needed
   - Follows the same pattern used for DCP, pooler models, encoder-decoder models

## Test Plan

The fix is config-level validation that runs at engine initialization. Manual testing with LMCache + `use_layerwise=True` + CUDA graphs should show the warning message and the mode being downgraded.

## Test Result

Pre-commit checks pass. The implementation follows existing patterns in the codebase for handling CUDA graph incompatibilities.

### DIFF
--- a/vllm/config/vllm.py
+++ b/vllm/config/vllm.py
@@ -919,6 +919,33 @@ class VllmConfig:
                         CUDAGraphMode.FULL_DECODE_ONLY
                     )
 
+            # Check if KV connector requires PIECEWISE mode for CUDA graphs
+            if (
+                self.kv_transfer_config is not None
+                and self.kv_transfer_config.is_kv_transfer_instance
+                and self.compilation_config.cudagraph_mode.has_full_cudagraphs()
+            ):
+                # Lazy import to avoid circular dependencies
+                from vllm.distributed.kv_transfer.kv_connector.factory import (
+                    KVConnectorFactory,
+                )
+
+                connector_cls = KVConnectorFactory.get_connector_class(
+                    self.kv_transfer_config
+                )
+                if connector_cls.requires_piecewise_for_cudagraph(
+                    self.kv_transfer_config.kv_connector_extra_config
+                ):
+                    logger.warning_once(
+                        "KV connector %s requires PIECEWISE CUDA graph mode "
+                        "due to layerwise async operations that cannot be "
+                        "captured in CUDA graphs. "
+                        "Overriding cudagraph_mode from %s to PIECEWISE.",
+                        connector_cls.__name__,
+                        self.compilation_config.cudagraph_mode.name,
+                    )
+                    self.compilation_config.cudagraph_mode = CUDAGraphMode.PIECEWISE
+
             # disable cudagraph when enforce eager execution
             if self.model_config is not None and self.model_config.enforce_eager:
                 logger.info("Cudagraph is disabled under eager mode")
--- a/vllm/distributed/kv_transfer/kv_connector/v1/base.py
+++ b/vllm/distributed/kv_transfer/kv_connector/v1/base.py
@@ -543,6 +543,28 @@ class KVConnectorBase_V1(ABC):
             )
         return None
 
+    @classmethod
+    def requires_piecewise_for_cudagraph(cls, extra_config: dict[str, Any]) -> bool:
+        """
+        Check if this connector requires PIECEWISE CUDA graph mode.
+
+        Connectors that use asynchronous layer-by-layer operations
+        (wait_for_layer_load/save_kv_layer) should override this method
+        to return True when those operations are enabled. These operations
+        cannot be captured in CUDA graphs and will be skipped during replay,
+        causing data races. PIECEWISE mode allows Python code to execute
+        between graph pieces, ensuring proper synchronization.
+
+        Args:
+            extra_config: The kv_connector_extra_config dict from
+                KVTransferConfig.
+
+        Returns:
+            True if this connector requires PIECEWISE CUDA graph mode,
+            False otherwise.
+        """
+        return False
+
     def get_finished_count(self) -> int | None:
         """
         Get the count of requests expected to complete send/receive operations
--- a/vllm/distributed/kv_transfer/kv_connector/v1/lmcache_connector.py
+++ b/vllm/distributed/kv_transfer/kv_connector/v1/lmcache_connector.py
@@ -70,6 +70,16 @@ class LMCacheKVEvents(KVConnectorKVEvents):
 
 
 class LMCacheConnectorV1(KVConnectorBase_V1):
+    @classmethod
+    def requires_piecewise_for_cudagraph(cls, extra_config: dict[str, Any]) -> bool:
+        """
+        LMCache requires PIECEWISE CUDA graph mode when layerwise
+        operations are enabled. The wait_for_layer_load and save_kv_layer
+        methods perform actual async synchronization that cannot be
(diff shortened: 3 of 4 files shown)


vllm/distributed/kv_transfer/kv_connector/v1/base.py
```diff
@@ -533,6 +533,28 @@ def get_required_kvcache_layout(cls, vllm_config: "VllmConfig") -> str | None:
...
+
+        Args:
+            extra_config: The kv_connector_extra_config dict from
+                KVTransferConfig.
+
+        Returns:
+            True if this connector requires PIECEWISE CUDA graph mode,
+            False otherwise.
+        """
+        return False
```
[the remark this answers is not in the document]
yashwantbezawada: Good catch, thanks! Added the override for MultiConnector to propagate the requirement from child connectors.

vllm/config/vllm.py
```diff
@@ -717,6 +717,32 @@ def has_blocked_weights():
...
+                # Lazy import to avoid circular dependencies
+                from vllm.distributed.kv_transfer.kv_connector.factory import (
+                    KVConnectorFactory,
+                )
+
+                connector_cls = KVConnectorFactory.get_connector_class(
+                    self.kv_transfer_config
+                )
+                if connector_cls.requires_piecewise_for_cudagraph(
+                    self.kv_transfer_config.kv_connector_extra_config
```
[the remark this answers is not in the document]
yashwantbezawada: Fixed in the latest push - MultiConnector now checks all child connectors.

yashwantbezawada:
Sounds good, thanks for the feedback! Will leave it as-is with the current warning.

vllm/config/vllm.py
```diff
@@ -717,6 +717,32 @@ def has_blocked_weights():
...
+                ):
+                    logger.warning_once(
+                        "KV connector %s requires PIECEWISE CUDA graph mode "
+                        "due to layerwise async operations that cannot be "
+                        "captured in CUDA graphs. "
+                        "Overriding cudagraph_mode to PIECEWISE.",
+                        connector_cls.__name__,
+                    )
```
ProExpertProg, line 946: Let's slightly improve the warning, something like this, feel free to edit:
```suggestion
                    logger.warning_once(
                        "KV connector %s requires PIECEWISE CUDA graph mode "
                        "due to layerwise async operations that cannot be "
                        "captured in CUDA graphs. "
                        "Overriding cudagraph_mode from %s to PIECEWISE, which "
                        "might reduce performance. "
                        "You may want to use a different kvcache connector "
                        "with support for full cudagraphs for better performance",
```
                        connector_cls.__name__,
                    )
```

**Review by ProExpertProg — APPROVED** [review]

**Comment by yashwantbezawada** [comment]
Hey @ProExpertProg, any chance this can be merged? Approved on Jan 5 and no conflicts. Thanks!

File: vllm/config/vllm.py [hunk 4]
```diff
@@ -849,6 +849,36 @@ def has_blocked_weights():
                         CUDAGraphMode.FULL_DECODE_ONLY
                     )
 
+            # Check if KV connector requires PIECEWISE mode for CUDA graphs
+            if (
+                self.compilation_config.cudagraph_mode.has_full_cudagraphs()
+                and self.kv_transfer_config is not None
+                and self.kv_transfer_config.is_kv_transfer_instance
```
NickLucche: nit: kv_transfer check first

vllm/config/vllm.py
```diff
@@ -849,6 +849,36 @@ def has_blocked_weights():
...
+                    self.kv_transfer_config.kv_connector_extra_config
+                ):
+                    logger.warning_once(
+                        "KV connector %s requires PIECEWISE CUDA graph mode "
+                        "due to layerwise async operations that cannot be "
+                        "captured in CUDA graphs. "
+                        "Overriding cudagraph_mode from %s to PIECEWISE, which "
+                        "might reduce performance. "
+                        "You may want to use a different kvcache connector "
+                        "with support for full cudagraphs for better performance",
```
NickLucche: I think this performance bit is a bit too "scary", suggesting that a connector that moves all the layers at once might actually be more performant than one that "streams" them.

NickLucche (commented):
Thanks for the follow-up @yashwantbezawada !
Left one minor note.

PS Re-running tests from main where fixes have hopefully now landed

vllm/config/vllm.py
yashwantbezawada: Good call, reordered to check kv_transfer_config first.

vllm/config/vllm.py
yashwantbezawada: Agreed, simplified the warning. Removed the performance suggestion entirely - just states the override now.

PR #31061, opened 2025-12-20
[Quantization] add marlin w4a8/w8a8 check

jinzhen-lin:
This PR add necessary checks for marlin w4a8 / w8a8.

fix https://github.com/vllm-project/vllm/issues/31033 (MXFP4 x INT8 is not supported, we just need a useful error message.)


mgoin (approved):
Thank you!

yewentao256 (approved):
LGTM, thanks for the work!

PR #31071, opened 2025-12-20
[Doc] Clarify FP8 KV cache computation workflow

westers:
## Summary
Fixes #31023 - Clarifies how FP8 KV cache quantization and computation works

Adds a detailed technical explanation to address user confusion about whether the attention-value multiplication operates on FP8 tensors directly or uses dequantized higher-precision values.

## Changes
Added new section "How FP8 KV Cache Works" explaining the complete workflow:

1. **Storage**: KV tensors quantized to FP8 for memory efficiency
2. **Retrieval**: Dequantized back to FP16/BF16 when needed
3. **Attention**: Computation uses higher-precision dequantized values

## Key Clarification
The documentation now explicitly states that:
- FP8 quantization reduces memory usage during storage
- Attention computation operates on dequantized higher-precision values (not FP8)
- This explains why there are "no fused dequantization and attention operations" currently

## Technical Investigation
- Reviewed `vllm/model_executor/layers/quantization/kv_cache.py` 
- Examined `csrc/cache.h` and found `gather_and_maybe_dequant_cache()` function
- Confirmed dequantization happens during cache retrieval before attention

This provides users with a clear understanding of the accuracy-performance tradeoffs.

### DIFF
--- a/docs/design/paged_attention.md
+++ b/docs/design/paged_attention.md
@@ -139,18 +139,18 @@ token data.
 const scalar_t* q_ptr = q + seq_idx * q_stride + head_idx * HEAD_SIZE;
 ```
 
-<figure markdown="span">
-  ![](../assets/design/paged_attention/query.png){ align="center" alt="query" width="70%" }
-</figure>
+<p align="center">
+  <img src="../assets/design/paged_attention/query.png" alt="query" width="70%" />
+</p>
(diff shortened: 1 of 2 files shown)


yewentao256 (approved):
LGTM, thanks for the work!

PR #30897, opened 2025-12-17
[NVFP4][Perf] Tune NVFP4 input quant kernel for small batch size

mgoin:
## Purpose

We discovered that the nvfp4 input quant operation was severely bottlenecking performance for dense models at small batch sizes. While investigating the flashinfer impl, I noticed that there is logic to [increase the grid size for small M](https://github.com/flashinfer-ai/flashinfer/blob/f0592411a3478d0548d62f7c877046499caf8d95/csrc/nv_internal/cpp/kernels/quantization.cu), based on the swizzled layout to help out with the padding ops. Adding that "effective row" calculation greatly improved bs<32 performance.

I also found that there was some redundant computation in the `cvt_quant_to_fp4_get_sf_out_offset` inner loop, so pulled that out and replaced many of the division operations that `nvcc` wasn't optimizing into bitwise ops.

I added a new benchmark script for the quantization op itself for easy microbenchmarks: `python benchmarks/kernels/bench_nvfp4_quant.py`

Screenshot 2025-12-17 at 5 01 17 PM
Screenshot 2025-12-17 at 5 01 56 PM

#### Dense model

Here is an end-to-end bs=1 decode benchmark for Qwen3-8B-NVFP4. The output token throughput grows from `220 -> 257 -> 276 toks/s`, a 25% increase.

```
vllm serve RedHatAI/Qwen3-8B-NVFP4
vllm bench serve --input-len 10 --output-len 1000 --num-prompts 1
```

Screenshot 2025-12-17 at 4 42 30 PM

#### MoE model

Here is an end-to-end bs=1 decode benchmark for Qwen3-30B-A3B-NVFP4. The output token throughput grows from `137 -> 145 toks/s`, a 6% increase.

```
vllm serve nvidia/Qwen3-30B-A3B-NVFP4
vllm bench serve --input-len 10 --output-len 1000 --num-prompts 1
```

Screenshot 2025-12-17 at 4 52 57 PM

#### Dense model

```
vllm serve RedHatAI/Qwen3-8B-NVFP4
python tests/evals/gsm8k/gsm8k_eval.py --port 8000

# MAIN
Results:
Accuracy: 0.854
Invalid responses: 0.001
Total latency: 11.811 s
Questions per second: 111.674
Total output tokens: 156997
Output tokens per second: 13292.225

# PR
Results:
Accuracy: 0.854
Invalid responses: 0.001
Total latency: 11.975 s
Questions per second: 110.147
Total output tokens: 157018
[...]

PR #31097, opened 2025-12-21
[FIX] FP4 quantization kernel padding initialization bug

danielafrimi:
### Summary
This PR fixes a bug in the FP4 quantization kernel (`nvfp4_quant_kernels.cu`) related to incorrect padding initialization for scale factors in padded tensor regions.

#### Unified kernel logic for padding handling
The original implementation had separate loops:
1. First loop: attempted to zero padding regions for scale factors (was a bug in the inner loop) 
2. Second loop: processed actual input data

**Problem:** The padding zeroing logic had addressing issues and didn't correctly initialize all necessary scale factor memory locations.

Created Unified into a single loop that iterates over all rows/columns including padded regions (`sf_m` x `num_padded_cols`), using zero-filled vectors for out-of-bounds regions and loading actual data for valid regions. All data is processed through quantization to ensure scale factors are properly computed, but output is only written for valid (non-padded) regions to avoid buffer overruns.

### Performance Impact
Benchmarked on **Nemotron Nano v3** with 10k requests, 1K/1K input/output, max concurrency of 512:

| Metric | Before | After |
|--------|--------|-------|
| **TPOT (mean)** | 30.54ms | 30.26ms 
| **ITL (mean)** | 30.51ms | 30.23ms 
| **Output Throughput** |  15991.81 tok/s  | 16099.50 tok/s

Performance remains effectively unchanged while fixing the correctness issue.

### Testing
- All related tests pass

Regarding the latest changes in main (following this [PR](https://github.com/vllm-project/vllm/pull/30897)) 
`numKTiles` is calculated from ACTUAL `numCols`, BUT it automatically accounts for the padding needed, since it calculates using `(numCols + 63) / 64` (always rounds UP).

The padding formula `num_padded_cols` = sf_n_int * 4 * 16` always rounds to a multiple of 64
`numCols` must be a multiple of 16, both formulas will always agree on the padded size.

@mgoin Do you agree 

In addition, as @mgoin mentioned, i ran the script `benchmarks/kernels/bench_nvfp4_quant.py`
and got the following performance: 
image

### DIFF
--- a/csrc/quantization/fp4/nvfp4_quant_kernels.cu
+++ b/csrc/quantization/fp4/nvfp4_quant_kernels.cu
@@ -77,6 +77,7 @@ __global__ void __launch_bounds__(512, VLLM_BLOCKS_PER_SM(512))
       int elem_idx = colIdx * CVT_FP4_ELTS_PER_THREAD;
 
       PackedVec in_vec;
+      int64_t inOffset = rowIdx * (numCols / CVT_FP4_ELTS_PER_THREAD) + colIdx;
 
       // If we are outside valid rows OR outside valid columns -> Use Zeros
       if (rowIdx >= numRows || elem_idx >= numCols) {
@@ -84,8 +85,6 @@ __global__ void __launch_bounds__(512, VLLM_BLOCKS_PER_SM(512))
 
       } else {
         // Valid Region: Load actual data
-        int64_t inOffset =
-            rowIdx * (numCols / CVT_FP4_ELTS_PER_THREAD) + colIdx;
         in_vec = reinterpret_cast<PackedVec const*>(in)[inOffset];
       }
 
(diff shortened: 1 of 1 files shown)


csrc/quantization/fp4/nvfp4_quant_kernels.cu
```diff
@@ -61,37 +61,51 @@ __global__ void __launch_bounds__(512, VLLM_BLOCKS_PER_SM(512))
...
+      auto out_val =
           cvt_warp_fp16_to_fp4<Type, UE8M0_SF>(in_vec, global_scale, sf_out);
+
+      // We do NOT write output for padding because the 'out' tensor is not
+      // padded.
+      if (rowIdx < numRows && elem_idx < numCols) {
+        // Get the output tensor offset.
+        // Same as inOffset because 8 elements are packed into one uint32_t.
+        int64_t inOffset =
+            rowIdx * (numCols / CVT_FP4_ELTS_PER_THREAD) + colIdx;
```
mgoin: nit: we are duplicating the inOffset calc here, we could just pull it out of the earlier conditional

mgoin (approved):
Thanks for cleaning this up! LGTM just a small nit

PR #31099, opened 2025-12-21
 [FIX] Always support TP > 4 for FP4 Gemm

danielafrimi:
## Summary

This PR enables FP4 (NVFP4) quantization to work with TP  >=  4

## Background

Previously, FP4 quantized models would fail to initialize with TP=4/8 due to kernel alignment requirements. The FlashInfer-CUTLASS FP4 GEMM kernels require the N/K-dimension to be divisible by 32.

## Changes

This PR makes the following changes to support TP >= 4, by padding the weights accordingly. In addition, in cases where we pad the K dim of the weights, we need to pad the activation in the forward pass. 

## Tests
Tested the following on Nemotron Nano v3 on both TP=4/8 on the mmlu dataset. 

This Fix is also related to the quant kernel fix - [PR](https://github.com/vllm-project/vllm/pull/31097)

More info:

1. The CUTLASS FP4 GEMM kernel (`nvfp4_scaled_mm_kernels.cu`) requires both K and N matrix dimensions to be divisible by 32 for aligned memory access and efficient tensor core operations. - without it, we get the error of` n % alignment == 0 (16 vs. 0) : Expected n to be divisible by 32` (in the file nvfp4_scaled_mm_kernels.cu line 274)

2. Pad the rows/columns by a multiple of 32 instead of taking the scales' dim
needs to pad the activation where we pad on the K dim of the weights, since it won't be compatible with the GEMM itself, therefore we pad it (it's happened in nemotron nano v3 for tp=8 )

### DIFF
--- a/vllm/model_executor/layers/quantization/modelopt.py
+++ b/vllm/model_executor/layers/quantization/modelopt.py
@@ -85,6 +85,9 @@ from vllm.model_executor.layers.quantization.utils.quant_utils import (
     kFp8StaticTokenSym,
     kNvfp4Dynamic,
     kNvfp4Static,
+    pad_nvfp4_activation_for_cutlass,
+    pad_nvfp4_weight_for_cutlass,
+    slice_nvfp4_output,
     swizzle_blockscale,
 )
 from vllm.model_executor.layers.quantization.utils.w8a8_utils import (
@@ -1276,9 +1279,16 @@ class ModelOptNvFp4LinearMethod(LinearMethodBase):
             layer.weight_scale = Parameter(weight_scale, requires_grad=False)
             layer.weight = Parameter(weight, requires_grad=False)
         else:
+            # Swizzle block scales and pad the packed NVFP4 weights for kernel
+            # alignment (CUTLASS/FlashInfer require K and N divisible by 32).
             swizzled_weight_scale = swizzle_blockscale(layer.weight_scale)
             layer.weight_scale = Parameter(swizzled_weight_scale, requires_grad=False)
-            layer.weight = Parameter(layer.weight.data, requires_grad=False)
+
+            weight, weights_padding_cols = pad_nvfp4_weight_for_cutlass(
+                layer.weight.data
+            )
+            layer.weights_padding_cols = weights_padding_cols
+            layer.weight = Parameter(weight, requires_grad=False)
 
     def apply(
         self,
@@ -1300,7 +1310,6 @@ class ModelOptNvFp4LinearMethod(LinearMethodBase):
             )
 
         output_dtype = x.dtype
-        output_shape = [x.shape[0], layer.weight.shape[0]]
 
         # quantize BF16 or FP16 to (FP4 and interleaved block scale)
         x_fp4, x_blockscale = scaled_fp4_quant(x, layer.input_scale_inv, self.backend)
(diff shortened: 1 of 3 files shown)


vllm/model_executor/layers/quantization/modelopt.py
```diff
@@ -1383,19 +1428,39 @@ def apply(
...
-        )
         if self.backend.startswith("flashinfer-"):
             backend_name = self.backend[len("flashinfer-") :]
+
+            # Match packed-K bytes between activations and weights using
+            # pre-calculated padding
+            pad_k_bytes = getattr(layer, "weights_padding_cols", 0)
+            output_shape = [x.shape[0], layer.output_size_per_partition]
+
+            x_fp4 = torch.nn.functional.pad(x_fp4, (0, pad_k_bytes)).contiguous()
```
pavanimajety: why do we need to `x_fp4`?

vllm/model_executor/layers/quantization/modelopt.py
```diff
@@ -1383,19 +1173,39 @@ def apply(
...
+                output_dtype,
+            )
             out = flashinfer_scaled_fp4_mm(*mm_args, backend=backend_name)
+
+            # Slice output to remove padding if weight was padded in N dimension
+            if out.shape[1] != output_shape[1]:
+                out = out[:, : output_shape[1]].contiguous()
         else:
             assert self.backend == "cutlass"
+            mm_args = (
```
pavanimajety: Why aren't we padding this path?

vllm/model_executor/layers/quantization/modelopt.py
```diff
@@ -1383,19 +1173,39 @@ def apply(
...
-        )
         if self.backend.startswith("flashinfer-"):
             backend_name = self.backend[len("flashinfer-") :]
+
+            # Match packed-K bytes between activations and weights using
+            # pre-calculated padding
+            pad_k_bytes = getattr(layer, "weights_padding_cols", 0)
+            output_shape = [x.shape[0], layer.output_size_per_partition]
+
+            x_fp4 = torch.nn.functional.pad(x_fp4, (0, pad_k_bytes)).contiguous()
```
pavanimajety: Can we check if there is a call to the torch function when pad_k_bytes = 0? if there is one, can we guard the call with `if pad_k_bytes > 0` ?

pavanimajety (commented):
Thanks for the PR @danielafrimi! 
Could you please clarify the PR title to mention "TP > 4 for NVFP4 padding scenarios"?

vllm/model_executor/layers/quantization/modelopt.py
danielafrimi: fixed

vllm/model_executor/layers/quantization/modelopt.py
danielafrimi: fixed as well

pavanimajety (dismissed):
Thanks for the fixes, @danielafrimi. 
DCO seems to be showing root with no author name when enabling auto-merge. Could you please do the following with your name, thanks! - 

```

git commit --amend --author="Your Name <your.email@example.com>" --signoff

git push --force-with-lease

```

mgoin (changes requested):
This issue isn't really to fix TP>4, it is just to fix any time a model's weight dim isn't divisible with TP for NVFP4. I also really dislike introducing activation padding to the inner dim since we have to do this `.contiguous()` call twice, which is going to hurt performance. It seems this limitation is not needed for the other kernels, so why don't we just dispatch to a different kernel that can support this without padding?

danielafrimi:
@mgoin Thanks for the review! 
Regarding the Nemotron issue, my concern is that other future models may have the same issues as Nemtron. So I think it's better not to conclude only this model. But if you think this is better i can do it. 
Meanwhile as you suggested added the functional support into the utils file, and added this logic into the `vllm/model_executor/layers/quantization/compressed_tensors/schemes/compressed_tensors_w4a4_nvfp4.py` as well. 

Let me know what you think

mgoin (approved):

PR #31104, opened 2025-12-22
[BugFix] LoRA: Support loading base_layer of experts

HollowMan6:
## Purpose

This PR fixes weight loading when LoRA is enabled, i.e., we have `base_layer` added to the:

`model.layers.0.mlp.experts.0.up_proj.weight` -> `model.layers.0.mlp.experts.0.up_proj.base_layer.weight`

Currently before this fix, the patched code will handled this as:
`model.layers.0.mlp.experts.w13_base_layer.weight`, which is wrong and
it should actually be `model.layers.0.mlp.experts.base_layer.w13_weight`

## Test Plan

Test on Qwen3 30B A3B

## Test Result

Looks good.

---
<details>

</details>

<sub>✨ Presented to you with <a href="https://macaron.im/mindlab">Mind Lab</a> - A Lab for Experiential Intelligence.</sub>


vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -2037,13 +2038,17 @@ def make_expert_params_mapping(
...
         return [
             # (param_name, weight_name, expert_id, shard_id)
             (
-                "experts.w13_"
+                f"experts.{base_layer}w13_"
                 if weight_name in [ckpt_gate_proj_name, ckpt_up_proj_name]
-                else "experts.w2_",
-                f"experts.{physical_to_logical_map[expert_id]}.{weight_name}.",
+                else f"experts.{base_layer}w2_",
+                f"experts.{physical_to_logical_map[expert_id]}.{weight_name}.{base_layer}",
```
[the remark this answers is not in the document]
HollowMan6: This is not correct

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -2037,13 +2038,17 @@ def make_expert_params_mapping(
...
 
+        base_layer = ""
+        if is_lora_enabled:
+            base_layer = "base_layer."
+
         return [
             # (param_name, weight_name, expert_id, shard_id)
             (
-                "experts.w13_"
+                f"experts.{base_layer}w13_"
```
jeejeelee, line 2038: Can you test   [test_qwen3moe_tp](https://github.com/vllm-project/vllm/blob/v0.13.0/tests/lora/test_qwen3moe_tp.py) locally? 
Generally, enabling LoRA doesn't require changing the base model's layer names

HollowMan6: Hi @jeejeelee, thank you for reviewing and pointing this out, originally I was intended for solutions like https://github.com/vllm-project/vllm/commit/f9008c92f6ead3b9d018f84924757e9787d186d9, but later I though this is not a good solution so I turn into something like https://github.com/vllm-project/vllm/commit/210bc7b4430864627748de262f6e7c1161220af1 (which is the solution this review is targeting at). Now I found that https://github.com/vllm-project/vllm/commit/210bc7b4430864627748de262f6e7c1161220af1 actually breaks [test_qwen3moe_tp](https://github.com/vllm-project/vllm/blob/v0.13.0/tests/lora/test_qwen3moe_tp.py), so I revert back to https://github.com/vllm-project/vllm/commit/f9008c92f6ead3b9d018f84924757e9787d186d9 since it's less intrusive (supports either with `base_layer` or without), please let me know what you think!

The issue here is that, for [verl](https://github.com/volcengine/verl/pull/4632), during the weight update (refit), with lora, when the weight has already loaded and we want to update the base layers, the current code is problematic as described in https://github.com/vllm-project/vllm/pull/31104#issue-3751854142, though weight loading during initialization is fine.

hmellor (dismissed):
We should not be duplicating this code in every model. It should be abstracted to a util.

Also, please make sure that the fix is also applied to https://github.com/vllm-project/vllm/blob/73cfb7a722e36c8429e7fdb5306be2d981c3f8dd/vllm/model_executor/layers/fused_moe/layer.py#L1366

vllm/model_executor/layers/fused_moe/layer.py
jeejeelee: > The issue here is that, for https://github.com/volcengine/verl/pull/4632, during the weight update (refit), with lora, when the weight has already loaded and we want to update the base layers, the current code is problematic as described in https://github.com/vllm-project/vllm/pull/31104#issue-3751854142, though weight loading during initialization is fine.

I understand that linear  layer should also have this issue. Have you tested it? If there's no similar issue as with MoE, is it because of [lora linear weight](https://github.com/vllm-project/vllm/blob/main/vllm/lora/layers/base_linear.py#L141)?

HollowMan6: Non-experts linear layers also have `.base_layer` in weight names, and they don't have any issue. The actual bug is not related to the implementation, but as mentioned in the PR description, https://github.com/vllm-project/vllm/pull/31104#issue-3751854142 The current weight remapping (naming handling) logic for experts linear layers, which is introduced for handling fusedMoE, is that, it didn't consider the possibility of having `.base_layer` in experts weight names. As a result, when a weight named `model.layers.0.mlp.experts.0.up_proj.base_layer.weight` comes in, the current weight remapping (naming handling) logic will remap this as: `model.layers.0.mlp.experts.w13_base_layer.weight`, but to make it correct, the weight remapping (naming handling) logic should remap this as `model.layers.0.mlp.experts.base_layer.w13_weight` to match what we have in `params_dict`. Non-experts linear layers don't have this remapping logic, so they are OK. @jeejeelee

jeejeelee: Thank you very much for your patient explanation. I misunderstood before, and now I think these modifications are ok, but could you test it locally? It seems this PR has a circular import issue.

HollowMan6: Thank you for pointing this out, @jeejeelee ! I didn't notice that importing `vllm.model_executor.model_loader.weight_utils` directly in `vllm/model_executor/layers/fused_moe/layer.py` will cause circular import. I was testing this by cherrypicking changes for only `vllm/model_executor/model_loader/weight_utils.py` and `vllm/model_executor/models/qwen3_moe.py` on top of my local vllm 0.13.0 installation and testing them using Qwen3-30B-A3B. I now fix this by importing `vllm.model_executor.model_loader.weight_utils` inside `load_weights` function for `vllm/model_executor/layers/fused_moe/layer.py`, and I have tested this again using both [test_qwen3moe_tp](https://github.com/vllm-project/vllm/blob/v0.13.0/tests/lora/test_qwen3moe_tp.py) and when integrate with https://github.com/volcengine/verl/pull/4632 and they all work fine.

jeejeelee (approved):
LGTM once CI is green

HollowMan6:
Thank you @jeejeelee! Now all the CIs should have already been passed, but auto-merge (squash) is not merging it, maybe this needs some manual work.

jeejeelee:
cc @hmellor

hmellor (commented):
My original concerns have been addressed so I'll dismiss my blocking review.

I have had a thought though. It seems like this logic should live in `FusedMoE.make_expert_params_mapping`, rather than being a separate utility that does additional remapping in `load_weights`.

Would it be possible to account for this LoRA special case in `make_expert_params_mapping` instead?

HollowMan6:
> Thanks for the context, what is the error you see when using the solution form [210bc7b](https://github.com/vllm-project/vllm/commit/210bc7b4430864627748de262f6e7c1161220af1)?

@hmellor During weight initialization loading from disk, those paramaters will be without `.base_layer` even if lora is enabled, and I don't think we have a robust way for checking that right in `make_expert_params_mapping`.

hmellor:
I see, so the issue is that `make_expert_params_mapping` is too early because LoRA hasn't actually added `.base_layer` yet?

hmellor:
I don't think this is the reason `make_expert_params_mapping` is called inside `load_weights`

HollowMan6:
Current CI failures don't seem to be caused by this PR.

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -2020,13 +2021,19 @@ def make_expert_params_mapping(
             )
         )
 
+        base_layer = (
+            "base_layer."
+            if any(".base_layer." in name for name, _ in model.named_parameters())
+            else ""
+        )
```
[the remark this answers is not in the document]
HollowMan6: no, this won't work as mentioned in https://github.com/vllm-project/vllm/pull/31104#issuecomment-3706127655, ".base_layer." should be safe.

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -2020,13 +2021,19 @@ def make_expert_params_mapping(
...
         return [
             # (param_name, weight_name, expert_id, shard_id)
             (
-                "experts.w13_"
+                f"experts.{base_layer}w13_"
                 if weight_name in [ckpt_gate_proj_name, ckpt_up_proj_name]
-                else "experts.w2_",
-                f"experts.{physical_to_logical_map[expert_id]}.{weight_name}.",
+                else f"experts.{base_layer}w2_",
+                f"experts.{physical_to_logical_map[expert_id]}.{weight_name}.{base_layer}",
```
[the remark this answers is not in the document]
HollowMan6: In vLLM, all-linear will be used for base models so that LoRA request can work, so this should be fine.

HollowMan6:
> What I don't like is that it means that every MoE model has to have this line added to it just for VeRL+LoRA.

@hmellor I think I just managed to remove this requirement by modifying `make_expert_params_mapping` only, so that now we determine them by checking if any parameter name contains `.base_layer.`, which suggests that LoRA is enabled. I have tested this with LoRA enabled, either with weight initialization loading or weight refit in https://github.com/volcengine/verl/pull/4632, and they work fine. Please let me know your comments.

cc: @jeejeelee

PR #31107, opened 2025-12-22
[Platform] Fix DP group init error

wangxiyuan:
When implement `stateless_init_device_torch_dist_pg` on a custom platform using non-gloo backend, we find that the process will hang. It's because that the communication check still using `cpu` even we changed the communication backend to non-gloo. This PR refactor the logic of `stateless_init_torch_distributed_process_group` to make sure non-gloo backend works as expect.

### DIFF
--- a/vllm/platforms/interface.py
+++ b/vllm/platforms/interface.py
@@ -11,13 +11,14 @@ from typing import TYPE_CHECKING, Any, NamedTuple, Optional
 
 import numpy as np
 import torch
+from torch.distributed import PrefixStore, ProcessGroup
 
 from vllm.attention.backends.registry import AttentionBackendEnum
 from vllm.logger import init_logger
+from vllm.utils.system_utils import suppress_stdout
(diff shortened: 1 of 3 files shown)


vllm/platforms/interface.py
```diff
@@ -612,16 +614,44 @@ def get_static_graph_wrapper_cls(cls) -> str:
...
         group_rank: int,
         group_size: int,
         timeout: timedelta,
     ) -> "ProcessGroup":
         """
-        Init platform-specific torch distributed process group.
-        """
-        raise NotImplementedError
+        Init platform-specific torch distributed process group. Gloo is used
+        by default.
```
ilmarkov, line 624: Is there possibility to choose non-default backend?

ilmarkov (changes requested):
Is there a way to add a test to verify that your changes actually fix the issue on custom platform with non-gloo backend?

wangxiyuan:
@ilmarkov of cause, i'll post more test result in these days. Please wait more.

PR #31125, opened 2025-12-22
[UX] improve profiler error message

BoyuanFeng:
Prior to this PR, when a user runs `vllm bench latency --profile`, the error would be `Exception: Call to profile method failed: Profiling is not enabled.` which does not show actionable suggestions.

This PR improves the log by giving an actionable example:

```
Exception: Call to profile method failed: Profiling is not enabled. Please set --profiler-config to enable profiling. A minimal example is '--profiler-config.profiler=torch --profiler-config.torch_profiler_dir=LOCAL_DIR_PATH'
```

cc @benchislett @hmellor


yewentao256 (approved):
LGTM, thanks for the work!

mgoin (approved):

PR #30200, opened 2025-12-07
[Frontend] Introduce Renderer for processing chat messages (using `ModelConfig`)

DarkLight1337:
Ready for review.

## Purpose

- Prototype an interface, `vllm.renderers.RendererLike`, to process chat messages into engine inputs.
- Introduce `RENDERER_REGISTRY` which lazily registers renderers to avoid circular import problem.
- Move implementation-specific chat utils to the corresponding renderer in `vllm.renderers`.
- Initialize the renderer in `InputPreprocessor`, replacing the tokenizer initialization inside `LLMEngine` and `AsyncLLM`.
- Replace `EngineClient.get_tokenizer()` with `EngineClient.renderer.get_tokenizer()` to avoid unnecessary async.
- Update tests accordingly, and move some tests into a new directory `tests/renderers` that is run under `Async Engine, Inputs, Utils, Worker, Config Test (CPU)`.

Towards #22880 and #23873

Future work:

- Merge `CompletionRenderer` with this implementation of `Renderer`.
- Move `apply_chat_template` from tokenizer into renderer.
- Move microbatch tokenizer into renderer.
- Since each renderer uses a specific tokenizer, we can deprecate `TokenizerRegistry` in favor of `RENDERER_REGISTRY` and rename `tokenizer_mode` to `renderer_mode`.
- Split out renderer-specific fields from `ModelConfig`.

## Test Result

---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit 3c3c2ab9f0d04c2d47d4e13d11a3abcbe6882d3d. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
[...]

PR #31138, opened 2025-12-22
[Mistral common] Ensure all functions are imported from the top & only use public methods

patrickvonplaten:
This PR makes sure that only public methods are used and that all imports are done at the top


vllm/tokenizers/mistral.py
```diff
@@ -3,17 +3,28 @@
...
+from mistral_common.tokens.tokenizers.tekken import Tekkenizer
+
 from vllm.entrypoints.chat_utils import ChatCompletionMessageParam
 from vllm.entrypoints.openai.protocol import ChatCompletionRequest
 from vllm.logger import init_logger
 
 from .protocol import TokenizerLike
 
 if TYPE_CHECKING:
-    from mistral_common.protocol.instruct.request import (
```
patrickvonplaten, line 13: mistral_common is very light weight (has no heavy dependencies) and is a necessary requirement in `common.txt` => so I think it's cleaner to directly import at the top

DarkLight1337 (commented):
I think this will cause a problem for people who don't have `mistral_common` installed. We currently still import `tokenizers.mistral` from various places (in particular chat-related entrypoints). Need to wait for #30200 to make those imports lazy.

vllm/tokenizers/mistral.py
```diff
@@ -271,36 +270,19 @@ def __init__(self, tokenizer: "MistralCommonBackend") -> None:
         self._vocab_dict = dict(sorted(self._vocab_dict.items(), key=lambda x: x[1]))
 
         # Cache special tokens for faster access.
+        self._vocab = self.tokenizer.vocab()
+        # Vocab sorted by token id.
+        self._max_token_id = self.vocab_size - 1
+
         self._special_token_ids = self._get_special_token_ids()
```
juliendenize:
```suggestion
        # Vocab sorted by token id.
        self._vocab = self.tokenizer.vocab()
        self._max_token_id = self.vocab_size - 1

        # Cache special tokens for faster access.
        self._special_token_ids = self._get_special_token_ids()
```

vllm/tokenizers/mistral.py
```diff
@@ -271,36 +270,19 @@ def __init__(self, tokenizer: "MistralCommonBackend") -> None:
...
-            special_ids = {t["rank"] for t in self.tokenizer._all_special_tokens}
-        elif self.is_spm:
-            assert isinstance(self.tokenizer, SentencePieceTokenizer), type(
-                self.tokenizer
-            )
-            special_ids = self.tokenizer._control_tokens
-        else:
-            raise ValueError(f"Unknown tokenizer type: {type(self.tokenizer)}")
-        return sorted(special_ids)
+        return sorted(self.tokenizer.is_special(i) for i in range(len(self._vocab)))
```
juliendenize:
```suggestion
        return [i for i in range(len(self._vocab)) if self.tokenizer.is_special(i)]
```

juliendenize (commented):
Left some comments ! Looks better with `is_special` API !

DarkLight1337 (approved):
Ok, let's merge this and see if anyone complains 😅

PR #31144, opened 2025-12-22
fix multiconnector for multi connector use push kv connector

### LINKED ISSUE #31145: [Bug]: [P/D] multi-connector cannot be used together with P2pNcclConnector that uses a put mode(push kv cache from P node to D node)
### fixpr
https://github.com/vllm-project/vllm/pull/31144
### 🐛 Describe the bug

```python
def get_num_new_matched_tokens(
        self,
        request: "Request",
        num_computed_tokens: int,
    ) -> tuple[int | None, bool]:
        to_return = (0, False)
        for i, c in enumerate(self._connectors):
            toks, load_async = c.get_num_new_matched_tokens(
                request, num_computed_tokens
            )
            # If there is a connector still looking up the matches,
            # we return None to indicate that we are not done yet.
            if toks is None:
                return (None, False)
            # The first connector that has new matched tokens will be assigned
            # to this request.
            if to_return[0] == 0 and toks > 0:
                self._requests_to_connector[request.request_id] = i
                to_return = (toks, load_async)
        return to_return

def update_state_after_alloc(
        self, request: "Request", blocks: "KVCacheBlocks", num_external_tokens: int
    ):
        chosen_connector = self._requests_to_connector.get(request.request_id, -1)
        empty_blocks = blocks.new_empty()
        for i, c in enumerate(self._connectors):
            if i == chosen_connector:
                # Forward call to the chosen connector (if any).
                c.update_state_after_alloc(request, blocks, num_external_tokens)
            else:
                # Call with empty blocks for other connectors.
                c.update_state_after_alloc(request, empty_blocks, 0)
```
**For these two functions in the multi-connector, if a KV connector using the push scheme is employed, get_num_new_matched_tokens will return (0, False) at the P node. As a result, update_state_after_alloc will pass an empty_blocks, preventing the P node from obtaining the requested local block IDs and causing the push KV cache operation to fail.**

```
--kv-transfer-config '{"kv_connector":"MultiConnector","kv_role":"kv_both","kv_connector_extra_config":{"connectors":[{"kv_connector":"P2pNcclConnector","kv_role":"kv_producer","kv_rank":1,"kv_parallel_size":2,"kv_buffer_size":"1e10","kv_port":"14580"},{"kv_connector":"SharedStorageConnector","kv_role":"kv_both","kv_connector_extra_config":{"shared_storage_path":"local_storage"}}]}}'
```

liziyu179:
## Purpose
a bugfix for multi connector use push kv connector，pass real blocks to kv connector.

Fixes https://github.com/vllm-project/vllm/issues/31145. 
## Test Result

---
<details>

</details>

### DIFF
--- a/vllm/distributed/kv_transfer/kv_connector/v1/multi_connector.py
+++ b/vllm/distributed/kv_transfer/kv_connector/v1/multi_connector.py
@@ -295,7 +295,11 @@ class MultiConnector(KVConnectorBase_V1):
         chosen_connector = self._requests_to_connector.get(request.request_id, -1)
         empty_blocks = blocks.new_empty()
         for i, c in enumerate(self._connectors):
-            if i == chosen_connector:
+            send_type = c._kv_transfer_config.get_from_extra_config("send_type", "GET")
+            if i == chosen_connector or (
+                send_type in ["PUT_ASYNC", "PUT"]
+                and c._kv_transfer_config.is_kv_producer
+            ):
                 # Forward call to the chosen connector (if any).
                 c.update_state_after_alloc(request, blocks, num_external_tokens)
             else:


NickLucche (commented):
Thanks for spotting this @liziyu179 !

Personally, I think this level of detail should be abstracted away from the multiconnector, but more generally I am not sure that prioritizing push connectors (with connector-specific format) is the way to go.
I also understand that given the inherent flexibility of the interface, cases like this where some connectors are not "compatible" may arise again.
Perhaps we should consider again some "priority" level of make the inherent one more clear?

@njhill do you have a better idea on how the MultiConnector behavior should be modified to accomodate this case?

cc @sdavidbd who may also be interested in the discussion

njhill:
I agree with what @sdavidbd wrote, the intention is for `get_num_new_matched_tokens` and `update_state_after_alloc` to be used on the loading/receiving side only, and to use other hooks for the sending/saving.

@liziyu179 it would help if you could provide an example of a "push" connector implementation of the kind you're talking about here.

> Related: it seems the NixlConnector's CPU save path will also run into the same issue.

@sdavidbd I don't quite understand what you mean by this?

PR #31146, opened 2025-12-22
Add util function for checking nesting of rope parameters

hmellor:
The name of this global var has changed in Transformers v5.

The same logic was also duplicated in a few places.

This PR handles the global var name for v4/v5 and centralises the logic for checking if RoPE parameters are nested.


DarkLight1337 (approved):

yewentao256 (approved):
LGTM, thanks for the work!

PR #31147, opened 2025-12-22
Add chat prefix completion feature to DeepSeek v3.2

PHOEBEMOON0802:
Add prefix continuation feature to DeepSeek v3.2


chaunceyjiang (commented):
Can you provide a usage example?

chaunceyjiang:
@PHOEBEMOON0802 Using this prefix parameter will not render the default template.
What is your actual scenario?

chaunceyjiang:
It seems that the official DeepSeek online API also does not provide this parameter.

So my question is: what are the real-world use cases for this `prefix` parameter?

chaunceyjiang (approved):
Thanks~

PR #31149, opened 2025-12-22
[Bugfix][ROCm][Dynamo][DS 3.1][FP8] fix unsupported hasattr call when Dynamo tracing for ROCm device

zejunchen-zejun:
Running DS FP8 with current vLLM on MI355 will fail with a dynamo error because of calling hasattr on a user defined variable in the aiter op aiter.ops.triton.gemm_a8w8_blockscale. Here is the associated log:
image

This PR registered the aiter.ops.triton.gemm_a8w8_blockscale as a custom op to torch.compile to avoid the aiter op being traced internally. With this PR, the DS FP8 can run successfully and the acc is good.
image

### DIFF
--- a/vllm/_aiter_ops.py
+++ b/vllm/_aiter_ops.py
@@ -379,6 +379,31 @@ def _rocm_aiter_gemm_a8w8_fake(
     return Y
 
 
+def _rocm_aiter_triton_gemm_a8w8_blockscale_impl(
+    A: torch.Tensor,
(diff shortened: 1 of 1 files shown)


tjtanaa (approved):
LGTM

zejunchen-zejun:
Hi, @tjtanaa  @gshtras 
Could you help review and merge this PR?
Thank you!

PR #31172, opened 2025-12-22
[NVIDIA] Add env var to override FlashInfer FP8 BMM

Kh4L:
## Purpose

Add `VLLM_FLASHINFER_FP8_BACKEND` environment variable to override the backend used for FlashInfer FP8 BMM operations
Supported options: `"auto", "cutlass", "cudnn", "trtllm"`
When not set, defaults to "auto" which lets FlashInfer choose the backend automatically.

### DIFF
--- a/vllm/envs.py
+++ b/vllm/envs.py
@@ -213,6 +213,7 @@ if TYPE_CHECKING:
     VLLM_USE_TRTLLM_ATTENTION: str | None = None
     VLLM_NVFP4_GEMM_BACKEND: str | None = None
     VLLM_FLASHINFER_DISABLE_Q_QUANTIZATION: bool = False
+    VLLM_FLASHINFER_FP8_BACKEND: str | None = None
     VLLM_HAS_FLASHINFER_CUBIN: bool = False
     VLLM_USE_FLASHINFER_MOE_MXFP4_MXFP8: bool = False
     VLLM_USE_FLASHINFER_MOE_MXFP4_BF16: bool = False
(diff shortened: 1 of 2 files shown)


vllm/envs.py
```diff
@@ -1403,6 +1404,12 @@ def get_vllm_port() -> int | None:
     "VLLM_FLASHINFER_DISABLE_Q_QUANTIZATION": lambda: bool(
         int(os.getenv("VLLM_FLASHINFER_DISABLE_Q_QUANTIZATION", "0"))
     ),
+    # Override the backend used for FlashInfer FP8 BMM operations.
+    # Supported options: "auto", "cutlass", "cudnn", "trtllm"
+    # If not set, defaults to "auto" which lets FlashInfer choose.
+    "VLLM_FLASHINFER_FP8_BACKEND": lambda: os.getenv(
+        "VLLM_FLASHINFER_FP8_BACKEND", None
+    ),
```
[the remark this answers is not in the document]
Kh4L: Addressed

vllm/envs.py
pavanimajety, line 1425: Could we maintain consistency with fp4 backend here?

vadiklyutiy (commented):
We have a lot of FI "backends". Attn, gemm, MLA. 
I propose to be more specific in naming of env var

vllm/envs.py
Kh4L: Are you talking about the naming?

As @vadiklyutiy  suggests, we should go for `VLLM_FLASHINFER_BMM_FP8_BACKEND`

PR #31175, opened 2025-12-22
[Bugfix] Properly apply v_scale for mimo_v2_flash

mgoin:
## Purpose

Noticed this when comparing with the Transformers implementation

Before:
```
vllm serve XiaomiMiMo/MiMo-V2-Flash --tensor-parallel-size 4 --trust-remote-code --tool-call-parser qwen3_xml --reasoning-parser qwen3 --generation-config vllm
python tests/evals/gsm8k/gsm8k_eval.py 
Results:
Accuracy: 0.756
Invalid responses: 0.000
Total latency: 105.251 s
Questions per second: 12.532
Total output tokens: 173189
Output tokens per second: 1645.484
```

After:
```
vllm serve XiaomiMiMo/MiMo-V2-Flash --tensor-parallel-size 4 --trust-remote-code --tool-call-parser qwen3_xml --reasoning-parser qwen3 --generation-config vllm
python tests/evals/gsm8k/gsm8k_eval.py 
Results:
Accuracy: 0.780
Invalid responses: 0.000
Total latency: 108.416 s
Questions per second: 12.166
Total output tokens: 171632
Output tokens per second: 1583.091
```

## Test Result

---
<details>

</details>


vllm/model_executor/models/mimo_v2_flash.py
```diff
@@ -332,13 +338,16 @@ def __init__(self, vllm_config: VllmConfig, prefix: str = "") -> None:
         rope_theta = getattr(config, "rope_theta", 1000000)
         max_position_embeddings = getattr(config, "max_position_embeddings", 32768)
 
+        v_scale = getattr(config, "attention_value_scale", None)
```
robertgshaw2-redhat, line 341: is it ever possible for this to be none?

- If so, I would suggest we raise an error here and avoid the optional typing throughout the model defintion
- If not, then can you add a comment about in which cases this is none

mgoin: I was just following the sglang implementation here https://github.com/sgl-project/sglang/blob/7759716786c481de1cc590f35c250acc488294e6/python/sglang/srt/models/mimo_v2_flash.py#L472
Unfortunately there is no mention of it in the transformer model's config class or modeling code, except for it added to the config.json https://huggingface.co/XiaomiMiMo/MiMo-V2-Flash/blob/main/config.json#L72
I don't have much insight beyond that.

robertgshaw2-redhat: got it. it seems to me that this parameter is part of the model architecture and therefore we should assume it is there. I would suggest validating v_scale is not None and raising an error if it is none asking for a GH report

this will make the typing for the rest of the model simpler

vllm/model_executor/models/mimo_v2_flash.py
```diff
@@ -304,6 +306,9 @@ def forward(
         q, k, v = qkv.split([self.q_size, self.k_size, self.v_size], dim=-1)
         q, k = self.rotary_emb(positions, q, k)
 
+        # Apply v_scale before attention
+        v = v * self.v_scale
```
Abatom:
```suggestion
        if self.v_scale is not None:
            v = v * self.v_scale
```
I think when `self.v_scale` equals 1.0, there's no need to perform the multiplication anymore—it can reduce computation.

vllm/model_executor/models/mimo_v2_flash.py
```diff
@@ -211,6 +211,7 @@ def __init__(
         num_kv_heads: int,
         head_dim: int,
         v_head_dim: int | None = None,
+        v_scale: float = 1.0,
```
Abatom:
```suggestion
        v_scale: float | None = None,
```

vllm/model_executor/models/mimo_v2_flash.py
mgoin: Agreed, let's revert back to None

jeejeelee (approved):

Abatom (commented):
LGTM

PR #31189, opened 2025-12-22
[CI/Build] Fix rlhf_colocate.py for AMD

rjrock:
## Purpose
To get rlhf_colocate.py to pass on AMD's CI.

## Test Plan
`RAY_EXPERIMENTAL_NOSET_HIP_VISIBLE_DEVICES="" VLLM_ALLOW_INSECURE_SERIALIZATION=1 RAY_DEDUP_LOGS=0 python3 rlhf_colocate.py`

## Test Result
Before: An import error on `from vllm.platforms import current_platform` due to ray spawning an Actor with no GPUs allocated. 

After: Exit code of 0.

---
<details>

</details>


vllm/platforms/rocm.py
```diff
@@ -102,8 +102,11 @@ def on_mi3xx() -> bool:
 
 
 @cache
+@with_amdsmi_context
 def on_gfx9() -> bool:
-    GPU_ARCH = torch.cuda.get_device_properties("cuda").gcnArchName
+    h = amdsmi_get_processor_handles()[0]
```
tjtanaa, line 113: @rjrock can you explain the background more about this changes? We have always been using `torch.cuda.get_device_properties("cuda").gcnArchName` without issue. Is there any edge case that `torch.cuda.get_device_properties("cuda").gcnArchName` is failing that we need to change to amdsmi? This is rather important as we would like to try to ensure we can reuse as many `torch` API as possible in coding without relying on this `amd-smi` toolkit for the ease of code development.

rjrock: Right, this change is needed due to using ray as the distributed executor backend, https://github.com/vllm-project/vllm/blob/8471b27df97c3eb79f891802fc0e858f8f7ac6a0/examples/offline_inference/rlhf_colocate.py#L200-L215. Spawning a ray Actor with `num_gpus=0` sets `CUDA_VISIBLE_DEVICES=""`, and vLLM toplevel imports result in resolving the platform as rocm, which results in a call to `on_gfx9`, https://github.com/vllm-project/vllm/blob/8471b27df97c3eb79f891802fc0e858f8f7ac6a0/vllm/platforms/rocm.py#L187-L188. If `CUDA_VISIBLE_DEVICES` is the empty string, then calls into `torch.cuda` fail.

I tried to resolve `supported_quantization` lazily, but ran into mypy issues.

Using `amd-smi` doesn't result in the same `torch.cuda` error case.

tjtanaa: ~can you file an issue to pytorch? Then add the issue link to where you add the amdsmi changes. And add a TODO to revert back to `torch.cuda.get_device_properties("cuda").gcnArchName` when the issue is fixed?~

On second note, we don't need to do this. This `gcnArchName` is exclusively only needed on ROCm. 

@rjrock  Let's return `false` if we know that `torch.cuda` is failing in `on_gfx9()`. Because semantically, there is no GPU given to the process. The process should not be able to know if it is on a GPU or CPU only machine. The current fix has a different meaning, it allows the 0-GPU process to find out details from the current GPU machine that it is living on.

Users can deploy the ray 
```
ray.remote( 
     num_cpus=0, 
     num_gpus=0, 
     scheduling_strategy=PlacementGroupSchedulingStrategy( 
         placement_group=pg, 
         placement_group_capture_child_tasks=True, 
     ), 
 )(MyLLM)
 ```
on a CPU machine and  
  
  
 ```
 .remote( 
     model="facebook/opt-125m", 
     enforce_eager=True, 
     worker_extension_cls="rlhf_utils.ColocateWorkerExtension", 
     tensor_parallel_size=2, 
     distributed_executor_backend="ray", 
     gpu_memory_utilization=0.4, 
     bundle_indices=bundle_indices, 
 ) 
 ```
 
 on a GPU machine.

tjtanaa: Can you also fix other `on_gfx` function in the `rocm.py` as well ?

rjrock: Sure thing

rjrock:
Switching away from using torch.cuda calls at toplevel contexts also enables using ray as data_parallel_backend for ROCm. That would fix this test I marked to skip, https://github.com/vllm-project/vllm/pull/29556

rjrock:
> From trying this approach in the past I remember that amdsmi is not torch.compile friendly. So currently there is no way around torch.cuda for getting gfx

In that case can we have something like `on_gfx9_torch` to be used in kernels and `on_gfx9_amdsmi` to be used host-side?

gshtras (changes requested):
Caching behavior is what you want for everything outside of testing, isn't it? Why work around it globally?

PR #31193, opened 2025-12-23
[Feature] Add iteration level logging and enhance nvtx marker

maxyanghu:
## Purpose
This PR adds iteration level logging for each scheduled iteration. It computes iteration details like number of context/generation requests and number of context/generation tokens. The definition of a context request is any requests that is still being processed in prefill phase and no output token has been generated for that request. It also logs elapsed time of iterations and keeps an index of the current iteration. As the index of iteration is recorded per `EngineCore`, it is kept per data parallel instance.

Here is a logging example:
```
(EngineCore_DP0 pid=2010410) INFO 12-22 21:44:40 [core.py:349] Iteration(515): 0 context requests, 0 context tokens, 20 generation requests, 20 generation tokens, iteration elapsed time: 10.99 ms
(EngineCore_DP0 pid=2010410) INFO 12-22 21:44:40 [core.py:349] Iteration(516): 0 context requests, 0 context tokens, 20 generation requests, 20 generation tokens, iteration elapsed time: 10.95 ms
(EngineCore_DP0 pid=2010410) INFO 12-22 21:44:40 [core.py:349] Iteration(517): 0 context requests, 0 context tokens, 20 generation requests, 20 generation tokens, iteration elapsed time: 10.79 ms
```
And this is a screenshot of the enriched nvtx iteration marker:
image

This feature turned off by default. You'll see a significant increase in server side logging if it's turned on. A flag `--enable-logging-iteration-details` is added to turn it on/off.

## Test Plan
```
vllm serve meta-llama/Llama-3.1-8B-Instruct --async-scheduling --enable-logging-iteration-details
vllm serve meta-llama/Llama-3.1-8B-Instruct --enable-logging-iteration-details
```
and then:
```
vllm bench serve --backend vllm  --model meta-llama/Llama-3.1-8B-Instruct  --dataset-name random  --num-prompts 20
```

## Test Result
correct logging

---
<details>

</details>


vllm/v1/core/sched/output.py
```diff
@@ -129,6 +129,22 @@ class CachedRequestData:
...
+
+        This cached property is safe because CachedRequestData instances
+        are created fresh each scheduling iteration and not mutated during
+        compuation of iteration details.
+        """
+        return dict(zip(self.req_ids, self.num_output_tokens))
+
+    def is_context_phase(self, req_id: str) -> bool:
+        num_output_tokens = self._req_id_to_num_output_tokens.get(req_id)
+        if num_output_tokens is not None:
```
nvpohanh: Line 144-146 can be simplified as: `return num_output_tokens is not None and num_output_tokens  == 0`

vllm/v1/engine/core.py
```diff
@@ -335,6 +336,28 @@ def callback(f, sched_output=scheduler_output):
...
+    @contextmanager
+    def log_iteration_details(self, scheduler_output: SchedulerOutput):
+        if not VLLM_LOG_ITERATION_DETAILS:
+            yield
+            return
+        iteration_details = compute_iteration_details(scheduler_output)
+        before = time.monotonic()
+        yield
+        logger.info(
+            "Iteration details: "
```
nvpohanh: does it make sense to also print out the iteration idx? we can have a `self.iteration_idx` stored in EngineCore, initilalized to 0 in `__init__()` and increment whenever this function is called.

nvpohanh:
> This feature turned on by default so you'll see a significant increase in server side logging.

I am a little worried about the CPU overhead this causes and whether vLLM users would think the logging is too verbose. Should we turn it off by default and add a scheduler flag to enable it?

vllm/v1/utils.py
```diff
@@ -412,3 +414,52 @@ def tensor_data(tensor: torch.Tensor) -> memoryview:
...
+    num_generation_tokens = 0
+    for req_id, num_tokens in scheduler_output.num_scheduled_tokens.items():
+        if scheduler_output.scheduled_cached_reqs.is_context_phase(req_id) or any(
+            new_req.req_id == req_id for new_req in scheduler_output.scheduled_new_reqs
+        ):
+            num_context_requests += 1
+            num_context_tokens += num_tokens
+        else:
+            num_generation_requests += 1
+            num_generation_tokens += num_tokens
```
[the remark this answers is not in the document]
maxyanghu: fixed

vllm/v1/worker/gpu_worker.py
```diff
@@ -546,11 +546,12 @@ def annotate_profile(self, scheduler_output):
...
 
-        num_new = len(scheduler_output.scheduled_new_reqs)
-        num_cached = len(scheduler_output.scheduled_cached_reqs.req_ids)
+        iteration_details = compute_iteration_details(scheduler_output)
 
         return self.profiler.annotate_context_manager(
-            f"execute_new_{num_new}_cached_{num_cached}"
+            f"execute_"
+            f"context_{iteration_details.num_ctx_requests}({iteration_details.num_ctx_tokens})_"
+            f"generation_{iteration_details.num_generation_requests}({iteration_details.num_generation_tokens})"
```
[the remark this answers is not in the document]
maxyanghu: fixed

vllm/v1/core/sched/output.py
maxyanghu: fixed

vllm/v1/worker/gpu_worker.py
```diff
@@ -540,18 +540,23 @@ def get_supported_tasks(self) -> tuple[SupportedTask, ...]:
...
 
         self.profiler.step()
 
-        num_new = len(scheduler_output.scheduled_new_reqs)
-        num_cached = len(scheduler_output.scheduled_cached_reqs.req_ids)
+        iteration_details = compute_iteration_details(scheduler_output)
 
-        return self.profiler.annotate_context_manager(
-            f"execute_new_{num_new}_cached_{num_cached}"
+        annotation = (
```
robertgshaw2-redhat: using fstring in core loop logging can add a lot of overhead. This is something we ran into issues with in the past

vllm/v1/engine/core.py
maxyanghu: added iteration index

vllm/v1/worker/gpu_worker.py
maxyanghu: changed all string building to `join()` to minimize overheads

maxyanghu:
@wangshangsam @nvpohanh @robertgshaw2-redhat It's turned off by default now.

vllm/v1/engine/core.py
```diff
@@ -335,6 +337,35 @@ def callback(f, sched_output=scheduler_output):
...
         return callback
 
+    @contextmanager
+    def log_iteration_details(self, scheduler_output: SchedulerOutput):
+        if not VLLM_LOG_ITERATION_DETAILS:
+            yield
+            return
+        iteration_details = compute_iteration_details(scheduler_output)
+        before = time.monotonic()
+        self.iteration_index += 1
```
nvpohanh: could we start the iteration index at 0?

vllm/envs.py
```diff
@@ -46,6 +46,7 @@
     VLLM_LOGGING_COLOR: str = "auto"
     NO_COLOR: bool = False
     VLLM_LOG_STATS_INTERVAL: float = 10.0
+    VLLM_LOG_ITERATION_DETAILS: bool = False
```
nvpohanh: @robertgshaw2-redhat do you think this should be controlled via env var or a server flag instead?

vllm/v1/engine/core.py
maxyanghu: right sorry, fixed.

vllm/v1/core/sched/output.py
```diff
@@ -129,6 +129,20 @@ class CachedRequestData:
...
+
+        This cached property is safe because CachedRequestData instances
+        are created fresh each scheduling iteration and not mutated during
+        computation of iteration details.
+        """
+        return dict(zip(self.req_ids, self.num_output_tokens))
+
+    def is_context_phase(self, req_id: str) -> bool:
+        num_output_tokens = self._req_id_to_num_output_tokens.get(req_id)
+        return num_output_tokens is not None and num_output_tokens == 0
```
pavanimajety, line 167: I am slightly confused why `num_output_tokens == 0`, is this for a single incomplete chunk? I was always of the opinion that we sample one token during prefill.

maxyanghu: This field is 0 for incomplete chunk. 

And it's also 0 for prefills that are expected to be finished in this iteration. Since this log is computed before we get the output tokens, `num_output_tokens` field of a `Request` is not updated yet and it's still 0 here.

pavanimajety: Thanks!

nvpohanh (approved):

vllm/envs.py
maxyanghu: @robertgshaw2-redhat Could you help take another look? Thanks!

robertgshaw2-redhat: can we have it as a flag? we should put it in the profiler config that @benchislett made I think

maxyanghu: How about putting it inside `ObservabilityConfig`. This is a logging option. But `ProfilerConfig` is for profilers.
I changed the code this way, please take another look. Thanks! @robertgshaw2-redhat

vllm/v1/worker/gpu_worker.py
```diff
@@ -545,18 +545,29 @@ def get_supported_tasks(self) -> tuple[SupportedTask, ...]:
 
     def annotate_profile(self, scheduler_output):
         # add trace annotation so that we can easily distinguish
-        # new/cached request numbers in each iteration
+        # context/generation request numbers in each iteration.
```
robertgshaw2-redhat, line 548: so will this now show up in the PT profiler?

vllm/v1/engine/core.py
```diff
@@ -337,6 +339,35 @@ def log_error_detail(self, scheduler_output: SchedulerOutput):
...
+                    str(iteration_details.num_generation_requests),
+                    " generation requests, ",
+                    str(iteration_details.num_generation_tokens),
+                    " generation tokens, iteration elapsed time: ",
+                    format((time.monotonic() - before) * 1000, ".2f"),
+                    " ms",
+                ]
+            )
+        )
+        self.iteration_index += 1
```
robertgshaw2-redhat: should we put some cap on this?

I know python can have arbitrary sized ints but I wonder if there is some footgun for billions of iterations

vllm/v1/worker/gpu_worker.py
maxyanghu: yes it will

vllm/v1/engine/core.py
maxyanghu: So even at 1,000 iterations/second (which is very fast), reaching 1 billion would take ~11.5 days. And I think putting a  cap or wrap it around would be confusing.

Plus this isn't turned on by default. It's a debugging functionality.

robertgshaw2-redhat: yeah, sounds reasonable to me

robertgshaw2-redhat: One small nit. Can you create the `self.iteration_index` inside this contextmanager?

I want to avoid people using it except for this debug functionality

robertgshaw2-redhat:
```python
self.iteration_index = getattr(self, "iteration_index", 0)
iteration_details = compute_iteration_details(scheduler_output)
```

robertgshaw2-redhat (approved):

PR #31171, opened 2025-12-22
[perf] Integrate flashinfer concat_mla_k

jiahanc:
## Purpose
Integrate Flashinfer concat_mla_k kernel

## Test Result
```
local-completions (base_url=http://0.0.0.0:8087/v1/completions,pretrained=/ds-models/DeepSeek-R1-0528-FP4-v2,model=/ds-models/DeepSeek-R1-0528-FP4-v2,add_bos_token=True,tokenized_requests=False,tokenizer_backend=None,num_concurrent=1024,timeout=60000,max_retries=5,trust_remote_code=True), gen_kwargs: (None), limit: None, num_fewshot: None, batch_size: auto
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9500|±  |0.0060|
|     |       |strict-match    |     5|exact_match|↑  |0.9469|±  |0.0062|
```
---
<details>
<summary> Essential Elements of an Effective PR Description Checklist </summary>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results
- [ ] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>

PR #30746, opened 2025-12-16
[SM100] Enable fp8 compute for prefill MLA

pavanimajety:
## Purpose

Adds a hook to run FP8 Prefill for Flashinfer Cutlass FMHA and TRTLLM Ragged Prefill Kernels for the Latent Attention models that use variable sequence length Prefill Kernels for SM100+.
This is triggered when `--kv-cache-dtype fp8` is enabled, similar to the decode path. 

Depends on https://github.com/flashinfer-ai/flashinfer/pull/2047/

This is step 0. To avoid the additional fp8 quant ops for q,k,v tensors, I plan on adding a rope + quant compilation pass that calls a custom kernel to fuse the operations.

## Test Plan
Accuracy evaluations with Math500 and MMLU Pro datasets.

## Test Result

GSM8k - Flashinfer Cutlass FMHA

```
export VLLM_USE_NCCL_SYMM_MEM=1
export NCCL_NVLS_ENABLE=1
export NCCL_CUMEM_ENABLE=1
export VLLM_ATTENTION_BACKEND=FLASHINFER_MLA
export VLLM_FLASHINFER_MOE_BACKEND=latency
export VLLM_USE_FLASHINFER_MOE_FP4=1

vllm serve /tmp/nvidia-DeepSeek-R1-FP4-v2  --kv-cache-dtype fp8 --tensor-parallel-size 1 --pipeline-parallel-size 1 --data-parallel-size 4 --enable-expert-parallel \
          --swap-space 16 --max-num-seqs 1024 --trust-remote-code --max-model-len 2176 --gpu-memory-utilization 0.9 --max-num-batched-tokens 8192 --no-enable-prefix-caching \
          --async-scheduling --compilation_config.pass_config.enable_fi_allreduce_fusion true --compilation_config.pass_config.enable_attn_fusion true \
          --compilation_config.max_cudagraph_capture_size 2048
```
Evals -
```
root@gb-nvl-081-compute05:/workspace/pm-vllm# python3 tests/evals/gsm8k/gsm8k_eval.py 
Running GSM8K evaluation: 1319 questions, 5-shot
Evaluating: 100%|██████████████████████████████████████████████████████| 1319/1319 [02:20<00:00,  9.37it/s]

Results:
Accuracy: 0.951
Invalid responses: 0.001
Total latency: 140.741 s
Questions per second: 9.372
Total output tokens: 125994
Output tokens per second: 895.219
```

[...]

PR #31195, opened 2025-12-23
[SM100] Resubmit FMHA FP8 prefill for MLA

pavanimajety:
## Purpose
Resubmit FP8 FMHA path for MLA Prefill. Clean up how kernel backends opt into FP8 Prefill. 
Use `--attention-config.use_prefill_query_quantization=true` to enable FP8 Prefill. This is currently guarded because it shows slightly lower perf than BF16 Prefill due to extra casts and quantizations although the kernel level performance is about ~1.5x better.

## Test Plan
CI tests and running DeepSeekR1-FP4 manually. 

## Test Result
Comparison of vLLM serving performance across different attention backend configurations.

---

## Performance Comparison Table

| Run | Date | Configuration | Precision | Accuracy | Duration (s) | Mean TTFT (ms) | Median TTFT (ms) | Req/s | Output tok/s | Peak tok/s | Total tok/s | Mean TPOT (ms) | Median TPOT (ms) | Mean ITL (ms) | Median ITL (ms) |
|-----|------|---------------|-----------|----------|--------------|----------------|------------------|-------|--------------|------------|-------------|----------------|------------------|---------------|-----------------|
| 1 | 2025-12-30 | Flashinfer Cutlass BatchPrefillWithRaggedKVCacheWrapper | FP8 | 0.959 | 43.70 | 5065.54 | 4911.86 | 11.72 | 11998.40 | 15351.00 | 23985.08 | 37.51 | 37.65 | 37.51 | 34.30 |
| 2 | 2025-12-30 | TRTLLM FP8 FMHA Ragged Prefill | FP8 | 0.955 | 43.07 | 5034.62 | 4873.71 | 11.89 | 12172.58 | 15421.00 | 24333.28 | 36.94 | 37.10 | 36.94 | 33.77 |
| 3 | 2025-12-30 | TRTLLM BF16 FMHA Ragged Prefill | BF16 | 0.951 | 41.60 | 4520.67 | 4730.66 | 12.31 | 12601.79 | 15872.00 | 25191.28 | 36.01 | 35.82 | 36.06 | 33.13 |
| 4 | 2025-12-30 | Flashinfer Cutlass BatchPrefillWithRaggedKVCacheWrapper | BF16 | 0.952 | 42.62 | 4700.33 | 4546.29 | 12.01 | 12301.39 | 15360.00 | 24590.77 | 36.82 | 36.97 | 36.82 | 33.81 |

---

### Client Configuration (Common for All Runs)
```bash
INPUT_LEN=1024
OUTPUT_LEN=1024
NUM_PROMPTS=512
CONC=2048
PORT=8000
HOST=0.0.0.0
MODEL=/tmp/nvidia-DeepSeek-R1-FP4-v2

vllm bench serve \
  --backend vllm \
  --host ${HOST} \
  --port ${PORT} \
  --model ${MODEL} \
  --num-prompts ${NUM_PROMPTS} \
  --dataset-name random \
  --random-input ${INPUT_LEN} \
  --random-output ${OUTPUT_LEN} \
  --max-concurrency ${CONC}
```

### Server Base Command (Common for All Runs)
```bash
export VLLM_USE_NCCL_SYMM_MEM=1
export NCCL_NVLS_ENABLE=1
export NCCL_CUMEM_ENABLE=1

vllm serve /tmp/nvidia-DeepSeek-R1-FP4-v2 --tensor-parallel-size 1 --pipeline-parallel-size 1 --data-parallel-size 4 --enable-expert-parallel \
          --swap-space 16 --max-num-seqs 1024 --trust-remote-code --max-model-len 2176 --gpu-memory-utilization 0.9 --max-num-batched-tokens 8192 \
          --no-enable-prefix-caching \
          --async-scheduling --compilation_config.pass_config.fuse_attn_quant true --compilation_config.pass_config.fuse_allreduce_rms true \
          --compilation_config.max_cudagraph_capture_size 2048 --attention-config.backend=FLASHINFER_MLA --compilation_config.custom_ops+=+rotary_embedding \
          [VARIABLE FLAGS BELOW]
```

### Run 1: Flashinfer Cutlass BatchPrefillWithRaggedKVCacheWrapper
**Flags:**
- `--attention-config.use_prefill_query_quantization=true`

### Run 2: TRTLLM FP8 FMHA Ragged Prefill
**Flags:**
- `--attention-config.use_prefill_query_quantization=true`
- `--attention-config.use_trtllm_ragged_prefill=true`

### Run 3: TRTLLM BF16 FMHA Ragged Prefill
**Flags:**
- `--attention-config.use_trtllm_ragged_prefill=true`

### Run 4: Flashinfer Cutlass BatchPrefillWithRaggedKVCacheWrapper
**Flags:**
- (No additional flags - BF16 default, no quantization, no TRTLLM ragged prefill)

---
<details>

- [n/a] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.
- [n/a] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>


vllm/v1/attention/backends/mla/common.py
```diff
@@ -1627,6 +1636,7 @@ def _run_prefill_context_chunk_trtllm_ragged(
             device=q.device,
             dtype=q.dtype,
         )
+
```
robertgshaw2-redhat: nit

vllm/v1/attention/backends/mla/common.py
```diff
@@ -1570,6 +1570,14 @@ def _run_prefill_new_tokens_trtllm_ragged(
 
         assert prefill.query_seq_lens is not None
         assert prefill.workspace_buffer is not None
+        # allocate BF16 / FP16 output tensor for TRT-LLM ragged attention
+        out = torch.zeros(
```
robertgshaw2-redhat: allocating zeros on the hotpath might not be a good idea. Can this be empty? We need to be careful about empty since it can lead to IMAs later

pavanimajety: zeros is a necessity for the trtllm FMHA kernels - we do the same for chunked prefill path in the next method

robertgshaw2-redhat: got it

pavanimajety: Acutally it is not necessary for the causal path. Will follow your suggestion- thanks!

robertgshaw2-redhat: do you think we should pass in the buffer from the function caller? it looks like there can be buffers provided sometimes

pavanimajety: size of q is determined during the `generate` call based on the length of prefill, so preallocating and preintializing a buffer of zeros that are later passed into this kernel might need some more work. I'll take that feedback into consideration and revert #30746 instead for the CI fix

pavanimajety: Passing in as a buffer still requires `preallocated_buffer.fill_(0.0)`, so I am changing the new tokens initialization to torch.empty and keeping it unchanged for chunked-prefill.

vllm/model_executor/layers/attention/mla_attention.py
```diff
@@ -2160,11 +2253,39 @@ def _context_parallel_compute_prefill_context(
                 toks=toks,
             )
 
+            # Allocate k and v together in layout: [k_nope | k_pe | v]
```
MatthewBonanni: This is also done below in `_forward_prefill` - can we factor this out into a helper?

vllm/model_executor/layers/attention/mla_attention.py
```diff
@@ -2038,26 +2109,48 @@ def _compute_prefill_context(
...
+            else:
+                # FP8 path: gather cache without dequantization
+                ops.cp_gather_cache(
+                    src_cache=kv_c_and_k_pe_cache,
+                    dst=workspace,
+                    block_table=prefill_metadata.block_table,
+                    cu_seq_lens=prefill_metadata.chunked_context.cu_seq_lens[i],
+                    batch_size=attn_metadata.num_prefills,
+                    seq_starts=prefill_metadata.chunked_context.starts[i],
+                )
```
MatthewBonanni, line 2352: Definitely out of scope for this PR, but we should really unify these into a single kernel that just decides whether or not to dequant based on the destination dtype. Also, I think it would make sense to bundle `cp_gather_and_upconvert_fp8_kv_cache` in the same python wrapper too, even if it remains a separate kernel on the backend

vllm/model_executor/layers/attention/mla_attention.py
```diff
@@ -2038,26 +2109,48 @@ def _compute_prefill_context(
         output = None
         iters = len(prefill_metadata.chunked_context.seq_tot)
         workspace = prefill_metadata.chunked_context.workspace
+
+        if attn_metadata.prefill.q_data_type == current_platform.fp8_dtype():
```
MatthewBonanni: nit: Could we define `use_fp8_prefill` and use that here and in [2118](https://github.com/vllm-project/vllm/pull/31195/changes#diff-e74ff3870f4b147beb297d6176081f70fd2bdf58fd38b182da58307e4da9eb2aR2118)

vllm/model_executor/layers/attention/mla_attention.py
```diff
@@ -1947,6 +2017,7 @@ def _run_prefill_new_tokens_trtllm_ragged(
             enable_pdl=False,
             is_causal=True,
             return_lse=return_softmax_lse,
+            out=out,
```
MatthewBonanni, line 2227: Is `ret` the same buffer as `out`? Is `out` being mutated and returned as `ret`? Seems like the same pattern as below in the context prefill, it's just a bit unclear

MatthewBonanni (commented):
Thanks for the contribution!

vllm/model_executor/layers/attention/mla_attention.py
pavanimajety: Ran some comparisons against #31171, seems like Flashinfer copy kernel is better choice throughout
```
root@gb-nvl-082-compute08:/workspace/pm-vllm# python3 ../cs-scripts/test_kv_build_paths.py
  Tokens |  Combined (ms) |  Concat (ms) | FlashInfer (ms)
      32 |         0.0967 |       0.0504 |          0.0518
      64 |         0.0956 |       0.0500 |          0.0510
     128 |         0.0946 |       0.0509 |          0.0540
     256 |         0.0947 |       0.0547 |          0.0509
     512 |         0.0933 |       0.0759 |          0.0503
    1024 |         0.1204 |       0.1380 |          0.0520
    2048 |         0.2322 |       0.2598 |          0.0833
    4096 |         0.4472 |       0.5046 |          0.1654
    8192 |         0.8645 |       0.9853 |          0.3130
```

Removing this logic in the upcoming change.

vllm/model_executor/layers/attention/mla_attention.py
pavanimajety: Agree with this. Created https://github.com/vllm-project/vllm/issues/33669 to keep track

vllm/model_executor/layers/attention/mla_attention.py
pavanimajety: done

vllm/model_executor/layers/attention/mla_attention.py
pavanimajety: I need to create `out` buffer to specify the `output_dtype` and we currently don't have `lse` as an output for `trtllm_ragged_attention_deepseek`, I believe.  So I need a different return variable that can be of type tuple. Please suggest if you have a better idea!

vllm/model_executor/layers/attention/mla_attention.py
```diff
@@ -1262,6 +1287,40 @@ def determine_chunked_prefill_workspace_size(vllm_config: VllmConfig) -> int:
...
+        model_dtype: torch.dtype,
+    ) -> torch.dtype:
+        """
+        Determine the query data type for prefill queries.
+        Return FP8 dtype if cache is FP8 and prefill query quantization
+        is enabled, else model dtype.
+        """
+        use_fp8 = (
+            vllm_config.cache_config.cache_dtype.startswith("fp8")
+            and vllm_config.attention_config.use_prefill_query_quantization
```
mgoin, line 1302: Is this plan to keep this as MLA only? It seems a bit strange to have this False by default in the config, have that Falseness by enforced by MLA, but not for GQA methods which will quantize prefill if FP8 kv cache is used and supported

mgoin (approved):

vllm/model_executor/layers/attention/mla_attention.py
pavanimajety: yes, the plan is to keep this for MLA only. GQA's prefill is quantized as long as trtllm kernels are available. For MLA, because of projections being enforced in BF16, the casts are incurring extra cost - hence the guard. In my follow up PRs, I can rename this to `use_prefill_query_quant_mla` or remove this entirely by reworking the projection quantization logic.

mgoin (approved):

PR #31196, opened 2025-12-23
[compile] Remove torch 2.9 patches

angelayi:
Since torch 2.9.1 integration has landed, we can remove the patches here!


ProExpertProg (changes requested):
We should do this but not until rocm moves past 2.9 (I think they're currently on 2.9.alpha?) - cc @tjtanaa @gshtras

BoyuanFeng:
since the patch only impacts pt 2.9, should we still keep it? Just in case some users want to use pt 2.9 for some reasons.

PR #31213, opened 2025-12-23
Add a support to disable Cutlass W8A8 kernels

houseroad:
Summary: To work around the issue on GB300 for now.

Test Plan: Run the Qwen3 Coder and check gsm8k results.

Differential Revision: D89704391


houseroad:
@pavanimajety , cutlass w8a8 block fp8 generates some inf numbers, only happens on GB300, not even on GB200.

PR #31225, opened 2025-12-23
fix(core): break circular reference in Request using weakref

kelvinvelasquez-SDE:
Fixes #31200.

This PR addresses a memory leak caused by a circular reference in the `Request` class initialization.
The `Request` object held a reference to `block_hasher` via `functools.partial`, which in turn held a reference back to `self`.
This cycle prevented immediate object reclamation via reference counting, relying instead on the garbage collector which may be delayed under load.

**Changes:**
- Used `weakref.proxy(self)` when binding `self` to `block_hasher` in `partial`.

**Verification:**
- Confirmed that `Request` objects are now reclaimed immediately upon deletion (refcount drops to 0) without requiring a GC collection cycle.

### DIFF
--- a/tools/pre_commit/validate_config.py
+++ b/tools/pre_commit/validate_config.py
@@ -149,11 +149,11 @@ def validate_file(file_path: str):
         print(e)
         raise SystemExit(1) from e
     else:
-        print("✅")
+        print("OK")
 
 
 def fail(message: str, node: ast.stmt):
-    raise ValueError(f"❌ line({node.lineno}): {message}")
+    raise ValueError(f"Error line({node.lineno}): {message}")
 
 
 def main():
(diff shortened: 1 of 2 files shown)


njhill (commented):
Thanks @kelvinvelasquez-SDE, and @frelam for reporting.

I think rather than using weakref it would be better to just avoid using `partial`: change the function type to `Callable[["Request"], list[BlockHash]]`, and adjust where it's called to pass `self` as an arg.

kelvinvelasquez-SDE:
Thanks for the feedback! I've updated the implementation to avoid using `weakref` and `partial`. instead, I'm now passing `self` explicitly to the `block_hasher`. Verified locally that the memory leak is still resolved.

PR #31228, opened 2025-12-23
Cleanup basic and entrypoint test organisation

hmellor:
Supersedes https://github.com/vllm-project/vllm/pull/27747:

- Don't call basic test files individually
- Move the unit tests into a `unit` directory so we don't have to manually ignore everything else
- Move `openai/tool_parsers` into the `unit` tests because that's what they are


chaunceyjiang (approved):
Thanks~

tests/entrypoints/openai/test_chat_utils.py
```diff
@@ -33,8 +33,8 @@
 from vllm.tokenizers.mistral import MistralTokenizer
 from vllm.utils.serial_utils import tensor2base64
 
-from ..models.registry import HF_EXAMPLE_MODELS
-from ..utils import VLLM_PATH
+from ...models.registry import HF_EXAMPLE_MODELS
+from ...utils import VLLM_PATH
```
chaunceyjiang: I suggest using `from tests.utils` here.

hmellor: I've updated all the imports from utils in `tests/entrypoints`

PR #31233, opened 2025-12-23
[Benchmark] Auto-infer dataset name from path for backward compatibility

yurekami:
## Summary
When `--dataset-path` is provided without `--dataset-name`, automatically infer the dataset type from the path to restore backward compatibility with the legacy `benchmark_serving.py` script.

This addresses the root cause of issue #24684 where users migrating from the old script found that `vllm bench serve` wasn't a drop-in replacement because it defaulted to "random" dataset instead of inferring "sharegpt".

## Changes
- Added `_infer_dataset_name_from_path()` function that analyzes the dataset path and returns the likely dataset type
- Modified `_ValidateDatasetArgs` to try inference before showing an error

## Inference Logic
| Path Pattern | Inferred Dataset |
|-------------|------------------|
| Contains "sharegpt" | sharegpt |
| Contains "sonnet" | sonnet |
| Contains "burstgpt" | burstgpt |
| Looks like HuggingFace ID (e.g., `org/dataset`) | hf |
| Ends with `.json` or `.jsonl` | sharegpt (legacy default) |

If inference fails, the existing helpful error message is shown prompting users to specify `--dataset-name` explicitly.

## Example
```bash
# Before: Error - "Cannot use 'random' dataset with --dataset-path"
vllm bench serve --dataset-path /path/to/ShareGPT.json

# After: Works! Prints "Inferred --dataset-name=sharegpt from --dataset-path=..."
vllm bench serve --dataset-path /path/to/ShareGPT.json
```

## Test Plan
- [ ] Test with ShareGPT dataset path → should infer "sharegpt"
- [ ] Test with sonnet dataset path → should infer "sonnet"
- [ ] Test with HuggingFace dataset ID → should infer "hf"
- [ ] Test with unrecognized path → should show error with guidance
- [ ] Explicit `--dataset-name` should still work as before

Fixes #24684

### DIFF
--- a/vllm/benchmarks/datasets.py
+++ b/vllm/benchmarks/datasets.py
@@ -1294,6 +1294,27 @@ class ShareGPTDataset(BenchmarkDataset):
         return samples
 
 
+def _is_local_hf_dataset(path: str) -> bool:
+    """Check if a local path is a downloaded HuggingFace dataset.
+
+    HuggingFace datasets when downloaded locally typically contain:
+    - dataset_info.json file
+    - Arrow files (.arrow) or parquet files
+    """
+    from pathlib import Path
+
+    p = Path(path).expanduser()
+    if not p.is_dir():
+        return False
+
+    # Check for HF dataset markers
+    if (p / "dataset_info.json").exists():
+        return True
+
+    # Check for arrow files (common in HF datasets)
+    return any(p.glob("*.arrow")) or any(p.glob("**/*.arrow"))
+
+
 def _infer_dataset_name_from_path(dataset_path: str) -> str | None:
     """Try to infer dataset type from the path.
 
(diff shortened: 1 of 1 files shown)


vllm/benchmarks/datasets.py
```diff
@@ -1294,6 +1294,51 @@ def sample(
...
+    provided. See GitHub issue #24684.
+
+    Returns:
+        Inferred dataset name, or None if inference failed.
+    """
+    path_lower = dataset_path.lower()
+
+    # Check for sharegpt patterns (most common case)
+    if "sharegpt" in path_lower:
+        return "sharegpt"
```
Isotr0py: I think this will return incorrect dataset name for HF datasets like `Aeala/ShareGPT_Vicuna_unfiltered`? Perhaps we need to make HF dataset highest priority.

vllm/benchmarks/datasets.py
```diff
@@ -1294,6 +1294,53 @@ def sample(
...
+    # This must come first to correctly handle HF datasets with names like
+    # "Aeala/ShareGPT_Vicuna_unfiltered" that contain keyword patterns
+    if (
+        "/" in dataset_path
+        and not dataset_path.startswith(("/", ".", "~"))
+        and not any(
+            path_lower.endswith(ext)
+            for ext in [".json", ".jsonl", ".txt", ".csv", ".parquet"]
+        )
+    ):
```
Isotr0py, line 1346: Hmmm, I think this would be invalid if user provides a local dwonloaded HF dataset... 🤔

We need a more robust condition to check HF dataset.

yurekami:
Thanks for the follow-up @Isotr0py! You're right - I've added a more robust check for locally downloaded HuggingFace datasets.

**New approach:**

Added `_is_local_hf_dataset()` function that checks for HF dataset markers before keyword matching:

1. **dataset_info.json** - Standard HF dataset metadata file
2. **Arrow files (.arrow)** - Common format used by HuggingFace datasets

**Priority order now:**
1. Local HF dataset (has `dataset_info.json` or `.arrow` files) → `hf`
2. Remote HF dataset ID (e.g., `org/dataset-name`) → `hf`
3. Path contains "sharegpt" → `sharegpt`
4. Path contains "sonnet" → `sonnet`
5. Path contains "burstgpt" → `burstgpt`
6. `.json`/`.jsonl` files → `sharegpt` (legacy default)

**Example:**
```
/data/Aeala/ShareGPT_Vicuna_unfiltered/  (with dataset_info.json)
→ Now correctly infers "hf" instead of "sharegpt"
```

PR #31241, opened 2025-12-23
[Bugfix] Fix eagle dp tests on A100

zou3519:
## Purpose
`TP_SIZE=1 DP_SIZE=2 pytest -v -s tests/v1/distributed/test_eagle_dp.py` fails on A100 for me before this PR.

Here's what I think is happening:
- the test is checking that the tokens produced by a model with eagle is identical to a model without eagle
- the model with eagle uses a draft model to produce draft tokens
- the target model takes all of the draft tokens and then does a forward pass to see how many of the tokens to accept/reject. The target model is using a batch_size > 1.
- the model without eagle just generates the tokens one-by-one, that is, it has batch_size = 1.
- For these two models to be *consistent*, we need batch invariance. So I turned on batch invariance (which also required the selection of an attention backend)

## Test Plan
`TP_SIZE=1 DP_SIZE=2 pytest -v -s tests/v1/distributed/test_eagle_dp.py` on A100

## Test Result
test passes

### DIFF
--- a/tests/v1/distributed/test_eagle_dp.py
+++ b/tests/v1/distributed/test_eagle_dp.py
@@ -16,7 +16,12 @@ DP_SIZE = int(os.getenv("DP_SIZE", 2))
 
 
 @pytest.mark.asyncio
-async def test_run_eagle_dp():
+async def test_run_eagle_dp(monkeypatch: pytest.MonkeyPatch):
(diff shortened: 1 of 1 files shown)


atalman (approved):
lgtm

LucasWilkinson (approved):
LGTM; thanks for the investigation

PR #31246, opened 2025-12-23
[Kernel] Add topk_sigmoid kernel

xyang16:
## Purpose

This PR add `topk_sigmoid` kernel. Currently fused topk only support softmax function, this PR add sigmoid function support, also support `e_score_correction_bias`.

* Support `sigmoid` function
* Support optional `bias`. If `bias` is not null, use biased values as weights for selection, but return the raw score as `topk_weights`, following torch reference implementation [here](https://github.com/vllm-project/vllm/blob/v0.13.0/vllm/model_executor/layers/fused_moe/fused_moe.py#L1185-L1193)
* ROCm aiter library supports `topk_sigmoid`, so use it for `fused_topk` op. But aiter `topk_sigmoid` doesn't support e_score_correction_bias, so use torch implementation for `fused_topk_bias` for ROCm.
* MiniMax-M2 model uses topk with [sigmoid](https://huggingface.co/MiniMaxAI/MiniMax-M2/blob/main/config.json#L115) function. Currently only `grouped_topk` kernel supports sigmoid, so it’s using grouped topk with [num_expert_group=1](https://github.com/vllm-project/vllm/blob/v0.13.0/vllm/model_executor/models/minimax_m2.py#L103-L105) to mimic `fused_topk`. Therefore this PR change it to use `topk_sigmoid` for MiniMax-M2 model to improve performance.
* e2e output token throughput improves 11%.
* Added unit test `test_fused_topk.py`
* Added micro benchmark `benchmark_fused_topk.py`

## Test Plan

```
pytest -s -v tests/kernels/moe/test_fused_topk.py
```

## Test Result

Unit tests passed.

## Profiling

Main:

Screenshot 2026-01-19 at 7 53 10 PM

PR:

Screenshot 2026-01-19 at 7 52 50 PM

Main: `grouped_topk_fused_kernel` is launched, totally takes around 11µs.

PR: After this PR, `topkGating` kernel is launched, takes around 6µs.

## Benchmark

Benchmarked MiniMax-M2 model e2e on H200.

```
vllm serve MiniMaxAI/MiniMax-M2 \
    --tensor-parallel-size 8 \
    --max-num-seqs 16 \
    --enable-expert-parallel \
    --trust-remote-code \
    --compilation_config '{"compile_sizes": [1,2,4,8,16]}' \
    --no-enable-prefix-caching
```

```
vllm bench serve \
  --model MiniMaxAI/MiniMax-M2 \
  --dataset-name sharegpt \
  --sharegpt-output-len 300 \
  --max-concurrency 16 \
  --num-prompts 1000 \
  --num-warmups 20 \
  --ignore-eos
```

Main:

```
Maximum request concurrency:             32        
Benchmark duration (s):                  181.18    
Total input tokens:                      225133    
Total generated tokens:                  300000    
Request throughput (req/s):              5.52      
Output token throughput (tok/s):         1655.81   
Peak output token throughput (tok/s):    1792.00   
Peak concurrent requests:                64.00     
Total token throughput (tok/s):          2898.41   
Mean TTFT (ms):                          155.76    
Median TTFT (ms):                        150.69    
P99 TTFT (ms):                           253.78    
Mean TPOT (ms):                          18.51     
Median TPOT (ms):                        18.56     
P99 TPOT (ms):                           19.20     
Mean ITL (ms):                           18.51     
Median ITL (ms):                         18.20     
P99 ITL (ms):                            19.69     
```

PR:
```
Maximum request concurrency:             32        
Benchmark duration (s):                  163.23    
Total input tokens:                      225133    
Total generated tokens:                  300000    
Request throughput (req/s):              6.13      
Output token throughput (tok/s):         1837.95   
Peak output token throughput (tok/s):    2048.00   
Peak concurrent requests:                64.00     
Total token throughput (tok/s):          3217.23   
Mean TTFT (ms):                          209.99    
Median TTFT (ms):                        209.79    
P99 TTFT (ms):                           269.37    
Mean TPOT (ms):                          16.44     
Median TPOT (ms):                        16.20     
P99 TPOT (ms):                           23.83     
Mean ITL (ms):                           16.44     
Median ITL (ms):                         15.98     
P99 ITL (ms):                            16.77     
```

e2e output token throughput improves 11%.

## Accuracy Testing

```
python3 tests/evals/gsm8k/gsm8k_eval.py
```

Main:
```
Results:
Accuracy: 0.943
Invalid responses: 0.000
Total latency: 183.035 s
Questions per second: 7.206
Total output tokens: 128382
Output tokens per second: 701.406
```

PR:
```
Results:
Accuracy: 0.943
Invalid responses: 0.000
Total latency: 159.336 s
Questions per second: 8.278
Total output tokens: 128101
Output tokens per second: 803.968
```

---
<details>

</details>

---

> [!NOTE]
> Enables fused MoE Top‑K routing with both `sigmoid` and `softmax` scoring, plus optional expert `bias` for selection, and integrates it across CUDA/ROCm and Python APIs.
> 
> - Refactors CUDA to unified `topkGating` supporting `sigmoid`/`softmax`, adds `moeSigmoid`, bias-applied selection while returning unbiased `topk_weights`; updates launcher/fallback paths
> - Exposes `topk_softmax(..., bias?)` and new `topk_sigmoid(..., bias?)` via torch bindings; Python adds `ops.topk_sigmoid`, extends `ops.topk_softmax`, and plumbs `scoring_func` through `fused_topk`/`fused_topk_bias`
> - Splits dispatch into `dispatch_topk_softmax_func` and `dispatch_topk_sigmoid_func`; registers/uses ROCm AITER `rocm_aiter_topk_sigmoid` (biased path falls back)
> - Updates routing config to allow sigmoid in non-grouped paths; switches `MiniMax-M2` to use fused `topk` with `sigmoid`
> - Adds unit tests (`tests/kernels/moe/test_fused_topk.py`, dispatch tests) and micro-benchmark (`benchmarks/kernels/benchmark_fused_topk.py`)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dbdefcb2252a488d984c409af784aaa06c1c3252. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


vllm/model_executor/models/minimax_m2.py
```diff
@@ -100,9 +100,6 @@ def __init__(
             num_experts=config.num_local_experts,
             top_k=config.num_experts_per_tok,
             scoring_func=config.scoring_func,
-            use_grouped_topk=True,
-            num_expert_group=1,
-            topk_group=1,
```
jeejeelee, line 105: From my understanding, some models, such as `GLM-4.7/Mino-V2-Flash` also need to be modified in this way. Perhaps a better approach would be to implement automatic dispatch within fuse_moe, rather than modifying the model implementation

csrc/moe/topk_softmax_kernels.cu
```diff
@@ -693,14 +792,55 @@ void topk_softmax(
...
+    bool renormalize,
+    std::optional<torch::Tensor> bias)
+{
+    const int num_experts = gating_output.size(-1);
+    const auto num_tokens = gating_output.numel() / num_experts;
+    const int topk = topk_weights.size(-1);
+
+    const bool is_pow_2 = (num_experts != 0) && ((num_experts & (num_experts - 1)) == 0);
+    const bool needs_workspace = !is_pow_2 || num_experts > 256;
+    const int64_t workspace_size = needs_workspace ? num_tokens * num_experts : 0;
```
yewentao256, line 826: Is this workspace_size int64_t needed?

vllm/model_executor/layers/fused_moe/fused_moe.py
```diff
@@ -1180,7 +1225,52 @@ def fused_topk_bias(
...
+                gating_output,
+                renormalize,
+                e_score_correction_bias,
+            )
+            return topk_weights, topk_ids
+        else:
+            raise ValueError(f"Unsupported scoring function: {scoring_func}")
+
     n_routed_experts = gating_output.shape[-1]
     scores = gating_output.softmax(dim=-1)
```
yewentao256: Here we should consider `scoring_func`, a potential bug

csrc/moe/topk_softmax_kernels.cu
```diff
@@ -636,39 +719,54 @@ void topkGatingSoftmaxKernelLauncher(
...
-    int num_tokens, int num_experts, int topk, bool renormalize, cudaStream_t stream)
+    int num_tokens, int num_experts, int topk, bool renormalize,
+    std::optional<torch::Tensor> bias,
+    int scoring_func, cudaStream_t stream)
 {
+    // Validate bias if provided - must always be float32
+    const float* bias_ptr = nullptr;
+    if (bias.has_value()) {
+      const torch::Tensor& bias_tensor = bias.value();
+      TORCH_CHECK(bias_tensor.dim() == 1, "bias tensor must be 1D");
```
yewentao256, line 737: We could also check cuda tensor and is_contiguous here

vllm/_aiter_ops.py
```diff
@@ -1236,6 +1262,19 @@ def topk_softmax(
...
 
+    @staticmethod
+    def topk_sigmoid(
+        topk_weights: torch.Tensor,
+        topk_indices: torch.Tensor,
+        token_expert_indices: torch.Tensor,
+        gating_output: torch.Tensor,
+        renormalize: bool,
+    ) -> tuple[torch.Tensor, ...]:
+        torch.ops.vllm.rocm_aiter_topk_sigmoid(
```
yewentao256, line 1297: Is `renormalize` needed here for rocm?

vllm/model_executor/layers/fused_moe/fused_moe.py
```diff
@@ -1121,32 +1121,63 @@ def vllm_topk_softmax(
...
+        return rocm_aiter_ops.topk_sigmoid
+    return vllm_topk_sigmoid
+
+
 def fused_topk(
     hidden_states: torch.Tensor,
     gating_output: torch.Tensor,
     topk: int,
     renormalize: bool,
+    scoring_func: str = "softmax",
```
yewentao256: Recommend adding this param after `indices_type`

csrc/moe/torch_bindings.cpp
```diff
@@ -5,9 +5,17 @@ TORCH_LIBRARY_EXPAND(TORCH_EXTENSION_NAME, m) {
...
-      "token_expert_indices, Tensor gating_output, bool renormalize) -> ()");
+      "token_expert_indices, Tensor gating_output, bool renormalize, Tensor? "
+      "bias) -> ()");
   m.impl("topk_softmax", torch::kCUDA, &topk_softmax);
 
+  // Apply topk sigmoid to the gating outputs.
+  m.def(
+      "topk_sigmoid(Tensor! topk_weights, Tensor! topk_indices, Tensor! "
+      "token_expert_indices, Tensor gating_output, bool renormalize, Tensor? "
+      "bias) -> ()");
```
yewentao256, line 16: Should we define `bias` a default None value?

yewentao256 (commented):
Nice work! A few thoughts

csrc/moe/topk_softmax_kernels.cu
xyang16: Thanks for reviewing! Yes I think it's not necessary to be int64_t. But since the `topk_softmax` also uses int64_t https://github.com/vllm-project/vllm/blob/v0.13.0/csrc/moe/topk_softmax_kernels.cu#L688, I just make it the same to be consistent.

csrc/moe/topk_softmax_kernels.cu
xyang16: Added check. Thanks!

csrc/moe/torch_bindings.cpp
xyang16: I make it `Tensor? bias` to specify it's optional tensor, but I can also make it `Tensor? bias=None` to give it a default value. Please let me know how you think. Thanks!

vllm/_aiter_ops.py
xyang16: Somehow aiter `topk_sigmoid` doesn't support `renormalize` https://github.com/ROCm/aiter/blob/v0.1.9/aiter/ops/moe_op.py#L36, although aiter `topk_softmax` support `renormalize` https://github.com/ROCm/aiter/blob/v0.1.9/aiter/ops/moe_op.py#L16

vllm/model_executor/layers/fused_moe/fused_moe.py
xyang16: Changed. Thanks!

vllm/model_executor/layers/fused_moe/fused_moe.py
xyang16: Good catch! Changed, thanks!

vllm/model_executor/models/minimax_m2.py
xyang16: @jeejeelee Thanks for reviewing!

I saw GLM-4.7 defined n_group, topk_group parameters in config https://huggingface.co/zai-org/GLM-4.7/blob/main/config.json#L24-L25. So I guess they want to use grouped_topk?


And Mino-V2-Flash specifies topk_method to be "noaux_tc", which is grouped_topk in config https://huggingface.co/XiaomiMiMo/MiMo-V2-Flash/blob/main/config.json#L205-L207. So I didn't change it for this model.

Please let me know how you think, thanks!

vllm/model_executor/models/minimax_m2.py
```diff
@@ -100,9 +100,6 @@ def __init__(
             num_experts=config.num_local_experts,
             top_k=config.num_experts_per_tok,
             scoring_func=config.scoring_func,
-            use_grouped_topk=True,
```
chaunceyjiang, line 103: Just a question, is this PR suitable for DS? Will it improve DS in any way?

https://huggingface.co/deepseek-ai/DeepSeek-V3.2/blob/main/config.json#L57
https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/deepseek_v2.py#L318-L320

csrc/moe/torch_bindings.cpp
mgoin: I would prefer not to have default values for function bindings here. I think we don't do this generally

csrc/moe/topk_softmax_kernels.cu
```diff
@@ -353,61 +386,87 @@ __launch_bounds__(WARPS_PER_CTA* WARP_SIZE_PARAM) __global__
...
 
 #pragma unroll
-    for (int ii = 0; ii < VPT; ++ii)
-    {
+      for (int ii = 0; ii < VPT; ++ii)
+      {
         row_chunk[ii] = row_chunk[ii] * reciprocal_row_sum;
+      }
+    } else if (SCORING_FUNC == SCORING_SIGMOID) {
+      for (int ii = 0; ii < VPT; ++ii)
```
mgoin, line 439: This is missing `#pragma unroll` to match the other case

csrc/moe/topk_softmax_kernels.cu
```diff
@@ -36,6 +36,9 @@
 #define MAX(a, b) ((a) > (b) ? (a) : (b))
 #define MIN(a, b) ((a) < (b) ? (a) : (b))
 
+static constexpr int SCORING_SOFTMAX = 0;
+static constexpr int SCORING_SIGMOID = 1;
```
mgoin: nit: consider an enum for ScoringFunc

tests/kernels/moe/test_fused_topk.py
```diff
@@ -0,0 +1,136 @@
...
+    gating_output: torch.Tensor,
+    topk: int,
+    renormalize: bool,
+    e_score_correction_bias: torch.Tensor = None,
+    scoring_func: str = "softmax",
+):
+    if scoring_func == "softmax":
+        scores = torch.softmax(gating_output.float(), dim=-1)
+    else:
+        scores = torch.sigmoid(gating_output.float())
```
mgoin, line 29: nit: assert `scoring_func == "sigmoid"`

vllm/model_executor/models/minimax_m2.py
xyang16: DeepSeek uses the noaux_tc / grouped_topk method https://huggingface.co/deepseek-ai/DeepSeek-V3.2/blob/main/config.json#L60. So this PR is not for DS. Thanks!

csrc/moe/topk_softmax_kernels.cu
xyang16: Thanks for catching this! Sorry to miss that.

csrc/moe/topk_softmax_kernels.cu
yewentao256: Make sense, perhaps you can have a following up PR updating this later.

vllm/model_executor/layers/fused_moe/layer.py
```diff
@@ -540,25 +540,6 @@ def __init__(
         self.apply_router_weight_on_input = apply_router_weight_on_input
         self.activation = activation
 
-        self._grouped_topk_impl: GroupedTopk | None = None
```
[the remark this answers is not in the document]
xyang16: Fixed

vllm/model_executor/models/minimax_m2.py
jeejeelee: I mean wether we use `topk_simoid` should be decided within FusedMoE, rather than change the model implementation by hardcoded way. Thus, GLM4.7 and similar models will also benefit from this PR

xyang16: I see. This makes sense. Since it will change a lot of files, I can do this in another PR.

csrc/moe/topk_softmax_kernels.cu
xyang16: @yewentao256 I have updated workspace_size to size_t. Please take a look. Thanks!

yewentao256 (commented):
Generally look good to me, please address the previous comments from @mgoin

csrc/moe/topk_softmax_kernels.cu
xyang16: @mgoin I have added ScoringFunc enum. Please take a look. Thanks!

vllm/model_executor/layers/fused_moe/router/fused_topk_bias_router.py
```diff
@@ -12,13 +14,98 @@
...
+                gating_output,
+                renormalize,
+                e_score_correction_bias,
+            )
+            return topk_weights, topk_ids
+        else:
+            raise ValueError(f"Unsupported scoring function: {scoring_func}")
+
     n_routed_experts = gating_output.shape[-1]
     scores = gating_output.softmax(dim=-1)
```
yewentao256: Should we consider scoring_func here?

yewentao256:
Generally LGTM, but also wants to get a stamp from @mgoin

vllm/model_executor/layers/fused_moe/router/fused_topk_bias_router.py
xyang16: Fixed, thanks a lot!

vllm/model_executor/layers/fused_moe/router/fused_topk_bias_router.py
```diff
@@ -12,15 +14,106 @@
...
+    return topk_weights, topk_indices
+
+
 def fused_topk_bias(
     hidden_states: torch.Tensor,
     gating_output: torch.Tensor,
     e_score_correction_bias: torch.Tensor,
     topk: int,
     renormalize: bool,
+    scoring_func: str = "softmax",
```
mgoin, line 63: I would rather not default to a value for scoring_func, it should always be explicit

csrc/moe/topk_softmax_kernels.cu
```diff
@@ -186,12 +220,12 @@ __launch_bounds__(TPB) __global__ void moeTopK(
             const bool should_process_row = row_is_active && node_uses_expert;
 
             const int idx = k * block_row + k_idx;
-            output[idx] = result_kvp.value;
+            output[idx] = inputs_after_softmax[thread_read_offset + expert];
```
mgoin, line 224: Might be worth a comment that you are using the original unbiased scores for output weights

xyang16: Added comment. Thanks!

vllm/model_executor/layers/fused_moe/router/fused_topk_bias_router.py
xyang16: Changed. Thanks!

tests/kernels/moe/test_fused_topk.py
xyang16: Added assert. Thanks!

mgoin (approved):

PR #31258, opened 2025-12-24
Update unquantized_fused_moe_method.py

aaab8b:
## Purpose
Change assignment of unquantized moe weights when using aiter on rocm, making it safer for reloading the weights. Solve the random output case after wake-up and reloading weights in reinforcement learning.
## Test Plan
Enabling VLLM_ROCM_USE_AITER=1 and VLLM_ROCM_USE_AITER_MOE=1 testing Qwen3-30BA3B (or a moe model) after wake-up and reloading weights in reinforcement learning.
## Test Result
If using previous assignment, the output will be random characters. If using my assignment, the output will be normal.
---
<details>
Hi, I'm a developer from AMD.
I've been doing some adaptation works for using vllm on ROLL (https://github.com/alibaba/ROLL/) with ROCm platforms. It turns out to generate random characters after vllm instance sleeps and wake-up and get model weights from reloading from train actors. Then I found out that AITER will shuffle the original weights to a better layout to be calculated, but it points to a new tensor, which will cause problem after capturing cuda graphs and reloading weights. 
This pull request will fix this issue in general usage of vllm when using AITER on ROCm platforms in reinforcement learning.

</details>

### DIFF
--- a/vllm/model_executor/layers/fused_moe/unquantized_fused_moe_method.py
+++ b/vllm/model_executor/layers/fused_moe/unquantized_fused_moe_method.py
@@ -211,16 +211,16 @@ class UnquantizedFusedMoEMethod(FusedMoEMethodBase, CustomOp):
         super().process_weights_after_loading(layer)
 
         # Padding the weight for better performance on ROCm
-        layer.w13_weight.data = self._maybe_pad_weight(layer.w13_weight.data)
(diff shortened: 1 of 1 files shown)


amd-fuweiy:
Hi team, we still think this PR is necessary, since it can solve the accuracy issue of ROCm when enabling AITER MOE for RL training. Without the PR, we have to disable AITER MOE, which harms the performance.

PR #31267, opened 2025-12-24
Fix skip_tokenizer_init=True crash for multimodal models

majiayu000:
## Summary
- Add validation to prevent `skip_tokenizer_init=True` with multimodal models
- Multimodal models require tokenizer for processor initialization
- Provides clear error message guiding users to set `skip_tokenizer_init=False`

Fixes #31123

## Test plan
- Added test case `test_skip_tokenizer_init_multimodal_not_allowed`
- Verified validation raises `ValueError` with descriptive message

Signed-off-by: majiayu000 <1835304752@qq.com>

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1050,3 +1050,13 @@ def test_scheduler_config_init():
     with pytest.raises(AttributeError):
         # InitVar does not become an attribute
         print(SchedulerConfig.default_factory().max_model_len)
+
+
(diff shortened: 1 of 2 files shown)


yewentao256 (approved):
LGTM, thanks for the work! https://github.com/vllm-project/vllm/pull/31291 @yurekami CC, seems duplicate

DarkLight1337 (commented):
This won't work because Terratorch models requires `skip_tokenizer_init`

PR #31294, opened 2025-12-24
[Bugfix] Remove dead `block_quant_to_tensor_quant` function

yurekami:
## Summary
Fixes #30098

- Removed the `block_quant_to_tensor_quant` function which was dead code (never called anywhere in the codebase)
- The function contained a bug: it passed a tensor `x_s` to `group_broadcast()` which expects a `GroupShape` tuple
- Removed the now-unused `group_broadcast` import

## Background

The original issue reported:
1. The docstring mentioned a "block size" parameter that didn't exist in the function signature
2. The function's logic was misleading

Investigation revealed the function was never used anywhere and also had a bug in its implementation (incorrect argument type).

## Test plan
- [ ] Linting passes locally
- [ ] No test changes needed (removing dead code)


yewentao256 (approved):
LGTM, thanks for the work!

PR #31307, opened 2025-12-24
fix: broadcast ngram/suffix draft tokens across TP ranks

yurekami:
## Summary

Fixes #31154

When using ngram or suffix speculative decoding with tensor parallelism (TP > 1), draft tokens could occasionally diverge across TP ranks due to rare non-determinism in the CPU-based proposers. This caused NCCL collective operations to hang because different ranks had different `num_tokens` values.

### Root Cause

The ngram and suffix proposers run independently on each TP rank using CPU-based computation (numba for ngram). While the random seed is set identically across ranks, rare non-determinism can cause draft tokens to diverge. When this happens:
1. Different ranks have different numbers of draft tokens
2. NCCL collectives expect the same tensor shapes across ranks
3. The operation hangs indefinitely

### Fix

Broadcast draft tokens from rank 0 to all other TP ranks after proposal. This ensures all ranks have identical draft tokens. The `broadcast_object` method already handles the single-GPU case (no-op when `world_size == 1`), so there's no overhead for non-TP configurations.

### Changes

- Added `get_tp_group().broadcast_object()` call after ngram proposer
- Added `get_tp_group().broadcast_object()` call after suffix proposer
- Eagle and Medusa proposers are GPU-based and already synchronized via NCCL, so they don't need this fix

## Test plan

This fix addresses a rare race condition that's difficult to reproduce deterministically. Testing approach:
1. Run ngram speculative decoding with TP=2+ under high load
2. Verify no NCCL hangs occur over extended periods
3. Verify correctness of generated outputs

The broadcast adds minimal overhead as it only transfers a small list of integers per batch.

### DIFF
--- a/vllm/config/model.py
+++ b/vllm/config/model.py
@@ -2,7 +2,7 @@
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 
 import warnings
-from collections.abc import Callable
+from collections.abc import Callable, Iterator
 from dataclasses import InitVar, field
 from functools import cached_property
 from typing import TYPE_CHECKING, Any, Literal, cast, get_args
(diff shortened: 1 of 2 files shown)


yurekami:
Closing as duplicate of #31288 which has all CI checks passing.

PR #31314, opened 2025-12-24
[Bugfix] Suppress UserWarning for non-writable buffer in binary2tensor

yurekami:
## Summary

Fixes #26781

When decoding base64-encoded embeddings to torch tensors, a `UserWarning` was emitted because `torch.from_numpy` was called on a read-only numpy array:

```
UserWarning: The given buffer is not writable, and PyTorch does not support non-writable tensors. This means you can write to the underlying (supposedly non-writable) buffer using the tensor. You may want to copy the buffer to protect its data or make it writable before converting it to a tensor.
```

### Root Cause
- `base64.b64decode()` returns immutable `bytes`
- `np.frombuffer()` on immutable bytes returns a read-only numpy array
- `torch.from_numpy()` on read-only arrays triggers the UserWarning

### Fix
Use `bytearray(binary)` to create a mutable copy before passing to `np.frombuffer`. This ensures the numpy array is writable while maintaining the efficient numpy-based conversion path.

This is more efficient than using `torch.frombuffer` directly because:
1. It avoids the warning without needing `warnings.filterwarnings`
2. It properly handles all embed_dtype formats (float32, float16, bfloat16, fp8_e4m3, fp8_e5m2) that numpy doesn't natively support
3. The bytearray copy is necessary anyway since the original bytes are immutable

### Changes
- `vllm/utils/serial_utils.py`: Use `bytearray(binary)` in `binary2tensor()`
- `tests/utils_/test_serial_utils.py`: Add test to verify no warning is emitted

## Test Plan
- Added `test_binary2tensor_no_warning` that verifies no UserWarning about non-writable buffers is emitted
- Existing `test_encode_and_decode` continues to pass

### DIFF
--- a/vllm/utils/serial_utils.py
+++ b/vllm/utils/serial_utils.py
@@ -107,7 +107,10 @@ def binary2tensor(
     torch_dtype = EMBED_DTYPE_TO_TORCH_DTYPE[embed_dtype]
     np_dtype = EMBED_DTYPE_TO_NUMPY_DTYPE_VIEW[embed_dtype]
 
-    np_array = np.frombuffer(binary, dtype=np_dtype).reshape(shape)
+    # Use bytearray to create a mutable copy of the binary data.
+    # This ensures np.frombuffer returns a writable array, avoiding
+    # UserWarning from torch.from_numpy on read-only arrays.
+    np_array = np.frombuffer(bytearray(binary), dtype=np_dtype).reshape(shape)
 
     if endianness != "native" and endianness != sys_byteorder:
         np_array = np_array.byteswap()
--- a/tests/utils_/test_serial_utils.py
+++ b/tests/utils_/test_serial_utils.py
@@ -38,3 +38,42 @@ def test_encode_and_decode(embed_dtype: str, endianness: str):
             name_1="new",
             tol=1e-2,
         )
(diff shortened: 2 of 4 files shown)


vllm/utils/serial_utils.py
```diff
@@ -107,7 +107,10 @@ def binary2tensor(
     torch_dtype = EMBED_DTYPE_TO_TORCH_DTYPE[embed_dtype]
     np_dtype = EMBED_DTYPE_TO_NUMPY_DTYPE_VIEW[embed_dtype]
 
-    np_array = np.frombuffer(binary, dtype=np_dtype).reshape(shape)
+    # Use bytearray to create a mutable copy of the binary data.
+    # This ensures np.frombuffer returns a writable array, avoiding
+    # UserWarning from torch.from_numpy on read-only arrays.
+    np_array = np.frombuffer(bytearray(binary), dtype=np_dtype).reshape(shape)
```
noooop, line 114: Our testing found that this zero-copy method is slightly faster, albeit with UserWarning.

noooop: I prefer using the warning_suppressed approach approved by @zou3519 in discussion #26781.

PR #31317, opened 2025-12-24
pin lora_b moe weights on cpu

gnovack:
## Purpose
The current weight loading logic in `FusedMoE3DWithLoRA` permutes the LoRA B weights on CPU before copying to GPU, this results in a slow copy of non-contigous, pageable memory (see below):
Screenshot 2025-12-24 at 9 32 25 AM

This PR moves the LoRA B weight permutation logic one layer up, into the model manager's `activate_adapter` function, so that it is only run the first time that the LoRA is loaded from disk, and will not run again if/when this adapter is swapped between CPU/GPU.

This ensures that all LoRA weight copies from CPU to GPU use pinned, contiguous memory, reducing the total transfer time by ~40%:
Image 12-24-25 at 9 38 AM

### Ran LoRA test cases for 3D MoE models:
```
pytest tests/lora/test_gptoss_tp.py tests/lora/test_qwen3moe_tp.py 
```

### Benchmarked e2e TTFT when loading LoRAs from CPU to GPU

**tl;dr** - benchmarking shows a ~40% decrease in median TTFT when frequently loading LoRA adapters from CPU to GPU.

Server command (loads 8 LoRAs, with only one on CPU, causing frequent swapping in from CPU to GPU during runtime):
```
	--trust-remote-code --enable-lora --max-loras 1 --max-cpu-loras 8 \
	-tp=1 --port 8001 --lora-modules \
		lora1=/mnt/models/rank-32-adapters/gpt-oss-20b-gsm8k-rank32-lr1e-5 \
		lora2=/mnt/models/rank-32-adapters/gpt-oss-20b-gsm8k-rank32-lr1e-5 \
		lora3=/mnt/models/rank-32-adapters/gpt-oss-20b-gsm8k-rank32-lr1e-5 \
		lora4=/mnt/models/rank-32-adapters/gpt-oss-20b-gsm8k-rank32-lr1e-5 \
		lora5=/mnt/models/rank-32-adapters/gpt-oss-20b-gsm8k-rank32-lr1e-5 \
		lora6=/mnt/models/rank-32-adapters/gpt-oss-20b-gsm8k-rank32-lr1e-5 \
		lora7=/mnt/models/rank-32-adapters/gpt-oss-20b-gsm8k-rank32-lr1e-5 \
		lora8=/mnt/models/rank-32-adapters/gpt-oss-20b-gsm8k-rank32-lr1e-5 \
	--max-lora-rank 32 \
	--max-num-seqs 8 --gpu-memory-utilization 0.95 \
	--no-enable-prefix-caching
```

Benchmark command:
```
vllm bench serve --model openai/gpt-oss-20b --backend vllm --endpoint /v1/completions --ignore-eos --random-range-ratio=0.05 --random-input-len=1600  --random-output-len=200 --max-concurrency 1 --lora-modules lora1 --lora-modules lora2 --lora-modules lora3 --lora-modules lora4 --lora-modules lora5 --lora-modules lora6 --lora-modules lora7 --lora-modules lora8 --num-prompts 100 --port 8001
```

Results before PR:
```
Successful requests:                     100       
Maximum request concurrency:             1         
Benchmark duration (s):                  163.72    
Total input tokens:                      160151    
Total generated tokens:                  20118     
Request throughput (req/s):              0.61      
Output token throughput (tok/s):         122.88    
Peak output token throughput (tok/s):    144.00    
Peak concurrent requests:                2.00      
Total token throughput (tok/s):          1101.09   
Mean TTFT (ms):                          232.66    
Median TTFT (ms):                        222.25    
P99 TTFT (ms):                           386.80    
Mean TPOT (ms):                          7.02      
Median TPOT (ms):                        7.01      
P99 TPOT (ms):                           7.15      
Mean ITL (ms):                           7.02      
Median ITL (ms):                         7.00      
P99 ITL (ms):                            7.42      
```

Results after PR:
```
Successful requests:                     100       
Maximum request concurrency:             1         
Benchmark duration (s):                  152.80    
Total input tokens:                      160151    
Total generated tokens:                  20118     
Request throughput (req/s):              0.65      
Output token throughput (tok/s):         131.66    
Peak output token throughput (tok/s):    144.00    
Peak concurrent requests:                2.00      
Total token throughput (tok/s):          1179.73   
Mean TTFT (ms):                          127.39    
Median TTFT (ms):                        134.17    
P99 TTFT (ms):                           138.23    
Mean TPOT (ms):                          7.00      
Median TPOT (ms):                        6.99      
P99 TPOT (ms):                           7.04      
Mean ITL (ms):                           7.00      
Median ITL (ms):                         6.99      
P99 ITL (ms):                            7.33      
```

---
<details>

</details>

### DIFF
--- a/vllm/lora/model_manager.py
+++ b/vllm/lora/model_manager.py
@@ -574,91 +574,8 @@ class LoRAModelManager:
             lora.optimize()
 
         for module_name, module in self.modules.items():
-            module_lora = self._get_lora_layer_weights(lora_model, module_name)
-            if not module_lora:
-                continue
-
-            # Note (gnovack) - If MOE lora weights are not split into
-            # num_experts chunks, we split them here
-            if isinstance(module, FusedMoE3DWithLoRA) and torch.is_tensor(
-                module_lora.lora_a
-            ):
-                # Handle PEFT file format where experts.base_layer is the
-                # gate_up_proj and experts is the down_proj
-                gate_up_proj_lora = self._get_lora_layer_weights(
-                    lora_model, module_name + ".base_layer"
-                )
-                down_proj_lora = module_lora
-                # FIXME Edge case where LoRA is not added to gate_up_proj
-                # or down_proj
-                assert gate_up_proj_lora is not None
-                assert down_proj_lora is not None
-                if self._is_3d_moe_model:
-                    num_experts = module.w13_lora_a_stacked[0].shape[1]
-
-                    # (num_experts,rank,input_size)
-                    gate_up_proj_lora.lora_a = gate_up_proj_lora.lora_a.reshape(
-                        num_experts, -1, gate_up_proj_lora.lora_a.shape[-1]
-                    )
-                    down_proj_lora.lora_a = down_proj_lora.lora_a.reshape(
-                        num_experts, -1, down_proj_lora.lora_a.shape[-1]
-                    )
-
-                    # (output_size,num_experts,rank)
-                    gate_up_proj_lora.lora_b = gate_up_proj_lora.lora_b.reshape(
-                        gate_up_proj_lora.lora_b.shape[0], -1, num_experts
-                    )
-                    down_proj_lora.lora_b = down_proj_lora.lora_b.reshape(
-                        down_proj_lora.lora_b.shape[0], -1, num_experts
-                    )
-
-                    # (num_experts,output_size,rank)
-                    gate_up_proj_lora.lora_b = gate_up_proj_lora.lora_b.permute(
-                        1, 0, 2
-                    ).contiguous()
-                    down_proj_lora.lora_b = down_proj_lora.lora_b.permute(
-                        1, 0, 2
-                    ).contiguous()
-
-                    module_lora.lora_a = [
-                        gate_up_proj_lora.lora_a,
-                        down_proj_lora.lora_a,
-                    ]
-                    module_lora.lora_b = [
-                        gate_up_proj_lora.lora_b,
-                        down_proj_lora.lora_b,
-                    ]
-                else:
-                    # Some 3D MoE models haven't added the `is_3d_moe_weight`
-                    # attribute yet, so fallback here
-                    num_experts = module_lora.lora_a.shape[0] // module_lora.rank
-
-                    gate_proj_a = gate_up_proj_lora.lora_a.chunk(num_experts, dim=0)
-                    up_proj_a = gate_up_proj_lora.lora_a.chunk(num_experts, dim=0)
-
-                    gate_proj_b = gate_up_proj_lora.lora_b[::2, ...].chunk(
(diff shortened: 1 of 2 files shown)


vllm/lora/model_manager.py
```diff
@@ -171,6 +171,37 @@ def activate_adapter(
                 assert gate_up_proj_lora is not None
                 assert down_proj_lora is not None
                 if self._is_3d_moe_model:
+                    num_experts = module.w13_lora_a_stacked[0].shape[1]
```
kfhfar: Could we have a separate function in model_manager to run this code inside the if_statement. It looks brittle for further extension.

gnovack:
> The set_lora calls are made for all the LoRA layers from model_manager. I wonder if this change could be applied to all of them and is not necessarily limited to 3DFusedMOE as other also have similar logic.

That's a good point. Based on my profiling, it seems that the linear layer adapters don't suffer from this issue (i.e. the CPU-to-GPU copies of these linear layers weights use pinned memory). I have not checked yet if the "non-3D" MoE layer adapters (e.g. DeepSeek MoE adapters) could benefit from a change like this.

vllm/lora/model_manager.py
jeejeelee: Agree. Perhaps we should add it [here](https://github.com/vllm-project/vllm/blob/main/vllm/lora/model_manager.py#L519), thus  we can  avoid repeated pin memory

gnovack: We could move it there, but that would require moving all of the MoE weight processing logic from `activate_adapter` to `_create_merged_loras_inplace`, including duplicating the loop over `self.modules.items()` and calls to `self._get_lora_layer_weights`. 

A simpler solution might be to defer any calls to `pin_memory` until just before the `set_lora` calls - this would ensure that pinning is only done once, and is not invalidated by any later permuting of the weights

gnovack: fyi - i moved this logic from the `activate_adapter` to the `_create_merged_loras_inplace` function. As mentioned above, this does introduce some duplication, but i do agree that it makes more sense to have this logic on the "add adapter" instead of the "activate adapter" path. @jeejeelee, let me know what you think

jeejeelee: >A simpler solution might be to defer any calls to pin_memory until just before the set_lora calls - this would ensure that pinning is only done once, and is not invalidated by any later permuting of the weights

Makes sense,and then maybe we should create a separate function to hanlde moe lora weights

gnovack: @jeejeelee i moved this logic into a separate function and rebased to include the accuracy bugfix changes as well. please take a look when you have some time. thanks!

jeejeelee (approved):
LGTM, thank you for this improvment. Hapy new year~

PR #31324, opened 2025-12-24
[CI] Fix flaky vision beam search test with flexible semantic validation

AndreasKaratzas:
Replaces brittle exact string matching in `test_single_chat_session_image_base64encoded_beamsearch` with flexible semantic term validation, fixing intermittent test failures and rendering the test platform-agnostic and compatible with other attention backends as well.

## Changes

- Remove platform-specific expected results (`EXPECTED_MM_BEAM_SEARCH_RES`, `EXPECTED_MM_BEAM_SEARCH_RES_ROCM`)
- Add `REQUIRED_BEAM_SEARCH_TERMS` with AND/OR logic for semantic validation
- Add `check_output_matches_terms()` helper function
- Add `warnings.warn()` to emit beam search outputs for debugging
- Unify test logic across CUDA and ROCm platforms

## Problem

The test was failing intermittently because it expected exact string matches like:
```
- The image contains a Venn diagram with three over
+ The image shows a colorful Venn diagram with
```

## Solution

Validate semantic correctness instead of exact wording. For example, the Venn diagram image now requires outputs to contain "venn" AND "diagram", allowing natural variation in phrasing.

### DIFF
--- a/tests/entrypoints/openai/test_vision.py
+++ b/tests/entrypoints/openai/test_vision.py
@@ -2,6 +2,7 @@
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 
 import json
+import warnings
 
 import openai
 import pytest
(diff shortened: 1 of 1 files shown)


AndreasKaratzas:
> Hmm how flaky is this test really? Buildkite shows it as non-flaky https://buildkite.com/organizations/vllm/analytics/suites/ci-1/tests?branch=main&period=14days&query=test_single_chat_session_image_base64encoded_beamsearch and I haven't noticed any flakiness either. Usually when this test starts to fail, it fails consistently.

@DarkLight1337 It is more flaky on ROCm. But most importantly, I thought that if more attention backends are tested here, then the hard-coded results are going to be different. So I was thinking that the solution that I am proposing is more scalable and maintainable. Also it is platform-agnostic since it covers both Triton (default ROCm backend) with flash attention for multi-modal, and flash attention for NVIDIA.

tests/entrypoints/openai/test_vision.py
```diff
@@ -341,8 +323,31 @@ async def test_single_chat_session_image_base64encoded_beamsearch(
...
     assert len(chat_completion.choices) == 2
-    for actual, expected_str in zip(chat_completion.choices, expected_res):
-        assert actual.message.content == expected_str
+
+    # Verify beam search produces two different non-empty outputs
+    content_0 = chat_completion.choices[0].message.content
+    content_1 = chat_completion.choices[1].message.content
+
+    # Emit warning with beam search outputs for debugging
+    warnings.warn(
```
DarkLight1337: Just print it? Why use warnings?

AndreasKaratzas: Honestly, the only reason that I have is that I saw similar logic being used in other tests (e.g.: `tests/models/multimodal/generation/`, specifically `test_pixtral.py::test_chat`) and I thought this was the "architecture" preferred. I am pushing a commit right now to make this use `logging`.

AndreasKaratzas: @DarkLight1337 Comment has been addressed in latest commit. Thank you for the review :)

DarkLight1337: A normal `print` statement should work already, since pytest will capture the printed output for the failed tests

AndreasKaratzas: @DarkLight1337 Done :)

DarkLight1337 (approved):

PR #31327, opened 2025-12-25
[ROCm] Migrate xgrammar to upstream release

AndreasKaratzas:
This PR updates the xgrammar dependency in `requirements/rocm-test.txt` from a custom fork to the upstream PyPI release.

### Changes
- Replaced `xgrammar @ git+https://github.com/divakar-amd/xgrammar@3272f7c...` with `xgrammar==0.1.29`

### Motivation
It's better practice to use upstream repositories instead of maintaining custom forks when possible. This is a follow-up to a comment from #29702 to eventually migrate to the upstream xgrammar repo.

### Testing
Verified on ROCm with the following tests passing:
```bash
pytest -v -s tests/v1/e2e/test_async_scheduling.py
pytest -v -s tests/v1/entrypoints/llm/test_struct_output_generate.py::test_structured_output_auto_mode[Qwen/Qwen2.5-1.5B-Instruct-auto]
pytest -v -s tests/v1/e2e/test_spec_decode.py
```

### DIFF
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -24,7 +24,7 @@ outlines_core == 0.2.11
 # required for outlines backend disk cache
 diskcache == 5.6.3
 lark == 1.2.2
-xgrammar == 0.1.27; platform_machine == "x86_64" or platform_machine == "aarch64" or platform_machine == "arm64" or platform_machine == "s390x" or platform_machine == "ppc64le"
+xgrammar == 0.1.29; platform_machine == "x86_64" or platform_machine == "aarch64" or platform_machine == "arm64" or platform_machine == "s390x" or platform_machine == "ppc64le"
 typing_extensions >= 4.10
 filelock >= 3.16.1 # need to contain https://github.com/tox-dev/filelock/pull/317
 partial-json-parser # used for parsing partial JSON outputs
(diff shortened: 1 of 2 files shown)


AndreasKaratzas:
**Update:** Bumped xgrammar to 0.1.29 in `common.txt`.

This resolves a dependency conflict with `rocm-test.txt` which required 0.1.29, while `common.txt` pinned 0.1.27. Since `rocm-test.txt` includes `common.txt` via `-r common.txt`, uv couldn't resolve both constraints simultaneously.

The reason we need 0.1.29 for ROCm: it includes a fix for the Triton kernel warp size on AMD GPUs (mlc-ai/xgrammar#476). Without this fix, structured output on MI300 fails with:

```
triton.runtime.errors.OutOfResources: out of resource: threads, Required: 2048, Hardware limit: 1024
```

The issue was that the Triton kernel hardcoded warp size to 64 (NVIDIA), but AMD GPUs use warp size 32. This caused the kernel to request more threads than available on AMD hardware.

AndreasKaratzas:
There is an AMD CI run for this PR before and after this addition, and it looks like everything is still the same:
Before: https://buildkite.com/vllm/amd-ci/builds/2103
After: https://buildkite.com/vllm/amd-ci/builds/2130

LucasWilkinson:
cc @tjtanaa

AndreasKaratzas:
@DarkLight1337 These b200 failures are not xgrammar related and are already known right? I have few of them already here: https://github.com/vllm-project/vllm/pull/30719

Issue #14003, opened 2025-02-28
[Feature]: Implement Concurrent Partial Prefills In V1 Engine


robertgshaw2-redhat:
### 🚀 The feature, motivation and pitch

In V0, we support concurrent partial prefills to avoid TTFT latency with long requests. Implement it in V1

cc @WoosukKwon 

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

plops655:
I can take this up.

ccw1996:
I'm interested in contributing this.

houseroad:
Shall we mark this as done? Since I feel in the context of chunked prefill, probably doesn't make sense to add support for max_num_partial_prefills and max_long_partial_prefills. cc: @comaniac

comaniac:
> Shall we mark this as done? Since I feel in the context of chunked prefill, probably doesn't make sense to add support for max_num_partial_prefills and max_long_partial_prefills. cc: [@comaniac](https://github.com/comaniac)

IIRC in v0 we still apply `max_num_partial_prefills` for example even when chunked prefill is enabled, but I'm not fully sure which one (`max_num_partial_prefills ` or `long_prefill_token_threshold ` has a higher priority). This is a topic worth to discuss. cc @joerunde @njhill

anirmohammadnoori72:
Hi! I’d like to take this issue and start working on it. I’ve already begun reviewing the code. Please let me know if it’s still available.

Thanks!

whuLames:
Is there any progress about the support of chunked prefill in V1

hibukipanim:
> Is there any progress about the support of chunked prefill in V1

V1 Supports chunk prefill, and it's enabled by default.

What is still not supported is "Concurrent Partial Prefills", which this PR for V0 explains well: https://github.com/vllm-project/vllm/pull/10235

pansicheng:
@comaniac @joerunde @njhill Hi! Mind taking a look at my PR (https://github.com/vllm-project/vllm/pull/21651)? It limits the number of concurrent long-text prefills processed per step by using the max_long_partial_prefills parameter in v1.

Csrayz:
[...]

PR #10235, opened 2024-11-11
[Core] Reduce TTFT with concurrent partial prefills

joerunde:
Replaces #10061, as inspired by @njhill and @comaniac's comments. Co-authored by @prashantgupta24 

Context: our customers running large multi-tenanted SaaS deployments of vLLM have a problem where high volumes of small-prompt requests are usually processed smoothly, but quickly pile up in a giant queue when a small number of large-prompt requests are submitted. We see the decoding throughput drop to zero on multiple replicas when this happens.

The current chunked prefill implementation only allows a single sequence to be partially prefilled at a time. This has a few limitations:
- Multiple medium-sized prompts must wait to be prefilled serially, increasing TTFT for those in the back of the queue
- A single very large prompt will block all other prompts from prefilling for many iterations. This can eventually starve decoding- for example a 130k token prompt with `—max-num-batched-tokens=512` will take about 250 iterations to prefill, in which time the currently decoding sequences may all finish. Send a few of these requests at once and very quickly nothing will be decoding.

This PR implements both
- An explicit setting for the number of sequences that can be partially prefilled concurrently. This can be configured with `--max-num-partial-prefills=N`
- A limit on the number of “very long prompt” sequences that can be prefilled concurrently. This can be configured with
  - `--max-long-partial-prefills=N` to set the limit on the number of long sequences that can be concurrently prefilled. This defaults to 1 sequence.
  - `--long-prefill-threashold=x%` to set a percentage of the context length that determines which sequences are considered "long". This defaults to 4%

This is implemented in the v0 scheduler. We’re aware that the v1 implementation is underway and will later become the default, but we need a fix for our customers soon and we hope that what we discover here may help inform a different, better solution in the v1 scheduler.

To test this we created three scenarios, a “medium request” case, a “large request” case, and a “mixed” case.

For the medium request case, we created a subset of the `sharegpt` dataset with 900 small requests (<50 prompt characters) and 100 of the largest requests (typically between 10k and 20k prompt characters, which we call “medium” sized). We modified the `benchmark_serving.py` test to not filter out any of the small or large requests, and ran it with this dataset. What we expect to find is similar throughput compared to the main branch, but much lower TTFT on the small requests. Since 10% of the requests are larger than the rest, we should see better TTFT at p90 and below, with comparable TTFT above p90.

[...]

PR #15419, opened 2025-03-24
[V1] Support long_prefill_token_threshold in v1 scheduler

houseroad:
To address https://github.com/vllm-project/vllm/issues/14003

To support concurrent partial prefill, we need to define long_prefill_token_threshold. Then the tokens from different prefill requests can be added. Since there is no difference between prefill and decode, just need to control the token threshold.

pytest tests/v1/core/test_scheduler.py
e2e test: VLLM_USE_V1=1 pytest tests/v1/core/test_scheduler_e2e.py

PR #31330, opened 2025-12-25
[Core] Concurrent partial prefills for V1

ppppqp:
## Purpose
Implements https://github.com/vllm-project/vllm/issues/14003, since we decide to include it in v1.
Referencing the implementation of the original PR: https://github.com/vllm-project/vllm/pull/10235

In short, concurrent partial prefill technique limits the number of large requests from starving small requests, improving throughput and TTFT overall. There are three key parameters for this strategy:
- `max_num_partial_prefills`: this controls how many requests that are **guaranteed** to progress in this run. At the start of each scheduler run, we make speculation and distribute the budget to each partial prefill slots evenly. The budget distribution makes sure that at least `max_num_partial_prefills` can get new tokens prefilled. If some request does not use all budgets distributed to it, it can happen that there are some budgets left, so more requests can be served.
- `max_long_partial_prefills`: this controls how many large requests can we put in the partial prefill slots. For example, if `max_num_partial_prefills=4 & max_long_partial_prefills=2`, we can have 2 large requests and 2 small requests each run. The default number is `1`.
- `long_prefill_token_threshold`: this controls the criteria for determination of "large request". If the number of token of the prompt exeeds the threshold, it is considered large request and will be limited by `max_long_partial_prefills`.

## Test Plan
Unit test with parity to the original PR.
I'm not sure about whether we should have parity to this unit test, because based on my testing it seems like the alignment gets abstracted away from scheduler. Would need some help here.
https://github.com/vllm-project/vllm/pull/10235/files#diff-2c6af6e25b8d1074f25ef5ad2901121b30bc1528de74d2b3625636fcb8181624R782-R831

### Benchmark plan
I followed the setup of the original PR with custom dataset I generated from [shareGPT creative writing](https://huggingface.co/datasets/Nitral-AI/Creative_Writing-ShareGPT/tree/main) dataset. The distribution of token count is shown below, showing a three groups of small/medium/large prompts:
image

I tested three versions on the dataset with A40, with the dataset: [benchmark-final.jsonl.zip](https://github.com/user-attachments/files/24355838/benchmark-final.jsonl.zip)
1. `main` branch
2. This branch with `max_num_partial_prefills=1`
3. this branch with `max_num_partial_prefills=4 & long_prefill_token_threshold=2048`
For each version, I also tested with `output_num=128`(which is default) and `output_num=1`.

```bash
vllm serve NousResearch/Hermes-3-Llama-3.1-8B [--max_num_partial_prefills=4 --long_prefill_token_threshold=2048]


vllm bench serve \
  --model NousResearch/Hermes-3-Llama-3.1-8B \
    --dataset-name custom \
  --dataset-path benchmark.jsonl \
  --num-prompts -1 \
  --metric-percentiles 80,85,90,95,99 \
  --request-rate 12 \
  --disable_shuffle \
  [--output_len=1]
```

Sorry that the chart is probably not organized in the clearest way. Please compare the stats in greyed column together and in white column together.
image

Some interesting observation:
1. The performance boost on TTFT for `output_len=128` experiment group is not significant (~20%). I did some investigation and I think it's because the decoding phase largely averaged out the TTFT since even after the small requests get prefilled, they still need to be in the queue for fairly large number of times, and therefore capped by `max_num_seqs` which is 128 by default. If we only consider the prefilling phase (i.e. we set `output_len` to be 1), the TTFT improvement is significant (~400%). I did an extra experiment to further confirm this issue (in the last column of the chart, where `max_num_seqs=1024`)
2. As we increase `max_num_prefills`(1->4->16), we increase the throughput pretty consistently.
3. There's a tradeoff for TTFT P99. The performance gets consistently worse as we increase the throughput.

# Considerations
Some best practice suggestion around this feature:
1. `long_prefill_token_threshold` is best used if around `max_num_batched_tokens`. If it's too high away, the single large request can still starve the queue. If it's lower, then the throughput for large requests get degraded quickly. In this case, if a large request is the only request in queue, it does not get full budget of the run.
2. `max_num_seqs` must be tuned up in accordance with the throughput improvement.

---
<details>

</details>

### DIFF
--- a/vllm/v1/core/sched/scheduler.py
+++ b/vllm/v1/core/sched/scheduler.py
@@ -340,6 +340,10 @@ class Scheduler(SchedulerInterface):
         partial_prefill_slot_budget = self._get_prefill_slot_budget(
             partial_prefill_metadata
         )
+        print(
+            f"Partial prefill metadata: {partial_prefill_metadata.active_prefills}, {partial_prefill_metadata.long_prefills}, {partial_prefill_metadata.schedulable_prefills}"
+        )
+        print(f"Partial prefill slot budget: {partial_prefill_slot_budget}")
 
         # First, schedule the RUNNING requests.
         req_index = 0
@@ -367,9 +371,11 @@ class Scheduler(SchedulerInterface):
                 + request.num_output_placeholders
                 - request.num_computed_tokens
             )
+            print(f"num_new_token_1 for request {request.request_id}: {num_new_tokens}")
             if 0 < self.scheduler_config.long_prefill_token_threshold < num_new_tokens:
                 num_new_tokens = self.scheduler_config.long_prefill_token_threshold
             num_new_tokens = min(num_new_tokens, token_budget)
+            print(f"num_new_token_2 for request {request.request_id}: {num_new_tokens}")
             if (
                 self.enable_concurrent_partial_prefill_scheduling
                 and partial_prefill_slot_budget is not None
@@ -379,12 +385,16 @@ class Scheduler(SchedulerInterface):
                 )
                 if prefill_state.is_prefill:
                     num_new_tokens = min(num_new_tokens, partial_prefill_slot_budget)
+                    print(
+                        f"num_new_token_3 for request {request.request_id}: {num_new_tokens}"
+                    )
 
             # Make sure the input position does not exceed the max model len.
             # This is necessary when using spec decoding.
             num_new_tokens = min(
                 num_new_tokens, self.max_model_len - 1 - request.num_computed_tokens
             )
+            print(f"num_new_token_4 for request {request.request_id}: {num_new_tokens}")
 
             # Schedule encoder inputs.
             encoder_inputs_to_schedule = None
@@ -1100,7 +1110,9 @@ class Scheduler(SchedulerInterface):
         for request in running_reqs:
             scheduled_tokens = num_scheduled_tokens.get(request.request_id, 0)
             num_tokens_after_step = request.num_computed_tokens + scheduled_tokens
-            print(scheduled_tokens, num_tokens_after_step, request.num_prompt_tokens)
+            print(
+                f"scheduled_tokens: {scheduled_tokens}, computed_token: {request.num_computed_tokens}, num_tokens_after_step: {num_tokens_after_step}, request.num_prompt_tokens: {request.num_prompt_tokens}"
+            )
             if num_tokens_after_step >= request.num_prompt_tokens:
                 finishing_prefills.append(request)
             else:
--- a/vllm/v1/core/sched/scheduler.py
+++ b/vllm/v1/core/sched/scheduler.py
@@ -115,6 +115,18 @@ class PartialPrefillMetadata:
             self.long_prefills += 1
 
 
+@dataclass
+class PrefillState:
+    """Lightweight state used to reason about a request's prefill status."""
+
+    # whether the request in in prefill phase
+    is_prefill: bool
+    # number of remaining tokens to prefill
+    remaining_tokens: int
+    # whether the prefill is considered a long prefill
+    is_long_prefill: bool
+
+
 class Scheduler(SchedulerInterface):
     def __init__(
         self,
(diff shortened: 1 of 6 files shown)


vllm/v1/core/sched/scheduler.py
```diff
@@ -57,6 +58,75 @@
...
+class PrefillState:
+    """Lightweight state used to reason about a request's prefill status."""
+
+    # whether the request in in prefill phase
+    is_prefill: bool
+    # number of remaining tokens to prefill
+    remaining_tokens: int
+    # whether the prefill is considered a long prefill
+    is_long_prefill: bool
+
```
ppppqp, line 127: Adding a new dataclass here as an abstraction, in case in future we want to implement more sophisticated cocurrency strategy (like the strategy to determine what is a long prefill)

vllm/v1/core/sched/scheduler.py
```diff
@@ -282,6 +371,15 @@ def schedule(self) -> SchedulerOutput:
...
             num_new_tokens = min(num_new_tokens, token_budget)
+            if (
+                self.enable_concurrent_partial_prefill_scheduling
+                and partial_prefill_slot_budget is not None
+            ):
+                prefill_state = self._get_request_prefill_state(
+                    request, request.num_computed_tokens
+                )
+                if prefill_state.is_prefill:
+                    num_new_tokens = min(num_new_tokens, partial_prefill_slot_budget)
```
ppppqp, line 386: Cap the computed `num_new_tokens` with the allocated budget. This ensures that all requests that are granted a partial prefill slot can progress.

vllm/v1/core/sched/scheduler.py
```diff
@@ -873,6 +1013,109 @@ def _make_cached_request_data(
...
+            request.num_output_tokens == 0
+            and num_computed_tokens < request.num_prompt_tokens
+        )
+
+    @staticmethod
+    def _remaining_prefill_tokens_with_tokens(
+        request: Request, num_computed_tokens: int
+    ) -> int:
+        """Get the number of remaining prefill tokens"""
+        return max(request.num_prompt_tokens - num_computed_tokens, 0)
```
ppppqp, line 1097: Not sure about these two helper function. Could you take a look and see if my understanding is correct?

robertgshaw2-redhat:
> > I believe we already support this functionality in V1. Specifically, the `--long_prefill_token_threshold`.
> 
> Hi! Are you talking about this [PR](https://github.com/vllm-project/vllm/pull/15419/files)? If so, I believe it only added the parameter but does not fully resolve the throughput issue. Consider this case: The budget for each run is `2048`, and if we set the threshold to be greater than `2048`, the threshold does nothing. If we set threshold to be lower than `2048`, say `512`, then if we see the requests coming in this order:
> 
> ```
> Request 1(100k tokens)
> Request 2(100k tokens)
> Request 3(100k tokens)
> Request 4(100k tokens)
> Request 5(5 tokens)
> ```
> 
> The Reuqest 1~4 is still gonna get all budgets for each run (512 * 4), and Request 5 will be starved.

Yes agreed - so the updated semantics would then limit the number of long prefills running at once?

chaunceyjiang:
> Hi! Do you mean the tuning the input_len to these numbers or I should actually generate a dataset to have the input prompt of that length? I'm not sure about whether the benchmark I've done is good enough or there's still need extra effort needed. Thanks!

i.e.
```
vllm bench serve --model XXXX --random-input-len 2048 --random-output-len 1024  --max-concurrency 120 --num-prompts 480 --port  8990
```

ppppqp:
> > Hi! Do you mean the tuning the input_len to these numbers or I should actually generate a dataset to have the input prompt of that length? I'm not sure about whether the benchmark I've done is good enough or there's still need extra effort needed. Thanks!
> 
> i.e.
> 
> ```
> vllm bench serve --model XXXX --random-input-len 2048 --random-output-len 1024  --max-concurrency 120 --num-prompts 480 --port  8990
> ```
@chaunceyjiang 
Actually, I don't think this makes sense. I tested and if I specify `--random-input-len 2048`, the lengths of all inputs are `2047`, where both cases should perform the same (since it's uniform distribution, partial prefilling should not observe any long request)
Let me know if there's anything else I need to benchmark on🙏.

If you intentionally need this benchmark of uniform distribution, I can also run the benchmark and provide the stats!

ppppqp:
@chaunceyjiang 
Hi! This is the benchmark result using
```
vllm serve NousResearch/Hermes-3-Llama-3.1-8B [--max_num_partial_prefills=4]

vllm bench serve --model NousResearch/Hermes-3-Llama-3.1-8B --random-input-len 2048 --random-output-len 1024  --max-concurrency 120 --num-prompts 480
```
- Comparing `main` branch and this branch with `max_num_partial_prefills=1`, the change in this PR has no performance regression on TTFT
- Comparing `main` branch and this branch with `max_num_partial_prefills=4`, The change in this PR causes about 1% slowdown on TTFT for uniformly distributed input.

image

Please let me know if I need to test on ` --random-input-len 16384/32768/65536`. It should take significantly longer to run the benchmark, but happy to do that as well if needed!

DmitriiShubin:
reopen

PR #31332, opened 2025-12-25
[BugFix] Fix async scheduling + reasoning with struct output

njhill:
Incorrect token accounting in `StructuredOutputManager.should_advance()` for async case.

### DIFF
--- a/vllm/v1/structured_output/__init__.py
+++ b/vllm/v1/structured_output/__init__.py
@@ -339,8 +339,9 @@ class StructuredOutputManager:
             return True
 
         # Check if reasoning ends in *this* step
+        delta_from = request.num_computed_tokens - request.num_output_placeholders
         if self.reasoner.is_reasoning_end_streaming(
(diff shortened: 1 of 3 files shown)


vllm/v1/structured_output/__init__.py
```diff
@@ -339,8 +339,9 @@ def should_advance(self, request: Request) -> bool:
             return True
 
         # Check if reasoning ends in *this* step
+        delta_from = request.num_computed_tokens - request.num_output_placeholders
```
chaunceyjiang, line 342: Thanks~ @njhill 

Indeed, I didn’t take this into consideration at the time.

chaunceyjiang (approved):
LGTM

vllm/v1/structured_output/__init__.py
njhill: Thanks @chaunceyjiang

PR #31336, opened 2025-12-25
[perf][async] support non cpu sync get logprob tensors for spec

izhuhaoran:
## Description:

This PR eliminates implicit CPU-GPU synchronization points in the `RejectionSampler` when computing logprobs during speculative decoding, specifically for the async scheduling path.

**Problem**
Previously, the rejection sampling logic used boolean masking to identify accepted tokens:
https://github.com/vllm-project/vllm/blob/ba25a6599212f01e6b877e22549612333e95b2c1/vllm/v1/sample/rejection_sampler.py#L191-L194
https://github.com/vllm-project/vllm/blob/ba25a6599212f01e6b877e22549612333e95b2c1/vllm/v1/sample/rejection_sampler.py#L203
Since the number of accepted tokens is dynamic and data-dependent, these operations required the CPU to know the output tensor sizes to alloc gpu mem, triggering implicit CPU-GPU synchronization. This stalled the GPU and negated the benefits of async scheduling.

image

image

**Solution**
1.  **Avoid Dynamic Shapes on GPU**: In async mode , the `RejectionSampler` now computes logprobs for **all** draft tokens (including rejected ones) using static tensor shapes. This removes the need for the CPU to query tensor sizes during GPU execution.
2.  **Deferred Filtering**: The filtering of rejected logprobs is moved to the `parse_output` stage (CPU side). 

### GPU Timeline

- Main

image

- This PR (No bubble)

image

### Bench Serve

- **Target**: Qwen/Qwen3-8B
- **Draft**: RedHatAI/Qwen3-8B-speculator.eagle3
- **num_speculative_tokens**: 2
- **bench args**: --dataset-name spec_bench --spec-bench-output-len 512 --logprobs 2 --max-concurrency 32 ----num-prompts 256

- Main (async_scheduling=True)

```
Successful requests:                     256       
Maximum request concurrency:             32        
Benchmark duration (s):                  50.40     
Total input tokens:                      72520     
Total generated tokens:                  127856    
Request throughput (req/s):              5.08      
Output token throughput (tok/s):         2537.00   
Peak output token throughput (tok/s):    1631.00   
Peak concurrent requests:                47.00     
Total token throughput (tok/s):          3975.99   
Mean TTFT (ms):                          222.49    
Median TTFT (ms):                        130.37    
P99 TTFT (ms):                           958.91    
Mean TPOT (ms):                          11.76     
Median TPOT (ms):                        11.93     
P99 TPOT (ms):                           14.38     
Mean ITL (ms):                           24.11     
Median ITL (ms):                         20.51     
P99 ITL (ms):                            110.74    
Mean E2EL (ms):                          6084.75   
Median E2EL (ms):                        6193.51   
P99 E2EL (ms):                           7744.30   
```

- This PR (async_scheduling=True)

```
Successful requests:                     256       
Maximum request concurrency:             32        
Benchmark duration (s):                  43.35     
Total input tokens:                      72520     
Total generated tokens:                  127987    
Request throughput (req/s):              5.91      
Output token throughput (tok/s):         2952.43   
Peak output token throughput (tok/s):    1888.00   
Peak concurrent requests:                49.00     
Total token throughput (tok/s):          4625.34   
Mean TTFT (ms):                          204.31    
Median TTFT (ms):                        114.88    
P99 TTFT (ms):                           891.48    
Mean TPOT (ms):                          10.15     
Median TPOT (ms):                        10.31     
P99 TPOT (ms):                           12.56     
Mean ITL (ms):                           20.84     
Median ITL (ms):                         17.28     
P99 ITL (ms):                            89.29     
Mean E2EL (ms):                          5268.62   
Median E2EL (ms):                        5400.22   
P99 E2EL (ms):                           6652.18   
```

### Unit Test

```
python -m pytest tests/v1/e2e/test_async_scheduling.py::test_with_spec_decoding -vvv -s
```
All passed

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit 708b9bd783d5bfe78455c05c511289dae491e662. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Improves async speculative decoding by eliminating GPU→CPU sync in rejection sampling logprobs and moving rejection filtering to the CPU.
> 
> - RejectionSampler now gathers logprobs for all draft positions (incl. rejected) with static indices; rejected entries are filtered later in `parse_output`
> - `parse_output` accepts optional `logprobs_tensors`, builds valid masks, filters, and returns `LogprobsLists` plus cumulative counts
> - Adds `LogprobsTensors.filter(mask)` to support CPU-side filtering
> - Updates GPU model runner to use new `parse_output` API in both async and non-async paths and set `ModelRunnerOutput.logprobs` accordingly
> - Minor refactor: expose `use_async_scheduling` earlier in runner init
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 708b9bd783d5bfe78455c05c511289dae491e662. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Removes implicit GPU→CPU sync in speculative decoding logprobs and defers filtering to the CPU.
> 
> - Compute logprobs for all draft positions with static indices in `RejectionSampler._get_logprobs_tensors`; rejected tokens are masked and filtered later
> - Extend `RejectionSampler.parse_output` to accept optional `logprobs_tensors` and return `LogprobsLists`, performing CPU-side masking via `valid_mask`
> - Add `LogprobsTensors.filter(mask)` for efficient CPU filtering and import `LogprobsLists`
> - Update GPU model runner to use new `parse_output` API and set `ModelRunnerOutput.logprobs` in both async and non-async paths
> - Minor refactor: expose `use_async_scheduling` earlier in runner init
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 75062c385ed60a5cddd04730aa032bcd28c85b58. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


izhuhaoran:
@njhill @benchislett  Could you please review this PR when you have time?

vllm/v1/sample/rejection_sampler.py
```diff
@@ -185,13 +186,30 @@ def _get_logprobs_tensors(
...
+            # all draft tokens, including the rejected ones. The rejected tokens will
+            # be filtered out in the `parse_output`.
+            logit_start_indices = cu_num_sampled_tokens
+            offsets = torch.arange(
+                sampled_token_ids.shape[-1], device=logit_start_indices.device
+            )
+            accepted_logit_indices = (
+                logit_start_indices.unsqueeze(1) + offsets.unsqueeze(0)
+            ).flatten()
+            accepted_logit_indices.clamp_(max=final_logits.shape[0] - 1)
```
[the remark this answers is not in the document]
izhuhaoran: The logic is correct for the async implementation. We are intentionally gathering logits for all tokens (including rejected and padded ones) to avoid dynamic masking on the GPU.

The `clamp_` prevents out-of-bounds errors for the padding tokens. Although these invalid tokens will get incorrect logits, they are discarded later during the filtering phase, so it doesn't affect the final result.

benchislett (commented):
Why only do this for async scheduling? If it's a more performant strategy, why not just do it whenever we can (even in sync-scheduled scenario)?

njhill:
@izhuhaoran I had the same thought, given that we are aiming for async-by-default (already true in non-spec-decode case), maybe it would be better to keep it simpler and accept the hit in non-async case? (unless the simplication isn't significant).

benchislett:
Yeah, I think it would be best to measure the hit to the sync-scheduled case, since generally having fewer syncs is better and the gpu work might turn out to be less significant than the cpu work on the critical path. 

If the perf is similar, I would prefer to unify the logic.

vllm/v1/sample/rejection_sampler.py
```diff
@@ -185,13 +185,20 @@ def _get_logprobs_tensors(
...
-        num_accepted_tokens = accepted_mask.sum(dim=-1)
-        accepted_logit_indices = accepted_mask.nonzero(as_tuple=True)[1]
-        accepted_logit_indices += cu_num_sampled_tokens.repeat_interleave(
-            num_accepted_tokens
+        # NOTE: To avoid cpu-gpu synchronization, we now simply compute indices for
+        # all draft tokens, including the rejected ones. The rejected tokens will
+        # be filtered out in the `parse_output`.
+        logit_start_indices = cu_num_sampled_tokens
+        offsets = torch.arange(
+            sampled_token_ids.shape[-1], device=logit_start_indices.device
```
njhill:
```suggestion
            sampled_token_ids.shape[-1],
            device=logit_start_indices.device,
            dtype=logit_start_indices.dtype,
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -238,17 +238,22 @@ def get_output(self) -> ModelRunnerOutput:
...
+                )
             )
 
         output = self._model_runner_output
         output.sampled_token_ids = valid_sampled_token_ids
-        if self._logprobs_tensors_cpu:
+        if logprobs_lists is not None:
+            output.logprobs = logprobs_lists
+        elif self._logprobs_tensors_cpu:
             output.logprobs = self._logprobs_tensors_cpu.tolists(cu_num_tokens)
```
njhill: don't need `cu_num_tokens` anymore, can move the non-spec logprobs inside `max_gen_len == 1` branch:
```python
            for i in self._invalid_req_indices:
                valid_sampled_token_ids[i].clear()
            logprobs_lists = None
            if self._logprobs_tensors_cpu is not None:
                logprobs_lists = self._logprobs_tensors_cpu.tolists()
        else:
            valid_sampled_token_ids, logprobs_lists = RejectionSampler.parse_output(
                self.sampled_token_ids_cpu,
                self.vocab_size,
                self._invalid_req_indices,
                logprobs_tensors=self._logprobs_tensors_cpu,
            )

        output = self._model_runner_output
        output.sampled_token_ids = valid_sampled_token_ids
        output.logprobs = logprobs_lists
```

vllm/v1/sample/rejection_sampler.py
```diff
@@ -185,13 +185,20 @@ def _get_logprobs_tensors(
...
         )
+        accepted_logit_indices = (
+            logit_start_indices.unsqueeze(1) + offsets.unsqueeze(0)
+        ).flatten()
```
njhill, line 199: can avoid an allocation
```suggestion
            offsets.unsqueeze(0).add_(logit_start_indices.unsqueeze(1)).flatten()
        )
```

vllm/v1/sample/rejection_sampler.py
```diff
@@ -185,13 +185,20 @@ def _get_logprobs_tensors(
...
+        accepted_logit_indices.clamp_(max=final_logits.shape[0] - 1)
+        accepted_tokens = sampled_token_ids.clone().flatten()
```
njhill, line 201: Why is the `clone()` needed here?

vllm/v1/sample/rejection_sampler.py
```diff
@@ -212,8 +218,8 @@ def parse_output(
         output_token_ids: torch.Tensor,
         vocab_size: int,
         discard_req_indices: Sequence[int] = (),
-        return_cu_num_tokens: bool = False,
-    ) -> tuple[list[list[int]], list[int] | None]:
+        logprobs_tensors: LogprobsTensors | None = None,
+    ) -> tuple[list[list[int]], list[int] | None, LogprobsLists | None]:
```
njhill: don't need to return `cu_num_tokens` anymore
```suggestion
    ) -> tuple[list[list[int]], LogprobsLists | None]:
```

vllm/v1/sample/rejection_sampler.py
```diff
@@ -232,14 +238,19 @@ def parse_output(
...
+            filtered_tensors = logprobs_tensors.filter(flat_mask)
+            output_logprobs = filtered_tensors.tolists(cu_num_tokens)
+
         if len(discard_req_indices) > 0:
             valid_mask[discard_req_indices] = False
         outputs = [
             row[valid_mask[i]].tolist() for i, row in enumerate(output_token_ids_np)
         ]
-        return outputs, cu_num_tokens
+        return outputs, cu_num_tokens, output_logprobs
```
njhill: can simplify a bit:
```python
        output_logprobs = None
        if logprobs_tensors is not None:
            cu_num_tokens = [0] + valid_mask.sum(axis=1).cumsum().tolist()
            filtered_tensors = logprobs_tensors.filter(valid_mask.flatten())
            output_logprobs = filtered_tensors.tolists(cu_num_tokens)
        if len(discard_req_indices) > 0:
            valid_mask[discard_req_indices] = False
        outputs = [
            row[valid_mask[i]].tolist() for i, row in enumerate(output_token_ids_np)
        ]
        return outputs, output_logprobs
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2794,13 +2802,18 @@ def _bookkeeping_sync(
...
-                    self.input_batch.vocab_size,
-                    discard_sampled_tokens_req_indices,
-                    return_cu_num_tokens=logprobs_tensors is not None,
+                valid_sampled_token_ids, cu_num_tokens, logprobs_lists = (
+                    RejectionSampler.parse_output(
+                        sampled_token_ids,
+                        self.input_batch.vocab_size,
+                        discard_sampled_tokens_req_indices,
+                        logprobs_tensors=logprobs_tensors,
+                    )
```
njhill:
```suggestion
                valid_sampled_token_ids, logprobs_lists = RejectionSampler.parse_output(
                    sampled_token_ids,
                    self.input_batch.vocab_size,
                    discard_sampled_tokens_req_indices,
                    logprobs_tensors=logprobs_tensors,
```

vllm/v1/worker/gpu_model_runner.py
```diff
@@ -2785,6 +2792,7 @@ def _bookkeeping_sync(
         logprobs_tensors = sampler_output.logprobs_tensors
         invalid_req_indices = []
         cu_num_tokens: list[int] | None = None
```
njhill:
```suggestion
```

njhill (commented):
Thanks again for the great work @izhuhaoran!

My comments are mainly simplifications/cleanup.

vllm/v1/sample/rejection_sampler.py
izhuhaoran: The clone() is needed because of the trick code `accepted_tokens[accepted_tokens == PLACEHOLDER_TOKEN_ID] = 0` where we replace rejected tokens (marked as PLACEHOLDER_TOKEN_ID = -1) with 0 to avoid a gather error in:
https://github.com/vllm-project/vllm/blob/d5ec6c056f2f806ea119b9129b07994f1d4fef4e/vllm/v1/sample/sampler.py#L237-L239

The gather operation requires valid indices in range [0, vocab_size). Without the clone, this replacement would modify the original `sampled_token_ids` tensor, causing `parse_output()` to incorrectly see rejected tokens as 0 instead of -1, which would break the filtering logic that depends on PLACEHOLDER_TOKEN_ID to identify rejected tokens.

njhill: Ah got it, missed that, thanks!

vllm/v1/sample/rejection_sampler.py
izhuhaoran: Thanks for the suggestion! However, I think the `add_()` in-place operation won't work here due to broadcasting constraints.

The issue is:
- `offsets.unsqueeze(0)` has shape `[1, M]`
- `logit_start_indices.unsqueeze(1)` has shape `[N, 1]`
- After broadcasting, the result would be `[N, M]`

The `add_()` in-place operation requires the left operand (the tensor being modified) to have a shape that can accommodate the result. Since `[1, M]` cannot store `[N, M]` in-place (unless N=1), PyTorch will raise a `RuntimeError`.

image

BTW, The original implementation is already fairly efficient:
```python
accepted_logit_indices = (
    logit_start_indices.unsqueeze(1) + offsets.unsqueeze(0)
).flatten()
```

Here, `unsqueeze` and `flatten()` only create views without allocating new memory, and the broadcast addition allocates just one `[N, M]` result tensor. The main way to further optimize would be fusing this into a Triton kernel, but that seems overkill for this operation.

njhill: OK, makes sense, sorry please ignore it was only a minor suggestion anyhow - thanks for all of the detail!

vllm/v1/sample/rejection_sampler.py
izhuhaoran: Thanks, done.

vllm/v1/worker/gpu_model_runner.py
izhuhaoran: done.

vllm/v1/sample/rejection_sampler.py
izhuhaoran: done.

vllm/v1/sample/rejection_sampler.py
izhuhaoran: done.

vllm/v1/worker/gpu_model_runner.py
izhuhaoran: done.

vllm/v1/worker/gpu_model_runner.py
izhuhaoran: done.

njhill (approved):
Thanks so much @izhuhaoran for working through the night to address the comments!

PR #31343, opened 2025-12-25
feat(frontend): add --default-chat-template-kwargs CLI argument

effortprogrammer:
Fixes #28070

## Purpose
Add server-level default chat_template_kwargs to control reasoning model behavior at deployment time. Request-level kwargs override these defaults. 

## Test Plan

This PR allows explicit control of reasoning/non-reasoning mode at the vllm serve command level using `--default-chat-template-kwargs`.

For reasoning models like Qwen3, you can now disable thinking mode server-wide by setting {"enable_thinking": false} as a default, eliminating the need to specify it in every request. Request-level chat_template_kwargs will override these server defaults when provided.

Manual test command:

```
vllm serve Qwen/Qwen3-8B --tensor-parallel-size 2 --served-model-name xionic-test --host 0.0.0.0 --port 8000
```

Minimal python code for test: 
```python
from openai import OpenAI
BASE_URL = "http://localhost:8000/v1"  # Change to your server
MODEL = "xionic-test"
client = OpenAI(api_key="EMPTY", base_url=BASE_URL)
# Same request to both servers
messages = [{"role": "user", "content": "What is 2+2?"}]
print("=" * 60)
print("WITHOUT --default-chat-template-kwargs (thinking enabled)")
print("=" * 60)
resp1 = client.chat.completions.create(model=MODEL, messages=messages, max_tokens=200)
print(resp1.choices[0].message.content)
print("\n" + "=" * 60)
print("WITH --default-chat-template-kwargs (thinking disabled)")
print("Or using client-side override (current workaround):")
print("=" * 60)
resp2 = client.chat.completions.create(
    model=MODEL,
    messages=messages,
    max_tokens=200,
    extra_body={"chat_template_kwargs": {"enable_thinking": False}},
)
print(resp2.choices[0].message.content)
print("\n" + "=" * 60)
print("SUMMARY")
print("=" * 60)
print(f"Response 1 length: {len(resp1.choices[0].message.content)} chars")
print(f"Response 2 length: {len(resp2.choices[0].message.content)} chars")
print(f"Has <think> tag in resp1: {'<think>' in resp1.choices[0].message.content}")
print(f"Has <think> tag in resp2: {'<think>' in resp2.choices[0].message.content}")
```

### WITHOUT --default-chat-template-kwargs (thinking enabled):

Result: <think>
Okay, the user is asking "What is 2+2?" That seems straightforward, but maybe they want a detailed explanation. Let me think. First, I should confirm the basic arithmetic. 2 plus 2 is 4. But maybe they're testing if I know the answer or if there's a trick. Sometimes people ask simple questions to see if the AI is reliable.

Wait, could there be a different interpretation? Like in some contexts, 2+2 might not be 4? For example, in modular arithmetic, if we're working modulo 3, 2+2 would be 1. But the question doesn't specify any context, so the default is standard arithmetic. 

Also, maybe they want to know the steps involved. Let me break it down. Starting with two units and adding another two units. So 2 + 2 equals 4. But perhaps they want a more detailed explanation, like using number lines or visual aids.

### WITH --default-chat-template-kwargs (thinking disabled):

Result: 2 + 2 equals 4.

<details>

</details>

### DIFF
--- a/tests/entrypoints/openai/test_cli_args.py
+++ b/tests/entrypoints/openai/test_cli_args.py
@@ -208,3 +208,38 @@ def test_middleware(serve_parser, cli_args, expected_middleware):
     """Ensure multiple middleware args are parsed properly"""
     args = serve_parser.parse_args(args=cli_args)
     assert args.middleware == expected_middleware
+
+
+def test_default_chat_template_kwargs_parsing(serve_parser):
+    """Ensure default_chat_template_kwargs JSON is parsed correctly"""
+    args = serve_parser.parse_args(
+        args=["--default-chat-template-kwargs", '{"enable_thinking": false}']
+    )
+    assert args.default_chat_template_kwargs == {"enable_thinking": False}
+
+
+def test_default_chat_template_kwargs_complex(serve_parser):
+    """Ensure complex default_chat_template_kwargs JSON is parsed correctly"""
+    kwargs_json = '{"enable_thinking": false, "custom_param": "value", "num": 42}'
+    args = serve_parser.parse_args(
+        args=["--default-chat-template-kwargs", kwargs_json]
+    )
+    assert args.default_chat_template_kwargs == {
+        "enable_thinking": False,
+        "custom_param": "value",
+        "num": 42,
+    }
+
+
+def test_default_chat_template_kwargs_default_none(serve_parser):
+    """Ensure default_chat_template_kwargs defaults to None"""
+    args = serve_parser.parse_args(args=[])
+    assert args.default_chat_template_kwargs is None
+
+
+def test_default_chat_template_kwargs_invalid_json(serve_parser):
+    """Ensure invalid JSON raises an error"""
+    with pytest.raises(SystemExit):
+        serve_parser.parse_args(
+            args=["--default-chat-template-kwargs", "not valid json"]
+        )
--- a/tests/entrypoints/openai/test_cli_args.py
+++ b/tests/entrypoints/openai/test_cli_args.py
@@ -221,9 +221,7 @@ def test_default_chat_template_kwargs_parsing(serve_parser):
 def test_default_chat_template_kwargs_complex(serve_parser):
     """Ensure complex default_chat_template_kwargs JSON is parsed correctly"""
     kwargs_json = '{"enable_thinking": false, "custom_param": "value", "num": 42}'
-    args = serve_parser.parse_args(
-        args=["--default-chat-template-kwargs", kwargs_json]
-    )
+    args = serve_parser.parse_args(args=["--default-chat-template-kwargs", kwargs_json])
     assert args.default_chat_template_kwargs == {
         "enable_thinking": False,
         "custom_param": "value",
--- a/vllm/entrypoints/openai/api_server.py
+++ b/vllm/entrypoints/openai/api_server.py
@@ -1079,6 +1079,7 @@ async def init_app_state(
             request_logger=request_logger,
             chat_template=resolved_chat_template,
             chat_template_content_format=args.chat_template_content_format,
+            default_chat_template_kwargs=args.default_chat_template_kwargs,
             trust_request_chat_template=args.trust_request_chat_template,
             return_tokens_as_token_ids=args.return_tokens_as_token_ids,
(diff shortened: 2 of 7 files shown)


vllm/entrypoints/openai/api_server.py
```diff
@@ -1089,6 +1089,7 @@ async def init_app_state(
             enable_force_include_usage=args.enable_force_include_usage,
             enable_log_outputs=args.enable_log_outputs,
             log_error_stack=args.log_error_stack,
+            default_chat_template_kwargs=args.default_chat_template_kwargs,
```
DarkLight1337: Can you move this line so it sticks with the other chat template CLI args?

vllm/entrypoints/openai/serving_engine.py
```diff
@@ -1183,6 +1184,8 @@ async def _preprocess_chat(
             tools=tool_dicts,
             documents=documents,
         )
+        if default_chat_template_kwargs:
+            _chat_template_kwargs.update(default_chat_template_kwargs)
```
chaunceyjiang, line 1188: 1. Why not merge `default_chat_template_kwargs` with `request.chat_template_kwargs` in advance?
2. When the same parameters are present, `request.chat_template_kwargs` should take higher priority.

effortprogrammer: Thanks for the review! @chaunceyjiang 
The current implementation already ensures that `request.chat_template_kwargs` takes higher priority:
```python
if default_chat_template_kwargs:
    _chat_template_kwargs.update(default_chat_template_kwargs)  # Apply server defaults first
_chat_template_kwargs.update(chat_template_kwargs or {})        # Then request kwargs override
```

Since dict.update() overwrites existing keys, request-level kwargs will always override server defaults when the same parameter is present.

For example:
- Server: --default-chat-template-kwargs '{"enable_thinking": false}'
- Request: chat_template_kwargs={"enable_thinking": true}
- Result: enable_thinking=true (request wins)

Regarding merging in advance: 

I kept it in `_preprocess_chat` to be consistent with how the existing `chat_template_kwargs` is handled, and to keep the merge logic in one place. 

Let me know if you'd prefer a different approach!

vllm/entrypoints/openai/api_server.py
effortprogrammer: Done [a821b47](https://github.com/vllm-project/vllm/pull/31343/commits/a821b47d928be54de8ae2e7b32efc1e391fd8a53)

vllm/entrypoints/openai/serving_engine.py
chaunceyjiang: OK. LGTM.

/cc @DarkLight1337 @hmellor PTAL.

DarkLight1337 (approved):

tests/entrypoints/openai/test_cli_args.py
```diff
@@ -208,3 +208,36 @@ def test_middleware(serve_parser, cli_args, expected_middleware):
     """Ensure multiple middleware args are parsed properly"""
     args = serve_parser.parse_args(args=cli_args)
     assert args.middleware == expected_middleware
+
+
+def test_default_chat_template_kwargs_parsing(serve_parser):
+    """Ensure default_chat_template_kwargs JSON is parsed correctly"""
+    args = serve_parser.parse_args(
+        args=["--default-chat-template-kwargs", '{"enable_thinking": false}']
```
chaunceyjiang, line 216: Hi @effortprogrammer Could you add this use case at the end of the `docs/features/reasoning_outputs.md` document?

It will help more people use this parameter, especially for models that support `enable_thinking`.

effortprogrammer: Done. [dda72c8](https://github.com/vllm-project/vllm/pull/31343/commits/dda72c86effea308dc7f83665b34aec43d7f52a3)

chaunceyjiang (approved):
Thanks~ @effortprogrammer

effortprogrammer:
@chaunceyjiang @DarkLight1337 It seems like current CI/CD failed does not relate with my current changes. Is there anything I should change for?

PR #18000, opened 2025-05-12
Use NVFP4 Marlin for CompressedTensorsW4A16Fp4

mgoin:
Update CT FP4 weight only path to use the nvfp4 dense marlin kernels from https://github.com/vllm-project/vllm/pull/17687 and remove slow emulation

PR #31346, opened 2025-12-25
[Bugfix] Remove spurious NVFP4 'GPU does not support FP4' warning

### LINKED ISSUE #27471: [Bug]: NVFP4A16 spurious warning that GPU doesn't support Fp4
### Your current environment

<details>

```text
vLLM API server version 0.11.1rc2.dev200+g250fb1b8e.d20251021
```

</details>

### 🐛 Describe the bug

When loading a NVFP4A16 quantized model we get a spurious warning

```
(Worker_TP0 pid=201) WARNING 10-24 09:50:14 [marlin_utils_fp4.py:137] Your GPU does not have native support for FP4 computation but FP4 quantization is being used. Weight-only FP4 compression will be used leveraging the Marlin kernel. This may degrade performance for compute-heavy workloads.
```

This is using 2x RTX Pro 6000, which do support Fp4.

That warning is coming from https://github.com/vllm-project/vllm/blob/3567816932e674abce3f44ceb0aff03f73b5aaff/vllm/model_executor/layers/quantization/utils/marlin_utils_fp4.py#L136-L142

I believe that warning was supposed to be removed when the emulation path for NVFP4A16 was removed in https://github.com/vllm-project/vllm/pull/18000. If so I can submit a simple PR to remove that part.

Otherwise does it mean there is a new marlin kernel planned that avoids software-dequantizing of NVFP4?

yurekami:
## Summary

Removes misleading warning from `prepare_fp4_layer_for_marlin` and `prepare_moe_fp4_layer_for_marlin` that incorrectly claims the GPU does not have native FP4 support.

## Problem

When loading NVFP4 quantized models, users see this warning even on GPUs with native FP4 support (like RTX Pro 6000):

```
WARNING marlin_utils_fp4.py:137 Your GPU does not have native support for FP4 computation but FP4 quantization is being used. Weight-only FP4 compression will be used leveraging the Marlin kernel. This may degrade performance for compute-heavy workloads.
```

This warning was added when there was a slow emulation path, but that path was removed in #18000. The Marlin kernel is now the standard efficient implementation for FP4 quantization, not a fallback for unsupported GPUs.

## Changes

- Remove spurious warning from `prepare_fp4_layer_for_marlin()`
- Remove spurious warning from `prepare_moe_fp4_layer_for_marlin()`

## Test Plan

- Verified linting passes
- The removed warning was purely informational and does not affect functionality

Fixes #27471

### DIFF
--- a/vllm/model_executor/layers/quantization/utils/marlin_utils_fp4.py
+++ b/vllm/model_executor/layers/quantization/utils/marlin_utils_fp4.py
@@ -146,13 +146,6 @@ def apply_fp4_marlin_linear(
 def prepare_fp4_layer_for_marlin(
     layer: torch.nn.Module, input_dtype: torch.dtype | None = None
 ) -> None:
-    logger.warning_once(
-        "Your GPU does not have native support for FP4 computation but "
-        "FP4 quantization is being used. Weight-only FP4 compression will "
-        "be used leveraging the Marlin kernel. This may degrade "
-        "performance for compute-heavy workloads."
-    )
-
     is_nvfp4 = hasattr(layer, "weight_scale_2")
     if input_dtype is not None and input_dtype.itemsize == 1:
         if is_nvfp4:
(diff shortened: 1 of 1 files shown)


mgoin:
Lucas is right, this message is intentional and ultimately this is still emulation as we aren't doing fp4 computation as expected

PR #31350, opened 2025-12-25
Add missing return type annotations to ray, distributed, and attention modules

yurekami:
## Purpose

This PR adds missing return type annotations to functions in `vllm/ray/lazy_utils.py`, `vllm/distributed/utils.py`, and `vllm/attention/utils/fa_utils.py` to improve code quality and IDE support.

### vllm/ray/lazy_utils.py

| Function | Return Type Added |
|----------|------------------|
| `is_ray_initialized()` | `bool` |
| `is_in_ray_actor()` | `bool` |

### vllm/distributed/utils.py

| Function | Return Type Added |
|----------|------------------|
| `sched_yield()` | `None` |
| `ensure_divisibility()` | `None` (also added parameter types: `int, int`) |
| `divide()` | `int` (also added parameter types: `int, int`) |

### vllm/attention/utils/fa_utils.py

| Function | Return Type Added |
|----------|------------------|
| `flash_attn_supports_mla()` | `bool` |

## Test Plan

This is a typing-only change with no runtime impact. All existing tests should pass.

### DIFF
--- a/vllm/model_executor/layers/quantization/utils/marlin_utils_fp4.py
+++ b/vllm/model_executor/layers/quantization/utils/marlin_utils_fp4.py
@@ -146,13 +146,6 @@ def apply_fp4_marlin_linear(
 def prepare_fp4_layer_for_marlin(
     layer: torch.nn.Module, input_dtype: torch.dtype | None = None
 ) -> None:
-    logger.warning_once(
-        "Your GPU does not have native support for FP4 computation but "
-        "FP4 quantization is being used. Weight-only FP4 compression will "
-        "be used leveraging the Marlin kernel. This may degrade "
-        "performance for compute-heavy workloads."
-    )
-
     is_nvfp4 = hasattr(layer, "weight_scale_2")
     if input_dtype is not None and input_dtype.itemsize == 1:
         if is_nvfp4:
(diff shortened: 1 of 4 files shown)


vllm/model_executor/layers/quantization/utils/marlin_utils_fp4.py
```diff
@@ -146,13 +146,6 @@ def apply_fp4_marlin_linear(
 def prepare_fp4_layer_for_marlin(
     layer: torch.nn.Module, input_dtype: torch.dtype | None = None
 ) -> None:
-    logger.warning_once(
```
yewentao256, line 149: Why do we want to remove this warning?

vllm/model_executor/layers/quantization/utils/marlin_utils_fp4.py
```diff
@@ -229,13 +222,6 @@ def prepare_fp4_layer_for_marlin(
 def prepare_moe_fp4_layer_for_marlin(
     layer: torch.nn.Module, input_dtype: torch.dtype | None = None
 ) -> None:
-    logger.warning_once(
```
yewentao256, line 232: Same here

yewentao256 (commented):
Thanks for the work!

LucasWilkinson:
Please consolidate all `add return type annotations` PRs into a single one larger one; no need to have such small PRs

PR #31366, opened 2025-12-25
feat(frontend): early-fail tokenization guard for user requests

scratch-ml:
## Purpose
To advance the early-fail truncation design proposed in [vllm#31229](https://github.com/vllm-project/vllm/issues/31229), these tests ensure default tokenization applies protective truncation and raises Exception immediately on overlength inputs, preventing the async tokenizer from being monopolized.

Default behavior: tokenization runs with protective truncation (`truncation=True`, `max_length=max_model_len+1`); if the encoded length hits the limit, it raises `ValueError `immediately to avoid monopolizing the async tokenizer on huge prompts. To override this, callers can explicitly set truncate_prompt_tokens=0 (or provide their own `truncation=False`/custom `max_length`), but then they are responsible for ensuring the input length won’t cause OOMs or stalls.

FIX #31229

## Test Plan
~~Unit: `pytest tests/entrypoints/openai/test_serving_engine_truncation.py -q` to verify default/disabled/custom truncation and the overlength prompt error path.~~
~~Unit: `pytest tests/inputs/test_preprocess_truncation.py -q `to verify tokenizer-side truncation behavior (_truncate_inputs) for both left/right truncation.~~

Updated `tests/renderers/test_completions.py` with new tests.

## Test Result
All added tests pass: ~~both suites complete successfully (4 passed in truncation serving tests; 5 passed in preprocess truncation tests).~~

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit ca965e7a53f2e90379c3e17fcb520d60bd695d6b. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Strengthens prompt length handling and safety across serving, renderer, and inputs.
> 
> - Introduces `_validate_text_prompt_char_length` with `VLLM_MAX_CHARS_PER_TOKEN` override; applied in OpenAI serving and renderer for early-fail on extreme text inputs
> - Updates `_validate_truncation_size` to always set `truncation=True` and `max_length` (`max_model_len+1` by default; negatives map to model max); raises if user value exceeds model limit
> - Refactors serving validation via `_validate_length_only` and avoids duplicate checks/decodes; minor error message cleanups
> - Renderer and `InputPreprocessor` now apply default protective truncation and raise on overlength prompts; add helpers (`apply_default_truncation`) and adjust tokenization kwargs flow
> - Adds unit tests for serving, renderer, and input preprocessing truncation behaviors; updates expectations for default truncation
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ca965e7a53f2e90379c3e17fcb520d60bd695d6b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit bbf0b0bfc84a059a447bd4f19094c6bd4bef34b2. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> Strengthens prompt length handling and safety across `serving_engine`, `renderer`, and inputs.
> 
> - Introduces `_validate_text_prompt_char_length` with `VLLM_MAX_CHARS_PER_TOKEN` (new env var) and applies it before tokenization in OpenAI serving and renderer
> - Updates `_validate_truncation_size` to always set `truncation=True` with `max_length=max_model_len+1` by default; negatives map to model max; raises on values > model max
> - Renderer and `InputPreprocessor` now apply default protective truncation; add `apply_default_truncation` and raise on overlength prompts when defaults are used
> - Refactors serving validation via `_validate_length_only`, avoids duplicate checks/decodes, and tweaks error messages
> - Adds unit tests covering default/explicit truncation and char-length guard; adjusts renderer tests to expect truncation kwargs
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bbf0b0bfc84a059a447bd4f19094c6bd4bef34b2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


vllm/inputs/preprocess.py
```diff
@@ -221,12 +239,25 @@ def _tokenize_prompt(
         tokenizer = self.get_tokenizer()
         tokenization_kwargs = self._get_tokenization_kw(tokenization_kwargs)
 
+        tokenization_kwargs, has_user_truncation = apply_default_truncation(
+            tokenization_kwargs, self.model_config
+        )
```
njhill: Better to put this logic in the existing `_get_tokenization_kw` method

vllm/inputs/preprocess.py
```diff
@@ -221,12 +239,25 @@ def _tokenize_prompt(
...
 
-        return tokenizer.encode(prompt, **tokenization_kwargs)
+        token_ids = tokenizer.encode(prompt, **tokenization_kwargs)
+
+        if not has_user_truncation and len(token_ids) > self.model_config.max_model_len:
+            raise ValueError(
+                f"This model's maximum context length is "
+                f"{self.model_config.max_model_len} tokens. The input text is too "
+                "long. Please shorten it or set `truncate_prompt_tokens`."
+            )
```
njhill: Suggest to reword the message:

"The tokenized prompt contains more than X tokens which is the maximum context length. Please shorten it or set `truncate_prompt_tokens`."

vllm/inputs/preprocess.py
```diff
@@ -42,6 +42,24 @@
...
+    has_user_truncation = (
+        "truncation" in tokenization_kwargs or "max_length" in tokenization_kwargs
+    )
+    if not has_user_truncation:
+        tokenization_kwargs["truncation"] = True
+        tokenization_kwargs["max_length"] = model_config.max_model_len + 1
+    elif tokenization_kwargs.get("truncation") and "max_length" not in (
+        tokenization_kwargs
+    ):
+        tokenization_kwargs["max_length"] = model_config.max_model_len
```
njhill: Suggest to force truncation in all cases:
```python
tokenization_kwargs["truncation"] = True
max_length = tokenization_kwargs.get("max_length")
if max_length is None or max_length > model_config.max_model_len:
   tokenization_kwargs["max_length"] = model_config.max_model_len + 1
```

vllm/entrypoints/openai/serving_engine.py
```diff
@@ -953,26 +954,32 @@ async def _normalize_prompt_text_to_input(
             prompt = prompt.lower()
 
         truncate_prompt_tokens = getattr(request, "truncate_prompt_tokens", None)
+        # When explicitly set to 0 or False, honor the caller and skip truncation
+        disable_truncation = (
+            truncate_prompt_tokens == 0 or truncate_prompt_tokens is False
+        )
```
njhill: Suggest to apply truncation in all cases here. It should not be possible to disable.

vllm/entrypoints/openai/serving_engine.py
```diff
@@ -994,6 +1001,9 @@ async def _normalize_prompt_tokens_to_input(
         else:
             input_ids = prompt_ids[-truncate_prompt_tokens:]
 
+        # Validate lengths before any decode to avoid extra work on very long inputs.
+        self._validate_length_only(request, len(input_ids))
```
njhill: If the validation is being called here, it doesn't need to be called again below (at end of this method)

njhill (changes requested):
Thanks @scratch-ml! I agree this is definitely something we should have.

We also need to apply this logic in the renderer: https://github.com/vllm-project/vllm/blob/b12cb38398a57a4abc4b8fb120fd39e97029ffa1/vllm/entrypoints/renderer.py#L44

Side node: we really need to fix the duplicated logic but not for this PR!

Also, we could consider having an additional guard for text prompts before even attempting tokenization, if the string length is > `(max_model_len * MAX_CHARS_PER_TOKEN)` (unless truncate_prompt_tokens == -1).

njhill:
We should also ideally unify with the logic here: https://github.com/vllm-project/vllm/blob/d63b969675f1e655ad318b629d9eb1516705a8c4/vllm/entrypoints/utils.py#L179

scratch-ml:
@njhill Thank you for your suggestions. I’ll carve out some time to make fixes and improvements

vllm/renderers/params.py
```diff
@@ -230,22 +231,56 @@ def get_encode_kwargs(self) -> dict[str, Any]:
         if max_length is not None and max_length < 0:
             max_length = self.max_input_tokens
 
+        # This prevents tokenization from taking up more resources than necessary
+        # while still failing `self._token_len_check` as expected by users
+        if max_length is None and self.max_input_tokens is not None:
```
njhill:
```suggestion
        elif max_length is None and self.max_input_tokens is not None:
```

vllm/renderers/params.py
```diff
@@ -230,22 +231,56 @@ def get_encode_kwargs(self) -> dict[str, Any]:
...
+                )
+
+        return text
+
+    def _text_lowercase(self, tokenizer: TokenizerLike | None, text: str) -> str:
+        """Apply lowercase to prompt text if necessary."""
         if self.do_lower_case:
             text = text.lower()
 
         return text
```
njhill: nit: simplify
```suggestion
        return text.lower() if self.do_lower_case else text
```

vllm/renderers/params.py
njhill: In what case would `self.max_input_tokens` be `None`? This should always be the model's max context length right? (possibly one less). Is this case just for a tokenization-only endpoint?

njhill: Also I see above that `self.max_input_tokens` defaults to `self.max_total_tokens - self.max_output_tokens` which looks wrong to me. This is true for a given request but not for the maximums (e,g, if max total tokens is 1024, both max input and output should be like 1023?)

vllm/renderers/params.py
```diff
@@ -230,22 +231,56 @@ def get_encode_kwargs(self) -> dict[str, Any]:
...
 
     def _validate_text(self, tokenizer: TokenizerLike | None, text: str) -> str:
         """Apply all validators to prompt text."""
-        # TODO: Implement https://github.com/vllm-project/vllm/pull/31366
-        for validator in (self._apply_lowercase,):
+        for validator in (
+            self._text_len_check,
+            self._text_lowercase,
```
njhill, line 277: This is modifying the text, not just validating?

njhill (commented):
Thanks @DarkLight1337

vllm/renderers/params.py
DarkLight1337: Yes, it is for tokenization-only endpoints

vllm/renderers/params.py
DarkLight1337: Yes, I still called it "validator" because Pydantic validators work in the same way

vllm/renderers/params.py
DarkLight1337: Prior to Renderer refactor, the max allowed input length was considered to be `model_config.max_model_len - request.max_tokens`. I agree the naming is a bit confusing, but `TokenizeParams` is created for each request so `max_input_tokens` indeed only refers to that request.

DarkLight1337: Done

vllm/renderers/params.py
DarkLight1337: Done

vllm/envs.py
```diff
@@ -1584,6 +1585,11 @@ def _get_or_set_default() -> str:
     # Disable PDL for LoRA, as enabling PDL with LoRA on SM100 causes
     # Triton compilation to fail.
     "VLLM_LORA_DISABLE_PDL": lambda: bool(int(os.getenv("VLLM_LORA_DISABLE_PDL", "0"))),
+    # Upper bound for characters per token used by the early-fail prompt length
+    # guard before tokenization.
+    "VLLM_MAX_CHARS_PER_TOKEN": lambda: int(
+        os.getenv("VLLM_MAX_CHARS_PER_TOKEN", "32")
+    ),
```
njhill: Can we initialize this from the tokenizer to avoid a new env var?

vllm/renderers/params.py
```diff
@@ -229,23 +230,53 @@ def get_encode_kwargs(self) -> dict[str, Any]:
         max_length = self.truncate_prompt_tokens
         if max_length is not None and max_length < 0:
             max_length = self.max_input_tokens
+        elif max_length is None and self.max_input_tokens is not None:
```
njhill, line 232: I think we should also cap to `self.max_input_tokens + 1` if `truncate_prompt_tokens` is provided?

DarkLight1337: This is unecessary because we already validate that `self.truncate_prompt_tokens <= self.max_input_tokens` in `__post_init__`

vllm/envs.py
DarkLight1337: Added a new pre-computed property `max_chars_per_token` to `TokenizerLike` interface

njhill: I guess we need to add to `DeepseekV32Tokenizer` too until that other PR is merged?

njhill (approved):

PR #22204, opened 2025-08-04
Optimize configuration access with LRU cache in custom ops

skyloevil:
Changes

  - Added get_cached_compilation_config() function with
  @lru_cache(maxsize=1) to cache compilation config access
  - Added cache invalidation in set_current_vllm_config() context
  manager to ensure cache freshness
  - Fixed code style (line length) issues

  Benefits

  - Reduces repeated configuration lookups in custom operations
  - Improves performance for frequently accessed compilation config
  - Maintains correctness with proper cache invalidation

  Files Modified

  - vllm/config.py: Added caching function and invalidation logic
  - vllm/model_executor/custom_op.py: Updated to use cached config access

PR #31376, opened 2025-12-26
[BugFix] Fix cache issue in compilation_config

BoyuanFeng:
#22204 added `get_cached_compilation_config()` with `@lru_cache(maxsize=1)` to cache compilation config access via
```python
@lru_cache(maxsize=1)
def get_cached_compilation_config():
    """Cache config to avoid repeated calls to get_current_vllm_config()"""
    return get_current_vllm_config().compilation_config
```

Since `get_cached_compilation_config()` has no argument, it would never be refreshed, unless we explicitly call `get_cached_compilation_config.cache_clear()`.

To refresh the config, #22204 implements `get_cached_compilation_config.cache_clear()` in `set_current_vllm_config`.

```python
@contextmanager
def set_current_vllm_config(vllm_config: VllmConfig, check_compile=False, prefix: Optional[str] = None):
    try:
        _current_vllm_config = vllm_config
        _current_prefix = prefix
        yield
   finally:
        _current_vllm_config = old_vllm_config
        _current_prefix = old_prefix
        # Clear the compilation config cache when context changes
        get_cached_compilation_config.cache_clear() # <----- explicitly clear the cache
```

However, a bug would happen if an `old_vllm_config` has been accessed and cached before `set_current_vllm_config(new_config)` is called. In other words

```python
config = old_config
with set_current_vllm_config(new_config):
    # old_config is accessed instead of new_config
```

This PR fixes the issue by clearing the cache when entering the context manager.


yewentao256 (commented):
LGTM, thanks for the work!

Is there any specific issue this PR could solve? (If there is a case that without this PR would cause trouble?)

BoyuanFeng:
@yewentao256 I added a unit test `test_cached_compilation_config` to show the issue.
tldr: prior to this pr, if I call `old_config = get_cached_compilation_config` first, and then call `with set_current_vllm_config(vllm_config):`, the `old_config.compilation_config` is used within the context manager instead of the new `vllm_config.compilation_config`.

yewentao256 (approved):
LGTM, thanks for the work!

ProExpertProg (approved):

yewentao256 (commented):
Merge this PR as all CI tests pass

PR #31386, opened 2025-12-26
[GLM-4.7] GLM Model support for GLM-Lite

zRzRzRzRzRzRzR:
using with transformers 5.0.0 with GLM-Lite model， transformers PR [here](https://github.com/huggingface/transformers/pull/43031)


vllm/model_executor/models/glm4_moe_lite.py
```diff
@@ -0,0 +1,874 @@
...
+                            param, "weight_loader", default_weight_loader
+                        )
+                        weight_loader(param, loaded_weight)
+            if not is_fusion_moe_shared_experts_layer:
+                loaded_params.add(name)
+
+        return loaded_params
+
+
+class Glm4MixtureOfExperts(MixtureOfExperts):
```
DarkLight1337: Maybe we can just modify the one inside `glm4_moe.py` and import that instead?

DarkLight1337: Otherwise we should rename this to `Glm4LiteMixtureOfExperts` to avoid confusion

vllm/model_executor/models/glm4_moe_lite.py
```diff
@@ -0,0 +1,874 @@
...
+        self.act_fn = SiluAndMul()
+
+    def forward(self, x):
+        gate_up, _ = self.gate_up_proj(x)
+        x = self.act_fn(gate_up)
+        x, _ = self.down_proj(x)
+        return x
+
+
+class Glm4MoeLite(nn.Module):
```
DarkLight1337: This looks very similar to the one from DeepSeek V2, what is the difference?

zRzRzRzRzRzRzR: Only the forward method uses Glm4MoeLiteMLP, but actually Glm4MoeLiteMLP can directly use deepseek MLP as well. vLLM doesn't have a modular structure like transformers, so it's either all inheritance or all independent implementation. What do you think?

DarkLight1337: We can add a `layer_type` argument to this module to reuse the implementation (refer to other similar models)

vllm/model_executor/models/glm4_moe_lite.py
zRzRzRzRzRzRzR: yes

zRzRzRzRzRzRzR:
@DarkLight1337 I made some modifications, is the implementation roughly like this?.

zRzRzRzRzRzRzR:
@DarkLight1337 now can review

tests/models/registry.py
```diff
@@ -271,6 +271,7 @@ def check_available_online(
     "GlmForCausalLM": _HfExamplesInfo("zai-org/glm-4-9b-chat-hf"),
     "Glm4ForCausalLM": _HfExamplesInfo("zai-org/GLM-4-9B-0414"),
     "Glm4MoeForCausalLM": _HfExamplesInfo("zai-org/GLM-4.5"),
+    "Glm4MoeLiteForCausalLM": _HfExamplesInfo("zai-org/GLM-4.7-Flash"),
```
DarkLight1337: The model isn't public yet so maybe you need to mark this `is_available_online=False` for now. Or do you prefer to merge the PR after model release?

zRzRzRzRzRzRzR: update

DarkLight1337: Update this as well

DarkLight1337 (approved):

PR #31388, opened 2025-12-26
[Voxtral] Fix speech transcription api

patrickvonplaten:
Make sure speech transcription api works.

Allows to do standard vllm serve + speech-to-text transcription API


vllm/model_executor/models/voxtral_streaming.py
```diff
@@ -205,39 +215,90 @@ def embed_multimodal(
...
+        to_language: str | None,
+    ) -> PromptType:
+        tokenizer = cached_tokenizer_from_config(model_config)
+        audio = Audio(audio, int(stt_config.sample_rate), format="wav")  # lossless
+
+        req = TranscriptionRequest(
+            model=model_config.model,
+            audio=RawAudio.from_audio(audio),
+            language=language,
+            streaming=StreamingMode.OFFLINE,
```
NickLucche, line 297: qq: is offline the right setting here?

vllm/entrypoints/openai/speech_to_text.py
```diff
@@ -477,15 +477,23 @@ async def _create_speech_to_text(
...
             text = ""
+            chunk_size_in_s = self.asr_config.max_audio_clip_s
             for idx, result_generator in enumerate(list_result_generator):
+                start_time: float
+                if chunk_size_in_s is not None:
+                    start_time = float(idx * chunk_size_in_s)
+                else:
+                    assert idx == 0, f"Cannot chunk audio input for {chunk_size_in_s=}"
+                    start_time = 0.0
+
```
NickLucche: I think it's clearer if we move this check outside the loop
```suggestion
            chunk_size_in_s = self.asr_config.max_audio_clip_s
            if chunk_size_in_s is None:
                assert len(list_result_generator) == 1, f"`max_audio_clip_s` is set to None, audio cannot be chunked"
            for idx, result_generator in enumerate(list_result_generator):
                start_time = float(idx * chunk_size_in_s) if chunk_size_in_s is not None else 0.0
```

vllm/config/speech_to_text.py
```diff
@@ -17,7 +17,7 @@ class SpeechToTextConfig:
     16kHz audio input. The input audio will be automatically resampled to this
     rate before processing."""
 
-    max_audio_clip_s: int = 30
+    max_audio_clip_s: int | None = 30
     """Maximum duration in seconds for a single audio clip without chunking.
     Audio longer than this will be split into smaller chunks if
     `allow_audio_chunking` evaluates to True, otherwise it will be rejected."""
```
NickLucche: we should mention None is no limit

NickLucche (commented):
Thanks @patrickvonplaten !

vllm/model_executor/models/voxtral_streaming.py
patrickvonplaten: Yeah it's still quite difficult to test actual "online" streaming given that this PR: https://github.com/vllm-project/vllm/pull/28973 is not yet merged - so for now I've only tested "offline" streaming which means that whole audio file is passed in one go.

Once https://github.com/vllm-project/vllm/pull/28973 is merged we however will need to slightly modify the methods here I think but maybe it's ok to do this after and for now just ensure "offline" streaming, a.k.a pass the whole audio file in one-go, works as expected?

vllm/model_executor/models/whisper.py
```diff
@@ -536,8 +538,11 @@ def forward_conv(
             hidden_states.append(embeds)
             input_is_batched = embeds.ndim > 2
         # Input to MHA must be B x T x D
-        if input_is_batched:
+        if input_is_batched or self.is_causal:
             # Models using WhisperEncoder may handle batching internally.
+            # If WhisperEncoder is causal sequences
```
patrickvonplaten:
```suggestion
            # If WhisperEncoder is causal, sequences
```

vllm/config/speech_to_text.py
patrickvonplaten:
```suggestion
    `allow_audio_chunking` evaluates to True, otherwise it will be rejected. 
    `None` means audio duration can be unlimited and won't be chunked."""
```

NickLucche (approved):

PR #31393, opened 2025-12-26
[CI] Replace `pre-commit` with `prek` for faster pre-commit checks

j178:
Signed-off-by: Jo <10510431+j178@users.noreply.github.com>

## Purpose

Replace [`pre-commit`](https://github.com/pre-commit/pre-commit) with [`prek`](https://github.com/j178/prek) for faster pre-commit checks.

On my Mac, this change reduced:

- **Cold run**  from **~55s** to **~17s**
- **Warm-cache** run from **~12s** to **~7s**

## Test Plan

Cold run (no cache)

```console
❯ hyperfine --warmup 2 --runs 5 \
        --prepare "prek cache clean" "prek run --all-files" \
        --prepare "pre-commit clean" "pre-commit run --all-files"
Benchmark 1: prek run --all-files
  Time (mean ± σ):     17.527 s ±  0.863 s    [User: 24.755 s, System: 18.198 s]
  Range (min … max):   16.270 s … 18.380 s    5 runs

Benchmark 2: pre-commit run --all-files
  Time (mean ± σ):     55.452 s ±  3.147 s    [User: 36.584 s, System: 21.337 s]
  Range (min … max):   51.810 s … 60.216 s    5 runs

Summary
  prek run --all-files ran
    3.16 ± 0.24 times faster than pre-commit run --all-files
```

Warm cache

```console
> hyperfine 'prek run -a' 'pre-commit run -a' --warmup 3
Benchmark 1: prek run -a
  Time (mean ± σ):      7.080 s ±  0.342 s    [User: 18.901 s, System: 14.657 s]
  Range (min … max):    6.696 s …  7.959 s    10 runs

Benchmark 2: pre-commit run -a
  Time (mean ± σ):     12.767 s ±  0.251 s    [User: 18.367 s, System: 13.410 s]
  Range (min … max):   12.310 s … 13.084 s    10 runs

Summary
  prek run -a ran
    1.80 ± 0.09 times faster than pre-commit run -a
```

## Background: What is prek?

prek is a reimagined version of pre-commit, built in Rust. It is designed to be a faster, dependency-free and drop-in alternative for it, while also providing some additional long-requested features.

prek has already been adopted by projects like [Apache Airflow](https://github.com/apache/airflow/issues/44995) and [FastAPI](https://github.com/fastapi/fastapi/pull/14572) (see more examples [here](https://github.com/j178/prek/?tab=readme-ov-file#who-is-using-prek)). CPython is also discussing adpotion: https://github.com/python/cpython/issues/143148

## Changes in this PR

- Updated the contributor documentation and development requirements to use prek.
- Updated the `pre-commit.yml` CI workflow to run prek instead of pre-commit.
- Added `priority: 1` to the slowest hooks (typos, pip-compile, mypy-local, png-lint, update-dockerfile-graph) so they can run in parallel, reducing end-to-end wall time.

Note: [priority](https://prek.j178.dev/configuration/#priority) is a prek feature (not supported by pre-commit). If you run pre-commit with this config, it will warn about “Unexpected keys”, but the hooks will still run; the warning is safe to ignore.

---
<details>

</details>

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -26,14 +26,14 @@ pull_request_rules:
         Hi @{{author}}, the pre-commit checks have failed. Please run:
 
         ```bash 
-        uv pip install pre-commit
-        pre-commit install
-        pre-commit run --all-files
+        uv pip install prek
+        prek install
+        prek run --all-files
         ```
 
         Then, commit the changes and push to your branch.
 
-        For future commits, `pre-commit` will run automatically on changed files before each commit.
+        For future commits, `prek` will run automatically on changed files before each commit.
 
         > [!TIP]
         > <details>
@@ -43,9 +43,9 @@ pull_request_rules:
         >
         > ```bash
         > # For mypy (substitute "3.10" with the failing version if needed)
-        > pre-commit run --hook-stage manual mypy-3.10
+        > prek run --hook-stage manual mypy-3.10
         > # For markdownlint
-        > pre-commit run --hook-stage manual markdownlint
+        > prek run --hook-stage manual markdownlint
         > ```
         > </details>
 
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -23,6 +23,6 @@ jobs:
     - run: echo "::add-matcher::.github/workflows/matchers/actionlint.json"
     - run: echo "::add-matcher::.github/workflows/matchers/markdownlint.json"
     - run: echo "::add-matcher::.github/workflows/matchers/mypy.json"
-    - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
+    - uses: j178/prek-action@91fd7d7cf70ae1dee9f4f44e7dfa5d1073fe6623 # v1.0.11
       with:
         extra_args: --all-files --hook-stage manual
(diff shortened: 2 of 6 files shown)


hmellor:
The cold start comparison is effectively comparing `uv pip` install speed with `pip` install speed. While it is a noticeable difference, this is a one time cost for local installs and a one time cost for CI (the python environment is cached).

As for the warm comparison, most of the speed difference comes from doing `mypy` in parallel with the rest of the hooks. `mypy` is very slow but will soon be replaced with `ty`, eliminating this issue. 

Also, it's not safe to parallelise many pre-commit hooks because they change files. i.e. `ruff` might make a change that fixes something that `mypy` would have failed. If you run them in parallel, `ruff` would fix the issue and fail, and `mypy` would fail with an error of the old state of the code. Which would be confusing.

hmellor:
> `prek` clones and installs hook envs in parallel

Good point, I hadn't considered that.

Regarding the parallelisation safety, it's not something that we have to think about when reviewing formatting/checking PRs, but switching to `prek` would mean we'd have to start.

Don't get me wrong, `prek` looks like a very cool project. I just don't think the potential issues justify switching to it for vLLM. Especially since `run -a` is very rarely run locally by users.

j178:
Thanks! I’m closing this. It was a good experiment anyway. And yeah, vLLM has a great contributing guide, btw, it made me getting started very easy.

PR #31397, opened 2025-12-27
implements register kv caches in lmcache connector

chunxiaozheng:
implements register kv caches in lmcache connector


chunxiaozheng:
@ApostaC @KuntaiDu could you help take a look?

ApostaC (approved):
LGTM!

PR #31411, opened 2025-12-27
[Core] Sort blocks by block_id in FreeKVCacheBlockQueue.append_n

majiayu000:
## Summary

Sort blocks by `block_id` when returning them to the free list in `FreeKVCacheBlockQueue.append_n` to enable contiguous allocation.

This improves I/O efficiency for layer-wise KV cache transfers by allowing multiple blocks to be merged into single operations. When blocks are allocated in ascending order (e.g., `[1, 2, 3, 4, 5]`), we can merge multiple I/O submissions into a single contiguous transfer, rather than requiring separate I/O submissions for each block.

## Changes

- Added `blocks.sort(key=lambda x: x.block_id)` before inserting blocks back into the free list
- Added test `test_free_kv_cache_block_queue_append_n_sorts_by_block_id` to verify the sorting behavior

## Impact

- Minimal overhead (sorting a small list of freed blocks)
- Significant potential benefit for systems performing batched or layer-wise KV cache transfers

Fixes #31371

PR #31412, opened 2025-12-27
[Core] Sort block IDs at I/O layer for contiguous memory access

### LINKED ISSUE #31371: [Feature]: Sort blocks by block_id in FreeKVCacheBlockQueue.append_n to enable contiguous allocation
### The feature, motivation and pitch

**Description:**

We are working on a layer-wise KV cache transfer system and have observed that maintaining forward-contiguous block allocation significantly improves I/O efficiency.

**The Problem:**

When blocks are allocated in ascending order (e.g., `[1, 2, 3, 4, 5]`), we can merge multiple I/O submissions into a single contiguous transfer (`[1-5]`). However, if blocks are allocated in arbitrary or reverse order (e.g., `[5, 4, 3, 2, 1]`), each block requires a separate I/O submission.

For long requests, this difference is substantial—potentially thousands of I/O submissions (number of blocks × number of layers) versus a handful of merged operations.

**Proposed Solution:**

Add a simple sort when returning blocks to the free list in `FreeKVCacheBlockQueue.append_n`:

```python
def append_n(self, blocks: list[KVCacheBlock]) -> None:
    """Put a list of blocks back into the free list."""
    if len(blocks) == 0:
        return

    blocks.sort(key=lambda x: x.block_id)  # <-- Add this line
    last_block = self.fake_free_list_tail.prev_free_block
    ...
```

**File location:** `vllm/v1/core/kv_cache_utils.py`

This ensures that when blocks are subsequently allocated, they tend to be contiguous, which benefits any downstream system that can exploit spatial locality in KV cache operations.

**Impact:**

- Minimal overhead (sorting a small list of freed blocks)
- Significant potential benefit for systems performing batched or layer-wise KV cache transfers

### Alternatives

_No response_

### Additional context

_No response_

majiayu000:
## Summary

Sort `src_to_dst` mapping by source block ID in `SingleDirectionOffloadingHandler.transfer_async()` to enable coalesced memory operations during KV cache transfers.

This is a redesign of #31411 based on reviewer feedback. The key insight is that sorting should happen at the I/O layer (after allocation, before transfer) rather than in the free list, which preserves the LRU eviction semantics while still achieving I/O efficiency benefits.

## Changes

- Added sorting by source block ID in `transfer_async()` before creating the `src_to_dst_tensor`
- Added unit test `test_expand_block_ids_with_sorting` to verify sorting behavior

## Why This Approach

The previous approach (#31411) sorted blocks in `FreeKVCacheBlockQueue.append_n()`, which:
1. Broke LRU eviction semantics
2. Only sorted incoming batches, not the entire queue

This new approach sorts at the I/O layer:
1. Preserves cache eviction policy completely
2. Directly benefits CUDA memory coalescing where it matters
3. Minimal overhead (sorting a small array before each transfer)

## Impact

- **No change** to block allocation or eviction semantics
- **Improved I/O efficiency** when blocks are accessed in sorted order by the CUDA kernel
- **Minimal overhead**: O(n log n) sort on the transfer mapping array

Fixes #31371

## Test plan

- [ ] Existing `test_transfer` tests verify data correctness after sorting
- [ ] New `test_expand_block_ids_with_sorting` verifies sorting logic

### DIFF
--- a/tests/v1/kv_offload/test_cpu_gpu.py
+++ b/tests/v1/kv_offload/test_cpu_gpu.py
@@ -189,3 +189,42 @@ def test_transfer(
                 torch.testing.assert_close(
                     dst_cache[dst_block].cpu(), expected_value.cpu()
                 )
+
+
+def test_expand_block_ids_with_sorting():
+    """Test that block ID expansion and sorting produces correct results.
+
+    This verifies that sorting src_to_dst by source block ID works correctly
+    with expand_block_ids, enabling contiguous memory access patterns.
+    """
+    import numpy as np
+
+    from vllm.v1.kv_offload.worker.cpu_gpu import expand_block_ids
+
+    # Test case: unsorted block IDs [3, 1, 2] with block_size_factor=2
+    src_blocks = np.array([3, 1, 2], dtype=np.int64)
+    dst_blocks = np.array([0, 1, 2], dtype=np.int64)
+    src_block_size_factor = 2
+    dst_block_size_factor = 2
+
+    dst_sub_block_count = dst_blocks.size * dst_block_size_factor
+
+    src_to_dst = np.empty((dst_sub_block_count, 2), dtype=np.int64)
+    expand_block_ids(src_blocks, src_block_size_factor, src_to_dst[:, 0])
+    expand_block_ids(dst_blocks, dst_block_size_factor, src_to_dst[:, 1])
+
+    # Before sorting: src blocks are [6,7, 2,3, 4,5] (from [3,1,2] * 2)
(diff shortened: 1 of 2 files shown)


orozery:
@majiayu000 thank you for your suggestion!
However, I think that this optimization will more likely hurt performance, for several reasons:
* I could not find any source that suggests that `cudaMemcpyAsync` calls can be coalesced automatically by the hardware.
* If you want to manually coalesce copies, this will only be possible if both the CPU and CPU block IDs are contiguous. The probability for that to happen is very very low.
* So basically the main effect here is adding of the sorting time, with no practical gains I see.

cc @robertgshaw2-redhat

PR #31415, opened 2025-12-27
[MoE Refactor][15/N] Apply Refactor to Fp8

robertgshaw2-redhat:
## Purpose

SUMMARY:
* continue refactoring fp8 moe
* CUTLASS MoE --> make CutlassExpertsFp8 manage the strides data
* CUTLASS MoE --> remove cutlass_moe_fp8 entrypoint
* factor out MoE kernel selection into a new module
* factor out process_weights_after_loading utilities to be shared by fp8 and ct
* apply mk refactor to ct

## Test Plan

- unit tests
```bash
pytest -v -x tests/kernels/moe/test_cutlass_moe.py
```

- benchmark scripts (note: these don't work on main)
```bash
python3 ./benchmark_grouped_gemm_cutlass.py --model "deepseek-ai/DeepSeek-V2-Lite" --tp-sizes 4 --batch-sizes 2 4
```

```bash
python benchmark_cutlass_moe_fp8.py  \
            --model "Llama-4-Maverick-17B-128E-Instruct-FP8"  \
            --tp-sizes 8 \
            --batch-size 2 4 8 \
            --per-act-token-opts false \
            --per-out-ch-opts false
```

- moe refactor integration tests

## Test Result

---
<details>

</details>


benchmarks/kernels/benchmark_cutlass_moe_fp8.py
```diff
@@ -120,85 +123,6 @@ def bench_run(
     # Force per-tensor quantization for all cases
     per_act_token = False
 
-    # Create stride tensors for CUTLASS
-    ab_strides1 = torch.full((num_experts,), k, dtype=torch.int64, device=device)
-    ab_strides2 = torch.full((num_experts,), n, dtype=torch.int64, device=device)
-    c_strides1 = torch.full((num_experts,), 2 * n, dtype=torch.int64, device=device)
-    c_strides2 = torch.full((num_experts,), k, dtype=torch.int64, device=device)
-
-    def run_triton_moe(
```
robertgshaw2-redhat, line 130: this function was unused

vllm/model_executor/layers/fused_moe/cutlass_moe.py
```diff
@@ -249,20 +249,28 @@ def run_cutlass_moe_fp8(
...
     ):
         assert quant_config.use_fp8_w8a8
         super().__init__(quant_config)
+
+        # E: num_experts
+        # N: intermediate size per partition
+        # K: hidden dim
+        ab_strides1_c_strides2 = torch.full((e,), k, device=device, dtype=torch.int64)
+        ab_strides2 = torch.full((e,), n, device=device, dtype=torch.int64)
+        c_strides1 = torch.full((e,), 2 * n, device=device, dtype=torch.int64)
```
bnellnm, line 267: I think this is more convenient from an API point of view but IIRC the strides tensors were pulled out so that they could be shared among multiple experts instances.

vllm/model_executor/layers/fused_moe/fallback.py
```diff
@@ -0,0 +1,118 @@
...
+        return (
+            self.experts.supports_chunking()
+            and self.fallback_experts.supports_chunking()
+        )
+
+    def supports_expert_map(self) -> bool:
+        return (
+            self.experts.supports_expert_map()
+            and self.fallback_experts.supports_expert_map()
+        )
```
bnellnm, line 50: Shouldn't the fallback and primary experts match behavior here as well?

vllm/model_executor/layers/fused_moe/oracle/fp8.py
```diff
@@ -0,0 +1,354 @@
...
+
+    use_deep_gemm = envs.VLLM_USE_DEEP_GEMM
+    if not is_deep_gemm_supported():
+        use_deep_gemm = False
+        logger.info_once(
+            "DeepGEMM is disabled because the platform does not support it.",
+            scope="local",
+        )
+
+    if use_deep_gemm and moe_use_deep_gemm and block_quant:
```
bnellnm, line 131: Should we check the block size too?

vllm/model_executor/layers/fused_moe/oracle/fp8.py
```diff
@@ -0,0 +1,354 @@
...
+    # all of the kernels in the TP case to use mk. Once this is
+    # done, then we will initialzie the TP case and DP/EP case
+    # via the same code path (i.e. via maybe_init_modular_kernel).
+    # NOTE(rob): in progress migrating all into this format.
+    use_inplace = True
+    if fp8_backend == Fp8MoeBackend.FLASHINFER_CUTLASS:
+        kernel = mk.FusedMoEModularKernel(
+            MoEPrepareAndFinalizeNoEP(
+                defer_input_quant=moe_quant_config.is_block_quantized
+            ),
```
bnellnm, line 290: Would it make more sense to have two different versions of this class instead of a flag?  I don't have any particular preference here.

vllm/model_executor/layers/fused_moe/triton_cutlass_moe.py
```diff
@@ -0,0 +1,75 @@
...
+        self,
+        hidden_states: torch.Tensor,
+        w1: torch.Tensor,
+        w2: torch.Tensor,
+    ) -> mk.FusedMoEPermuteExpertsUnpermute:
+        # Small batch fallback for sm100.
+        if self.is_sm100 and hidden_states.shape[0] <= 8:
+            return self.fallback_experts
+        else:
+            return self.experts
```
bnellnm: Would it make more sense to have a `select_experts` (or maybe `use_primary`) method that returns a bool instead that indicates primary or fallback experts?  Then it could be used by `workspace_shapes`, `apply`, etc?  This would prevent duplicating the selection logic in one or more spots.

vllm/model_executor/models/llama4.py
```diff
@@ -432,6 +432,7 @@ def load_moe_expert_weights(
 
         # Whether the MoE expert weights are loaded successfully.
         expert_param_loaded = False
+        loaded_weight = loaded_weight.to("cuda")
```
bnellnm: Was this a bug?

vllm/model_executor/layers/fused_moe/oracle/fp8.py
robertgshaw2-redhat: is there a specific size that is supported? I just copied over the existing logic here

vllm/model_executor/layers/fused_moe/oracle/fp8.py
robertgshaw2-redhat: yeah I think that could work. I would prefer to make the change in a separate PR though as this one already has a lot going on

vllm/model_executor/layers/fused_moe/triton_cutlass_moe.py
robertgshaw2-redhat: I originally had it that way, but I thought there might be times that the workspace_shapes selection and gemm impl selection might not be the same. I think that the TritonOrDeepGEMM experts had a comment about this

vllm/model_executor/layers/fused_moe/fallback.py
robertgshaw2-redhat: yeah I think you are right

vllm/model_executor/models/llama4.py
robertgshaw2-redhat: I meant to remove this. There is an issue with where loading Llama4 checkpoints for ModelOpt takes a very long time and this makes it go faster

vllm/model_executor/layers/fused_moe/triton_cutlass_moe.py
robertgshaw2-redhat: okay, I remember why I could not do this

The reason is that for the DeepGEMM case, we have different logic for workspace vs gemm selection and the gemm selection requires the weights to be there

I think we should fix this up, but leave it for another Pr since this is a big change and this PR has a lot going on

vllm/model_executor/layers/fused_moe/cutlass_moe.py
robertgshaw2-redhat: Arg

robertgshaw2-redhat: I think ~10MB per layer for llama maverick. So something like ~0.5GB

Given this kernel is not the most critical, I think we should leave it as is

vllm/model_executor/layers/fused_moe/oracle/fp8.py
robertgshaw2-redhat: I tend to agree that the oracle should validate. That being said, the spirit of this PR is to move from several places that make the selection to a single centralized place.

I plan to do a more extensive validation for all kernels that they support the features in a follow up. I would prefer to do that all at once

vllm/model_executor/layers/fused_moe/cutlass_moe.py
bnellnm: Maybe you can use the buffer manager for this?  i.e. the same one used for workspaces

vllm/model_executor/layers/fused_moe/oracle/fp8.py
```diff
@@ -0,0 +1,354 @@
...
+            logger.info_once(_make_log_backend("FlashInfer CUTLASS"))
+            return Fp8MoeBackend.FLASHINFER_CUTLASS
+
+    # weight-only path for older GPUs without native FP8
+    use_marlin = (
+        not current_platform.has_device_capability(89)
+        or envs.VLLM_TEST_FORCE_FP8_MARLIN
+    )
+    if current_platform.is_rocm():
+        use_marlin = False
```
mgoin: nit: just check for cuda first
```python
    use_marlin = (
        (current_platform.is_cuda()
        and not current_platform.has_device_capability(89))
        or envs.VLLM_TEST_FORCE_FP8_MARLIN
    )
```

vllm/model_executor/layers/fused_moe/oracle/fp8.py
mgoin: In reality I think nothing has been tested e2e except for 128x128, so we can safely assert that for now

vllm/model_executor/layers/fused_moe/cutlass_moe.py
robertgshaw2-redhat: yeah I think thats a good follow up item

vllm/model_executor/layers/fused_moe/triton_cutlass_moe.py
```diff
@@ -0,0 +1,75 @@
...
+                M,
+                N,
+                K,
+                topk,
+                global_num_experts,
+                local_num_experts,
+                expert_tokens_meta,
+            )
+
+    def select_gemm_impl(
```
mgoin: Shouldn't this be `_select_experts_impl` since `_select_experts_impl` is marked as abstract in `FallbackExperts`

tests/evals/gsm8k/configs/moe-refactor/Llama-4-Scout-Fp8-ModelOpt-fi-cutlass.yaml
```diff
@@ -2,7 +2,7 @@ model_name: "nvidia/Llama-4-Scout-17B-16E-Instruct-FP8"
 accuracy_threshold: 0.92
 num_questions: 1319
 num_fewshot: 5
-server_args: "--enforce-eager --max-model-len 8192 --tensor-parallel-size 2"
+server_args: "--enforce-eager --max-model-len 8192 --tensor-parallel-size 2 --disable-uvicorn-access-log"
```
mgoin: What is this? Do you just want to add it to the eval script here https://github.com/vllm-project/vllm/blob/c907d22158dbf31508e73e89cb2bb272bb59e37e/tests/evals/gsm8k/test_gsm8k_correctness.py#L60-L65

vllm/model_executor/layers/fused_moe/oracle/fp8.py
```diff
@@ -0,0 +1,354 @@
...
+    )
+    from vllm.model_executor.layers.fused_moe.fused_marlin_moe import (
+        MarlinExperts,
+    )
+    from vllm.model_executor.layers.fused_moe.prepare_finalize import (
+        MoEPrepareAndFinalizeNoEP,
+    )
+    from vllm.model_executor.layers.fused_moe.rocm_aiter_fused_moe import (
+        AiterExperts,
+    )
```
mgoin: I think we should just put each import within each conditional, rather than importing all

vllm/model_executor/layers/fused_moe/triton_cutlass_moe.py
robertgshaw2-redhat: yeah, I must not be triggering this code in the CI

robertgshaw2-redhat: nice catch

vllm/model_executor/layers/fused_moe/oracle/fp8.py
robertgshaw2-redhat: thanks

vllm/model_executor/layers/fused_moe/triton_cutlass_moe.py
robertgshaw2-redhat: addressed. I hadnt actually hooked up CutlassOrTritonExperts

I also added a test to ensure it runs on B200

vllm/model_executor/layers/fused_moe/triton_cutlass_moe.py
bnellnm: Do you know if it needs the actual weights or just the weight shapes?  Either way, I'm fine with deferring this til later.

robertgshaw2-redhat: I think we could avoid having the weights. We need to cleanup the logic for deepgemm selection

tests/evals/gsm8k/configs/moe-refactor/Llama-4-Scout-Fp8-ModelOpt-fi-cutlass.yaml
robertgshaw2-redhat: yeah, thats a good idea. It makes the logs much easier to read by not logging out /completions every request

bnellnm (approved):
Nice refactor!

vllm/model_executor/layers/quantization/utils/flashinfer_utils.py
```diff
@@ -301,3 +266,104 @@ def is_flashinfer_supporting_global_sf(backend: FlashinferMoeBackend | None) ->
...
+
+    # FI TRT-LLM FP8 per-tensor MoE kernel requires weight shuffle
+    # and registration of alpha scales. Note that we do not register
+    # as nn.Parameters since they are not needed for weight-reloading.
+    if is_trtllm and not block_quant:
+        assert w13_input_scale is not None
+        assert w2_input_scale is not None
+
+        rotate_weights_for_fi_trtllm_fp8_per_tensor_moe(w13, w2)
+        register_scales_for_trtllm_fp8_per_tensor_moe(
```
zyongye, line 361: Should we register this inside the `FusedMoEMethod?` instead of in the utils?

zyongye (commented):
The overall structure LGTM. But we will need to revisit the fallback experts abstraction in the future as discussed.

vllm/model_executor/layers/quantization/utils/flashinfer_utils.py
robertgshaw2-redhat: there is some work to make TRTLLM a modular kernel. once that is done, we can revisit it

mgoin (approved):

mgoin:
Great work!

Issue #16340, opened 2025-04-09
[Bug]: Missing "type":"function" in OpenAI-Compatible Streaming Tool Calls with specific tool_choice


iGmainC:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>

```text
PyTorch version: 2.6.0+cu124
Is debug build: False
CUDA used to build PyTorch: 12.4
ROCM used to build PyTorch: N/A

OS: Ubuntu 22.04.4 LTS (x86_64)
GCC version: (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Clang version: Could not collect
CMake version: Could not collect
Libc version: glibc-2.35

Python version: 3.12.4 | packaged by Anaconda, Inc. | (main, Jun 18 2024, 15:12:24) [GCC 11.2.0] (64-bit runtime)
Python platform: Linux-5.4.0-100-generic-x86_64-with-glibc2.35
Is CUDA available: True
CUDA runtime version: 12.4.131
CUDA_MODULE_LOADING set to: LAZY
GPU models and configuration: GPU 0: NVIDIA GeForce RTX 4090
Nvidia driver version: 535.129.03
cuDNN version: Probably one of the following:
/usr/lib/x86_64-linux-gnu/libcudnn.so.9.1.0
/usr/lib/x86_64-linux-gnu/libcudnn_adv.so.9.1.0
/usr/lib/x86_64-linux-gnu/libcudnn_cnn.so.9.1.0
/usr/lib/x86_64-linux-gnu/libcudnn_engines_precompiled.so.9.1.0
/usr/lib/x86_64-linux-gnu/libcudnn_engines_runtime_compiled.so.9.1.0
/usr/lib/x86_64-linux-gnu/libcudnn_graph.so.9.1.0
/usr/lib/x86_64-linux-gnu/libcudnn_heuristic.so.9.1.0
/usr/lib/x86_64-linux-gnu/libcudnn_ops.so.9.1.0
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Architecture:                    x86_64
CPU op-mode(s):                  32-bit, 64-bit
Address sizes:                   43 bits physical, 48 bits virtual
Byte Order:                      Little Endian
CPU(s):                          128
On-line CPU(s) list:             0-127
Vendor ID:                       AuthenticAMD
Model name:                      AMD EPYC 7542 32-Core Processor
CPU family:                      23
[...]

Issue #10781, opened 2024-11-29
[Bug]: Streaming w/ tool choice auto often truncates the final delta in the streamed arguments


cedonley:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>

```text
PyTorch version: 2.5.1+cu124
Is debug build: False
CUDA used to build PyTorch: 12.4
ROCM used to build PyTorch: N/A

OS: Ubuntu 22.04.5 LTS (x86_64)
GCC version: (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Clang version: Could not collect
CMake version: version 3.22.1
Libc version: glibc-2.35

Python version: 3.11.10 (main, Oct  3 2024, 07:29:13) [GCC 11.2.0] (64-bit runtime)
Python platform: Linux-5.15.0-122-generic-x86_64-with-glibc2.35
Is CUDA available: True
CUDA runtime version: 12.4.131
CUDA_MODULE_LOADING set to: LAZY
GPU models and configuration:
GPU 0: NVIDIA RTX A6000
GPU 1: NVIDIA RTX A6000
GPU 2: NVIDIA GeForce RTX 3090
GPU 3: NVIDIA GeForce RTX 3090

Nvidia driver version: 555.42.06
cuDNN version: Could not collect
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Architecture:                         x86_64
CPU op-mode(s):                       32-bit, 64-bit
Address sizes:                        46 bits physical, 48 bits virtual
Byte Order:                           Little Endian
CPU(s):                               24
On-line CPU(s) list:                  0-23
Vendor ID:                            GenuineIntel
Model name:                           13th Gen Intel(R) Core(TM) i7-13700K
CPU family:                           6
Model:                                183
Thread(s) per core:                   2
Core(s) per socket:                   16
Socket(s):                            1
Stepping:                             1
CPU max MHz:                          5400.0000
CPU min MHz:                          800.0000
BogoMIPS:                             6835.20
[...]

PR #31438, opened 2025-12-27
[Bugfix] Preserve tool call id/type/name in streaming finish chunk

amittell:
## Summary

When streaming tool calls, the finish chunk code in `serving_chat.py` overwrites the tool parser's `DeltaMessage` with a stripped-down version containing only `index` and `function.arguments`. This loses the `id`, `type`, and `function.name` fields.

This fix preserves those fields from the original `delta_message` when building the finish chunk.

## Issue

Fixes #31437

## What was wrong

```python
delta_message = DeltaMessage(
    tool_calls=[
        DeltaToolCall(
            index=index,
            function=DeltaFunctionCall(
                arguments=remaining_call
            ).model_dump(exclude_none=True),
        )
    ]
)
```

The `id`, `type`, and `function.name` fields were lost.

## The fix

```python
original_tc = delta_message.tool_calls[0]
original_fn = original_tc.function if original_tc else None
delta_message = DeltaMessage(
    tool_calls=[
        DeltaToolCall(
            index=index,
            id=original_tc.id if original_tc else None,
            type=original_tc.type if original_tc else None,
            function=DeltaFunctionCall(
                name=original_fn.name if original_fn else None,
                arguments=remaining_call,
            ),
        )
    ]
)
```

## Testing

Tested with GLM-4.7-AWQ using `--tool-call-parser glm47`. Before fix: ~20% of streaming tool call responses had `id=None`. After fix: 100% success rate.

## Related

- #16340 - Similar symptoms, different root cause
- #10781 - Mentions delta not being submitted correctly

### DIFF
--- a/vllm/entrypoints/openai/serving_chat.py
+++ b/vllm/entrypoints/openai/serving_chat.py
@@ -1208,13 +1208,19 @@ class OpenAIServingChat(OpenAIServing):
                             # check to see if there's anything left to stream
                             remaining_call = expected_call.replace(actual_call, "", 1)
                             # set that as a delta message
+                            # preserve id/type/name from original delta_message
+                            original_tc = delta_message.tool_calls[0]
+                            original_fn = original_tc.function if original_tc else None
                             delta_message = DeltaMessage(
                                 tool_calls=[
                                     DeltaToolCall(
                                         index=index,
+                                        id=original_tc.id if original_tc else None,
+                                        type=original_tc.type if original_tc else None,
                                         function=DeltaFunctionCall(
-                                            arguments=remaining_call
-                                        ).model_dump(exclude_none=True),
+                                            name=original_fn.name if original_fn else None,
+                                            arguments=remaining_call,
+                                        ),
                                     )
                                 ]
                             )
--- a/vllm/entrypoints/openai/serving_chat.py
+++ b/vllm/entrypoints/openai/serving_chat.py
@@ -1209,7 +1209,9 @@ class OpenAIServingChat(OpenAIServing):
                             remaining_call = expected_call.replace(actual_call, "", 1)
                             # set that as a delta message
                             # preserve id/type/name from original delta_message
-                            original_tc = delta_message.tool_calls[0]
+                            original_tc = next(
+                                (tc for tc in delta_message.tool_calls
+                                 if tc.index == index), None)
                             original_fn = original_tc.function if original_tc else None
                             delta_message = DeltaMessage(
                                 tool_calls=[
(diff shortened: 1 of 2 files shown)


vllm/entrypoints/openai/serving_chat.py
```diff
@@ -1208,13 +1208,28 @@ async def chat_completion_stream_generator(
...
+                            # preserve id/type/name from original delta_message
+                            original_tc = next(
+                                (
+                                    tc
+                                    for tc in delta_message.tool_calls
+                                    if tc.index == index
+                                ),
+                                None,
+                            )
+                            original_fn = original_tc.function if original_tc else None
```
chaunceyjiang: Thanks~ @amittell 

I noticed that #31451 seems to be addressing the same issue.

What do you think about #31451?

vllm/entrypoints/openai/serving_chat.py
```diff
@@ -1208,13 +1208,28 @@ async def chat_completion_stream_generator(
                             # check to see if there's anything left to stream
                             remaining_call = expected_call.replace(actual_call, "", 1)
                             # set that as a delta message
+                            # preserve id/type/name from original delta_message
```
chaunceyjiang: I suggest extracting lines L1211–L1234 into a separate function.

amittell:
Thanks for the pointer. I looked at #31451 - it skips delta creation when `remaining_call` is empty or matches current args.

One difference: when `remaining_call` is non-empty and different from current args, #31451 still creates a stripped delta without id/type/name. My fix preserves those fields in all cases by matching on index.

Refactored into a helper function per your suggestion:

```python
@staticmethod
def _create_remaining_args_delta(
    delta_message: DeltaMessage,
    remaining_call: str,
    index: int,
) -> DeltaMessage:
    """
    Create a delta message for remaining tool arguments, preserving
    id/type/name from the original delta.
    """
    original_tc = next(
        (tc for tc in delta_message.tool_calls if tc.index == index),
        None,
    )
    original_fn = original_tc.function if original_tc else None
    return DeltaMessage(
        tool_calls=[
            DeltaToolCall(
                index=index,
                id=original_tc.id if original_tc else None,
                type=original_tc.type if original_tc else None,
                function=DeltaFunctionCall(
                    name=original_fn.name if original_fn else None,
                    arguments=remaining_call,
                ),
            )
        ]
    )
```

Tested with GLM-4.7-AWQ on a 4-node cluster - 100% success rate. I can push this refactor if it looks good.

vllm/entrypoints/openai/serving_chat.py
amittell: Looked at #31451 - one difference is when `remaining_call` is non-empty and different from current args, it still creates a stripped delta without id/type/name. My fix preserves those fields in all cases. Happy to discuss if I'm missing something.

vllm/entrypoints/openai/serving_chat.py
amittell: Done - pushed fa59b42 with the helper function refactor.

chaunceyjiang (commented):
Thanks~ @amittell

chaunceyjiang (approved):

PR #31452, opened 2025-12-28
[LoRA] Add `--lora-target-modules` to selectively apply LoRA layers

laohyx:
## Purpose
This commit introduces a new feature, `--lora-target-modules`, and `--lora-exclude-modules`, which allows users to specify a regular expression to filter the modules where LoRA layers are applied. This is particularly useful for conserving memory by avoiding the initialization of LoRA layers for parts of the model that are not being fine-tuned, such as the FFN layers.
    
The changes include:
- Added lora_target_modules to LoRAConfig and the corresponding CLI argument.
- Implemented the filtering logic in LoRAModelManager to apply LoRA layers only to modules matching the regex.
- Added validation for the regex syntax.
- Added a check to ensure at least one LoRA layer is applied when a regex is provided.
- Added a warning when loading a LoRA adapter if it contains layers that do not match the provided regex.
- Updated documentation to explain the new feature and its memory-saving benefits.

## Test Plan
1.  Run models in both offline/serve mode, without new param `lora_target_modules` (baseline behavior), and do memory profiling.
2.  Run models in both model, with `--lora-target-modules ".*\\.self_attn\\..*"`, and do memory profiling.
3. `--lora-target-modules "NOT_EXSITING_LAYER"` to get a friendly error message.

## Test Result
1 & 2: tested on Qwen/Qwen3-VL-32B-Instruct-FP8, params:
```
--model /root/.cache/modelscope/hub/models/Qwen/Qwen3-VL-32B-Instruct-FP8 --max_model_len=1000 --enforce-eager --enable-lora
```

Both cases works, here is the difference of Pytorch memory result:

#### Test case1: Without `lora-target-modules`:
Without `lora_target_modules`

#### Test case 2, add ` --lora-target-modules ".*\\.self_attn\\..*"`
add ` --lora-target-modules`

#### Test case 3
Run with `--lora-target-modules "NOT_EXSITING_LAYER"`

Got a friendly error message:
```
(EngineCore_DP0 pid=3768408)   File "/app/data/heyuxuan/vllm/vllm/v1/worker/lora_model_runner_mixin.py", line 194, in maybe_dummy_run_with_lora
(EngineCore_DP0 pid=3768408)     with (
(EngineCore_DP0 pid=3768408)   File "/app/data/heyuxuan/miniconda3/envs/vllm013/lib/python3.11/contextlib.py", line 137, in __enter__
(EngineCore_DP0 pid=3768408)     return next(self.gen)
(EngineCore_DP0 pid=3768408)            ^^^^^^^^^^^^^^
(EngineCore_DP0 pid=3768408)   File "/app/data/heyuxuan/vllm/vllm/v1/worker/lora_model_runner_mixin.py", line 121, in maybe_setup_dummy_loras
(EngineCore_DP0 pid=3768408)     self.lora_manager.add_dummy_lora(lr, rank=lora_warmup_rank)
(EngineCore_DP0 pid=3768408)   File "/app/data/heyuxuan/vllm/vllm/lora/worker_manager.py", line 169, in add_dummy_lora
(EngineCore_DP0 pid=3768408)     dummy_lora = self._adapter_manager.create_dummy_lora(
(EngineCore_DP0 pid=3768408)                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=3768408)   File "/app/data/heyuxuan/vllm/vllm/lora/model_manager.py", line 554, in create_dummy_lora
(EngineCore_DP0 pid=3768408)     raise ValueError(
(EngineCore_DP0 pid=3768408) ValueError: No layers in the model have LoRA applied. It may be caused by incorrect target modules (--lora-target-modules), or all modules are excluded (--lora-exclude-modules). Please check your settings.
```

---
<details>

        The PR itself reduces memory usage about unused LoRA modules. 
</details>

---

> [!NOTE]
> Introduces selective LoRA initialization to reduce VRAM by limiting which modules receive LoRA layers.
> 
> - Adds `lora_target_modules` to `LoRAConfig`, included in hash and exposed via `--lora-target-modules`
> - Applies filtering in `LoRAModelManager` to only wrap modules matching list/regex; raises a clear error if no layers match
> - Logs warnings in `WorkerLoRAManager` when loaded adapters contain non-targeted modules that will be ignored
> - Updates docs (`docs/features/lora.md`) with guidance and examples (suffix list and regex) highlighting memory savings
> - Adds tests validating selective replacement behavior
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3b84348580c8ec45e03dea9174c050c0ae88989e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit 59dee6e08e94b66c70c0a24b0d869991c3617a3d. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/docs/configuration/conserving_memory.md
+++ b/docs/configuration/conserving_memory.md
@@ -201,13 +201,13 @@ By using `--lora-target-modules` to target only the self-attention layers, you c
 
 You can specify the target modules as a list of strings (matched as suffix or exact) or a regex string (matched anywhere).
 
-**Example: List of modules (suffix matching)**
+##### Example: List of modules (suffix matching)
 
 ```bash
 --lora-target-modules q_proj v_proj k_proj o_proj
 ```
 
-**Example: Regex**
+##### Example: Regex
 
 ```bash
 --lora-target-modules '.*\.self_attn\..*'
--- a/docs/configuration/conserving_memory.md
+++ b/docs/configuration/conserving_memory.md
@@ -183,3 +183,40 @@ llm = LLM(
     mm_processor_kwargs={"max_dynamic_patch": 4},  # Default is 12
 )
 ```
+
+## Selective LoRA with `lora_target_modules`
+
+When `--enable-lora` is configured, vLLM allocates GPU memory for all potential LoRA adapter layers at initialization time, based on the `--max-lora-rank`. This includes layers for self-attention, Feed-Forward Networks (FFN), Mixture-of-Experts (MoE), etc. This allocation happens even if the adapters you subsequently load do not modify all these components (e.g., they lack an FFN part).
+
+This default behavior can lead to a large memory consumption, especially since typical PEFT fine-tuning often only requires adjusting the self-attention layers, leaving the large FFN/MoE parts untouched.
+
+To optimize memory usage, you can use the `--lora-target-modules` command-line parameter to selectively initialize only the required LoRA layers.
+
+For example, with a model like `Qwen3-VL-32B-Instruct` and a `lora_rank=64`, the memory footprint for the initialized LoRA adapter (calculated with bf16) could be:
+
+- Self-attention: 51MiB
+- FFN (router + experts): 2406MiB
+
+By using `--lora-target-modules` to target only the self-attention layers, you can avoid allocating memory for the FFN/MoE LoRA layers, saving over 2.4 GiB of VRAM in this specific example.
+
+You can specify the target modules as a list of strings (matched as suffix or exact) or a regex string (matched anywhere).
+
+**Example: List of modules (suffix matching)**
+
+```bash
+--lora-target-modules q_proj v_proj k_proj o_proj
+```
+
+**Example: Regex**
+
+```bash
+--lora-target-modules '.*\.self_attn\..*'
+```
+
+You can also use `--lora-exclude-modules` to exclude specific modules.
+For example, to exclude the large FFN/MoE layers (which might be named `mlp`, `gate_up_proj`, `down_proj`, or `gate`):
+
+```bash
+--lora-exclude-modules '.*\.mlp\.*' gate_up_proj down_proj gate
+```
+
--- a/docs/configuration/conserving_memory.md
+++ b/docs/configuration/conserving_memory.md
@@ -201,13 +201,13 @@ By using `--lora-target-modules` to target only the self-attention layers, you c
 
 You can specify the target modules as a list of strings (matched as suffix or exact) or a regex string (matched anywhere).
 
-##### Example: List of modules (suffix matching)
+Example: List of modules (suffix matching)
 
 ```bash
 --lora-target-modules q_proj v_proj k_proj o_proj
 ```
 
-##### Example: Regex
+Example: Regex
 
 ```bash
 --lora-target-modules '.*\.self_attn\..*'
@@ -219,4 +219,3 @@ For example, to exclude the large FFN/MoE layers (which might be named `mlp`, `g
 ```bash
 --lora-exclude-modules '.*\.mlp\.*' gate_up_proj down_proj gate
 ```
-
--- a/docs/configuration/conserving_memory.md
+++ b/docs/configuration/conserving_memory.md
@@ -210,12 +210,5 @@ Example: List of modules (suffix matching)
 Example: Regex
 
 ```bash
---lora-target-modules '.*\.self_attn\..*'
-```
-
-You can also use `--lora-exclude-modules` to exclude specific modules.
-For example, to exclude the large FFN/MoE layers (which might be named `mlp`, `gate_up_proj`, `down_proj`, or `gate`):
-
-```bash
---lora-exclude-modules '.*\.mlp\.*' gate_up_proj down_proj gate
-```
++--lora-target-modules '.*\.self_attn\..*'
++```
--- a/docs/configuration/conserving_memory.md
+++ b/docs/configuration/conserving_memory.md
@@ -183,32 +183,3 @@ llm = LLM(
     mm_processor_kwargs={"max_dynamic_patch": 4},  # Default is 12
 )
 ```
-
-## Selective LoRA with `lora_target_modules`
-
-When `--enable-lora` is configured, vLLM allocates GPU memory for all potential LoRA adapter layers at initialization time, based on the `--max-lora-rank`. This includes layers for self-attention, Feed-Forward Networks (FFN), Mixture-of-Experts (MoE), etc. This allocation happens even if the adapters you subsequently load do not modify all these components (e.g., they lack an FFN part).
-
-This default behavior can lead to a large memory consumption, especially since typical PEFT fine-tuning often only requires adjusting the self-attention layers, leaving the large FFN/MoE parts untouched.
-
-To optimize memory usage, you can use the `--lora-target-modules` command-line parameter to selectively initialize only the required LoRA layers.
-
-For example, with a model like `Qwen3-VL-32B-Instruct` and a `lora_rank=64`, the memory footprint for the initialized LoRA adapter (calculated with bf16) could be:
-
-- Self-attention: 51MiB
-- FFN (router + experts): 2406MiB
-
-By using `--lora-target-modules` to target only the self-attention layers, you can avoid allocating memory for the FFN/MoE LoRA layers, saving over 2.4 GiB of VRAM in this specific example.
-
-You can specify the target modules as a list of strings (matched as suffix or exact) or a regex string (matched anywhere).
-
-Example: List of modules (suffix matching)
-
-```bash
---lora-target-modules q_proj v_proj k_proj o_proj
-```
-
-Example: Regex
-
-```bash
-+--lora-target-modules '.*\.self_attn\..*'
-+```
--- a/docs/features/lora.md
+++ b/docs/features/lora.md
@@ -371,3 +371,32 @@ vllm serve model --enable-lora --max-lora-rank 64
 # Bad: unnecessarily high, wastes memory
 vllm serve model --enable-lora --max-lora-rank 256
 ```
+
+### Selective LoRA with `lora_target_modules` to Reduce Memory Usage
+
+When `--enable-lora` is configured, vLLM allocates GPU memory for all potential LoRA adapter layers at initialization time, based on the `--max-lora-rank`. This includes layers for self-attention, Feed-Forward Networks (FFN), Mixture-of-Experts (MoE), etc. This allocation happens even if the adapters you subsequently load do not modify all these components (e.g., they lack an FFN part).
+
+This default behavior can lead to a large memory consumption, especially since typical PEFT fine-tuning often only requires adjusting the self-attention layers, leaving the large FFN/MoE parts untouched.
+
+To optimize memory usage, you can use the `--lora-target-modules` command-line parameter to selectively initialize only the required LoRA layers.
+
+For example, with a model like `Qwen3-VL-32B-Instruct` and a `lora_rank=64`, the memory footprint for the initialized LoRA adapter (calculated with bf16) could be:
+
+- Self-attention: 51MiB
+- FFN (router + experts): 2406MiB
+
+By using `--lora-target-modules` to target only the self-attention layers, you can avoid allocating memory for the FFN/MoE LoRA layers, saving over 2.4 GiB of VRAM in this specific example.
+
+You can specify the target modules as a list of strings (matched as suffix or exact) or a regex string (matched anywhere).
+
+Example: List of modules (suffix matching)
+
+```bash
+--lora-target-modules q_proj v_proj k_proj o_proj
+```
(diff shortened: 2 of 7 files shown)


laohyx:
> I think this is the right direction, but I believe we should support not only regex but also specifying a layer list.

hi @jeejeelee , how about keep align with `peft`'s [target_modules and exclude_modules](https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/config.py#L460-L478)? 

In vLLM, it could be `--lora-target-modules`, accepting list of strings or a single string, and `--lora-exclude-modules`.

Reference code (peft's implementation): 
https://github.com/huggingface/peft/blob/main/src/peft/tuners/tuners_utils.py#L1655-L1724

laohyx:
hi @jeejeelee @cjackal 

Could you please raise some comments on this PR, and push this feature into vLLM? 

This really helps users to fine-tune memory usage with LoRA adapters.

Thanks!

docs/configuration/conserving_memory.md
```diff
@@ -183,3 +183,32 @@ llm = LLM(
     mm_processor_kwargs={"max_dynamic_patch": 4},  # Default is 12
 )
 ```
+
+## Selective LoRA with `lora_target_modules`
```
jeejeelee: Maybe we should move this docs to https://docs.vllm.ai/en/latest/features/lora/

laohyx: OK, this section is moved to docs/features/lora.md

linitra24:
Hi @laohyx, If `--lora-target-modules` is not set, could we simply fall back to the `"target_modules"` defined in adapter_config.json and use that as the default?

vllm/lora/worker_manager.py
```diff
@@ -140,6 +140,20 @@ def _load_adapter(self, lora_request: LoRARequest) -> LoRAModel:
             # For BadRequestError
             raise e
 
+        if self.lora_config.lora_target_modules:
```
jeejeelee, line 143: We can get all the modules that support LoRA from the above `expected_lora_modules`, and if `lora_target_modules` is not None, we should reset `expected_lora_modules = lora_target_modules`

docs/features/lora.md
```diff
@@ -371,3 +371,32 @@ vllm serve model --enable-lora --max-lora-rank 64
...
+
+```bash
+--lora-target-modules q_proj v_proj k_proj o_proj
+```
+
+Example: Regex
+
+```bash
++--lora-target-modules '.*\.self_attn\..*'
++```
```
jeejeelee, line 402: If we set `--lora-target-modules self_attn`, but an adapter containing MLP LoRA layers is passed, this should raise an error. We should document this risk to make users aware of it.

vllm/config/lora.py
```diff
@@ -60,6 +60,15 @@ class LoRAConfig:
     of multimodal models will be enabled. This is an experimental feature and 
     currently only supports some MM models such as the Qwen VL series. The default 
     is False."""
+    lora_target_modules: str | list[str] | None = None
+    """List of module names or regex expressions of the module names to replace
+    with LoRA. If not specified, all supported modules will be enabled for 
+    possible future LoRA adapters. If you only want to enable self-attention's 
+    q_proj and v_proj layers for LoRA adapters, you can set this to a list of module
+    names like ['q_proj', 'v_proj'], or a regex expression like
```
jeejeelee, line 68: Does `--lora-target-modules` specify HuggingFace modules or vLLM modules? I believe it should specify vLLM modules. For example, in llama's self-attention, the QKV projections are three separate modules in HuggingFace (`q_proj, k_proj, v_proj`), but vLLM fuses them into a single module called `qkv_proj`.

laohyx: I did test on Llama-3.2-3B-Instruct.
My debug result is: `expected_lora_modules` is for HF modules.

Shall we ephesis in comment, that the module names or layers are align with HuggingFace layers?

image

vllm/lora/worker_manager.py
laohyx: I understand that modifying `expected_lora_modules` is a good way to load and check with user-specified `lora_path`.

But its value is:
- Llama-3.2-3B-Instruct: `{'o_proj', 'lm_head', 'embed_tokens', 'down_proj', 'k_proj', 'gate_proj', 'q_proj', 'up_proj', 'v_proj'}`
- Qwen3-VL-30B-A3B-Instruct: `{'proj', 'v_proj', 'experts', 'o_proj', 'k_proj', 'linear_fc2', 'linear_fc1', 'qkv', 'q_proj', 'gate'}`

They are the __last part__ of layer names, whose __full name__ is like `model.language_model.layers.0.self_attn.k_proj`.
Currently, `lora_target_modules` matches against __full__ layer names, instead of "k_proj". It seem __*improper*__ to name as "target modules".

Here are two options:
1. keep the name `lora_target_modules`, but remove regex, only compare with module names like {'proj', 'v_proj', 'experts', 'o_proj', 'k_proj', 'linear_fc2', 'linear_fc1', 'qkv', 'q_proj', 'gate'}
2. rename to `lora_target_layers`, and keep current implementation: match against full layer names (like `model.language_model.layers.0.self_attn.k_proj`)

vllm/config/lora.py
jeejeelee: For example, q_proj, k_proj, and v_proj are merged into `qkv_proj` in vLLM. We have no way to specify only one or two of them individually, so I think it might be better to align with vLLM's naming convention.

docs/features/lora.md
jeejeelee: Please reponse this comment

vllm/config/lora.py
cjackal: Just a few thoughts on UX perspective, I am leaving them as a reply here as the weight naming convention is related:

Changing the syntax of a CLI flag after it has been documented and public may leave some vllm recipes obsolete and cause some claims, we have already observed a similar one [in the case of `--rope-theta` flag removal](https://github.com/vllm-project/vllm/issues/29026). As the memory budget gain from this PR is pretty substantial, I'd expect that model providers with LoRA adapters as part of the model checkpoint (not a lot at the moment, but phi-4 is one instance, and my team also has some real-world examples) are likely in favor of guiding the users to this new CLI flag. So I wonder if we design the syntax for this new feature with care so that most AI researchers feel natural with the convention?

I think most open-weight adapters are in huggingface PEFT format, so it seems @linitra24 's opinion is attractive in the sense that we may just pass the would-be `adapter_config.json` format as JSON input of the CLI flag. I guess many LoRA weights that benefits from this PR have a full attention + some selective MLP/MoE layers which have both `target_modules`(for full attentions) and `target_parameters`(for selective MLP/MoE layers) in its `adapter_config.json`. So I wonder how the following format (or similar) feels like:

```shell
--adapter-config.target_modules 'q_proj,k_proj,v_proj,o_proj' --adapter-config.target_paramters '11.mlp.experts.gate_up_proj,11.mlp.experts.down_proj,22.mlp.experts.gate_up_proj,22.mlp.experts.down_proj,33.mlp.experts.gate_up_proj,33.mlp.experts.down_proj'
```
(this one comes from a gpt-oss-120b LoRA adapter)

Of course we need the PEFT -> vllm weight name map in this case, but I think it may be better than exposing the vllm-internal weight names to the users (teaching the users the vllm internal sounds like a big learning curve IMO).

And I don't think there is a practical usecase of only some of qkv LoRA weights trained, so it sounds safe to just raise with proper error message for this rare case?

docs/features/lora.md
laohyx: Sure. Currently I printed a warning message in vllm/lora/worker_manager.py (in _load_adapter function):
```
        if self.lora_config.lora_target_modules:
            for module_name in lora.loras:
                if not self._adapter_manager._check_target_module_exists(
                    self.lora_config, module_name
                ):
                    logger.warning(
                        "LoRA module '%s' in adapter '%s' is not targeted by "
                        "the current configuration (lora_target_modules). "
                        "These parameters will be ignored, which may cause "
                        "abnormal model behavior.",
                        module_name,
                        lora_request.lora_path,
                    )

```

Indeed I can change it to raising an exception (ValueError). I will make this change once other comments have decisions. : )

varun-sundar-rabindranath: I think a warning is good here. This will help us in perf analysis / ablation studies.

vllm/config/lora.py
varun-sundar-rabindranath: >Of course we need the PEFT -> vllm weight name map in this case, but I think it may be better than exposing the vllm-internal weight names to the users (teaching the users the vllm internal sounds like a big learning curve IMO).

Agree. 

> And I don't think there is a practical usecase of only some of qkv LoRA weights trained, so it sounds safe to just raise with proper error message for this rare case?

I agree with raising an error for configurations that don't make sense. 

about : 
> `--adapter-config.target_modules 'q_proj,k_proj,v_proj,o_proj' --adapter-config.target_paramters '11.mlp.experts.gate_up_proj,11.mlp.experts.down_proj,22.mlp.experts.gate_up_proj,22.mlp.experts.down_proj,33.mlp.experts.gate_up_proj,33.mlp.experts.down_proj'`

I think it is better to keep a single argument that handles both cases. i.e., just have 
```
--adapter-config.target_modules 'q_proj,k_proj,v_proj,o_proj, 11.mlp.experts.gate_up_proj,11.mlp.experts.down_proj,22.mlp.experts.gate_up_proj,22.mlp.experts.down_proj,33.mlp.experts.gate_up_proj,33.mlp.experts.down_proj'
```
I understand that it would overload the term "target_modules" - but that is okay in the sense that it would avoid dev errors where we have to treat the lists differently. Also, the users dont have to learn what the 2 different lists mean, instead they put everything in one list.

varun-sundar-rabindranath (commented):
IIUC, 
- `supported_lora_modules` is a coarse filter (we guarantee vLLM can support LoRA for this particular module)
- `target_modules` is fine-grained where the user has more control over what module is  loaded. 

also, based on the comments on the PR we want the `target_modules` support to be versatile (which I agree). some examples, 
 * `['q_proj', 'k_proj']` - Last name support (If the model has q_proj / k_proj modules treat it as a LoRA module)
 * `['.*11.mlp.experts.gate_up_proj']` - Only treat this module as LoRA
 * `['model.language_model.layers.[0-4].self_attn.[qkv]_proj']` - Treat attention, qkv proj modules from layers 0-4 as LoRA modules.
  * `[.*attn.*]` - Treat all supported modules in attention as LoRA modules
  * ['.*mlp.*'] - Treat all supported modules in MLP as LoRA modules 
  * ['11.attn.q_proj,11.attn.k_proj,11.attn.v_proj,11.attn.o_proj,11.mlp.experts.gate_up_proj,11.mlp.experts.down_proj,22.mlp.experts.gate_up_proj,22.mlp.experts.down_proj,33.mlp.experts.gate_up_proj,33.mlp.experts.down_proj']  - Just list everything you are interested in explicitly. Last resort if everything else fails (in case there is a bug with selection logic) 

I believe the PR in its current form handles all these cases. @laohyx  please correct me if  I am wrong. 

The distinction between `target_modules` and `target_parameters` - thinking about this a bit, I think we can treat them the same. we can make an argument `target_parameters` are still LoRA modules that live deeper in the hierarchy. 

Either way, I think  we should land the PR given the benefits. I am not taking the UI changes lightly, but we can always deprecate the UI parameter if we find the need.

If we think this needs more thought, we can land https://github.com/vllm-project/vllm/pull/34984/ which doesn't have regex support, but still lets the user have some fine grained filtering abilities (last name filtering). 

cc @jeejeelee  @cjackal @laohyx wdyt ?

vllm/config/lora.py
cjackal: > I think it is better to keep a single argument that handles both cases. i.e., just have
> 
> ```
> --adapter-config.target_modules 'q_proj,k_proj,v_proj,o_proj, 11.mlp.experts.gate_up_proj,11.mlp.experts.down_proj,22.mlp.experts.gate_up_proj,22.mlp.experts.down_proj,33.mlp.experts.gate_up_proj,33.mlp.experts.down_proj'
> ```
> 
> I understand that it would overload the term "target_modules" - but that is okay in the sense that it would avoid dev errors where we have to treat the lists differently. Also, the users dont have to learn what the 2 different lists mean, instead they put everything in one list.

Yeah I think the nomenclature of `target_modules` and `target_parameters` are from huggingface model architecture's `nn.Linear` and `nn.Parameters`, in vllm model weight loader the distinction between modules and parameters are vague. In #34984 we took vllm internal weight names, so there is little motivation to separate the weight names into two any longer I think.

PR #31460, opened 2025-12-28
[CI]Test Group 'NixlConnector PD accuracy tests' is fixed

qli88:
On ROCm platform we must use RIXL instead of NIXL. This PR is to add RIXL/ucx into ROCm docker file and replace NIXL with RIXL for CI tests.

### DIFF
--- a/requirements/kv_connectors_rocm.txt
+++ b/requirements/kv_connectors_rocm.txt
@@ -1,0 +1,2 @@
+tblib
+lm_eval[api]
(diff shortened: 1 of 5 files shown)


robertgshaw2-redhat (approved):

robertgshaw2-redhat:
LGTM

requirements/kv_connectors_rocm.txt
```diff
@@ -0,0 +1,2 @@
+tblib
+lm_eval[api]
```
NickLucche, line 2: @qli88 are these actual runtime requirements? These look more like tests requirements. If that's the case we should just install them inline

PR #26189, opened 2025-10-03
[Model] Gemma3: Fix GGUF loading and quantization

lucianommartins:
## Purpose

Fix Gemma3 GGUF quantization support in vLLM, resolving gibberish output and enabling true Q4_0 compression.

**Issues Resolved:**
- Fixes #14753 
- Fixes #15480

**Problems Fixed:**
1. Model type mismatch: HuggingFace uses `gemma3_text` but GGUF expects `gemma3`
2. Gibberish output: Incorrect RMSNorm weight handling due to architectural difference
3. Missing quantization: Models loaded in BF16 instead of compressed Q4_0 format
4. Shape warnings: Hundreds of "GGUF SHAPE MISMATCH" warnings during loading

### Test Environment
- Python 3.13
- CUDA 12.4
- NVIDIA A100-80GB GPU
- vLLM built from source with modifications

### Test Command
```bash
python -c "
from vllm import LLM, SamplingParams
llm = LLM(model='path/to/gemma-3-1b-it-Q4_0.gguf', max_model_len=2048)
outputs = llm.generate(['Hello, my name is'], SamplingParams(max_tokens=20))
print(outputs[0].outputs[0].text)
"
```

### Models Tested
Comprehensive validation across 8 Gemma3 variants:
- gemma-3-1b-pt (Q4_0)
- gemma-3-1b-it (Q4_0)
- gemma-3-4b-pt (Q4_0)
- gemma-3-4b-it (Q4_0)
- gemma-3-12b-pt (Q4_0)
- gemma-3-12b-it (Q4_0)
- gemma-3-27b-pt (Q4_0)
- gemma-3-27b-it (Q4_0)

### Before Fix
- Model loading failed: `ValueError: GGUF model with architecture gemma3 is not supported yet`
- Alternative load produced gibberish: "������������������"
- Hundreds of shape mismatch warnings
- Using BF16 (2x memory vs Q4_0)

### After Fix
- All 8 models load successfully
- Coherent text generation (e.g., "Hello, my name is Alice and I am a...")
- No shape mismatch warnings
- True Q4_0 compression active (~50% memory reduction)

### Performance Metrics

| Model | File Size | GPU Memory (Q4_0) | Memory vs BF16 | Status |
|-------|-----------|-------------------|----------------|-----------|---------|
| 1B-pt | 665 MB | ~0.37 GB | 55% | PASS |
| 1B-it | 665 MB | ~0.37 GB | 55% | PASS |
[...]

PR #31464, opened 2025-12-29
[Bugfix] Apply RMSNorm weight correction for Gemma2 GGUF models

kitaekatt:
## Summary

llama.cpp adds 1 to RMSNorm weights during GGUF conversion (see [convert_hf_to_gguf.py#L3397-L3400](https://github.com/ggml-org/llama.cpp/blob/be7c3034108473beda214fd1d7c98fd6a7a3bdf5/convert_hf_to_gguf.py#L3397-L3400)), but vLLM expects original values. Without this correction, Gemma2 GGUF models produce gibberish output.

This fix applies the same correction that was added for Gemma3 in PR #26189 - subtracting 1 from norm weights during GGUF loading.

## Root Cause

When loading Gemma2 GGUF models, the RMSNorm weights are +1 higher than expected because llama.cpp's GGUF conversion adds 1 to these weights. The Gemma3 model already handles this (added in #26189), but Gemma2 was missing the same correction.

## Changes

- Add `_process_weights` generator to `Gemma2ForCausalLM.load_weights()` that subtracts 1 from norm weights when loading GGUF models
- Pattern matches existing Gemma3 implementation

## Testing

Tested with `bartowski/gemma-2-2b-it-GGUF` on RTX 5090:

**Before fix**: Gibberish output (excessive script mixing: ARABIC, CJK, CYRILLIC, GREEK, LATIN)
**After fix**: Coherent output, 40% accuracy on MMLU benchmark, 344 tok/s throughput

```
Single-Thread Baseline Results: 40.00% accuracy (4/10 correct)
Throughput: 343.70 tok/s
```

### DIFF
--- a/vllm/model_executor/models/gemma2.py
+++ b/vllm/model_executor/models/gemma2.py
@@ -428,8 +428,24 @@ class Gemma2ForCausalLM(nn.Module, SupportsLoRA, SupportsPP):
         return logits
 
     def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
+        # Apply GGUF RMSNorm weight correction before loading.
+        # llama.cpp adds 1 to RMSNorm weights during GGUF conversion,
+        # so we need to subtract 1 to restore original values.
+        # See: https://github.com/ggml-org/llama.cpp/blob/be7c3034108473beda214fd1d7c98fd6a7a3bdf5/convert_hf_to_gguf.py#L3397-L3400
+        def _process_weights(
+            weights: Iterable[tuple[str, torch.Tensor]],
+        ) -> Iterable[tuple[str, torch.Tensor]]:
+            for name, loaded_weight in weights:
+                if (
+                    self.quant_config
+                    and self.quant_config.get_name() == "gguf"
+                    and name.endswith("norm.weight")
+                ):
+                    loaded_weight = loaded_weight - 1
+                yield name, loaded_weight
+
         loader = AutoWeightsLoader(
             self,
             skip_prefixes=(["lm_head."] if self.config.tie_word_embeddings else None),
         )
-        return loader.load_weights(weights)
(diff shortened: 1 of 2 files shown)


hmellor (changes requested):
Instead of modifying the weights can we just say that `if self.quant_config.get_name() == "gguf"` we use `RMSNorm` instead of `GemmaRMSNorm`?

vllm/model_executor/models/gemma2.py
```diff
@@ -217,14 +217,21 @@ def __init__(
...
             config.hidden_size, eps=config.rms_norm_eps
         )
-        self.pre_feedforward_layernorm = GemmaRMSNorm(
+        self.pre_feedforward_layernorm = norm_cls(
             config.hidden_size, eps=config.rms_norm_eps
         )
-        self.post_feedforward_layernorm = GemmaRMSNorm(
+        self.post_feedforward_layernorm = norm_cls(
             config.hidden_size, eps=config.rms_norm_eps
         )
```
hmellor, line 236: Could you use this pattern as it's more compact? (and use it in the other places too)

```suggestion
        # GGUF stores RMSNorm weights with +1 baked in (llama.cpp convention).
        # GemmaRMSNorm adds 1 in its forward pass, so use plain RMSNorm for GGUF.
        quant_name = quant_config.get_name() if quant_config else None
        rms_norm_cls = RMSNorm if quant_name == "gguf" else GemmaRMSNorm
        rms_norm_kwargs = dict(hidden_size=config.hidden_size, eps=config.rms_norm_eps)
        self.input_layernorm = rms_norm_cls(**rms_norm_kwargs)
        self.post_attention_layernorm = rms_norm_cls(**rms_norm_kwargs)
        self.pre_feedforward_layernorm = rms_norm_cls(**rms_norm_kwargs)
        self.post_feedforward_layernorm = rms_norm_cls(**rms_norm_kwargs)
```

PR #31465, opened 2025-12-29
fixed mypy warnings for files vllm/v1/attention with TEMPORARY workaround

MrIceCreamMan:
## Purpose
Follow-up to https://github.com/vllm-project/vllm/pull/26448

This PR improves type checking coverage for the attention module by:
- Moving `vllm/v1/attention` from SEPARATE_GROUPS to FILES in mypy configuration
- Fixing type errors in the attention module to enable strict type checking
- This ensures better code quality and catches type-related bugs earlier

This is part of the effort to enable comprehensive type checking across the vLLM codebase.

### Dynamic Module Type Checking Workaround

This PR includes `# type: ignore[attr-defined]` annotations for imports from the `vllm.vllm_flash_attn` module. This is necessary because `vllm_flash_attn` is a dynamically built C extension module that behaves differently in CI and local environments:

**Local environment:** The module is compiled during build/installation, making all attributes available to mypy.

**CI environment:** Pre-commit hooks run on a fresh checkout before the build step, leaving `vllm_flash_attn` as an empty stub, causing mypy to report missing attribute errors.

**Future improvement:** The ideal solution would be to update the CI workflow to build extensions before running type checking:
```yaml
- name: Build vLLM extensions
  run: python setup.py build_ext --inplace

- name: Run pre-commit
  run: pre-commit run --all-files
```

This would eliminate the need for type ignore comments while maintaining comprehensive type checking coverage.

## Test Plan
1. Type checking:
```bash
pre-commit run --hook-stage manual mypy-3.10 -a
```

2. Unit tests for attention module:
```bash
pytest tests/v1/attention/ -v -k "not (mla and (decode or mixed or spec_decode)) and not sparse_mla"
```

3. Pre-commit checks:
```bash
rm -rf .mypy_cache/
pre-commit clean
pre-commit run --all-files --hook-stage manual
```

## Test Results
**Type checking:**
```
```

**Unit tests:**
```
======================= 109 passed, 22 skipped, 88 deselected, 493 warnings in 114.47s (0:01:54) =======================
```

**Pre-commit hooks:** All checks passed ✓

---
<details>

- [ ] (Optional) Release notes update. If your change is user facing, please update the release notes draft in the [[Google Doc](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0)](https://docs.google.com/document/d/1YyVqrgX4gHTtrstbq8oWUImOyPCKSGnJ7xtTpmXzlRs/edit?tab=t.0).
</details>


yewentao256 (commented):
Thanks for the work! Will take a look later

yewentao256 (commented):
The pre-commit fails for some reason, please take a look and fix manually if automatic one doesn't work in your env, also please take a look what Gemini suggests

vllm/v1/attention/backends/mla/common.py
```diff
@@ -251,13 +252,18 @@ class QueryLenSupport(Enum):
...
 
+    flash_attn_varlen_func = _vllm_fa
     is_vllm_fa = True
 except ImportError:
     # For rocm use upstream flash attention
     if current_platform.is_rocm():
-        from flash_attn import flash_attn_varlen_func
+        from flash_attn import flash_attn_varlen_func as _upstream_fa
+
+        flash_attn_varlen_func = _upstream_fa
```
yewentao256: Why this is needed?

vllm/v1/attention/backends/mla/common.py
```diff
@@ -251,13 +252,18 @@ class QueryLenSupport(Enum):
 
 
 try:
-    from vllm.vllm_flash_attn import flash_attn_varlen_func
+    from vllm.vllm_flash_attn import (  # type: ignore[attr-defined]
+        flash_attn_varlen_func as _vllm_fa,
+    )
 
+    flash_attn_varlen_func = _vllm_fa
```
yewentao256: same here

vllm/v1/attention/backends/mla/flashattn_mla.py
```diff
@@ -68,7 +71,7 @@ def supports_combination(
         head_size: int,
         dtype: torch.dtype,
         kv_cache_dtype: CacheDType | None,
-        block_size: int,
+        block_size: int | None,
```
yewentao256: Will the caller pass a None here? I think we should update in the caller side

vllm/v1/attention/backends/mla/flashmla.py
```diff
@@ -71,7 +71,7 @@ def supports_combination(
         head_size: int,
         dtype: torch.dtype,
         kv_cache_dtype: CacheDType | None,
-        block_size: int,
+        block_size: int | None,
```
yewentao256: same here

vllm/v1/attention/backends/mla/rocm_aiter_mla.py
```diff
@@ -270,7 +270,7 @@ def _forward_decode(
             o,
             self.scale,
             attn_metadata.decode.qo_indptr,
-            attn_metadata.decode.max_qo_len,
+            attn_metadata.decode.max_qo_len or 0,
```
yewentao256: Try not to use or 0 which is kind of confusing

vllm/v1/attention/backends/flash_attn.py
```diff
@@ -709,11 +713,13 @@ def forward(
...
                     alibi_slopes=self.alibi_slopes,
-                    window_size=self.sliding_window,
+                    window_size=list(self.sliding_window)
+                    if self.sliding_window is not None
+                    else None,
                     block_table=block_table,
                     softcap=self.logits_soft_cap,
                     scheduler_metadata=scheduler_metadata,
-                    fa_version=self.vllm_flash_attn_version,
+                    fa_version=self.vllm_flash_attn_version or 0,
```
yewentao256: Same here

vllm/v1/attention/backends/flash_attn.py
```diff
@@ -709,11 +713,13 @@ def forward(
                     softmax_scale=self.scale,
                     causal=attn_metadata.causal,
                     alibi_slopes=self.alibi_slopes,
-                    window_size=self.sliding_window,
+                    window_size=list(self.sliding_window)
+                    if self.sliding_window is not None
+                    else None,
```
yewentao256: extract the if else outside

yewentao256 (commented):
Thanks for the work! A few thoughts

vllm/v1/attention/backends/mla/common.py
MrIceCreamMan: This pattern is required to avoid a mypy `no-redef` error. When importing `flash_attn_varlen_func` directly in both the try and except blocks, mypy flags this as:
```
vllm/v1/attention/backends/mla/common.py:260: error: Name "flash_attn_varlen_func" already defined (possibly by an import)  [no-redef]
```
Mypy treats each direct import as a definition, so importing the same name in both branches is considered a redefinition. By importing to intermediate aliases (`_vllm_fa` and `_upstream_fa`) and then assigning to `flash_attn_varlen_func`, we avoid this issue since the assignment is not considered a new definition by mypy's type checker.

vllm/v1/attention/backends/mla/flashattn_mla.py
MrIceCreamMan: Updated

vllm/v1/attention/backends/mla/flashmla.py
MrIceCreamMan: updated

vllm/v1/attention/backends/mla/rocm_aiter_mla.py
MrIceCreamMan: updated

vllm/v1/attention/backends/flash_attn.py
MrIceCreamMan: updated

vllm/v1/attention/backends/flash_attn.py
MrIceCreamMan: updated

vllm/v1/attention/backends/mla/common.py
MrIceCreamMan: Same rationale as above.

vllm/v1/attention/backends/mla/common.py
yewentao256: Is there any other way to fix this? Eg. type ignore.
`from flash_attn import flash_attn_varlen_func as _upstream_fa` looks confusing and we'd better no to use this

vllm/v1/attention/backends/gdn_attn.py
```diff
@@ -75,8 +75,9 @@ def __init__(
         self.compilation_config = vllm_config.compilation_config
         self.speculative_config = vllm_config.speculative_config
         self.kv_cache_spec = kv_cache_spec
+
         if self.speculative_config:
-            self.num_spec = self.speculative_config.num_speculative_tokens
+            self.num_spec: int = self.speculative_config.num_speculative_tokens or 0
```
yewentao256: NIT: still a `or 0` here

yewentao256 (commented):
Thanks for the work! Let's run CI as well

vllm/v1/attention/backends/gdn_attn.py
MrIceCreamMan: resolved

yewentao256 (approved):
LGTM, thanks for the work!

vllm/v1/attention/backends/flash_attn.py
```diff
@@ -599,6 +603,9 @@ def forward(
               We use torch's .expand() to avoid duplicating values
         """
         assert output is not None, "Output tensor must be provided."
+        assert self.vllm_flash_attn_version is not None, (
```
tjtanaa, line 606: Does mypy allow us to add this assertion after line in https://github.com/vllm-project/vllm/blob/3212278ebbf4231f1333bc945b351b088262cd22/vllm/v1/attention/backends/flash_attn.py#L557 since this backend requires vllm_flash_attn to be installed?  This way, we don't need to call it in every forward passes.

vllm/v1/attention/backends/flash_attn.py
```diff
@@ -697,6 +704,11 @@ def forward(
                 )
                 return output
             else:
+                sliding_window_size = (
```
tjtanaa, line 707: Is it possible for use to do it in the `__init__` function to avoid redundant calls every forward pass. Like after this line https://github.com/vllm-project/vllm/blob/3212278ebbf4231f1333bc945b351b088262cd22/vllm/v1/attention/backends/flash_attn.py#L546

vllm/v1/attention/backends/flash_attn_diffkv.py
```diff
@@ -113,6 +113,9 @@ def forward(
               We use torch's .expand() to avoid duplicating values
         """
         assert output is not None, "Output tensor must be provided."
+        assert self.vllm_flash_attn_version is not None, (
```
tjtanaa, line 116: These lines are not needed after you move the assertion into `FlashAttentionImpl`

vllm/v1/attention/backends/flash_attn_diffkv.py
```diff
@@ -214,6 +217,11 @@ def forward(
                 )
                 return output
             else:
+                sliding_window_size = (
```
tjtanaa, line 220: These lines are not needed after you move them into `FlashAttentionImpl`

vllm/v1/attention/backends/mla/rocm_aiter_mla.py
```diff
@@ -233,7 +233,7 @@ def _flash_attn_varlen_diff_headdims(
             k=k,
             v=v,
             softmax_scale=softmax_scale,
-            return_lse=return_softmax_lse,
+            return_softmax_lse=return_softmax_lse,
```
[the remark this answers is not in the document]
tjtanaa: @MrIceCreamMan  can you revert this back? Because the parameter name can be different from CUDA. Gemini suggestion could be wrong. Please retain what is already there for ROCm.

CC @yewentao256

vllm/v1/attention/backends/mla/aiter_triton_mla.py
```diff
@@ -62,7 +62,7 @@ def _flash_attn_varlen_diff_headdims(
             k,
             v,
             softmax_scale=softmax_scale,
-            return_lse=return_softmax_lse,
```
vllmellm, line 65: please keep `return_lse=return_softmax_lse` the gemini suggestion was wrong. rocm aiter backend the function is different and the arguments names are different.

vllm/v1/attention/backends/mla/rocm_aiter_mla.py
```diff
@@ -236,7 +236,7 @@ def _flash_attn_varlen_diff_headdims(
             k=k,
             v=v,
             softmax_scale=softmax_scale,
-            return_lse=return_softmax_lse,
```
vllmellm, line 239: please keep `return_lse=return_softmax_lse` the gemini suggestion was wrong. rocm aiter backend the function is different and the arguments names are different.

